PR #27, opened 2017-12-06
➖ Remove `babel-preset-es2015` as a dev dependency

yeskunall:
Looks like it's not required any more as we're using `babel-preset-env` in its place.


devongovett:
Thanks!

PR #38, opened 2017-12-06
change reBuildTimeout from 100ms to 1000ms

Seojiwoong:
change reBuildTimeout from 100ms to 1000ms

100ms is too short see what file has changed.


devongovett:
Was there a particular error you saw because of the lower timeout?

Seojiwoong:
@devongovett  There is no error. But  It is hard to see what file has changed.
100ms is too short

PR #35, opened 2017-12-06
Add Travis, Prettier, Contribution docs, and init Flow

thejameskyle:
cc @aweary

PR #45, opened 2017-12-06
Setup Travis, add build script, init contributing docs

thejameskyle:
Gonna get some of these changes in immediately because #35 got complicated with Babel changes


thejameskyle:
@devongovett Tests have been very flakey for me, and they take too long in CI. We're gonna have to figure out something to make them more reliable

StarpTech:
Could you also add appveyor?

StarpTech:
here a template https://github.com/hemerajs/hemera/blob/master/.appveyor.yml

devongovett:
@thejameskyle yeah... wonder if we need to mock the filesystem or something. I think there are race conditions in the tests, especially with the watcher too.

CONTRIBUTING.md
```diff
@@ -0,0 +1,36 @@
+# Contributing
+
+## Overview
+
+* `yarn install` - install all dependencies
+* `yarn build` - run src/ through [babel] into lib/
+* `yarn test` - run all tests in repo
+* `yarn format` - run [prettier] on all files
```
aweary: Since it's listed here, can we add `prettier` and the `format` script to this PR too?

aweary:
@devongovett maybe just write a script that can load all the integrations in memory, make it cacheable, and then just avoid using `fs` all together in the tests?

devongovett:
Gonna merge this so we can get the contributing guide in. Will work on the tests later.

CONTRIBUTING.md
devongovett: Will be done in #35.

thejameskyle:
This may help with the fixtures: https://github.com/Thinkmill/jest-fixtures

note: It's poorly named, there's nothing specific to Jest inside it other than it following Jests file convention

We should probably switch to Jest or Ava though so that we can parallelize the tests

devongovett:
Raised the test timeout and fixed some other test flakiness here #74

PR #48, opened 2017-12-06
Feature/#11 windows path issue

StarpTech:


thejameskyle:
We'll need to do this in many more places. The tests are all using posix paths

src/utils/config.js
```diff
@@ -3,7 +3,7 @@ const path = require('path');
 
 const existsCache = new Map;
 
-async function resolve(filepath, filenames, root = '/') {
+async function resolve(filepath, filenames, root = path.parse(filepath).root) {
```
pke: or use `root = path.delimiter`?

StarpTech: On windows you don't get this `\` or `\\` or `/` it depends on your volume

StarpTech: Look at https://nodejs.org/api/path.html#path_path_parse_path

pke: Good catch! Will you also fix the tests?

devongovett:
Thanks, I'll get this released shortly. Want to check other places that might need to change too, @StarpTech?

src/utils/config.js
StarpTech: Tests are hard to fix because I want to know whether it works on linux. I will wait for https://github.com/parcel-bundler/parcel/pull/35

Issue #43, opened 2017-12-06
tries to import datauri as a file if it is in quotes


chee:
```html
<!--index.html-->
<!doctype html>
<link rel="stylesheet" href="./lol.css">
```

works!:
```css
/* lol.css */
body {
  background: url(data:image/gif;base64,something);
}
```

error:
```css
/* lol.css */
body {
  background: url("data:image/gif;base64,something");
}
```

(`🚨  /Users/chee/projects/nothing/lol.css:2:3: Cannot resolve dependency './data:image/gif;base64,something'`)

thejameskyle:
[Relevant code is here](https://github.com/parcel-bundler/parcel/blob/ad5135adfda84e7bf49a9aee477deae7d83a3588/src/assets/CSSAsset.js#L61)

ssuman:
I'll work on this.

devongovett:
Fixed by #50.

nijikokun:
Awaiting release of this fix :+1:

nijikokun:
I tried using master since I couldn't wait and I am still encountering this issue:

```
/Users/nijikokun/ui.css:1:21599: Cannot resolve dependency './data:image/svg+xml;charset=utf-8,%3Csvg width='10' height='8' vie
>
```

The line in question is (notice the url inside, maybe that is triggering it):

```
background:url("data:image/svg+xml;charset=utf-8,%3Csvg xmlns='http://www.w3.org/2000/svg' width='24' height='24' viewBox='0 0 24 24'%3E%3Cpath d='M10 6L8.59 7.41 13.17 12l-4.58 4.59L10 18l6-6z'/%3E%3C/svg%3E")
```

Here is me checking the contents of `is-url.js` to confirm that the code I am using has the changes which it does:

```
global/node_modules/parcel-bundler
▶ cat src/utils/is-url.js
const isURL = require('is-url');

// Matches anchor (ie: #raptors)
const ANCHOR_REGEXP = /^#/;

// Matches scheme (ie: tel:, mailto:, data:)
const SCHEME_REGEXP = /^[a-z]*\:/i;

module.exports = function (url) {
  return isURL(url) || ANCHOR_REGEXP.test(url) || SCHEME_REGEXP.test(url);
};

global/node_modules/parcel-bundler
▶
```

nijikokun:
I have found the issue, there was a `.cache` directory that needs to be blown away first. It works as intended.

chee:
thank you, everyone.

[...]

PR #50, opened 2017-12-06
Do not try to import data: tel: mailto: something: URLs

oliger:
Parcel is trying to "import" data URLs. This PR ignore URLs when they start with `data:`.

### DIFF
--- a/src/utils/is-url.js
+++ b/src/utils/is-url.js
@@ -1,5 +1,5 @@
 const isURL = require('is-url');
 
 module.exports = function (url) {
-  return isURL(url) || /^#/.test(url) || /^data\:/.test(url);
+  return isURL(url) || /^#/.test(url) || /^[a-z]*\:/i.test(url);
 };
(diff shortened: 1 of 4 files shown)


oliger:
So it seems that this PR fixes #43... And that it was already taken by @ssuman... Feel free to close.

oliger:
Found out that Parcel needs to ignore all `something:` URLs. I made one more commit to accept URLs that start with `mailto:`, `tel:`.

src/utils/is-url.js
```diff
@@ -1,5 +1,5 @@
 const isURL = require('is-url');
 
 module.exports = function (url) {
-  return isURL(url) || /^#/.test(url);
+  return isURL(url) || /^#/.test(url) || /^[a-z]*\:/i.test(url);
```
thejameskyle: Could we hoist these regexes into constants and add descriptions?

davidnagli: +1

oliger:
Yep 👍. Try to do this tonight. 

Envoyé de mon iPhone

src/utils/is-url.js
oliger: @thejameskyle Done

devongovett:
Thanks!

PR #59, opened 2017-12-06
Fix issue with normalizeOptions and publicUrl

### LINKED ISSUE #21: How do I update public URL?
Heyo. I've seen the flag `--public-url` but not exactly sure how it behaves. I am trying to `--public-url=.` and I've also tried `--public-url=http://localhost:8080`.

It would be nice if I could just modify it. Especially for cases where I'd throw everything that's on `dist` in a S3 bucket with static hosting enabled. Does that make sense? am I misunderstanding something?

Thanks! Very nice stuff, tho!

❤️

micopc:
Fixes #21. normalizeOptions can now support publicUrl and publicURL.

### DIFF
--- a/src/Bundler.js
+++ b/src/Bundler.js
@@ -46,7 +46,7 @@ class Bundler extends EventEmitter {
 
   normalizeOptions(options) {
     const isProduction = options.production || process.env.NODE_ENV === 'production';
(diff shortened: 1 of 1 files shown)


devongovett:
Thanks!

micopc:
Thank you for this awesome tool!

Issue #30, opened 2017-12-06
do u support cmd?


Sobranier:
index.js:

```
import main from './main';
//
main();
```

main.js:

```
export default () => {
    console.log(classes.main);
};
```

result:

<img width="253" alt="2017-12-06 11 43 04" src="https://user-images.githubusercontent.com/7440073/33643905-b80dae96-da7c-11e7-89f8-1e6b4cd64ff6.png">

thejameskyle:
This is not enough information to go off of, please provide something that we can reproduce and I'll reopen this issue.

PR #63, opened 2017-12-06
Fix default only exports

RichieAHB:
This fixes an issue where modules which only had a default export would not be marked as ES6 and wouldn't then have it's module syntax converted to CommonJS. So something like the below would fail:

```javascript
const a = 1;
const b = 2;
export default () => a + b;
```
Default exports would be transpiled _if_ they had imports, as the visitor for the import statement would flag the module as ES6 (this is how the ‘Hello World’ example works on the home page).

I think this also fixes https://github.com/parcel-bundler/parcel/issues/30

There is also a yarn.lock format update that I brought in.

### DIFF
--- a/src/visitors/dependencies.js
+++ b/src/visitors/dependencies.js
@@ -23,6 +23,10 @@ module.exports = {
     addDependency(asset, node.source);
   },
 
+  ExportDefaultDeclaration(node, asset) {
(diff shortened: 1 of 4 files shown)


devongovett:
Awesome, thanks!

Issue #62, opened 2017-12-06
Make current parcel version part of cache keys


thejameskyle:
It would be good if newer versions of Parcel invalidated the cache

thejameskyle:
Moving to #67

PR #67, opened 2017-12-07
Make current Parcel version part of cache key.

nithinpeter:
This pull request addresses #62. Updated `FSCache` to also include current Parcel version also in the package key.


thejameskyle:
Fixed the typo by just renaming it to `pkg`, "pjson" sounded like some special JSON format.

devongovett:
Awesome, thanks!

thejameskyle:
We might want to add the current version of Node as well. This bites me all the time with Jest because I switch between versions of Node in between test runs.

PR #75, opened 2017-12-07
Merge postcss-modules config

isbasex:


src/transforms/postcss.js
```diff
@@ -24,12 +26,16 @@ async function getConfig(asset) {
   }
 
   config = config || {};
-  config.plugins = loadPlugins(config.plugins, asset.name);
+  const postcssModulesConfig = Object.assign(get(config.plugins, 'postcss-modules') || {}, {
```
thejameskyle: Why not just `config.plugins['postcss-modules']`?

isbasex: type safety

package.json
```diff
@@ -23,6 +23,8 @@
     "htmlnano": "^0.1.6",
     "is-url": "^1.2.2",
     "js-yaml": "^3.10.0",
+    "lodash.get": "^4.4.2",
+    "lodash.omit": "^4.5.0",
```
devongovett: I'd rather avoid adding more dependencies if we can. omit could just be `delete config.plugins['postcss-modules']`, and get could be just be `config.plugins['postcss-modules']`.

isbasex:
@devongovett Now, just pure js

PR #78, opened 2017-12-07
fix for bundle-loader when using ESM

kingdaro:
The previous behavior only returned default exports and discarded all others when the imported module is an ES module.
```js
export const a = 1
export const b = 2
```
```js
const nums = await import('./nums')
console.log(nums.a + nums.b) // error, nums is undefined
```

This PR removes the ES module check and returns all of the available exports in all cases, which should be the correct behavior. References: [1](http://2ality.com/2017/01/import-operator.html) [2](http://2ality.com/2017/01/import-operator.html)


devongovett:
ah this is a little unfortunate, but makes sense. would be nice to avoid `.default` everywhere, but if that's what the spec says that's what we should do.

will need a documentation update as well.

kingdaro:
I actually agree, but it's the way webpack works, and it's the way it works in chrome natively. Not sure if there's a way to be able to access both default and other named exports at the same time without exposing them explicitly. The use could use `module.exports = ...` if they really wanted, but seems kinda dirty if the rest of the code is ESM. Oh well 🙁

thejameskyle:
Created a docs PR https://github.com/parcel-bundler/website/pull/29

PR #84, opened 2017-12-07
Basic TypeScript Support

DeMoorJasper:
Potential implementation of basic typescript support #3


src/Parser.js
```diff
@@ -10,6 +10,7 @@ class Parser {
     this.registerExtension('js', './assets/JSAsset');
     this.registerExtension('jsx', './assets/JSAsset');
     this.registerExtension('es6', './assets/JSAsset');
+    this.registerExtension('ts', './assets/TypeScriptAsset');
```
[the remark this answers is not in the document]
DeMoorJasper: done

src/assets/TypeScriptAsset.js
```diff
@@ -0,0 +1,23 @@
...
+    this.ast = await this.parse(this.contents);
+    this.isAstDirty = true;
+  }
+
+  async parse(code) {
+    // require typescript, installed locally in the app
+    let typescript = localRequire('typescript', this.name);
+
+    let parserOptions = {
+      module: typescript.ModuleKind.CommonJS
```
felipe-gdr: you might need to add `jsx: true` to the options

src/assets/TypeScriptAsset.js
```diff
@@ -0,0 +1,27 @@
...
+    this.ast = await this.parse(this.contents);
+    this.isAstDirty = true;
+  }
+
+  async parse(code) {
+    // require typescript, installed locally in the app
+    let typescript = localRequire('typescript', this.name);
+
+    let transpilerOptions = {
+      compilerOptions: {
```
ds300: Might make sense to use this already to load a project's tsconfig.json file: https://www.npmjs.com/package/tsconfig

If you do, be sure to overwrite `noEmit` to be false, so that people can use `tsc` as a kind of type-linter by setting `noEmit` to true in their tsconfig.json.

DeMoorJasper: will def look into this

src/assets/TypeScriptAsset.js
```diff
@@ -0,0 +1,34 @@
+const JSAsset = require('./JSAsset');
+const config = require('../utils/config');
+const localRequire = require('../utils/localRequire');
+
+class TypeScriptAsset extends JSAsset {
+  async transform() {
+    this.ast = await this.parse(this.contents);
+    this.isAstDirty = true;
+  }
```
devongovett: This method shouldn't be necessary, and overriding it here disables other transforms like babel (post typescript) if needed, and uglify. `parse` should already be called by the super `Asset` class.

src/assets/TypeScriptAsset.js
```diff
@@ -0,0 +1,34 @@
...
+
+  async parse(code) {
+    // require typescript, installed locally in the app
+    let typescript = localRequire('typescript', this.name);
+
+    let transpilerOptions = {
+      compilerOptions: {
+        module: typescript.ModuleKind.CommonJS,
```
devongovett: Does this still support ES2015 modules as well?

src/assets/TypeScriptAsset.js
DeMoorJasper: At first i tried without overwriting transform and it didn't parse the code. I'll try look into it a bit more

src/assets/TypeScriptAsset.js
DeMoorJasper: It's configurable by adding a tsconfig.json file, but i can default it to ES2015 if that is preferred?

devongovett: Sorry, I guess I should ask what this option is for. Does it control parsing behavior or code generation? If codegen, we want commonjs for sure.

DeMoorJasper: It's codegen

src/assets/TypeScriptAsset.js
DeMoorJasper: In my latest addition to this PR i tried to implement it in a better way, hope you find this to be the correct way

src/assets/TypeScriptAsset.js
```diff
@@ -0,0 +1,35 @@
...
+        jsx: true
+      },
+      fileName: this.basename
```
felipe-gdr: Maybe I'm missing something, but why do we need `fileName` here?

DeMoorJasper: If i'm reading the TypeScript transpile API docs correctly it should autodetect wether something is ts or tsx based on filename, this is why i included it.
``Usually TypeScript compiler uses file extension to determine if file should be parsed as '.tsx' or '.ts'. The same rule is applied during single file transpilation if the file name is provided.``

felipe-gdr: According to the docs, `fileName` is an optional parameter of the `ts#transpile` and not a property of the tsconfig. If you take a look at the Json schema you won't find a fileName property there: http://json.schemastore.org/tsconfig

DeMoorJasper: The interface TranspileOptions which this is, does contain a FileName as far as i can see in their source code
``export interface TranspileOptions {
    compilerOptions?: CompilerOptions;
    fileName?: string;
    reportDiagnostics?: boolean;
    moduleName?: string;
    renamedDependencies?: Map<string>;
}

export function transpileModule(input: string, transpileOptions: TranspileOptions): TranspileOutput``

felipe-gdr: Source code doesn't lie

devongovett:
Awesome thanks for your work on this!! I cleaned up a couple things here and added a test for `tsconfig.json` support here https://github.com/parcel-bundler/parcel/commit/c27c5c81986e54703c9742c63e8714f65be8c555

imvetri:
Really good. I went through the PR I would also like to know understand how this was done. The code change was minimal but functionality wise I couldn't understand. Is there any doc that says explains how to extend support?

DeMoorJasper:
It extends the JSAsset class, which just takes over all functionality.
Overwrites parse function so that this gets used instead of the JSAsset one, to transpile it into JS.
At the end of parse it replaces this.contents with TS one which makes sure all the other functions in transform still work (because they are all expecting JS)
After that in calls the super.parse which just makes sure it is valid to the babelrc config and puts it into AST format.
That's mainly everything that changed

PR #104, opened 2017-12-07
Notify browser on error

### LINKED ISSUE #100: Report build errors to the browser?
This has happened to me more than a couple times:

 - make some changes
 - switch to the browser
 - wonder why my changes are showing up
 - switch back and forth, refresh
 - checked the console and realized the build was failing

Perhaps parcel could notify the browser on build failures via the HMR socket, or refuse to serve the build assets if the build failed?

spalger:
Closes #100

 - Emit `error` events on the HMR socket when the bundler fails
 - Emit `error` events on new HMR sockets when the bundler has not completed successfully since the last error
 - Emit `error-resolved` events on the HMR socket when the bundler passes after failing
 - Serve 500s when the build fails

### DIFF
--- a/src/HMRServer.js
+++ b/src/HMRServer.js
@@ -6,6 +6,12 @@ class HMRServer {
       this.wss = new WebSocket.Server({port: 0}, resolve);
     });
 
+    this.wss.on('connection', (ws) => {
+      if (this.unresolvedError) {
+        ws.send(JSON.stringify(this.unresolvedError))
+      }
+    });
+
     return this.wss._server.address().port;
   }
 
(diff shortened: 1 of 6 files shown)


src/HMRServer.js
```diff
@@ -6,15 +6,63 @@ class HMRServer {
...
+
     return this.wss._server.address().port;
   }
 
   stop() {
     this.wss.close();
   }
 
+  emitError(err) {
+    let message = typeof err === 'string' ? err : err.message;
```
spalger: I mostly copied this code from [the logger](https://github.com/parcel-bundler/parcel/blob/ebadce69cd7ec003ad72a0c99d87675f4df50f7b/src/Logger.js#L46-L62) so that error messages/handling would be consistent. Perhaps it should be moved into a utility of some sort? Maybe `normalizeError(error, { colors: boolean }): { message, stack }`?

src/HMRServer.js
```diff
@@ -6,15 +6,63 @@ class HMRServer {
...
+    }
+
+    this.broadcast(this.unresolvedError)
+  }
+
   emitUpdate(assets) {
-    let msg = JSON.stringify({
+    if (this.unresolvedError) {
+      this.unresolvedError = null
+      this.broadcast({
```
spalger: Initially tracked this in the `hmr-runtime`, but when I realized it needed to be tracked here too I thought it might make more sense to keep the event handling in `hmr-runtime` mostly stateless.

src/Server.js
```diff
@@ -35,6 +37,11 @@ function middleware(bundler) {
       }
     }
 
+    function send500() {
+      res.writeHead(500);
```
spalger: I'd love if this responded with a trimmed down bundle that included the hmr-runtime and reloaded when the build succeeded but I couldn't figure out how to make it work. Any ideas?

src/HMRServer.js
devongovett: Done in https://github.com/parcel-bundler/parcel/commit/67317ab5b4c618ed3492da236647a0a161d87562

devongovett:
This is sweet! Thanks for your work on it. 🎉

PR #114, opened 2017-12-08
Print stack traces on error

### LINKED ISSUE #113: Print stack traces on error
🐛 bug report

Parcel suppresses stack traces, leading to confusion:

https://github.com/parcel-bundler/parcel/blob/a1a190bb19a2c625ad3af9f4072fbbc8f342f746/src/Logger.js#L67

Users end up seeing "Cannot resolve dependency foo.bar" rather than the actual error.

One solution:

```js
      this.write(((this.color && err.highlightedCodeFrame) || err.codeFrame) + '\n\n' + err.stack);
```

shawwn:
Closes #113.

### DIFF
--- a/src/Bundler.js
+++ b/src/Bundler.js
@@ -227,20 +227,18 @@ class Bundler extends EventEmitter {
     return bundle;
   }
 
-  async resolveAsset(name, parent) {
+  async resolveAsset(name, parent, dep) {
     try {
       var {path, pkg} = await this.resolver.resolve(name, parent);
     } catch (err) {
-      err.message = `Cannot resolve dependency '${dep.name}'`;
+      err.message = `Cannot resolve dependency '${name}'`;
 
       // Generate a code frame where the dependency was used
-      if (dep.loc) {
-        await asset.loadIfNeeded();
+      if (dep && dep.loc) {
         err.loc = dep.loc;
-        err = asset.generateErrorMessage(err);
       }
 
-      err.fileName = asset.name;
+      err.fileName = parent;
       throw err;
     }
     if (this.loadedAssets.has(path)) {
@@ -258,7 +256,7 @@ class Bundler extends EventEmitter {
   }
 
   async resolveDep(asset, dep) {
-    return await this.resolveAsset(dep.name, asset.name);
+    return await this.resolveAsset(dep.name, asset.name, dep);
   }
 
   async loadAsset(asset) {
--- a/src/Bundler.js
+++ b/src/Bundler.js
@@ -227,20 +227,8 @@ class Bundler extends EventEmitter {
     return bundle;
   }
 
-  async resolveAsset(name, parent, dep) {
-    try {
-      var {path, pkg} = await this.resolver.resolve(name, parent);
-    } catch (err) {
-      err.message = `Cannot resolve dependency '${name}'`;
-
-      // Generate a code frame where the dependency was used
-      if (dep && dep.loc) {
-        err.loc = dep.loc;
-      }
-
-      err.fileName = parent;
-      throw err;
-    }
+  async resolveAsset(name, parent) {
+    let {path, pkg} = await this.resolver.resolve(name, parent);
     if (this.loadedAssets.has(path)) {
       return this.loadedAssets.get(path);
     }
@@ -256,7 +244,23 @@ class Bundler extends EventEmitter {
   }
 
   async resolveDep(asset, dep) {
-    return await this.resolveAsset(dep.name, asset.name, dep);
+    try {
+      return await this.resolveAsset(dep.name, asset.name);
+    } catch (err) {
+      if (err.message.indexOf('Cannot find module ') === 0) {
+        err.message = `Cannot resolve dependency '${dep.name}'`;
+
+        // Generate a code frame where the dependency was used
+        if (dep.loc) {
+          await asset.loadIfNeeded();
+          err.loc = dep.loc;
+          err = asset.generateErrorMessage(err);
+        }
+
+        err.fileName = asset.name;
+      }
+      throw err;
+    }
   }
 
   async loadAsset(asset) {
--- a/src/Bundler.js
+++ b/src/Bundler.js
@@ -247,7 +247,7 @@ class Bundler extends EventEmitter {
     try {
       return await this.resolveAsset(dep.name, asset.name);
     } catch (err) {
-      err.message = `Cannot resolve dependency '${dep.name}'`;
+      err.message = `Cannot resolve dependency '${dep.name}': ${err.message}`;
 
       // Generate a code frame where the dependency was used
       if (dep.loc) {
--- a/src/Bundler.js
+++ b/src/Bundler.js
@@ -228,7 +228,21 @@ class Bundler extends EventEmitter {
   }
 
   async resolveAsset(name, parent) {
-    let {path, pkg} = await this.resolver.resolve(name, parent);
+    try {
+      var {path, pkg} = await this.resolver.resolve(name, parent);
+    } catch (err) {
+      err.message = `Cannot resolve dependency '${dep.name}'`;
+
+      // Generate a code frame where the dependency was used
+      if (dep.loc) {
+        await asset.loadIfNeeded();
+        err.loc = dep.loc;
+        err = asset.generateErrorMessage(err);
+      }
+
+      err.fileName = asset.name;
+      throw err;
+    }
     if (this.loadedAssets.has(path)) {
       return this.loadedAssets.get(path);
     }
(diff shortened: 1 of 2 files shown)


devongovett:
So the stack trace is suppressed on purpose - most of the time, users don't care where in the compiler (e.g. babel) something failed, especially if we can give them a code frame pointing at the location (the case you changed). If there is no code frame (e.g. in the case of an internal error rather than a user code error), then we print the stack. Make sense?

thejameskyle:
Maybe we could add a debug flag of some sort to do both?

shawwn:
I was adding CoffeeScript support, but forgot to `yarn add coffeescript`. Therefore `require('coffeescript')` in CoffeeAsset.js was failing.

The only thing that showed up was "Cannot resolve dependency './local.coffee'", and it pointed to `parcel/test/integration/coffee/index.js:1:20`.

The actual error was:

```
ReferenceError: YAMLAsset is not defined
    at Object.<anonymous> (parcel/shawwn/parcel/src/assets/CoffeeAsset.js:64:18)
```

I really think it would be a better idea to just print the actual error. And in fact, this PR is incomplete: we stomp on the real error message by overriding `err.message`, throwing away the original error. I'll add a commit.

shawwn:
Ok, here's what users will see now:

```
$ bin/cli.js test.html 
test/integration/coffee/index.js:1:20: Cannot resolve dependency './local.coffee': Cannot find module 'coffeescript'
> 1 | var local = require('./local.coffee');
    |                     ^
  2 | 
  3 | module.exports = function () {
  4 |   return local.a + local.b.c;

Error: Cannot resolve dependency './local.coffee': Cannot find module 'coffeescript'
    at Function.Module._resolveFilename (module.js:536:15)
    at Function.Module._load (module.js:466:25)
    at Module.require (module.js:579:17)
    at require (node_modules/v8-compile-cache/v8-compile-cache.js:159:20)
    at Object.<anonymous> (src/assets/CoffeeAsset.js:45:14)
    at Module._compile (node_modules/v8-compile-cache/v8-compile-cache.js:178:30)
    at Object.Module._extensions..js (module.js:646:10)
    at Module.load (module.js:554:32)
    at tryModuleLoad (module.js:497:12)
    at Function.Module._load (module.js:489:3)

```

As a contributor, it was essential for the error to pinpoint line 2 of CoffeeAsset.js as the problem:

```js
var coffee = require('coffeescript'); // this require was failing

class CoffeeAsset extends Asset {
  constructor(name, pkg, options) {
...
```

shawwn:
Without this PR, parcel only shows the following error:

```
$ bin/cli.js test.html 
test/integration/coffee/index.js:1:20: Cannot resolve dependency './local.coffee'
> 1 | var local = require('./local.coffee');
    |                     ^
  2 | 
  3 | module.exports = function () {
  4 |   return local.a + local.b.c;
```

That wasn't enough to figure out that I failed to `yarn add coffeescript` and hence `require('coffeescript')` was failing.

davidnagli:
I agree with @thejameskyle We should keep this behind a debug flag. 

I think the idea to suppress the useless error information was actually pretty neat and I think removing it defeats the purpose for the majority of users who don’t plan to add their own custom Assets to Parcel.

shawwn:
I think I'm not being clear. Let me try again.

I was adding CoffeeScript support. This is equivalent to adding a new plugin, so essentially I was a plugin author.

I added a CoffeeAsset.js file, which contained an error. That error was suppressed; it was impossible to figure out what was going on. All I could see was "failed to resolve import './local.coffee'", and the code frame did not point to the right place. Worse, there was no stack trace, so I had no idea where to look. 

I had to open up Parcel in Webstorm debugger just to figure out what was going wrong.

So, seeing as this is an error, not a warning, and seeing as the compilation is failing, it's important for us to give all available information to the user. They shouldn't need to run in debug mode just to see that no one ran `npm i coffeescript` yet.

shawwn:
It's very hard to add new asset types without this PR, because during development you make a lot of mistakes, and Parcel suppresses the actual error messages. It doesn't print the real error message, and it doesn't point to the actual spot causing the error.

The only thing Parcel prints out is this:

```
$ bin/cli.js test.html
🚨  parcel/test/integration/coffee/index.js:1:20: Cannot resolve dependency './local.coffee'
> 1 | var local = require('./local.coffee');
    |                     ^
  2 | 
  3 | module.exports = function () {
  4 |   return local.a + local.b.c;
```

... Even though the error is in `local.coffee` on line 2, not `index.js` on line 1.

devongovett:
can you make your changes behind a debug flag? then we can add `debug: true` to the tests here https://github.com/parcel-bundler/parcel/blob/master/test/utils.js#L15-L20 as well as the CLI. when contributing to parcel, you could use this flag to debug, but normal users don't need to see the long stack traces.

davidnagli:
I see your point. But as an Asset Devloper why can’t you just add an extra debug flag when you run Parcel. This way we can give the users a pretty output, and the Asset authors a detailed one. 

I see your point entirely, I just think it might be neat to differentiate the user experience between the two by hiding the detailed error output behind a flag.

davidnagli:
To help illustrate the other perspective: Imagine that you are a regular user of Parcel (not an Asset Author). Let’s say you install parcel and run it on your files. But **you** scew up by mistyping the module name in your import statement. 

According to your PR, that new user should see a full stack trace showing him exactly which line of code WITHIN Parcel called the function that broke his build.

What Daven did (how it is currently), is when **the user** screws up and gets an error, they just see where **they** screwed up, not the full stack trace showing them exactly what parts of Parcel were getting called when **their code** caused an error.

That’s how I see it :)

shawwn:
@davidnagli Good point.

Here's the current state of the PR:

```sh

# When the user tries to import an invalid filename:
$ bin/cli.js test.html 
parcel/test/integration/coffee/index.js:1:20: Cannot resolve dependency './bogus.coffee'
    at parcel/node_modules/browser-resolve/node_modules/resolve/lib/async.js:55:21
    at load (parcel/node_modules/browser-resolve/node_modules/resolve/lib/async.js:69:43)
    at onex (parcel/node_modules/browser-resolve/node_modules/resolve/lib/async.js:92:31)
    at parcel/node_modules/browser-resolve/node_modules/resolve/lib/async.js:22:47
    at FSReqWrap.oncomplete (fs.js:152:21)

# When a plugin has an internal error:
$ bin/cli.js test.html 
YAMLAsset is not defined
    at Object.<anonymous> (parcel/src/assets/CoffeeAsset.js:68:18)
    at Module._compile (parcel/node_modules/v8-compile-cache/v8-compile-cache.js:178:30)
    at Object.Module._extensions..js (module.js:646:10)
    at Module.load (module.js:554:32)
    at tryModuleLoad (module.js:497:12)
    at Function.Module._load (module.js:489:3)
    at Module.require (module.js:579:17)
    at require (parcel/node_modules/v8-compile-cache/v8-compile-cache.js:159:20)
    at Parser.findParser (parcel/src/Parser.js:48:45)
    at Parser.getAsset (parcel/src/Parser.js:56:22)
    at Bundler.resolveAsset (parcel/src/Bundler.js:248:29)
    at <anonymous>:null:null
```

This seems exactly what the user would expect.

Node itself prints out traces on error, so both of these cases seem to fall in line with standard behavior.

shawwn:
Is there a better way to detect a user import error than this?

```js
if (err.message.indexOf(`Cannot find module '${dep.name}'`) === 0) {
```

Otherwise, this PR is ready to merge. The behavior is as you wanted it: If the user tries to import a bogus file, it generates a code frame and suppresses the stack trace.

In all other cases, the error is printed with a stack trace.

shawwn:
```sh

# plugin author failed to install a dependency
~/parcel$ bin/cli.js test.html
🚨  Cannot find module 'coffeescript'
    at Function.Module._resolveFilename (module.js:536:15)
    at Function.Module._load (module.js:466:25)
    at Module.require (module.js:579:17)
    at require (parcel/node_modules/v8-compile-cache/v8-compile-cache.js:159:20)
    at Object.<anonymous> (parcel/src/assets/CoffeeAsset.js:45:14)
    at Module._compile (parcel/node_modules/v8-compile-cache/v8-compile-cache.js:178:30)
    at Object.Module._extensions..js (module.js:646:10)
    at Module.load (module.js:554:32)
    at tryModuleLoad (module.js:497:12)
    at Function.Module._load (module.js:489:3)

# user tried to import a bogus file
~/parcel$ bin/cli.js test.html
🚨  parcel/test/integration/coffee/index.js:1:20: Cannot resolve dependency './bogus.coffee'
> 1 | var local = require('./bogus.coffee');
    |                     ^
  2 | 
  3 | module.exports = function () {
  4 |   return local.a + local.b.c;

# plugin had an internal error
~/parcel$ bin/cli.js test.html
🚨  YAMLAsset is not defined
    at Object.<anonymous> (parcel/src/assets/CoffeeAsset.js:68:18)
    at Module._compile (parcel/node_modules/v8-compile-cache/v8-compile-cache.js:178:30)
    at Object.Module._extensions..js (module.js:646:10)
    at Module.load (module.js:554:32)
    at tryModuleLoad (module.js:497:12)
    at Function.Module._load (module.js:489:3)
    at Module.require (module.js:579:17)
    at require (parcel/node_modules/v8-compile-cache/v8-compile-cache.js:159:20)
    at Parser.findParser (parcel/src/Parser.js:48:45)
    at Parser.getAsset (parcel/src/Parser.js:56:22)
    at Bundler.resolveAsset (parcel/src/Bundler.js:236:29)
    at <anonymous>:null:null
```

moos:
Previous reference by @moos by mistake!  Meant to ref 144.

PR #118, opened 2017-12-08
Initial CoffeeScript support

shawwn:
Poor test coverage, but other than that I modeled this after #84. Seems to work great.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ dist
 lib
 !test/**/node_modules
 .vscode/
+.idea/
--- a/src/Parser.js
+++ b/src/Parser.js
@@ -12,6 +12,7 @@ class Parser {
     this.registerExtension('es6', './assets/JSAsset');
     this.registerExtension('ts', './assets/TypeScriptAsset');
     this.registerExtension('tsx', './assets/TypeScriptAsset');
+    this.registerExtension('coffee', './assets/CoffeeScriptAsset');
     this.registerExtension('json', './assets/JSONAsset');
     this.registerExtension('yaml', './assets/YAMLAsset');
     this.registerExtension('yml', './assets/YAMLAsset');
@@ -23,7 +24,6 @@ class Parser {
     this.registerExtension('scss', './assets/SASSAsset');
 
     this.registerExtension('html', './assets/HTMLAsset');
-    this.registerExtension('coffee', './assets/CoffeeScriptAsset');
 
     let extensions = options.extensions || {};
     for (let ext in extensions) {
--- a/src/Parser.js
+++ b/src/Parser.js
@@ -23,6 +23,7 @@ class Parser {
     this.registerExtension('scss', './assets/SASSAsset');
 
     this.registerExtension('html', './assets/HTMLAsset');
+    this.registerExtension('coffee', './assets/CoffeeScriptAsset');
 
     let extensions = options.extensions || {};
     for (let ext in extensions) {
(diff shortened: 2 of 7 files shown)


src/Parser.js
```diff
@@ -23,6 +23,7 @@ class Parser {
     this.registerExtension('scss', './assets/SASSAsset');
 
     this.registerExtension('html', './assets/HTMLAsset');
+    this.registerExtension('coffee', './assets/CoffeeScriptAsset');
```
thejameskyle: Probably better to group this up with JS/TS

src/assets/CoffeeScriptAsset.js
```diff
@@ -0,0 +1,22 @@
...
+const config = require('../utils/config');
+const localRequire = require('../utils/localRequire');
+
+class CoffeeScriptAsset extends JSAsset {
+  async transform() {
+    super.transform();
+
+    await this.parseIfNeeded();
+    this.isAstDirty = true;
+  }
```
DeMoorJasper: This shouldn't be necessary, after the TypeScript PR this has been fixed by @devongovett in a better way to parse every non js file automatically. This would also break Uglify support.

shawwn:
Thanks! I'll take care of this tomorrow.

src/assets/CoffeeScriptAsset.js
```diff
@@ -0,0 +1,22 @@
+const JSAsset = require('./JSAsset');
+const config = require('../utils/config');
+const localRequire = require('../utils/localRequire');
+
+class CoffeeScriptAsset extends JSAsset {
+  async transform() {
```
DeMoorJasper: This shouldn't be necessary, after the TypeScript PR this has been fixed by @devongovett in a better way to parse every non js file automatically. This would also break Uglify support.

.gitignore
```diff
@@ -9,4 +9,5 @@ coverage
 dist
 lib
 !test/**/node_modules
-.vscode/
\ No newline at end of file
+.vscode/
+.idea/
```
antony: IDE specific ignores like .vscode and .idea probably belong in your own, machine local ignore file, or you'll end up with a huge list in a public project.

src/assets/CoffeeScriptAsset.js
shawwn: I tried removing the async transform block, but then it stopped working. I'm not sure why it works for TS but not for this.

src/Parser.js
shawwn: Done.

shawwn:
Welp. I screwed up the merge. One moment.

src/assets/CoffeeScriptAsset.js
DeMoorJasper: It should look like this:
```Javascript
class CoffeeScriptAsset extends JSAsset {
  async parse(code) {
    // require coffeescript, installed locally in the app
    let coffee = localRequire('coffeescript', this.name);

    // Transpile Module using CoffeeScript and parse result as ast format through babylon
    this.contents = coffee.compile(code, {});
    return await super.parse(this.contents);
  }
}
```

test/javascript.js
```diff
@@ -193,4 +193,18 @@ describe('javascript', function() {
     assert.equal(typeof output, 'function');
     assert.equal(output(), 3);
   });
+
+  it('should support requiring CoffeeScript files', async function () {
```
DeMoorJasper: Would maybe be a good idea to add some more tests for coffeescript in a seperate coffeescript test script

devongovett: that would be excellent! @shawwn can you follow up with that?

PR #122, opened 2017-12-08
Add Code of Conduct

thejameskyle:


shawwn:
How about the Ruby CoC? https://www.ruby-lang.org/en/conduct/

---

This document provides community guidelines for a safe, respectful, productive, and collaborative place for any person who is willing to contribute to the Ruby community. It applies to all “collaborative space”, which is defined as community communications channels (such as mailing lists, submitted patches, commit comments, etc.).

- Participants will be tolerant of opposing views.
- Participants must ensure that their language and actions are free of personal attacks and disparaging personal remarks.
- When interpreting the words and actions of others, participants should always assume good intentions.
- Behaviour which can be reasonably considered harassment will not be tolerated.

---

The Contributor Covenant is [quite controversial](https://www.allegro.cc/forums/thread/616388), and there are various forces on both sides that will try to inject politics into an otherwise neutral medium. It may be better to avoid that.

See the heated Ruby discussion: https://bugs.ruby-lang.org/issues/12004

If anyone in the parcel-bundler org expresses disagreeable views on Twitter, we may have people showing up demanding for them to be removed, like the Opal project experienced.

And we may be legally bound to do so, since merging this PR may constitute entering into a contract: https://github.com/ContributorCovenant/contributor_covenant/issues/284

That said, if this CoC is most in line with the project's goals, we should merge it.

pke:
I never understood why basic behavioural rules like those should be written down in the first place. And I don't remember seeing them in "the old days" of GitHub.

PR #128, opened 2017-12-08
CSS Hotreload if put in html

### LINKED ISSUE #57: Css not auto-updating if required from html
🙋 feature request
Include css-loader (auto-refresh css) in the bundle if the css is declared in the html

🔦 Context
So for example if someone would use one big global css file in index.html like this, it would not auto-update
```html
<link rel="stylesheet" href="./global.css" />
```

DeMoorJasper:
Potential fix for closes #57


src/builtins/hmr-runtime.js
```diff
@@ -22,6 +22,10 @@ if (!module.bundle.parent) {
     if (data.type === 'update') {
       for (let asset of data.assets) {
         hmrApply(global.require, asset);
+        // CSS should always update
+        if (asset.generated && asset.generated.css) {
+          asset.isNew = false;
+        }
```
devongovett: Not sure I understand why this was added. Can you explain?

src/packagers/HTMLPackager.js
```diff
@@ -44,6 +60,27 @@ class HTMLPackager extends Packager {
...
+    let head = find(tree, 'head');
+    if (!head) {
+      let html = find(tree, 'html');
+      head = {tag: 'head'};
+      html.content.unshift(head);
+    }
+
+    if (!head.content) {
+      head.content = [];
+    }
```
devongovett: maybe the above code can be moved to another method/function and shared with the css version?

src/builtins/hmr-runtime.js
DeMoorJasper: First time css gets edited it doesn't have it inside modules so it only updates from second edit up.
This should fix that, there is potentially a better fix for it though but thought this fix wasn't that bad because every generated js for css updates is the same anyway.

src/packagers/HTMLPackager.js
DeMoorJasper: Probably a good idea, i'll look into that

src/builtins/hmr-runtime.js
devongovett: Ah interesting. That's an interesting issue. Seems that the HMR code is not actually getting bundled into the JS output. Will look into it a bit more.

devongovett:
@DeMoorJasper refactored slightly. Now ensuring that the HMR JS produced by the CSSAsset is actually included in the JS bundle rather than adding a CSS specific hack to the runtime. Also merged the sibling bundle insertion code into a single pass, and added a test. Let me know if it works for you!

DeMoorJasper:
@devongovett works perfectly 👍

Issue #131, opened 2017-12-08
The code that parcel inserts is not processed with babel


chee:
🐛/🙋, could go either way.

I think this is related to #8.

if i have this `.babelrc`: 

```json
{"presets": ["es2015"]}
```

and i have this code:

```js
// index.js
import('./lol').then(console.log)
```

```js
// lol.js
export default 'lol'
```

and i do `parcel build index.js --no-minify`, i get this handy guy in the code:

```js
function requireModule(id) {
  let res = require(id);
  if (res.__esModule) {
     return res.default;
  }
  return res;
}
```

the `let` here is problematic in the environment i'm building for, and testing on.

Similarly, when using  `parcel index.html`, the websocket code generating for hot reloading has an arrow function in it.

Possible solution: transpile these generated snippets with the current babel config

reel:
If you use `env` instead of `es2015` and add a lower browser target, does it work?

See https://babeljs.io/docs/plugins/preset-env/

chee:
@reel 

No, 'cause the builtins are written in as they appear in their files. For instance, `hmr` in here: https://github.com/parcel-bundler/parcel/blob/3e97493690eb4b9a87b899b30bc0e1b19f0f84f1/src/packagers/JSPackager.js

they aren't transformed along with everything else, so [this arrow function](https://github.com/parcel-bundler/parcel/blob/3e97493690eb4b9a87b899b30bc0e1b19f0f84f1/src/builtins/hmr-runtime.js#L19) for instance, stays an arrow

PR #136, opened 2017-12-08
Use `var` over `let` in bundle loader built-in

chee:
related to #131 

The rest of the code in this file uses `var`, and this was probably a typo.


chee:
i'd like to add a commit to this to remove the couple arrow functions from the HMR Runtime built-in too, but i'd prefer to fork off @thejameskyle's prettier branch to do that

chee:
oh, the builtins are .prettierignore`d, so i don't need to do that

chee:
that's not the right approach. leaving this in pr though.

PR #147, opened 2017-12-08
added version option to command description

philipodev:
Hi, awesome project you got there.

Noticed, when looking for it, there's no description in `--help` how to output the current version so i added it.

### DIFF
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -14,6 +14,7 @@ program.command('serve [input]')
   .option('--public-url <url>', 'set the public URL to serve on. defaults to the same as the --out-dir option')
   .option('--no-hmr', 'disable hot module replacement')
   .option('--no-cache', 'disable the filesystem cache')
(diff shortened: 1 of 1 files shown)


philipodev:
@yeskunall hmm, you're right. no point in having the description in anything other than --help. i have commited changes for this. thanks

devongovett:
`parcel --version` already works.

philipodev:
@devongovett yes it does, but this PR simply adds it to the --help. I was personally looking for it when -v (lowercase) didn't work.

yeskunall:
@philipodev with your current changes, won't this only work for the `serve` command? I'm just curious 😅

philipodev:
@yeskunall yes, but parcel is using that command as `--help`. But if you type `parcel watch --help` you get the help for watch command and then I don't think printing the version option is relevant. :)

PR #153, opened 2017-12-09
Handle call expression during global insertion. Fixes #117

### LINKED ISSUE #117: process.env transformation is breaking leaving only ".NODE_ENV"
I think I found a bug where `(process.env.NODE_ENV)` in the source file was transformed to `(.NODE_ENV)` in my bundle.  Parcel doesn't complain, but when loading the page the js error is thrown.

I traced where this is happening to around:
https://github.com/parcel-bundler/parcel/blob/3e97493690eb4b9a87b899b30bc0e1b19f0f84f1/src/visitors/globals.js#L21

If I insert an early return in that function, then it builds and works.

You can repro the bug here:
https://github.com/joenoon/example-react-transition-group-stack-navigator

Then temporarily change node_modules/parcel-bundler/src/visitors/globals.js inserting the early return, yarn start, and the page will load correctly.

Raathigesh:
The global insertion was failing if the global parameter was passed as an argument to a function.

```javascript
function doSomethingAwesome(env) {}

doSomethingAwesome(process.env.NODE_ENV); // This would fail and this PR fixes it
```

### DIFF
--- a/src/visitors/globals.js
+++ b/src/visitors/globals.js
@@ -102,10 +102,19 @@ function matchesPattern(member, match) {
 
 // Replaces the key in `parent` whose value is `from` with `to`
 function replace(parent, from, to) {
-  for (let key in parent) {
-    if (parent[key] === from) {
-      parent[key] = to;
-      break;
+  if (types.isCallExpression(parent)) {
+    for (let key in parent.arguments) {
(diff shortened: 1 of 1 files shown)


devongovett:
fixed by #193.

PR #157, opened 2017-12-09
Change to uglify es & output uglify errors

### LINKED ISSUE #8: Unable to `build` index.html
Firstly, super excited for this. I was surprised at how quickly I was able to get a dev environment setup. You are doing some awesome work. 

I have `parcel-bundler`, `react` and `react-dom` installed. 

I have a simple `index.html` file and an `index.js` file which is just outputting "Hello World" to a new div on the index.html file. 

```
<body>
    <div id="root"></div>
     <script src="./index.js"></div>
</body>
```

I have no problems running `parcel index.html` or `parcel watch index.html`

however when I run `parcel build index.html` I get the following error:

```
./node_modules/.bin/parcel build index.html
🚨  /Users/azhar/projects/react/index.js: Cannot read property 'start' of undefined
    at module.exports (/Users/azhar/projects/react/node_modules/babel-to-estree/babylon-to-espree/attachComments.js:56:42)
    at module.exports (/Users/azhar/projects/react/node_modules/babel-to-estree/babylon-to-espree/index.js:35:3)
    at exports.toEstree (/Users/azhar/projects/react/node_modules/babel-to-estree/index.js:9:3)
    at module.exports (/Users/azhar/projects/react/node_modules/parcel-bundler/src/transforms/uglify.js:10:39)
    at <anonymous>
```

---

# Edit: [Temporary workaround](https://github.com/parcel-bundler/parcel/issues/8#issuecomment-349422021)

### LINKED ISSUE #15: Use babel-minify instead of the old Uglify JS
https://github.com/babel/minify

Update: UglifyJS caused build to fail in this issue: #8

### LINKED ISSUE #124: The packaged file is too LARGE
🐛 the 

here is my codes:
index.js
```jsx
console.log('hello, from parcel.')
require('./css/index.css')

import React, {Component} from 'react'
import ReactDOM from 'react-dom'

// import Test from './js/components/test'
const Test = () => {
    return <div>Hey, here is Test Cwefweomponent!!!!!!</div>
}

ReactDOM.render((
    <div>
        <h1>Hello React</h1>
        <h2>Hello Parcel</h2>
        <Test />
    </div>
), document.getElementById('root'))
```

### 😯 Current Behavior

after running: ` NODE_ENV=production && parcel build index.html --no-minify`
if without `--no-minify`, it doesn't work #8 
```bash
total 1544
-rw-r--r--  1 lynnleung  staff   120B 12  8 17:37 068be14c44c5675d76e18130404f6790.css
-rw-r--r--  1 lynnleung  staff   762K 12  8 17:37 068be14c44c5675d76e18130404f6790.js
-rw-r--r--  1 lynnleung  staff   488B 12  8 17:37 index.html
```

### 🌍 Your Environment

| Parcel           |  1.03
| Node             | 8.9.1
| Yarn         | 1.3.2
| Operating System | macOS 10.13.1

### LINKED ISSUE #145: You should be using uglify-es.
Using uglify-js is likely to leave your users of some modern browsers with zero-configuration errors. 

F.e. you can find many bugs like this one:

https://github.com/meteor/meteor/issues/9224

Just use uglify-es

DeMoorJasper:
If uglify encountered errors it didn't get thrown, due to a change in uglify v3.X.
~~I'm aware the current uglify bug people reported in #8 is due to this bug: https://github.com/lightscript/babel-to-estree/pull/1
EDIT: This will now also fix #8 due to no longer using estree [FIXED]~~
I think parcel should be using Uglify ES to support more modern babel configs & js versions to prevent future issues

**Performance tests** (Based on running the tests 3 times and taking the avg - On Macbook Pro dual core mid-2014)

|  | **UglifyJS**  | **UglifyES** | **UglifyES without estree** (ES6+ compatible) |
| ---------- | ------- | -------- | --- |
| JavaScript | 226ms | 263ms | 252ms |
|TypeScript | 100ms | 126ms | 133ms |

**Related to**
closes #145 closes #124 closes #15 closes #8

### DIFF
--- a/src/transforms/uglify.js
+++ b/src/transforms/uglify.js
@@ -3,7 +3,7 @@ const {toEstree} = require('babel-to-estree');
 const types = require('babel-types');
 const walk = require('babylon-walk');
 
-module.exports = async function (asset) {
+module.exports = async function(asset) {
   await asset.parseIfNeeded();
 
   // Convert to UglifyJS AST
--- a/src/transforms/uglify.js
+++ b/src/transforms/uglify.js
@@ -1,13 +1,10 @@
 const {minify} = require('uglify-es');
-const generate = require('babel-generator').default;
 
 module.exports = async function(asset) {
   await asset.parseIfNeeded();
 
   // Convert AST into JS
-  let code = asset.isAstDirty
-    ? generate(asset.ast).code
-    : asset.outputCode || asset.contents;
+  let code = asset.generate().js;
 
   let options = {
     warnings: true,
--- a/src/transforms/uglify.js
+++ b/src/transforms/uglify.js
@@ -1,11 +1,15 @@
 const {minify} = require('uglify-es');
-const types = require('babel-types');
-const walk = require('babylon-walk');
+const generate = require('babel-generator').default;
 
 module.exports = async function(asset) {
   await asset.parseIfNeeded();
 
-  let result = minify(asset.generate().js, {
+  // Convert AST into JS
+  let code = asset.isAstDirty
+    ? generate(asset.ast).code
+    : asset.outputCode || asset.contents;
+
+  let result = minify(code, {
     toplevel: true
   });
 
--- a/src/transforms/uglify.js
+++ b/src/transforms/uglify.js
@@ -5,8 +5,7 @@ const walk = require('babylon-walk');
 module.exports = async function(asset) {
   await asset.parseIfNeeded();
 
-  // Minify with uglify-es
-  var result = minify(asset.generate().js, {
+  let result = minify(asset.generate().js, {
     toplevel: true
   });
 
--- a/src/transforms/uglify.js
+++ b/src/transforms/uglify.js
@@ -5,7 +5,7 @@ const walk = require('babylon-walk');
 module.exports = async function(asset) {
   await asset.parseIfNeeded();
 
-  // Convert to UglifyES AST
+  // Minify with UglifyES
   var result = minify(asset.generate().js, {
     toplevel: true
   });
--- a/src/transforms/uglify.js
+++ b/src/transforms/uglify.js
@@ -1,4 +1,4 @@
-const {AST_Node, minify} = require('uglify-js');
+const {AST_Node, minify} = require('uglify-es');
 const {toEstree} = require('babel-to-estree');
 const types = require('babel-types');
 const walk = require('babylon-walk');
@@ -6,7 +6,7 @@ const walk = require('babylon-walk');
 module.exports = async function(asset) {
   await asset.parseIfNeeded();
 
-  // Convert to UglifyJS AST
+  // Convert to UglifyES AST
   var ast = AST_Node.from_mozilla_ast(toEstree(asset.ast, asset.contents));
   var result = minify(ast, {
     toplevel: true
--- a/src/transforms/uglify.js
+++ b/src/transforms/uglify.js
@@ -32,7 +32,7 @@ module.exports = async function(asset) {
     });
   }
 
-  // Uglify did our code generation for us, so remove the old AST
+  // babel-generator did our code generation for us, so remove the old AST
   asset.ast = null;
   asset.outputCode = result.code;
   asset.isAstDirty = false;
--- a/src/transforms/uglify.js
+++ b/src/transforms/uglify.js
@@ -12,6 +12,8 @@ module.exports = async function(asset) {
     toplevel: true
   });
 
+  if (result.error) throw result.error;
+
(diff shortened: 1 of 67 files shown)


src/transforms/uglify.js
```diff
@@ -1,17 +1,19 @@
...
 const {toEstree} = require('babel-to-estree');
 const types = require('babel-types');
 const walk = require('babylon-walk');
 
 module.exports = async function(asset) {
   await asset.parseIfNeeded();
 
-  // Convert to UglifyJS AST
+  // Convert to UglifyES AST
   var ast = AST_Node.from_mozilla_ast(toEstree(asset.ast, asset.contents));
```
kzc: This won't work for ES6+ code.

You either have to use the built-in `uglify-es` parser instead of ESTree or add ES6+ support to `uglify-es`.

It's an known issue that only ES5 ESTree constructs are currently converted by `uglify-es`: https://github.com/mishoo/UglifyJS2/issues/968

DeMoorJasper: Ow damn that kinda sucks, i'll read through that issue for sure

src/transforms/uglify.js
```diff
@@ -5,7 +5,7 @@ const walk = require('babylon-walk');
 module.exports = async function(asset) {
   await asset.parseIfNeeded();
 
-  // Convert to UglifyES AST
+  // Minify with UglifyES
```
kzc: Suggest to change to "Minify with uglify-es" to match the npm package name. The term "UglifyES" is not used by its maintainers.

DeMoorJasper: Will do, tnx for the suggestion 💯

kzc: ...or better yet - drop the comment altogether.

kzc:
I'm not surprised that the timings are not impacted very much for ES6. Converting an AST is an expensive operation.

Regarding the timings for Typescript - is it faster now because an AST no longer has to be generated?

DeMoorJasper:
I have no clue why TypeScript became so fast. It normally goes through the babel parser as well (generating the AST).
The TS example used in the tests is smaller than the JS one normally so that explains compared to JS though.

kzc:
With this PR is the babel AST still needed? 

https://github.com/parcel-bundler/parcel/blob/548eef92e11711db007b7613ba5530de508d21a0/src/transforms/babel.js#L23-L25

DeMoorJasper:
I'm pretty sure it is, it's being used in different places

kzc:
I'm kind of puzzled by https://github.com/parcel-bundler/parcel/pull/157/commits/9d59def0c4a36d1dfd3722cc7cc14fa15d033f17. `babel-generator` and `uglify-es` are like oil and water unless you add ES6+ ESTree support to `uglify-es`. Is this for non-JS assets?

DeMoorJasper:
I just copied that piece of code from JSAsset.generate() to uglify, so that it would only generate code from the AST @kzc 
I'm still trying to figure out all ways to improve upon this though

kzc:
Okay, I see. Uglify is only fed ES source code, not ESTree.

src/transforms/uglify.js
```diff
@@ -1,18 +1,37 @@
...
+  };
+
+  let result = minify(code, options);
+
+  if (result.error) throw result.error;
+
+  // Log all warnings
+  if (result.warnings) {
+    result.warnings.forEach(warning => {
+      // TODO: warn this using the logger
```
DeMoorJasper: Would probably be better to use the logger to send warnings to the console, not really sure how to address the logger from here though

src/transforms/uglify.js
```diff
@@ -1,18 +1,37 @@
...
 
-  // Convert to UglifyJS AST
-  var ast = AST_Node.from_mozilla_ast(toEstree(asset.ast, asset.contents));
-  var result = minify(ast, {
-    toplevel: true
-  });
+  // Convert AST into JS
+  let code = asset.isAstDirty
+    ? generate(asset.ast).code
+    : asset.outputCode || asset.contents;
```
devongovett: maybe just call `asset.generate().js` here?

src/transforms/uglify.js
devongovett: hmm yeah, the logger isn't really accessible from here, it lives only in the main process right now, not workers.

src/transforms/uglify.js
DeMoorJasper: I extracted this bit from generate because i saw TODO: Source maps and thought this would affect performance in the future. I'll change it back to asset.generate though :)

Issue #96, opened 2017-12-07
Imported image paths during development


jaredpalmer:
I am having issues with image asset paths during development and production. 

During dev, if i do `import logo from './logo.svg` this becomes => `1230981023.png`.  
During build, `import logo from './logo.svg` this becomes => `/dist/1230981023.png`. 

This may have to do with a public URL issue during dev but i'm not sure.

devongovett:
url is generated here https://github.com/parcel-bundler/parcel/blob/master/src/assets/RawAsset.js#L10

thejameskyle:
Also see https://github.com/parcel-bundler/parcel/blob/7b899b38930a7a30943a6148f475a5007ce2c625/src/Asset.js#L140-L157

RakanNimer:
Hey, 

Thanks for the great work on parcel, going through the code was fun !

I wrote a fix for this that's working on my machine.

I'm cleaning it up and will send a PR in a bit if no one is working on this.

Cheers !

gisderdube:
I had the same issue with importing assets (jpg|svg), but after fiddling with the `--public-url` it worked out just fine. Perhaps you can share some of your project structure?

devongovett:
Should be fixed by https://github.com/parcel-bundler/parcel/pull/160. You just need to set `--public-url` when building for prod.

samsch:
I think I'm having the same issue using Parcel 1.2.1 development server (`parcel src/index.html`). My imported url (`import image from './path/to/image.jpg';`) is `/24e15f924c17a873dc81435468e5287b.jpg`, while it would need to be `/dist/24e15f924c17a873dc81435468e5287b.jpg` to work.

QuantumInformation:
I have an issue here with a path that is correct for importing an image

<img width="823" alt="screen shot 2019-01-24 at 11 53 42" src="https://user-images.githubusercontent.com/216566/51676518-ba402200-1fce-11e9-8b50-e3a5f0256a53.png">

mischnic:
> I have an issue here with a path that is correct for importing an image

Does it build/run correctly?
Then it's only TypeScript not recognizing it, does this fix work: https://github.com/Microsoft/TypeScript-React-Starter/issues/12#issuecomment-326370098

QuantumInformation:
It builds but i get

⠧ Building SpaceInvaders.ts...🚨  /Users/nikos/WebstormProjects/youtube-space-invaders/src/SpaceInvaders.ts(14,21)
[...]

PR #160, opened 2017-12-09
Fix RawAsset loading.

RakanNimer:
Hello ! 

## Fixes : 
https://github.com/jaredpalmer/react-parcel-example/issues/6 https://github.com/parcel-bundler/parcel/issues/96 #186 

## Problem : 
In RawAsset, the generated bundle name was not prepended by the public url.

## This solution : 

Prepend RawAsset generated bundle name with options.publicURL (defaults to dist). Updated tests to reflect the change.

## Tests : 

Passing. I also, tested my fork against https://github.com/jaredpalmer/react-parcel-example and it works in both dev and prod mode. 

Thanks for the great work on Parcel ! 👍 
Please feel free to close if this code affects other parts negatively.

### DIFF
--- a/src/assets/RawAsset.js
+++ b/src/assets/RawAsset.js
@@ -6,8 +6,12 @@ class RawAsset extends Asset {
   load() {}
 
   generate() {
+    const pathToAsset = path.join(
+      this.options.publicURL,
+      this.generateBundleName()
+    );
     return {
-      js: `module.exports=${JSON.stringify(this.generateBundleName())};`
+      js: `module.exports=${pathToAsset};`
     };
   }
 }
--- a/src/assets/RawAsset.js
+++ b/src/assets/RawAsset.js
@@ -7,7 +7,7 @@ class RawAsset extends Asset {
 
   generate() {
     const pathToAsset = JSON.stringify(
-      `${this.options.publicURL}/${this.generateBundleName()}`
+      path.join(this.options.publicURL, this.generateBundleName())
     );
     return {
       js: `module.exports=${pathToAsset};`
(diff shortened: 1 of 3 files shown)


test/javascript.js
```diff
@@ -63,10 +63,12 @@ describe('javascript', function() {
...
-      childBundles: [{
-        assets: ['local.js'],
-        childBundles: []
-      }]
+      childBundles: [
+        {
+          assets: ['local.js'],
+          childBundles: []
+        }
+      ]
```
davidnagli: Is this an actual change, or just an accidental reformatting?

src/assets/RawAsset.js
```diff
@@ -6,8 +6,11 @@ class RawAsset extends Asset {
   load() {}
 
   generate() {
+    const pathToAsset = JSON.stringify(
+      path.join(this.options.publicURL, this.generateBundleName())
```
devongovett: will this cause a `\` to be used on windows instead of `/`?

devongovett: you probably want `url.resolve` https://nodejs.org/api/url.html#url_url_resolve_from_to

RakanNimer: Thanks  for the heads up. On it.

RakanNimer: Done. I had misunderstood how modules are loaded. Thanks for

drejohnson:
Not sure if this was supposed to fix images not loading in dev environment but I'm still having this issue on v1.2.0

davidnagli:
@drejohnson I think you should make a new issue or comment on one of the closed ones if they pertain to the bug your experiencing. We definitely need to look into this if it’s still a problem! 

PR #161, opened 2017-12-09
Added https development server support.

mbaranovski:
Added support for hosting files over https.

On the first run with `--https` parameter it generates `.pem` and `.cert` file and saves it into the `.cache` folder and serves files over https. During the next runs it checks if `.pem` and `.cert` file is cached or not and use them. 

Example: `parcel index.html --https`


davidnagli:
Wow this is awesome

bin/cli.js
```diff
@@ -19,6 +19,7 @@ program
     '--public-url <url>',
     'set the public URL to serve on. defaults to the same as the --out-dir option'
   )
+  .option('--https', 'serves files over HTTPS')
```
devongovett: Rather than making this an option, could we just start the server on both http and https so they both work at the same time?

devongovett: though i suppose they'd need to be on different ports :/

davidnagli: I would imagine that generating TLs Crete is fairly expensive, which is a problem when we’re going for efficiency.

davidnagli: **TLS certs

devongovett: If that's an actual problem we could always cache them.

brandon93s: Looking at `src/utils/generateCertificate.js`, it already caches the generated cert and serves it from the fs automatically. Paying the price *once* on first-run might not be such a big deal. It does raise these questions, though:

- Which does `--open` launch?
- Do we log both `address:port` combinations?
- Do we have a `--no-https` to disable it?

mbaranovski: Yes I do cache the generated Certificate. We can have both servers running and redirect to https, or have only https running all the time and only disable it by adding `--no-https` option.

yeskunall: Would there ever be a reason to _not_ serve over HTTPS? I say we default to HTTPS by default and add a `--no-https` to disable it.

brandon93s: > Would there ever be a reason to not serve over HTTPS?

It's a self-signed cert, which the browser is going to warn you about with a big scary ⚠️ message. I think making https opt-in is ideal. Use it when the use case warrants it.

yeskunall:
Just a thought: what if we used something like ngrok? [Here's](https://github.com/bubenshchykov/ngrok) a Node wrapper for it. I personally use ngrok for all my local HTTPS needs. 🤗

eubenesa:
Any update on this PR? Any way I can help?

mbaranovski:
I'll try to find some time to resolve conflicts today. As I understand we stick with the current approach and we leave https as optional.

brandon93s:
@mbaranovski I'm in agreement, stick with it as optional.

shawwn:
This is probably not a good idea. It's extremely simple for nginx to act as a frontend and route requests over to parcel, so if some user hypothetically wanted to do this, that'd be a better approach.

Beyond that, no users have asked for this feature. Why are we adding it?

Ahahaha...

Never mind. I just researched it.

Apparently users do in fact want this, regardless of how absurd it is: https://stackoverflow.com/questions/26663404/webpack-dev-server-running-on-https-web-sockets-secure

Webpack dev server went through the same process: https://github.com/webpack/webpack-dev-server/issues/10

That dude used his webpack dev server as a SharePoint Online / Office 365 Server. A true Microsoft solution if I've ever seen one. (Just kidding; love you, Microsoft devs.)

Careful to note the potential pitfalls here: https://stackoverflow.com/a/42050155

Specifically, we might need to support loading a specific cert.pem and key.pem file, and then when we do that, we'll need to support reading passphrases from the key.

Without those CLI options, there's not much point of just having HTTPS, is there? Users' browsers will always throw an error message warning about HTTPS insecurity.

BrieucP:
@shawwn : Why adding it ? To enable service worker and other thing that requires https while working on real mobile device.

mbaranovski:
@brandon93s do you happen to know where .cache folder has been moved now? I no longer see it.

mbaranovski:
Branch updated and conflicts resolved. Did several tests within my private project and everything seems to be working as expected. Please verify and merge @devongovett

devongovett:
Going to merge this, but it would be cool if you could follow up with a couple things:

1. Support for passing a custom `--cert` and `--key`. This will be needed for example if you have a trusted certificate already registered on your system (e.g. keychain on mac).
2. Support for running the HMR server over HTTPS as well. Currently it still runs on HTTP even when the main server is running on HTTPS, which probably isn't great.

Thanks!

mbaranovski:
Okay, will look into this two things.

devongovett:
Also added a test in ed11a2e1d21fec175e6fd457b174d8bb9c8d4a4a

wescoder:
Using this with a proxy server like `*.lvh.me` calls the ws server without `wss`.
For example starting `parcel src/index.html --https`
And trying to access `https://my-project.lvh.me:1234` should work as fine as `https://localhost:1234`, but it make attempts to access `ws://my-project.lvh.me` instead of `wss://my-project.lvh.me` throwing:
```
Mixed Content: The page at 'https://my-project.lvh.me:1234/' was loaded over HTTPS, but attempted to connect to the insecure WebSocket endpoint 'ws://my-project.lvh.me:41351/'. This request has been blocked; this endpoint must be available over WSS.

Uncaught DOMException: Failed to construct 'WebSocket': An insecure WebSocket connection may not be initiated from a page loaded over HTTPS.
    at Object.require.0 (https://my-project.lvh.me:1234/dist/f0df77a449d3efcf7d04e9b3934c86ee.js:1136:12)
    at newRequire (https://my-project.lvh.me:1234/dist/f0df77a449d3efcf7d04e9b3934c86ee.js:42:24)
    at require.4 (https://my-project.lvh.me:1234/dist/f0df77a449d3efcf7d04e9b3934c86ee.js:67:5)
    at https://my-project.lvh.me:1234/dist/f0df77a449d3efcf7d04e9b3934c86ee.js:72:3
```

mbaranovski:
Is it working pointing to wss when used w/o proxy?

wescoder:
Yeah, acessing `https://localhost:1234/` points to `wss` properly.

wescoder:
If you want to se the configs I'm using you can check this [repo](https://github.com/wescoder/tavern) I'm working on.

theomessin:
I second @wescoder. I have the same issue as I use a localhost development server with `https`. Tried using `parcel` with `--https` and I get `wss` when on `localhost:1234` but `ws` on my custom development server. Would be great if `parcel watch` had an `--https` option that switched to `wss`.

jraoult:
Are you sure about that @wescoder ?:
 > Yeah, acessing https://localhost:1234/ points to wss properly.

For me it is still `ws://` but it is working since the same origin policies for localhost are relaxed. Looking at the code I don't think there is a way right now to get any other URL than the `ws` one (cf. https://github.com/parcel-bundler/parcel/blob/v1.5.1/src/builtins/hmr-runtime.js#L19)

We would need to pass the https server instance as a parameter (cf.https://github.com/websockets/ws/blob/master/examples/ssl.js) when instiantiating the [ws server](https://github.com/parcel-bundler/parcel/blob/v1.5.1/src/HMRServer.js#L7) to make it secure and switch the protocol in the URL when `--https` option is on.

@mbaranovski do you want me to have a look at that?

wescoder:
In a couple of hours I'll double check this...

wescoder:
@jraoult I tested in a project from scratch and as you said it goes trough `ws://` and not `wss://`, I guess I checked it wrong on the first test.

jraoult:
@wescoder ok it is coherent with my observations.

@mbaranovski I can try to come up with a fix tomorrow unless you are already on it?

niicojs:
Done in #708

mbaranovski:
@jraoult sorry for being unresponsive.
It looks like @niicojs PR is going to be merged.

PR #164, opened 2017-12-09
User friendly server errors and automatic port switching

### LINKED ISSUE #111: Better failover for EADDRINUSE
🙋 feature request

I’ve noticed people complaining about this issue before, and I’ve encountered it accidentally several times. If you already have something else using `:1234` (like another process running parcel) you get a pretty nasty error message.

```
⏳  Building...
events.js:183
      throw er; // Unhandled 'error' event
      ^

Error: listen EADDRINUSE :::1234
    at Object._errnoException (util.js:1024:11)
    at _exceptionWithHostPort (util.js:1046:20)
    at Server.setupListenHandle [as _listen2] (net.js:1351:14)
    at listenInCluster (net.js:1392:12)
    at Server.listen (net.js:1476:7)
    at Object.serve (/Users/davidnagli/.config/yarn/global/node_modules/parcel-bundler/src/Server.js:50:49)
    at Bundler.serve (/Users/davidnagli/.config/yarn/global/node_modules/parcel-bundler/src/Bundler.js:421:19)
    at Command.bundle (/Users/davidnagli/.config/yarn/global/node_modules/parcel-bundler/bin/cli.js:67:13)
    at Command.listener (/Users/davidnagli/.config/yarn/global/node_modules/commander/index.js:315:8)
    at emitTwo (events.js:126:13)
```
For people who use Node all the time, this error is pretty familiar, but for someone others it can be pretty confusing.

Can we just wrap a simple try/catch or check for `err` in the callback or whatever it is, so that we display a more readable message to the user?

Maybe we can even automatically choose a different port…

DeMoorJasper:
As discussed in closes #111 when server fails to start the error stack is very non user-friendly and it has no fallback if it fails.
This Pull Request provides both more user friendly errors for the server and a fallback to a random free port.


src/utils/customErrors.js
```diff
@@ -0,0 +1,14 @@
+let serverErrorList = {
+  EACCES: "You don't have access to bind the server to port {port}.",
+  EADDRINUSE: 'There is already a process listening on port {port}.'
+};
+
+function serverErrors(err, port) {
+  let desc = serverErrorList[err.code].replace(/{port}/g, port);
+  if (!desc) {
+    desc = `${err.code} accured while setting up server on port ${port}`;
```
brandon93s: "accured" should be "occurred"

DeMoorJasper: Yup indeed, i'll change it

src/Server.js
```diff
@@ -54,7 +54,22 @@ function middleware(bundler) {
   };
 }
 
-function serve(bundler, port) {
+function getFreePort(port) {
+  return new Promise((resolve, reject) => {
+    let server = http.createServer().listen(port);
+    server.once('error', err => {
+      resolve(0);
```
yeskunall: I think because you're resolving the `Promise` on error, it will return early and `port` will now be 0. So essentially, `getFreePort` will always return 0 for `port` if the provided port is in use.

Correct me if I'm wrong tho 😅

DeMoorJasper: You're correct this is the point though because node.js's net library sees port 0 as pick a free port which resolves into a random free port

DeMoorJasper: I'll change some stuff around to make it more clear in the code

yeskunall: TIL a bit more about `net`. Thanks for that @DeMoorJasper! 😁

DeMoorJasper: @yeskunall u can read all about it here: https://nodejs.org/api/net.html#net_server_listen_port_host_backlog_callback =D

src/Bundler.js
```diff
@@ -448,12 +449,31 @@ class Bundler extends EventEmitter {
...
+            )}`
+          : '';
+      this.logger.persistent(
+        `Server running at ${this.logger.chalk.cyan(
+          `http://localhost:${server.address().port}`
+        )} ${addon}\n`
+      );
+    });
+
+    return server;
```
devongovett: let's put this code in `Server.js` instead of here. Trying to avoid adding too much more to the `Bundler` :)

src/Server.js
```diff
@@ -54,7 +54,25 @@ function middleware(bundler) {
...
+    server.once('error', err => {
+      resolve(false);
+    });
+    server.once('listening', connection => {
+      server.close(() => {
+        resolve(true);
+      });
+    });
+  });
+}
```
devongovett: Do we need to create a whole separate server here? Could we use the normal server made below to do this, and handle errors?

rahulcs:
@DeMoorJasper any reason why you chose to make this change inside `src/bundler.js` instead of `bin/cli.js`? i feel that `bin/cli.js` is a more appropriate place to make this change. 
you can also check out [detect-port](https://github.com/node-modules/detect-port) which projects like [create-react-app](https://github.com/facebookincubator/create-react-app) are using.

rahulcs:
with this change, if i run the command:
```bash
$ parcel index.html --port=2345
```
i get a message
`Server running at http://localhost:2345 - configured port 2345 could not be used.`
even when port 2345 is free.

src/Bundler.js
```diff
@@ -448,12 +449,31 @@ class Bundler extends EventEmitter {
     return Server.middleware(this);
   }
 
-  serve(port = 1234) {
-    this.logger.persistent(
-      'Server running at ' + this.logger.chalk.cyan(`http://localhost:${port}`)
-    );
+  async serve(port = 1234) {
```
rahulcs: this change will break `--open` option on `bin/parcel`

src/Server.js
yeskunall: IMO, moving the logic to check for a free port to its own method seems apt, although that's just a matter of preference. 😅

brandon93s: If we're not opposed to _another_ dependency, something like [get-port](https://www.npmjs.com/package/get-port) would simplify things. It lets you pass in the default, and returns an ephemeral port if it isn't available. 

```js
const port = getPort({port: 1234});
```

DeMoorJasper: @brandon93s code wise this package does exactly the same so it should not really make a difference and would clean up the code a bit i'll pivot towards it, keeps things clean

src/Bundler.js
DeMoorJasper: fixed in aaa5308c2dac6ddb3b637d0c9bbaff8f12ebca97

src/Server.js
DeMoorJasper: fixed in aaa5308c2dac6ddb3b637d0c9bbaff8f12ebca97

src/Bundler.js
DeMoorJasper: This should now be fixed

PR #166, opened 2017-12-09
fix: replacing process.env member node with value node

maciej-ka:
Hi, 
Thanks for this project!

Fixes: https://github.com/parcel-bundler/parcel/issues/117 The bug is happening because replace function doesn't work with a CallExpression node.

`replaceIn(parent, from, to)` dives deep into `parent` object properties and arrays until the first occurrence of argument `from` which it replaces into argument `to`. I made an assumption, about AST, that nodes are present in parent at most once.

Some ideas how to improve this: 
- make `replaceIn` and `deepEquals` as general node utility functions.
- implement in Node type hierarchy, one per type. 
- compare node only on base of some id, maybe it's enough to compare `loc` and `type`?

### DIFF
--- a/src/visitors/globals.js
+++ b/src/visitors/globals.js
@@ -24,7 +24,7 @@ module.exports = {
       let key = types.toComputedKey(node);
       if (types.isStringLiteral(key)) {
         let val = types.valueToNode(process.env[key.value]);
-        replace(ancestors[ancestors.length - 2], node, val);
+        replaceIn(ancestors[ancestors.length - 2], node, val);
         delete node.object; // prevent traversing into this node
         asset.isAstDirty = true;
       }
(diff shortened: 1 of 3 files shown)


joenoon:
@maciej-ka awesome!  I noticed in my issue #117 there was also another user who seems to have fixed this in their branch: https://github.com/Raathigesh/parcel/commit/e292ee065f10a7c8fb6a4a67dd7c5a4831c8b09b

These look like different approaches to me and I have barely any babel/complier experience, so just wanted to mention that other commit so you and others who know what they are doing can decide how to proceed.

devongovett:
how does this relate to #193?

maciej-ka:
These pull requests are alternatives, if you pick one then close the rest: #153, #166, #193

devongovett:
fixed by #193.

PR #169, opened 2017-12-09
Make hmr-runtime builtin ES3-compatible

### LINKED ISSUE #131: The code that parcel inserts is not processed with babel

chee:
This is the only builtin that uses non-es3 code. 
My use-case is that I'm debugging our site in IE11, and I can't use `parcel watch` for it at present, as the HMR-runtime builtin causes the script to die.

this closes #131 as far as i'm concerned

### DIFF
--- a/src/builtins/hmr-runtime.js
+++ b/src/builtins/hmr-runtime.js
@@ -18,11 +18,11 @@ if (!module.bundle.parent) {
   var ws = new WebSocket('ws://localhost:{{HMR_PORT}}/');
   ws.onmessage = function(event) {
     var data = JSON.parse(event.data);
-    var key, asset;
+    var index, asset;
 
     if (data.type === 'update') {
-      for (key in data.assets) {
-        asset = data.assets[key];
+      for (index = 0; index < data.assets.length; index++) {
+        asset = data.assets[index];
         hmrApply(global.require, asset);
         if (!asset.isNew) {
           hmrAccept(global.require, asset.id);
--- a/src/builtins/hmr-runtime.js
+++ b/src/builtins/hmr-runtime.js
@@ -1,4 +1,4 @@
-var global = (1,eval)('this');
+var global = (1, eval)('this');
 var OldModule = module.bundle.Module;
 function Module() {
   OldModule.call(this);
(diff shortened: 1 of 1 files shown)


src/builtins/hmr-runtime.js
```diff
@@ -16,15 +16,14 @@ module.bundle.Module = Module;
...
   var ws = new WebSocket('ws://localhost:{{HMR_PORT}}/');
-  ws.onmessage = (e) => {
-    var data = JSON.parse(e.data);
+  ws.onmessage = function(event) {
+    var data = JSON.parse(event.data);
+    var key, asset;
 
     if (data.type === 'update') {
-      for (let asset of data.assets) {
+      for (key in data.assets) {
```
devongovett: not a good idea to use a `for..in` loop over an array. can you either make this a normal `for` loop or `data.assets.forEach(function (asset) { ... })`?

davidnagli:
Why should we go through the pain of maintaining an ES3 compatible codebase? Why can't we just use Babel to transpile the Parcel code before shipping it?

Also, we use `let` and `class`, and `async` everywhere, so I don't see how this part could be the only "non-es3 code" 😂

src/builtins/hmr-runtime.js
chee: yep, i'll do that now.

chee:
@davidnagli these builtins don't get run through babel.
perhaps a better solution would be to process these with babel before they get inserted, but for now this seemed a fine approach as only one of them was non-es3

DeMoorJasper:
@chee It wouldn't be a big change to just run them through babel but would possibly impact performance though

tonkotsuboy:
I guess IE doesn't support template literal.

https://github.com/parcel-bundler/parcel/pull/169/commits/ff2641ac8165a636aca4a07ea68fe9e37afe9ded#diff-84b7f03297bb09457de68d252f8af43aR38

- [Template literal](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals)

PR #170, opened 2017-12-09
Package.json can't be found bugfix

### LINKED ISSUE #167: Plugins are not loaded if the entry is not on the same directory as package.json
This is a 🐛 bug report.

### 🎛 Directory structure

```
- package.json
+ src
  - main.js
+ node_modules
  + parcel-bundler
  + parcel-plugin-plugin
```

### 🤔 Expected Behavior
Parcel should be able to load the plugins from the `package.json`.

### 😯 Current Behavior
It does not if the entry file is not in the root directory of the project.

The error is silenced but if we log it we get :
```
Error: Cannot find module './package.json' from '[project]/src'
at Function.module.exports [as sync] ([project]/node_modules/resolve/lib/sync.js:40:15)
at module.exports ([project]/node_modules/parcel-bundler/src/utils/localRequire.js:11:24)
at Bundler.loadPlugins ([project]/node_modules/parcel-bundler/src/Bundler.js:86:17)
at new Bundler ([project]/node_modules/parcel-bundler/src/Bundler.js:44:10)
at Command.bundle ([project]/node_modules/parcel-bundler/bin/cli.js:66:19)
at Command.listener ([project]/node_modules/commander/index.js:315:8)
at emitTwo (events.js:126:13)
at Command.emit (events.js:214:7)
at Command.parseArgs ([project]/node_modules/commander/index.js:653:12)
at Command.parse ([project]/node_modules/commander/index.js:475:21) code: 'MODULE_NOT_FOUND'
```

### 💁 Possible Solution

I dirty-fixed it by replacing `./package.json` to `../package.json`, but the best solution should be to traverse the directory tree until a `package.json` is found.

### 🔦 Context

Found this while working on a plugin

### 🌍 Your Environment

| Parcel           |  1.1.0
| Node             | 8.9.1
| npm/Yarn         | 1.3.2
| Operating System | macOS 10.13.2

DeMoorJasper:
Currently the package.json file is only searched for within the directory and not in parent directories, this should fix this.

As discussed here: closes #167

### DIFF
--- a/src/Bundler.js
+++ b/src/Bundler.js
@@ -87,16 +87,21 @@ class Bundler extends EventEmitter {
     this.packagers.add(type, packager);
   }
 
-  loadPlugins() {
+  loadPlugins(location = this.mainFile) {
     try {
-      let pkg = localRequire('./package.json', this.mainFile);
+      let pkg = localRequire('./package.json', location);
       let deps = Object.assign({}, pkg.dependencies, pkg.devDependencies);
       for (let dep in deps) {
         if (dep.startsWith('parcel-plugin-')) {
-          localRequire(dep, this.mainFile)(this);
+          localRequire(dep, location)(this);
         }
       }
     } catch (err) {
+      let currPath = Path.dirname(this.mainFile);
+      currPath = currPath.substring(0, currPath.lastIndexOf('/'));
+      if (currPath.length > 1) {
+        return this.loadPlugins(currPath);
+      }
       this.logger.warn(err);
     }
   }
--- a/src/Bundler.js
+++ b/src/Bundler.js
@@ -88,8 +88,15 @@ class Bundler extends EventEmitter {
   }
 
   loadPlugins(location = this.mainFile) {
+    let pkg;
     try {
-      let pkg = localRequire('./package.json', location);
+      pkg = localRequire('./package.json', location);
+    } catch (err) {
+      if (err.code === 'MODULE_NOT_FOUND' && location.length > 1) {
+        return this.loadPlugins(Path.join(location, '..'));
+      }
+    }
+    try {
       let deps = Object.assign({}, pkg.dependencies, pkg.devDependencies);
       for (let dep in deps) {
         if (dep.startsWith('parcel-plugin-')) {
@@ -97,9 +104,6 @@ class Bundler extends EventEmitter {
         }
       }
     } catch (err) {
-      if (err.code === 'MODULE_NOT_FOUND' && location.length > 1) {
-        return this.loadPlugins(Path.join(location, '..'));
-      }
       this.logger.warn(err);
     }
   }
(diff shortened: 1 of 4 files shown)


src/Bundler.js
```diff
@@ -87,16 +87,21 @@ class Bundler extends EventEmitter {
...
       let deps = Object.assign({}, pkg.dependencies, pkg.devDependencies);
       for (let dep in deps) {
         if (dep.startsWith('parcel-plugin-')) {
-          localRequire(dep, this.mainFile)(this);
+          localRequire(dep, location)(this);
         }
       }
     } catch (err) {
+      let currPath = Path.dirname(this.mainFile);
+      currPath = currPath.substring(0, currPath.lastIndexOf('/'));
```
brandon93s: May be more robust to use `Path` to traverse up a directory.

```js
currPath  = Path.join(currPath, '..');
```

src/Bundler.js
```diff
@@ -87,16 +87,21 @@ class Bundler extends EventEmitter {
...
-      let pkg = localRequire('./package.json', this.mainFile);
+      let pkg = localRequire('./package.json', location);
       let deps = Object.assign({}, pkg.dependencies, pkg.devDependencies);
       for (let dep in deps) {
         if (dep.startsWith('parcel-plugin-')) {
-          localRequire(dep, this.mainFile)(this);
+          localRequire(dep, location)(this);
         }
       }
     } catch (err) {
```
brandon93s: Is `err` always because it isn't found, or could it be something like access denied? Wouldn't want to keep traversing if it is found but just couldn't read it. That's a legitimate error.

DeMoorJasper: I'll look into it

marcioj:
@DeMoorJasper nice work! Do you think it is possible to add tests for this issue? In order to avoid future regressions?

DeMoorJasper:
@marcioj i'll look into it, not sure how to write tests for this though

src/Bundler.js
```diff
@@ -87,16 +87,19 @@ class Bundler extends EventEmitter {
...
+      if (err.code === 'MODULE_NOT_FOUND' && location.length > 1) {
```
devongovett: we probably only want to do this if the require of `package.json` failed, not one of plugins.

what if we used the `config.load` util function that already exists to load the package.json file instead? that will already do the traversing for us.

DeMoorJasper: Ow didn't know that i'll change it to config.load

marcioj:
> @marcioj i'll look into it, not sure how to write tests for this though

@DeMoorJasper I'd try to create something based on this [test](https://github.com/parcel-bundler/parcel/blob/master/test/plugins.js#L5). You can then copy the`test/integration/plugins` and change to a subfolder like structure. Like src/index.js and src/test.txt ..

DeMoorJasper:
Tests should handle both traversing up as well

PR #172, opened 2017-12-09
Add Travis, npm-version, npm-dm and twitter badges.

gagan0723:


DeMoorJasper:
Would be nice if you could add dependancy status as well
[Dependancies](https://david-dm.org/parcel-bundler/parcel)

gagan0723:
Done

Issue #174, opened 2017-12-09
Tests Fail on Windows | +AppVeyor


brandon93s:
**Choose one:**   &nbsp;&nbsp;&nbsp;&nbsp;  🐛

### 🤔 Expected Behavior

Tests pass on all platforms.

### 😯 Current Behavior

Currently, ~50% of the tests are failing on Windows ([see AppVeyor results](https://ci.appveyor.com/project/brandon93s/parcel)). Some of these are due to a failure in a `beforeEach` hook, but others are legitimate failures. **The test failures could mean that `parcel` isn't working correctly on Windows. (e.g. #86)**

### 💁 Possible Solution

1) Add AppVeyor alongside Travis.ci
2) Start working through and correcting the failures

### 🌍 Your Environment

Occurring on Windows 10 (local) and Windows Server 2012/2016 (AppVeyor).
Occurring on Node.js v9.2.0 (local) and Node.js v8.x.x (AppVeyor).

brandon93s:
Working through these now.

brandon93s:
Stylus tests are failing with the same issues as seen in #158. Stylus may not be working on windows at all at the moment.

aicest:
@brandon93s some details added to #159.

PR #175, opened 2017-12-09
Add AppVeyor for Windows builds

brandon93s:
[AppVeyor Results](https://ci.appveyor.com/project/brandon93s/parcel)
See #174 for additional details

Should we go ahead and get the builds running, and then try to fix all of them? Or fix them all first (a lot of them)? Note that this PR does not include any fixes yet, just the tests running.

### DIFF
--- a/test/utils.js
+++ b/test/utils.js
@@ -7,7 +7,11 @@ const path = require('path');
 const WebSocket = require('ws');
 
 beforeEach(function() {
-  rimraf.sync(path.join(__dirname, 'dist'));
+  try {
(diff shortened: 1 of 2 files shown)


test/utils.js
```diff
@@ -7,7 +7,11 @@ const path = require('path');
 const WebSocket = require('ws');
 
 beforeEach(function() {
```
brandon93s: This fails on Windows, and is fatal when it does. To allow the tests to continue executing past `hmr`, this is catching and logging the error. This can / should be removed once the tests are fixed on Windows.

PR #180, opened 2017-12-09
Uses ESLint to catch errors and fixes ESLint errors #173

### LINKED ISSUE #173: Introduce ESLint for linting
🙋

N/A

### 🤔 Expected Behavior

To be able to catch errors like #165, and overall introduce linting into `parcel`.

### 😯 Current Behavior

Since there is no setup for linting, we miss out on being able to enhance code quality, and just catch errors in general (which I believe ESLint does a solid job)

### 💁 Possible Solution

Introduce ESLint in some way/shape/form. Either use existing configs like [`eslint-config-airbnb`](https://www.npmjs.com/package/eslint-config-airbnb) or just write our own config.

### 🔦 Context

Not affected me __yet__, but I think adding some form of linting will overall benefit `parcel`.

### 🌍 Your Environment

N/A

Let me know your thoughts on this. 💖

/cc @davidnagli

ghusse:
Adds ESLint to the build process:

- Adds `.eslintrc.json` files to configure ESLint at different levels
- Fixes some bugs / warnings revealed by ESLint
- Adds inline comments for disabling ESLint in some situations
- Launches ESLint in Travis CI
- Launches ESLint in the precommit hook

For some files in integration tests, I had to disable ESLint because the syntax is a mix of the `import` `export` from ES6, and the support of a special function named `import`. I suppose that this is revealing an issue with the current syntax supported by parcel, but in the meantime, I just ignored these files as they cannot be parsed by ESLint.

For the moment, only the recommended rules from ESLint have been applied. Feel free to propose customizations.

----

Closes #173


src/Bundler.js
```diff
@@ -262,19 +261,21 @@ class Bundler extends EventEmitter {
     try {
       return await this.resolveAsset(dep.name, asset.name);
     } catch (err) {
-      if (err.message.indexOf(`Cannot find module '${dep.name}'`) === 0) {
-        err.message = `Cannot resolve dependency '${dep.name}'`;
+      let thrown = err;
```
ghusse: Fixes [no-ex-assign](https://eslint.org/docs/rules/no-ex-assign)

src/transforms/uglify.js
```diff
@@ -1,7 +1,5 @@
 const {AST_Node, minify} = require('uglify-js');
 const {toEstree} = require('babel-to-estree');
-const types = require('babel-types');
```
ghusse: Are these `require` statements required for babel? I removed them because the variables are not used.

src/worker.js
```diff
@@ -1,14 +1,8 @@
 require('v8-compile-cache');
-const fs = require('./utils/fs');
 const Parser = require('./Parser');
-const babel = require('./transforms/babel');
 
 let parser;
 
-function emit(event, ...args) {
```
ghusse: This function does not seem to be called

src/Bundler.js
```diff
@@ -2,7 +2,6 @@ const fs = require('./utils/fs');
 const Resolver = require('./Resolver');
 const Parser = require('./Parser');
 const WorkerFarm = require('./WorkerFarm');
-const worker = require('./utils/promisify')(require('./worker.js'));
```
ghusse: Is it normal that `worker` was not used?

src/Asset.js
```diff
@@ -73,7 +72,7 @@ class Asset {
 
     let resolved = path
       .resolve(path.dirname(from), url)
-      .replace(/[\?#].*$/, '');
+      .replace(/[?#].*$/, '');
```
yeskunall: I'm not very good with regex, so simply ignore if I'm being stupid here, but looks like ESLint removed `\`, which could potentially change the result of `resolved`. Wouldn't that mean that it would now also remove `?` and `#` from the path? 🤔

src/FSCache.js
```diff
@@ -70,7 +71,9 @@ class FSCache {
     try {
       await fs.unlink(this.getCacheFile(filename));
       this.invalidated.delete(filename);
-    } catch (err) {}
+    } catch (err) {
```
yeskunall: Did ESLint complain about this?

Also, never really a good idea to have anything fail silently, imo. But that's another discussion.

src/assets/CSSAsset.js
```diff
@@ -1,12 +1,10 @@
 const Asset = require('../Asset');
 const postcss = require('postcss');
 const valueParser = require('postcss-value-parser');
-const path = require('path');
-const md5 = require('../utils/md5');
 const postcssTransform = require('../transforms/postcss');
 const CssSyntaxError = require('postcss/lib/css-syntax-error');
 
-const URL_RE = /url\s*\(\"?(?![a-z]+:)/;
+const URL_RE = /url\s*\("?(?![a-z]+:)/;
```
yeskunall: Here too. ESLint seems to have removed some bit of regex. Again, ignore if I'm wrong. 😅

src/builtins/bundle-url.js
```diff
@@ -22,7 +22,7 @@ function getBundleURL() {
 }
 
 function getBaseURL(url) {
-  return ('' + url).replace(/^((?:https?|file|ftp):\/\/.+)\/[^\/]+$/, '$1') + '/';
+  return ('' + url).replace(/^((?:https?|file|ftp):\/\/.+)\/[^/]+$/, '$1') + '/';
```
yeskunall: Here too. ESLint seems to have removed some bit of regex. Again, ignore if I'm wrong. 😅

src/packagers/Packager.js
```diff
@@ -17,6 +17,7 @@ class Packager {
 
   async start() {}
 
+  // eslint-disable-next-line no-unused-vars
```
yeskunall: Why is this particular one needed? I don't see any unused vars.

src/utils/is-url.js
```diff
@@ -4,7 +4,7 @@ const isURL = require('is-url');
 const ANCHOR_REGEXP = /^#/;
 
 // Matches scheme (ie: tel:, mailto:, data:)
-const SCHEME_REGEXP = /^[a-z]*\:/i;
+const SCHEME_REGEXP = /^[a-z]*:/i;
```
yeskunall: Here too. ESLint seems to have removed some bit of regex. Again, ignore if I'm wrong. 😅

.eslintrc.json
```diff
@@ -0,0 +1,10 @@
+{
+  "extends": "eslint:recommended",
+  "parserOptions": {
+    "ecmaVersion": 8
```
yeskunall: Setting this to 8 will result in some features __not__ supported by Node 6 to no longer be caught by ESLint, so we should beware of this.

yeskunall:
@ghusse noticed a trend where ESLint just removed some parts of the regex. I've mentioned that in quite a few places in my review (sorry about that), but other than that, this looks really good. Appreciate it. 💖

kingdaro:
I believe those are in places where the escape is unnecessary, so it should be fine 🙂

yeskunall:
Yup, that's what ESLint says. But I pushed some code once that removed those escapes without checking and my tests started failing. Admittedly, I didn't know what I was doing with regex then, and I still don't. 😅

Thanks tho @kingdaro 💖

src/Asset.js
brandon93s: In this case, they are the same. 

```
[\?#]
=> match any item in the list: 
- literal ?
- literal #

[?#]
=> match any item in the list:
`` ?#

File: src/assets/CSSAsset.js (hunk 7 above)
Comment (brandon93s) [reply src/assets/CSSAsset.js]: Same here. The `\` was an unnecessary character escape. Same functionality.

File: src/builtins/bundle-url.js (hunk 8 above)
Comment (brandon93s) [reply src/builtins/bundle-url.js]: Same here. The \ was an unnecessary character escape. Same functionality.

File: src/utils/is-url.js (hunk 10 above)
Comment (brandon93s) [reply src/utils/is-url.js]: Same here. The \ was an unnecessary character escape. Same functionality.

File: src/Asset.js (hunk 5 above)
Comment (yeskunall) [reply src/Asset.js]: Yeah @brandon93s I'm the type that has too look up SO for the simplest regex's. 😆 but thanks for looking into it!

File: bin/cli.js [hunk 12]
```diff
@@ -62,13 +62,15 @@ program
   });
 
 program.on('--help', function() {
+  /* eslint-disable no-console */
```
brandon93s: Since this is a CLI, should we just allow console globally in the `eslintrc`?

yeskunall:
So the general consensus is that I'm just bad at regex. Feel free to ignore all comments I made about the regex issues.

However, @ghusse I did notice a lot of `/* eslint-enable no-console */` (or something along the lines) during my initial review, but forgot to mention them. I think that for all purposes, this should be safe to turn off, or at the very list, switched to only warn the user.

package.json
```diff
@@ -71,6 +73,10 @@
     "node": ">= 6.0.0"
   },
   "lint-staged": {
-    "*.{js,json,md}": ["prettier --write", "git add"]
+    "*.{js,json,md}": [
+      "prettier --write",
+      "eslint --fix",
```
brandon93s: Running `eslint` on `md` files?

src/packagers/Packager.js
brandon93s: `asset` is unused. Guessing it was left to be self-documenting since this requires implementation by extending classes.

package.json
yeskunall: Yeah, it should instead do a `npm run lint` before `lint-staged` in the precommit hook.

.eslintrc.json
ghusse: Actually, babel seem used everywhere in parcel. It allows to use `async`/`await` for instance, which is not supported by Node 6 and si part of ecmaScript 2017 or 8. That's why this version here is needed.

bin/cli.js
ghusse: Ok, I'll change that

package.json
ghusse: woops. Thanks. I'll change that

src/Asset.js
ghusse: Yep: ESLint was complaining about unecessary escapes.

src/FSCache.js
ghusse: Yes, that's why I added a comment: I did not want to deactivate this rule

src/packagers/Packager.js
ghusse: Exactly: I prefered not to remove it, because it seems to be useful for documentation

.eslintrc.json
yeskunall: Yup, you're right. The whole codebase is later transpiled using Babel. This should be okay. Thanks for the clarification 😅

ghusse:
I pushed new modifications to this PR. Let me know if there are other changes you'd like to see.

brandon93s:
Are all of the sub  `eslintrc` necessary? Can they be consolidated into the parent by promoting the rules? I see they're extending one another, but would want to reduce as much maintenance overhead as possible.

ghusse:
I created a sub-config file everywhere the context changed, or different rules have to be applied. That's the way eslint is handling these cases.

I can comment every file to explain why I created it if you want.

.eslintrc.json
```diff
@@ -0,0 +1,10 @@
+{
```
ghusse: That is the master configuration file

bin/.eslintrc.json
```diff
@@ -0,0 +1,6 @@
+{
```
ghusse: I created this one in the `bin` directory to deactivate the `no-console` rule just here.

I propose not to deactivate it everywhere, because usage of `console.log` can be signs of forgotten test code. But if you prefer I can move it to the general configuration (but deactivate it in tests)

src/builtins/.eslintrc.json
```diff
@@ -0,0 +1,9 @@
+{
```
ghusse: I needed to specify the context used by files in `builtins`, as a browser context. Some files are using features specific to browsers (such as `WebSocket` for what I remember)

test/.eslintrc.json
```diff
@@ -0,0 +1,6 @@
+{
```
ghusse: Needed to specify that in this repository, the environment is different (it is using mocha)

test/integration/.eslintrc.json
```diff
@@ -0,0 +1,10 @@
+{
```
ghusse: Integration test code is using browser-specific features and special functions `output` and `import` that are used in different files. This is not the case in the rest of the code.

test/integration/dynamic-esm/.eslintrc.json
```diff
@@ -0,0 +1,9 @@
+{
```
ghusse: This directory contains code that tests the ES6 syntax for importing and exporting code. ESLint needs a specific configuration to support this syntax, and we cannot use inline-configuration comments for this.

test/integration/es6-default-only/.eslintrc.json
```diff
@@ -0,0 +1,6 @@
+{
```
ghusse: Needed to support the parsing of ES6 import/export syntax

test/integration/es6/.eslintrc.json
```diff
@@ -0,0 +1,6 @@
+{
```
ghusse: Needed to support the parsing of ES6 import/export syntax

test/integration/fs-import/.eslintrc.json
```diff
@@ -0,0 +1,6 @@
+{
```
ghusse: Needed to support the parsing of ES6 import/export syntax

ghusse:
Did I introduce regressions? It seems that this build is pretty new, are the test passing on master?

ghusse:
Ok, I need to investigate why tests are failing. It seems that this PR introduces a regression somewhere, as test are passing on master.

ghusse:
Ok, indeed, I introduced regressions with my fixes on prelude.js. I just pushed a version that both work, and statisfies ESLint rule of not creating a function inside a if statement.

ghusse:
Ok, all tests are passing, and all the warnings from master are fixed in this branch.

src/builtins/.eslintrc.json
brandon93s: Please add the following:

```js
"parserOptions": {
  "ecmaVersion": 5
}
```

This will protect against future regressions (see #169, #227) since these aren't currently processed by Babel.

ghusse: Ok, I'll do it :-)

ghusse: Done, and I also replaced one `const` and one `let` by `var` in `src/builtins/index.js`.

ghusse:
Anything else to change on this PR?

yeskunall:
At this point, you should ping one of the members @ghusse 😁

davidnagli:
@yeskunall No need to ping us, we’re always watching 👀

yeskunall:
@davidnagli

CONTRIBUTING.md
```diff
@@ -71,4 +66,4 @@ Thank you to all our sponsors! (please ask your company to also support this ope
 <a href="https://opencollective.com/parcel/sponsor/6/website" target="_blank"><img src="https://opencollective.com/parcel/sponsor/6/avatar.svg"></a>
 <a href="https://opencollective.com/parcel/sponsor/7/website" target="_blank"><img src="https://opencollective.com/parcel/sponsor/7/avatar.svg"></a>
 <a href="https://opencollective.com/parcel/sponsor/8/website" target="_blank"><img src="https://opencollective.com/parcel/sponsor/8/avatar.svg"></a>
-<a href="https://opencollective.com/parcel/sponsor/9/website" target="_blank"><img src="https://opencollective.com/parcel/sponsor/9/avatar.svg"></a>
\ No newline at end of file
+<a href="https://opencollective.com/parcel/sponsor/9/website" target="_blank"><img src="https://opencollective.com/parcel/sponsor/9/avatar.svg"></a>
```
davidnagli: Did I miss something, or are these exactly the same?

yeskunall: EOL mismatch between OSes? Sure looks like it.

ghusse: Sh******

Thanks, I'll change that

ghusse: I checked the file, it seems that this file is now encoded using LF. I suppose that it was not the case on master. Do you want me to revert changes on that file?

davidnagli: I don’t think it makes any difference

ghusse:
@davidnagli @yeskunall Anything I can do for this PR to be merged?

yeskunall:
I don't have push access @ghusse 😟 (at this point you should really ping someone who has the access)

ghusse:
ping @brandon93s ?

brandon93s:
Looks ready to go to me.

devongovett:
Looks good. can you fix the conflicts?

ghusse:
Conflicts solved (I had to fix other errors introduced by recent commits)

yeskunall:
Hey @ghusse this is killer! Thanks for not giving up on this PR. Do you also mind squashing your commits? It's just a suggestion. 😄

bin/.eslintrc.json
aryaNmlp: It was a great time with you

devongovett:
Awesome, thanks for doing this!

No need to squash btw, I'll do that on merge.

PR #193, opened 2017-12-10
fix: global env problem by morphing a Node object

maciej-ka:
Closes: https://github.com/parcel-bundler/parcel/issues/117

This is a line from current code
```replace(ancestors[ancestors.length - 2], node, val);```

Parameters are called
```replace(parent, from, to)```

I just realized that `from` parameter is referentially same (will pass `obj === obj` comparison) as one of the fields somewhere in `parent`. So instead of manipulating in `parent`, we can change MemberExpression (or whatever node type of `from` is) into StringLiteral by just replacing its parameters while keeping its object identity.

### DIFF
--- a/src/visitors/globals.js
+++ b/src/visitors/globals.js
@@ -24,7 +24,7 @@ module.exports = {
       let key = types.toComputedKey(node);
       if (types.isStringLiteral(key)) {
         let val = types.valueToNode(process.env[key.value]);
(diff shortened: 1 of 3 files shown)


devongovett:
how does this relate to #166?

maciej-ka:
These pull requests are alternatives, if you pick one then close the rest: #153, #166, #193

PR #195, opened 2017-12-11
Updated dependencies

davidnagli:
Now that we added a [Dependency Badge](https://david-dm.org/parcel-bundler/parcel), we should make sure our dependencies are up-to-date.

1 had a security vulnerability (it would never affect anybody since it's on the dev server), and 2 were just out of date.


davidnagli:
Also closes #8 (according to @DeMoorJasper, upgrading to `v0.0.3` of `babel-to-estree` will fix the infamous problem)

PR #197, opened 2017-12-11
Normalize path for windows os

xcatliu:
This pr fixed the path issue in windows os.

When exec `parcel build site/index.html`, the assets in html will be transformed to another path in `dist/index.html` like this:

```html
<script src="\docs\89ee6b6452b6649c72b85a19637b8362.js"></script>
```

However, the `\` separator is not a standard separator for web. With this fix, it will be transformed to:

```html
<script src="/docs/89ee6b6452b6649c72b85a19637b8362.js"></script>
```

[normalize-path](https://www.npmjs.com/package/normalize-path) is a wildly used module to normalize the path
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,6 +26,7 @@
     "micromatch": "^3.0.4",
     "mkdirp": "^0.5.1",
     "node-libs-browser": "^2.0.0",
+    "normalize-path": "^2.1.1",
     "opn": "^5.1.0",
     "parse-json": "^4.0.0",
     "physical-cpu-count": "^2.0.0",
--- a/package.json
+++ b/package.json
@@ -26,7 +26,6 @@
     "micromatch": "^3.0.4",
     "mkdirp": "^0.5.1",
     "node-libs-browser": "^2.0.0",
-    "normalize-path": "^2.1.1",
     "opn": "^5.1.0",
     "parse-json": "^4.0.0",
     "physical-cpu-count": "^2.0.0",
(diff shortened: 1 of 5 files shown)


devongovett:
can you just use the builtin `url.normalize` for this instead of adding another dependency? https://nodejs.org/api/url.html#url_url_resolve_from_to

brandon93s:
That module seems to only fix escaped backslashes `\\`. Is that going to cover every uses case? Other areas of interest (probably more):

brandon93s:
_accidental close_

It actually looks like `url.resolve` might not handle this correctly, either: [repl.it](https://repl.it/repls/WelltodoClearcutAsiansmallclawedotter).

[normalize-path](https://www.npmjs.com/package/normalize-path) seems to only handle `\\` and not `\`. 

We may want to just roll our own utility for this...

kingdaro:
Correct me if I'm wrong, but, as I understand it, the `\` as it is in a JS string wouldn't ever show up in a path, right? Because `\\` in JS means `\`, but there's no equivalent for `\` since it's a special character, and not an output character. So `url.resolve` should work fine here

xcatliu:
@brandon93s I tried normalize-path and `url.resolve`, both of them can handle this correctly. As @kingdaro  said, `\` is an escape symbol in JS string. So `\\` means `\` in string.

```js
const url = require('url');

console.log(url.resolve('/base', 'path/to/resource'));
// will log /path/to/resource

console.log(url.resolve('/base/', 'path/to/resource'));
// will log /base/path/to/resource

console.log(url.resolve('\\base\\', 'path\\to\\resource'));
// will log /base/path/to/resource

console.log(url.resolve('\\path\\to\\resource', ''));
// will log /path/to/resource
```

PR #204, opened 2017-12-11
Hot reload all the parent modules. Fixes #177

### LINKED ISSUE #177: Hot reloading fails for nested components

This is a bug report.

Full example in a minimal repo available here: https://github.com/saadq/misc
Steps to reproduce are listed in the **readme.md** of that repo.

### 🤔 Expected Behavior
Hot reloading should work (without any need to refresh)

### 😯 Current Behavior
Hot reloading works for some components, fails for other components

### 💁 Possible Solution
I think the issue may be related to the the fact that components aren't used directly, but instead I import components from an **index.js** file which happen to import the desired components and reexport them.

### 🔦 Context
I love everything about Parcel, but hot reloading is kinda crucial to workflow nowadays.

### 🌍 Your Environment

| Parcel           | 1.1.0
| Node             | v8.3.0
| npm/Yarn         | 5.5.1
| Operating System | macOS Sierra 10.12.6

Raathigesh:
More discussion available on https://github.com/parcel-bundler/parcel/issues/177

I'm not quite sure whether this is the right fix but proposing this change here so we can get the discussion started regarding the issue.


devongovett:
the reason it is `some` instead of `forEach` is so that it stops at the module that accepts the update (via `module.hot.accept()`) and doesn't traverse higher in the tree. however, you are right that this is slightly incorrect since it will fail to update sibling parents.

what we need to do instead is update both parents, and then traverse higher ONLY if neither accepted the update. you will probably need to break up `hmrAccept` a bit to get this working.

Raathigesh:
@devongovett To prevent traversing higher if any of the parents accepted the update, Should I break up the `getParents()` method and first apply only to the sibling parents and if they don't accept the update then proceed with the other parents or is it related to the following lines in the `hmrAccept` method it-self.

https://github.com/parcel-bundler/parcel/blob/b7b2991d69b960d9f2951828b8145a6d9396ee4e/src/builtins/hmr-runtime.js#L94-L96

Any pointers would help as I'm don't quite understand the parent-child hierarchy yet.

TimNZ:
My HMR in a large React project is broken.  hmrAccept gets stuck in a loop of calls until I hit maximum stack size error.  I can then do a browser page reload (Chrome doesn't seem to allow interrupting JS code that doesn't yield somehow), or kill the Helper process if I'm in a rush.

I can instrument hmr-runtime to spit out debug info on the chain to at least see what's happening.

Given the requests for other HMR runtime control, is it time to provide hooks into HMR as it is meant to be a debug/dev only thing anyway?

zepplock:
Any update?

kjellski:
I have exactly the same problem as @TimNZ

devongovett:

kyleslight:
> the reason it is `some` instead of `forEach` is so that it stops at the module that accepts the update (via `module.hot.accept()`) and doesn't traverse higher in the tree. however, you are right that this is slightly incorrect since it will fail to update sibling parents.
> 
> what we need to do instead is update both parents, and then traverse higher ONLY if neither accepted the update. you will probably need to break up `hmrAccept` a bit to get this working.

I'm facing the same problem today.

Still wondering how to implement `traverse higher ONLY if neither accepted the update` against `2.x`.

PR #207, opened 2017-12-11
[WIP] Logger improvements

DeMoorJasper:
* Change logger into a global accessible Singleton
* Improve message outputting to not overwrite and glitch
* Add the ability to make any message persistent
* Full control over the console


albinotonnina:
I see you found the solution for your logger issue in #157 👍

yeskunall:
+1 for converting `Logger` into a `Singleton` instance.

src/Bundler.js
```diff
@@ -98,7 +97,9 @@ class Bundler extends EventEmitter {
         }
       }
     } catch (err) {
-      this.logger.warn(err);
+      setTimeout(() => {
+        logger.warn(err);
+      }, 500);
```
brandon93s: Whats the 0.5 second delay for?

DeMoorJasper: So that the warning doesn't get removed when the builder actually starts and does logger.clear()

DeMoorJasper: Nvm this is no longer necessary in the latest version...

DeMoorJasper:
All conflicts should be fixed now @brandon93s

devongovett:
Thanks @DeMoorJasper! I tried this out locally. A couple things:

* I like the idea of making it a global singleton. The problem is that it isn't really a singleton since different instances will be accessed by each worker. This will cause the history to be overwritten by each worker. To fix this, I think we should use IPC to send messages from worker instances to the master, which will actually do the logging. Should be fairly easy.
* This seems to cause the logger to take over the whole terminal rather than write inline. e.g. if you had other history in your terminal prior to running `parcel`, it would be overwritten when parcel starts. I'd rather if it remained inline so that we only ever overwrite other messages from Parcel itself and not other commands.

DeMoorJasper:
@devongovett i removed the singleton pattern, just wondering how u see implementing the IPC, as far as i've thought about it i think it would overwrite messages by parcel or cause the glitches i tried to fix in the first place with keeping an array with all messages (overwriting status every time and having full control over the output on clear and rewrites)

devongovett:
Sorry I wasn't clear. I DID like the singleton pattern you had, we just need to add something so that the instances in workers don't overwrite each other.

Maybe something like this could work? The child would intercept writes to the logger and send them to the master.

```javascript
write(message) {
  if (process.send) {
    // we're in a worker
    process.send({type: 'log', message});
  } else {
    // do actual write here
  }
}

process.on('message', msg => {
  if (msg.type === 'log') {
    Logger.write(msg.message);
  }
});
```

src/Logger.js
```diff
@@ -54,60 +104,60 @@ class Logger {
...
 
-  status(emoji, message, color = 'gray') {
-    if (this.logLevel < 3) {
-      return;
-    }
-
-    let hasStatusLine = this.statusLine != null;
-    if (!hasStatusLine) {
-      this.statusLine = this.lines;
+  handleMessage(options) {
```
shawwn: `handleMessage` could be simplified as:

```js
handleMessage(options) {
  this[options.method](...options.args);
}
```

src/Logger.js
```diff
@@ -54,60 +104,60 @@ class Logger {
...
+let loggerInstance;
+function getLogger() {
+  if (!loggerInstance) {
+    loggerInstance = new Logger();
   }
+  return loggerInstance;
 }
 
-module.exports = Logger;
+module.exports = getLogger();
```
shawwn: We should probably export Logger rather than an instance, so that it's possible to inherit from `Logger` or extend it with mixins.

DeMoorJasper: I'm not sure if this is really necessary but i exported both in new improvements

src/Logger.js
DeMoorJasper: Thanks for this, changed it

DeMoorJasper:
Closing this in favor of #462 once i figure out how to make the statuses not glitch i'll just open up a new PR.

pmurias:
A problem here on xterm + bash + Ubuntu.
I have to write to pipe to cat to avoid this horrible error message maiming.

Issue #211, opened 2017-12-11
🐛 Cannot resolve dependency when I use <a> tag with virtual path


garrydzeng:
🐛 bug report

### 🎛 Configuration (.babelrc, package.json, cli command)

package.json
```
{
  "name": "parcel-test",
  "version": "0.1.0",
  "devDependencies": {
    "parcel-bundler": "git://github.com/parcel-bundler/parcel.git"
  }
}
```
index.html
```
<html>
  <body>
    <a href="/b">CLICK ME</a>
  </body>
</html>
```
And run command:
```
parcel build src/index.html
```

### 🤔 Expected Behavior

No error

### 😯 Current Behavior

Thrown an error with message `Cannot resolve dependency './..\..\..\..\b'`

### 🌍 Your Environment

| Software         | Version(s)
| ---------------- | ----------
| Parcel           |  master branch
| Node             |  8.9.3
| npm/Yarn         |  Yarn 1.2.1
| Operating System |  Windows 10 x64

DeMoorJasper:
Will look into this

MIFind:
I have a bug as same as u. waiting for the resolved.

davidnagli:
Is this similar to #191?

DeMoorJasper:
Nah #191 had to do with something else, i'm not sure why it accured and guess it got fixed through a PR somewhere along the way

nakoo:
```
<a href="/b">CLICK ME</a>
```
😡

```
<a href="./b">CLICK ME</a>
```
😙

garrydzeng:
@nakoo I tried but still error.

PR #212, opened 2017-12-11
Fix virtual paths

DeMoorJasper:
Fix for #211 
* Only change link attribute of existing files
* Remove deprecated fs.exists from ./utils/fs.js


src/Asset.js
```diff
@@ -73,6 +73,11 @@ class Asset {
     let resolved = path
       .resolve(path.dirname(from), url)
       .replace(/[\?#].*$/, '');
+
+    if (!fs.existsSync(resolved)) {
```
devongovett: mmm, rather not be doing sync file io here. also, this doesn't take into account that the resolver may do additional steps to find the path (e.g. checking file extensions, aliases, etc.). finally, it will prevent errors from occurring in legitimate cases where the file should really be there.

let's think about how to solve this in a different way.

DeMoorJasper:
Closing up untill someone or i finds a better fix

PR #216, opened 2017-12-11
Detect asset ext by regex

gongpeione:
In order to add extensions by regex

### DIFF
--- a/src/Parser.js
+++ b/src/Parser.js
@@ -2,6 +2,9 @@ const path = require('path');
 const RawAsset = require('./assets/RawAsset');
 const GlobAsset = require('./assets/GlobAsset');
 const glob = require('glob');
+const toType = require('./utils/toType');
+
+const isRegexp = /^\/.+?\/[gimuy]+?/;
 
 class Parser {
   constructor(options = {}) {
@@ -33,11 +36,32 @@ class Parser {
   }
 
   registerExtension(ext, parser) {
-    if (!ext.startsWith('.')) {
+    const type = toType(ext);
+
+    if (type === 'regexp' || isRegexp.test(ext)) {
+      ext = ext.toString();
+    } else if (!ext.startsWith('.')) {
       ext = '.' + ext;
     }
 
     this.extensions[ext] = parser;
+  }
+
+  // find extension key
+  findExt(ext) {
+    const keys = Object.keys(this.extensions);
+
+    for (let i = keys.length - 1; i >= 0; i--) {
+      const key = keys[i];
+
+      if (isRegexp.test(key) && new RegExp(key, 'i').test(ext)) {
+        return key;
+      } else if (ext === key) {
+        key;
+        return key;
+      }
+      isRegexp.lastIndex = 0;
+    }
   }
 
   findParser(filename) {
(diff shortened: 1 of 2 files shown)


src/utils/toType.js
```diff
@@ -0,0 +1,8 @@
+module.exports = function(obj) {
+  const toString = Object.prototype.toString;
+  const type = toString
+    .call(obj)
+    .replace(/[\[\]]/g, '')
+    .split(/\s/)[1];
+  return type.toLowerCase();
+};
```
brandon93s: A bit more simple alternative:

```js
const toString = Object.prototype.toString;

module.exports = function(obj) {
     const type = toString.call(obj).slice(8, -1);
     return type.toLowerCase();
}
```

src/Parser.js
```diff
@@ -33,19 +36,41 @@ class Parser {
   }
 
   registerExtension(ext, parser) {
-    if (!ext.startsWith('.')) {
+    const type = toType(ext);
+
+    if (type === 'regexp' || isRegexp.test(ext)) {
```
brandon93s: Is the type check alone not sufficient?

src/Parser.js
```diff
@@ -33,19 +36,41 @@ class Parser {
...
+  findExt(ext) {
+    const keys = Object.keys(this.extensions);
+
+    for (let i = keys.length - 1; i >= 0; i--) {
+      const key = keys[i];
+
+      if (isRegexp.test(key) && new RegExp(key, 'i').test(ext)) {
+        return key;
+      } else if (ext === key) {
+        key;
```
brandon93s: `key;`  <- intentional?

src/Parser.js
```diff
@@ -33,19 +36,41 @@ class Parser {
...
     this.extensions[ext] = parser;
   }
 
+  // find extension key
+  findExt(ext) {
+    const keys = Object.keys(this.extensions);
+
+    for (let i = keys.length - 1; i >= 0; i--) {
+      const key = keys[i];
+
```
brandon93s: Might do the simple check first `ext === key` since it's the cheaper and more common operation.

src/utils/toType.js
gongpeione: lol yea my way is too complex, thank u~

src/Parser.js
gongpeione: sorry forgot to remove test code 😔  I will fixed these issuses asap

src/Parser.js
gongpeione: Do you mean the expression cannot tell if ext is a regex or not? Could you give me some example?

brandon93s: I meant are both the `type === 'regexp' and `isRegexp.test(ext)` checks necessary. Seems they're both checking if it's a regex.

```js
if (type === 'regexp') {
     return key
} else if (ext === key) {
// ...
```

gongpeione: Oh I see, I just wanna deal with this situation:  
```javascript
bundler.addAssetType('/ext/', require.resolve('./MyAsset'));
```
In some cases type name may generated by a function:
```javascript
() => `/${['jpg', 'jpeg'].map(type => `(${type})`).join('|')}/`
```
So they donot need to cover string to RegExp.

devongovett:
Why is this needed? Why not just register each extension? What is the usecase?

gongpeione:
Oh I just want to offer another option, I think this makes plugin registration a little bit easier. For example if you wanna process handlerbar files, you donnot need to regisiter `hbs` and `handlebars`, just `this.registerExtension(/(handlebars)|(hbs)/, './assets/HandlebarsAsset');`.

PR #223, opened 2017-12-11
Load assets in meta tags

### LINKED ISSUE #220: 🙋 Support image meta tags
# Feature Request

### 🤔 Expected Behavior

If I include a `<meta name="og:image" content="./cover.jpg">`, in an `index.html`, i'd expect that the source path gets properly updated after a successful build. 

```
<meta name="og:image" content="./cover.jpg" />
```

would become

```
<meta name="og:image" content="./123414523412341234.jpg" />
```

### 😯 Current Behavior

meta tags with references to assets are not updated

### 🌍 Your Environment

| Parcel           |  1.1.0
| Node             | 8
| npm/Yarn         | 1.3 
| Operating System | macOs

DeMoorJasper:
close #220


src/assets/HTMLAsset.js
```diff
@@ -23,9 +23,12 @@ const ATTRS = {
     'embed'
   ],
   href: ['link', 'a'],
-  poster: ['video']
+  poster: ['video'],
+  content: ['meta']
 };
 
+const META_ASSETS = ['og:image'];
```
brandon93s: Based on [The Open Graph protocol](http://ogp.me/) we should include:

- 'og:url'
- 'og:image'
- 'og:image:url'
- 'og:image:secure_url'
- 'og:audio'
- 'og:audio:url'
- 'og:audio:secure_url'
- 'og:video'
- 'og:video:url'
- 'og:video:secure_url'

brandon93s:
Instead of checking for meta tags with specific names, what if we loop through every attribute when we see a meta tag and check them all with `isUrl`? The list of *know* names we'd have to check for is pretty large, not to mention the *unknown*.

DanielRuf:
And what about twitter cards and other solutions?

DeMoorJasper:
isUrl would not be an option, because it doesn't see local files as a url, this check is to avoid trying to bundle public resources, i'll look into a better solution didn't know there were that many meta tags containing ressources

brandon93s:
>  this check is to avoid trying to bundle public resources

The current `isFile` implementation could return true for full URIs. Is this intended?

image

DanielRuf:
Probably not. But these are also valid files for possible meta tags (at least pictures). A host check + absolute / relative check might make sense, similar like `path.resolve()`.

DeMoorJasper:
@brandon93s It wouldn't make a lot of difference to improve the regex to not include public ressources, cuz isUrl is checked right after it anyway

src/assets/HTMLAsset.js
```diff
@@ -46,6 +48,9 @@ class HTMLAsset extends Asset {
         for (let attr in node.attrs) {
           let elements = ATTRS[attr];
           if (elements && elements.includes(node.tag)) {
+            if (node.tag === 'meta' && !isFile(node.attrs[attr])) {
```
devongovett: what does the `isFile` check fix that `isUrl` doesn't already check for?

DeMoorJasper: It can also be a local file with isFile, meta tags can contain a lot more than just locations so i added a check to make sure it doesn't try to include other meta content

devongovett: Seems like this would break for anything that has a `.` in it: e.g. `<meta name="something" content="a.b.c">`. I can't think of a good workaround. We may need to maintain a list of known meta tag names...

DeMoorJasper:
Closing this off as it is not a decent fix for this

PR #226, opened 2017-12-11
Asset hashing tests

### LINKED ISSUE #66: Config files do not invalidate cache
Create a new Parcel application with CSS modules and start it via `parcel index.html` but DO NOT create a `.postcssrc` config file yet.

Quit the Parcel server then create a `.postcssrc` file with `{ "modules": true }` as the content.

Start the Parcel server again.

Note that CSS modules do *not* work correctly at this point.

In order to make CSS modules work now you must `rm -rf .cache` and restart the Parcel server.

Example repo:

https://gist.github.com/wavebeem/c3afffcb4d757bf9d771fe1f0a9cf7e4

DeMoorJasper:
Tests the asset hashes


brandon93s:
Looks like this won't cover `postcss` config changes. Adding `asset.hashAddon = config;` in the [transform](https://github.com/parcel-bundler/parcel/blob/master/src/transforms/postcss.js#L12) may do the trick.

DeMoorJasper:
Thanks @brandon93s forgot about that one, i'll add a fix.
Will also add some tests to test if everything works

DeMoorJasper:
Just realised the asset hash is not used by caching in any way, so pivoted this to just testing the functionality

DeMoorJasper:
Closing this unless anyone would find this usefull

PR #234, opened 2017-12-12
feat: change order of hmr accept to start from children

maciej-ka:
This MR will change the order of HMR acceptance.

Say, we have a files:

[module id:1] index.js
```js
var func = require('./func').func;
var value = require('./value').value;
output(value + func());
```

[module id:3] func.js
```js
var value = require('./value').value;
exports.func = () => value;
```

[module id:5] value.js
```js
exports.value = 1;
```

In current master, when I change value.js to export `2` and hot reload this change, the order of applying changes will be modules 5, 1, 3 and the history of output calls will be: 

[2, 3, 4]

In this PR, the order of applying changes will be 5, 3, 1, and output will be:

[2, 4, 4]

Additionally, this PR is adding a test for recent a HMR problem when module had two parents.


devongovett:
Thanks for the PR. I'll need to look into it a bit more. Is the output of `getParents` always just in reverse order? Seems like this would depend on the asset ids that are assigned, which aren't always deterministic.

maciej-ka:
I tried to mess things by changing order of imports in index.js:
```
 +var func = require('./func').func;
 +var value = require('./value').value;
```

And test was always [2, 4, 4]. However I don't know for sure that it isn't non-deterministic.

Alternative solution could be: for each module, we could calculate the longest possible distance to a root (think of it as an anti-Dijkstra algorithm) and use this to set an order in update phase.

Btw. wouldn't it be great for updates to be just [2, 4], two steps (not the current [2, 4, 4] or [2, 3, 4]).

devongovett:

PR #238, opened 2017-12-12
Add CLI option to enable FSWatcher polling

### LINKED ISSUE #221: [Feature] Automatically detect safe-write
### 🐛 Bug report 

While the parcel dev server or the watcher is running, changes in a file are sent to the browser correctly, but only the first time I save changes in the file. The second time, parcel doesn't rebuild (there is no console output) and the module doesn't reload in the browser. Other files will still reload, but also only once.

Even when refreshing in the browser, the changes aren't shown. I have to restart parcel for it to show the latest changes again.

### Reproduction

Here's a minimal sample that shows the problem for me: https://github.com/Edootjuh/parcelrepro

To reproduce: 
  * Start parcel with `parcel index.html`
  * Load `http://localhost:1234` in browser. The page shows `a b`
  * Edit module-a and save. The page changes. Edit it again. The page doesn't change.
  * Edit module-b and save. The page changes. etc.

The same happens with `parcel watch index.html`

### 🌍 Your Environment

| Parcel           |  Latest on master (commit 2d680c0)
| Node             | node v8.9.1
| npm/Yarn         | yarn 1.2.1
| Operating System | Ubuntu 17.10

DeMoorJasper:
As discussed here: closes #221 Some IDE's have safewrite which makes it hard for FSWatcher to detect the files changes, this CLI option should allow users to still use safewrite while not affecting performance for people who do not use safewrite.


devongovett:
this doesn't seem like something users should need to know about. it seems like an issue with the IDEs not with parcel. is there a way to detect the need for polling? should we just document how to fix your IDE like webpack does?

DeMoorJasper:
I couldn't find a way to detect this, probably best solution is to add it to the docs than

davidnagli:
@DeMoorJasper Can you please make an issue for this in the website repo

shawwn:
I think this is worth doing. It's nice to have a fallback mode for users that need it.

DeMoorJasper:
i'll close this off untill we find a way to detect this

PR #239, opened 2017-12-13
Add graphql support

tb:


src/assets/GraphqlAsset.js
```diff
@@ -0,0 +1,21 @@
+const Asset = require('../Asset');
+const gql = require('graphql-tag');
```
devongovett: this should be a local dependency in the project. you can use the localRequire util function to require relative to the project directory. see here for an example: https://github.com/parcel-bundler/parcel/blob/master/src/assets/SASSAsset.js#L10

test/javascript.js
```diff
@@ -248,4 +249,32 @@ describe('javascript', function() {
     assert.equal(typeof output, 'function');
     assert.equal(output(), 3);
   });
+
+  it('should support requiring graphql files', async function() {
```
devongovett: can you move this to a separate test file for graphql?

brandon93s:
@tb Looks like there is just a few feedback items. Have capacity to tackle them? Otherwise, looks ready to land as a newly supported asset type!

test/javascript.js
tb: Ok, I will move

src/assets/GraphqlAsset.js
tb: Ok, will use that

tb:
@brandon93s @devongovett PR is rebased, feedback applied

src/assets/GraphqlAsset.js
```diff
@@ -0,0 +1,22 @@
...
+const localRequire = require('../utils/localRequire');
+
+class GraphqlAsset extends Asset {
+  constructor(name, pkg, options) {
+    super(name, pkg, options);
+    this.type = 'js';
+  }
+
+  async parse(code) {
+    let gql = await localRequire('graphql-tag', this.name);
```
devongovett: Question: does this work if `graphql` and `graphql-tag` are not already installed? In that case `graphql-tag` would be autoinstalled by parcel, but not `graphql` which is a peer dependency. Would that cause it to break?

Perhaps we will need to extend the autoinstall mechanism to also install peer dependencies... Or I guess we could just `localRequire('graphql')` here as well, which would force an install.

brandon93s: @devongovett Ah, good catch. No, it won't work if they aren't already installed. Fortunately, the error message will be along the lines of "cannot find `graphql`". My preference is to enhance auto install to check and handle peer dependencies. 

Adding `localRequire('graphql')` would be an unclear side-effect in the code. If `graphql-tag` ever drops it as a peer dependency, we'd keep installing it for no reason.

devongovett: Ok, let's do that as a separate PR.

PR #185, opened 2017-12-10
Add `--open` option to automatically launch in default browser

brandon93s:
Add a new option `-o` `--open` to automatically launch the dev server in the default browser. 

Using [opn](https://www.npmjs.com/package/opn) to launch the browser, which is also used by [webpack-dev-server](https://github.com/webpack/webpack-dev-server/blob/0a8f8965fba4a46a862497f6e7d1ad9377040a15/bin/webpack-dev-server.js#L468), [browser-sync](https://github.com/BrowserSync/browser-sync/blob/3aced708d47f032905721d01dd435d809d8feb33/lib/utils.js#L175-L189), [create-react-app](https://github.com/facebookincubator/create-react-app/blob/2e82ebb3371731a5c4e346f310848ddb23fd0976/packages/react-dev-utils/openBrowser.js#L100), and a lot [more](https://www.npmjs.com/browse/depended/opn).

PR #240, opened 2017-12-13
Add "copy to clipboard" functionality to parcel-cli

yeskunall:
As per https://github.com/parcel-bundler/parcel/pull/185#issuecomment-350607109. There's just one problem, however:

screen shot 2017-12-12 at 7 24 01 pm

... yeah. The logger writes (almost) the same info. 😬

Thoughts?

### DIFF
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -4,6 +4,7 @@ require('v8-compile-cache');
 const chalk = require('chalk');
 const program = require('commander');
 const version = require('../package.json').version;
+const copy = require('clipboardy');
 
 program.version(version);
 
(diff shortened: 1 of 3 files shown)


devongovett:
@yeskunall cool! maybe rather than having two messages in the terminal we could just add to the existing "Server running at http://localhost:1234" message?

yeskunall:
@devongovett I would have already done that, but the second message comes from `Server.js`, and I see no easy way of adding it to that message.

The only way I can think of is explicitly passing the `command` param from here:

https://github.com/parcel-bundler/parcel/blob/b7b2991d69b960d9f2951828b8145a6d9396ee4e/bin/cli.js#L94

... into the server object created in `Server#serve`. 😓

Then print the appropriate message according to what the value of `command.open` is.

ykpythemind:
Whenever I run parcel command, my clipboard is overwritten?? 
I don't think it's a good idea 🤔

DeMoorJasper:
This should probably be an option, because it might get annoying for people who have anything in their clipboard.
But i guess this is kinda the same as what --open was aiming for, to make going to that location easier?

yeskunall:
I feel like we're adding way too many options. It's all about saving a few key strokes. So I say we either opt to keep it as it is, or we don't introduce this change at all.

@DeMoorJasper yeah, it is. But I would rather keep this over `--open`. I would never use that option, but I _would_ like this one. This is obviously open for discussion.

DeMoorJasper:
Imo this is more of a choice between --open or --copy for the user, so it wouldn't add any option it would just supply a choice

brandon93s:
With the availability of `--open` and the desire not to fiddle with the users clipboard, we're going to hold off on this for now.

PR #249, opened 2017-12-13
[WIP] cli.js show helpful message

chocnut:
Instead of showing these error message.
screen shot 2017-12-13 at 8 41 22 pm
screen shot 2017-12-13 at 8 41 35 pm

Show them a helpful message.
screen shot 2017-12-13 at 8 49 35 pm
screen shot 2017-12-13 at 8 49 51 pm

### DIFF
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -1,5 +1,4 @@
 #!/usr/bin/env node
-
 require('v8-compile-cache');
 const chalk = require('chalk');
 const program = require('commander');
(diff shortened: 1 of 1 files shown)


bin/cli.js
```diff
@@ -73,13 +73,32 @@ program.on('--help', function() {
   console.log('');
 });
 
-// Make serve the default command
 var args = process.argv;
-if (!args[2] || !program.commands.some(c => c.name() === args[2])) {
+if (!args[2]) {
```
yeskunall: What happens if I just do `parcel build`? Won't this display a warning message specific to the `serve` command?

bin/cli.js
```diff
@@ -73,13 +73,32 @@ program.on('--help', function() {
...
+  // no entry point.
+  displayWarningMessage();
+} else if (args[2] && args[2] === 'serve') {
```
chocnut: @yeskunall good point. I think that checking should be here as well.

bin/cli.js
```diff
@@ -76,21 +75,21 @@ program.on('--help', function() {
 var args = process.argv;
 if (!args[2]) {
   // no entry point.
-  displayWarningMessage();
-} else if (args[2] && args[2] === 'serve') {
-  // serve but with no entry point.
-  displayWarningMessage('serve');
+  showInstructions();
+} else if (args[2] === 'serve' || args[2] === 'build') {
```
yeskunall: This is still not fully correct. `build` doesn't require you to provide an entry point.

Also, you need to remove the first `if` altogether. It makes no sense.

devongovett:
An entry point is actually not required: if not passed, the `main` field from `package.json` will be used instead.

Issue #40, opened 2017-12-06
🐛 `require is not defined`


benbonnet:
Is that just me ? [the simplest example](https://parceljs.org/getting_started.html) does not even work; breaking on the very first variable in ./build/index.js :
```
"use strict";
require = function (_require) { etc…
```
Or transforms/babelization/etc…  (not shown in the example) are to be added first ?

handoing:
I have the same problem.

deesx:
Just tried out this one : https://github.com/davidsonsns/parcel-react-app

Was on node6, upgrading to node8 let me run this example without issues.

node8 is obviously required; so is a minimal .babelrc setup

labnol:
Getting the same issue on Node 6

thejameskyle:
We're relying on non-strict mode. Some tool (most likely Babel is adding in the `"use strict"`.

We should probably update the [prelude.js](https://github.com/parcel-bundler/parcel/blob/ad5135adfda84e7bf49a9aee477deae7d83a3588/src/builtins/prelude.js) file to find the correct global to attach to.

@ljharb What's the best way to get `global` (whether it be `window/self/global/this`) today?

bigtunacan:
Just chiming in; I was getting the a different error #31, but after that issue was fixed I'm now getting this error with node6.

ljharb:
In strict or sloppy mode, you can always use `Function(‘return this’)()` in every engine except CSP-enabled browsers (however, for those, you can use `window || self`.

handoing:
I upgraded node v9.2.0 and solved it.

marcioj:
I was looking into the issue but I think even fixing the use strict problem we will have problems with babel transforming `require = (function (modules, cache, entry) { ... })` to `require = function (modules, cache, entry) { ... };` when the project is built. 
So in node 7 and further, when the script is compiled by parcel and the browser loads, we have correctly `require = (function (modules, cache, entry) { ... })([modules])` which is a iife. Because it uses the files under src/* which are not compiled it uses the original file. But in node 6 and before it generates `require = function (modules, cache, entry) { ... };([modules])`. Which makes the page render nothing.

Not sure how to fix that. Maybe we could make babel ignore this specific file and just copy it to lib?

Also is it a bad idea to just not support node 6? I think people using parcel are mostly testing it or creating new stuff, in this case they can at least use a more up to date node version.

bigtunacan:
I just ended up upgrading to Node 8x series since it is on LTS

[...]

PR #253, opened 2017-12-13
[WIP] Node 6 Support

ghusse:
I investigated on the Node 6 issue, and I found some interesting things. Unfortunately I did not fix all the issues (yet) but I think there are interesting things to know in this Pull Request.

Plus, if anyone has a clue about the origin of the regressions/issues left, I'm interested.

Things I noted:

1. Files in `/builtins` are processed by babel before being used. Which leads to strange behaviors, because `prelude.js` declares a global variable in strict environment.
2. Babel does not process `prelude.js` the same way in Node 6 and Node 8: an extra semi-column is added to the file in Node 6. Making it totally non-functional.

In this PR, I tried to fix these two problems by:

1. Make babel ignore files in `/builtins`, and use them only from `/src`
2. Change `prelude` to declare a function named `register`
3. Change `prelude` to write the result of this function call into `window.require` instead of using `require` without declaring it
4. Change the tests to create a global variable named `window` in the context of the vm
5. Change the way `JSPackager` combines different JS files, to create an IIFE around all the code and correctly call the function `register`
6. Call babel on test files (to be able to launch unit tests on Node 6)

I think all these remarks can be useful for people knowing more about the project than me (even if I learned A LOT about how parcel works in the process).

For the moment, more tests are passing on Node 6 than before my changes, but it seems that I introduced some regressions in Node 8. Any help is welcome.


package.json
```diff
@@ -63,10 +63,11 @@
   "scripts": {
     "test": "cross-env NODE_ENV=test mocha",
     "format": "prettier --write './{src,bin,test}/**/*.{js,json,md}'",
-    "build": "babel src -d lib",
+    "build": "babel src -d lib --ignore src/builtins/*",
```
DeMoorJasper: Could u add this to .babelrc instead?

ghusse: Sure, thanks

davidnagli:
Will this fix #40?

ghusse:
Yes. But for the moment I need to investigate on regressions introduced by these changes.

rondonjon:
@ghusse I really like that you took the time to enumerate and explain your changes. :+1:

ghusse:
I found another issue with the current code.

Previous modifications caused some tests to fail on both node 8 and node 6. I did not understand why when I first created this PR. But since, I investigated and the results might interest you.

1/ Errors are ignored when running tests

First of all, `Bundler.bundle` totally hide errors during the bundle function. Which rends the debugging process very difficult.

I made a modification in this method to rethrow errors in test mode (I don't know why it should not be the case in other modes, but I did not want to introduce other regressions).

2/ Throwing errors revealed issues in some async functions

Throwing errors revealed the following error:

```
keyword super unexpected here
```

This error message seems to be caused by babel, not being able to correctly manage calls to `super` in `async` methods.

One workaround consisted of changing babel's configuration, the other consisted of using `ParentClass.prototype.functionName.call(this, [args])` instead of `super`.

I changed calls to super in async methods, and tests are now passing!

Tests are passing on Node 8, only 1 test is failing in node 6!

ghusse:
I need help on this last issue: there is only 1 test failing on Node 6.

On node 8, both tests on `fs-destructuring` & `fs-destructuring-assignment` are passing.

On node 6, `fs-destructuring-assignment` is not passing, whereas `fs-destructuring` is almost similar, and is passing.

When debugging the code, it seems that, on passing tests, calls to `fs.readFileSync` are replaced by the actual content of the file.

But on node 6 + `fs-destructuring-assignment`, the call is not replaced (but code in `fs-destructuring` is correctly replaced "hello"). This issue in replacement causes the test to fail.

Is there anyone that can point me out the piece of code I should take a look at?

Thanks

ghusse:
It seems that parcel uses babel to parse the JS Files, in order to replace calls to `readFileSync` with the actual file's content.

Babel does not parse files the same way with Node 6 and Node 8. Is there someone with more knowledge of the current parsing code, I'm interested by his/her insights on this subject.

ghusse:
@devongovett? It seems that you wrote most of `src/visitor/fs`. Any clue?

ghusse:
Up, I'd definitively need help on this last point. Is there someone knowing how Babel works internally and how it is used to replace calls to fs.readFileSync?

devongovett:
> Babel does not parse files the same way with Node 6 and Node 8.

In what way? Like the resulting AST is different?

ghusse:
Exactly, AST are different in both cases. I don't have the details right now, but i'll copy/paste them as soon as possible.

ghusse:
On node8, in `visitors\fs.js`, the function `AssignmentExpression` is called twice:

I called console.log on `path.node`, here are the outputs in the console
```
// NODE 8
// first call with path.node =
Node {
  type: 'AssignmentExpression',
  start: 20,
  end: 50,
  loc:
   SourceLocation {
     start: Position { line: 2, column: 1 },
     end: Position { line: 2, column: 31 } },
  operator: '=',
  left:
   Node {
     type: 'ObjectPattern',
     start: 20,
     end: 34,
     loc: SourceLocation { start: [Object], end: [Object] },
     properties: [ [Object] ] },
  right:
   Node {
     type: 'CallExpression',
     start: 37,
     end: 50,
     loc: SourceLocation { start: [Object], end: [Object] },
     callee:
      Node {
        type: 'Identifier',
        start: 37,
        end: 44,
        loc: [Object],
        name: 'require' },
     arguments: [ [Object] ] },
  extra: { parenthesized: true, parenStart: 19 } }

// Second call on
Node {
  type: 'AssignmentExpression',
  start: 54,
  end: 116,
  loc:
   SourceLocation {
     start: Position { line: 3, column: 0 },
     end: Position { line: 3, column: 62 } },
  operator: '=',
  left:
   Node {
     type: 'MemberExpression',
     start: 54,
     end: 68,
     loc: SourceLocation { start: [Object], end: [Object] },
     object:
      Node {
        type: 'Identifier',
        start: 54,
        end: 60,
        loc: [Object],
        name: 'module' },
     property:
      Node {
        type: 'Identifier',
        start: 61,
        end: 68,
        loc: [Object],
        name: 'exports' },
     computed: false },
  right:
   Node {
     type: 'CallExpression',
     start: 71,
     end: 116,
     loc: SourceLocation { start: [Object], end: [Object] },
     callee:
      Node {
        type: 'Identifier',
        start: 71,
        end: 83,
        loc: [Object],
        name: 'readFileSync' },
     arguments: [ [Object], [Object] ] } }
```

On node 6, same test, same console.log

```
// NODE 6
// First call
{ type: 'AssignmentExpression',
  operator: '=',
  left:
   Node {
     type: 'Identifier',
     start: 21,
     end: 33,
     loc:
      SourceLocation {
        start: [Object],
        end: [Object],
        identifierName: 'readFileSync' },
     name: 'readFileSync',
     __clone: [Function: __clone] },
  right:
   { type: 'MemberExpression',
     object: { type: 'Identifier', name: '_require' },
     property:
      Node {
        type: 'Identifier',
        start: 21,
        end: 33,
        loc: [Object],
        name: 'readFileSync' },
     computed: false } }

// Second call
Node {
  type: 'AssignmentExpression',
  start: 54,
  end: 116,
  loc:
   SourceLocation {
     start: Position { line: 3, column: 0 },
     end: Position { line: 3, column: 62 } },
  operator: '=',
  left:
   Node {
     type: 'MemberExpression',
     start: 54,
     end: 68,
     loc: SourceLocation { start: [Object], end: [Object] },
     object:
      Node {
        type: 'Identifier',
        start: 54,
        end: 60,
        loc: [Object],
        name: 'module' },
     property:
      Node {
        type: 'Identifier',
        start: 61,
        end: 68,
        loc: [Object],
        name: 'exports' },
     computed: false },
  right:
   Node {
     type: 'CallExpression',
     start: 71,
     end: 116,
     loc: SourceLocation { start: [Object], end: [Object] },
     callee:
      Node {
        type: 'Identifier',
        start: 71,
        end: 83,
        loc: [Object],
        name: 'readFileSync' },
     arguments: [ [Object], [Object] ] } }
```

As you can see, the method `AssignmentExpression` is called twice in both cases, but on node 8 and node 6, input parameters are different.

<table>
<tr>
<th>Node 6 </th>
<th> Node 8</th>
</tr>
<tr><td><pre>{ type: 'AssignmentExpression',
  operator: '=',
  left:
   Node {
     type: 'Identifier',
     start: 21,
     end: 33,
     loc:
      SourceLocation {
        start: [Object],
        end: [Object],
        identifierName: 'readFileSync' },
     name: 'readFileSync',
     __clone: [Function: __clone] },
  right:
   { type: 'MemberExpression',
     object: { type: 'Identifier', name: '_require' },
     property:
      Node {
        type: 'Identifier',
        start: 21,
        end: 33,
        loc: [Object],
        name: 'readFileSync' },
     computed: false } }</pre></td>
<td><pre>Node {
  type: 'AssignmentExpression',
  start: 20,
  end: 50,
  loc:
   SourceLocation {
     start: Position { line: 2, column: 1 },
     end: Position { line: 2, column: 31 } },
  operator: '=',
  left:
   Node {
     type: 'ObjectPattern',
     start: 20,
     end: 34,
     loc: SourceLocation { start: [Object], end: [Object] },
     properties: [ [Object] ] },
  right:
   Node {
     type: 'CallExpression',
     start: 37,
     end: 50,
     loc: SourceLocation { start: [Object], end: [Object] },
     callee:
      Node {
        type: 'Identifier',
        start: 37,
        end: 44,
        loc: [Object],
        name: 'require' },
     arguments: [ [Object] ] },
  extra: { parenthesized: true, parenStart: 19 } }</pre></td></tr></table>

ghusse:
@devongovett any idea?

PR #256, opened 2017-12-14
Enable JSON5-flavored .babelrc

yamitzky:
Enable JSON5 .babelrc, including trailing comma and comments.

It is same that babel-core is handling json5.
https://github.com/babel/babel/blob/a89171910fa2e5b026ec87aa5bcfc1e0aa3aefa2/packages/babel-core/src/config/loading/files/configuration.js#L195


devongovett:
Thanks!

Issue #255, opened 2017-12-14
🙋 Environment Variables


rj254:
# 🙋 Feature request

Define process.env variables within the bundle (similar to Webpack's EnvironmentPlugin, which relies on DefinePlugin).

### 🤔 Expected Behavior

Add a plugin that reads environment variables and injects them into the bundle such that code can rely on `process.env.VAR` in browser. 

### 😯 Current Behavior

There doesn't seem to be a simple way to define environment variables, this would add a simple way to add environment variables.

### 💁 Possible Solution

This plugin could use dotenv so that in order to define environment variables you can define a `.env` or `.env.development` or `.env.production` in the root of the project. This would make it pretty straight forward to use environment variables; however, it does raise the question of how to pass environment variables into a project without relying on dotenv (host environment variables).

### 🔦 Context

An internal library relies on a variable being set in process.env - with Webpack, we use the EnvironmentPlugin to add it in. 

### 🌍 Your Environment

| Software         | Version(s)
| ---------------- | ----------
| Parcel           |  1.2.0
| Node             | ^8
| npm/Yarn         | Yarn 1.3.2
| Operating System | macOS

devongovett:
You can just set environment variables in your shell when building. e.g.

```
SOME_VAR=test parcel index.html
```

and `process.env.SOME_VAR` will automatically be replaced with the value.

bogdansoare:
It will be nice if we could have a similar way to set custom environment variables like in Create react app https://github.com/facebookincubator/create-react-app/blob/master/packages/react-scripts/template/README.md#adding-custom-environment-variables

kolya-ay:
May be just [babel-plugin-transform-inline-environment-variables](https://github.com/babel/minify/tree/master/packages/babel-plugin-transform-inline-environment-variables) to not bload the code base?

Siyfion:
@devongovett that works, but as the original post commented, it'd really be good to have production/development env vars picked up automatically based on whether it's a `build` or not. I use this feature heavily in CRA and find it invaluable.

Having said that, I think this project looks _incredibly_ promising, so keep up the good work!

devongovett:
moving to #258. See additional discussion there.

PR #258, opened 2017-12-14
Add basic .env support.

gongpeione:
#255


src/assets/JSAsset.js
```diff
@@ -24,6 +28,26 @@ class JSAsset extends Asset {
     this.isAstDirty = false;
     this.isES6Module = false;
     this.outputCode = null;
+
+    // decode .env
+    this.envTable = {};
+    const envRegex = /^(\w+)\s*=\s*([\s\S]+)/;
+    const envFile =
+      process.env.NODE_ENV === 'development' ? ['.dev.env'] : ['.env'];
```
yeskunall: This is kind of tricky. `NODE_ENV` is something that was popularized by Express. I know most of us would probably end up using `NODE_ENV`, but it's really not required to do so. If parcel is going to look for that _exact_ prop, then can you also add it in the documentation somewhere?

gongpeione: Yes I agree. And I am not sure if it is a good idea seperate `.env` and `.dev.env` this way.

yeskunall: @gongpeione as long as it is documented, I think it's okay. 😄

cksac: How to handle env other than dev and prod? e.g. dev/uat/prod.

devongovett:
I'm on board with this idea (loading environment variables from a file). Can we use this module to actually load and parse the files though? https://github.com/motdotla/dotenv We can do that once on bundler startup, rather than for each asset. It will set the environment variables in the node.js environment, and the existing environment variable substitution in parcel should just work after that. I'd  recommend doing this in the `Bundler` class immediately after startup.

Paths that we should load should follow this pattern: `.env.${process.env.NODE_ENV}`. We should load these paths: https://github.com/bkeepers/dotenv#what-other-env-files-can-i-use. That's what  create-react-app does as well: https://github.com/facebookincubator/create-react-app/blob/master/packages/react-scripts/config/env.js

devongovett:
Thanks for your work on this. 👏

I refactored it slightly to move the code to a util file, and use the config resolution system to find .env files (searching upward from main file until they are found). Also added a test.

cloud-walker:
this has not been published yet, right?

DeMoorJasper:
@cloud-walker It's already in the master branch but not yet on npm

juliocanares:
waiting for this 🔥

andywer:
As far as I can tell there is still no support for environment vars when using TypeScript, right? 😕

A distinct parcel-plugin-env might be helpful after all...

luizfilipe:
This is already on npm? I'm also trying to read something regarding this on the docs, but there's no information about dotenv on it.

Can anyone help me here? I've place `.env` file at the same folder as entry file in this case `index.html` I'm using babel and I'm also using React for front-end I've set inside of `.env`  file the environment variable that stores an URL like this `API_URL=http://localhost:8081` and I'm doing a `console.log` on the script (`<script src="src/index.js"></script>`) that I'm calling on `index.html`.
But I'm still getting `undefined`.

Here is the js file looks like

```
'use strict';

import React from 'react';
import {render} from 'react-dom';
import {BrowserRouter as Router} from 'react-router-dom';
import {App} from './app'

console.log(process.env.API_URL);

render((
  <Router>
    <App/>
  </Router>
), document.getElementsByName('body'));

if (module.hot) {
  module.hot.accept();
}
```

FDiskas:
@luizfilipe place your `.env` file in root directory where `package.json` is located

brotzky:
Not working for me as well.

```
"parcel-bundler": "^1.7.0",
```

`.env` in root with variables `VAR=something`

```
console.log(process.env) // {}
```

leriel:
For anyone else that may come across this wondering - do not rely on `process.env` object to have values inside. Instead, rely directly on `process.env.KEY` as those are going to be replaced directly in the code, example:

`.env`
```
KEY=val
```
`index.js`
```
console.log(process.env.KEY); // val
console.log(process.env); // {}
```

frankieshakes:
> Can anyone help me here? I've place `.env` file at the same folder as entry file in this case `index.html` I'm using babel and I'm also using React for front-end I've set inside of `.env` file the environment variable that stores an URL like this `API_URL=http://localhost:8081` and I'm doing a `console.log` on the script (`<script src="src/index.js"></script>`) that I'm calling on `index.html`.
> But I'm still getting `undefined`.

I was experiencing the same thing. Restarting the Parcel server didn't seem to help. Running this command did the trick for me:

`npx parcel --no-cache src/index.html`

Hopefully this helps out the next person.

risanto:
> npx parcel --no-cache src/index.html

This works wonder. Thanks a lot!

PR #269, opened 2017-12-14
added missing parse-json dep to package.json

philipodev:
as title says, added the missing parse-json dependency.

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,6 +28,7 @@
     "mkdirp": "^0.5.1",
     "node-libs-browser": "^2.0.0",
     "opn": "^5.1.0",
+    "parse-json": "^4.0.0",
     "physical-cpu-count": "^2.0.0",
     "postcss": "^6.0.10",
     "postcss-value-parser": "^3.3.0",


DeMoorJasper:
Shouldn't the current json-parse requires simply be replaced by JSON5?

DeMoorJasper:
Here: https://github.com/parcel-bundler/parcel/blob/548eef92e11711db007b7613ba5530de508d21a0/src/FSCache.js#L6
Here: https://github.com/parcel-bundler/parcel/blob/0eb7a930ffcd4fc77b5b6c75e490299f92ca8a8e/test/hmr.js#L9

philipodev:
@DeMoorJasper im not sure, i just saw that master wasn't working and instead threw "Unhandled promise rejection (rejection id: 1): Error: Cannot find module 'parse-json'". But when `npm install parse-json` it worked as expected

DeMoorJasper:
I know i just mean replacing all parse-json requires with json5 might make json parsing more consistent across the codebase and wouldn't require an additional package.
Doesn't matter that much

philipodev:
@DeMoorJasper oh i see. If that's possible it would be better I agree.

DeMoorJasper:
u can have a look at #256 to see how json5 works

philipodev:
thanks @DeMoorJasper for pointing this out. i will now close this PR in favour of https://github.com/parcel-bundler/parcel/pull/272 where i changed the remaining `parse-json` requires to `json5`.

PR #270, opened 2017-12-14
Root file shouldn't be hashed.

ssuman:
If we have 
index.html
   -other.html

Other.html
  -index.html

In other.html index.html is being replaced with <hash>.html. This shouldn't happen.

### DIFF
--- a/src/Asset.js
+++ b/src/Asset.js
@@ -77,16 +77,9 @@ class Asset {
       './' + path.relative(path.dirname(this.name), resolved),
       Object.assign({dynamic: true}, opts)
     );
-    let asset = this.options.parser.getAsset(
-      resolved,
-      this.package,
-      this.options
-    );
-
-    if (this.isMainFile(resolved)) {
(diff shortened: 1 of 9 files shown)


ssuman:
All tests passed locally. But one test failed here. Any idea ?

DeMoorJasper:
That test just fails from time to time

src/Asset.js
```diff
@@ -77,9 +77,16 @@ class Asset {
...
-    return this.options.parser
-      .getAsset(resolved, this.package, this.options)
-      .generateBundleName();
+    let asset = this.options.parser.getAsset(
+      resolved,
+      this.package,
+      this.options
+    );
+
+    if (this.isMainFile(resolved)) {
```
devongovett: Can you do this check inside `generateBundleName` instead of here? Maybe get rid of the boolean parameter all together and replace with this check?

ssuman:
@devongovett - Changes done.

Issue #219, opened 2017-12-11
🐛 AppVeyor test HTML: should support bundling HTML fails


DeMoorJasper:
🐛 bug report
On AppVeyor i can see this test is lately failing a lot.
```shell
1) html should support bundling HTML:
      AssertionError [ERR_ASSERTION]: 3 == 2
      + expected - actual
      -3
      +2
      
      at assertBundleTree (test\utils.js:117:12)
      at Context.<anonymous> (test\html.js:9:5)
      at <anonymous>
      at process._tickCallback (internal/process/next_tick.js:188:7)
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
Command exited with code 1
```

### 🎛 Configuration (.babelrc, package.json, cli command)
AppVeyor config of this Repo

### 🤔 Expected Behavior
Should pass

### 😯 Current Behavior
Doesn't pass

pomle:
Hello @DeMoorJasper!
Why was this closed? 

I have this problem and seems to be an indeterministic test. 

Here's some output from ~10 runs of `while true; do yarn test -- --grep 'should support bundling HTML'; done`

About half fail.

https://gist.github.com/pomle/283c6f16d0b4e4e5e4947ad58fce1abb#file-parcel-indeterministic-html-test-txt


Also seen problems with `watcher should only re-package bundles that changed` which looks like a drifting timestamp issue.

```
1) watcher should only re-package bundles that changed:

      AssertionError [ERR_ASSERTION]: [ 1513331454, 1513331454 ] deepEqual [ 1513331454, 1513331455 ]
      + expected - actual

       [
         1513331454
      -  1513331454
      +  1513331455
       ]

      at Context.<anonymous> (test/watcher.js:122:12)
      at <anonymous>
```

DeMoorJasper:
This was a mistake indeed, (didn’t accure the day i closed it)

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

pomle:
I've looked into this but need more information.

In https://github.com/parcel-bundler/parcel/commit/fc8ce69e3735550018c0b915df69ff17092da31c#diff-08e28e98a118d512b13f4fc7423da77aR668 the HTML test is introduced, and the integration fixtures looks looks like
```
html
├── index.css
├── index.html
├── index.js
└── other.html
```

[...]

PR #271, opened 2017-12-14
Display absolute path on failed dependency resolve.

pomle:
When a file path can not be resolved the error message is cleaned
up but the relative path to file is displayed.

In cases where a mistake have been made this path can be very confusing,
like `Cannot resolve dependency './../../../../../../main.js'` for when
a path starts with `/`.

This PR resolves path into an absolute path in hope of
reducing confusion and minimize cognitive overhead of `../` chasing.

The above example would now display 
```
Cannot resolve dependency '/main.js' at '/main.js'
```
and for relative unresolvable path
```
Cannot resolve dependency './even-more/x.js' at '/mnt/c/Users/pom/Development/pom/parcel-implementation/src/script/more/even-more/x.js'
```

### DIFF
--- a/src/Bundler.js
+++ b/src/Bundler.js
@@ -273,7 +273,8 @@ class Bundler extends EventEmitter {
       return await this.resolveAsset(dep.name, asset.name);
     } catch (err) {
       if (err.message.indexOf(`Cannot find module '${dep.name}'`) === 0) {
-        err.message = `Cannot resolve dependency '${dep.name}'`;
+        const absPath = Path.resolve(dep.name);
+        err.message = `Cannot resolve dependency '${absPath}'`;
 
         // Generate a code frame where the dependency was used
         if (dep.loc) {
--- a/src/Bundler.js
+++ b/src/Bundler.js
@@ -273,7 +273,7 @@ class Bundler extends EventEmitter {
       return await this.resolveAsset(dep.name, asset.name);
     } catch (err) {
       if (err.message.indexOf(`Cannot find module '${dep.name}'`) === 0) {
-        const absPath = Path.resolve(dep.name);
+        const absPath = Path.resolve(Path.dirname(asset.name), dep.name);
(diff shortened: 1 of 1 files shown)


devongovett:
The path that it currently logs is exactly the path that is in the code of the parent module. What do people think about changing this? On one hand it is nice that the error message tells you what in the code is incorrect, but you're right that relative dependencies like that can be hard to parse.

pomle:
I thought this thru and it is possible this can be misleading due to
resolve function origin path may vary.

Let's put it on the back burner for now. I'll get back shortly.

pomle:
@devongovett I looked into it a bit more and revised it. Was obviously a bit more complicated to conclude something, but my changes so far include resolving absolute path relative to current asset. Before it was working dir. 

Not sure it is the right thing to do for all use cases. The error message print system also seems to suffer from a bug where the message is clipped when spanning multiple lines. The clipping bug is really the one that bit me,

Principally, I believe exposing absolute search paths is crucial to easy debugging, but rewriting the error message might not be the correct way.

Let me know if you think this is worth spending more time on right now and I'll be happy to proceed.

kingdaro:
I'm onboard with this change. Using absolute paths is also good for editors with hyperclick support in terminals and other places, where the path wouldn't have been resolvable the way it is now.

pomle:
Test is failing due to an indeterministic test issue:
https://github.com/parcel-bundler/parcel/issues/219

Perhaps this ticket should be re-opened? Test errors still persist on c008bb0ac492dbac38e2f13017e3143f40359934

devongovett:
Nice, best of both worlds. 👍

devongovett:
Modified slightly in https://github.com/parcel-bundler/parcel/commit/bd1d6e806522fcb539a9a50b3cb6bfdf46ea2888 to only include the absolute path if the dependency was specified as a relative one. That way you don't get incorrect absolute paths for npm dependencies that aren't installed (e.g. `require('foo')`).

PR #278, opened 2017-12-15
Add --host option to CLI

### LINKED ISSUE #276: 🙋 Allow to specify a hostname for hosted IDEs
I use [Cloud9](c9.io) which is a hosted IDE site. They do some host/port mapping stuff so they want me to use an environment variable whenever I start a server. I understand that I can pass the `$PORT` variable to the `-p` argument, but there doesn't appear to be an argument for the host, or ip address which I could pass the `$IP` variable to. 

### 💁 Possible Solution
`npx parcel-bundler index.html -p $PORT -h $IP`

brandon93s:
Add a new option `-h --host` to override the host / IP of the dev server.

Closes #276


bin/cli.js
```diff
@@ -10,7 +10,15 @@ program.version(version);
 program
   .command('serve [input]')
   .description('starts a development server')
-  .option('-p, --port <port>', 'set the port to serve on. defaults to 1234')
+  .option(
+    '-p, --port <port>',
+    'set the port to serve on. defaults to 1234',
+    parseInt
```
brandon93s: Added coercion to the port. Node handles it as-is, but it was resulting in a false error message on Windows:

image

bin/cli.js
```diff
@@ -94,9 +102,19 @@ async function bundle(main, command) {
...
-    const server = await bundler.serve(command.port || 1234);
+    const server = await bundler.serve(command.port, command.host);
+
     if (command.open) {
-      require('opn')(`http://localhost:${server.address().port}`);
+      const open = () => {
+        require('opn')(
+          `http://${command.host || 'localhost'}:${server.address().port}`
+        );
+      };
```
brandon93s: The below fixes a bug. Timing wasn't an issue when using `localhost` but `listening` appears to be slower (at least on Windows) when providing a host. This results in an attempted `server.address().port` call before `listening`. Simply defer if not `listening` yet.

bin/cli.js
```diff
@@ -94,9 +102,19 @@ async function bundle(main, command) {
   const bundler = new Bundler(main, command);
 
   if (command.name() === 'serve') {
-    const server = await bundler.serve(command.port || 1234);
```
brandon93s: Default port is already handled:

https://github.com/parcel-bundler/parcel/blob/333c3aa5d20f98a5f3c52635751032d12854c13c/src/Bundler.js#L461-L465

src/Bundler.js
```diff
@@ -458,9 +458,9 @@ class Bundler extends EventEmitter {
     return Server.middleware(this);
   }
 
-  async serve(port = 1234) {
-    this.bundle();
-    return await Server.serve(this, port);
+  async serve(port = 1234, host) {
+    await this.bundle();
+    return await Server.serve(this, port, host);
```
brandon93s: This was resulting in output like this:

image

Instead of like this:

image

Seems the `persistent` server logging should come at the end, and not be shown until everything is ready.

brandon93s: Caused by #164, build output was on a single line prior to that. The _build_ time will remain the same with this change, but there could be a minor delay added between end of build and dev server availability due to the blocking. Not a big deal IMO, it should be on the magnitude of milliseconds.

Alternatives:
- Revert this, leave the broken up logging
- Refactor logging of server messages until build is finished

bin/cli.js
rondonjon: You are setting the default host here, but not the default port ...

src/Bundler.js
rondonjon: ... whereas here you are setting the default port but not the default host. Can't the defaults be in the same place?

brandon93s: @rondonjon - good call! Updated to handle defaults with `commander`. This also gives us nicer help commands.

src/cli.js
```diff
@@ -10,14 +10,13 @@ program
   .description('starts a development server')
   .option(
     '-p, --port <port>',
-    'set the port to serve on. defaults to 1234',
-    parseInt
+    'set the port to serve on',
+    x => parseInt(x, 10),
```
brandon93s: I've seen this fail without the radix: 

image

devongovett:
hmm, not sure I understand the usecase here... why is this needed? doesn't the default (i.e. no host passed to the server) listen on all interfaces already?

brandon93s:
Good question. I (incorrectly) thought the default was `127.0.0.1` and not `0.0.0.0`. The orignal ask (#276) was in relation to cloud9, but after reviewing their [docs](https://docs.c9.io/docs/run-an-application) they're asking users to bind to `0.0.0.0` anyhow. With that in mind, it should be working as expected without these changes. In fact, these changes make the behavior _more_ restrictive by default.

superhawk610:
@devongovett @brandon93s the dev server itself works fine in its current form, but hot reloading does not work when viewing from hosts other than localhost (i.e. - when developing on a remote machine), as the WebSocket is hardcoded to look for localhost - see comment by @Fluidbyte on #276.

jarlef:
I'm currently working on a project that needs this feature. We have a lot of sites running on subdomains including our local development sites. Cookies are often attached to the top domain (*.foo.com) and also CORS is also enabled for subdomains in the dev environment. We are basically running our local dev sites under a subdomain  e.g local.foo.com etc. So being able to set `-h local.foo.com -open' would be great to avoid manually changing the hostname in the address bar.

PR #285, opened 2017-12-15
Remove filename-unsafe but package-safe characters from bundle name

### LINKED ISSUE #283: 🐛 Package.json name with "/" throws error
Possibly a bug, don't know it's related to Parcel itself or what.

I am running the https://parceljs.org/getting_started.html example. And when running `parcel index.html`, it throws:

```sh
⏳  Building...
⏳  Building index.js...
events.js:182
      throw er; // Unhandled 'error' event
      ^

Error: ENOENT: no such file or directory, open '/Users/changyan/Documents/repos/test_parcel/dist/@changyan/test_parcel.js'
```

```js
{
  "your": { "config": "here" }
}
```

### 🌍 Your Environment

| Parcel           |  1.2.0
| Node             | 8.6.0
| npm/Yarn         | 
| Operating System | MacOS

chee:
closes #283 

Originally I was stripping off the namespace, and then hyphening everything else. Then I was using a module to strip out the filename unsafe characters.

but then, of course most filename-unsafe characters are already invalid package.json names, so now i'm only stripping out the remaining url/file unsafe characters which are still package.json safe. and that's only the `/` in the namespace and the `*` allowed in old package names.

so there

### DIFF
--- a/src/utils/sanitizeFilename.js
+++ b/src/utils/sanitizeFilename.js
@@ -1,5 +1,0 @@
-const sanitizeFilename = require('sanitize-filename');
-
-module.exports = function sanitize(name, replacement = '-') {
-  return sanitizeFilename(name, {replacement});
-};
--- a/src/utils/sanitizeFilename.js
+++ b/src/utils/sanitizeFilename.js
@@ -1,0 +1,5 @@
+const sanitizeFilename = require('sanitize-filename');
+
+module.exports = function sanitize(name, replacement = '-') {
+  return sanitizeFilename(name, {replacement});
+};
--- a/src/utils/sanitizePackageName.js
+++ b/src/utils/sanitizePackageName.js
@@ -1,0 +1,5 @@
+const unsafeRegex = /[/*]/g;
+
+module.exports = function sanitizePackageName(name, replacement = '-') {
+  return name.replace(unsafeRegex, replacement);
+};
--- a/src/utils/sanitizePackageName.js
+++ b/src/utils/sanitizePackageName.js
@@ -1,5 +1,0 @@
-const unsafeRegex = /[/*]/g;
-
-module.exports = function sanitizePackageName(name, replacement = '-') {
-  return name.replace(unsafeRegex, replacement);
-};
--- a/package.json
+++ b/package.json
@@ -33,6 +33,7 @@
     "postcss-value-parser": "^3.3.0",
     "posthtml": "^0.10.1",
     "resolve": "^1.4.0",
+    "sanitize-filename": "^1.6.1",
     "serve-static": "^1.12.4",
     "uglify-es": "^3.2.1",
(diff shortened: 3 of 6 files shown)


src/utils/sanitizePackageName.js
```diff
@@ -0,0 +1,5 @@
+const unsafeRegex = /[/*]/g;
```
brandon93s: Let's go ahead and replace _all_ invalid filename characters to avoid future issues:

`/, \, <, >, :, ", |, ?, *`

On windows, a filename also cannot end in `.`

src/utils/sanitizePackageName.js
```diff
@@ -0,0 +1,5 @@
...
+
+module.exports = function sanitizePackageName(name, replacement = '-') {
```
brandon93s: Could make this a more generic `sanitizeFileName` so that it can be used elsewhere if an issue presents itself.

src/utils/sanitizePackageName.js
chee: okay. if we want to create a more generic sanitizeFileName then we should use ‘sanitize-filename’ itself from npm, i think. as there are a bunch more illegal windows names.

i’ll do this in a couple hours time when i get home

chee: i prefer the the approach of trusting the package name to be a valid package name. I understand the want for a more generic approach, in case at some point you'd want to use this for things other than the package name.
But, if that was to happen, it would take no more work to include the login then that it would now, and trusting the package name isn't locking us in to anything that is complex or would mean a lot of rework should a future decision mean it needs changed.
But, i'm also fine with it and have done it and it's done now.

src/utils/sanitizeFilename.js
```diff
@@ -0,0 +1,5 @@
+const sanitizeFilename = require('sanitize-filename');
+
+module.exports = function sanitize(name, replacement = '-') {
+  return sanitizeFilename(unsafeRegex, {replacement});
```
brandon93s: Is this un-tested?

`unsafeRegex` is no longer defined.
`name` is not passed to `sanitizeFilename`.

Doesn't look like this code would work as-is. The helper is probably unnecessary now, too. Can just use `sanitize-filename` directly.

chee: oh shoot, it was unpushed ! i'd amended the commit, and had not forced push.

i thought it preferable to keep the indirection in case more was ever desired of sanitize filename than simple replacement (such as removing the namespace, or simply removing all non alpha-numerics). but fair enough. will do.

devongovett:
Would be great to add a test for this! Can you follow up with that @chee? Thanks!!

chee:
@devongovett sure thing !

chee:
@devongovett i'm not sure where this test should go !

PR #288, opened 2017-12-15
Fix dynamic imports exports

### LINKED ISSUE #112: Dynamic import doesn't work?
when i try code splitting, get a error:
```
  Uncaught Error: Cannot find module '7'
```
index.js:
```js
import('./m').then(fn => {
    console.log(fn(1,2));
});
```
m.js:
```js
export default (a, b) => a + b;
```
no more config...

| Parcel           |  1.0.3
| Node             |  9.2.0
| npm/Yarn         | npm
| Operating System | macOS 10.12.6

fathyb:
The current implementation of dynamic imports inside Parcel does not return module exports in the resulting promise, contrary to what is described in the  [Parcel documentation](https://parceljs.org/code_splitting.html) :

```js
// pages/about.js
export function render() {
  // Render the page
}
```
```js
import('./pages/about').then(function (page) {
  // page is undefined here
  // Render page
  page.render();
});
```

This PR implements dynamic imports as defined in [TC-39 Dynamic import Proposal](https://tc39.github.io/proposal-dynamic-import/).

## Notable changes

- It uses the `fetch` API, so I added a polyfill for that. I also updated the tests to be compatible with this changes by updating the test context and the bundled files. It can be replaced with XMLHTTPRequest.

- It also depends on `Blob` and `URL.createObjectURL`, we use it to inject a small function triggered where the main entry has been fully loaded. I assumed it was OK since we already depends on `Promise` and we were already depending on `document`. It can be replaced with Data URIs, but will use more memory and will be synchronous.

- In the case of multiple entries the last one will be used.

- This will change how Parcel handles defaults imports. [The spec defines that defaults exports are exported as named](https://github.com/tc39/proposal-dynamic-import/issues/21). Example :

    ```js
    const {default: foo, bar} = await import('module')
    ```
    is the async equivalent of :
    ```js
    import foo, {bar} from 'module'
    ```

It fixes #112 .

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,6 +19,7 @@
     "commander": "^2.11.0",
     "cssnano": "^3.10.0",
     "get-port": "^3.2.0",
+    "fetch-ponyfill": "^4.1.0",
     "glob": "^7.1.2",
     "htmlnano": "^0.1.6",
     "is-url": "^1.2.2",
--- a/src/builtins/bundle-loader.js
+++ b/src/builtins/bundle-loader.js
@@ -38,25 +38,76 @@ function loadBundle(bundle) {
   }
 }
 
+var jsBundleCallbacks = {};
+var jsBundleIdCounter = 0;
+
+// fetch polyfill for Node.js and the browser
+var localFetch = typeof fetch === 'undefined' ?
+  require('fetch-ponyfill')({Promise: Promise}).fetch :
+  fetch;
+
 function loadJSBundle(bundle) {
-  return new Promise(function (resolve, reject) {
-    var script = document.createElement('script');
-    script.async = true;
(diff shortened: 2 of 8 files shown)


steveblue:
This bugfix should allow Angular routes to be code split and lazy loaded.

devongovett:
I'm not sure I understand why this change is necessary? The dynamic imports should already work as described. There are tests for it too: https://github.com/parcel-bundler/parcel/blob/master/test/integration/dynamic/index.js

fathyb:
Oh, I was wrong to assume it was not implemented, turns out the problem is that when you use CSS in `index.html` dynamic imports do not work anymore. I'll search for the issue instead of reimplementing it.

Issue #298, opened 2017-12-15
🐛 d3 import: Could not statically evaluate fs call


MoritzStefaner:
Thanks for parcel! Looks great so far! 

I just hit one wall so far:
Importing d3.js (https://github.com/d3/d3) results in build error

```
node_modules/xmlhttprequest/lib/XMLHttpRequest.js: Could not statically evaluate fs call
    at evaluate (/usr/local/lib/node_modules/parcel-bundler/src/visitors/fs.js:164:11)
    at CallExpression.path.get.map.arg (/usr/local/lib/node_modules/parcel-bundler/src/visitors/fs.js:31:21)
    at Array.map (<anonymous>)
    at JSAsset.CallExpression (/usr/local/lib/node_modules/parcel-bundler/src/visitors/fs.js:31:10)
    at NodePath._call (/usr/local/lib/node_modules/parcel-bundler/node_modules/babel-traverse/lib/path/context.js:76:18)
    at NodePath.call (/usr/local/lib/node_modules/parcel-bundler/node_modules/babel-traverse/lib/path/context.js:48:17)
    at NodePath.visit (/usr/local/lib/node_modules/parcel-bundler/node_modules/babel-traverse/lib/path/context.js:105:12)
    at TraversalContext.visitQueue (/usr/local/lib/node_modules/parcel-bundler/node_modules/babel-traverse/lib/context.js:150:16)
    at TraversalContext.visitSingle (/usr/local/lib/node_modules/parcel-bundler/node_modules/babel-traverse/lib/context.js:108:19)
    at TraversalContext.visit (/usr/local/lib/node_modules/parcel-bundler/node_modules/babel-traverse/lib/context.js:192:19)
``` 

Any ideas? Thanks!

fathyb:
This is because d3.js declares it's node.js version in the `main` field and the browser version in the `module` (and other) field. I made a PR to make Parcel uses these field in priority, it should totally work with d3.

mbasso:
Hi,
I'm facing the same issue with `.js` and `.asm.js` files compiled with [emscripten](http://kripken.github.io/emscripten-site/) from C++. It seems that the compiled code contains something like this:

```js
function(a,b){var c=R(a);c||(hb||(hb=require("fs")),ib||(ib=require("path")),a=ib.normalize(a),c=hb.readFileSync(a));return b?c:c.toString()};
```

that, unfortunately, is in contrast with the doc:

[...]

PR #299, opened 2017-12-15
Resolve "module", "jsnext:main" and "browser" before "main"

fathyb:
This is needed for a lot of libraries who ship their Node.js version in the `main` field (see #298). This will also use the ES6 modules version of some libraries (instead of UMD) and might lead to a better bundling/sharing of dependencies.


devongovett:
Would be good to add a test for this! Would you like to follow up with that @fathyb? Thanks!

fathyb:
I'll make another PR for the tests later today. I also found a bug a few minutes ago where some libraries define an object inside of module, I'll make another to prevent that 😅

PR #302, opened 2017-12-16
Parse --port string as an integer

mattdesl:
The `--port` command was being passed as a string, causing an equality check to fail since Node is returning it as an integer, and the `options.port` was a string. With this PR, the port is parsed as an integer, and the warning message goes away.

Before:
```sh
Server running at http://localhost:9966 - configured port 9966 could not be used
```

After:
```sh
Server running at http://localhost:9966 
```


brandon93s:
This is fixed by #278.

Issue #213, opened 2017-12-11
[RFC] New Asset support via Plugins or Core?


davidnagli:
# 💬 RFC 

Do you guys think support for new things should be added to the core, or via plugins?

Currently, we have a weird mix of both:
* `Svetle` support is via a plugin (thanks to @DeMoorJasper)
* `Stylus` is in the core
* `TypeScript` is in the core
* etc

What do you guys think?

To sum up what I heard in the slack:
* Perspective 1 - Keep the core lean, we should have a lean, minimal, and performant core. We should rely on plugins to integrate with all “non-standard” file extensions (so things like `.jsx` vs `.js`)
* Perspective 2 - If we have everything through plugins, we aren’t living up to our mission of being 0 config. We end up being the same as Webpack, where users need to install a billion plugins
* @albinotonnina Thinks that we should use plugins, but provide an easy CLI tool for importing them

It’s probably going to end up being  a combination of all of these…

If so, we need to decide which files should be considered “core” and which “plugins”

davidnagli:
What are your guys’ thoughts?

dschnare:
And if asset types are indeed going to be part of core, a way to list them. I tried looking for TypeScript support and I didn't know it was in core until reading this issue.

albinotonnina:
To elaborate more on that, I imagine something along the lines of:

- on

 Project Init, somewhere after my init command:

![choose](https://raw.githubusercontent.com/SBoudrias/Inquirer.js/HEAD/assets/screenshots/checkbox-prompt.png)

- Later on
`$ parcel supports jpg svg`

I would like this sort of granular control

EduardoRFS:
"The core", don't need to be the "parcel-bundler" package, perhaps the default setup contains a "lot" of plugins, like stylus, ts, jsx, and exists a "clean" package without everything, perhaps even the default css, html can be removed.

But for the tool `supports` a default option to setup vue and react fully support it's a cool feature

artyomtrityak:
what about Flowtype? Pracel uses Babel preset-env, can babel-preset-flow be added too?

davidnagli:
@artyomtrityak This doesn’t really have to do with Parcel plugins… Parcel supports `.babelrc`, so you can add any plugins you want. 😃

thejameskyle:
I think that assets should probably be implemented exactly the same as plugins even if they are built-in. That way they people can use them as references

brandon93s:
[...]

PR #306, opened 2017-12-16
Automagically install missing dependancies

DeMoorJasper:
Attempt to achieve what's been discussed here: #213


devongovett:
Oh this is clever.

src/utils/installPackage.js
```diff
@@ -0,0 +1,51 @@
+const {spawn} = require('child_process');
+const config = require('./config');
+
+module.exports = async function(dir, name) {
+  let yarn;
+  try {
+    yarn = await config.resolve(dir, ['yarn.lock']);
```
devongovett: maybe instead of trying to detect yarn based on yarn.lock we should just look for the command?

src/utils/installPackage.js
```diff
@@ -0,0 +1,51 @@
...
+  } catch (e) {}
+
+  return new Promise((resolve, reject) => {
+    let options = {
+      cwd: dir
+    };
+
+    let install = spawn('yarn', ['add', name, '--dev'], options);
```
devongovett: should they be dev dependencies or normal dependencies? I could see arguments for both.

src/utils/localRequire.js
```diff
@@ -1,16 +1,29 @@
...
   let resolved = cache.get(key);
   if (!resolved) {
-    resolved = resolve.sync(name, {basedir});
+    try {
+      resolved = resolve.sync(name, {basedir});
+    } catch (e) {
+      if (e.code === 'MODULE_NOT_FOUND' && triedInstall === false) {
+        // TODO: Make this use logger
+        console.log(`INSTALLING ${name}...\n`);
+        await install(basedir, name);
```
devongovett: maybe we only want to do this for the top-level project, not for packages inside node_modules?

src/utils/installPackage.js
DeMoorJasper: I think because it's a bundler it should only happen at development not sure though

src/utils/localRequire.js
DeMoorJasper: Isn't this the responsibility of localRequire?

DeMoorJasper: Than wouldn't this break if a dependancy isn't use a base language?

src/utils/installPackage.js
DeMoorJasper: This sounds like a more solid approach, will change it

src/utils/localRequire.js
EduardoRFS: but some packages like request-promise don't install by default dependencies

DeMoorJasper: How do you mean @EduardoRFS ? This PR is just for auto installing missing parsers, through changing localRequire, ofcourse this can be used for way more things in the future

src/utils/installPackage.js
```diff
@@ -0,0 +1,63 @@
+const {spawn} = require('child_process');
+const config = require('./config');
+
+function testYarn(options) {
```
brandon93s: An alternative would be to check for a `yarn.lock` file in the `cwd`. It's probably a quicker check than spawning a whole process, and worst case we fall back to npm. This may also be desirable for the scenario that the user _has_ yarn but the project is using _npm_.

DeMoorJasper: Checking for a yarn file in comparison with spawning an instance is an insane speed difference i'll pivot back towards my initial setup, thank you for your input.

brandon93s: Ah, didn't realize you'd started with that setup!

DeMoorJasper:
This just automatically installed dependencies not plugins, at least not yet

davidnagli:
Oh wow that makes it even cooler:)

DeMoorJasper:
It's not a wip, it's just a function that can be used for installing packages instead of throwing "Module not found" Atm if a plugin is not installed it will just not parse the file the correct way, even with this merged.
For this i'm also waiting on what we eventually decide on.
@davidnagli

fritx:
Why use yarn instead of npm?

fritx:
Ok, would pick up yarn if there is a [yarn.lock](https://github.com/eddyerburgh/parcel-vuejs-template/blob/master/template/yarn.lock) even it's been deprecated or regardless

Issue #280, opened 2017-12-15
🙋 how to keep  html name


songlipeng2003:
```
<!DOCTYPE html>
<html>
<head>
	<title>home page</title>
  <script src="./scripts/index.js"></script>
</head>
<body>
	<img src="./images/logo.png">

	<a href="page.html">Link</a>
</body>
</html>
```
this page will compile  to

```
<!DOCTYPE html>
<html>
<head>
	<title>home page</title>
  <script src="/dist/c4f6ff1a9fe4f3634b9272ba1ddaaf50.js"></script>
</head>
<body>
	<img src="/dist/98d8f0da23c4c8c6fe7083e3f62c1fdf.png">

	<a href="/dist/2f84526286378f5180fc0841e2f1a1d9.html">Link</a>
</body>
</html>
```

I want to keep html name, my expectation：

```
<!DOCTYPE html>
<html>
<head>
	<title>home page</title>
  <script src="/dist/c4f6ff1a9fe4f3634b9272ba1ddaaf50.js"></script>
</head>
<body>
	<img src="/dist/98d8f0da23c4c8c6fe7083e3f62c1fdf.png">

	<a href="/dist/page.html">Link</a>
</body>
</html>
```

Please tell me how to config ?

| Software         | Version(s)
| ---------------- | ----------
| Parcel           |   1.2.0
| Node             |  1.2.1
| npm/Yarn         | 
| Operating System |  macos

philipodev:
this is somewhat related to https://github.com/parcel-bundler/parcel/pull/270

DeMoorJasper:
This shouldn't be a hard fix, related code:
https://github.com/parcel-bundler/parcel/blob/c008bb0ac492dbac38e2f13017e3143f40359934/src/Asset.js#L146

ssuman:
Should I include this in my change ?

DeMoorJasper:
@ssuman if you want, go for it :)

ssuman:
Sure will include this in my PR.

Munter:
In assetgraph-builder we also hash filenames, but exclude any files matching the following from renaming:

- Any graph entry point (usually html)
- Any asset linked to with an `<a href>`
- Any asset linked to with a `<meta http-equiv="refresh">`
- Serviceworkers (must keep consistent file name across builds)
- `humans.txt`, `robots.txt`, '.htaccess`
- Cache manifests, rss and atom feeds
- `favicon.ico`

[...]

PR #307, opened 2017-12-16
[WIP] Keep file names

ssuman:
This commit when applied should allow builds to keep their original name in their build outputs.
#280

### DIFF
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -23,6 +23,7 @@ program
   .option('--no-hmr', 'disable hot module replacement')
   .option('--no-cache', 'disable the filesystem cache')
   .option('-V, --version', 'output the version number')
+  .option('--keepFileName', 'disable output name modification')
   .action(bundle);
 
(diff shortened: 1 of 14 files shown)


devongovett:
This won't work for cases where there are multiple files with the same name, since parcel outputs a flat dist folder.

e.g. if I have `app/a/index.js` and `app/b/index.js`, `dist/index.js` would be generated twice. That's why parcel generates the output filenames as a hash.

ssuman:
@devongovett Maintaining relative folder names should work right ? 

```
dist/app/a/index.js
dist/app/b/index.js
```

zeakd:
Are there any updates in this pr?

bricss:
Yes, please!

devongovett:
Should be solved by #1025 which retains filenames and folder structure for linked files, and content hashes static assets. Please help test using the master branch - a release will hopefully come next week!

PR #309, opened 2017-12-16
Deterministic Bundle Trees

brandon93s:
Potential fix for #219.

`Promise.all` guarantees order. This delays the `set` operations until after processing so that they always occur in the same order.

### DIFF
--- a/src/Bundler.js
+++ b/src/Bundler.js
@@ -323,21 +323,32 @@ class Bundler extends EventEmitter {
     }
 
     // Process asset dependencies
-    await Promise.all(
+    const applyFunctions = await Promise.all(
       dependencies.map(async dep => {
+        let applyFunction;
         let assetDep = await this.resolveDep(asset, dep);
         if (dep.includedInParent) {
           // This dependency is already included in the parent's generated output,
           // so no need to load it. We map the name back to the parent asset so
           // that changing it triggers a recompile of the parent.
-          this.loadedAssets.set(dep.name, asset);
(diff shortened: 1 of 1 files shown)


src/Bundler.js
```diff
@@ -323,22 +323,33 @@ class Bundler extends EventEmitter {
...
-          asset.depAssets.set(dep.name, assetDep);
+          applyFunction = () => {
+            asset.dependencies.set(dep.name, dep);
+            asset.depAssets.set(dep.name, assetDep);
+          };
           await this.loadAsset(assetDep);
         }
+        return applyFunction;
       })
     );
```
pomle: Is `applyFunctions` only job to be run later. Then would it be leaner to chain `.forEach(fn => fn())` do you think?

pomle:
👍

src/Bundler.js
brandon93s: That would work as well. I felt the current was more readable.

pomle: All good. Confirmed test is stable on my end now.

devongovett:
Thanks for looking into this! I refactored it slightly to be two loops: one that resolves and loads assets, and the second which assigns them back to deps. https://github.com/parcel-bundler/parcel/commit/539ade1820c3e644586e3161f6b2357a68784142

brandon93s:
Awesome, definitely a bit cleaner. Hopefully the last we see of this test failing!

Issue #34, opened 2017-12-06
WebAssembly Support


lastmjs:
It would be amazing to support WASM, so that C/C++/Rust and eventually many other languages could be automatically supported. Webpack is already working on this, and I've been working on it too here: https://github.com/lastmjs/zwitterion

I already have initial support for including basic C and C++ files, and WASM is now or will be ES module integratable.

It's possible and it would be awesome for this project to provide support.

lastmjs:
You know, support for different file types besides just js should probably be discussed. This might be a major infrastructure change and decision that this library might want to make now. We're already talking about supporting at least ts files here: https://github.com/parcel-bundler/parcel/issues/3

lastmjs:
And I am pleased to find out that this was already though of, and parcel is file-type agnostic. Perfect. Then, let's support C/C++/Rust/wasm files. How do we do this?

devongovett:
Yep totally want to support wasm. I did some initial experiments with this a while ago, and it should be pretty easy. I think WASM should definitely be supported in parcel core. Less sure about languages that compile to WASM like Rust since they usually depend on large toolchains, but perhaps it could work assuming installation instructions are documented well.

albizures:
I can take a look at the implementation of rust, but yes, I agree with @devongovett, this shouldn't come out of the box, maybe a plugin?

davidnagli:
@albizures I think you misread/miss-interpreted what Devon said. He said that it **should** be part of parcel core!

Parcel is trying to stay away from plugins, and trying to add all essential stuff directly to the core, to avoid the configuration mess of other tools like webpack.

albizures:
@davidnagli 👍, I already made some progress but I have some problems...

I already upload it and I created the PR  although it is not completed, maybe someone has an idea :)

PR: #312

devongovett:
Progress: just merged #312 into the `wasm` branch!

I think we should go for a few milestones here. 

1. First, let's get a `WASMAsset` merged that supports only precompiled `.wasm` files. This will be the base for other languages (e.g. rust) that compile to wasm.
2. Work on adding Rust support. Either autoinstall or warn if the proper toolchain isn't installed (rustc, `wasm32-unknown-unknown`, `wasm-gc`, etc.).
[...]

PR #312, opened 2017-12-16
[WIP] Support to rust/wasm

albizures:
Hi, this PR is not complete yet, currently, I'm struggling with two things:

- the first one is when there is a change in the entry rust file, the new wasm file is being generated but the file in the build folder doesn't change. I think it's something related to the cache.
- and the second one is related to the secondary rust files when there is a change in these files parcel doesn't detect the change. I think I have to do something like in  StylusAssets, use a custom evaluator, but I'm not sure.

Any ideas?

Thanks in advance

related: #34


devongovett:
Awesome start, really excited about this! I'll give some more feedback in a bit! 🎉

softprops:
Awesome to find some one is _already_ working on this. So fast!

src/assets/RustAsset.js
```diff
@@ -0,0 +1,93 @@
+const url = require('url');
+const path = require('path');
+const toml = require('toml');
+const {exec} = require('child-process-promise');
+
+const fs = require('../utils/fs');
+const JSAsset = require('./JSAsset');
+const localRequire = require('../utils/localRequire');
+
+const rustTarget = `wasm32-unknown-emscripten`;
```
devongovett: I think maybe we'll want to use wasm32-unknown-unknown, which doesn't require emscripten. Maybe checkout [rustify](https://github.com/browserify/rustify) which does something similar for browserify, and [wasm-experiments](https://github.com/killercup/wasm-experiments) to see how it all works.

src/assets/RustAsset.js
```diff
@@ -0,0 +1,93 @@
...
+    // Do nothing. Dependencies are collected by cargo :).
+  }
+
+  async parse(code) {
+    this.invalidateBundle();
+    this.invalidate();
+    this.dependencies.clear();
+    const {name, encoding, options} = this;
+    const release = process.env.NODE_ENV === 'production';
+    const cmd = `wargo build ${release ? ' --release' : ''}`;
```
devongovett: did you mean `cargo`?

src/assets/RustAsset.js
```diff
@@ -0,0 +1,93 @@
...
+const toml = require('toml');
+const {exec} = require('child-process-promise');
+
+const fs = require('../utils/fs');
+const JSAsset = require('./JSAsset');
+const localRequire = require('../utils/localRequire');
+
+const rustTarget = `wasm32-unknown-emscripten`;
+
+class RustAsset extends JSAsset {
```
devongovett: I think we should probably make a `WASMAsset` class to load precompiled `.wasm` files, and use it as a base class for this one. That way the runtime can be shared across multiple wasm languages.

src/assets/RustAsset.js
```diff
@@ -0,0 +1,93 @@
...
+  collectDependencies() {
+    // Do nothing. Dependencies are collected by cargo :).
+  }
+
+  async parse(code) {
+    this.invalidateBundle();
+    this.invalidate();
+    this.dependencies.clear();
```
devongovett: why is this necessary?

src/assets/RustAsset.js
```diff
@@ -0,0 +1,93 @@
...
+    );
+
+    const Module = {
+      // Path in the built project to the wasm file
+      wasmBinaryFile,
+      // Indicates that we are NOT running in node, despite 'require' being defined
+      ENVIRONMENT: 'WEB'
+    };
+
+    if (options.fromHtml) {
```
devongovett: Why is it different when loaded from HTML?

src/assets/RustAsset.js
albizures: oh my bad, this was me trying to fix the rebuild as mentioned in the description, I'll remove it, thanks

src/assets/RustAsset.js
albizures: the difference is that with this approach you can use rust as a standalone executable, without javascript.

I got the idea from this [article](https://hoverbear.org/2017/04/06/the-path-to-rust-on-the-web/)

src/assets/RustAsset.js
albizures: no, wargo is a library that setup your environment and makes easy to compile rust into wasm

src/assets/RustAsset.js
albizures: right now I'm using wargo and for now, wargo just uses wasm32-unknown-emscripten 

but sure, I'm going to take a look at rustify

src/assets/RustAsset.js
Phrohdoh: Does this PR verify that wargo is installed and accessible or fail violently?

albizures: right now it's failing violently, 😄  but I was working on an api more friendly to use wargo directly instead of using wargo through child_process. 
However, I was thinking about to stop using wargo since wargo is using `wasm32-unknown-emscripten` and I think `wasm32-unknown-unknown` is a better option...

shawwn:
Hey, I can set aside some time over the holidays to help finish the PR if you'd like. If you want a hand, ping me in parcel's Slack.

albizures:
Sure thing @shawwn thanks!

albizures:
Hey @devongovett, I made some changes

So, I got rid of wargo, and now I'm using `wasm32-unknown-unknown` directly also these changes fix my first point in the description 🎉 

what is next? 
- tackle my second point in the description
- validate the existence of `rustc`, `wasm32-unknown-unknown` target and create a useful alert
- use [wasm-gc](https://github.com/alexcrichton/wasm-gc), in production mode, if it exists

Do you have something else in mind?

also, I made a [demo](https://github.com/albizures/parcel-rust)

cc: @shawwn

devongovett:
🎉 Awesome work @albizures! Your list sounds good.

I think you can use `rustc` to determine the list of imported files: use the `--emit dep-info` flag and it will generate a file with all the deps in it that we could parse. These can be sent to parcel as dependencies with the `includedInParent: true` flag, which will cause parcel to watch those files but not try to parse them as they've already been compiled by rust.

About validating the existence of `rustc`, I wonder if we could just install it for you if you don't have it installed. Especially since we depend on the `nightly` toolchain, a specific backend (`wasm32-unknown-unknown`), and `wasm-gc`, we could just install those for you if they aren't already. We do this for other compilers like typescript, but it's slightly easier since those are in npm.

Also, we should probably support using `cargo build` rather than `rustc` directly, if we detect a `Cargo.toml` file. This will allow some more complex builds to happen. Not sure if we can detect dependencies in that case though.

Let's also start thinking about what we want the API for importing wasm modules to be. I think there should be a small runtime built into Parcel to load and instantiate wasm modules. Basically we should take care of the `WebAssembly.instantiate` call and the import should resolve to `instance.exports`.

There are two cases to think about.

Synchronous import, wasm is either inlined into JS or somehow loaded prior to the JS running:

```javascript
import {add} from './add.rs';

console.log(add(2, 3));
```

Asynchronous import, wasm is placed in a separate file and loaded on demand:

```javascript
let {add} = await import('./add.rs');

console.log(add(2, 3));
```

Sorry for the long comment. All of this definitely doesn't need to happen in this PR, just getting the ideas flowing! 😄

devongovett:
Going to merge this into the `wasm` branch on the main repo. I've added you as a collaborator @albizures so you should be able to work off that branch.

albizures:
oh great, thanks 🎉 

and I agree with you with all of your points.

I'll let you know any news

Issue #301, opened 2017-12-15
🙋 PWA support


Olian04:
_This is both a question and a :raising_hand_woman: feature request._

__Question:__ <br>
Since a PWA requires a set of resources to exists along side its `manifest.webmanifest` file. My question is, is there a way to include a file without having its name "scrambled"? This would apply to both the manifest file and its resources. _(Related to https://github.com/parcel-bundler/parcel/issues/280)_

__Feature request:__ <br>
It would be nice to have parcel look through the manifest file for files to include in the build.
For example (partial files):
```html
<!-- index.html -->
<link rel="manifest" href="./manifest.webmanifest">
``` 
```js
// manifest.webmanifest
"icons": [{
      "src": "pwa/icons/48.png",
      "sizes": "48x48",
      "type": "image/png"
    }, {
      "src": "pwa/icons/72.png",
      "sizes": "72x72",
      "type": "image/png"
    }
]
```
Should result in both `manifest.webmanifest` as well as `pwa/icons/48.png` and `pwa/icons/72.png` being included in the build.

Olian04:
The name "scrambling" is a problem when working with service workers as well.
```js
if ('serviceWorker' in navigator) {
    navigator.serviceWorker.register('./service-worker.js')
        .then(function(registration) {
            console.log('Registration successful, scope is:', registration.scope);
        })
        .catch(function(error) {
            console.log('Service worker registration failed, error:', error);
        });
}
```

kennetpostigo:
Given the control and information parcel has about the app it's working on, might be the perfect opportunity for parcel to generate a service worker for us.

Olian04:
@kennetpostigo When you say `generate a service worker` what part of using a service worker do you suggest parcel should generate? IMO a service worker is a script like any other, so its contents couldn't/shouldn't be generate. And the registration of a service worker is a js event (as seen in my previous comment) witch will have to be "subscribable" by the programmer, so this couldn't/shouldn't be generate either. So what part exactly are you suggesting parcel to generate?

kennetpostigo:
@Olian04 well seeing as it knows the assets a site uses, it could cache assets for us automagically.

Olian04:
@kennetpostigo Yes that is true, however caching responses is only one part of a service worker. And since we can't register multiple service workers for the same site the caching should probably be left to the programmer.

kennetpostigo:
[...]

PR #313, opened 2017-12-17
[WIP - Feedback required] Webmanifest support for PWA

ssuman:
PWA requires a set of resources to exists along side its manifest.webmanifest.

#301


mischnic:
I guess this can be closed because #691 got merged?

PR #322, opened 2017-12-17
Normalize filenames to nfc for hfs+

disjukr:
https://github.com/webpack/webpack/pull/1227#issuecomment-120780108


devongovett:
is this only a problem in GlobAsset or elsewhere as well?

PR #324, opened 2017-12-17
Removed package-lock from .gitignore

davidnagli:
According to the [npm documentation](https://docs.npmjs.com/files/package-lock.json):

> This file is intended to be committed into source repositories

&nbsp;
We already do this for `yarn.lock`. It is not in `.gitignore` and is currently checked in to our repo.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,5 @@
 node_modules/
 .DS_Store
-package-lock.json
 test/dist
 test/input
 coverage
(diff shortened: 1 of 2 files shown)


yeskunall:
I think I will quote what Sindre says in an answer to a question he was asked:

> The package-lock.json files in dependencies are ignored, so the lockfile only applies when users run npm install in the package repo. If you use a lockfile for packages, your local dependency tree will not match the dependency tree of users having your package as a dependency. This can potentially cause problems if one of your package dependencies breaks your package in a patch release. The lockfile will prevent you from seeing the problem locally, but it will affect your users.

[Ref](https://github.com/sindresorhus/ama/issues/479#issuecomment-310661514).

Issue #327, opened 2017-12-18
Error when scss file is empty


gongpeione:
**Choose one:** is this a 🐛 bug report or 🙋 feature request?
🐛 bug report

### 🤔 Expected Behavior
Built successful

### 😯 Current Behavior
Build failed, `No input specified: provide a file name a source string to process`

### 💁 Possible Solution
pass a `space` when .scss file is empty

### 🔦 Context
I know scss file is not empty normally, but I think maybe it should not show an error message when it is empty. And I can create a pr if you want.

nnnikolay:
I just took to play around with the latest version and I end up with not working HRM, apparently. When scss has been changed the process either stops or changes are not visible unless I refresh the page. It happens also if I make some big changes, for example, refactoring (incl. npm modules installation) from simple js file to react based 'hello world'. Complete re-run `npm start` helps though.

davidnagli:
@nnnikolay I’m not really understanding how that applies to this issue...

@gongpeione Wow great catch! This definetly needs to get fixed. Can you please make a PR?

nnnikolay:
@davidnagli sorry if it's not, I just started with an empty file as well, and then almost every file creation, module installation etc... leads me to the Error, I thought it might be connected.

brandon93s:
Is there a valid use case for an empty sass file, or are we solving for an edge case here?

davidnagli:
@brandon93s Looks like an edge case

rondonjon:
@brandon93s it is definitely a bug. An importer should be able to handle empty files unless empty files are invalid by the file type specification.

it may be an edge case in daily development, but considering the trouble that it causes and the little effort that it takes to solve it, I would definitely take the change. If this helps 50 people each save 30 minutes of bug search, it's worth it.

brandon93s:
I don't disagree, I just think this should be solved to handle all empty assets. The proposed solution fixed SASS only. I support empty asset handling at a higher level.

gongpeione:
Yes, I agree empty files should be handling at a higher level.

devongovett:
Closing. Should be fixed by #340.

PR #340, opened 2017-12-19
Fix: Error when scss file is empty #327

gongpeione:


brandon93s:
Would like to understand why this is needed ([issue](https://github.com/parcel-bundler/parcel/issues/327#issuecomment-352614344)). This fixes things currently, but there is no guarantee that an empty string fixes it in future versions of `node-sass`. Ideally, we'd skip `node-sass` if this is truly needed.

yeskunall:
@brandon93s while there is no use-case, it shouldn't break `parcel`'s bundling process. Similar issue has been brought up at webpack-contrib/sass-loader#403, and I believe this check should be in [src/Asset.js#parse](https://github.com/parcel-bundler/parcel/blob/master/src/Asset.js#L93) instead.

All assets extend from the base class, so we can avoid duplication by adding the check (or something along the lines of that) added by @gongpeione there.

devongovett:
What is the error you're getting with an empty file?

yeskunall:
Looks like the specific error message is

> No input specified: provide a file name a source string to process

... taken from the issue.

gongpeione:
Yes like @yeskunall said, the error message is `No input specified: provide a file name a source string to process`, cause node-sass treat empty file like null
```javascript
if (options.data) {
    binding.render(options);
} else if (options.file) {
    binding.renderFile(options);
} else {
    cb({status: 3, message: 'No input specified: provide a file name or a source string to process' });
}
```

src/Asset.js
```diff
@@ -50,7 +50,7 @@ class Asset {
   async getDependencies() {
     await this.loadIfNeeded();
 
-    if (this.mightHaveDependencies()) {
+    if (this.mightHaveDependencies() && this.contents) {
```
yeskunall: `mightHaveDependencies` already checks for `this.contents`, so this additional check is not required.

src/Asset.js
```diff
@@ -82,7 +82,11 @@ class Asset {
       .generateBundleName();
   }
 
-  mightHaveDependencies() {
+  async mightHaveDependencies() {
+    await this.loadIfNeeded();
+    if (!this.contents) {
```
yeskunall: If `this.contents` is passed along a "falsy" value, such as an empty string, then `mightHaveDependencies` will return `false`, which might cause unintended behaviour.

src/Asset.js
gongpeione: Some Asset class like CSSAsset override `getDependencies` method or some others may override `getDependencies`, this is wanna double check if contents needs to be parsed, trying not to modifiy other Asset class.

src/Asset.js
gongpeione: I am a little confused, this is indeed wants to skip `parse()` if `this.contents` is an empty string.

src/Asset.js
devongovett: swap the order of these conditions so `mightHaveDependencies` doesn't run if `this.contents` is empty.

src/Asset.js
```diff
@@ -83,6 +83,9 @@ class Asset {
   }
 
   mightHaveDependencies() {
+    if (!this.contents) {
+      return false;
+    }
```
devongovett: remove this check as it's already done above

PR #342, opened 2017-12-19
Add Reason asset type

rrdelaney:
This adds the [Reason](https://reasonml.github.io) asset type to `.re` files.

Right now this is pretty basic and only loads the recommended Reason configuration. I can update it to use other configurations if you want, but overall I think this is a good starting point.

The code for this is pretty simple -- a lot of the work was already done in the Reason commuity's wrapper for the build tool, [`bsb-js`](https://github.com/reasonml-community/bs-loader/tree/master/packages/bsb-js). That should handle requiring the build tool per-project and error handling already, so little logic other than "run the build" and "read the result" are needed.


davidnagli:
Nice!

src/Parser.js
```diff
@@ -12,6 +12,7 @@ class Parser {
     this.registerExtension('es6', './assets/JSAsset');
     this.registerExtension('jsm', './assets/JSAsset');
     this.registerExtension('mjs', './assets/JSAsset');
+    this.registerExtension('re', './assets/ReasonAsset');
```
brandon93s: What about `.ml` files?

src/assets/ReasonAsset.js
```diff
@@ -0,0 +1,32 @@
+const bsb = require('bsb-js');
+const fs = require('fs');
+const JSAsset = require('./JSAsset');
+const config = require('../utils/config');
+const localRequire = require('../utils/localRequire');
```
brandon93s: Appear to be unused:
```js
const config = require('../utils/config');
```

src/assets/ReasonAsset.js
```diff
@@ -0,0 +1,32 @@
...
+
+class ReasonAsset extends JSAsset {
+  async parse(code) {
+    // This runs BuckleScript - the Reason to JS compiler.
+    // Other Asset types use `localRequire` but the `bsb-js` package already
+    // does that internally. This should also take care of error handling in
+    // the Reason compilation process.
+    await bsb.runBuild();
+
+    this.contents = await new Promise((resolve, reject) => {
```
brandon93s: Could use `../utils/promisify` on `fs.readFile` instead of `new Promise`.

brandon93s:
See [parcel-plugin-bucklescript](https://github.com/jihchi/parcel-plugin-bucklescript) ([npm](https://www.npmjs.com/package/parcel-plugin-bucklescript)) for an existing plugin. Similar implementation, but may provide a bit more on the config front.

davidnagli:
I just tried it out and I’m consistently getting the error:
```
🚨  script: bsb: No such file or directory
```

And yes, I do have a file called hello.re:
```
> ls
    dist     hello.re
```


Full output:
```
> parcelfork hello.re
    ⏳  Building...
    🚨  script: bsb: No such file or directory

^C
```
`parcelfork` is just a little alias I set up to run the version of Parcel that I have forked locally (in this case switched to your PR HEAD). It’s essentially just calling `bin/cli.js`

If you’re reproducing this, here’s the exact code I use (taken from [here](https://reasonml.github.io/try/?reason)):

hello.re
```
type tree = Leaf | Node(int, tree, tree);

let rec sum =
  fun
  | Leaf => 0
  | Node(value, left, right) => value + sum(left) + sum(right);

let myTree =
  Node(
    1,
    Node(2, Node(4, Leaf, Leaf), Node(6, Leaf, Leaf)),
    Node(3, Node(5, Leaf, Leaf), Node(7, Leaf, Leaf))
  );

sum(myTree) |> Js.log;
```

src/Parser.js
rrdelaney: Added for OCaml files too 😄

src/assets/ReasonAsset.js
rrdelaney: Yep, they were. Thanks!

rrdelaney:
@brandon93s Thanks for the link! It looks like the implementation they used adapted from [`bs-loader`](https://github.com/reasonml-community/bs-loader), a Webpack loader I wrote for BuckleScript. bs-loader has a few problems and complexities, so we're actively trying to move away from it. By only supporting a small set of BuckleScript configurations the code can be simpler, faster, and more predictable.

@davidnagli You'll need to run the following commands in your project to get set up with Reason:

```
$ npm install --dev bs-platform
$ touch bsconfig.json
$ touch src/index.re
```

 and then copy this into `bsconfig.json`:

```json
{
  "name": "reason-parcel-example",
  "sources": [
    "src"
  ],
  "refmt": 3,
  "package-specs": {
    "module": "commonjs",
    "in-source": true
  },
  "suffix": ".bs.js"
}
```

rrdelaney:
Uhh, I don't think that last one should have failed. The HTML test failed for some reason?

src/assets/ReasonAsset.js
```diff
@@ -0,0 +1,28 @@
...
+    // does that internally. This should also take care of error handling in
+    // the Reason compilation process.
+    if (process.env.NODE_ENV !== 'test') {
+      await bsb.runBuild();
+    }
+
+    // This is a simplified use-case for Reason - it only loads the recommended
+    // BuckleScript configuration to simplify the file processing.
+    const outputFile = this.name.replace(/\.(re|ml)$/, '.bs.js');
+    const outputContent = await readFile(outputFile);
```
devongovett: is there no way to do this without writing to a temporary file first?

package.json
```diff
@@ -14,6 +14,7 @@
     "babylon": "^6.17.4",
     "babylon-walk": "^1.0.2",
     "browser-resolve": "^1.11.2",
+    "bsb-js": "^1.0.1",
```
devongovett: Let's not take this on as a dependency - people can install it locally in their projects. can you move it to a dev dep for the tests and use `localRequire` below?

src/assets/ReasonAsset.js
```diff
@@ -0,0 +1,28 @@
...
+const promisify = require('../utils/promisify');
+const readFile = promisify(fs.readFile);
+
+class ReasonAsset extends JSAsset {
+  async parse(code) {
+    // This runs BuckleScript - the Reason to JS compiler.
+    // Other Asset types use `localRequire` but the `bsb-js` package already
+    // does that internally. This should also take care of error handling in
+    // the Reason compilation process.
+    if (process.env.NODE_ENV !== 'test') {
```
devongovett: why not building in the tests?

test/integration/reason/local.bs.js
```diff
@@ -0,0 +1,11 @@
+// Generated by BUCKLESCRIPT VERSION 1.9.2, PLEASE EDIT WITH CARE
```
devongovett: doesn't seem like this should be committed...

brandon93s:
@rrdelaney -  That test experiences random failures; a fix for it is now in master. 

package.json
rrdelaney: bsb-js is a thin wrapper over finding and running a BuckleScript executable, kind of like the logic in the TS asset. I developed he package for use cases like this, adding Reason support easily to a new bundler. 

This doesn’t actually depend on BuckleScript, so users would still have to install that themselves. Requiring users to install it themselves wouldn’t make that much sense because it’s a package meant to be used by tooling. 

With this being the case, it is ok to add it as a dependency? If it helps at all, I’m the maintainer of that package.

src/assets/ReasonAsset.js
rrdelaney: Building during test would require adding BuckleScript as a dev dependency. BS is a big native binary that compiles itself on install, which takes about 7 minutes.

The wrapper around BuckleScript, bsb-js, is pretty well tested, so the tests for Reason only test the file-loading logic.

src/assets/ReasonAsset.js
rrdelaney: Sorry, what do you mean by writing to a temp file?

test/integration/reason/local.bs.js
rrdelaney: I hard coded the result of compiling the Reason file with BuckleScript so that we don’t have to run it when running the tests. BuckleScript is a binary that compiles itself during installation which could take a while. By hard coding the result we can test just the loading logic, because the logic to run BuckleScript is contained in another package, bsb-js.

src/assets/ReasonAsset.js
devongovett: I mean, it seems like `bsb-js` or whatever that uses is writing its output to a temp file, which we then need to read here. Is there a way to avoid that?

src/assets/ReasonAsset.js
devongovett: i see, ok.

src/assets/ReasonAsset.js
devongovett: e.g. the `compileFile` function as used by parcel-plugin-bucklescript seems to return the output directly to a variable

package.json
devongovett: After #306 lands this won't be such a big deal since deps like this will be automatically installed by parcel as needed.

src/assets/ReasonAsset.js
rrdelaney: compileFile basically does the same thing here — reading the result of the output. Currently BuckleScript has no way of compiling Reason like that without seriously compromised performance and maintainability.

rrdelaney:
@brandon93s Thanks 😄 I just pulled from upstream

package.json
rrdelaney: Even so, `bsb-js` is still a pretty small dependency to add. It doesn't include the entire compiler toolchain and has no dependencies itself. The package should only add [~2.7k](https://bundlephobia.com/result?p=bsb-js@1.0.1)

devongovett: It's a slippery slope I think. We don't do it for other asset types, so I don't want to start now.

devongovett:
I moved bsb-js to a dev dependency and used localRequire. That should get it autoinstalled when used once #306 lands.

PR #353, opened 2017-12-19
[WIP] Fix CSP issue while obtaining ref to global

### LINKED ISSUE #335: 🐛 CSP 'unsafe-eval' not allowed
In my packaged non-minified build, there are eval calls, which yield the following EvalError in my specific use case, where 'unsafe-eval' Content Security Policy is not allowed.

```
Uncaught EvalError: Refused to evaluate a string as JavaScript because 'unsafe-eval' is not an allowed source of script in the following Content Security Policy directive: "script-src 'self'
```

**Is there any way I can fix unsafe eval calls, e.g. using a custom packager?**

My deps:

```
  "dependencies": {
    "babel-polyfill": "^6.26.0",
    "mo-js": "^0.288.2",
    "react": "^16.2.0",
    "react-dom": "^16.2.0",
    "uiw": "^1.5.3",
    "uiw-iconfont": "^1.2.6"
  },
  "devDependencies": {
    "babel-plugin-transform-object-rest-spread": "^6.26.0",
    "babel-preset-env": "^1.6.1",
    "concurrently": "^3.5.1",
    "less": "^2.7.3",
    "node-sass": "^4.7.2",
    "parcel-bundler": "^1.2.0",
  }
```

.babelrc
```
{
  "presets": ["env", "react"],
  "plugins": ["transform-class-properties", "transform-object-rest-spread"]
}
```

For further context:

The application runs in an iframe with the following sandbox values:

```
sandbox="allow-forms allow-scripts allow-same-origin allow-popups allow-popups-to-escape-sandbox"
```

Thank you!

mBeierl:
Prevent violating Content Security Policy

fixes #335

### DIFF
--- a/src/visitors/globals.js
+++ b/src/visitors/globals.js
@@ -7,7 +7,7 @@ const VARS = {
     asset.addDependency('process');
     return 'var process = require("process");';
   },
-  global: () => 'var global = (1,eval)("this");',
+  global: () => 'var global = typeof global === "object" ? global : typeof window === "object" ? window :typeof self === "object" ? self : this;',
   __dirname: asset =>
     `var __dirname = ${JSON.stringify(Path.dirname(asset.name))};`,
   __filename: asset => `var __filename = ${JSON.stringify(asset.name)};`,


mBeierl:
It appears that this breaks the [globals integration-test](https://github.com/parcel-bundler/parcel/blob/master/test/integration/globals/index.js), as `global: !!global.document` evaluates to `global: false`. I don't really see why, tbh. 

I'd love to get a reviewer's eye on this, as the proposed changed in this PR did fix my [CSP 'unsafe-eval' issue](https://github.com/parcel-bundler/parcel/issues/335) and the packaged code still worked.

src/visitors/globals.js
```diff
@@ -7,7 +7,7 @@ const VARS = {
     asset.addDependency('process');
     return 'var process = require("process");';
   },
-  global: () => 'var global = (1,eval)("this");',
+  global: () => 'var global = typeof global === "object" ? global : typeof window === "object" ? window :typeof self === "object" ? self : this;',
```
brandon93s: Other implementations seem to default to `{}` instead of `this`. Trade-offs / concerns with either approach?

mBeierl: In global context `this` always refers to the global object. As we check before with `typeof global === "object"` whether the global object exists, it might be a better solution to use `{}`. I don't feel expert enough to give a definitive answer though.

CvX: A cosmetic nitpick: there's a space missing in `:typeof`.

CvX:
FWIW the tests pass on macOS (node 8.9.3 and 9.6.1) and Linux (node 9.6.1).

CvX:
@mBeierl You'll also have to replace `eval` here: [src/builtins/hmr-runtime.js:1](https://github.com/mBeierl/parcel/blob/6358a6f34a5a2acee044be1f98bc74e636ca3e9b/src/builtins/hmr-runtime.js#L1)

mrcoles:
I’m running into this issue trying to use parcel to build a Chrome extension and all Chrome extensions do not allow 'unsafe-eval'.

ygnoh:
I'm having the same problem with @mrcoles. I think this problem is very critical.

infoeon:
Can this be replaced with the following instead:

`global: () => 'var global = (function(){return this;}).call(null);',`

devongovett:
Closing in favor of #1133

Issue #334, opened 2017-12-18
🐛 Parcel does not minify built-ins (prelude.js)


bartw:
🐛 bug report

When I create a simple html and js like in the hello-world example and run "parcel build src/index.html" or "parcel build src/index.js", the html and js get bundled but only the html is minified, not the js.

### 🤔 Expected Behavior
I expect parcel to minify my js for a production build.

### 😯 Current Behavior
The js is not minified

### 🌍 Your Environment

| Software         | Version(s)
| ---------------- | ----------
| Parcel           |  1.2.0
| Node             | 9.3.0
| npm/Yarn         |  npm 5.6.0
| Operating System | Ubuntu 17.10 and Windows 10

### Github repo with my code

https://github.com/bartw/hello-parcel

DeMoorJasper:
Parcel does minify JS, minified version i got when running build of ur script was:
```
var t="123";const s="test2",a=665;
```
Currently minifying does not happen on the piece of JS Parcel includes every time because that doesn't get parsed

bartw:
@DeMoorJasper where did the console.log statements end up?

DeMoorJasper:
They get removed, why would you want console.log in build?

bartw:
Sorry, you are correct.

davidnagli:
> Currently minifying does not happen on the piece of JS Parcel includes every time because that doesn’t get parsed

Can you elaborate on that this @DeMoorJasper? Not fully understanding…

nfantone:
Probably related: what I'm seeing happening is that the `require` polyfill that precedes the actual app code is _not_ minified, while the rest _is_.

So the build output (after `parcel build index.js`) becomes:

```js
// modules are defined as an array
// [ module function, map of requires ]
//
// map of requires is short require name -> numeric require
//
// anything defined in a previous bundle is accessed via the
// orig method which is the require for previous bundles

require = (function (modules, cache, entry) {
  // Save the require from previous bundle to this closure if any
  var previousRequire = typeof require === "function" && require;

  function newRequire(name, jumped) {
    if (!cache[name]) {
      if (!modules[name]) {
        // if we cannot find the module within our internal map or
        // cache jump to the current global require ie. the last bundle
        // that was added to the page.
        var currentRequire = typeof require === "function" && require;
        if (!jumped && currentRequire) {
          return currentRequire(name, true);
        }

// ... unminified content continues <----

// Override the current require with this new one
  return newRequire;

[...]

PR #354, opened 2017-12-19
Minify the prelude.js file as well.

ssuman:
Currently prelude.js file isn't being minified. This PR will address that issue by moving the minification process from JSAsset to JSPackager. #334

### DIFF
--- a/src/assets/JSAsset.js
+++ b/src/assets/JSAsset.js
@@ -9,7 +9,6 @@ const insertGlobals = require('../visitors/globals');
 const fsVisitor = require('../visitors/fs');
 const babel = require('../transforms/babel');
 const generate = require('babel-generator').default;
-const uglify = require('../transforms/uglify');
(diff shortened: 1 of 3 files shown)


ssuman:
Tests are successful locally. It failed here. Any idea ?

devongovett:
Minification is done in JSAsset instead of JSPackager so that it can be done in parallel. This should be MUCH faster - not just because of the parallelism, but because minifying a bunch of small files is much faster than minifying one big file.

devongovett:
To fix prelude.js not being minified, I'd suggest doing this as a build step before we publish to npm.

nfantone:
@devongovett So, was this fixed after all? I was kinda waiting on this one. Do we have another issue to track progress?

nfantone:
For anyone else interested in this, progress in being tracked on https://github.com/parcel-bundler/parcel/issues/334.

PR #361, opened 2017-12-20
Add `.editorconfig`

yeskunall:
This is to avoid having to deal with line endings on different OSes.

I was going to instead make a PR adding `.editorconfig` but then realized that not everyone might have an extension/support for it.

Let me know your thoughts. Cheers! 💖

<hr />

__EDIT__: After review, opting for `.editorconfig` instead. 😅

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,15 @@
+# http://editorconfig.org
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+indent_size = 2
+indent_style = space
+insert_final_newline = true
+max_line_length = 80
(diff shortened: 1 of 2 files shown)


shawwn:
I'm a fan of .editorconfig personally. Webstorm supports it, and I think a bunch of others do too.

yeskunall:
Me too @shawwn. I'm more than willing to replace `.gitattributes` with an `.editorconfig` instead.

/cc @devongovett @DeMoorJasper thoughts?

devongovett:
.editorconfig seems widely supported. let's do it.

PR #363, opened 2017-12-21
Switch emojis to log-symbols for terminal compatibility

### LINKED ISSUE #362: Improved CLI Icons for Windows
🙋 Feature Request

### 🤔 Expected Behavior

Window's should fall back to icons that it supports and can render.

### 😯 Current Behavior

Window's renders colored boxes instead of ⌛️ ✨:

image

### 💁 Possible Solution

We could leverage something like [log-symbols](https://www.npmjs.com/package/log-symbols) to provide fallbacks. Alternatively, just detect Windows and conditionally show the existing ⌛️ ✨ vs other compatible icons.

### 🌍 Your Environment

| Parcel           |  N/A
| Node             |  N/A
| npm/Yarn         |  N/A
| Operating System |  Windows 10

shawwn:
Closes #362.

### DIFF
--- a/src/utils/emoji.js
+++ b/src/utils/emoji.js
@@ -1,0 +1,6 @@
+const supportsEmoji = process.platform !== 'win32' || process.env.VSCODE_PID;
+
+// Fallback symbols for Windows from https://en.wikipedia.org/wiki/Code_page_437
+exports.progress = supportsEmoji ? '⏳' : '∞';
+exports.success = supportsEmoji ? '✨' : '√';
+exports.error = supportsEmoji ? '🚨' : '×';
--- a/package.json
+++ b/package.json
@@ -25,6 +25,7 @@
     "is-url": "^1.2.2",
     "js-yaml": "^3.10.0",
     "json5": "^0.5.1",
+    "log-symbols": "^2.1.0",
     "micromatch": "^3.0.4",
     "mkdirp": "^0.5.1",
     "node-forge": "^0.7.1",
--- a/package.json
+++ b/package.json
@@ -25,7 +25,6 @@
     "is-url": "^1.2.2",
     "js-yaml": "^3.10.0",
     "json5": "^0.5.1",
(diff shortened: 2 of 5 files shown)


shawwn:
@brandon93s Ok, I switched out the emojis with log-symbols. It'll be up to @devongovett whether he's fond of the existing hourglass / star emojis though.

Alternatively, we could use log-symbols on windows, and the current emojis everywhere else.

brandon93s:
Yep. 

> Alternatively, just detect Windows and conditionally show the existing ⌛️ ✨ vs other compatible icons.

devongovett:
yeah let's keep the emojis by default and fall back

shawwn:
Ok, I'll preserve our emoji integrity.

#savemojis

shawwn:
Yep, agreed. I just got distracted by holidays and focusing on impactful
work (like abstracting the FS layer). Feel free to submit your own PR if
you get time to tackle this, otherwise I'll eventually get around to it.

devongovett:
Updated to use emoji but have fallback for windows. Apparently VS Code does support emoji though thanks to electron, so we try to detect that.

@brandon93s since you use windows, could you verify? Thanks!

brandon93s:
@devongovett  - great find on the emoji support in VS Code! Windows test outputs below:

PowerShell:

image

CMD:

image

VSCode (PowerShell):

image

I've tested failures as well. All three states (processing, success, failure) look good.

src/utils/emoji.js
```diff
@@ -0,0 +1,6 @@
+const supportsEmoji = process.platform !== 'win32' || process.env.VSCODE_PID;
```
brandon93s: 👍 I like the dedicated utility; should make maintenance easier!

devongovett:
Awesome, thanks for testing @brandon93s! Merry Christmas. 🎄

brandon93s:
Same to you! 🎄🎅🏼

Issue #372, opened 2017-12-21
🐛 HTML minification removes spaces around <span> tags in paragraphs


phocks:
🐛 bug report

### 🎛 Configuration (.babelrc, package.json, cli command)

```js
{
  "presets": ["env"]
}
```

### 🤔 Expected Behavior

Spaces should be preserved for span tags in paragraphs. This is the behaviour of most other HTML minifiers online it seems.

### 😯 Current Behavior

When I build for production the minified HTML removes the space before the <span> tags in paragraphs so the words run into each other.

### 💁 Possible Solution

Change settings in HTML minifier? Sorry I wasn't sure where to find this.

### 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Parcel           | 1.2.1
| Node             | 9.0.0
| npm/Yarn         | 5.5.1
| Operating System | macOS High Sierra

PR #373, opened 2017-12-22
Conserve the html spacing between tags

ssuman:
#372 

Currently, HTML minification strips the space between two <i> tags. The result might become "helloworld"

```
<i>hello</i> <i>world</i>
```
This PR would prevent that from happening.

### DIFF
--- a/test/integration/html/index.html
+++ b/test/integration/html/index.html
@@ -11,5 +11,6 @@
   <a href="tel:+33636757575">Tel link</a>
   <script src="index.js"></script>
   <script src="https://unpkg.com/parcel-bundler"></script>
+  <i>hello</i> <i>world</i>
 </body>
 </html>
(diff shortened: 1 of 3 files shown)


test/integration/html/index.html
```diff
@@ -11,5 +11,6 @@ <h1>Hello world</h1>
   <a href="tel:+33636757575">Tel link</a>
   <script src="index.js"></script>
   <script src="https://unpkg.com/parcel-bundler"></script>
+  <i>hello</i> <i>world</i>
```
brandon93s: To elaborate on this scenario, since you wouldn't write the above html in practice, consider the following:

```html
<div>
<span class="a">Hello in style a</span> <span class="b">World in style b</span>
</div>
```

ssuman: @brandon93s Should I update my PR ? Or is this just for info ?

brandon93s: Just info for reviewers.

devongovett: hmm, wonder why this is the default behavior for htmlnano? wonder if we should bring this up with them.

PR #374, opened 2017-12-22
Use `cross-spawn` to perform autoinstall

brandon93s:
Use [cross-spawn](https://www.npmjs.com/package/cross-spawn) to spawn process.

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,6 +17,7 @@
     "chalk": "^2.1.0",
     "chokidar": "^1.7.0",
     "commander": "^2.11.0",
+    "cross-spawn": "^5.1.0",
     "cssnano": "^3.10.0",
     "get-port": "^3.2.0",
     "glob": "^7.1.2",
--- a/src/utils/installPackage.js
+++ b/src/utils/installPackage.js
@@ -1,4 +1,4 @@
-const {spawn} = require('child_process');
+const spawn = require('cross-spawn');
 const config = require('./config');
 const path = require('path');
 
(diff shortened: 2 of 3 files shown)


brandon93s:
Well.. it fixed the `spawn` issue. Looks like `appveyor` and `yarn` aren't playing nicely now. Investigating.

brandon93s:
Commented out the `yarn.lock` check for now. The `package.json` check is succeeding, so we know the install is occurring. Seems `appveyor` has some global settings that are resulting in the `yarn.lock` not being updated. Will investigate separately,

shawwn:
This entire idea of automatically installing dependencies is a very risky
decision that may bite us hard in the future.

It's good to make things convenient, but coddling the programmer probably
isn't necessary. And the yarn/npm checks are just one reason why it's too
complex.

If someone wants to make another package manager, similar to yarn, how
would they even tell parcel how to use it? They can't. So we'd have to add
support for that or just not support that case at all.

Best leave it up to users.

I'm a fan of removing code. The smaller the codebase, the more effective it
usually is. But I know people have strong feelings about this so it's up to
everyone else.

It's just an experience thing. I used to try to get fancy with my programs
like this too, but ornament ends up causing problems.

Ditto for the emojis,by the way. Webstorm can automatically create
hyperlinks from error stack traces printed to the Webstorm console log,
meaning you can just click on the stack trace to be taken exactly to that
point in your codebase. It's incredibly handy and very fast for debugging.

But guess what, it doesn't work for Parcel because we decided normal error
reporting wasn't good enough and tried to make it super fancy and super
emojified, and webstorm has no idea how to interpret that.

And long error messages get cut off.

And all of these concerns simply vanish if we don't try to get fancy. :)

Issue #326, opened 2017-12-18
console 'Cannot find module' when html and js import a same image


BoltDoggy:
**Choose one:** is this a 🐛 bug report or 🙋 feature request?

🐛

### 🎛 Configuration (.babelrc, package.json, cli command)

`parcel src/index.html`

### 🤔 Expected Behavior

no Error
### 😯 Current Behavior

```
0b34c49e9b73a6078941be22bb88f26c.js:32 Uncaught Error: Cannot find module '2'
    at newRequire (0b34c49e9b73a6078941be22bb88f26c.js:32)
    at localRequire (0b34c49e9b73a6078941be22bb88f26c.js:38)
    at Object.require.4../logo.png (0b34c49e9b73a6078941be22bb88f26c.js:71)
    at newRequire (0b34c49e9b73a6078941be22bb88f26c.js:46)
    at require.4../logo.png (0b34c49e9b73a6078941be22bb88f26c.js:63)
    at 0b34c49e9b73a6078941be22bb88f26c.js:68
```

### 🔦 Context

https://github.com/lc60005457/parcel-demo-for-issues/tree/issues-326

```html
<!-- src/index.html -->
    <img src="./logo.png" alt="">
    <script src="./index.js"></script>
```

```js
// src/index.js
import './logo.png';
```

and a `src/logo.png`

### 🌍 Your Environment

| Software         | Version(s)
| ---------------- | ----------
| Parcel           |  v1.2.0
| Node             | v8.9.0
| npm/Yarn         | v5.5.1
| Operating System | Mac(10.13.2)

davidnagli:
So you want to get an error?? 🤔

DeMoorJasper:
He probably flipped it around by accident

davidnagli:
Ok, I went ahead and flipped it so it makes sense (I edited the issue)

maximilianschmitt:
I'm having the same issue with `scss` files. If I only `import` `scss` files from my `js` files, everything works as expected. If I import one `scss` file from `index.html` and other files from a `js` file, I get the "cannot find module <some number>" error. The error goes away if I turn off HMR.

benhutton:
do you think that this is describing the same thing? https://github.com/parcel-bundler/parcel/issues/294

(Just at a more abstract level)

ssuman:
@benhutton Yes this is similar to #294. Since common bundle and parent bundle types are different, parcel currently ignores one of the bundles.

[...]

Issue #294, opened 2017-12-15
🐛 Parcel Bundling issues


ssuman:
🐛 bug report
Parcel currently builds a bundle tree. While building the tree, if it finds a common module, it finds the common ancestor bundle and places the common module in that ancestor bundle. Now this only works if the asset's parent bundle and the common bundle are the same type. For example: both are of type JS.

There are various cases where the asset's parent bundle type is not the same as the common bundle type. Few cases shown below - 

Case (i):
<img width="376" alt="screen shot 2017-12-14 at 12 39 10 pm" src="https://user-images.githubusercontent.com/2545586/34017432-bc404a64-e0f3-11e7-95e5-84c69ee8c244.png">

case (ii):
<img width="385" alt="screen shot 2017-12-14 at 5 26 47 pm" src="https://user-images.githubusercontent.com/2545586/34017509-09a12292-e0f4-11e7-8966-b5096b754b62.png">

case (iii): 
<img width="396" alt="screen shot 2017-12-14 at 5 27 39 pm" src="https://user-images.githubusercontent.com/2545586/34017551-3467d9f8-e0f4-11e7-8fcd-f441b92b1c13.png">

### 🤔 Expected Behavior

case (i): Util.js should be bundled in both bundles B2 and B3.

case (ii): Not sure, but I think a hashed Util.js file should be referenced in index.html. Additionally, util.js should also be present in Bundles B2 and B3.

case (iii): A stylesheet should be appended to the head of both html files ( index.html and other.html).

### 😯 Current Behavior
case (i): Util.js is present only in bundle B2. So we get module not found exception.

case (ii): Same as case i. In addition, html file has a hash reference to util.js but is never created.

case (iii): CSS stylesheet is appended only to either other.html or index.html based on the order of execution.

### 💁 Possible Solution
I'm not 100% sure. I think we need to treat each html nodes as boundaries. The common chuck can't be moved beyond the boundary. It should be treated as an isolated entity. Although I'm not sure if this would address all the cases. But would love to work on this PR after some discussion,

### 🌍 Your Environment

| Software         | Version(s)
| ---------------- | ----------
| Parcel           |  
| Node             | 8.4.0
| npm/Yarn         | 5.3.0
| Operating System | MacOS Sierra 10.12.1

mBeierl:
[...]

PR #380, opened 2017-12-22
Potential fix to parcel bundling issues when asset parent & commonBundle types differ

ssuman:
#294 #326 
This PR potentially fixes the issues mentioned above. It does the following
case (i):
If the below is true => i.e both types are the same (parent Bundle is JS and commonBundle is also JS) move it the common ancestor.
```
if (
          asset.parentBundle !== commonBundle &&
          asset.parentBundle.type === commonBundle.type
        ) {
          this.moveAssetToBundle(asset, commonBundle);
          return;
        }
```
case 2:
If they aren't the same, for eg: parentBundle is JS and commonBundle is HTML. Then allow it to continue further so that it adds this asset to the immediate parent.

case 3: 
If both parent bundles are equal, then just return.
```
if (asset.parentBundle !== bundle) {
```

### DIFF
--- a/src/Bundler.js
+++ b/src/Bundler.js
@@ -373,10 +373,9 @@ class Bundler extends EventEmitter {
           asset.parentBundle.type === commonBundle.type
         ) {
           this.moveAssetToBundle(asset, commonBundle);
+          return;
         }
-      }
-
-      return;
+      } else return;
     }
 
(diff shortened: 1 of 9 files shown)


devongovett:
Thanks for tracking this down and fixing! 🎉

PR #398, opened 2017-12-25
Add url dependency for serviceWorker.register calls

jdanyow:
Adds support for `serviceWorker.register('foo.js')`.

Maybe we should support `self.importScripts('foo.js', 'bar.js', ...);` too?

### DIFF
--- a/src/visitors/dependencies.js
+++ b/src/visitors/dependencies.js
@@ -2,9 +2,11 @@ const types = require('babel-types');
 const template = require('babel-template');
 const urlJoin = require('../utils/urlJoin');
 const isURL = require('../utils/is-url');
+const matchesPattern = require('./matches-pattern');
 
 const requireTemplate = template('require("_bundle_loader")');
 const argTemplate = template('require.resolve(MODULE)');
+const serviceWorkerPattern = ['navigator', 'serviceWorker', 'register'];
 
 module.exports = {
   ImportDeclaration(node, asset) {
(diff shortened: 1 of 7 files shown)


src/visitors/dependencies.js
```diff
@@ -51,6 +54,25 @@ module.exports = {
...
+      callee.property.name === 'register' &&
+      callee.object !== undefined &&
+      callee.object.property !== undefined &&
+      callee.object.property.name === 'serviceWorker' &&
+      types.isStringLiteral(args[0]);
+
+    if (isRegisterServiceWorker) {
+      let assetPath = asset.addURLDependency(args[0].value);
+      if (!isURL(assetPath)) {
+        assetPath = urlJoin(asset.options.publicURL, assetPath);
```
devongovett: maybe we should just make `addURLDependency` include the publicURL?

src/visitors/dependencies.js
```diff
@@ -51,6 +54,25 @@ module.exports = {
...
+      return;
+    }
+
+    const isRegisterServiceWorker =
+      callee.property !== undefined &&
+      callee.property.name === 'register' &&
+      callee.object !== undefined &&
+      callee.object.property !== undefined &&
+      callee.object.property.name === 'serviceWorker' &&
+      types.isStringLiteral(args[0]);
```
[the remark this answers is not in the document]
jdanyow: :+1:

src/visitors/dependencies.js
jdanyow: not sure- CSSAsset, LESSAsset, SASSAsset and StylusAsset use `addURLDependency`'s return value without prefixing with the public url. Will prefixing with the public url by default cause problems in these asset types?

PR #400, opened 2017-12-25
[WIP] Uglify JSON modules for production

### LINKED ISSUE #399: 🐛 JSON files don't get minified

https://github.com/K900/parcel-test

No special configuration changes needed, run `yarn run parcel build main.js`, then look at `dist/main.js`.

### 🤔 Expected Behavior

JSON content is minified, all extra whitespace is removed.

### 😯 Current Behavior

JSON content is included as-is, preserving whitespace.

### 💁 Possible Solution

Run JSON files through uglify-es just like normal JS files before bundling.

### 🔦 Context

Saving some bytes.

### 🌍 Your Environment

| Parcel           | 1.3.1
| Node             | 9.3.0

K900:
Saves a few bytes, fixes #399

### DIFF
--- a/src/assets/JSONAsset.js
+++ b/src/assets/JSONAsset.js
@@ -1,4 +1,5 @@
 const JSAsset = require('./JSAsset');
+const uglify = require('../transforms/uglify');
 
 class JSONAsset extends JSAsset {
   async load() {
@@ -8,7 +9,11 @@ class JSONAsset extends JSAsset {
   parse() {}
   collectDependencies() {}
   pretransform() {}
-  transform() {}
+  async transform() {
+    if (this.options.minify) {
+      await uglify(this);
+    }
+  }
 }
 
 module.exports = JSONAsset;


brandon93s:
Also worth keeping in mind that the `.uglifyrc` will now apply to both JS and JSON files. This shouldn't be an issue, but is worth be aware of.

kzc:
https://github.com/mishoo/UglifyJS2/issues/812#issuecomment-143431988

K900:
Something weird is happening here - I managed to get it to _not_ minify once, but now it's working correctly. I'll look into it more after the holidays.

devongovett:
Yeah just `JSON.stringify(JSON.parse(contents))` should get rid of the extra whitespace, but won't get rid of unnecessary quotes around keys. Maybe use something like https://github.com/yeoman/stringify-object?

kzc:
If you don't care whether the result is JSON compliant just do this:
```js
$ echo '{ "foo": 1, "bar": [2, 3], "hello world": false }' | bin/uglifyjs --parse expression
{foo:1,bar:[2,3],"hello world":false}
```
or the `minify()` equivalent.

PR #401, opened 2017-12-25
Resolve Dependency correctly if the target asset is URI-encoded

zacky1972:
This patch fixes the bug that parcel can't read assets including encoded URI such as:

```html
<img src="images/%e6%97%a5%e6%9c%ac%e8%aa%9e.jpg" />
```


devongovett:
I think this should be specific to HTML assets. We don't want to double decode in the case where something has a legit percent symbol in the filename.

brandon93s:
I agree on the HTML asset only piece. Let's limit this to the use case. 

A few more thoughts:

- `decodeURIComponent` only supports UTF-8. If someone chooses to use ISO-8859-1 or other on their site, this is going to throw an exception. We need to handle this or provide a sensible error message. The asset failure shouldn't cascade into a decode failure. 

- With that in mind, the current implementation could throw in the catch block for encoding that isn't UTF-8, which would not be desirable. 

- I'd prefer error messages show the original asset name, not a decoded name. When debugging, I'd want to see the values present in my source not their decoded representations.

zacky1972:
Thank you for your comments.

I'd like to reorganize this patch into the HTML asset, but I can't understand where I should insert the patch into the source code of the HTML asset.  Help me!

shawwn:
Hi, if you hop on our slack and send me a message (I'm @shawwn) I can help
you sort through the codebase.

zacky1972:
Thank you! I sent a Slack message, just now.

zacky1972:
I've just pushed PR worked with shawwn. 
Please review it, again!

shawwn:
@devongovett @brandon93s This PR looks good to me now. Any further thoughts?

I think the UTF-8 concern is good, but we might want to merge this for now. Supporting at least one Japanese user is better than supporting all hypothetical foreign users, and @zacky1972 has spent a lot of time on this already.

devongovett:
Would be awesome to add a test if possible, but it looks good to me otherwise!

zacky1972:
OK! I'd like to add a test.

zacky1972:
I add a test code.

But, I found another problem. This patch assumes encoding of the file system is UTF-8.
If encoding is CP932, which is used in Windows in Japan, or EUC-JP, which is used in old UNIX system in Japan, the patch might be unworkable.

I'd like to test such systems, but I don't have them.

zacky1972:
I call for help by Twitter and Facebook.

https://twitter.com/zacky1972/status/949342180430237697
https://www.facebook.com/zacky1972/posts/1799791536720470

> parcel ユーザーで，ファイルシステムが UTF-8 でない Windows や Linux 等をお使いの方いらっしゃいませんか？ テストにご協力ください！#拡散希望

This means:
 
Help us, if you use parcel and encoding of the file system of your PC (Windows or Linux etc.) is not UTF-8!

devongovett:
I think Node will handle converting from UTF8 and back for us, but it would be good to verify. Some info here: https://nodejs.org/api/fs.html#fs_buffer_api

zacky1972:
Oh, really! 
I hope this patch works in non-UTF-8 file systems...

brandon93s:
`decodeURIComponent` itself only handles UTF-8 though (and will throw otherwise), even in node. More details: https://github.com/sindresorhus/got/issues/420.

We may want to take a "try-decode" approach here, continuing with the raw value if decoding fails.

devongovett:
Let's get this in and deal with any other encoding issues as they come up. Thanks for your work on this @zacky1972!

zacky1972:
You're welcome!
I'd like to continue to support parcel, not only about this issue but also about any other issues if I can!

zacky1972:
I've confirmed it on Windows PC of the Japanese edition,  and it works well.

Issue #106, opened 2017-12-07
Babel (babel-preset-env) breaks inlining via fs.readFileSync(...)


simoncave:
Repro project: https://github.com/simoncave/parcel-babel-inline-test

### 🤔 Expected Behavior

According to [Parcel's "Assets" documentation](https://parceljs.org/assets.html):

> If you want to inline a file into the JavaScript bundle instead of reference it by URL, you can use the Node.js `fs.readFileSync` API to do that. The URL must be statically analyzable, meaning it cannot have any variables in it (other than `__dirname` and `__filename`).

When successful, the inlining should directly replace `fs.readFileSync(...)` with the asset being inlined. In the case of [this example](https://github.com/simoncave/parcel-babel-inline-test/blob/master/sayHello.js#L4), it should replace the call with `"World"`.

This inlining feature should work regardless of how the code is transformed, provided the transformation is not doing anything that should explicity modify the usage pattern expressed above (in short: it should work with [babel-preset-env]).

### 😯 Current Behavior

This appears to break when using Babel, or at least when using [babel-preset-env] with module transpilation enabled. The assets transformation _does not_ replace the call to `fs.readFileSync(...)`. Instead, the code is left untransformed, resulting in the following error:

```
TypeError: _fs2.default.readFileSync is not a function. (In '_fs2.default.readFileSync(__dirname + '/name.txt', 'utf8')', '_fs2.default.readFileSync' is undefined)
```

### Workaround

This issue can be avoided by disabling the `es6-modules-amd` transformation. Example `.babelrc`:

```json
{
  "presets": [
    ["env", {
      "modules": false
    }]
  ]
}
```

Parcel does its own module transpilation, so doing so with Babel is redundant and also seems to be the cause of this issue. One way to resolve this issue would be for Parcel to detect module transpilation in a user's `.babelrc` and provide instructions for disabling it. Another way would be to improve Parcel's matching of `fs.readFileSync(...)` such that it can succeed after module transpilation.

### 🎛 Configuration

- **Deps**

    ```
    parcel-inline-test@0.0.0
    ├── babel-preset-env@1.6.1
    ├── node-pre-gyp@0.6.39 extraneous
    └── parcel-bundler@1.0.3
    ```
- **.babelrc**

    ```json
    {
      "presets": ["env"]
    }
    ```
- **Code**

    ```js
    const name = fs.readFileSync(__dirname + '/name.txt', 'utf8');
[...]

PR #404, opened 2017-12-26
Support babel's _interopRequireDefault for inlining fs.readFileSync

d4rkr00t:
Closes: #106 

This PR fixes an issue with babel transpiled import for fs.readFileSync:
```js
var _fs = require("fs");

var _fs2 = _interopRequireDefault(_fs);

function _interopRequireDefault(obj) { return obj && obj.__esModule ? obj : { default: obj }; }

var name = _fs2.default.readFileSync(__driname + '/name.text');
document.body.innerHTML = "Hello, " + name + "! Inlining via fs.readFileSync is working!";
```

It sort of simplifies `referencesImport` method, but at the same time it's not the "most performant" solution (traverse couple of times) :( I ran tests several times and couldn't see any performance regressions, but they are too small to be able to reveal any perf impact.

Also i re-enabled fs-import test. It wasn't included for, i guess, the same reason.

Another option could be detecting babel presets/plugins that transpile imports and show error message.


devongovett:
hmm, perhaps we should just turn off babel transpiling modules. we could automatically set the `modules: false` option to babel-preset-env etc. pre dependency detection. we already transpile only modules if needed post detection, so this shouldn't break anything. thoughts?

d4rkr00t:
Agree, it shouldn't break anything. i'll update my PR soonish, thanks :)

src/transforms/babel.js
```diff
@@ -8,22 +8,47 @@ module.exports = async function(asset) {
...
+    (await config.load(asset.name, ['.babelrc', '.babelrc.js'])) || {};
+  let babelConfig = Object.assign(babelrc, {
     code: false,
-    filename: asset.name
-  };
+    filename: asset.name,
+    babelrc: false
+  });
+
+  // TODO: Support .babelignore
```
d4rkr00t: In order to set modules: false to babel-preset-env I had to disabled automatic loading of .babelrc which as well disables loading of .babelrcignore :(

d4rkr00t: I'll add support for that if you are happy with everything else

d4rkr00t:
### Updates:

Instead of changing `referencesImport` logic I make babel transform load `.babelrc`, update opts for presets and use it instead of babel looking up configs on its own.

devongovett:
This appears to have been fixed by other changes.

PR #426, opened 2017-12-28
patch(hmr): use hostname for WS instantiation

Fluidbyte:
This is in response to #276 and removes hard-coded `localhost` for websockets to use the `window.location.hostname` instead, allowing users to run this on remote hosts or different domains.


devongovett:
Looks good, the tests are failing though. You'll need to shim `window.location.hostname` in https://github.com/parcel-bundler/parcel/blob/master/test/utils.js#L78

Fluidbyte:
Yeah, a bit hasty push on my part, will get that fixed and pushed.

Fluidbyte:
Tests passing, lmk if there's anything else needed.

Issue #437, opened 2017-12-29
🐛 Cannot resolve dependency for `itms-apps://...`


hkwid:
When I tried to use the app store link like `itms-apps://itunes.apple.com/en/app/awesomeApp/id1234567890` as `href` attr.
Get `Cannot resolve dependency` error.

```html
<!-- index.html -->
...
<a href="itms-apps://itunes.apple.com/en/app/awesomeApp/id1234567890">App Store Link</a>
...
```

### 🎛  Configuration

```bash
$ parcel src/index.html
```

### 🤔  Expected Behavior

When the URI schema is valid, it should not try to resolve the dependency.

### 😯  Current Behavior

Even the URI schema is valid, it tries to resolve the dependency.
The error log is like below

```bash
/path/to/the/project/src/index.html: Cannot resolve dependency './itms-apps:/itunes.apple.com/en/app/awesomeApp/id1234567890' at '/path/to/the/project/src/itms-apps:/itunes.apple.com/en/app/awesomeApp/id1234567890'
at /path/to/the/project/node_modules/browser-resolve/node_modules/resolve/lib/async.js:55:21
at load (/path/to/the/project/node_modules/browser-resolve/node_modules/resolve/lib/async.js:69:43)
at onex (/path/to/the/project/node_modules/browser-resolve/node_modules/resolve/lib/async.js:92:31)
at /path/to/the/project/node_modules/browser-resolve/node_modules/resolve/lib/async.js:22:47
at FSReqWrap.oncomplete (fs.js:152:21)
```

### 💁 Possible Solution

According to wiki [https://en.wikipedia.org/wiki/Uniform_Resource_Identifier#Syntax](https://en.wikipedia.org/wiki/Uniform_Resource_Identifier#Syntax), URI schema can include `.`, `-`, `+` ... etc.

Maybe tweak `SCHEME_REGEXP` in `src/utils/is-url.js` to cover such a schemas.

devongovett:
Should be fixed by #438 which will be in a release shortly.

PR #438, opened 2017-12-29
handle app store url scheme

jdanyow:
saw this in the chat: https://parcel-bundler.slack.com/archives/C8B26V8RF/p1514579681000034

### DIFF
--- a/src/utils/is-url.js
+++ b/src/utils/is-url.js
@@ -3,8 +3,8 @@ const isURL = require('is-url');
 // Matches anchor (ie: #raptors)
 const ANCHOR_REGEXP = /^#/;
 
-// Matches scheme (ie: tel:, mailto:, data:)
-const SCHEME_REGEXP = /^[a-z]*:/i;
+// Matches scheme (ie: tel:, mailto:, data:, itms-apps:)
+const SCHEME_REGEXP = /^[a-z-]*:/i;
 
 module.exports = function(url) {
   return isURL(url) || ANCHOR_REGEXP.test(url) || SCHEME_REGEXP.test(url);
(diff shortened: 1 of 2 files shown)


leonardosnt:
Related: https://github.com/parcel-bundler/parcel/issues/437

Why not use `/^[\w-+.]*:/i` ? (or `^[a-z][\w-+.]*:`) It will handle more use cases.
> The scheme, consisting of a sequence of characters beginning with a letter and followed by any combination of letters, digits, plus (+), period (.), or hyphen (-)

https://en.wikipedia.org/wiki/Uniform_Resource_Identifier#Syntax

hkwid:
According to here https://www.iana.org/assignments/uri-schemes/uri-schemes.xhtml
URI schema can include numbers like `z39.50`.

How about `/^[a-z][a-z0-9\+-\.]*:/i`?

BTW, for integration test, we can add `<a href="itms-apps://...">App Store Link</a>` to `test/integration/html/index.html`

```html
...
  <a href="tel:+33636757575">Tel link</a>
  <a href="itms-apps://itunes.apple.com/en/app/awesomeApp/id1234567890">App Store Link</a>
  <script src="index.js"></script>
...
```

leonardosnt:
`\w` already cover alphanumeric (a-z0-9)

hkwid:
@leonardosnt I guess `\w` is included `A-Z` too. URI schema seems like must be lowercase.

> Although schemes are case-insensitive, the canonical form is lowercase and documents that specify schemes must do so with lowercase letters. It is followed by a colon (:). Examples of popular schemes include http(s), ftp, mailto, file, data, and irc. URI schemes should be registered with the Internet Assigned Numbers Authority (IANA), although non-registered schemes are used in practice.
[https://en.wikipedia.org/wiki/Uniform_Resource_Identifier#Syntax](https://en.wikipedia.org/wiki/Uniform_Resource_Identifier#Syntax)

leonardosnt:
Yeah, you are right. \w also includes underscore. `a-z0-9` is better

Issue #289, opened 2017-12-15
🙋 Regular Browser Reload on File Save


mattdesl:
🙋

Hi there! I'd like to start exploring Parcel as a workflow for creative coding (WebGL, Canvas2D, etc). I'm trying to migrate some programs I've written for non-HMR workflow to this new HMR workflow, and I'm hitting huge performance issues since the code (which is doing things like creating a WebGL/Canvas context) was not designed to be re-run multiple times.

### 🎛 Configuration 

Take a simple program like this:

```js
const canvas = document.createElement('canvas');
const ctx = canvas.getContext('2d');

document.body.appendChild(canvas);
```

### 🤔 Expected Behavior

I'd like to develop with the same way that a JS file runs/loads in the browser (i.e. once, not many times). If possible, I'd like a way to replace JavaScript HMR with a simple `window.location.reload()` functionality. However, other features (like CSS) should still use HMR / inject without hard reload.

### 😯 Current Behavior

Currently the above code, when saved several times, will create several canvas elements in the body.

### 💁 Possible Solution

A way of turning on/off regular hot reload. I am assuming this may already exist, but I couldn't find it, so perhaps it's more an issue of documentation?

brandon93s:
Give `--no-hmr` a try when you call parcel. It sounds like it may not be exactly what you're looking for, but it could be an improvement.

mattdesl:
Thanks! But that just turns off reloading altogether... 😆

mattdesl:
For now, I have a branch with the option `--reload` which will trigger a hard-reload on any non-CSS assets.

https://github.com/mattdesl/parcel/tree/feature/add-js-reload

davidnagli:
Honestly I don’t think this has any practical use cases outside of the one you described.

Is there anywhere else where somebody would actually want this option? If so, feel free to submit a PR with your `add-js-reload` branch.

mattdesl:
> Honestly I don’t think this has any practical use cases outside of the one you described.

You mean, creating a new HTML element and adding it to the body? It's really one of the most basic things you can do with JS...

I am genuinely curious how I could change my application to better support HMR.

Here is another example: creating a loop with `requestAnimationFrame`:

```js
window.requestAnimationFrame(render);
console.log('Starting loop...');

function render (time) {
  window.requestAnimationFrame(render);
  console.log('Rendering', time);
}
```

If you save the file twice you will end up with two requestAnimationFrame loops running simultaneously.

[...]

PR #443, opened 2017-12-30
Add --reload option for hard reload on non-CSS assets

mattdesl:
With this new `--reload` option, parcel will trigger hard page reloads on JS and HTML assets, but should continue to inject updates with CSS assets.

See discussion here:
https://github.com/parcel-bundler/parcel/issues/289

Personally I feel this should be the default rather than opt-in (HMR is not even working currently) but to each their own. 🍻


src/HMRServer.js
```diff
@@ -47,8 +51,11 @@ class HMRServer {
       });
     }
 
-    const containsHtmlAsset = assets.some(asset => asset.type === 'html');
-    if (containsHtmlAsset) {
+    const isReload = this.options.reload;
+    const isReloadAsset = assets.some(asset => {
+      return isReload ? asset.type !== 'css' : asset.type === 'html';
```
TennyZhuang: It seems that the `isReload` condition check can move out of the `some` loop

devongovett:
Is this option really needed? Could you just hook into the HMR system yourself to do a reload?

in your main JS file:
```javascript
if (module.hot) {
  module.hot.accept(function () {
    window.reload();
  });
}
```

brandon93s:
+1 on the "is this needed" front. The reload suggestion above works, and building your code to perform a DOM "upsert" (see issue suggestions) also works. 

This does however, seem to be desired by quite a few.

src/HMRServer.js
xu3u4: Actually I hope the modification for styles can also reload the page, because it is related to DOM.
So I am thinking would it be better to be like `isReload ? asset.type === 'js' : (asset.type === 'html' || asset.type === 'css')`?

xu3u4:
Agree.
The HMR only appends the new content, which means the old content still exists.
So the number of times we save our js module, the same number of different old version script exists. And when the corresponding action is triggered, it runs through all the appended scripts, from the oldest to the newest.
I think it is a little bit odd.
Adding `--reload` might be a quicker and safer solution.

src/HMRServer.js
mattdesl: It sounds like you want reload to happen in any situation, so:

```js
const isReloadAsset = isReload ? true : assets.some(a => a.type === 'html');
```

During the PR I assumed it was desirable to have CSS apply without page reloads, since it doesn't depend on JavaScript state and the like. But because it's using `import`, maybe it runs into the same problems, and it's better to just hard reload the page?

xu3u4: Oh yes. Maybe hot reload every time is more like what I need 😆

OrganicPanda:
Looks like this has been approved but has fallen in to conflict. Would love to see this merged though - @mattdesl do you have any time to fix the conflicts?

filipesabella:
hey @mattdesl if you feel you can't find time to see this one through, I'd be happy to take on solving the conflicts.

mattdesl:
See the latest comment in the original discussion thread - it sounds like a cleaner fix than the command line flag in this PR. 

https://github.com/parcel-bundler/parcel/issues/289#issuecomment-379529297

filipesabella:
Ah I see, thanks for the reply.

Well, to be honest, I get browser freezes with HMR on with the two apps I'm working on.
On is a vanilla typescript+react+redux and the other one is a canvas thing.
As it stands now I have to disable HMR and refresh the browser every time, which's a a little unfortunate.

As much as I like the idea of HMR, I'd love to just have the option you introduced in this PR, but I understand that it doesn't match the project's vision.

OrganicPanda:
Same here! HMR gives me limited confidence so I end up refreshing manually anyway haha

marcushellberg:
I find myself needing this in projects using web components. Calling `customElements.define()` fails if the element has already been defined.

devongovett:

PR #444, opened 2017-12-30
[WIP] add elm support

rtorr:
I saw reasonml support, so I suspect something like this could be wanted.

Used some code from https://github.com/ssuman/parcel-plugin-elm

cc @ssuman

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "parcel-bundler",
-  "version": "1.3.1",
+  "version": "1.4.0",
   "main": "index.js",
   "license": "MIT",
   "repository": {
@@ -72,8 +72,11 @@
   "scripts": {
     "test": "cross-env NODE_ENV=test mocha",
     "format": "prettier --write './{src,bin,test}/**/*.{js,json,md}'",
-    "build": "babel src -d lib",
+    "build":
+      "yarn minify && babel src -d lib && cp -R src/builtins lib/builtins",
     "prepublish": "yarn build",
+    "minify":
+      "uglifyjs -c -m -o src/builtins/prelude.min.js src/builtins/prelude.js",
     "precommit": "npm run lint && lint-staged",
     "lint": "eslint .",
     "postinstall":
--- a/package.json
+++ b/package.json
@@ -49,6 +49,7 @@
     "bsb-js": "^1.0.1",
     "coffeescript": "^2.0.3",
     "cross-env": "^5.1.1",
+    "elm": "^0.18.0",
     "eslint": "^4.13.0",
     "graphql": "^0.11.7",
     "graphql-tag": "^2.6.0",
--- a/package.json
+++ b/package.json
@@ -44,6 +44,7 @@
   },
   "devDependencies": {
     "babel-cli": "^6.26.0",
+    "babel-eslint": "^8.1.2",
     "babel-preset-env": "^1.6.1",
     "bsb-js": "^1.0.1",
     "coffeescript": "^2.0.3",
@@ -57,6 +58,7 @@
     "mocha": "^3.5.0",
     "ncp": "^2.0.0",
     "nib": "^1.1.2",
+    "node-elm-compiler": "^4.4.1",
     "node-sass": "^4.5.3",
     "nyc": "^11.1.0",
     "postcss-modules": "^0.8.0",
--- a/package.json
+++ b/package.json
@@ -44,7 +44,6 @@
   },
   "devDependencies": {
     "babel-cli": "^6.26.0",
-    "babel-eslint": "^8.1.2",
     "babel-preset-env": "^1.6.1",
     "bsb-js": "^1.0.1",
     "coffeescript": "^2.0.3",
--- a/src/assets/ElmAsset.js
+++ b/src/assets/ElmAsset.js
@@ -1,0 +1,23 @@
+const JSAsset = require('./JSAsset');
+const localRequire = require('../utils/localRequire');
+const process = require('process');
+
+class ElmAsset extends JSAsset {
+  getParserOptions() {
+    const defaultOptions = {
+      yes: true,
+      cwd: process.cwd()
+    };
+    return defaultOptions;
+  }
+
+  async parse() {
+    let elmCompiler = await localRequire('node-elm-compiler', this.name);
+    const options = this.getParserOptions();
+    const data = elmCompiler.compileToStringSync(this.name, options);
+    this.contents = data.toString();
+    return await super.parse(this.contents);
+  }
+}
+
+module.exports = ElmAsset;
--- a/src/assets/ElmAsset.js
+++ b/src/assets/ElmAsset.js
@@ -1,6 +1,7 @@
 const JSAsset = require('./JSAsset');
 const localRequire = require('../utils/localRequire');
 const process = require('process');
+const config = require('../utils/config');
 
 // Currently no pretty way to install Elm tools on behalf of the user :/
 async function compile(name, options) {
@@ -20,11 +21,21 @@ async function compile(name, options) {
 
 class ElmAsset extends JSAsset {
   async getParserOptions() {
-    const defaultOptions = {
+    let defaultOptions = {
       yes: true,
-      cwd: process.cwd(),
-      verbose: true
+      cwd: process.cwd()
     };
+
+    // grab other optios from package.elm.json
+    const elmOptions = await config.load(this.name, ['elm-package.json']);
+
+    if (elmOptions.nodeElmCompilerOptions) {
+      defaultOptions = Object.assign(
+        elmOptions.nodeElmCompilerOptions,
+        defaultOptions
+      );
+    }
+
     // Use a local copy of elm-make when in test
     if (process.env.NODE_ENV === 'test') {
       defaultOptions.pathToMake = require.resolve('elm/binwrappers/elm-make');
--- a/src/assets/ElmAsset.js
+++ b/src/assets/ElmAsset.js
@@ -3,22 +3,6 @@ const localRequire = require('../utils/localRequire');
 const process = require('process');
 const config = require('../utils/config');
 
-// Currently no pretty way to install Elm tools on behalf of the user :/
-async function compile(name, options) {
-  try {
-    const elmCompiler = await localRequire('node-elm-compiler', name);
-    // compileToStringSync does not throw an Error, even when there is one.
-    // Other methods will throw an error, but do not bubble up to Parcel
-    const data = elmCompiler.compileToStringSync(name, options);
-    if (!data) {
-      throw 'You either have an error in your code, or are missing the Elm tools here https://guide.elm-lang.org/install.html.';
-    }
-    return data;
-  } catch (error) {
-    throw error;
-  }
-}
-
 class ElmAsset extends JSAsset {
   async getParserOptions() {
     let defaultOptions = {
@@ -45,8 +29,15 @@ class ElmAsset extends JSAsset {
 
   async parse() {
     const options = await this.getParserOptions();
-    const data = await compile(this.name, options);
-    this.contents = data.toString();
+
+    const elm = await localRequire('node-elm-compiler', this.name);
+
+    try {
+      this.contents = await elm.compileToString(this.name, options);
+    } catch (err) {
+      throw new Error(err);
+    }
+
     return await super.parse(this.contents);
   }
 }
--- a/src/assets/ElmAsset.js
+++ b/src/assets/ElmAsset.js
@@ -2,19 +2,39 @@ const JSAsset = require('./JSAsset');
 const localRequire = require('../utils/localRequire');
 const process = require('process');
 
+// Currently no pretty way to install Elm tools on behalf of the user :/
+async function compile(name, options) {
+  try {
+    const elmCompiler = await localRequire('node-elm-compiler', name);
+    // compileToStringSync does not throw an Error, even when there is one.
+    // Other methods will throw an error, but do not bubble up to Parcel
+    const data = elmCompiler.compileToStringSync(name, options);
+    if (!data) {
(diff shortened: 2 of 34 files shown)


rtorr:
Hrmmm. Tests pass locally. Could this be because I am on windows?

rtorr:
Ooops, its missing dependencies

shawwn:
Hi,

I looked over your work. It seems pretty good! If you need any help with anything or have any questions, hop on our Slack channel (I'm @shawwn on there).

Feel free to reopen this PR if you want – my understanding is that Parcel is erring on the side of "more support = better", Elm integration should be ok, especially because it's pretty lightweight.

rtorr:
@shawwn did you test this out locally?

shawwn:
unfortunately no, I'm currently busy adding some other features. But I can
help answer specific questions – glancing over the code, it looks pretty
reasonable.

rtorr:
This is the output im getting :)

Something is not right.

```javascript
// modules are defined as an array
// [ module function, map of requires ]
//
// map of requires is short require name -> numeric require
//
// anything defined in a previous bundle is accessed via the
// orig method which is the require for previous bundles

require = (function (modules, cache, entry) {
  var previousRequire = typeof require === "function" && require;

  function newRequire(name, jumped) {
    if (!cache[name]) {
      if (!modules[name]) {
        // if we cannot find the module within our internal map or
        // cache jump to the current global require ie. the last bundle
        // that was added to the page.
        var currentRequire = typeof require === "function" && require;
        if (!jumped && currentRequire) {
          return currentRequire(name, true);
        }

        // If there are other bundles on this page the require from the
        // previous one is saved to 'previousRequire'. Repeat this as
        // many times as there are bundles until the module is found or
        // we exhaust the require chain.
        if (previousRequire) {
          return previousRequire(name, true);
        }

        var err = new Error('Cannot find module \'' + name + '\'');
        err.code = 'MODULE_NOT_FOUND';
        throw err;
      }
      
      localRequire.resolve = resolve;

      var module = cache[name] = new newRequire.Module;

      modules[name][0].call(module.exports, localRequire, module, module.exports);
    }

    return cache[name].exports;

    function localRequire(x){
      return newRequire(localRequire.resolve(x));
    }

    function resolve(x){
      return modules[name][1][x] || x;
    }
  }

  function Module() {
    this.bundle = newRequire;
    this.exports = {};
  }

  newRequire.Module = Module;
  newRequire.modules = modules;
  newRequire.cache = cache;
  newRequire.parent = previousRequire;

  for (var i = 0; i < entry.length; i++) {
    newRequire(entry[i]);
  }

  // Override the current require with this new one
  return newRequire;
})({4:[function(require,module,exports) {
module.exports="/dist/74a93514c957e96dddff8f578e33c5f9.elm";
},{}],3:[function(require,module,exports) {
console.log('wat');

},{}],2:[function(require,module,exports) {
"use strict";

var _HelloWorld = require("./HelloWorld.elm");

var _HelloWorld2 = _interopRequireDefault(_HelloWorld);

require("./test.ts");

function _interopRequireDefault(obj) { return obj && obj.__esModule ? obj : { default: obj }; }

console.log(_HelloWorld2.default);
},{"./HelloWorld.elm":4,"./test.ts":3}],0:[function(require,module,exports) {
var global = (1, eval)('this');
var OldModule = module.bundle.Module;
function Module() {
  OldModule.call(this);
  this.hot = {
    accept: function (fn) {
      this._acceptCallback = fn || function () {};
    },
    dispose: function (fn) {
      this._disposeCallback = fn;
    }
  };
}

if (!module.bundle.parent && typeof WebSocket !== 'undefined') {
  var ws = new WebSocket('ws://' + window.location.hostname + ':59435/');
  ws.onmessage = function(event) {
    var data = JSON.parse(event.data);

    if (data.type === 'update') {
      data.assets.forEach(function (asset) {
        hmrApply(global.require, asset);
      });

      data.assets.forEach(function (asset) {
        if (!asset.isNew) {
          hmrAccept(global.require, asset.id);
        }
      });
    }

    if (data.type === 'reload') {
      ws.close();
      ws.onclose = function () {
        window.location.reload();
      }
    }

    if (data.type === 'error-resolved') {
      console.log('[parcel] ✨ Error resolved');
    }

    if (data.type === 'error') {
      console.error('[parcel] 🚨  ' + data.error.message + '\n' + 'data.error.stack');
    }
  };
}

function getParents(bundle, id) {
  var modules = bundle.modules;
  if (!modules) {
    return [];
  }

  var parents = [];
  var k, d, dep;

  for (k in modules) {
    for (d in modules[k][1]) {
      dep = modules[k][1][d];
      if (dep === id || (Array.isArray(dep) && dep[dep.length - 1] === id)) {
        parents.push(+k);
      }
    }
  }

  if (bundle.parent) {
    parents = parents.concat(getParents(bundle.parent, id));
  }

  return parents;
}

function hmrApply(bundle, asset) {
  var modules = bundle.modules;
  if (!modules) {
    return;
  }

  if (modules[asset.id] || !bundle.parent) {
    var fn = new Function('require', 'module', 'exports', asset.generated.js);
    asset.isNew = !modules[asset.id];
    modules[asset.id] = [fn, asset.deps];
  } else if (bundle.parent) {
    hmrApply(bundle.parent, asset);
  }
}

function hmrAccept(bundle, id) {
  var modules = bundle.modules;
  if (!modules) {
    return;
  }

  if (!modules[id] && bundle.parent) {
    return hmrAccept(bundle.parent, id);
  }

  var cached = bundle.cache[id];
  if (cached && cached.hot._disposeCallback) {
    cached.hot._disposeCallback();
  }

  delete bundle.cache[id];
  bundle(id);

  cached = bundle.cache[id];
  if (cached && cached.hot && cached.hot._acceptCallback) {
    cached.hot._acceptCallback();
    return true;
  }

  return getParents(global.require, id).some(function (id) {
    return hmrAccept(global.require, id)
  });
}
},{}]},{},[0,2])
```

shawwn:
It's hard to reverse-engineer the problem without the sources; it looks like you're requiring test.ts? If so, you'll need to make a tsconfig file.

If you send me your input files, I'll try running this locally.

test/elm.js
```diff
@@ -0,0 +1,14 @@
+const assert = require('assert');
+const {bundle, run} = require('./utils');
+
+describe('elm', function() {
+  it('should produce a bundle', async function() {
+    let b = await bundle(__dirname + '/integration/elm/index.js');
```
shawwn: Oh. This won't work. Create an index.html file that looks like this:

```
<html>
<body>
</body>
</html>
```

Then change that line to

```
 +    let b = await bundle(__dirname + '/integration/elm/index.html');
```

shawwn: Ignore that, sorry.

rtorr:
It’s a mix of a few files, just to test a few different ones. The interesting bit is `module.exports="/dist/74a93514c957e96dddff8f578e33c5f9.elm";`

Just used the example index.html, imported the elm file from index.js.

rtorr:
If that still does not make sense, I’ll post some files when I’m back at a computer.

rtorr:
Ok, I was able to test this on a clean machine, I think my other computer had issues linking my local copy of `parcel` to really test this.

IT WORKS!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!

rtorr:
```html
<!doctype html>
<html>
<head>
<meta charset="utf-8">

</head>
<body>
  <div id="main">
  </div>
</body>
</html>
```

```javascript
// index.js
import('./main.elm').then(Elm => {
  var mountNode = document.getElementById('main');
  var app = Elm.Main.embed(mountNode);
});
```
```elm
import Html exposing (..)
import Html.Events exposing (onClick)

model = 0

view model =  div [] 
    [ button [onClick Decrement] [text "-"] 
    ,text (toString model)
    , button [onClick Increment] [text "+"]
    ]

type Message = Increment | Decrement

update msg model = 
    case msg of
        Increment -> model + 2
        Decrement -> model - 1

main = beginnerProgram
        {
            model = model,
            view = view,
            update = update
        }
```

rtorr:
This branch should be ready now 💃

rtorr:
After further testing, there are a few issues:

* If elm-make is not installed in your machine, this silently fails

Will look in to this soon

package.json
```diff
@@ -44,10 +44,12 @@
   },
   "devDependencies": {
     "babel-cli": "^6.26.0",
+    "babel-eslint": "^8.1.2",
```
devongovett: needed?

src/assets/ElmAsset.js
```diff
@@ -0,0 +1,43 @@
+const JSAsset = require('./JSAsset');
+const localRequire = require('../utils/localRequire');
+const process = require('process');
+
+// Currently no pretty way to install Elm tools on behalf of the user :/
+async function compile(name, options) {
```
devongovett: Can we just inline this function into `parse` below?

src/assets/ElmAsset.js
```diff
@@ -0,0 +1,43 @@
...
+  try {
+    const elmCompiler = await localRequire('node-elm-compiler', name);
+    // compileToStringSync does not throw an Error, even when there is one.
+    // Other methods will throw an error, but do not bubble up to Parcel
+    const data = elmCompiler.compileToStringSync(name, options);
+    if (!data) {
+      throw 'You either have an error in your code, or are missing the Elm tools here https://guide.elm-lang.org/install.html.';
```
devongovett: Would be nice to have an actual error if possible from the compiler... you mentioned something about that in your comment. What issue did you have?

Either way, don't throw a string. Throw an Error object, for stack traces. `throw new Error('...')`

src/assets/ElmAsset.js
```diff
@@ -0,0 +1,43 @@
...
+    }
+    return data;
+  } catch (error) {
+    throw error;
```
devongovett: Why cat the error at all if you are just going to throw it again?

src/assets/ElmAsset.js
```diff
@@ -0,0 +1,43 @@
...
+  }
+}
+
+class ElmAsset extends JSAsset {
+  async getParserOptions() {
+    const defaultOptions = {
+      yes: true,
```
devongovett: is this really an option?

Dtatoo: `yes` option is to reply "yes" to all automated prompts.
Here are the full list of options available for node-elm-compiler.
```
var defaultOptions  = {
  emitWarning: console.warn,
  spawn:      spawn,
  cwd:        undefined,
  pathToMake: undefined,
  yes:        undefined,
  help:       undefined,
  output:     undefined,
  report:     undefined,
  warn:       undefined,
  debug:      undefined,
  verbose:    false,
  processOpts: undefined,
};
```
Is there any way to customize these options?

src/assets/ElmAsset.js
rtorr: Sorry, almost all of these are updated in a local branch with your concerns. I was hopeful that the WIP label made it obvious that this code is trash atm.

src/assets/ElmAsset.js
rtorr: Yep, have this fixed, need to push my latest. The elm compilers errors will still make it through to the console, but there is a case where a missing dependency does not bubble up an error :/

src/assets/ElmAsset.js
rtorr: Will do

src/assets/ElmAsset.js
Dtatoo: @rtorr I've created a pull request on your master regarding to adding user options to `node-elm-compiler` https://github.com/rtorr/parcel/pull/1

rtorr: accepted

src/assets/ElmAsset.js
```diff
@@ -0,0 +1,45 @@
...
+
+  async parse() {
+    const options = await this.getParserOptions();
+
+    const elm = await localRequire('node-elm-compiler', this.name);
+
+    try {
+      this.contents = await elm.compileToString(this.name, options);
+    } catch (err) {
+      throw new Error(err);
```
Dtatoo: I am not sure how I can write a test for this case. Would someone able to provide some insight on how I can write a test for this Error object?

Dtatoo:
I am unsure of why CI is failing. I've pull the master into my working branch and ran `npm test` and it seems to be working in order (all tests are passing).

goldenMetteyya:
how far off is this?
thanks:)

goldenMetteyya:
Hello,

so on my Mac I downloaded and installed this branch and ran the tests and it all passes:) I am able to link the demo to that copy and it works. The only thing is that when modifications are made on the elm files it doesn't trigger a recompile unlike the html file which does. I am not familiar with parcel but I'm sure there is a way to hook into their watched files changes and force elm to recompile again

src/assets/ElmAsset.js
```diff
@@ -0,0 +1,45 @@
...
+      );
+    }
+
+    // Use a local copy of elm-make when in test
+    if (process.env.NODE_ENV === 'test') {
+      defaultOptions.pathToMake = require.resolve('elm/binwrappers/elm-make');
+    }
+    return defaultOptions;
+  }
+
```
splodingsocks: I think you'll want to override `getDependencies` here and provide Parcel with a list of Elm deps. I've made a PR for `parcel-plugin-elm` that does just that right here:

https://github.com/ssuman/parcel-plugin-elm/pull/3#issuecomment-357263751

devongovett:
@rtorr how do you feel about finishing this up? Maybe you can collaborate with @ssuman to integrate some of his work on parcel-plugin-elm?

rtorr:
I looked at his work for prior art. I just haven’t had the time to get back to this. I’d love for this to get done. I’ll try for next weekend. Don’t feel bad about closing if this gets too stale.

robotarmy:
exciting work, commenting to monitor thread.

PR #446, opened 2017-12-31
Add support for the action attribute of the HTML form tag

jgkim:
Since the `action` attribute of the `form` tag can be a relative path, it should be added to the list of all HTML attributes that should produce a dependency in `HTMLAsset.js`.

### DIFF
--- a/src/assets/HTMLAsset.js
+++ b/src/assets/HTMLAsset.js
@@ -20,7 +20,8 @@ const ATTRS = {
     'embed'
   ],
   href: ['link', 'a'],
-  poster: ['video']
+  poster: ['video'],
+  action: ['form']
 };
 
 class HTMLAsset extends Asset {


devongovett:
Doesn't this usually point to a server endpoint, not another HTML file? Parcel won't be able to handle those.

jgkim:
@devongovett Yes. The `action` attribute usually points to a server endpoint, but I guess it could also point to another client-side page or program. In a very rare case, there might be someone, who wants to send the form data to another non-SPA page using the `GET` method. In which case, the URI parameters would be handled in the client-side code. Yes, it would be very rare, but I think it could be possible.

devongovett:
What is your usecase for this? Seems to me like it will break things more often than not, e.g. if you had `action="submit.php"`, parcel is not going to know what to do with that.

DeMoorJasper:
I don't think this should be part of Parcel, it would cause more issues than it'll solve.

jgkim:
I agree that this could be used in some rare use cases, and that this would make issues. I’m closing this PR. Actually, my use case was just demonstrating the flow of some mocked pages without using Javascript.

Issue #409, opened 2017-12-27
When I use the --public-url option, There is a problems with 'https://'


limingyang2012:
**Choose one:** is this a 🐛 bug report or 🙋 feature request?

### 😯 Current Behavior

my code is:

```js

  --public-url https://a.b.com/c.d.com/storage/projectname/

```

but I get: 

```js

  https:/a.b.com/c.d.com/storage/projectname/****

```

there is just one '/' after https:

### 🤔 Expected Behavior

I need get: 

```js

  https://a.b.com/c.d.com/storage/projectname/****

```

### 💁 Possible Solution

I think the code in `urlJoin.js` is wrong,

```js

const url = require('url');
const path = require('path');

module.exports = function(publicURL, assetPath) {
  // Use url.resolve to normalize path for windows
  // from \path\to\res.js to /path/to/res.js
  return url.resolve(path.join(publicURL, assetPath), '');
};

```

examples:

`url.resolve(path.join('https://www.baidu.com', '/df'), '')` => `https:/www.baidu.com/df`

`path.join('https://www.baidu.com', '/df')` => `https:/www.baidu.com/df`

`url.resolve('https://www.baidu.com', '/df')` => `https://www.baidu.com/df`

1120340041:
🐛

brandon93s:
Hello @limingyang2012 

The `--public-url` is intended to be a path and not a fully qualified URI. The [documentation](https://parceljs.org/production.html#set-the-public-url-to-serve-on) could probably use some clarifications. This option is used to alter the resource path, relative to your domain (e.g. `--public-url ./storage/projectname`.

1120340041:
@brandon93s Sometimes we may have multiple domain names, and then we want to js and css assets into another domain name, so we want to  a fully qualified URI

devongovett:
Should be fixed by #447.

1120340041:
@devongovett hi，parcel v1.4.1 is not solved this problem

Issue #405, opened 2017-12-26
sources url in the packaged html files is incorrect


JZLeung:
my project has multiple pages

index.html can jump to second.html

if i set the `--public-url` as `./`, the packged files can work well

but if i set the `--public-url` as a relative path, like `public`, the second.html will not be able to introduce the correct resources, because the js file is in `example.com/public/xxx.js` instead of `example.com/public/public/xxx.js`
```
GET http://192.168.2.3:8000/public/public/1fad9c454e4ab4186605464b35256085.js 404 (Not Found)
``` 
I think all the html files can be placed in the project root directory, not in the resource directory.

robinp:
Slightly related problem I stumbled upon (parcel 1.3.1).

```
$ cat index.html
<html>
<body>
 <script src="./output/Main/index.js"></script>
</body>
</html>
```
```
$ parcel --public-url 'http://rigo:8000' -p 8000 -d dist index.html
Server running at http://localhost:8000
Built in 4.70s.
```

```
$ curl 'http://rigo:8000/'
<html>
<body>
 <script src="http:/rigo:8000/4bfee3649edec6cdb33724f88316be5d.js"></script>
</body>
</html>
```

```
$ curl 'http:/rigo:8000/4bfee3649edec6cdb33724f88316be5d.js'
<html>
<body>
 <script src="http:/rigo:8000/4bfee3649edec6cdb33724f88316be5d.js"></script>
</body>
</html>
```

So the js path served in the main html is not correct (I tried other seemingly trivial paths, like `/dists/....js`, without success). If I try without `--public-url`, it works.

devongovett:
Should be fixed by #447.

PR #447, opened 2017-12-31
fix: when the publicUrl is url not a path of os

fansenze:
#409  
#405 

if we do, i can make a pr next

### DIFF
--- a/src/utils/urlJoin.js
+++ b/src/utils/urlJoin.js
@@ -9,5 +9,5 @@ module.exports = function(publicURL, assetPath) {
 
   return isUrl(publicURL)
     ? url.resolve(publicURL, assetPath)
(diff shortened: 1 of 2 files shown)


DeMoorJasper:
@fansenze some tests are failing, could you fix it so they pass?

devongovett:
This breaks when the url has a path in it:

```javascript
> require('url').resolve('http://test.com/foo', 'bar')
'http://test.com/bar'
```

I'd expect `http://test.com/foo/bar`.

fansenze:
@devongovett  
i fix and i write test

devongovett:
Simplified slightly to just always use URL.parse and URL.format, and use posix path joining for the paths.

Issue #357, opened 2017-12-20
🙋: Ignore certain code sections


agauniyal:
🙋

Sometimes there are code portions that I don't want parcel to parse and completely ignore them while acting normally on rest of the file. An example is embedding [monaco-editor](https://github.com/Microsoft/monaco-editor) where index.html code looks like -

```html
<!DOCTYPE html>
<html>
<head>
	<meta http-equiv="X-UA-Compatible" content="IE=edge" />
	<meta http-equiv="Content-Type" content="text/html;charset=utf-8" >
</head>
<body>

<div id="container" style="width:800px;height:600px;border:1px solid grey"></div>

<!-- I want this section to be ignored by parcel -->
<script src="monaco-editor/min/vs/loader.js"></script>
<script>
	require.config({ paths: { 'vs': 'monaco-editor/min/vs' }});
	require(['vs/editor/editor.main'], function() {
		var editor = monaco.editor.create(document.getElementById('container'), {
			value: [
				'function x() {',
				'\tconsole.log("Hello world!");',
				'}'
			].join('\n'),
			language: 'javascript'
		});
	});
</script>
<!-- Section End -->

<!-- This should work normally -->
<script src="./index.js"></script>

</body>
</html>
```

### 🤔 Expected Behavior

Parcel should be able to ignore certain sections.

### 😯 Current Behavior

It'll try parsing the file but would not do it the right way so browser errors out. This is because monaco-editor has a long standing issue here - https://github.com/Microsoft/monaco-editor/issues/18 and I manually copy node_modules/monaco-editor to `public` directory in my build step.

### 💁 Possible Solution

Use comments to inform parcel.js to ignore these sections(like eslint does)..

### 🔦 Context

Showstopper for me because I can't migrate to parcel.

### 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Parcel           | 1.2.1
| Node             | 9.3.0
| npm/Yarn         | 1.3.2
| Operating System | Arch Linux

rondonjon:
Sounds very much like an edge-case to me. You could achieve the same with a custom preprocessor.

shawwn:
During html parsing, we could skip nodes that set the attribute `data-parcel="false"`.

That would be a decent design choice because we could extend it to `data-parcel="process.env.NODE_ENV === 'production'"` if necessary.

jdanyow:
[...]

Issue #333, opened 2017-12-18
Some dependencies that rely on window don't work


mix3d:
BUG: 
An older website that imports jQuery and Bootstrap from locally downloaded files don't have their dependencies properly running; Bootstrap complains that `jQuery` and `$` are not found. It seems that either the order from the HTML is not respected, or that JS that installs to the root `window` are not working correctly. 

### 🎛 Configuration (.babelrc, package.json, cli command)

```html
        <script src="LIBS/jquery.min.js"></script>
        <script src="LIBS/bootstrap.min.js"></script>
```
`$> parcel index.html`

### 🤔 Expected Behavior
Imported scripts run correctly

### 😯 Current Behavior
Bootstrap complains about not finding jquery. Not sure if import order or `window` problem

### 🔦 Context
Can't use Parcel if this doesn't work for older / retrofit websites.

### 🌍 Your Environment

| Software         | Version(s)
| ---------------- | ----------
| Parcel           |   1.2.0
| Node             |  v8.9.1
| npm/Yarn         |  npm v5.5.1
| Operating System |  Windows 10

benhutton:
We're seeing a lot of this, too, trying to get some legacy code working with Parcel

brandon93s:
This seems like a valid use case for a `parcelignore`. They're being included in the correct order, but I'm guessing the closure scope is preventing them from mounting to the window correctly.

benhutton:
some of my problems here seem to go away by using jquery off of CDN instead of bundling in a node package. Probably more performant that way, anyways... Still investigating to see if that fixes everything.

mix3d:
Sometimes corporate policy requires self hosting all dependencies

On Dec 27, 2017 10:01 AM, "Ben Hutton" <notifications@github.com> wrote:

some of my problems here seem to go away by using jquery off of CDN instead
of bundling in a node package. Probably more performant that way,
anyways... Still investigating to see if that fixes everything.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/parcel-bundler/parcel/issues/333#issuecomment-354125970>,
or mute the thread

wywywywy:
Hi.

I tried using a CDN, a local link, and node modules, nothing works.

This is a very simple HTML that I used to test.

```
<!DOCTYPE html>
<html lang="en">
<head>
	<title>Test</title>
	<meta charset="utf-8">
	<meta name="viewport" content="width=device-width, initial-scale=1">
[...]

Issue #141, opened 2017-12-08
🐛 @webcomponents Cannot resolve dependency 'vertx'


sbesh91:
**Choose one:** is this a 🐛 bug report or 🙋 feature request? 🐛

I've been giving parcel a go as a replacement for my webpack config.
It's running into the same issue I had with webpack where it had an issue with the webcomponents polyfill
### 🎛 Configuration (webpackconfig)
This is the line that would fix a webpack config
```js
{
    new webpack.IgnorePlugin(/vertx/)
}
```

### 🤔 Expected Behavior
The expectation is for a way to ignore vertx

### 😯 Current Behavior
@webcomponents/webcomponentsjs/webcomponents-lite.js:27:271: Cannot resolve dependency 'vertx'
That is the error message I get upon running a build

### 🔦 Context
I'd like to replace my webpack config with parcel if at all possible.  This is the error that is currently holding me back because my component model heavily relies on web components

### 🌍 Your Environment

| Software         | Version(s)
| ---------------- | ----------
| Parcel           |  1.0.3
| Node             | 6.12.0
| npm/Yarn         | 3.10.10
| Operating System | ubuntu 17.10

davidnagli:
So put more generally, you're asking for a way to ignore certain file name patterns in Parcel?

sbesh91:
Yeah I think that sounds right to me.

davidnagli:
I think we need to implement a CLI option called `--ignore`

It should work like this (once implemented):
```
parcel --ignore *.vertx
```

Do you guys think we should use wildcards for the ignore, or regex?

```
parcel --ignore /\.vertex/
```

I guess we could also support both...

DeMoorJasper:
It would probably be a good idea to allow multiple entries, for example if anyone would use a minified library (for example jQuery) and not want it to get included in the bundle.
This is kinda already included in babelrc though, don't know if it is being used in the source codes though
```json
{
  "ignore": [
    "foo.js",
    "bar/**/*.js"
  ]
}
```

devongovett:
why is `vertx` required if it isn't a dependency?

sbesh91:
The original answer I got to this problem for my webpack solution came here.  
https://github.com/webcomponents/webcomponentsjs/issues/794
I can't say vertx is a dependency of my app.

chrismbeckett:
[...]

PR #448, opened 2017-12-31
[WIP/RFC] .parcelignore

DeMoorJasper:
So i gave ``.parcelignore`` a shot because it's been requested in some issues.

Some questions
* Should the ignore, ignore the files completely or include them as is without parsing?
* ~Currently i'm having issues by not wanting to read the ``.parcelignore`` synchronous and seeing that all getDependencies/addDependency logic is synchronous, so if i would change it into an asynchronous function I would change a lot of core logic which is not really a good idea I think. Is there a better way to handle this or should i write ``.parcelignore`` logic to be synchronous?~
* Does this collide with the whole point of 0-config?

Progress:
* [x] Create ignoreChecker following gitignore logic
* [x] ~Find the best way to alter getDependencies~ Should end up in the resolver
* [ ] Ignore files that are flagged by ignoreChecker

related #141 #357 #333


src/utils/checkIgnore.js
```diff
@@ -0,0 +1,23 @@
+const ignore = require('ignore');
+const path = require('path');
+const config = require('./config');
+const fs = require('./fs');
+
+async function getIgnoreFile(location) {
+  let fileLocation = await config.resolve(location, ['.parcelignore']);
+  let ignoreFile = '';
+  if (fileLocation) ignoreFile = (await fs.readFile(fileLocation)).toString();
```
sabeurthabti: maybe some sort of error handling here? (`try/catch`)

devongovett:
Good questions! Also related: #474. Not sure I have the answers yet. I think we need a bit more input from others on how this should work.

On your second point, I don't think I'd do this in `addDependency`. I think it's ok if the dependency is still parsed. Instead, you could check in the Resolver, and return null if the file is ignored. That will cause a few other changes to check for null in Bundler, but should be easier than trying to change the logic in every asset type.

thejameskyle:
I would like to discuss this in the context of a more complete (`.parcelrc`) config format. I think it's possible that this would not be necessary.

mikkiek:
Hello,

Any ideas when can i expect parcelignore "miracle"?

agauniyal:
In my case(#357) I wanted to ignore certain sections, not the whole file itself 😅, would that be possible with this PR?

mathiasrw:
A central use case would be to link from your HTML to a legacy entry point. 

Let's say you have a nice static website up and running with parceljs but you would like to link to your crazy old PHP3 system located at `/login` - at the current version `<a href="/login">login</a>` will result in an error because it cant find `/../../../../../login` (at the root of the drive)

I suggest adding `/login` to the `.parcelignore` would ignore it as a dependency and just move along without any output.

DeMoorJasper:
@mathiasrw I just closed it due to inactivity if parcelignore ever gets approved i’ll just make a new pr

dinvlad:
+1 on special URLs used by Firebase

goyney:
Would like to see this reopened. Here's a PERFECT use case for it: https://github.com/moment/moment/issues/2517#issuecomment-393704231

DeMoorJasper:
Please do not comment on an old closed PR, open up a new issue instead.

PR #449, opened 2017-12-31
Make bundler middleware usage intuitive

### LINKED ISSUE #442: Bundler middleware throws a TypeError when trying to serve bundle
🐛 bug report 🐛

Not using custom babel configuration.

### 🤔 Expected Behavior

Using bundler middleware with Express server should serve the bundled file on request.

### 😯 Current Behavior

Bundler middleware throws the following TypeError when it tries to serve the file:

```
TypeError: Cannot read property 'type' of undefined
    at sendIndex (/home/fluidbyte/workspace/sandbox/react-parcel/node_modules/parcel-bundler/src/Server.js:33:29)
    at respond (/home/fluidbyte/workspace/sandbox/react-parcel/node_modules/parcel-bundler/src/Server.js:23:16)
    at /home/fluidbyte/workspace/sandbox/react-parcel/node_modules/parcel-bundler/src/Server.js:15:7
    at Layer.handle [as handle_request] (/home/fluidbyte/workspace/sandbox/react-parcel/node_modules/express/lib/router/layer.js:95:5)
    at trim_prefix (/home/fluidbyte/workspace/sandbox/react-parcel/node_modules/express/lib/router/index.js:317:13)
    at /home/fluidbyte/workspace/sandbox/react-parcel/node_modules/express/lib/router/index.js:284:7
    at Function.process_params (/home/fluidbyte/workspace/sandbox/react-parcel/node_modules/express/lib/router/index.js:335:12)
    at next (/home/fluidbyte/workspace/sandbox/react-parcel/node_modules/express/lib/router/index.js:275:10)
    at expressInit (/home/fluidbyte/workspace/sandbox/react-parcel/node_modules/express/lib/middleware/init.js:40:5)
    at Layer.handle [as handle_request] (/home/fluidbyte/workspace/sandbox/react-parcel/node_modules/express/lib/router/layer.js:95:5)
```

The middleware [attempts to check the `type` prop on `bundler.mainAsset`](https://github.com/parcel-bundler/parcel/blob/c8a1156d4c2dd5902ba500aa9c25547bfab53eac/src/Server.js#L34), but `bundler.mainAsset` will be undefined [until the bundler bundles at least once](https://github.com/parcel-bundler/parcel/blob/c8a1156d4c2dd5902ba500aa9c25547bfab53eac/src/Bundler.js#L136).

### 💁 Possible Solution

I was able to work around this by calling `await bundler.bundle()`, [which defines the `mainAssset` property here](https://github.com/parcel-bundler/parcel/blob/c8a1156d4c2dd5902ba500aa9c25547bfab53eac/src/Bundler.js#L136), before using the middleware. A fix would involve at least defining the `bundler.mainAsset` property when using middleware like the current implementation does when calling `await bundler.bundle()`.

### 🔦 Context

I was trying to create a setup where I can run both my server and client at the same time and have requests from the client get to the server without having to create any proxies or hardcode URLs (I use relative `/path/to/api` paths instead). The solution was to run my regular express server and have it connect the Parcel middleware when in development mode.

### 💻 Code Sample

This does not work:

```ts
if (Constants.ISLIVE) {
    // Redirect http requests to https when live
    app.use(httpsRedirect());
} else {
    // Run Parcel bundler for client development
    const bundler = new Bundler(path.resolve(__dirname, "../client/index.html"))
    app.use(bundler.middleware());
}
```

But by calling `await bundler.bundle()` before using the middleware, this *does* work:

```ts
if (Constants.ISLIVE) {
    // Redirect http requests to https when live
    app.use(httpsRedirect());
} else {
    // Run Parcel bundler for client development
    const bundler = new Bundler(path.resolve(__dirname, "../client/index.html"))
    // Bugfix: parcel bundler must call bundle at least once before using middleware
    await bundler.bundle();
    app.use(bundler.middleware());
}
```

### 🌍 Your Environment

| Parcel           | 1.3.1 |
| Node             | 8.9.3 |
| npm/Yarn         | 5.5.1/1.3.2 |
| Operating System | Windows 10 FCU |

kootoopas:
Fixes #442.

### DIFF
--- a/src/Bundler.js
+++ b/src/Bundler.js
@@ -473,6 +473,7 @@ class Bundler extends EventEmitter {
   }
 
   middleware() {
+    this.bundle();
     return Server.middleware(this);
   }
 
--- a/test/bundler.js
+++ b/test/bundler.js
@@ -1,0 +1,14 @@
+const assert = require('assert');
+const {bundler} = require('./utils');
+
+describe('bundler', function() {
+  it('should bundle once before exporting middleware', function(done) {
+    let b = bundler(__dirname + '/integration/commonjs/index.js');
+    b.middleware();
+
+    setTimeout(() => {
+      assert(b.mainAsset);
+      done();
+    }, 300);
+  });
+});
--- a/test/bundler.js
+++ b/test/bundler.js
@@ -3,7 +3,7 @@ const {bundler} = require('./utils');
 
 describe('bundler', function() {
   it('should bundle once before exporting middleware', function(done) {
-    let b = bundler(__dirname + '/integration/commonjs/index.js');
+    let b = bundler(__dirname + '/integration/bundler-middleware/index.js');
     b.middleware();
 
     setTimeout(() => {
--- a/test/integration/bundler-middleware/index.js
+++ b/test/integration/bundler-middleware/index.js
@@ -1,0 +1,5 @@
+var local = require('./local');
+
+module.exports = function () {
+  return local.a + local.b;
+};
--- a/test/integration/bundler-middleware/local.js
+++ b/test/integration/bundler-middleware/local.js
@@ -1,0 +1,2 @@
+exports.a = 1;
+exports.b = 2;


kootoopas:
Hey there! I didn't use `await` for `this.bundle()` because that would require `bundler.middleware()` to be called with `await`.

We shouldn't care about waiting for `this.bundle()` to finish anyway since middleware returns a handler function that is fired on requests.

src/Bundler.js
```diff
@@ -473,6 +473,7 @@ class Bundler extends EventEmitter {
   }
 
   middleware() {
+    this.bundle();
```
DeMoorJasper: You should probably also remove this.bundle() from the server function as this would cause bundle to be called twice on built-in server

kootoopas: @DeMoorJasper Hmm, I think that `Server.serve` calls `Server.middleware` and not `Bundler.middleware`. Therefore bundle is never called twice.

kootoopas: @DeMoorJasper Plus the `serve` in `Server.middleware()` just uses the `serve-static` pkg. :)

DeMoorJasper: Ow damn sorry for the mixup, looks good than, kinda weird that one of the tests is failing though

kootoopas: Indeed, considering that it's not related to the changes:
```
  99 passing (33s)
  1 failing
  1) css should produce two bundles when importing a CSS file:
     evalmachine.<anonymous>:87
},{"./local":9,"./index.css":14}]},{},[5]) module == 'object' && module &&
                                           ^^^^^^
SyntaxError: Unexpected identifier
      at createScript (vm.js:80:10)
      at Object.runInContext (vm.js:119:10)
      at run (test\utils.js:90:6)
      at Context.<anonymous> (test\css.js:24:18)
      at <anonymous>
```

kootoopas:
@DeMoorJasper Turns out that it was my test that was messing up the css one. 😅 You can read the commit description for more info... Also, Happy New Year's! 🎉

Issue #279, opened 2017-12-15
🙋 Allow to define a specific port for HMRServer


real34:
I am trying to use Parcel inside a docker container and could not make HMR work with the hardcoded `port: 0` value, since the Websocket server selects a random available port every time.

https://github.com/parcel-bundler/parcel/blob/333c3aa5d20f98a5f3c52635751032d12854c13c/src/HMRServer.js#L7

Hardcoding a port value here and exposing it through Docker allows to make HMR work, but it would mean adding a configuration.

What I would need is to find a way to expose the ws port used on the host machine, or find a better way to connect to the socket from the client.

Would you agree to add a new option similar to the current `-p` for the server port? What name would you recommend?

If that sound good to you, I could try to do a PR for this (inspired by #278 for instance)

davidnagli:
I’m trying to understand why you would need to specify the ws port for the HMRServer, if the HMR client handles everything itself. What kind of issue do you get when you use the default port?

mbrobbel:
When running multi-container docker applications with docker-compose, you explicitly expose specific ports to other services (https://docs.docker.com/compose/compose-file/#expose) and to the host (https://docs.docker.com/compose/compose-file/#ports).
In order to make sure that the clients are able to connect to a HMRServer running in a docker container, one would have to expose all ports for that container, which is considered bad practice.

Fluidbyte:
I have the same problem with docker (I build and test all my apps in containers). Additionally if I wanted to serve my WIP I would have to open a massive range of ports on my firewall, whereas if I could set this I could explicitly define a single port.

Fluidbyte:
^ PR for review

devongovett:
Closing. Should be fixed by #450.

thedull:
Hi guys, any plans to release soon? I'm in desperate need of this feature.

brandon93s:
@thedull - Should be a release next week.

marcklei:
If anyone else is looking for this option. It's implemented but with the following args `-h, --hmr-port <port>`

moonlight16:
I'm getting this error:

```
> docker_web_app@1.0.0 start /usr/src/app
> parcel index.html --hmr-port 1234

Server running at http://localhost:1234 
⠋ Building...events.js:183
      throw er; // Unhandled 'error' event
      ^

Error: listen EADDRINUSE :::1234
    at Server.setupListenHandle [as _listen2] (net.js:1360:14)
[...]

PR #450, opened 2017-12-31
feat(hmr-ws-port): adds ability to explicitly set hmr ws port

Fluidbyte:
This PR is to address issues from #279 in allowing a specific port to be set for the HMR websocket. This allows the use, either in containers, behind firewalls, or in other circumstances where allowing broad port access is not viable, to specifically set the port:

```
-w, --websocket-port <port>    set the port to serve HMR websockets, defaults to random
```

Ensured _non-breaking_ by defaulting to `0` as was previously hard-coded into the WS and HMR configuration.


src/cli.js
```diff
@@ -13,6 +13,11 @@ program
     'set the port to serve on. defaults to 1234',
     parseInt
   )
+  .option(
+    '-w, --websocket-port <port>',
```
devongovett: can we call the option `--hmr-port` so it's clear what websocket is being referred to?

src/HMRServer.js
```diff
@@ -2,9 +2,9 @@ const WebSocket = require('ws');
 const prettyError = require('./utils/prettyError');
 
 class HMRServer {
-  async start() {
+  async start(websockerPort) {
```
devongovett: misspelled. just call it `port`.

src/cli.js
Fluidbyte: yeah, do you want it to be `-h, --hmr-port`?

devongovett: sounds good.

PR #453, opened 2017-12-31
Expose entry points exports

eXon:
The way the code is bundled is perfect for the browser. It is also perfect for Node.js if you want to do server-side rendering. However, it was impossible to expose functions that you could import from Node.js by requiring the bundle. The expected behaviour would be to access the entry point exports. Here is an example of what I was trying to accomplish:

test.js
```js
module.exports = function test(req) {
  req.send();
};
```

index.js
```js
const req = {
  send() {
    console.log('MEOW!');
  }
};

const test = require('./dist/test');
test(req);
```

 - Step 1: bundling: `parcel build test.js`
 - Step 2: run `node index`

The expected behaviour would be to have printed on the console `MEOW!`. Right now, it would throw `test is undefined`.

There is many benefits of working this way when you do server-rendering. You still have zero-configuration. The hashed asset names are already converted within the bundle. And many more.

I am not sure if there is only 1 entry point and it is always the first. Maybe someone that knows the code better would be able to confirm if this approach is correct.

### DIFF
--- a/src/builtins/prelude.js
+++ b/src/builtins/prelude.js
@@ -7,7 +7,7 @@
 // orig method which is the require for previous bundles
 
 // eslint-disable-next-line no-global-assign
-parcelRequire = (function (modules, cache, entry) {
+parcelRequire = (function (modules, cache, entry, globalName) {
   // Save the require from previous bundle to this closure if any
   var previousRequire = typeof parcelRequire === 'function' && parcelRequire;
   var nodeRequire = typeof require === 'function' && require;
@@ -73,6 +73,27 @@ parcelRequire = (function (modules, cache, entry) {
 
   for (var i = 0; i < entry.length; i++) {
     newRequire(entry[i]);
+  }
+
+  if (entry.length) {
+    // Expose entry point to Node, AMD or browser globals
+    // Based on https://github.com/ForbesLindesay/umd/blob/master/template.js
+    var mainExports = newRequire(entry[entry.length - 1]);
+
+    // CommonJS
+    if (typeof exports === "object" && typeof module !== "undefined") {
+      module.exports = mainExports;
+
+    // RequireJS
+    } else if (typeof define === "function" && define.amd) {
+     define(function () {
+       return mainExports;
+     });
+
+    // <script>
+    } else if (globalName) {
+      this[globalName] = mainExports;
+    }
   }
 
   // Override the current require with this new one
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -11,6 +11,7 @@
     "es6": true
   },
   "globals": {
-    "parcelRequire": true
+    "parcelRequire": true,
+    "define": true
   }
 }
--- a/src/Bundler.js
+++ b/src/Bundler.js
@@ -99,6 +99,7 @@ class Bundler extends EventEmitter {
         options.hmrHostname ||
         (options.target === 'electron' ? 'localhost' : ''),
       detailedReport: options.detailedReport || false,
+      global: options.global,
       autoinstall:
         typeof options.autoinstall === 'boolean'
           ? options.autoinstall
--- a/src/cli.js
+++ b/src/cli.js
@@ -41,6 +41,7 @@ program
     '--public-url <url>',
     'set the public URL to serve on. defaults to the same as the --out-dir option'
   )
+  .option('--global <variable>', 'expose your module through a global variable')
   .option('--no-hmr', 'disable hot module replacement')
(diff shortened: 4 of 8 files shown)


eXon:
@davidnagli Does it align with Parcel being just a simple bundler? I've tried hard to make sure there is 0 side-effect on any environment (browser, node.js, electron, etc).

It would make server-rendering much easier. The only missing piece after this would be supporting code-splitting.

shawwn:
Hi @eXon,

Interesting PR. I'm trying to run it using these test files: https://github.com/parcel-bundler/parcel/commit/900e034ca161ac416de8214cb34209fef75a942b

Do those files look correct? When I run `cd test && ../bin/cli.js build --no-cache integration/node/index.js && node dist/index.js`, I get no output.

If you have Slack, hop on and shoot me a message, or join the #contributors channel.  https://slack.parceljs.org/

eXon:
Hey @shawwn 

I've updated the PR with a more appropriate integration test. The goal is not to bundle the whole thing for Node.js but rather make the exports on the entry point available when you require the bundle from Node.js.

If you run within the `node-exports` folder `npm run build && npm start` it should print `Test!` within the console.

eXon:
@shawwn After testing exactly your code, I can see the `MEOW!` only if I disable minifying. When minification is enabled, the `console.log` is stripped away.

screen shot 2018-01-07 at 11 14 43 am


Here is the code with minification enabled:
```js
const e={send(){}},s=require("./test");s(e);
```

However, in my test the `console.log` comes from Node.js node the bundle.

eXon:
Anything else I can do to make this PR merged?

eXon:
@shawwn anything preventing this PR from getting merged? I know everyone is probably busy but I think this would improve parcel and shouldn't cause any problem.

Thanks!

src/builtins/prelude.js
```diff
@@ -34,12 +37,18 @@ require = (function (modules, cache, entry) {
...
+
       localRequire.resolve = resolve;
 
       var module = cache[name] = new newRequire.Module;
 
       modules[name][0].call(module.exports, localRequire, module, module.exports);
+
+      // Expose the exports if we are on the entry point by using the
+      // previous defined module.
+      if (previousModule && name === 1) {
```
lbguilherme: Checking `name === 1` would break if we ever change the way IDs are generated (such as #780). Not sure how else this can be checked, though. @devongovett  ?

devongovett: Yeah. Also in split bundles, the entry module does not have `id === 1`. You can look in the `entry` argument passed to this function above to get the entry module for the bundle. It is an array, but the last element will be the id of entry module. See here: https://github.com/parcel-bundler/parcel/blob/595252c4135c9f9d6cbb8aa5f3f2e94771cc9712/src/packagers/JSPackager.js#L197-L199.

test/integration/node-exports/src/app.js
```diff
@@ -0,0 +1,5 @@
+function middleware(req) {
+  req.send('Test!');
+}
+
+module.exports = middleware;
```
devongovett: Thanks for adding these fixtures! Can you also add a test that uses them in `test/javascript.js`?

devongovett:
Perhaps we should just support generating a UMD bundle instead of something specific for node? That would make it compatible with CommonJS, AMD, globals, etc. https://github.com/umdjs/umd

Here's a template: https://github.com/ForbesLindesay/umd/blob/master/template.js

eXon:
@devongovett After reading your comments I've added the missing test and ajusted the entry point to use the correct value. I've also managed to support RequireJS.

While I was reading about UMD examples, I also saw that in order to be UMD, we also need the global export. However, that exporrt needs a name. Just like jquery is on `window.jquery`, a bundle would be on `window.myBundleName`. However, as I was trying things out, I was wondering how could we name that? 

I feel like this is an impossible task with the current information the file has. @devongovett Do you have an insight about this?

Thanks!

eXon:
While I was playing with UMD, I think I found the most elegant solution. CommonJS and RequireJS would have the entry point exposed out of the box. Then, the only missing part to be 100% UMD is the browser globals.

By adding a new option `browser-global`, it will allow any author to expose his library by using a global variable and he can choose which one. For example, I'm the author of `react-cookie` and I want to expose the library on `window.reactCookie`, I would set `--browser-global reactCookie` and I would have a 100% UMD build for the browser or Node :).

@devongovett @shawwn @ntomsic @danawoodman
Let me know what you think!

ntomsic:
Any traction on this? This is a much needed feature.

PR #460, opened 2018-01-01
De-clutter root of the repository

yeskunall:
Moved `CONTRIBUTING.md` and `CODE_OF_CONDUCT.md` to `.github` to de-clutter the root of the repository.

The repo seems to be cluttered with dotfiles/configs ATM, so I think this is justified. 😅


brandon93s:
Won't this break the contribute link in the readme? The root is the standard location for these. Keeping them in the root helps people find them.

yeskunall:
@brandon93s I can update the link in the README. In general, these files _can_ go in the `.github` folder, as suggested by GitHub. But it's also okay to put them in the root of the repository 🤷‍♂️ and like I said, I only made this PR because I started seeing a lot of dotfiles/configs and all of that stuff.

PR #462, opened 2018-01-01
Logger singleton

DeMoorJasper:
Converts Logger into a singleton, so everything outside of bundler can use it.
Also created IPC for logger so that assets can send messages to the workerfarm, instead of overloading logger singleton.

After seeing #207 was too big of change and failing times and times again to fix the glitches with status i thought it be a better idea to just extract the singleton part of that PR and make a fresh clean new PR, without the conflicts, ...

### DIFF
--- a/src/Bundler.js
+++ b/src/Bundler.js
@@ -9,7 +9,7 @@ const FSCache = require('./FSCache');
 const HMRServer = require('./HMRServer');
 const Server = require('./Server');
 const {EventEmitter} = require('events');
-const Logger = require('./Logger');
+const logger = require('./Logger');
 const PackagerRegistry = require('./packagers');
 const localRequire = require('./utils/localRequire');
 const config = require('./utils/config');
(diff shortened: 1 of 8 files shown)


devongovett:
Sorry for the delay getting this in. I just cleaned up the IPC a little, so we aren't adding logging things to the Asset class. Instead, a proxy class is exported by the Logger in workers that sends IPC messages. That way you can use the normal logger "singleton" as usual but the messages will actually get sent to the main process. Seems to work well!

brandon93s:
@devongovett - clever idea with the proxy logger! 👍

ioss:
@devongovett this is a breaking change (not sure if this project follows semver?). 

I suggest to put the logger back as a member to the bundler; so plugins that use bundler.logger do not break.
Also the logger should be available as Bundler.Logger for new/updated plugins, so that there is no need for a plugin developer to do the src vs. lib check, which if done wrong (or if parcel should change it's check) could lead to working with two "Singletons" ;)

Issue #336, opened 2017-12-18
🙋 Support project-relative import paths


rondonjon:
Projects with a deep folder structure often have import paths with lots of "../../../../../a/b/c/d/actualFile.js" in them because import paths are kind of restricted.

One workaround in webpack is to define custom aliases for the resolver, but this approach would break the zero-configuration approach of parcel.

Would it be possible to have at least an additional notation for "project-relative" imports in parcel, so that the dotdot madness at the beginning could be stopped?

### 🎛 Configuration (.babelrc, package.json, cli command)

N/A

### 😯 Current Behavior

Apparently (??) no support for project-relative import paths.

### 💁 Possible Solution

- Resolve './abc' relatibe to the current file (as is)
- Resolve '/abc' absolute to filesystem (or webroot? as is)
- Resolve 'x/abc' relative to module 'x' (as is)
- Resolve '-/abc' or '.../abc' or similar relative to current or nearest parent folder with a package.json file in it (package root)

davidnagli:
Hmm I actually really like this idea!  😮 😃 👏

This would be **supper** useful

davidnagli:
>Resolve ‘-/abc’ or '.../abc' or similar relative to current or nearest parent folder with a package.json file in it (package root)

Not sure about the syntax for package root though…
* I don’t really like like  `-/abc` since `-` looks a lot like `~` (maybe this was intentional, but it could make the difference hard to spot)
* `…/abc` wouldn’t work either since it kinda implies “go back 3 directories”, or at least it does for anyone using ZSH as their terminal 0_O

**I propose using `@` to represent “package root”:**
```
@/abc
```
As far as I know the “@“ simple doesn’t have any reserved meaning in file paths (@brandon93s maybe it does on Windows?).
…

If not we could also use `#`:
```
#/abc
```

brandon93s:
`@` is good to go on Windows. This seems to be widely accepted as the "package root" character, especially among webpack users which would ease adoption.

psimoneau22:
I believe react-native(or one of the tools for it)allows you to define a package.json in any folder.   The 'name' property defined in that json file can then be used as an alias for the folder, as if it was in node_modules, just like webpack alias.

For example:   in app/someDir2/someDir2, i put a package.json file with name = 'myDep', now 'anywhere' in the folder hierarchy i can simply import ... from 'myDep/someFile', the same way webpack alias works.

psimoneau22:
[...]

PR #290, opened 2017-12-15
[RFC] Fix for absolute paths

DeMoorJasper:
Fixes absolute paths in html and other assets closes #86 
Removes the need for other project relative solutions closes #336 closes #466 

* Is able to detect and resolve absolute paths, unix, node_modules and project relative
* Supports both ~ and /

PR #466, opened 2018-01-02
[WIP] resolve '~/' to root path of project

fansenze:
####  resolve '~' to root path of project
i support use ~

#336

### DIFF
--- a/src/Resolver.js
+++ b/src/Resolver.js
@@ -9,6 +9,18 @@ class Resolver {
   constructor(options = {}) {
     this.options = options;
     this.cache = new Map();
+    this.pathIndicators = {
+      // use ~ instead of root path for project
+      '~': process.cwd()
+    };
+  }
+
+  resolvePath(filename) {
+    for (let [indicator, exactPath] of Object.entries(this.pathIndicators)) {
+      if (filename.startsWith(indicator))
+        return path.resolve(exactPath, filename.slice(1));
+    }
+    return filename;
   }
 
   async resolve(filename, parent) {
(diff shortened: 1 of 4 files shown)


fansenze:
@DeMoorJasper  
i think `/` may be relative path , also i think can use `~/`
`~` only related to version of package, has nothing to do with path

src/Resolver.js
```diff
@@ -9,6 +9,19 @@ class Resolver {
   constructor(options = {}) {
     this.options = options;
     this.cache = new Map();
+    this.pathIndicators = {
+      // resolve '~/' to root path of project
+      '~/': path.dirname(this.options.cacheDir)
```
DeMoorJasper: This would break if source is in another directory than where u executed the parcel command.
This kinda is the same as your previous cwd() implementation

src/Resolver.js
```diff
@@ -9,6 +9,19 @@ class Resolver {
...
+    };
+  }
+
+  resolvePath(filename) {
+    for (let [indicator, exactPath] of Object.entries(this.pathIndicators)) {
+      if (filename.startsWith(indicator)) {
+        return path.resolve(exactPath, filename.slice(indicator.length));
```
DeMoorJasper: Will this also handle urlImports in for example html?

fansenze: it cann't handle in html which is entry

src/Resolver.js
fansenze: I think most projects only need to run under their current directory as the root directory on it, because of the need for convenient reference to resource files

nehaleem:
I think, this is not needed. Isnt better just use https://github.com/tleunen/babel-plugin-module-resolver ?

DeMoorJasper:
@nehaleem a root resolver should be in Parcel by default, also for html assets which do need to support '/' but this PR could be easily achieved with `babel-plugin-module-resolver`.
You can have a look at #290

PR #471, opened 2018-01-02
Minify JSON

### LINKED ISSUE #399: 🐛 JSON files don't get minified

DeMoorJasper:
closes #400 closes #399


src/assets/JSONAsset.js
```diff
@@ -8,7 +9,15 @@ class JSONAsset extends JSAsset {
   parse() {}
   collectDependencies() {}
   pretransform() {}
-  transform() {}
+  async transform() {
+    if (this.options.minify) {
+      this.contents = minify(this.contents, {
```
kzc: Should check the `minify()` result for `error`.

https://github.com/mishoo/UglifyJS2/tree/harmony#api-reference

kzc: for example:

https://github.com/parcel-bundler/parcel/blob/68258b1e53d086f173d2888d9fa367fcba891807/src/transforms/uglify.js#L25-L28

brandon93s: https://github.com/parcel-bundler/parcel/blob/065a49e8f673922e514c5279d79df74f052a1558/src/transforms/uglify.js#L25-L36

brandon93s: Are we intentionally skipping any custom config for JSON?

https://github.com/parcel-bundler/parcel/blob/065a49e8f673922e514c5279d79df74f052a1558/src/transforms/uglify.js#L10

Seems a lot of this could be consolidated with the uglify transform.

kzc: `result.warnings` is not interesting in parsing JS expressions - `result.error` is sufficient as it's a binary pass/fail operation. The warnings apply to the uglify compressor which is not applicable here.

DeMoorJasper: @brandon93s i don't think anyone would want to change this behaviour, as it just removing some quotes and whitespace

DeMoorJasper: @kzc i'll throw the errors, forgot about those

devongovett:
Refactored slightly to make JSONAsset not inherit from JSAsset anymore. Seems like that had the potential to break more things than it helped with as we add more stuff to JSAsset that doesn't apply to JSON.

PR #297, opened 2017-12-15
Cache bugfixes/improvements

DeMoorJasper:
A potential way to fix most current caching issues, closes #66 
* Invalidate cache on plugin installs/updates
* Invalidate cache on Parcel update
* Invalidate cache on Config changes
* Invalidate cache on package namechange

Performance test (New build = this PR, old build = v1.2.1)
screen shot 2017-12-15 at 23 00 12

PR #473, opened 2018-01-02
WASM bundle loader

devongovett:
This refactors the way async bundle loaders work so that custom formats can be registered. Previously, there were only CSS and JS bundle loaders for dynamic `import()`s hard coded. Now you can call e.g. `bundler.addBundleLoader('ext', '/path/to/my/loader.js')` to add a bundle loader for some custom type. These will automatically be included in the bundle by the JSPackager if that type is dynamically imported.

This also makes it possible to have other bundles be preloaded prior to a JS bundle being executed. This is for the case of e.g. `wasm` where we don't want to inline the binary into the JS, but the import was synchronous: e.g. `import {add} from './add.wasm'`. In this case, the JSPackager notices that the module is in an external file, and adds some code to preload it prior to the JS being run. This is only done for root JS bundles, because for async/dynamic imports the wasm file would be loaded in parallel with the JS already.

This is a somewhat large change, and I ended up refactoring a bunch of other stuff to make it work. Please test out this branch and report any bugs you find! Thanks! 😍


albizures:
Hi @devongovett great job!

I have been testing it out but I get this error:

in chrome and firefox, in safari works perfectly

and I found that [compileStreaming](https://github.com/WebAssembly/design/blob/master/Web.md#webassemblycompilestreaming
) checks the MIME type, so it's a matter of adding the MIME type to the response.
but also I tried changing loadWASMBundle to:
```javascript
module.exports = function loadWASMBundle(bundle) {
  return fetch(bundle)
    .then(response => response.arrayBuffer())
    .then(bytes => WebAssembly.instantiate(bytes, {}))
    .then(results => results.instance.exports)
};
```
and it's works on chrome, firefox and safari!
I don't know what the real difference is between this two ways

devongovett:
Thanks for testing! Ok, I guess we need to use something like https://www.npmjs.com/package/mime to serve correct mime types in the dev server.

devongovett:
@albizures added the correct mime type for wasm, and it seems to be working for me now in chrome.

devongovett:
One current issue is with HMR - the port is now injected by an environment variable since the HMR runtime is a normal asset now and not injected by the packager. The problem is that env vars don't currently invalidate the cache when they change, and by default the HMR port changes on every run. Going to fix this as part of #297 hopefully.

albizures:
the MIME type error is gone but now I'm getting this error
captura de pantalla 2018-01-04 a la s 9 45 24 p m
again in chrome 😞,  in firefox and safari works perfect

btw I'm MacOS

devongovett:
@albizures interesting. I've changed it to use `WebAssembly.instantiateStreaming` and `WebAssembly.instantiate` instead. Let me know if that works.

albizures:
it works perfectly! ✨

devongovett:
Awesome, going to merge this. 💥

PR #475, opened 2018-01-02
[WIP] Use babel 7

xtuc:
I also removed `matchesPattern` which is now in `@babel/types`.

**Edit**:
I managed to break most of your tests:
```js
bundler(file, opts).bundle();
```

returns `undefined` now. Do you have any idea why?

I work continue to work on this tomorrow.


xtuc:
Most of the tests are failing with:
```
     TypeError: Cannot read property 'name' of undefined
      at assertBundleTree (test/utils.js:96:39)
      at Context.<anonymous> (test/watcher.js:131:5)
      at <anonymous>
```

Do you have any idea why?

src/transforms/babel.js
```diff
@@ -17,7 +17,8 @@ module.exports = async function(asset) {
     config.plugins = [require('@babel/plugin-transform-modules-commonjs')];
   }
 
-  let res = babel.transformFromAst(asset.ast, asset.contents, config);
+  let res = babel.transformFromAstSync(asset.ast, asset.contents, config);
```
DeMoorJasper: Maybe u could use the async one and convert it into a promise and use await babel.transformAst?
As the function this is called in is asynchronous anyway.

xtuc: Yes, do you have a preferred way for promisifing (like a helper) or can I create a Promise here?

DeMoorJasper: There is this, https://github.com/parcel-bundler/parcel/blob/master/src/utils/promisify.js

devongovett:
Cool, thanks for working on this! 👍 

I'm not sure we should upgrade yet until Babel 7 final is released. Probably shouldn't depend on the alpha.

Question: is Babel 7 backward compatibile with Babel 6 plugins? Is the `.babelrc` format the same as in Babel 6? If either of those is no, we will need to talk about how we can support both versions depending on which one the project has installed.

xtuc:
@devongovett 

Some API have been changed between these two versions and we have a few incompatibilities issues.

The Babel configuration stays the same apart from the addition of `.babelrc.js`.

thejameskyle:
I don't think we should use Babel 7 yet, I've been using it a lot at work and it's not been very stable

babakness:
Would be nice to have this. At least in Beta. Babel 7 has been working well for me.

pmurias:
Any updates on the upgrade? I want to use the new bigints syntax: 100n and that needs the new babel.

DeMoorJasper:
Gonna close this due to it becoming more and more conflicted with current branch and will eventually take longer to update than re-implement

xtuc:
FYI the first rc release has been published a couple of minutes again.

DeMoorJasper:
@xtuc Thanks for the heads up. We should probably start figuring out how to support both 6 and 7 soonish.

Discussion for that goes here: #868

PR #476, opened 2018-01-02
Make HMR work in Worker contexts

arianon:
The `window` object is not available in the context of Web Workers
and Service Workers, therefore `window.location` won't work, the way
to make HMR work is by accessing `location` through the implicit
global context.

### DIFF
--- a/src/builtins/hmr-runtime.js
+++ b/src/builtins/hmr-runtime.js
@@ -15,7 +15,7 @@ function Module() {
 module.bundle.Module = Module;
 
 if (!module.bundle.parent && typeof WebSocket !== 'undefined') {
-  var ws = new WebSocket('ws://' + location.hostname + ':{{HMR_PORT}}/');
+  var ws = new WebSocket('ws://' + global.location.hostname + ':{{HMR_PORT}}/');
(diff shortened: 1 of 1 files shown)


src/builtins/hmr-runtime.js
```diff
@@ -15,7 +15,7 @@ function Module() {
 module.bundle.Module = Module;
 
 if (!module.bundle.parent && typeof WebSocket !== 'undefined') {
-  var ws = new WebSocket('ws://' + window.location.hostname + ':{{HMR_PORT}}/');
+  var ws = new WebSocket('ws://' + location.hostname + ':{{HMR_PORT}}/');
```
brandon93s: Could we use `self.location.hostname` to avoid the implicit global? I think this will work in both contexts.

arianon: I tried that, but then the tests throw `ReferenceError: self is not defined`, however I found out that I can reference `location` from the `global` object, which you define on the top level, I'll use that.

devongovett:
Closing in favor of #529. `global` won't currently work in the HMR runtime due to the way it is injected.

PR #483, opened 2018-01-03
Transform HTML prior to collecting dependencies

ipmb:
Parcel assumes it is working with complete HTML when collecting dependencies, but until it is transformed, not all the assets may be visible. The test includes an example with `posthtml-include`, but I also had issues with `posthtml-extend`.


ipmb:
Committed a fix, but not sure if it's the best one. This transforms the HTML twice, once before collecting dependencies and once after.

PR #484, opened 2018-01-04
Use verbose output during testing

shawwn:
Currently, we use `logLevel: 0` when running tests.

`logLevel: 3` has several advantages, most notably that it doesn't
suppress errors.

(If you run `yarn test` with `logLevel: 3`, you'll see we currently
have an error in test/html.js that went unnoticed, for example.)

### DIFF
--- a/test/utils.js
+++ b/test/utils.js
@@ -36,7 +36,7 @@ function bundler(file, opts) {
         cache: false,
         killWorkers: false,
         hmr: false,
-        logLevel: 0
+        logLevel: 3
       },
       opts
     )


DeMoorJasper:
Maybe something like this in the logger would be better?
Although this might need some changes inside the tests
```Javascript
error(err) {
    if (process.env.NODE_ENV === 'test') {
      throw (err);
    }
```

devongovett:
yeah IMO here we should just throw the error in test mode: https://github.com/parcel-bundler/parcel/blob/master/src/Bundler.js#L162

```javascript
} else if (process.env.NODE_ENV === 'test') {
  throw err;
}
```

shawwn:
Yeah, throwing in Bundler#L162 would help.

It's quite nice having verbose output during tests. Maybe we could add an
environment variable? PARCEL_VERBOSE or some such.

brandon93s:
@shawwn - why do we need a new environment variable? 

Let's just throw (re-thtow) when testing and move on.

devongovett:
Closing. If you want to make another PR with the re-throw, please do.

PR #486, opened 2018-01-04
Improve CSS loader

shawwn:
When implementing Vue support, I ran into an issue with our existing
CSS loader. When I made changes to CSS files, the results wouldn't
show up for >5 seconds.

This CSS loader is derived from either vueify or vue-loader (I forget
which). Either way, it's battle tested and simple. There is no delay
whenever CSS code changes.

### DIFF
--- a/src/assets/CSSAsset.js
+++ b/src/assets/CSSAsset.js
@@ -98,16 +98,18 @@ class CSSAsset extends Asset {
   }
 
   generate() {
-    let css = this.ast ? this.ast.render() : this.contents;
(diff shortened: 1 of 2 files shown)


devongovett:
One potential issue: this doesn't look like it removes the original `link` element that loaded the stylesheet, it just adds a `style` element with the new css in it. So imagine you edited the CSS file to remove a selector. A new style element would be appended with the new CSS, but the selector would not get removed since it is still in the original stylesheet loaded in the `link`. That's why I originally went with the approach of just reloading the links.

PR #492, opened 2018-01-04
Add ability to specify HMR hostname

suevalov:
I've got a problem after #426 being merged. I'm using `parcel` for Chrome Plugins development and cannot rely on `window.location.hostname` value. 

Url of the page i'm working on is `chrome-extension://<extension-id>` and HMR doesn't work, because `window.location.hostname` is equal to `<extension-id>`.

This PR adds new `hmrHostname` option. If it's not set, `window.location.hostname` is being used.


suevalov:
Any update or feedback?

devongovett:
You probably need to add this to the CLI as an option too, right? https://github.com/parcel-bundler/parcel/blob/master/src/cli.js#L16-L20

suevalov:
@devongovett Good catch, thanks. Added.

PR #495, opened 2018-01-05
[RFC] Get rid of multiprocessing – parcel becomes twice as fast

shawwn:
🐛 bug report

`yarn test`. I've also tested this with a large codebase.

### 🤔 Expected Behavior

Intuitively, multiprocessing should increase performance, but Parcel runs much faster after turning it off.

### 😯 Current Behavior

Performance tests using the current HEAD:

`parcel --no-cache demo/index.html`, multiprocessing / worker processes ON:

image

`parcel build --no-cache demo/index.html`, multiprocessing / worker processes ON:
image

`time yarn test`, multiprocessing / worker processes ON:

image

After applying the following patch to disable worker processes, you can see a sharp increase in performance:

image

`parcel --no-cache demo/index.html`, single process:

image

`parcel build --no-cache demo/index.html`, single process:

image

`time yarn test`, single process:

image

### 🔦 Context

Benefits of stripping out multiprocessing:

`yarn test` runs 2.4x faster (60% time decrease).

`parcel foo.html` starts ~50% faster

`parcel build foo.html` is slightly faster

IDE debugging becomes much easier, since breakpoints can be placed in Asset classes. (Normally this functionality runs in a separate process.)

The code becomes far less complicated, since this removes the need for an RPC interface between child processes and the host.

### 🌍 Your Environment

| Parcel           | HEAD
| Node             | v9.3.0
| Operating System | macOS

### DIFF
--- a/src/WorkerFarm.js
+++ b/src/WorkerFarm.js
@@ -1,6 +1,4 @@
 const {EventEmitter} = require('events');
-const os = require('os');
-const Farm = require('worker-farm/lib/farm');
 const promisify = require('./utils/promisify');
 
 let shared = null;
--- a/src/WorkerFarm.js
+++ b/src/WorkerFarm.js
@@ -5,25 +5,14 @@ const promisify = require('./utils/promisify');
 
 let shared = null;
 
-class WorkerFarm extends Farm {
-  constructor(options) {
-    let opts = {
-      autoStart: true,
-      maxConcurrentWorkers: getNumWorkers()
-    };
-
-    super(opts, require.resolve('./worker'));
-
+class WorkerFarm {
+  constructor() {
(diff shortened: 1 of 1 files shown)


shawwn:
Ok, if anyone wants to see what I mean, run this:

```
hub checkout https://github.com/parcel-bundler/parcel/pull/495
yarn && yarn link
time yarn test
```

As you can see, the unit tests are way faster, build times are way lower, and parcel watch/serve starts up ~instantly. 

Unless someone can show a performance test demonstrating that multiprocessing is ever faster than this PR, in any situation, then we should consider axing the workerfarm.

devongovett:
Closing for now. For large project with e.g. > 1000 modules, we are seeing a 2x speedup using workers.

PR #499, opened 2018-01-05
[RFC] Workerfarm performance experiment

DeMoorJasper:
related to #495
I switched out ``workerFarm`` for ``worker-nodes`` as worker-nodes advertised to be faster, in reality i only noticed small performance improvements however. I'm convinced there is a way to up this to be closer to or equal to the synchronise building.

== Benchmarks ==
Synchronise: 3.64 secs
Parcel master branch: 9.56 secs
worker-nodes: 7.38 secs
jest-worker: 12.64secs


devongovett:
cool, how about `jest-worker`?

DeMoorJasper:
@devongovett jest-worker vs current master branch
```
jest-worker: 12.64secs
Master branch: 6.73secs
`
``

PR #506, opened 2018-01-07
JS Source Maps support

### LINKED ISSUE #68: Feature request: source map support
It would be great if parcel managed source maps.

DeMoorJasper:
Adds source map support to Javascript, Typescript & Coffeescript.
Performance impact appears to be very limited to my surprise (as every transpiler was probably already creating a sourcemap in the background)

Closes #68

TODO:
* [x] Get source maps to work
* [x] ~Generate custom source maps instead of using babel (so we can continu using babylon)~ EDIT: Used babel-generator as it's already being used
* [x] Find a better way to offset (Possibly offset on JSPackage, as this will be most reliable? For this i'd have to update the code to package main asset type before the sibling bundles)
* [x] Support source maps in typescript, coffeescript,...
* [x] Fix the tests (to expect source maps)
* [x] Write source maps tests
* [x] Make it blazing fast! 🚀
* [ ] SourceMaps in Production mode? [RFC]


src/packagers/JSPackager.js
```diff
@@ -20,6 +21,7 @@ class JSPackager extends Packager {
   async start() {
     this.first = true;
     this.dedupe = new Map();
+    this.sourcemapPackager = new SourcemapPackager(this.bundle, this.bundler);
```
devongovett: If you register the SourcemapPackager in [here](https://github.com/parcel-bundler/parcel/blob/master/src/packagers/index.js) you won't need to construct it from here. `addAsset` will automatically get called as well, you don't need to call it from here. This way the source map packager will automatically work for non-JS langs as well if we generate maps.

src/packagers/JSPackager.js
```diff
@@ -87,7 +93,12 @@ class JSPackager extends Packager {
       entry.push(this.bundle.entryAsset.id);
     }
 
-    await this.dest.end('},{},' + JSON.stringify(entry) + ')');
+    await this.dest.write('},{},' + JSON.stringify(entry) + ')');
+    await this.dest.write(
+      `\n//# sourceMappingURL=${this.sourcemapPackager.url}`
```
devongovett: Since you won't have access to the source map packager, the URL will be the name of the JS bundle with `.js` replaced with `.map`.

src/assets/JSAsset.js
```diff
@@ -116,7 +116,8 @@ class JSAsset extends Asset {
     }
 
     return {
-      js: code
+      js: code,
+      map: this.sourcemap
```
devongovett: We'll also need to generate 1:1 source maps for files that we don't process with babel.

src/packagers/SourcemapPackager.js
```diff
@@ -0,0 +1,74 @@
...
+        return false;
+      }
+      this.generator.addMapping({
+        source: mapping.source,
+        original: {
+          line: mapping.originalLine,
+          column: mapping.originalColumn
+        },
+        generated: {
+          line: mapping.generatedLine,
```
devongovett: don't we need to offset this line by the current line number in the file since we're concatenating a bunch of things together? and what about the extra lines added by the JSPackager... hmm

src/assets/JSAsset.js
DeMoorJasper: Will work on that once the JS maps are more on point

src/packagers/JSPackager.js
DeMoorJasper: How do i go about this without overwriting the jsPackager?

devongovett: You'd register for type `map` not type `js` so it shouldn't overwrite.

```javascript
this.add('map', SourceMapPackager);
```

Parcel should already be creating a bundle of type `map` since you return one from JSAsset#generate.

src/packagers/JSPackager.js
DeMoorJasper: Alright, will fix this

src/packagers/SourcemapPackager.js
DeMoorJasper: We probably should, i'll dive deeper into this

src/packagers/JSPackager.js
DeMoorJasper: I've tried this out and it appears bundles are created based on asset.type not asset.generated.

devongovett: oh right, I changed this recently. See https://github.com/parcel-bundler/parcel/blob/master/src/Bundler.js#L413-L420. Looks like your branch doesn't have that change.

DeMoorJasper:
@devongovett I got the big portions of source maps working, my main concern now is the offsets.
I was thinking of adding this to bundle's:
```
this.offsets = new Map();

function addOffset(assetname, line, column) {
  this.offsets.set(assetname, line, column);
}
```
And this to jsPackage
```
function insertModule(asset) {
  ...
  this.bundle.addOffset(asset.relativename, this.lineCount, 0);
  this.dest.write(....);
  ...
}
```
This way i could just extract the very accurate offsets from the bundle, this would also support css in the future, if cssPackager just adds offsets as well.

The current issue with this is that the map packager is triggered before the js packager, which would require a few edits before this to work, so i'm wondering if this is the best approach before changing/adding all this code.

DeMoorJasper:
@devongovett sourcemaps should be ready to go

albinotonnina:
**Oh I'm getting all emotional now** 😥🤧

screen shot 2018-01-11 at 15 48 28

src/assets/JSAsset.js
DeMoorJasper: done

devongovett:
Awesome work @DeMoorJasper! I'll take a look at this shortly.

hjanuschka:
❤️  i am really looking forward to this PR.

however i tried it on one of my parcel projects.:

```
node_modules/symbol-observable/es/index.js: original.line and original.column are not numbers -- you probably meant to omit the original mapping entirely and only map the generated position. If so, pass null for the original mapping instead of an object with empty or null values.
at SourceMapGenerator_validateMapping [as _validateMapping] (/home/deploy/parcel/node_modules/source-map/lib/source-map-generator.js:276:15)
    at SourceMapGenerator_addMapping [as addMapping] (/home/deploy/parcel/node_modules/source-map/lib/source-map-generator.js:110:12)
    at /home/deploy/parcel/node_modules/source-map/lib/source-map-generator.js:72:17
    at BasicSourceMapConsumer.SourceMapConsumer_eachMapping [as eachMapping] (/home/deploy/parcel/node_modules/source-map/lib/source-map-consumer.js:157:14)
    at Function.SourceMapGenerator_fromSourceMap [as fromSourceMap] (/home/deploy/parcel/node_modules/source-map/lib/source-map-generator.js:48:24)
    at getGenerator (/home/deploy/parcel/src/utils/SourceMaps.js:32:39)
    at offsetEmptyMap (/home/deploy/parcel/src/utils/SourceMaps.js:52:19)
    at combineSourceMaps (/home/deploy/parcel/src/utils/SourceMaps.js:88:14)
    at Object.offsetSourceMap (/home/deploy/parcel/src/utils/SourceMaps.js:38:15)
```


any ideas?

DeMoorJasper:
@hjanuschka Thank you for testing this out! I'll look into this issue in a bit

hjanuschka:
@DeMoorJasper  seems that column and line are `null` - let me know if i can provide more details - i really like this PR

DeMoorJasper:
@hjanuschka It should be fixed in the latest version, could u give it a go? If it's not fixed i would like a test repo so i can experiment and debug a bit

hjanuschka:
@DeMoorJasper you made my day! yes the PR is working fine! looking forward to get this into `master`

src/Bundler.js
```diff
@@ -68,6 +68,11 @@ class Bundler extends EventEmitter {
       logLevel: typeof options.logLevel === 'number' ? options.logLevel : 3,
       mainFile: this.mainFile,
       hmrPort: options.hmrPort || 0,
+      rootDir: Path.dirname(this.mainFile),
+      sourcemaps:
+        typeof options.sourcemaps === 'boolean'
+          ? options.sourcemaps
+          : !isProduction,
```
devongovett: Why should we disable sourcemaps in production? Isn't that where they are most useful since code is minified? Let's get some feedback on this.

src/assets/JSAsset.js
```diff
@@ -117,16 +118,41 @@ class JSAsset extends Asset {
...
-      ? generate(this.ast).code
-      : this.outputCode || this.contents;
+    let code;
+    if (this.isAstDirty) {
+      let opts = {
+        sourceMaps: this.options.sourcemaps,
+        sourceFileName: this.options.sourcemaps ? this.relativename : undefined
+      };
+      let generated = generate(this.ast, opts, this.contents);
+      if (this.sourcemap && generated.map) {
```
devongovett: When/why does this case occur (there is already a source map)?

src/packagers/JSPackager.js
```diff
@@ -87,7 +91,15 @@ class JSPackager extends Packager {
       entry.push(this.bundle.entryAsset.id);
     }
 
-    await this.dest.end('},{},' + JSON.stringify(entry) + ')');
+    await this.dest.write('},{},' + JSON.stringify(entry) + ')');
+    // Add sourcemap url
+    await this.dest.write(
```
devongovett: This should only be added if sourcemaps are enabled right?

src/packagers/SourcemapPackager.js
```diff
@@ -0,0 +1,38 @@
...
+    });
+  }
+
+  getOffsets(asset) {
+    let parent = this.bundle.parentBundle;
+    return parent.offsets.get(asset.relativename) || {line: 0, column: 0};
+  }
+
+  async addAsset(asset) {
+    if (asset.generated.map) {
```
devongovett: If an asset is added to a `map` bundle, it should definitely have a generated map. did you find otherwise?

devongovett:
One question: how does this work with HMR? Ideally, on asset update the source map would also update. Not sure if that's even possible.

Also, I did some performance testing and unfortunately this slows down Parcel a lot. 6s build time on my project without source maps, and 53s with source maps enabled. I profiled and looks like it's mostly in the text utils. Let's see if we can make this faster!

screen shot 2018-01-15 at 3 57 54 pm

devongovett:
Maybe read https://medium.com/@david.aurelio/medium-fast-source-map-generation-for-react-native-ea5549007c18 and see if we can apply any of the learnings from the perf work that went into react native even if we can't use the code (because of the license)

src/Bundler.js
hjanuschka: i think there are usecases for disabling sourcemaps in production. how about enabling it by default and adding an option `--disable-sourcemaps` - so we can choose at build-time?

src/assets/JSAsset.js
DeMoorJasper: This is when babel has to generate the code from ast after for example Typescript parsed the code. Typescript sets a sourcemap, although babel changes the sourcemap in these particular cases, and this function offsets it correctly again.
This happens in the typescript env test for example

src/packagers/JSPackager.js
DeMoorJasper: Ow yes, will change this

src/packagers/SourcemapPackager.js
DeMoorJasper: This is probably a remain from the initial approach, i'll remove this and test a bit to make sure it doesn't break

DeMoorJasper:
@devongovett the long split time in your profiling probably comes from the lineCounter which is being used to make a 1:1 map, I haven't seen this in any of my tests. But I will make it faster non the less, have seen some stuff i can improve a lot upon.

DeMoorJasper:
Most performance flaws should now be fixed.

Things i did to improve performance since Devon's review:
* Removing all use of source-map generator (except for stringifying), replacing most of it's functionality with the SourceMap class.
* Using babel-generator's rawMappings, removing the need to encode and decode the mappings
* Reduced the amount of textUtils usage, by including getEmptyMap in SourceMap (as this does support 1:1 mappings and the source-map generator doesn't) class and using consumers correctly for extendSourceMap

One function that can still cause a bit of performance issues is `SourceMap.extendSourceMap` as it uses SourceMapConsumer's and doesn't yet use the rawMappings from babel. Also any non-js asset that supports sourcemaps and don't support some sort of rawMappings still need an extra decoding operation costing some extra time and ressources.

(I also tried out the buffer instead of source-map generator that was proposed by the article about metro-source-maps but it appears to not cause a big performance difference in this case. If there still appears to be a major performance penalty I will experiment further with this. https://github.com/DeMoorJasper/parcel/tree/feature/source-map/experimental-buffer-encoding)

geowarin:
@DeMoorJasper Sourcemaps seem off on a typescript + react simple project.

localhost


Repro: https://github.com/geowarin/test-parcel

TimNZ:
Apologies for noob question.
I did:

* git clone of your fork
* yarn install
* yarn build
* cd to root dir of my app
* run '../parcel-sourcemaps/bin/cli build src/index.js' (same params as currently do for global installed parcel), and getting 

```
Cannot read property 'env' of undefined
    at JSAsset.shouldInvalidate (/Users/timshnaider/Desktop/Development/packagers/parcel/sourcemap/src/assets/JSAsset.js:33:30)
    at Bundler.loadAsset (/Users/timshnaider/Desktop/Development/packagers/parcel/sourcemap/src/Bundler.js:370:29)
    at <anonymous>
```

Node 8.4.0, OS X

Am I doing something wrong?

TimNZ:
Source map reloading:
https://github.com/facebookincubator/create-react-app/issues/2407

devongovett:
@TimNZ delete the `.cache` folder or run with `--no-cache`. Normally this isn't an issue unless you're installing from github.

TimNZ:
@devongovett That did it, thanks.

Sorry this is even more noob question: my src files aren't in the map, only node_modules.

TimNZ:
Production react is being imported for debug build using fork.

Development react imported using parcel.

devongovett:
Nice work @DeMoorJasper! 🥇 

With your latest changes I get 6.85s without sourcemaps, and 7.68s with. Much improved!

src/utils/textUtils.js
```diff
@@ -0,0 +1,5 @@
+function lineCounter(string) {
+  return string.split('\n').length;
```
devongovett: You could implement this as a simple for loop to check count each newline character instead of using split. That would avoid allocations and probably be faster. Want to test?

DeMoorJasper: I wrote a jsPerf for this in it appears regex is faster than both split and for loop.
https://jsperf.com/linecounter/1
I'll change the function into the regex method

TimNZ:
@devongovett With the env issue I commented above needing clearing of cache, something you should catch and show a message?  Error handling in general ... :)

TimNZ:
@DeMoorJasper Did a git pull on your fork and rm node_modules and reinstall and seems to have sorted my issues at quick glance.  Thanks.

DeMoorJasper:
After people have mentioned the new improved source-map and reading a blog post about the many performance improvements i decided to branch off this pr and make a source-map 0.7 compatible version for when it lands: https://github.com/DeMoorJasper/parcel/tree/source-map-0.7
The performance impact is pretty limited due to this being maximised for performance and preventing mapping parsing whenever possible, this was the main performance issue with source-map 0.6.1

devongovett:
One problem: source-map 0.7 requires Node 8. I know we're not there yet, but I think we are aiming to support Node 6 in parcel as well...

DeMoorJasper:
@devongovett Should i downgrade back to source-map 0.6.1? I see no way around this as webassembly is not supported below node.js 8.0

devongovett:
Yeah for now. Filed an issue to see if they can include the JS only version for Node 6 users: https://github.com/mozilla/source-map/issues/313

src/SourceMap.js
```diff
@@ -0,0 +1,192 @@
...
+  copyConstructor(map) {
+    let sourcemap = new SourceMap();
+    sourcemap.mappings = map.mappings;
+    sourcemap.sources = map.sources;
+    sourcemap.file = map.file;
+    return sourcemap;
+  }
+
+  isConsumer(map) {
+    return map && map.computeColumnSpans;
```
devongovett: Can we just do `map instanceof sourceMap.SourceMapConsumer`? This check currently seems somewhat brittle since they could change the internal implementation.

PR #521, opened 2018-01-09
Invalidate cache on config and environment variable changes

### LINKED ISSUE #66: Config files do not invalidate cache

devongovett:
Closes #297. Fixes #66.

This PR does three things:

1. Adds config files as dependencies. This means that they are automatically watched, so for example if you change your `.babelrc`, Parcel will immediately recompile all files that depend on it. Also, if you restart Parcel after making changes, those files will also be recompiled (since #514).
2. Adds a `cacheData` object to assets, and a `shouldInvalidate` method which accepts a cacheData object to compare. If it returns false, the asset is re-generated. JSAsset uses this to implement environment variable cache invalidation. If your JS file has `process.env.MY_ENV`, and you change `MY_ENV` to a new value, the cache will be invalidated.
3. Fixes an issue where assets that have multiple parents that include them weren't recompiled on change. For example, if you had a stylus file that was @imported by several parents, only one of them would get recompiled when that file changed. This also affected .babelrc and other config changes, etc.

### DIFF
--- a/src/assets/JSAsset.js
+++ b/src/assets/JSAsset.js
@@ -25,6 +25,17 @@ class JSAsset extends Asset {
     this.isAstDirty = false;
     this.isES6Module = false;
     this.outputCode = null;
+    this.cacheData.env = {};
+  }
+
+  shouldInvalidate(cacheData) {
+    for (let key in cacheData.env) {
+      if (cacheData.env[key] !== process.env[key]) {
+        return true;
+      }
+    }
+
+    return false;
   }
 
   mightHaveDependencies() {
--- a/src/assets/JSAsset.js
+++ b/src/assets/JSAsset.js
@@ -10,7 +10,6 @@ const fsVisitor = require('../visitors/fs');
 const babel = require('../transforms/babel');
 const generate = require('babel-generator').default;
 const uglify = require('../transforms/uglify');
-const config = require('../utils/config');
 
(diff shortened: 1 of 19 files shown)


src/assets/JSAsset.js
```diff
@@ -26,6 +25,17 @@ class JSAsset extends Asset {
     this.isAstDirty = false;
     this.isES6Module = false;
     this.outputCode = null;
+    this.cacheData.env = {};
+  }
+
+  shouldInvalidate(cacheData) {
```
DeMoorJasper: This causes an error on existing builds (as they already have cache files) and cacheData is undefined in these files.

devongovett: hmm yeah, but should be ok with the npm version since the parcel version is part of the cache key right? so when people update parcel, everything will be invalidated already.

DeMoorJasper: Ow forgot about that, yeah should be fine

Issue #474, opened 2018-01-02
Ability to ignore modules at least in well-known environments


gimenete:
This is a 🙋 feature request.

I was trying to use parcel.js for an electron application. The goal is to have a lightweight HMR solution without the need to set up something complex. Besides I love the fine grained control you can get with `module.hot.dispose` and `module.hot.accept`.

```bash
parcel index.html
```

### 🤔 Expected Behavior

parcel.js should not bundle the `electron` package.

### 😯 Current Behavior

```
🚨  /Users/gimenete/projects/unicycle/unicycle/node_modules/electron/index.js: Could not statically evaluate fs call
 ```

Basically parcel is trying to bundle the `electron` package itself. Internally electron does things with the file system. I cannot control that, but even more: there's no need to bundle `electron` itself with the code.

### 💁 Possible Solution

I like the idea of having well-known environment configurations, such as [eslint does](https://eslint.org/docs/user-guide/configuring#specifying-environments). So we could have:

```bash
parcel --env electron index.html
```

This would apply some additional rules to parcel due to the selected environment. For example ignoring `electron` or even ignoring all `node_modules`.

### 💻 Code Sample

As soon as you import electron itself:

```html
<!DOCTYPE html>
<html>

<head>
  <meta charset="UTF-8">
</head>

<body>
  <script>
    window.nodeRequire = require;
    delete window.require;
    delete window.exports;
    delete window.module;
  </script>
  <script src="./hello.js"></script>
</body>

</html>
```

```javascript
// hello.js
import * as electron from "electron";

console.log('hello world');
```

The build fails. The additional `<script>` with those `delete`s is due to this: https://github.com/parcel-bundler/parcel/issues/321 Which is another thing the `--env` idea could fix.

### 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Parcel           | 1.4.1 |
| Node             | v8.9.1 |
| npm/Yarn         | npm 5.5.1 |
| Operating System | macOS High Sierra |

gimenete:
I've created a repo to reproduce the problem: https://github.com/gimenete/electron-parcel-bug-report

bardiainjast:
any news on this ? unable to use parcel in electron environments :(

maccelerated:
[...]

PR #523, opened 2018-01-09
Use browser-resolve to ignore fs inlining

### LINKED ISSUE #496: 🐛 Could not statically evaluate fs call
🐛 bug report

zero config

```js
{
  "your": { "config": "here" }
}
```

### 🤔 Expected Behavior

work

### 😯 Current Behavior

```
node_modules/options/lib/options.js: Could not statically evaluate fs call

at evaluate (/Users/victor/.config/yarn/global/node_modules/parcel-bundler/src/visitors/fs.js:163:11)
    at CallExpression.path.get.map.arg (/Users/victor/.config/yarn/global/node_modules/parcel-bundler/src/visitors/fs.js:30:21)
    at Array.map (<anonymous>)
    at JSAsset.CallExpression (/Users/victor/.config/yarn/global/node_modules/parcel-bundler/src/visitors/fs.js:30:10)
    at NodePath._call (/Users/victor/.config/yarn/global/node_modules/babel-traverse/lib/path/context.js:76:18)
    at NodePath.call (/Users/victor/.config/yarn/global/node_modules/babel-traverse/lib/path/context.js:48:17)
    at NodePath.visit (/Users/victor/.config/yarn/global/node_modules/babel-traverse/lib/path/context.js:105:12)
    at TraversalContext.visitQueue (/Users/victor/.config/yarn/global/node_modules/babel-traverse/lib/context.js:150:16)
    at TraversalContext.visitMultiple (/Users/victor/.config/yarn/global/node_modules/babel-traverse/lib/context.js:103:17)
    at TraversalContext.visit (/Users/victor/.config/yarn/global/node_modules/babel-traverse/lib/context.js:190:19)
```

### 💻 Code Sample

```
import Gdax from 'gdax'
const publicClient = new Gdax.PublicClient()

publicClient
  .getProducts()
  .then(data => {
    console.log(data)
  })
  .catch(error => {
    console.log(error)
  })
```

### 🌍 Your Environment

| Parcel           | 1.4.1
| Node             | 9.3.0
| npm/Yarn         | 1.3.2
| Operating System | macOS 10.13.3

fathyb:
[`browser-resolve` has an option to skip `fs` and other modules](https://github.com/defunctzombie/node-browser-resolve#skip). This PR makes the `fs` visitor check if `fs` is ignored before processing it.

This makes `aws-sdk-js` usable amongst other libraries using this feature.

Fixes #496 and maybe #474


devongovett:
Cool! another option would be to just not error on non-static file paths (e.g. with variables), and just skip inlining. The only thing I'd be concerned about would be that it might reduce debuggability for legit coding errors. thoughts?

fathyb:
I agree on the debuggability point, it could be frustrating to not understand why it does not inline. What about passing an additional option to `fs.readFileSync`, something like : 
```js
fs.readFileSync(Math.random().toString(), {
  parcel: false
})
```

That way it can be ignored on node.js and can hint the visitor if the call wasn't meant to be parsed. It might not be applicable on certain use case tho.

devongovett:
I don't think we should be changing the API. Also that won't work for existing node_modules that are already broken with Parcel.

fathyb:
Alright, then I'd recommend to display a warning when it happens. #448 could be useful here to silent warnings on some files.

SheetJSDev:
Encountered the `Could not statically evaluate fs call` error while trying to import the [`xlsx` module](https://www.npmjs.com/package/xlsx).  It uses the [`browser` field of package.json](https://github.com/SheetJS/js-xlsx/blob/master/package.json#L30) to suppress for browserify.

>  another option would be to just not error on non-static file paths (e.g. with variables), and just skip inlining.

@devongovett even if a specific `fs` call could be statically evaluated, this patch is still important.  `false` entries in package.json `browser` field unambiguously communicate a desire to suppress behavior in a browser-based deployment.  The general expectation from the perspective of library authors is that `require("fs")` would either error or evaluate to `{}` if suppressed.  

Considering prior art, neither browserify nor webpack attach special meaning to the `fs` module; the filesystem meaning is conferred by modules like `brfs`.  You can manually verify this with a basic `fs.readFileSync` example in browserify and webpack.

@fathyb I personally like the concise expression 

```js
let ignore = (((asset||{}).package||{}).browser||{}).fs === false
```

but the explicit form is probably clearer.

sterproton:
i have problem that after use browser-resolve and it skip fs module
but i still want to use the fs module and it just become {}??

PR #529, opened 2018-01-10
Do not use window global in HMR runtime

### LINKED ISSUE #528: HMR runtime references global by wrong name in Worker
🐛 bug report

I've created a very simple project with a web worker. It looks like the script reference in the Worker constructor call is picked up automatically by Parcel and compiled into a separate bundle, which is great. However, the generated bundle by default includes the hot module reloading runtime, which contains code referencing `window.location`. In a Worker scope, the global is called `self`, and the bundle code throws a ReferenceError before the actual worker script can run.

### 🎛 Configuration / code sample

Create the following files and run `parcel serve index.html`

#### index.js
```js
var w = new Worker('./worker.js');
w.onmessage = e => console.log(e.data);
setInterval(() => w.postMessage('test'), 1000);
```

#### worker.js
```js
self.onmessage = e => self.postMessage(e.data + ' from worker');
```

### index.html
```html
```

### 🤔 Expected Behavior

Expected code to load and work correctly. 

### 😯 Current Behavior

Uncaught ReferenceError: window is not defined

The code in worker.js does not run and no "test" messages are logged to the console.

### 💁 Possible Solution

The error is thrown in what looks to be the following code, included in the Worker bundle: https://github.com/parcel-bundler/parcel/blob/06d5ffc33dac19fa9ca730bbf939052b500a34ec/src/builtins/hmr-runtime.js#L18 

The HRM runtime should either detect it is running in a Worker scope and use the self global, or be disabled for Worker bundles if there are other issues that prevent it from working in that context.

### 🌍 Your Environment

| Parcel           | 1.4.1
| Node             | 8.4.0
| npm/Yarn         | npm 5.6.0

bwrrp:
The name of the global object may differ based on the context in which the
script is running. For example, web workers use "self" instead of "window". As
location is a property of the global, it can be referenced directly as a global
variable, which avoids having to deal with this naming inconsistency.

Fixes #528

### DIFF
--- a/src/builtins/hmr-runtime.js
+++ b/src/builtins/hmr-runtime.js
@@ -15,7 +15,7 @@ function Module() {
 module.bundle.Module = Module;
 
 if (!module.bundle.parent && typeof WebSocket !== 'undefined') {
-  var ws = new WebSocket('ws://' + window.location.hostname + ':{{HMR_PORT}}/');
+  var ws = new WebSocket('ws://' + location.hostname + ':{{HMR_PORT}}/');
   ws.onmessage = function(event) {
     var data = JSON.parse(event.data);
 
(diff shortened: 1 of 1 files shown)


DeMoorJasper:
duplicate of #476 but this seems to be a more complete fix

Issue #491, opened 2018-01-04
🙋 moment.js zh-cn fail


echowant:
Using the following code to do the Sinization of the time,
it is failed by parcel. But by webpack，it is ok.
Can you help me?

```
import  moment from 'moment';
import 'moment/locale/zh-cn'
moment.locale('zh-cn');
```
![tim 20180104180226](https://user-images.githubusercontent.com/18191880/34558483-49ac78b0-f179-11e7-857e-915df73ca61c.png)

PlayMa256:
What is the error that parcel throws? Or none?

niicojs:
I do have a similar issue. Parcel works fine, but the second import does not do what it should.
It is supposed to modify the moment global variable to add stuff into it, but I guess it cannot access moment as it's not global?

fathyb:
I think it's related to #418

PlayMa256:
check pr #530 and see if it works.

fathyb:
@PlayMa256 it shouldn't work on this case.

**The problem** :
- `import moment from 'moment'` is resolved to an ES6 module, according to the `module` `package.json` field
- `import 'moment/locale/zh-cn'` is resolved to an UMD module, which uses the moment commonjs module

**What it means** : you are using ES6 moment, the locale uses CommonJS moment, it's two different instances, so the locale will not apply.

**Workaround** (pick one) :
- Use the ES6 locale :
  ```js
  import locale from 'moment/src/locale/zh-cn'

  moment.locale('zh-cn', locale)
  ```
- Use the CommonJS moment :
  ```js
  import moment from 'moment/moment'
  ```

Caused by #299, related to #418

niicojs:
in my case, this is what fix it:
```
import * as moment from 'moment/moment';
import 'moment/locale/fr';
```

DeMoorJasper:
Should have been fixed

PR #530, opened 2018-01-10
[WIP/RFC] Fix ES6 wildcard import of functions

### LINKED ISSUE #418: 🐛 Upgrading from 1.2.0 to 1.3.0 breaks moment.js
I'm using typescript and when i upgrade parcel from 1.2.0 to 1.3.1 _(I've tried with 1.3.0 as well)_ the `moment.js` library starts complaining about `moment` not being a function.

```js
// package.json
{
  "scripts": {
    "start": "parcel index.html --out-dir temp/"
  },
  "dependencies": {
    "moment": "^2.20.1",
    "parcel-bundler": "1.3.0",
    "typescript": "^2.6.2"
  }
}
```

```html
<!-- index.html -->
<!DOCTYPE html>
<html lang="en">
<head>
    <title>Document</title>
</head>
<body>
    <script src="./app.ts"></script>
</body>
</html>
```

```ts
//app.ts
import * as moment from 'moment';

console.log( moment() );
```

_No tsconfig.json_

### 🤔 Expected Behavior

I would expect a `moment` object to be printed to the console.

### 😯 Current Behavior

With parcel `v1.2.0` the `moment` object is printed just fine. But with parcel `v1.3.0` or `v1.3.1` it throws a TypeError. 
```
b6623a1acd83678d58392c21229a4b42.js:6473 Uncaught TypeError: moment is not a function
    at Object.require.2.moment (b6623a1acd83678d58392c21229a4b42.js:6473)
    at newRequire (b6623a1acd83678d58392c21229a4b42.js:41)
    at require.4 (b6623a1acd83678d58392c21229a4b42.js:66)
    at b6623a1acd83678d58392c21229a4b42.js:71
```

### 🔦 Context

Gist of all files used to recreate bug: https://gist.github.com/Olian04/ce4e23c11a17d60c9a39247a879e54c7

### 🌍 Your Environment

| Parcel           | 1.2.0 & 1.3.0 & 1.3.1
| Node             | 9.3.0
| npm/Yarn         | 5.5.1 / 1.3.2 _(I've tried with both)_

fathyb:
`module.js` : 
```js
module.exports = () => 'world'
```

`index.js` :
```js
import * as world from './module'

// throws "world is not a function"
// where world is {default: () => 'world'}
console.log('hello %s', world())
```

Also covers the `moment` problem :
`module.js`
```js
export default () => 'world'
```
`index.js`
```js
import * as world from './module'

// throws "world is not a function"
// world is {default: () => 'world'}
console.log('hello %s', world())
```

Fixes #418

### DIFF
--- a/src/transforms/babel.js
+++ b/src/transforms/babel.js
@@ -14,9 +14,7 @@ module.exports = async function(asset) {
 
   if (asset.isES6Module) {
     config.babelrc = false;
-    config.plugins = [
-      require('babel-plugin-transform-es2015-modules-commonjs')
-    ];
+    config.plugins = [require('./commonjs')];
   }
 
   let res = babel.transformFromAst(asset.ast, asset.contents, config);
--- a/src/transforms/commonjs.js
+++ b/src/transforms/commonjs.js
@@ -1,0 +1,669 @@
+/*
+  Copy-pasted from
+  https://github.com/babel/babel/blob/master/packages/babel-plugin-transform-modules-commonjs/src/index.js
+*/
+const {basename, extname} = require('path');
+const template = require('babel-template');
+const babelPluginTransformStrictMode = require('babel-plugin-transform-strict-mode');
+
+const t = require('babel-types');
+
+const buildRequire = template(`
+  require($0);
+`);
+
+const buildExportsModuleDeclaration = template(`
+  Object.defineProperty(exports, "__esModule", {
+    value: true
+  });
+`);
+
+const buildExportsFrom = template(`
+  Object.defineProperty(exports, $0, {
+    enumerable: true,
+    get: function () {
+      return $1;
+    }
+  });
+`);
+
+const buildLooseExportsModuleDeclaration = template(`
+  exports.__esModule = true;
+`);
+
+const buildExportsAssignment = template(`
+  exports.$0 = $1;
+`);
+
+const buildExportAll = template(`
+  Object.keys(OBJECT).forEach(function (key) {
+    if (key === "default" || key === "__esModule") return;
+    Object.defineProperty(exports, key, {
+      enumerable: true,
+      get: function () {
+        return OBJECT[key];
+      }
+    });
+  });
+`);
+
+// Parcel specific modification
+const buildParcelWildcardInterop = template(`
+  function parcelInteropWildcard(obj) {
+    // When the object is a primitive we just return it with a default property
+    if(typeof obj !== 'object') {
+      obj.default = obj
+
+      return obj
+    }
+
+    // Use the default export as a base if it exists
+    var newObj = (obj && obj.default) || {}
+
+    // Copy each property to the object (exotic namespace)
+    for(var key in obj) {
+      if(Object.prototype.hasOwnProperty.call(obj, key)) {
+        newObj[key] = obj[key]
+      }
+    }
+
+    // set the default export if it's an ES module
+    if(obj && !obj.__esModule) {
+      obj.default = obj
+    }
+
+    return newObj
+  }
+`);
+
+const THIS_BREAK_KEYS = [
+  'FunctionExpression',
+  'FunctionDeclaration',
+  'ClassProperty',
+  'ClassMethod',
+  'ObjectMethod'
+];
+
+module.exports = function() {
+  const REASSIGN_REMAP_SKIP = Symbol();
+
+  const reassignmentVisitor = {
+    ReferencedIdentifier(path) {
+      const name = path.node.name;
+      const remap = this.remaps[name];
+      if (!remap) return;
+
+      // redeclared in this scope
+      if (this.scope.getBinding(name) !== path.scope.getBinding(name)) return;
+
+      if (path.parentPath.isCallExpression({callee: path.node})) {
+        path.replaceWith(t.sequenceExpression([t.numericLiteral(0), remap]));
+      } else if (path.isJSXIdentifier() && t.isMemberExpression(remap)) {
+        const {object, property} = remap;
+        path.replaceWith(
+          t.JSXMemberExpression(
+            t.JSXIdentifier(object.name),
+            t.JSXIdentifier(property.name)
+          )
+        );
+      } else {
+        path.replaceWith(remap);
+      }
+      this.requeueInParent(path);
+    },
+
+    AssignmentExpression(path) {
+      let node = path.node;
+      if (node[REASSIGN_REMAP_SKIP]) return;
+
+      const left = path.get('left');
+      if (left.isIdentifier()) {
+        const name = left.node.name;
+        const exports = this.exports[name];
+        if (!exports) return;
+
+        // redeclared in this scope
+        if (this.scope.getBinding(name) !== path.scope.getBinding(name)) return;
+
+        node[REASSIGN_REMAP_SKIP] = true;
+
+        for (const reid of exports) {
+          node = buildExportsAssignment(reid, node).expression;
+        }
+
+        path.replaceWith(node);
+        this.requeueInParent(path);
+      } else if (left.isObjectPattern()) {
+        for (const property of left.node.properties) {
+          const name = property.value.name;
+
+          const exports = this.exports[name];
+          if (!exports) continue;
+
+          // redeclared in this scope
+          if (this.scope.getBinding(name) !== path.scope.getBinding(name))
+            return;
+
+          node[REASSIGN_REMAP_SKIP] = true;
+
+          path.insertAfter(
+            buildExportsAssignment(t.identifier(name), t.identifier(name))
+          );
+        }
+      } else if (left.isArrayPattern()) {
+        for (const element of left.node.elements) {
+          if (!element) continue;
+          const name = element.name;
+
+          const exports = this.exports[name];
+          if (!exports) continue;
+
+          // redeclared in this scope
+          if (this.scope.getBinding(name) !== path.scope.getBinding(name))
+            return;
+
+          node[REASSIGN_REMAP_SKIP] = true;
+
+          path.insertAfter(
+            buildExportsAssignment(t.identifier(name), t.identifier(name))
+          );
+        }
+      }
+    },
+
+    UpdateExpression(path) {
+      const arg = path.get('argument');
+      if (!arg.isIdentifier()) return;
+
+      const name = arg.node.name;
+      const exports = this.exports[name];
+      if (!exports) return;
+
+      // redeclared in this scope
+      if (this.scope.getBinding(name) !== path.scope.getBinding(name)) return;
+
+      const node = t.assignmentExpression(
+        path.node.operator[0] + '=',
+        arg.node,
+        t.numericLiteral(1)
+      );
+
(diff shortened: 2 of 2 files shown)


devongovett:
hmm seems weird that we need to change the default behavior of babel. what is the exact difference? should we contribute back to babel?

fathyb:
The Babel current behaviour is totally correct and respects the ES6 spec. Importing a default export with a wildcard is illegal, so the errors here are legit.

However, people (#418) and libraries (`antd`, `@blueprintjs/core`, etc..) are still relying on this. The idea is to be more tolerant by relaxing the spec on exotic namespaces.

robcrocombe:
This didn't work for me. Using `import * as moment from 'moment';` still gives me an error (e.g. `moment is not a function`).

Hope I've installed it correctly, I'm using `npm install git://github.com/fathyb/parcel.git#fix/primitive-wildcard`.

fathyb:
@robcrocombe here are the complete instructions to be sure you are correctly using the PR :

- `rm -rf .cache`
- `npm install git://github.com/fathyb/parcel.git#fix/primitive-wildcard`
- `(cd node_modules/parcel-bundler && npm install)`
- use `node_modules/parcel-bundler/bin/cli.js` instead of `parcel` in your scripts (eg. `node_modules/bin/cli.js entry` instead of `parcel entry`)

If it still does not work could you create a reproduction repository? Thanks.

gomezhyuuga:
Great! Thanks a lot, I was having the same issue but @fathyb your instructions solved my problem.
Is this going to be solved in the future? For the moment is a dirty-patch right?

robcrocombe:
> use node_modules/parcel-bundler/bin/cli.js instead of parcel in your scripts (eg. node_modules/bin/cli.js entry instead of parcel entry)

How does this work when I import Parcel directly? E.g. `import parcel from 'parcel-bundler'`. Trying `import parcel from './node_modules/parcel-bundler/bin/cli'` didn't work for me.

fathyb:
@gomezhyuuga It is probably going to be resolved in the future. The implications needs to be discussed first and we need to be sure it does not impact other users negatively.

@robcrocombe if you are using the API then you can keep `import parcel from 'parcel-bundler'`. Keep in mind that this fix does not work with already processed libraries (eg. a npm module already compiled with Babel).

robcrocombe:
Here is a small repo that reproduces what I see in a larger project: https://github.com/robcrocombe/parcel-import-example

Hope it helps.

fathyb:
@robcrocombe  Ah, I didn't knew you were using TypeScript. You need to use `"module": "es6"` for fix to work correctly (the fix parses ES6 `import` only). Don't forget to remove the `.cache`.
```json
{
  "compilerOptions": {
    "module": "es6"
  }
}
```

@DeMoorJasper is there any reason we use `commonjs` as `module` for TypeScript by default? What do you think of using `es6` instead?

devongovett:
we don't want to use typescript es6 module output since that would require extra processing by babel (slow).

robcrocombe:
@fathyb it worked for that example repo, but I haven't had success implementing this in my large project. However I did find that using:

```js
import * as moment from 'moment/moment';
```

from [this comment](https://github.com/parcel-bundler/parcel/issues/491#issuecomment-358406066) seems to be working!

jsiebern:
jQuery & Moment were unusable (with Typescript anyway, no change of doing `moment/moment` without changing the type defs), with this PR it works nicely, any chance of it being merged?

jjh456:
@jsiebern I'm using 1.2.1 for now. Hopefully it works for you in the meantime as well.

TimNZ:
What's the status of this?  The PR is 2 months behind master

davidnagli:
@TimNZ Do you wanna test out this PR and see if it fixes your issue?

@fathyb Is this still a WIP? When do you think this will be done?

TimNZ:
@davidnagli mate I am trying but since the PR is based off 1.4.1 it doesn't include the ignoring .babelrc changes and I'm having to fix those and I don't have time to go down that hole right now.

jsiebern:
I have a fork where I applied this PR Code manually and it seems to fix all my issues I have had with the imports (esp. in Typescript)

TimNZ:
Doesn't make a difference, moment is still returned as an es6 module.

You should be able to replicate this
```
import React from 'react';
import ReactDOM from 'react-dom';
import {useStrict} from 'mobx';
import {Provider, useStaticRendering} from 'mobx-react';
import {BrowserRouter, Route, Switch} from 'react-router-dom';
useStrict(true)
import 'flag-icon-css/css/flag-icon.min.css';
import 'font-awesome/css/font-awesome.min.css';
import 'simple-line-icons/css/simple-line-icons.css';
import '../scss/core/_dropdown-menu-right.scss'
const routerBaseName = process.env.NODE_ENV == 'production' ? '/client' : '/'
import ReactDateTime from 'react-datetime'

ReactDOM.render(
  <BrowserRouter basename={routerBaseName}>
  <Provider>
      <ReactDateTime  />
    </Provider>
  </BrowserRouter>,
  document.getElementById('app')
);

```

.babelrc
```
{
    "presets": [
        ["env", {
            "targets": {
                "browsers": ["last 2 versions", "safari >= 7"]
            }
        }],
        "react"
    ],
    "plugins": [
    ]
}
```

juliocanares:
🙏 we need this 🙏

jdalton:
> import * as world from './module'

That is not a function. It's a namespace object. When you use the namespace import syntax `import * as foo from "mod"` you're requesting the namespace object of bindings from `"mod"`. If you want the default export (which may be a function) then the correct syntax is `import word from './module`.

fathyb:
@jdalton [see my comment](https://github.com/parcel-bundler/parcel/pull/530#issuecomment-356734460) :
> The Babel current behaviour is totally correct and respects the ES6 spec. Importing a default export with a wildcard is illegal, so the errors here are legit.

> However, people (#418) and libraries (antd, @blueprintjs/core, etc..) are still relying on this. The idea is to be more tolerant by relaxing the spec on exotic namespaces.

It's terribly phrased but we're aware of it. People keep complaining so we have to do something at one point, other bundlers normalised this behaviour so people/libraries are relying on it.
We can't just say "guys we don't care if the most popular date library doesn't work, at least we're standard".

jdalton:
The `moment` library should work fine when imported correctly using `import moment from "moment"`. When imported as a namespace object it makes sense that is, in fact, an object.

jdalton:
\cc TypeScript devs for their take too @DanielRosenwasser @weswigham and moment core @maggiepint

weswigham:
> edit : and I think esModuleInterop from TypeScript does the same.

Naw, that's the flag that explicitly closes the callable-namespace-import hole and issues errors if an import is used wrong. It's by default that TS allows namespace imports to be called, and is more of a "bug TS can't outright fix because so many people rely on it" than a feature. However prior to adding the "esModuleInterop" flag TS also had no other option (than calling a namespace import) that would get you the callable thing, since the whole "automatically put stuff on the default" bit we didn't support the emit for (since that's not technically standard anywhere either, and we had hoped for better), which in turn caused proliferation of the namespace hack even as TS discouraged using it (and have long allowed a non-esm `import = require("moment")` statement instead provided the target wasn't set to esm).

fathyb:
Thanks a lot @jdalton and @weswigham for your insights, this is the kind of discussion I wanted for this RFC. Lot of issues were reported with this bug and I was (or still am) confused.

I really agree that this shouldn't be done in the first place (I really don't want to merge this), but let me explain the PR origin story :

- Bundler A (could be `browserify`, or node.js) :
  - is using `require.resolve`, it doesn't care about ES6
  - resolves `moment` to the UMD module `node_modules/moment/moment.js` which uses `module.exports = function moment() {}`
- Bundler B (could be Parcel, Rollup or Webpack) : 
  - is using the `jsnext:main` or `module` fields to get ES6 modules, mainly for tree-shaking and fallbacks to `main`
  - resolves `moment` to the ES6 module `node_modules/moment/src/moment.js` which uses `export default function moment() {}`

The problem is interoperability. How can a library import `moment` in way that works on both Bundler A and B?

- `import moment from 'moment'` will not work on Bundler A or Node.js
  - It works with the Babel CommonJS plugin
  - It doesn't using TypeScript, it works with `"esModuleInterop": true` though
- `const moment = require('moment')` will not work on Bundler B

IIRC TypeScript 2.7 wasn't released when the PR was created, but now that `"esModuleInterop"` is available I think we don't need this anymore, and working with library developers is a way safer bet.

-------

# To all the people affected : 
- **The PR will not be merged**
- If you are using TypeScript : 
  - Set `module` to `commonjs` or nothing
  - Set `esModuleInterop` to `true`
- If you are using `import * as moment from 'moment'`
  - Use `import moment from 'moment'` instead
- If you are not using `moment` directly but depend on a library which depends on `moment`
  - Open an issue on the library repository explaining `import moment as` should be prefered to `import * as moment` and link it to [moment#4488](https://github.com/moment/moment/issues/4488)
- If this doesn't resolve your issue feel free to open an issue or ask for help on the Slack, it's a complex problem and it really depends on what are your dependencies (please don't comment wih your problem here, leave the PR for discussion only)

TimNZ:
Thanks for the tip on jsnext:main, I removed that from the moment package.json for building and does the trick for the time being.

TimNZ:
@fathyb  I've been ruminating about this for a few days and have a growing belief this doesn't explain why the moment scenario is broken, at least in one case.

My app
-> import 'react-datetime'
----> react-datetime: package.json 'main:./datetime.js' -> line: require('moment')
---------> package.json from moment: 'jsnext:./src/main.js'

The packager (Parcel) is drilling deep into the import/require chain, applying jsnext:* from moment package.json (or directing Babel to do so, I don't know the implementation)

Should react-datetime be getting an ES6 object with 'default' for moment when the caller (react-datetime) is doing a require and not an import?

I can't see any comments in this issue that clarify that.

I had a play with babel-plugin-module-resolver to see if I could override where moment gets resolved from but doesn't look like it works with inner modules, or I'm specifying it wrong.

```
"plugins": [
        "transform-decorators-legacy",
        "transform-class-properties",
        "transform-es2015-modules-commonjs",
        "add-module-exports",       
        ["module-resolver", {
            "root": ["./src"],
            "alias": {
                "moment": "./node_modules/moment/moment.js"
            }
        }]
    ]
```

Thoughts?

brandonros:
```
$ parcel -V
1.7.0
```

package.json

```
{
  ...
  "dependencies": {
    "moment": "^2.22.1"
  }
  ...
}
```

util.js

```
var moment = require('moment');

moment('timestamp').format('format);
```

moment is not a function...

Is this related?

DeMoorJasper:
@brandonros probably not, this was a typescript bug. Possibly just a wrong way of importing? Might be moment specific?

devongovett:
moment exports an ES6 module via the `jsnext:main` field in its package.json: https://github.com/moment/moment/blob/develop/package.json#L26. If you want to `require` it, you'll need to access the default export yourself:

```javascript
var moment = require('moment').default;
```

PR #532, opened 2018-01-10
Fix garbled output files

### LINKED ISSUE #531: JS output files are sometimes garbled
🐛 bug report

JS output files are sometimes garbled:

image

(Notice that the content is bogus.)

The problem goes away by changing `src/Bundle.js` as follows:

image

shawwn:
Closes #531

### DIFF
--- a/src/Bundle.js
+++ b/src/Bundle.js
@@ -65,16 +65,14 @@ class Bundle {
     let hash = this.getHash();
     newHashes.set(this.name, hash);
 
-    let promises = [];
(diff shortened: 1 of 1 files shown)


shawwn:
@devongovett This PR is probably worth investigating.

Since Packager kicks off multiple async/awaits in parallel, it's unsafe to write to the same output stream. I'm seeing partial writes occasionally, which garbles the output file.

devongovett:
Hmm, this seems like a different problem. Why are multiple bundles being produced that write to the same filename?

PR #533, opened 2018-01-10
Fix --no-cache crash

shawwn:


DeMoorJasper:
This fix is obsolete, as cache is invalidated when version number gets upped

PR #534, opened 2018-01-10
Generate DOT graphs

shawwn:
When running e.g. `parcel index.html`, Parcel now generates `dist/index.dot` in addition to `dist/index.html`.

You can then use `dot` or http://www.webgraphviz.com/ to render the graph.

test/dist/index.dot
```
digraph assets {
  "test/\ninput/\nindex.js" -> "test/\ninput/\nfoo.js" [label="./\nfoo"]
}
```

Copy-paste into webraphviz.com:

image

### DIFF
--- a/src/Bundler.js
+++ b/src/Bundler.js
@@ -248,6 +248,17 @@ class Bundler extends EventEmitter {
       asset.invalidateBundle();
     }
 
+    // Graph the asset tree.
+    {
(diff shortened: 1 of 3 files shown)


shawwn:
If you want me to refactor this code to a separate file, let me know.

devongovett:
Could this be written as a plugin? I think it would be possible to listen for the `bundled` event, which emits a Bundle object, and traverse everything from there.

Issue #303, opened 2017-12-16
🙋 Static Site Folder Structure?


mattdesl:
🙋 

I'm looking to create a project with a folder structure similar to this:

```sh
package.json
src/
  index.js
app/
  index.html  <-- not generated
  bundle.js   <-- generated
  assets/
    foo.png   <-- not generated via import
```

Where the code is in one folder (`src`), and the static files are in another (`app`). The same static folder is then used during dev and production.

The main goal of this is a clean repo without unnecessarily duplicated folders/files and clean relative URLs to PNG/JPG/etc assets (not all of which can be imported). The structure is ideal for `gh-pages` demos or basically any small static site.

For an example of how an entire repo and static site will look with this structure:
https://github.com/mattdesl/webgl-wireframes

### 🎛  Test Case

For example, `index.js` which just tries to load an image with a runtime (not imported) URL.

```js
const getRuntimeURL = () => 'assets/foo.jpg';

const url = getRuntimeURL();
const img = new Image();
img.onload = () => {
  console.log('Loaded!');
};
img.onerror = () => {
  console.error('Fail...');
}
img.src = url;
```

### 🎛  Failed Attempts

For now, I'm trying with `index.html` in the `lib` folder.

> `parcel ./lib/index.html --out-dir app/`

This is OK, but the URL needs to be `'/app/assets/foo.jpg'` which means the static site will have to be under `/app`.

> `parcel ./lib/index.html --out-dir app/ --public-url ./`

This gives me a more correct URL pathing in the served HTML, but I get "Uncaught SyntaxError: Unexpected token <" since it's responding with the HTML instead of JavaScript.

### 🤔 Expected Behavior

A way to structure static sites with minimal duplication and clean URLs.

I'm also open to other ways of structuring my site with parcel, just looking for something that could work nicely with the above JavaScript test.

DeMoorJasper:
duplicate of #233

mattdesl:
I've been looking through that issue, and I don't think it's a duplicate.

That issue is asking for an option to specify a *sub-folder* for the output assets. Whereas I am proposing a change in architecture of how parcel handles its paths & static folders to support a simple static site structure, either in the root folder (not possible right now with parcel) or in a sub folder (seems challenging right now).

IMHO there should just be a flag `--dir` that does two things:

- Be the directory that generated files are written to
- Be the root directory that is passed to `serve-static`

[...]

PR #536, opened 2018-01-11
Fix: serve static files that exist in the outDir even when they don't…

naturalethic:
… begin with the publicUrl (#303)


sompylasar:
@naturalethic Hi, could you please update this PR to resolve the conflict?
@DeMoorJasper Hi, what's blocking this besides the merge conflict?

Thanks!

devongovett:
Going to close this. I think it has been resolved in the meantime. I'm not sure how files would get into the outDir that parcel didn't generate the URL for.

PR #543, opened 2018-01-11
Add watch script to make build easier in development

ksashikumar:
Adding watch script will make development easier.


package.json
```diff
@@ -75,6 +75,8 @@
     "test-ci": "yarn test-coverage && yarn report-coverage",
     "format": "prettier --write './{src,bin,test}/**/*.{js,json,md}'",
     "build": "yarn minify && babel src -d lib && ncp src/builtins lib/builtins",
+    "watch":
+      "yarn minify && babel --watch src -d lib && ncp src/builtins lib/builtins",
```
devongovett: Won't it never get to the `ncp` command because the watch sits there forever?

PR #489, opened 2018-01-04
Fix recursive dependencies

shawwn:
Closes #463

PR #548, opened 2018-01-11
Fix circular bundles, and never hoist the entry asset

devongovett:
Slightly simpler, more generic version of #489.

There were two issues:

1. Infinite recursion on circular bundles. This is solved by tracking the current bundle stack.
2. HTML files getting merged into one bundle. This was because Parcel tried to hoist the common dependencies. However, this should never be done for the entry asset of a bundle since those assets were explicitly requested to be placed in a separate bundle (deps marked as `dynamic`).

Added test covers both cases. cc. @shawwn @ipmb

### DIFF
--- a/src/Bundler.js
+++ b/src/Bundler.js
@@ -389,7 +389,7 @@ class Bundler extends EventEmitter {
     this.buildQueue.delete(asset);
   }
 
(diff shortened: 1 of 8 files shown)


ipmb:
works for me!

Issue #433, opened 2017-12-29
🙋 Keep folder structure and filename


zeakd:
Thanks for making this amazing library.

### 🤔 Expected Behavior

``` html
<!DOCTYPE html>
<html lang="en">
<head>
  <meta charset="UTF-8">
  <meta name="viewport" content="width=device-width, initial-scale=1.0">
  <meta http-equiv="X-UA-Compatible" content="ie=edge">
  <title>Document</title>
</head>
<body>
  <a href="./a.html"></a>
  <script src='./index.js'></script>
  <script src='./main.js'></script>
  <script src='./b/index.js'></script>
</body>
</html>
```

Currently, all folder are flatten and filename become hashed with '/dist'. but how about keep the folder structure and filename? like,

``` html
<!DOCTYPE html>
<html lang="en">
<head>
  <meta charset="UTF-8">
  <meta name="viewport" content="width=device-width, initial-scale=1.0">
  <meta http-equiv="X-UA-Compatible" content="ie=edge">
  <title>Document</title>
</head>
<body>
  <a href="/a.html"></a>
  <script src='/index.js'></script>
  <script src='/main.js'></script>
  <script src='/b/index.js'></script>
</body>
</html>
```
### 😯 Current Behavior

``` html
<!DOCTYPE html>
<html lang="en">
<head>
  <meta charset="UTF-8">
  <meta name="viewport" content="width=device-width, initial-scale=1.0">
  <meta http-equiv="X-UA-Compatible" content="ie=edge">
  <title>Document</title>
</head>
<body>
  <a href="/dist/328a03237686de9183d07e1b3ecb7768.html"></a>
  <script src="/dist/example-app.js"></script>
  <script src="/dist/81b87a177e4cfca53b661600c0bf5cd2.js"></script>
  <script src="/dist/b2d29b73372f261f4f9869fe092e7d54.js"></script>
</body>
</html>
```

### 🔦 Context

#303 #323 #307 #280 are related. 

I think It have several advantages.

- easy to understand. it doesn't need many options. 
- dist folder can be served as it is 
- don't need to concern about multiple files with the same name

Maybe I missed your intentions, but in my opinion this feature make parcel more flexible.

zeakd:
I made a PR for this. #557

devongovett:
Should be solved by #1025 which retains filenames and folder structure for linked files, and content hashes static assets. Please help test using the master branch - a release will hopefully come next week!

PR #557, opened 2018-01-13
Keep folder structure and filename

zeakd:
Hi. this PR is related to #433 

This PR does two things.

1. keep filepath (folder structure and filename) 
2. dist folder can be served as it is

### DIFF
--- a/src/Asset.js
+++ b/src/Asset.js
@@ -1,7 +1,6 @@
 const path = require('path');
 const fs = require('./utils/fs');
 const objectHash = require('./utils/objectHash');
-const md5 = require('./utils/md5');
 const isURL = require('./utils/is-url');
 const sanitizeFilename = require('sanitize-filename');
 const config = require('./utils/config');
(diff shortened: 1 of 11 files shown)


devongovett:
See also #307.

TheBlackBolt:
What is the status on this pull request? Also where can I view upcoming release dates?

pauldariye:
Is there update on when this will land?

zeakd:
@TheBlackBolt #307 It seems like they use hashed name for output files because of unique name.  IMO hashed filename that currently used is not usable if parcel keep filename.

dbrw:
IMO this feature is actually a must have.
It is very convenient to have the original directory structure and file names.

marcklei:
Also really need this feature. +1
I will try this branch in my project and share my experience.

Chathula:
waiting for this PR!  this is a must!

Slapbox:
Agreed. Not having this is really causing me problems. You can work around not having it, but it's not a good workflow at all.

bricss:
I'm going `bananas` without this feature.

devongovett:
Should be solved by #1025 which retains filenames and folder structure for linked files, and content hashes static assets. Please help test using the master branch - a release will hopefully come next week!

alexpi:
So, if I understand correctly, it is not possible to keep folder structure for static assets?

MindaugasNarbutas:
same question about keeping folder structure for build? is there any progress?

DeMoorJasper:
Please open up a new issue to discuss this.

Issue #13, opened 2017-12-05
Discuss .babelrc detection, engines field, etc.


devongovett:
Based on twitter convo:

https://twitter.com/dan_abramov/status/938090483166924800

> I’d suggest to reconsider this (or at least limit it to compiling with -env preset only). This encourages authors to publish packages with stage 0 proposals, and ultimately makes builds slower for everyone. Also .babelrc won’t be compatible between Babel versions.

4rlekin:
As i understand the fundamental issue here is to detect how much transpilation (if any) is needed for each third party dependency, correct?

Stop me if it's stupid but couldn't we somehow check `package.json`s of dependencies?
I mean is there really any other way to programatically check what is needed by given project?
Though i imagine it could be potentially costly... :/

benjamn:
Whether or not to honor `.babelrc` files in nested `node_modules` directories is a hot topic over at the Babel repo, too: https://github.com/babel/babel/issues/6766

In short, most npm packages are fully compiled before they are published, so it is wasteful/inappropriate to recompile them (even using their own `.babelrc` files). It would be great if those packages put `.babelrc` in their `.npmignore` files, but that doesn't always happen.

For the minority of npm packages that _do_ need to be recompiled for the browser, the application developer needs to be able to intervene somehow, to enable recompilation selectively. I can imagine various ways of configuring that behavior, but it may be hard/impossible to make the right choices without increasing the configuration burden somewhat.

Taking a step back, Parcel seems to be designed as an alternative to Webpack, which is known for using configuration to solve any and all problems. While "zero-configuration" may be more of a goal than a reality (you still have to configure packagers, e.g.), I think it's a really valuable goal, so it's worth thinking hard about how to solve this problem with minimal additional configuration.

4rlekin:
@benjamn So do I understand correctly that the issue lies in whether it is necessary to recompile given dependency and if so using which `.babelrc`? Because if that's the case maybe we could choose the default behaviour, and allow the other via some CLI flag?

It does seem lika a band-aid of some kind though...

thejameskyle:
I'm generally against doing any kind of transforms on `node_modules`.

Basing things off of `package.json#engines#node` and only using `babel-preset-env` would be an improvement, but we'd still have the problem of unsupported APIs in older versions of Node.

I suppose if someone shipped:

```json
{
  "engines": {
    "ecmascript": "2017"
  }
}
```

[...]

PR #559, opened 2018-01-15
Automatically transpile dependencies with babel-preset-env

devongovett:
Potential fix for #13. This disables `.babelrc` inside node_modules, but provides a way to automatically compile dependencies to the target browsers of the app. It compares the browser support matrix for the source module and the target app, and automatically runs only the necessary babel plugins. This allows modules to be published to npm as ES2015 or higher, and automatically get transpiled only to the level that the app desires rather than all the way down to ES5.

Modules can specify the browser and node versions that they were written to support via `package.json`, `.browserslistrc` file, etc. Apps also specify their target browser/node version support the same way. Parcel calls babel-preset-env to generate a list of plugins for each, and does a diff to only include the necessary plugins to get from the source language version to the target.

Full target location list:

* package.json `engines.node` field
* package.json `engines.browsers` field
* package.json `browserslist` field
* `browserslist` or `.browserslistrc` files
* `.babelrc` or `.babelrc.js` files with `babel-preset-env` preset config (questionable)

Currently the list of locations Parcel searches is the same for both modules and the target app. I'm not sure if this is correct or whether only `engines` should be used for modules. Will searching in browserslist or babelrc files cause issues if they are published to npm? Theoretically modules will already be pre-compiled to the level of the published babelrc/browserslist so it seems relatively safe to use that as the source support matrix along with engines etc. Slightly worried about babelrc format changing in the future though.

Would be interested in feedback here!

To do:

- [x] Fix tests
- [x] Add more tests
- [x] Test performance
- [x] Test in real world apps

cc. @gaearon @thejameskyle


src/transforms/babel.js
```diff
@@ -26,19 +21,81 @@ module.exports = async function(asset) {
...
+  let config = await getEnvConfig(asset);
+
+  // If this is the app module, and there is a .babelrc, merge it into the config.
+  // Don't load .babelrc for node_modules. https://github.com/parcel-bundler/parcel/issues/13.
+  if (!asset.name.includes('/node_modules/')) {
+    let babelrc =
+      (asset.package && asset.package.babel) ||
+      (await asset.getConfig(['.babelrc', '.babelrc.js']));
+    if (babelrc) {
+      config = Object.assign({}, babelrc, {
```
devongovett: Should we merge the babelrc with the generated env config or just override if you specify a babelrc?

src/transforms/babel.js
```diff
@@ -26,19 +21,81 @@ module.exports = async function(asset) {
...
+  let sourceEnv = await getEnvPlugins(sourceEngines);
+  let targetEnv = await getEnvPlugins(targetEngines);
+
+  if (!sourceEnv || !targetEnv) {
+    return;
+  }
+
+  // Do a diff of the returned plugins. We only need to process the remaining plugins to get to the app target.
+  // If this is the app module, the source and target will be the same, so just compile everything.
+  if (asset.name.includes('/node_modules/')) {
```
devongovett: Need a nicer way to detect files that are in the app rather than a module

src/transforms/babel.js
```diff
@@ -26,19 +21,81 @@ module.exports = async function(asset) {
...
+  return config;
+}
+
+module.exports.getConfig = getConfig;
+
+async function getEnvConfig(asset) {
+  // Load supported source engines (node and browser versions) for source module and target app,
+  // and generate a babel-preset-env config for each.
+  let sourceEngines = await getTargetEngines(asset);
+  let targetEngines = await getTargetEngines(asset, asset.options.mainFile);
```
devongovett: We should cache this somehow so we don't recompute the target engines for every single file

src/transforms/babel.js
Andarist: there might not be any, we've discussed this recently at babel and it seems there is no better solution to see if a file is node_module or not than to look at its path, and we went with this approach for this kind of detection

gaearon: In CRA we're currently just processing anything that's *not* in application `src` folder with `-env`. Is that wrong?

yavorsky:
BTW, We already have a PR for preset-env that adds `engines` from the `package.json` support: https://github.com/babel/babel-preset-env/pull/114

It presumes `engines.node` field, but we could extend it to consider `engines.browsers` field also.

Unfortunately, we've focused on the other things, but I can finish it in the near future and merge if it will be useful for your intention.
Then, you can just reuse logic provided in preset-env, which already can [search browserslist in package.json and browserslist config files](https://github.com/babel/babel-preset-env/pull/161).

src/transforms/babel.js
devongovett: That's fine if you know something about the file structure (that there is a src folder), but Parcel needs to work for all types of projects.

aaqibkhaan:
Hi


Can you please unsubscribe me from all the parcel-bundler threads please.


thanks,

Aaqib


________________________________

devongovett:
@aaqibkhaan hit the unwatch button at the top of the repo, or on an individual thread.

src/transforms/babel.js
gaearon: Thanks for explaining!

src/transforms/babel.js
```diff
@@ -26,19 +21,81 @@ module.exports = async function(asset) {
...
+    return;
+  }
+
+  // Do a diff of the returned plugins. We only need to process the remaining plugins to get to the app target.
+  // If this is the app module, the source and target will be the same, so just compile everything.
+  if (asset.name.includes('/node_modules/')) {
+    let sourcePlugins = new Set(sourceEnv.map(p => p[0]));
+    targetEnv = targetEnv.filter(plugin => {
+      return !sourcePlugins.has(plugin[0]);
+    });
```
gaearon: I wonder if there's any possible way to make a Babel preset handle this automatically. For example we'd like our ejected users to just have `babel-preset-react-app/dependencies` for non-app code, and have that automatically do the diffing.

devongovett: Yeah totally, I was thinking that too. The only piece of information it needs is the entry point to the bundler (`asset.options.mainFile` in above in parcel's case) in order to resolve the target app's browser list. If this information were somehow provided to babel plugins via bundlers it would be possible to do in a babel preset (or even in env itself). Maybe just resolving based on the `process.cwd` would be good enough though.

devongovett: The other thing we need in Parcel's case is to track the config files we load (e.g. `package.json`, `.browserslistrc`, etc.) so they can be added to the file watcher and cache correctly. That way when you change your browser support file, we can recompile files automatically. Would be cool if Babel presets could emit dependencies as well.

Andarist: I think I we are open for this stuff at babel, just need to specify the requirements, how this would work etc. Dunno if anyone want to focus on that at the moment as we'd like to release stable babel@7 - we could always use an outside contribution for this though 😉 

The problem is that babel at the moment has no notion of project, it operates on single files within a directory structure. And it comes with support for nested config files, overrides and stuff, which could complicate things.

But I guess what is really interesting here is a single "targets" field/config for a project, usually kept at the root of the project which is a first package.json NOT within a node_modules, it should be quite easy to locate it without much hassle and without introducing much complexity to the codebase. 

Let's open an issue on babel boards to discuss this further, we'd want to have a short summary of the problem and proposed solution for it as a discussion starter.

gaearon: >But I guess what is really interesting here is a single "targets" field/config for a project, usually kept at the root of the project which is a first package.json NOT within a node_modules, it should be quite easy to locate it without much hassle and without introducing much complexity to the codebase.

Just to clarify, I don't think the *targets* are relevant here. Env preset already can determine target browsers desired by the **app** (through browserslist).

What seems missing from Babel's side is inferring a **library's** supported `engines` and then using *those* as "minimal guaranteed" environment heuristic.

For example if my app specifies `IE 11` as a target via `browserslist`, we know ES6 needs to be compiled down. But if React says via `engines` that it runs on Node `>= 0.12` we know that it can't possibly include ES6 code. Therefore we don't have to run React through ES6 transforms. For this to work, we need presets to know they're dealing with React package (and what `engines` it has in `package.json`).

Does this make sense? Sorry if you said the same and I misunderstood.

Andarist: We are on the same page on this, I just forgot to mention need for a field in packages - hence the confusion. 

There are related discussions about similar topics on babel boards, cannot find them on mobile so will post links 2morrow.

Im not sure if it has been discussed here but we'd have to standardize a new field for package.json - smth like "source". Using "main"/"module" at the moment would cause too much problems at this point

gaearon: Why do we need that field? I'm not quite sure I'm following. At least in CRA I don't plan to compile third-party packages from source—maybe I misunderstood your intention?

Andarist: Well, I've chosen the unfortunate name for this just before hitting the bed. I didnt mean the literal "source", transpiling is becoming much more than just new shiny (often experimental) syntax with typings, macros, prevals and other cool stuff. We do not want to ship all that stuff obviously to npm.

We'd have to recommend shipping (somehow) "transpiled down to latest (?) syntax" to npm. It's a problem as latest is a moving target. What if one sets up his build tools with let's say babel@7 in place and everything works, but babel@8 gets out and "latest" is different - new syntax got standardized and people start publishing their libraries with babel@8 in mind? I believe we can figure out this problem, but we need to evaluate how it can be solved.

Many people believe that fields like "main" shouldn't require much tooling and preferably they should be runnable at least in all LTS versions of node. Using "main" for this whole idea would pretty much force consumer to use babel.

We've started some conversations about similar topics [here](https://github.com/babel/babel/issues/6328).

src/transforms/babel.js
thejameskyle: I don't think that we should allow people to specify Babel transforms for `node_modules`, I think it should be exclusively `babel-preset-env` and nothing else.

Otherwise we're going to end up with all these things in the ecosystem which tell you that you need to be compiling their package in a certain way. Package authors will either say:

- You need to add this Babel plugin in order to use this package
- You cannot use these Babel plugins in this way or it will break this package

If we want to change this restriction in the future, it's easier to loosen up than to make more strict. But I think this is the right way to go

src/transforms/babel.js
gaearon: Sorry, I believe I completely lost you 😥 

Let me reiterate how I view history:

* npm ecosystem used to publish things in ES5 because Node used ES5
* Client-side apps started depending on npm ecosystem and also used ES5
* Adventurous folk started using ES6 in npm packages, but since Node runs ES5, they had to Babel it down to ES5
* Node started supporting ES6
* Maintainers of popular Node packages no longer want to have a build step before publish since they mainly care about Node users and think that client-side developers can compile themselves (and some client-side developers also are fine with shipping ES6 in the first place)

And how I see the solution:

* Package authors continue publishing JavaScript (any version!) to npm as long as it's only pure valid JavaScript that runs on Node and not some JSX/Flow/stage-X. There's no separate entry point necessary for "new" or "old" code, they just publish the JS code like they always did.
* Client-side app developers compile `node_modules` using `babel-preset-env` that considers their target (e.g. "IE 9"). We already know `node_modules` always contain *only* pure JS so this always works, and code in packages never depends on specific Babel versions.
* As a potential *optimization*, we could *skip* compiling some packages down to our app's target (i.e. "iOS 10") if we know they already claim to be compatible with a "lower" target (e.g. Node 0.12). I want to reiterate this is an *optimization* and does not fundamentally change anything, except it lets us run less transformation. The output wouldn't change because we assume those features that we planned to compile don't exist in the package (per its claim of compatibility with a known lower target).

Where does my picture not match yours? I don't quite see how the `source` (or `latest` etc) entry point fits into this picture.

src/transforms/babel.js
gaearon: Agreed

src/transforms/babel.js
gaearon: In other words, what I'm trying to say is that I don't think we should solve the problem of "new Node version comes out that supports ES10, and thus popular packages released ES10 versions, making them incompatible with consumers who expect ES9". I think this is up to every specific maintainer to choose which versions of Node they want to support in their package.

To the Node consumers, there is a choice: either migrate to that version of Node, or to use the previous major version of the package, or to compile it with `preset-env` and never have to think about that.

The client-side consumers can always compile with `-env`. As an *optimization*, they can skip compiling a package if claims to only contain a code written against an old enough engine (e.g. Node 4) that we know for sure it's a subset of the version our app targets (e.g. Chrome latest).

Of course, all of this only works if we agree that packages on npm can only contain pure JS code.

src/transforms/babel.js
devongovett: @thejameskyle see check above. This only happens in the top-level app, not node_modules. In that case you could have a babelrc that we need to respect, as well as a a browserslist file which we use to enable the env transforms needed. In that case do we merge, or only use the babelrc?

src/transforms/babel.js
ljharb: Compiling "native javascript" comes with caveats whose safety only the author of the code can truly understand - for example, what if the code depends on `(() => {}).prototype` being `undefined`, but the transpilation can't offer that? It's simply not safe to transpile code you didn't write.

gaearon: That's a great point, thanks for mentioning it! I asked in a few places but haven't considered this case.

jaydenseric:
@yavorsky I had been waiting for that PR to merge for ages, [looks like it did not get migrated to the mono-repo](https://github.com/babel/babel/pulls?q=is%3Aopen+is%3Apr+label%3A%22pkg%3A+preset-env%22). It would be awesome if you could get the ball rolling again.

I have been manually importing the node version from `package.json` `engines` [in .babelrc.js](https://github.com/jaydenseric/apollo-upload-client/blob/v7.0.0-alpha.2/.babelrc.js#L1).

src/transforms/babel.js
devongovett: Even if we can't transpile perfectly in some edge cases, I think its still more valuable more of the time than not transpiling anything at all and having dependencies that don't work in an app's target engines. If a dependency is written in ES6, it won't work on IE 9 either way, but it is more likely to work with transpiling than without.

ljharb: Yet, if that dependency was transpiled prepublish, the likelihood is even higher - which is why the community best practice remains, despite a few outlier packages and the react-native ecosystem, to publish a "main" that's CJS ES5.

Andarist: @gaearon 
I have in mind pretty much exactly what you have said, regarding the history, recent practices, proposed approaches. Except for this whole "main"/"some_other_field" point.

I dont exactly disagree with you on that, I only have seen this from slightly different angle. I just often see people considering "main" as some kind of "stable" es5 version. 

You envision this being an enhancement for consumers, letting package authors to chose what they want to support etc. I thought of this (not that it is any vision of mine, and I dont agree with your approach, just had different goal in mind while we've started discussing this) more of an "enhancement" provided by package authors (with different field in the package.json) for consumers. So I have considered more opt-in situation here, whereas your solution with `main` makes it more obligatory to use tooling supporting `node_modules` transpilation (babel-preset-env). 

My original thought hasnt really contradicted yours that much. With your approach this makes it kinda "ok, lets ship whatever you want, whatever you find appropriate, we can take care of the rest". With my original approach it was more like "let's ship whatever you want as main, we cant strictly forbid that, we'll take care of the rest, BUT if you want to support older envs right out of the box please use `main` for completely transpiled version of your code and `some_other_field` as an enhancement for users who'd like to benefit from less transpilation".

gaearon: Now it clicks, thanks so much for explaining! Yeah I can see how this could be valuable although indeed the shifting baseline is a bit confusing.

Andarist: I just expect most libraries supporting at least all current LTS versions which in a sense might mean "overtranspiling" `main` for a part of the user base that would like to target newer envs.

src/transforms/babel.js
thejameskyle: I'm not a fan of adding things to Babel configs without people actually specifying them, it's something that really annoys me with Jest

devongovett:
Updated this PR to implement the strategy described in https://github.com/parcel-bundler/parcel/issues/13#issuecomment-361095861. It continues to work the same way as above, but in addition allows modules to specify that they want to use a specific babel config. This allows e.g. npm linked modules to be compiled correctly without a manual build step.

To summarize, you can use the following config in package.json to enable babel on node_modules:

```json
{
  "bundlerOptions": {
    "babel": true
  }
}
```

In addition, the babel config is merged with the babel-preset-env generated config that takes into account the app's browser target. Unnecessary presets and plugins are removed, so even if a module specified the `es2015` preset, modules will only be actually be compiled to the app's defined browser target. I think this is fairly safe to do, but let me know if there is a case where this wouldn't work.

The locally installed version of `babel-core` specified by the module is used to do the compilation. This should allow for future babel updates while maintaining compatibility with existing modules that don't update.

devongovett:
Check out the tests to see some examples of how it works.

src/utils/getTargetEngines.js
```diff
@@ -0,0 +1,89 @@
...
+  }
+
+  if (targets.browsers) {
+    targets.browsers = browserslist(targets.browsers).sort();
+  }
+
+  return targets;
+}
+
+function getMinSemver(version) {
```
jaydenseric: Related: https://github.com/npm/node-semver/pull/208. This is actually a pretty tricky thing to do.

devongovett: Yep. I think what I have here might be good enough for what we're using it for though, which is only parsing the `package.engines.node` field. It's unlikely to contain any super complex semver versions in there.

devongovett:
Ok removed `bundlerOptions` for now while we debate future options (e.g. `source` field etc.) in #13. Going to merge this as is. This disables `.babelrc` for node_modules, and enables automatic babel-preset-env compilation.

piscisaureus:
My project uses typescript, which already knows how to target various ecmascript versions.
There is no need to have babel go over the javascript that this produced by the typescript compiler.
Is there a way to turn this feature off?

src/transforms/babel.js
timwis: Is there a conclusion or a next chapter to this convo somewhere? I ran into the issue this morning and had a very similar idea. I wonder if we can borrow insight from browserify's approach of libraries [explicitly defining](https://github.com/browserify/browserify-handbook/blob/master/readme.markdown#browserifytransform-field) their transforms in package.json and including them as devDeps so consumers don't need to worry about it. I imagine you're still agreeing on a baseline target to some degree, but this could let downstream compilers opt out of certain transforms. Anyway, wanted to get thoughts on that and see if this convo has developed into a best practise somewhere. Is it recommended to transpile your node_modules directory with preset-env now, both in dev and prod?

RPDeshaies:
Following @piscisaureus  ' comment, is there a way to use the Typescript compiler to bundle node modules instead since it's already there  ?

PR #560, opened 2018-01-15
[RFC/WIP] Only starts workers on big projects

DeMoorJasper:
This speeds up performance on small projects, while maintaining the same performance (minus a small delay due to getting folder size)


src/WorkerFarm.js
```diff
@@ -2,8 +2,10 @@ const {EventEmitter} = require('events');
 const os = require('os');
 const Farm = require('worker-farm/lib/farm');
 const promisify = require('./utils/promisify');
+const path = require('path');
 
 let shared = null;
+const SYNC_TRESHOLD = 50000; // Treshold to start workers in bytes
```
devongovett: Curious how you came up with this threshold 😜 

Also, is it faster to compute file size or just count the number of files? Which is a better heuristic?

brandon93s: Where did this number come from? 0.05 Megabyte doesn't seem like a very "big" project. Adding just a single image to the project would cross the threshold.

yarn-error.log
```diff
@@ -0,0 +1,5734 @@
+Arguments:
```
devongovett: let's not commit this file

src/WorkerFarm.js
brandon93s: Gah. Same comment, same time!

brandon93s:
@DeMoorJasper - Is there a few "real world" projects we've been using the benchmark the various proposed WorkerFarm enhancements? I'd like to benchmark them vs master on Windows 10.

yarn-error.log
DeMoorJasper: Totally looked over that file, will remove it

DeMoorJasper:
@brandon93s i used react-news for my tests as it's about a second faster with workerfarm https://github.com/echenley/react-news

src/WorkerFarm.js
DeMoorJasper: I'm pretty sure it'll be faster to count them, i'll rework it, to also include node_modules and do some benchmarks comparing count vs size

DeMoorJasper: That's kind of an issue with this approach it counts in images, i'll probably pivot towards counting for this reason and maybe performance. According to my benchmarks this is near the treshold of where things start to get quicker but i just noticed it doesn't take node_modules into consideration so...

DeMoorJasper:
So on a fairly large codebase i tested the performance of `du` vs the fileCounter i added.
```
du: 1546ms
fileCounter.js= 1031ms
`
``
I'm not sure if this is such a solid solution after adding node_modules to the mix it more likely became a bottleneck than an actual performance improvement

devongovett:
Yeah the file counting is a bit slow. You could maybe optimize it a bit to only count up to the threshold rather than continuing after that.

I made an alternative PR though which seems to be fairly promising. Have a look: https://github.com/parcel-bundler/parcel/pull/567

DeMoorJasper:
Your pr seems to be the proper solution i'll close this one

PR #562, opened 2018-01-15
Improve Code Coverage

brandon93s:
Goals:

- Improve code coverage
- Start testing some classes directly
- Add sinon to begin mocking

Notes:

- There is some low hanging fruit, namely testing failure and exception branches, that can be tested.


src/Logger.js
```diff
@@ -108,6 +108,10 @@ class Logger {
       this.lines++;
     }
   }
+
```
brandon93s: Internal _log function to help with stubbing, and to avoid stubbing console.log directly (nasty side-effects).

src/utils/customErrors.js
```diff
@@ -1,15 +1,17 @@
-let serverErrorList = {
+const serverErrorList = {
   EACCES: "You don't have access to bind the server to port {port}.",
   EADDRINUSE: 'There is already a process listening on port {port}.'
 };
 
 function serverErrors(err, port) {
```
brandon93s: Code was previously built to handle unknown errors `if (!desc) ...`, but an exception on the `.replace` call would prevent this code path from ever executing. 

Exposed by tests, fixed.

src/utils/fs.js
```diff
@@ -5,6 +5,7 @@ const mkdirp = require('mkdirp');
 exports.readFile = promisify(fs.readFile);
 exports.writeFile = promisify(fs.writeFile);
 exports.stat = promisify(fs.stat);
+exports.unlink = promisify(fs.unlink);
```
brandon93s: See notes in fs-cache tests.

test/fs-cache.js
```diff
@@ -0,0 +1,142 @@
...
+
+  it('should return null for invalidated resources', async () => {
+    const cache = new FSCache({cacheDir: cachePath});
+    cache.invalidate(__filename);
+
+    let cached = await cache.read(__filename);
+    assert.equal(cached, null);
+  });
+
+  it('should remove file on delete', async () => {
```
brandon93s: `cache.delete` was not working due to `fs.unlink` missing from the `fs` utility. The implementation catches the error and fails silently. Does not appear to be used today, but is now working correctly. 

Exposed by tests, fixed.

test/fs-cache.js
```diff
@@ -0,0 +1,142 @@
...
+          name: __filename
+        }
+      ]
+    });
+
+    const cached = await cache.read(__filename);
+    assert.equal(cached.dependencies[0].mtime, mtime);
+    assert.equal(cached.dependencies[1].mtime, undefined);
+  });
+
```
brandon93s: Added tests for #514.

test/integration/posthtml/.posthtmlrc.js
```diff
@@ -1,7 +1,7 @@
 module.exports = {
```
brandon93s: Gives us coverage on the array branch of `loadPlugins`:

https://github.com/parcel-bundler/parcel/blob/b56b2a9f3c3ff6db2dd27a086b409a8d4af6f2bd/src/utils/loadPlugins.js#L4-L7

Other posthtmlrc is object, for complete coverage.

devongovett:
Awesome, thanks for working on this! 🥇

brandon93s:
Tests have been updated with master. Notes & explanations can be found in the current and outdated self-review.

PR #563, opened 2018-01-16
Add out-file option to set the output file name

jkrehm:
Add out-file option to set the output file name

If one is not set, default to current behavior (sanitized version of package name).

In a project I'm working on, we require the output files to have a specific name, and changing the package.json name is a less desirable solution vs expanding the API to accommodate.

I couldn't find any tests that reference `outDir` to copy to test this option, so I did not include any tests. I'm not familiar enough with the project or Mocha to greenfield it.

### DIFF
--- a/src/Asset.js
+++ b/src/Asset.js
@@ -174,9 +174,11 @@ class Asset {
 
     // If this asset is main file of the package, use the sanitized package name
     if (this.name === main) {
-      const packageName = sanitizeFilename(this.package.name, {
-        replacement: '-'
-      });
+      const packageName =
+        this.options.outFile ||
+        sanitizeFilename(this.package.name, {
+          replacement: '-'
+        });
       return packageName + ext;
     }
 
--- a/src/cli.js
+++ b/src/cli.js
@@ -29,6 +29,10 @@ program
     'set the output directory. defaults to "dist"'
   )
   .option(
+    '-f, --out-file <filename>',
+    'set the output filename. defaults to the package.json "name"'
+  )
+  .option(
     '--public-url <url>',
     'set the public URL to serve on. defaults to the same as the --out-dir option'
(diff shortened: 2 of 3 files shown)


muminoff:
I would love this PR too!

src/Asset.js
```diff
@@ -174,9 +174,11 @@ class Asset {
 
     // If this asset is main file of the package, use the sanitized package name
     if (this.name === main) {
-      const packageName = sanitizeFilename(this.package.name, {
-        replacement: '-'
-      });
+      const packageName =
+        this.options.outFile ||
```
devongovett: This currently only gets used when the asset matches `main` in package.json. I guess I'd expect it to also work when it doesn't match, or there is no `main` defined (below case). Maybe add an additional case above this if statement to always return `this.options.outfile` if it is defined?

src/cli.js
```diff
@@ -28,6 +28,10 @@ program
     '-d, --out-dir <path>',
     'set the output directory. defaults to "dist"'
   )
+  .option(
+    '-f, --out-file <filename>',
```
devongovett: Can we make the alias `-o` instead? I see it's already aliased to `--open`, but IMO that doesn't need an alias and `-o` should alias to `--out-file` instead.

jkrehm:
As I stated in my initial comment, I'm not familiar enough with Mocha or the codebase of this project to write a test from scratch, and as far as I could tell, there is not a test for `outDir` for me to pattern.

In addition, I've switched away from Parcel so I'm happy to make small changes to the commit, perhaps it would be better for a current user to take over, especially if it results in breaking changes to the API.

brandon93s:
@jkrehm - Thanks for getting this started! I've continued these efforts in #745.

PR #565, opened 2018-01-16
Custom bundle loaders + WASM support -> master

devongovett:
See #473. This omits Rust -> WASM support for now - will come in a separate PR.

cc. @albizures


src/builtins/loaders/js-loader.js
```diff
@@ -0,0 +1,20 @@
+module.exports = function loadJSBundle(bundle) {
+  return new Promise(function (resolve, reject) {
+    var script = document.createElement('script');
+    script.async = true;
+    script.type = 'text/javascript';
```
niieani: You don't need to set explicit `async` or `type`, these are the defaults.

PR #567, opened 2018-01-16
Warm up worker processes before switching to them

devongovett:
While the workers are starting up, we already process assets on the main thread. But it turns out that even after the workers have started, it takes a bit of time for them to warm up. The JS engine needs to do JIT things, optimize the code of some giant libraries like babel, etc.

We now send assets to both the worker and main process simultaneously in order to warm up the workers, and then transition to the workers only once they are warmed up. This seems to cause a pretty big improvement in perf for small apps, while large apps remain pretty much the same.

**Small app (`import 'react'`):**

Without patch (with workers): 752ms
Without workers: 357ms
With patch: 487ms

**Large app:**

Without patch (with workers): 6.76s
Without workers: 15.53s
With patch: 6.77s

cc. @DeMoorJasper @brandon93s Want to do some testing?

### DIFF
--- a/src/WorkerFarm.js
+++ b/src/WorkerFarm.js
@@ -8,7 +8,6 @@ let shared = null;
 class WorkerFarm extends Farm {
   constructor(options) {
     let opts = {
-      autoStart: true,
       maxConcurrentWorkers: getNumWorkers()
     };
 
@@ -18,6 +17,7 @@ class WorkerFarm extends Farm {
     this.remoteWorker = this.promisifyWorker(this.setup(['init', 'run']));
 
     this.started = false;
+    this.warmWorkers = 0;
     this.init(options);
   }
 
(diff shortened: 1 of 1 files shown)


brandon93s:
👍 Seeing a consistent ~15% speed improvement [examples/Vue](https://github.com/parcel-bundler/examples/tree/master/Vue) and ~45% improvement on [examples/hyperapp](https://github.com/parcel-bundler/examples/tree/master/hyperapp) on my PC. No improvement for [examples/react](https://github.com/parcel-bundler/examples/tree/master/react) - maybe a slight slowdown if anything.

DeMoorJasper:
I've tested in on react-news (which is my go to repo for the worker tests) there i noticed a difference of -0.2secs, so it appears this also speeds up medium sized application building.
On a supertiny application it took half as long as on the master branch, so about as fast as synchronous.
Seems to be working extremely well 💯

PR #573, opened 2018-01-17
Add pug support

danielpa9708:


src/assets/PugAsset.js
```diff
@@ -0,0 +1,18 @@
+const HTMLAsset = require('./HTMLAsset');
+const localRequire = require('../utils/localRequire');
+
+class TypeScriptAsset extends HTMLAsset {
```
brandon93s: `TypeScriptAsset`  still here from copied file.

brandon93s:
@danielpa9708 - Have you seen @Ty3uK 's [parcel-plugin-pug](https://github.com/Ty3uK/parcel-plugin-pug)? It seems to be a fair bit more comprehensive (resolving dependencies, filter support, pug & jade, etc.) and may suit your needs.

src/assets/PugAsset.js
danielpa9708: My bad, couldn't test it because I have node v6

danielpa9708:
@brandon93s , I didn't know there where plugins for parcel, why is it not include in the main repository ?

danielpa9708:
Sorry for the commit and fail test mess, I'm using node v6 so can't test locally.

Will this PR interfere with parcel-plugin-pug ?

DeMoorJasper:
@danielpa9708 it will not interfere with the pug plugin as it simply takes the last registered asset handler and plugins get registered at the end.
About the why is it not included in the core, it's probably because it's not so wildly adopted as some other asset types we have inside the core (i might be mistaken here)

Ty3uK:
@danielpa9708 wihout manual walking through AST you cannot trigger rebuild when user change included or extended file.

danielpa9708:
@Ty3uK , I don't know how to work around that, I just heard about parcel yesterday, it would be great if you or someone more experienced than me could fill the gaps of this PR

Ty3uK:
@danielpa9708 you can check https://github.com/Ty3uK/parcel-plugin-pug

danielpa9708:
Is it ok if I just replace my PR with parcel-plugin-pug code

brandon93s:
@danielpa9708 you should be able to just use parcel-plugin-pug directly without needing to add it to parcel:

```shell
# install parcel and pug (or use npm)
yarn add parcel-bundler --dev
yarn add parcel-plugin-pug --dev

# add scripts to package.json
{
   "scripts": {
		"start": "parcel ./index.html --open"
   }
}

# start parcel. it will automatically load the plugin for you. pug should now be supported:
yarn start
```

Issue #544, opened 2018-01-11
🐛 Tags that do not belong to HTML should be ignored


highsea:
For services rendered as the template file, the template engine ```<%%>```,```{{}}``` marker package should be ignored, but the label disrupted

### 💻 Demo Code

```
<!DOCTYPE html>
<html>
<head>
	<title><%=title%></title>
	<link href="./style.css">
</head>
<body>
	<h1>Welcome : </h1>
<% if(data){ %>
	<p><%=data%></p>
<%}else{%>
	<p><%=request.session._flash||' Error '%></p>
<%}%>
</body>
<script src="./index.js"></script>
</html>
```

### 🤔 Expected Behavior

```
<!DOCTYPE html> <html> <head> <title><%=title%></title> <link href="/dist/1fb0a95e00274611850900f80a9d9d1e.css"> </head> <body> <h1>Welcome : </h1> <% if(data){ %> <p><%=data%></p> <%}else{%> <p><%=request.session._flash||' Error ' %></p> <%}%> </body> <script src="/dist/e2e07bfc7be96fd730ba9a0cd2160e84.js"></script> </html>
```

### 😯 Current Behavior

```
<!DOCTYPE html> <html> <head> <title><%=title%></%=title%></title> <link href="/dist/1fb0a95e00274611850900f80a9d9d1e.css"> </head> <body> <h1>Welcome : </h1> <% if(data){="" %=""> <p><%=data%></%=data%></p> <%}else{%> <p><%=request.session._flash||' Error="" '%=""></%=request.session._flash||'></p> <%}%> </%}%></%}else{%></%></body> <script src="/dist/e2e07bfc7be96fd730ba9a0cd2160e84.js"></script> </html>
```

### 💁 Possible Solution

```
parcel build test/demo.html -d ./test/dist/ --mangle-tag=<%%>,{{}}
```

### 🌍 My Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Parcel           | 1.4.1 |
| Node             | v8.8.0 |
| npm/Yarn         | Yarn |
| Operating System | MacOS 10.12.6 |

okyantoro:
maybe we can add add custom tags for ```{{}}``` and ```<% %>``` in ```HTMLAsset``` when parsing and rendering HTML file to solve this issue

DeMoorJasper:
[...]

PR #578, opened 2018-01-18
feat: ignore "<%" server html tags #544

maciej-ka:
Upgrades posthtml-render to get a recent feature that was added to close issue #544.


devongovett:
Closing. I don't think you should be running template files through parcel - you probably want to run them through the template engine first.

PR #580, opened 2018-01-19
[WIP/RFC] Rust support

albizures:
What does this PR add?

* install wasm32-unknown-unknown
* use rustc or cargo to build the wasm file
* install and use wasm-gc for builds
* add the cdylib crate-type if Cargo.toml file doesn't have it yet

What is it missing?
* test in windows and linux
* test in a free-Rust environment


devongovett:
Awesome work @albizures!! 🏅 I refactored a few things in #623 , but overall this was really well done!

Merging. Let's follow up with support for importing JS from WASM/Rust as well. We will need to discuss how to implement that! 🚀

PR #587, opened 2018-01-19
[WIP/RFC] Warn when a fs call cannot be evaluated

fathyb:
screen shot 2018-01-19 at 16 42 09

This PR will warn instead of throwing when a `fs` call cannot be evaluated.

Remarks :
- I added a commit to create a logger singleton, we might want to wait for #462 to land first and then remove the commit.
- It warns using a code frame, maybe it's too much, waiting for your feedback for that
- The warning will not be propagated to the browser console
- If the call cannot be evaluated then the `fs` module will be required as an empty module (`Uncaught TypeError: fs.readFileSync is not a function` on the runtime)
- See the `nodeNotEvaluated` var in `fs.js`, I don't know if it's elegant
- Should we handle the case where the filename is not a string? (eg. `readFileSync({notAString: true})`)

TODO:
- [x] warn
- [x] add a sweet emoji right before the start of the warning (#588)
- [x] add tests
- [x] needs #462 for the logger

### DIFF
--- a/src/Logger.js
+++ b/src/Logger.js
@@ -44,12 +44,16 @@ class Logger {
     this.write(this.chalk.bold(message), true);
   }
 
-  warn(message) {
+  warn(err) {
     if (this.logLevel < 2) {
       return;
     }
 
+    let {message, stack} = prettyError(err, {color: this.color});
     this.write(this.chalk.yellow(`${emoji.warning}  ${message}`));
+    if (stack) {
+      this.write(stack);
+    }
   }
 
   error(err) {
--- a/src/visitors/fs.js
+++ b/src/visitors/fs.js
@@ -2,6 +2,8 @@ const t = require('babel-types');
 const Path = require('path');
 const fs = require('fs');
 const template = require('babel-template');
+const codeFrame = require('babel-code-frame');
+const logger = require('../Logger');
 
 const bufferTemplate = template('Buffer(CONTENT, ENC)');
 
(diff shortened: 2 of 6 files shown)


src/Logger.js
```diff
@@ -3,7 +3,25 @@ const readline = require('readline');
 const prettyError = require('./utils/prettyError');
 const emoji = require('./utils/emoji');
 
+const loggerMap = new WeakMap();
+
 class Logger {
+  static forAsset(asset) {
```
DeMoorJasper: Won't this bug the logger if multiple loggers exist?
Mainly like glitches with the status

fathyb: Oh yeah, this is just to quickly showcase the feature. A IPC system is way better like you did in the PR, like said in the description I'll remove this commit and use your PR once it gets merged.

However I'm not sure if Node.js scrambles the logs when `stdout` is shared across processes with `child_process` (it sure does when each `stdout` has it's own descriptor).

DeMoorJasper: Ow alright, thought this was another way to implement a singleton, guess i misunderstood.
About the workers stdout, I think it's best if we either don't use it or catch it on the master process and redirect it to the logger to make sure everything is streamlined to prevent console glitches

test/integration/fs-options-non-evaluable/index.js
```diff
@@ -0,0 +1,5 @@
+const dir = __dirname
```
fathyb: This is weird. This does not work :

```js
module.exports = require('fs').readFileSync(__dirname + '/test.txt', {
  encoding: (typeof Date.now()).replace(/number/, 'utf-8')
})

```

While `const dir = __dirname` works. I think it may be (not sure at all) a [babylon-walk](https://github.com/pugjs/babylon-walk) bug, but I'm not really sure about that. I checked the `Identifier` visitor and it is not called with the node, but it seems to be in the AST. I can't reproduce it outside of this test 😕

src/Logger.js
fathyb: I'm more for just using the IPC and not messing too much around `stdout` (Windows + `stdout` + multiple processes don't blend well in my experience).


`[bit off-topic]`
And I really like the IPC idea, all my plugins are needing an IPC at some point and I ended up making my own using UNIX sockets. If we could streamline this and make it accessible to plugin developers it could really help.
`[/bit off-topic]`

src/visitors/fs.js
```diff
@@ -31,9 +37,37 @@ module.exports = {
         __dirname: Path.dirname(asset.name),
         __filename: asset.basename
       };
-      let [filename, ...args] = path
-        .get('arguments')
-        .map(arg => evaluate(arg, vars));
```
devongovett: Maybe instead of changing the logic to be a loop that exists early as below, you could just wrap this in a `try...catch` block and do the logging in the catch. That way you'd also not need the `nodeNotEvaluated` thing. Might clean up the code a bit.

src/Logger.js
devongovett: Jasper's logger changes have been merged to master, so the IPC should just work now. You can just use the Logger singleton within the worker and things will be transparently proxied.

fathyb:
screen shot 2018-01-31 at 22 38 08

I pushed a new commit according to the review and added a code frame where `fs.readFileSync` throws.

devongovett:
Thanks! Cleaned it up just slightly to take advantage of the existing `generateErrorMessage` function on JSAsset. That will also provide a non-highlighted code frame for CLIs that don't support color.

Also needed to fix a couple things in WorkerFarm so that we don't log things when workers are warming up - caused duplicate messages to appear.

PR #588, opened 2018-01-19
Add a warning emoji ⚠️ to the logger

fathyb:
because we all love emojis, don't we?

### DIFF
--- a/src/Logger.js
+++ b/src/Logger.js
@@ -42,7 +42,7 @@ class Logger {
       return;
     }
 
-    this.write(this.chalk.yellow(message));
+    this.write(this.chalk.yellow(`${emoji.warning} ${message}`));
   }
 
   error(err) {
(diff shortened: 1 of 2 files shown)


src/Logger.js
```diff
@@ -42,7 +42,7 @@ class Logger {
       return;
     }
 
-    this.write(this.chalk.yellow(message));
+    this.status(emoji.warning, message, 'yellow');
```
brandon93s: This introduces a logging defect. If I'm using a `logLevel` of 2, `warn` should write. With this change, it will be silenced by the check in `this.status` which requires a `logLevel` of 3. 

A test for such a regression is included in #562.

src/utils/emoji.js
```diff
@@ -4,3 +4,4 @@ const supportsEmoji = process.platform !== 'win32' || process.env.VSCODE_PID;
 exports.progress = supportsEmoji ? '⏳' : '∞';
 exports.success = supportsEmoji ? '✨' : '√';
 exports.error = supportsEmoji ? '🚨' : '×';
+exports.warning = supportsEmoji ? '⚠️' : '⚠';
```
brandon93s: The provided Windows fallback `'⚠'` doesn't actually work on Windows and will render as `[]`. Please use `‼` (203C).

fathyb: Thanks, I pushed a new commit with both issues fixed.

PR #589, opened 2018-01-19
Use PARCEL_WORKERS environment variable

shawwn:
Per the slack convo, `PARCEL_WORKERS` environment variable controls how many workers to spawn, and `PARCEL_WORKERS=0` makes all code run on the main thread.

`PARCEL_WORKERS=0 yarn test` runs about 33% faster.

### DIFF
--- a/src/WorkerFarm.js
+++ b/src/WorkerFarm.js
@@ -45,7 +45,9 @@ class WorkerFarm extends Farm {
     }
 
     await Promise.all(promises);
-    this.started = true;
+    if (this.options.maxConcurrentWorkers > 0) {
+      this.started = true;
+    }
   }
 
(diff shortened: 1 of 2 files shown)


src/WorkerFarm.js
```diff
@@ -97,6 +99,9 @@ for (let key in EventEmitter.prototype) {
 }
 
 function getNumWorkers() {
+  if (process.env.PARCEL_WORKERS) {
+    return parseInt(process.env.PARCEL_WORKERS);
```
fathyb: I recommend you to specify the `radix` argument, else it'll be infered. Eg. `parseInt('0x0101') = 257` while `parseInt('0x0101', 10) = 0`.

PR #591, opened 2018-01-19
Fix for ipc closed error in tests

DeMoorJasper:
In the tests there sometimes accurs an ipc closed while sending data error, this is due to the main thread being killed before the children died.
This should force-kill the children, preventing this from happening. Another solution is awaiting the super.end using a promise or callback but this caused many tests to fail and in theory this fix would be just as solid and will not wait for unused data to be received.

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
     "tomlify-j0.4": "^3.0.0",
     "uglify-es": "^3.2.1",
     "v8-compile-cache": "^1.1.0",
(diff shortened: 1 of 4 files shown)


fathyb:
Fixes the issue for me, good job!

devongovett:
The error still occurred for me, but in the child worker process not in the master. I added something to worker.js to handle the promise rejections and just ignore that error.

PR #595, opened 2018-01-20
Fix hot reloading of raw assets

shawwn:
After implementing Vue support, I noticed raw assets don't properly hot reload.

(E.g. if I have a Vue component containing ``, and I flip logo.png vertically and save it, I don't see the change until restarting the server.)

This PR fixes that.


src/assets/RawAsset.js
```diff
@@ -18,9 +18,16 @@ class RawAsset extends Asset {
     );
 
     return {
-      js: `module.exports=${JSON.stringify(pathToAsset)};`
+      js: `module.exports=${JSON.stringify(pathToAsset+'?'+this.generateHash(true))};`
     };
   }
+
+  generateHash(regenerate) {
```
DeMoorJasper: Why not just change the filename instead of appending the hash?
And perhaps use mtime or the hash of the file instead of Date.now() as this might invalidate cache too quickly

shawwn: - The filename is generated by `Asset.generateBundleName()`, which is called from several different places in the codebase. It would be risky to try to change it.

- A raw asset doesn't necessarily have an mtime, if it's a virtual sub-asset. (E.g. you can define an asset that simply returns `{raw: /* png data */}` in its `.generate()` method. It doesn't always correspond to a file on disk.)

- It might be a bad idea to use the hash of the file contents. If a project has a lot of raw assets (> 500MB), each rebuild would have to read all of those assets from disk in order to compute the hashes.

The cache concern is valid, though. What if we use file hashes for `parcel build`s, and timestamps for `parcel serve`s? That would give the best of both worlds.

DeMoorJasper: We should have a solid fix that works for both.
The raw data seems like a thing parcel doesn't do atm? I'm still a big supporter of using mtime, u could however combine the best of both worlds, and use mtime if it's a real file and a hash of the content if it's not (as we already have the content in this case, this is very rare and hashing md5's is blazing fast).
Although i'm still not so convinced about the ?<hash> approach, i don't really see how we could go around it. So i guess it's the proper way to go

src/assets/RawAsset.js
```diff
@@ -18,9 +18,16 @@ class RawAsset extends Asset {
...
+    if (regenerate || !this.hash) {
+      this.hash = ('t=' + Date.now());
```
DeMoorJasper: Why don't u just hash Date.now(), this is a bit confusing imo as it is not a hash but a timestamp?

PR #599, opened 2018-01-21
Fixes bug where double slashes isn't parsed as host.

### LINKED ISSUE #577: The js and css cited in the build index.html do not support absolute paths.
I've tried executing the command **parcel index.html - public-url //cdn.test.com/**, but the final result reference is **/cdn.test.com/xxx** and the result I want is **//cdn. test.com/xxx**, would like to ask how to solve this problem?

iifeoluwa:
Fixes #577

### DIFF
--- a/src/utils/urlJoin.js
+++ b/src/utils/urlJoin.js
@@ -6,7 +6,7 @@ const path = require('path');
  * e.g. from \path\to\res.js to /path/to/res.js.
  */
 module.exports = function(publicURL, assetPath) {
-  const url = URL.parse(publicURL);
+  const url = URL.parse(publicURL, false, true);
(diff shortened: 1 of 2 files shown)


src/utils/urlJoin.js
```diff
@@ -6,7 +6,7 @@ const path = require('path');
  * e.g. from \path\to\res.js to /path/to/res.js.
  */
 module.exports = function(publicURL, assetPath) {
-  const url = URL.parse(publicURL);
+  const url = URL.parse(publicURL, false, true);
```
brandon93s: [Documentation](https://nodejs.org/api/url.html#url_url_parse_urlstring_parsequerystring_slashesdenotehost)

- 2nd param is `parseQueryString` which already defaults to false. 
- 3rd param is `slashesDenoteHost` which this changes to `true`

PR #607, opened 2018-01-21
Improve Windows emoji support detection

brandon93s:
I've started using [Hyper](https://hyper.is/) on Windows, which supports emoji. This PR enhances the support check to detect `xterm-256color` to enable emoji's for all platforms using [xterm.js](https://github.com/xtermjs/xterm.js) instead of just VS Code (which uses xterm.js).

Hyper:

image

VS Code:

image


devongovett:
Nice!

PR #612, opened 2018-01-22
minimal svg sprite support

jdanyow:
Hello! I'm looking for feedback on adding some basic support for svg sprites to parcel... Here's an example:

Assuming there's a svg sprite sheet named `icons.svg`, it would be nice if HTMLAsset added url dependencies when it encountered these types of references:

```html
...
<body>
  ...
  <svg>
    <use xlink:href="icons.svg#icon-repo-pull"></use>
  </svg>
  ...
</body>
...
```

For this to work the url's hash needs to be preserved. This got me thinking- should HTMLAsset preserve each url dependency's search and hash components in all cases? For example, bundling html with `<a href="test.html?foo=bar#baz">test</a>` results in `<a href="/dist/eb3da15aa2712be35d36a6920165de75.html">test</a>` which seems like a bug/missing feature.

Thoughts?

### DIFF
--- a/src/assets/HTMLAsset.js
+++ b/src/assets/HTMLAsset.js
@@ -20,7 +20,8 @@ const ATTRS = {
     'embed'
   ],
   href: ['link', 'a'],
-  poster: ['video']
+  poster: ['video'],
+  'xlink:href': ['use']
 };
 
 class HTMLAsset extends Asset {
--- a/test/html.js
+++ b/test/html.js
@@ -11,6 +11,11 @@ describe('html', function() {
       assets: ['index.html'],
       childBundles: [
         {
+          type: 'svg',
+          assets: ['icons.svg'],
+          childBundles: []
+        },
+        {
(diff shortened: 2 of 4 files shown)


DeMoorJasper:
This should indeed be preserved in all cases not just svg's. Same with '?' and '&' if this was also not the case

devongovett:
hmm, I thought it would retain the hash already. There is a test here for that: https://github.com/parcel-bundler/parcel/blob/master/test/url-join.js#L45-L55

jdanyow:
I think urlJoin's hash/search preservation logic might be reversed. It's preserving the publicUrl's hash/search instead of the assetPath's hash/search. #621 switches the urlJoin logic and makes sure addURLDependency preserve's the hash and search, let me know what you think :dancer:

fritx:
How to use svg-sprite within parcel workflow?
I just knew svg-sprite-loader for webpack.

PR #617, opened 2018-01-22
Uglify sourcemaps support

DeMoorJasper:
* Sourcemap support for builds (supporting uglify)
* Added functions to search closest mapping for both original and generated mappings
* Tried to make it as fast as possible (Based on binary search) 🚀


src/transforms/uglify.js
```diff
@@ -14,18 +15,36 @@ module.exports = async function(asset) {
     },
     compress: {
       drop_console: true
-    }
+    },
+    sourceMap: asset.options.sourceMaps
+      ? {
+          filename: asset.relativeName
+        }
```
devongovett: I think uglify has a way to pass an input source map to it so you don't have to manually map it like you did below. See https://github.com/mishoo/UglifyJS2#source-map-options. Looks like the `content` option?

DeMoorJasper: I've seen it but that would add an extra stringify and parse operation to the building of the new sourcemap, which is the largest performance problem with source-map 0.6.1

sant123:
¿Any advance on this? Just asking 😝

DeMoorJasper:
@sant123 as far as i know it's ready to go, just waiting on the approval and merge from Devon

DeMoorJasper:
@devongovett this should be ready to go

src/transforms/uglify.js
```diff
@@ -11,18 +13,39 @@ module.exports = async function(asset) {
...
+      asset.sourceMap = await new SourceMap().extendSourceMap(
+        asset.sourceMap,
+        result.map
+      );
+    } else {
+      asset.sourceMap = result.map;
+    }
+  }
+
+  // Log all warnings
```
devongovett: I'd rather not log all of these. Uglify logs a lot of warnings and they are pretty much useless. I removed this in master already.

src/transforms/uglify.js
```diff
@@ -11,18 +13,39 @@ module.exports = async function(asset) {
...
   }
 
   let result = minify(code, options);
+
   if (result.error) {
     throw result.error;
   }
 
+  if (result.map) {
+    result.map = await new SourceMap().addMap(JSON.parse(result.map));
```
devongovett: I think it's possible to get the map without converting to JSON first. https://github.com/mishoo/UglifyJS2/blob/master/lib/sourcemap.js#L94

Might be faster - maybe you could even access the original mappings from there so you don't need to decode it.

PR #618, opened 2018-01-22
Discourage global parcel install

DeMoorJasper:
As global installing of parcel brings some unwanted issues and generally is a bad practise (in most cases) it should probably not be encouraged, this PR changes the commands into local installs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,13 +28,13 @@
 1. Install with yarn:
 
 ```shell
-yarn global add parcel-bundler
+yarn add parcel-bundler
 ```
 
 or with npm:
 
 ```shell
-npm install -g parcel-bundler
+npm install parcel-bundler --save
 ```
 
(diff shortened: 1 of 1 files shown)


shawwn:
Disagree. The reason this is tricky is because of lockfiles.

If we go this route, users will not get the latest parcel changes by default. Upgrading will become something they have to manually remember. Worse, they commit lock files referencing the *old* parcel, so users will never be using the latest version of parcel by default (after awhile).

This is both a good thing and a bad thing; it's good because theoretically that version of parcel won't break, but it's bad because parcel is already a hefty install.

Now that I write this out, I'm not sure if I agree or disagree. Someone else decide.

devongovett:
Going to close this. Parcel is primarily a CLI tool, and installing globally makes it much easier to get started. Otherwise we'd need to explain that to run it you have to type `./node_modules/.bin/parcel` and that's annoying.

PR #620, opened 2018-01-23
Publish src/builtins/prelude.min.js

shawwn:
We should commit this and publish it to NPM.

The reason for committing this, however, is because when users run `yarn add parcel-bundler/parcel#vue` to use my vue branch,
they don't get any files under `lib/`, because it's a `git clone` of the vue branch, not an npm publish.

In other words, without this, it's very annoying to ship a "beta" version of parcel to users, because
they would have to run `yarn && cd node_modules/parcel-bundler && yarn` instead of simply running `yarn`.

If we just ship the minifed prelude, that allows everyone with node >= 7 to test out the beta version
without having to bother with building the minified file or the lib folder.


shawwn:
That was a bit scatterbrained. Let me try again:

Right now, users can run `yarn add parcel-bundler/parcel#vue` to test out the Vue branch.

However, that is just a `git clone`. It doesn't get any files under `lib/`, nor does it get `prelude.min.js`. Meaning users run into errors related to "missing prelude.min.js".

If we simply commit the minified prelude, they don't have to worry about this, and it's low cost for us.

devongovett:
I don't really like committing build artifacts to git. Maybe instead we could have a `postinstall` script that runs the build on install only if the files aren't already there?

PR #621, opened 2018-01-23
preserve query and hash when adding url dep

jdanyow:
related to #612 

1. updates addURLDependency to preserve asset path's query and hash
2. updates urlJoin to preserve asset's query and hash instead of preserving the public url's query and hash.

### DIFF
--- a/src/Asset.js
+++ b/src/Asset.js
@@ -1,3 +1,4 @@
+const URL = require('url');
 const path = require('path');
 const fs = require('./utils/fs');
 const objectHash = require('./utils/objectHash');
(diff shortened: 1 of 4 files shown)


src/Asset.js
```diff
@@ -78,15 +78,24 @@ class Asset {
       from = this.name;
     }
```
brandon93s: Can be simplified if we go ahead and use `URL.parse()`.

```js
let resolved = path.resolve(path.dirname(from), url).replace(/[?#].*$/, ''); // same as today

const parsed = URL.parse(url);

return (
  this.options.parser
    .getAsset(resolved, this.package, this.options)
    .generateBundleName() + parsed.search + parsed.hash
);
```

jdanyow: sure- was overthinking the perf implications I guess.

jdanyow: before I go any further- what should we do about this:

HTMLAsset calls `addURLDependency`, followed by `urlJoin`:
https://github.com/parcel-bundler/parcel/blob/5c5d5f8af634c0e0aa8e8a3542892febe7c27e85/src/assets/HTMLAsset.js#L50-L55

`urlJoin` preserves the query and hash of the publicUrl, not the assetPath...
https://github.com/parcel-bundler/parcel/blob/5c5d5f8af634c0e0aa8e8a3542892febe7c27e85/src/utils/urlJoin.js#L8-L12

should be the other way around right?

jdanyow: (went ahead and updated the PR with the necessary changes to urlJoin)

devongovett: Yeah this seems kinda brittle. I think using URL.parse + URL.stringify would be better probably.

jdanyow: updated!

devongovett:
Awesome, thanks!

PR #622, opened 2018-01-23
Auto-install peerDependencies 

### LINKED ISSUE #419: 🙋 Install peer dependencies of autoinstalled modules
As a follow up to https://github.com/parcel-bundler/parcel/pull/239#discussion_r158898325, we need to potentially autoinstall peer dependencies of packages that are `localRequire`ed. e.g. `graphql-tag` has a peer dependency on `graphql` that won't be autoinstalled currently, but `graphql-tag` will. Users must manually fix by installing `graphql` in their projects.

brandon93s:
`peerDependencies` are now installed automatically when `installPackage` is run. Only direct-peers are currently installed (e.g. peers of peers are ignored).

Fortunately, the `postcss-cssnext` tests could be tapped into to verify the peer functionality without adding more installation tests... which are slow. 

Independently verified the `graphql-tag` `graphql` use-case as well by running the integration test scenario ([here](https://github.com/parcel-bundler/parcel/tree/master/test/integration/graphql)) with a package.json. Both install as expected! 

---
Closes #419

### DIFF
--- a/src/utils/installPackage.js
+++ b/src/utils/installPackage.js
@@ -1,8 +1,21 @@
 const spawn = require('cross-spawn');
 const config = require('./config');
 const path = require('path');
+const resolve = require('resolve');
 
-module.exports = async function(dir, name) {
+const installPeerDependencies = async (dir, name) => {
+  let basedir = path.dirname(dir);
+
+  const resolved = resolve.sync(name, {basedir});
+  const pkg = await config.load(resolved, ['package.json']);
+  const peers = pkg.peerDependencies || {};
+
+  for (const peer in peers) {
+    await install(dir, `${peer}@${peers[peer]}`, false);
+  }
+};
+
+const install = async function(dir, name, installPeers = true) {
   let location = await config.resolve(dir, ['yarn.lock', 'package.json']);
 
   return new Promise((resolve, reject) => {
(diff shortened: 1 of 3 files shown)


src/utils/installPackage.js
```diff
@@ -1,8 +1,21 @@
...
-module.exports = async function(dir, name) {
+const installPeerDependencies = async (dir, name) => {
+  let basedir = path.dirname(dir);
+
+  const resolved = resolve.sync(name, {basedir});
+  const pkg = await config.load(resolved, ['package.json']);
+  const peers = pkg.peerDependencies || {};
+
+  for (const peer in peers) {
+    await install(dir, `${peer}@${peers[peer]}`, false);
```
devongovett: I think you can pass multiple packages to install at once to npm or yarn rather than doing each one in series. Does that not work?

devongovett: Also, do we also need to install peers of peers etc? Not sure.

brandon93s: > I think you can pass multiple packages to install at once to npm or yarn rather than doing each one in series. Does that not work?

This should work, just didn't think about it at the time. 

> Also, do we also need to install peers of peers etc? Not sure.

I'm explicitly blocking it right now to prevent following a theoretically endless dependency chain. I doubt we'd see this "in the wild" though. Consider the following:

- Package A lists Package B as a peer dep
- Package B lists Package C as a peer dep

If Package B _requires_ Package C to run, then the author of Package A is essentially guaranteeing their module won't work by just installing it's own listed dependencies. This seems unlikely...

brandon93s:
Updated. Now installs all peer dependencies in a single call to npm/yarn. Adds a bit of code complexity (arrays of modules), but should perform better.

PR #639, opened 2018-01-24
Purify sourcemaps (Prevent babel from giving invalid mappings)

### LINKED ISSUE #634: [Source Maps] TypeError: Cannot read property 'line' of undefined
**Choose one:** this a 🐛 bug report

I get an error when I build my project with Parcel 1.5.0 with source maps enabled. With source maps disabled, it works fine (as in 1.4.1).

Using TypeScript, `tsconfig.json`:

```json
{
  "compileOnSave": false,
  "compilerOptions": {
    "noImplicitAny": false,
    "removeComments": false,
    "target": "es2015",
    "sourceMap": true,
    "allowJs": true,
    "listFiles": true,
    "noUnusedParameters": true,
    "noUnusedLocals": true,
    "module": "commonjs"
  }
}
```

Calling Parcel directly (i.e. `import parcel from 'parcel-bundler';`) with a single `index.ts` as the entry point. Entry file imports other TS files and npm modules with the `import` syntax.

### 🤔 Expected Behavior

It should compile and run.

### 😯 Current Behavior

I get this error in the terminal:

```
(node:7050) UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'line' of undefined
    at extension.eachMapping.mapping (/Users/.../node_modules/parcel-bundler/src/SourceMap.js:127:32)
    at SourceMap.eachMapping (/Users/.../node_modules/parcel-bundler/src/SourceMap.js:93:19)
    at SourceMap.extendSourceMap (/Users/.../node_modules/parcel-bundler/src/SourceMap.js:125:15)
    at <anonymous>
(node:7050) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
(node:7050) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

Exception occurs [here](https://github.com/parcel-bundler/parcel/blob/e9596d61f8ec534896359a3820817574c8fb1ea4/src/SourceMap.js#L127) in the Parcel source:

```js
extension.eachMapping(mapping => {
  let originalMapping = original.originalPositionFor({
    line: mapping.original.line, // <- HERE
    column: mapping.original.column
  });
```

This appears multiple times (I'm guessing because I have a lot of files).

### 🌍 Your Environment

| Parcel           | 1.5.0 |
| TypeScript       | 2.5.2 |
| Node             | 9.3.0 |
| npm         | 5.5.1 |
| Operating System | macOS Sierra 10.12.6 |

Thanks!

DeMoorJasper:
Will probably fix #634


robcrocombe:
Tested locally with my project, it seems to have fixed the issue, thanks! I hope we get this live soon :)

test/sourcemaps.js
```diff
@@ -61,6 +61,35 @@ describe('sourcemaps', function() {
     assert.equal(output.env(), process.env.NODE_ENV);
   });
 
+  it('should create a valid sourcemap as a child of a nexted TS bundle', async function() {
```
brandon93s: `nexted` -> `nested`

src/SourceMap.js
```diff
@@ -3,11 +3,28 @@ const lineCounter = require('./utils/lineCounter');
...
 class SourceMap {
   constructor(mappings, sources) {
-    this.mappings = mappings || [];
+    this.mappings = this.purifyMappings(mappings) || [];
     this.sources = sources || {};
     this.lineCount = null;
   }
 
+  purifyMappings(mappings) {
+    if (mappings && mappings.length) {
```
brandon93s: **a)** While we're filtering, it may be worthwhile to check for `mapping` itself before accessing properties on it:

```js
return (
   mapping &&
   mapping.source &&
   ...
)
```

**b)** The above if would be a bit more clear as:
` if (Array.isArray(mappings)) {
...
`

**c)** Assuming column should always be a number, could use `typeof mapping.original.column === 'number'` for these checks.

test/sourcemaps.js
DeMoorJasper: Fixed

src/SourceMap.js
DeMoorJasper: Should be more strict now

src/SourceMap.js
```diff
@@ -3,16 +3,37 @@ const lineCounter = require('./utils/lineCounter');
...
+
+    return [];
+  }
+
   async getConsumer(map) {
     if (map instanceof SourceMapConsumer) {
       return map;
     }
-
+    map = typeof map === 'string' ? JSON.parse(map) : map;
```
devongovett: When does this case occur? Doesn't typescript asset already parse it?

test/sourcemaps.js
```diff
@@ -61,6 +61,35 @@ describe('sourcemaps', function() {
...
+      .toString();
+    let map = fs
+      .readFileSync(path.join(__dirname, '/dist/index.map'))
+      .toString();
+    mapValidator(raw, map);
+
+    let output = run(b);
+    assert.equal(typeof output.env, 'function');
+    assert.equal(output.env(), process.env.NODE_ENV);
+  });
```
devongovett: This test seemed to pass even without the fix for me... How can we reproduce the issue?

DeMoorJasper: I was unfortunately not able to reproduce the issue, I wrote this test in the hopes it would fail. It didn't and i just left it in, because of better code coverage.
Maybe @brandon93s or @robcrocombe can provide us with a better test?

src/SourceMap.js
DeMoorJasper: Uglify outputs a stringified sourcemap, probably also usefull for plugins

PR #641, opened 2018-01-24
Add htm extension

### LINKED ISSUE #636: parcel fails silently without any error reporting
reproduce: create a file and name it index.htm (not html!)

    > parcel index.htm

parcel reports "Server running at http://localhost:1234", but this url is unaccessible.

would be nice to have error reporting.

DeMoorJasper:
closes #636


DeMoorJasper:
Some tests are failing in Travis, but are passing locally

Issue #564, opened 2018-01-16
🙋watch bundle add "poll" option


yalishizhude:
In Docker container on Windows, it cannot rebuild when files changed, so I need poll files to know whether they are changed. It is similar to webpack watch's option "poll".

refrence: https://webpack.js.org/configuration/watch/#watchoptions-poll

### 🤔 Expected Behavior

Just like webpack, I can config a param named "poll",which is a millisecond.
### 😯 Current Behavior

Now there is no option.

### 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Parcel           |  0.0.1
| Node             | 8.9.3
| npm/Yarn         | 5.5.1
| Operating System | windows 10

devongovett:
Can we automatically detect the need to poll somehow? I would prefer not adding a million options if we can choose the right default automatically.

jchv:
Without getting pretty deep into the weeds, it may be hard to detect the conditions where you would need to use polling. The problem is that inotify WILL work, but it will only see local changes. So for example, if you exec into the container and touch a file, it will catch it, but if you change it externally nothing will happen.

The best solution I can imagine would be UNIX specific and would basically consist of the following:
- When watching a file, detect the mount point it is nearest.
- Detect the filesystem of the mountpoint.
- If the filesystem is known to have broken inotify support, punt back to polling for that file. Examples: `cifs`, `nfs`, `vboxsf` will all have trouble with inotify.

This may actually be worth an option though...

marcklei:
It is possible to set polling for chokidar via environment variables. I am in a similiar situation where I need to work on a windows machine with a docker development setup.

An example for setting polling via docker-compose.yml:
```
version: '2'
services:
  web:
    build: .
    command: yarn start
    ports:
     - "1234:1234"
     - "49359:49359"
    volumes:
    - .:/code
    environment:
    - CHOKIDAR_USEPOLLING=1
```
see [chokidar docs](https://github.com/paulmillr/chokidar#performance)

Even if this solves my issues, if a poll option should be added anyway I'd like to start working on this issue.

brandon93s:
@marcklei - that's a great find, thanks for sharing! Parcel also supports `dotenv` if you'd rather define the environment variable alongside you project as opposed to your machine configuration.

jchv:
Wow, I looked at the Chokidar source code and 100% missed this even though I was looking for it. Thanks for the tip, this is hugely useful for me!

[...]

PR #649, opened 2018-01-24
Add --use-polling option.

jchv:
I know there's a good chance this will not get accepted, because it adds a new configuration option, but I figured I'd propose it anyways.

There really doesn't seem to be a good, reasonably portable way to detect that filesystem events will be unreliable. You could do it for a specific platform and mostly be right, but having an override seems like the way to go here.

Related to #564, but without option to set the interval.

### DIFF
--- a/src/Bundler.js
+++ b/src/Bundler.js
@@ -85,7 +85,9 @@ class Bundler extends EventEmitter {
         typeof options.sourceMaps === 'boolean'
           ? options.sourceMaps
           : !isProduction,
-      hmrHostname: options.hmrHostname || ''
+      hmrHostname: options.hmrHostname || '',
+      usePolling:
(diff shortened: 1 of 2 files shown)


brandon93s:
Could we detect the process running inside of a container and auto-enable polling? There's some relatively reliable ways to detect if a process is running inside docker. There is also a `containerized` module. 

This won't cover every scenario that needs polling, but would be a start and avoids the need for a CLI flag.

marcklei:
As i mentioned in #564 chokidar polling can also be enabled through environment vars. eg. `CHOKIDAR_USEPOLLING=1`

brandon93s:
Closing in favor of using `CHOKIDAR_USEPOLLING=1` which seems favorable to the flag anyways since it can be set in a config file. Feel free to open a separate PR documenting the availability of this behavior.

Issue #648, opened 2018-01-24
UnhandledPromiseRejectionWarning


gweinfurther:
I just installed parcel. When running parcel in live or build mode, I get this message:

UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 1): SyntaxError: 'super' keyword unexpected here

Running on Windows 10 in Chrome 63.
Parcel 1.5

ZWkang:
i got it too~

akosasante:
Same here
Running on Ubuntu 16.04
Parcel 1.50

sinout:
Same here. Encountered after update from v1.4.1

DeMoorJasper:
@sinout node.js 6 is currently not supported we are working to support it in the future

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

srph:
Having it as well. Just installed `v1.5`. Using node `v7.5.0`

DeMoorJasper:
@srph node 7 isnt supported yet either only node 8+ is currently supported

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

mdchristians:
Having the same issue.
Fresh install of Parcel on a fresh project.  Node v8.9.3.

katsos:
For anyone who has the same problem, you can use nvm for your project with "v8" wrote in its .nvmrc file until it is fixed.

mdchristians:
@katsos My project has an .nvmrc implemented already using 8.9.3...

katsos:
@mdchristians don't forget to execute `nvm use` inside the folder from the same terminal that you are going to execute anything related to Node.

mdchristians:
@katsos My .nvmrc file is being used.  Running` node -v` outputs 8.9.3 within the project directory.

mdchristians:
It's working with Node v8.5.0 so now I'm pretty sure it's an issue with something else...

DepickereSven:
I have this problem to. 

I have:
Node v8.9.4
Fresh install of Parcel

Every time I run it I get an other number in front of the error some examples:
`(node:16656) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 1): SyntaxError: 'super' keyword unexpected here`
`(node:4332) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 1): SyntaxError: 'super' keyword unexpected here`
[...]

PR #650, opened 2018-01-24
fix `super` error when node version <8 #648

ZWkang:
when  the build ,we will get two source folder.lib and src.
index.js will help us to choose the work folder.
the babel setting.target with 6.
then,will get some problem in the CoffeeScriptAsset/ReasonAsset/TypeScriptAsset.
 I found the problem the key in the babel converted code, when async/await transform, after await such as
```
await super.end ()
Will become
return _asyncToGenerator (function * () {
     yield super.end ()
})
```
This is not legal
To my surprise, the typescriptAsset coffescriptAsset in assets is always the same.
Probably a minority is using a node version less than eight

I think this is why parcel does not fully support the low version.
In the past, if you do not apply the corresponding asset extension error will not exist, but at this time if you do not make this change, then the lower version will no longer be used, then babel transform will be no meaning in parcel.

Either support better or not.
I think so

this edit is a demo show, we have many ways to make that problem resolved.~~

sorry my bad english.

### DIFF
--- a/src/packagers/SourceMapPackager.js
+++ b/src/packagers/SourceMapPackager.js
@@ -17,7 +17,10 @@ class SourceMapPackager extends Packager {
   async end() {
     let file = path.basename(this.bundle.name);
     await this.dest.write(this.sourceMap.stringify(file));
-    await super.end();
+    this.stopend();
+  }
+  stopend(){
+    return super.end();
   }
 }
 


src/packagers/SourceMapPackager.js
```diff
@@ -17,7 +17,10 @@ class SourceMapPackager extends Packager {
   async end() {
     let file = path.basename(this.bundle.name);
     await this.dest.write(this.sourceMap.stringify(file));
-    await super.end();
+    this.stopend();
```
DeMoorJasper: what about just doing `return super.end();` here?

ZWkang: when the babel transform .the await super change to function *(){ super.....}
so make a function .do super job. and then use this to called that function.
just a one of the way to edit them.have many ways to do that....
i think if we want to make it support node version <8.
we should do better and think more things~~~
Otherwise it does not support version <8 just fine

devongovett:
This seems like a bug we should report to Babel. `super` within transpiled async functions should work.

DeMoorJasper:
Apparently this was an issue with babel and has been fixed in Babel 7 https://github.com/babel/babel/issues/3930

PR #652, opened 2018-01-25
Add "--target=node" option to produce a NodeJS-friendly bundle, and also "--target=electron"

### LINKED ISSUE #192: node bundling
🙋 feature request

Other bundlers (like webpack) can target node environment instead of browser. That largely involves leaving things like `require('fs')` and `process` intact in the code.

Is there a way to do the same in parcel?
If no, is it on the roadmap?

Ideally I'd want to test it by parcelling the parcel itself, and getting a bundled single JS file.

### LINKED ISSUE #474: Ability to ignore modules at least in well-known environments

lbguilherme:
This adds `--target=node` and `--target=electron` options to the `build` command. Four things changes:

1. Dependencies that are not relative (such as `import express from 'express';`) are assumed to be from `node_modules` and are left as is. Relative dependencies, such as `import car from './car'` are still bundled.

2. `fs.readFileSync` is left as is.

3. Globals such as `Buffer` or `__dirname` are left as is.

4. Hot Module Loading runtime is not added (for `--target=node`)

The default is `--target=browser`.

Fix #192.
Fix #474.


devongovett:
This is reasonable. Perhaps we should be a little more generic in how we specify this option though. Maybe in the future we'll want to support other targets, like electron or something. So maybe make the option `--target=node`?

lbguilherme:
@devongovett I changed from `--node` to `--target=node`, where the default is `--target=browser`. This is more generic.

lbguilherme:

davidnagli:
@lbguilherme Is this ready to merge?

lbguilherme:
As far as I can tell, it is. 


On 27 Jan 2018 5:50 p.m., David Nagli <notifications@github.com> wrote:

@lbguilherme Is this ready to merge?

devongovett:
This looks great. Would be awesome if you could add a few tests for this behavior as well. Are you up for it?

src/visitors/dependencies.js
```diff
@@ -86,6 +86,15 @@ module.exports = {
 };
 
 function addDependency(asset, node, opts = {}) {
+  if (asset.options.target !== 'browser') {
+    const isRelativeImport =
+      node.value.startsWith('/') ||
+      node.value.startsWith('./') ||
+      node.value.startsWith('../');
```
devongovett: You could do this with a regex maybe? `/^[./]/.test(node.value)`.

Also, I wonder if this is always the desired behavior. Are there cases where you'd want to bundle node_modules into a single file when targeting node?

lbguilherme: I believe using startsWith will be faster than using a regex, and it also cleaner to read. If you disagree I can change it still.

The logic is taken from typescript documentation: https://www.typescriptlang.org/docs/handbook/module-resolution.html, see the section about "Relative vs. Non-relative module imports".

I think there are cases where you would want to bundle everything, but it would break a lot of packages that expect to dynamically read a file from itself, or to list other packages like parcel itself does for plugins. Too many will break. Perhaps it can be solved in the future, wrapping some fs calls. But it's much more complicated than this solution. And for who is using node, having the modules there is very ok.

gimenete:
I think it should be easy to add support for electron here by using `--target=electron`.

- In electron HMR should be enabled
- In electron non-relative packages should be ignored

I made a repo to test https://github.com/gimenete/electron-parcel-bug-report

And that will close https://github.com/parcel-bundler/parcel/issues/474

src/visitors/dependencies.js
devongovett: OK, this is a good start. Let's see if bundling module sources for node comes up later.

lbguilherme:
@gimenete Just added a commit with `--target=electron`. I tested it in your repo, it works fine!

@devongovett I'm not sure how to add tests for this.

gimenete:
@lbguilherme does the HMR work for you on electron? If I change `renderer.js` it doesn't refresh.

lbguilherme:
@gimenete I just noticed it doesn't. Trying to figure out why

gimenete:
@lbguilherme I've also tried to require a file from `renderer.js` and see if changing that imported file refreshes. But it doesn't. I wanted to be sure that it was not an issue with the root file being processed and whether it worked with dependencies of that file or not.

lbguilherme:
@gimenete `rm -rf .cache` first.

lbguilherme:
@gimenete This is fixed now. :tada:

gimenete:
@lbguilherme I've just tested it and it works! 💯

lbguilherme:
Is there anything else I should do to get this approved?

devongovett:
Could you add some tests please? You can copy some existing ones for e.g. fs, importing node modules, and globals, set the target to node, and assert that the opposite of the current tests.

lbguilherme:
@devongovett I have added some tests.

src/visitors/dependencies.js
Toxicable: >And for who is using node, having the modules there is very ok.

I disagree. Currently I bundle all my Nodejs apps' node_modules so I have a single deployable which makes a redeployment easy and avoid any possibility of transitive module breakages.

However I also understand that this isn't an easy problem for a bundler to solve since there are a lot of edge cases working with modules that interact of with binary files and filesystems.
But it would be great to see some work after this lands to help solve some of these issues.

lbguilherme:
I don't need, but I definitely prefer to bundle node_modules for my projects as well. I plan to work on that after this pull request is merged. 


On 4 Feb 2018 3:47 p.m., Fabian Wiles <notifications@github.com> wrote:

@Toxicable commented on this pull request.

_____________________________________________

In src/visitors/dependencies.js:

> @@ -86,6 +86,15 @@ module.exports = { }; function addDependency(asset, node, opts = {}) { + if (asset.options.target !== 'browser') { + const isRelativeImport = + node.value.startsWith('/') || + node.value.startsWith('./') || + node.value.startsWith('../'); 

And for who is using node, having the modules there is very ok.

I disagree. Currently I bundle all my Nodejs apps' node_modules so I have a single deployable which makes a redeployment easy and avoid any possibility of transitive module breakages.

However I also understand that this isn't an easy problem for a bundler to solve since there are a lot of edge cases working with modules that interact of with binary files and filesystems.
But it would be great to see some work after this lands to help solve some of these issues.

src/builtins/hmr-runtime.js
```diff
@@ -14,7 +14,8 @@ function Module(moduleName) {
 
 module.bundle.Module = Module;
 
-if (!module.bundle.parent && typeof WebSocket !== 'undefined') {
+if (!global.hmrReady && typeof WebSocket !== 'undefined') {
+  global.hmrReady = true;
```
devongovett: Why is this global variable necessary?

lbguilherme: Because on electron there is a native `require`, so `module.bundle.parent` will never be `undefined` as happens on the browser. I understand that this conditional is here to make sure the HMR is only connected once, so I changed to a global flag, that works on both browser and electron.

test/hmr.js
```diff
@@ -61,6 +61,47 @@ describe('hmr', function() {
...
+      watch: true,
+      hmr: true,
+      target: 'node'
+    });
+    await b.bundle();
+
+    ws = new WebSocket('ws://localhost:' + b.options.hmrPort);
+
+    let err = await nextEvent(ws, 'error');
+    assert.equal(err.code, 'EADDRNOTAVAIL');
```
lbguilherme: I get `ECONNREFUSED` on my machine. Perhaps it should just assert that *some* error happen, without checking which?

lbguilherme: @devongovett I get `ECONNREFUSED` on my machine. Perhaps it should just assert that *some* error happen, without checking which?

devongovett: hmm ok. maybe it's node version specific

src/builtins/hmr-runtime.js
devongovett: ah i see. maybe we can check if there is no parent bundle, or the parent bundle isn't a parcel bundle. don't really want to add more global variables.

lbguilherme: How to check if the parent bundle is not a parcel bundle?

devongovett: I just added an `isParcelRequire` flag to the parcel require function

test/hmr.js
devongovett: done

src/builtins/hmr-runtime.js
lbguilherme: :+1:

JayStackKrisz:
Hi guys,
So I don't get it yet. Is it possible to define the target at this moment with ParcelJS?

lbguilherme:
This is just not released yet. It will be possible on the next release.

danr:
@JayStackKrisz you could try referring to the git repo in your package.json: https://stackoverflow.com/questions/43411864/how-to-install-package-from-github-repo-in-yarn

briandk:
Is there any documentation on how `target=electron` works?

There’s nothing for it under the “recipes” section of the docs, and I’m unclear what kind of project structure it expects. For example, where should my main.js, renderer.js, and index.html files be with respect to one another?

I would be happy to issue a PR to add this documentation, but I need some help understanding how the electron target is supposed to be used for development and for building to production 😇

src/visitors/dependencies.js
pravdomil: @Toxicable https://github.com/parcel-bundler/parcel/pull/1196

mubaidr:
This pull request suggests that correct usage is: 
`--target=electron`
while documentation states that it should be used as: 
`--target electron`

Also, is this only available in `build` script or development too?
BTW I have tried both options.

I am getting this error:
```
Uncaught TypeError: fs.existsSync is not a function
    at Object.parcelRequire.46.fs (..\..\..\node_modules\electron\index.js:6)
    at newRequire (main.4f9b08dc.js:48)
    at localRequire (main.4f9b08dc.js:54)
    at Object.parcelRequire.54.electron (components\Templates\NavBar.vue:31)
    at newRequire (main.4f9b08dc.js:48)
    at localRequire (main.4f9b08dc.js:54)
    at Object.parcelRequire.15../components/Templates/NavBar (App.vue:18)
    at newRequire (main.4f9b08dc.js:48)
    at localRequire (main.4f9b08dc.js:54)
    at Object.parcelRequire.2.vue (main.js:12)
```

Hammster:
@briandk `--target=electron` does not change on how to set the start file, there also also no set way on how to structure your project.

You can do multiple things like:

- Use parcel for the renderer only, any you need the `--target=electron` for fs calls etc. Which is the most sane way for electron IMHO.

- Use for main and renderer by importing the renderer html in the window creation script.

- Use the API and split the 2 processes whith individual settings eg `target: "electron"` for the main process and `target: "browser"` for the renderer etc.

For the documentation for what `--target=electron` or `node` exactly does, it parses it to babel to generate electron friendly enviroments, and also does some adjustments on how the HMR works.  

HMR change:
https://github.com/parcel-bundler/parcel/blob/28b87cf03425ca4214639bde9fbc334ed04d25b8/src/packagers/JSPackager.js#L25-L31

Babel target docs:
https://babeljs.io/docs/plugins/preset-env/

src/visitors/dependencies.js
texastoland: Hey I'm looking for a few key features relative to Microbundle. It looks like tilde imports don't work?

It may have changed since this PR but the logic should be checking that it starts with `/`, `~`, `.` which should _not_ be measurably slower with a regex `/^[/~.]/`.

https://github.com/parcel-bundler/parcel/blob/f431b78047bc4fde720be1d7f1411cc19710eda6/src/Resolver.js#L121-L147

That function makes me wonder if the logic should be duplicated though.

PR #654, opened 2018-01-25
Inject HTML tag if missing

### LINKED ISSUE #633: 🐛 issue with css outside of head tag

This is a bug.

If I try to reference to my css outside of the `head` tag, it throws an error:
`Cannot read property 'content' of undefined`

My babel configuration is:
```js
{
  "presets": ["env", "stage-2"]
}
```

### 🤔 Expected Behavior

This should parse and add css no matter where it is referenced.

### 😯 Current Behavior

throws `Cannot read property 'content' of undefined`.

### Full error in console:

```
Cannot read property 'content' of undefined
    at HTMLPackager.getHeadContent (/usr/local/lib/node_modules/parcel-bundler/src/packagers/HTMLPackager.js:30:12)
    at HTMLPackager.insertSiblingBundles (/usr/local/lib/node_modules/parcel-bundler/src/packagers/HTMLPackager
.js:41:21)
    at /usr/local/lib/node_modules/parcel-bundler/node_modules/posthtml/lib/index.js:109:53
    at PostHTML.process (/usr/local/lib/node_modules/parcel-bundler/node_modules/posthtml/lib/index.js:104:18)
    at HTMLPackager.addAsset (/usr/local/lib/node_modules/parcel-bundler/src/packagers/HTMLPackager.js:19:9)
    at Bundle._addDeps (/usr/local/lib/node_modules/parcel-bundler/src/Bundle.js:146:20)
    at <anonymous>
    at process._tickCallback (internal/process/next_tick.js:160:7)
```

### 🌍 Your Environment

| Parcel           | 1.5.0
| Node             | 9.4.0
| npm/Yarn         | yarn v1.3.2
| Operating System | Mac OS Sierra 10.12.6

brandon93s:
Fixes #633 if we want to support this case.

### DIFF
--- a/src/packagers/HTMLPackager.js
+++ b/src/packagers/HTMLPackager.js
@@ -3,6 +3,18 @@ const posthtml = require('posthtml');
 const path = require('path');
 const urlJoin = require('../utils/urlJoin');
 
+// https://www.w3.org/TR/html5/dom.html#metadata-content-2
+const metadataContent = new Set([
+  'base',
+  'link',
+  'meta',
+  'noscript',
+  'script',
+  'style',
+  'template',
+  'title'
+]);
+
 class HTMLPackager extends Packager {
   async addAsset(asset) {
     let html = asset.generated.html || '';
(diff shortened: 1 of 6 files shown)


farskid:
So basically this would inject the `html` tag if not found? is so, is it possible to check for `body` and `head` too?

brandon93s:
@farskid - That's correct. 

Parcel already handles injecting the head for you if it isn't found. The body would also be possible. The body isn't required by the html5 spec though; it's left as an implementation detail for the browser to handle it. 

Up for debate if we should handle this or just encourage spec compliant HTML.

devongovett:
We should probably handle it. Browsers do, so HTML like this exists in the wild.

Question: in [the example](https://github.com/farskid/guessing-game/blob/develop/index.html) from the issue, there is no `<html>`, `<head>`, or `<body>`. In this PR, Parcel will add the `<html>`, and a `<head>` tag inside it. But content that's already in there like the `<link>` tag won't get moved into the `<head>` and the other stuff won't get moved into a `<body>`. It's unclear what actually happens to that stuff.

Perhaps we should actually fix it as browsers do? There is probably a spec for this to determine how to insert the elements. Otherwise, since the tags are optional, we could just insert the element into the root content and not even bother adding an `<html>` or `<head>` tag.

devongovett:
For reference: https://www.w3.org/TR/2011/WD-html5-20110525/syntax.html#optional-tags

farskid:
@brandon93s  @devongovett Good point. in addition to what you said, in the example on issue, if I added the `head` tag even without `html` tag, parcel would bundle successfully. isn't it against the spec too?

brandon93s:
After some research and digging into [WebKit](https://github.com/WebKit/webkit/blob/master/Source/WebCore/html/parser/HTMLTreeBuilder.cpp#L563-L576), it looks like the `html`, `head` and `body` are _all_ optional in an HTML5 document. I'm of the opinion that parcel shouldn't unnecessarily alter the user's html but instead insert the bundle tags in a location that produces explainable and consistent output. 

Browsers will automatically break up your html into `body` and `head` tags using some relatively straightforward logic. The gist of it is: if no head is found, it scans nodes until it finds the first node that is not _allowed_ to be in the head. Once it encounters that node, everything above it is placed into a header and the rest is candidate for the body. We can use this information to place the bundles in the tree where we know they'll end up in the head, without having to manually add the `head`, `html` or `body` to the users document. 

Adding `html`, `body` and `head` could also have unintended side-effects down the line if an html-partial were to pass through here. 

With the above in mind, I propose the following strategy:

diagram 1

This would remove the current code that adds the `head` node, and implements additional logic to find the correct position to insert the bundle tag if no head is found. 

Thoughts? Will take a shot at implementation pending feedback. @devongovett

devongovett:
Yeah that sounds like the right strategy.

brandon93s:
The above discussed strategy has been implemented.

PR #663, opened 2018-01-25
Fix rust projects with hyphenated names

derekdreery:
Currently, parcel will not successfully bundle rust code if the cargo project name has a hyphen in it. This PR fixes this bug.

In Rust, hyphens in project names are converted to underscores both in code and in output file names.

### DIFF
--- a/src/assets/RustAsset.js
+++ b/src/assets/RustAsset.js
@@ -153,7 +153,8 @@ class RustAsset extends Asset {
 
     // Get output file paths
     let outDir = path.join(cargoDir, 'target', RUST_TARGET, 'release');
-    let rustName = cargoConfig.package.name;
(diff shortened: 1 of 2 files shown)


src/assets/RustAsset.js
```diff
@@ -153,7 +153,8 @@ class RustAsset extends Asset {
 
     // Get output file paths
     let outDir = path.join(cargoDir, 'target', RUST_TARGET, 'release');
-    let rustName = cargoConfig.package.name;
+    // Rust converts '-' to '_' when outputting files.
+    let rustName = cargoConfig.package.name.replace('-', '_');
```
devongovett: This will only replace the first dash. We should make sure to replace every instance.

PR #664, opened 2018-01-25
Do not throw when loadedAssets=null

### LINKED ISSUE #344: When I change my code, the server adds on content to existing content on the page (repeats content instead of replacing old content)
This is a potential bug report.

When I start up a parcel server, everything works fine, but when I update the code (make any changes like changing the text on my header), the window shows the old content as well as the new content. [screenshot here](https://cl.ly/3H0H1n0P3w3D)

.babelrc
```js
{
    "presets": [
        "env"
    ]
}
```

package.json
```js
{
  "name": "sketchvids-new",
  "version": "1.0.0",
  "description": "",
  "main": "index.js",
  "scripts": {
    "test": "echo \"Error: no test specified\" && exit 1",
    "start": "parcel index.html"
  },
  "keywords": [],
  "author": "Adam Rasheed",
  "license": "ISC",
  "devDependencies": {
    "parcel-bundler": "^1.2.0"
  },
  "dependencies": {
    "babel-preset-env": "^1.6.1",
    "node-sass": "^4.7.2"
  }
}
```

### 🤔 Expected Behavior

When I run the parcel server, and make changes, it should update the existing content, not actually add content.

### 😯 Current Behavior

When I run the parcel server, it actually adds the changed content onto the dom.

### 🌍 Your Environment

| Parcel           | 1.2
| Node             | 9.2.1
| npm/Yarn         | 5.5.1
| Operating System | MacOS

fritx:
related to comment https://github.com/parcel-bundler/parcel/issues/344#issuecomment-360464971

### DIFF
--- a/src/Bundler.js
+++ b/src/Bundler.js
@@ -176,7 +176,9 @@ class Bundler extends EventEmitter {
       // Emit an HMR update for any new assets (that don't have a parent bundle yet)
       // plus the asset that actually changed.
       if (this.hmr && !isInitialBundle) {
-        this.hmr.emitUpdate([...this.findOrphanAssets(), ...loadedAssets]);
+        if (loadedAssets) {
+          this.hmr.emitUpdate([...this.findOrphanAssets(), ...loadedAssets]);
+        }
       }
 
       // Invalidate bundles


devongovett:
Could you provide a small example that reproduces the issue? I'm not sure how it could ever be null. The promise should always resolve with a Set: https://github.com/parcel-bundler/parcel/blob/master/src/utils/PromiseQueue.js#L63 `this.processed` is never set to null in that file.

MarcusAhlfors:
I can confirm that I see this exact problem with 1.5.1 and that the patch fixes the problem. Unfortunately the project is quite large and I don't have the time/possibility right now to minimize it down to a reproducable public example.

alesdrobysh:
I can observe this issue when
1) one of my npm dependencies is linked to a folder
2) parcel watcher  is started
3) the linked npm package is updated

then parcel rebuilds the app and fails with

```
🚨  loadedAssets is not iterable
    at Bundler.bundle (/home/**/ui/node_modules/parcel-bundler/src/Bundler.js:179:61)
    at <anonymous>
    at process._tickCallback (internal/process/next_tick.js:188:7)
```

hope that info is helpful

pselden:
I've got that same issue and I did a little digging. I haven't narrowed down the exact scenario needed to reproduce this but I have determined that the error occurs here https://github.com/parcel-bundler/parcel/blob/master/src/Bundler.js#L176 when the promise queue has no jobs left to process and run is called.

The following set of events happens:

0) this.runPromise is created and NOT null - https://github.com/parcel-bundler/parcel/blob/master/src/utils/PromiseQueue.js#L31
1) _next() is called https://github.com/parcel-bundler/parcel/blob/master/src/utils/PromiseQueue.js#L62
2) this.processing.size === 0 - https://github.com/parcel-bundler/parcel/blob/master/src/utils/PromiseQueue.js#L62
3) promise is resolved - https://github.com/parcel-bundler/parcel/blob/master/src/utils/PromiseQueue.js#L63
4) reset is called, which nulls out the promise before it can be returned to the caller! - https://github.com/parcel-bundler/parcel/blob/master/src/utils/PromiseQueue.js#L64
5) this.runPromise is returned, but it's null -- await null just returns null, so loadedAssets is null
6) doing a ...null blows everything up - https://github.com/parcel-bundler/parcel/blob/master/src/Bundler.js#L181

Whether the root fix is checking for null here, or somehow fixing it so runPromise cannot be returned null.. I do not know.

PR #681, opened 2018-01-27
Implement support for <img srcset="...">

lbguilherme:
It will collect dependencies from ``, just like already happens with ``.


lbguilherme:

src/assets/HTMLAsset.js
```diff
@@ -41,6 +41,28 @@ class HTMLAsset extends Asset {
...
+              }
+              pair[0] = assetPath;
+
+              newSources.push(pair.join(' '));
+            }
+
+            node.attrs[attr] = newSources.join(',');
+            this.isAstDirty = true;
+
+            continue;
```
devongovett: Could we move this logic for parsing srcset to a separate function? Maybe we could have a map somewhere of custom handlers for certain attrs?

src/assets/HTMLAsset.js
```diff
@@ -41,6 +41,28 @@ class HTMLAsset extends Asset {
...
+              const pair = source.trim().split(' ');
+              if (pair.length === 0) continue;
+
+              let assetPath = this.addURLDependency(
+                decodeURIComponent(pair[0])
+              );
+              if (!isURL(assetPath)) {
+                assetPath = urlJoin(this.options.publicURL, assetPath);
+              }
+              pair[0] = assetPath;
```
devongovett: The logic of the above few lines could be shared with the copy below which was already there. Maybe make a function for it?

lbguilherme: Done.

src/assets/HTMLAsset.js
lbguilherme: I moved this logic to its own function. About custom handlers, perhaps its better to wait until there is at least another one?

PR #682, opened 2018-01-27
Enable minification of yaml

lbguilherme:


lbguilherme:

lbguilherme:
Actually, YAML support is simply broken. YAML has types that can't be represented as JSON, such as binary data or timestamps. It will probably need to be rewritten without `JSON.stringify`.

src/assets/YAMLAsset.js
```diff
@@ -12,8 +13,19 @@ class YAMLAsset extends Asset {
...
+    let code = `module.exports = ${JSON.stringify(this.ast, false, 2)};`;
+
+    if (this.options.minify) {
+      let minified = minify(code);
+      if (minified.error) {
+        throw minified.error;
+      }
+
+      code = minified.code;
+    }
```
devongovett: This looks the same as the code in JSONAsset. Maybe we could make a util function to JSON stringify and optionally minify and then use it in both places?

lbguilherme: Actually, I'll rewrite this without JSON.stringify, because YAML can contain dates.

lbguilherme: Please check the new commit

brandon93s:
👍 on extracting `serializeObject`. Can drop it into the `TOMLAsset` with ease!

PR #683, opened 2018-01-27
Minify PNG images with pngquant

lbguilherme:


DeMoorJasper:
See https://github.com/DeMoorJasper/parcel-plugin-imagemin

DeMoorJasper:
I'll close this down as the plugin already supports this and also svg, jpg and gif images

devongovett:
Do you guys think this should be included in core by default?

lbguilherme:
First time I used Parcel I expected it to be core, because pretty much every site will include images of some kind. At the time (a week ago) there wasn't a plugin to this _yet_. I still think it would be better in the core, but having a plugin for this is already very good. For my usecase, it's solved.

Perhaps would be interesting to have a directory of plugins on the site. Maybe generated from npm and ranked by github stars or npm downloads.

PR #684, opened 2018-01-27
[bug fix] hmr doesn't crash app if it fails

niicojs:
If hmr fails (in my case for example I was testing through ngrok), the whole site doesn't break.
I've just catch error and print it to console so the app can keep loading


DeMoorJasper:
@davidnagli the hmr-runtime is a built-in (gets packaged in the bundle) and has no access to the Logger class?

niicojs:
also, this is client side, `Logger` is while generating right?

davidnagli:
Oh ya sorry didn’t realize this was the HMR runtime 😂

src/builtins/hmr-runtime.js
```diff
@@ -15,38 +15,43 @@ function Module() {
...
 
-    if (data.type === 'error') {
-      console.error('[parcel] 🚨  ' + data.error.message + '\n' + 'data.error.stack');
-    }
-  };
+      if (data.type === 'error') {
+        console.error('[parcel] 🚨  ' + data.error.message + '\n' + data.error.stack);
+      }
+    };
+  } catch (e) {
```
devongovett: Do we need the try/catch around the entirety of this code or is only a certain line failing? Also what is the issue you had? Connection failure or something? What about a `ws.onerror` handler?

niicojs: I was using ngrok through https, so Chrome did not want to allow connection to non-secure websocket used by hmr. and it crashed my whole app.
for me, hmr can't crash the underlying app, that's why I made a (too?) broad try / catch.

DeMoorJasper: Isn't this fixed by the recent https enabled websocket?

devongovett:

PR #687, opened 2018-01-28
Adds linking to local parcel guideline to contributing file

farskid:
Adds multiple options for linking to local copy of parcel during development, to the contributing guidelines.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -67,11 +67,11 @@ While contributing to parcel, you may need to run a local version of parcel and 
 
 To do this, you have multiple options:
 
-* if you had **parcel** installed globally, you can run `yarn global remove parcel` and then `cd` to your own copy of parcel and run `yarn link`. This way, the global parcel in your PATH would reference to the local copy. To make sure you have it installed globally, just run `parcel --version`.
+* If you had **parcel** installed globally, you can run `yarn global remove parcel` and then `cd` to your own copy of parcel and run `yarn link`. This way, the global parcel in your PATH would reference to the local copy. To make sure you have it installed globally, just run `parcel --version`.
 
 * Another option would be to run `yarn link` in your own copy of parcel directory and then go to another directory and run `yarn link parcel-bundler`, so that locally your copy of parcel is the one used in its npm scripts.
 
-* Another way which is probably the best way is to **alias** your local parcel. This is also the recommended way of developing tools suggested by [Yarn](https://yarnpkg.com/en/org/contributing/). To achieve this, simply run `alias parcel="PATH_TO_PARCEL_BIN_CLI_FILE"`.
+* Another way, which we highly recommend if you plan to switch between different parcel sources often, is to create some terminal aliases. This way you can have different aliases for each of the versions of parcel you wish to maintain on your machine. For example, you can make an alias called `parcelfork` pointing to your fork of parcel for development, and then `parceldev` pointing to a clone of the parcel repo for experimenting with the latest changes, and then then still keep the original `parcel` command pointing to the npm installed stable version. To do this, add some aliases to your terminal’s startup file (usually .bashrc) in the format `alias [name]=“node [path_to_new_parcel_bin_file]”` for each of the sources you want to use. For example: `alias parcelfork=“node ~/my/path/to/fork/bin/cli.js”`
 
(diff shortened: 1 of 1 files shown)


farskid:
@davidnagli CI Fails on a test case for rust packager. According to the test description, `rust should use wasm-gc to minify output`. is there anything I should do here to pass the test?

davidnagli:
No dw about the CI fail, that’s from something else. We’re currently working on getting that fixed.

CONTRIBUTING.md
```diff
@@ -61,6 +61,20 @@ You can set `PARCEL_WORKERS` to the number of worker processes to spawn.
...
+
+While contributing to parcel, you may need to run a local version of parcel and set it as the global cli option in your command line, so that when you run `parcel ...`, it will find your new local variant of parcel instead of the global package installed before (if you had it installed globally before).
+
+To do this, you have multiple options:
+
+* if you had **parcel** installed globally, you can run `yarn global remove parcel` and then `cd` to your own copy of parcel and run `yarn link`. This way, the global parcel in your PATH would reference to the local copy. To make sure you have it installed globally, just run `parcel --version`.
+
+* Another option would be to run `yarn link` in your own copy of parcel directory and then go to another directory and run `yarn link parcel-bundler`, so that locally your copy of parcel is the one used in its npm scripts.
+
+* Another way which is probably the best way is to **alias** your local parcel. This is also the recommended way of developing tools suggested by [Yarn](https://yarnpkg.com/en/org/contributing/). To achieve this, simply run `alias parcel="PATH_TO_PARCEL_BIN_CLI_FILE"`.
```
davidnagli: I think we should specify that they should save that alias definition to their .bashrc or something like that

CONTRIBUTING.md
```diff
@@ -61,6 +61,20 @@ You can set `PARCEL_WORKERS` to the number of worker processes to spawn.
...
 
+## Link to local parcel build
+
+While contributing to parcel, you may need to run a local version of parcel and set it as the global cli option in your command line, so that when you run `parcel ...`, it will find your new local variant of parcel instead of the global package installed before (if you had it installed globally before).
+
+To do this, you have multiple options:
+
+* if you had **parcel** installed globally, you can run `yarn global remove parcel` and then `cd` to your own copy of parcel and run `yarn link`. This way, the global parcel in your PATH would reference to the local copy. To make sure you have it installed globally, just run `parcel --version`.
+
+* Another option would be to run `yarn link` in your own copy of parcel directory and then go to another directory and run `yarn link parcel-bundler`, so that locally your copy of parcel is the one used in its npm scripts.
```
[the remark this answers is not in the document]
farskid: I agree.Maybe I should put it into more clear words. __another directory__ should mean the project that you wanna test your copy of parcel in.

CONTRIBUTING.md
farskid: Agree. is it also necessary to point to files such as `.zshrc` or `.bashrc` is sufficient?

davidnagli: Good question… not sure 😬

farskid: I'm not sure about it either. Variants of profile managements such as `zsh` are used by many developers. but I assume `zsh` is most used alternative. Also, if you're using any alternative such as `zsh`, you can still use `bashrc` and `bash_profile` files.

davidnagli: Here’s how I would put it:
* Another way, which we highly recommend if you plan to switch between different parcel sources often, is to create some terminal aliases. This way you can have different aliases for each of the versions of parcel you wish to maintain on your machine. For example, you can make an alias called `parcelfork` pointing to your fork of parcel for development, and then `parceldev` pointing to a clone of the parcel repo for experimenting with the latest changes, and then then still keep the original `parcel` command pointing to the npm installed stable version. To do this, add some aliases to your terminal’s startup file (usually .bashrc) in the format `alias [name]=“node [path_to_new_parcel_bin_file]”` for each of the sources you want to use. For example: `alias parcelfork=“node ~/my/path/to/fork/bin/cli.js”`

farskid: I love it! it's super clear how to use alias and the trick to have multiple instances of it on a single machine. Should I replace it right now?

CONTRIBUTING.md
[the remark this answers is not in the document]
farskid: @davidnagli Changed it to a more clear and concise version.

Issue #685, opened 2018-01-28
rust wasm-gc test failure on Windows


brandon93s:
Windows build:

![image](https://user-images.githubusercontent.com/4730164/35478015-577b24bc-03a0-11e8-8a03-cb6b5f4398e0.png)

I _think_ the bundle being generated is null (error) for this test. Will need to install `rustup` on Windows to check. Any rust users able to debug?

PR #688, opened 2018-01-28
Add 'id' to the module variable

### LINKED ISSUE #619: "HOT" module API is not standard
**Choose one:** is this a  🙋 feature request? (actually, this is a bug)

module.hot.accept is too different from webpack or electron ones. Especially not the hot API, but module API. 

### 🤔 Expected Behavior

API should be consistent among different bundles.

### 😯 Current Behavior

Parcel's one is too simple.

### 💁 Possible Solution

1. Provide a module.id. Actually, this is the only change I am looking for.
2. In ideal - do not dispose the old accept handlers, and allow more that one accept callback.

### 🔦 Context

I come here from a React-Hot-Loader's issue (#500) -> version 4 is incompatible with parcel.
The problem - v4 introduces a new API, named `hot`. A customer could just call `hot(module)` and RHL will do the rest.

Meanwhile, a new logic bound to `module.id`, as long we are using it to track component instances, fabricated from the exported class to update them when the time comes.

You have that id, but as a part of module record.

theKashey:
Just letting `module.id` to exists.

Partially fixes #619.
Making parcel compatible with React-hot-loader v4 (https://github.com/gaearon/react-hot-loader/pull/818)

### DIFF
--- a/src/builtins/hmr-runtime.js
+++ b/src/builtins/hmr-runtime.js
@@ -1,7 +1,7 @@
 var global = (1, eval)('this');
 var OldModule = module.bundle.Module;
-function Module() {
-  OldModule.call(this);
+function Module(moduleName) {
+  OldModule.call(this, moduleName);
   this.hot = {
     accept: function (fn) {
       this._acceptCallback = fn || function () {};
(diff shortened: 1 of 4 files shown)


theKashey:
🙇tests are green on localhost 🤷‍♂️

davidnagli:
The CI test fail is from something else (see #685)

devongovett:
Thanks!

PR #690, opened 2018-01-28
[not working] Handle pwa manifest.json and prevent converting to js

niicojs:
detecting manifest.json file in httmlasset is a dirty hack.
but the current way of assigning asset only per extension pushed me in that direction


devongovett:
Should we even support `.json` as a manifest file? The spec says the extension should be `.webmanifest`: https://www.w3.org/TR/appmanifest/#media-type-registration

niicojs:
most people are doing `manifest.json`, but I'm good with `.webmanifest` if this is the way to do it with parcel.
checking why it doesn't work, it seems I would need a "don't process it further" in the Asset definition cause I'm generating a `.json` file which gets handler further by `JSONAsset`

davidnagli:
How does this relate to #313?

niicojs:
#691 fix #313 if you name the ressource with `.webmanifest` extension.  
This one would allow more flexibility with naming, like `manifest.json` which currently fails even with #691

HenrikJoreteg:
@devongovett I work with PWAs constantly and have never once seen someone do ‘.manifest‘ especially since the content is JSON it feels natural and is better with editor code highlighting etc, to do JSON.


Anyway, I'm fine with switching it to .manifest as a workaround, personally. But I feel like this will keep coming up.

devongovett:
This might not be easy to get working with the current architecture in which an asset is only compiled once. With this feature, assets may need to get compiled more than once to different formats. For example, if you had your `<link rel="manifest" href="manifest.json" />` as well as `require('./manifest.json')` in your JS, we'd need to produce a JSON output (handled as a `.webmanifest` file) as well as compile the JSON to JS for inclusion in the JS bundle.

I'll think more about it and see if there is a way to enable this. Open to suggestions!

devongovett:
Closing. We will come back to this in Parcel 2.

ptitjes:
Is there an issue to follow that for Parcel 2 ?
Just a reminder about what the specification says:

> NOTE: manifest.webmanifest or manifest.json?
> The official file extension for the manifest is .webmanifest. Some web servers recognize this extension and transfer the file using the standardized MIME type for a manifest (application/manifest+json). Developers can also choose a different extension (e.g. .json) or none at all (e.g. /api/GetManifest), but are strongly encouraged to transfer the manifest using the application/manifest+json MIME type.

Cf. https://www.w3.org/TR/appmanifest/#using-a-link-element-to-link-to-a-manifest

ptitjes:
OK, found it: https://github.com/parcel-bundler/parcel/issues/3362

PR #691, opened 2018-01-28
[feature] Add WebManifestAsset to handle W3C webmanifest

niicojs:
Allow pwa site to define a manifest.webmanifest according to https://www.w3.org/TR/appmanifest/.
Currently, you have to name it like that as if you name it with .json, the JsonAsset get triggered and convert it to js.
I have another PR for that, but it's not currently working (#690)


devongovett:
Awesome! Looks like there are some other fields supported by webmanifest files that we should probably be adding as dependencies as well. For example, `serviceworker.src`, `screenshots`. WDYT?

niicojs:
sure, I'll check that

devongovett:
Thanks, this looks great! Can you add a test as well?

niicojs:
I've added test coverage (that's a first for me, so I hope I did it right...).
Also, I've seen #313 which I missed before. I did not create a packager myself, did I miss something?

devongovett:
🏆

PR #695, opened 2018-01-29
Add JSON5 support

aicest:


devongovett:
Nice! This looks great. Thanks for the tests too!

Looks like this will also treat `.json` files as JSON5, as well as `.json5` files. Is this the desired behavior?

aicest:
@devongovett Thanks for reminding me. I have corrected it.

PR #707, opened 2018-01-30
Safe cssnano transforms by default

### LINKED ISSUE #698: @font-face files omitted when building with parcel build index.html
```javascript
import { h, app } from 'hyperapp';
import 'normalize.css';
import 'typeface-varela-round';
import './index.css';
```

This works fine when running `parcel index.html` but it does not work using `parcel build index.html`
Somehow, the typeface file gets omitted. The only way it'll work is if I manually copy the @font-face declaration to another css file that already contains some other css.

olizilla:
Pass `{safe: true}` to cssnano by default, to avoid the dangers of the advanced transforms that it would enable by default in cssnano@3.x

Adds a test (well co-opts an existing test) to ensure that @font-face files make it through to the final css bundle when `{ production: true }`

Once cssnano v4 is finally released as stable, the default options can be removed, as cssnano 4 will default to safe transforms only.

Fix #698

### DIFF
--- a/src/transforms/postcss.js
+++ b/src/transforms/postcss.js
@@ -51,7 +51,15 @@ async function getConfig(asset) {
 
   if (asset.options.minify) {
     config.plugins.push(
-      cssnano((await asset.getConfig(['cssnano.config.js'])) || {})
+      cssnano(
+        (await asset.getConfig(['cssnano.config.js'])) || {
+          // Only enable safe css transforms by default.
+          // See: https://github.com/parcel-bundler/parcel/issues/698
+          // Note: Remove when upgrading cssnano to v4
+          // See: https://github.com/ben-eb/cssnano/releases/tag/v4.0.0-rc.0
+          safe: true
+        }
+      )
     );
   }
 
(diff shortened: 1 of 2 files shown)


src/transforms/postcss.js
```diff
@@ -51,7 +51,15 @@ async function getConfig(asset) {
 
   if (asset.options.minify) {
     config.plugins.push(
-      cssnano((await asset.getConfig(['cssnano.config.js'])) || {})
+      cssnano(
+        (await asset.getConfig(['cssnano.config.js'])) || {
```
DeMoorJasper: Maybe change this to Object.assign?
So config doesn't overwrite defaults but appends them

olizilla: That's reasonable, but how would we communicate the net result to users. I think it's easier to reason about if we just use the `cssnano.config.js` as provided (_advanced user, let them choose_), and otherwise, just use safe defaults.

DeMoorJasper: Alright, fair enough

PR #708, opened 2018-01-30
🐛 launch https websocket server if --https

### LINKED ISSUE #704: 🐛 Accessing an app with a non-loopback address when using --https prevents the JS code from running
Following a discussion in https://github.com/parcel-bundler/parcel/pull/161#issuecomment-361304497. We found out that an error in JS makes it impossible to debug (or even run) an app using `--https` while accessing it with a non-loopback address (ie. not `localhost` and `127.0.0.1`).

### 🎛 Configuration

`parcel --https index.html`

### 🤔 Expected Behavior

The JS should run and we should be able to use HMR whatever address we use. It is necessary especially when debugging from a mobile browser.

### 😯 Current Behavior

Nothing shows up and there is an error in the console:

In Chrome:
> Mixed Content: The page at 'https://192.168.1.13:1234/' was loaded over HTTPS, but attempted to connect to the insecure WebSocket endpoint 'ws://192.168.1.13:57476/'. This request has been blocked; this endpoint must be available over WSS.

In FF:
> SecurityError: The operation is insecure.

### 💁 Possible Solution

As I described in the comment, I think we would need to pass the https server instance as a parameter (cf.https://github.com/websockets/ws/blob/master/examples/ssl.js) when instantiating the [ws server](https://github.com/parcel-bundler/parcel/blob/v1.5.1/src/HMRServer.js#L7) to make it secure and switch the protocol in the URL when `--https` option is on.

EDIT:
~~It might be good also to make sure that even if the HMR snippet fails in the browser it does not prevent the rest of the code from running~~ Will be fixed by #684 

I'm happy to try to submit a fix for that if we think it is a good idea.

niicojs:
Should fix #704

### DIFF
--- a/src/HMRServer.js
+++ b/src/HMRServer.js
@@ -1,10 +1,18 @@
+const http = require('http');
+const https = require('https');
 const WebSocket = require('ws');
 const prettyError = require('./utils/prettyError');
+const generateCertificate = require('./utils/generateCertificate');
 
 class HMRServer {
-  async start(port) {
+  async start(options = {}) {
     await new Promise(resolve => {
-      this.wss = new WebSocket.Server({port}, resolve);
+      let server = options.https
+        ? https.createServer(generateCertificate(options))
+        : http.createServer();
+
+      this.wss = new WebSocket.Server({server});
+      server.listen(options.hmrPort, resolve);
     });
 
     this.wss.on('connection', ws => {
--- a/src/Bundler.js
+++ b/src/Bundler.js
@@ -77,6 +77,7 @@ class Bundler extends EventEmitter {
       minify:
         typeof options.minify === 'boolean' ? options.minify : isProduction,
       hmr: typeof options.hmr === 'boolean' ? options.hmr : watch,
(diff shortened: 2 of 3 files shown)


src/HMRServer.js
```diff
@@ -1,10 +1,18 @@
...
 const prettyError = require('./utils/prettyError');
+const generateCertificate = require('./utils/generateCertificate');
 
 class HMRServer {
-  async start(port) {
+  async start(options = {}) {
     await new Promise(resolve => {
-      this.wss = new WebSocket.Server({port}, resolve);
+      let server = options.https
+        ? https.createServer(generateCertificate(options))
```
devongovett: Doesn't this result in generating two certificates? One for the HMR server and one for the normal web server? Perhaps we should share them...

niicojs: `generateCertificate` cache certificates. It seems I only get one in my .cache folder.

mbaranovski: Yes, the certificate is being cached.

PR #721, opened 2018-02-01
Support all Bundler options

### LINKED ISSUE #709: [Feature] Make cache directory path configurable
Maybe I missed some information somewhere. It seems the `.cache` folder cannot be changed by configuration, that is something what should be possible.

brandon93s:
The intent here is to support _all_ Bundler options ([here](https://github.com/parcel-bundler/parcel/blob/master/src/Bundler.js#L61-L91)) via the CLI without the need for adding a new explicit CLI flag to commander every time. The CLI interface is becoming crowded, often to handle edge cases. This change would automatically expose any new options as well. Ideally, this functionality would be added to documentation noting that "any Bundler option is supported" via the CLI. 

This would close #709. The following has been tested:

```shell
parcel .\index.html --cache-dir ".demo"
```

Pros:
- Support "all" options
- Support new options automatically 
- Keep CLI "interface" minimal for 99% use-case

Cons:
- Undocumented (via parcel help) functionality

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,8 @@
     "uglify-es": "^3.2.1",
     "v8-compile-cache": "^1.1.0",
     "worker-farm": "^1.5.2",
-    "ws": "^3.3.3"
+    "ws": "^3.3.3",
+    "yargs-parser": "^9.0.2"
   },
   "devDependencies": {
     "babel-cli": "^6.26.0",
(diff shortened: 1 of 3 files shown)


devongovett:
What options are not supported now that this enables?

lbguilherme:
-1 for this. `--help` is importante for every option. And this would allow things like `parcel --super-speed-mode index.html` to simply run without error.

brandon93s:
This would add support for `--cache-dir`, `--log-level` and `--kill-worers`.

PR #731, opened 2018-02-02
[RFC] Treeshaking experiment

DeMoorJasper:
This is an experimental implementation of treeshaking, currently very basic and only supports commonJS requires.

Main reason i threw this in a PR is for feedback, following are the main things i'm struggling with:
* Generating would happen twice, once before bundling and once after bundling, this should happen right before packaging in the case of treeshaking
* ast isn't being shared from worker back to bundler (which is needed to treeshake at rapid speed)
* Workers aren't being used for bundling, which could potentially slow down the building as bundle now treeshakes and calls generate


DeMoorJasper:
This was an experiment closing in favor of the upcoming scope hoisting

PR #733, opened 2018-02-02
Log bundle metrics

### LINKED ISSUE #415: 🙋 Report of generated bundle size + breakdown of contents
Hiya is it possible to get parcel build to output more info about final bundle size + a breakdown of modules that contributed to the bundles size?

DeMoorJasper:
Feature requested in closes #415

Minimal report (default behaviour)
screen shot 2018-02-02 at 17 15 25
Detailed report (cli flag `--detailed-report`)
screen shot 2018-02-02 at 17 15 07


src/utils/bundleReport.js
```diff
@@ -0,0 +1,106 @@
...
+  return res;
+}
+
+function prettifySize(size) {
+  let type = 0;
+  while (size > 1024) {
+    size = size / 1024;
+    type++;
+  }
+  return `${size.toFixed(2)} ${sizeTypes[type]}`;
```
brandon93s: While unlikely, you could overflow your array in which case you'd end up with `1.54 undefined`. Should at least safe-guard against it and stop dividing by 1024.

src/utils/bundleReport.js
```diff
@@ -0,0 +1,106 @@
+const path = require('path');
+const sizeTypes = ['B', 'KB', 'MB', 'GB'];
```
brandon93s: Should `KB` be `kB`?

src/utils/bundleReport.js
```diff
@@ -0,0 +1,106 @@
...
+
+function spaces(amount) {
```
brandon93s: Can use `String.prototype.repeat()` here:

```js
function spaces(count) {
	return ' '.repeat(count);
}
```

It also correctly returns an empty string for `0`.

There is also `String.prototype.padEnd()` in Node 8+ which may simplify some code.

brandon93s:
Curious why the total time is so much greater (as a %) than the bundle time in both of the screenshots.

src/utils/bundleReport.js
DeMoorJasper: U are correct it should be kibiByte (kB)

src/packagers/CSSPackager.js
```diff
@@ -23,6 +23,7 @@ class CSSPackager extends Packager {
     }
 
     await this.dest.write(css);
+    this.bundle.addAssetSize(asset, this.dest);
```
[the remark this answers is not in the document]
DeMoorJasper: I can possibly move it to asset, but `stat` is an extra operation and it wouldn't keep in mind things like treeshaking and uglifying. So i guess this is the correct implementation, keeping in mind that some Packagers have a different way of writing files, for example sourcemap writes everything at the end and rawPackager writes the entire file the same way it got it (without a writestream)
I did my best to abstract it using `bundle.addAssetSize`

src/Bundle.js
```diff
@@ -18,6 +18,14 @@ class Bundle {
     this.siblingBundles = new Set();
     this.siblingBundlesMap = new Map();
     this.offsets = new Map();
+    this.startTime = new Date().getTime();
```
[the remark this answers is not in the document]
DeMoorJasper: I will change it

src/utils/bundleReport.js
DeMoorJasper: I'll use the repeat as babel didn't catch padEnd when i tried to build

src/utils/bundleReport.js
DeMoorJasper: Ow yeah, thought it wouldn't be an issue as a bundle larger than 1024 GB would be insane for a web application (or any other application)

src/utils/bundleReport.js
```diff
@@ -0,0 +1,106 @@
...
+function prettifySize(size) {
+  let type = 0;
+  while (size > 1024) {
+    size = size / 1024;
+    type++;
+  }
+  return `${size.toFixed(2)} ${sizeTypes[type]}`;
+}
+
+function prettifyTime(time) {
```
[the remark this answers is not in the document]
DeMoorJasper: Will change

DeMoorJasper:
@brandon93s the bigger time is because the total time in the metrics is the total processing time, basically the time it would've taken if ran on a single thread/worker.
The total time that it took to build is the conceived time.
So in theory the metrics total time would be near `total bundle time * threads` on large projects

src/utils/bundleReport.js
```diff
@@ -0,0 +1,95 @@
+const path = require('path');
+const prettifyTime = require('./prettifyTime');
+const sizeTypes = ['B', 'KB', 'MB', 'GB'];
```
lbguilherme: What about using `KiB`, `MiB`, `GiB` or changing the factor to 1000 instead of 1024, for correctness?

devongovett:
Question: should we only do this on `parcel build` and not in development mode? Seems like bundle sizes aren't that useful in development, and will be wrong anyway since code isn't minified/optimized in dev mode anyway. Thoughts?

src/utils/bundleReport.js
DeMoorJasper: I'm not sure, this is the correct notation as used in almost every OS

lbguilherme: I know that Windows uses the wrong KB notation, and that most Linux distros use KiB notation. Don't know about Mac, Android, iOS and others.
(nice!)

devongovett:
Ok I ended up reworking this a bit to make it a bit prettier. Here's what it looks like:

screen shot 2018-02-13 at 9 53 32 pm

When you want a detailed report it looks like this:

screen shot 2018-02-13 at 9 51 42 pm

As you can see, it's clearer which items are bundles, and which items are assets within bundles. Bundles are bold, and assets are dimmed and indented. Plus it's colorful! Directory names are dimmed and filenames highlighted. 🎨

It will also warn when you have a large bundle size with a warning triangle and yellow text:

screen shot 2018-02-13 at 9 52 38 pm

I moved the table rendering stuff to the logger class, and used the widely used `filesize` module to format file sizes. I also made it so each packager doesn't need to manually add each asset size - the bundler handles that as it is packaging. There is a default `getSize` function in the packager that returns the current output size of the file, and the base Packager implements the default version of this. Only RawPackager needed to override it.

Let me know what you think!

PR #734, opened 2018-02-02
[RFC] Support custom TypeScript transformers

### LINKED ISSUE #699: Support for Custom Typescript transformers
Any plans to support custom typescript transformers? So I can use https://github.com/deamme/ts-transform-css-modules-next, https://github.com/deamme/ts-transform-inferno etc.

References:
https://github.com/Microsoft/TypeScript/pull/13940
https://github.com/TypeStrong/ts-loader/pull/535

fathyb:
Fixes #699 

**Configuration proposal** :
- A `parcel` field is added to the `tsconfig.json`.
- This field can contain the `transformers` property, the property should be a path relative to the `tsconfig.json` to a Common JS module.
- This module should export a function that returns an object conform to the TypeScript transformer API, it should follow this interface :
```ts
import {SourceFile, TransformerFactory} from 'typescript'

type TransformerModule =
  () => {
    after?: TransformerFactory<SourceFile>[]
    before?: TransformerFactory<SourceFile>[]
  }

```

**Example** : 
`tsconfig.json` :
```json
{
  "parcel": {
    "transformers": "./custom-transformer"
  }
}
```

`custom-transformer.js` :
```js
const {anotherTransformer, coolTransformer} = require('somewhere')

module.exports =
  () => ({
    before: [coolTransformer()],
    after: [anotherTransformer()]
  })

```

### DIFF
--- a/src/Asset.js
+++ b/src/Asset.js
@@ -93,9 +93,11 @@ class Asset {
     return URL.format(parsed);
   }
 
-  async getConfig(filenames) {
-    // Resolve the config file
-    let conf = await config.resolve(this.name, filenames);
+  getConfigPath(filenames) {
+    return config.resolve(this.name, filenames);
+  }
+
+  async loadConfig(conf, filenames) {
     if (conf) {
       // Add as a dependency so it is added to the watcher and invalidates
       // this asset when the config changes.
@@ -104,6 +106,12 @@ class Asset {
     }
 
     return null;
+  }
+  async getConfig(filenames) {
+    // Resolve the config file
+    let conf = await this.getConfigPath(filenames);
+
+    return this.loadConfig(conf, filenames);
   }
 
   mightHaveDependencies() {
--- a/src/assets/TypeScriptAsset.js
+++ b/src/assets/TypeScriptAsset.js
@@ -1,3 +1,4 @@
+const path = require('path');
 const JSAsset = require('./JSAsset');
 const localRequire = require('../utils/localRequire');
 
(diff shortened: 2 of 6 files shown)


devongovett:
Are these transformers specific to parcel? I'm not sure we should have parcel specific things inside the typescript config file. Is there a way to specify custom transformers in tsconfig already?

fathyb:
> Are these transformers specific to parcel?

They are not specific to Parcel, they are part of the TypeScript public API. However, the file exporting the transformers (what's referenced by `parcel.transformers` in the `tsconfig.json`) is specific to Parcel.
It's a function so in the future we can pass additionnal parameters (eg. [ts-transformer-keys](https://github.com/kimamula/ts-transformer-keys) needs the type-checker passed as an argument).
This feature is already supported by the two most popular TypeScript loaders for Webpack : `awesome-typescript-loader` and `ts-loader`.

> I'm not sure we should have parcel specific things inside the typescript config file.

I totally agree on that and I'm open to suggestions as to where we should specify this.

> Is there a way to specify custom transformers in tsconfig already?

TypeScript does not specify any fields for the transformers in `tsconfig.json` and `tsc` does not support transformers, even if it's a public API. There is [an opened issue about this](https://github.com/Microsoft/TypeScript/issues/14419#issuecomment-283721389), but their position is pretty clear :

  > We do not plan on exposing compiler plugin system in the short term. The transforms are exposed as part of the public API as you noted, and we are in the process of writing documentation and samples for these.

I think that their idea is that plugins may make TypeScript fragmented and less portable.

zerkalica:
To support transforms you just need replace typescript to ttypescript in [TypeScriptAsset](https://github.com/parcel-bundler/parcel/blob/master/src/assets/TypeScriptAsset.js#L12)

```js
let typescript = await localRequire('ttypescript', this.name);
```

[ttypescript](https://github.com/cevek/ttypescript) wraps typescript and loads transforms from tsconfig.json.

I can create separate TTypeScriptAsset, but may be add simple condition to TypeScriptAsset? Like this:

```js
let typescript = await localRequire(isTTypescriptInDeps ? 'ttypescript' : 'typescript', this.name);
```

0xcaff:
Why is this closed? It seems that this PR was abandoned in favor of using ttypescript but this change hasn't happened yet.

PR #736, opened 2018-02-03
Added -o option to readme

danmarshall:
I was going to make a feature request, then happily discovered that this feature already exists - but was undocumented. Until now. 🍻


devongovett:
Thanks, I just changed it to list the full `--open` option instead of `-o` as we are probably going to change `-o` to mean something else (output file name) soon.

PR #738, opened 2018-02-04
fix PromiseQueue returning null when there are no jobs

pselden:
When PromiseQueue is called with no jobs it returns null instead of a promise that is already resolved with an empty set. This causes "await run()" to evaluate to null, and a "Cannot read property 'Symbol(Symbol.iterator)' of null" error.

Fixes #664


devongovett:
Ah, great catch! Thank you! 🙏

PR #742, opened 2018-02-04
Add `.toml` asset support

brandon93s:
`toml` was added as a dependency in #623. This exposes it as an asset, similar to YAML support. 

Easy win! 🎉


brandon93s:
On a related note... should `js-yaml` be a devDependency and leverage `localRequire`? We're not currently using it outside of `YAMLAsset`.

src/assets/TOMLAsset.js
```diff
@@ -0,0 +1,21 @@
...
+    this.type = 'js';
+  }
+
+  parse(code) {
+    return toml.parse(code);
+  }
+
+  generate() {
+    return {
+      js: `module.exports=${JSON.stringify(this.ast, false, 2)};`
```
devongovett: Looks like toml can contain more types than JSON, e.g. dates. Maybe we should use https://github.com/commenthol/serialize-to-js like #682 is going to do for YAML?

brandon93s: 👍 yeah, probably a good idea. Will update this once it lands.

PR #751, opened 2018-02-05
Add no-config support for jsx

sheeldotme:
Parcel aims to be a zero configuration web application bundler yet it currently requires configuration for JSX support.  This PR changes that.

For context:
@amPerl on slack stated:

> do you mean automatically including react presets detecting jsx? in which case there might be other frameworks that use similar syntax with different presets (edited)

and my response was:

> Great point @amPerl, the transform-react-jsx plugin actually allows you to define a custom jsx pragma as follows:

```/** @jsx dom */
/** @jsxFrag DomFrag */
```

> So in reality, one can use this adaptation with any library that utilizes jsx with no further configuration files required.  If one does prefer to use a configuration file to override the pragma, of course that option is always available.  So I really don't see the downside here, with considerable value for those who came looking for "zero configuration" which is what parcel promises.

If there's a different way you'd like to have this implemented let me know.  If you have doubts that this should be included as part of parcel I'd love to hear your objections.

Thanks,

Sheel


devongovett:
Awesome, thanks @sheeldotme! 🏆 I added a few tests.

I've also added support for JSX in normal `.js` files as well, including support for other react-like libraries like preact. Parcel looks for them in the package dependencies and sets the JSX pragma automatically. This only happens in application code, not in node_modules.

Issue #753, opened 2018-02-06
🙋 Add --rev option to `build` command for revved filenames


gamebox:
### 🤔 Expected Behavior

When using `parcel build`, one should be able to supply a `--rev` option at the cli to have the resulting bundles filename be `bundlename-LAST_SIX_OF_MD5_HASH.ext`

### 😯 Current Behavior

No such option exists, requiring a multistep build process for production deployments

### 💁 Possible Solution

I think the cli flag discussed above is the best solution.

### 🔦 Context
We like to have all of our application bundles be "revved" with a chunk of the MD5 hash of the file appended to the bundle name for easy cache invalidation for the client.  We can do this now, but it requires yet another step in the build process.

gamebox:
If it helps, I've went through the source and I _think_ I could put in a PR for this relatively easily.

devongovett:
Should be solved by #1025 which generates content-hashed filenames for static assets. Please help test using the master branch - a release will hopefully come next week!

PR #756, opened 2018-02-06
Add --rev option to allow for bundle revisions

gamebox:
This is related to #753


lbguilherme:
1. I don't think this behavior need to be behind a flag. Is there a downside of always adding the hash?
2. The outer bundle probably shouldn't have hash, as it is referred by name (the index.html).
3. RawPackager needs work too, as it doesn't use `setup` from Packager.

devongovett:
Should be solved by #1025 which generates content-hashed filenames for static assets. Please help test using the master branch - a release will hopefully come next week!

PR #769, opened 2018-02-08
Pug support

shawwn:
This PR adds support for Pug (formerly known as Jade).

It's a necessary step toward supporting Vue, since both `vue-loader` and `vueify` support pug/jade.

Custom filters and local variables can be defined by creating a `.pugrc.js` file like so:

```js
let options = {
  locals: { youAreUsingPug: true }
};
options.filters = {
  'my-own-filter': function (text, options) {
    if (options.addStart) text = 'Start\n' + text;
    if (options.addEnd)   text = text + '\nEnd';
    return text;
  }
};
module.exports = options;
```

See the [pug – reference](https://pugjs.org/api/reference.html) for the complete list of options.

Markdown and other filters are supported. See [pug – filters](https://pugjs.org/language/filters.html#custom-filters)


src/Parser.js
```diff
@@ -34,6 +34,7 @@ class Parser {
 
     this.registerExtension('html', './assets/HTMLAsset');
     this.registerExtension('htm', './assets/HTMLAsset');
+    this.registerExtension('pug', './assets/PugAsset');
```
brandon93s: Should support `.jade` too, it's still common enough.

Ty3uK:
@shawwn how do you trigger rebuild pug template when user changed included or extended file?

shawwn:
@Ty3uK Good point! I don't.

Is there a pug method to get a list of dependencies to be watched?

Ty3uK:
@shawwn nope, you'll need to build AST from template and mark all include and extends as dependency for parcel. That's covered by my [plugin](https://github.com/Ty3uK/parcel-plugin-pug).

devongovett:
@Ty3uK how would you feel about contributing your plugin to parcel core?

Ty3uK:
@devongovett it’s would be great, but now plugin has some unresolved issues.

laosb:
Is there any progress, or something I can help with? Sounds like we're not going to accept this PR but make @Ty3uK 's plugin into core?

It said that #5 depends on this. I'm really looking forward to use Vue with Parcel.

davidnagli:
@Ty3uK Any updates?

@devongovett It looks like this whole pug thing is holding up Vue support. Should we just merge this PR for now, adding at least initial Pug support, and then once parcel-plugin-pug works out it's issues we can consider merging it into core instead? 

This PR works, and although it doesn't fully support HMR, I think at least some Pug support is better than none.

Ty3uK:
@davidnagli yep, just need to fix some small issues. Give me one day, 🙏

Ty3uK:
@davidnagli check out latest version, all critical issues has been resolved

devongovett:
Fixed by #1051.

CareyToboo:
@shawwn Is it possible to pass custom locals when using pug? and my pacel version is 1.7.1.

I see the code in 'parcel/src/assets/PugAsset.js',  there only three properties 'filters/filterOptions/filterAliases' in pug config file are useable. 

I want to use pug locals to complie the index.pug.   

.

PR #772, opened 2018-02-09
Fix shared config objects

shawwn:
This fixes situations where `config` objects are shared across assets
and its properties are being mutated on a per-asset basis.


devongovett:
Maybe we should just do the cloning in the `config.load` helper instead of every asset?

shawwn:
That won't work because of shared `this.package` objects:

https://github.com/parcel-bundler/parcel/pull/772/files#diff-11ce184cb188e058ad296dd1aa693ae0R12

Those should be cloned too.

PR #773, opened 2018-02-09
Include data- attributes in HTML dependency processing

### LINKED ISSUE #771: How to import an image from any attribute (like data-srcset)?
Hello, and thanks for the amazing work!

I'm using `lazysizes`, and I need to set my images in a `data-srcset` attribute, but...

**this is working:**
``

**this is not:**
``

Couldn't find an answer in the documentation...

Thank you!

brandon93s:
Expands on #681 and closes #771.

Adds support for custom (`data-*`) attributes:

`data-src`
`data-href`
`data-poster`
`data-srcset`

These are commonly used by JS libraries that provide common functionality, like the popular https://github.com/aFarkas/lazysizes.


devongovett:
I hesitate to do this because those attributes aren't really standard. WDYT?

brandon93s:
Had the same thought, but the only scenario it'd cause a problem is if they we're being used for something completely unrelated to the standard attribute. While possible, seems unlikely they'd be used for non-asset purposes.

SkyGuardian42:
It looks like this was closed without a merge, is there any way we can still use this?

DeMoorJasper:
@SkyGuardian42 It didn't get merged because this isn't standard html, someone could eventually make a plugin

mcrider:
@DeMoorJasper any guidance on how to make this into a plugin? Would it be through an html packager or an asset type? I've been studying the docs and code and am having a tough time getting going :\ Basically I want to rename my output file based on a link's data attribute and this looked like a good starting point.

alvaaz:
I found a plugin, I hope I help you [parcel-lazy-lazy](https://github.com/parcel-bundler/parcel/pull/773)

Issue #735, opened 2018-02-03
🙋 Deterministic builds


termhn:
I want to use Parcel in my project [RippleWarpWallet](https://github.com/termhn/ripplewarpwallet) and possible future derivatives. One thing blocking that happening right now is that even given the same exact input code, building twice will produce slightly different outputs. It is really nice to be able to have a deterministic output given the same inputs because this lets people verify that the live online version of the tool is really the same as what is directly built from the source code (verifiable builds).

### 🎛 Configuration (.babelrc, package.json, cli command)

In my `package.json` I have this script which runs to do the preliminary parcel build

```json
    "parcelbuild": "parcel build index.html --no-minify --public-url ./ --out-dir ./build"
```

I then do some post processing in a script to directly insert the bundled code into the HTML document because this is convenient for my purposes.

### 🤔 Expected Behavior

Given the same input code, two builds one after the other should produce the same output code bundles.

### 😯 Current Behavior

There are small differences in the code that are insignificant to functionality but result in different SHA hashes and hits in diff tools, making verifying builds much harder for the average user. An example is these two builds which were done successively:

https://gist.github.com/termhn/ddaf437fc20fa50afe0378256869c901

Which have different SHA256 hashes and if compared in a text diffing tool like https://www.diffchecker.com/ result in several minor changes.

### 💁 Possible Solution

Not sure. Haven't yet dug into the code to see where these number changes are coming from or possible solutions.

### 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Parcel           | 1.5.1
| Node             | 8.1.3
| npm/Yarn         | yarn 0.27.5
| Operating System | Windows 10

gamebox:
I'd also support looking into this as I just filed an issue(#753 ) for file "rev"ing which obviously requires the hash to remain consistent for the same input.

termhn:
Right so I got a little chance to play around with the code. I figured out a way to get this to work, but it's sort of a patch on top... I haven't quite figured out the real source of the issue. The problem is that Asset IDs are not consistently assigned between builds, and the Asset ID is what is referenced for the require polyfills. This has something to do with the fact that when assets are not fully cached (and maybe even when they are for larger projects?), there are multiple duplicate assets created for the same files. For example, in a basic setup with an `index.html` that references `index.js` and then `index.js` references `include.js`, there are two distinct `Asset`s created for the `index.html`, three distinct assets created for `index.js` and two distinct assets created for `include.js`

[...]

PR #780, opened 2018-02-10
Make builds deterministic 

termhn:
Fix/improvement based on #735 

Makes asset ids the last 5 characters of the md5 hash of the path of that asset.


termhn:
wip... still looking to improve this and pass all the tests

termhn:
Alright this passes tests now. Main downside is that it could increase minified file sizes a small amount and it may very slightly affect performance, maybe noticeable for projects with a very large number of assets.

jeanfortheweb:
I had a similar approach a few days ago, but I used the Asset name itself since it's already the absolute asset path itself which should be perfect for unique ID's. Anyway, I also discovered that assets are forced to be unique created inside the bundler, therefore the current ID generation should already deterministic as long as the assets do not change. If that's not the case, then the files are not getting read in the same order every time. 

At least I would think about moving the asset creation from Bundler.js:259 (on 1.4.1) to the Asset itself using a factory pattern. Guess that would make it a lot easier to reason about.

termhn:
That's what I did originally, but they change across different machines/operating systems. Using a normalized relative path seems to be the best options as far as that goes. See #735 for discussion/my findings about why this isn't happening--ultimately I reached a similar conclusion to you. Something's going on where assets are being loaded multiple times and not always in the same order which is why the current system doesn't work properly.. possibly because of the way workers are used to parallelize the work.

jeanfortheweb:
Just a guess but it shouldn't happen as load time, since this is happening in the main thread on the bundler... hmmm, well if the loading order depending on the native file system layer can't be deterministic then these IDs will also never be while I still actually like the simplicity of numeric IDs.

If we move the asset creation onto a static factory function on the asset class itself, we could store the already loaded asset on a map there.

PR #791, opened 2018-02-12
Bugfix/doubled assets

### LINKED ISSUE #790: Doubled Assets in Output
### 🤔 Expected Behavior

Hoisted assets and all of its dependencies should not get rendered twice when they are already present on ancestor level.

### 😯 Current Behavior

Hoisted assets and all of its dependencies are getting rendered inside the ancestor and children.

### 💁 Possible Solution

I've found several solutions to fix this behavior but always knowing that it must be possible a way earlier than package time. I've finally made it work with some tweaks to createBundleTree and moveAssetsToBundle which I provide as an PR to 1.4.1.

### 🔦 Context

While integrating parcel into one of our projects, with lots of dependencies and sources as also complex requirements to each other, I've noticed that the rendered asset bundles a way bigger than they should as also that React and other libraries are getting initialized more than once causing wierd behaviors.

### 💻 Code Sample

Taking the dynamic-hoist test from parcel and changing it slightly will trigger the bug.
Inside `integration/dynamic-hoist`, require `common.js` inside `index.js`file, as also removing `common.js`from `a.js` replacing the exported value with a static `2`.

Then we've created a case where a dependency is present in a child while it is already present in the common ancestor or parent. When running the test now, it will fail since the `common.js` asset will be rendered in both, `b.js` and `index.js`. Same applies for all of it's dependencies.

### 🌍 Your Environment

| Parcel           | 1.4.1
| Node             | 8.x
| npm/Yarn         | 0.27.x
| Operating System | Windows/OSX

jeanfortheweb:
Fixes #790, by calling `moveAssetToBundle` always when we know that the current bundle of the asset is different.

Instead, `moveAssetToBundle` will be aware of that certain assets could already be present, but also that they could also present in bundles where there shouldn't anymore.


jeanfortheweb:
Shouldn't go into master.
This fix is based on 1.4.1 and should go there first.

PR #794, opened 2018-02-12
Allow dots in lazy loaded file names

jsiebern:
I stumbled across this issue using a plugin to rename the output files


brandon93s:
Thanks! 🏅

PR #796, opened 2018-02-12
Bundle node modules on --target=node/electron

lbguilherme:
This adds the option `--bundle-node-modules` to force bundling all files even on `--target=node`. This *does not work with all modules* , but should work for most simpler cases. It will break apart if a module try to do filesystem operations or depend on some internal files. Things such as `puppeteer` or `lsmod` will certainly fail.

This is ready to merge (please review!).

In a followup pull request I'll bring support for native node modules. Then we can start checking what packages don't build and what can be done for them.

This still isn't capable of bundling Parcel itself, but eventually it should be.

---

The major change is a rewrite of the Resolver, that now have to conditionally handle `pkg.browser` (the old resolver had no way to disable that). The other files touched are minimal.


devongovett:
I've rewritten the resolver in #850. It should be easier to disable builtins with that.

lbguilherme:
I'll rebase this once #850 is merged. Much better with the new resolver.

lpil:
Hi all. Would it be possible to get this rebased and merged?

devongovett:
Updated this in #1690. Closing.

Stradivario:
@lbguilherme where it is implemented ? 

From a long time ago i try and looking for solution but this option is unavailable.

I am getting 

```
error: unknown option `--bundle-node-modules'
```

Any ideas ?

PR #798, opened 2018-02-13
Recognize path.join when handling readFileSync

lbguilherme:
Recognize and handle `fs.readFileSync(path.join(__dirname, 'test.txt'), 'utf8')`.

Parcel does it itself on JSPackager.js.


devongovett:

PR #805, opened 2018-02-14
Automatically install missing dependencies, part 2

### LINKED ISSUE #376: 🙋 Automatically install missing dependencies, part 2
Now that #306 has landed, it would be awesome to take this even further. Currently we automatically install dependencies loaded by the internal `localRequire` helper. These are things like compilers (e.g. typescript, coffeescript), and plugins (e.g. postcss plugins, parcel plugins, etc.). 

It would be super cool to take this even further and transparently install any npm dependencies you require in your code for you. e.g. if you add `require('react')` in your code, and it's not installed, Parcel can install it for you.

Thoughts?

davidnagli:
Closes #376 

- [x] Core functionality (should install any missing dependencies through npm)
- [x] Add CLI flag to disable it
- [x] Limit to only run in dev mode


src/Bundler.js
```diff
@@ -328,26 +329,43 @@ class Bundler extends EventEmitter {
...
-          await asset.loadIfNeeded();
-          thrown.loc = dep.loc;
-          thrown = asset.generateErrorMessage(thrown);
+        let isLocalFile = dep.name.startsWith('.');
+
+        if (!isLocalFile) {
+          try {
+            try {
+              logger.status(emoji.progress, `Installing ${dep.name}...`);
+              await npm.install([dep.name], {
```
devongovett: Any reason you can't use the `installPackage` helper that we already have? That also supports yarn.

davidnagli: Didn’t know it existed 😬 😂

davidnagli: The only problem is that it saves dependencies as a dev dependencies by default with no way to override that, so I’m gonna need to change it and refactor the code that uses it.

devongovett: Sure. You could just make dev dependency the default, and have an option to override it. Then you wouldn't need to change the existing calls.

src/utils/installPackage.js
```diff
@@ -13,12 +14,18 @@ async function install(dir, modules, installPeers = true) {
...
-      install = spawn('yarn', ['add', ...modules, '--dev'], options);
-    } else {
-      install = spawn('npm', ['install', ...modules, '--save-dev'], options);
+    let args = ['add', ...modules];
+    if (saveDev) {
+      args.push('-D');
     }
 
+    let command = 'npm';
+    if (commandExists('yarn')) {
```
lbguilherme: Better preserve the old logic of `location && path.basename(location) === 'yarn.lock'`. Just because `yarn` is installed, doesn't mean it is being used.

davidnagli: Interesting... is there a case where you wouldn’t want to use yarn if you already have it?

davidnagli: Interesting... is there a case where you wouldn’t want to use yarn if you already have it?

lbguilherme: Sure there is! I have yarn and I don't use it for every project. Specially when working with a team, everyone must use the same package manager for that project.

davidnagli: Got it :)

We should keep both though, right? So we should check for the yarn lock file **AND** the yarn command, that way if they have a yarn lock but no yarn they can still use npm. Or is that also not necessary?

lbguilherme: I see no need for keeping both checks, but I see no harm as well. Your call.

src/utils/installPackage.js
```diff
@@ -3,8 +3,9 @@ const config = require('./config');
 const path = require('path');
 const promisify = require('./promisify');
 const resolve = promisify(require('resolve'));
+const commandExists = require('command-exists').sync;
```
brandon93s: This is used in an async method. Should use the promise version exposed by `command-exists` with `await`.

src/utils/installPackage.js
brandon93s: +1 for keeping the `yarn.lock` check. I'm OK with both, but the lock check is going to be an indication of what the team / project is using.

src/Bundler.js
```diff
@@ -332,26 +333,42 @@ class Bundler extends EventEmitter {
...
+              logger.status(emoji.progress, `Installing ${dep.name}...`);
+              let dir = Path.dirname(asset.name);
+              await installPackage(dir, [dep.name], true, false);
+            } catch (npmError) {
+              logger.error(npmError.message);
+            }
+
+            return await this.resolveAsset(dep.name, asset.name);
+          } catch (e) {
+            if (dep.optional) {
```
brandon93s: Should "optional" dependencies be installed automatically, or should we just skip these like before? Thoughts?

src/Bundler.js
```diff
@@ -332,26 +333,42 @@ class Bundler extends EventEmitter {
...
-        }
-
-        // Generate a code frame where the dependency was used
-        if (dep.loc) {
-          await asset.loadIfNeeded();
-          thrown.loc = dep.loc;
-          thrown = asset.generateErrorMessage(thrown);
+        let isLocalFile = dep.name.startsWith('.');
+
+        if (!isLocalFile) {
```
brandon93s: Invert the `if` to reduce the nesting.

src/Bundler.js
```diff
@@ -332,26 +333,42 @@ class Bundler extends EventEmitter {
...
+          try {
+            try {
```
brandon93s: Hmm. Nested `try` statements doesn't feel right. Can they be sequential?

```js
// Install
try {
  logger.status(emoji.progress, `Installing ${dep.name}...`);
  let dir = Path.dirname(asset.name);
  await installPackage(dir, [dep.name], true, false);
} catch (npmError) {
  logger.error(npmError.message);
}

// Resolve
try {
  return await this.resolveAsset(dep.name, asset.name);
} catch (e) {
  if (dep.optional) {
    return;
  }
```

Or handle all in a single catch based on the error?

src/Bundler.js
```diff
@@ -332,26 +333,42 @@ class Bundler extends EventEmitter {
...
+              logger.status(emoji.progress, `Installing ${dep.name}...`);
+              let dir = Path.dirname(asset.name);
+              await installPackage(dir, [dep.name], true, false);
+            } catch (npmError) {
+              logger.error(npmError.message);
```
brandon93s: If install fails, should we continue or should it bail out? If the build would fail, could just bail out. If that's the case, could probably consolidate into a single `try...catch` .

brandon93s:
Great start! 👍

src/utils/installPackage.js
davidnagli: Ya that’s exactly what I originally wanted to do that 😃 

But... because of the way the `command-exists` package is written, it’s doesn’t really work out. 😕 

Pretty much, `commandExists()` conveys the result of whether or not it found the command by resolving the returned promise for “package found”, and rejecting the promise for “package not found”. This is a problem if we’re using async/await because `await` handles promise rejections by throwing. So pretty much whenever yarn _isn’t_ found it’ll throw an error instead of nicely informing us.

I guess it’s possible to do it using a try/catch, but that’s really messy. The only other way to use the async version of `command-exists` is to just directly use the promise, but that’s really complicated since `promise.finally` hasn’t landed in Node yet and it would probably require nesting all the rest of the install code.

So I can still use the async version... just it won’t be as clean, and I don’t really think there’s that much of a performance loss from doing it synchronously, especially considering that the function it’s running in is gonna be executing asynchronously anyway. 

Your call.

src/Bundler.js
davidnagli: Oh ya good catch!

davidnagli: Not necessary once I used your try/catch split idea

devongovett:
IMO we should only do this in development mode (e.g. not `parcel build`). In CI environments, we definitely shouldn't be auto installing things for example.

src/Bundler.js
davidnagli: By the looks of it, optional deps are pretty much dependencies that _are_ supposed to get installed/bundled, but aren’t supposed to throw an error if they aren’t able to be resolved or installed.

So ideally we should attempt to install any optional deps, but we shouldn’t block the build if the install fails.

davidnagli:
Ready for review 😃

src/cli.js
```diff
@@ -39,6 +39,12 @@ program
   .option('--no-hmr', 'disable hot module replacement')
   .option('--no-cache', 'disable the filesystem cache')
   .option('--no-source-maps', 'disable sourcemaps')
+  .option('--no-autoinstall', 'disable autoinstall')
+  .option(
+    '-m, --package-manager <packageManager>',
```
brandon93s: Is this option really necessary? Our approach to inferring the package manager seems pretty solid.

src/Bundler.js
```diff
@@ -332,14 +335,28 @@ class Bundler extends EventEmitter {
       let thrown = err;
 
       if (thrown.message.indexOf(`Cannot find module '${dep.name}'`) === 0) {
+        let isLocalFile = dep.name.startsWith('.');
```
brandon93s: Could the dependency be an absolute path, which would still be a local file? Trying to install in that scenario would blow up on `npm install`.

src/utils/installPackage.js
```diff
@@ -3,22 +3,67 @@ const config = require('./config');
...
-      install = spawn('yarn', ['add', ...modules, '--dev'], options);
+    let packageManagerToUse;
+    if (packageManager) {
+      packageManagerToUse = packageManager;
+    } else {
+      // If no package manager specified, try to figure out which one to use:
+      // Default to npm
+      packageManagerToUse = 'npm';
+      // If the yarn command exists and we find a yarn.lock, use yarn
+      if (commandExists('yarn')) {
```
brandon93s: `if (commandExists('yarn') && await fs.exists(...))`

src/Bundler.js
```diff
@@ -332,14 +335,28 @@ class Bundler extends EventEmitter {
...
+        // Attempt to install missing npm dependencies
```
brandon93s: Mind adding tests for this?

src/Bundler.js
```diff
@@ -332,14 +335,28 @@ class Bundler extends EventEmitter {
...
+        if (!isLocalFile && this.options.autoinstall) {
+          logger.status(emoji.progress, `Installing ${dep.name}...`);
+          await installPackage(
```
brandon93s: Might be time for a config object to `installPackage`... 5 parameters!

src/utils/installPackage.js
brandon93s: We're already inside of a promise chain (`return new Promise...`) so you could use the Promise API with a bit of refactoring. Alternatively, you could create a `hasYarnInstalled` method:

```js
hasYarnInstalled = async () => {
	try {
		return await commandExists('yarn');
	} catch (err) {
		return false;
	}
}

if (await hasYarnInstalled() && fs.exists(...)) {
   ...
}
```

src/Bundler.js
davidnagli: Ya you’re right.

I would also have to add support for tilde paths etc now that #850 landed.

src/Bundler.js
davidnagli: Yep. I’ve tried a few times but I’m gonna need to wait for #881 to land. Atm the second I add tests which use the test `/input` directory it causes a bunch of other flaky tests to go crazy.

src/cli.js
davidnagli: That’s true, but I don’t really see any downsides to giving the users manual control if necessary.

One scenario that I would imagine will be particularly problematic is if a user is starting a brand new project with Parcel and has no dependencies yet. They try to add their first import statement, and Parcel automatically uses npm because it doesn’t see a `yarn.lock`, even though the user would rather use yarn. In that case they would want to use the `-m yarn` flag to force it to use yarn. 

(I mean I suppose they could just `touch yarn.lock` to generate an empty one, but that’s kinda hacky and just another point for confusion)

I can remove it if you think we should, but I just don’t see any downsides to leaving it.

src/utils/installPackage.js
davidnagli: There’s a reason I separated them. I only want the `else` to apply if `await fs.exists(...)` is **false** **AND** `commandExists(‘yarn’)` is **true**.

Pretty much only if the user has yarn, but no `yarn.lock`, then I want to trigger the warning. If they don’t have yarn, then no need to warn them.

brandon93s: Ah I see. The warning may be unnecessary. If they haven't explicitly asked for yarn, and there is no yarn.lock, then using npm without a warning seems valid - expected even.

src/Bundler.js
devongovett: You'll need something like `let isLocalFile = /^[/~.]/.test(dep.name);` in order to support the new resolver. See https://github.com/parcel-bundler/parcel/blob/master/src/Resolver.js#L88-L114

src/Bundler.js
devongovett: You'll need something like let isLocalFile = /^[/~.]/.test(dep.name); in order to support the new resolver. See https://github.com/parcel-bundler/parcel/blob/master/src/Resolver.js#L88-L114

src/Bundler.js
davidnagli: Done

src/utils/installPackage.js
davidnagli: Pretty much the same edge case that I mentioned before applies. If you’re starting a new project, you’d want to know that you need to create a `yarn.lock` / `--yarn` and that Parcel isn’t using yarn.

I see your point though... if you want, I can remove it

src/Bundler.js
davidnagli: Done

davidnagli:
Done...

test/autoinstall.js
```diff
@@ -0,0 +1,56 @@
...
+
+    let expectedModulePath = inputDirPath + '/node_modules/' + pkgName;
+    assert(fs.existsSync(expectedModulePath), 'lodash is in node_modules');
+
+    let pkg = fs.readFileSync(inputDirPath + '/package.json');
+    pkg = JSON.parse(pkg);
+    assert(pkg.devDependencies[pkgName], 'lodash is saved as a dev dep');
+  });
+
+  it('should install lodash using yarn and save dev dependency to package.json', async function() {
```
DeMoorJasper: Perhaps also add a test that tests the autoinstall from require.
For example just have a script like this that gets bundled and check if it's installed after test ran.
```
const _ = require('lodash');
// Do something with lodash to check later...
```

davidnagli: 👍

src/cli.js
devongovett: Going to remove it for now. I think our approach is pretty solid. If someone wants to open a bug with a good reason to add an option for it, they can.

src/Bundler.js
```diff
@@ -325,38 +327,70 @@ class Bundler extends EventEmitter {
...
+  async resolveDep(asset, dep, install = true) {
     try {
       return await this.resolveAsset(dep.name, asset.name);
     } catch (err) {
       let thrown = err;
 
       if (thrown.message.indexOf(`Cannot find module '${dep.name}'`) === 0) {
+        // Check if dependency is a local file
+        let isLocalFile = /^[/~.]/.test(dep.name);
+        let fromNodeModules = asset.name.includes(
```
devongovett: Added a check here to make sure we don't auto install dependencies of things inside node_modules, only application code. If a dependency is missing from within a node_modules folder, that is a bug in the package and so we should error.

src/Bundler.js
```diff
@@ -325,38 +327,70 @@ class Bundler extends EventEmitter {
...
-        thrown.fileName = asset.name;
+        await this.throwDepError(asset, dep, thrown);
       }
+
       throw thrown;
     }
   }
 
+  async installDep(asset, dep) {
+    let [moduleName] = this.resolver.getModuleParts(dep.name);
```
devongovett: Used this function to get the module name to install. If you required e.g. `lodash/cloneDeep` in your app, we want to install `lodash`, not the entire dep name. The exception is scoped packages, e.g. `@scope/module`. In that case, we pass the whole thing. the `getModuleParts` function from the resolver handles both of these cases.

src/utils/installPackage.js
```diff
@@ -1,54 +1,55 @@
...
+  }
+
+  // npm doesn't auto-create a package.json when installing,
+  // so create an empty one if needed.
+  if (packageManager === 'npm' && !packageLocation) {
+    await fs.writeFile(path.join(cwd, 'package.json'), '{}');
+  }
+
+  try {
+    await pipeSpawn(packageManager, args, {cwd});
```
devongovett: Switched to using the `pipeSpawn` utility for this.

src/utils/installPackage.js
```diff
@@ -1,54 +1,55 @@
...
+  let args = [commandToUse, ...modules];
+  if (saveDev) {
+    args.push('-D');
+  } else if (packageManager === 'npm') {
+    args.push('--save');
+  }
+
+  // npm doesn't auto-create a package.json when installing,
+  // so create an empty one if needed.
+  if (packageManager === 'npm' && !packageLocation) {
```
devongovett: Added this to make sure we create a package.json when needed.

src/utils/installPackage.js
```diff
@@ -59,8 +60,47 @@ async function installPeerDependencies(dir, name) {
...
+  if (!hasYarn || configName === 'package-lock.json') {
+    return 'npm';
+  }
+
+  return 'yarn';
+}
+
+let hasYarn = null;
+async function checkForYarnCommand() {
+  if (hasYarn != null) {
```
devongovett: Caching if we have yarn or not

src/utils/installPackage.js
```diff
@@ -59,8 +60,47 @@ async function installPeerDependencies(dir, name) {
...
+    'yarn.lock',
+    'package-lock.json'
+  ]);
+  let hasYarn = await checkForYarnCommand();
+
+  // If Yarn isn't available, or there is a package-lock.json file, use npm.
+  let configName = configFile && path.basename(configFile);
+  if (!hasYarn || configName === 'package-lock.json') {
+    return 'npm';
+  }
```
devongovett: Changed this logic slightly. Now we default to `yarn` if it is installed, but use `npm` if a `package-lock.json` file is found and `yarn.lock` is not, or `yarn` is not installed.

src/utils/installPackage.js
```diff
@@ -59,8 +60,47 @@ async function installPeerDependencies(dir, name) {
...
+  try {
+    hasYarn = await commandExists('yarn');
+  } catch (err) {
+    hasYarn = false;
+  }
+
+  return hasYarn;
+}
+
+let queue = new PromiseQueue(install, {maxConcurrent: 1, retry: false});
```
devongovett: Added a queue for install calls so that we can never have more than one install happening at the same time. Otherwise, npm will fail trying to write over the same files simultaneously.

src/utils/pipeSpawn.js
```diff
@@ -1,16 +1,26 @@
...
+  const cp = spawn(cmd, params, Object.assign({
+    env: Object.assign({
+      FORCE_COLOR: logger.color,
+      npm_config_color: logger.color ? 'always': '',
+      npm_config_progress: true
+    }, process.env)
+  }, opts));
+
+  cp.stdout.setEncoding('utf8').on('data', d => logger.writeRaw(d));
+  cp.stderr.setEncoding('utf8').on('data', d => logger.writeRaw(d));
```
devongovett: `pipeSpawn` now uses the logger to write its output rather than directly piping to stdout. This creates a better experience when installing packages: when the install is complete, the output from yarn or npm is cleared since it is no longer relevant and the build continues. When an error occurs, it stays up to allow debugging.

devongovett:
Thanks for your work on this! Really excited about it. 🚀

test/autoinstall.js
devongovett: This would be good to do in a followup PR. Maybe we can cover a few more of the different cases as well.

Issue #814, opened 2018-02-15
🐛 on page reload: message.split is not a function


mischnic:
🐛 bug report

### 😯 Current Behavior

First load of page works. But when refreshing the page:

```
yarn run v1.3.2
$ standard && parcel src/index.html -p 3000 --open
Server running at http://localhost:3000
✨  Built in 32.58s.
/Users/niklas/Sites/hyperapp-todo-parcel/node_modules/parcel-bundler/src/Logger.js:27
      this.lines += message.split('\n').length;
                            ^

TypeError: message.split is not a function
    at Logger.write (/Users/niklas/Sites/hyperapp-todo-parcel/node_modules/parcel-bundler/src/Logger.js:27:29)
    at Logger.log (/Users/niklas/Sites/hyperapp-todo-parcel/node_modules/parcel-bundler/src/Logger.js:38:10)
    at WebSocket.handleSocketError (/Users/niklas/Sites/hyperapp-todo-parcel/node_modules/parcel-bundler/src/HMRServer.js:88:12)
    at WebSocket.onError (/Users/niklas/Sites/hyperapp-todo-parcel/node_modules/ws/lib/event-target.js:128:16)
    at WebSocket.emit (events.js:160:13)
    at WebSocket.finalize (/Users/niklas/Sites/hyperapp-todo-parcel/node_modules/ws/lib/websocket.js:190:41)
    at Socket.emit (events.js:165:20)
    at emitErrorNT (internal/streams/destroy.js:64:8)
    at process._tickCallback (internal/process/next_tick.js:152:19)
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```
### 💁 Possible Solution

Looks like a error with HMR, because it works with `--no-hmr`.

Maybe because of version bumping? 

### 💻 Code Sample

Repo: https://github.com/marcusasplund/hyperapp-todo-parcel

```
yarn
yarn upgrade parcel-bundler
yarn start
```
@marcusasplund

### 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Parcel           | 1.6.0
| Node             | 9.5.0
| npm/Yarn         |1.3.2
| Operating System |macOS

FalkoJoseph:
Got the same error suddenly.

ryan-edge-ck:
I am getting the same error on a React project.

MarcusAhlfors:
+1

[...]

PR #815, opened 2018-02-15
Update HMRServer handleSocketError for ErrorEvent

brianchung808:
It looks like `ws` changed the error passed to `onerror` to be an `ErrorEvent` (https://github.com/websockets/ws/pull/1261) in 4.0, which doesn't have `code` at the top-level.

The guard for ECONNRESET was falling through when closing the tab in chrome and throwing an error at the log line since err isn't a string which ends up killing the server.

### DIFF
--- a/src/HMRServer.js
+++ b/src/HMRServer.js
@@ -81,11 +81,11 @@ class HMRServer {
   }
 
   handleSocketError(err) {
-    if (err.code === 'ECONNRESET') {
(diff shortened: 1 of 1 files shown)


brianchung808:
This addresses #814

devongovett:
Released to npm in v1.6.1.

brandon93s:
Whoops! Thanks @brianchung808!

brianchung808:
Happy to help!

PR #820, opened 2018-02-15
Changed function to arrow function

thevoidf:
Just looks nicer


devongovett:
Sorry, I prefer normal top-level functions.

PR #821, opened 2018-02-16
add JSX support for "Nerv"

g-plane:
Ref: https://github.com/NervJS/nerv


g-plane:
Sorry for missing it. I will add it later.

PR #822, opened 2018-02-16
Add server reference to bundler

fusepilot:
I need to get the development server's port number for use in a plugin. Attaching the server to the bundler allows me to call `bundler.server.address().port`. This seems to be the only way to access it since the server's port can vary if the user provided port is already being used.

### DIFF
--- a/src/Bundler.js
+++ b/src/Bundler.js
@@ -574,10 +574,9 @@ class Bundler extends EventEmitter {
   }
 
   async serve(port = 1234, https = false) {
-    let server = await Server.serve(this, port, https);
-    this.server = server;
+    this.server = await Server.serve(this, port, https);
     this.bundle();
-    return server;
+    return this.server;
   }
 }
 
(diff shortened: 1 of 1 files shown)


brandon93s:
Actually, `.serve()` returns the server. Is that not sufficient?

fusepilot:
Wouldn't `.serve()` spin up a new server instance? I tried it and the build errors with
````
loadedAssets is not iterable
  at Bundler.bundle (.../parcel-bundler/src/Bundler.js:179:61)
````

**Comment by brandon93s** [comment]
> for use in a plugin

Missed this part. Expecting the CLI to continue to call `serve` and having the server be available to your plugin by keeping a reference to it. 👍

File: src/Bundler.js [hunk 1]
```diff
@@ -575,6 +575,7 @@ class Bundler extends EventEmitter {
 
   async serve(port = 1234, https = false) {
     let server = await Server.serve(this, port, https);
+    this.server = server;
     this.bundle();
     return server;
```
lbguilherme: Do:

```js
     this.server = await Server.serve(this, port, https);
     this.bundle();
     return this.server;
```

DeMoorJasper:
@fusepilot it's a rust test that fails from time to time, no need to worry about it

PR #825, opened 2018-02-16
Add JSX support for Hyperapp.

JorgeBucaran:
Hi @devongovett! This PR adds [Hyperapp](https://github.com/hyperapp/hyperapp) support to Parcel!

PR #828, opened 2018-02-16
Hyperapp support

Art2B:
I've added hyperapp support out of the box with the use of `h` jsx pragma.


devongovett:
This was already done in https://github.com/parcel-bundler/parcel/pull/825.

PR #829, opened 2018-02-16
WIP: Content-hash bundle names

### LINKED ISSUE #188: File hash does not change after its content updates
This is a 🙋 feature request.

### 🤔 Expected Behavior
An output file hash should change when its content updates.

### 😯 Current Behavior
File hash does not change.

### 🔦 Context
```
<!-- index.html -->
<html>
<body>
  <script src="./main.js"></script>
</body>
</html>
```
```js
// main.js
console.log(1);
```

When build, 'b695675d84099f097ec37d68c8c83fce.js' generates.

```bash
parcel build --no-cache --no-minify index.html
```

And then, change main.js

```js
// main.js
console.log(2);
```

Build again.

```bash
parcel build --no-cache --no-minify index.html
```

The javascript file name is still 'b695675d84099f097ec37d68c8c83fce.js'.
I am not sure it is the expected behavior or not. However, when I using webpack, the output file hash will change every time its content updates.

### 🌍 Your Environment

| Parcel           |  v1.1.0
| Node             | v8.9.1
| npm/Yarn         | yarn v1.3.2
| Operating System | macOS High Sierra 10.13.1

### LINKED ISSUE #717: 🙋 Use MD5 of File Content As Name To Bust Caches
This is something my team is interested in implementing, but want to make sure we're not heading in the wrong direction...

### 🤔 Expected Behavior

In order to serve assets through a CDN, we need to use a hash of the file contents as part of the filename. This is the only way to ensure that cache busts at the right time and only at the right time. 

 - `mtime` changes based on developer machine / git checkout
 - etags don't work with CDN
 - query string approaches don't work with all proxies
 - the current approach never cache busts on content change
 - some sort of random string (current time, or build time, or something) would bust too often
 - manual versioning of the entire build (ie, output to /dist/v2) is annoying and busts too often
 - manual versioning of individual files (ie, app-v2.js) is really annoying

This issue has been discussed extensively here: https://github.com/parcel-bundler/parcel/issues/188.
We have discussed it a bit on slack start here: https://parcel-bundler.slack.com/archives/C8QC15PNY/p1515609163000373

To be specific, I am proposing files be named `${contentHash}.${ext}}`

### 😯 Current Behavior

Currently, the filename is a hash based on the _name_ of the file (and some options like whether or not hmr is enabled, I believe). Something like `${fileNameHash}.${ext}}`

### 💁 Possible Solution

@shawwn proposes parts of the solution here: https://github.com/parcel-bundler/parcel/issues/188#issuecomment-353141569

@Munter describes how tree traversal needs to work to make the cache bust in all the right places here: https://github.com/parcel-bundler/parcel/issues/188#issuecomment-357072153

One key piece of the puzzle will be to only enable this behavior on `build` (ie, when building for production). Leave it off in the development modes, so you don't have to worry about 

### 🔦 Context

We want to serve our assets (css, js, images) through a CDN like CloudFlare or CloudFront. If something about the url doesn't change, then new versions of those assets will never make it to the end user. If too much changes, then we bust the cache too aggressively.

benhutton:
Still a long way to go to get this ready for review. And the javascript you see here is that of a Rubyist stumbling into javascript and its asynchronicity. We'll get this cleaned up in the next few days and try to write a few tests and submit a proper PR template.

Fixes #717 
Fixes #188 

/cc @shanebo

### DIFF
--- a/src/Asset.js
+++ b/src/Asset.js
@@ -200,7 +200,7 @@ class Asset {
       const packageName = sanitizeFilename(this.package.name, {
         replacement: '-'
       });
-      return packageName + ext;
+      return md5(packageName) + ext;
     }
 
     // If this is the entry point of the root bundle, use the original filename
--- a/src/utils/packageHasher.js
+++ b/src/utils/packageHasher.js
@@ -1,0 +1,55 @@
+const logger = require('../Logger');
+const fs = require('fs');
+const md5File = require('md5-file');
+const walkedFiles = new Map();
+const hashMap = new Map();
+const Path = require('path');
+
+async function packageHasher(mainBundle) {
+  walkBundles(mainBundle, '--');
+}
+
+module.exports = packageHasher;
+
+async function walkBundles(bundle, indent) {
+  for (let child of bundle.childBundles) {
+    if (!walkedFiles.has(child.name)) {
+      walkBundles(child, indent + '--');
+    }
+  }
+
+  renameReferences(bundle.name, hashMap);
+
+  let oldHash = Path.basename(bundle.name, Path.extname(bundle.name));
+  let hash = md5File.sync(bundle.name);
+  walkedFiles.set(bundle.name, hash);
+  hashMap.set(oldHash, hash);
+  // logger.log(indent + bundle.name)
+  logger.log(indent + bundle.entryAsset.name);
+
+  let newLocation = `${bundle.entryAsset.options.outDir}/${hash}.${
+    bundle.entryAsset.type
+  }`;
+  // logger.log(indent + newLocation + " (move here)");
+
+  // If it's not the entry file, rename it
+  if (bundle.parentBundle) {
+    fs.renameSync(bundle.name, newLocation);
+  }
+}
+
+async function renameReferences(fileName, hashMap) {
+  if (fs.existsSync(fileName)) {
+    let data = fs.readFileSync(fileName, 'utf8');
+
+    hashMap.forEach(function(value, key) {
+      data = data.replace(new RegExp(key, 'g'), value);
+    });
+
+    fs.writeFileSync(fileName, data, 'utf8', function(err) {
+      if (err) return console.log(err);
+    });
+  } else {
+    console.log(fileName + ' not found. Must have already been moved!');
+  }
+}
--- a/package.json
+++ b/package.json
@@ -37,6 +37,7 @@
     "is-url": "^1.2.2",
     "js-yaml": "^3.10.0",
     "json5": "^0.5.1",
+    "md5-file": "^3.2.3",
     "micromatch": "^3.0.4",
     "mkdirp": "^0.5.1",
     "node-forge": "^0.7.1",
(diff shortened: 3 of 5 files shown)


src/utils/packageHasher.js
```diff
@@ -0,0 +1,29 @@
...
+
+module.exports = packageHasher;
+
+async function walkBundles(bundle, indent) {
+  for (let child of bundle.childBundles) {
+    if ( !walkedFiles.has(child.name)) {
+      walkBundles(child, indent + '--')
+    }
+  }
+  hash = md5File.sync(bundle.name);
```
benhutton: Above this line, we need to do a find and replace on the current file (`bundle.name`) with the contents of `walkedFiles` (which has a map from old hashes to new hashes).

src/utils/packageHasher.js
```diff
@@ -0,0 +1,52 @@
...
+    fs.renameSync(bundle.name, newLocation);
+  }
+}
+
+async function renameReferences(fileName, hashMap) {
+  if (fs.existsSync(fileName)) {
+    data = fs.readFileSync(fileName, 'utf8')
+
+    hashMap.forEach(function(value, key) {
+      data = data.replace(new RegExp(key, 'g'), value);
```
devongovett: This seems somewhat dangerous and would probably end up replacing things we don't want. For example, if you had a file called `index.js` but some string in the app also contained that but wasn't a reference, we'd end up replacing it too.

Instead of doing it this way, maybe we can store a list of reference locations as part of the asset (e.g. part of the options to `addDependency`)?

benhutton: @devongovett in my limited experience, `bundle.name` has only ever been my entry file (`entry.html` in my case) or an md5 hash. What are the other possibilities?

benhutton: @devongovett in my testing, `hashMap` only contains md5 hashes of things. Are there other non-md5 names that can get assigned to a bundle?

benhutton: @devongovett okay....

If a name isn't hashed in `generateBundleName()`, would we want this new code to hash it? What's the reason for leaving it as-is there and not just hashing everything?

devongovett: My point is mostly that doing a search and replace over output files for filenames is somewhat dangerous in that we might end up replacing things we didn't intend. Perhaps `generateBundleName` could generate something random always which is much less likely to appear in the source code, and we could use that to replace with the final names here.

benhutton: @devongovett yeah, if it's safe for `generateBundleName` to _always_ use hashing, that would be nice.

src/Asset.js
```diff
@@ -200,7 +200,7 @@ class Asset {
       const packageName = sanitizeFilename(this.package.name, {
         replacement: '-'
       });
-      return packageName + ext;
+      return md5(packageName) + ext;
```
benhutton: @devongovett is all I need to do in this function? My intent is to not do any renaming on the entry file: https://github.com/parcel-bundler/parcel/pull/829/files#diff-784c5f1cd4c785db2834e618a52b438eR35

devongovett:
Closing in favor of #1025.

PR #831, opened 2018-02-16
Add GLSL assets support

mvlabat:
Greetings! Thanks for the awesome instrument :)

This PR adds support for GLSL assets. I assume people using WebGL may find it useful. 
Requiring GLSL asset will return its source. The following extensions are supported:
- *.glsl
- *.vert
- *.frag

This feature doesn't give much. For instance, it can't provide checking that shaders are valid and can be compiled. But the most important thing here is watching source files and rebuilding them automatically.


lbguilherme:
As there is no way to compile or verify the source without getting a WebGL context from the user, perhaps there is no need for these assets to be importable assets, as you can already import a source file with `fs`. Not against merging this, though.

(The only way I see of doing it is if the module would export a function that takes the context and generates the compiled program already. But this is too much for little gain).

Note: The same feature can already be accomplished by `require("fs").readFileSync(__dirname + "/shader.glsl")`.

mvlabat:
My bad, didn't know about this: `require("fs").readFileSync(__dirname + "/shader.glsl")`. :D It appeared far more interesting to dig into Parcel codebase and implement a useless feature, than to read just one page of Parcel documentation. :)

> As there is no way to compile or verify the source without getting a WebGL context from the user

Well, we could at least verify it, if using, for instance, something like this: https://github.com/stackgl/headless-gl. But it doesn't support WebGL2 at the moment. And for some reason this package doesn't work for windows (fails on `yarn install`). So maybe this could be something for future.

> The only way I see of doing it is if the module would export a function that takes the context and generates the compiled program already. But this is too much for little gain

Yeah, indeed, I think it's not Parcel's task to provide functions for compiling shaders.

But if anyone comes up with and idea what this asset loader could do, I'll try to implement it.
Or maybe we should close this PR for now.

devongovett:
Cool, thanks for working on this! What do you think of doing something like [glslify](https://github.com/glslify/glslify) in Parcel so you could import other GLSL modules from within a shader?

mvlabat:
@devongovett, that's a good idea indeed, thanks! I'll try to do it. Marking as `WIP`. :)

mvlabat:
Seems to be ready now. @devongovett, would be glad for review :)

src/assets/GLSLAsset.js
```diff
@@ -0,0 +1,62 @@
...
+
+    const basedir = path.dirname(this.name);
+    const glsl = glslify(this.contents, {basedir});
+
+    return new GLSLAst(glsl, {basedir});
+  }
+
+  async collectDependencies() {
+    const glslifyDeps = require('glslify-deps');
+    const depper = glslifyDeps({cwd: this.ast.basedir});
```
devongovett: Won't this cause us to read and parse the files twice: once above as part of `glslify` transform, and once again here? I wonder if we can only do it once.

src/assets/GLSLAsset.js
```diff
@@ -0,0 +1,62 @@
...
+const path = require('path');
+
+class GLSLAsset extends Asset {
+  constructor(name, pkg, options) {
+    super(name, pkg, options);
+    this.type = 'js';
+  }
+
+  async parse() {
+    await localRequire('glsl-token-inject-block', this.name);
```
devongovett: Why is this being required but not used?

mvlabat: For some reason `yarn install` doesn't fetch this dependency (it's a dependency of `glsl-inject-defines`), but `npm install` does. As soon as we merge https://github.com/parcel-bundler/parcel/pull/805, it won't be needed. So this is a temporary workaround.

src/assets/GLSLAsset.js
mvlabat: Yeah, unfortunately it will. I tried to find the way to do it only once, but `glslify` doesn't export the needed functions in order to split it up conveniently. Maybe we could accomplish this with copy-pasting some `glslify` code, but I think it's not desirable.
Anyway, I believe that scanning for dependencies twice won't be a huge overhead, since it doesn't really compile the files, but just traverses them with regexps.

src/assets/GLSLAsset.js
davidnagli: @mvlabat Can you elaborate on why `yarn install` doesn’t work but `npm install` does. Not really sure if we should leave it broken like that, and not really sure if #805 is gonna solve anything anyways...

mvlabat: @devongovett, @davidnagli, after rebasing with the latest master I couldn't reproduce the bug. So I've just removed that `localRequire`

mvlabat:
After updating `glslify` to `6.1.1` the issue with resolving its dependencies was resolved. Now I think it's ready to be reviewed again )
@devongovett, @davidnagli

src/assets/GLSLAsset.js
mvlabat: The actual compile process, that involves concatenating GLSL files and other stuff, will be performed only once

devongovett:
Updated to just use `glslify-deps` and `glslify-bundle` directly rather than the `glslify` package. This results in doing a single parse of the dependencies rather than two. I also updated it to use the Parcel resolver, which enables features like aliases and tilde paths to be used just like when importing JS, etc.

mvlabat:
@devongovett, thank you for your time improving this and merging. That's awesome!

PR #838, opened 2018-02-17
Server index improvements

### LINKED ISSUE #714: 🐛 --public-url is setted on build and watch
When I work with `parcel index.html` the resulting files have public-url set to `./`, when I use build or watch the public url is no longer `./` but `./build`.

## Reproducing the error:

https://github.com/lrn2prgrm/parcel-bug/tree/master

{
  "name": "cli",
  "version": "1.0.0",
  "license": "MIT",
  "scripts": {
    "dev-serve": "parcel src/index.html -d build --no-cache", 
    "watch": "parcel watch src/index.html -d build --no-cache", 
    "build": "parcel build src/index.html -d build --no-minify --no-cache",
    "start": "yarn build"
  },
  "devDependencies": {
    "parcel-bundler": "^1.5.1",
    "typescript": "^2.6.2"
  }
}

### 🤔 Expected Behavior

watch and build produced files should be generated as if public url is `./`

either with
* "prestart": "yarn build",
* "prestart": "yarn watch",

*dev-serve behaves correctly*

### 😯 Current Behavior
watch and build produced files should be generated as if public url is `./build`

either with
* "prestart": "yarn build",
* "prestart": "yarn watch",

*dev-serve behaves correctly*

### 💁 Possible Solution
No idea about implementation but the code that adds this public url should be found and modified

### 🌎 Environment
| Software   | Version(s) 
| Parcel        | 1.5.1
| Node         | 8.9.2
| Yarn           | 1.2.1
| OS             | Windows 10 Pro

#### Found as I closed this issue thanks to niicojs: https://github.com/parcel-bundler/parcel/issues/706

### LINKED ISSUE #782: 🐛 publicURL with development server
`parcel src/index.html --public-url ./`
or 
`parcel src/index.html --public-url /`

### 🤔 Expected Behavior

The site should be served at `localhost:1234`.

### 😯 Current Behavior

- with `parcel src/index.html --public-url ./`:
	- `http://localhost:1234/` properly serves `index.html`
	- every request serves index.html (i.e. `http://localhost:1234/423958b25b168623f5a455b67f81b487.js` returns the `index.html` file)

- with `parcel src/index.html --public-url /`:

	- http://localhost:1234/ returns a 404
	- everything works when navigating to `http://localhost:1234/index.html`

### 🌍 Your Environment

| Parcel           |1.5.1
| Node             |v9.5.0
| npm/Yarn         |yarn 1.3.2
| Operating System |macOS 10.12.6

DeMoorJasper:
* Closes #782 
* closes #714 
* Closes #974 
* Changes default publicURL to / which is probably a better default
* Changes index bahaviour to be more inline with a standard webserver only showing index if the request url is publicURL or below publicURL (e.g. `publicURL = /this/is/a/directory` than it will give index on any path up to `/this/is/a/directory/` once you go above that for example `/this/is/a/directory/index.js` it'll return an asset or 404 depending on it's existance)


src/Bundler.js
```diff
@@ -66,7 +66,7 @@ class Bundler extends EventEmitter {
     const publicURL =
       options.publicUrl ||
       options.publicURL ||
-      '/' + Path.basename(options.outDir || 'dist');
+      '/' + Path.basename(options.outDir || '');
```
mightyiam: Why set it to `options.outDir`? What does that provide? Why are these two related? I notice that this is the part that is not changed. I think we should change it, as I described in #932. To simply `/` by default.

src/Server.js
```diff
@@ -43,8 +43,11 @@ function middleware(bundler) {
     function respond() {
       if (bundler.errored) {
         return send500();
-      } else if (!req.url.startsWith(bundler.options.publicURL)) {
-        // If the URL doesn't start with the public path, send the main HTML bundle
+      } else if (
+        !req.url.startsWith(bundler.options.publicURL) ||
+        req.url === bundler.options.publicURL
+      ) {
+        // If the URL doesn't start with the public path or is the public path, send the main HTML bundle
```
mightyiam:
```
function respond() {
      if (bundler.errored) {
        return send500();
      } else if (!req.url.startsWith(bundler.options.publicURL)) {
        return sendIndex();
      } else if (req.url === bundler.options.publicURL) {
        return sendIndex();
      } else {
        // Otherwise, serve the file from the dist folder
        req.url = req.url.slice(bundler.options.publicURL.length);
        return serve(req, res, send404);
      }
    }
```

DeMoorJasper: I don’t think this is a better solution as it just adds an extra unnecessary if

src/Bundler.js
DeMoorJasper: I didn’t want to change it too much, just fix a bug and prevent future issues regarding the default. Although i do agree with you, would be nice to have Devon’s opinion on these changes

src/Server.js
mightyiam: More readable for me.

davidnagli: I do agree with @mightyiam that the multiline `||` is pretty hard to read / reason about, but I also agree with @DeMoorJasper that the split `if`s looks redundant and unnecessary since both `if`s do exactly the same thing.

I would write it like this, which solves both of those problems, and get's rid of some of the confusing negation, making it much easier to understand and reason about:

```javascript
function respond() {
  // If an error occurred in the bundler, send a 500 error
  if (bundler.errored) return send500();

  // If the URL start with the public path, serve the file from the dist folder
  if (req.url.startsWith(bundler.options.publicURL)) {
    // Check that its not the publicURL itself
    if (req.url !== bundler.options.publicURL) {
      req.url = req.url.slice(bundler.options.publicURL.length);
      return serve(req, res, send404);
    }
  }

  // Otherwise send the main HTML bundle
  return sendIndex();
}
```

<details>
<summary>Explanation</summary>
This converts the "or" into a negated "and" ([using this fancy Boolean algebra rule](https://en.wikipedia.org/wiki/De_Morgan%27s_laws)), then flips the resulting if-else to clean up the negation, and then splits the resulting `&&` into separate nested `if`s (thank you Webstorm for doing all that in the click of a button 😛). Oh, and I also got rid of all the `else`s, since they're redundant with the `return`s
</details>
&nbsp;

IMO, my version is _slightly_ cleaner and easier to read / reason about, but tbh it doesn't really matter  so I'll leave it up to you.

davidnagli: I do agree with @mightyiam that the multiline `||` is pretty hard to read / reason about, but I also agree with @DeMoorJasper that the split `if`s looks redundant and unnecessary since both `if`s do exactly the same thing.

I would write it like this, which solves both of those problems, and get's rid of some of the confusing negation, making it much easier to understand and reason about:

```javascript
function respond() {
  // If an error occurred in the bundler, send a 500 error
  if (bundler.errored) return send500();

  // If the URL start with the public path, serve the file from the dist folder
  if (req.url.startsWith(bundler.options.publicURL)) {
    // Check that its not the publicURL itself
    if (req.url !== bundler.options.publicURL) {
      req.url = req.url.slice(bundler.options.publicURL.length);
      return serve(req, res, send404);
    }
  }

  // Otherwise send the main HTML bundle
  return sendIndex();
}
```

<details>
<summary>Explanation</summary>
This converts the "or" into a negated "and" ([using this fancy Boolean algebra rule](https://en.wikipedia.org/wiki/De_Morgan%27s_laws)), then flips the resulting if-else to clean up the negation, and then splits the resulting `&&` into separate nested `if`s (thank you Webstorm for doing all that in the click of a button 😛). Oh, and I also got rid of all the `else`s, since they're redundant with the `return`s
</details>
&nbsp;

IMO, my version is _slightly_ cleaner and easier to read / reason about, but tbh it doesn't really matter  so I'll leave it up to you.

DeMoorJasper: I tried improving readability, also changed how index works slightly (so it's more like how other webserver work, only giving index if it's in publicURL or a sub-directory of publicURL. Would also prevent issues with the future naming conventions inside parcel and any issue with parcel not giving an asset but the index page

devongovett:
This one seems to break going to e.g. `/foo` when `--public-url=/`. In this case, the index HTML file should be served. This is so SPAs that do client side routing can work correctly. Going to merge #974 instead which solves the same issue, but still allows virtual URLs to work correctly. Since Parcel never generates extensionless files, I think this is safe.

devongovett:
Also changing the default public url to `/` in #1040.

PR #841, opened 2018-02-17
Use babel-register for tests only with node < 8

ronami:
Hey,

#789 added `babel-register` before running `mocha` tests. Using `babel-register` slows down tests, adds flakiness and it's redundant for Node v8.x.x.

This PR adds `babel-regsiter` only for when Node's version is <8.

### DIFF
--- a/test/babel-register.js
+++ b/test/babel-register.js
@@ -1,0 +1,3 @@
+if (process.version.match(/^v(\d+)\.\d+\.\d+$/)[1] < 8) {
+  require('babel-register');
+}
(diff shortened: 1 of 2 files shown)


test/babel-register.js
```diff
@@ -0,0 +1,3 @@
+if (process.version.match(/^v(\d+)\.\d+\.\d+$/)[1] < 8) {
```
lbguilherme: Better: `parseInt(process.versions.node, 10) < 8`

ronami: Nice, thanks 👍

Issue #316, opened 2017-12-17
Url() without quotes is not supported in sass


StarpTech:
According to https://developer.mozilla.org/en-US/docs/Web/CSS/url there are three formats. 
```sass
div#test {
  background-image: url(./images/test.jpg)
}
```
```js
{
  "name": "parcel-demo",
  "version": "1.0.0",
  "description": "",
  "main": "index.js",
  "scripts": {
    "start": "parcel index.html"
  },
  "author": "Dustin Deus",
  "license": "ISC",
  "devDependencies": {
    "autoprefixer": "^7.2.3",
    "babel-preset-env": "^1.6.1",
    "node-sass": "^4.7.2",
    "postcss-modules": "^1.1.0",
    "uglify-es": "^3.2.2"
  }
}
```

## Exptected behaviour
`./images/test.jpg` should be in `dist` folder
## Current behaviour
`./images/test.jpg` is not in `dist` folder

Parcel: 1.2.0
Windows 10

josephch405:
It seems like the image is actually there - just renamed and put at the root directory. If I'm not mistaken, a key part of using a bundler in the first place is that assets such as images would be processed like so - so instead of having a `./images/test.jpg`, the bundler replaces references to it with `7286d30e8328e17dbf8fa60c2c4b1ffb.jpg` (as an example). Not sure if actually an issue.

davidnagli:
Looks like it works correctly according to @josephch405. Feel free to reopen this if you thinks it’s still an issue :)

StarpTech:
The issue is not about naming the image is missing completely. The image is only there when I use quotes.

StarpTech:
Please reopen.

davidnagli:
@StarpTech Oh I see, so you’re saying that the image doesn’t appear in the `dist` directory at all?

StarpTech:
Yes!

brandon93s:
This is related to _sass_ specifically. This works as expected when using vanilla _css_, but when converting to `sass` and adding `node-sass` the issue is reproduced.

brandon93s:
This is occurring because the `url` function isn't executing when files do not have quotes, and therefore are not being registered as dependencies:

https://github.com/parcel-bundler/parcel/blob/c008bb0ac492dbac38e2f13017e3143f40359934/src/assets/SASSAsset.js#L26-L31

[...]

PR #850, opened 2018-02-19
New faster resolver supporting absolute and tilde paths, and aliases

### LINKED ISSUE #25: Add an aliasing system that works for all asset types
Is there a way to define alias ? Like webpack config:

### LINKED ISSUE #86: 🐛 link to static html could'not be resolved, win10, parcel 1.0.3
index.html
```html
            <p>
                <a href="/src/somePathToChecker/index.html">Check</a>
            </p>
```

```
 Cannot resolve dependency './..\..\..\..\..\src\somePathToChecker\index.html'
    at C:\Users\Aleksandr_Kanunnikov\AppData\Roaming\npm\node_modules\parcel-bundler\node_modules\browser-resolve\node_modules\resolve\lib\async.js:55:21
    at load (C:\Users\Aleksandr_Kanunnikov\AppData\Roaming\npm\node_modules\parcel-bundler\node_modules\browser-resolve\node_modules\resolve\lib\async.js:69:43)
    at onex (C:\Users\Aleksandr_Kanunnikov\AppData\Roaming\npm\node_modules\parcel-bundler\node_modules\browser-resolve\node_modules\resolve\lib\async.js:92:31)
    at C:\Users\Aleksandr_Kanunnikov\AppData\Roaming\npm\node_modules\parcel-bundler\node_modules\browser-resolve\node_modules\resolve\lib\async.js:22:47


```

### LINKED ISSUE #336: 🙋 Support project-relative import paths

### LINKED ISSUE #832: 🙋 Aliasing modules -- react -> preact-compat etc.
It is impossible to use parcel with preact-compat because it requires being able to alias all references to 'react' and 'react-dom' to 'preact-compat'. It is possible to do this for local package using `browser` in `package.json`, but does not work when including external React components.

### 🤔 Expected Behavior

Add an alias field in, say, `package.json`, like

```
"alias": {
    "^react$": "preact-compat",
    "react-dom$": "preact-compat",
    "react-dom/server": "preact-compat/server"
}
```

Which would then replace all references to `react` with `preact-compat` etc.

### 😯 Current Behavior

Using `browser` field, I can replace local import/requires successfully, but when importing, say `react-apollo`, it doesn't replace there.

### 🔦 Context

I know there was an issue #25 for this but it got locked and I haven't seen any further discussion about it, but this is quite inconvenient for people trying to use preact. I think it was a bad decision to completely delete all discussion about current workarounds as now people have no reference on what to do until this gets implemented natively.

### Workaround for now

The workaround I found is to make symbolic links in the `node_modules` folder so that `node_modules/react` and `node_modules/react-dom` both link to `node_modules/preact-compat`.

devongovett:
This replaces the `browser-resolve` and `resolve` modules with our own custom resolver which is both faster and more extendible. It especially speeds up building from cache: on one project my build time went from 4.93s to 1.40s. 

This performance increase is due to severely fewer `fs.stat` calls compared to the `resolve` implementation: for example, resolving one dependency I had in my project caused 258 `fs.stat` calls in `resolve` whereas it takes just 9 in the new resolver.

In addition to performance, there is also some new features:

1. Absolute paths (e.g. /foo) resolved relative to the project root (even from within node_modules - is this a bad idea?). Closes #336, closes #290, closes #86.
2. Tilde paths (e.g. ~/foo) resolved relative to the nearest package root in node_modules, or the project root if not in node_modules. Closes #336, closes #466.
3. Aliases are supported using a `package.alias` field in package.json. It works identically to the `browser` field, but when found in the project root package.json it applies globally rather than to only the local module. The alias field also works within node_modules, but does not apply globally. Closes #25, closes #832.

As an example, you could declare the following in your project's package.json to alias react to preact globally:

```json
{
  "alias": {
    "react": "preact-compat",
    "react-dom": "preact-compat"
  }
}
```

This can also be used to alias modules that don't exist in node_modules at all:

```json
{
  "alias": {
    "fake-module": "./some/deep/path"
  }
}
```

Now you can `require('fake-module')` and it will point to the path you've specified.


JounQin:
What about alias modules under specific directory? Or it can be alias one by one?

Like: https://webpack.js.org/configuration/resolve/#resolve-modules

devongovett:
@JounQin yeah, I don't think we are going to support something like that (similar to `NODE_PATH` environment variable). I think it's a bad idea to do this it makes it much harder to reason about where the code you read in a project is coming from. At least when you explicitly list the modules you want to alias in package.json there is one place to look for to see the mapping.

Slapbox:
Does this support mapping of named exports? For example webpack supports this syntax:

`alias: ['nodeModuleName, exportName]


`ipcRenderer: ['electron', 'ipcRenderer']`

devongovett:
This is a mapping of files only. If you wanted to do something like that I guess you could do it by mapping to your own file which re-exports the thing you want.

```json
{
  "alias": {
    "ipcRenderer": "./electron-ipc.js"
  }
}
```

in electron-ipc.js:

```javascript
module.exports = require('electron').ipcRenderer;
```

Cliff122:
I have a SCSS file containing `@import "~bootstrap/scss/bootstrap";` which I import into my main `app.js` like this: `import '../scss/styles.scss'`. I still however get this error in the console when trying to compile.
`C:\Users\david\Coding\bootstrap-test\src\scss\styles.scss:1:1: File to import not found or unreadable: ~bootstrap/scss/bootstrap.`

What am I doing wrong?

DeMoorJasper:
@Cliff122 This hasn’t been released to npm. You could link to github. ‘~bootstrap...’ should probably also be ~/bootstrap...

Cliff122:
Alright, I took your advice and cloned the repo and built it locally on my machine. But even when I run the project using that version the error still appears. It looks like the error is thown by the `node-sass` module, if that is relevant.

DeMoorJasper:
@Cliff122 Aha! I remember a similar issue, SASS uses it's internal resolver unfortunately, as far as i'm aware, you can read up the whole discussion here: https://github.com/parcel-bundler/parcel/issues/316 it basically means it's out of parcel's control.

Cliff122:
Ok, that makes sense. When importing directly in the javascript file however, like this: `import '~/bootstrap/scss/bootstrap.scss'`, SASS shouldn't have anything to do with resolving the path, right? When I try that though I instead get the error `C:\Users\david\Coding\bootstrap-test\src\scripts\app.js:1:7: Cannot resolve dependency '~/bootstrap/scss/bootstrap.scss'`.

DeMoorJasper:
@Cliff122 hmmm not sure why u get an error, might be u are using it incorrectly, your import would resolve like this I think.
`~/bootstrap/scss/bootstrap.scss` => `C:\Users\david\Coding\bootstrap-test\src\bootstrap/scss/bootstrap.scss`
at least if the entrypoint is inside `C:\Users\david\Coding\bootstrap-test\src\`

Cliff122:
Oh, my bad! I must have misread the description of the pull request. I thought tildes would resolve to the `node_modules` directory, but looking closer I realize it won't. I managed to solve it however by installing the `node-sass-tilde-importer` module from npm and creating a `.sassrc.js` file in my project directory with the following content.
```javascript
// .sassrc.js
const importer = require('node-sass-tilde-importer')

module.exports = {
  importer
}
```

3stacks:
Hey folks, any idea when this one will be published to npm? I'm currently pointing to latest on github and it's causing issues with importing scoped packages for me. Loving the bundler though

ellsclytn:
Does this permit recursive module resolution within an alias? I.e. if I have `alias.bar: 'foo/bar'` in my `package.json`, can I then `import qux from 'bar/baz/qux'` and have that resolve to `foo/bar/baz/qux`? A practical case would be if someone wanted to use Lodash over Underscore, or vice-versa, and wanted to use the nested module style of import, rather than destructuring.

ckken:
preact-compat alias in package.json is no work for me .

devongovett:
@ellsclytn yes, it works as you describe.

@ckken this PR hasn't been released to npm yet. That should work in the next release.

lifeiscontent:
@devongovett I can't seem to get tilde paths to work... which version of parcel is this?

davidnagli:
@lifeiscontent It hasn’t been released yet... Big release coming soon!

derolf:
Aliasing from HTML files using the tilde syntax doesn‘t work. Will this also come?

derolf:
I am working on the „externals“ issue. It would be very helpful if we could move the resolution step into the parse phase. This way aliasing works from ALL assets, i.e. you could write <script src=‘react/cjs/minified.js’> in HTML.

Brouilles:
@davidnagli An approximation for the release of this update? So excited!

Siyfion:
Does this also allow imports for modules residing in a monorepo root node_modules folder?

i.e. Will `~/example` search up the directory tree until it finds a node modules folder containing the example module, or will it stop when it gets to the project root folder?

wesleyboar:
[1.70 has been released](https://github.com/parcel-bundler/parcel/releases). I am very much looking forward to [some documentation](https://github.com/parcel-bundler/website/issues/132).

---

I am trying to use `~/` for protject-root-relative paths—and I am failing…

```
/omg/absolute/path/scripts/src/site/site.js:21:28: Cannot resolve dependency '~/src/site/visual/is-logged-in.js' at '/omg/absolute/path/scripts/src/site/~/src/site/visual/is-logged-in.js'
```

The nearest `package.json` is at `/omg/absolute/path/scripts/package.json`.

freeall:
I have the same issue as @wesleyboar. I have been used to doing something like `import Foo from ~/components/Foo`. Even in node core I remember this having been a longer discussion at one point.

To me it has some great advantages:
1. It allows us to quickly write requires/imports without thinking about how many `../` we have to put first.
2. When moving code around you don't also have to look through requires/imports and change all that code, which can lead to a reason not to move code to the appropriate place.

Since you've already implemented reading `alias` from `package.json` it would feel natural to add it to that api as well as something like this:
``` json
{
  "alias": {
    "~": "./src/app/"
  }
}
```

3stacks:
^ Maybe submit a new issue for this. Your request might get lost in a merged PR

Stradivario:
Any news about this ?

goyney:
It seems that folders aliased via the package.json file are not being transpiled by babel.

```
  "alias": {
    "clone-object": "../modules/clone-object",
    ....
   }
```

Instructing babel via a .babelrc file to transpile down to es5 doesn't apply to these aliased folders. How can we make that work?

rocinante42:
> It seems that folders aliased via the package.json file are not being transpiled by babel.
> 
> ```
>   "alias": {
>     "clone-object": "../modules/clone-object",
>     ....
>    }
> ```
> 
> Instructing babel via a .babelrc file to transpile down to es5 doesn't apply to these aliased folders. How can we make that work?

Did you solve this somehow @goyney ?

goyney:
Yeah, we solved it by moving to Webpack. 😄

derolf:
Me too.

mischnic:
@derolf @goyney Can you post a more complete code sample of your problem?

goyney:
No, I cannot. We literally switched away from Parcel over a year ago.

PR #852, opened 2018-02-19
Fix HMR runtime error call stack message

mischnic:
bildschirmfoto 2018-02-19 um 14 37 56


mischnic:
Why is that rust minification test always failing?

Issue #761, opened 2018-02-07
Prevent that "require" is exposed as global function


klipstein:
Why does `parcel build` create build files that expose a global `require` function and is there a way to prevent that?

```
// beginning of js build file
require=function(r,e,n) {...
```

### 🤔 Expected Behavior

When parcel is used to build libraries that are loaded on 3rd party websites, it should prevent to expose a global `require` function, because other scripts on that site could rely on a global `require` function.

### 😯 Current Behavior

When loading a JS file, which was built with parcel, it exposes require on the window object.

### 💁 Possible Solution

Add the possibility to configure, if require gets exposed or not.

### 🔦 Context

Loaded a JS build on a foreign website which relied on a different global require.

termhn:
I think parcel is quite focused on being a monolithic bundler for web apps, not for building libraries.

klipstein:
@termhn also thought that this is probably the reason for this global `require`.

My use-case is not necessarily a library. I use parcel to package code, that provides a web component, which then can be used on other websites. I used browserify, closure-compiler and webpack before and I really like the ease-of-use of parcel (zero-config, watcher included, handling node dependencies properly) and would like to stick to it.

I solved my particular use-case by wrapping the build result into my own IIFE:

```js
(function(require){
  // parcel build result
  require=...
}())
```

jsiebern:
Unfortunately that introduces a new problem for dynamic imports as the require chain get's broken! I have a use case where I need several packages on the same page with dynamic imports and it got completely messed up. I wrote a small plugin that helps me transform `require` into `require_{HASH}`.

@devongovett Imo the ideal solution to this problem would be to create a singleton named `requireParcel` for example (introduced by the prelude) where the modules get registered by identification of the hash of the entryAsset. So the require calls would probably look like `requireParcel('entry-asset-hash', './some/module');`. Does that make any sense in relation to the long term goals of parcel?

jsiebern:
@klipstein This is highly tailored towards my needs but maybe you can write your own for now to automate it: https://github.com/jsiebern/parcel-plugin-multi-require/blob/master/plugins/MultiRequirePlugin.js

mubaidr:
This is the reason node modules are not available in Electron when parcel is used, details: https://github.com/parcel-bundler/parcel/issues/1244

DeMoorJasper:
This has been fixed by now if I'm not mistaken

klipstein:
Thanks for taking care. The solution works for me.

[...]

PR #854, opened 2018-02-19
Change require logic to a per bundle basis

jsiebern:
Solves the issue discussed in [this issue](https://github.com/parcel-bundler/parcel/issues/761). This is probably not the best approach but it worked fine in my first use-cases. I would much rather create a singleton but I'm lacking the time to implement it as it is a bit more sophisticated.

We could probably write a cli flag for it as well (disabled by default).

Any thoughts on this are more than welcome :)

### DIFF
--- a/src/Asset.js
+++ b/src/Asset.js
@@ -41,6 +41,21 @@ class Asset {
     this.bundledSize = 0;
   }
 
+  topHash() {
+    let pBundle = this.parentBundle;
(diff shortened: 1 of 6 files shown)


jsiebern:
That still would not solve the issue I've had, but I have since moved back to Webpack for the production bundle (definitely staying with parcel for dev though :) ).

PR #861, opened 2018-02-20
Add instruction for install fsevents

hayeah:
Without `fsevents`, parcel bundler would silently operate in polling mode, causing high CPU usage at rest.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,12 @@ or with npm:
 npm install -g parcel-bundler
 ```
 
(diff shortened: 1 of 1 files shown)


brandon93s:
`fsevents` is an optional dependency of `chokidar` and should be installed automatically on supported platforms. There shouldn't be any reason to need to install it globally. 

Have you experienced something different?

hayeah:
huh, i tried nuking the global yarn directory and reinstalling parcel from scratch, and ya `fsevents` was installed by default.

not sure what happend before, sorry guys~

PR #878, opened 2018-02-21
Add useBuiltIns babel-preset-env option (fix #843)

### LINKED ISSUE #843: Missing useBuiltIns (babel-polyfill env transform) support.
This a 🐛 bug report

```json
{
  "presets": [
    [
      "env",
      {
        "targets": {
          "browsers": [
            "last 2 Chrome versions",
            "last 2 Edge versions",
            "last 2 Firefox versions",
            "last 2 iOS versions"
          ]
        },
        "useBuiltIns": true,
        "modules": false
      }
    ]
  ],
  "plugins": [
    "syntax-dynamic-import",
    ["transform-react-jsx", { "pragma": "h" }]
  ],
  "env": {
    "test": {
      "plugins": ["dynamic-import-node"]
    }
  }
}
```
[Test app repo](https://github.com/samsch/hyperapp-parcel/tree/2784bc3bcb122d1c21976481e7bdc00c1d827027). Note that the .babelrc file env browserlist does not include IE, nor should it include regenerator or compile generator functions.

### 🤔 Expected Behavior

The [useBuiltIns babel-preset-env option](https://babeljs.io/docs/plugins/preset-env/#usebuiltins) transforms `import 'babel-polyfill';` into only the necessary individual polyfill imports. Notably, if your browserlist has support for generators, it does not import the regenerator runtime.

### 😯 Current Behavior

Running `npx babel-cli src/index.js | less` (Linux command) shows the correct output, however, if you view/search in the compiled output from Parcel (I've been testing with `npx parcel build-dev --no-cache src/index.html`), you can see that while the babel transform is correct (generators are not compiled to use regenerator), the regenerator runtime is still included from babel-polyfill.

### 💁 Possible Solution

I'm not familiar with Parcel internals, but one possible thought is that Parcel might be transforming `import 'babel-polyfill';` into something babel-preset-env doesn't recognize.

### 🔦 Context

The main problem here is large unused code. Without polyfills, my example app is 4.25KB. Importing babel-polyfill (and getting all of it) brings the app to 148.23KB. If I manually use the individual polyfill imports (copied from babel-cli output), I get 31.08KB.

(I'm only picking on regenerator here because it's the one I remember the easiest, and it's easy to search for in the output (no false positives), so it's an easy test.)

### 🌍 Your Environment

[Package versions in package.json](https://github.com/samsch/hyperapp-parcel/blob/2784bc3bcb122d1c21976481e7bdc00c1d827027/package.json)

| Node             |v9.3.0
| npm |v5.6.0
| Operating System |Xubuntu 17.10

samsch:
Enables babel-preset-env useBuiltIns option for source packages (and not node_modules packages). Fix for #843

No tests were harmed in the making of this pull request. (Prettier got excited and added a newline though.)

I added babel-polyfill as an explicit devDependency for a test.

As per any test dependent on external sources (browserlist), this isn't perfectly future-proof.


src/transforms/babel.js
```diff
@@ -239,7 +239,7 @@ async function getEnvConfig(asset, isSourceModule) {
 
 const envCache = new Map();
 
-async function getEnvPlugins(targets) {
+async function getEnvPlugins(targets, useBuiltIns) {
```
2hu12: According to the [babel preset-env doc](https://babeljs.io/docs/plugins/preset-env/#usebuiltins), the default value of `useBuiltIns` is `false`, so what about make the second parameter of `getEnvPlugins` with a default `false`, like
```js
async function getEnvPlugins(targets, useBuiltIns = false) { ... }
```

samsch: There does seem to be a little precedent to use a default argument, and it does seem to make usage a bit clearer. This does not actually change any functionality, since the argument is used for a truthy check, and undefined is falsy.

FDiskas:
This will fix errors like `src/index.js: Unknown version 62 of android` ?

samsch:
@FDiskas Probably not, no. This shouldn't fix any errors, it decreases the bundle size by replaying an import of `babel-polyfill` with separate imports of just the necessary polyfills.

devongovett:
Thanks for doing this. When we upgrade to Babel 7, we should also switch to the new `usage` option for `useBuiltins` which makes it so you don't even need to import babel-polyfill at all.

Issue #863, opened 2018-02-20
🙋 Replace chokidar with fb-watchman


hayeah:
**Choose one:** is this a 🐛 bug report or 🙋 feature request?

`chokidar` only supports native file-system notification for MacOS (a limitation because it relies on fsevents).

[watchman](https://facebook.github.io/watchman/docs/install.html) provides efficient fs notification for all platforms. Watchaman is used by the `react-native` bundler, which seems to work pretty well.

devongovett:
Interesting. Are there nodejs bindings that don't require installation via `brew` or another package manager but just via npm?

If someone wants to experiment with this, it looks cool!

mrcoles:
Also, FWIW, `create-react-app` sees changes for me in this situation.

setup:

```bash
$ create-react-app -V
1.5.2
$ create-react-app test-create-react-app
$ cd test-create-react-app/
$ yarn start
```

Then edit `src/App.js` and http://localhost:3000/ auto-updates.

davidnagli:
I think this a really interesting idea, agree with @devongovett here, we should definitely experiment with this 👍 

I found that for some reason it’s a bit hard to find information on google about using watchmen with Node-JS, so here are some links to use as starting points:

[Watchmen NodeJS Docs](https://facebook.github.io/watchman/docs/nodejs.html)
[Watchmen NPM Package](https://www.npmjs.com/package/fb-watchman)

davidnagli:
@mrcoles Are you talking about the dropbox thing?

yakotika:
I tried to replace the library in the library, but I got the error 

> Watchman was not found in PATH.

yakotika:
I tried to replace the chokidar in the fb-watchman, but I got the error:

> Watchman was not found in PATH.

I tried to do like in this [webpack-watchman-plugin](https://github.com/researchgate/webpack-watchman-plugin/blob/master/src/WatchmanConnector.js)

brandon93s:
From the `webpack-watchman-plugin` readme:

> Make sure watchman is installed and ready to use on your system.

You have to install it on your machine on your own for the plugin to work. This is going to be the biggest barrier to switching.

DeMoorJasper:
I'm not sure it's worth switching as installing watchman to enable the watcher would sort of defy the whole `zero-config` experience.
[...]

PR #889, opened 2018-02-23
[WIP/PoC] Watchman backend

mischnic:
This pull request adds a watchman backend, currently intended mainly for (performance) testing (@davidnagli). The api of the watchman wrapper (at src/Watchman.js) is the same as the one from `require('chokidar').FSWatcher`, so deciding which to use at runtime should be quiet easy. 
I added more `await`s because all commands are sent to watchman asynchronosly.
There could be room for optimization.

To use watchman, set the environment variable `PARCEL_WATCHMAN=1`. The tests pass with watchman and chokidar:
```sh
yarn test test/watcher.js
PARCEL_WATCHMAN=1 yarn test test/watcher.js
```

For the original discussion see #863.

This code doesn't check if watchman is actually installed! (Will be added via `fb-watchman`'s `client.capabilityCheck` if this gets well received.)

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,6 +30,7 @@
     "cssnano": "^3.10.0",
     "deasync": "^0.1.12",
     "dotenv": "^5.0.0",
+    "fb-watchman": "^2.0.0",
     "filesize": "^3.6.0",
     "get-port": "^3.2.0",
     "glob": "^7.1.2",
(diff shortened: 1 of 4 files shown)


mischnic:
Is there still interest now that the Watcher was revamped?

devongovett:
Yeah I think probably. While the new one is better, I think we may want to fully switch off of chokidar. It's too slow for large projects. I'm looking into alternatives at the moment.

devongovett:
Going to close this for now. The watcher has changed significantly in the meantime. While we do eventually want to replace chokidar, it will likely be another alternative and not watchman since we don't want to force users to install another tool on their system.

Alternatives worth considering:

https://github.com/Axosoft/nsfw
https://github.com/atom/watcher

Issue #787, opened 2018-02-11
JS file (vuetify) not bundled correctly, causing 404


wlcx:
🐛 bug report

When trying to use parcel on a project that uses [vuetify](https://github.com/vuetifyjs/vuetify), `vuetify.js` is not bundled into the dist folder under the filename that is inserted into index.html.  

### 🎛 Configuration (.babelrc, package.json, cli command)

Create new folder and enter it
`$ mkdir reproduce-the-bug && cd $_`
Install vuetify
`$ npm install vuetify`

Create **index.html**
```
<!DOCTYPE html>
<html lang="en">
<head>
    <meta charset="UTF-8">
    <title></title>
</head>
<body>
    <script src="./node_modules/vuetify/dist/vuetify.js"></script>
</body>
</html>
```

`$ parcel index.html`

### 🤔 Expected Behavior

* `vuetify.js` would be bundled into dist under a long (hashed?) filename.
* The bundled filename would be inserted into `index.html`
* The page would load

### 😯 Current Behavior

* `vuetify.js` is not renamed in the `dist` folder
```
$ ls dist
index.html  vuetify.js  vuetify.map
```
* Although, this file is different to the referenced one in `node_modules`, so some processing has happened...

* **However**, a bundled filename is inserted into `index.html`: 
```
) cat dist/index.html
<!DOCTYPE html>
<html lang="en">
<head>
    <meta charset="UTF-8">
    <title></title>
</head>
<body>
    <script src="/dist/3f0b13654a52acc5af1e96bddb0a7840.js"></script>
</body>
</html>
```

* **If the line `"main": "dist/vuetify.js"` is deleted from `node_modules/vuetify/package.json`, it works as expected** (this took some trial and error!)
### 🔦 Context
Just trying to build my stuff ;)

### 💻 Code Sample
See above!

### 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Parcel           |1.5.1
| Node             |v8.9.4
| npm/Yarn         |5.6.0
| Operating System |MacOS 10.13.2

DeMoorJasper:
This appears to be related to how generatebundleName works, related code: https://github.com/parcel-bundler/parcel/blob/master/src/Asset.js#L181
Not sure why the hashed name is used inside html though

DeMoorJasper:
On further investigating i discovered that inside Asset it uses Asset.addURLDependency(...), to include it inside html, this uses 
```Javascript
this.options.parser
[...]

PR #891, opened 2018-02-24
[RFC] ipc between bundler and assets

DeMoorJasper:
This PR is a RFC regarding putting an ipc link between worker/asset and the corresponding bundler.
This would be helpful in some cases and could potentially speed up parcel, as i've tested ipc speeds and they should not form a bottleneck. The case I tried to fix with this PR is #787.

* A streamlined way to communicate `Worker <=> WorkerFarm` without having to write an implementation for every seperate class/function we want to address (possibly with a custom WorkerFarm, it kinda makes sense as we've already expanded the WorkerFarm quite a lot - potentially using promises instead of callbacks to be more in line with the rest of parcels code)


devongovett:
This is really cool, but we don't have a reason for it since #934. If we ever need something like it, we can come back to this! 😄

Issue #872, opened 2018-02-20
[RFC] File naming strategy


devongovett:
This is a meta issue to discuss a number of things that have come up about file naming in Parcel.

1. Keep original filenames for HTML files: #433, #280, #557, #307
2. Hashing assets based on file content for cache busting: #717, #188, #753, #756, #829
3. Putting assets in a separate folder from HTML: #233 

I think we can come up with a cohesive file naming strategy that meets all of these needs.

* We hash all assets based on file contents to produce filenames like `index.a8b29e.js`, except in the following cases (taken from the rules outlined by @Munter [here](https://github.com/parcel-bundler/parcel/issues/280#issuecomment-353894742)). In those cases, we use the original filenames.
  - Any graph entry point (usually html)
  - Any asset linked to with an `<a href>`
  - Any asset linked to with a `<meta http-equiv="refresh">`
  - Serviceworkers (must keep consistent file name across builds)
  - `humans.txt`, `robots.txt`, `.htaccess`, `favicon.ico`
  - Cache manifests, rss and atom feeds
* Place hashed assets (things not matched by above rules) in `dist/assets` e.g. `dist/assets/index.a8b29e.js`. This would be flattened as it is currently, so `src/some/path/something.js` would be placed in e.g. `dist/assets/something.fd5se2.js`.
* Place non-hashed assets (things matched by the above rules) in the root, and create directories as needed to match the original paths. For example, if an HTML file were linked to from `<a href="/some/path/something.html">` the output file would be `dist/some/path/something.html`.
* We also support the `-o` or `--out-file` CLI option, which would override the default name for the entry file. If not provided, and the entry file matches `main` in package.json, use the package name.

The only case where this breaks down is if the input path started with `/assets` - which is the folder we're already using for static things. Not sure what to do about that: I guess we could try to generate a unique name for the assets folder or something. Open to suggestions here!

Otherwise, I think this strategy solves most of the issues listed above. Please let me know your feedback and make any suggestions you think would improve this strategy!

cc. @zeakd @songlipeng2003 @ssuman @Munter @benhutton @shanebo @leeching @gamebox @npup

jsiebern:
I think its a great strategy, though I'd like a CLI option for not putting the assets in a subfolder! Is `index._hash_.js` based on the root file name or on the budle file name?

Munter:
[...]

PR #892, opened 2018-02-24
asset.addURLDependency, incorrect package.json fix

### LINKED ISSUE #787: JS file (vuetify) not bundled correctly, causing 404

DeMoorJasper:
closes #787

Less drastic alternative than #891, although this is slightly slower.


devongovett:
Ah, good catch! Seems like this would be solved by just not using the package name in `generateBundleName`. Not sure we really need that anyway with #872.

DeMoorJasper:
@devongovett Isn't it still necessary for that RFC?
`We also support the -o or --out-file CLI option, which would override the default name for the entry file. If not provided, and the entry file matches main in package.json, use the package name.`

If i'm mistaken it could still be used as a temporary patch untill the RFC gets into implementation phase I guess.

devongovett:
Yeah, but actually not sure how useful that will be if we are hashing things anyway. Closing this one, should be fixed by #934.

Issue #802, opened 2018-02-13
Debugging Dedupe


natew:
🐛 bug report + 🙋 feature request

I'm getting more than one instance of a package loaded, but am a bit lost as to how to debug it.

### 🤔 Expected Behavior

Basically not to have duplicates, or to see why duplicates are appearing if so. I've checked around and it seems it should be singly-loaded. I have a few apps but all sub-dependencies define as peerDeps. There is a copy of of the module *above* the app itself, so it could be coming from that perhaps.

### 💁 Possible Solution

One idea would be to have something like a Parcel object in the console to check through and see information on things.

Just opening this to see if there may already be some way, didn't see anything in docs/issues.

davidnagli:
Didn’t fully understand what you’re asking, but from what I got, you’re asking how to debug Parcel, and not how to solve your whole duplicate problem itself….

In that case, I recommend setting the `PARCEL_WORKERS` environment variable to `0` and just using a debugger to step through the code and see what it’s doing and where it’s going wrong.

natew:
Basically I'm seeing a duplicated package that I don't think I should be, and unsure how I can debug it. It may be my fault.

In a huge bundle, stepping through seems pretty hard. Having somewhere to inspect the parcel state would allow much better debugging I'd think.

jeanfortheweb:
Possible duplicate of #790 ?

DeMoorJasper:
Closing, as it seems a duplicate of #790

PR #895, opened 2018-02-26
Bugfix/dynamic hoist

jeanfortheweb:
Bugfix for #790 and possible #802.
There has been changes made since 1.4.1 which unfortunately fixed the issue only partially.
This time, I made the PR for the master, since the "freeze" issue since 1.5.1 for at least virtual windows machines has been found, but will be opened in a separate case.


jeanfortheweb:
There are two HTML tests failing on appveyor which do not fail local.
Also, my code changes do not intersect with these tests. Any suggestions?

DeMoorJasper:
I can confirm that the tests fail on windows, the first html test results in a bundle that lacks the JS bundle

jeanfortheweb:
Guess I may should take a look at it, since my other PR is building multiple times without any issues...

devongovett:
It is actually possible and correct for an asset to appear in more than one bundle. For example, imagine you had `a.html` and `b.html` and they import `a.js` and `b.js` respectively. If both `a.js` and `b.js` import `c.js`, then `c.js` should appear in the bundled output for both `a.js` and `b.js`. This PR would  change that behavior so that it would only appear in one of the two - which would break the other page.

Can you provide more details and possibly a reduced test case on the exact duplication you are seeing?

jeanfortheweb:
I will check your fear through tests. I was thinking that a different HTML file will get it's own tree, since they have independent entry points. But it shouldn't be too complicated to create a test case for this. With this patch I was able to reduce the bundle output by almost 5MB while maintaining it's functionality.

Issue #896, opened 2018-02-26
Less 3.0.0 update


bricss:
Any plans for update? -> https://github.com/less/less.js/blob/master/CHANGELOG.md#300

DeMoorJasper:
I'll dig into the stuff that broke and try to get a PR with an update out soonish

loki1404:
555-555-0199@example.com

devongovett:
We use the project local version of less, so it should just work if you install less 3.0.0 in your project.

PR #897, opened 2018-02-26
Update Less to v3.0.1

DeMoorJasper:
As requested here: #896 

It appears parcel is not affected by any of the breaking changes.


loki1404:
555-555-0199@example.com

devongovett:
Shouldn't matter too much since we use the project local version of less. This is just a dev dependency, but good to run the tests against it I guess.

PR #901, opened 2018-02-26
Allow adjustment of maxConcurrentCallsPerWorker through env variable

### LINKED ISSUE #900: Freeze of bundler since 1.5.1
### 🤔 Expected Behavior

The bundler should finish on any system it runs on.

### 😯 Current Behavior

On specific systems, in our case a virtual Windows Server environment, the bundler stops and random files without any errors.

### 💁 Possible Solution

My investigation showed that it is some wierd scenario of memory issues, causing certain workers to become zombies. Therefore, these workers will never stop working and the bundler stops. It's possible to fix this issue by adjusting the "maxConcurrentCallsPerWorker" settings of the worker-farm. In our case, we reduced this value from 10 to 5 to get a stable result.

To become maximum flexible, I would suggest to make this setting available as env variable, just the like worker count, keeping the default value at 10.

### 🌍 Your Environment

| Parcel           | ^1.5.1
| Node             | ^8
| npm/Yarn         | ^1.3
| Operating System | Virtual Windows Server

jeanfortheweb:
Fixes #900 by allowing to set maxConcurrentCallsPerWorker through env variables. On affected systems, the maxConcurrentCallsPerWorker value can then be reduced:

`PARCEL_WORKER_CALLS=5 parcel`


src/WorkerFarm.js
```diff
@@ -9,7 +9,10 @@ let shared = null;
 class WorkerFarm extends Farm {
   constructor(options) {
     let opts = {
-      maxConcurrentWorkers: getNumWorkers()
+      maxConcurrentWorkers: getNumWorkers(),
+      maxConcurrentCallsPerWorker: process.env.PARCEL_WORKER_CALLS
+        ? parseInt(process.env.PARCEL_WORKER_CALLS)
+        : 10
     };
```
brandon93s: A few changes:
1. Provide a radix to `parseInt`
2. Only define `maxConcurrentCallsPerWorker` if the env variable is defined so that `worker-farm` can continue to own the default.
3. worker-farm supports `Infinity` for `maxConcurrentCallsPerWorker`, let's handle that

```js
let opts = {
  maxConcurrentWorkers: getNumWorkers()
};

if (!isNaN(process.env.PARCEL_WORKER_CALLS)) {
	opts.maxConcurrentCallsPerWorker = isFinite(process.env.PARCEL_WORKER_CALLS)
		? parseInt(process.env.PARCEL_WORKER_CALLS, 10)
		: Infinity;
}

```

brandon93s: May also consider naming it with the concurrency in mind. Current name implies the work is called x times. 

`PARCEL_WORKER_CONCURRENT_CALLS`

devongovett:
Why is this actually necessary to configure? It seems like solving a symptom rather than a real problem. Can we figure out what the root problem is and fix that instead?

jeanfortheweb:
On the virtual window server, i logged everything possible, from the worker before and after processing to everything the worker farm does output. I disabled the Logger class for this completely so that I can see every single log and no one gets swallowed by accident. The result is, that some workers are just stop responding. I would love to have proofs for my claims but I don't. I claim that node has certain memory problems on windows, causing workers with to much load at the same time to break but not to quit. They just freeze. So they never finish jobs which is the reason that the worker farm does not kill them. The only solution i've found, and maybe it's kind of a bruteforce solution, is to reduce the number of calls made to a single worker.

src/WorkerFarm.js
jeanfortheweb: Done! Sorry for the late response

juliocanares:
we need this, we tried the PR an it works 🔥

devongovett:
hmm, maybe we should just set a good default? Is there a point at which the performance benefits of concurrent calls flatten off? At what point does increasing the concurrent calls cause this issue? Maybe we can find the intersection.

fathyb:
Isn't this related to nodejs/node#7657 and libuv/libuv#1099?

I agree with @devongovett that we should find the root cause first, but if we don't quickly we need a workaround as this issue makes Parcel unusable on Windows.

DeMoorJasper:
If this is a windows bug in node.js perhaps we could add a default value for windows?

pke:
So does this mean on Windows we would have to specify a env var to reduce the number of workers? Otherwise this patch still falls back to 10 workers which would still cause the same hangs. @jeanfortheweb

jeanfortheweb:
Sorry for the late response @pke. Yes, this is how this PR works. Quiet idiomatic: If your workers go zombie mode on windows, adjust them until they work on your system. It's never the same value. It depends on the windows version, cpu count and the memory available.

PR #923, opened 2018-03-01
Load spinner

DeMoorJasper:
This changes the hourglass emoji into a animated clock emoji and an animated block for windows


pke:
Just wanted to note, that git bash under Windows may support emoji when using a truetype font. Haven't tested it but I saw some fancy symbols used by other npm packages there already.

PR #293, opened 2017-12-15
[WIP]: HTML bundle loader dynamically through import statement. (#178)

abiduzz420:
- [x] All tests passing.
- [x] Prettier format done

### Changes included:
- added `@webcomponents/webcomponentsjs` dependency as a polyfill
- created `loadHTMLBundle` method in `bundle-loader.js`
- included html-import directory in test/integrations.
- wrote a test to support dynamic imports in HTML

_This PR is still Work in Progress_

PR #926, opened 2018-03-01
Import/require html from js

### LINKED ISSUE #178: Add a html bundleloader
🙋 feature request

Add a html bundleloader to allow html fragments to be loaded dynamically, like already possible with js or css content.

### 🤔 Expected Behavior

Add a html bundleloader to allow html fragments to be loaded dynamically, like already possible with js or css content.

import('view.html')

### 😯 Current Behavior

Currently the following exception is thrown as no html bundleloader is implemented yet:

bundleLoaders[type] is not a function

### 💁 Possible Solution

Add a new html bundleloader to the existing bundleloaders:

https://github.com/parcel-bundler/parcel/blob/c8295de9b7a1955c70f3619ccdced93284e6f601/src/builtins/bundle-loader.js#L24

### 🔦 Context

I'm migrating a small spa from requirejs to parcel-bundler.

### LINKED ISSUE #848: AngularJS Routing + Parcel = Error
# Angular version
angular 1.6.9

# Angular route version
angular-route 1.6.9

# Parcel version
parcel-bundler 1.6.1

Hey! I'm writing this issue tonight because i've experienced some errors with Angular 1.6.9 routing and Parcel Bundler.

I'm not sure if this issue is related to Angular or Parcel, but feel free to take me back to the right way if i'm wrong.

As a developer, I like to split each blocks of code in separate files for better organization and since I don't like configuration, Parcel seems to be something I'll use a lot.

Here's my router.js file:
```javascript
export default function routing($routeProvider, $locationProvider) {
    $routeProvider
        .when('/', {
            controller: 'HomeCtrl'
        })
        .otherwise({redirectTo: '/'})
    $locationProvider.html5Mode(true);
}
```

Inside a route, I have to specify which template or templateUrl to use. I've tried many ways to do this because if I specify a templateUrl like this `../pages/home.html` my application instantly crash.

However, if I use instead a simple `template:` key and enter something like this: `require('../pages/home.html)` it doesn't work.

I've made some researches and found nothing about that issue. That's why i'm not sure if its related to AngularJS or Parcel.

The only way to make it work is to create a `home.js` file and to export it:
```javascript
export default `
    <h1>asdasdasd</h1>
    <p>Test routing on / {{message}}</p>
`
```

I personnaly don't think that this is the best way to do that. Let me know if you have any informations about that. I may provide more details if needed.

Thanks!

**EDIT: Sorry for my bad english!**

pierredavidbelanger:
This is the first time I contribute to this project, so I am sorry if I missed something important.

I wanted to do something like what is described here #848 
In short, I wanted to be able to `const template = require('./other.html');`

You will see that I have no idea what I'm doing ;)
I am especially not proud of how I ended up [butchering src/packagers/HTMLPackager.js](https://github.com/pierredavidbelanger/parcel/commit/dedb0ee0dd4be54091a4972b12e0b5563c06836e#diff-742d394996519d0939508ba378368eff)

Also, all 232 tests are still passing.

Fixes #848 
Closes #178

### DIFF
--- a/src/builtins/loaders/text-loader.js
+++ b/src/builtins/loaders/text-loader.js
@@ -1,4 +1,4 @@
-module.exports = function loadHtmlBundle(bundle) {
+module.exports = function loadTextBundle(bundle) {
   return fetch(bundle).then(function (res) {
     return res.text();
   });
--- a/src/Bundler.js
+++ b/src/Bundler.js
@@ -44,6 +44,10 @@ class Bundler extends EventEmitter {
       'css',
       require.resolve('./builtins/loaders/css-loader')
     );
+    this.addBundleLoader(
+      'html',
+      require.resolve('./builtins/loaders/raw-loader')
+    );
     this.addBundleLoader('js', require.resolve('./builtins/loaders/js-loader'));
 
     this.pending = false;
--- a/src/Bundler.js
+++ b/src/Bundler.js
@@ -46,7 +46,7 @@ class Bundler extends EventEmitter {
     );
     this.addBundleLoader(
       'html',
-      require.resolve('./builtins/loaders/raw-loader')
+      require.resolve('./builtins/loaders/text-loader')
     );
     this.addBundleLoader('js', require.resolve('./builtins/loaders/js-loader'));
 
--- a/src/builtins/loaders/raw-loader.js
+++ b/src/builtins/loaders/raw-loader.js
@@ -1,0 +1,5 @@
+module.exports = function loadHtmlBundle(bundle) {
+  return fetch(bundle).then(function (res) {
+    return res.text();
+  });
+};
--- a/src/packagers/HTMLPackager.js
+++ b/src/packagers/HTMLPackager.js
@@ -1,6 +1,8 @@
 const Packager = require('./Packager');
 const posthtml = require('posthtml');
 const path = require('path');
+const fs = require('../utils/fs');
+const url = require('url');
 const urlJoin = require('../utils/urlJoin');
 
 // https://www.w3.org/TR/html5/dom.html#metadata-content-2
@@ -16,6 +18,8 @@ const metadataContent = new Set([
 ]);
 
 class HTMLPackager extends Packager {
+  setup() {}
+
   async addAsset(asset) {
     let html = asset.generated.html || '';
 
@@ -31,8 +35,23 @@ class HTMLPackager extends Packager {
       ).process(html, {sync: true}).html;
     }
 
-    await this.dest.write(html);
+    let name = this.bundle.name;
+    if (asset !== this.bundle.entryAsset) {
+      name = url.resolve(
+        path.join(path.dirname(this.bundle.name), asset.generateBundleName()),
+        ''
+      );
+    }
+
+    this.size = html.length;
+    await fs.writeFile(name, html);
   }
+
+  getSize() {
+    return this.size || 0;
+  }
+
+  end() {}
 
   addBundlesToTree(bundles, tree) {
     const head = find(tree, 'head');
--- a/src/packagers/JSPackager.js
+++ b/src/packagers/JSPackager.js
@@ -35,7 +35,7 @@ class JSPackager extends Packager {
   }
 
   async addAsset(asset) {
-    if (this.dedupe.has(asset.generated.js)) {
+    if (!asset.generated || this.dedupe.has(asset.generated.js)) {
       return;
     }
 
--- a/src/packagers/JSPackager.js
+++ b/src/packagers/JSPackager.js
@@ -35,7 +35,7 @@ class JSPackager extends Packager {
   }
 
   async addAsset(asset) {
-    if (!asset.generated || this.dedupe.has(asset.generated.js)) {
+    if (this.dedupe.has(asset.generated.js)) {
       return;
     }
 
@@ -138,6 +138,7 @@ class JSPackager extends Packager {
     }
 
     if (bundleLoader) {
+      await bundleLoader.process();
       await this.addAssetToBundle(bundleLoader);
     } else {
       return;
--- a/test/html.js
+++ b/test/html.js
@@ -27,7 +27,7 @@ describe('html', function() {
         },
         {
           type: 'js',
-          assets: ['bundle-loader.js', 'raw-loader.js', 'index.js'],
+          assets: ['bundle-loader.js', 'text-loader.js', 'index.js'],
           childBundles: [
             {
               type: 'map'
@@ -48,7 +48,7 @@ describe('html', function() {
               assets: [
                 'bundle-loader.js',
                 'bundle-url.js',
-                'raw-loader.js',
+                'text-loader.js',
                 'index.js'
               ],
               childBundles: [
--- a/test/html.js
+++ b/test/html.js
@@ -27,7 +27,7 @@ describe('html', function() {
         },
         {
           type: 'js',
-          assets: ['index.js'],
+          assets: ['bundle-loader.js', 'raw-loader.js', 'index.js'],
           childBundles: [
             {
               type: 'map'
@@ -45,7 +45,12 @@ describe('html', function() {
             },
             {
               type: 'js',
-              assets: ['index.js'],
+              assets: [
+                'bundle-loader.js',
+                'bundle-url.js',
+                'raw-loader.js',
+                'index.js'
+              ],
               childBundles: [
                 {
                   type: 'map'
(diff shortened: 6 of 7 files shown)


bricss:
Yes, please!

pierredavidbelanger:
I initially assumed some things wrong. So I just pushed a new commit.

All tests are still passing.
Also, I am able to use the new feature in my simple angularjs projets.

Still, I do not really know what I am doing, I essentially just randomly hack here and there.

I do not know why I get [this same error with my OnsenUI project](https://github.com/OnsenUI/OnsenUI/issues/2360). 
I hope it is not related to my parcel hack.
I cannot tel for now.

Anyways, I hope this PR can help someone :)

davidnagli:
How does this relate to #293?

pierredavidbelanger:
@davidnagli I am actually using such a custom bundle loader to implements the HTML loading part. 

But, in retrospect, and to be honest, I am not sure I am heading the right way with this PR.

I guess, one want to `fetch(require('./other.html')).then(...)` 
but not `const content = require('./other.html');` as this PR does.

What do your think ?
I can discuss this further on the Slack if you want.

BrOrlandi:
I want this feature too! :+1:

lbguilherme:
@pierredavidbelanger You want `import('./other.html').then(...)`, which this PR seems to cover

arabsight:
This will enable me to use aurelia, I was hoping something like this would be added :+1:

src/builtins/loaders/text-loader.js
```diff
@@ -0,0 +1,5 @@
+module.exports = function loadTextBundle(bundle) {
+  return fetch(bundle).then(function (res) {
+    return res.text();
```
davidnagli: I would use an arrow function here 😀 

```javascript
module.exports.loadTextBundle = bundle => fetch(bundle).then(res => res.text())
```

pierredavidbelanger: The linter (in the pre-commit hook) does not seems to like arrow functions:

```
src/builtins/loaders/text-loader.js
  1:26  error  Parsing error: Unexpected token >

✖ 1 problem (1 error, 0 warnings)


husky > pre-commit hook failed (add --no-verify to bypass)
```

pierredavidbelanger: I was trying to commit this:

```js
module.exports = bundle => fetch(bundle).then(res => res.text())
```

davidnagli: Oh ya builtins are supposed to be in ES5 👍

unt0xication:
Is this PR is supposed to be released with the next release?

davidnagli:
Just realized the broader implications of merging this PR...

Currently HTML file imports are treated like RawAssets, in that they just return the resolved URL pointing to the file and not the actual contents of the file itself. 

However, by changing this behavior to allow HTML imports, we effectively break the way it previously worked.

That means that anything like:
```jsx
import loginPage from 'login.html'

<a href=loginPage>Link to login page</a>

```
would suddenly break, since now loginPage would be the actual contents of the login page rather than the url to the login page as it was before.


Maybe I'm wrong? Does your code magically handle this somehow?


Most likely we're gonna need to carefully design how this will play out and potentially introduce a breaking release when the time comes (Parcel 2.0?). Let me know if I'm wrong, but if not we should open a separate RFC issue to discuss / design everything and then come back to this once that's ready.

pierredavidbelanger:
@davidnagli You are absolutely right. This PR breaks the way it previously worked. And my code does NOT magically handle this :)

I had a bad felling a couple days ago about heading the wrong way with this PR.
Now you proved it.

If was just me, I would close this one, until we know what we need and, of course, when Parcel is ready for a breaking release.

pierredavidbelanger:
@davidnagli just to be sure though, 

right now, with the master branch, your snippet is not working:

```
import loginPage from 'login.html'

<a href=loginPage>Link to login page</a>
```

for at least 2 reasons:
- the generated HTML file name (the the file system) is not the same as the dependency name required() in the parent .js file
- the HTMLAsset is **not** treated like a RawAssets, so no `module.exports="/whateverf171619ff7dcf2171a706973.html";` is generated in the parent .js like it should with, say, a .txt file

I can try to have a look at this.

pierredavidbelanger:
@davidnagli I created the simplest app that demonstrate the problem:

``` sh
git clone https://github.com/pierredavidbelanger/parcel-html-test
cd parcel-html-test
yarn install
yarn run build
ls dist && tail dist/*.js
```

```sh
510368565fb792405acc6a86524a444c.html 510368565fb792405acc6a86524a444c.js   index.html
```

```js
LazyPromise.prototype.then = function (onSuccess, onError) {
  return this.promise || (this.promise = new Promise(this.executor).then(onSuccess, onError));
};

LazyPromise.prototype.catch = function (onError) {
  return this.promise || (this.promise = new Promise(this.executor).catch(onError));
};
},{"./bundle-url":10}],0:[function(require,module,exports) {
var b=require(8);b.load([["6b4a33c18b69460d577328b29c800143.html",6],4]);
},{}]},{},[0])
```

Just in the little snippet here you can see both points (file name VS dep name, raw URL VS bundle-loader) I mentioned earlier.

pierredavidbelanger:
I am sorry, it may seems spaghetti at first, but I tried to keep the concern separated, in case you accept one PR ant not the other.

I submit two PRs, for both the points I mentioned in a previous comment. 

I suggest you try my simple app that demonstrate the problem, then pull #1016 and #1017 into your local parcel, and try again my simple app; the problem is gone.

I think that #1016 and #1017 are better fixes than this PR for the problem of `require('./other.html')`;

adrian7:
FYI when all you need is to import a html file inside a string (e.g. like a nav menu or a logo with link) - 
 ```javascript
//Import html as string
import templateString from './template.html';

//Parse string to dom node
let template       = document.createElement('template');
templateString     = templateString.trim(); // Never return a text node of whitespace as the result
template.innerHTML = templateString;

//Return DOM Node
console.log(template.content.firstChild);
return template.content.firstChild;

```
 You can use the babel plugin: [babel-plugin-transform-html-import-to-string](https://www.npmjs.com/package/babel-plugin-transform-html-import-to-string)

bricss:
`FIY` -> It's impossible to use Parcel almost in any project without this feature.

BrOrlandi:
@adrian7 That's the perfect solution. I think that would be nice to have it by default in Parcel bundler.
@bricss It's not true, I'm using Parcel in projects without needing to import HTML files. 
I wanted to import HTML files to be able to develop my own Router, it would be a nice use case.

I have implemented a sample project with @adrian7 solution:
https://github.com/BrOrlandi/ParcelImportHTML

mjgerace:
I think this is a pretty crucial feature to merge in. Voicing support for this feature.

jluisgarcia:
yes, import/include (whatever you want to call it). It would be Awesome
html includes typically to share navigation across multiple pages.
Edit in one place and parcel would take care to include across all pages
i.e.      <include src="./nav.html"></include>

bricss:
Simple question to ask about this PR. What if I don't want to put `HTML`, `CSS` and `JS` in one file, e.i. `JSX`? How to build projects then? Therefore, `Parcel` is mainly _useful_ for a projects using `React` and similar libraries.

Elazul:
@adrian7 I owe you a big thanks. Your solution is perfect for importing AngularJS 1.6 component templates.

stevengill:
Thanks @adrian7! The babel plugin worked at getting a string representation of the html. 

One issue it doesn't handle is resolving the dependencies/assets in the html file. For instance, if i have ``, `./imgPath.png` isn't copied into `dist`.

stevengill:
I've posted an updated PR based on this work over at https://github.com/parcel-bundler/parcel/pull/1528.

jpenna:
It would be nice to have the solution @adrian7 suggested (`babel-plugin-transform-html-import-to-string`) in the docs. I've been strungling with this for a long time already and this not just fixed my problem but also simplified a lot the code, without the readFileSync (which only runs in the entry JS, for what I observed).
If you think it is valid to have it there I can do the PR.

devongovett:
@jpenna this issue has been solved, and the html bundle loader will be available in the next release.

JudeOsborn:
Is this supposed to be working? When I do this:

`import someFile from './somefile.html`

I get this:

`Uncaught SyntaxError: Unexpected token '<'`

I suppose that's what I would expect with a typical import because I'm importing an HTML file and it's expecting a module, but this ticket gives me the impression it should work with parcel. Am I misunderstanding?

PR #929, opened 2018-03-02
Stop minifying svg

bencooper222:
htmlnano has a really irritating error (it's possibly intentional, hard to tell) where it ruins svg tags that use the `<use>` tag. You can view an example I created [here](https://glitch.com/edit/#!/lying-equinox?path=transform.js:52:1).

1. No minification: https://lying-equinox.glitch.me/
2. Minification with minifySvg set to false: https://lying-equinox.glitch.me/minify_good
3. Minification with minifySvg set to default (which is true): https://lying-equinox.glitch.me/minify_bad

As you can see by inspect element, the minification that the `/minify_bad` path uses totally butchers the contents of the `<use>` tag.

I suggest this PR because I think Parcel should aim for the minimally disruptive behavior. That's important for encouraging new people to use the project. Power users can, of course, create a `.htmlnanorc` in their `/` directory to override the Parcel default setting.

### DIFF
--- a/src/transforms/posthtml.js
+++ b/src/transforms/posthtml.js
@@ -36,7 +36,8 @@ async function getConfig(asset) {
         collapseWhitespace: 'conservative',
         minifyCss: {
           safe: true
-        }
+        },
+        minifySvg: false
       };
 
     config.plugins.push(htmlnano(htmlNanoConfig));


devongovett:
Should we fix this bug in posthtml-minifier instead of just disabling it?

bencooper222:
@devongovett I submitted an [issue](https://github.com/posthtml/htmlnano/issues/47) to the htmlnano repo and they haven't responded to it at all. I don't want to create a PR before knowing if they want it or not. IMO we should just go for damage control right now.

davidnagli:
Ya looks like there’s only 2 people watching that repo 😬 

 I added a comment and mentioned a few more people from the PostHTML project, so hopefully we get a response soon 👍 🚀

maltsev:
It was an issue in [SVGO](https://github.com/svg/svgo/), which [htmlnano](https://github.com/posthtml/htmlnano) uses for SVG minification. I updated it and released new version [htmlnano@0.1.7](https://github.com/posthtml/htmlnano/releases/tag/0.1.7), where `<use>` tags are handled well.

So you could delete this PR and release new version with updated `htmlnano` to solve that issue.

PR #934, opened 2018-03-02
Do not try to use the sanitized package name as the bundle name

pierredavidbelanger:
It does not work because not all part of the code agree on this.
Also, I guess we ***want*** the hash anyways, as a cache buster.

Take this simple example:
 
`$ tree src`
```sh
src
└── index.html
```

`$ cat src/index.html`
```html
<script type="text/javascript" src="../node_modules/jquery/dist/jquery.js"></script>
```

### Without this PR 

```sh
$ ./node_modules/.bin/parcel build src/index.html
```

`$ tree dist`
```sh
dist
├── index.html
├── jquery.js
└── jquery.map
```

`$ cat dist/index.html`
```html
<script src="/dist/1d486a6abe13f93fbb4e2e63028ada7c.js"></script>
```

This will cause a 404!

### With this PR

```sh
$ ./node_modules/.bin/parcel build src/index.html
```

`$ tree dist`
```sh
dist
├── 1d486a6abe13f93fbb4e2e63028ada7c.js
├── 1d486a6abe13f93fbb4e2e63028ada7c.map
└── index.html
```

`$ cat dist/index.html`
```html
<script src="/dist/1d486a6abe13f93fbb4e2e63028ada7c.js"></script>
```

Fantastic!

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,7 +50,6 @@
     "posthtml-parser": "^0.4.0",
     "posthtml-render": "^1.1.0",
     "resolve": "^1.4.0",
-    "sanitize-filename": "^1.6.1",
     "semver": "^5.4.1",
     "serialize-to-js": "^1.1.1",
     "serve-static": "^1.12.4",
--- a/src/Asset.js
+++ b/src/Asset.js
@@ -4,7 +4,6 @@ const fs = require('./utils/fs');
 const objectHash = require('./utils/objectHash');
 const md5 = require('./utils/md5');
 const isURL = require('./utils/is-url');
-const sanitizeFilename = require('sanitize-filename');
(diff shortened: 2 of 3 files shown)


devongovett:
Thanks. Could you remove `sanitize-filename` from package.json and yarn.lock as well? Just run `yarn remove sanitize-filename` and commit the changes.

pierredavidbelanger:
Indeed, you are right, I totally forgot this.

I just ran it (`yarn remove sanitize-filename`), but this yield ***way*** more changes than what I expected. 

It seems like `yarn` re-formatted all of `package.json` and `yarn.lock`.

I will commit anyways, but I am suspicious.

PR #936, opened 2018-03-02
Use minified version of prelude only if it exists ...

pierredavidbelanger:
... which does not when using the src only, and this leads to:

```
(node:56732) UnhandledPromiseRejectionWarning: Error: ENOENT: no such file or directory, open '/Users/pierredavidbelanger/Dev/parcel/src/builtins/prelude.min.js'
    at Object.fs.openSync (fs.js:663:18)
    at Object.fs.readFileSync (fs.js:568:33)
    at Object.<anonymous> (/Users/pierredavidbelanger/Dev/parcel/src/packagers/JSPackager.js:12:6)
    at Module._compile (/Users/pierredavidbelanger/Dev/parcel/node_modules/v8-compile-cache/v8-compile-cache.js:178:30)
    at Object.Module._extensions..js (module.js:671:10)
    at Module.load (module.js:573:32)
    at tryModuleLoad (module.js:513:12)
    at Function.Module._load (module.js:505:3)
    at Module.require (module.js:604:17)
    at require (/Users/pierredavidbelanger/Dev/parcel/node_modules/v8-compile-cache/v8-compile-cache.js:159:20)
```

With this PR, one can use the fresh out-of-the-box parcel source to build.

### DIFF
--- a/src/packagers/JSPackager.js
+++ b/src/packagers/JSPackager.js
@@ -4,15 +4,16 @@ const Packager = require('./Packager');
 const urlJoin = require('../utils/urlJoin');
 const lineCounter = require('../utils/lineCounter');
 
-const prelude = {
-  source: fs
-    .readFileSync(path.join(__dirname, '../builtins/prelude.js'), 'utf8')
-    .trim(),
-  minified: fs
-    .readFileSync(path.join(__dirname, '../builtins/prelude.min.js'), 'utf8')
+const prelude = {};
+const preludePath = path.join(__dirname, '../builtins/prelude.js');
+prelude.minified = prelude.source = fs.readFileSync(preludePath, 'utf8').trim();
+const preludeMinPath = path.join(__dirname, '../builtins/prelude.min.js');
+if (fs.existsSync(preludeMinPath)) {
+  prelude.minified = fs
+    .readFileSync(preludeMinPath, 'utf8')
     .trim()
-    .replace(/;$/, '')
-};
+    .replace(/;$/, '');
+}
 
 class JSPackager extends Packager {
   async start() {


DeMoorJasper:
This should never be an issue, u can use `yarn build` to build the minified prelude, this always is the case in the published version

pierredavidbelanger:
Can I disagree ?

It can be really useful not to have to build parcel when using a dev version of it.

For example, with a `package.json` like this (a fresh local clone)

```json
{
  "dependencies": {
    "parcel-bundler": "file:../parcel"
  }
}
```

or even more, like this (a github clone):

```json
{
  "dependencies": {
    "parcel-bundler": "pierredavidbelanger/parcel#mybranch"
  }
}
```

This PR fixes the only thing that prevent someone from doing it :)

DeMoorJasper:
Looks very similar to #620 and if people use it as a dependency from github it should automatically be built  by yarn?

pierredavidbelanger:
@DeMoorJasper no, here on my dev machine at least, the github dependency is not automatically built, nor is the local dependency, even if my projet uses `yarn` instead of `npm`.

Whatever I do, `prelude.min.js` is never present after a `npm install` in my project.

```sh
$ yarn --version
1.0.2
$ npm --version
5.6.0
$ node --version
v9.3.0
```

DeMoorJasper:
Sorry for the confusion, I found this issue, and apparently it doesn't happen for some reason. Thought it happened as it does happen after a local install, would make sense... as that's what the prepublish script is meant for. https://github.com/npm/npm/issues/3055
Not really a fan of this solution but I suppose there is no other way around it

pierredavidbelanger:
@DeMoorJasper I agree this feels like a patch that is useful only when using a dev (unpublished) version of parcel. Though, I prefer this solution over having to commit a built artifact (`prelude.min.js`)

BTW. I certainly can live without this. It is your call after all.

pierredavidbelanger:
An other (less smelly than this PR ?) option would be to always minify (and maybe cache) prelude.js at runtime.

Though this solution will requires a runtime dependency on uglifyjs.

DeMoorJasper:
@pierredavidbelanger uglify is already a dependency as we use it to minify production assets, minifying the file if not found would be possibly result in smaller packages for people who depend on the github version and possibly be better than this fix, but it would add even more complexity.

pierredavidbelanger:
@DeMoorJasper you can have a look at #937 . I tried what we discussed here (runtime minified prelude). You can now close one or the other, or both :)

DeMoorJasper:
@pierredavidbelanger awesome, i’ll let that decision to Devon. I was just trying to find a better solution :)

jkrehm:
I just ran into this issue today and can give what I feel is a valid use case. While I'm waiting for a new release and for a couple of PR's to be merged, I'm using a git branch and have added it to my package.json via Yarn. To make this work I had to build and commit the prelude.min.js file to that branch because otherwise it resulted in the "no such file" error. Sure it's a bit of an edge case, but it's a pain point and can be remediated pretty easily.

gnijuohz:
I feel this is a valid use case as well. Some users are installing it from GitHub and may get confused when `npm/yarn install` alone doesn't work as they expected, for example, see the comments in #987.

At least, maybe we could add a note in the readme for people who install from github directly? (not ideal, I know...)

davidnagli:
I agree with @gnijuohz. We should just update `CONTRIBUTING.md` and/or `README` with instructions for building and installing everything.

Can we close this PR in favor of that?

devongovett:
There is another step in the `build` command, which is to run babel over the source files: https://github.com/parcel-bundler/parcel/blob/master/package.json#L101. This would not be done when installing from git either, so it wouldn't work in Node 6.

I think ideally we'd just publish beta versions to npm much more frequently (like every commit to the master branch maybe?). We could probably automate that with Jenkins. We could even tag the releases so you could `npm install parcel-bundler@next` to get the latest beta version. 

Going to close this PR. If someone wants to work on the above, that would be awesome!

davidnagli:
@devongovett It's actually possible to configure Travis or AppVeyor to do automatic releases, we just need to add a more few lines to our config files.

[Travis NPM Release Docs](https://docs.travis-ci.com/user/deployment/npm/)

I think you're the only one with publish access to the npm package, so you would have to be the one to set it up.


For Travis:
```yml
deploy:
  provider: npm
  email: "YOUR_EMAIL_ADDRESS"
  api_key: "YOUR_AUTH_TOKEN"
  tag: next
```

Kerrick:
Any progress on this? I had a tough time figuring out how to get a build to work after installing via `yarn add --dev parcel-bundler/parcel#master` until I found this thread. Need a bug fix that made it into `master` recently, but the latest version is 20 days ago. Would love to see either automatic beta releases to get features faster, or this PR to be shipped so we can track `master`.

PR #937, opened 2018-03-02
minify prelude at runtime if pre-minified version does not exists

pierredavidbelanger:
As discussed with @DeMoorJasper  in #936 

I compared the CLI minified version with the runtime minified version, they looks identical (indeed, its the same lib).

I hope I did not introduced too much complexity for this rare corner case usage ;)

### DIFF
--- a/src/packagers/JSPackager.js
+++ b/src/packagers/JSPackager.js
@@ -3,16 +3,7 @@ const path = require('path');
 const Packager = require('./Packager');
 const urlJoin = require('../utils/urlJoin');
 const lineCounter = require('../utils/lineCounter');
-
-const prelude = {
-  source: fs
-    .readFileSync(path.join(__dirname, '../builtins/prelude.js'), 'utf8')
-    .trim(),
-  minified: fs
-    .readFileSync(path.join(__dirname, '../builtins/prelude.min.js'), 'utf8')
-    .trim()
-    .replace(/;$/, '')
-};
+const {minify} = require('uglify-es');
 
 class JSPackager extends Packager {
   async start() {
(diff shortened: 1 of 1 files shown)


davidnagli:
I’m personally against adding all this complexity just for this extreme corner case that will never affect real users (only useful for Parcel developers, and a small fraction of those developers at that). 

Can you maybe elaborate on when somebody would want to directly reference the Parcel repo from their npm build?

Because for the most part, you have 2 groups of people:
* Users (people using Parcel) - get Parcel from npm using `yarn add parcel` or `npm install parcel`
* Developers (people developing Parcel) - fork/clone Parcel directly from GitHub

I’m not really seeing why anybody would want to get Parcel from the Github source files **through** NPM. It just doesn’t make sense for developer _or_ users. 

Maybe I’m missing something? I'd love to hear your perspective... 😃

pierredavidbelanger:
@davidnagli Indeed, this PR is too complex , as I feared.

You can have a look at #936 , it solves the same problem, with a definitely simpler patch.

Having that said, I have nothing really to elaborate more, You have correctly summarized why this PR is a rare corner case. I certainly can live without this feature :)

I did it anyways because I ran into the rare use case of wanting to use a feature of parcel in a project, but the feature is still not merged, so I created a branch on my parcel clone and use this branch in my projet (something like this #620 ).

laosb:
@pierredavidbelanger @davidnagli Since Parcel is not releasing fast and has some crucial features (like Vue integration) not merged, I'm forced to use a fork. This minification step caused way too much problem in our CI process. I'm even kind of missing Webpack for it's even sounds easier to just copy-paste some configuration files and start working immediately. I'm getting tired of running into problems that has been fixed in later versions, and I still need to use Vue branch, which is ancient compared to quick development of Parcel itself.

This feature wouldn't be this important if Parcel try to release faster (with fewer commits introduced in a release), because if we really do so, bugs can be found more quickly and people will be excited to get new features quicker. I'm seeing many people can't wait for some features like the new path resolver. One minor release with a new feature also means, if they found the feature buggy, they can safely and easily revert to use a previous version without worrying about losing another working feature, and update to a patch release when a fix is available.

PR #938, opened 2018-03-02
Load plugins concurrently

fathyb:
**Before**: 
- Plugins are loaded sequentially
- When a plugin throws it warns the user and then stop loading other plugins

**After**:
- Plugins are loaded concurrently
- When a plugin throws it skips it, warns the user and continues to load other plugins


davidnagli:
We plan to make some pretty big changes to how plugins work soon... See: #918 #919

PR #944, opened 2018-03-05
Use const instead of var

limichange:


devongovett:
Sorry, but I don't think this really improves the code. `const` is for constants. If anything we should use `let`, but I don't really see a reason to change it.

PR #955, opened 2018-03-06
Allow handling of external modules and scripts

derolf:
This patch allows two related things for handling externals:

# Externals in package.json

A package.json might now have an external section. An external module will not be bundled, but rather a module stub generated that (re)exports the global variable.

Example:

```
{
 ...
 "externals": {
    "cesium": "Cesium"
  }
}
```
This example will create a module `cesium` that exports `Cesium`.

# External scripts in URLs

If an dependency URL (e.g. in HTML) starts with a forward slash `/`, the reference is treated to be external and to point to a file in the output directory. The URL is then prefixed with the public URL.

Example:
```
<html>
<body>
  <script src="/lib/Cesium/Cesium.js"></script>
</body>
</html>
```

Here, Parcel will assume that there exists a file `lib/Cesium/Cesium.js` in the output directory. The file will not be bundled. But, rather Parcel will warn if the file does NOT exist.

### DIFF
--- a/src/Asset.js
+++ b/src/Asset.js
@@ -5,6 +5,7 @@ const objectHash = require('./utils/objectHash');
 const md5 = require('./utils/md5');
 const isURL = require('./utils/is-url');
 const config = require('./utils/config');
+const logger = require('./Logger');
 
 let ASSET_ID = 1;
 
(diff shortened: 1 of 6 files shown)


src/Asset.js
```diff
@@ -81,15 +82,24 @@ class Asset {
...
+    // external src?
+    if (parsed.pathname.startsWith('/')) {
+      // warn if the file does not exist
+      if (!fs.existsSync(path.join(this.options.outDir, parsed.pathname))) {
+        logger.warn(`'${from}: Cannot find '${parsed.pathname}'`);
+      }
+    } else {
+      const resolved = path.resolve(path.dirname(from), parsed.pathname);
+      this.addDependency(
+        './' + path.relative(path.dirname(this.name), resolved),
```
davidnagli: Won’t the `/` cause issues on Windows?

derolf: That has not been changed by me. It was already like this in the original Asset.js.

derolf: Ready to merge? :-)

danielpa9708:
I'd really like this feature and I'm a bit anxius about it but seems to have 12 days stuck, what's the status on this PR ?

devongovett:
We cannot rely on absolute paths being external. Absolute paths are valid and can be resolved by the new Parcel resolver relative to the project directory. Also I'm not sure why you'd need those files to not be processed by Parcel anyway: I can see wanting external URLs to point to a CDN but then those wouldn't be absolute paths on your own webserver probably.

derolf:
Let’s say you copy prebuilt scripts to your output folder and just want to reference them there.

Right now, Parcel does not allow “local” external references, but only to CDN.

derolf:
So what’s your proposal to reference local external scripts (or other files) you do not want to be processed by parcel.

If you make a good proposal, I will adopt. Otherwise, I will just abandon this PR.

PR #960, opened 2018-03-08
Feature/open browser

shrynx:
[**edited**]
Parcel cli has an option `--open`, which opens default browser.
There are two aspects that i would like to change

1. Allow providing custom browser option (eg: “parcel  ...  --open ‘firefox’ “)
2. Try best to re use existing open tab.

The first one is fairly simple, for the second one, ~~i have seen a reasonable solution in create-react-app, they have a custom applescript, which on macos for only chrome, tries (and in most cases successfully does) reusing existing tab.~~
~~https://github.com/facebook/create-react-app/blob/next/packages/react-dev-utils/openChrome.applescript~~

i implemented a macOS automation script written in javascript to re use chrome tab in macOS.
This has a benifit that we donot introduce applescript, but the catch is this would be supported on macOS version >= 10.10 ([refer](https://developer.apple.com/library/content/releasenotes/InterapplicationCommunication/RN-JavaScriptForAutomation/Articles/OSX10-10.html#//apple_ref/doc/uid/TP40014508-CH109-SW1))

This PR implements both 1 and 2.
I implemented them in separate commits to make it easy to cherry pick as i am not sure if it is in scope of parcel.

Additionally, can also log which browser the cli is opening it in.
something like (`opening ${url} in ${browser} ...`)

### DIFF
--- a/src/openInBrowser.js
+++ b/src/openInBrowser.js
@@ -1,0 +1,14 @@
+const opn = require('opn');
+
+const openInBrowser = (url, browser) => {
+  try {
+    const options = typeof browser === 'string' ? {app: browser} : undefined;
+
+    opn(url, options).catch(() => {}); // Prevent `unhandledRejection` error.
+    return true;
+  } catch (err) {
+    return false;
+  }
+};
+
+module.exports = openInBrowser;
--- a/src/openInBrowser.js
+++ b/src/openInBrowser.js
@@ -1,8 +1,37 @@
 const opn = require('opn');
+const execSync = require('child_process').execSync;
 
 const openInBrowser = (url, browser) => {
+  const OSX_CHROME = 'google chrome';
+
+  const isBrowserString = typeof browser === 'string';
+
+  const shouldTryOpenChromeWithAppleScript =
+    process.platform === 'darwin' &&
+    (!isBrowserString || browser === OSX_CHROME);
+
+  if (shouldTryOpenChromeWithAppleScript) {
+    try {
+      // check if chrome is running
+      execSync('ps cax | grep "Google Chrome"');
+      const command = `osascript -l JavaScript reuseChromeTabInMacOS.js ${encodeURI(
+        url
+      )}`;
+      // try our best to reuse existing tab
+      execSync(command, {
+        cwd: __dirname,
+        stdio: 'ignore'
+      });
+      return true;
+    } catch (err) {
+      // Ignore errors.
+    }
+  }
+
+  // Fallback to opn
+  // (It will always open new tab)
   try {
-    const options = typeof browser === 'string' ? {app: browser} : undefined;
+    const options = isBrowserString ? {app: browser} : undefined;
 
     opn(url, options).catch(() => {}); // Prevent `unhandledRejection` error.
     return true;
--- a/src/cli.js
+++ b/src/cli.js
@@ -23,7 +23,10 @@ program
     'set the hostname of HMR websockets, defaults to location.hostname of current window'
   )
   .option('--https', 'serves files over HTTPS')
-  .option('--open', 'automatically open in default browser')
+  .option(
+    '--open [browser]',
+    'automatically open in specified browser, defaults to default browser'
+  )
   .option(
     '-d, --out-dir <path>',
     'set the output directory. defaults to "dist"'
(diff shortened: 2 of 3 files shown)


davidnagli:
👍 I agree your first point would actually be a pretty useful feature. I often test websites in a browser that’s not my default, and this would certainly make that much easier.

Not a really big fan of you second point though :/

A few questions:

* Will your implementation still support the old behavior of just using `--open` without any arguments? It’s important that we make sure that it still works the other way too.
* Also, is it possible to accomplish your second point without AppleScript?

shrynx:
@davidnagli 
1. yes , passing only `--open` without a specific browser name works the same way, opening default browser.
2. i do re run parcel command often and end up having a lots of open tabs (which i personally find annoying), but i do agree the implementation for 2 is not satisfying as it just works for chrome on osx and that too requires applescript.
At this time, I don't have a better approach for it. i'll dig around a little more though.

shrynx:
@davidnagli i changed the apple script to javascript, having the same behavior.

shrynx:
😩  sorry for the multiple ci builds, i was working while traveling and kept pushing every little change i did, not realizing that i am triggering ci rebuilds.

src/openInBrowser.js
```diff
@@ -0,0 +1,43 @@
...
+
+  const shouldTryOpenChromeWithAppleScript =
+    process.platform === 'darwin' &&
+    (!isBrowserString || browser === OSX_CHROME);
+
+  if (shouldTryOpenChromeWithAppleScript) {
+    try {
+      // check if chrome is running
+      execSync('ps cax | grep "Google Chrome"');
+      const command = `osascript -l JavaScript reuseChromeTabInMacOS.js ${encodeURI(
```
davidnagli: This is still using AppleScript (even though it’s in JavaScript), so this won’t work on other platforms etc.

I would much rather have us implement a nice and neat cross-platform solution that works for everyone. Could we maybe try to implement this using some kind of “polling” on the HMR client that starts when the connection get’s closed? That way we can just check for those incoming connections to the HRR server to determine if a tab is already open.

To clarify: My idea is that when the connection to the HMR server get’s closed, the HMR client should keep trying to reconnect or at least sending some kind or request. Then, we can just prevent the “open” feature from going off whenever we see these incoming connections.

shrynx: sorry, maybe i forgot to mention this is just a work around specifically for macOS on chrome.
i am not sure how to implement your suggestion and also if focusing on the correct browser window/tab has a pure js cross platform solution.
i liked this feature from create-react-app and was trying to replicate it.

i can look into into it later though.

shrynx: @davidnagli I searched more about it. 
- Just refreshing disconnected connection won't be enough, the browser window or the tab maybe in background.
Bringing the window and tab to front is important otherwise user might not know that the tab has been refreshed and will eventually open a new tab, thus leading to bad UX/DX.
- Supporting other browser and platform
  - MacOS
    - Applescript or mac os automation by javascript (**not using applescript**) can help bringing tab to focus.
    - Chrome is quite scriptable, thus we can provide this feature.
    - Firefox doesn't expose tabs api.
    - Safari does expose tab api but no way to bring it to front (activate). or directly reloading it
  - Windows
    - Can't automate chrome or firefox with vbscript
    - IE automation is possible, but i didn't look much into it.
  - Linux
    - Don't have access to a linux machine, maybe someone can look into.

I completely agree with you that the current solution is too specific (chrome on mac), but isn't it better to give atleast a few dev a good DX  than having no solution at all. 
If and when someone figure out a better way, this implementation can get replaced.

davidnagli: How about we split these features into two seperate PRs.

It looks like the `--open [browser]` feature looks ready, and I’m a huge fan of it so I’d love to see that get merged as soon as possible. But I think we should discuss the other feature a bit further to work out the kinks.

Do you mind moving the “tab reuse feature” into a separate PR so we can discuss it separately from the “browser selection feature"?

shrynx: Sounds good. i'll make a PR soon

shrynx:
a pull request https://github.com/parcel-bundler/parcel/pull/978 for just customizing `--open` to support custom browser option.

@davidnagli should we close this one ? and revisit it once https://github.com/parcel-bundler/parcel/pull/978 is merged ?

PR #966, opened 2018-03-08
Add support for specifying SSL certificate & key

jkrehm:
Adds CLI options to specify `--cert` and `--key` alongside `--https`. If they are set and the files found, they are used. If they are not set but `--https` is, then the `generateCertificate` logic is called. If `--https` is not specified, then an insecure server will start.


jkrehm:
Anything I can do to improve this PR and increase its chances of making it into the project?

src/utils/getCertificate.js
```diff
@@ -0,0 +1,12 @@
+const fs = require('fs');
+
+function getCertificate(options) {
+  const key = fs.existsSync(options.key) && fs.readFileSync(options.key);
```
DeMoorJasper: Perhaps use async reads here?

jkrehm: I can add that. I patterned its behavior after `generateCertificate` which uses sync calls. I'll update the PR to use `utils/fs.js` and `await`.

jkrehm:
Changes made to load the files asynchronously.

devongovett:
Updated to make the `--https` option not necessary when using `--cert` and --key` - we can infer https from those keys.

jkrehm:
Thanks!

PR #971, opened 2018-03-09
Set TypeScript option esModuleInterop to true by default

### LINKED ISSUE #418: 🐛 Upgrading from 1.2.0 to 1.3.0 breaks moment.js

fathyb:
This will make TypeScript behave more like the Babel CommonJS plugin. See [TypeScript 2.7 release notes](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-7.html) for more info. It'll be ignored if Typescript < 2.7 is used. Users can still opt-out by setting it to false in their `tsconfig.json`.

Fixes #418 and is a better alternative to #530 for TypeScript users.

Please note that it may break existing code bases. From the release note :
> Note: The new behavior is added under a flag to avoid unwarranted breaks to existing code bases. We highly recommend applying it both to new and existing projects. For existing projects, namespace imports (import * as express from "express"; express();) will need to be converted to default imports (import express from "express"; express();).


fathyb:
@DeMoorJasper any thoughts on also setting `importHelpers` to `true` by default?

DeMoorJasper:
@fathyb won't babel catch most of the exotic syntaxes?
Not sure as i've never used importHelpers, but i guess it should be fine as is

PR #974, opened 2018-03-09
HTML bundle serving bug.

### LINKED ISSUE #782: 🐛 publicURL with development server

ry:
When serving HTML bundle with publicUrl="/", allow the server to send static assets other than html.

Fixes #782

### DIFF
--- a/src/Server.js
+++ b/src/Server.js
@@ -51,6 +51,11 @@ function middleware(bundler) {
       } else {
         // Otherwise, serve the file from the dist folder
         req.url = req.url.slice(bundler.options.publicURL.length);
+        // If we're serving an html bundle, and the requested path has not file
+        // extension, serve the index.html. Otherwise go to the file server.
+        if (bundler.mainAsset.type === 'html' && path.extname(req.url) == "") {
+          return sendIndex();
+        }
         return serve(req, res, send404);
       }
     }
--- a/src/Server.js
+++ b/src/Server.js
@@ -45,17 +45,16 @@ function middleware(bundler) {
     function respond() {
       if (bundler.errored) {
         return send500();
-      } else if (!req.url.startsWith(bundler.options.publicURL)) {
-        // If the URL doesn't start with the public path, send the main HTML bundle
+      } else if (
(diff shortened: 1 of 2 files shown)


src/Server.js
```diff
@@ -49,6 +50,11 @@ function middleware(bundler) {
       } else {
         // Otherwise, serve the file from the dist folder
         req.url = req.url.slice(bundler.options.publicURL.length);
+        // If we're serving an html bundle, and the requested path has not file
+        // extension, serve the index.html. Otherwise go to the file server.
+        if (bundler.mainAsset.type === 'html' && path.extname(req.url) == "") {
```
DeMoorJasper: This would break any file that doesn't have an extension?

ry: Currently only HTML is served, so it wouldn't change the behavior. But it's true that it will not serve files without extensions - but I'd argue that is desired behavior for single page apps.

davidnagli:
@ry @DeMoorJasper Is this a duplicate of #838?

I noticed that #838 closes this PR 🤔 🤷‍♂️

devongovett:
Thanks!

src/Server.js
pulpdood: I need to verify my domain for Apple Pay for my frontend Stripe Elements app, and this requires serving a static file without a file extension, so there is a use case for this.

PR #978, opened 2018-03-11
customize open in browser feature

shrynx:
This PR allows cli command `--open` to take an optional argument `--open [browser]` to open in a specific browser.

If browser is not specified then it i will open in the default browser.


src/openInBrowser.js
```diff
@@ -0,0 +1,14 @@
+const opn = require('opn');
+
+const openInBrowser = (url, browser) => {
+  try {
+    const options = typeof browser === 'string' ? {app: browser} : undefined;
```
davidnagli: More readable alternative:
```javascript
const options = {app: browser || ''};
```

src/openInBrowser.js
```diff
@@ -0,0 +1,14 @@
...
+
+    opn(url, options).catch(() => {}); // Prevent `unhandledRejection` error.
```
davidnagli: If you make `openInBrowser` an async function you can await `opn`, so that any errors will get caught in the `try/catch`.

src/openInBrowser.js
```diff
@@ -0,0 +1,14 @@
...
+    return true;
+  } catch (err) {
+    return false;
```
davidnagli: Don’t you think it would be better if we told the user about the error? We probably can’t use the Logger class since it hasn't been initialized yet, but we can at least do something simple like:

```javascript
console.error('Unexpected error when opening the browser:', err)
```

That way we don't throw (which would prevent bundling), but we still inform the user about the error.

src/openInBrowser.js
shrynx: the problem is when we pass `--open` without any browser, the `command.open` passed here as `browser` takes on value as `true`. therefore i did a simple type check to see if it is a string or boolean.
This was because passing `{app: true}` to open , doesn't do anything at all.

src/openInBrowser.js
shrynx: ahhh right, i keep forgetting source is transpiled, i'll make the fix

src/openInBrowser.js
shrynx: sure sound good.

shrynx:
@davidnagli Sorry for the silence, was on vacation.
I replied to your comments and updated the pr.

src/openInBrowser.js
```diff
@@ -0,0 +1,14 @@
+const opn = require('opn');
```
devongovett: could you move this file to `utils/openBrowser.js`? Thanks!

shrynx: @devongovett sure., updated the PR.

shrynx:
Thanks for merging.
@davidnagli as for reusing same tab, should i open i new PR for discussion or reopen the [old one](https://github.com/parcel-bundler/parcel/pull/960) ?

davidnagli:
Let's open a new PR with only the reuse tab feature 👍

shrynx:
done https://github.com/parcel-bundler/parcel/pull/1031

FatehAK:
Is it possible to supply addition arguments as well such as --disable-extension or --remote-debugging-port=9222 ?

Issue #758, opened 2018-02-06
🐛 Code splitting should duplicate modules used by web workers


bwrrp:
**Choose one:** this is a 🐛 bug report

Web workers run in an entirely separate JS context. It seems that the current approach treats a `new Worker(path)` expression as a normal code splitting point. The result of this is that modules shared by multiple workers end up in the main bundle instead of in each worker's `.js` file.

### 🎛 Configuration (.babelrc, package.json, cli command)

Seen on a very minimal project, see example code below. Create those files, install parcel-bundler and run `npx parcel index.html --no-hmr`. Disabling HMR is not necessary to trigger this behavior, but it makes the bundled code a bit easier to read.

### 🤔 Expected Behavior

I expected the code from `shared.js` to end up duplicated in each web worker's .js bundle.

### 😯 Current Behavior

The code for `shared.js` is included in the `parcel.js` bundle running in the web page context. The bundles generated for `w1.js` and `w2.js` contain the require call for './shared', but do not contain the module itself. This results in the following error being thrown in each of the workers:

```
Uncaught Error: Cannot find module '12'
newRequire	@	d2f19dd….js:33
localRequire	@	d2f19dd….js:48
require.9../shared	@	w2.js:1
newRequire	@	d2f19dd….js:42
require.9../shared	@	d2f19dd….js:67
(anonymous)	@	d2f19dd….js:72
```

### 💁 Possible Solution

Workers should probably considered a special case in code splitting, as all modules required by one should actually be duplicated into the resulting bundle.

### 🔦 Context

I'm building an application that requires communication between several web workers. These workers have different responsibilities, and therefore a different entry point, but do share large parts of the codebase. I don't really want to maintain a separate copy of these shared modules for each worker.

### 💻 Code Sample

*index.html*
```html
<script src="index.js"></script>
```

*index.js*
```js
new Worker("./w1.js");
new Worker("./w2.js");
```

*w1.js*
```js
import shared from "./shared";
console.log("w1", shared);
```

*w2.js*
```js
import shared from "./shared";
console.log("w2", shared);
```

*shared.js*
```js
export default { test: 123 }
```

### 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Parcel           | 5.6.0
| Node             | 8.1.3
| npm         | 5.6.0
| Operating System | Linux

jsilveira:
[...]

PR #979, opened 2018-03-11
Web Worker bundle does not include all dependencies

dhcode:
### What I did
I extended the Worker test in javascript.js to show that a common dependency, that is required by the index.js and the worker.js should be contained in the worker bundle, too.

### Story
PR #441 added support for web workers with `asset.addURLDependency`. But it didn't take into account, that the `Bundler.createBundleTree` optimizes the tree and possibly moves common assets to another bundle.

### Requirement
The worker bundle must have a copy of all assets, because it runs in the browser in a different context and can't access what is provided by other bundles.

### What I tried
Unfortunately my tries to fiddle with the Bundler to support this were not successful and depending on the order of imports. That is why there is a second test 'should support bundling workers with different order'.

Maybe someone more familiar with how parcel works, can suggest a solution?

### DIFF
--- a/src/Bundle.js
+++ b/src/Bundle.js
@@ -91,6 +91,14 @@ class Bundle {
     return this.assets.size === 0;
   }
 
+  get isIsolated() {
+    return (
+      this.entryAsset &&
+      this.entryAsset.parentDeps &&
+      Array.from(this.entryAsset.parentDeps.values()).some(dep => dep.isolated)
+    );
+  }
+
   getBundleNameMap(contentHash, hashes = new Map()) {
     let hashedName = this.getHashedBundleName(contentHash);
     hashes.set(Path.basename(this.name), hashedName);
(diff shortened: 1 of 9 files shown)


davidnagli:
cc @DeMoorJasper

dhcode:
Related issues: #758 and webpack/webpack/issues/6472

dhcode:
I implemented now an isolated flag for the dependencies. If the entry asset of a bundle should be isolated, it is handled now in a way that common assets can be in multiple bundles.
* Common dependencies are not removed from the isolated bundle.
* Common dependencies are added to the isolated bundle.

Yes, it can produce duplicate assets in the isolated bundle for now. But it was easier for me now, than supporting importScripts in the js-loader.js and somehow generating a commons bundle, that is loaded by multiple bundles.

The CI builds seem to be unstable at the moment and I guess it has nothing todo with my changes?

DeMoorJasper:
@dhcode The tests are indeed pretty instable, as long as they pass locally you should be fine

devongovett:
Closing in favor of #1257 which is based on this PR.

PR #981, opened 2018-03-12
Fix builtin loaders with --target=node

### LINKED ISSUE #969: 🐛 Fetch API is included in the codes generated by `parcel build -t node`
This a 🐛 bug report.

#### `/index.js`
```js
import { add } from './math.rs'

console.log(add(1, 2))

```

#### `/math.rs`
```rust
#[no_mangle]
pub fn add(a: i32, b: i32) -> i32 {
    a + b
}
```

### 🤔 Expected Behavior
```bash
$ parcel build -t node -o dist/bundle.js index.js
$ node dist/bundle.js
3
```

### 😯 Current Behavior
```bash
$ parcel build -t node -o dist/bundle.js index.js
$ node dist/bundle.js
/Users/mizdra/src/gomi/parcel-wasm-for-node/dist/bundle.js:8
module.exports=function(t){return fetch(t).then(function(t){return WebAssembly.instantiateStreaming?WebAssembly.instantiateStreaming(t):t.arrayBuffer().then(function(t){return WebAssembly.instantiate(t)})}).then(function(t){return t.instance.exports})};
                           ^

ReferenceError: fetch is not defined
    at module.exports (/Users/mizdra/src/gomi/parcel-wasm-for-node/dist/bundle.js:8:28)
    at u (/Users/mizdra/src/gomi/parcel-wasm-for-node/dist/bundle.js:6:610)
    at Array.map (<anonymous>)
    at Function.t [as load] (/Users/mizdra/src/gomi/parcel-wasm-for-node/dist/bundle.js:6:303)
    at Object.require.0 (/Users/mizdra/src/gomi/parcel-wasm-for-node/dist/bundle.js:10:50)
    at t (/Users/mizdra/src/gomi/parcel-wasm-for-node/dist/bundle.js:1:408)
    at /Users/mizdra/src/gomi/parcel-wasm-for-node/dist/bundle.js:1:646
    at Object.<anonymous> (/Users/mizdra/src/gomi/parcel-wasm-for-node/dist/bundle.js:1:663)
    at Module._compile (module.js:660:30)
    at Object.Module._extensions..js (module.js:671:10)
```

### 💁 Possible Solution
- Use `fs.readFile` instead of `fetch`
- Use `node-fetch`

### 🔦 Context
I want to test `.wasm` with unit testing frameworks on Node.js and run on browsers. 

### 💻 Code Sample
[https://gist.github.com/mizdra/e71b62f0ee839c8ad58ef679f2db747b](https://gist.github.com/mizdra/e71b62f0ee839c8ad58ef679f2db747b)

### 🌍 Your Environment

| Parcel           | 1.6.2 |
| Node             | 9.5.0 |
| npm/Yarn         | 5.6.0/1.3.2 |
| Operating System | macOS Sierra 10.12.6 |

fathyb:
Fixes #969.

Add support for dynamic imports, WebAssembly and Rust with `--target=node`.

cc @lbguilherme @albizures 

- [ ] the `global` modifications on `prelude.js` are ugly, any suggestion? Maybe make a Node.js prelude and a browser one?


src/builtins/loaders/node/js-loader.js
```diff
@@ -0,0 +1,9 @@
+module.exports = function loadJSBundle(bundle) {
+  return new Promise(function(resolve) {
+    // require in the next event loop tick to let the main module load
+    setTimeout(function() {
+      require(__dirname + bundle)
+      resolve()
```
lbguilherme: Should resolve to the return of require. `resolve(require(__dirname + bundle))`

fathyb: Are you sure? The original `js-loader.js` doesn't resolve anything : 
https://github.com/parcel-bundler/parcel/blob/0dc062f77e36a19297dac542be33cded04580e89/src/builtins/loaders/js-loader.js#L13-L16

If `require` is called here it's just to register the `__parcel_require` so next `require` calls can require numeric IDs here : 

https://github.com/parcel-bundler/parcel/blob/0dc062f77e36a19297dac542be33cded04580e89/src/builtins/bundle-loader.js#L27-L30

I'm not very familiar with that part though so I may have missed something 😅

devongovett:
Instead of introducing another global to store the node require function, I just renamed the parcel require function to `parcelRequire` instead of `require` so it doesn't conflict with node's require function. This should also fix #761, so that parcel doesn't conflict with the `require` function exposed by other bundlers or scripts like require.js which might also be on the page. We fall back to checking `require` if we cannot find the module in parcel's require system.

src/Bundler.js
```diff
@@ -37,15 +37,13 @@ class Bundler extends EventEmitter {
...
-    this.addBundleLoader(
-      'wasm',
-      require.resolve('./builtins/loaders/wasm-loader')
-    );
-    this.addBundleLoader(
-      'css',
-      require.resolve('./builtins/loaders/css-loader')
-    );
-    this.addBundleLoader('js', require.resolve('./builtins/loaders/js-loader'));
+    const loadersPath = `./builtins/loaders/${
```
devongovett: would be cool to add support for registering bundle loaders for specific environments. this would allow third party plugins to do the same:

```javascript
bundler.addBundleLoader('wasm', {
  node: require.resolve('./node/wasm-loader'),
  browser: require.resolve('./browser/wasm-loader')
});
```

Want to make a followup PR to support that?

PR #1004, opened 2018-03-15
Exit when STDIN closes

rzane:
This PR adds a new flag to the watch command that instructs Parcel to exit when STDIN closes. This is consistent with the behavior in Webpack and Brunch.

Here's a really clear synopsis from [this Brunch issue](https://github.com/brunch/brunch/issues/920):
> I would like to automatically run `parcel watch` as part of my application but, when my application shuts down, the `parcel` process persists and continue running. I have observed that my application does close STDIN, however, brunch watch is not checking if STDIN was closed or not. These two lines solve the issue:

```javascript
process.stdin.on('end', () => process.exit(0));
process.stdin.resume();
```

> I am running on the Erlang VM which does not expose the OS Process API as node.js does (in fact, neither does the Java VM last time I checked. It is common for VMs to not expose those operations as it is hard to provide an unified API across multiple OS and hardware).

As such, it is common for Unix processes to exit when STDIN closes.

Relevant links:

* https://github.com/webpack/webpack/pull/1311
* https://github.com/brunch/brunch/issues/920
* https://github.com/brunch/brunch/pull/1003


rzane:
Force pushed to trigger a CI rebuild. I suspect that the Node 6 tests are flaking.

davidnagli:
@rzane Ya, we have some flakiness issues with Node 6 tests

rzane:
I am writing a [Phoenix application](http://phoenixframework.org/). Phoenix is a web framework written in Elixir.

Elixir doesn't have a way to send a kill signal to a process. It is common for VMs to not expose those operations as it is hard to provide an unified API across multiple OS and hardware. As such, it is quite common for UNIX tools to exit when STDIN closes.

Specifically, my problem is that Phoenix provides the ability to start an asset bundler when the server starts. And, when the Phoenix server shuts down, it closes STDIN. Unfortunately, because parcel doesn't exit when STDIN closes, parcel doesn't die. It just stays alive.

As mentioned in the original description, Webpack and Brunch currently offer this functionality.

devongovett:
hmm, ok. Is there any reason to not make this default behavior instead of adding an option?

rzane:
Yes. If you run parcel with `--stdin`, then press `Ctrl+z` to suspend, then `bg` to run in the background, parcel will exit. Since this could affect the workflow of current parcel users, it's probably wise to put this functionality under a flag.

Here's the issue report of that specific problem: https://github.com/brunch/brunch/issues/998.

rzane:
I think @dparnell's plugin is probably an adequate and simple solution for this specific problem.

devongovett:
You could also make a small wrapper script like this:

```sh
#!/bin/sh
"$@" &
pid=$!
while read line ; do
  :
done
kill -KILL $pid
```

https://hexdocs.pm/elixir/Port.html#module-zombie-processes

I'm going to close this since it is possible to work around. I don't think this is Parcel's problem.

Issue #793, opened 2018-02-12
🐛 Incorrect module resolution when one import name matches along the parent path (only on `build`)


jsiebern:
### 🎛 Configuration (.babelrc, package.json, cli command)

```js
{
  "scripts": {
    "build": "parcel build src/index.js"
  }
}
```

```js

{
    "presets": [
        [
            "env",
            {
                "modules": false
            }
        ]
    ]
}
```

I have a directory structure of imports like so:
```
index.js
sub/test/index.js
sub/test/tabs.js
sub/test.landscape/index.js
sub/test.landscape/tabs.js
```

The top `index.js` imports both lower `index.js` and each of them have the content:
```js
export { default } from './tabs';
```

### 🤔 Expected Behavior

In the outputted file I expect the ids for `"./sub/test"` and `"./sub/test.landscape"` to be different, wich they even are when *not* using the `build` command.

For example:
```js
{"./sub/test":3,"./sub/test.landscape":4}
```

### 😯 Current Behavior

Using the build command it outputs:
```js
{"./sub/test":2,"./sub/test.landscape":2}
```

So the same id, that causes my code to break in production where in development it runs fine.

### 💁 Possible Solution

I only had a brief glance at the code debugging this, it might have something todo with the common ancestor search? Totally not sure on that one, sorry.

### 🔦 Context

As this is only a "beautification" for filenames (putting the main file of the subfolder as `tabs.tsx` for example instead of `index.tsx`, its not a big deal to work around. I still believe it to be a bug though.

### 💻 Code Sample

I have attached some example code, just run `yarn; yarn build` and check the `dist/index.js` for the faulty result.

[error_example.zip](https://github.com/parcel-bundler/parcel/files/1716549/error_example.zip)

jeyj0:
Hi, I was just looking into this, and tried to reproduce it using your given example. 
* Using the Version in the example I can indeed verify that there is the described, buggy behavior. 
* However, using the newest version of parcel-bundler (1.9.7) I got: `{"./sub/test":"C2SW","./sub/test.landscape":"BZFw"}` which seems to be correct as far as I understand the issue.

I would therefore suggest closing this issue as being resolved. :smile:

PR #1011, opened 2018-03-17
fixes #933 : JSPackager deduplication now accounts for differences in absolute dependency paths

### LINKED ISSUE #933: 🐛Cache issue with identical files with relative import
It seems like there is a cache issue when you have two identical files in two different folder, but they have relative import path. As an example, I have `folder1/subFolder/index.js` and `folder2/subFolder/index.js` importing `../index.js`. However, `folder1/index.js` and `folder2/index.js` are two different files. 

It looks like Parcel is thinking they are both the same but they are not because of the relative path.

### 😯 Current Behavior

Here is a proof of concept: https://github.com/exon/parcel-cache-bug

If you build and run it, it will print:
```
FOLDER1!!!
FOLDER3!!!
```

### 🤔 Expected Behavior

However, the expected result were:
```
FOLDER1!!!
FOLDER2!!!
FOLDER3!!!
```

### 💁 Possible Solution

I don't know the Parcel cache system but I guess it should transform the import path into something more absolute starting from the project root.

pcattori:
fixes #933 #1152

### DIFF
--- a/src/packagers/JSPackager.js
+++ b/src/packagers/JSPackager.js
@@ -35,13 +35,13 @@ class JSPackager extends Packager {
   }
 
   async addAsset(asset) {
-    if (this.dedupe.has(asset.generated.js)) {
+    if (this.dedupe.has(this.dedupeKey(asset))) {
       return;
     }
 
     // Don't dedupe when HMR is turned on since it messes with the asset ids
     if (!this.options.hmr) {
-      this.dedupe.set(asset.generated.js, asset.id);
+      this.dedupe.set(this.dedupeKey(asset), asset.id);
     }
 
     let deps = {};
@@ -60,7 +60,7 @@ class JSPackager extends Packager {
         deps[dep.name] = bundles;
         this.bundleLoaders.add(mod.type);
       } else {
-        deps[dep.name] = this.dedupe.get(mod.generated.js) || mod.id;
+        deps[dep.name] = this.dedupe.get(this.dedupeKey(mod)) || mod.id;
 
         // If the dep isn't in this bundle, add it to the list of external modules to preload.
         // Only do this if this is the root JS bundle, otherwise they will have already been
@@ -91,6 +91,17 @@ class JSPackager extends Packager {
     }
 
     return name;
+  }
+
+  dedupeKey(asset) {
+    // cannot rely *only* on generated JS for deduplication because paths like
+    // `../` can cause 2 identical JS files to behave differently depending on
+    // where they are located on the filesystem
+    let deps = Array.from(asset.depAssets.keys(), dep =>
+      this.bundler.resolver.resolveFilename(dep.name, asset.name)
+    );
+    deps.sort();
+    return JSON.stringify([asset.generated.js, deps]);
   }
 
   async writeModule(id, code, deps = {}, map) {
--- a/src/packagers/JSPackager.js
+++ b/src/packagers/JSPackager.js
@@ -3,6 +3,7 @@ const path = require('path');
 const Packager = require('./Packager');
 const urlJoin = require('../utils/urlJoin');
 const lineCounter = require('../utils/lineCounter');
+const objectHash = require('../utils/objectHash');
 
 const prelude = {
   source: fs
@@ -35,13 +36,14 @@ class JSPackager extends Packager {
   }
 
   async addAsset(asset) {
-    if (this.dedupe.has(this.dedupeKey(asset))) {
+    let key = this.dedupeKey(asset);
+    if (this.dedupe.has(key)) {
       return;
     }
 
     // Don't dedupe when HMR is turned on since it messes with the asset ids
     if (!this.options.hmr) {
-      this.dedupe.set(this.dedupeKey(asset), asset.id);
+      this.dedupe.set(key, asset.id);
     }
 
     let deps = {};
@@ -97,11 +99,8 @@ class JSPackager extends Packager {
     // cannot rely *only* on generated JS for deduplication because paths like
     // `../` can cause 2 identical JS files to behave differently depending on
     // where they are located on the filesystem
-    let deps = Array.from(asset.depAssets.keys(), dep =>
-      this.bundler.resolver.resolveFilename(dep.name, asset.name)
-    );
-    deps.sort();
-    return JSON.stringify([asset.generated.js, deps]);
+    let deps = Array.from(asset.depAssets.values(), dep => dep.name).sort();
+    return objectHash([asset.generated.js, deps]);
   }
 
   async writeModule(id, code, deps = {}, map) {
--- a/test/integration/js-different-contents/hello.js
+++ b/test/integration/js-different-contents/hello.js
@@ -1,0 +1,2 @@
+const value = 'Hello'
+export default value
--- a/test/integration/js-different-contents/index.js
+++ b/test/integration/js-different-contents/index.js
@@ -1,0 +1,4 @@
+import hello from './hello.js'
+import world from './world.js'
+
+export default `${hello} ${world}!`
--- a/test/integration/js-different-contents/world.js
+++ b/test/integration/js-different-contents/world.js
@@ -1,0 +1,2 @@
+const value = 'World'
+export default value
--- a/test/integration/js-same-contents-different-dependencies/hello/index.js
+++ b/test/integration/js-same-contents-different-dependencies/hello/index.js
@@ -1,0 +1,3 @@
+// hello/index.js and world/index.js are exactly the same content-wise, but in different locations
+import value from './value'
+export default value
--- a/test/integration/js-same-contents-different-dependencies/hello/value.js
+++ b/test/integration/js-same-contents-different-dependencies/hello/value.js
@@ -1,0 +1,2 @@
+const value = 'Hello'
+export default value
--- a/test/integration/js-same-contents-different-dependencies/index.js
+++ b/test/integration/js-same-contents-different-dependencies/index.js
@@ -1,0 +1,4 @@
+import hello from './hello'
+import world from './world'
+
+export default `${hello} ${world}!`
--- a/test/integration/js-same-contents-different-dependencies/world/index.js
+++ b/test/integration/js-same-contents-different-dependencies/world/index.js
@@ -1,0 +1,3 @@
+// hello/index.js and world/index.js are exactly the same content-wise, but in different locations
+import value from './value'
+export default value
--- a/test/integration/js-same-contents-different-dependencies/world/value.js
+++ b/test/integration/js-same-contents-different-dependencies/world/value.js
@@ -1,0 +1,2 @@
+const value = 'World'
+export default value
--- a/test/integration/js-same-contents-same-dependencies/hello1.js
+++ b/test/integration/js-same-contents-same-dependencies/hello1.js
@@ -1,0 +1,2 @@
+const value = 'Hello'
+export default value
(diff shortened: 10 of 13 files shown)


src/packagers/JSPackager.js
```diff
@@ -14,6 +14,20 @@ const prelude = {
     .replace(/;$/, '')
 };
 
+const dedupKey = asset => {
+  // cannot rely only on generated JS for deduplication as paths like `../` can
+  // cause 2 identical JS files to be different depending on filesystem locations
+  return {
```
devongovett: We cannot use objects as keys for the dedupe map - they will never match since JS compares keys by object identity. You'll need to generate a string key.

pcattori:
@devongovett addressed your review comments and refactored to use `Resolver.resolveFilename` 😄

src/packagers/JSPackager.js
```diff
@@ -93,6 +93,17 @@ class JSPackager extends Packager {
     return name;
   }
 
+  dedupeKey(asset) {
+    // cannot rely *only* on generated JS for deduplication because paths like
+    // `../` can cause 2 identical JS files to behave differently depending on
+    // where they are located on the filesystem
+    let deps = Array.from(asset.dependencies.keys()).map(dep =>
```
devongovett: Instead of using `asset.dependencies` here, you could use `asset.depAssets` which is a list of resolved assets rather than dependencies. So you won't even need to call `resolveFilename`, it will already have the resolved filename as its `name` property.

pcattori:
Regarding tests, my plan would be to have the following 3 tests:

1. 2 files with different contents should not be dedupe'd
2. 2 files with same contents *AND* same dependency paths *should* be dedupe'd
2. 2 files with same contents *BUT* with different dependency paths *should not* be dedupe'd

@devongovett I can definitely make the quick change to use `asset.depAssets`. For the tests, it may take me some time to figure out how to best write tests for this repo, so if you can help to implement the tests it would greatly speed things up! 🚀 but if not, I'll get around to it when I have some time 😄

henri-hulski:
Fixes also the bug in #1152.

henri-hulski:
@pcattori When running the javascript tests locally with your fix I get the following errors:
```
  1) javascript should split bundles when a dynamic import is used with --target=browser:
     TypeError: Path must be a string. Received { browser: '/home/pacha/web/cerebral/parcel/src/builtins/loaders/browser/js-loader.js',
  node: '/home/pacha/web/cerebral/parcel/src/builtins/loaders/node/js-loader.js' }
      at assertPath (path.js:28:11)
      at Object.isAbsolute (path.js:1225:5)
      at Resolver.resolve (src/Resolver.js:67:14)
      at <anonymous>

  2) javascript should split bundles when a dynamic import is used with --target=node:
     TypeError: Path must be a string. Received { browser: '/home/pacha/web/cerebral/parcel/src/builtins/loaders/browser/js-loader.js',

  node: '/home/pacha/web/cerebral/parcel/src/builtins/loaders/node/js-loader.js' }
      at assertPath (path.js:28:11)
      at Object.isAbsolute (path.js:1225:5)
      at Resolver.resolve (src/Resolver.js:67:14)
      at <anonymous>

  3) javascript should dynamic import files which import raw files:
     TypeError: Path must be a string. Received { browser: '/home/pacha/web/cerebral/parcel/src/builtins/loaders/browser/js-loader.js',
  node: '/home/pacha/web/cerebral/parcel/src/builtins/loaders/node/js-loader.js' }
      at assertPath (path.js:28:11)
      at Object.isAbsolute (path.js:1225:5)
      at Resolver.resolve (src/Resolver.js:67:14)
      at <anonymous>

  4) javascript should return all exports as an object when using ES modules:
     TypeError: Path must be a string. Received { browser: '/home/pacha/web/cerebral/parcel/src/builtins/loaders/browser/js-loader.js',
  node: '/home/pacha/web/cerebral/parcel/src/builtins/loaders/node/js-loader.js' }
      at assertPath (path.js:28:11)
      at Object.isAbsolute (path.js:1225:5)
      at Resolver.resolve (src/Resolver.js:67:14)
      at <anonymous>

  5) javascript should hoist common dependencies into a parent bundle:
     TypeError: Path must be a string. Received { browser: '/home/pacha/web/cerebral/parcel/src/builtins/loaders/browser/js-loader.js',
  node: '/home/pacha/web/cerebral/parcel/src/builtins/loaders/node/js-loader.js' }
      at assertPath (path.js:28:11)
      at Object.isAbsolute (path.js:1225:5)
      at Resolver.resolve (src/Resolver.js:67:14)
      at <anonymous>
```
I have also created a test which checks
> 3. 2 files with same contents BUT with different dependency paths should not be dedupe'd
It's failing now but passing with this fix.
Should I create a new PR for the test?

src/packagers/JSPackager.js
henri-hulski: @devongovett I think it still needs `resolveFilename` as the `name` property contains only the relative file name.
But something like this should work:
```js
    let deps = Array.from(asset.depAssets.keys()).map(dep => {
      return this.bundler.resolver.resolveFilename(dep.name, asset.name)
    });
```

henri-hulski:
Added a test for the scenario described in #933 in PR #1198.

PS.: Deleted previous comment as that was an issue coming from an outdated branch and which is already fixed.

henri-hulski:
@pcattori Are you still planning to finish this PR?

pcattori:
@henri-hulski unfortunately, I have not had time to work on this recently. I do intend on continuing work on this after this week, though I'm not sure if this work will be [irrelevant once 1.8 is released](https://github.com/parcel-bundler/parcel/issues/1152#issuecomment-380362580).

If 1.8 is going to be released soon, I may wait to see if the issue is fixed then.

DeMoorJasper:
@pcattori treeshaking will be optional (opt-out or opt-in) so this will still be relevant as far as I know

henri-hulski:
@pcattori For me the PR looks good. Just switch to `asset.depAssets` like in my example above and it should be ready.

src/packagers/JSPackager.js
felixrabe: @henri-hulski It should probably be `...resolveFilename(dep, asset.name)` in your suggestion as `dep` is a string.

henri-hulski: @felixrabe here `dep` is not a string. I've logged it out and also tested this fix. It works fine.

olizilla:
Bump. The underlying issues with the dedupe logic that this PR fixes are important. I'm seeing module require ids being incorrectly re-used for multiple dependencies that are not the same when using parcel `build` on master, but are correctly resolved when tested against this PR.

I think also fixes https://github.com/parcel-bundler/parcel/issues/793

pcattori:
@olizilla @henri-hulski just writing to let you know that I haven't forgotten about this and have some time set aside this week to get this in. I did some work last weekend to get robust tests for this up and almost have those ready.

src/packagers/JSPackager.js
felixrabe: Ah right, depAssets is a Map, and its key is not a string but a dep (see Bundler.js where it does `asset.depAssets.set(...)`). Still need to get used to Maps having non-string keys it seems :)

olizilla:
Another datapoint for this PR. I've tracked down why the existing dedupe logic breaking when bundling [IPFS](https://github.com/ipfs/js-ipfs) as a dependency; two of it's deps have identical index.js files, as they follow a pattern of requiring the same two files and deferring the differences to other files. See:

See:

https://github.com/ipld/js-ipld-dag-cbor/blob/e892dc987d32bdad9dba0f2d86d27ebc254fe967/src/index.js

and

https://github.com/ipld/js-ipld-git/blob/20f7a66c2fc916aeb4d7e6c7141bfe275033472f/src/index.js

When these two unrelated files are used as keys in the dedupe map, they are the same, so they get deduped, even though they come from different modules. As they are the entry point to a dep, that means the whole require is swapped out for the wrong module.

pcattori:
Alright! This now includes the 3 tests I mentioned before as well as all the suggestions left by reviewers so far.

Note that the 2nd test is the important one here (`should not dedupe imports with same content but different absolute dependency paths`). This is the test that goes from red to green due to the changes in `JSPackager.js` (the other 2 tests are just there to make sure those changes didn't break other existing behaviors).

henri-hulski:
@pcattori Looks good to me, but in Windows ([AppVeyor](https://ci.appveyor.com/project/devongovett/parcel/build/1.0.1816/job/ycfydopa9t2eee58#L641)) one test is failing:
```
1) javascript
       should dedupe imports with same content and same dependency paths:
      AssertionError: false == true
      + expected - actual
      -false
      +true
      
      at Context.<anonymous> (test\javascript.js:871:5)
      at undefined.next (native)
      at step (test\javascript.js:3:191)
      at C:\projects\parcel\test\javascript.js:3:361
      at process._tickCallback (internal/process/next_tick.js:109:7)
  
error An unexpected error occurred: "Command failed.
Exit code: 1
Command: C:\\windows\\system32\\cmd.exe
Arguments: /d /s /c cross-env NODE_ENV=test mocha
Directory: C:\\projects\\parcel
Output:
".
info If you think this is a bug, please open a bug report with the information provided in "C:\\projects\\parcel\\yarn-error.log".
```

henri-hulski:
Maybe a problem with slashes vs. backslashes.

pcattori:
Yep, looks like making the paths xplatform did the trick. Now the only tests that are failing are some flakey `wasm`-related tests (which I think is a known issue?)

olizilla:
It'd be great to get this merged. The "files as Map key" issue causes imports to incorrectly resolve to unrelated modules which happen to have the same shaped index.js file as any other dependency in your tree, which is causing hard to debug errors, and only for production builds. 

See https://github.com/ipfs/js-ipfs/issues/1344

pcattori:
@devongovett just a friendly ping 😄

asyncx:
Any idea when this will be merged?

olizilla:
Is it the test failure that's blocking this? It looks like it's unrelated 

```
  314 passing (4m)
  2 failing
  1) rust
       should use wasm-gc to minify output:
     Error: cargo failed.
      at ChildProcess.<anonymous> (src/utils/pipeSpawn.js:1:3757)
      at maybeClose (internal/child_process.js:925:16)
      at Process.ChildProcess._handle.onexit (internal/child_process.js:209:5)
  
  2) wasm
       --target=node
         should load a wasm file in parallel with a dynamic JS import:
     Error: Cannot find module '1678'
      at newRequire (evalmachine.<anonymous>:39:19)
      at newRequire (<anonymous>:31:1)
```

pcattori:
@olizilla yea definitely unrelated. those tests were sporadically failing before I made my changes as well.

gaspard:
Please unlock this PR @devongovett. Your requests for changes seems to be blocking.

src/packagers/JSPackager.js
```diff
@@ -93,6 +93,17 @@ class JSPackager extends Packager {
     return name;
   }
 
+  dedupeKey(asset) {
+    // cannot rely *only* on generated JS for deduplication because paths like
+    // `../` can cause 2 identical JS files to behave differently depending on
+    // where they are located on the filesystem
+    let deps = Array.from(asset.depAssets.keys(), dep =>
+      this.bundler.resolver.resolveFilename(dep.name, asset.name)
```
devongovett: if you use `depAssets.values()`, you can just map to `dep.name` which is already resolved

devongovett:
Updated to fix the last comment (use pre-resolved deps), and switched it to use objectHash instead of JSON.stringify (seemed slightly faster in my tests due to shorter string keys).

devongovett:
sorry for the delay!

PR #1017, opened 2018-03-18
Export path to a js dep when it has no loader associated

pierredavidbelanger:
This fixes the fact that an HTML dependency of a JS asset was wrongly written in the JS as an (not working) bundle loader that could make an app crash at runtime with a message like : `Cannot find module '6'`.

This is one of the two points (required to be able to do require('./other.html')) mentioned in [this comment](https://github.com/parcel-bundler/parcel/pull/926#issuecomment-373782658)


pierredavidbelanger:
I am not sure I did this one correctly.

You can see this only as a proof that this problem can be fixed and to have a better view at it.
I guess someone with a better understanding of the codebase will be able to refactor my mess with something beautiful.

For now I, myself, see it as a good enough temporary fix ;)

PR #1022, opened 2018-03-18
Fix svg dependencies not being found when using minification

### LINKED ISSUE #1019: 🐛"build" breaks dependencies detection in <svg><use/>

🐛Bug

SVG files linked with `<use>` are detected correctly with `parcel watch` but they're left untouched when using `parcel build`. 

No config

```html
<!-- index.html -->
<!DOCTYPE html>
<svg xmlns:xlink="http://www.w3.org/1999/xlink">
	<use xlink:href="file.svg#all"/>
</svg>
```

```svg
<!-- file.svg -->
<svg xmlns="http://www.w3.org/2000/svg" id="all"><rect x="0" y="0" width="100" height="100"/></svg>
```

```sh
parcel watch --no-cache
```
```sh
parcel build --no-cache
```

### 🤔 Expected Behavior

The linked SVG is detected and processed.

```html
<!-- index.html -->
<!DOCTYPE html><svg xmlns:xlink="http://www.w3.org/1999/xlink"><use xlink:href="/dist/fe03c3efee964b0b65876cb335d8782f.svg#all"></use></svg>
```

So two files should be generated.

```sh
$ tree dist
dist
├── e4746ad4854ea1f42b9990e63012fe62.html
└── fe03c3efee964b0b65876cb335d8782f.svg

0 directories, 2 files
```

This works correctly in `parcel watch`

### 😯 Current Behavior

The `use` tag in the `index.html` is left untouched:

```html
<!DOCTYPE html><svg xmlns:xlink="http://www.w3.org/1999/xlink"><use xlink:href="file.svg#all"/></svg> 
```

And only one file is generated:

```sh
$ ./node_modules/.bin/parcel build --no-cache
✨  Built in 1.06s.

dist/e4746ad4854ea1f42b9990e63012fe62.html    102 B    1.00s

$ tree dist
dist
└── e4746ad4854ea1f42b9990e63012fe62.html

0 directories, 1 file
```

### 💻 Code Sample

```sh
# create the two files
$ echo '<!DOCTYPE html><svg xmlns:xlink="http://www.w3.org/1999/xlink"><use xlink:href="file.svg#all"/></svg>' > index.html
'<svg xmlns="http://www.w3.org/2000/svg" id="all"><rect x="0" y="0" width="100" height="100"/></svg>'
$ echo '<svg xmlns="http://www.w3.org/2000/svg" id="all"><rect x="0" y="0" width="100" height="100"/></svg>' > file.svg

# build creates one file
$ parcel build

# watch creates 2 files
$ parcel watch
```

### 🌍 Your Environment

| Parcel           | 1.6.2
| Node             | 5.6.0
| npm/Yarn         | 5.7.1
| Operating System | macOS 10.13.3

DeMoorJasper:
Adds a walker as a plugin that collects svg dependencies before htmlnano is being ran, this enables svg dependency collecting as svgo stringifies the entire svg ast tree removing our ability to walk through it.

Closes #1019


src/assets/HTMLAsset.js
```diff
@@ -19,10 +19,9 @@ const ATTRS = {
     'iframe',
     'embed'
   ],
-  href: ['link', 'a', 'use'],
+  href: ['link', 'a'],
   srcset: ['img', 'source'],
   poster: ['video'],
-  'xlink:href': ['use'], // Deprecated since SVG 2, throws error in svgo
```
bfred-it: FYI Chrome doesn’t support the plain `href` so it can be “deprecated” but support is 0 as far as I’m concerned. Does this change still work?

DeMoorJasper: It's still supported in this PR, don't wanna break cross-browser support.
Dependency collection is just moved from `collectDependencies` into `svgDependencyWalker` in this pr (as htmlnano stringifies svgs...)

devongovett:
Instead of introducing a separate dependency walker just for SVG, we could do this the same way JSAsset does it. There are both `pretransform` and `transform` hooks: `pretransform` happens before collecting dependencies and should apply posthtml, and `transform` happens after collecting deps and should apply htmlnano.

PR #1025, opened 2018-03-19
Content hash output file names

devongovett:
This implements the strategy described in #872. In production mode, a hash of bundle content is appended to names of files that aren't entry points (e.g. HTML files, things linked to a `<a href>`, service workers, etc.). Unhashed files retain their original name, and path relative to the main entry point.

In development mode, rather than using content hashes, the hash of the filename is used as before, so that filenames don't change during development. The hashes are necessary at all to avoid name collisions since the file structure is flattened. In addition, the filenames look more similar to the real production ones - the hashes just don't change.

It works by using a temporary filename (md5 of asset path) as references (which should be unique), and at the end, once file contents is known for all bundles, it replaces those temporary names with the real content-hashed names during packaging.

To do:

- [x] fix hashing of raw assets (which aren't in memory)
- [x] more unit tests
- [ ] test in real world apps.

Please help me test! cc. @zeakd @songlipeng2003 @ssuman @Munter @benhutton @shanebo @leeching @gamebox @npup

### DIFF
--- a/src/Asset.js
+++ b/src/Asset.js
@@ -148,7 +148,7 @@ class Asset {
       await this.getDependencies();
       await this.transform();
       this.generated = await this.generate();
-      this.hash = this.generateHash();
+      this.hash = await this.generateHash();
     }
 
     return this.generated;
--- a/src/Asset.js
+++ b/src/Asset.js
@@ -175,27 +175,9 @@ class Asset {
   }
 
   generateBundleName() {
-    const ext = '.' + this.type;
-
-    const isEntryPoint = this.name === this.options.mainFile;
-
-    // If this is the entry point of the root bundle, use outFile filename if provided
-    if (isEntryPoint && this.options.outFile) {
-      return (
-        path.basename(
-          this.options.outFile,
-          path.extname(this.options.outFile)
-        ) + ext
-      );
-    }
-
-    // If this is the entry point of the root bundle, use the original filename
-    if (isEntryPoint) {
-      return path.basename(this.name, path.extname(this.name)) + ext;
-    }
-
-    // Otherwise generate a unique name
-    return md5(this.name) + ext;
+    // Generate a unique name. This will be replaced with a nicer
+    // name later as part of content hashing.
+    return md5(this.name) + '.' + this.type;
   }
 
   generateErrorMessage(err) {
--- a/src/Bundle.js
+++ b/src/Bundle.js
@@ -91,6 +91,59 @@ class Bundle {
     return this.assets.size === 0;
   }
 
+  getBundleNameMap(contentHash, hashes = new Map()) {
+    let hashedName = this.getHashedBundleName(contentHash);
+    hashes.set(Path.basename(this.name), hashedName);
+    this.name = Path.join(Path.dirname(this.name), hashedName);
+
+    for (let child of this.childBundles.values()) {
+      child.getBundleNameMap(contentHash, hashes);
+    }
+
+    return hashes;
+  }
+
+  getHashedBundleName(contentHash) {
+    // If content hashing is enabled, generate a hash from all assets in the bundle.
+    // Otherwise, use a hash of the filename so it remains consistent across builds.
+    let ext = Path.extname(this.name);
+    let hash = (contentHash
+      ? this.getHash()
+      : Path.basename(this.name, ext)
+    ).slice(-8);
+    let entryAsset = this.entryAsset || this.parentBundle.entryAsset;
+    let name = Path.basename(entryAsset.name, Path.extname(entryAsset.name));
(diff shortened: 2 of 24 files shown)


jamiebuilds:
This is a really nice way to do it, we should make a chart of the implications of the different ways to import resources:

| Language | Syntax | Dynamic (possibly code split) | Public URL (keeps name) |
| ---:| --- |:---:|:---:|
| JavaScript | `import 'x'` |  |  |
| JavaScript | `import('x')` | ✅ | 
| JavaScript | `SW.register('x')` | ✅ | ✅ |
| HTML | `<iframe src="x"` | ✅ | ✅ |
| HTML | `<a href="x">` | ✅ | ✅ |
| CSS | `@import "x"` | ✅ |  |

devongovett:
iframes do not currently retain the filename (they are hashed) since they aren't linked to directly. do you think we should keep the name?

jamiebuilds:
Hmm... Yeah, I think they should be. I would treat them as public resources that could be linked to elsewhere and not partials

jamiebuilds:
🎉

TimNZ:
LOVING the new filenames.
Along with simplifying cache busting deployment will make debugging easier.
Thanks!

rogeriochaves:
is there any way to opt out of hashing at all? I'm currently building a browser extension, and I have to explicitly list every js I'll need to load on the manifest.json file, so I can't have filenames changing all the time

TheAlexPorter:
I'm doing the same thing as @rogeriochaves . It would be nice to be able to turn off the hashing so that tools, like Chrome Extensions, get static names that don't change so we can manage them better.

felixrabe:
@rogeriochaves @TheAlexPorter Please find or open an issue for this and mention it here. Thx!

denis1stomin:
Hey guys,
I have a few questions for you regarding the strategy.
I have a folder somewhere inside my repo some/path/webui.
This folder contains standard package structure and src folder in it.
I moved index,html file into src folder since for me it is a source code.
(I'm a pro developer but really new in those all frontend things :))
Now when I run "npx parcel build" (production mode) or globally "parcel build" it produces a file like
dist/src.4320c2f6.html. Looks like it treats all src/* files like "aren't entry points".
For me it looks like a bug.
So the questions are:
1. Is it a bug for you and can it be easily fixed in future versions?
2. If it is "by design", how you'd describe (your vision) where html files should be places and why?

Thanks

DeMoorJasper:
@denis1stomin html files are not content hashed if they are the entry-point or required from html. So not sure why this file is content hashed?

imWildCat:
I really need the support for disabling hashing for a specific path, for example `./images/`.

I am struggling to use parcel with `lottie-web`,

This way `<a href="x">` does not support wildcard paths, such as `<a href="./images/*">`.

Currently I have to do this in a very ugly way:

```html
        <a href="./images/img_0.png" />
        <a href="./images/img_1.png" />
        <a href="./images/img_2.png" />
        <a href="./images/img_3.png" />
        <a href="./images/img_4.png" />
        <a href="./images/img_5.png" />
        <a href="./images/img_6.png" />
        <a href="./images/img_7.png" />
        <a href="./images/img_8.png" />
        <a href="./images/img_9.png" />
        <a href="./images/img_10.png" />
        <a href="./images/img_11.png" />
```

micopiira:
Would it be possible to hash the filenames when using a JS entrypoint?

For example: `parcel build index.js test.js` would create`dist/index.da39a3.js`and a `dist/test.ee5e6b.js`, possibly with a JSON file containing the mappings

```json
{
  "index.js": "dist/index.da39a3.js",
  "test.js": "dist/test.ee5e6b.js"
}
```

This feature would make working with a Maven/Java web project a lot easier (for me, atleast..)

equinusocio:
Hi, is there a way to disable hashing? I need to distribuite a single css file on NPM and the hash is a problem.

LLe27:
Hey everyone,

I apologize if my question is not clear and concise. I am relatively new to the parceljs bundler and progressive web apps. I am currently trying to create an offline experience but I need to specify specific JS and CSS files that are needed for this experience in production. How do I, or how would I detect the exact filenames considering they are appended with hashes? This is needed for the cache file list for my service worker.

Thanks!

mischnic:
@LLe27 Some options:
- pass `--no-content-hash` to Parcel so that the hash stays constant (is only dependent on the path and not the content).
- generate the Service worker cache list after building based on the actual output names

LLe27:
@mischnic 

Thanks, i'll give this a try! Worse case i'll just implement the Workbox library.

joeosburn:
@devongovett We've tested this and the bundle files do get updated with new filename hashes when their contents change. However, the entry file does not get updated to point to the newly hashed files.

PR #1031, opened 2018-03-20
added support for re using same browser tab for google chrome on osx

shrynx:
This PR aims to added support for re using existing browser tab for google chrome on osx.
This is implemented using [Mac OS automation scripting](https://developer.apple.com/library/content/documentation/LanguagesUtilities/Conceptual/MacAutomationScriptingGuide/index.html) using javascript.

Although this is very specific case of just supporting chrome on osx, but supporting other browser is currently not possible.

Prior discussions https://github.com/parcel-bundler/parcel/pull/960#discussion_r173371973

### DIFF
--- a/src/utils/openInBrowser.js
+++ b/src/utils/openInBrowser.js
@@ -1,8 +1,37 @@
 const opn = require('opn');
+const execSync = require('child_process').execSync;
 
 const openInBrowser = async (url, browser) => {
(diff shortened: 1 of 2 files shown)


devongovett:
While I can see how this might be useful, I think it is far too limited in its current form (only macOS Yosemite + Chrome). Maybe we could revisit in the future.

Issue #1032, opened 2018-03-20
SCSS resouce file possible?


th3fallen:
I'm trying to migrate from webpack to parcel but in webpack i had my scss builder to add a resource.scss file to the head of every scss file in my project for variables/mixin's ect.

is that possible with parcel and im just missing it?

```
{
            loader: 'sass-loader',
            options: {
              data: '@import "styles/resources";',
              includePaths: [path.join(cwd, 'app')],
            },
          },
```

steffenmllr:
Same question here. That keeps me from migrating at the moment.

steffenmllr:
I'm using a `.sassrc.js` at the moment which works for `includePaths` but not for `data`:

```js
const path = require('path');
const cwd = process.cwd();

module.exports = {
    data: '@import "_globals";',
    includePaths: [
        path.resolve(cwd, 'node_modules/susy/sass'),
        path.resolve(cwd, 'node_modules/normalize.scss'),
        path.resolve(cwd, 'node_modules'),
        path.resolve(cwd, 'static', 'scss')
    ]
};
```

devongovett:
Fixed by #1035.

th3fallen:
@steffenmllr how are you including that js file? just in your index?

PR #1035, opened 2018-03-20
Add data so it is possible to include globals scss variables

steffenmllr:
This is a fix for #1032 - gives the possibility to use globals variables within the `data` attribute - same as https://github.com/webpack-contrib/sass-loader/blob/master/lib/normalizeOptions.js#L24


Dobby89:
@steffenmllr Can I use this to inject JavaScript variables into my scss?

steffenmllr:
@Dobby89 nope you can setup variables in scss (define them in a eg. globals.scss) and those variables are available in all your required scss files

PR #1051, opened 2018-03-23
Added Pug support from parcel-plugin-pug

Ty3uK:
As discussed in [#769](https://github.com/parcel-bundler/parcel/pull/769). Also I've a question - can I extend HTML tests by creating pug templates in same directories (`html, html-css, etc.`)?


DeMoorJasper:
@Ty3uK Probably better to create seperate tests for pug

src/assets/PugAsset.js
```diff
@@ -0,0 +1,121 @@
+const Asset = require('../Asset');
+const HTMLAsset = require('./HTMLAsset');
+
+const load = require('pug-load');
```
DeMoorJasper: Probably best to use localRequire('..') for the pug dependencies, instead of shipping them with parcel

Ty3uK: Need to replace `pug-*` packages with `localRequire('...', this.name)`?

Ty3uK: Fixed

davidnagli:
Yay 🎉 

Do you mind adding tests for the pug asset?

Ty3uK:
@davidnagli of course, but I'm not so good at this 😔

src/assets/PugAsset.js
```diff
@@ -0,0 +1,124 @@
...
+            const elements = ATTRS[attr.name];
+            if (elements && elements.indexOf(node.name) > -1) {
+              if (PURE_STRING_REGEX.test(attr.val)) {
+                this.addURLDependency(
+                  attr.val.substring(1, attr.val.length - 1)
+                );
+              }
+            }
+          }
+        }
```
devongovett: rather than duplicate the above logic, perhaps we should just let the HTMLAsset process these attributes?

src/assets/PugAsset.js
```diff
@@ -0,0 +1,124 @@
...
+    super(name, pkg, options);
+    this.type = 'html';
+  }
+
+  async parse(code) {
+    const load = await localRequire('pug-load', this.name);
+    const lexer = await localRequire('pug-lexer', this.name);
+    const parser = await localRequire('pug-parser', this.name);
+    const linker = await localRequire('pug-linker', this.name);
+    const filters = await localRequire('pug-filters', this.name);
```
devongovett: that's a lot of packages... I wonder if there is a way we can just require the `pug` package rather than all of these separately.

src/assets/PugAsset.js
```diff
@@ -0,0 +1,124 @@
...
+    const result = generateCode(this.ast, {
+      compileDebug: false,
+      pretty: !this.options.minify
+    });
+
+    return {html: wrap(result)()};
+  }
+
+  shouldInvalidate() {
+    return false;
```
devongovett: this is the default. no need to override this method.

src/assets/PugAsset.js
Ty3uK: Agree!

src/assets/PugAsset.js
Ty3uK: Those packages needed for handling internal dependencies, such `include` and `extend`. Calling `pug` onto source file will not be reload those files at changing. That's why we need parse AST and do other things.

src/assets/PugAsset.js
Ty3uK: Thanks!

Ty3uK:
@devongovett fixed issues, added test and added custom filters support

src/assets/PugAsset.js
devongovett: hmm, it seems like `pug.compile` already returns a list of dependencies. Can we just use that rather than re-implementing the ast walking ourselves? https://github.com/pugjs/pug/blob/master/packages/pug/lib/index.js#L192

Ty3uK: @devongovett Ok, I'll check it

Ty3uK:
@davidnagli @devongovett @DeMoorJasper thanks for helping me! 🤝

PR #1053, opened 2018-03-23
Check browserslist prop in package.json for environments

### LINKED ISSUE #1026: Parsing objects for browserslist fails
Parcel doesn't interpret environments in a browserslist config.

Using this config

```json
  "browserslist": {
    "development": ">2%, last 1 edge versions, not ie > 0, not op_mini all",
    "production": ">2%, last 1 edge versions, not ie > 0, not op_mini all"
  }
```

Results in

```sh
🚨  <file>: Browser queries must be an array. Got object.
    at browserslist (.../node_modules/browserslist/index.js:207:11)
    at getTargetEngines (.../node_modules/parcel-bundler/src/utils/getTargetEngines.js:69:24)
    at <anonymous>
```

It looks like the main `browserslist(queries, opts)` API that's exported isn't meant to know about environments (it's left to the user to determine which environment to supply it with).

nicolaisueper:
This will fix #1026.

Please let me know if I should change anything.

Regards
Nico


devongovett:
Awesome, thanks for working on this. Could you add a test to cover this change?

nicolaisueper:
I encountered a problem testing this feature/fix.

[Over here](https://github.com/parcel-bundler/parcel/pull/1053/files#diff-62939a6d4567e7b9e2741ec83de4c4e4R41) I determine the browserlist-prop to use by checking `NODE_ENV` for any value. When running the tests, `NODE_ENV` equals to undefined.

I also have no idea how to reach the options I pass the bundler inside the tests.

Would you please give me a hint how to continue? I pushed a WIP test case w/ my last commit.

nicolaisueper:
I had to refactor things to get the tests running.
I'd be happy if anyone can review this. I'm not very happy with the solution but I didn't find another way.

src/Bundler.js
```diff
@@ -72,6 +72,7 @@ class Bundler extends EventEmitter {
       outDir: Path.resolve(options.outDir || 'dist'),
       outFile: options.outFile || '',
       publicURL: publicURL,
+      isProduction,
```
DeMoorJasper: Could u rename this to `production` instead of `isProduction` as we used the key `production` in the vue branch

src/Parser.js
```diff
@@ -73,7 +74,7 @@ class Parser {
   getAsset(filename, pkg, options = {}) {
     let Asset = this.findParser(filename);
     options.parser = this;
-    return new Asset(filename, pkg, options);
+    return new Asset(filename, pkg, options, this.isProdBuild);
```
DeMoorJasper: You don't need to pass this as it's part of the options?

src/Parser.js
```diff
@@ -46,6 +46,7 @@ class Parser {
     for (let ext in extensions) {
       this.registerExtension(ext, extensions[ext]);
     }
+    this.isProdBuild = options.isProduction;
```
DeMoorJasper: You don't need to set this as it's part of the options?

src/assets/JSAsset.js
```diff
@@ -19,14 +19,15 @@ const SW_RE = /\bnavigator\s*\.\s*serviceWorker\s*\.\s*register\s*\(/;
...
-  constructor(name, pkg, options) {
+  constructor(name, pkg, options, isProdBuild) {
     super(name, pkg, options);
     this.type = 'js';
     this.globals = new Map();
     this.isAstDirty = false;
     this.isES6Module = false;
     this.outputCode = null;
     this.cacheData.env = {};
+    this.isProdBuild = isProdBuild;
```
DeMoorJasper: This should normally be part of the this.options

src/transforms/babel.js
```diff
@@ -26,8 +26,8 @@ const JSX_PRAGMA = {
   hyperapp: 'h'
 };
 
-async function babelTransform(asset) {
-  let config = await getConfig(asset);
+async function babelTransform(asset, isProdBuild) {
```
DeMoorJasper: shouldn't be necessary as asset.options.production should return the same as `isProdBuild `

src/Parser.js
nicolaisueper: The options don't get stored inside the parser object. Should I do `this.options = options` instead?

src/assets/JSAsset.js
nicolaisueper: This is just the asset options, not the application options afaik

nicolaisueper: Oh, you're right

src/transforms/babel.js
nicolaisueper: Refactored 👌

Grawl:
There is no documentation for this feature on Parcel website.

Issue #993, opened 2018-03-13
🙋Add verbose logging mode


bfred-it:
**Choose one:** is this a 🐛 bug report or 🙋 feature request?

🙋 feature request

It's impossible to debug an issue or to show anyone if I can't see what's happening. All logging messages are replaced one the build is done.

### 🎛 Configuration (.babelrc, package.json, cli command)

```sh
parcel watch
```

### 🤔 Expected Behavior

```sh
$ parcel watch --verbose
Server running at http://localhost:1234 
⏳  Building index.html...
⏳  Building index.js...
✨  Built in 3.78s.
⏳  Building index.js...
✨  Built in 0.78s.
⏳  Building index.html...
✨  Built in 0.1s.
```

### 😯 Current Behavior

```sh
$ parcel watch
Server running at http://localhost:1234 
✨  Built in 3.78s.
# what happened?! what did it do?
```

### 🔦 Context

Mostly for debugging purposes

### 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Parcel           | 1.6.2
| Node             | 8.6.0
| npm/Yarn         | 5.7.1
| Operating System | macOS 10.13.3

davidnagli:
This should be VERY easy to implement 😃 😇 🚀 

We already have an option called `logLevel` that we use internally. So it's just a matter of exposing that in `src/cli.js`.

The easiest way to do it is to make the CLI option called `--logLevel [level]` which would just take the that level as a parameter directly (however this would probably require us to document how those log levels actually work). 

If you just add that to the commands in `src/cli.js`, it should just automatically start working, since the code that parses that option already exists 👍

DeMoorJasper:
@davidnagli We don't have a verbose level though, this would require someone to add an extra level to the loglevel value for example go up to lvl 4 instead of 3, 4 being verbose

davidnagli:
@DeMoorJasper How is lvl 3 different from the verbose mode that @bfred-it is describing? 

What would be new in lvl 4?

davidnagli:
Gochya 👍 

We should probably document the log levels somewhere. From what I pieced together:

| logLevel | Behavior |
| ------- | ---------- |
| 0 | No output
| 1  | Errors
| 2 | Warnings + Errors
| 3 |  All
| 4 | [Proposed] Verbose

TomVance:
Is this issue still being worked on? I'd be interested in starting to contribute too this project and this seems like a good starting point.

[...]

PR #1055, opened 2018-03-23
Add option for log level

nicolaisueper:
Related to #993.

Please let me know if I should add / change something.

Regards
Nico


davidnagli:
#993 can use this once we define a logLevel 4

PR #1065, opened 2018-03-25
Implement pipelines to compose multiple asset types together

devongovett:
Rather than asset type classes extending one another, Parcel now takes care of composing asset types together. The output returned by an asset type's `generate` method is now recursively processed. This means, for example, that the `TypeScriptAsset` just returns JavaScript from its generate method and Parcel will automatically pass this to the `JSAsset` for further processing.

The API for `generate` changed slightly, but remains backward compatible. Now, instead of using object keys to return multiple renditions from an asset, an array is returned. This allows more options to be specified along with the actual value.

```javascript
generate() {
  return [{
    type: 'css',
    value: ...
  }, {
    type: 'js',
    value: ...
  }];
}
```

There are a couple options that allow some control over how the composition works:

1. The `final` option can be specified, which means the code is final and should not be processed by other asset types.
2. The `hasDependencies` option can be set to `false`, which means the code is guaranteed not to have any further dependencies in it. This is useful as an optimization. For example, the SASS asset type uses the SASS AST to collect dependencies, so the CSS asset type doesn't need to do it.

More options could be added in the future.

This change should be fully backward compatible. If the old API is used, composition doesn't happen at all so it should work exactly the same way as it does today.

This pipeline approach should be much more extendible, especially for plugins which no longer need to rely on the internal asset types in order to extend them. It will also allow us to do things like Vue support much more easily.

### DIFF
--- a/src/Asset.js
+++ b/src/Asset.js
@@ -26,7 +26,7 @@ class Asset {
     this.type = path.extname(this.name).slice(1);
 
     this.processed = false;
-    this.contents = null;
+    this.contents = options.rendition ? options.rendition.value : null;
     this.ast = null;
     this.generated = null;
     this.hash = null;
@@ -58,6 +58,13 @@ class Asset {
   }
 
   async getDependencies() {
+    if (
+      this.options.rendition &&
+      this.options.rendition.hasDependencies === false
+    ) {
+      return;
+    }
+
     await this.loadIfNeeded();
 
     if (this.contents && this.mightHaveDependencies()) {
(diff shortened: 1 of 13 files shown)


DeMoorJasper:
Maybe in the future we could add a map property to the parts?
As css and js can both have a sourcemap it'll be hard to distinguish these 2 in a clean way without explicitly telling the packager which one belongs to which asset type.

devongovett:
Yep, definitely need to do that. Also want to allow multiple renditions of the same type (e.g. multiple JS sections). However, that would require changing the way `asset.generated` is set, and a lot of existing code relies on the current format.

PR #1068, opened 2018-03-25
Fix localRequire install to not install certain packages in parallel

DeMoorJasper:
There has been a bug in parcel ever since I introduced the automatic install of localRequires.
This bug causes some yarn or npm installs to fail due to parcel trying to install it in parallel (running an install per worker at the same time, for example processing 2 sass files at the same time => installing sass twice at the same time in parallel).
This PR should fix this behaviour by moving all parser installs to the master worker/bundler.


package.json
```diff
@@ -71,6 +71,8 @@
     "coffeescript": "^2.0.3",
     "cross-env": "^5.1.1",
     "eslint": "^4.13.0",
+    "glsl-token-assignments": "^2.0.2",
```
DeMoorJasper: Apparently this is even a bug in the current master branch if looking at the test results

src/assets/GLSLAsset.js
```diff
@@ -10,8 +10,15 @@ class GLSLAsset extends Asset {
     this.type = 'js';
   }
 
+  async installParserDependencies() {
+    await localRequire('glslify-deps', this.name);
+    await localRequire('glslify-bundle', this.name);
+    await localRequire('glsl-token-properties', this.name);
```
DeMoorJasper: Apparently this is even a bug in the current master branch if looking at the test results

devongovett: would be cool if this were possible to do in one call rather than 4 separate installs. should be faster that way.

src/assets/RustAsset.js
```diff
@@ -35,7 +40,7 @@ class RustAsset extends Asset {
   }
 
   async parse() {
-    // Install rust toolchain and target if needed
+    // Checks if rust is installed otherwise throw error
     await this.installRust();
```
devongovett: does this still need to be done again here since you moved it to `installParserDependencies`?

src/assets/CoffeeScriptAsset.js
```diff
@@ -2,9 +2,13 @@ const JSAsset = require('./JSAsset');
 const localRequire = require('../utils/localRequire');
 
 class CoffeeScriptAsset extends JSAsset {
+  async installParserDependencies() {
+    await localRequire('coffeescript', this.name);
```
devongovett: Seems a little weird to call localRequire to install something. What do you think about having the method return a list of dependencies instead, which would be installed by the base Asset class.

```javascript
async getParserDependencies() {
  return ['coffeescript'];
}
```

You could also still have `installParserDependencies` as a method of the base Asset class which called that method to get a list of deps. And it could be overridden in the case of e.g. Rust where the deps are not npm. WDYT?

src/assets/RustAsset.js
DeMoorJasper: I guess not, wanted to keep that for the if installed check but it throws so it won't be necessary, i'll remove it

DeMoorJasper:
All merge errors have been resolved.
However due to the new pipeline for combining assets and not having to extend assets,... this now no longer works in case of these assets.
Maybe this can be resolved using something like #891 ?

devongovett:
Hmm, yeah. It also wouldn't work for loading plugins: https://github.com/parcel-bundler/parcel/blob/6f493f030852f557c299b2c26a565c99e1e9de66/src/utils/loadPlugins.js#L22. That uses `localRequire` but now that it doesn't trigger the install, it won't work.

Seems like we will need your work on IPC after all. I guess we should leave `localRequire` calling `installPackage` how it was before, but in `installPackage` if we detect that we're in a worker (`process.send` exists), we send a message over IPC to do the install on the main process. There, we already use a queue to check if multiple installs are happening simultaneously.

PR #1074, opened 2018-03-26
Add overlay for build errors when using hmr

### LINKED ISSUE #314: [HMR] How to keep redux store in HMR?
### 🤔 Expected Behavior
Keep redux  store in HMR.

### 😯 Current Behavior
Lost redux store in HMR.

### 💁 Possible Solution

Intergrate with https://github.com/gaearon/react-hot-loader

### 🌍 Your Environment

| Parcel           |  1.2.0
| Node             | 8.9.3
|Yarn         | 1.3.2
| Operating System | macOS 10.12

### LINKED ISSUE #413: 🙋 Browser error overlay similar to react-error-overlay
Firstly, thanks for all the hard work on Parcel. I think this is going to be quite a force throughout 2018!

🙋 feature request

### 🎛 Configuration

N/A

### 🤔 Expected Behavior

Create-React-App has an error overlay that, when an error/warning occurs, shows details about the issue on the browser (overlaying the web app). Also this error is usually clickable, jumping straight to the file and line of the issue. I find this very useful and noticed I missed it when trying out Parcel. 

### 😯 Current Behavior

Error shown in terminal and browser console, but app remains in last good state.

### 💁 Possible Solution

I believe react-error-overlay uses webpack, so maybe that can be ported over to Parcel? Sorry, my knowledge of build tools isn't great. I'm sure this would be quite a project and I'm in no way asking for it to be added tomorrow; let's see if there's interest.

### 🔦 Context

Whilst the console errors are detailed, some users (and me) may tend to look to their browser first to see what has been affected by their code changes. This helps notify that an error has been found.

### 🌍 Your Environment

| Parcel           | 1.3.1 |
| Node             | 9.3.0 |
| npm/Yarn         | Yarn 1.3.2 |
| Operating System | Windows/macOS |

lmontopo:
Fixes #413

Worked as a team with @sibartlett and @narsaynorath

screen shot 2018-03-26 at 3 12 17 pm

### DIFF
--- a/src/builtins/hmr-runtime.js
+++ b/src/builtins/hmr-runtime.js
@@ -55,7 +55,7 @@ if ((!parent || !parent.isParcelRequire) && typeof WebSocket !== 'undefined') {
       removeErrorOverlay();
 
       var overlay = createErrorOverlay(data);
-      document.body.append(overlay);
+      document.body.appendChild(overlay);
     }
   };
 }
@@ -69,47 +69,22 @@ function removeErrorOverlay() {
 
 function createErrorOverlay(data) {
   var overlay = document.createElement('div');
-  var message = document.createElement('div');
-  var pre = document.createElement('pre');
-  var err = document.createElement('span');
-  var errorIcon = document.createElement('span')
-
   overlay.id = OVERLAY_ID;
 
-  overlay.append(err);
-  overlay.append(errorIcon);
-  overlay.append(message);
-  overlay.append(pre);
-
-  err.innerText = 'ERROR';
-  errorIcon.innerText = '🚨';
+  // html encode message and stack trace
+  var message = document.createElement('div');
+  var stackTrace = document.createElement('pre');
   message.innerText = data.error.message;
-  pre.innerText = data.error.stack;
+  stackTrace.innerText = data.error.stack;
 
-  overlay.style.background = 'black';
-  overlay.style.fontSize = '16';
-  overlay.style.color = 'white';
-  overlay.style.position = 'fixed';
-  overlay.style.height = '100%';
-  overlay.style.width = '100%';
-  overlay.style.top = '0';
-  overlay.style.left = '0';
-  overlay.style.padding = '30px';
-  overlay.style.opacity = '0.85';
-  overlay.style.fontFamily = 'Menlo, Consolas, monospace';
-  overlay.style.zIndex = '9999';
-
-  err.style.background = 'red';
-  err.style.padding = '2px 4px';
-  err.style.borderRadius = '2px';
-
-  errorIcon.style.top = '2px';
-  errorIcon.style.marginLeft = '5px';
-  errorIcon.style.position = 'relative';
-
-  message.style.fontSize = '18';
-  message.style.fontWeight = 'bold';
-  message.style.marginTop = '20px';
+  overlay.innerHTML = (
+    '<div style="background: black; font-size: 16px; color: white; position: fixed; height: 100%; width: 100%; top: 0px; left: 0px; padding: 30px; opacity: 0.85; font-family: Menlo, Consolas, monospace; z-index: 9999;">' +
+      '<span style="background: red; padding: 2px 4px; border-radius: 2px;">ERROR</span>' +
+      '<span style="top: 2px; margin-left: 5px; position: relative;">🚨</span>' +
+      '<div style="font-size: 18px; font-weight: bold; margin-top: 20px;">' + message.innerHTML + '</div>' +
+      '<pre>' + stackTrace.innerHTML + '</pre>' +
+    '</div>'
+  );
 
   return overlay;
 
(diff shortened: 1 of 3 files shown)


devongovett:
Siiiiick! Nice work 🎉

lmontopo:
Any idea why these tests are failing? We've been trying to run the tests locally and different tests are failing.  The one that's consistent is the `hmr should log emitted errors` test, but I'm confused as to why its raising a `Uncaught TypeError: document.getElementById is not a function`.  If you have any ideas that'd be sweet! :)

DeMoorJasper:
@lmontopo In the test environment some tests just fail from time to time.

To fix the Uncaught typeError however I think you have to change this part:
https://github.com/parcel-bundler/parcel/blob/master/test/hmr.js#L309-L315

Into something like this:
```Javascript
run(bundle, {
      console: {
        error(msg) {
          logs.push(msg);
        }
      },
      document: {
        getElementById(id) {
          // return a mock element object
        },
        createElement(element) {
          // return a mock element object
        }
      }
    });
```

sibartlett:
I’m guessing: `document` is undefined because the tests are out ran outside the context of a web browser.

So we probably need input on what to do here...

* Do we add a check to see if `document` is defined?
* Do we try mocking/stubbing `document`?
* Do we take a dependency on phantomjs (or something similar)?

I think that last option is a bit heavy though.

lmontopo:
Thanks @DeMoorJasper, that makes sense (mocking those methods).  We'll look into doing this soon (probably tomorrow) and then we'll push updates! :)

src/builtins/hmr-runtime.js
```diff
@@ -43,14 +45,76 @@ if ((!parent || !parent.isParcelRequire) && typeof WebSocket !== 'undefined') {
...
+
+  errorIcon.style.top = '2px';
+  errorIcon.style.marginLeft = '5px';
+  errorIcon.style.position = 'relative';
+
+  message.style.fontSize = '18';
+  message.style.fontWeight = 'bold';
+  message.style.marginTop = '20px';
+
+  return overlay;
```
devongovett: what do you think about just using `innerHTML` for most of this? Might be easier to read?

lmontopo: just pushed up a fix to this! Took some time to figure out how to make it safe (i.e. how to ensure the message and stack trace are html encoded) but we figured it out.  Let me know what you think.

src/builtins/hmr-runtime.js
```diff
@@ -43,14 +45,53 @@ if ((!parent || !parent.isParcelRequire) && typeof WebSocket !== 'undefined') {
...
+      removeErrorOverlay();
     }
 
     if (data.type === 'error') {
       console.error('[parcel] 🚨  ' + data.error.message + '\n' + data.error.stack);
+
+      removeErrorOverlay();
+
+      var overlay = createErrorOverlay(data);
+      document.body.append(overlay);
```
sibartlett: We should use `appendChild` instead of `append`.

src/builtins/hmr-runtime.js
```diff
@@ -43,14 +45,53 @@ if ((!parent || !parent.isParcelRequire) && typeof WebSocket !== 'undefined') {
...
+  var html = (
+    '<div style="background: black; font-size: 16px; color: white; position: fixed; height: 100%; width: 100%; top: 0px; left: 0px; padding: 30px; opacity: 0.85; font-family: Menlo, Consolas, monospace; z-index: 9999;">' +
+      '<span style="background: red; padding: 2px 4px; border-radius: 2px;">ERROR</span>' +
+      '<span style="top: 2px; margin-left: 5px; position: relative;">🚨</span>' +
+      '<div style="font-size: 18px; font-weight: bold; margin-top: 20px;">' + message.innerHTML + '</div>' +
+      '<pre>' + stackTrace.innerHTML + '</pre>' +
+    '</div>'
+  )
+
+  overlay.innerHTML = html;
```
sibartlett: Instead of wrapping the overlay in another div, we could try using`outerHTML` instead:

```js
function createErrorOverlay(data) {
  var overlay = document.createElement('div');
  
  var html = (
    '<div id="' + OVERLAY_ID + ' ...' +
    '</div>'
  )

  overlay.outerHTML = html;

  return overlay;
}
```

lmontopo: good idea will do!

src/builtins/hmr-runtime.js
lmontopo: 👍

src/builtins/hmr-runtime.js
lmontopo: We decided to get rid of the html var as you suggested, but we kept the outer div because it just seemed more straightforward / easy to read and understand.

sibartlett: 👍

lmontopo:
yay! thanks for adding the tests @devongovett. Super psyched this was merged!

sibartlett:
Yep, thanks for merging! I'm glad to have contributed to an awesome tool :)

devongovett:
Thanks for contributing it! I'm excited to use it. :)

PR #1085, opened 2018-03-28
Update addBundleLoader to accept multiple targets

fathyb:
Follow up to #981 based on [@devongovett comment](https://github.com/parcel-bundler/parcel/pull/981#discussion_r177645431) :

> would be cool to add support for registering bundle loaders for specific environments. this would allow third party plugins to do the same:
>
>```javascript
> bundler.addBundleLoader('wasm', {
>  node: require.resolve('./node/wasm-loader'),
>  browser: require.resolve('./browser/wasm-loader')
>});
>```

### DIFF
--- a/src/Bundler.js
+++ b/src/Bundler.js
@@ -37,13 +37,18 @@ class Bundler extends EventEmitter {
     this.delegate = options.delegate || {};
     this.bundleLoaders = {};
 
-    const loadersPath = `./builtins/loaders/${
-      options.target === 'node' ? 'node' : 'browser'
-    }/`;
-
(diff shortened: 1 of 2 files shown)


src/Bundler.js
```diff
@@ -122,16 +127,26 @@ class Bundler extends EventEmitter {
     this.packagers.add(type, packager);
   }
 
-  addBundleLoader(type, path) {
-    if (typeof path !== 'string') {
-      throw new Error('Bundle loader should be a module path.');
+  addBundleLoader(type, paths) {
+    if (typeof paths !== 'object') {
```
devongovett: If it's a string, we could just assign the same file to all targets.

PR #1086, opened 2018-03-28
Wait for the next event loop iteration when loading WASM

fathyb:
The test `should load a wasm file in parallel with a dynamic JS import` was failing randomly since #981 was merged, this fixes it.

Combined with #1068 the tests should be reliable again.


fathyb:
Honestly I don't fully understand it either, I did it from a gut feeling that the parent isn't loaded and it worked.

Even weirder: replace `setImmediate` with `process.nextTick` and it doesn't work. Could've made sense if we weren't using `readFile` but like you said `fs.readFile` is already async.

PR #1088, opened 2018-03-28
WIP: Don't serve on non html main assets

### LINKED ISSUE #1005: [RFC] Don't start dev server for Node bundling?

nicolaisueper:
This PR will fix #1005.

I currently do a strict check, whether the main entry asset has the `.html` file extension or not.
I don't know if anyone still does this today, but this would fail on `.htm` or `.html5` main files.
This could be solved using a RegExp, like e.g. `/htm|html|html5/`.

But is it intended to support main `.hbs` or other template language files as main assets?

I'd be happy if anyone has a suggestion / tip for this problem.

Regards
Nico

### DIFF
--- a/src/cli.js
+++ b/src/cli.js
@@ -177,8 +177,8 @@ async function bundle(main, command) {
   }
 
   const bundler = new Bundler(main, command);
-
-  if (command.name() === 'serve') {
+  const mainFileExtension = main.split('.').pop();
+  if (command.name() === 'serve' && mainFileExtension === 'html') {
     const server = await bundler.serve(command.port || 1234, command.https);
     if (command.open) {
       await require('./utils/openInBrowser')(


DeMoorJasper:
What about checking for asset type and only serving if asset type is html?

nicolaisueper:
Is there already a way to do this? I found nothing similiar.

nicolaisueper:
I could check if `bundler.parser.findParser(ext) instance RawAsset`. Is that the way to go?

DeMoorJasper:
There is not and i'm not really sure how you'd do it, but this approach would break any non-html file as entrypoint that gets converted into html like pug for example.

matthewmueller:
Personally, I'd like to see an HTML page generates for non-HTML assets rather than failing to serve. Something like the way that https://github.com/jantimon/html-webpack-plugin works.

Issue #1096, opened 2018-03-29
Serving html files with links to static assets (pdf, jpg) produces error "ENOENT: no such file or directory"


prosellen:
🐛 bug report 

#### cli command
```bash
parcel src/index.html
```

#### src/index.html
```html
<html>
    <head>
    </head>
    <body>  
        <a href="./downloads/hello.pdf" target="_blank">This is a link to a PDF-File</a>   
    </body>
</html>
```

#### Project structure
```bash
├── dist
├── src
│   ├── downloads
│   │   ├── hello.pdf
│   ├── index.html
├── package.json
```

### 🤔 Expected Behavior
```parcel src/index.html``` should spin up the internal server
```parcel build src/index.html --public-url ./``` should build the project into the /dist/ folder

### 😯 Current Behavior
As long as the html file contains an anchor link to a static resource like a pdf or an image, Parcel throws the following error: 
`×  ENOENT: no such file or directory, open 'C:\Users\path\to\project\dist\downloads\hello.pdf'`

The same goes for images. This will produce an error:
```<a href="./downloads/cat.jpg" target="_blank">Linktext</a>```  

__However__ if I add the linked image as an `<img>` it will work just fine:
```<a href="./downloads/cat.jpg" target="_blank"><img src="./downloads/cat.jpg" /></a>```  

This seems to be new since I updated to v1.7 - I didn't get the errors when using v1.6.2

### 🔦 Context
Trying to serve/build a project

### 💻 Code Sample
https://github.com/prosellen/parcel-enoent-text

### 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Parcel           | 1.7
| Node             | 8.9.0
| npm         | 5.5.1
| Operating System | Windows 10

**Comment by malonehedges** [comment]
@Ty3uK This isn't fixed when using `import()` statements with #1097. 

```js
// Works
const x = require('./someDir/*')
console.log(x)

// Throws error attached below
import('./someDir/*').then((stuff) => {
  console.log(stuff)
})
```

Error:
```
ENOENT: no such file or directory, open '/path/to/project/src/someDir/*'
```

DeMoorJasper:
I'll close this issue as it seems to be resolved through the PR.

Feel free to open a new issue @malonehedges as this seems to be an unrelated issue

malonehedges:
[...]

PR #1097, opened 2018-03-29
Fix writing files in subfolders inside dist directory

Ty3uK:
Related to #1096. If static files placed in subfolders, parcel throws error in `fs.writeFile` because those subfolders doesn't created inside dist.


Ty3uK:
Sorry for trash commits, merging upstream gives me conflicts 😖

src/packagers/RawPackager.js
```diff
@@ -12,6 +13,12 @@ class RawPackager extends Packager {
       contents = await fs.readFile(contents ? contents.path : asset.name);
     }
 
+    // Create output path if not exist
+    const dir = path.dirname(this.bundle.name);
+    if (!await fs.exists(dir)) {
```
fathyb: I think you can directly do `fs.mkdirp`, `mkdirp` will internally check if the directory exists. It will save a few I/O calls

Ty3uK: Thanks, I'll fix that ASAP

yarn.lock
```diff
@@ -6628,6 +6628,8 @@ void-elements@^2.0.1:
   version "2.0.1"
   resolved "https://registry.yarnpkg.com/void-elements/-/void-elements-2.0.1.tgz#c066afb582bb1cb4128d60ea92392e94d5e9dbec"
 
+<<<<<<< HEAD
```
devongovett: looks like merge issue

Ty3uK: @devongovett my bad :(

devongovett:
Updated slightly to ignore the `mkdirp` call if there are no path separators in the output filename. Should be slightly faster.

PR #1100, opened 2018-03-30
Added --source-root option for sourcemaps

teabyii:


devongovett:
Wait, why is this necessary? I believe we already inline the full source code into the source map so dev tools doesn't need to request it.

Hammster:
@devongovett indeed, i did not know about the rootDir option.

But i can also think of a usecase for setting the sourceRoot option manually to your network adress for when using remote debug for smartphones/TVs/web-devices. But this is more of a edgecase since normally your local browser should resolve the remote debugging information.

devongovett:
`rootDir` is automatically computed based on the input files.

felixfbecker:
`sourceRoot` always being relative is not correct. Chrome devtools will resolve the `sourceRoot` URL relative to the URL of the bundled file that was requested. The source file usually is not hosted at this location. The contents may be included in the source map, but when you have two bundles that have the same file structure within their project, the resolved URIs can collide.

Example: We bundle Sourcegraph extensions with Parcel, and all extensions typically have an `src/extension.ts` as entrypoint. They are loaded from an endpoint like `https://sourcegraph.com/static/extension/{extension id}.js` The source for these will always get resolved to `https://sourcegraph.com/static/src/extension.ts` (for all extensions). This means that when you have multiple extensions loaded and you open devtools, all you see is dozens `src/extension.ts` with no way to know which is which extension.

The way to solve this if you only use source maps in local dev would be to supply a `--source-root` option with `file:///path/to/extension/project/src/` so that each extension has an absolute URL that points to the actual location on disk.

We also want to publish source maps in production, so we would actually set it to our raw API `https://raw.githubusercontent.com/{owner}/{repo}/{commit}/src/` or `https://sourcegraph.com/github.com/{owner}/{repo}@{commit}/-/raw/src/`, so it doesn't depend on local folder structure.

Another use case for setting `sourceRoot` is when you want to publish with source maps for mapped stack traces (especially when tracking errors with e.g. Sentry), but not publish the source contents to protect your IP. You would need unique URLs for each file for the stack traces or you run into the same problem, and you want to jump from the stack trace to the actual file contents later, which depends on the revision you deployed so you want to encode that into the source root URL.

Please add a `--source-root` option.

felixfbecker:
See the source map spec:

> #### Resolving Sources
> If the sources are not absolute URLs after prepending of the “sourceRoot”, the sources are resolved relative to the SourceMap (like resolving script src in a html document).

https://sourcemaps.info/spec.html

felixfbecker:
Another thing is that `sourceRoot` should be set in a way so that it is truly the highest common ancestor of the source files, i.e. no element in the `sources` array should have to start with `../`. This is because the `sourceRoot` gets prepended to `sources` (not resolved) and Chrome will keep any `..` segments in the URL, resulting in a confusing tree in dev tools:

image

Note how the tree starts at `.../raw/extension/src`, then goes up one level `..`, then goes into `src` again. If `sourceRoot` was set to the highest common ancestor (`.../raw/`), and all `sources` were relative to that instead of starting with `..`, this tree would be as expected.

mischnic:
@felixfbecker Please open a new issue, otherwise this will be forgotten.

PR #1090, opened 2018-03-28
Give priority to styl files in StylusAsset resolver

brandon93s:
Stylus's base case expects a `.styl` file. If a folder has two or more files with the same name, but different extensions, any `styl` file should take precedence for the sake of resolution. This sorts `styl` to the top of the list so that it is checked first. 
Closes #1110

PR #1119, opened 2018-04-02
Add support for multiple entry points

### LINKED ISSUE #189: Support for multiple entry points.
Does it support multiple page ? or only single page ?

devongovett:
Closes #189.

This PR adds support for multiple entry points to Parcel. It works by simply adding an additional fake node at the root of the bundle tree and attaching each entry asset to it as a child. If there is only one child, that child bundle is returned as it was before for convenience and to maintain backward compatibility.

Entry points can be passed to parcel as a list of files, or as a glob or list of globs. e.g. `parcel *.html` will run parcel over all HTML files in the current directory. Common dependencies between the entry points, such as a shared JS file, are deduped and produce a single output file that is linked to by both HTML entry points.


davidnagli:
Umm doesn’t this break backwards-compatibility for anyone using Parcel programmatically (requiring Parcel and then manually instanciating Bundler)?? 🤔

devongovett:
Should be backward compatible. It will check if the input is a string rather than an array, and convert it for you.

https://github.com/parcel-bundler/parcel/blob/81aaf9b25a8eebac9307933f57ca2c649575c911/src/Bundler.js#L65-L80

davidnagli:
Oh cool didn’t notice that 👍

briandk:
@devongovett I would love to see this feature! Unfortunately, It looks like the Appveyor build is [failing a Vue.js integration test][1]:

> ```
> vue should produce a vue bundle using preprocessors:
> C:\projects\parcel\test\integration\vue-preprocessors\pre-processors.vue:5:22
> ```

I'm kind of a first-timer here, but is there anything I can do to help?

[1]: https://ci.appveyor.com/project/devongovett/parcel/build/1.0.1409/job/5sialu1y6t7arlmb#L603

FGRibreau:
For those who wish to try it:

```json
"scripts": {
    "postinstall": "(cd node_modules/parcel-bundler && npm i)
},
"devDependencies": {
    "parcel-bundler": "git://github.com/parcel-bundler/parcel.git#multi-entry"
}
```

note: prelude.min.js was not there (generated through `yarn minify` from parcel npm `build` step so I had to force it).

icai:
@devongovett  support middleware as the following?
```
const bundler = new Bundler('./pages/**/*.html', {
  outFile: '*.html',
  publicUrl: './'
});

app.use(bundler.middleware());

```

https://en.parceljs.org/api.html#middleware

nicolasdelfino:
@fathyb is it in fact #1090 or you think it's something else preventing the merge?

tylerlong:
But now I have to access `http://localhost:1234/index.html` instead of `http://localhost:1234`.

Could it be a little smarter to know that I am want `index.html`?

yamsellem:
I don't find any mention of this (great) feature in the doc.
Do I miss something?

If I don't, maybe adding a line or two in the [api](https://en.parceljs.org/cli.html) section will greatly help (_I've searched among github issues before discovering the feature was already there_).

redbulled:
@devongovett thank you for suck an amazing bundler.

> Entry points can be passed to parcel as a list of files, or as a glob or list of globs. e.g. parcel *.html

pls write it somewhere on your site. Because it is such a great future and I was amazed when I found it

matthewmueller:
> But now I have to access http://localhost:1234/index.html instead of http://localhost:1234.

Second this. `http://localhost:1234` with `parcel index.html` will resolve to `http://localhost:1234/index.html` but as soon as you add `parcel index.html app.html`, this no longer works: `http://localhost:1234`

sgf:
waiting for use this feature  😂

elsheepo:
when I bundle multiple entry points, I get SCSS style and working JavaScript on the first html file in alphabetical order, and the rest only HTML, no style, no js, why?
I have tried both of these...
`parcel *.html` && `parcel *.html index.js`
fyi: each page (index.html, about.html, etc) links to index.js like so...
`<script src="./index.js"></script>`
and the contents of index.js are...
`import './vendor/bootstrap/css/bootstrap.min.css'
import './vendor/fontawesome-free/css/all.min.css'
import './scss/custom.scss'
import './vendor/jquery/jquery.min.js'
import './vendor/bootstrap/js/bootstrap.bundle.min.js'
import './js/custom.js'`

ivanmalagon:
@elsheepo I have the same issue but in my case I could solve it writing `<link href="../src/style/custom.scss" rel="stylesheet">` in each HTML instead of setting `import './style/custom.scss` in a JS file.
This way, the SCSS gets processed and added to every HTML file that uses it.

wesbos:
I just added this to the docs. Would love to hear if anyone has a solution to the index.html no longer working with multiple entry points

foobarbecue:
Deduplication seems to be broken when I do `parcel build *.html`. My resulting js bundle has multiple copies of everything -- one for each .html file that referenced it. If this is not a known issue, I should have time to make a minimal example this weekend.

shaunakv1:
Looks like this is still an issue 

`parcel index.html report.html --https --open`  opens: `https://localhost:1234` which doesn't work. 

`https://localhost:1234/index.html` does however. It's a minor inconvenience, which only effects in dev mode, but def one that confuses others developers on the team when adding the multiple entry points for the first time. Plus now you have to go to the browser and type `index.html`

ProLoser:
Doesn't multiple entrypoints require different respective exit points? I do not want all my entrypoints to yield one dist file.

Sometimes I'm transpiling a module or library and then converting it to a string to be injected at runtime. This is currently something I'm unable to do for parcel. Perhaps with a custom prefix? But then I cannot ship this code if it's consumed as esm

Alternatively, I use transpilation to pre-compile static versions of my page namespaced under a categorization folder and would like to run those in parallel.

I can do this with rollup (awkwardly) but cannot do it in Parcel

devongovett:
You can do it with targets. Each target can have a source field that defines its entry point, and a top level field for the output or a distDir within the target. https://parceljs.org/features/targets/#package.json%23targets.*.source

ProLoser:
Can you use glob patterns to handle a parallel series of sources with parallel targets (like statically generated variations of a website)

rohit404404:
> 

Any resolution to this issue yet?

rohit404404:
> Looks like this is still an issue
> 
> `parcel index.html report.html --https --open` opens: `https://localhost:1234` which doesn't work.
> 
> `https://localhost:1234/index.html` does however. It's a minor inconvenience, which only effects in dev mode, but def one that confuses others developers on the team when adding the multiple entry points for the first time. Plus now you have to go to the browser and type `index.html`

Any resolution to this issue yet? there is solution discussed here. HOwever, seems it does not work. https://stackoverflow.com/questions/69029045/deploying-problem-with-multiple-html-files-parcel

wesbos:
No fix - I ended up having to move to Vite just to get this feature unfortunately

ProLoser:
Vite is just roll-up with a bunch of defaults. 

I ended up going back to basics and doing everything in roll-up and gulp for more granular control. [KISS](https://en.wikipedia.org/wiki/KISS_principle)

Tired of wasting most of my time with build tooling instead of just building.

PR #1133, opened 2018-04-04
Remove `eval` usage. Fixes CSP cases.

### LINKED ISSUE #335: 🐛 CSP 'unsafe-eval' not allowed

CvX:
Picks up on work done by @mBeierl (closes #353).
Fixes #335.

### DIFF
--- a/src/visitors/globals.js
+++ b/src/visitors/globals.js
@@ -7,7 +7,7 @@ const VARS = {
     asset.addDependency('process');
     return 'var process = require("process");';
   },
-  global: () => 'var global = (1,eval)("this");',
+  global: () => 'var global = typeof global === "object" ? global : typeof window === "object" ? window :typeof self === "object" ? self : this;',
   __dirname: asset =>
     `var __dirname = ${JSON.stringify(Path.dirname(asset.name))};`,
   __filename: asset => `var __filename = ${JSON.stringify(asset.name)};`,
--- a/src/visitors/globals.js
+++ b/src/visitors/globals.js
@@ -7,6 +7,7 @@ const VARS = {
     asset.addDependency('process');
     return 'var process = require("process");';
   },
+  global: () => 'var global = arguments[3];',
   __dirname: asset =>
     `var __dirname = ${JSON.stringify(Path.dirname(asset.name))};`,
   __filename: asset => `var __filename = ${JSON.stringify(asset.name)};`,
--- a/src/visitors/globals.js
+++ b/src/visitors/globals.js
@@ -7,7 +7,6 @@ const VARS = {
     asset.addDependency('process');
     return 'var process = require("process");';
   },
-  global: () => 'var global = typeof global === "object" ? global : typeof window === "object" ? window :typeof self === "object" ? self : this;',
   __dirname: asset =>
     `var __dirname = ${JSON.stringify(Path.dirname(asset.name))};`,
   __filename: asset => `var __filename = ${JSON.stringify(asset.name)};`,
--- a/src/builtins/hmr-runtime.js
+++ b/src/builtins/hmr-runtime.js
@@ -1,6 +1,5 @@
 var OVERLAY_ID = '__parcel__error__overlay__';
 
-var global = (1, eval)('this');
 var OldModule = module.bundle.Module;
 
 function Module(moduleName) {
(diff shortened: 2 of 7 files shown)


src/visitors/globals.js
```diff
@@ -7,7 +7,8 @@ const VARS = {
     asset.addDependency('process');
     return 'var process = require("process");';
   },
-  global: () => 'var global = (1,eval)("this");',
+  global: () =>
+    'var global = typeof global === "object" ? global : typeof window === "object" ? window : typeof self === "object" ? self : {};',
```
devongovett: I don't think this is necessarily very reliable. What if someone declares `window` or `self` in the local scope of their module?

CvX: That's a good point. Replaced it with a more reliable (and succinct!) solution: https://github.com/parcel-bundler/parcel/pull/1133/commits/1fe753ed5d276eeaf8619100d6bd22aa15eb27fe

devongovett: Yeah but won't `Function` constructor have the same problem as `eval` with CSP?

CvX: You're, right, my bad.
I don't know why I assumed that another variant of arbitrary code execution would fly. 🙃  Should have read [W3C Working Draft for CSP](https://www.w3.org/TR/CSP/#directive-script-src)…

About the case of `window` and `self` in a local scope – you're talking about a situation where there's a module that either:
1. uses `global` object, redefines `window`, and is used in a browser
2. uses `global` object, redefines `self`, and is used in a Web Worker

Do we have to choose between supporting those two cases and supporting CSP?

devongovett: Yeah... we may be able to get it working by moving that code to get the global object you originally added outside of each module and into the prelude where we can guarantee that nothing will mess with those variables. In there, the global object should be `this` (unless strict mode is enabled). Then we can attach it to `module` or something, and here assign `var global = module.global`, or maybe pass it into the module function. What do you think?

See this article for more info: https://www.contentful.com/blog/2017/01/17/the-global-object-in-javascript/

CvX:
@devongovett Moving acquisition of `global` to the prelude is a great idea. I've updated the PR – let me know what you think. It's using `this` since prelude seems to be in a non-strict scope. Is there any case where it would become strict?

Btw, it would be a good idea to run some of the tests in an actual browser environment, maybe via Puppeteer. I'll have to look into that. Some actual CSP tests would be nice.

devongovett:
There was one additional case to handle: when you redeclare `global` as a variable inside a module, e.g. `const global = 2`. This would cause a syntax error complaining that `global` was already defined (in the module function wrapper). I solved this by accessing `global` from `arguments[3]` when needed instead of always passing to the function.

RichardJECooke:
I use latest Parcel 2 and I still see `eval` in my JS build. How do I disable the `eval` so I can use Parcel to build my browser script that runs in Electron without warnings please?

Package:

```js
  "devDependencies": {
    "electron": "^39.2.2",
    "parcel": "^2.16.1",
    "@parcel/transformer-vue": "^2.16.1"
  },
```

Output JS:

```js
// eval may also be disabled via CSP, so do a quick check.
    var supportsSourceURL = false;
    try {
        (0, eval)('throw new Error("test"); //# sourceURL=test.js');
    } catch (err) {
        supportsSourceURL = err.stack.includes('test.js');
    }
```

PR #1135, opened 2018-04-04
Scope hoisting for ES6 and CommonJS modules

### LINKED ISSUE #392: 🙋 Tree Shaking
We should add tree shaking to Parcel in production mode so we have smaller bundle sizes. Because this requires cross-file optimizations, it will be harder to do with Parcel's parallel architecture. This basically means that tree-shaking must be done in a Packager.

There are a few approaches we could take:

* Use [Rollup](http://rollupjs.org). I tried creating a RollupPackager to replace the current JSPackager in production mode, which just runs everything through Rollup. This worked and was relatively simple to implement, but was extremely slow. e.g. a 3s build went to 15s. It also only really works well with ES6 modules, not CommonJS which is much more common (haha) still especially in Node modules.
* Use something like [common-shake](https://github.com/indutny/common-shake), which is a library for doing tree-shaking on CommonJS modules. I haven't tried it, but it looks neat and is a library rather than a full bundler.
* Implement something ourselves. Sounds like a lot of work, but we may be able to tune it specifically for Parcel's architecture, e.g. doing some work in the Asset workers, and some in the Packager.

Goals we should aim for:

* Support ES6 modules
* Support CommonJS modules
* Eliminate the module wrappers/runtime that JSPackager uses as much as possible by e.g. concatenating modules like Rollup does (scope hoisting)
* Be relatively fast still
* Support something like webpack 4's `sideEffects: false` [option](https://github.com/webpack/webpack/tree/next/examples/side-effects) in `package.json` to enable us to not parse or include all code that is imported but not used.

Would love to hear some feedback on this, and if you have ideas on how we should implement this or features we should support please comment below! 😍

### LINKED ISSUE #1104: [RFC] Support module concatenation (scope hoisting)
- `a.js`:
```js
import {foo} from './b'

console.log('foo', foo())
```

- `b.js`:
```js
import barFactory from 'bar-module'

export const foo = () => 'hello'
export const bar = () => barFactory
```

- `bar-module`:
```js
export default 'bar'
```

### Build

- `bundle-parcel.js` (simplified)
```js
require = (function (modules, cache, entry) {
  // prelude.js
})({
  0: [
    function(require, module, exports) {
      const foo = require('./b').foo;

      console.log('foo', foo())
    },
    {'./b': 1}
  ],
  1: [
    function(require, module, exports) {
      const barFactory = require('bar-module').default;

       exports.foo = () => 'hello'
       exports.bar = () => barFactory
    },
    {'bar-module': 2}
  ],
  2: [
    function(require, module, exports) {
      exports.default = 'bar'
    },
    {}
  ]
}, {}, [0])
```

- hypothetical `bundle-module-concat.js` using module concatenation :
```js
void function () {
  // bar-module
  const $module$bar_module$default = 'bar'

  // b.js
  const barFactory = $module$bar_module$default

  const $module$b_js$foo = () => 'hello'
  const $module$b_js$bar = () => barFactory

  // a.js
  const foo = $module$b_js$foo

  console.log('foo', foo())
}()
```

### Minification using `babili` 0.1.3 (prettified)

- `bundle-parcel.js` :
  - **Input size**: 2.36 kB (gzip: 870 B) 
  - **Minified size**: 803 B (gzip: 463 B) 
  - **Saved**: 65.97% (gzip: 46.78%)

```js
require = function(a, b, c) {
    function d(c, f) {
        function g(a) {
            return d(g.resolve(a))
        }

        function h(b) {
            return a[c][1][b] || b
        }
        if (!b[c]) {
            if (!a[c]) {
                var i = "function" == typeof require && require;
                if (!f && i) return i(c, !0);
                if (e) return e(c, !0);
                var j = new Error("Cannot find module '" + c + "'");
                throw j.code = "MODULE_NOT_FOUND", j
            }
            g.resolve = h;
            var k = b[c] = new d.Module(c);
            a[c][0].call(k.exports, g, k, k.exports)
        }
        return b[c].exports
    }
    var e = "function" == typeof require && require;
    d.isParcelRequire = !0, d.Module = function(a) {
        this.id = a, this.bundle = d, this.exports = {}
    }, d.modules = a, d.cache = b, d.parent = e;
    for (var f = 0; f < c.length; f++) d(c[f]);
    return d
}({
    0: [function(a) {
        const b = a("./b").foo;
        console.log("foo", b())
    }, {
        "./b": 1
    }],
    1: [function(a, b, c) {
        const d = a("bar-module").default;
        c.foo = () => "hello", c.bar = () => d
    }, {
        "bar-module": 2
    }],
    2: [function(a, b, c) {
        c.default = "bar"
    }, {}]
}, {}, [0]);
```

- `bundle-module-concat.js` :
  - **Input size**: 297 B (gzip: 164 B) 
  - **Minified size**: 54 B (gzip: 74 B) 
  - **Saved**: 81.82% (gzip: 54.88%)

```js
void function(){
  console.log('foo',(()=>'hello')())
}()
```

### Using Prepack

- [`bundle-parcel.js`: doesn't work](https://prepack.io/repl.html#E4UwjgrglqAEC8sAUAzCA7AxgFyge3WQFs8ATCAGxAGcAaWTAQ0wAsR6R1tgBPASlgBvAFCxYAenGwAyowBuIWNjaxQkGIpTA8RWAAdQc-BGqwARhlJUleJSyinMFPNQhwoKWI3Q9RsOYzA+obG1ABK4NBwiNg8eiB4nmpRivBpsABEaFi4BBmwAGQFqpEaANzCftk4+IToIADuEeqgSOiMROywAFYQRPGkAiJiYh7IAIRMrCAA2u2dALpDfiOwY0jjJORU1HMdIEtCK6sSUmMNikzo6HjYsChQ6KR2iluUig1Qyo+weG5rXBAwHaFFgREYel+wGOq0kDGYKl6-RsLwYblAXFgAHNnGZGKDkho1iAAHSoiiMah3CxPKgwkZw5SMO4NSleUikEDPbC2ZSKPSMLGk+liAJBTDozjYZopBBKOIJJKlaLpLIYGp5QrFQmgConEbrcZIgZatHADHS5UgZb61agbBuQgS81SmUaNr7ejcCDWvW2gC+lVtp1gAElPHy4IFFLc2EEaVYaL9CN9TAKhaidZptLo+SKQwYQEY-qYCIoHLBqPIuSiAOSF4smN2gWtkiLxZl2CuU-Nw8E+JRQTqmNmRxTR8yWHawDC4UF8sFkd5rUwoP5PKG9qQXWAgAAeLEYJjuC6zDEPjxJ+fWDdCzetR2DYntjuCRbvVo9nS9wB9fD9+qBvmYq7uacr1A0sAAKLmngwBILWADC3g3HcDwbm81gADq1rWsAANSwPMiiEbWOG1v++ZAsAJKYGQqSwLWACyADyAAiACqAAyUEAPoAHIsQAKrxABiLEcfxbG1gBJzKNokHUbJYhAfqzhMBQ94kqA1B4BQChyjpekKBU9IgZhDFTGweyLOBjREY0WlMUuVBfr6QYnBZuzEQsMwAAwLLR+IUEgFkkvuehwdgdCwOp+L3vQFmJS5pIRVF1CUasqkjC+wLwtMNkHOFe6RcA0Wmas1S5IQcWaZ+e58MMJy5XUjmfrVWlGfpIBIA1mUjNlYhVbUJS6d1vWNfSLWLtsNCFb5ACMvl7gssAAD5rbAe6yYGKkecNBCwM5s2hSl-H7DaIypiSUDPIgFnnZ0snXQm1iIBB97PfY1DFaV0VyoI-p+tlH1Wjd1AAAqBJgIB1S0DHeiAfqg-DJLHcuiDo1QyNtajXlyl5ONNGDVmWQiSN+CjKQkgKFpyreJafftcHICBUByn5ZRrLAAA8u5cLwJJUOgWLKFzUD4fhl1iFT7pSrwMxQEswMeXCLEKOat2KAuzp02enzKF2pgQcmIB+NNsu6sI-p8EgTV+QAXLAMzHAd6BIFmyWzRwJXpdLIx0egVL3HgtiIFmCEkuIZgUSSa54BVJyB2NpLOFiCHx7W9Dx0gfB8Mc-q0McgittHtZOwtu2wAsRdiAtTsu5V6rVR7Vpe+8Pt-Rlj5JwQwd4sAonMDyvCGZ+tYDwAtBZsecigR4UNgicnGlZU-fHcq5wgAB8jFsBQzi1iKq-RSSA+bwI8C7wPQ84HBvhZbXIwl1PM9OwATFXNd+O-Deu83tRW7w3blQTufse6rBPj9OeC87iIAnoEI+j9i5f2tvQQG9B-JLGEEAA)
- [`bundle-module-concat.js`](https://prepack.io/repl.html#G4ewlgJgBAZgrgOwMYBcwgVAFASigbwCgooB6UqAIwEMAnAWgFsQI4AbAU2KiQwGcUUACTNWnITVoB9Uew5CIHGNXaCAvFADkkzYW7kqAOgBWfbrwQCqdAGLVUIWgE8oGkSzkS6Mj+MXLVPRILK3cxeUopUyEYEBBXbDw1AD4tAAsONjYQXWD+QTDPSOjJBNxXVMk7B2cgsgpqEzM8y0FY+LdZcWK+GLi6kJBOQ2yAcyxNds0AGlg43BxCAF9cIA) : 
```js
console.log("foo", "hello");
```

-------

### Implemented by

- [Webpack as a plugin](https://webpack.js.org/plugins/module-concatenation-plugin/), [example](https://github.com/webpack/webpack/tree/master/examples/scope-hoisting) (they have really good docs about it)
- [Rollup](https://github.com/rollup/rollup)
- [Google Closure Compiler](https://github.com/google/closure-compiler) (we have a Parcel plugin to do this but it's pretty heavy and relies on the Java JNI)

### Pro

- Faster runtime startup
- Smaller bundles
- Easier to debug stacktraces
- Implements tree-shaking by design and probably in a better way using minifiers (side-effects)

### Cons

- A real pain to implement correctly given the current state of the ecosystem (es6/cjs, see Webpack bailouts)
- Makes parallelism more difficult

----

Some questions :

- How would we expose this to the user? Only on prod build? Behind a CLI option?
- How would we implement it technically?
- Is it worth it?

fathyb:
Closes #1104.
Fixes #392.

- [x] `eval`
- [x] UMD
- [x] Pure ES6 support
- [x] Import CommonJS from ES6
- [x] Import ES6 from CommonJS
- [x] Dynamic `import`
- [x] Wildcards (`export * from`)
  - [x] ES6
  - [x] CommonJS
- [x] minify using Uglify
- [x] update the non-HMR tests to use scope-hoisting
- [x] make scope hoisting the default when HMR is disabled
- [x] [`sideEffects`](https://github.com/stereobooster/package.json#sideeffects) package field (partial support)
- [x] Source map

Output : 

- `a.js`
```js
import {hello, world} from './b'

console.log(`${hello} ${world}`)
```

- `b.js`
```js
export * from './c'
```

- `c.js`
```js
const localHello = 'hello'
export {localHello as hello}

export const world = 'world'
export const bar = 'tree-shake me if you can'
```

#### `--no-minify`
```js
(function () {
  /* ASSET: 5 - c.js */

  var $5$export$hello = 'hello';

  var $5$export$world = 'world';

  var $5$export$bar = 'tree-shake me if you can';

  /* ASSET: 3 - b.js */


  /* ASSET: 1 - a.js */
  console.log($5$export$hello + ' ' + $5$export$world);
})();
```

#### `--minify`
```js
console.log("hello world");
```


devongovett:
Oh, awesome! I was actually working on something similar. Will take a look in a bit

src/visitors/hoist.js
```diff
@@ -177,9 +187,124 @@ module.exports = {
...
+      );
+    }
+  },
+  ImportDeclaration(path, asset) {
+    let {source, specifiers} = path.node;
+
+    if (source && specifiers.length > 0) {
+      if (BREAK_COMMON_JS) {
+        path.replaceWith(
+          t.variableDeclaration(
```
devongovett: I think it should be possible to avoid these extra variables by just renaming the local references to point to the imported one, e.g.

```javascript
path.scope.rename(specifier.local, IMPORT_NAME_HERE);
```

src/assets/JSAsset.js
```diff
@@ -173,7 +174,10 @@ class JSAsset extends Asset {
 
     return {
       js: code,
-      map: this.sourceMap
+      map: this.sourceMap,
+      '@reflect': {
```
devongovett: Rather than adding these to `generated` and then removing them later, you could just add them to `asset.cacheData`. That already gets written to the cache and you should be able to access it from the packager.

src/visitors/hoist.js
fathyb: Better! Fixed

src/assets/JSAsset.js
fathyb: I tried `cacheData` and I didn't get it to work (`cacheData.exports` is empty). Should I manually update it? I pushed it

src/packagers/JSConcatPackager.js
```diff
@@ -54,13 +54,27 @@ class JSConcatPackager extends Packager {
...
 
-      js = js
-        // $[asset.id]$named_import$a_js => $[module.id]$named_exports
-        .split('$' + asset.id + '$named_import$' + t.toIdentifier(dep.name))
-        .join('$' + mod.id + '$named_export')
-        // $[asset.id]$expand_exports$a_js => $parcel$expand_exports([module.id])
-        .split('$' + asset.id + '$expand_exports$' + t.toIdentifier(dep.name))
-        .join('$parcel$expand_exports(' + mod.id + ',' + asset.id + ')');
+
+      if (dep.isES6Import) {
```
devongovett: We track imports to see if they are from ES6, as well as exports to mark a module as ES6. If they are both ES6, then we can just replace the import names with exports. If a commonjs module is imported with ES6, then the `default` import resolves to `module.exports`, and other names resolve to `module.exports.NAME`.

src/packagers/JSConcatPackager.js
```diff
@@ -154,36 +168,36 @@ class JSConcatPackager extends Packager {
...
-
-          // if there is at least two occurences of the exports binding
-          if (this.buffer.split(exportBinding).length > 2) {
-            return `${binding} ${exportBinding}.${name} =`;
-          }
-
-          return binding;
-        })
-    );
+    // super.write(
```
devongovett: I think these re-exports are going to be problematic since they are supposed to be live bindings rather than a copy. Commented for now while we figure it out.

src/visitors/hoist.js
```diff
@@ -183,122 +182,201 @@ module.exports = {
...
-              asset.id +
-              '$named_import$' +
-              t.toIdentifier(source.value) +
-              '$' +
-              specifier.imported.name
-          )
+    // For each specifier, rename the local variables to point to the imported name.
+    // This will be replaced by the final variable name of the resolved asset in the packager.
+    for (let specifier of path.node.specifiers) {
+      if (t.isImportDefaultSpecifier(specifier)) {
```
devongovett: support more types of specifiers

src/visitors/hoist.js
```diff
@@ -183,122 +182,201 @@ module.exports = {
...
+          specifier.local.name,
+          getName(
+            asset,
+            'import',
+            path.node.source.value,
+            specifier.imported.name
           )
         );
+      } else if (t.isImportNamespaceSpecifier(specifier)) {
+        path.scope.rename(
```
devongovett: importing a namespace e.g. `import * from 'blah'` just does the same thing as a commonjs require.

src/visitors/hoist.js
```diff
@@ -183,122 +182,201 @@ module.exports = {
...
     }
+
+    path.remove();
   },
-  ExportNamedDeclaration(path, asset) {
-    let {declaration, source, specifiers} = path.node;
 
-    if (!source) {
-      let declarations;
+  ExportDefaultDeclaration(path, asset) {
```
devongovett: Added default export support

src/visitors/hoist.js
```diff
@@ -183,122 +182,201 @@ module.exports = {
...
+    asset.cacheData.exports[identifier.name] = 'default';
+
+    // Mark the asset as an ES6 module, so we handle imports correctly in the packager.
+    asset.cacheData.isES6Module = true;
+  },
+
+  ExportNamedDeclaration(path, asset) {
+    let {declaration, source, specifiers} = path.node;
+
+    if (source) {
```
devongovett: Support re-exports, e.g. `export {foo} from 'bar'`. Again, this will be problematic I realized, and the code to insert a variable here probably won't work right.

src/visitors/hoist.js
```diff
@@ -183,122 +182,201 @@ module.exports = {
...
         );
       }
 
-      if (declarations.length) {
-        path.replaceWith(t.variableDeclaration('var', declarations));
+      path.remove();
+    } else if (declaration) {
+      path.replaceWith(declaration);
+
+      if (
```
devongovett: Added support for a bunch of different types of declarations. Still more to support, see todos.

src/visitors/hoist.js
fathyb: for stuff like this I thought about a Babel transform we could do in the worker, where basically we would replaced wildcard imports to named if possible (bailout if the namespace if called). We could also do the same to turn for example `const {foo} = require('foo')` to `import {foo} from 'foo'`, it may simplify our logic an should be safe enough

src/visitors/hoist.js
```diff
@@ -183,122 +182,201 @@ module.exports = {
...
+  asset.cacheData.exports[getName(asset, 'export', exported.name)] =
+    exported.name;
 
-  return NAMED_EXPORT_TEMPLATE({
-    NAME: t.identifier(varName),
-    EXPORTS: getExportsIdentifier(asset),
-    BINDING: name,
-    INIT: init
-  }).declarations[0];
+  path.insertAfter(
```
devongovett: For each export, we don't create a new variable, we just rename the local variable within the module to the exported name. Also, we add a property on the exports object for use in namespaces and commonjs imports of ES6 modules.

devongovett:
Did a few things, and added some inline comments to describe them. Hope you don't mind me committing to your branch, sorry about that. 😜

Currently the only problematic cases are re-exports:

* `export * from 'foo'`
* `export {foo} from 'bar'`

Currently, we were handling those as variable replacements: we'd make a new variable in the context of the module that is doing the re-exporting, and then dependencies of that module would use the renamed variable instead of the original one.

The problem with this is that if the value of the original binding changes, the renamed one will not. So what we really need to do is resolve re-exported names to their original names in the modules that use them.

Here's an example that illustrates the problem:

a.js
```javascript
export var foo = 2;
export function changeFoo(x) {
  foo = x;
}
```

b.js
```javascript
export * from './a.js'
```

c.js
```javascript
import {foo, changeFoo} from './b.js';

console.log(foo) // => 2
changeFoo(10);
console.log(foo) // => 2 (should be 10)
```

src/visitors/hoist.js
devongovett: Yeah that could be done. I wonder if minifiers will be able to do that for us though. We're already creating an object with the namespace for every module, so it seems like the minifier should be able to statically resolve those names off the object...

fathyb:
Thanks for the update, saved me some time 👍  I will look into details later but I noticed it broke tree-shaking and `export *` support, it now compiles to : 
```js
(function () {

/* ASSET: /Users/fathy/Documents/Git/test-parcel/c.js */
var $5$exports = {};
var $5$export$foo = 'hello';
$5$exports.foo = $5$export$foo;
var $5$export$bar = 'tree-shake me if you can';
$5$exports.bar = $5$export$bar;

/* ASSET: /Users/fathy/Documents/Git/test-parcel/b.js */
export * from './c';

/* ASSET: /Users/fathy/Documents/Git/test-parcel/a.js */
console.log('foo', $3$export$foo);
})();
```

IMHO we should not do `$5$exports.bar = $5$export$bar;` unless `$5$exports` is used somewhere else in the code. Expect for Closure, minifiers aren't smart enough to trim it if it has property assignments (tried using Uglify 3 and Babili). We should really focus on making the code as simple as possible to minify.

About the `foo` `changeFoo` situation I think we could just parse and transform the final bundle using Babel, we could then correctly replace references to a named export to the original variable name. If we end up using `babel/minify` we could parse the final code using Babylon, transform it and directly pass it to it. So we get clean, non-regexy way of "expanding" exports and stuff like that.

devongovett:
Yeah I commented out the `export *` support for now because of the above mentioned reason. I played around with it for a bit and realized that we need to do some further renaming in the packager once we have all the modules.

Huh, that's strange about it not removing the unused object. When I tested with uglify on another input, it did seem to remove them. On further testing, it looks like uglify will remove the object but only if there is only one property added to it? Weird.

Here's an example. `test3.js` has one export and it's unused. uglify totally removes it from the output. As soon as you add another property to `$3$exports`, it won't remove it anymore.

```javascript
(function () {

/* ASSET: test3.js */
var $3$exports = {};
function $3$export$default() {
  return 4;
}

$3$exports.default = $3$export$default;

/* ASSET: index.js */
var $1$exports = {};

function $1$var$add(a, b) {
  return a + b;
}

var $1$export$foo = $1$var$add(1, 2);
$1$exports.foo = $1$export$foo;

console.log($1$export$foo);
})();
```

output:

```javascript
!function(){var o=1+2;console.log(o)}();
```

fathyb:
I pushed a commit with a fix for the wildcard, but it breaks commonjs, I will fix it later. It's messy and I reverted some of your changes, feel free to revert the commit 😄 

What I'm trying to do :
- resolve the exports ahead of time using Babel (`src/transforms/concat.js`) to get a fully optimisable interop with cjs, umd, es6
- find if an import should be replaced with `$id$export$name` or `$id$exports.name`
- add the `$id$exports.name = $id$export$name` assignment only when needed (based on your comment on Uglify and multiple props)


#### Output
- `a.js`
```js
import {foo} from './b'

console.log('foo', foo)
```

- `b.js`
```js
export * from './c'
```

- `c.js`
```js
export let foo = 'hello'
export const bar = 'trim me'

export function changeFoo(next) {
    foo = next
}
```

- `bundle.js` (love the prelude-less bundles 😍)
```js
(function () {

  /* ASSET: /~/test-parcel/c.js */
  var $5$export$foo = 'hello';
  var $5$export$bar = 'trim me';

  function $5$export$changeFoo(next) {
    $5$export$foo = next;
  }

  /* ASSET: /~/test-parcel/b.js */


  /* ASSET: /~/test-parcel/a.js */
  console.log('foo', $5$export$foo);
})();
```

devongovett:
Thought it would be useful to summarize all of the possible cases of import/export syntax, and what happens for both ES6 and commonjs modules. Eventually we should have tests for all combinations of these.

Imports:

| syntax                           | action                              |
| -------------------------------- | ----------------------------------- |
| `import foo from 'bar'`            | rename `foo` to `$bar_id$export$default`    |
| `import {foo} from 'bar'`          | rename `foo` to `$bar_id$export$foo`        |
| `import {foo as bar} from 'bar'`   | rename `bar` to `$bar_id$export$foo`        |
| `import * as foo from 'bar'`       | rename `foo` to `$bar_id$exports`           |
| `var foo = require('bar')`          | `var foo = $bar_id$exports`                       |

Exports:

| syntax                               | es6                                                                              | commonjs                                           |
| ------------------------------------ | -------------------------------------------------------------------------------- | -------------------------------------------------- |
| `export default 4`                   | `var $id$export$default = 4`                                                     | `$id$exports.default =  $id$export$default`        |
| `export default function foo() {}`   | rename `foo` to `$id$export$default`                                             | `$id$exports.default = $id$export$default`         |
| `export var foo = 4`                 | rename `foo` to `$id$export$foo`                                                 | `$id$exports.foo =$id$export$foo`                  |
| `export {foo}`                       | rename `foo` to `$id$export$foo`                                                 | `$id$exports.foo = $id$export$foo`                 |
| `export {foo} from 'bar'`            | rename `$id$export$foo` to `$bar_id$export$foo`                                  | `$id$exports.foo = $bar_id$export$foo`             |
| `export * from 'bar'`                | rename all exports in bar from `$id$export$NAME` to `$bar_id$export$NAME`        | `Object.assign($id$exports, $bar_id$exports)`      |
| `module.exports = 4`                 | `$id$exports = 4`                                                                | `$id$exports = 4`                                  |
| `exports.foo = 4`                    | `$id$exports.foo = 4`                                                            | `$id$exports.foo = 4`                              |

devongovett:
I changed the way export all is implemented a bit. Rather than making identifiers with the ids of the two modules, I just keep track of whether a dependency was produced by an export all and replace the variables in all subsequent modules pointing to exports from the parent module. Since assets are added to the bundle in order of use, this should be safe.

The other thing I did was create the commonjs exports all at once at the end of the file if possible rather than creating an empty object at the start and incrementally assigning to it. This should make it possible for minifiers to prune them more easily. We do incrementally assign to it if the `module` or `exports` variables are referenced in the module though, so you could have both commonjs and es6 in the same module (probably more common than you'd think because babel).

The only problematic one is again export all, which requires an `Object.assign` to add the imported module to exports. The problem is that this causes minifiers to include both exports objects and all of their functions even if neither are used anywhere else. We might need to do that part in the packager only if we detect that a module with export all is referenced by a commonjs module or namespace import.

All these string variable replacements do seem somewhat dangerous, but since we're using unique names should be relatively safe. I somewhat like the idea of running babel over everything again in the packager (and we might anyway for minification), but I'm worried about performance. Should probably measure on some larger apps once we get this working fully.

devongovett:
Actually `Object.assign` is not even gonna work for export all since they need to be live bindings... Here's what babel does for `export * from 'foo'`:

```javascript
var _foo = require('foo');

Object.keys(_foo).forEach(function (key) {
  if (key === "default" || key === "__esModule") return;
  Object.defineProperty(exports, key, {
    enumerable: true,
    get: function get() {
      return _foo[key];
    }
  });
});
```

Also apparently `default` is excluded from re-exports...

fathyb:
Turns out Babylon AST can be fully (de)serialised using JSON 😮 
```js
const ast = JSON.parse(JSON.stringify(babylon.parse(code)))
```

DeMoorJasper:
@fathyb I've tested this inside my treeshaking experiment a lil while back and it had circular properties when I've tested it (might have been a non-babel ast that caused the circular props, as I didn't only stringify babel asts). Just wanted to let you know as you might run into the same issue in the future.

fathyb:
@devongovett  Love the syntax matrices 👍 

@DeMoorJasper I was a bit afraid of this 😕 thanks for letting me know

I tweaked the Babel packager implementation to generate correct exports using the scope (`Binding.constant`). The pure ES6 code is still perfectly tree-shaked, and the CommonJS one is a bit better. The exports are generated on the fly when needed and it chooses property assignment or getter based on the var side-effects : 

- `a.js`
```js
const {foo} = require('./b')

require('./d')

console.log('foo', foo)
```

- `b.js`
```js
export * from './c'

export const some = 'thing'
```

- `c.js`
```js
export let foo = 'hello'
export const bar = 'trim me'

export function changeFoo(next) {
    foo = next
}
```

- `d.js`
```js
console.log('require(./b)', require('./b'))
console.log('require(./c)', require('./c'))
```

- `bundle.js`
```js
(function () {

  /* ASSET: /~/c.js */
  var $7$export$foo = 'hello';
  var $7$export$bar = 'trim me';

  function $7$export$changeFoo(next) {
    $7$export$foo = next;
  }

  /* ASSET: /~/b.js */

  var $3$export$some = 'thing';

  /* ASSET: /~/d.js */
  var $3$exports = {
    changeFoo: $7$export$changeFoo,

    get foo() {
      return $7$export$foo;
    },

    bar: $7$export$bar,
    some: $3$export$some
  };
  console.log('require(./b)', $3$exports);
  var $7$exports = {
    changeFoo: $7$export$changeFoo,

    get foo() {
      return $7$export$foo;
    },

    bar: $7$export$bar
  };
  console.log('require(./c)', $7$exports);

  /* ASSET: /~/a.js */
  var $1$var$_require = $3$exports,
      $1$var$foo = $1$var$_require.foo;

  var $5$exports = {};
  $5$exports;

  console.log('foo', $1$var$foo);
})();
```

devongovett:
Yeah I was worried about `toIdentifier` for filenames. Maybe instead we could just hash the filename?

fathyb:
It's getting better! By using Babel's scope we can get `assign`/copy free and deep wildcard with tree-shaking. Example (no exports object are created for the wildcards, and a variable is used for `foo` and `some` while `moduleName` uses a property access) : 

- `a.js`
```js
import {foo, moduleName, some} from './d'

console.log('foo', foo, moduleName, some)
```

- `b.js`
```js
export * from './c'

export const some = 'thing'
```

- `c.js`
```js
export let foo = 'hello'
export const bar = 'trim me'

export function changeFoo(next) {
    foo = next
}
```

- `d.js`
```js
export * from './e'
```

- `e.js`
```js
export * from './b'

exports.moduleName = 'e.js'
```

- `bundle.js`
```js
(function () {

  /* ASSET: /~/c.js */
  var $9$export$foo = 'hello';
  var $9$export$bar = 'trim me';

  function $9$export$changeFoo(next) {
    $9$export$foo = next;
  }

  /* ASSET: /~/b.js */
  var $7$export$some = 'thing';

  /* ASSET: /~/e.js */
  var $7$exports = {};

  $7$exports.moduleName = 'e.js';

  /* ASSET: /~/d.js */

  /* ASSET: /~/a.js */
  console.log('foo', $9$export$foo, $7$exports.moduleName, $7$export$some);
})();
```

- `bundle.min.js` (Uglify)
```js
!function(){
  var o="hello";
  var e={
    moduleName: "e.js"
  };
  console.log("foo", o, e.moduleName, "thing")
}();
```
- `bundle.min.js` (`babel/minify`)
```js
(function() {
  var b = "hello",
    e = {};
  (e.moduleName = "e.js"), console.log("foo", b, e.moduleName, "thing");
})();
```

src/packagers/JSConcatPackager.js
```diff
@@ -72,6 +72,9 @@ class JSConcatPackager extends Packager {
 
       // If this was an ES6 export all (e.g. export * from 'foo'), resolve to the original exports.
       if (dep.isExportAll) {
+        asset.cacheData.exports[`$${asset.id}$exports`] = `$${mod.id}$exports`;
```
devongovett: Is this true? What if you re-exported multiple modules?

```javascript
export * from './foo';
export * from './bar';
```

Then `$id$exports` should be the combination of both modules...

fathyb: Good catch, it's a tricky case, I'll prolly fix it by fixing the commonjs import

src/packagers/JSConcatPackager.js
```diff
@@ -72,17 +69,17 @@ class JSConcatPackager extends Packager {
 
       // If this was an ES6 export all (e.g. export * from 'foo'), resolve to the original exports.
       if (dep.isExportAll) {
-        asset.cacheData.exports[`$${asset.id}$exports`] = `$${mod.id}$exports`;
-        this.exports.set(`$${asset.id}$exports`, `$${mod.id}$exports`);
+        // asset.cacheData.exports[`$${asset.id}$exports`] = `$${mod.id}$exports`;
+        // this.exports.set(`$${asset.id}$exports`, `$${mod.id}$exports`);
```
devongovett: commented these for now. they were causing issues for me, and don't seem quite right anyway.

src/transforms/concat.js
```diff
@@ -122,7 +125,7 @@ module.exports = (code, exports, moduleMap) => {
         // let id = Number(match[1]);
 
         if (!path.scope.hasBinding(name) && !addedExports.has(name)) {
-          let {bindings} = path.scope;
+          let exports = moduleMap.get(+match[1]).cacheData.exports;
```
devongovett: Previously, all variables in the current scope would be placed on the exports namespace object, not just the actual exports of the module. Changed to get the exports from the module.

devongovett:
Added a few more improvements, along with a test suite which should test most of the cases in the table above. A few of the tests are currently failing, so we'll need to fix those and add more tests as well.

src/transforms/concat.js
```diff
@@ -91,6 +91,24 @@ module.exports = (code, exports, moduleMap) => {
...
+      }
+
+      let match = object.name.match(EXPORTS_RE);
+      if (match) {
+        let exportName = '$' + match[1] + '$export$' + property.name;
+        if (path.scope.hasBinding(exportName)) {
+          path.replaceWith(t.identifier(exportName));
+        }
+      }
+    },
```
devongovett: This means code like this:

```javascript
import * as test from './foo';

console.log(test.b);
```

will compile to:

```javascript
var $4$export$b = 1;

console.log($4$export$b);
```

No namespace object needed!

fathyb: Awesome!

fathyb:
good job on the tests! 👍 I reworked the wildcards to use a `wildcard` map instead of the `exports` list

test/integration/scope-hoisting/es6/re-export-multiple/c.js
```diff
@@ -1 +1,2 @@
 export var foo = 2, bar = 3;
+export * from './d'
```
fathyb: not sure if this was intended?

devongovett: ah no, good catch

src/visitors/hoist.js
```diff
@@ -355,16 +327,6 @@ function addExport(asset, path, local, exported) {
...
-  if (path.scope.hasGlobal('module') || path.scope.hasGlobal('exports')) {
-    path.insertAfter(
-      EXPORT_ASSIGN_TEMPLATE({
-        EXPORTS: getExportsIdentifier(asset),
-        NAME: t.identifier(local.name),
-        LOCAL: getIdentifier(asset, 'export', exported.name)
-      })
-    );
-  }
-
```
devongovett: I think removing this causes an issue where if you reference commonjs exports before an es6 export it will insert the es6 export in the wrong place (before it is initialized, so it will be undefined). I've added a failing test case for that.

```javascript
(function () {
  var $4$exports = {
    b: $4$export$b
  };


  /* ASSET: 4 - b.js */
  $4$exports.a = 2;
  var $4$export$b = 3;

  /* ASSET: 2 - a.js */
  var $2$var$_require = $4$exports,
      $2$var$a = $2$var$_require.a,
      $2$var$b = $2$var$_require.b;

  output = $2$var$a + $2$var$b;
})();
```

As you can see, `$4$export$b` is referenced before it is initialized since the commonjs export comes first.

fathyb: The easiest way to fix this would be to use getters for all exports (and ignore if the binding is constant). 

```js
var $4$exports = {
  get b() { return $4$export$b }
};
```

I have no idea if it has an impact on the final size though.

devongovett: Yeah that would work... All those getters would probably hurt minifiers being able to statically resolve the values of object properties and potentially get rid of the object all together though right?

Why not just do what we did before which was insert the empty object at the top if needed, and assign to it as you go?

kzc: Not necessarily:
```
$ echo 'var o = {a: 1, b: 2}; console.log(o.a, o.b, o.a + o.b);' | bin/uglifyjs --toplevel -mc passes=2
console.log(1,2,3);
```

kzc: Curiously without the IIFE it will be simplified:
```
$ cat t2.js 

    var willNeverBeTrimmed = {};
    willNeverBeTrimmed.one = 'it does not';
    willNeverBeTrimmed.two = 'get trimmed';
    console.log(willNeverBeTrimmed.one);
    function nothingToDoWithTheRest() {
        this_fn_should(make_uglify(bailout));
    }

$ cat t2.js | bin/uglifyjs --toplevel -mc passes=3
console.log("it does not");
```
However, with a surrounding IIFE it is not simplified. That's probably a minor scope issue that could be reported to the uglify project.

It's worth noting that Rollup uses objects for its default exports and handles some real world cases pretty well:

https://github.com/rollup/rollup/issues/501#issuecomment-355654923

The only real point of my comments is that there are more optimization opportunities without getters.

kzc: @fathyb I'd recommend using the following uglify options for advanced optimization:
```js
compress: {
    passes: 3,
    pure_getters: true, // avoid if code has getters with side effects
    toplevel: true, // not needed if all code wrapped in function scope
    unsafe: true, // optional: actually safe for sane JS not modifying builtins
}
```
Keep in mind that each additional pass increases the minification time. Something to be aware of when selecting defaults.

src/visitors/hoist.js
```diff
@@ -330,7 +330,10 @@ module.exports = {
     } else if (declaration) {
       path.replaceWith(declaration);
 
-      let identifiers = t.getBindingIdentifiers(declaration);
+      let identifiers = t.isIdentifier(declaration.id)
+        ? [declaration.id]
+        : t.getBindingIdentifiers(declaration);
```
fathyb: Did this because `getBindingIdentifiers` also returns `params`, it caused arguments to be treated as exports

fathyb:
I pushed a basic wildcard support for CommonJS using the `/*#__PURE__*/` annotation. This cuts the size cost of the assignments.

```js
(function () {
  function $parcel$exportWildcard(dest, source) {
    Object.keys(source).forEach(function (key) {
      if (key !== 'default') {
        Object.defineProperty(dest, key, {
          enumerable: true,
          get() {
            return source[key];
          }
        });
      }
    });

    return dest;
  }

  var a = {foo: 'bar'}
  var b$0 = {foo: 'baz'}

  var b = /*#__PURE__*/$parcel$exportWildcard(b$0, a);

  console.log('Foo', b.foo)
})()
```

Because `b` is referenced the output is : 
```js
!function(){var o=function(o,e){return Object.keys(e).forEach(function(n){"default"!==n&&Object.defineProperty(o,n,{enumerable:!0,get:()=>e[n]})}),o}({foo:"baz"},{foo:"bar"});console.log("Foo",o.foo)}();
```

If `b.foo` is changed to `a.foo` the output becomes : 
```js
console.log("Foo","bar");
```

If the  `/*#__PURE__*/` comment get removed the output gets big again : 
```js
!function(){var o,e,f={foo:"bar"};o={foo:"baz"},e=f,Object.keys(e).forEach(function(f){"default"!==f&&Object.defineProperty(o,f,{enumerable:!0,get:()=>e[f]})}),console.log("Foo",f.foo)}();
```

fathyb:
⚠️ I force pushed to rebase and resolved conflicts with `master`.

fathyb:
🚀 

screen shot 2018-04-21 at 20 43 28

test/integration/scope-hoisting/commonjs/require-re-export-default/b.js
```diff
@@ -1,2 +1,2 @@
-export foo from './c';
+export {default as foo} from './c';
```
fathyb: Not sure if `export foo from './c'` is valid, Babylon parses it but Babel throws. VSCode also shows invalid syntax errors.

test/integration/scope-hoisting/commonjs/require-re-export-namespace/b.js
```diff
@@ -1,2 +1,4 @@
-export * as c from './c';
```
fathyb: Same as above

devongovett:
Awesome! Next step is to run all of the tests and make sure they all pass with scope hoisting enabled (except HMR which won’t work). Then we can enable scope hoisting by default whenever HMR is disabled.

src/visitors/hoist.js
```diff
@@ -209,7 +209,10 @@ module.exports = {
...
     let {callee, arguments: args} = path.node;
 
-    if (t.isIdentifier(callee, {name: '$parcel$exportWildcard'})) {
+    if (
+      t.isIdentifier(callee, {name: '$parcel$exportWildcard'}) ||
+      asset.package.sideEffects === false
+    ) {
       // This hints Uglify and Babel that this CallExpression does not have any side-effects.
       // This will make unsused CommonJS wildcards removed from the minified builds.
       path.addComment('leading', '#__PURE__');
```
devongovett: probably only want to add this to top-level calls right? e.g. not calls inside another function.

devongovett: Seems like that should be removed still with `sideEffects: false`. We know there are no side effects because of that flag, so removing the outer call should be safe. And since the inner one is inside a function called by the outer one, that should be safe too.

I was mostly concerned with marking calls inside function declarations as pure:

```javascript
function test() {
  return foo();
}
```

In that example, `foo` shouldn't be marked as pure since it is inside another function.

Maybe rather than checking if it's a top-level statement you could check the scope instead to make sure the call is not inside a function/class method scope? i.e. you could have a call inside a block scope (e.g. if statement) in the top-level scope that should still be marked as pure.

src/packagers/JSConcatPackager.js
```diff
@@ -0,0 +1,249 @@
...
+            asset.cacheData.exports[key] = id;
+            this.exports.set(
+              key,
+              this.exports.get('$' + mod.id + '$export$' + id) || exp
+            );
+          }
+        }
+      }
+
+      if (dep.isES6Import) {
```
devongovett: let's see if we can move this mess into the babel plugin instead of doing this string manipulation

fathyb:
I just pushed fixes to make [react-news](https://github.com/echenley/react-news) fully compile and run!
It includes a lot a potentially ~problematic~ complex libraries like Firebase, moment, react-router.

screen shot 2018-04-24 at 00 13 42

src/packagers/JSConcatPackager.js
```diff
@@ -0,0 +1,249 @@
...
+    if (this.needsPrelude) {
+      if (this.bundle.entryAsset) {
+        this.exposedModules.add(this.bundle.entryAsset);
+      }
+
+      await this.write(prelude + '(function (require) {\n');
+    } else {
+      await this.write(`
+(function () {
+  function $parcel$exportWildcard(dest, source) {
```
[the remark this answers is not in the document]
fathyb: Fixed

src/transforms/concat.js
```diff
@@ -0,0 +1,174 @@
...
+        return t.identifier(computedSymbol);
+      }
+
+      if (exports.has(computedSymbol)) {
+        return t.identifier(exports.get(computedSymbol));
+      }
+
+      // if there is a wildcard for the module
+      // default exports are excluded from wildcard exports
+      if (wildcards.has(id) && name !== 'default') {
```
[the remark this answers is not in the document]
fathyb: You're right, removed it 👍

src/assets/RawAsset.js
```diff
@@ -18,9 +18,12 @@ class RawAsset extends Asset {
...
 
-    return {
-      js: `module.exports=${JSON.stringify(pathToAsset)};`
-    };
+    return [
+      {
+        type: 'js',
+        value: `module.exports=${JSON.stringify(pathToAsset)};`
+      }
+    ];
```
fathyb: Did this because I ended up with `module.exports = 'somepath'`. This requires each asset to do it, is there another way?

devongovett:
Awesome! How does the minified bundle size compare with the old version? How about build performance?

fathyb:
### Size

| Mode    | Vanilla | PR         |
| ------- | ------- | ---------- |
| `serve` | 1,3 MB  | 1,4 MB     |
| `build` | 536 KB  | **427 KB** |

### Time

| Mode    | Vanilla     | PR     |
| ------- | ----------- | ------ |
| `serve` | **7.53s**   | 13.65s |
| `build` | 8.78s       | 24.41s |

Somes notes : 
- `react-news` doesn't really take advantage of tree-shaking, its biggest libraries (Firebase, moment, react, react-router) aren't tree-shakables
- the tree-shaking seems to be broken with wildcards but I can't reproduce it in small scale

fathyb:
Uglify (even with hardcore options) still doesn't trim all exports objects, don't sure if it's the config, a bug or intended. I pushed a transform to trim exports object when they aren't used.

`react-news` goes from 427 KB to 422 KB. This small `lodash` script goes from 101 KB (same as without tree-shaking) to 88 KB : 

```js
import {chunk} from 'lodash-es'

console.log('output : %s', chunk('foobar', 3))
```

kzc:
> Uglify (even with hardcore options) still doesn't trim all exports objects, don't sure if it's the config, a bug or intended.

When in doubt compare the unminified bundle to what Rollup or Webpack 4 produces. If you post a gist of the uglify input I can tell you why a certain export is retained. I don't wish to build anything.

BTW, this actively maintained [fork](https://www.npmjs.com/package/@fabiosantoscode/uglify-es) of `uglify-es` contains a number of bug fixes:

https://github.com/fabiosantoscode/uglify-es

fathyb:
> If you post a gist of the uglify input I can tell you why a certain export is retained

```js
(function () {
  function $parcel$interopDefault(a) {
    return a && a.__esModule ? a.default : a;
  }

  function $parcel$exportWildcard(dest, source) {
    Object.keys(source).forEach(function (key) {
      if (key === "default" || key === "__esModule") {
        return;
      }

      Object.defineProperty(dest, key, {
        enumerable: true,
        get: function get() {
          return source[key];
        }
      });
    });

    return dest;
  }

  /* ASSET: 3 - b.js */
  var $3$exports = {};
  var $3$export$foo = ' ';
  $3$exports.foo = $3$export$foo;
  var $3$export$bar = 'trim me';

  $3$exports.bar = $3$export$bar;
  function $3$export$hello(world) {
    return [$3$export$hello, world].join($3$export$foo);
  }
  $3$exports.hello = $3$export$hello;

  /* ASSET: 1 - a.js */


  console.log('foo', $3$export$hello('world'));
})();
```

```js
!function(){var o={},n=" ";function r(o){return[r,o].join(n)}o.foo=n,o.bar="trim me",o.hello=r,console.log("foo",r("world"))}();
```

kzc:
@fathyb Unfortunately this line causes the `'trim me'` string to be retained by uglify:
```js
  $3$exports.bar = $3$export$bar;
```
Is it possible to generate code in the following manner?
```js
$ cat ex2.js 
(function () {
  function $parcel$interopDefault(a) {
    return a && a.__esModule ? a.default : a;
  }

  function $parcel$exportWildcard(dest, source) {
    Object.keys(source).forEach(function (key) {
      if (key === "default" || key === "__esModule") {
        return;
      }

      Object.defineProperty(dest, key, {
        enumerable: true,
        get: function get() {
          return source[key];
        }
      });
    });

    return dest;
  }

  /* ASSET: 3 - b.js */
  var $3$export$foo = ' ';
  var $3$export$bar = 'trim me';

  function $3$export$hello(world) {
    return [$3$export$hello, world].join($3$export$foo);
  }

  // export { foo, bar, hello };
  var $3$exports = {
      foo: $3$export$foo,
      bar: $3$export$bar,
      hello: $3$export$hello,
  };

  /* ASSET: 1 - a.js */


  console.log('foo', $3$export$hello('world'));
})();
```
```js
$ bin/uglifyjs ex2.js --toplevel -mc passes=9 -b
!function() {
    var o = " ";
    function n(r) {
        return [ n, r ].join(o);
    }
    console.log("foo", n("world"));
}();
```

fathyb:
No, I originally tried it but @devongovett noted that will break in cases like this one : 

```js
exports.bar = function() { ... }

export const foo = exports.bar('test')
```

kzc:
You're mixing commonjs `exports` assignments with ES6 `export` statements? If not, what does the original ES6 code look like?

src/transforms/concat.js
```diff
@@ -133,16 +132,28 @@ module.exports = packager => {
...
+                })
+              : node
           );
+        } else if (t.isIdentifier(node)) {
+          path.replaceWith(node);
         } else {
-          path.replaceWith(replaceExportNode(mod.id, name.value, path));
+          let relativePath = relative(packager.options.rootDir, mod.name);
+
+          throw new Error(`${relativePath} does not export '${name.value}'`);
```
fathyb: What about a warning instead of an error, like how we do for `fs`? I have to comment this to get code using `readFileSync` working

fathyb:
@kzc We need to support mixing CommonJS and ES6. I tried implementing this but after [this discussion](https://github.com/parcel-bundler/parcel/pull/1135#discussion_r179999336) and a bit of experimentation I switched back to member expressions.

test/integration/scope-hoisting/es6/re-export-default/b.js
```diff
@@ -1,2 +1,2 @@
-export foo from './c';
+export {default as foo} from './c';
```
devongovett: This was to test export extensions: https://github.com/tc39/proposal-export-default-from

fathyb: Oh sorry, didn't know! Reverted.

kzc:
>  if you reference commonjs exports before an es6 export it will insert the es6 export in the wrong place (before it is initialized, so it will be undefined). 

@fathyb I don't think rollup allows mixing CJS and ES6 exports. How does webpack accomplish it? Does this fringe case in webpack produce optimal DCE with uglify?

I could be mistaken, but it just seems to be a generated statement ordering issue to produce an uglify DCE friendly form.

tigp:
@fathyb may I ask you if this PR would remove the code splitting overhead #1428 ?

fathyb:
@tigp yes and no.

Yes because we now have to notion of "prelude-less", if you only have a `foo.js` file with `console.log("Hello world")` you get a final bundle with only `console.log("Hello world")`, no prelude.

No because the prelude is still needed with splitted dependencies. It's still an interesting idea, I'll try on this PR 🙂

tigp:
@fathyb that would be amazing! Thank you for your quick reply!

devongovett:
I think this is almost ready! Just doing some final code cleanup/moving around. 🚀 

How do you think we should release it? I'm thinking maybe under a flag at first, e.g. `--experimental-scope-hoisting` before we enable it by default - I'm worried it won't be perfectly stable at first. Curious to hear what people think.

devongovett:
In my tests, on large apps the performance is pretty good. Most of the minimization work now happens in workers, with only a top-level shake of exports and name mangling happening in the packager. The slowest part, by far, is the source map generation and I'm not sure how to make it faster... :/

src/scope-hoisting/hoist.js
```diff
@@ -432,28 +432,13 @@ module.exports = {
...
-      oldName = path.scope.generateDeclaredUidIdentifier(exportsName.name);
-
-      rename(path.scope, exportsName.name, oldName.name);
-    }
-
-    path.scope.push({
-      id: exportsName,
-      init: EXPORT_ALL_TEMPLATE({
-        OLD_NAME: oldName,
+    path.replaceWith(
```
fathyb: This will break CommonJS export all DCE: 

```js
var b = /*!#__PURE__*/test()

b = /*!#__PURE__*/test2(b) // doesn't get removed
```


```js
var _b = /*!#__PURE__*/test()
var b = /*!#__PURE__*/test2(_b) // b and _b are removed
```

fathyb: Ah, I should've looked more in details, great 👍

kennetpostigo:
You could always just use `scope-hoisting` flag, but not document it, or add a note that it's experimental in the documentation and a note in the CLI when it's being used.

joseluisq:
Awesome! 🥇

Siyfion:
Great work @fathy 🥇 🏅

fathyb:
Haha thanks! But @devongovett did most of the work, so congrats and thanks to him for spending so much time working on this! 🥇

src/scope-hoisting/mangler.js
```diff
@@ -0,0 +1,69 @@
+const rename = require('./renamer');
+const t = require('babel-types');
+
+const CHARSET = (
+  'abcdefghijklmnopqrstuvwxyz' + 'ABCDEFGHIJKLMNOPQRSTUVWXYZ$_'
+).split('');
+
+/**
+ * This is a very specialized mangler designer to mangle only names in the top-level scope.
```
vigneshshanmugam: Just curious, why does the mangler fail for topLevel scope? Babel minify supports toplevel mangling and I am interested to know what was the reason to have similar mangling step again here? or is it done for uglify since you are using it as the default minifier?

src/scope-hoisting/mangler.js
```diff
@@ -0,0 +1,69 @@
...
+ * Mangling of names in other scopes happens at a file level inside workers, but we can't
+ * mangle the top-level scope until scope hoisting is complete in the packager.
+ */
+function mangleScope(scope) {
```
vigneshshanmugam: I can see babel minify license being violated here, Parcel is using the [babel minify's mangler](https://github.com/babel/minify/blob/master/packages/babel-plugin-minify-mangle-names/src/charset.js)code which is licensed(which is MIT). Can you please mention link to the project and the license?

src/scope-hoisting/mangler.js
devongovett: The point of this mangler is to ONLY mangle the top level scope once all files are concatenation. All other scopes are mangled on a per file level.

Issue #1140, opened 2018-04-06
Built source map name is invalid


nick-woodward:
this a 🐛 bug report

### 🎛 Configuration (.babelrc, package.json, cli command)

```sh
$ parcel build index.html
```

```js
{
  // ...
  "browserslist": [
    "last 2 Chrome versions",
    "last 2 Firefox versions",
    "last 2 Safari versions",
    "last 2 Edge versions"
  ]
}
```

### 🤔 Expected Behavior

`parcel-bundler` should build files with the correct source map reference.

### 😯 Current Behavior

`parcel-bundler` is including the current files hash as the source map reference name, rather then the source maps hash.

**src.7f7f6500.js**

```js
// my source code here ...
//# sourceMappingURL=/src.7f7f6500.map
```

**src.eb1f688a.map**

```js
// source map here ...
```

### 🔦 Context

This issue is currently making it impossible to use source maps for error reporting tools (sentry).

### 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Parcel           | 1.7.0
| Node             | 9.5.0
| npm/Yarn         | 1.3.2
| Operating System | MacOS High Sierra 10.13.3

DeMoorJasper:
I've written a test that tests this, but can't seem to reproduce could you provide any test repo?

nick-woodward:
Currently having issues reproducing this outside of my project. 

I'll continue to try and reproduce it, however I'm guessing the issue 
is heavily tied to my use of `flat: true` within yarn.

nick-woodward:
@DeMoorJasper So I've managed to narrow down what causes it to happen.

It seems that if you import a `json` file into ` js` file it generates different hashes for the source and its respective map.

Here's a small example on https://glitch.com that you can check out.

https://glitch.com/edit/#!/parceljs-sourcemap-example

nick-woodward:
Hey @DeMoorJasper, any ETA on when #1148 will be published to NPM?

PR #1141, opened 2018-04-06
Sourcemap rootDir and improved tests

### LINKED ISSUE #1078: 🙋 sourceRoot support for sourcemaps
Currently there is no way to define the sourceRoot property via parcel

### 💁 Possible Solution

Adding a option for the API, which tells the sourceRoot

### 🔦 Context

The only way to to debug my electron application that has been bundled with parcel is by overriding the sourcemaps paths through the vscode debugger. 

My .vscode/launch.json
```
{
      "type": "node",
      "request": "launch",
      "name": "Electron: Main",
      "protocol": "inspector",
      "runtimeExecutable": "yarn",
      "runtimeArgs": ["start", "-- --remote-debugging-port=9223"],
      "cwd": "${workspaceRoot}",
      "sourceMaps": true,
      "outFiles": [ "${workspaceRoot}/dist/electron/**/*.js" ],
      "sourceMapPathOverrides": {
        "*": "${workspaceRoot}/src/electron/*",
      }
    }
```

Normally this should be handled via the `sourceRoot` property of a sourcemap.

### 💻 Code Sample

```
const options = {
  outDir: './dist', // The out directory to put the build files in, defaults to dist
  outFile: 'index.html', // The name of the outputFile
  publicUrl: './dist/',
  target: 'electron',
  sourceMaps: true,
  sourceRoot: './src/'
};
```

DeMoorJasper:
Closes #1100 Closes #1078 
Test related to #1140


src/packagers/SourceMapPackager.js
```diff
@@ -16,7 +16,7 @@ class SourceMapPackager extends Packager {
 
   async end() {
     let file = path.basename(this.bundle.name);
-    await this.write(this.sourceMap.stringify(file));
+    await this.write(this.sourceMap.stringify(file, this.options.rootDir));
```
devongovett: Should the sourceRoot be a relative path rather than absolute? seems like maybe `path.relative(options.outDir, options.rootDir)`?

DeMoorJasper: Updated to relative from outDir path

test/sourcemaps.js
```diff
@@ -26,8 +27,24 @@ describe('sourcemaps', function() {
...
     mapValidator(raw, map);
+    let mapObject = JSON.parse(map);
+    assert(
+      mapObject.sourceRoot ===
+        path.relative(b.options.outDir, b.options.rootDir),
+      'sourceRoot should be the root of the source files, relative to the output directory.'
+    );
+    assert(
+      fs.existsSync(
+        path.resolve(
```
felixfbecker: This is not a sufficient test, because it passes as long as `path.resolve()` can resolve it. But `sourceRoot` is a URL, so you need to test that resolving it as URL results in the right location. Which it currently doesn't, because `sourceRoot` doesn't contain a trailing slash, and trailing slashes are significant in URLs while they don't matter to `path.resolve()`.

```ts
new URL('index.js', 'file:///my/project')  // => 'file:///my/index.js' (WRONG)
new URL('index.js', 'file:///my/project/') // => 'file:///my/project/index.js' (CORRECT)
```

felixfbecker: See the source map spec:

> #### Resolving Sources
> If the sources are not absolute URLs after prepending of the “sourceRoot”, the sources are resolved relative to the SourceMap (like resolving script src in a html document).

https://sourcemaps.info/spec.html

test/sourcemaps.js
```diff
@@ -26,8 +27,24 @@ describe('sourcemaps', function() {
       .readFileSync(path.join(__dirname, '/dist/index.map'))
       .toString();
     mapValidator(raw, map);
+    let mapObject = JSON.parse(map);
+    assert(
+      mapObject.sourceRoot ===
+        path.relative(b.options.outDir, b.options.rootDir),
```
felixfbecker: This test is incorrect on Windows because `sourceRoot` should always be a URL with forward slashes, but on Windows this will assert that it's a path with backslashes

mischnic:
@felixfbecker Please open a new issue, otherwise this will be forgotten.

PR #1160, opened 2018-04-09
Add Markdown asset extension to core

### LINKED ISSUE #970: Import markdown file as a multi-line js string.

```js
// .babelrc
{
  "presets": [
    "env",
    "react"
  ],
  "plugins": [
    ["transform-class-properties", { "spec": true }]
  ]
}
```

### 🤔 Expected Behavior

```js
import mardownContent from '../contents/help.md';
console.log(mardownContent); // `# Title \n Paragraph`
```

### 😯 Current Behavior

```js
import mardownContent from '../contents/help.md';
console.log(mardownContent); // 12d7fdaf521be8a4e6ac363931673970.md
```

### 💁 Possible Solution

Allow this natively or make it a plugin?

### 🔦 Context

Isolating content from code seems like a fair idea.

### 🌍 Your Environment

| Parcel           | ^1.4.1  |
| Node             | v8.0.0 |
| npm/Yarn         | 5.0.0 |
| Operating System | Linux |

zzarcon:
As requested on #970 this adds Markdown support to parcel core instead of having to use a plugin for it.

Please let me know if Im missing something! 💃 

Closes #970


DeMoorJasper:
Imo this is a tricky one to implement, what if people want to import markdown and expect html?

Perhaps something like this would be better?
Markdown file:
```Javascript
export.html = ''; // Html version of the markdown
export.md = ''; // The original markdown string
```
Import usage:
```
const {md, html} = require('./somefile.md');
```

zzarcon:
Good point about the html version! I think we can easily have both named exports if we use a markdown converter like https://github.com/showdownjs/showdown. We might need to do the html conversion on demand to avoid unnecessary work.

About 
```
<a href="./somepage.md"></a>
```

I'd say we want to provide the html version, that's what I will expect, but I'm happy to see other approaches :)

I think we need to decide first on, before continue:

* Which parser we want to use
* Whats the default behavior for md links

thoughts?

DeMoorJasper:
@zzarcon Sounds good, we can always swap out the parser if it's not sufficient.
I think md links in html should transpile to html.

zeakd:
Here's some markdown parsers
 - [markdown-it](https://github.com/markdown-it/markdown-it) (used in stackedit.io)
 - [showdown](https://github.com/showdownjs/showdown)
 - [marked](https://github.com/markedjs/marked)  (used in vue example)
 - [remarkable](https://github.com/jonschlinkert/remarkable) (used in react example) but it maybe not maintained.

I think md should be transpile to html, too. But the problem is how parcel can pass `renderer` to MarkdownAsset. Without renderer options, MarkdownAssets would not be useful. So,, which is better way? 

- Pass renderer or other options to `<link />` attribute
- external Markdown options like .markdownrc. 
- else ?

devongovett:
Going to close this. I think if we support markdown it should be more useful than just inlining the original string. For that, you can always use `fs.readFileSync`.

Feel free to open a PR that compiles markdown to HTML. Then maybe we can use something like #1528 to load it.

PR #1166, opened 2018-04-10
Ignore noop dependencies

fathyb:
Ignore dependencies in a falsy branch, example `if(false)` or `if('')`.

This is needed when consuming Webpack generated UMD modules. For example in `react-data-grid`: 

```js
if (false) {
    var REACT_ELEMENT_TYPE = (typeof Symbol === 'function' &&
    Symbol.for &&
    Symbol.for('react.element')) ||
    0xeac7;

    var isValidElement = function(object) {
    return typeof object === 'object' &&
        object !== null &&
        object.$$typeof === REACT_ELEMENT_TYPE;
    };

    // By explicitly using `prop-types` you are opting into new development behavior.
    // http://fb.me/prop-types-in-prod
    var throwOnDirectAccess = true;
    module.exports = require('./factoryWithTypeCheckers')(isValidElement, throwOnDirectAccess);
}
```

Triggers this error : 

```
C:\~\node_modules\react-data-grid\dist\react-data-grid.js:93:28: Cannot resolve dependency './factoryWithTypeCheckers'
  91 | a  // http://fb.me/prop-types-in-prod\react-data-
  92 |    var throwOnDirectAccess = true;
> 93 |    module.exports = require('./factoryWithTypeCheckers')(isValidElement, throwOnDirectAccess);
     |                             ^
  94 |  } else {
  95 |    // By explicitly using `prop-types` you are opting into new production behavior.
  96 |    // http://fb.me/prop-types-in-prod
```

----

Reported by `Jindřich Pergler`

### DIFF
--- a/src/visitors/dependencies.js
+++ b/src/visitors/dependencies.js
@@ -38,7 +38,8 @@ module.exports = {
       callee.name === 'require' &&
       args.length === 1 &&
       types.isStringLiteral(args[0]) &&
-      !hasBinding(ancestors, 'require');
+      !hasBinding(ancestors, 'require') &&
+      !hasNoOpParent(ancestors);
 
     if (isRequire) {
       let optional = ancestors.some(a => types.isTryStatement(a)) || undefined;
@@ -88,6 +89,19 @@ module.exports = {
     }
   }
 };
+
+function hasNoOpParent(ancestors) {
+  return ancestors.some(node => {
+    if (
+      types.isIfStatement(node) &&
+      types.isLiteral(node.test) &&
+      !node.test.value
+    ) {
+      // only ignore if it's in the consequent branch
+      return ancestors.indexOf(node.consequent) !== -1;
+    }
+  });
+}
 
 function hasBinding(node, name) {
   if (Array.isArray(node)) {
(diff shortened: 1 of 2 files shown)


fathyb:
@devongovett sounds good, would this only apply to `process.env.NODE_ENV` or other globals? I feel like  `__dirname`, `__filename` and `process.env` is enough but your call 🙂 

@mischnic I'm not sure to understand

mischnic:
I was referring to @devongovett 's comment. `"development" === "production"` should evaluate to `false` just as `''` does.

fathyb:
@mischnic Oh, gotcha. The replacement is done before `src/visitors/globals` is executed (`collectDependencies` is called before `transform`) so it does not evaluate it currently : 
https://github.com/parcel-bundler/parcel/blob/3435c4c10c7234f292ac6809c5fbd680c6a915c8/src/assets/JSAsset.js#L90-L109

jpergler:
Hi @fathyb, will there be any progress with this issue? Thank you.

devongovett:
Updated to move environment variable inlining to a separate stage prior to dependency collection, and now evaluate the if statement test to determine whether it is true or false rather than only supporting literals.

PR #1167, opened 2018-04-10
Introduce .webmanifestrc

lwakefield:
I have no idea if this is the best approach for this problem, but it seemed to (kind of) fit into the same pattern used for `.posthtmlrc.js`, and `.postcssrc.js`.

`.webmanifestrc.js` is expected to export a single function which will have the webmanifest object passed to it. This allows modifying the output `webmanifest` at compile time.

The use case that I had in mind was compiling `webmanifest` as a handlebars (or other) template, so that I can set `name` and other variables from an `.env` file - this is kind of demonstrated in `test/html.js`

### DIFF
--- a/src/assets/WebManifestAsset.js
+++ b/src/assets/WebManifestAsset.js
@@ -1,4 +1,5 @@
 const Asset = require('../Asset');
+const webmanifestTransform = require('../transforms/webmanifest');
 
 class WebManifestAsset extends Asset {
   constructor(name, pkg, options) {
@@ -28,6 +29,10 @@ class WebManifestAsset extends Asset {
         this.ast.serviceworker.src
       );
     }
+  }
+
+  async pretransform() {
+    await webmanifestTransform(this);
   }
 
   generate() {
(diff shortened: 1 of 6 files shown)


DeMoorJasper:
What about having some defaults, in that case we could document it and add the config to overwrite it in case of certain edge cases?

lwakefield:
I don't have a good idea of what a default might look like - are you thinking something along the lines of:

```javascript
let config =
   require('./default-webmanifestrc') || // move import to top, just for demonstration purposes
   asset.package.webmanifest ||
   (await asset.getConfig([
     '.webmanifestrc',
     '.webmanifestrc.js',
     'webmanifest.config.js'
   ]));
```

I would imagine that an un-opinionated would probably just be an identity function - what do you think? Is there something more useful that isn't too opinionated?

devongovett:
Why not just generate your webmanifest file with another script prior to running parcel?

devongovett:
I'm going to close this since it seems to be possible to achieve the desired result in another way.

PR #1193, opened 2018-04-16
[Fix] Prevent build from breaking when .scss file is empty

crhayes:
I was attempting to set up my build with SASS according to the guides. In order to get the build working as quickly as possible I created an empty `app.scss` file and imported it into my main module. I expected the build to complete, but instead received this error:

image

Not sure how you feel about this solution, but I traced the problem down to `/src/Asset.js:70`. 

```js
if (this.contents && this.mightHaveDependencies()) {
  await this.parseIfNeeded();
  await this.collectDependencies();
}
```

`this.contents` is equal to an empty string, therefore this check evaluates falsy and the contents are not parsed. The end result is that `this.ast` is not set and remains `null`.

Because of this, `/src/SASSAsset.js:51` blows up:

```js
generate() {
  return [
    {
      type: 'css',
      value: this.ast.css.toString() // this.ast is null
      hasDependencies: false
    }
  ];
}
```

### DIFF
--- a/src/assets/SASSAsset.js
+++ b/src/assets/SASSAsset.js
@@ -48,7 +48,7 @@ class SASSAsset extends Asset {
     return [
       {
         type: 'css',
-        value: this.ast.css.toString(),
+        value: this.ast ? this.ast.css.toString() : '',
         hasDependencies: false
       }
     ];
--- a/test/integration/scss-empty/index.js
+++ b/test/integration/scss-empty/index.js
@@ -1,0 +1,5 @@
+require('./index.scss');
+
+module.exports = function () {
+  return 2;
+};
--- a/test/sass.js
+++ b/test/sass.js
@@ -82,6 +82,32 @@ describe('sass', function() {
     assert(css.includes('.base'));
   });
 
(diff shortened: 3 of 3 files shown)


crhayes:
@gnijuohz Thanks for the heads up, I fixed it!

gnijuohz:
@crhayes great! I was about to submit a PR to your repo haha. Thanks!

DeMoorJasper:
I think a better fix would be setting a check for empty files and not transpiling or returning anything in that case inside ‘Asset.js’ which is the parent class.
This might even speed up slightly as than we wouldn’t do anything with an empty asset

src/assets/LESSAsset.js
```diff
@@ -33,7 +33,7 @@ class LESSAsset extends Asset {
     return [
       {
         type: 'css',
-        value: this.ast.css,
+        value: this.ast ? this.ast.css : '',
```
DeMoorJasper: You could fallback to `this.contents` instead of an empty string, it would end up being the same though. Might be faster as you have it right now.
Would only be really useful if ast can ever be undefined/null while content is not.

PR #1196, opened 2018-04-16
Allow building standalone bundles for node.js

pravdomil:
**Proposal**
Create new option `bundleAll` to create standalone bundle for `node.js` and `electron` environments.

**Actual situation**
Currently `node_modules` files are not bundled. That's not full standalone bundle.


devongovett:
dup of #798

PR #1198, opened 2018-04-16
Add test to not dedupe imports with same contents but different dependency paths

henri-hulski:
This PR adds a test for the scenario described in #933 and fixed in #1011.
At the moment it's failing, but after merging #1011 it will pass.


henri-hulski:
Closing. Tests are now already included in #1011.

PR #1201, opened 2018-04-17
ouput local ip address

### LINKED ISSUE #1121: [feature request]output ip address
🙋 feature request

### 🔦 Context

It would be nice to have this feature. there are plenty times I want to visit my local serve through my phone, or let my coworker look at it while developing.  something like this would be enough.

image

lili21:
fix #1121


lili21:
image

src/Server.js
```diff
@@ -109,18 +110,26 @@ async function serve(bundler, port, useHTTPS = false) {
...
-      );
+      const protocol = useHTTPS ? 'https' : 'http';
+      const localAddr = `${logger.chalk.cyan(
+        `${protocol}://localhost:${_port}`
+      )}`;
+      const localIpAddr = `${logger.chalk.cyan(
+        `${protocol}://${ip.address()}:${_port}`
+      )}`;
+
+      logger.persistent('Server running at\n');
```
DeMoorJasper: Perhaps remove the newlines they seem a bit obsolete, definitely after looking at the screenshot.

lili21: 👌

src/Server.js
```diff
@@ -109,18 +110,21 @@ async function serve(bundler, port, useHTTPS = false) {
...
+      )}`;
+      const localIpAddr = `${logger.chalk.cyan(
+        `${protocol}://${ip.address()}:${_port}`
+      )}`;
+
+      logger.persistent('Server running at');
+      logger.persistent(`Local:             ${localAddr}`);
+      logger.persistent(`On Your Network:   ${localIpAddr}`);
+      if (_port !== port) {
+        logger.persistent(`${logger.chalk.yellow(`configured port ${port} could not be used.`)}`);
```
DeMoorJasper: Moved this into an if statement, otherwise it calls logger.persistent even without anything to print

lili21:
any updates ?

devongovett:
I think this is a useful feature in certain instances, but I hesitate to enable it by default as it might be confusing for beginners to have multiple urls in their terminal. I'd rather keep the noise to a minimum. It's not to hard to look up your computer's IP address by other means when you need to do that.

PR #1202, opened 2018-04-17
fix: just bundle for non-html entry

### LINKED ISSUE #1005: [RFC] Don't start dev server for Node bundling?

lili21:
fix #1005

### DIFF
--- a/src/cli.js
+++ b/src/cli.js
@@ -178,7 +178,7 @@ async function bundle(main, command) {
 
   const bundler = new Bundler(main, command);
 
-  if (command.name() === 'serve') {
+  if (command.name() === 'serve' && /\.html$/.test(main)) {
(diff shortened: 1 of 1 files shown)


DeMoorJasper:
What about htm or something that gets transpiled to html?

lili21:
I can do that with parcel ?

lili21:
any suggestion ?

DeMoorJasper:
Just wondering, this isn't related to the issue? The issue was about never starting a dev server if target is not browser as that makes no sense.
But this wouldn't start a server even if the target is browser which does make sense even for js entryAssets (although in practise perhaps a very niche use-case, this is why `parcel watch` exists)

devongovett:
I think there are valid use cases for running the server when the entry point isn't an HTML file. Maybe your HTML is served separately, or generated dynamically from a different server, for example.

PR #1204, opened 2018-04-17
Don't server if target isn't browser

### LINKED ISSUE #1005: [RFC] Don't start dev server for Node bundling?

DeMoorJasper:
Closes #1088 Closes #1202 Closes #1005


src/Bundler.js
```diff
@@ -643,7 +643,9 @@ class Bundler extends EventEmitter {
   }
 
   async serve(port = 1234, https = false) {
-    this.server = await Server.serve(this, port, https);
+    if (this.options.target === 'browser') {
```
devongovett: I think we should move this check into the CLI specifically. If someone is using the Parcel API, and they call `serve`, we should always start a server even if the target is node. In the CLI, we can detect the target and just call `bundle` instead of `serve` there.

src/cli.js
```diff
@@ -178,9 +178,14 @@ async function bundle(main, command) {
 
   const bundler = new Bundler(main, command);
 
-  if (command.name() === 'serve') {
+  console.log(command.target);
+
+  if (
+    (command.name() === 'serve' && command.target === undefined) ||
+    'browser'
```
devongovett: not sure I understand this logic

DeMoorJasper: that should've been `if (command.name() === 'serve' && (command.target === undefined ||  'browser')) {}` guess I made a typo

felixrabe:
Not sure this is the right way to go. I use `parcel serve --target electron` for HMR. With this change (IIUC - didn't run it, just looking at the diff) that would not be possible anymore.

devongovett:
HMR still starts in watch mode. serve mode is just for running a web server as well.

Issue #670, opened 2018-01-25
Support Web Workers coming from node_modules


wojtekmaj:
**Choose one:** is this a 🐛 bug report or 🙋 feature request?

Both, I guess?

### 🤔 Expected Behavior

* node_modules should be also checked when looking for a file

### 😯 Current Behavior

* trying to refer to a Service Worker file from node_modules fails with the following error:

```
×  C:\Users\wojtekmaj\Projekty\react-pdf\sample\parcel\node_modules\react-pdf\build\entry.parcel.js: Cannot resolve dependency './pdfjs-dist\build\pdf.worker.js' (...)
```

what it is trying to do is look for the path in `./` but not in `node_modules`.

### 💻 Code Sample

I'm currently working on support for Parcel in [React-PDF](https://github.com/wojtekmaj/react-pdf). That requires me to handle loading a service worker which is not mine.

In Webpack, the following code does the job:

```js
var pdfjs = require('react-pdf/build/pdf.js');
var PdfjsWorker = require('worker-loader!react-pdf/build/pdf.worker.js');

if (typeof window !== 'undefined' && 'Worker' in window) {
  pdfjs.PDFJS.workerPort = new PdfjsWorker();
}
```

I was trying to reproduce the same using Parcel, and here's what I got so far:

```js
const pdfjs = require('pdfjs-dist');

if (typeof window !== 'undefined' && 'Worker' in window) {
  pdfjs.PDFJS.workerPort = new Worker('pdfjs-dist/build/pdf.worker.js');
}
```

This results in an error described below. Changing the line to

```js
pdfjs.PDFJS.workerPort = new Worker('../node_modules/pdfjs-dist/build/pdf.worker.js');
```

stops the error, but:

* I'm not comfortable with relative path in a package like that
* This caused.. a 404 error! Parcel starts referring to a file it did not produced.

### 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Parcel           |1.5.0
| Node             |9.3.0
| npm/Yarn         |npm, newest
| Operating System |Windows 10

wojtekmaj:
Here's my full code at the moment if you'd like to have a look:

https://github.com/wojtekmaj/react-pdf/tree/parcel/sample/parcel

devongovett:
[...]

PR #1206, opened 2018-04-17
Fix generate names when outside of the entry directory

### LINKED ISSUE #1205: 🐛 Registering ServiceWorker fails with weird path

No configurations (see example repo)
Command: `parcel packages/a/examples/live/index.html`

### 🤔 Expected Behavior

Parcel should find calls to `navigator.serviceWorker.register` and resolve the given path correctly to a public http url accessible by the browser.

### 😯 Current Behavior

Parcel finds the `register()` call, but resolves the path wrong. The browser can not load the serviceworker file, therefore the registration fails.

The URL produced by parcel in the example repository is `http://localhost:1234/......%08sw.js`

Also note that the serviceworker file does not appear in the `dist` folder.

_Console output_
image

_Service Workers panel_
image

### 💁 Possible Solution

The path should point to the correct address of the file. Parcel should put the file into the dist folder to make it accessible to the browser.

### 🔦 Context

-> See example repository

### 💻 Code Sample

Repository to reproduce the bug: https://github.com/matthiasbayer/parcel-sw

Steps to reproduce:

- `yarn install`
- `parcel packages/a/examples/live/index.html`
- open http://localhost:1234
- check console and Application -> Service Workers panel

### 🌍 Your Environment

| Parcel           | 1.7.1
| Node             | 9.11.1
| npm/Yarn         | yarn
| Operating System | Win10 & Linux

fathyb:
Fixes #1205 and maybe #670.

When a service worker is outside of the `mainFile` directory it doesn't get bundled. This PR replaces `../` in the name with `__/`. I originally just trimmed `../` but it caused collisions, eg. `../../b` and `../b`.

I noticed while adding tests that the old service worker files in `integration` weren't actually used 🤔 
I don't know if it's the best fix (it's prolly not), feel free to edit or suggest alternative solutions!


matthiasbayer:
Looks like this should work on linux/mac, but it fails on my windows machine due to the `\` directory separators. 

A solution would be to also `replace(/\\/g, '/')`, but i have not enough knowledge about the parcel internals to tell if this causes any side effects.

fathyb:
Thanks, I pushed a commit to also parse `\`

matthiasbayer:
I tested the fix, but there is still a problem with these windows paths. If the path looks like this `..\..\..\a\b.js`, the result of the replace operation will be `__/__/__/b\sw.js`, the browser will read the path as `__/__/__/bsw.js`. So i think we have to replace all occurences of `\` here (if that's the right place to do so)

fathyb:
Thanks @devongovett, I forgot this one. Looks good to me

Kagami:
>When a service worker is outside of the mainFile directory it doesn't get bundled. This PR replaces ../ in the name with __/. I originally just trimmed ../ but it caused collisions, eg. ../../b and ../b.

Unfortunately that doesn't work properly because of worker scope:

>The path of the provided scope ('/') is not under the max scope allowed ('/__/'). Adjust the scope, move the Service Worker script, or use the Service-Worker-Allowed HTTP header to allow the scope.

Worker has to be in www root in order to e.g. intercept fetch events for root urls.

PR #1212, opened 2018-04-18
Allow spaces in filenames

### LINKED ISSUE #873: 🐛 Parcel fails to find dependencies if it contains spaces on the file name
### 🎛 Configuration

```
parcel index.html
```

### 🤔 Expected Behavior

No errors.

### 😯 Current Behavior

```
$ parcel index.html
🚨  /Users/user/dev/test/index.html: Cannot resolve dependency './test%20file.js' at '/Users/user/dev/test/test%    at /usr/local/lib/node_modules/parcel-bundler/node_modules/browser-resolve/node_modules/resolve/lib/async.js:55:21
    at load (/usr/local/lib/node_modules/parcel-bundler/node_modules/browser-resolve/node_modules/resolve/lib/a✨  Built in 8ms.
```

Browser:
```
🚨 Build error, check the console for details.
```

### 💁 Possible Solution

Accept spaces (` `) on the file name, checking the file system before URL-encoding it to %20.

### 🔦 Context

I've received a front-end to work on, however it has a long CSS file with some background-images pointing to images with spaces on the file name.

### 💻 Code Sample

index.html
```
<script src="test file.js"></script>
```

test file.js
```
document.write("test");
```

### 🌍 Your Environment

| Parcel           | 1.6.2 |
| Node             | 8.9.1 |
| npm/Yarn         | npm 5.6.0 |
| Operating System | macOS X 10.13.2 |

DeMoorJasper:
Closes #873


src/Resolver.js
```diff
@@ -27,7 +27,7 @@ class Resolver {
   }
 
   async resolve(input, parent) {
-    let filename = input;
+    let filename = decodeURIComponent(input);
```
devongovett: I don't think this is the right place to do this. We should only decode when we know that the input is a URL. For example, you could legitimately have a filename called "my%20file" where %20 is not a space.

I think we should do this in `addURLDependency` where we know the input is supposed to be a URL. Note that we already handled this for HTML assets, so we should move that call into `addURLDependency` itself so that it applies to e.g. CSS as well. https://github.com/parcel-bundler/parcel/blob/master/src/assets/HTMLAsset.js#L90

DeMoorJasper: Fixed it in the latest commit, now it supports url encoded files from any asset, and doesn't pass url encoded strings to asset.dependencies

src/Asset.js
```diff
@@ -87,10 +87,11 @@ class Asset {
       from = this.name;
     }
 
-    const parsed = URL.parse(url);
+    const parsed = URL.parse(decodeURIComponent(url));
     const resolved = path.resolve(path.dirname(from), parsed.pathname);
     this.addDependency(
-      './' + path.relative(path.dirname(this.name), resolved),
+      './' +
+        path.relative(path.dirname(this.name), decodeURIComponent(resolved)),
```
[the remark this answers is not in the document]
DeMoorJasper: `URL.parse` url encodes the parsed string.
I've pushed a fix as originally I thought the parsed url's encoding was important (therefore decoding twice), but it isn't as we create new filenames anyway.
Now it only decodes once saving some ms of building.

PR #1216, opened 2018-04-19
[WIP] Externals

### LINKED ISSUE #144: How do i mark some modules to external?

🙋 feature request

### 🤔 Expected Behavior

Don't include external module in bundled file everywhere. 
Like rollup `globals` option.
https://rollupjs.org/#big-list-of-options

### 🌍 Your Environment

| Parcel           |  1.0.3
| Node             | 9.2.0
| npm/Yarn         | Yarn 1.2.1

DeMoorJasper:
```Json
"externals": {
  "react": "externalReact"
}
```

```Javascript
const React = require('react');
```
Adds this as a dependency
```Javascript
module.exports = externalReact;
```

SIDENOTE: This changes the resolvers rootPackage logic, now it gets the rootPackage inside bundler and adds it to the options object. This could potentially add a few ms to the ipc transfer times

Closes #144


fathyb:
I'm not sure `const React = React` works though : 

```js
var React = {foo: 'bar'}

// Uncaught ReferenceError: React is not defined
void function() {
  const React = React

  console.log('React', React)
}()

// logs "React undefined"
void function() {
  var React = React

  console.log('React', React)
}()
```

Couldn't we just replace the `react` module with a simple `module.exports = React` or `module.exports = externalReact`? It'll be compatible with both ES6 and CommonJS and doesn't need a supplemental AST traversal.

DeMoorJasper:
@fathyb Ow didn't think about that, that'll make it way easier.
I'll update the PR to that strategy in a bit

DeMoorJasper:
@fathyb Just realised there isn't a way to add an asset based on content.

fathyb:
@DeMoorJasper ah, had same problem for the Angular integration and its virtual fs. Alternatively we could just internally handle it in `Asset.js`.

https://github.com/parcel-bundler/parcel/blob/bb7f4aa576ca65d26331412db05cf6cef08a7f5e/src/Asset.js#L60-L74

It could implement using something like :
```js
 async getDependencies() { 
   if ( 
     this.options.rendition && 
     this.options.rendition.hasDependencies === false 
   ) { 
     return; 
   }

   if(isExternal(this.name)) {
     this.contents = `module.exports = ${getExternal(this)}`

     // No need to parse dependencies or transform the AST
     return
   }
  
   await this.loadIfNeeded(); 
  
   if (this.contents && this.mightHaveDependencies()) { 
     await this.parseIfNeeded(); 
     await this.collectDependencies(); 
   } 
 } 
```

I don't know if changing the content on the fly doesn't have any side-effects (content hash, cache...). Thoughts?

DeMoorJasper:
@fathyb Seems a bit hacky, not sure how to resolve this in a proper way.
Perhaps we could introduce virtual fs to parcel?

fathyb:
@DeMoorJasper Totally for it! 👍 I could also bring performance improvement if it's abstracted enough to build a in-memory fs like `webpack-dev-server` for example.

mohsen1:
Can we do this inline system for now and when in memory fs is in place optimize it later?

```
contents = `module.exports = ${getExternal(this)}`
```

DeMoorJasper:
@mohsen1 Feel free to open up a PR that implements this.

PR #1233, opened 2018-04-23
Minor change to mkHandle in workerfarm

DeMoorJasper:
mkHandle used to be responsible for exposing functions for remoteWorkers, this PR updates this behaviour to work on local workers as well.
This way we pave a way for a more universal/extendible workerfarm as we can just call mkHandle to register a new function on both local and remote workers


src/workerfarm/WorkerFarm.js
```diff
@@ -25,24 +25,56 @@ class WorkerFarm extends EventEmitter {
...
     this.localWorker = require(this.options.workerPath);
-    this.remoteWorker = {
-      run: this.mkhandle('run')
-    };
+    this.run = this.mkhandle('run');
 
     this.init(options);
   }
 
+  callPromiseFactory(resolve, reject, method, args) {
```
devongovett: Why not just make this construct and return a Promise rather than duplicating twice below? Even better, merge this function with `addCall` and have it do that.

DeMoorJasper: Cleaned up in ae6fa72cec76c263cd7ad4726bbb63f765eebe2d

src/workerfarm/WorkerFarm.js
```diff
@@ -25,24 +25,46 @@ class WorkerFarm extends EventEmitter {
...
-          method,
-          args: args,
-          retries: 0,
-          resolve,
-          reject
+      // Child process workers are slow to start (~600ms).
+      // While we're waiting, just run on the main thread.
+      // This significantly speeds up startup time.
+      if (this.shouldUseRemoteWorkers()) {
+        return new Promise((resolve, reject) => {
```
devongovett: cool. any reason why the `new Promise` couldn't also go in `addCall`?

DeMoorJasper: Ow damn, totally overlooked that one. f24d3f9a25e3803e6a62f0e4787933ce0e6b70a3

PR #1235, opened 2018-04-23
Add ability to inject functions into the workers

DeMoorJasper:
Changes the way workers are used from registering functions/file to just requesting a function with it's path. Using mkHandle

Example:
```Javascript
let something = workerfarm.mkhandle(require.resolve('./worker.js'), 'doSomething');
let result = await something(...args);
```


devongovett:
Most plugins are asset types and already run in workers right?

src/workerfarm/WorkerFarm.js
```diff
@@ -23,10 +22,14 @@ class WorkerFarm extends EventEmitter {
...
-
-    this.localWorker = require(this.options.workerPath);
-    this.run = this.mkhandle('run');
-
+    this.workerFunctions = {};
+    
+    this.localWorker = require('./childBootstrap');
+    this.registerWorkerFunctions(this.options.workerPath || '../worker', [
+      'run',
+      'init'
```
devongovett: Maybe we could autodetect these instead of needing to list them? Basically `Object.keys(require(workerPath))`?

DeMoorJasper:
@devongovett Yes that's true, the typescript one utilises workers for speeding up error handling. It would be faster and probably better if the plugin wouldn't have to start and handle it's own workers.
Not sure how this exactly works I haven't dug into the code of parcel-plugin-ts

devongovett:
Also seems like this would potentially lead to different modules overwriting each other if they had the same function names. What if instead we just made the `processRequest` of the WorkerFarm work on both ends. So when you made a call to the worker farm, you'd specify both a module path and a function name rather than only having one module loaded into the children. This would work exactly the same way as when we call from children to parents. What do you think?

src/workerfarm/WorkerFarm.js
DeMoorJasper: I can change it, but think it'll be more precise if it has to be defined when being called.
For example if a worker has more than just the worker functions as exports we don't add a handler for all of them.
If we document this eventually it should be fine to use export keys I guess.

DeMoorJasper:
@devongovett If this doesn't impact performance or code readability it's probably a great idea. Would definitely solve the problem of 2 plugins possibly having a naming conflict

PR #1237, opened 2018-04-24
Update getPackageMain logic to retrieve the first entry that exists

rocketspacer:
Currently, node_modules package entry is resolved in the following order.
pkg.module -> pkg['jsnext:main'] -> pkg.browser -> pkg.main -> `${pkg.pkgdir}/index.js`

I keep the same order but will fallback to the first EXISTS entry. That way we can avoid common library author mistakes, which results in a lot of `cannot resolve dependency...` issues.

### DIFF
--- a/src/Resolver.js
+++ b/src/Resolver.js
@@ -185,7 +185,7 @@ class Resolver {
       pkg = await this.readPackage(dir);
 
       // First try loading package.main as a file, then try as a directory.
-      let main = this.getPackageMain(pkg);
+      let main = await this.getPackageMain(pkg);
       let res =
         (await this.loadAsFile(main, extensions, pkg)) ||
         (await this.loadDirectory(main, extensions, pkg));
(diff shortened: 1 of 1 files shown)


src/Resolver.js
```diff
@@ -217,19 +217,29 @@ class Resolver {
...
+      const entry = entries[i];
 
-    // Default to index file if no main field find
-    if (!main || main === '.' || main === './') {
-      main = 'index';
+      try {
+        const stat = await fs.stat(entry);
+        if (stat.isDirectory()) return path.resolve(entry, 'index');
+        return entry;
+      } catch (e) {
```
rocketspacer: I think we should just blow-up here if an entry record declared in package.json doesn't exists instead of handling it.

src/Resolver.js
```diff
@@ -217,19 +217,29 @@ class Resolver {
...
+
+    // main is the first entry that exists
+    for (let i = 0; i < entries.length; ++i) {
+      const entry = entries[i];
 
-    // Default to index file if no main field find
-    if (!main || main === '.' || main === './') {
-      main = 'index';
+      try {
+        const stat = await fs.stat(entry);
```
devongovett: Maybe instead of doing the actual resolution here (and then duplicating it again later), we could just have `getPackageMain` return an array of possible paths. Then, `loadDirectory` could just loop through them and try each one: https://github.com/parcel-bundler/parcel/blob/master/src/Resolver.js#L188-L191

devongovett:
Closing in favor of #1577.

PR #1238, opened 2018-04-24
Update default engines to > 0.25%

devongovett:
Don’t use “last 2 versions”. See https://jamie.build/last-2-versions. 

I think this should be an ok default query, see http://browserl.ist/?q=%3E0.25%25. I don't think we should drop IE 11 by default yet though, so this still means we are always running all plugins in babel-preset-env. A query like `> 0.5%, not ie 11, not android 4.4` would reduce that from 24 to 9 plugins... hmm. Thoughts?

cc. @jamiebuilds

### DIFF
--- a/src/utils/getTargetEngines.js
+++ b/src/utils/getTargetEngines.js
@@ -2,7 +2,7 @@ const browserslist = require('browserslist');
 const semver = require('semver');
 
 const DEFAULT_ENGINES = {
-  browsers: ['> 1%', 'last 2 versions', 'Firefox ESR'],
+  browsers: ['> 0.25%'],
   node: '6'
 };
 


jamiebuilds:
I sorta think that we should drop IE 11 by default, it's easy enough to turn back on

devongovett:
ok, WDYT about `> 1%, not ie 11`: http://browserl.ist/?q=%3E+1%25%2C+not+ie+11. That's 83.98% global coverage compared to 92.6% for the `> 0.25%` query, and results in just 2 plugins being applied (crazy): babel-plugin-transform-function-name, and babel-plugin-transform-destructuring vs all 24. Even async functions are supported in all of those browsers natively now, awesome. Other than IE11, that drops a bunch of old versions of Chrome, Safari, and Android browser, all with very small usage numbers.

jamiebuilds:
hmm i wonder why babel-plugin-transform-function-name is being run

jamiebuilds:
Oh, and I'm okay with that as a default. We should definitely document it well though

devongovett:
Ok, I did this [twitter poll](https://twitter.com/devongovett/status/989151366596608006) and I think there are enough people still targeting IE 11 that we shouldn't remove it by default yet. Even though I'd like to encourage it, I don't really want to deal with the issues that come in from people that don't know why parcel bundles aren't working in IE. Hopefully soon.

PR #1241, opened 2018-04-24
Fix autoinstall infinite loop

DeMoorJasper:
Autoinstall got triggered on too many errors, this PR handles the use-case of invalid/failed resolves for existing node_modules. Preventing it to trigger an install


src/Resolver.js
```diff
@@ -73,10 +74,8 @@ class Resolver {
     }
 
     if (!resolved) {
-      let err = new Error(
-        "Cannot find module '" + input + "' from '" + dir + "'"
-      );
-      err.code = 'MODULE_NOT_FOUND';
+      let err = new Error(`Cannot find module '${input}' from '${dir}'`);
+      err.code = isRelativeFile ? 'MODULE_NOT_FOUND' : 'NODE_MODULE_NOT_FOUND';
```
devongovett: unfortunately, this error code is standard as part of the node require resolution algorithm, so I don't think we can change it https://nodejs.org/api/modules.html#modules_file_modules

src/Bundler.js
```diff
@@ -366,40 +366,34 @@ class Bundler extends EventEmitter {
...
-        // If the dep is optional, return before we throw
-        if (dep.optional) {
-          return;
+      if (err.code === 'NODE_MODULE_NOT_FOUND' && install) {
+        let resolved;
+        if (dep.name.indexOf('/') > -1) {
+          // Check if module exists, prevents infinite install loop
+          try {
+            resolved = await this.resolver.resolve(
+              dep.name.substring(0, dep.name.indexOf('/')),
```
devongovett: this won't work for scoped modules, e.g. `@scope/module`. You should use `resolver.getModuleParts` to get the package name.

src/Bundler.js
```diff
@@ -366,40 +366,34 @@ class Bundler extends EventEmitter {
...
 
-        // If the dep is optional, return before we throw
-        if (dep.optional) {
-          return;
+      if (err.code === 'NODE_MODULE_NOT_FOUND' && install) {
+        let resolved;
+        if (dep.name.indexOf('/') > -1) {
+          // Check if module exists, prevents infinite install loop
+          try {
+            resolved = await this.resolver.resolve(
```
devongovett: You could move this call to `installDep` and return early if it fails.

One potential issue with trying to check if the package exists this way is with packages that don't have a `package.main` set. For example, imagine a package where there is no main entry point, but you instead need to request a sub-module every time e.g. `require('package')` doesn't work but `require('package/module')` does. It's not required to have a main entry point in a module. In this case, when someone did `require('package/bad-module')` where `package` is already installed but `package/bad-module` doesn't exist, we'd still get into a install loop.

Instead, we could just check if `node_modules/${this.resolver.getModuleParts(dep.name)[0]}` exists in any node_modules folder up to the root. In fact, we already do that check here: https://github.com/parcel-bundler/parcel/blob/master/src/Resolver.js#L149. In this case, we just don't want the code below which actually loads the package. Maybe we could extract the code that finds the node_modules folder containing the package from the rest of the code so we can reuse that function here. What do you think?

src/Resolver.js
DeMoorJasper: Yeah, probably not a good idea in that case

src/Bundler.js
DeMoorJasper: I've changed it into your suggestions, however it will still install when it's a failed alias. Not sure how to properly handle that

PR #1249, opened 2018-04-25
Htmlnano don't remove redundant attributes

### LINKED ISSUE #1245: "parcel build" removes type="text"
Bug report, maybe feature report after reading Jasper's comment. :smiley: 

Running the following command from shell:
      parcel build index.html
Inside index.html file, there is one line:
`<input type="text" name="demo" value="Hello">`
After parcel build, the above line becomes:
`<input name="demo" value="Hello">`     
type="text" is removed.

DeMoorJasper:
This will change the defaults of htmlnano to keep default values inside form elements, that could cause css selectors to fail.

Closes #1245


DeMoorJasper:
@devongovett Not that I know of, it doesn't mention it in the docs.
I've opened up an issue, with the feature proposal.

PR #1257, opened 2018-04-27
Fix worker bundle hoisting

DeMoorJasper:
This is a branched of PR from #979 

Main difference is that this PR adds options to the bundle that sets bundle.isolated, instead of looping through all dependencies every time you add a new one. (Should be slightly faster and more straightforward)
All credit goes to @dhcode for this, just wanted to bring some new life into this issue.

Closes #979

### DIFF
--- a/src/Bundler.js
+++ b/src/Bundler.js
@@ -585,7 +585,7 @@ class Bundler extends EventEmitter {
       asset.parentDeps.add(dep);
     }
 
-    if (asset.parentBundle) {
+    if (asset.parentBundle && !bundle.isolated) {
       // If the asset is already in a bundle, it is shared. Move it to the lowest common ancestor.
       if (asset.parentBundle !== bundle) {
         let commonBundle = bundle.findCommonAncestor(asset.parentBundle);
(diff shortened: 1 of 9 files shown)


src/Bundler.js
```diff
@@ -532,7 +538,7 @@ class Bundler extends EventEmitter {
       bundle = Bundle.createWithAsset(asset);
     } else if (dep && dep.dynamic) {
       // Create a new bundle for dynamic imports
-      bundle = bundle.createChildBundle(asset);
+      bundle = bundle.createChildBundle(asset, dep);
```
owap: Is this the magical line that will allow me to 

```jsx
import { debounce } from 'lodash';
```

...in my web workers?  We're very keen on being able to bundle external assets in our webworker scripts

DeMoorJasper: This PR should fix it

sebastiencs:
Thanks for your great work on parcel @DeMoorJasper 

I have tested the branch `fix/worker-bundle` and I still get the error:
`Uncaught Error: Cannot find module '99'`
when I try to import a module in a worker:
```js
import throttle from "lodash.throttle";
```

Am I missing something ?

DeMoorJasper:
@sebastiencs that's strange could you create a minimal example repo to reproduce that?

sebastiencs:
@DeMoorJasper  Turns out that if I change my worker filename from `worker.tsx` to `worker.js` it works, without changing anything else in the code. I will try to make a minimal repo to reproduce the error.

PR #1270, opened 2018-05-01
standalone bundles for node

joelnordell:
This is a more complete version of #1196. It adds an optional flag, `--bundle-all` (or `-A`) when targeting `node` or `electron` which will make parcel produce a standalone `.js` output which bundles all node modules. This can significantly reduce deployed code size for node applications by eliminating the need to deploy `node_modules`.

This does not attempt to solve the issue of node modules that dynamically load files from the filesystem, as mentioned in comments on #652. For applications using such modules, this feature likely will not work.

### DIFF
--- a/src/Bundler.js
+++ b/src/Bundler.js
@@ -104,6 +104,7 @@ class Bundler extends EventEmitter {
       minify:
         typeof options.minify === 'boolean' ? options.minify : isProduction,
       target: target,
+      bundleAll: options.bundleAll || false,
       hmr:
         target === 'node'
           ? false
--- a/src/Bundler.js
+++ b/src/Bundler.js
@@ -105,6 +105,7 @@ class Bundler extends EventEmitter {
         typeof options.minify === 'boolean' ? options.minify : isProduction,
       target: target,
       bundleAll: options.bundleAll || false,
+      ignore: options.ignore ? new Set(options.ignore.split(',')) : new Set(),
       hmr:
         target === 'node'
           ? false
@@ -399,6 +400,11 @@ class Bundler extends EventEmitter {
 
   async resolveDep(asset, dep, install = true) {
     try {
+      // Check for ignored modules
+      if (this.options.ignore.has(dep.name)) {
+        return;
+      }
+
       return await this.resolveAsset(dep.name, asset.name);
     } catch (err) {
       // If the dep is optional, return before we throw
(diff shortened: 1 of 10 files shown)


joelnordell:
This is not ready for merge yet, I plan to add a couple more fixes to this branch, rebase to the latest master, and add some tests.

joelnordell:
Still working on this -- closing this PR for now, will open a new one when this is actually ready for prime time.

mischnic:
So this would bundle non-Node modules (e.g. React, ....) and leave Node modules (async_hooks, fs, ...) untouched ? Great!

joelnordell:
@mischnic correct, node built-in modules will not be bundled but all other modules will be.

I don't love the workarounds I needed to do here for certain packages though. In general it ignores the "browser" attribute in package.json just like node does, but some packages don't bundle properly with parcel, and I found myself needing the prebuilt "browser" version. (e.g. `when`)

Also this still needs tests.

One of these days I'll hopefully have time to address those issues and reopen this pull request.

mischnic:
> I don't love the workarounds I needed to do here for certain packages though. In general it ignores the "browser" attribute in package.json just like node does, but some packages don't bundle properly with parcel, and I found myself needing the prebuilt "browser" version. (e.g. when)

How does webpack solve this when using `target: "node"`? (The `externals` config option is their `--ignore` flag.)

joelnordell:
Probably worth spending some time digging into webpack a bit to see how it handles packages like `when`.

As for the `--ignore` option -- I know the idea behind parcel is to be zero-configuration, but do you (or anyone else) have any thoughts on whether it might make sense to introduce this as some configuration in the `package.json` rather than relying on a (potentially long and ugly) command line option?

My specific use case is to ignore `./lib-cov` which is conditionally required in one my dependencies, for example the `lodash-inflection` package, but only if the `JSCOV` environment variable is set. (Without the ignore option, parcel attempts to bundle a package from that nonexistent path, and therefore fails.)

mischnic:
> I know the idea behind parcel is to be zero-configuration, but do you (or anyone else) have any thoughts on whether it might make sense to introduce this as some configuration in the `package.json` rather than relying on a (potentially long and ugly) command line option?

Yes, I have a feeling for quite some time now that the "no own configuration" choice might be hindering the development of more advanced features.

PR #1289, opened 2018-05-04
Custom less filemanager

### LINKED ISSUE #1281: LESS custom resolver syntax not supported
**Choose one:**  🐛 bug report

Using custom resolver syntax for LESS `@import '~...'` to import LESS files from node_modules/ doesn't work and gives an error.

### 🤔 Expected Behavior

Compile without issues. This works in webpack using `less` and `less-loader`.

### 😯 Current Behavior

```
~/project/src/styles.global.less: '~semantic-ui-less/semantic.less' wasn't found. Tried - ~/project/src/~semantic-ui-less/semantic.less,~semantic-ui-less/semantic.less
```

### 🔦 Context

Can't compile using less files from packages in node_modules/

### 💻 Code Sample
Install `semantic-ui-less`

```less
@import '~semantic-ui-less/semantic.less'
```
```js
import './styles.less'
```

### 🌍 Your Environment

| Parcel           | 1.8.0
| Node             | 8.11.1
| npm/Yarn         | npm 6.0.0
| Operating System | OSX Sierra

DeMoorJasper:
This PR adds a less filemanager, to support the custom resolver syntax that got added to parcel.

Closes #1281


src/visitors/lessFileManager.js
```diff
@@ -0,0 +1,44 @@
+const Resolver = require('../Resolver');
```
devongovett: I don't think this file should go in `src/visitors`. Maybe just put it inside LESSAsset for now. Soon we will be splitting these assets out into their own packages, so multiple files per asset type will make more sense.

src/visitors/lessFileManager.js
```diff
@@ -0,0 +1,44 @@
...
+const syncPromise = require('../utils/syncPromise');
+const path = require('path');
+const fs = require('../utils/fs');
+
+module.exports = function(less, options) {
+  const FileManager = less.FileManager;
+  const resolver = new Resolver({
+    extensions: ['.less'],
```
devongovett: less supports importing css too right?

DeMoorJasper: Yes it does, i'll change it

src/visitors/lessFileManager.js
DeMoorJasper: Alright, just wanted to keep it clean, and wasn't sure where else to put it.

PR #1304, opened 2018-05-06
Don’t pass package.json and options over IPC

devongovett:
Previously, we would get the package.json from the resolver and pass it over the IPC to workers for use in assets along with the full bundler options object. These were fairly large objects, so passing them across the wire every time was slow.

We now no longer pass those things over the IPC. The package.json object is resolved as needed within the asset using the `getPackage` method. Additionally a `packageKey` option was added to the `getConfig` method. The options object is already passed to all of the workers during initialization, so doesn't need to be passed for every asset.

### DIFF
--- a/src/Asset.js
+++ b/src/Asset.js
@@ -15,12 +15,11 @@ let ASSET_ID = 1;
  * for subclasses to implement.
  */
 class Asset {
-  constructor(name, pkg, options) {
+  constructor(name, options) {
     this.id = ASSET_ID++;
     this.name = name;
     this.basename = path.basename(this.name);
     this.relativeName = path.relative(options.rootDir, this.name);
-    this.package = pkg || {};
     this.options = options;
     this.encoding = 'utf8';
     this.type = path.extname(this.name).slice(1);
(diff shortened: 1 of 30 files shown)


DeMoorJasper:
Awesome!
But this will be a breaking change in the API right?

devongovett:
@DeMoorJasper technically yes, but `asset.package` was never documented (https://parceljs.org/asset_types.html) so I don't feel too bad about considering it a "private" field that we are removing. What do you think? Have you seen plugins relying on it?

DeMoorJasper:
@devongovett My imagemin plugin depends on it, but I guess it should be fine to see as a non-breaking change if you look at it as a private property.
Main reason I saw it as a non-private property is because it used to be a part of the constructor, which gets overwritten by a lot of plugins

devongovett:
hmm yeah that's true... I guess we could just pass an empty object there instead for now and remove it later? or we could try to shim it as a getter and throw a deprecation warning

DeMoorJasper:
Throwing a deprecation warning sounds like the best plan.

PR #1316, opened 2018-05-07
Pass .posthtmlrc options to posthtml-parse

### LINKED ISSUE #1103: 🐛Vue self-closing tags broken on build but fine on dev
When a `div` is self-closed in vue template, it will not correctly closed causing things below get embed in that self-closing tag.

| Code | Result |
|------|--------|
| image | image |

And strangely it's a production-build-only (`parcel build`) bug. Works fine on dev.

Included in repro.

### 💁 Possible Solution

No idea, even don't know it's a Vue issue or a Parcel issue.

### 💻 Code Sample

https://github.com/laosb/parcel-vue-test/commit/60baa3f7e1de3b963a5af1aeddba10d5e24a49a6

### 🌍 Your Environment

| Parcel           | 1.7.0 |
| Node             |  9.8.0 |
| npm/Yarn         | Yarn 1.5.1 |
| Operating System | macOS 10.13.3 |

### LINKED ISSUE #1123: Parcel lowercasing Vue attributes
🐛 bug report

I noticed that my logo component in Vue is huge when I run `parcel index.html`. I looked at both the Webpack output code and the Parcel output code and the only thing that was different was that Parcel caused my SVG's `viewBox` attribute to be all lowercase (`viewbox`) which breaks it (at least in the latest version of FireFox).

After testing it out some more, I noticed it doesn't change the capitalization when I remove the `use` elements and `xlink`.

```js
{
	"presets": [
		["env", {
			"modules": false,
			"targets": {
				"browsers": ["> 1%", "last 2 versions", "not ie <= 8"]
			}
		}],
		"stage-2"
	],
	"plugins": [
		"transform-runtime"
	],
	"env": {
		"test": {
			"presets": ["env", "stage-2"],
			"plugins": ["istanbul"]
		}
	}
}
```

luikore:
This pull request makes HTML parsing options configurable, so HTMLAsset parsing / transform behavior can be more consistent.

With this change we can config `.posthtmlrc` with the following options (they are consumed by posthtml-parse, then htmlparser2. See also https://github.com/fb55/htmlparser2/wiki/Parser-options )

- `xmlMode: true`
- `lowerCaseAttributeNames: false`
- `lowerCaseTags: false`
- `recognizeSelfClosing: true`
- ...

If we config `recognizeSelfClosing: true` in `.posthtmlrc`,  we can also fix #1103 

Allow setting `false` to lower case options maybe can also fix #1123 . (not tested yet)

### DIFF
--- a/src/assets/HTMLAsset.js
+++ b/src/assets/HTMLAsset.js
@@ -1,5 +1,4 @@
 const Asset = require('../Asset');
-const parse = require('posthtml-parser');
 const api = require('posthtml/lib/api');
 const urlJoin = require('../utils/urlJoin');
 const render = require('posthtml-render');
@@ -79,8 +78,8 @@ class HTMLAsset extends Asset {
     this.isAstDirty = false;
   }
 
-  parse(code) {
-    let res = parse(code, {lowerCaseAttributeNames: true});
+  async parse(code) {
+    let res = await posthtmlTransform.parse(code, this);
     res.walk = api.walk;
     res.match = api.match;
     return res;
@@ -151,7 +150,7 @@ class HTMLAsset extends Asset {
   }
 
   async pretransform() {
-    await posthtmlTransform(this);
+    await posthtmlTransform.transform(this);
   }
 
   async transform() {
--- a/src/transforms/posthtml.js
+++ b/src/transforms/posthtml.js
@@ -1,7 +1,17 @@
 const loadPlugins = require('../utils/loadPlugins');
 const posthtml = require('posthtml');
+const posthtmlParse = require('posthtml-parser');
 
-module.exports = async function(asset) {
+async function parse(code, asset) {
+  var config = await getConfig(asset);
+  if (!config) {
+    config = {};
+  }
+  config = Object.assign({lowerCaseAttributeNames: true}, config);
+  return posthtmlParse(code, config);
+}
+
+async function transform(asset) {
   let config = await getConfig(asset);
   if (!config) {
     return;
@@ -12,13 +22,16 @@ module.exports = async function(asset) {
 
   asset.ast = res.tree;
   asset.isAstDirty = true;
-};
+}
 
 async function getConfig(asset) {
-  let config = await asset.getConfig(
-    ['.posthtmlrc', '.posthtmlrc.js', 'posthtml.config.js'],
-    {packageKey: 'posthtml'}
-  );
+  let config =
+    (await asset.getPackage()).posthtml ||
+    (await asset.getConfig([
+      '.posthtmlrc',
+      '.posthtmlrc.js',
+      'posthtml.config.js'
+    ]));
   if (!config && !asset.options.minify) {
     return;
   }
@@ -38,3 +51,5 @@ async function getConfig(asset) {
   config.skipParse = true;
   return config;
 }
+
+module.exports = {parse, transform};


luikore:
@devongovett 
The vue support of parcel is still basically unusable due to so many bugs.
I hope some one can really pay some time to have a review about this?
Or just say you prefer other way for the fix so I can change or close this?

src/transforms/posthtml.js
```diff
@@ -1,7 +1,17 @@
 const loadPlugins = require('../utils/loadPlugins');
 const posthtml = require('posthtml');
+const posthtmlParse = require('posthtml-parser');
 
-module.exports = async function(asset) {
+async function parse(code, asset) {
+  var config = await getConfig(asset);
```
DeMoorJasper: Is the configuration format identical in parser and transform?

src/transforms/posthtml.js
```diff
@@ -11,20 +21,28 @@ module.exports = async function(asset) {
...
   }
 
   config = Object.assign({}, config);
   config.plugins = await loadPlugins(config.plugins, asset.name);
   config.skipParse = true;
-  return config;
+  return config
+}
+
+module.exports = {
```
DeMoorJasper: this can be written as `module.exports = {parse, transform};`

src/transforms/posthtml.js
```diff
@@ -11,20 +21,28 @@ module.exports = async function(asset) {
...
+  asset.isAstDirty = true
+}
 
 async function getConfig(asset) {
-  let config = await asset.getConfig(
-    ['.posthtmlrc', '.posthtmlrc.js', 'posthtml.config.js'],
-    {packageKey: 'posthtml'}
-  );
+  let config =
+    asset.getPackage().posthtml ||
```
fathyb: `asset.getPackage()` returns a `Promise` so your expression always evaluate to `undefined` :

https://github.com/parcel-bundler/parcel/blob/16149185bdd386cf478b08629fd2776dce76c943/src/Asset.js#L127-L133

Use `await` :

```js
(await asset.getPackage()).posthtml || ...
```

src/transforms/posthtml.js
```diff
@@ -11,20 +21,28 @@ module.exports = async function(asset) {
...
+    ]));
   if (!config && !asset.options.minify) {
     return;
   }
 
   config = Object.assign({}, config);
   config.plugins = await loadPlugins(config.plugins, asset.name);
   config.skipParse = true;
-  return config;
+  return config
```
fathyb: Looks like Prettier did not format (`--no-verify`?), you can manually run it by running `yarn format` or `npm run format`.

src/assets/HTMLAsset.js
```diff
@@ -73,14 +72,14 @@ const OPTIONS = {
 };
 
 class HTMLAsset extends Asset {
-  constructor(name, options) {
-    super(name, options);
+  constructor(name, pkg, options) {
+    super(name, pkg, options);
```
fathyb: We don't pass `pkg` to the `constructor` anymore for perf reasons : 

```js
constructor(name, options) {
  super(name, options)
```

src/transforms/posthtml.js
luikore: Updated.

src/transforms/posthtml.js
luikore: Did not know this before, updated

src/assets/HTMLAsset.js
luikore: Updated.

src/transforms/posthtml.js
luikore: Yes, but different options are used. Do you prefer these configs to be different objects?

DeMoorJasper: As long as they can’t conflict it should be fine but it sounds like they can

luikore: How about putting these options into a new htmlparser2rc ?

DeMoorJasper: I was thinking more of something like this:

DeMoorJasper: But as you said they won't conflict I think it's safe for now to just let this PR be as is

src/transforms/posthtml.js
```diff
@@ -12,13 +22,16 @@ module.exports = async function(asset) {
...
-    ['.posthtmlrc', '.posthtmlrc.js', 'posthtml.config.js'],
-    {packageKey: 'posthtml'}
-  );
+  let config =
+    (await asset.getPackage()).posthtml ||
+    (await asset.getConfig([
+      '.posthtmlrc',
+      '.posthtmlrc.js',
+      'posthtml.config.js'
+    ]));
```
devongovett: Did `packageKey` not work?

src/transforms/posthtml.js
devongovett: I think this will result in getting the config twice... once for parse, and once for transform. Maybe ok because of the cache? Just pointing that out.

PR #1321, opened 2018-05-08
add cache-dir option to cli

### LINKED ISSUE #709: [Feature] Make cache directory path configurable

DeMoorJasper:
Closes #709

### DIFF
--- a/src/cli.js
+++ b/src/cli.js
@@ -57,6 +57,7 @@ program
     'set the log level, either "0" (no output), "1" (errors), "2" (warnings + errors) or "3" (all).',
     /^([0-3])$/
   )
+  .option('--cache-dir <path>', 'set the cache directory. defaults to ".cache"')
   .action(bundle);
 
 program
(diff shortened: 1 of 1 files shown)


ekoeryanto:
why it still here?
Please merge it.

devongovett:
what if we just call it `.parcel-cache` always, then it won't overwrite things?

ekoeryanto:
@devongovett or ".cache/parcel"

mohsen1:
`cache/parcel` will work with existing `.gitignore` rules

devongovett:
I still think we should change the default so it conflicts less, but I merged anyway because I think it will be important e.g. for CI environments to be able to control the cache dir, for example if you wanted to put it outside a container in a shared location.

If someone wants to make a PR to change the default dir name, that would be great as well!

devongovett:
released in v1.9.3.

PR #1322, opened 2018-05-08
Split cache into subdirs for faster fs

DeMoorJasper:
Splits cache dir structure into subdirectories for slightly faster fs if there are a lot of cache files


jean343:
Would it be even faster to have all data in one single cache file to reduce disk IO?
For info, my project has about 3000 small files for a total of 150MB.

DeMoorJasper:
@jean343 that would probably flood memory and require virtualfs to work properly. Also wouldn’t be very redundant as the moment that file gets corrupt the entire cache is

PR #1324, opened 2018-05-08
Fix absolute and tilde paths

marcosbozzani:
Fix addURLDependency resolution in Asset.js

Minimal reproducible repo: https://github.com/marcosbozzani/parcel-bug-tilde-abs-paths

Just clone the repo, `yarn install` and `yarn tilde` or `yarn slash`

Tilde error: `Cannot resolve dependency './~\style.css'`
Slash error: `Cannot resolve dependency './..\..\..\..\..\..\`

### DIFF
--- a/src/Asset.js
+++ b/src/Asset.js
@@ -7,6 +7,7 @@ const isURL = require('./utils/is-url');
 const config = require('./utils/config');
 const syncPromise = require('./utils/syncPromise');
 const logger = require('./Logger');
+const Resolver = require('./Resolver');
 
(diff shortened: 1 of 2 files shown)


marcosbozzani:
Maybe I can create a logic that checks whether the assets already has a resolver or else use the default one.
I'll analyse the code and try to come up with a solution.

absentees:
@marcosbozzani does this work for you now? I'm still getting the same errors on your example repo, and on my own.

viztor:
still getting the same errors

PR #1326, opened 2018-05-08
Fixing bugs related to filewatchers and symlinks

### LINKED ISSUE #998: 🐛No hot module replacement with symlinks
🐛 bug report

I start up Parcel's development server with a command like this:
```
parcel ./public/symlink-index.html
```
I have a symlink at ./public/symlink-index.html which points to an index.html file (in any location). The hot module replacement does not occur when changes are made to the symlinked index.html file. Hot module replacement only occurs when the symlink itself changes.

My .babelrc
```js
{
  "plugins": [
    "transform-class-properties"
  ],
  "presets": ["env", "stage-0", "react"]
}
```

### 🤔 Expected Behavior

When a change is made to the symlinked index.html file, hmr should occur.

### 😯 Current Behavior

When a change is made to the symlinked index.html file, no hmr occurs. Dev server must be restarted for changes to take effect.

### 🔦 Context

We are using Bazel to build all of our projects. Bazel runs the parcel binary in a sandbox where it creates symlinks to all of the files in our app. 

### 🌍 Your Environment

| Parcel           | 1.6.2
| Node             | 9.7.1
| npm/Yarn         | 5.6.0
| Operating System | High Sierra

Tom-Julux:
📦🚀Parcel is awesome, but it did not supports *symlinks* very well, as `chokidar`, the filewatcher used by parcel, handles these transparently, which collides with the way assets are stored in parcel.

This pull request resolves #998 and hmr now works as expected.


Tom-Julux:
From what I can see the failing build have nothing to do with my changes.

DeMoorJasper:
Would be nice if you could add some tests

Tom-Julux:
Of course tests are always important and I already have created one, but the problem lies in the low-level nature of symlink. 

Most importantly they cannot be checked in with git and work reliable on another computer. Eg. My *ntfs windows 10* symlink will not work on any ext4 system.

The alternative to checking one in with git would be to create on dynamicly when the test is executed. This is only possible if the tests are run by an `administrator` (on Windows) / with `sudo` (on unix).

I am not sure it is a good idea to require these privileges, as it is one of nodes biggest strength that most packages can be installed without worrying about privileges.

So I am not sure if adding a test for this is a good idea.

DeMoorJasper:
It should be possible if you use relative symlinks like here: https://github.com/parcel-bundler/parcel/tree/master/test/integration/resolver/node_modules

But it's not that big of a deal.

Tom-Julux:
I now added a test. Using relativ symlinks was the solution.

src/Bundler.js
```diff
@@ -377,6 +379,9 @@ class Bundler extends EventEmitter {
     if (!this.watcher) {
       return;
     }
+    if (fs.existsSync(path) && fs.lstatSync(path).isSymbolicLink()) {
```
DeMoorJasper: I think we can turn watch into an async function and do this asynchronous?

Tom-Julux: This is probably a good idea. I will take a look at it.

Tom-Julux: I do not think a race condition could occur here, as the stats and exists function usually only take a few ms to execute, while the bundling takes one to two magnitudes longer.

Tom-Julux:
I can confirm that now all tests pass on my windows machine. I think while pushing my symlinks up I accidentally broke the existing ones in the repo

DeMoorJasper:
Closing in favor of a smaller version of this Pr #1408

PR #1327, opened 2018-05-08
Support `--no-autoinstall` for `build`

dobesv:
When running `build` you might want to disable autoinstall.


DeMoorJasper:
Isn’t autoinstall disabled by default for build?

dobesv:
Ah you are right.  It was auto-installing `node-sass` but I guess `--no-autoinstall` doesn't cover that.

DeMoorJasper:
Yeah build tools are autoinstalled because they cant cause side effects

PR #1336, opened 2018-05-09
[WIP] [RFC] Plugins v2 v2

fathyb:
## Principles

- Composition over inheritance, declarative API
- Polymorphic, multiple plugin can extend the same asset type. Reducers are used to handle async and multiple results
- Nothing should be imported to create a basic asset
- Plugins can't interfere between each other, each plugin has a `state` object allocated for each instance, acting as a proxy to `Asset`
- Every internal asset is implemented as a Plugin
- Every API used by an internal asset should be public

## Status

- [x] extend `Asset`
- [x] rewrite internal assets as plugins
- [ ] extend `Packager`
- [ ] expose `SourceMap` and `Resolver`
- [ ] load external plugins

#### YAML Asset
```js
const yaml = require('js-yaml');
const serializeObject = require('../../utils/serializeObject');

const YAMLAsset = {
  type: 'js',

  parse(code) {
    return yaml.safeLoad(code);
  },

  generate(ast, state) {
    return {
      js: serializeObject(ast, state.options.minify)
    };
  }
};

module.exports = {
  Asset: {
    yaml: YAMLAsset,
    yml: YAMLAsset
  }
};
```

#### GraphQL Asset
```js
const GraphqlAsset = {
  type: 'js',

  async parse(code, state) {
    let gql = await state.require('graphql-tag');
    return gql(code);
  },

  generate(ast) {
    return {
      js: `module.exports=${JSON.stringify(ast, false, 2)};`
    };
  }
};

module.exports = {
  Asset: {
    gql: GraphqlAsset,
    graphql: GraphqlAsset
  }
};
```

#### 🦄
```js
const traverse = require('babel-traverse')
const Unicorn = {
  transform(ast) {
    traverse(ast, {
      BlockStatement({scope}) {
        Object
          // get all bindings
          .keys(scope.bindings)
          // ignore already unicornified variables
          .filter(name => name[0] !== '🦄')
          // rename ${name} to 🦄_${name}_🦄
          .forEach(name => scope.rename(name, `🦄_${name}_🦄`))
      }
    })
  }
}
module.exports = {
  Asset: {
    js: Unicorn
  }
}
```

Related #919


fathyb:
@devongovett @DeMoorJasper you guys have any feedback on this one?

DeMoorJasper:
Looks better than before.
But I think Devon and Jamie are working on a new plugin system so not much use in reviewing this.
Definitely seems way more extendable this way though.

fathyb:
Ah, didn't know an effort was already started

PR #1373, opened 2018-05-16
WIP: Enable/disable node-sass sourcemaps (closes #903)

### LINKED ISSUE #903: Enable/disable node-sass sourcemaps
👋

Is there a way to add sourcemaps support for `node-sass`  ? 

Only way i've found is to create a `.sassrc` with this config (inline-sourcemaps here) for example:
```
{
  "sourceMap": true,
  "sourceMapEmbed": true
}
```
But no way to disable it when running `parcel build`command.

Thanks for your work.

oligot:
Enable/disable sourcemaps based on the CLI flag for sourcemaps that is already used for JS.

Note that sass test are failing for now, and I don't really understand why.
This seems to be related to the way sass/scss files are imported: when imported in a js file via `require('./file.scss')`, it fails; when "imported" in a html file or when it is the main entry (as is the case in the latest sass test), it works...

If someone with a deeper understanding of the Parcel code base than me could take a look and see what's wrong, this would help me a lot !


devongovett:

Issue #1363, opened 2018-05-13
Cannot have multiple Vue components in template


SteffenL:
# 🐛 bug report

When I use multiple Vue components in my Vue template, only the first child component of a parent component appears to be displayed after minification. This happens when building the code for production, and does not happen when building for development or when using the `--no-minify` parameter.

## 🎛 Configuration (.babelrc, package.json, cli command)

.babelrc: N/A

package.json:
```js
{
  "name": "testapp",
  "version": "1.0.0",
  "description": "testapp",
  "private": true,
  "scripts": {
    "test": "echo \"Error: no test specified\" && exit 1",
    "build": "parcel build src/index.html",
    "dev": "parcel src/index.html"
  },
  "author": "Steffen",
  "license": "ISC",
  "dependencies": {
    "vue": "^2.5.16"
  },
  "devDependencies": {
    "@vue/component-compiler-utils": "^1.2.1",
    "vue-template-compiler": "^2.5.16"
  }
}
```

cli command:
```
parcel build src/index.html
```

## 🤔 Expected Behavior

All of the components should be displayed.

In my example code, I expect `A B C D` to be displayed in the UI.

## 😯 Current Behavior

Only the first child component of a parent component appears to be displayed.

In my example code, `A B` is displayed in the UI. This means that component A did not display component B and C, and the App component did not display component D.

## 🔦 Context

I use multiple components to build my UI, so only some of the UI is displayed right now.

## 💻 Code Sample

Please see the example code here:

https://github.com/SteffenL/parcel-vue-bug-repro/tree/vue-components

Here is a smaller example template:
```html
<template>
<div id="app">
    <A /> <!-- displayed -->
    <B /> <!-- not displayed -->
</div>
</template>

<script>
import A from './A.vue';
import B from './B.vue';

export default {
    components: {
        A,
        B
    }
};
</script>
```

Building the code:
```
npm install
npm run build
```

Serve the files located in the `dist/` directory.

## 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Parcel           | 1.8.1
| Node             | 8.11.1
| npm/Yarn         | 5.6.0
| Operating System | Windows 10

SteffenL:
FYI, this `.uglifyrc` config is not helping:
```js
{
    compress: false,
    mangle: false
}
```

SteffenL:
In `VueAsset#compileTemplate`, I have found that the value of `html.value` is unexpected.

[...]

Issue #1294, opened 2018-05-04
Issue with vue.js and building for production


carbontwelve:
This is a 🐛 bug report.

When using parcel to build for production it breaks my vue.js application. This only happens when building for production, using parcel in development works with no issue.

### 🎛 Configuration (.babelrc, package.json, cli command)

`package.json`:
```js
{
  "dependencies": {
    "vue": "^2.5.16",
    "vue-hot-reload-api": "^2.3.0"
  },
  "devDependencies": {
    "@vue/component-compiler-utils": "^1.1.0",
    "parcel-bundler": "^1.8.1",
    "vue-template-compiler": "^2.5.16"
  }
}
```

`.babelrc`
```js
{
  "presets": [
    "env"
  ]
}
```

### 🤔 Expected Behavior

When building https://github.com/photogabble/vuejs-incremental-game-tutorial using the command `parcel build index.html --public-url=./ --out-file=index.html` and uploading to http://builds.photogabble.co.uk/parcel-js-vue-bug-report/ the app should work as expected (when it does work locally in dev mode.)

Below shows a screenshot of the expected behavior:

![image](https://user-images.githubusercontent.com/464699/39635785-60cf28a8-4fb6-11e8-8b27-dc89622676f2.png)

### 😯 Current Behavior

The app doesn't work as expected. Instead certain buttons are broken as shown in below screen shot (or if you visit that page):

![image](https://user-images.githubusercontent.com/464699/39635667-0f267d4e-4fb6-11e8-9df5-9678ee0a5159.png)

In vue the component is loaded via `<Resource name="Ore Reserves" units="tons" v-on:doBuy="buyOre" v-on:doSell="sellOre" v-bind:amount="ore" v-bind:buyPrice="oreBuy" v-bind:sellPrice="oreSell" v-bind:credits="credits" transactional></Resource>` and the buy/sell prices are passed through from `App.vue` as not being the zero being displayed.

Given that this only happens when parcel bundles for production I feel that this is a bug with parcel and not with my code.

### 💁 Possible Solution

The only solution I have found is to run `parcel index.html --public-url=./ --out-file=index.html` and then upload the development files from the `dist` folder. This can be seen working at http://builds.photogabble.co.uk/vuejs-incremental-game-tutorial-p1/. 

This tells me that there is something wonky going on with the build process for production.

### 🔦 Context

I was just trying to get a demo of my project online to share and it wasn't working, which is odd to me because it does work when built for development.

[...]

PR #1382, opened 2018-05-17
change Uglify to Terser

### LINKED ISSUE #1381: New dependency name: uglify-es => terser #1
* 🙋 feature request*

There has been development on a new fork of uglify-es, which is now called terser and has its own package!

    "terser": "^3.7.3"

Please upgrade to our new version.

DeMoorJasper:
Uglify-ES has been discontinued and Terser is a continuation of this package. This PR changes to Terser

Close #1381


devongovett:
hmm... I'm thinking of just switching to babel-minify. Most of the time we already process the asset with babel, so stringifying and re-parsing for uglify is probably not faster than just using babel-minify. And we can help contribute perf fixes to babel-minify where needed.

devongovett:
I've been experimenting with that for scope hoisting btw.

kzc:
> Most of the time we already process the asset with babel, so stringifying and re-parsing for uglify is probably not faster than just using babel-minify.

That's fine of course, but be aware that @fabiosantoscode just recently added ES6+ ESTree support in `@fabiosantoscode/uglify-es` (also present in`terser`). You might give that a spin before deciding.

kzc:
`babel-minify` is much faster than it used to be, but there's still a noticeable perf and size difference:

```
$ terser -V
terser 3.7.5

# both mangle and compress enabled
$ /usr/bin/time terser d3.js -mc | gzip | wc -c
        6.03 real         7.04 user         0.17 sys
   70954

# only mangle enabled
$ /usr/bin/time terser d3.js -m | gzip | wc -c
        1.44 real         1.62 user         0.06 sys
   72730
```

     '0.5.0-alpha.01eac1c3': '2018-05-17T23:39:29.993Z' 

```
$ babel-minify -V
0.5.0-alpha.01eac1c3

$ /usr/bin/time babel-minify d3.js | gzip | wc -c
       14.17 real        14.45 user         0.20 sys
   72238
```
I imagine a mangle-only run of `babel-minify` would also be comparably faster, but I don't know how to configure it.

kzc:
> I think we should focus our efforts around improving babel-minify

Absolutely - do whatever you think makes sense for your project. 

> avoid an additional generate + parse step in order to get from babel to uglify.

`terser` is just a maintained fork of `uglify-es`. An extra generate/parse step for ES6+ would not be required in `terser` as was the case in `uglify-es` because `terser` supports converting babel-compatible ES6+ ESTree AST to `terser` AST and back again, which is comparatively quick by comparison.

> That only made it into babel 7, so you may not be seeing it in your benchmarks if you're running 6 

`babel-minify` timing was made using latest babel 7 beta and latest alpha minify transforms:

```
├─┬ babel-minify@0.5.0-alpha.01eac1c3
│ ├─┬ @babel/core@7.0.0-beta.47
│ │ ├─┬ @babel/code-frame@7.0.0-beta.47
...
```
Looking beyond AST conversion, some algorithm(s) in `babel-minify` may still have O(N^2) complexity judging by slower timings with larger inputs:

```
$ /usr/bin/time terser app.js -mc | gzip | wc -c
       21.08 real        22.41 user         0.35 sys
  400385
```
```
$ /usr/bin/time babel-minify app.js | gzip | wc -c
       76.05 real        76.75 user         0.63 sys
  406098
```

But I'm sure with enough effort someone will be able to identify these remaining bottlenecks to benefit all `babel-minify` users.

devongovett:
btw, you can explicitly disable that behavior by calling `path.skip()` after replacing/inserting a node that doesn't need to be processed further. Maybe doing that in a few key places would help.

kzc:
@fathyb That's interesting. If that's the case, is there any way to effectively limit the number of passes in `babel-minify`? You get greatly diminishing returns after a couple of passes in a minifier. 

BTW, there are still a number of ESTree (a.k.a. "spidermonkey") AST conversion issues in `terser` for ES6+ code to be worked out. Probably just a few tweaks are required. No one is using this functionality at the moment. Should you change your mind and decide to use it, feel free to report any issues to the `terser` project.

devongovett:
I tried to use https://www.npmjs.com/package/babel-to-estree and then uglify's converter from estree early on but it was really buggy since babel's ast is so different, so we switched to converting to a string and re-parsing.

DeMoorJasper:
@fathyb resolved all merge conflicts. This PR should be ready to be merged, apparently the issue regarding name squatting has been resolved.

kzc:
I appreciate that the plan is to replace `uglify-es` with `babel-minify`, but I don't see any harm in this drop-in replacement pull request in the meantime.

`uglify-es` has been unmaintained for several months and has a number of significant bugs which have been corrected by `terser`. I'm surprised that Parcel users haven't encountered them.

fathyb:
We all have a direct interest in letting babel-minify and Terser compete for better size and speed. By letting users to choose which one to use we make comparing each other more easy and thus encourage competition.

@devongovett what do you think of merging this and implement babel-minifiy and Terser on the scope-hoisting PR? We could use babel-minify by default for speed and let user switch to Terser by using a `.uglifyrc` or `.terserrc`.

kzc:
Competition is always good. But I'd encourage this project to perform timings on real-life scenarios before choosing a default minifier. Very few bother to change the defaults in practise - particularly for a zero config bundler.

Morgul:
> uglify-es has been unmaintained for several months and has a number of significant bugs which have been corrected by terser. I'm surprised that Parcel users haven't encountered them.

@kzc Actually, I've run into a few bugs that seemed `uglify-es` related. The 'fix' for me was to turn off mangling in parcel; a terrible "work around". (#1294, and #1363 are two off the top of my head.) I would be really interested to know if this merge request would just solve those issues.

kzc:
>  I would be really interested to know if this merge request would just solve those issues.

Dozens of bugs have been fixed in `terser`. Either test against this PR directly, or use a `yarn` alias to replace `uglify-es` with `terser`. Google to find out how.

PR #1383, opened 2018-05-17
Update objectHash.js

odykyi:


odykyi:
@DeMoorJasper only refactoring, no issues

PR #1384, opened 2018-05-17
Update Asset.js

odykyi:


odykyi:
only refactoring, no issues

mischnic:
> @fathyb only refactoring, no issues

Could you please elaborate why you made that change?

PR #1390, opened 2018-05-18
Make CSS assets async

### LINKED ISSUE #1331: Bundler freezes since 1.8.0
# 🐛 Bug report

## 🤔 Expected Behavior

The bundler should finish the process.

## 😯 Current Behavior

Bundler freezes on build and does not finish at all. Previously I had version 1.7.0 and sometimes it was frozen as well but that was only several times and when restarted it mostly worked. After upgrade to version 1.8.1 I tried to build approx. 20 times but it did not finish at all. No changes in project. I tried version 1.8.0 and it did not work either.

## 🔦 Context

My project is just index.html with references scripts. Typescript is used in project. App based on React and Bootstrap.

## 🌍 Your Environment

| Parcel           | ^1.8.0 |
| Node             | 9.11.1 |
| npm         | 5.6.0 |
| Operating System | Windows 10 |

fathyb:
Remove `deasync` from CSS assets.
Fixes #1331 and bring some performance improvements.

#### Status

- [x] Sass
- [x] Less
- [x] Stylus


DeMoorJasper:
What about creating: Resolver.resolveSync inside the Resolver, that would remove all deasync calls instantly (except for get package in Asset.js, but that's just for backwards compatibility)?

fathyb:
@DeMoorJasper oh yeah, I was thinking the same. What do you think of making three classes :
- `Resolver`: the commons stuffs shared
- `AsyncResolver`: async implementation
- `SyncResolver`: sync implementation

I'm trying to find a good pattern where we could reuse maximum code and only have to implement `fs` stuff in the `Resolver` implementations.

What do you think of implementing a `Thenable` interface for each resolver :

```ts
type MaybeThenable<T> = T | Thenable<T>

interface Thenable<T> {
  then<U>(
    do: (v: T) => MaybeThenable<U>,
    catch: (err: Error) => MaybeThenable<U>
  ): Thenable<U>
}

interface SyncThenable<T> extends Thenable<T> {
  get(): T

  static from(value: T): Thenable<T>
}
```

And in each resolver: 

- `SyncResolver.js`
```js
class SyncResolver extends Resolver {
  stat(file) {
    return Thenable.from(fs.statSync(file))
  }

  readFile(file) {
    return Thenable.from(fs.readFileSync(file))
  }

  resolve(input, parent) {
    // Returns the sync result of the Thenable
    return super.resolve(input, parent).get()
  }
}
```

- `AsyncResolver.js`
```js
class AsyncResolver extends Resolver {
  stat(file) {
    return fsUtils.stat(file)
  }

  readFile(file) {
    return fsUtils.readFile(file)
  }
}
```

DeMoorJasper:
I think the resolver should contain them both as a function, so that in the bundler we don't have to create a second instance if we would ever need a synchronous version

devongovett:
hmm this was originally added for Stylus, which requires the resolution be synchronous. Even if we switched to `readFileSync` etc. it would still block the event loop. I wanted to avoid having two completely separate implementations of the resolver for async and sync, and since this has the same effect it seemed ok to me. If you can figure a way to make the Stylus resolution happen asynchronously without too much hassle that would be better, but until then deasync seemed like the easiest way.

DeMoorJasper:
Hope you don't mind me committing to this branch but I found a way to make the less filemanager async and pushed it to this branch @fathyb

fathyb:
Awesome @DeMoorJasper, I don't mind at all, thanks!

For Stylus I think I found a way to make it async using a double-traversal : 
- Parse the Stylus AST using `stylus.Parser`
- Use a `stylus.Visitor` to traverse nodes and collect dependencies
- pre-resolve and `await` dependencies promises
- Call the custom `stylus.Evaluator` using the dependencies we already have resolved

fathyb:
Just pushed an async `StylusAsset`, no need for a sync `Resolver` anymore 😄

dobesv:
I tried to run this branch but I get an error that I can't quite explain:

```
$ /home/ubuntu/vaultdropweb/node_modules/.bin/parcel watch app/roovy/static/html/index.html app/roovy/static/assets.assets -d app/roovy/static/gen --public-url /gen --no-cache --hmr-port 23693 --hmr-hostname 1.tcp.ngrok.io --no-autoinstall      
internal/modules/cjs/loader.js:550
    throw err;
    ^

Error: Cannot find module 'v8-compile-cache'
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:548:15)
    at Function.Module._load (internal/modules/cjs/loader.js:475:25)
    at Module.require (internal/modules/cjs/loader.js:598:17)
    at require (internal/modules/cjs/helpers.js:11:18)
    at Object.<anonymous> (/home/ubuntu/parcel/src/cli.js:1:63)
    at Module._compile (internal/modules/cjs/loader.js:654:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:665:10)
    at Module.load (internal/modules/cjs/loader.js:566:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:506:12)
    at Function.Module._load (internal/modules/cjs/loader.js:498:3)

```

To test the branch I used `yarn link` to link a clone of the repo to my project.

dobesv:
I tried it out and it ran to completion for me.

jpergler:
I can also confirm that the build finishes on previously problematic project.

src/assets/StylusAsset.js
```diff
@@ -50,18 +50,56 @@ class StylusAsset extends Asset {
...
+        deps.set(path, resolver.resolve(path, asset.name).then(m => m.path));
+      }
+    }
+  }
+
+  let options = {
+    cache: true,
+    filename: asset.name
+  };
+  let parser = new Parser(code, options);
```
devongovett: unfortunately I think this will result in the code being parsed twice for each file: once here, and once when actually processing the code later. Do you think there is a way they could share the same AST?

fathyb: I tested by putting some logs in Stylus' code and (ironically) `cache: true` broke the Parser cache (the cache is disabled if it isn't `'memory'`).

I fixed it and now it'll parse in `getDependencies` and use the cached AST in `generate`.

PR #1396, opened 2018-05-19
Use config.locals to render pug template

### LINKED ISSUE #1369: render index.pug file with locals variables.
# 🙋 feature request

render index.pug file with locals variables. 

## 🤔 Expected Behavior

I set "locals" filed in "pug.config.js" file. And then In the index.pug file , I can use these locals variables.

## 😯 Current Behavior

i can't use the locals in index.pug and there is no way i can resolve it.  

## 💁 Possible Solution

https://github.com/parcel-bundler/parcel/blob/e7df99decb2484d08b4f613ec6dc2ebe4bf9e969/src/assets/PugAsset.js#L35
update to
```
compiled(config.locals);
```

## 💻 Examples

```javascript
// pug.config.js
module.exports = {
  locals: {
    hello: 'world'
  }
}
// index.pug
doctype html
html(lang='zh')
  head
    body
    h1 hello #{hello} // access "hello" value in locals
```

buckle2000:
Closes #1369

### DIFF
--- a/src/assets/PugAsset.js
+++ b/src/assets/PugAsset.js
@@ -32,7 +32,7 @@ class PugAsset extends Asset {
       }
     }
 
-    return compiled();
+    return compiled(config.locals);
   }
 }
 


sebald:
Is there a way to already use/test this feature? :)

PR #1407, opened 2018-05-22
Resolve absolute URL relative to options.rootDir

### LINKED ISSUE #1352: Current `basedir` setting for pug seems redundant, and makes templating hard
Some might consider this to be a bug, but it doesn't violate documented behaviour so I'll file as a feature request instead.

# 🙋 feature request

The current `basedir` setting for rendering Pug ([here](https://github.com/parcel-bundler/parcel/blob/e7df99decb2484d08b4f613ec6dc2ebe4bf9e969/src/assets/PugAsset.js#L19)) is set to the directory containing the file being rendered.

If I'm interpreting the pug [docs](https://pugjs.org/language/includes.html) correctly, then this is somewhat redundant:

> If the path is absolute (e.g., include /root.pug), it is resolved by prepending options.basedir. Otherwise, paths are resolved relative to the current file being compiled.

This makes it hard to reference assets from templates, if the concrete `pug` files which extend those templates exist in different levels of the directory hierarchy, for example:

```
/src/templates/master.pug
/src/templates/product.pug [extends master]

/src/about/index.pug [extends master]
/src/about/... (assets specific for this page)

/src/contact/index.pug [extends master]
/src/contact/... (assets specific for page)
```

All fine at this point.

```
/src/product/potato/index.pug [extends product]

/src/product/tomato/index.pug [extends product]
```

Rendering these files fails if they extend master.pug (either directly or via a sub-template), as master.pug's assets are resolved relative to these files, and even if using absolute paths in master.pug, the paths are still resolved relative to the concrete file since it's path is supplied to pug for `basedir`

## 🤔 Expected Behavior

I'd expect that `basedir` would be configurable via some optional config file e.g. `.pugrc.js`, and that it would fall back to a sane default (e.g. project root path).  Also that either way, every template has the same `basedir` supplied to it, so that Pug's absolute paths can be useful.

## 😯 Current Behavior

Currently, `basedir` is set to the directory containing the `pug` file being rendered.
[At least if I'm interpreting this line correctly anyway.](https://github.com/parcel-bundler/parcel/blob/e7df99decb2484d08b4f613ec6dc2ebe4bf9e969/src/assets/PugAsset.js#L19)

While this doesn't cause anything that I'd consider a "bug", it does make absolute paths quite useless and enforces restrictions on directory structuring for pug pages and templates.

## 💁 Possible Solution

Somewhat dirty suggestion to convey the idea:

In [PugAsset.js](https://github.com/parcel-bundler/parcel/blob/e7df99decb2484d08b4f613ec6dc2ebe4bf9e969/src/assets/PugAsset.js#L19) around line 19, change

```
basedir: path.dirname(this.name),
```

to

```
basedir: path.resolve(config.basedir || ''),
```

While this is a breaking change to anyone currently using absolute paths in their `pug` files, I'd suspect that (nearly-)nobody is affected since (as far as I can tell) the current `basedir` causes absolute paths to be totally redundant and to behave the exact same way as relative paths.

## 🔦 Context

I'm trying to build pages where the template includes a corporate logo in the top navbar.

The absolute path that I'd need to specify in the template depends on the directory level of the page which ultimately invoked the template, which means all pages must be at the same level in the directory hierarchy.

## 💻 Examples

I'm happy to provide an example to trigger the problem if this isn't already clear enough, and I'll provide a PR if we agree on a solution.

### LINKED ISSUE #1398: Option to resolve absolute path relative to options.rootDir
# 🙋 feature request

Add an option to set "relative" absolute path.

## 🤔 Expected Behavior

If option `--root-dir` is given to `parcel`, absolute path should resolve relative to that path.

Example: `parcel src/index.html --root-dir src`

## 😯 Current Behavior

Any absolute path in assets is resolved to filesystem root `/`.

## 💁 Possible Solution

When resolving dependencies, change all `path.resolve` to the following function:

```js
const rootDir = '/home/user/some/path'

function resolvePath(...paths) {
    const rawPath = path.resolve(...paths)
    if (rootDir !== undefined) {
        if (path.isAbsolute(rawPath)) {
            return path.join(rootDir, rawPath)
        } else {
            return rawPath
        }
    } else {
        return rawPath
    }
}
```

For example, change
```const absPath = Path.resolve(Path.dirname(asset.name), dep.name);```
to 
```const absPath = resolvePath(Path.dirname(asset.name), dep.name);```

## 🔦 Context

Closes #1352 

## 💻 Examples

<!-- Love parcel? Please consider supporting our collective:
👉  https://opencollective.com/parcel/

`a.html` must reference `site.css`
`sub/b.html` must reference `../site.css`

However, the two HTML files are generated from the same template, so they must not reference different files.

buckle2000:
Will not close #1352 
Closes #1398

### DIFF
--- a/src/Asset.js
+++ b/src/Asset.js
@@ -91,10 +91,13 @@ class Asset {
     }
 
     const parsed = URL.parse(url);
-    const resolved = path.resolve(
-      path.dirname(from),
-      decodeURIComponent(parsed.pathname)
-    );
+
+    const subPath = decodeURIComponent(parsed.pathname);
+
+    const resolved =
+      subPath[0] === '/'
+        ? path.resolve(this.options.rootDir, subPath.slice(1))
+        : path.resolve(path.dirname(from), subPath);
     this.addDependency(
       './' + path.relative(path.dirname(this.name), resolved),
       Object.assign({dynamic: true}, opts)


DeMoorJasper:
Would probably be nice to run these dependencies through the resolver, this way it also resolves `~/`

DeMoorJasper:
How does this close #1398 ? This PR doesn't add any flag to the cli?

buckle2000:
https://github.com/parcel-bundler/parcel/pull/1407#issuecomment-390998887
What do you mean?

buckle2000:
https://github.com/parcel-bundler/parcel/pull/1407#issuecomment-391006582
I think this should be the default behaviour without any CLI flag. No one will use absolute path resolved to filesystem root.

DeMoorJasper:
I meant this: #1324 
You're definitely right that this should be the default behaviour though

DeMoorJasper:
merged #1324

PR #1408, opened 2018-05-22
Fix bugs related to watching symlinks

### LINKED ISSUE #998: 🐛No hot module replacement with symlinks

DeMoorJasper:
A cleaned up version of #1326

closes #1326
closes #998


src/Bundler.js
```diff
@@ -377,24 +378,33 @@ class Bundler extends EventEmitter {
...
     if (!this.watcher) {
       return;
     }
 
-    if (!this.watchedAssets.has(path)) {
-      this.watcher.watch(path);
-      this.watchedAssets.set(path, new Set());
-    }
+    if (await fs.exists(path)) {
+      if ((await fs.lstat(path)).isSymbolicLink()) {
```
fathyb: We should probably put `watch` and `unwatch` in a noop `try`/`catch` clause in case an error is thrown (eg. file locked by the anti-virus on Windows) to prevent unhandled promise rejections.

DeMoorJasper: Yeah was thinking the same but than thought people would open issues for some failed watches.

DeMoorJasper: What about just skipping (await fs.lstat(path)).isSymbolicLink() and just using await fs.realpath(path) instead? As it returns the original path if it's not a symbolic link? Seems to be faster as well

src/utils/fs.js
```diff
@@ -7,7 +7,16 @@ exports.writeFile = promisify(fs.writeFile);
 exports.stat = promisify(fs.stat);
 exports.readdir = promisify(fs.readdir);
 exports.unlink = promisify(fs.unlink);
-exports.realpath = promisify(fs.realpath);
+exports.realpath = async function(path) {
```
DeMoorJasper: Not sure if this is safe as it overwrites a native fs function.

src/Bundler.js
fathyb: 👍 that's even better!

curran:
CI is failing.

PR #1409, opened 2018-05-22
node-sass accepts importer as single function or array of functions

amrnt:
The current implementation sets an importer, so you cant add other importers via `.sassrc.js, ...`.

### DIFF
--- a/src/assets/SASSAsset.js
+++ b/src/assets/SASSAsset.js
@@ -38,7 +38,9 @@ class SASSAsset extends Asset {
       }
     });
 
-    opts.importer = (url, prev, done) => {
+    opts.importer = opts.importer || [];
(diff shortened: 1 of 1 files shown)


src/assets/SASSAsset.js
```diff
@@ -38,7 +38,9 @@ class SASSAsset extends Asset {
       }
     });
 
-    opts.importer = (url, prev, done) => {
+    opts.importer = opts.importer || [];
+    opts.importer = Array.isArray(opts.importer) ? [].concat(opts.importer) : [opts.importer];
```
DeMoorJasper: Why is `[].concat(opts.importer)` necessary, shouldn't it be safe to append to this array?

amrnt: oh my bad! This might be some remaining of my testing code... I'll fix it 👍

PR #1410, opened 2018-05-22
Fix wildcard deps in cache

DeMoorJasper:
If an asset had a dep that was a wildcard require, caching threw an error. 
Due to passing a non-existent path to fs.stat. 

This PR changes it to use the mtime of the directory of a wildcard dependency

Closes #1397

PR #1414, opened 2018-05-22
Glob asset dependency cache invalidating

DeMoorJasper:
This PR adds a test as a follow-up to #1410 
Also fixes the actual behaviour as it didn't invalidate


DeMoorJasper:
Now it should work perfectly.

Does parcel also support `some-location/**`? than the `getLastModified` function should also go into folders recursively.

devongovett:
Yeah and there are other types of globs too... https://www.npmjs.com/package/glob

PR #1417, opened 2018-05-22
Handle invalidating cache if dependency is a globAsset

### LINKED ISSUE #1546: Error ENOENT and hang on parcel 1.9.0 using Vuetify À-la-carte
# 🐛 bug report

We are using Vuetify [À la carte](https://vuetifyjs.com/en/guides/a-la-carte), upgraded to parcel 1.9.0 and the build fails with an ENOENT: no such file or directory error. Once the error appears, parcel hangs. This means `parcel watch` does not update anymore.

This may be a bug related to #1397.

package.json (updated):

```js
{
  "name": "parcel-vuetify-error",
  "version": "1.0.0",
  "description": "Bug in parcel-bundler (Error writing to cache)",
  "main": "index.html",
  "author": "Aldrin Martoq",
  "license": "MIT",
  "private": true,
  "dependencies": {
    "vue": "^2.5.16",
    "vue-hot-reload-api": "^2.3.0",
    "vuetify": "^1.0.19"
  },
  "devDependencies": {
    "@vue/component-compiler-utils": "^2.0.0",
    "parcel-bundler": "^1.9.0",
    "pug": "^2.0.3",
    "stylus": "^0.54.5",
    "stylus-loader": "^3.0.2",
    "vue-template-compiler": "^2.5.16"
  }
}
```

## 🤔 Expected Behavior

There should be no errors writing to the cache, here is Vuetify without a la carte:

```
$ rm -fr .cache; node_modules/.bin/parcel build --log-level 3 src/index.full.html | tee
⏳  Building...

⏳  Building index.full.html...
⏳  Building main.full.js...
⏳  Building vue.runtime.esm.js...
⏳  Building vuetify.js...
⏳  Building vuetify.min.css...
⏳  Building App.vue...
⏳  Producing bundles...
⏳  Packaging...
✨  Built in 8.71s.

dist/main.full.4243f518.js     294.18 KB    8.00s
dist/main.full.841574f8.css    244.46 KB    7.79s
dist/index.full.html               394 B    541ms
dist/main.full.e237299c.map          0 B     82ms
```

## 😯 Current Behavior

Here is Vuetify using a la carte, with the error:

```
$ rm -fr .cache; node_modules/.bin/parcel build --log-level 3 src/index.a-la-carte.html | tee
⏳  Building...

⏳  Building index.a-la-carte.html...
⏳  Building main.a-la-carte.js...
⏳  Building vue.runtime.esm.js...
⏳  Building app.styl...
⏳  Building App.vue...
⏳  Building index.js...
⏳  Building index.js...
⏳  Building index.js...
⏳  Building index.js...
⏳  Building index.js...
⏳  Building index.js...
🚨  /Users/amartoq/proj/vuejs/sandbox/parcel-vuetify-error-writing-to-cache/node_modules/vuetify/src/stylus/app.styl: ENOENT: no such file or directory, open '/Users/amartoq/proj/vuejs/sandbox/parcel-vuetify-error-writing-to-cache/node_modules/vuetify/src/stylus/settings/*
ENOENT: no such file or directory, open '/Users/amartoq/proj/vuejs/sandbox/parcel-vuetify-error-writing-to-cache/node_modules/vuetify/src/stylus/settings/*'
```

And **parcel hangs** here.

## 💁 Possible Solution

I'm sorry, I don't know how to help with this issue.

## 🔦 Context

We have to go back to parcel 1.8.1, because `parcel watch` hangs. This may be a bug related to #1397.

## 💻 Code Sample

I just upgraded the packages from the same sample code:

[https://github.com/aldrinmartoq/parcel-vuetify-error-writing-to-cache](https://github.com/aldrinmartoq/parcel-vuetify-error-writing-to-cache)

## 🌍 Your Environment

| Parcel           | 1.9.0
| Node             | 8.9.4
| Yarn         | 1.7.0
| Operating System | Mac OS X 10.13.5

Thanks.

DeMoorJasper:
Duplicate of #1414 

This is needed to properly invalidate and cache styl files as they allow glob requires.

But I kind of messed up a rebase, so here is a clean version

Closes #1546

### DIFF
--- a/src/FSCache.js
+++ b/src/FSCache.js
@@ -69,8 +69,7 @@ class FSCache {
     // If any of them changed, invalidate.
     for (let dep of data.dependencies) {
       if (dep.includedInParent) {
-        let mtime = await this.getLastModified(dep.name);
-        if (mtime > dep.mtime) {
+        if ((await this.getLastModified(dep.name)) > dep.mtime) {
           return false;
         }
       }
--- a/src/FSCache.js
+++ b/src/FSCache.js
@@ -32,20 +32,16 @@ class FSCache {
   }
 
   async getLastModified(filename) {
-    let mtime = 0;
     if (isGlob(filename)) {
       let files = await glob(filename, {
         strict: true,
         nodir: true
       });
-      mtime = (await Promise.all(
+      return (await Promise.all(
         files.map(file => fs.stat(file).then(({mtime}) => mtime.getTime()))
       )).reduce((a, b) => Math.max(a, b), 0);
-    } else {
-      let stats = await fs.stat(filename);
-      mtime = stats.mtime.getTime();
     }
-    return mtime;
+    return (await fs.stat(filename)).mtime.getTime();
   }
 
   async writeDepMtimes(data) {
--- a/src/FSCache.js
+++ b/src/FSCache.js
@@ -4,6 +4,7 @@ const md5 = require('./utils/md5');
 const objectHash = require('./utils/objectHash');
 const pkg = require('../package.json');
 const logger = require('./Logger');
+const {isGlob, glob} = require('./utils/glob');
 
 // These keys can affect the output, so if they differ, the cache should not match
 const OPTION_KEYS = ['publicURL', 'minify', 'hmr', 'target', 'scopeHoist'];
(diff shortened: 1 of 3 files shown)


src/FSCache.js
```diff
@@ -30,16 +31,31 @@ class FSCache {
...
+        strict: true,
+        nodir: true
+      });
+      for (let file of files) {
+        let fileStats = await fs.stat(file);
+        if (fileStats.mtime > mtime) {
+          mtime = fileStats.mtime.getTime();
+        }
+      }
+    } else {
```
fathyb: nit-pick: use `Promise.all` for better perf:
```js
if (isGlob(filename)) {
    let files = await glob(filename, {
        strict: true,
        nodir: true
    });

    mtime = (
        await Promise.all(
            files.map(file =>
                fs.stat(file).then(stat => stat.mtime.getTime())
            )
        )
    ).reduce((a, b) => Math.max(a, b), 0)
}
```

DeMoorJasper: I'll change it

devongovett:
Maybe it would simplify things if we just marked all GlobAssets as non-cacheable? That way the cache wouldn't need to have specific logic about them. Alternatively we could use the `shouldInvalidate` hook in GlobAsset to check that...

Right now we always need to run `glob` for GlobAssets every time, and that's the expensive part. So if that's the case, we might as well just not cache them at all. WDYT?

DeMoorJasper:
@devongovett Won't this invalidate the parent asset as well as the main edit in this PR is dependencies.
Not sure if this would still be faster if the parent fairly large js file in production mode.

devongovett:
When you `require('./foo/*.js')` for example, a virtual file (the GlobAsset) is created which looks something like this:

```javascript
module.exports = {
  test: require('./foo/test.js'),
  bar: require('./foo/bar.js')
};
```

That's the "file" that you are needing to invalidate when files are added/removed. So if we just don't cache that at all, it should just work for free. Not sure we would gain anything by caching it since we need to check it every time anyway.

PR #1423, opened 2018-05-23
Add amp-* tag to HTMLAsset to parse its own src attribute

takashi:
Currently, `amp-*` tag's src does not be parsed by `HTMLAsset`.

### DIFF
--- a/src/assets/HTMLAsset.js
+++ b/src/assets/HTMLAsset.js
@@ -18,7 +18,9 @@ const ATTRS = {
     'source',
     'track',
     'iframe',
-    'embed'
+    'embed',
+    'amp-img',
+    'amp-anim'
   ],
   href: ['link', 'a', 'use'],
   srcset: ['img', 'source'],


devongovett:
hmm not sure about this one. they aren't standard attributes so I hesitate to handle them in parcel

takashi:
@fathyb Hi thanks for reviewing, I'll add a test tomorrow.

@devongovett I know it's not standard attributes.  And I can understand your opinion. If there is  a some way to extend `ATTR` or some new plugin is a better way to do this.

devongovett:
I don't have a good solution to this. I guess you could write a plugin to override the HTMLAsset with something else.

A better solution might be for Parcel to support a config file where users could specify non-standard attributes they want to handle, e.g. `.htmlrc`. Thoughts on that idea?

Going to close this for now and we can follow up about the ideas above separately.

takashi:
@devongovett thanks for considering this.

PR #1429, opened 2018-05-24
Always invalidate globAssets

DeMoorJasper:
This PR is an alternative to #1417.

* Invalidate globAssets every time, as it takes just as long to check mtimes as to generate a new globAsset (but caching happens in master process, so actually even slows down the process)...
* Never invalidate a parent of globAsset, due to the glob Dependency (This should speed up builds, as before it always invalidated due to an error and globAsset deps don't really have an influence on the content of the parent asset)
* Makes glob generating async, might result in a minor performance improvement

Related to the comment by @devongovett on #1417

### DIFF
--- a/src/FSCache.js
+++ b/src/FSCache.js
@@ -4,7 +4,7 @@ const md5 = require('./utils/md5');
 const objectHash = require('./utils/objectHash');
 const pkg = require('../package.json');
 const logger = require('./Logger');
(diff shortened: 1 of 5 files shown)


DeMoorJasper:
After investigating the original case for this, it seems to be related to Styl and this would potentially break invalidating styl files correctly

PR #1435, opened 2018-05-25
Adding an bundlestart event

Tom-Julux:
Plugins should be able to run code before a build process starts.

### DIFF
--- a/src/Bundler.js
+++ b/src/Bundler.js
@@ -207,6 +207,7 @@ class Bundler extends EventEmitter {
         });
       });
     }
+    
     this.emit('buildStart', this.entryFiles)
     
     let isInitialBundle = !this.entryAssets;
--- a/src/Bundler.js
+++ b/src/Bundler.js
@@ -207,7 +207,7 @@ class Bundler extends EventEmitter {
         });
       });
     }
-    this.emit('bundlestart', this.entryFiles)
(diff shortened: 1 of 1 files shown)


src/Bundler.js
```diff
@@ -207,7 +207,8 @@ class Bundler extends EventEmitter {
         });
       });
     }
-
+    this.emit('bundlestart', this.entryFiles)
```
fathyb: We currently have a `buildEnd` event when `Bundler` is done, it would be more consistent to name this one `buildStart`

Tom-Julux: Definitely makes sense

DeMoorJasper:
Just wondering, this won't block any action, so what's the actual use-case for this?

Tom-Julux:
There are no use cases when running parcel via cli. But for embedding parcel in a larger scope it can be helpful to know when a new build started. For example I use a build server and used this event to set up a webhook.
As this feature is non-blocking, I saw no reason to not implement it. However it is not that important...

DeMoorJasper:
@Tom-Julux Alright :)

I'll merge this in as it makes sense as we already print a message that build starts to cli, makes it available in the api as well :)

dsych:
I am just curious, when can this make it into the official release?

DeMoorJasper:
It’ll be part of 1.9

PR #1436, opened 2018-05-25
[WIP] Improve logging user experience

fathyb:
Just an idea, I opened this PR prematurely to get some feedback.

- [x] Use relative name instead of base name (`Building index.js` isn't really expressive)
  - [ ] Use a npm based name instead (`@tensorflow/tfjs-core/dist-es6/kernels/backend_webgl.js` instead of `../node_modules/@tensorflow/tfjs-core/dist-es6/kernels/backend_webgl.js`)
- [ ] Improve emojis (⚡️ for transform, 💥 for generate)
- [ ] Batch logger calls (no need to render 400 lines/seconds)

**Before**

The logger shows `Building quad.ts` for two seconds, given its content I highly doubt it's representative : 

- `quad.ts`
```ts
import {StaticBuffer} from '../objects/buffer'

const BufferData = new Float32Array([
    // (x, y)

    // Vertices
    -1,  1,
    -1, -1,
     1,  1,
     1, -1,
     1,  1,
     1, -1,
     1,  1,
     1, -1,

     // Texture coordinates
     0,  0,
     0,  1,
     1,  0,
     1,  1,
     0,  0,
     0,  1,
     1,  0,
     1,  1,
])

export class QuadGeometry extends StaticBuffer {
    public readonly position = 0
    public readonly texture = 16 * Float32Array.BYTES_PER_ELEMENT

    constructor(gl: WebGLRenderingContext) {
        super(gl, BufferData)
    }
}
```

**After**

### DIFF
--- a/src/Asset.js
+++ b/src/Asset.js
@@ -6,6 +6,7 @@ const md5 = require('./utils/md5');
 const isURL = require('./utils/is-url');
 const config = require('./utils/config');
 const syncPromise = require('./utils/syncPromise');
+const emoji = require('./utils/emoji');
 const logger = require('./Logger');
 const Resolver = require('./Resolver');
 
@@ -51,6 +52,7 @@ class Asset {
 
   async loadIfNeeded() {
     if (this.contents == null) {
+      logger.status(emoji.progress, `Loading ${this.relativeName}...`);
       this.contents = await this.load();
     }
   }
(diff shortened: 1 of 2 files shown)


DeMoorJasper:
The relative path logging is definitely a must.
What do you mean by batching? Is it the status that sets 400 lines/second, in that case I guess it'll be safe to simply give status a cooldown, this way you don't need to batch just return early in case the status gets set, before the cooldown ended.

fathyb:
@DeMoorJasper Ah batch wasn't probably the right word, I meant cooldown 🙂

DeMoorJasper:
Really like the other idea to show progress per process and per worker.
This would give you a very nice overview and show progress a lot better as it now looks the same the entire build progress, which is kind of confusing for new users.
Would also help with debugging to see if a worker freezes or where it exactly failed

PR #1439, opened 2018-05-25
Use map in parser

mohsen1:
Does this make sense so I spend time clean it up and add test that actually test something?

### DIFF
--- a/src/Parser.js
+++ b/src/Parser.js
@@ -5,7 +5,11 @@ const glob = require('glob');
 
 class Parser {
   constructor(options = {}) {
-    this.extensions = {};
+    /**
+     * A map of extension names to either path to Asset class or Asset class itself
+     * @type {Map<string, string|Asset}
+     */
+    this.extensions = new Map();
 
     this.registerExtension('js', './assets/JSAsset');
     this.registerExtension('jsx', './assets/JSAsset');
@@ -58,7 +62,7 @@ class Parser {
       ext = '.' + ext;
     }
 
-    this.extensions[ext.toLowerCase()] = parser;
+    this.extensions.set(ext.toLowerCase(), parser);
   }
 
   findParser(filename) {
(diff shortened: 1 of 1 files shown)


DeMoorJasper:
What exactly changed? Could you add a bt more info on why this change was made?

src/Parser.js
```diff
@@ -5,7 +5,11 @@ const glob = require('glob');
 
 class Parser {
   constructor(options = {}) {
-    this.extensions = {};
+    /**
+     * A map of extension names to either path to Asset class or Asset class itself
+     * @type {Map<string, string|Asset}
+     */
+    this.extensions = new Map();
```
mohsen1: @DeMoorJasper this is the core of the change. I'll clean up unnecessary changes momentary

DeMoorJasper: I see you changed it into a map but wondering what issue this fixes?

DeMoorJasper:
The only reason I could think of for this PR is performance?

I've made a JSPerf and it's about the same (1% faster)
https://jsperf.com/map-vs-object-11116

src/Parser.js
```diff
@@ -58,18 +62,22 @@ class Parser {
...
 
     let extension = path.extname(filename).toLowerCase();
-    let parser = this.extensions[extension] || RawAsset;
+    // console.log({extension});
+    // console.log('registered extensions', ...this.extensions.keys());
+    let parser = this.extensions.get(extension) || RawAsset;
+    // console.log({parser});
     if (typeof parser === 'string') {
-      parser = this.extensions[extension] = require(parser);
+      parser = this.extensions.set(extension, require(parser)).get(extension);
```
DeMoorJasper: If performance is the point of this PR than this actually could be faster if you write it like this:
```Javascript
parser = require(parser);
this.extensions.set(extension, parser);
```

Test case: https://jsperf.com/assign-vs-getting-map

mohsen1:
@DeMoorJasper Thanks for your comments. I cleaned up the code a bit more. 

Not sure how to unit test this. My [first attempt](https://github.com/mohsen1/parcel/blob/21f2bba106fed8454e155d3f844c4dd1d6303591/test/parser.js#L53-L76) didn't really test anything so I removed it.

DeMoorJasper:
Not sure why this actually decreased test coverage this much, hope this can be resolved

mohsen1:
Changes from https://github.com/parcel-bundler/parcel/pull/1442 is also here. if coverage is line based we're adding a lot of new lines with those prettier formatted code. Lets merge #1442 first so this change list becomes what it says

PR #1442, opened 2018-05-26
Enforce Prettier (check if prettier is run in lint script)

mohsen1:
I noticed running `yarn format` in master will leave some diffs behind. This change in package.json makes sure running Prettier will leave no diff behind. 

The rest of changes are result of running `yarn format` in master.


mohsen1:
Weird. It's probably a bug with Prettier CLI. Tried 87efabf to see if that resolves it. I'll report to Prettier otherwise

fathyb:
Thanks @mohsen1! 👍 I pushed some commits to fix CI tests, hope you don't mind

PR #1444, opened 2018-05-27
Add unit tests for line counter

mohsen1:
I'm curious if not using `require('os').EOL` for splitting lines in lineCounter is problematic or not. Since I don't have a Windows machine I'm testing this in your CI! :D


DeMoorJasper:
Wait what is codecov just not working properly?
This should increase test coverage right?

mohsen1:
my bad. I left the `.only` there :( 

Suggestion: Your lint rules should be a bit stricter as this project grows.

Issue #1389, opened 2018-05-18
Pipeline should bundle all parts of generated with the same type instead of only the last


DeMoorJasper:
# 🐛 bug report

The asset pipeline drops the first JS output from generated if a second one is also in the generated Array.
This mainly occurs when having an asset that has both css and js, as in hmr mode css adds a js file to the generated array

## 🎛 Configuration (.babelrc, package.json, cli command)

Zero Config, using parcel-plugin-svelte

## 🤔 Expected Behavior

Should add both scripts to the bundle

## 😯 Current Behavior

Adds the latest script to the bundle

## 💁 Possible Solution

Concat the scripts, add both or create a seperate HMR bundle that gets added to the html in a seperate script tag that replaces scripts, css and reloads html. Instead of bundling inside asset bundles

## 🔦 Context

https://github.com/DeMoorJasper/parcel-plugin-svelte/issues/15

## 💻 Code Sample

https://github.com/DeMoorJasper/parcel-plugin-svelte/issues/15

## 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Parcel           | 1.8.1
| Node             | v10
| npm/Yarn         | latest
| Operating System | OS X High Sierra

PR #1456, opened 2018-05-29
Process inline styles and scripts

### LINKED ISSUE #541: background image using html inline styles is not working

### LINKED ISSUE #1231: parcel build & inline JS issue
🐛 bug report

No parcel config

```html
<textarea v-model="input" rows="8" cols="200" onkeydown="if(event.keyCode===9){var v=this.value,s=this.selectionStart,e=this.selectionEnd;this.value=v.substring(0, s)+'\t'+v.substring(e);this.selectionStart=this.selectionEnd=s+1;return false;}"></textarea>
```

### 🤔 Expected Behavior

```
<textarea v-model="input" rows="8" cols="200" onkeydown="if(event.keyCode===9){var v=this.value,s=this.selectionStart,e=this.selectionEnd;this.value=v.substring(0, s)+'\t'+v.substring(e);this.selectionStart=this.selectionEnd=s+1;return false;}"></textarea>
```

### 😯 Current Behavior

```html
<textarea v-model="input" rows="8" cols="200" onkeydown="if(9===event.keyCode){var t=this.value,e=this.selectionStart,i=this.selectionEnd;return this.value=t.substring(0,e)+" \t"+t.substring(i),this.selectionStart="this.selectionEnd=e+1,!1}""></textarea>
```

`onkeydown` value is escaped incorrectly and code above has syntax error.

### 🔦 Context

`parcel index.html` is OK while `parcel build index.html` is not.

### 🌍 Your Environment

| Parcel           |  1.7.1
| Node             |  9.11.1
| npm/Yarn         |  yarn 1.5.1
| Operating System | macOS 10.13.4

### LINKED ISSUE #1297: 🐛 Build alter inlined javascript
🐛 bug report 

Hello,
When I'm using parcel-cli "**parcel build src/index.html --public-url ./ --out-dir build/**" the output alter inline JS leading to not working code. This does not happen when using **parcel src/index.html** (bundled server).

🤔 Expected Behavior

The inlined JS should not be changed.

😯 Current Behavior

The inlined JS in my HTML is being wrongly altered.
How it is in my source file :
```html
<link rel="stylesheet" href="less/incarna.less" media="none" onload="if(media!='all')media='all';">
```
The output in build/index.html:
```html
<link rel="stylesheet" href="incarna.3defd595.css" media="none" onload=""all"!=media&&(media="all")"> 
```
As you can see, **if(media!='all')media='all'**; as been translated to **"all"!=media&&(media="all")**

💁 Possible Solution
I have no idea!

🔦 Context
Doing a simple landing page using UIKIT.

🌍 Your Environment
I'm using VIsualStudioCode.

### LINKED ISSUE #1491: Script tags are not executed in the global scope and do not work after bundling
# 🐛 bug report

Included vanilla javascript scripts are not executed in the global scope the way they are in a non-bundled, statically-served site.

No configuration.

Bundling a vanilla, static website.

## 🤔 Expected Behavior

A vanilla bundled website should behave the same before and after it's bundled.

#### includedScript.js
```js
var testVariable = 1000;
```
#### index.html
```html
<head><html>
    <script src="./includedScript.js" />
    <script>
        // Prints "1000" in static server
        // Undefined variable error when bundled
        console.log(testVariable);
    </script>
</html><body>
</body></html>
```

## 💁 Possible Solution

Execute normal included script tags in the global scope instead of scoping them. Required or otherwise imported scripts can still be scoped accordingly.

## 🔦 Context

Trying to bundle a static website for production and use older libraries that rely on global scope.

## 🌍 Your Environment

| Parcel           | 1.8.1
| Node             | 8.6.0
| npm         | 5.3.0

DeMoorJasper:
This PR changes the behaviour of inline scripts and styles in html to be processed by Parcel, similar to how we process Vue Assets.

Adds the ability to use files in css and modern syntax in JS.

Also removes the need for any minification for JS or CSS from htmlnano

Closes #541 Closes #1297 Closes #1231 Closes #1491


src/assets/HTMLAsset.js
```diff
@@ -160,8 +165,71 @@ class HTMLAsset extends Asset {
...
+      return node;
+    });
+
+    return parts;
+  }
+
+  async postProcess(generated) {
+    // Hacky way of filtering out the css hmr JS
+    // Related: https://github.com/parcel-bundler/parcel/issues/1389
+    generated = generated.filter(
```
DeMoorJasper: This line might cause issues, related to this issue: https://github.com/parcel-bundler/parcel/issues/1389 
There is currently no way to tell in generated you don't want the hmr code for css assets.

src/assets/HTMLAsset.js
```diff
@@ -160,8 +165,71 @@ class HTMLAsset extends Asset {
...
-  generate() {
-    return this.isAstDirty ? render(this.ast) : this.contents;
+  async generate() {
+    let parts = [];
+    this.ast.walk(node => {
+      if (node.tag === 'script' || node.tag === 'style') {
+        if (node.content) {
+          let value = node.content.join('').trim();
+          if (value) {
+            parts.push({
```
mohsen1: This is a bit too clever for my IQ 😅

stevengill:
This PR looks great!

One thing I noticed it doesn't handle is inline `background-image: url(./someImg.png)`. If i have that code in a css file, it works fine (realizes `someImg.png` is a dependency, hashes it and moves it over to `dist`, updates references to it)

I **think** it is handled in `SASSAsset` at https://github.com/parcel-bundler/parcel/blob/master/src/assets/SASSAsset.js#L41-L46 and `CSSAsset` at https://github.com/parcel-bundler/parcel/blob/master/src/assets/CSSAsset.js#L61-L77

DeMoorJasper:
@stevengill Not sure what you mean, I got it to work in a new test I've added, could you create a small test-case that fails?

stevengill:
my mistake @DeMoorJasper! Sorry about that. It is definiltey working. I tried it yesterday for a while and couldn't get it working. I think it was because I was using quotes in `url("./SomeImg.png");`

Thanks again!

stevengill:
Alright, I've played around more with the branch and run into some issues. 

I've got a [sample repo](https://github.com/stevengill/parcelBackgroundBug) with a index.html that renders fine with master branch (minus the `background-image:url('./someImg.png')`.

With this branch, the styles are all messed up. I haven't tracked down what in the html is causing this to happen with this branch yet. 

Please ignore the (text) in the html 😛. It is an internal email newsletter we sent out a while ago at work.

I'll continue to play around with this and see if I can figure out what is going on. It would be really nice if I could actually add some console.logs to `HTMLAsset` so I could debug this. My logs never show up though (`parcel-bundler` is `yarn link`ed)

devongovett:
This is cool. How are we going to handle imports/requires inside the inline scripts/styles though? We'd need to run the packager to merge the modules together somehow, and then put the result back in the HTML...

What happens right now when you put a `require()` call in your inline script?

devongovett:
Or maybe we don't try to inline `require` at all, and make another external `<script>` tag before the inline one to load it?

DeMoorJasper:
@devongovett I originally made this to prevent any posthtml issues with js or css transformations. If you'd like this to handle requires as well it should probably inline the full bundle with requires, as a magically appearing external script seems a bit strange to me.

Currently it works perfectly with requiring images and such in css which was one of the issues this PR resolves.

For JS it currently just creates the JS bundle and throws an error, the bundle doesn't get added to html using the script tag

devongovett:
I do like the idea of this feature, we just have to make sure that it is not inconsistent with how the rest of parcel works. Extracting dependencies from inline CSS is obviously a good thing, but the fact that `require`/`import`/`@import` don't work inside `script` and `style` tags might be weird for people. That would also be the hardest thing to implement...

Maybe we start with this for now and design packagers in a way that's more compassable for Parcel 2 so that we can just use the CSS and JS packagers inside the HTML one for inline stuff. Then we can fully support `require` etc. at that time. In the mean time, we will still get the benefits of at least some of this.

DeMoorJasper:
Awesome, I’ll try to resolve those issues soonish

test/html.js
```diff
@@ -277,11 +277,6 @@ describe('html', function() {
       )
     );
 
-    // minifyJson
```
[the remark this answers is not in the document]
DeMoorJasper: Because htmlnano is no longer responsible for JSON minifying, although this caused the tests to not show that it actually changed json into JS so I've put it back

DeMoorJasper:
Should be working in most cases now, anyone who'd like to test this. Please do.

devongovett:
Simplified slightly so that we actually pass through the original nodes to the pipeline so we can more easily replace their contents in postProcess without hackiness to detect things like css hmr.

Also added some temporary handling for `@import` in <style> tags, and `import`/`require` in <script> tags. For CSS, the imports are not inlined as usual, but modified to point to the correct URL for a separate bundle. For JS, we just throw an error saying imports and requires aren't supported in inline scripts yet. This should be clearer than just not working at runtime.

PR #1457, opened 2018-05-29
Surface Bundler error to browser

mohsen1:
Instead of sending a generic error with the 500 response, print the error message and stack trace in the response HTML payload. 

Feel free to comment. 

screen shot 2018-05-31 at 4 23 03 am

### DIFF
--- a/src/Bundler.js
+++ b/src/Bundler.js
@@ -60,7 +60,7 @@ class Bundler extends EventEmitter {
     this.watcher = null;
     this.hmr = null;
     this.bundleHashes = null;
-    this.errored = false;
+    this.error = null;
     this.buildQueue = new PromiseQueue(this.processAsset.bind(this));
     this.rebuildTimeout = null;
 
@@ -213,7 +213,7 @@ class Bundler extends EventEmitter {
     let isInitialBundle = !this.entryAssets;
     let startTime = Date.now();
     this.pending = true;
-    this.errored = false;
+    this.error = null;
 
     logger.clear();
     logger.status(emoji.progress, 'Building...');
@@ -290,7 +290,7 @@ class Bundler extends EventEmitter {
       this.emit('bundled', this.mainBundle);
       return this.mainBundle;
     } catch (err) {
-      this.errored = true;
+      this.error = err;
       logger.error(err);
       if (this.hmr) {
         this.hmr.emitError(err);
@@ -493,7 +493,7 @@ class Bundler extends EventEmitter {
       return;
     }
 
-    if (!this.errored) {
+    if (!this.error) {
       logger.status(emoji.progress, `Building ${asset.basename}...`);
     }
 
@@ -713,7 +713,11 @@ class Bundler extends EventEmitter {
 
   async serve(port = 1234, https = false) {
     this.server = await Server.serve(this, port, https);
-    this.bundle();
+    try {
+      await this.bundle();
+    } catch (e) {
+      // ignore: server can still work with errored bundler
+    }
     return this.server;
   }
 }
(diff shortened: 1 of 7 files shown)


DeMoorJasper:
Seems like an awesome idea, as we already have the overlay

mohsen1:
Currently I'm stuck with pretty printing output of babel-code-frame in HTML. I know that https://new.babeljs.io/repl is doing something similar (without syntax highlighting)but I'm not sure how they print the code frame. I need to look into it. 


Should I consolidate this code with the overlay? Overlay is manually putting together the HTML as well. Should we depend on a templating library for generating the HTML?

DeMoorJasper:
@mohsen1 I think it’s safe to write a general function that generates the html and just pass a boolean or something that rewrites the css very slightly to be overlay or static

src/Bundler.js
```diff
@@ -713,7 +713,11 @@ class Bundler extends EventEmitter {
 
   async serve(port = 1234, https = false) {
     this.server = await Server.serve(this, port, https);
-    this.bundle();
```
mohsen1: Previously if this line was throwing an error we had a silent unhandled promise rejection

mohsen1:
@DeMoorJasper @devongovett Can you guys take a look? This is working :)

src/Server.js
```diff
@@ -71,10 +75,28 @@ function middleware(bundler) {
...
 
-    function send500() {
-      res.setHeader('Content-Type', 'text/plain; charset=utf-8');
+    function send500(error) {
+      res.setHeader('Content-Type', 'text/html; charset=utf-8');
       res.writeHead(500);
-      res.end('🚨 Build error, check the console for details.');
+      let errorMesssge = '<h1>🚨 Build Error</h1>';
+      if (process.env.NODE_ENV === 'production') {
+        errorMesssge += '<p><b>Check the console for details.</b></p>';
```
DeMoorJasper: Not sure if this is necessary, the built-in parcel server should never be used in production

src/Server.js
```diff
@@ -71,10 +75,28 @@ function middleware(bundler) {
...
+      } else {
+        const {message, stack} = prettyError(error, {color: true});
+        errorMesssge += `<p><b>${message}</b></p>`;
+        if (stack) {
+          errorMesssge += `<div style="background: black; padding: 1rem;">${ansiToHtml.toHtml(
+            stack
+          )}</div>`;
+        }
+      }
+      res.end(
```
DeMoorJasper: Why not put use one template string here instead of an array?

src/Server.js
DeMoorJasper: Parcel can be used as middleware in a production server, so I guess it's good to check this though.

src/Server.js
mohsen1: This is to make sure we're not sending extra whitespaces or not ending up with awkward code indentations. Depending on project coding style I can update.

src/Server.js
mohsen1: Yeah it should never happen but just in case...

src/Server.js
DeMoorJasper: Ah, alright, I don't think the weird indentation matters, but that's personal taste.

DeMoorJasper:
Looks good to me, I'm just not 100% convinced and it's a pretty large change so I'll let someone else review it

test/server.js
```diff
@@ -23,14 +23,13 @@ describe('server', function() {
...
         res => {
-          if (res.statusCode !== 200) {
-            return reject(new Error('Request failed: ' + res.statusCode));
-          }
-
           res.setEncoding('utf8');
           let data = '';
           res.on('data', c => (data += c));
           res.on('end', () => {
+            if (Math.floor(res.statusCode / 100) !== 2) {
```
DeMoorJasper: Does this ever respond with a code not equal to 200?

mohsen1: Yes. My new test case has 500 response

mohsen1: if you ever respond with 201 or other 2xx you'll not need to update this line. I'm open to reverting back this to comparing strictly against 200 if you feel strongly about it

mohsen1:
@devongovett can you take a look? Thanks!

PR #1472, opened 2018-06-01
Minor logger improvement and a bugfix

### LINKED ISSUE #349: 🐛 Long error messages gets cut off
### 😯 Current Behavior

image
image

As you can see in the first image the error message reaches the end of the row and gets cut off instead of wrapping to the next line. And as you can see in the second image the error message gets cut off at a slightly different position when the terminal is smaller.

### 🤔 Expected Behavior

The entire error message should always be displayed.

### 🌍 Your Environment

| Parcel           | 1.2.0
| Node             | 9.3.0
| Yarn         | 1.3.2

### LINKED ISSUE #1293: 🐛IntelliJ terminal output not clearing lines
When running parcel from inside of Webstorm's Terminal or a Run Configuration, parcel's "Building ..." messages all show up on a single line. Running in a regular terminal, they are clearing the previous line and overwriting it.

This happens with any site I've tried with parcel and does not see to be configuration based.

### 🤔 Expected Behavior

I expect either to have new lines between the "Building..." lines, or for it to work the exact same as in a regular terminal.

### 😯 Current Behavior

As you can see from this screen shot:

image

It is not correctly clearing the previous line.

### 💁 Possible Solution

This is, arguably an IntelliJ bug. However, I don't know what parcel is doing that's different from any of the other tools (ex: `npm`, `yarn`) that clear the line and replace it (all of which work). I would love to see parcel either use the same mechanism as other packages that do this correctly, or a CLI option for parcel to forego clearing the line and instead putting a new line between messages.

### 🔦 Context

It makes it incredibly difficult to read error messages output during the parcel build. (I have to scroll right and look for any red text.)

### 🌍 Your Environment

I've tried with Webstorm, IntelliJ IDEA, and PyCharm; all have this issue.

DeMoorJasper:
This PR adds a spinner using `ora`, this replaces the status code.
As this is already battletested it should fix some issues we've been seeing with status.

This PR also changes error logging to log the message using console.log instead of status

This also fixes a bug in lineCounter, as it wasn't being ansiStripped, which caused incorrect lineCounts (mainly for error Stacks)

screen shot 2018-06-01 at 20 11 20

Closes #923 Closes #349 Closes #1293

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,6 +51,7 @@
     "node-forge": "^0.7.1",
     "node-libs-browser": "^2.0.0",
     "opn": "^5.1.0",
+    "ora": "^2.1.0",
     "physical-cpu-count": "^2.0.0",
     "postcss": "^6.0.19",
     "postcss-value-parser": "^3.3.0",
--- a/src/Bundler.js
+++ b/src/Bundler.js
@@ -222,7 +222,7 @@ class Bundler extends EventEmitter {
     this.error = null;
 
     logger.clear();
-    logger.status(emoji.progress, 'Building...');
+    logger.status('spinner', 'Building...');
 
     try {
       // Start worker farm, watcher, etc. if needed
@@ -513,7 +513,7 @@ class Bundler extends EventEmitter {
     }
 
     if (!this.error) {
-      logger.status(emoji.progress, `Building ${asset.basename}...`);
+      logger.status('spinner', `Building ${asset.basename}...`);
     }
 
     // Mark the asset processed so we don't load it twice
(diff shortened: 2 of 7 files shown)


devongovett:
Made a slight refactor to create separate `logger.progress` and `logger.success` methods, instead of one `logger.status`. This should be easier to use since it handles the emoji/spinner for you, and the code is a bit simpler.

sebastian-schmitt:
it seems as if there is a problem with the parcel-ora-integration when running parcel inside a docker container. This may be related to [this](https://github.com/sindresorhus/ora/issues/58) and [this](https://github.com/sindresorhus/ora/issues/9).

I'm sorry for just commenting here instead of opening a new issue - as soon as I get to a "real" keyboard I will do so.

Issue #1412, opened 2018-05-22
Add option --host to development server, currently only localhost


kwmiebach:
# 🙋 feature request

The dev server should be able to listen on a non localhost interface. There should be a --host option to allow setting a different IP or 0.0.0.0 for all interfaces. 

## 🤔 Expected Behavior

When you start a dev server with `parcel index.html --host 10.1.1.10` the dev server should listen on the 10.1.1.10 interface. (I guess also the websocket interface has to be changed in this case.)

When you start a dev server with `parcel index.html --host 0.0.0.0` the dev server should listen on all interfaces.

When you omit the --host option, the dev server should only serve on localhost, like before. This is important for security reasons and backwards compatibility.

## 😯 Current Behavior

No --host option is possible and the dev server always listens on localhost. Some people believe it listens on 0.0.0.0 (which means 'listen on all interfaces/IPs') but this is not true. And it would be insecure.

## 💁 Possible Solution

Implement a --host option and pass its value to the http server. Probably in this file as second parameter after the port parameter:

https://github.com/parcel-bundler/parcel/blob/3988ffcf66033722c41d767e714ed1a01b572b27/src/Server.js#L103

I believe this needs also to be changed in the websocket server implementation:

https://github.com/parcel-bundler/parcel/blob/master/src/HMRServer.js

## 🔦 Context

You won't need this if you only develop on your laptop / local machine.

This can be helpful if the development machine is a different machine in the same network, like a development VM. And there are special setups where you need it, see examples below.

## 💻 Examples

There can be situations like with vagrant or docker, or on continous integration servers or hosted web editors where you need to set a different interface. This has been mentioned in some other issues.

Another common situation is, when you are testing with a real phone as the client. Or if you develop on a lets say linux machine and want to see it live reloading on multiple parallel heads, like a windows machine with IE/Edge, a Mac, etc.

Many development servers have it: 

#### Websocket

https://github.com/webpack/webpack-dev-server/blob/master/examples/cli/host-and-port/README.md#ipv4

`npm run webpack-dev-server -- --open --port 5000 --host 0.0.0.0`

#### Angular Cli

https://github.com/angular/angular-cli/#usage

`ng serve --host 0.0.0.0 --port 4201`

#### Vue.js CLI

https://github.com/vuejs/vue-cli/blob/dev/docs/cli-service.md#serve

`vue-cli-service serve --host 0.0.0.0`

kwmiebach:
reference #278 
reference #924

[...]

PR #1482, opened 2018-06-02
Added new --host option and updated relevant tests

victor-cordova:
Hello! This PR should add the new option --host as requested on [1412](https://github.com/parcel-bundler/parcel/issues/1412). Based on the comments I assumed perhaps ```HMRServer.js``` had to be updated but after testing, that change seems unecessary.

## Rundown of my changes
- A new cli option called --host which receives the host to listen on (defaults to localhost)
- Updated server tests

## How I tested
1. Created an Ubuntu guest virtual machine
2. Run ```yarn build``` in the host machine after changing the code
3. Run ```bin\cli.js -p 999 --host <LOCAL_INTERFACE_IP> <SOME_LOCAL_ENTRY_POINT>``` in the host machine
4. Connected successfully from the guest machine to the host machine via <LOCAL_INTERFACE_IP>:999

### DIFF
--- a/src/Bundler.js
+++ b/src/Bundler.js
@@ -711,8 +711,8 @@ class Bundler extends EventEmitter {
     return Server.middleware(this);
   }
 
-  async serve(port = 1234, https = false) {
-    this.server = await Server.serve(this, port, https);
+  async serve(port = 1234, host = 'localhost', https = false) {
+    this.server = await Server.serve(this, port, host, https);
     this.bundle();
     return this.server;
   }
--- a/src/Server.js
+++ b/src/Server.js
@@ -88,7 +88,7 @@ function middleware(bundler) {
   };
 }
 
-async function serve(bundler, port, useHTTPS = false) {
+async function serve(bundler, port, host, useHTTPS = false) {
   let handler = middleware(bundler);
   let server;
   if (!useHTTPS) {
@@ -100,7 +100,7 @@ async function serve(bundler, port, useHTTPS = false) {
   }
 
   let freePort = await getPort({port});
-  server.listen(freePort);
+  server.listen(freePort, host);
 
   return new Promise((resolve, reject) => {
     server.on('error', err => {
@@ -115,10 +115,13 @@ async function serve(bundler, port, useHTTPS = false) {
               `configured port ${port} could not be used.`
             )}`
           : '';
+      const hostAddress = host || 'localhost';
 
       logger.persistent(
         `Server running at ${logger.chalk.cyan(
-          `${useHTTPS ? 'https' : 'http'}://localhost:${server.address().port}`
+          `${useHTTPS ? 'https' : 'http'}://${hostAddress}:${
+            server.address().port
+          }`
         )} ${addon}`
       );
 
--- a/src/cli.js
+++ b/src/cli.js
@@ -13,6 +13,7 @@ program
     'set the port to serve on. defaults to 1234',
     parseInt
   )
+  .option('--host <host>', 'set the host to listen to. defaults to localhost')
   .option(
     '--hmr-port <port>',
     'set the port to serve HMR websockets, defaults to random',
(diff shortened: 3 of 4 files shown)


victor-cordova:
@DeMoorJasper after doing some testing I'm not sure if websockets would be the main issue
I've performed the following test and it seems to be working ok (please let me know if you can think of a different scenario that's not considered in the changes I made):

1. Run the following command from a mac on my local network ```bin\cli.js -p 1234 --host <LOCAL_INTERFACE_IP> index.html```
2. Connected to <LOCAL_INTERFACE_IP>:1234 from another windows machine in my local network
3. Changed a css file (references by index.html) from the mac machine
**Result**: Windows machine updated styles without reloading the page

DeMoorJasper:
@victor-cordova alright sounds good :) Just want to make sure

@kwmiebach does this PR fix the original issue?

devongovett:
by default the server already listens on all hosts. is there a usecase to restrict that for some reason?

victor-cordova:
@devongovett I believe the use case would be the one described here [Comment](https://github.com/parcel-bundler/parcel/pull/278#issuecomment-354047330). What do you think?

devongovett:
there is already `--hmr-hostname` for specifying the hostname of the HMR server.

kwmiebach:
I did not find in my tests that the server listens by default on all interfaces. It only listens on localhost, which is normally 127.0.0.1.

Btw. if it would do that, that would be a security issue.

And shouldn't `hostname` and `interface/IP address` be distinguished? A server cannot 'listen' on a hostname, it can only listen on an interface/IP address.

(But it could resolve a hostname to an IP address and listen on that IP address. Or it could listen on an interface but only answer requests with a specific hostname on that interface. But both would not be necessary for a development environment)

kwmiebach:
@devongovett you said that already in another thread but it is not true. it only listens on localhost

devongovett:
According to the [node docs](https://nodejs.org/dist/latest-v8.x/docs/api/net.html#net_server_listen_port_host_backlog_callback):

> If host is omitted, the server will accept connections on the unspecified IPv6 address (::) when IPv6 is available, or the unspecified IPv4 address (0.0.0.0) otherwise.

kwmiebach:
@devongovett thank you for looking into this. give me some time and I will repeat my tests

kwmiebach:
I repeated my tests and I can confirm now that parcel listens on all interfaces.

I was wrong, sorry for the confusion and work it cost some of you.

But: Maybe the feature that @victor-cordova implemented in his pull request could still be useful.

The default behaviour to listen on all interfaces could be seen as a security issue in some circumstances. In fact many development servers only listen to localhost by default. But at least it would now be possible to choose the interface where parcel listens, so a developer that does not want to expose the dev server to the whole network, could choose not to. 

-----

And there is something I noticed:

When you start parcel, it tells you it is only listening to localhost, which is wrong, as everybody knows now. This is definitely a security issue, because parcel listens to more interfaces than it tells you. Example:

```
$ parcel index.html
✨  Built in 1.91s.

``` 

This would be easy to fix here, where the string `localhost` is hardcoded: https://github.com/parcel-bundler/parcel/blob/3988ffcf66033722c41d767e714ed1a01b572b27/src/Server.js#L121

DeMoorJasper:
@kwmiebach I agree this `--host` flag would still be usefull for security reasons.
Also the output should probably print `listening on port 1234` or `Listening on 0.0.0.0:1234` the latter might be confusing for some users though.

It has been requested a lot as well and if we close this again, I think the question will pop up again.

I'm wondering however if we could merge `--hmr-hostname` and `--hostname` to just `--hostname` as they both set the same value eventually anyway?

victor-cordova:
I agree with @DeMoorJasper in that the new flag would be useful for security reasons. I'm not sure if merging them would be a good idea at this point. Mainly because it would be a breaking change for people already using ```--hmr-hostname``` (unless ```--hmr-hostname``` gets deprecated). Also, I really don't know if most people prefer having a separate hostname and hmr-hostname (for me it would make sense though).

devongovett:
Going to close this. The parcel server is not meant to be used in production. I don't really see why this feature would be useful for a development server.

The `--hmr-hostname` option was added not to control the interfaces the server listens on, but to fix cases where `window.location.hostname` did not work to get the hostname from the client. See https://github.com/parcel-bundler/parcel/pull/492.

ross-pfahler:
I think this would be useful for cases that certain people have with running local servers on different hostnames.

As you know @DeMoorJasper, certain use cases at large companies require a certain hostname, eg `localhost.corp.bigcompany.com:1234` to be used due to CORS restrictions.

I'm hoping we can reconsider this PR -- would be helpful in reducing confusion!

Issue #1039, opened 2018-03-21
Support WebExtensions manifest.json


sntran:
# 🙋 feature request

### 🤔 Expected Behavior

Parcel takes a `manifest.json` (not to be confused with normal web app `manifest.json`) as the entry point, and handles the assets (HTML, JS, CSS, image files) defined in there. It copies them, as well as referenced assets inside, to `./dist`. Of course, assets will be processed if necessary, i.e., TS to JS, LESS to CSS.

### 😯 Current Behavior

Parcel only copies the `manifest.json` to `./dist`.

### 💁 Possible Solution

Implements a new Asset handler for this.

### 🔦 Context

When developing an extension for browsers, there are many different files in different places, i.e., `background/index.ts`, `popup/index.html`, `devtools/index.html`, `images/icon32.png`, `inject/inject.less`.... However, they are all defined in the main `manifest.json`, which makes it suitable for being the entry point.

Besides, Parcel does not handle multiple entry points nicely, yet.

fathyb:
I'd really love to see WebExtensions support in Parcel. Assets are currently per extension, and `manifest.json` is too broad and could be used in a totally different context.

We would need something to discriminate WebExtensions manifest from other `json` files. `webmanifest` is used for Web Application manifests, perhaps using a dedicated extension? Not a big fan since it's not standard, but I don't see much alternatives. Any thoughts?

davidnagli:
@fathyb I don't think we should be trying to guess weather the user is trying to bundle extensions, since both having a `manifest.json` file doesn't necessarily mean it's a browser extension. Instead, I think we should have a `--target` option which tells Parcel explicitly that you intend to package an extension.

I do agree that the lack of standardization is gonna be an issue since it's gonna be different across Chrome, Safari, Firefox, Opera, etc. So I think for now we should focus on **chrome extensions only** (they have the largest market share of extensions, and MUCH better documentation)

---

Side note: In the far future it would be really cool if Parcel could build a single extension into multiple browser targets, transpiling it for each browser to match the API differences.

sntran:
I agree with @davidnagli . The user should not have to rename the extension. They should specify what the input type is.

And yes, we definitely want to follow WebExtensions API. There are some differences between browsers, but we can always base the support on Chrome as the initial prototype, and work out the differences later on.

ZY-Ang:
Is this feature being worked on? Would love to have this for an extension I'm working on.

kevincharm:
[...]

PR #1484, opened 2018-06-03
WebExtension Support

kevincharm:
Hi all! Loving the work on Parcel. This PR adds support for using `manifest.json` as an entry point for a WebExtension project, as per the [MDN documentation](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/manifest.json). I've been using this experimental version locally for my WebExtension project and it's working pretty well. Looking for comments/suggestions on improving the implementation!

I'd like to thank @helpermethod for his initial input.

## tl;dr
- ~Adds a `--web-extension` option to the cli. It was suggested in #1039 to add a new target; i.e., `--target=web-extension`. I tried this at first but I was running into a lot of issues with resolving modules. I figured the target is still actually `browser` so this extra option is my current solution.~ Now testing for mandatory keys as per the MDN documentation to determine whether a `manifest.json` is indeed a WebExtension manifest.
- Implements new asset handler `WebExtensionManifestAsset`
- Does not bundle `browser-polyfill.js` (import it yourself)
- PWA `manifest.json` is now also handled as a `WebManifest`
- Expressions in the form of `browser.extension.getURL('some-script.js')` are not resolved (not sure how hard this would be to do)


devongovett:
What's the difference between this and the existing WebManifestAsset? https://github.com/parcel-bundler/parcel/blob/master/src/assets/WebManifestAsset.js

kevincharm:
@devongovett I believe `WebManifestAsset` handles `manifest.webmanifest` files which are a totally different thing to do with PWAs ([MDN - Web App Manifest](https://developer.mozilla.org/en-US/docs/Web/Manifest))

devongovett:
hmmmmm, this is going to be interesting. I believe `manifest.json` is also used for PWA web manifests as well. Currently we only support the `.webmanifest` extension but it seems much more common to use `.json` as the extension - there are a bunch of issues here about that.

We may need to merge the two into a single asset type that can handle both formats. WDYT?

kevincharm:
Is it the intention that a file extension (or a specific filename in this case) should be uniquely handled by a single `Asset` class? We can certainly combine them into a `ManifestJsonAsset` but I would still lean towards returning separate modules for `WebManifestAsset` and `WebExtensionManifestAsset` 
 -- for readability's sake -- because they are unrelated. I'll play around with it to see if I can make it more cohesive.

src/utils/isWebExtensionManifest.js
```diff
@@ -0,0 +1,32 @@
...
+ * @param {string} filename Filename to test.
+ * @returns {boolean}
+ */
+function isWebExtensionManifest(filename) {
+  if (path.basename(filename) !== 'manifest.json') {
+    return false;
+  }
+
+  try {
+    const manifestFile = fs.readFileSync(filename, {encoding: 'utf8'});
```
kevincharm: Kept this `readFileSync` because using the async version would have introduced a lot of changes (`addURLDependency` becomes `async`). This code path only runs if the filename is `manifest.json`.

kevincharm:
@devongovett what do you think of this? It will now be able to discern what kind of `manifest.json` (WebExt or PWA) an asset is and return the appropriate parser for it. Also `.webmanifest` assets are still handled.

src/assets/ManifestAsset/index.js
```diff
@@ -0,0 +1,11 @@
+const isWebExtensionManifest = require('../../utils/isWebExtensionManifest');
+
+function resolveManifestAsset(filename) {
+  if (isWebExtensionManifest(filename)) {
+    return require('./WebExtensionManifestAsset');
+  }
+
+  return require('./WebManifestAsset');
```
kevincharm: Alternatively, we can just chuck these two classes (`WebExtensionManifestAsset` and `WebManifestAsset`) into this one file and export them all (`WebManifestAsset` is still being referenced individually by the parser for `.webmanifest` file ext).

src/utils/isWebExtensionManifest.js
DeMoorJasper: What about always resulting in a shared asset in addUrlDependency, and checking in the shared asset which type of manifest this is? Than it'll definitely be async

al-the-x:
I was literally about to suggest the same feature! Thanks @kevincharm!

src/assets/ManifestAsset.js
```diff
@@ -0,0 +1,168 @@
...
+
+    const icons = this.ast[nodeName];
+    for (const size of Object.keys(icons)) {
+      icons[size] = this.processSingleDependency(icons[size]);
+      this.isAstDirty = true;
+    }
+  }
+
+  collectDependenciesForWebExtension() {
+    for (const nodeName of Object.keys(this.ast)) {
```
al-the-x: Love the breakout of individual concerns here. Would you consider a lookup strategy for extracting and processing the parts?

```js
const processors = {
  'background': this.proceesBackground,
  'content_scripts': this.processContentScripts,
  'web_accessible_resources': this.processWebAccessibleResources,
  // . . .
}

for (const nodeName of Object.keys(this.ast)) {
  processors[nodeName] && processors[nodeName].call(this, this.ast[nodeName])
}
```

The `process*` methods would change a little to accept the `node` from the `ast` instead of extracting it themselves:

```js
processContentScripts (contentScripts) {
  // . . . as above . . .
}
```

src/assets/ManifestAsset.js
```diff
@@ -0,0 +1,168 @@
...
+ * A shared asset that handles:
+ * - PWA .webmanifest
+ * - PWA manifest.json
+ * - WebExtension manifest.json
+ */
+class ManifestAsset extends Asset {
+  constructor(name, options) {
+    super(name, options);
+    this.type =
+      path.basename(name) === 'manifest.json' ? 'json' : 'webmanifest';
```
al-the-x: Consider extracting to separate method, e.g. `this.type = this.typeFor(name)`

We may want to test the file beyond just it's `basename` to determine the `type` in future, for example `manifest.js` or `manifest.json5` that gets transformed to `manifest.json` on the way out.

src/assets/ManifestAsset.js
```diff
@@ -0,0 +1,168 @@
...
+    }
+
+    const background = this.ast[nodeName];
+    if (Array.isArray(background.scripts)) {
+      background.scripts = this.processMultipleDependencies(background.scripts);
+      this.isAstDirty = true;
+    }
+    if (background.page) {
+      background.page = this.processSingleDependency(background.page);
+      this.isAstDirty = true;
```
al-the-x: You might consider pulling `this.isAstDirty = true` into `processSingleDependency` to reduce duplication. All the other `process*` methods funnel through `processSingleDependency`, so it seems like a good place to set the dirty check.

src/assets/ManifestAsset.js
```diff
@@ -0,0 +1,168 @@
...
+      );
+    }
+  }
+
+  hasWebExtensionManifestKeys() {
+    const requiredKeys = ['manifest_version', 'name', 'version'];
+    const presentKeys = Object.keys(this.ast).filter(key =>
+      requiredKeys.includes(key)
+    );
+    return presentKeys.length === requiredKeys.length;
```
al-the-x: Could this be shortened as `return requiredKeys.every(key => !!this.ast[key])` i.e. "return `true` if _every_ key in `requiredKeys` has a truthy value in `this.ast`"...?

al-the-x:
The tests in appveyor are freaking out about a circular dependency serializing an object to JSON in one of the `beforeEach` blocks for the HTML asset type:

```
  html
    √ should support bundling HTML (78ms)
    √ should find href attr when not first
    √ should support transforming HTML with posthtml
    √ should find assets inside posthtml (46ms)
    √ should add dependencies referenced by posthtml-include (47ms)
    1) "before each" hook

  1) "before each" hook:
     Uncaught TypeError: Converting circular structure to JSON
      at JSON.stringify (<anonymous>)
      at ChildProcess.target._send (internal/child_process.js:675:23)
      at ChildProcess.target.send (internal/child_process.js:586:19)
      at Worker.send (src\workerfarm\Worker.js:1:11249)
      at queueCopy.forEach.entry (src\workerfarm\Worker.js:2:363)
      at result.child.send.error (src\workerfarm\Worker.js:2:292)
      at WriteWrap.req.oncomplete (internal/child_process.js:690:13)
```

Is this the [known failure on Win32](https://github.com/parcel-bundler/parcel/blob/181a63f156e1d55056fee96ea5841b749a55470f/test/utils.js#L15-L17)?

kevincharm:
Thanks for the review @al-the-x! I will go through your comments and do some of the suggested cleanup when I get some free time this week.

In the mean time, I managed to extract all this logic into a [parcel plugin](https://github.com/kevincharm/parcel-plugin-web-extension) which you can use with the release version of parcel. Should've started with this in the first place 😀

kevincharm:
Going to close this and move development to the [plugin](https://github.com/kevincharm/parcel-plugin-web-extension). I'll open another PR once it's had time to mature a bit more. Thanks everyone for the input so far!

PR #1488, opened 2018-06-04
Add support for Cargo workspaces in Rust integration

hobofan:
Without the PR Parcel tries to infer the output directory for a cargo build in a less flexible way. However, since the project structure of Cargo projects that make use of workspaces feature is slightly different from the normal one, that fails.

This PR uses the `cargo metadata` command to infer the correct target directory, which is should be fairly robust.


devongovett:
Thanks!

PR #1505, opened 2018-06-07
Fix duplicated content due to symlinks

### LINKED ISSUE #1125: Symlink realpaths
Currently, (as of 1.7) if you have a project that looks like this:

| File | Contents |
| --- | --- |
| `index.js` | `require('./target');`<br>`require('./symlink')`
| `target.js` | `...` |
| `symlink.js` |  _(symlink `./target.js`)_ |

The contents and dependencies of `target.js` will be duplicated.

This is because Parcel does not `fs.realpath` file paths as they are resolved, and for good reason: If you have a `npm link`-d package in your `node_modules` and you `fs.realpath` it before resolving its dependencies, then you can get multiple copies of the same dependencies. Basically, Parcel is following Node's `--preserve-symlinks` option.

Dependency resolution should continue to work this way, but before we add a module to the graph, we should then `fs.realpath` it and if it already exists we should not re-add it. This also matches Node's behaviour because symlinked modules do not get re-evaluated in Node, where right now they do (in effect) in Parcel. 

We should:

- Continue resolving dependencies of modules using the non-realpath'd file path.
- Call `fs.realpath` on modules before we add them to the graph and avoid them ending up in our bundles multiple times.

padmaia:
fixes #1125


padmaia:
Yeah, I'm realizing I was too zoomed in on this. I'll spend some more time on it this week.

padmaia:
I'm going to close this PR because based on my findings outlined in [this comment](https://github.com/parcel-bundler/parcel/issues/1125#issuecomment-401451734) I don't think the related issue should be considered a bug to be fixed. If we decide to add an option to accommodate normal Node module resolution I will open up another PR.

PR #1506, opened 2018-06-07
Allow Aliasing of Folder relative to project folder (#1485)

### LINKED ISSUE #1485: Allow Aliasing of Folder relative to project folder
# 🙋 feature request

I would like to propose to add the possibility to alias folders relative to the project directory.

## 🤔 Expected Behavior

I want to be able to import modules/files via the aliased name and the relative path to the target file from the aliased directory, for example:

I want to define:

`
  "alias": {
    "base": "./src/pages/base"
  }
`

I want to import files like this:

`import Button from 'base/Button'`

The import should be resolved to the file "<projectdir>/src/pages/base/Button.ts".

There exists the possibility, that there is module with the same name in the node_modules folder, in this case the alias should probably be the preferred resolution of the import.

## 😯 Current Behavior

Importing fails.

## 💁 Possible Solution

Change the resolving strategy to allow this special case.

## 🔦 Context

Custom Aliases help separating parts of ones application and makes therefore the code and imports a lot cleaner.

Webpack has support for this type of imports: https://webpack.js.org/configuration/resolve/ (See resolve.alias)

## 💻 Examples

I created a repo which tries to implement the requested functionality.

https://github.com/theodm/test-case-for-parceljs

theodm:
This patch adds the possibility to alias folders in a project.

Following new aliasing options are available:

```
alias: {
  "abc": "./src/abc",
  "cde": "/src/cde"
}

import "abc" -> resolves to /src/abc/index.js
import "cde" -> resolves to /src/cde/index.js

import "abc/test.js" -> resolves to /src/abc/test.js
import "cde/test.js" -> resolves to /src/cde/test.js
```

Aliases to another module of node_modules and aliases to folders are distinguished with a path specifier "alias": "./<path>" or "alias": "/<path>" and no path specifier "<alias>": "leftpad".

Behavior is inspired by Webpack: https://webpack.js.org/configuration/resolve/

Closes #1485


src/Resolver.js
```diff
@@ -354,13 +354,24 @@ class Resolver {
...
+        const isAliasToFolder = alias.startsWith('.') || alias.startsWith('/');
+
+        if (isAliasToFolder) {
+          alias = alias + '/' + path.join(...parts.slice(1));
+
+          if (alias.startsWith('/')) {
+            alias = '.' + alias;
+          }
+        } else {
+          // It is a alias to a node module !
```
DeMoorJasper: Typo `It is an alias to a node module!`

test/integration/resolver/package.json
```diff
@@ -1,9 +1,11 @@
-{
-  "name": "resolver",
-  "private": true,
-  "alias": {
-    "aliased": "foo",
-    "aliased-file": "./bar.js",
-    "something": "./nested/test.js"
-  }
-}
+{
```
DeMoorJasper: Can this be cleaned up? Seems like you have changed the formatting of the file or something, so it now overwrites the entire file

src/Resolver.js
```diff
@@ -354,13 +354,24 @@ class Resolver {
...
+      alias = aliases[parts[0]];
+
+      // If the alias starts with . or ./ it doesn't point
+      // to a node module instead to a folder or file of the project
+      // convention of webpack https://webpack.js.org/configuration/resolve/
+      if (typeof alias === 'string') {
+        const isAliasToFolder = alias.startsWith('.') || alias.startsWith('/');
+
+        if (isAliasToFolder) {
+          alias = alias + '/' + path.join(...parts.slice(1));
```
DeMoorJasper: What about doing it without a spread (Seems less complex to me):
`alias = path.join(alias, '/', parts.slice(1).join('/'));`
You can probably also write it like this (although it will be less error-prone):
`alias = alias + '/' + parts.slice(1).join('/');`

src/Resolver.js
```diff
@@ -354,13 +354,24 @@ class Resolver {
...
+
+          if (alias.startsWith('/')) {
+            alias = '.' + alias;
```
DeMoorJasper: Shouldn't this resolve relative to the project root?

src/Resolver.js
theodm: fixed

test/integration/resolver/package.json
theodm: should be fixed now: wrong line endings :/

src/Resolver.js
theodm: The following alternative is not equivalent, since it will strip the path prefix "./" and "/" which is expected by the code a few lines further down.
 
    alias = path.join(alias, '/', parts.slice(1).join('/'));

code further down

    // If the alias is a relative path, then resolve
    // relative to the package.json directory.
    if (alias && alias[0] === '.') {
      return path.resolve(dir, alias);
    }

The second works though, I will use that.

src/Resolver.js
theodm: This sounds like a good idea to me. I have to confess, I just learned the difference about distinction of root directory / project directory in parcel. Thank you very much 👍 :)

scottwarren:
Sorry, I don't mean to be impatient, but @DeMoorJasper do you know when this will be merged? If there is anything I can do to help, let me know!

alidcastano:
Hey  @theodmand @DeMoorJasper, glad to see this coming 

I'm following up on this tweet (https://twitter.com/devongovett/status/1014209304419655680) 

I'm wondering, if you configure a `~/` path, will this override Parcels built-in one?

DeMoorJasper:
@alidcastano Not sure why you'd want to change `~/` as `./src` is already linked to `/`

devongovett:
Thanks for contributing @theodm! I simplified this slightly:

* Always use `resolveFilename` to resolve the aliased paths. This supports tilde and absolute paths already, and also works correctly on windows where the path separators are different.
* Added a missing test for aliasing a specific sub-path of a module. This was supported before by the first lookup which checked whether the whole path was aliased.
* Always use the `lookupAlias` function to get aliases, for consistency. This also enables glob sub-paths of modules.

Issue #927, opened 2018-03-02
Cannot read property '__esModule' of null


willin:
**Choose one:**  🐛 bug report 

### 🎛 Configuration (.babelrc, package.json, cli command)

```js
{
  "presets": [
    ["env", {
      "targets": {
        "browsers": ["last 2 versions"]
      }
    }]
  ],
  "plugins": [
    ["transform-react-jsx", { "pragma": "h" }]
  ]
}
```

### 🤔 Expected Behavior

when i edit my source codes it should be rebuild successfully (but not).

### 😯 Current Behavior

```
yarn example
yarn run v1.3.2
$ parcel example/index.html --no-cache
Server running at http://localhost:1234
🚨  /Users/willin/Documents/w2fs/i18n/src/index.js: Cannot read property '__esModule' of null
    at Function.normalisePlugin (/Users/willin/Documents/w2fs/i18n/node_modules/babel-core/lib/transformation/file/options/option-manager.js:142:21)
    at /Users/willin/Documents/w2fs/i18n/node_modules/babel-core/lib/transformation/file/options/option-manager.js:184:30
    at Array.map (<anonymous>)
    at Function.normalisePlugins (/Users/willin/Documents/w2fs/i18n/node_modules/babel-core/lib/transformation/file/options/option-manager.js:158:20)
🚨  /Users/willin/Documents/w2fs/i18n/src/index.js: Cannot read property '__esModule' of null    at Function.normalisePlugin (/Users/willin/Documents/w2fs/i18n/node_modules/babel-core/li
b/transformation/file/options/option-manager.js:142:21)    at /Users/willin/Documents/w2fs/i18n/node_modules/babel-core/lib/transformation/file/opti
ons/option-manager.js:184:30    at Array.map (<anonymous>)
    at Function.normalisePlugins (/Users/willin/Documents/w2fs/i18n/node_modules/babel-core/lib/transformation/file/options/option-manager.js:158:20)
🚨  /Users/willin/Documents/w2fs/i18n/src/index.js: Cannot read property '__esModule' of null    at Function.normalisePlugin (/Users/willin/Documents/w2fs/i18n/node_modules/babel-core/li
b/transformation/file/options/option-manager.js:142:21)
    at /Users/willin/Documents/w2fs/i18n/node_modules/babel-core/lib/transformation/file/opti
ons/option-manager.js:184:30    at Array.map (<anonymous>)
[...]

PR #1515, opened 2018-06-09
Fix babel config in package.json

sj26:
I figured out #927.

Babel seems to mutate config. More specifically, presets seem to append plugins every time babel is run. Because the asset caches `asset.package`, when using a package config like `asset.package.babel`, babel mutates the cached package config. Next time it needs the config it gets the same config, and mutates it again.

Instead, if loading config using a packageKey, clone the config before returning so mutations won't be persisted on the cached copy.


sj26:
So quick! Thanks. 🙇🏻‍♂️

PR #1551, opened 2018-06-15
Add .npmignore to reduce install size

mohsen1:
This is mostly to make @styfle happy! 😅

Current parcel bundle size:

https://packagephobia.now.sh/result?p=parcel-bundler


.npmignore
```diff
@@ -0,0 +1,4 @@
+**/*
+!index.js
+!lib/**/*
+!src/**/*
```
DeMoorJasper: Shouldn't yarn.lock & package.json be added to this list as well?

I think this list is a bit extreme, it's probably a better idea to exclude certain folders than excluding everything except a couple files

mohsen1: package.json and README.md are included automatically included. I think taking the opt-in approach guarantees we don't include other files.

fathyb:
@mohsen1 What file will be ignored after this PR? This will exclude the `bin` directory. We also already have a `files` field in `package.json` : 

https://github.com/parcel-bundler/parcel/blob/cc6018ebcb6bb068246ef77c0aa2ee227e11c7b4/package.json#L11-L16

I see 615kB for the publish size, sounds pretty reasonable.

mohsen1:
Completely missed the files entry

PR #1558, opened 2018-06-16
Remove physical-cpu-count

### LINKED ISSUE #1554: physical-cpu-count performance
@jamiebuilds pointed out that Parcel uses `physical-cpu-count` to calculate the amount of workers to spawn. I had a look at the [source](https://gist.github.com/brandon93s/a46fb07b0dd589dc34e987c33d775679) and they're doing `execSync` on almost all platforms. The overhead of which is probably pretty significant. Feel free to close this if you don't care, just wanted to flag it in case it wasn't on your radar.

fathyb:
As noted in #1554 `physical-cpu-count` has some overhead.

Is there any motivation behind not using the number of logical cores? Parcel benefits from HyperThreading, on all my i7 machines I get slightly better performance.

---

Closes #1554

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,7 +50,6 @@
     "node-forge": "^0.7.1",
     "node-libs-browser": "^2.0.0",
     "opn": "^5.1.0",
-    "physical-cpu-count": "^2.0.0",
     "postcss": "^6.0.19",
     "postcss-value-parser": "^3.3.0",
     "posthtml": "^0.11.2",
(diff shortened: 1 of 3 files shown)


DeMoorJasper:
In the description of `physical-cpu-count` it states that in certain workloads it might be slower to use `logical cpus`/`threadcount`.

As far as I know modern operating systems and cpu's are perfectly able to handle any workload using the threads instead of physical cores.

On large projects removing `physical cpu count` might benefit parcel

devongovett:
hmm not really sure how big a deal the physical-cpu-count perf hit really is. we only run it once, on startup, so it's probably not affecting build perf much.

Also we definitely want to use the physical core count, not the logical cpu count. I found it to be much slower to spin up more workers than physical cores.

fathyb:
> hmm not really sure how big a deal the physical-cpu-count perf hit really is

To be honest I don't think it's even noticeable 😅

---

For the rest I agree with @merlinnot, I really don't see why we don't use a technology I paid 400$ extra for. Especially when it gives a consistent performance increase on all big real-world projects.

The only argument I've seen that makes sense is that spawning workers is slower. In this case spawning `physical-cpu-count / 2` is even faster, why not do it?

---

I'm not going to argue more (even if I love to haha), I mainly just wanted to know to motivations behind it.

PR #1559, opened 2018-06-16
Watcher ignore node_modules

### LINKED ISSUE #1557: Why is node_modules watched?
# ❔ Question

Why is node_modules watched?

Intuitively, only source files defined by the user should be watched.

## 🔦 Context

Errors are cropping up of the form

```
"Error: ENOSPC: no space left on device, watch 'my/project/path/node_modules/some/random/file.js'
```

I'm experiencing this error, and it's also been reported in #1438 and #1427.

It appears that every file in node_modules (recursively) is being watched for changes. Intuitively, this should not be necessary. Only the source files should be watched. Right?

Is there any way to disable watching all of `node_modules`? It is possible to [tell Chokidar not to watch node_modules](https://github.com/paulmillr/chokidar/issues/478).

## 💻 Code Sample

index.html:
```html
<html>
<head>
	<title>Parcel Example</title>
	<meta charset="UTF-8" />
</head>
<body>
	<div id="app"></div>
	<script src="src/index.js"></script>
</body>
</html>
```

src/index.js:

```js
console.log('hello Parcel');
```

## 🌍 Your Environment

Ubuntu 

| Parcel           |1.9.0
| Node             |v9.11.1
| npm/Yarn         |6.0.0
| Operating System |Ubuntu 17.10

DeMoorJasper:
Currently parcel is watching node modules. As a result a big node_modules folder causes performance issues and errors (mainly on linux).

This PR changes this, improving watch performance and fixing some issues with watching too many files

Closes #1557


devongovett:
the problem with this is that you can symlink node_modules into your project and work on them simultaneously. with this change they won't update

DeMoorJasper:
@devongovett If we would resolve paths before watching this will be prevented I think? Something like this: #1408

curran:
@DeMoorJasper Nice work!

I wonder how difficult it would be to make this a command line option instead of the default.

DeMoorJasper:
If #1408 would get merged there is probably no need for a cli option, but otherwise I'm open to adding a cli option, so these errors don't occur by default

chikamichi:
Hi, any news on that front? A tough regression on Linux :)

DeMoorJasper:
This Pr should do the trick, as symlinks are being followed this shouldn't even give issues with watching symlinked modules.

So I guess this is ready for merge.

I can always add a cli flag if it's something that might be usefull for quickly fixing a bug in a non symlinked node module.

@devongovett what do you think? Do I need to add a cli flag?

chikamichi:
Hi @devongovett

This PR relates to bug #1557 which seems to qualify as a regression, but no fix has been released since it got reported. The bug report had been closed before a PR was even considered. A few days after, this PR #1559 was proposed but is kind of lingering, and it may be so because it's not very clear what the definitive stand on the matter is:

- Parcel 1.9.x made a decision to watch node_modules by default (which breaks things)
- you said it's more of an edge-case, a quite useful one, but probably not the main use-case (quoting you: "watching node_modules is pretty useful, especially when you have things in there that are symlinked for local development. […] sometimes its really useful to be able to edit a file in node_modules to test something out.")
- this PR would revert the default behaviour to not watching node_modules, and it seems like you would agree merging it, but it does not seem to provide a way to opt-in for watching modules (globally or through symlinks)

On the short term, would it be possible to:

- clarify what Parcel's position is on the matter (is watching node_modules considered a default behaviour, for the time being and in the forthcoming releases?)
- provide an opt-out mechanism (for instance an ENV variable)

Thank you! :)

devongovett:
Parcel has always watched node_modules, so turning it off completely would be a regression. I'd like to come to a middle ground.

We switched to watching directories instead of files in #1279 because of EMFILE errors (running out of file descriptors) when watching individual files. This was especially seen on macOS. However, it seems that watching directories is not supported on Linux, so Chokidar is recursively watching all individual files in node_modules. This is likely more files than just the ones we were watching individually before, since it will include things in node_modules that aren't in the build (e.g. tests, etc.). It also lines up with https://github.com/parcel-bundler/parcel/issues/1557#issuecomment-400522578, which states that it works when turning off the directory watching behavior.

To solve this issue, I think we should disable `shouldWatchDirs` except on macOS. Chokidar has special support for macOS via FSEvents, and on other platforms I believe it resorts to watching individual files and even polling in some cases. It would be great if someone could check this on Windows, to see if it performs better with watching directories or files.

Long term, I think we need to replace Chokidar with a better file watching solution that works well with large numbers of files on more platforms. I will make a followup ticket to discuss that.

chikamichi:
Thank you @devongovett for this detailed feedback :+1:

nicoandresr:
Thank u @devongovett for you explanation, we are wait for the replace of Chokidar. 👏

devongovett:
Made the above changes in #1665. We now only watch directories on macOS.

PR #1577, opened 2018-06-18
get only existing package main

### LINKED ISSUE #1568: Can't resolve some dependencies
# 🐛 bug report
When trying to import a certain dependency i get: ```Cannot resolve dependency '...'```

The error comes when i try to build this package https://github.com/HenrikJoreteg/redux-bundler.
It works if i install it as usual, but i can't build it when it's linked.
I mean, if i install it like this, it works:
```yarn install redux-bundler```,
but if i do this:
```
git clone https://github.com/HenrikJoreteg/redux-bundler
cd redux-bundler
yarn link
cd ../my-app
yarn link redux-bundler
```
It can't resolves the ```redux-persist-middleware```.

## 🤔 Expected Behavior

Parcel should resolve ```redux-persist-middleware``` as usual
## 😯 Current Behavior

Cannot resolve dependency 'redux-persist-middleware'`

## 🔦 Context

I'm trying to test the changes i made to my fork of redux-bundler.
The error doesn't show when i install it regularly because in that case, parcel looks for the compiled redux-bundler package (compiled using microbundle), which doesn't need any dependencies

## 🌍 Your Environment

| Parcel           | 1.9.1
| Node             | 9.11.2
| npm/Yarn         |yarn 1.7.0
| Operating System | elementary os 0.4 (based on ubuntu linux 16.04)

### PR DESCRIPTION
**Pull request by ranfdev**
Sometimes a package has in his package.json a source field, but the package is already bundled/compiled, so there isn't any source file inside that package. Now, before returning any package file, parcel checks if the file exists. If it doesn't exists, parcel tries to use another file, until a real file is found. fixes #1568 

I still need to fix the error...

### DIFF
--- a/src/Resolver.js
+++ b/src/Resolver.js
@@ -228,16 +228,11 @@ class Resolver {
     try {
       pkg = await this.readPackage(dir);
 
-      let fileList = this.listPossibleFiles(pkg);
-      for (let file of fileList) {
-        if (typeof file != 'string') {
-          continue;
-        }
-        if (file == '.' || file == './') {
-          file = 'index';
-        }
-        file = path.resolve(pkg.pkgdir, file);
+      // Get a list of possible package entry points.
+      let entries = this.getPackageEntries(pkg);
 
+      for (let file of entries) {
+        // First try loading package.main as a file, then try as a directory.
         const res =
           (await this.loadAsFile(file, extensions, pkg)) ||
           (await this.loadDirectory(file, extensions, pkg));
@@ -278,10 +273,12 @@ class Resolver {
     return pkg;
   }
 
-  listPossibleFiles(pkg) {
-    let {browser} = pkg;
+  getBrowserField(pkg) {
+    let target = this.options.target || 'browser';
+    return target === 'browser' ? pkg.browser : null;
+  }
 
-  getPackageMain(pkg) {
+  getPackageEntries(pkg) {
     let browser = this.getBrowserField(pkg);
     if (browser && typeof browser === 'object' && browser[pkg.name]) {
       browser = browser[pkg.name];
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: src/Resolver.js [hunk 1]
```diff
@@ -276,16 +276,14 @@ class Resolver {
...
-      entry => typeof entry === 'string'
-    );
-
-    // Default to index file if no main field find
-    if (!main || main === '.' || main === './') {
-      main = 'index';
+    for (let source of [pkg.source, pkg.module, browser, pkg.main, 'index']) {
+      if (
+        typeof source === 'string' &&
+        (await fs.exists(path.resolve(pkg.pkgdir, source)))
```
RafalFilipek: Hi!

According to node.js docs `fs.exists` method [is deprecated](https://nodejs.org/api/fs.html#fs_fs_exists_path_callback). In order to check if files exists we can use [`fs.access`](https://nodejs.org/api/fs.html#fs_fs_access_path_mode_callback) method or sync version [`fs.accessSync`](https://nodejs.org/api/fs.html#fs_fs_accesssync_path_mode).

ranfdev:
The errors are not related to this pr

src/Resolver.js
```diff
@@ -276,15 +276,37 @@ class Resolver {
...
+      main = possibleMain[i];
+      if (typeof main !== 'string') {
+        // if the current main is not a string, trash it
+        possibleMain.pop();
+        continue;
+      }
+      try {
+        // else, check if the file main exists/is accessible. if it does, return the path of the file.
+        const resolvedPath = path.resolve(pkg.pkgdir, main);
+        await fs.access(resolvedPath);
```
DeMoorJasper: Isn't this the responsibility of the library anyone is using?

Shouldn't main always be accesible? Otherwise it's a bug in package.json from the pkg that's being processed

ranfdev: >   // else, check if the file main exists/is accessible
here i'm referring to the main variable in general, not only the pkg.main file. The value of the main variable in that moment could be pkg.source. 
pkg.source is not always present in compiled packages, that's why parcel needs to check if it's accessible. If it's not, parcel should try to resolve the next possibleMain (e.g. pkg.module), until it finds an existing one.
Before this pr, parcel would try to load pkg.source even if it isn't accessible

ranfdev:
Any news?

DeMoorJasper:
Could you resolve the conflicts?

ranfdev:
Ok

src/Resolver.js
devongovett: I think we should just have `getPackageMain` return an array of possible paths, and have `loadDirectory` loop through them and try each one. Since it is already reading files there, we would avoid checking twice.

ranfdev:
@devongovett this should do the job

PR #1578, opened 2018-06-18
fix 'Cannot read property 'posthtml' of null

### LINKED ISSUE #1574: .../index.html: Cannot read property 'posthtml' of null
# 🐛 bug report

```
$ parcel serve index.html
🚨  [redacted]/index.html: Cannot read property 'posthtml' of null
    at getConfig (/Users/stevebennett/npm/lib/node_modules/parcel/src/transforms/posthtml.js:29:32)
```

The error seems to be caused by the lack of package.json.

Directory containing only index.html and index.js.

## 🤔 Expected Behavior

A friendly error message ("Parcel requires a package.json in order to build.")

## 😯 Current Behavior

Stack trace above.

## 🔦 Context

This is a very unwelcoming start to using Parcel, sort of contradicts the "It just works!" messaging.

ranfdev:
fixes #1574

### DIFF
--- a/src/transforms/posthtml.js
+++ b/src/transforms/posthtml.js
@@ -26,7 +26,7 @@ async function transform(asset) {
 
 async function getConfig(asset) {
   let config =
-    (await asset.getPackage()).posthtml ||
+    ((await asset.getPackage()) || {}).posthtml ||
     (await asset.getConfig([
       '.posthtmlrc',
       '.posthtmlrc.js',


src/transforms/posthtml.js
```diff
@@ -26,7 +26,7 @@ async function transform(asset) {
 
 async function getConfig(asset) {
   let config =
-    (await asset.getPackage()).posthtml ||
+    ((await asset.getPackage()) || {}).posthtml ||
```
DeMoorJasper: This is not the correct fix, there is an option you can pass to asset.getConfig, that already prevents this behaviour. Better to use this than to reimplement it in every place.

ranfdev: You mean: js```await asset.getConfig([ 
   '.posthtmlrc', 
   '.posthtmlrc.js', 
   'posthtml.config.js' 
], {
  'packageKey': 'posthtml'
})```? I tried, but this doesn't work. 
The problem is not in asset.getConfig(). The error comes from line 29:32, this line

DeMoorJasper: Yes I mean that, the original asset.getPackage becomes obsolete when using getConfig options. You can always check out how asset.getConfig works to understand

ranfdev: Ok, thanks!

ranfdev:
Now it should be ok, the error from AppVeyor is not related to this

DeMoorJasper:
Windows seems to be instable since tree shaking

Issue #809, opened 2018-02-14
[Bug] HMR and PostCSS modules, styles disappear


pavel:
🐛 bug report

### 🎛 Configuration

.babelrc
```json
{
	"presets": [
		"react",
		["env", {
			"targets": {
				"browsers": [ "last 2 versions" ]
			},
			"debug": true
		}]
	]
}
```

.postcssrc.js
```js
module.exports = {
	modules: true,
	plugins: {
		"autoprefixer": {}
	}
};
```

browserslist
```
last 2 versions
```

index.html
```html
<!doctype html>
<html>
	<head>
		<meta charset="utf-8">
		<meta name="viewport" content="width=device-width, initial-scale=1">
		<title>React App</title>
	</head>
	<body>
		<div id="app"></div>

		<script src="./index.js"></script>
	</body>
</html>
```

index.js
```js
require("babel-polyfill");

const styles = require("./app.styl");

console.log("styles???", styles);

const React = require("react");
const ReactDOM = require("react-dom");

const App = () => (
	<div className={ styles.app }>Hello!!</div>
);

ReactDOM.render(<App/>, document.getElementById("app"));
```

app.styl
```styl
.app
	color: blue
	transform: scale(0.95)
	transition: transform 0.4s ease
```

CLI: `./node_modules/.bin/parcel index.html`

### 🤔 Expected Behavior

Changing `app.styl` should update styles properly.

### 😯 Current Behavior

Styles disappear after change to `app.styl`.
`index.js` (which requires `app.styl`) is not reloaded when the `app.styl` file is changed.

### 💻 Code Sample

Source code: 
[parcel-postcss-modules-bug.tar.gz](https://github.com/parcel-bundler/parcel/files/1725533/parcel-postcss-modules-bug.tar.gz)

### 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Parcel           | 1.5.1 and `master` at 6deec80b3491dc5ac690da6550323b51deec6530
| Node             | 9.5.0
| npm         | 5.6.0
| Operating System | Linux x86_64

yiooxir:
Have the same issue

joakin:
I think I've hit this issue with a typescript + css modules repo. See https://github.com/joakin/parcel-bundler-test/ for minimal example.

Here's a gif that shows the broken behavior:

[...]

PR #1589, opened 2018-06-20
Prevent postcss-modules plugin config from being deleted after first run

maxholman:
When the postcss config object is being copied with `Object.assign`, the nested `plugins` array or object remains as a reference and not a copy.

As the postcss transform deletes the plugins config for `postcss-modules` each time, it inadvertently deletes the original config by reference.

As a result, postcss transforms that occur after the first time have no `postcss-modules` plugins config and therefore parcel produces a different result for all subsequent transforms.

Related: #809

### DIFF
--- a/src/transforms/postcss.js
+++ b/src/transforms/postcss.js
@@ -28,7 +28,15 @@ async function getConfig(asset) {
     return;
   }
 
-  config = Object.assign({}, config);
+  config = Object.assign(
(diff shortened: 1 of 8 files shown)


devongovett:
Simplified slightly to always clone config objects that we load with `require`, since they are cached. Should solve this situation and others.

PR #1595, opened 2018-06-21
Use PARCEL_MAX_CONCURRENT_CALLS environment variable

albinotonnina:
I had an issue with Parcel freezing during bundling.
@DeMoorJasper was so kind to help debugging the issue.
Turns out lowering the amount of concurrent calls per worker solves my issue.

This PR let users set that value with an env variable.

Thanks

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -61,6 +61,8 @@ You can set `PARCEL_WORKERS` to the number of worker processes to spawn.
 
 **NOTE:** When developing plugins or new Asset types, run with `--no-cache` (or pass `cache: false` to `Bundler` options). Parcel uses caching by default, but during development you'll normally pass incomplete results into the cache. This can leave you wondering why you're constantly seeing old results.
 
+You can set `PARCEL_MAX_CONCURRENT_CALLS` to change the limit of concurrent calls per worker.
+
 ## Link to local parcel build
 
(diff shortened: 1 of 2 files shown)


KeineLimonade:
Still the default value of 10 should be used if no environment variable is set.

albinotonnina:
@KeineLimonade of course! silly me, fixed

devongovett:
do we know what is causing the freeze? I'd rather fix that than make everyone configure this.

DeMoorJasper:
This seems to happen only on projects that are fairly large and use scss, So my guess is it's scss related.

Although I haven't dug very deep into why exactly it is happening, in the example @albinotonnina gave me I could eliminate the freeze by removing all scss imports.

I've done some debugging on the workerfarm and processing 10 assets at the same time makes some assets take q very long time to process. so I think a lower default would be a better idea, although it would probably have very little performance effect (besides from putting less stress on the workers)

albinotonnina:
It would be an advanced setting, like the other env variable `PARCEL_WORKERS`.
I guess most people will rarely use it but it could be a good un-blocker while looking for a solution in cases such as mine. It surely makes me continue working!
Btw thanks again @DeMoorJasper. @devongovett you should take him with you at Adobe! :P

Hammster:
@DeMoorJasper i can confirm that after removing all scss in my example it builds again too. https://github.com/Hammster/parcel-1.9.x-freeze

And @albinotonnina reasoning is totally valid, we are stuck using 1.7.1 because the build freezez, having this option would at least allow us to use the latest version.

Issue #1451, opened 2018-05-28
Parcel not tracking for changes on scss files


natqe:
# 🐛 bug report

## 🎛 Configuration (.babelrc, package.json, cli command)
package.json:
```Json
{
  "name": "any",
  "version": "1.0.0",
  "description": "any",
  "main": "./app/index.htm",
  "scripts": {
    "start": "parcel app/index.htm  -p 506",
    "https": "parcel app/index.htm --no-cache --https -p 506",
    "watch": "parcel watch app/index.htm --public-url ./",
    "build": "parcel build app/index.htm --no-cache --public-url ./",
    "test": "parcel watch app/index.scss --no-cache"
  },
  "repository": {
    "type": "git",
    "url": "git+https://any@bitbucket.org/any/any.git"
  },
  "author": "natan farkash",
  "license": "ISC",
  "homepage": "https://bitbucket.org/any/any#readme",
  "devDependencies": {
    "autoprefixer": "^8.5.0",
    "node-sass": "^4.9.0",
    "parcel-bundler": "git+https://github.com/parcel-bundler/parcel.git",
    "postcss-modules": "^1.1.0",
    "typescript": "^2.8.3"
  },
  "dependencies": {
    "angular": "^1.7.0",
    "angular-animate": "^1.7.0",
    "angular-aria": "^1.7.0",
    "angular-material": "^1.1.9",
    "angular-messages": "^1.7.0",
    "angular-route": "^1.7.0",
    "element-closest": "^2.0.2",
    "lodash": "^4.17.10"
  }
}
```

## 🤔 Expected Behavior

Tracking my files for changes and updating the bundle.css

## 😯 Current Behavior

I'm need to change something in index.scss for parcel refresh the bundle with the new code from the partials, and without that is not recognize my changes in _partial.scss

## 💻 Code Sample
//index.scss
@import 'partials/selector';

project example:
https://github.com/natqe/parcel_bug_watcher_sass_example

## 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Parcel           | 1.0.9
| Node             | 10.3.0
| npm         | 6.0.1
| Windows 10 |

DeMoorJasper:
Could you please provide with a test repo?
There is an open issue already related to this issue. But I'd like to have a test repo to see why this happens.

natqe:
@DeMoorJasper 
I sanded to your mail an example, i hope you got it.

DeMoorJasper:
[...]

PR #1602, opened 2018-06-22
Replaced fwd slashes with backslashes for win to fix sass deps watch

lysla:
Running `parcel watch` on Windows wouldn't re-build scss files when only partial sass dependencies, imported in the main .scss file, were modified. The reason is Windows uses forward slashes `/` in the path of these dependencies. #1451


src/assets/SASSAsset.js
```diff
@@ -65,6 +65,9 @@ class SASSAsset extends Asset {
 
   collectDependencies() {
     for (let dep of this.ast.stats.includedFiles) {
+      if (os.platform() === 'win32') {
```
DeMoorJasper: What about moving this to:
https://github.com/parcel-bundler/parcel/blob/master/src/Bundler.js#L394
and
https://github.com/parcel-bundler/parcel/blob/master/src/Bundler.js#L704

This way this applies to every asset type, not just scss

devongovett: maybe you could use `path.normalize` for this https://nodejs.org/api/path.html#path_path_normalize_path

lysla: I just did another commit, moving the fix in Bundler.js like suggested and using path.normalize. I'm not sure I put it in the right positions, but as far as I could test it, it works. Thanks for your efforts!

devongovett:
actually I believe that since we are calling `fs.realpath` on the path now prior to watching it, that will cause it to be normalized already.

lysla:
Thank you for merging @devongovett keep up the awesome work.

PR #1603, opened 2018-06-22
Refactor normalizeOptions

albinotonnina:
Hello!

I think this cleans up a bit `Bundler.js` and `normalizeOptions()`. 
I've added a few tests.


src/utils/normalizeOptions.js
```diff
@@ -0,0 +1,56 @@
+const Path = require('path');
+const getRootDir = require('./getRootDir');
+
+function normalizeOptions(options = {}, entryFiles = []) {
+  const {
```
DeMoorJasper: This seperate options object, seems a bit overkill to me, most of these options could probably be normalised in the return statement as how it was done in the original code.

albinotonnina: Ciao @DeMoorJasper !
Overkill in terms of performance or lines of code?
I had that `{options}` with default values separated from the returned obj mostly for readability. I found separating calculations and defaultValues an improvement in DX, making both of them a bit more explicit. The cost of it is slightly more lines but the advantage is a more flat structure IMO.
The options are quite a lot and some of them are purely derived, some depends on other values and some of them have defaults. This PR intends to highlight that and make stuff easier to scan.

albinotonnina: Visually speaking:

image

albinotonnina: I could reduce the amount of code by doing something like

```javascript

return Object.assign(options, {
production: isProduction,
cacheDir: Path.resolve(cacheDir),
sourceMaps: sourceMaps && !scopeHoist
...
...
})

```

It would save like a good dozen line but it could penalize a bit readability.

DeMoorJasper: Overkill in complexity, I don't think performance would really matter here as it only happens once every time the command is being ran

RafalFilipek: I think we can make it more readable just by introducing simple `getBool` helper :)

```
function normalizeOptions(options = {}, entryFiles = []) {
  const isProduction = process.env.NODE_ENV === 'production';

  const getBool = (value, fallback) => typeof value === 'boolean' ? value : fallback;
  return {
    production: isProduction,
    outDir: Path.resolve(options.outDir || 'dist'),
    outFile: options.outFile || '',
    publicURL: options.publicUrl || options.publicURL || '/',
    watch: !isProduction,
    cache: getBool(options.cache, true),
    cacheDir: Path.resolve(options.cacheDir || '.cache'),
    killWorkers: getBool(options.killWorkers, true),
    minify: isProduction,
    target: options.buildTarget || 'browser',
    hmr: options.buildTarget === 'node' ? false : options.hmr,
    hmrPort: options.hmrPort || 0,
    hmrHostname: options.buildTarget === 'electron' ? 'localhost' : '',
    https: getBool(options.https, false),
    logLevel: isNaN(options.logLevel) ? 3 :options.logLevel,
    entryFiles,
    rootDir: getRootDir(entryFiles),
    sourceMaps: options.sourceMaps && !options.scopeHoist,
    detailedReport: getBool(options.detailedReport, false),
    global: options.global,
    autoinstall: getBool(options.autoinstall, !isProduction),
    scopeHoist: getBool(options.scopeHoist, false),
    contentHash: getBool(options.contentHash, isProduction)
  }
}
```

What you think?

albinotonnina: Hello @RafalFilipek!
Thanks! With `getBool` the code looks better already although, and this is my opinion, the code still look a bit too 'horizontal' to me.
what do you think about splitting it in a different module? For sure it's easier to test it.

RafalFilipek: Probably `utils/getBool.js` :) There are few places where `typeof foo === 'boolean' is used so might be helpful. We can also write this function in more readable form:

```
function getBool (value, fallback) {
  return typeof value === 'boolean' ? value : fallback;
} 
```

devongovett: I like the idea of splitting `normalizeOptions` out into a separate module for testing. I think the destructuring assignment here makes it a bit hard to read though - I had to use the babel repl to look at what it compiles to to understand it. Also, for the boolean options, it doesn't check if they are actually a boolean, just if it is undefined.

I like @RafalFilipek's idea of a `getBool` helper - maybe we can just put it in the same file for now.

albinotonnina:
Ok people! I don’t think this PR is a priority so if I do the changes in a couple of weeks when I’m back from holidays it should be fine! Right?

devongovett:
all good!

devongovett:

Issue #1613, opened 2018-06-25
When includePaths is a relative path, SASS dependencies can not be watched.


2zH:
# 🐛 bug report

When includePaths is a relative path, SASS dependencies can not be watched.

## 🎛 Configuration (.sassrc)

```json
{
  "includePaths": [
    "node_modules"
  ],
}
```

## 🤔 Expected Behavior

Work fine.

## 😯 Current Behavior

It will be locked in bundling code. Tracing it, I got an endless loops.
https://github.com/parcel-bundler/parcel/blob/master/src/Watcher.js#L31

```javascript
  getWatchedParent(path) { // path = 'node_modules/xxx/xxx.scss'
    path = Path.dirname(path); // path = 'node_modules/xxx'

    let root = Path.parse(path).root; // root = ''
    while (path !== root) {
      if (this.watchedDirectories.has(path)) {
        return path;
      }

      path = Path.dirname(path); // At last time, path = '.', It will never be equal to ''.
    }

    return null;
  }
```

## 💁 Possible Solution
https://github.com/parcel-bundler/parcel/blob/master/src/assets/SASSAsset.js#L27
```javascript
  async parse(code) {
    // ...
    opts.includePaths = (
        opts.includePaths
            ? opts.includePaths.map(path => Path.resolve(path))
            : []
    ).concat(
      path.dirname(this.name)
    );
    // ...
  }
```

## 🔦 Context

```sh
yarn parcel index.html --no-cache
```
```
  ┌── src
  │     └── index.js
  ├── index.html
  ├── node_modules
  │     └── foo
  │          ├── src
  │          │        └── foo.scss
  │          ├── .sassrc
  │          ├── node_modules
  │          │                └── bar
  │          │                           ├── bar.scss
  │          │                           └── package.json
  │          └── package.json
  └── package.json
```
```javascript
// index.js
import 'foo/src/foo.scss'
// ...
```
```sass
// node_modules/foo/src/foo.scss
@import 'bar/bar.scss'
// ...
```

## 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Parcel           | 1.9.2 |
| Node             | 10.5.0 |
| npm/Yarn         | Yarn |
| Operating System | Arch Linux |

devongovett:
Fixed by #1619.

PR #1619, opened 2018-06-26
Fix Sass dependencies can not be watched when includePaths is a relative path

2zH:
For #1613 
I have reviewed the contributing guidelines and glad to give a pr to fix it.
Hope this is a suitable solution.

### DIFF
--- a/src/assets/SASSAsset.js
+++ b/src/assets/SASSAsset.js
@@ -24,9 +24,10 @@ class SASSAsset extends Asset {
     let opts =
       (await this.getConfig(['.sassrc', '.sassrc.js'], {packageKey: 'sass'})) ||
       {};
-    opts.includePaths = (opts.includePaths || []).concat(
-      path.dirname(this.name)
-    );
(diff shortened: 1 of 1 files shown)


src/assets/SASSAsset.js
```diff
@@ -24,9 +24,10 @@ class SASSAsset extends Asset {
     let opts =
       (await this.getConfig(['.sassrc', '.sassrc.js'], {packageKey: 'sass'})) ||
       {};
-    opts.includePaths = (opts.includePaths || []).concat(
-      path.dirname(this.name)
-    );
+    opts.includePaths = (opts.includePaths
+      ? opts.includePaths.map(includePath => path.resolve(includePath))
```
devongovett: this resolves paths relative to the current working directory. is this right, or do we want to resolve relative to the sass file being compiled?

2zH: Yes, only reative to the current working directory will not change original behavior. It seems that the sass compiler has the same logic for reative path in `includePaths`.

PR #1620, opened 2018-06-26
feat: allow overriding logger output stream

alexlafroscia:
I'm writing a VSCode plugin that wraps Parcel, and it would be useful to be able to receive all of the logging from Parcel programmatically rather than allowing it to go to `stdout` of the VSCode plugin. 

This should allow the output stream to optionally be "dependency injected" when creating a new `Bundler` with something I can define and turn into useful messages within VSCode.


alexlafroscia:
I'm not sure why that test is failing... it was failing locally when I ran the tests against `master`, too

KeineLimonade:
Tests are failing for most PRs. That is most likely not related to your changes.

I think this is a seemingly good addition. It is non intrusive if no output stream is provided and it is totally backwards compatible.

alexlafroscia:
Thanks! I did notice that I need to whitelist the property when normalizing the options passed to the Bundler — I’ll update my PR tomorrow. 

devongovett:
One potential issue here is that the logger is currently a singleton. So if there were multiple bundler instances at once in your app, each with different logger output streams, this would not work. Perhaps not a huge deal for most uses, but just thought I'd point it out.

peanutbother:
@devongovett what if we would implement a mechanism which defaults to the singleton but creates new instance of logger, if specified?

pseudo:

logger => singleton 
+ getInstance(id = defaultID)
+ createNew(id = instances.length < 1 ? defaultID : randomID )

peanutbother:
another approach could be implementing something like log channels where parcel uses one logger(the singleton) instance and based on its process-id you could observe logger and grab every message adressed to chalnnel #`proccess.id`

devongovett:

PR #1641, opened 2018-06-29
Fix sourcemap file size in report

### LINKED ISSUE #1635: --detailed-report always reports 0 bytes for size of sourcemap .map files
Somewhat similar to #1151, when I'm building a bundle and ask for a `--detailed-report`, I'm (erroneously) told that the generated sourcemap files were 0 bytes:

```
> filer@0.0.29 release /Users/humphd/repos/filer
> parcel build --global Filer src/index.js --out-file filer.min.js --detailed-report

✨  Built in 308ms.

dist/filer.min.js                                                     70.15 KB    79ms
├── node_modules/node-libs-browser/node_modules/buffer/index.js       19.52 KB    41ms
├── src/filesystem/implementation.js                                  17.96 KB    39ms
├── lib/intercom.js                                                    3.75 KB    39ms
├── src/filesystem/interface.js                                        3.12 KB     7ms
├── src/shell/shell.js                                                 3.06 KB    38ms
├── src/providers/websql.js                                            2.46 KB    12ms
├── lib/nodash.js                                                      2.33 KB    37ms
├── src/providers/indexeddb.js                                         1.95 KB    12ms
├── src/path.js                                                        1.69 KB     9ms
├── node_modules/node-libs-browser/node_modules/process/browser.js     1.68 KB     3ms
└── + 21 more assets

dist/filer.min.map                                                         0 B    73ms
~/repos/filer (parceljs *) $ ls -lah dist
total 2096
drwxr-xr-x   7 humphd  staff   224B 28 Jun 11:18 .
drwxr-xr-x  22 humphd  staff   704B 28 Jun 11:07 ..
-rw-r--r--   1 humphd  staff    70K 28 Jun 11:18 filer.min.js
-rw-r--r--   1 humphd  staff   278K 28 Jun 11:18 filer.min.map
``` 

The actual file on disk is `278K`.  It would be nice to know this, even if only so you didn't wrongly assume that that sourcemap didn't generate properly.

DeMoorJasper:
Fixes #1635

### DIFF
--- a/src/Bundle.js
+++ b/src/Bundle.js
@@ -196,6 +196,8 @@ class Bundle {
 
     await packager.end();
 
+    this.totalSize = packager.getSize();
+
     let assetArray = Array.from(this.assets);
     let assetStartTime =
       this.type === 'map'
--- a/src/packagers/JSConcatPackager.js
+++ b/src/packagers/JSConcatPackager.js
@@ -505,7 +505,7 @@ class JSConcatPackager extends Packager {
       output = '(function () {' + output + '})();';
     }
 
-    this.bundle.totalSize = output.length;
+    this.size = output.length;
 
     let {sourceMaps} = this.options;
     if (sourceMaps) {


src/Bundle.js
```diff
@@ -196,6 +196,8 @@ class Bundle {
 
     await packager.end();
 
+    this.totalSize = packager.getSize();
```
DeMoorJasper: This line updates `bundle.totalSize` to the actual final amount of written bytes instead of relying on `addAsset` to be called for every packager

devongovett: I think we will need to adjust the packager for tree shaking here or we'll get incorrect sizes:

https://github.com/parcel-bundler/parcel/blob/3a8f38cc0d8b5d71d158eeb7c7526e01be746c28/src/packagers/JSConcatPackager.js#L508

Can you change that to `this.size = output.length`? That way when we call `getSize` here after the bundle has finished we'll get the correct size.

DeMoorJasper: Fixed it in the latest commit

Issue #1632, opened 2018-06-27
🐛 Object spread is not working 


mohsen1:
# 🐛 Object spread (`...obj`) is failing to work with basic setup 

## 🤔 Expected Behavior

No error

## 😯 Current Behavior
Parcel `build` **hangs** with this syntax error 

```
yarn run v1.7.0
$ parcel build index.html
🚨  Desktop/parcel-spread-bug/src/components/SplitPane.js:74:8: Unexpected token (74:8)
  72 |       // top
  73 |       styleA = {
> 74 |         ...baseStyleVertical,
     |         ^
  75 |         top: 0,
  76 |         height: dividerPos + '%',
  77 |         paddingBottom: 3,
````

## 🔦 Context

Faced this bug here 
https://github.com/fkling/astexplorer/pull/330

## 💻 Code Sample

Repo to reproduce 

https://github.com/mohsen1/parcel-spread-bug

**Comment by mohsen1** [comment]
Interesting: 

Removing Google Analytics inline script will resolve the hanging issue 

https://github.com/mohsen1/parcel-spread-bug/pull/1

**Comment by mohsen1** [comment]
I reduced this bug to simply:

**index.html**
```html
<!DOCTYPE html>
<html lang="en">

<head>
</head>

<body>
    <script src="./index.js"></script>
</body>

</html>
```

**index.js**
```js
var a = {
  foo: 1
};

var b = {
  ...a,
  bar: 2
};
```

**`yarn build index.html`** prints:

```
$ parcel build index.html
🚨  /parcel-spread-bug/index.js:6:2: Unexpected token (6:2)
  4 |
  5 | var b = {
> 6 |   ...a,
    |   ^
  7 |   bar: 2
  8 | };
  9 |
```

tripodsan:
note, it works with `--target node`

buronnie:
Parcel out of box includes `babel-preset-env` and `babel-preset-react`. The object rest spread transform is not part of either. So you need to install `babel-plugin-transform-object-rest-spread` and define it in your `.babelrc` file.

mohsen1:
That seems to work:

https://github.com/mohsen1/parcel-spread-bug/pull/2

I wonder why this works in node target though?

DeMoorJasper:
Owkeey, so this is not a bug, closing this.

texastoland:
[...]

PR #1643, opened 2018-06-30
fix build not exiting in dev env

### LINKED ISSUE #1626: The process does not end after the completion of the custom environment variable packaging.
image
The environment variable is configured to define the interface address
```
let baseUrl = 'http://test-xxx.com';

if(process.env.NODE_ENV === 'production') baseUrl = 'http://xxx.com';
```
But after packaging , the process does not end;

Although my environment variable is not production, my command is build, so should end the process.

My project was deployed online through shell scripts, but your process was not closed, and I could not tell if the packaging was finished.

Hope to solve it quickly.

gorangajic:
When NODE_ENV is set to something else instead of `production` build was
not exiting. This bug was introduced in 1b6a93f3efa1d8a4e9e04beda1a5545770e9fb07

Closes #1626

### DIFF
--- a/src/cli.js
+++ b/src/cli.js
@@ -177,7 +177,7 @@ async function bundle(main, command) {
   const Bundler = require('../');
 
   if (command.name() === 'build') {
-    command.watch = false;
+    command.production = true;
     process.env.NODE_ENV = process.env.NODE_ENV || 'production';
   } else {
     process.env.NODE_ENV = process.env.NODE_ENV || 'development';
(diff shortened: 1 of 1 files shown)


src/cli.js
```diff
@@ -177,6 +177,7 @@ async function bundle(main, command) {
   const Bundler = require('../');
 
   if (command.name() === 'build') {
+    command.watch = false;
```
devongovett: I think we should set `command.production = true` here actually, which would implicitly disable watching as well. I believe #1600 was meant to keep environment variables as set by the user, not control actual build. That can be done with other command line flags, e.g. if you want to disable minification, use `--no-minify`.

mohsen1:
FYI 
The hang did happen for me when running `parcel build` here: https://github.com/parcel-bundler/parcel/issues/1632

PR #1650, opened 2018-07-02
Use fast glob

DeMoorJasper:
Adds a performance boost to the usage of globs


src/utils/glob.js
```diff
@@ -0,0 +1,9 @@
+const isGlob = require('is-glob');
+
+exports.isGlob = function(filename) {
+  return isGlob(filename, {
+    strict: true
```
devongovett: looks like strict is the default: https://github.com/micromatch/is-glob/blob/master/index.js#L24

DeMoorJasper:
Removed the util

PR #1665, opened 2018-07-04
Only watch directories on macOS

### LINKED ISSUE #1557: Why is node_modules watched?

devongovett:
Fixes #1557.
Closes #1559.

We switched to watching directories instead of files in #1279 because of EMFILE errors (running out of file descriptors) when watching individual files. This was especially seen on macOS. However, it seems that watching directories is not supported on Linux, so Chokidar is recursively watching all individual files in node_modules. This is likely more files than just the ones we were watching individually before, since it will include things in node_modules that aren't in the build (e.g. tests, etc.).

This PR changes the behavior of the watcher to only watch directories on macOS, where Chokidar has special support via fs-events. Long term, we should consider moving off of Chokidar but this should "solve" the problem temporarily.

### DIFF
--- a/src/Watcher.js
+++ b/src/Watcher.js
@@ -2,14 +2,17 @@ const FSWatcher = require('fswatcher-child');
 const Path = require('path');
 
 /**
- * This watcher wraps chokidar so that we watch directories rather than individual files.
+ * This watcher wraps chokidar so that we watch directories rather than individual files on macOS.
  * This prevents us from hitting EMFILE errors when running out of file descriptors.
+ * Chokidar does not have support for watching directories on non-macOS platforms, so we disable
+ * this behavior in order to prevent watching more individual files than necessary (e.g. node_modules).
  */
 class Watcher {
   constructor() {
     // FS events on macOS are flakey in the tests, which write lots of files very quickly
     // See https://github.com/paulmillr/chokidar/issues/612
-    this.shouldWatchDirs = process.env.NODE_ENV !== 'test';
+    this.shouldWatchDirs =
+      process.platform === 'darwin' && process.env.NODE_ENV !== 'test';
     this.watcher = new FSWatcher({
       useFsEvents: this.shouldWatchDirs,
       ignoreInitial: true,


cpotemski:
Sorry for reopen, but i still have some problems with the watcher.

I'm runnig a small new project with parcel and it is a pleasure compared to the webpack config.

My problem is, that i run my project inside docker on the default node:10 docker image.
If the watcher runs inside the container, the CPU of my 1 year old MacBook Pro runs at 50-60% in idle mode (without installing or changing files).

If I change the `ignored` option of `src/Watcher.js:20` to ignore `node_modules` with
```
ignored: /\.cache|\.git|node_modules/
```
the container runs silent at 1-2% CPU.

So it would be nice if you could help me to fix that problem or give us the ability to override that default behavior (like configure `ignored` or give us a environment variable).

I would like to add a PR if we need to change something.

philkunz:
My CPU also goes crazy which led me to implement my own filewatcher and server that then runs parcel build when needed.

PR #1669, opened 2018-07-05
Avoid createBundleTree if same asset

terrierscript:
Thank you for great product.
I checkout master branch, and got Maximum call stack size exceeded in master branch when below code.

```
window.Buffer = require("buffer/").Buffer
```

```
🚨  Maximum call stack size exceeded
    at Bundler.createBundleTree (/***/src/Bundler.js:583:19)
    at Bundler.createBundleTree (/***/src/Bundler.js:652:12)
    at Bundler.createBundleTree (/***/src/Bundler.js:652:12)
    at Bundler.createBundleTree (/***/src/Bundler.js:652:12)
    at Bundler.createBundleTree (/***/src/Bundler.js:652:12)
    at Bundler.createBundleTree (/***/src/Bundler.js:652:12)
    at Bundler.createBundleTree (/***/src/Bundler.js:652:12)
    at Bundler.createBundleTree (/***/src/Bundler.js:652:12)
    at Bundler.createBundleTree (/***/src/Bundler.js:652:12)
    at Bundler.createBundleTree (/***/src/Bundler.js:652:12)
```

If my approach is not good, please close this PR.

### DIFF
--- a/src/Bundler.js
+++ b/src/Bundler.js
@@ -648,7 +648,9 @@ class Bundler extends EventEmitter {
     parentBundles.add(bundle);
 
     for (let [dep, assetDep] of asset.depAssets) {
-      this.createBundleTree(assetDep, bundle, dep, parentBundles);
+      if (asset.name !== assetDep.name) {
+        this.createBundleTree(assetDep, bundle, dep, parentBundles);
+      }
     }
 
     parentBundles.delete(bundle);
--- a/src/Bundler.js
+++ b/src/Bundler.js
@@ -648,7 +648,7 @@ class Bundler extends EventEmitter {
     parentBundles.add(bundle);
 
     for (let [dep, assetDep] of asset.depAssets) {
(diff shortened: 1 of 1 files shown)


devongovett:
hmm I cannot reproduce this error... is there any other code that might be causing it? can you provide a minimal test case that reproduces it?

terrierscript:
I can make little project.
https://github.com/terrierscript/parcel-call-stack-bug

This bug occur when `new Worker`.

```index.js
// index.js
const worker = new Worker("./worker.js");
```

```js
// worker.js
window.Buffer = require("buffer/").Buffer;
```

devongovett:
Ah, thanks for finding this bug. It is specific to workers (or really, isolated bundles in general, where assets may be duplicated into multiple bundles). I fixed it in a slightly different way in f2deb5c, which fixes indirect circular deps as well.

PR #1587, opened 2018-06-20
Workerfarm improvements

DeMoorJasper:
* Passes bundler options to workerfarm
* Persists these bundler options to every worker
* Moved all previous initRemote code to worker
* Improved workerfarm end function (to 100% sure kill all processes that are idle)
* Less concurrent workerfarm calls for faster execution (due to less freezing and overloading)
* Don't use the workerfarm on single-threaded machines
* Assign the custom worker functions to the process

Closes #1548 Closes #1286 Closes #1319

PR #1671, opened 2018-07-06
fix a case where null options might cause issues

Jimexist:
otherwise when using 3rd party plugin this might cause issue with empty options

### DIFF
--- a/src/worker.js
+++ b/src/worker.js
@@ -10,7 +10,10 @@ function setChildReference(childReference) {
 }
 
 function init(options) {
-  pipeline = new Pipeline(options || {});
+  if (!options) {
+    options = {};
+  }
+  pipeline = new Pipeline(options);
   Object.assign(process.env, options.env || {}, {
     PARCEL_WORKER_TYPE: child ? 'remote-worker' : 'local-worker'
   });
--- a/src/worker.js
+++ b/src/worker.js
@@ -17,8 +17,12 @@ function init(options) {
   Object.assign(process.env, options.env || {}, {
     PARCEL_WORKER_TYPE: child ? 'remote-worker' : 'local-worker'
   });
-  process.env.HMR_PORT = options.hmrPort;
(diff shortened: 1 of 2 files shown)


DeMoorJasper:
Options should always be defined this isn’t gonna fix the core issue of this bug

pke:
well, options seems to be undefined for some reason though.

pke:
No haven't tried that yet. Would have to check out this PR branch and build the bundler myself?

DeMoorJasper:
@pke yes

Jimexist:
@DeMoorJasper i understand your concern. will #1587 be merged soon?

devongovett:
@DeMoorJasper it's possible there could be a race here: https://github.com/parcel-bundler/parcel/blob/master/src/worker.js#L35. That's the only place we call `WorkerFarm.getShared` and don't pass any options. If the shared instance hasn't been created yet, it will fail.

devongovett:
@Jimexist did #1587 work for you?

DeMoorJasper:
@devongovett The worker always gets called from the workerfarm so once the worker reaches https://github.com/parcel-bundler/parcel/blob/master/src/worker.js#L35 it should already be set right?

Unless it gets called from a remote worker in that case that shouldn't even be happening.

#1587 should fix most of the worker dying and not being properly initialised issue, so there's a big chance it fixes this. But I have no clue as this is an issue I have never been able to reproduce. I have been able to reproduce similar errors and those have been fixed by the PR as mentioned in the PR.

DeMoorJasper:
@devongovett Ow right that's probably the reason. Totally didn't know that addCall is being called outside of the asset processing workers.

Jimexist:
if this PR is no longer relevant, should i close it now?

devongovett:
Yeah, going to close it. We can tackle the fix for the problem described earlier in a separate PR, or part of #1587.

PR #1672, opened 2018-07-06
adding support for node 10?

Jimexist:
Do we want to add node version 10 support?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ language: node_js
 node_js:
   - '6'
   - '8'
+  - '10'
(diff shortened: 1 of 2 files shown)


Jimexist:
i don't think the test failures are due to my changes. any way to stabilize the tests?

Jimexist:
@devongovett done!

Jimexist:
still don't know why the CI fails

devongovett:
Yeah it's a bit flakey. Seems like some of the rust tests are failing currently.

PR #1674, opened 2018-07-07
Watch real path of case-insensitive files

### LINKED ISSUE #1673: Some files are not watched in case-insensitive filesystems
# 🐛 bug report

I am trying out parcel bundler and with help from other people online have produced the following repository:
https://github.com/markwylde/parcel-react-demo

It's a working one page example app that builds successfully using react.

However when I do `parcel watch ./src/index.html` or `parcel ./src/index.html` and then change a file, it's not actually rebuilding anything.

This was just one of a few other problems I've been having. In general parcel just seems very buggy and is throwing a lot of errors. But I will submit that it could just be react specific stuff, or the fact I have never used parcel before.

I have uploaded a video of my experience to youtube.

[video of my experience](https://youtu.be/XzHsgyaJkZk)

`.babelrc`:
```
{
  "presets": ["react"]
}
```

`package.json`:
```
{
  "name": "egr",
  "version": "1.0.0",
  "description": "",
  "main": "index.js",
  "scripts": {
    "test": "echo \"Error: no test specified\" && exit 1"
  },
  "keywords": [],
  "author": "",
  "license": "ISC",
  "dependencies": {
    "node-sass": "^4.9.1",
    "react": "^16.4.1",
    "react-dom": "^16.4.1"
  },
  "devDependencies": {
    "babel-core": "^6.26.3",
    "babel-preset-react": "^6.24.1"
  }
}
```

## 🤔 Expected Behavior

I expect the dist folder to be updated when I change a source file.

## 😯 Current Behavior

When I change a source file the dist folder is not updating.

## 💁 Possible Solution

No idea, I've only just started looking into parcel today.

## 🔦 Context

With 7 second build times it would be impossible to use this for an actual project.

## 💻 Code Sample

https://github.com/markwylde/parcel-react-demo

## 🌍 Your Environment

| Parcel           | 1.9.4 |
| Node             | v10.3.0 |
| npm/Yarn         | 6.1.0 |
| Operating System | macOS High Sierre |

fathyb:
Fixes #1673 

Not sure how to test this though

### DIFF
--- a/src/Bundler.js
+++ b/src/Bundler.js
@@ -20,6 +20,7 @@ const installPackage = require('./utils/installPackage');
 const bundleReport = require('./utils/bundleReport');
 const prettifyTime = require('./utils/prettifyTime');
 const getRootDir = require('./utils/getRootDir');
+const trueCasePath = require('./utils/trueCasePath');
 const glob = require('fast-glob');
 
 /**
@@ -402,6 +403,7 @@ class Bundler extends EventEmitter {
     }
 
     path = await fs.realpath(path);
+    path = await trueCasePath(path);
 
     if (!this.watchedAssets.has(path)) {
       this.watcher.watch(path);
--- a/src/utils/trueCasePath.js
+++ b/src/utils/trueCasePath.js
@@ -1,0 +1,46 @@
+const glob = require('glob');
+const path = require('path');
+
+// Async version of https://github.com/barsh/true-case-path
+function trueCase(fsPath) {
+  // Normalize the path so as to resolve . and .. components.
+  // !! As of Node v4.1.1, a path starting with ../ is NOT resolved relative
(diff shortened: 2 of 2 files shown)


devongovett:
hmm do we really want to support this? IMO supporting case-insensitivity just makes code non-portable across file systems. e.g. you build an app on OS X, and use incorrect casing, but then it doesn't build on Linux. Also a bit worried about performance of running a glob on every file...

devongovett:
> I totally agree on this, but we already support case-insensitivity

Not really, we just pass through exactly what was given to us to the filesystem and it happens to work.

I'm not even sure how we would detect that we need to display a warning besides actually checking if every filename is correct, via glob or e.g. calling `readdir` on each directory.

fathyb:
Got you, closing this then.

PR #1675, opened 2018-07-07
Watch PostHTML plugins dependencies

### LINKED ISSUE #1670: Changes to files used by PostHTML plugins do not trigger rebuild
# ❔ Question

When I use the `posthtml-extend` plugin to have my `index.html` "extend" a `base.html` file, it works on first build, but changes to the `base.html` file don't trigger a rebuild. It also seems that `base.html` gets cached, so even after a change to a file that *is* being watched, the changes to `base.html` are not present.

Is this a limitation that Parcel has with all plugins for PostHTML/PostCSS, or is this something that the `posthtml-extend` package could be changed to support?

## 🔦 Context

If I make changes to the base template I have to quit the dev env, delete the .cache dir, and start it back up. 

## 🌍 Your Environment

| Parcel           | latest
| Node             | latest
| npm/Yarn         | latest

fathyb:
Fixes #1670

I didn't find much about `tree.messages`, it is in [`posthtml` tests](https://github.com/posthtml/posthtml/blob/788a306b33b6930b9bbfbcb659484e75cf3ee940/test/messages.js).

Also found in `posthtml-include` :
https://github.com/posthtml/posthtml-include/blob/e4f2a57c2e52ff721eed747b65eddf7d7a1451e3/index.js#L18-L26

### DIFF
--- a/package.json
+++ b/package.json
@@ -93,6 +93,7 @@
     "node-sass": "^4.7.2",
     "nyc": "^11.1.0",
     "postcss-modules": "^1.1.0",
+    "posthtml-extend": "^0.2.1",
     "posthtml-include": "^1.1.0",
     "prettier": "^1.13.0",
     "pug": "^2.0.3",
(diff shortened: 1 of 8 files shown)


devongovett:
looks like the test you added is failing

fathyb:
Should be fixed 🙂

PR #1683, opened 2018-07-08
Fix tree-shaking named import on wrapped module

fathyb:
Fixes de-optimized modules like `redux-bundler`

### DIFF
--- a/src/scope-hoisting/concat.js
+++ b/src/scope-hoisting/concat.js
@@ -339,7 +339,7 @@ module.exports = (packager, ast) => {
           let {identifier} = findExportModule(match[1], name, path);
 
           // Check if $id$export$name exists and if so, replace the node by it.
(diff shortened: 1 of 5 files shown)


devongovett:
maybe we could just ignore es6 exports if it is not an es6 module. We already set the `isES6Module` flag to false in the hoister when we wrap the module (https://github.com/parcel-bundler/parcel/blob/master/src/scope-hoisting/hoist.js#L114), but we don't check it in concat (https://github.com/parcel-bundler/parcel/blob/master/src/scope-hoisting/concat.js#L46).

I think if we either clear the exports in the hoister, or check the flag in concat it should fix this issue. Then we wouldn't need to have all this extra fallback code.

fathyb:
@devongovett yup I overthought that, pushed a simpler solution

PR #1685, opened 2018-07-08
Don't minify in asset when using scope-hoisting

### LINKED ISSUE #1664: FontAwesome 5 and tree-shaking
# 🐛 bug report

```js
{
  "presets": [
    "env"
  ],
  "plugins": [
    "transform-class-properties",
    ["styled-components", {
      "displayName": true
    }],
    "polished"
  ]
}
```

## 🤔 Expected Behavior

I've followed the [latest instructions](https://fontawesome.com/how-to-use/with-the-api/other/tree-shaking) for tree-shaking in the FontAwesome docs. So my icon component looks something like this;

```js
import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
import { library } from '@fortawesome/fontawesome-svg-core'
import {
  faArchive,
  faChevronDown,
  faChevronLeft,
  faChevronRight,
  faChevronUp,
  faEllipsisH,
  faPencil,
  faPlus,
  faSearchMinus,
  faSearchPlus,
  faTrash
} from '@fortawesome/pro-solid-svg-icons'
import styled from 'styled-components'

library.add(
  faArchive,
  faChevronDown,
  faChevronLeft,
  faChevronRight,
  faChevronUp,
  faEllipsisH,
  faPencil,
  faPlus,
  faSearchMinus,
  faSearchPlus,
  faTrash
)

const Icon = styled(FontAwesomeIcon)``

export default Icon
```

## 😯 Current Behavior

When I inspect the generated bundle with the analyzer, I see that FontAwesome takes most of the size. A further look into the minified output, I see a lot of references to icons I haven't imported. The command I run to build is `parcel build src/index.html --experimental-scope-hoisting`.

## 🔦 Context

The bundle size is very large as the FontAwesome pro import takes roughly 518kb. It would decrease significantly when shaken correctly.

## 💻 Code Sample

See above

## 🌍 Your Environment

| Parcel           | 1.9.3
| Node             | 8.9.4
| npm/Yarn         | 1.6.0
| Operating System | MacOS 10.13.5

fathyb:
Should bring a speed and size boost

Fixes #1664

### DIFF
--- a/src/assets/JSAsset.js
+++ b/src/assets/JSAsset.js
@@ -146,7 +146,7 @@ class JSAsset extends Asset {
       }
     }
 
-    if (this.options.minify) {
+    if (this.options.minify && !this.options.scopeHoist) {
       await terser(this);
     }
   }


devongovett:
hmm, I think this will just completely turn off minification for tree shaking. the minification that happens in the concatenation phase only operates on the top-level scope, so disabling terser will disable minification of inner scopes.

fathyb:
Oh, thought it was a merge error. Is there any reason to do this? Giving the minifier the complete bundle can only improve minification

devongovett:
Performance. This way, we get the best of both worlds. Terser minifies everything, but skips mangling the top-level scope by default. Then all we have to do over the whole bundle is tree shake (which we have a custom implementation for anyway), and mangle the top-level scope only.

devongovett:
@kzc what specific optimizations would benefit from running over the whole bundle besides the top-level DCE and mangling that parcel already does in combination with running terser on each file?

devongovett:
In my testing, the difference in size between running terser on the whole bundle vs on individual files with our additional top-level optimizations was negligible.

kzc:
> what specific optimizations would benefit from running over the whole bundle besides the top-level DCE and mangling that parcel already does in combination with running terser on each file?

Running the Terser `compress` phase at the individual file level and not globally will produce less optimal results in the average case, and depending on the project it could be quite large. I appreciate that Parcel defaulting to a "quick build" mode using the hybrid minify scheme outlined above could be beneficial to most users. But some users may be willing to opt in to have Terser produce smaller bundles globally.

If Terser is run globally (ideally with multiple `passes`) it can:

* Perform global constant variable propagation.
* Inline single use functions globally. If mangling is done at the file level it will impede this in some cases. 
* Do additional object literal property to variable hoisting.
* Additionally drop unused `/*@__PURE__*/` annotated function calls.
* Perform additional DCE as the cumulative effects of all these transforms take place.
* Optional [property mangling](https://github.com/fabiosantoscode/terser#mangle-properties-options) can only really be done globally unless a name cache is used. While not recommended for novice use, when property mangling is done properly bundles can be further reduced in size by as much as 40%. However, a name cache won't work correctly when read/written in parallel.

If for speed/parallelism reasons you wish to have parcel minify each file individually then you should disable mangle in this phase, and only perform mangle globally. I do recommend at least one `compress` pass globally. 

Also decide what phase(s) `.terserrc` applies to. Does it apply to the global phase or the local file phase? Each phase serves a different purpose and should have different minify settings for expert use.

fathyb:
@devongovett to complete @kzc answer here is a simple example : 

- `a.js`
```js
import {b} from './b'

console.log(b.foo)
console.log(b.bar)
```

- `b.js`
```js
export const b = {
  foo: 'bar',
  bar: 'foo'
}
```

- `out.js` on `master`:
```js
(function () {
  var a = {
    foo: "bar",
    bar: "foo"
  };
  console.log(a.foo),
  console.log(a.bar);
})();
```

- `out.js` by using Terser on the complete bundle 
```js
console.log("bar"), console.log("foo");
```

---

Another one using property mangling, pretty common in the real world with frameworks like React : 
- `index.js`
```js
import A from './a'
import B from './b'
import React from './react'

React.render(
  React.createElement([
    A, B
  ])
)
```
- `a.js`
```js
import React from './react'

export default React.createElement('div')
```

- `b.js`
```js
import React from './react'

export default React.createElement('div')
```

- `react.js` 
```js
export default {
  createElement: (...args) => React.createElement(...args),
  render: (elem) => ReactDOM.render(elem)
}
```

- `out.js` on `master`
```js
(function() {
    var a = {
        createElement: function() {
            var e;
            return (e = React).createElement.apply(e, arguments)
        },
        render: function(e) {
            return ReactDOM.render(e)
        }
    };
    var b = a.createElement("div");
    var c = a.createElement("div");
    a.render(a.createElement([b, c]));
})();
```

- `out.js` by using Terser on the complete bundle 
```js
!function() {
    var e, n = function() {
        var e;
        return (e = React).createElement.apply(e, arguments);
    };
    e = n([ n("div"), n("div") ]), ReactDOM.render(e);
}();
```

kzc:
> Another one using property mangling

@fathyb The react example above demonstrated object literal property hoisting.

Here are examples of property mangling:

https://github.com/fabiosantoscode/terser#cli-mangling-property-names---mangle-props

Property mangling is not enabled by default.

PR #1689, opened 2018-07-09
WIP: Add PostCSS's SugarSS recognition by default

DoumanAsh:
Personally I find [SugarSS](https://github.com/postcss/sugarss) is pretty nice simplified version of CSS syntax.

Since its addition is configured by setting corresponding parse in PostCSS's config, I wonder if it would be ok to add its extension by default?

### DIFF
--- a/src/assets/Ast.js
+++ b/src/assets/Ast.js
@@ -1,0 +1,18 @@
+class CSSAst {
+  constructor(css, root) {
+    this.css = css;
+    this.root = root;
+    this.dirty = false;
+  }
+
+  render() {
+    if (this.dirty) {
+      this.css = '';
+      postcss.stringify(this.root, c => (this.css += c));
+    }
+
+    return this.css;
+  }
+}
+
+module.exports.CSSAst = CSSAst;
--- a/src/assets/Ast.js
+++ b/src/assets/Ast.js
@@ -1,3 +1,5 @@
+const postcss = require('postcss');
+
 class CSSAst {
   constructor(css, root) {
     this.css = css;
--- a/src/assets/CSSAsset.js
+++ b/src/assets/CSSAsset.js
@@ -3,6 +3,7 @@ const postcss = require('postcss');
 const valueParser = require('postcss-value-parser');
 const postcssTransform = require('../transforms/postcss');
 const CssSyntaxError = require('postcss/lib/css-syntax-error');
+const CSSAst = require('./Ast.js').CSSAst;
 
 const URL_RE = /url\s*\("?(?![a-z]+:)/;
 const IMPORT_RE = /@import/;
@@ -162,23 +163,6 @@ class CSSAsset extends Asset {
     }
 
     return err;
-  }
-}
-
-class CSSAst {
-  constructor(css, root) {
-    this.css = css;
-    this.root = root;
-    this.dirty = false;
-  }
-
-  render() {
-    if (this.dirty) {
-      this.css = '';
-      postcss.stringify(this.root, c => (this.css += c));
-    }
-
-    return this.css;
   }
 }
 
--- a/src/assets/SSSAsset.js
+++ b/src/assets/SSSAsset.js
@@ -1,10 +1,10 @@
 const postcss = require('postcss');
 
 const localRequire = require('../utils/localRequire');
-const CSSAsset = require('./CSSAsset.js');
+const Asset = require('../Asset');
 const CSSAst = require('./Ast.js').CSSAst;
 
-class SSSAsset extends CSSAsset {
+class SSSAsset extends Asset {
   constructor(name, options) {
     super(name, options);
     this.type = 'css';
--- a/src/assets/SSSAsset.js
+++ b/src/assets/SSSAsset.js
@@ -1,0 +1,21 @@
+const postcss = require('postcss');
+
+const localRequire = require('../utils/localRequire');
+const CSSAsset = require('./CSSAsset.js');
+const CSSAst = require('./Ast.js').CSSAst;
+
+class SSSAsset extends CSSAsset {
+  constructor(name, options) {
+    options.css_parser = "sugarss";
+    super(name, options);
+    this.type = 'css';
+  }
+
+  async parse(code) {
+    let sugarss = await localRequire('sugarss', this.name);
+    let root = postcss.parse(code, {from: this.name, to: this.name, parser: sugarss});
+    return new CSSAst(code, root);
+  }
+}
+
+module.exports = SSSAsset;
--- a/src/assets/SSSAsset.js
+++ b/src/assets/SSSAsset.js
@@ -6,7 +6,6 @@ const CSSAst = require('./Ast.js').CSSAst;
 
 class SSSAsset extends CSSAsset {
   constructor(name, options) {
-    options.css_parser = "sugarss";
     super(name, options);
     this.type = 'css';
   }
--- a/src/assets/SSSAsset.js
+++ b/src/assets/SSSAsset.js
@@ -15,6 +15,16 @@ class SSSAsset extends CSSAsset {
     let root = postcss.parse(code, {from: this.name, to: this.name, parser: sugarss});
     return new CSSAst(code, root);
   }
+  
+  generate() {
+    return [
+      {
+        type: 'css',
+        value: this.ast ? this.ast.render() : this.contents
+      }
+    ];
+  }
+
 }
 
 module.exports = SSSAsset;
--- a/src/transforms/postcss.js
+++ b/src/transforms/postcss.js
@@ -53,15 +53,6 @@ async function getConfig(asset) {
     config.plugins.push(postcssModules(postcssModulesConfig));
   }
 
-  if (config.parser === undefined && asset.options.css_parser !== undefined) {
-      const name = asset.options.css_parser;
-
-      if ((typeof name !== "string")) throw "Internal Error: Invalid PostCSS parser type. Should be string";
-
-      let parser = await localRequire(name, asset.name);
-      config.parser = parser;
-  }
-
   if (asset.options.minify) {
     let [cssnano, {version}] = await Promise.all(
       ['cssnano', 'cssnano/package.json'].map(name =>
--- a/src/transforms/postcss.js
+++ b/src/transforms/postcss.js
@@ -53,6 +53,15 @@ async function getConfig(asset) {
     config.plugins.push(postcssModules(postcssModulesConfig));
   }
 
+  if (config.parser === undefined && asset.options.css_parser !== undefined) {
+      const name = asset.options.css_parser;
+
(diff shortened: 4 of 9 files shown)


Kompwu:
@DoumanAsh Can you help me ?
I'm trying to get `.sss` bundled...

DoumanAsh:
@Kompwu Well as far as I understand currently you'd need to register this extension to be recognized as CSS asset:

```js
const Bundler = require('parcel-bundler');

let bundler = new Bundler('input.js');
bundler.addAssetType('.sss', './assets/CSSAsset');
```

If I understand this API correctly(I'm currently decided to use plain CSS with nesting support

Kompwu:
@DoumanAsh Thanks,
How can i make the same thing without js ?

DoumanAsh:
@Kompwu Sorry I'm not sure as I'm new to parcel myself

Kompwu:
@DeMoorJasper Can you merge it pls ?

DeMoorJasper:
It would be nice if the sugarSS asset actually installed and configured postCSS for the user, this way it works `zero-config`, you can probably create an asset that runs sss through postCSS and then passes it to css. (Hope this makes any sense, lemme know if you need any help)

DoumanAsh:
@DeMoorJasper Ok, I'll try to approach with this suggestion. But what should we do if user already has config?

DeMoorJasper:
@DoumanAsh You can probably load PostCSS and check for SugarSS and just forward it to CSS if it's already installed and configured I guess?

Kompwu:
@DoumanAsh How's going ?

DoumanAsh:
I tried to give it a go by extending existing CSS asset, but it makes me think that it is rather difficult to do.

The problem is not only to load postcss config, but there is also `parse` method that requires you to specify syntax so I'd need to load sugarss in two places it seems

Kompwu:
@DoumanAsh Do what you can and we'll help you with everything else.

Kompwu:
@DoumanAsh How's going?

DoumanAsh:
@Kompwu Sorry, I've been a bit busy lately, I'll try to resume working on it as soon as I have time

DoumanAsh:
Currently I have issue with PostCSS API usage as it seems to cause errors when I attempt to parse `.sss`:
```CssSyntaxError: Missed semicolon````

Though I did set `let root = postcss.parse(code, {from: this.name, to: this.name, syntax: sugarss})` corresponding parser.

I was thinking that I could parse SugarSSS into CSS and then rely on existing `CSSAsset` but it seems I misunderstood the API as it ignores provided by me syntax.

**Comment by Kompwu** [comment]
@DoumanAsh That's weird... i have no idea mate.

**Comment by Kompwu** [comment]
@DoumanAsh Any news ?
@devongovett Hey, can you help us ?

**Comment by DoumanAsh** [comment]
@Kompwu Sorry, recently I'm busy with other things and didn't look into PostCSS

**Comment by DoumanAsh** [comment]
@Kompwu Sorry I'm not actively using parcel as of the moment and kinda busy so I'm not planning to resume work on it until need arise again.

I don't mind someone to carry on, or leave issue open so that me or someone else could continue later on.

**Comment by Kompwu** [comment]
@ai Hi, can you take a look ?

File: src/assets/SSSAsset.js [hunk 1]
```diff
@@ -0,0 +1,35 @@
...
+
+class SSSAsset extends CSSAsset {
+  constructor(name, options) {
+    options.css_parser = "sugarss";
+    super(name, options);
+    this.type = 'css';
+  }
+
+  parse(code) {
+    let root = postcss.parse(code, {from: this.name, to: this.name, syntax: sugarss});
```
ai: `syntax` => `parser`.

We need to set SugarSS only as input language (parser) and keep output language (stringifier) as default CSS

DoumanAsh: I think I also tried it initially, but for some reason postcss still parsed code as if it was CSS

ai: I will try to run tests after Def Con

Kompwu: I can try if you want.

ai: Yeap, it should be `parser:`.

Kompwu: So all that remains is to automate the configuration?

Kompwu:
@DeMoorJasper Are you sure automating configuration is a good thing?
I don't think Parcel should do it.

DeMoorJasper:
@Kompwu We are not automating config, we're adding a default?

I just thought it might be useful to actually have something with extension sugarSS be transpiled to css by default (it should still read the config and potentially even just pass along the sugarss file to postCSS without any transforms if it's configured). The rest of the PostCSS transforms will still apply as far as I know...

Kompwu:
@DoumanAsh Can you finish?
All that remains is to add a default configuration if there is none.

DoumanAsh:
@Kompwu To be honest I'm not sure why, but even when I place `parser`, it still fails in simple integration test.

The current commit also makes `sugarss` required dependency, would it be possible to make it optional?

DeMoorJasper:
How does it add it as a dependency, you could use localRequire, that's how it's supposed to be used.

src/assets/SSSAsset.js
```diff
@@ -0,0 +1,35 @@
...
+
+class SSSAsset extends CSSAsset {
+  constructor(name, options) {
+    options.css_parser = "sugarss";
+    super(name, options);
+    this.type = 'css';
+  }
+
+  parse(code) {
+    let root = postcss.parse(code, {from: this.name, to: this.name, parser: sugarss});
```
DeMoorJasper: let sugarss = await localRequire('sugarss')

src/assets/SSSAsset.js
```diff
@@ -0,0 +1,35 @@
+const postcss = require('postcss');
+const sugarss = require('sugarss');
```
DeMoorJasper: should be a local require

src/assets/SSSAsset.js
```diff
@@ -0,0 +1,35 @@
...
+    this.type = 'css';
+  }
+
+  parse(code) {
+    let root = postcss.parse(code, {from: this.name, to: this.name, parser: sugarss});
+    return new CSSAst(code, root);
+  }
+}
+
+class CSSAst {
```
DeMoorJasper: Split this up into a seperate file, so both CSS and SugarSS Asset use it

src/transforms/postcss.js
```diff
@@ -48,6 +48,15 @@ async function getConfig(asset) {
     config.plugins.push(postcssModules(postcssModulesConfig));
   }
 
+  if (config.parser === undefined && asset.options.css_parser !== undefined) {
+      const name = asset.options.css_parser;
+
+      if (!(typeof name === "string")) throw "Internal Error: Invalid PostCSS parser type. Should be string";
```
DeMoorJasper: `typeof name !== "string"`

src/transforms/postcss.js
```diff
@@ -48,6 +48,15 @@ async function getConfig(asset) {
...
+
+      let parser = await localRequire(name, asset.name).catch(() => require(name));
```
DeMoorJasper: why does it catch with a require, that's what localRequire already does?

localRequire's should just throw if it fails, because require is just the same thing. (Unless you want to require something from inside parcel, but that should happen inside the transform not be a fallback as this could have unwanted side-effects)

src/assets/SSSAsset.js
```diff
@@ -0,0 +1,35 @@
+const postcss = require('postcss');
+const sugarss = require('sugarss');
+const CSSAsset = require('./CSSAsset.js');
+
+class SSSAsset extends CSSAsset {
+  constructor(name, options) {
+    options.css_parser = "sugarss";
```
DeMoorJasper: Why is this necessary?

DoumanAsh: I wanted to propagate to `transform` to automatically load parser, if it is necessary.

But not sure if it is good idea

DoumanAsh:
Thanks for review, I applied majority of comments.

I'm not sure though how to automate configuration for PostCSS though.

My simple integration test is still failing me:

```
  sugarss
    1) should correctly parse SugarSS asset


  0 passing (478ms)
  1 failing

  1) sugarss
       should correctly parse SugarSS asset:
     CssSyntaxError: Missed semicolon
      at Input.error (node_modules\postcss\lib\input.js:119:22)
      at Parser.checkMissedSemicolon (node_modules\postcss\lib\parser.js:544:26)
      at Parser.decl (node_modules\postcss\lib\parser.js:269:50)
      at Parser.other (node_modules\postcss\lib\parser.js:168:18)
      at Parser.parse (node_modules\postcss\lib\parser.js:84:26)
      at Function.parse (node_modules\postcss\lib\parse.js:24:16)
      at SSSAsset.parse (src\assets\SSSAsset.js:16:24)
```

src/assets/Ast.js
```diff
@@ -0,0 +1,18 @@
+class CSSAst {
```
DeMoorJasper: Rename the file to CSSAst, as js and html also have an ast...

src/assets/Ast.js
```diff
@@ -0,0 +1,18 @@
...
+    if (this.dirty) {
+      this.css = '';
+      postcss.stringify(this.root, c => (this.css += c));
+    }
+
+    return this.css;
+  }
+}
+
+module.exports.CSSAst = CSSAst;
```
DeMoorJasper: Just make this the default export `module.exports = CSSAst;`

src/assets/SSSAsset.js
DeMoorJasper: You already do a localRequire in transform so not sure why this is necessary at all.

DeMoorJasper:
You could check if the postCSS config already has a sugarss transform setup and skip this transform entirely if it has. But in the test it probably shouldn't have any postCSS config. Otherwise you're basically just testing postCSS

src/assets/CSSAsset.js
```diff
@@ -3,6 +3,7 @@ const postcss = require('postcss');
 const valueParser = require('postcss-value-parser');
 const postcssTransform = require('../transforms/postcss');
 const CssSyntaxError = require('postcss/lib/css-syntax-error');
+const CSSAst = require('./Ast.js').CSSAst;
```
DeMoorJasper: Do the same thing for CSSAsset

DeMoorJasper:
I've changed some stuff around and it should be working better now. You might want to cleanup and debug it a bit more as it's not actually working right now, and def not clean

src/assets/Ast.js
DoumanAsh: I was thinking that it would be better for future to have common source of asts?

DeMoorJasper: In the future every asset will have it's own package so it will be seperate anyways

DeMoorJasper:
Gonna open up a new PR that bases of off this one, with fixed code

PR #1690, opened 2018-07-09
Support optionally bundling node_modules for --target=node

devongovett:
Closes #796.

This is an updated version of #796.

* Adds a `--bundle-node-modules` flag, which turns on bundling of node_modules for `--target=node` and `--target=electron`.
* Disables `browser` field resolving for non `--target=browser`
* Disables bundling node builtin modules (e.g. `fs`, `zlib`, etc.) for `--target=node`.


mvlabat:
Hi! Thanks for the really nice feature. It would also be very cool to specify which modules to bundle and which ones not. In my project I want to bundle only several packages (my internal libs), but not all the packages.

lpil:
@mvlabat Sounds useful, but could we make that a later enhancement?

It'd be nice to avoid adding additional features in this PR so that it can go through more quickly :)

mohsen1:
How does this work for modules that compile binaries per platform using Gyp? If I bundle on Linux, is there a chance that program will not run in MacOS?

devongovett:
we don't handle native node modules yet. that would be a separate feature.

taschmidt:
This is great!  Any idea when 0.10.0 will be pushed?

GuiRitter:
I like this, apparently solves a problem I'm trying to fix. I said "apparently" because another problem showed up, and maybe Parcel can be improved to fix it.

I'm trying to bundle [text-to-svg](https://www.npmjs.com/package/text-to-svg) as a dependency. That package comes with a `fonts/ipag.ttf` file that I suspect is not bundled. I'm trying to find a workaround to get this font loaded.

Any chance Parcel can be improved to bundle resources as well?

purkhusid:
Any timeline on when a release with this feature will be published?

tripodsan:
@devongovett is it possible to control which modules to bundle? e.g. when creating an package to run inside a node8 docker container, I only want to bundle the ones the container doesn't provide.

tripodsan:
Hi @Stradivario. thank you very much for your explanation. unfortunately, I don't want to build an entire dockercontainer; just reusing one.

here's an example:

let's say, I build an openwhisk action, using the default [node 6](https://github.com/apache/incubator-openwhisk/blob/master/docs/actions-node.md#nodejs-version-6-environment) container.
For example, they already include lodash. So, in my javasccript, when I require lodash, I don't need to inline it.

in webpack, I could easily configure this and declare lodash as external.

peanutbother:
@tripodsan you could alias these modules to a local stub pointing to your container module to do that theoretically. saying theoretically since I did not test such a scenario, but had local stubs in aliases for having client and server common config using config module and a local stub.

peanutbother:
> let's say, I build an openwhisk action, using the default [node 6](https://github.com/apache/incubator-openwhisk/blob/master/docs/actions-node.md#nodejs-version-6-environment) container.
> For example, they already include lodash. So, in my javasccript, when I require lodash, I don't need to inline it.

@tripodsan you could alias these modules to a local stub pointing to your container module to do that theoretically. saying theoretically since I did not test such a scenario, but had local stubs in `aliases ` (package.json) for having client and server common config using config module and a local stub.

You can find out about aliases [here](https://parceljs.org/module_resolution.html#aliasing)

tripodsan:
cool. thanks for the hint, @peanutbother

devongovett:
@tripodsan seems like you want https://github.com/parcel-bundler/parcel/issues/144

devongovett:
@pontusab @purkhusid just published a beta that includes this feature: v1.10.0-beta.0. Please test it out and let me know if you find issues. :)

tony-g:
How difficult/possible would it be to extend this to integrate with yarn workspaces?

e.g. in developing an electron app with workspaces, I'd like parcel to watch for changes to workspace packages like `shared-module-one`, but not `fs` etc

```
├── package.json
└── packages
    ├── electron-app
    │   ├── index.js
    │   └── package.json
    ├── shared-module-one
    │   ├── index.js
    │   └── package.json
    ├── etc
```

```js
/* electron-app/index.js */
const fs = require('fs')
const sharedModuleFromWorkspace = require('shared-module-one')

/* shared-module-one/index.js */
// parcel won't auto-rebuild changes here when target=electron
// but would be nice if it did!
```

PR #1694, opened 2018-07-10
Deterministic asset ids

### LINKED ISSUE #735: 🙋 Deterministic builds

### LINKED ISSUE #1656: Uncaught TypeError: (xxxx) is not a function
# 🐛 bug report

When I start the parcel dev server after a night or something, and open up my browser, randomly I get a "Uncaught TypeError: **** is not a function". It happend with 3 different things the past days, and it's getting annoying. Workaround `rm -rf .cache dist`

.babelrc
```js
{
  "presets": ["stage-0", "env", "react"],
  "plugins": ["transform-class-properties"]
}
```

package.json
```js
{
  "name": "frontend-graphql",
  "version": "0.1.0",
  "private": true,
  "dependencies": {
    "apollo-cache-inmemory": "^1.2.2",
    "apollo-client": "^2.3.2",
    "apollo-link": "^1.2.2",
    "apollo-link-context": "^1.0.8",
    "apollo-link-error": "^1.1.0",
    "apollo-link-http": "^1.5.4",
    "babel-polyfill": "^6.26.0",
    "babel-preset-env": "^1.7.0",
    "babel-preset-stage-0": "^6.24.1",
    "graphql": "^0.13.2",
    "graphql-tag": "^2.9.2",
    "react": "^16.4.0",
    "react-apollo": "^2.1.4",
    "react-dom": "^16.4.0",
    "react-router-dom": "^4.3.1"
  },
  "scripts": {
    "start": "parcel public/index.html"
  },
  "devDependencies": {
    "parcel-bundler": "^1.9.4"
  }
}
```

## 🤔 Expected Behavior

__not__ error and work correctly

## 😯 Current Behavior

read the report ^^

image

## 💻 Code Sample

https://gitlab.com/homework-manager/hwm-frontend

## 🌍 Your Environment

| Parcel           | 1.9.4 |
| Node             | 10.5.0 |
| npm/Yarn         | Yarn 1.7.0 |
| Operating System | Arch Linux // kernel 4.17.2-1-ARCH |

devongovett:
Closes #780. Fixes #735. Fixes #1656.

Before this change, asset ids were integers that were dependent on resolution order. There were two problems with this:

1. Builds were not deterministic (#735). This meant that if you did two builds without changing any files, the output might be different, since the asset ids might change. This is not great for long-term cachability, build verification, etc.
2. Since scope hoisting was merged, we started depending on the asset ids as part of variable identifiers. We started caching these asset ids so that references between modules were kept the same if you rebuilt from cache. The problem with caching asset ids is that if you actually changed a file and it was invalidated in the cache, a new asset id was generated. Then, references using the old ids would be invalid, and id collisions were also possible. This caused issues like #1656 to appear.

This PR changes asset ids to be deterministic, based on the relative path to that file from the root project directory. There are two cases:

* In production, we use a short 4-character base64 md5 hash as the id. While slightly longer than the numeric ids of the past, this should provide enough uniqueness for fairly large apps. 
* In development, we simply use the relative path as the id without modification. This avoids the cost of hashing, and also could help with debugging bundles since you can see the original filenames right in the output.

I **think** I changed all the places where we assumed asset ids were numbers and not strings, and it passes all tests, but this could use a bit of real world testing.

### DIFF
--- a/src/Asset.js
+++ b/src/Asset.js
@@ -10,8 +10,6 @@ const syncPromise = require('./utils/syncPromise');
 const logger = require('./Logger');
 const Resolver = require('./Resolver');
 
-let ASSET_ID = 1;
-
 /**
  * An Asset represents a file in the dependency tree. Assets can have multiple
  * parents that depend on it, and can be added to multiple output bundles.
@@ -20,7 +18,7 @@ let ASSET_ID = 1;
  */
 class Asset {
   constructor(name, options) {
-    this.id = ASSET_ID++;
+    this.id = null;
     this.name = name;
     this.basename = path.basename(this.name);
     this.relativeName = path.relative(options.rootDir, this.name);
@@ -188,6 +186,17 @@ class Asset {
   }
 
   async process() {
+    // Generate the id for this asset, unless it has already been set.
+    // We do this here rather than in the constructor to avoid unnecessary work in the main process.
+    // In development, the id is just the relative path to the file, for easy debugging and performance.
+    // In production, we use a short hash of the relative path.
+    if (!this.id) {
+      this.id =
+        this.options.production || this.options.scopeHoist
+          ? md5(this.relativeName, 'base64').slice(0, 4)
+          : this.relativeName;
+    }
+
     if (!this.generated) {
       await this.loadIfNeeded();
       await this.pretransform();
--- a/src/Bundler.js
+++ b/src/Bundler.js
@@ -525,8 +525,7 @@ class Bundler extends EventEmitter {
     let processed = this.cache && (await this.cache.read(asset.name));
     let cacheMiss = false;
     if (!processed || asset.shouldInvalidate(processed.cacheData)) {
-      processed = await this.farm.run(asset.name, asset.id);
-      processed.id = asset.id;
+      processed = await this.farm.run(asset.name);
       cacheMiss = true;
     }
 
--- a/src/Pipeline.js
+++ b/src/Pipeline.js
@@ -11,15 +11,13 @@ class Pipeline {
     this.parser = new Parser(options);
   }
 
-  async process(path, id, isWarmUp) {
+  async process(path, isWarmUp) {
     let options = this.options;
     if (isWarmUp) {
       options = Object.assign({isWarmUp}, options);
     }
 
     let asset = this.parser.getAsset(path, options);
-    asset.id = id;
-
     let generated = await this.processAsset(asset);
     let generatedMap = {};
     for (let rendition of generated) {
@@ -27,6 +25,7 @@ class Pipeline {
     }
 
     return {
+      id: asset.id,
       dependencies: Array.from(asset.dependencies.values()),
       generated: generatedMap,
       hash: asset.hash,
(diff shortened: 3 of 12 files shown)


rafaelrinaldi:
This PR might actually solve [an issue I'm experiencing with tree shaking and JSX](https://spectrum.chat/?t=811287bb-429b-4843-9f37-40c576a1e0d6).

sebastiencs:
This branch fixes #1656 for me. Thanks !

safareli:
On `parcel build` I'm still getting multiple`blob:http://localhost:8080/3818d181-16ea-4946-9409-badb9c52681d:1 Uncaught ReferenceError: e is not defined`.
screenshot 2018-07-11 13 25 36

This is contents of the [blob](https://gist.github.com/safareli/f6222ba84692a97f1f8f05b5c83879b1)

On `parcel start` i'm not getting this errors but if i change scss I see:
`Uncaught Error: Cannot find module '../node_modules/parcel-bundler/lib/builtins/css-loader.js'`

fathyb:
@safareli did you remove the `.cache` directory or tried using `--no-cache`? It's needed to take the new ids

safareli:
With parcel build, tried with no cache but I still got same e is not defined error, with parcel start tho, I  didn't got errors as before and scss change also didn't produced errors but, styles were not live reloaded (as before, I think it's separate issue introduced from 1.8)

fathyb:
@safareli thanks, just to confirm: you're only getting this error when using this PR right?

safareli:
No i'm also getting similar errors on other versions too (https://github.com/parcel-bundler/parcel/issues/1656#issuecomment-403474399) (if i add `--no-minify` then i wouldn't get the error) maybe it's related to some diferent error  ¯\_(ツ)_/¯.

fathyb:
@safareli Alright, this PR brings a heavy change to Parcel so we'd like to only discuss issues directly introduced by it here.

I think you're experiencing something different, could you open a new issue with a minimal reproduction?

rafaelrinaldi:
Just experienced another issue that might be solved by this PR. Even though I import `babel-polyfill` at the very top of my application entry point, Parcel will not respect this on the production bundle. So my application broke in browsers that need it.

devongovett:
@safareli I think this is a separate issue. Where are these `blob:` urls coming from?

devongovett:
Released in v1.9.5.

PR #1700, opened 2018-07-11
Use minify prelude only minify option

terrierscript:
I found JSConcatPackager use minify file when `option.minify = false`


DeMoorJasper:
In theory prelude should never be debugged by the end user as it should be working perfectly. Therefore it wouldn't matter as it's pre-minified and doesn't really affect speed.

Although if it ever does cause a bug it might be hard to debug so I guess it's a good idea after all.

Issue #1699, opened 2018-07-10
Tree shaking and React


rafaelrinaldi:
# 🐛 bug report

I am experiencing a faulty behavior when generating a production bundle using the experimental tree shaking feature.

[I have already logged this in the Spectrum channel](https://spectrum.chat/?t=811287bb-429b-4843-9f37-40c576a1e0d6) but decided to open a proper bug report as well.

## 🎛 Configuration (.babelrc, package.json, cli command)

I put together a repository so you can reproduce the same behavior I am experiencing: https://github.com/rafaelrinaldi/parcel-tree-shaking-jsx-error

## 🤔 Expected Behavior

Tree shaking should work just fine.

## 😯 Current Behavior

Tree shaking kinda works but seems to bypass React, logging "React is not defined" in the console.

## 💁 Possible Solution

I believe that this could be two of the following:

1. "React" is imported whenever JSX is used but not necessarily is being referenced, so the tree shaking might be skipping it for that reason (even though the `babel-plugin-transform-react-jsx` plugin added by the `babel-preset-react` should be replacing JSX nodes by `React.createElement()` calls, therefore notifying the tree shaking algorithm that the module is being used)
2. It might be an module load ordering/evaluation issue in Parcel. Perhaps even solved by #1694 

## 🔦 Context

I'm trying to get the experimental tree shaking feature to work in a project that uses a route-based code splitting strategy using react router and react loadable.

## 💻 Code Sample

I put together a repository so you can reproduce the same behavior I am experiencing: https://github.com/rafaelrinaldi/parcel-tree-shaking-jsx-error

## 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Parcel           | 1.9.4 |
| Node             | v10.6.0 |
| npm         | 6.1.0 |
| Operating System | macOS High Sierra |

rafaelrinaldi:
@buronnie Thanks for chiming in. Yes, that possibility ran through my mind today as well. The order that the transpilation steps happen will mislead the tree shaking algo in some cases.

rafaelrinaldi:
I've updated the test repository to use Parcel `v1.9.5`. Now I get a different error:

<details>
<summary>Error (from the browser console)</summary>

```
Uncaught TypeError: ($Mgd$init(...) , $Mgd$exports) is not a function
    at $iqAm$init (parcel-test.7f7f4a64.js:942)
    at $DGX$init (parcel-test.7f7f4a64.js:1153)
[...]

PR #1735, opened 2018-07-15
Fix hoisting of "var" variables on wrapped modules

fathyb:
Part of the fixes for #1699

Fix issues with some React apps. One of the problematic files (`loggedTypeFailures`) : 
https://github.com/facebook/prop-types/blob/cfc7f43af3d74978ec82052fe6bb8212f09daca0/checkPropTypes.js

### DIFF
--- a/src/scope-hoisting/hoist.js
+++ b/src/scope-hoisting/hoist.js
@@ -155,6 +155,37 @@ module.exports = {
         ) {
           scope.push({id: exportsIdentifier, init: t.objectExpression([])});
         }
+
+        // Move all "var" variables to the top-level to prevent out of order definitions when wrapped.
+        for (let name in scope.bindings) {
+          let binding = scope.getBinding(name);
+
+          if (binding.path.scope !== scope && binding.kind === 'var') {
+            let {node} = binding.path;
+            let {parentPath} = binding.path;
+
+            path.unshiftContainer('body', [
+              t.variableDeclaration('var', [t.variableDeclarator(node.id)])
+            ]);
+
+            if (!parentPath.removed) {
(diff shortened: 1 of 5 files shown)


src/scope-hoisting/hoist.js
```diff
@@ -135,6 +135,36 @@ module.exports = {
         ) {
           scope.push({id: exportsIdentifier, init: t.objectExpression([])});
         }
+
+        // Move all "var" variables to the top-level to prevent out of order definitions when wrapped.
+        for (let name in scope.bindings) {
```
devongovett: hmm yeah we do something like this here already: https://github.com/parcel-bundler/parcel/blob/master/src/packagers/JSConcatPackager.js#L282-L319 but it wouldn't get variables inside an if statement like that.

Is there a performance hit to doing this on every module? Should we only do it on wrapped modules (in the packager)?

fathyb: > Is there a performance hit to doing this on every module?

I did not found any noticeable change on build time on a medium size app, same for vue-hackernews. We only do it when a `var` variable is not in the root block of the current scope, which is relatively rare.

> Should we only do it on wrapped modules (in the packager)?

I guess you could either :
- build a scope using `babel-traverse` and do the same as here => slow
- do a deep AST traversal => slow and error prone

Just found out other some issues with this PR, and I think I should also do it when `path.getData('shouldWrap') == true`,  I'm putting it in WIP.

fathyb: Looks like we don't need to do it when shouldWrap is true, ready to merge!

fathyb:
This breaks React, closing it.

PR #1737, opened 2018-07-15
Prevent nameclashes with internal variables

### LINKED ISSUE #1699: Tree shaking and React

fathyb:
Fixes #1699

Found in [`core-js` code](https://cdn.jsdelivr.net/npm/core-js/modules/_wks.js) 
```js
var $exports = module.exports = function (name) {
  return store[name] || (store[name] =
    USE_SYMBOL && Symbol[name] || (USE_SYMBOL ? Symbol : uid)('Symbol.' + name));
};

$exports.store = store;
```

### DIFF
--- a/src/scope-hoisting/hoist.js
+++ b/src/scope-hoisting/hoist.js
@@ -121,7 +121,7 @@ module.exports = {
         // Rename each binding in the top-level scope to something unique.
         for (let name in scope.bindings) {
           if (!name.startsWith('$' + t.toIdentifier(asset.id))) {
-            let newName = getName(asset, 'var', name, '$');
+            let newName = getName(asset, 'var', name);
             rename(scope, name, newName);
           }
         }
--- a/src/scope-hoisting/hoist.js
+++ b/src/scope-hoisting/hoist.js
@@ -121,7 +121,7 @@ module.exports = {
         // Rename each binding in the top-level scope to something unique.
         for (let name in scope.bindings) {
           if (!name.startsWith('$' + t.toIdentifier(asset.id))) {
-            let newName = getName(asset, 'var', name);
+            let newName = getName(asset, 'var', name, '$');
             rename(scope, name, newName);
           }
         }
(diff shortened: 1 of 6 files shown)


src/scope-hoisting/hoist.js
```diff
@@ -121,7 +121,7 @@ module.exports = {
         // Rename each binding in the top-level scope to something unique.
         for (let name in scope.bindings) {
           if (!name.startsWith('$' + t.toIdentifier(asset.id))) {
-            let newName = getName(asset, 'var', name);
+            let newName = getName(asset, 'var', name, '$');
```
devongovett: I don't think this should clash with anything. This would generate e.g. `$id$var$exports` vs our internal which is just `$id$exports`. Seems like the problem might be something else though I'm not sure why this would fix it...

devongovett:
Turns out the regex matching exports variables was wrong. Fixed it.

PR #1741, opened 2018-07-16
feat(builtins): load min versions if available

martell:
helpers.js prelude2.js and prelude.js can have min versions.
I think we should load these instead of blindly checking for the non min versions.
Thoughts?


DeMoorJasper:
Perhaps write a helper function to make it cleaner. Not sure if this feature is actually required or usefull as minified version should always be available in released version or built version

martell:
@DeMoorJasper I created a file `src/utils/getExisting.js` with the other utils which should be cleaner and is used for all 3 builtins.

src/packagers/JSConcatPackager.js
```diff
@@ -1,27 +1,15 @@
...
-    .readFileSync(path.join(__dirname, '../builtins/prelude2.min.js'), 'utf8')
-    .trim()
-    .replace(/;$/, '')
-};
-
-const helpers =
-  fs
-    .readFileSync(path.join(__dirname, '../builtins/helpers.js'), 'utf8')
-    .trim() + '\n';
+const prelude = getExisting(path.join(__dirname, '../builtins/prelude2'));
```
DeMoorJasper: Sorry for being difficult, but I'd prefer this to be more explicit, like use the full paths, because this overcomplicates it a little in my opinion

martell: Understood. I did it in a separate commit.
I'll squash after we get this into an approved state.

src/utils/getExisting.js
```diff
@@ -0,0 +1,21 @@
+const fs = require('fs');
+/**
+ * Creates an object that contains both source and minified source.
+ * e.g. builtins.min.js and builtins.js.
+ * If there is only one argument and it does not have an extension
+ * we use min.js and .js (with .js as the minified fallback).
+ */
+module.exports = function(minified, source) {
+  if(minified.split('.').length < 2 && source === undefined) {
```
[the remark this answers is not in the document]
DeMoorJasper: Yeah sure seems obsolete now that you've written them explicitly

martell:
done, this should be good to go. I also squashed it.

src/packagers/JSPackager.js
```diff
@@ -1,19 +1,14 @@
...
-  source: fs
-    .readFileSync(path.join(__dirname, '../builtins/prelude.js'), 'utf8')
-    .trim(),
-  minified: fs
-    .readFileSync(path.join(__dirname, '../builtins/prelude.min.js'), 'utf8')
-    .trim()
-    .replace(/;$/, '')
-};
+const prelude = getExisting(
+  path.join(__dirname, '../builtins/prelude.min.js')
```
DeMoorJasper: It appears you forgot a `,` here though

martell: thanks, done

PR #1749, opened 2018-07-18
Load minified version if available

DeMoorJasper:
Improvement on #1741


DeMoorJasper:
Probably linking using git, I think we might just merge is as people keep requesting it and this seems like the cleanest way to support this.

PR #1755, opened 2018-07-18
Fix sourcemaps failing on refresh

### LINKED ISSUE #997: sourcemaps works incorrectly after any change in dev mode
index.html
```html
<!DOCTYPE html>
<html lang="en">
<head>
    <meta charset="UTF-8">
    <title>Title</title>
</head>
<body>
</body>
</html>
```

```js
//index.js
import { foo } from "./foo";

console.log("index");
foo();
```
```js
//foo.js
export function foo() {
  console.log("foo");
}
```
Run `parcel index.html`

After first build source maps works correctly
console output:
```
index         index.js:3
foo           foo.js:2
```

But after some changes in any of file sourcemaps turns into a mess:
```
index         index.js:3
foo           parcel-sourcemap.js:82
```

### Environment

| Parcel           | 1.6.2
| Node             | 9.2.1
| npm/Yarn         | 5.5.1
| Operating System | osx

DeMoorJasper:
Sourcemaps inside assets were being overwritten due to not cloning the actual mappings array before merging and manipulating them inside the packager. Which caused the sourcemaps to get their position + bundle offset, resulting in incorrect mappings

Closes #997

### DIFF
--- a/src/SourceMap.js
+++ b/src/SourceMap.js
@@ -1,5 +1,6 @@
 const {SourceMapConsumer, SourceMapGenerator} = require('source-map');
 const lineCounter = require('./utils/lineCounter');
+const clone = require('clone');
 
 class SourceMap {
   constructor(mappings, sources) {
@@ -38,6 +39,8 @@ class SourceMap {
   }
 
   async addMap(map, lineOffset = 0, columnOffset = 0) {
+    map = clone(map);
+
     if (!(map instanceof SourceMap) && map.version) {
       let consumer = await this.getConsumer(map);
 
--- a/src/SourceMap.js
+++ b/src/SourceMap.js
@@ -1,6 +1,5 @@
 const {SourceMapConsumer, SourceMapGenerator} = require('source-map');
 const lineCounter = require('./utils/lineCounter');
-const clone = require('clone');
 
 class SourceMap {
   constructor(mappings, sources) {
@@ -39,8 +38,6 @@ class SourceMap {
   }
 
   async addMap(map, lineOffset = 0, columnOffset = 0) {
-    map = clone(map);
-
     if (!(map instanceof SourceMap) && map.version) {
       let consumer = await this.getConsumer(map);
 
@@ -82,12 +79,14 @@ class SourceMap {
   }
 
   addMapping(mapping, lineOffset = 0, columnOffset = 0) {
-    mapping.generated = {
-      line: mapping.generated.line + lineOffset,
-      column: mapping.generated.column + columnOffset
-    };
-
-    this.mappings.push(mapping);
+    this.mappings.push({
+      source: mapping.source,
+      original: mapping.original,
+      generated: {
+        line: mapping.generated.line + lineOffset,
+        column: mapping.generated.column + columnOffset
+      }
+    });
   }
 
   addConsumerMapping(mapping, lineOffset = 0, columnOffset = 0) {


src/SourceMap.js
```diff
@@ -38,6 +39,8 @@ class SourceMap {
   }
 
   async addMap(map, lineOffset = 0, columnOffset = 0) {
+    map = clone(map);
```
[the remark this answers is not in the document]
DeMoorJasper: We could probably move it to before line 62, as the actual reuse of mappings happens on the lines after that as far as I know. I haven't tested moving it, but it should work in theory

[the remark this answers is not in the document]
DeMoorJasper: I'm not able to compare performance as I don't really have a project which is big enough to notice a difference

devongovett:
Tested perf on a larger app, and it was noticeable - about 5 seconds slower build time. Changed it to manually clone each mapping, and this reduced the time to about the same again.

DeMoorJasper:
Awesome, thanks for testing it out :)

src/SourceMap.js
```diff
@@ -79,12 +79,14 @@ class SourceMap {
   }
 
   addMapping(mapping, lineOffset = 0, columnOffset = 0) {
-    mapping.generated = {
-      line: mapping.generated.line + lineOffset,
-      column: mapping.generated.column + columnOffset
-    };
-
-    this.mappings.push(mapping);
+    this.mappings.push({
```
loganfsmyth: Not familiar with this API, but does this `mapping` have a `.name`? I see it used in `addConsumerMapping`, but not this one.

src/SourceMap.js
```diff
@@ -79,12 +79,14 @@ class SourceMap {
...
+      source: mapping.source,
+      original: mapping.original,
+      generated: {
+        line: mapping.generated.line + lineOffset,
+        column: mapping.generated.column + columnOffset
+      }
+    });
   }
 
   addConsumerMapping(mapping, lineOffset = 0, columnOffset = 0) {
```
loganfsmyth: FYI, the handling below here seems questionable. Mappings with no original location are still important because the terminate the end of a conceptual range over the generated file by saying that a generated location points at nothing. By discarding mappings with no original location, you're essentially force-extending every generated mapping to instead extend to the end of the line/next original-location mapping.

src/SourceMap.js
DeMoorJasper: Ow it should, thanks for noticing

src/SourceMap.js
loganfsmyth: Not quite. Sourcemaps are conceptually ranges over the generated code, but the actual mappings passed from the consumer are single points of data, from one generated line/col to an original line/col (optionally).

It's the fact that the original part is option that is critical here, and that is getting lost with the check in

```
     if (
       !mapping.source ||
       !mapping.originalLine ||
       (!mapping.originalColumn && mapping.originalColumn !== 0)
     ) {
       return;
     }
```

Imagine a generated piece of code like this:

    console.log

If I want to map _just_ the `console` to something in the original JS file, setting no original location on the `.log` part, I'd generate two mappings. One would say

```
{
  generated: {
    line: 0,
    column: 0,
  },
  original: {
    line: 0,
    column: 0,
  },
  source: "input.js",
  name: "console",
}
```
to say that the start of the word "console" (line 0, column 0) maps to the original file, also at line 0 column 0.

So far so good, but nothing sets an end position, so that mapping on its own would also apply to the `.log` part.

That's where non-`source` mappings come in. You'd also include a mapping as

```
{
  generated: {
    line: 0,
    column: 7,
  },
  original: null
  source: null,
  name: null,
}
```
to say that, starting at line 0, column 7, the code maps to nothing in the original file.

By discarding that information, you've essentially made the original line 0 column 0 mapping apply to the whole of `console.log`.

DeMoorJasper: Ow I never saw that as an issue but I'll definitely look into it.
Now that you've explained it, I can see it could and will cause issues.

Thank you

loganfsmyth: No problem, I'm happy to answer any other sourcemap questions if you have them. Babel has had it's fair share of questionable mapping logic over the ears too.

makmm:
when is this supposed to be fixed? i need this and it still breaks in     "parcel-bundler": "1.10.0-beta.1

DeMoorJasper:
@makmm it should’ve been fixed

QuantumInformation:
Still having this issue with new typescript project and fresh install of parcel

PR #1766, opened 2018-07-20
Support array in package.json's sideEffects property

### LINKED ISSUE #1566: Support fine grained sideEffects field in package.json
The current implementation of tree shaking supports `sideEffects: false`, but some libraries need more fine grained control to allow a subset of files to include side effects. webpack allows this by supporting an array of files which contain side effects.

```json
{
  "sideEffect": ["./foo.js", "./bar.js"]
}
```

For compatibility with libraries that depend on this, we should support it in Parcel as well.

ahocevar:
This pull request implements array support for the `package.json`'s `sideEffects` property, like webpack does. Example:
```json
"sideEffects": ["./a.js"]
```
Fixes #1566.

### DIFF
--- a/src/Asset.js
+++ b/src/Asset.js
@@ -125,7 +125,7 @@ class Asset {
 
   async getPackage() {
     if (!this._package) {
-      this._package = await this.getConfig(['package.json']);
+      this._package = await this.resolver.findPackage(path.dirname(this.name));
     }
 
     return this._package;
--- a/src/packagers/JSConcatPackager.js
+++ b/src/packagers/JSConcatPackager.js
@@ -141,8 +141,8 @@ class JSConcatPackager extends Packager {
     this.addedAssets.add(asset);
     let {js} = asset.generated;
 
-    // If the asset's package has the sideEffects: false flag set, and there are no used
-    // exports marked, exclude the asset from the bundle.
+    // If the asset has no side effects according to the its package's sideEffects flag,
+    // and there are no used exports marked, exclude the asset from the bundle.
     if (
       asset.cacheData.sideEffects === false &&
       (!asset.usedExports || asset.usedExports.size === 0)
--- a/src/scope-hoisting/hoist.js
+++ b/src/scope-hoisting/hoist.js
@@ -1,4 +1,6 @@
+const path = require('path');
 const matchesPattern = require('../visitors/matches-pattern');
+const mm = require('micromatch');
 const t = require('babel-types');
 const template = require('babel-template');
 const rename = require('./renamer');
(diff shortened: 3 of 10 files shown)


src/packagers/JSConcatPackager.js
```diff
@@ -23,6 +24,31 @@ const helpers =
...
+    case 'undefined':
+      return true;
+    case 'boolean':
+      return sideEffects;
+    case 'string':
+      if (process.platform === 'win32') {
+        sideEffects = sideEffects.replace(/\\/g, '/');
+      }
+      return mm.isMatch(
+        path.relative((await asset.getPackage()).pkgdir, asset.name),
```
fathyb: You can save some I/O and `async` overhead by setting `pkgdir` in `cacheData` preemptively in `hoist.js` like we do for `sideEffects` :
https://github.com/parcel-bundler/parcel/blob/7423541be5bb5974e13862190ba82899ebaa0367/src/scope-hoisting/hoist.js#L34-L38

src/packagers/JSConcatPackager.js
```diff
@@ -23,6 +24,31 @@ const helpers =
     .readFileSync(path.join(__dirname, '../builtins/helpers.js'), 'utf8')
     .trim() + '\n';
 
+async function hasSideEffects(asset, sideEffects) {
```
fathyb: nit: defaulting `sideEffects` using `{sideEffects} = asset.cacheData` could make the `if` conditions using it cleaner (`await hasSideEffects(asset)`)

src/Asset.js
```diff
@@ -125,7 +125,7 @@ class Asset {
 
   async getPackage() {
     if (!this._package) {
-      this._package = await this.getConfig(['package.json']);
+      this._package = await this.resolver.findPackage(path.dirname(this.name));
```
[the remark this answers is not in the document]
ahocevar: Because `findPackage()` will add a `pkgdir´ property to the package, which is needed later for globbing the `sideEffects` array entries.

ahocevar:
@fathyb Thanks for the review. I have addressed all your comments.

src/packagers/JSConcatPackager.js
```diff
@@ -23,6 +24,25 @@ const helpers =
     .readFileSync(path.join(__dirname, '../builtins/helpers.js'), 'utf8')
     .trim() + '\n';
 
+function hasSideEffects(asset, {sideEffects} = asset.cacheData) {
```
devongovett: this function could probably go in `hoist.js`. That way `asset.cacheData.sideEffects` could always be a boolean computed using this function. Might be slightly faster, especially since we can cache the results.

src/Asset.js
```diff
@@ -126,6 +126,11 @@ class Asset {
   async getPackage() {
     if (!this._package) {
       this._package = await this.getConfig(['package.json']);
+      if (this._package) {
+        this._package.pkgdir = path.dirname(
+          await config.resolve(this.name, ['package.json'])
+        );
+      }
```
devongovett: I actually liked your original solution using the resolver. This one will result in two lookups.

ahocevar:
Thanks @devongovett for your review. I have addressed your comments as well now.

PR #1770, opened 2018-07-21
Support sideEffect: false with CommonJS

fathyb:
Part of the fixes for #1699.

We currently can't `require` a module with `sideEffects: false`. This PR adds a `$require$NAME` entry to `cacheData.import` so we can treat both `require` like it is a wildcard import.

### DIFF
--- a/src/scope-hoisting/hoist.js
+++ b/src/scope-hoisting/hoist.js
@@ -280,8 +280,9 @@ module.exports = {
     }
 
     if (t.isIdentifier(callee, {name: 'require'})) {
+      let source = args[0].value;
       // Ignore require calls that were ignored earlier.
-      if (!asset.dependencies.has(args[0].value)) {
+      if (!asset.dependencies.has(source)) {
         return;
       }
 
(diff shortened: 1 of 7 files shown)


src/scope-hoisting/hoist.js
```diff
@@ -296,9 +297,11 @@ module.exports = {
           p.isSequenceExpression()
       );
       if (!parent.isProgram() || bail) {
-        asset.dependencies.get(args[0].value).shouldWrap = true;
+        asset.dependencies.get(source).shouldWrap = true;
       }
 
+      asset.cacheData.imports['$require$' + source] = [source, '*'];
```
devongovett: This could potentially conflict with a real import, right? e.g. `import $require$foo from './foo'`

fathyb: hmm, in imports we would use `$id$import$$require$foo` as key :
```js
asset.cacheData.import = {
  "$require$foo": ['foo', '*'],
  "$id$import$$require$foo": ['./foo', 'default']
}
```

or is there another place where it would conflict?

devongovett: oh right...

rafaelrinaldi:
Awesome work, thank you for your efforts.

PR #1775, opened 2018-07-23
Add support for vue-i18n single-file components

iptq:
Hi! I'm in a project where we use `vue-i18n` for localization. In particular, we're taking advantage of the single-file feature where you can include a `<i18n>` tag in a particular component rather than using global localization sheets. Typically, with webpack, you would use `vue-i18n-loader`, [as documented here](https://kazupon.github.io/vue-i18n/guide/sfc.html#installing-vue-i18n-loader), to inject the `<i18n></i18n>` tags into the Vue component.

After digging into the `vue-i18n` package a bit, it seems like it's able to detect the existence of an `__i18n` option on the component and automatically load the injected content, which is what [`vue-i18n-loader` is doing](https://github.com/kazupon/vue-i18n-loader/blob/dev/src/index.js). I figured since there isn't even a dependency problem, this could be implemented relatively easily into parcel as well.

Let me know if there's anything else that's needed or if this isn't the right place to discuss things. Thanks!

### DIFF
--- a/src/assets/VueAsset.js
+++ b/src/assets/VueAsset.js
@@ -30,6 +30,11 @@ class VueAsset extends Asset {
     let descriptor = this.ast;
     let parts = [];
 
+    let i18nBlock = descriptor.customBlocks.find(r => r.type === 'i18n');
+    if (i18nBlock) {
+      this.i18nData = JSON.parse(i18nBlock.content);
+    }
+
     if (descriptor.script) {
       parts.push({
         type: descriptor.script.lang || 'js',
@@ -95,6 +100,15 @@ class VueAsset extends Asset {
         ${optsVar} = ${optsVar}.options;
       }
     `;
+
+    if (this.i18nData) {
+      supplemental += `
+        ${optsVar}.__i18n = ${optsVar}.__i18n || [];
+        ${optsVar}.__i18n.push(${JSON.stringify(
+        JSON.stringify(this.i18nData)
+      )});
+      `;
+    }
 
     supplemental += this.compileTemplate(generated, scopeId, optsVar);
     supplemental += this.compileCSSModules(generated, optsVar);


evanleck:
I was _just_ looking for this last night!

devongovett:
This seems kinda specific to a particular library. Should we be hard coding things like that here or is there a better way? How does vue-loader handle stuff like this?

iptq:
To be honest, I'm not too familiar with webpack or parcel's bundling systems, but from a couple hours of just prodding through code it seems to me like `vue-loader` allows you to load additional loaders that get run somewhere in between the many passes.

On the other hand, it seems like the burden of catering to the service should be placed upon parcel, being a zero-configuration bundler and all. This does raise some concerns about trying to support too many features, but for this i18n package in particular, there's not too many changes that need to be made since the library itself is already very flexible in this regard (supporting the `__i18n` option)

iptq:
I figure the other option would be for parcel to actually have an optional configuration file for "anything else you want to run" in the form of npm packages. I just don't want it to be _impossible_ to use this feature if my project is using parcel.

evanleck:
Would it be possible to put this in a plugin? Sorry, I'm not at all familiar with the internals of Parcel right now.

evanleck:
Bump? Seems relatively innocuous, though I understand the argument that it's quite specific to a single library.

devongovett:
Going to close this. I don't think it's a good idea to support a library specific transform in Parcel core.

That said, I do think it should be possible to add custom block transforms in Parcel via a Vue specific config file. Perhaps the Vue community can come up with a format for a `.vuerc` file which could define transforms for custom blocks and work across multiple tools (e.g. Webpack, Parcel, Browserify, etc.). Ideally that would be part of the `@vue/component-compiler-utils` package which Parcel and others already use to compile `.vue` files, rather than something bundler specific.

cc. @ yyx990803

PR #1788, opened 2018-07-25
Support html5 history mode routing

snaptopixel:
While developing my first Parcel project I noticed that html5 history urls/routes were not working when reloading the page or changing the url manually. This is due to a default setting in `serve-static` which redirects directory urls to `/`. I made this simple change to my locally installed parcel-bundler and all works as expected 🎉

### DIFF
--- a/src/Server.js
+++ b/src/Server.js
@@ -36,6 +36,7 @@ function enableCors(res) {
 function middleware(bundler) {
   const serve = serveStatic(bundler.options.outDir, {
     index: false,
+    redirect: false,
     setHeaders: setHeaders
   });
 


snaptopixel:
@DeMoorJasper this fixes SPAs 😄 

_Without_ this change, if you are using html5 history mode in an SPA and navigate to a route like:

```
http://localhost:1234/some/route
```

If you refresh your browser, or manually change the url, it will always redirect you to `localhost:1234` losing the `some/route` part of the url.

This change will simply keep the url in place. It will still load the root `index.html` for the SPA and all should work as expected.

DeMoorJasper:
Ow alright, didn't expect that to happen from reading the docs. Thanks for this

PR #1794, opened 2018-07-25
Fix security vulnerability

### LINKED ISSUE #1783: A vulnerability found in parcel-bundler
Hi, I found a vulnerability in parcel-bundler, how do I report it to you?

DeMoorJasper:
After having a short discussion with @chromium1337 we found a fix for the security vulnerability #1783 

The vulnerability mainly had to do with people being able to steal your code as the origin of requests wasn't checked by websocket server.

However the CORS header in the static server comes down to the same vulnerability and as there is no general hostname flag, we can't really secure that server.

Unless we would agree, that both these hostnames would always be the same and deprecate `--hmr-hostname` and create a `--hostname` flag instead that adds a cors limitation to both the static and websocket server. This way allowing parcel to be run on a server without having to worry about any security risks (related to CORS).

Fixes #1783


DeMoorJasper:
@chromium1337 Wrote a page about it explaining why it's an issue and origin would fix this. I'll DM you the page

PR #1803, opened 2018-07-26
Fix issue with duplicate ids

DeMoorJasper:
Although very unlikely it was possible to have duplicate ids.

This PR makes the bundler check for a duplicate and pick a new id if there is a duplicate.

Only necessary on production builds as development uses relative paths and scope hoisting minifies everything anyways


gregpardo:
Was having a TypeError issue and this PR seems to fix it for me. Was only happening in production.

devongovett:

Issue #1808, opened 2018-07-27
Can resolve font files but wrong src in the generated css?


slmjkdbtl:
# 🐛 bug report

Generated css have the wrong src to the font file

# config 

```js
const Bundler = require("parcel-bundler");

const entries = [
	"./public/scripts/main.js",
	"./public/styles/main.scss"
]

const bundler = new Bundler(entries, {

	outDir: "./public/dist",
	publicUrl: "/dist",
	cache: process.env.NODE_ENV !== "production",
	minify: process.env.NODE_ENV === "production",
	watch: process.env.NODE_ENV !== "production"

});

bundler.bundle()
```
.
├── dist
│   ├── SourceHanSans.daafd6fb.otf
│   ├── scripts
│   │   ├── main.js
│   │   └── main.map
│   └── styles
│       ├── main.css
│       └── main.map
├── fonts
│   └── SourceHanSans.otf
├── scripts
│   └── main.js
└── styles
    ├── global.scss
    └── main.scss

## 😯 Current Behavior

the font could be resolved, but the generated css has `src: url("SourceHanSans.daafd6fb.otf")`, but the generated css file is in the `styles` directory and the font file is in the root `dist` directory

## 💻 Code Sample

```scss
@font-face {
    font-family: SourceHanSans;
    src: url("../fonts/SourceHanSans.otf")
}
```

## 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Parcel           | 1.9.7
| Node             | 10.7.0
| npm/Yarn         | 1.7.0
| Operating System | MacOS 10.13.6

Issue #1592, opened 2018-06-20
css assets path not correct


slmjkdbtl:
# 🐛 bug report

I'm using a font with `@font-face` and `url` in my scss, the css file generated are in `/dist/styles/main.css`, and the font file is in `/dist/font.ttf`, and the generated css file has the wrong `url` of the font file. Also the build gets stucked at building `processing.scss` forever, but if i commented other entries all out, `processing.scss` gets successfully bundled. Also is it possible to put all bundled files in one big folder when the entries are from different folders (instead of creating `scripts` and `styles` subdirectories in `dist`)

## 🎛 Configuration (.babelrc, package.json, cli command)

```js
const Bundler = require("parcel-bundler");

const entries = [
	"./public/scripts/main.js",
	"./public/styles/main.scss",
	"./public/scripts/track.js",
	"./public/scripts/md.js",
	"./public/styles/md.scss",
	"./public/styles/resume.scss",
	"./public/styles/processing.scss",
	"./public/styles/links.scss"
]

const bundler = new Bundler(entries, {

	outDir: "./public/dist",
	publicUrl: "/dist",
	cache: process.env.NODE_ENV !== "production",
	minify: process.env.NODE_ENV === "production",
	watch: process.env.NODE_ENV !== "production"

});

bundler.bundle()
```

## 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Parcel           | 1.9.2
| Node             | 10.4.1
| npm/Yarn         | 1.7.0
| Operating System | macOS High Sierra 10.13.4

slmjkdbtl:
possible duplicate of #1303

mischnic:
Is this still an issue? Can you provide a reproduction?

mischnic:
> I'm using a font with `@font-face` and `url` in my scss, the css file generated are in `/dist/styles/main.css`, and the font file is in `/dist/font.ttf`, and the generated css file has the wrong `url` of the font file.

This part is described in issue #1801.

PR #1809, opened 2018-07-27
Change Asset#addURLDependency to include publicURL

### LINKED ISSUE #1801: Wrong relative path to URL assets when bundle lives in subdirectory
# 🐛 Bug Report

When two bundles in different directories with the same file name are generated, the directory structure is reflected in the output. This causes an issue with URL dependencies because their paths are relative to the current directory.

## 🎛 Configuration

Given the following directory structure:

```
assets
├── a
│   └── bundle.css
├── b
│   └── bundle.css
└── img
    └── foo.png
```

With the contents of `assets/a/bundle.css` being:

``` css
body {
    background-image: url('../img/foo.png');
}
```

And the following Parcel command:

```
parcel build --no-cache --no-source-maps 'assets/*/bundle.css'
```

The following directory tree is correctly generated:

```
dist
├── a
│   └── bundle.css
├── b
│   └── bundle.css
└── foo.d7b28491.png
```

So far so good. However...

## 🤔 Expected Behavior

I would expect the contents of `dist/a/bundle.css` to be either:

``` css
body{background-image:url(../foo.d7b28491.png)}
```

Or:

``` css
body{background-image:url(/foo.d7b28491.png)}
```

## 😯 Current Behavior

The contents of `dist/a/bundle.css` is currently:

``` css
body{background-image:url(foo.d7b28491.png)}
```

Which means the browser will request `/a/foo.png`, but the file is located at the web root – not in the "a" directory.

## 💁 Possible Solution

I'm really not familiar enough with Parcel to make an educated recommendation as to how is should be fixed, but I'm guessing the best way to fix the issue is by somehow making calculating the relative path to the URL asset from the bundle. So that (using the example code above), the output URL would be `url(../foo.[md5].png)`

Another possible solution is to join the URL with the publicURL option. This could be achieved by modifying the `Asset#addURLDependency()` method's return value, like so:

``` diff
+ const urlJoin = require('./utils/urlJoin');

...

- return URL.format(parsed);
+ return urlJoin(this.options.publicURL, URL.format(parsed));
```

This method seems to work, but as said before, I'm not familiar enough with Parcel to know if it's a viable solution. It does require modifying some tests, which indicates it could be a breaking change. Would love some feedback.

I have [made a commit](https://github.com/kalfheim/parcel/commit/b2233dc610576684b63728dcbb73ae63f42e673f) and will definitely create a PR if the solution is approved.

## 🌍 Your Environment

| Parcel           | 1.9.7
| Node             | 10.7.0
| npm/Yarn         | Yarn 1.7.0
| Operating System | macOS 10.13.6

kalfheim:
Fixes #1801 #1808 #1592

This PR changes the return value of Asset#addURLDependency() to be prefixed with the publicURL option.


test/asset.js
```diff
@@ -40,6 +40,7 @@ describe('Asset', () => {
     const bundleName = 'xyz';
     const options = {
       rootDir: '/root/dir',
+      publicURL: '/',
```
DeMoorJasper: isn't `/` the default?

devongovett:
I think it already gives a relative path. This makes the output more portable depending on where you move it to. Why does this not work?

Issue #1477, opened 2018-06-01
Import directive within GraphQL files


hmaurer:
Hi!

I am opening this issue to ask whether the addition of an import directive in GraphQL (.graphql) files has been considered and/or is currently being worked on. To be more specific, I am referring to the `#import` syntax [used by `webpack-graphql-loader`](https://www.npmjs.com/package/webpack-graphql-loader#import-statements-in-graphql-files), like so:

```
#import "./fragments.graphql"
 
query {
  ...a
  ...b
}
```

[The implementation by `webpack-graphql-loader`](https://github.com/samsarahq/graphql-loader/blob/master/src/loader.ts#L71) is a little dirty but pragmatic (they essentially do a search-and-replace for "#import" directives). 

Since this is also the [recommended approach in Apollo's documentation](https://www.apollographql.com/docs/react/advanced/fragments.html#webpack-importing-fragments), I think that it would be a useful addition to Parcel.

I did not open this issue as a feature request because if this is not something currently being worked on I would be happy to contribute myself.

DeMoorJasper:
Haven't seen anyone working on it.
Would be nice if you could implement it :)

hmaurer:
HI @agonbina ! No, sorry :( I moved away from `.graphql` files to defining the queries/fragments in my `.js` files and so I didn't have a pressing need for this feature. I do agree it seems pretty much essential when working with `.graphql` files though but I don't have the time to implement it right now.

agonbina:
There is a workaround to get this to work and still keep your fragments in `.graphql` files. Here is an example with `react-apollo`
```
import { concatAST } from 'graphql'
import { graphql } from 'react-apollo'

import customerFragment from './customer.fragment.graphql'
import customerQuery from './customer.graphql'

export const withCustomer = graphql(concatAST([customerFragment, customerQuery]), {
  name: 'customerQuery'
})
```
It is more cumbersome but on the other hand you don't need to use the `#import` directive which is not party of the graphql spec anyway. And the visual studio code extension for GraphQL is able to find and understand the fragments that you use in your queries/mutations.

nabiltntn:
I created PR https://github.com/parcel-bundler/parcel/pull/1817 to add this feature

PR #1817, opened 2018-07-29
Use graphql-tag webpack loader code to allow using #import for fragment

nabiltntn:
Allow using the **#import** directive to import fragments inside **.graphql** files

I reused the [code of graphql-tag webpack loader](https://github.com/apollographql/graphql-tag/blob/master/loader.js)  that parses **.graphql** files to generate modules that include fragments with queries definitions.

This PR addresses issue https://github.com/parcel-bundler/parcel/issues/1477


src/assets/GraphqlAsset.js
```diff
@@ -9,11 +46,141 @@ class GraphqlAsset extends Asset {
...
+          outputCode += `
+          module.exports["${opName}"] = oneQuery(doc, "${opName}");
+          `;
+        }
+      }
+    }
+
+    const importOutputCode = expandImports(code, doc);
+    const allCode =
+      headerCode + os.EOL + importOutputCode + os.EOL + outputCode + os.EOL;
```
DeMoorJasper: You can probably just use `\n`, windows should be able to handle that as well

nabiltntn:
I removed the copied code and i used `loader` function directly from 'graphql-tag' package.

DeMoorJasper:
Yeah I saw you changed it, not sure if I like adding a webpack plugin as a dependency and using something internal that is outside the scope of semantic versioning

PR #1823, opened 2018-07-30
[WIP] Resolve imports from paths also specified in LESS config

### LINKED ISSUE #1818: [LESSAsset] Respect included paths in LESS config
# 🐛 bug report

When specifying a config for LESS compiler/renderer, the `paths` property in the config is not respected/resolved the way LESS does itself. You can specify the property to lessc with `--include-path=path1:path2:path3` or via the config (.lessrc / .lessrc.js / package.json key "less"): `{ "paths": ["path1", "path2", "path3"] }`

## 🤔 Expected Behavior

When you have included a path in the config that is outside of entry file or `rootDir`, you would want to search/visit those directories - like LESS does.

When running: `lessc --include-path=src/:node_modules/semantic-ui-less/ src/index.less` I get the following:

```
FileError: 'theme.less' wasn't found. Tried - src/theme.less,node_modules/semantic-ui-less/the
me.less,theme.less in src/index.less on line 1, column 1:
1 @import 'theme.less';
```

## 😯 Current Behavior

LESSAsset doesn't respect the `paths` property in the config, so when I run Parcel I get:

```
×  src/index.less: Cannot find module './theme.less' from 'src'
```

## 💁 Possible Solution

The solution to this problem, would be to change the `loadFile` function in `LESSAsset.js` and add the `options` parameter (which is the LESS option/config, containing the LESS config object retrieved earlier).

```
async loadFile(filename, currentDirectory) {
    filename = parseCSSImport(filename);
    let resolved = await resolver.resolve(
    filename,
    path.join(currentDirectory, 'index')
    );
    return {
    contents: await fs.readFile(resolved.path, 'utf8'),
    filename: resolved.path
    };
}
```

## 🔦 Context

Semantic UI has this weird way of merge'ing default themes and variable files with customized, by importing files relatively and depending on paths from the upper context. That's why you have to tell LESS to search for those files in another directories, because they don't exist from relative location where they are imported. See image below.

## 🌍 Your Environment

| Parcel           | 1.9.7
| Node             | 10.7.0
| npm/Yarn         | 1.7.0
| Operating System | Win10

gaui:
Use similar file resolution strategy as [LESS FileManager](https://github.com/less/less.js/blob/v2.7.3/lib/less-node/file-manager.js#L23)

Fixes #1818


gaui:
Might have to modify a little, it's based on the resolution strategy from [2.7.3](https://github.com/less/less.js/blob/v2.7.3/lib/less-node/file-manager.js#L23), but the version in 3.0.1 (which Parcel uses) uses a [slightly modified version](https://github.com/less/less.js/blob/master/lib/less-node/file-manager.js#L19).

src/assets/LESSAsset.js
```diff
@@ -80,17 +83,65 @@ function getFileManager(less, options) {
...
+      }
+
+      if (!path.isAbsolute(filename) && !paths.includes('.') === -1) {
+        paths.push('.');
+      }
+
+      // promise is guaranteed to be asyncronous
+      // which helps as it allows the file handle
+      // to be closed before it continues with the next file
+      return new Promise(function(resolve, reject) {
```
DeMoorJasper: Why are you returning a new Promise here?

Async functions are just functions that automatically return a promise already.

src/assets/LESSAsset.js
```diff
@@ -80,17 +83,65 @@ function getFileManager(less, options) {
...
+        (async function tryPathIndex(i) {
```
DeMoorJasper: You can probably split this up into a seperate function and await it instead of wrapping it in a promise in a IFFE.
It would make the code more readable and probably smaller

Something like this
```Javascript
const tryPathIndex = async i => {
  // the code of the function...
}
return await tryPathIndex(0);
```

src/assets/LESSAsset.js
```diff
@@ -80,17 +83,65 @@ function getFileManager(less, options) {
...
+                resolve(resolution);
+              } else {
+                throw new Error();
+              }
+            } catch (err) {
+              filenamesTried.push(fullFilename);
+              tryPathIndex(i + 1);
+            }
+          } else {
+            reject({
```
DeMoorJasper: You can just use await and use `throw new Error(...)` or `throw {// the object}` in the catch section

DeMoorJasper:
Apparently this also breaks existing less tests, this should probably not happen

devongovett:

johanberonius:
History got replaced so it is no longer possible to see the change proposed by this PR.
I posted another possible fix as a comment to #1818

PR #1827, opened 2018-07-31
POC: adding tests for custom html processing and cascading plugins

tripodsan:
I created this PR to illustrate the problems I had trying to solve "cascading plugins" in #1828.
I found now a solution to this problem, but it's not "nice" as it messes with the HTMLAsset behavior, such as it generates the new rendition already in generate and not in postProcess:
parcel-bundler/parcel/pull/1827/files#diff-ac1519a4d6f8aff0627d99d7a204acc4R12

I wanted to ask the devs if this is the correct approach to do this.


DeMoorJasper:
Could you add a description, describing what this does and why.

It's easier for us to understand the problem that way.

PR #1829, opened 2018-07-31
Show warning message when entry file not provided (#1797)

### LINKED ISSUE #1797: Should default to index.html, or show an error
# 🙋 feature request

## 🤔 Expected Behavior

parcel should default to using index.html, or error out.

## 😯 Current Behavior
1. I have a project with index.html
1. I type "parcel" not "parcel index.html"
1. parcel starts up happily
1. but when i load the site, i get nothing.

## Expected
1. I have a project with index.html
1. I type "parcel" not "parcel index.html"

either
* parcel starts up happily by using index.html by default.
or
* parcel throws an error.

rajikaimal:
Fixes #1797

### DIFF
--- a/src/Bundler.js
+++ b/src/Bundler.js
@@ -79,6 +79,8 @@ class Bundler extends EventEmitter {
 
     // If no entry files provided, resolve the entry point from the current directory.
     if (!entryFiles || entryFiles.length === 0) {
+      // Show warning message if arguments are not passed to command
+      logger.warn('Entry file not provided explicitly');
       entryFiles = [process.cwd()];
     }
 
@@ -301,6 +303,7 @@ class Bundler extends EventEmitter {
       let buildTime = Date.now() - startTime;
       let time = prettifyTime(buildTime);
       logger.success(`Built in ${time}.`);
+
       if (!this.watcher) {
         bundleReport(this.mainBundle, this.options.detailedReport);
       }


DeMoorJasper:
They are still checking it, but it shouldn't throw a warning. Parcel can be used without entrypoint it's optional, it should just throw a user friendly error if it can't find one.

mohsen1:
@rajikaimal I think @DeMoorJasper means `package.json` main field, not a file named `pkg.json` :)

rajikaimal:
@mohsen1 Yeah that's what I meant by `pkg.json`. 

@DeMoorJasper Isn't a warning log the best for this scenario, rather than an error log?

DeMoorJasper:
Users of parcel are allowed to use non explicit entrypoints at their own risk, we can log the entrypoints if that's really necessary.

But the most important thing is if parcel can't resolve an entrypoint it should throw a nice error, otherwise users think it's a parcel bug or something. Because users expect too much magic.

jchiatt:
@DeMoorJasper Is this what you're looking for?

~~~js
 for (let entry of this.entryFiles) {
          try {
            let asset = await this.resolveAsset(entry);
            this.buildQueue.add(asset);
            this.entryAssets.add(asset);
          } catch (err) {
            const errMsg = `Failed to load entrypoint. ${err}`
            logger.error(errMsg);
            process.exit(1);
          }
        }
~~~

DeMoorJasper:
Sort of the process.exit seems a bit hacky to me though as in the catch of build we already have something similar

rajikaimal:
@jchiatt Hey, did you test the code with `try-catch`? This didn't resolve the bug for me.

DeMoorJasper:
@rajikaimal what bug?

rajikaimal:
Execution doesn't go to catch block at all.

DeMoorJasper:
That’s strange than it can resolve it right?

DeMoorJasper:
I've opened a PR with what I had in mind, see #1848

Issue #1422, opened 2018-05-23
Failed to resolve existing module `rmc-steps`


FourLeafClover:
![image](https://user-images.githubusercontent.com/8449788/40405548-9ee27680-5e8f-11e8-84ea-4ade57760fde.png)

lordGuan:
you may need babel in your project

FourLeafClover:
@lordGuan   sames, where  have some problems when i import antd-mobile。 I have npm install rmc-steps  

![image](https://user-images.githubusercontent.com/8449788/40406445-466baad6-5e93-11e8-9a13-85f1833d3542.png)

DeMoorJasper:
Object spread is ES2018 not ES6

FourLeafClover:
@DeMoorJasper  I added the babel。 but there have another problem。please see the previous image。

DeMoorJasper:
If the dependency is installed I'm not sure why it can't get resolved

DeMoorJasper:
An example repo or a bit more info would be helpfull. There is a bug template if you file an issue, would be nice if you use that to reformat the issue :)

fathyb:
@FourLeafClover did you try with `--no-cache` or remove the `.cache` directory?

cncolder:
Example here:

package.json
```json
{
  "dependencies": {
    "antd-mobile": "^2.2.1"
  }
}
```

index.html
```html
<html>
<body>
    <script src="./index.jsx"></script>
</body>
</html>
```

index.jsx
```jsx
import Steps from "antd-mobile/es/steps";
```

`parcel index.html`
```bash
parcel-antd-mobile/node_modules/antd-mobile/es/steps/index.js:7:20: Cannot resolve dependency 'rmc-steps'
   5 | import _inherits from 'babel-runtime/helpers/inherits';
   6 | import React from 'react';
>  7 | import RcSteps from 'rmc-steps';
     |                     ^
   8 | import Icon from '../icon';
   9 |
  10 | var Steps = function (_React$Component) {
```

The `rmc-steps` stay there. But parcel just not seen.

`ls node_modules/rmc-steps`
```bash
HISTORY.md   LICENSE.md   README.md    assets       lib          package.json
```

cncolder:
rmc-steps has a mistake field `"module": "./es/index"` in `package.json`. It not bundle es code into npm.

[...]

PR #1835, opened 2018-08-02
Enable Object Rest Spread by default

### LINKED ISSUE #1221: can`t user es7 destructuring assignments

mian.js
```js
var { a, b, ...other } = { a: 1, b: 2, c: 3, d: 4 }
```

.babelrc
```js
{
    "presets": [
        "es2015",
        "react"
    ]
}
```
### 😯 Current Behavior
parcel ./main.js

### 🌍 Your Environment

| Parcel           | 1.7.1
| Node             | 8.11.1
| npm/Yarn         | npm 5.6.0
| Operating System | windows10

arv:
This uses Babel's plugin:
https://babeljs.io/docs/en/babel-plugin-transform-object-rest-spread

Spec:
https://tc39.github.io/ecma262/#sec-object-initializer-runtime-semantics-propertydefinitionevaluation

Fixes #1221, #1422

### DIFF
--- a/src/transforms/babel.js
+++ b/src/transforms/babel.js
@@ -265,6 +265,14 @@ async function getEnvPlugins(targets, useBuiltIns = false) {
     {},
     {targets, modules: false, useBuiltIns: useBuiltIns ? 'entry' : false}
   ).plugins;
+
+  // babel-preset-env version 6.x does not cover object-rest-spread so always
+  // add it.
+  plugins.push([
+    require('babel-plugin-transform-object-rest-spread'),
+    {useBuiltIns}
+  ]);
+
   envCache.set(key, plugins);
   return plugins;
 }
--- a/test/javascript.js
+++ b/test/javascript.js
@@ -28,6 +28,21 @@ describe('javascript', function() {
     assert.equal(output.default(), 3);
   });
 
+  it('should produce a basic JS bundle with object rest spread support', async function() {
+    let b = await bundle(__dirname + '/integration/es6/object-rest-spread.js');
+
+    assert.equal(b.assets.size, 1);
+
+    let output = await run(b);
+    assert.equal(typeof output, 'object');
+    assert.equal(typeof output.default, 'function');
+
+    let res = output.default();
+    assert.equal(res.y, 'a');
+    assert.deepEqual(res.z, {y: 'a', b: 'b'});
+    assert.deepEqual(res.ys, {b: 'b'});
+  });
+
   it('should bundle node_modules on --target=browser', async function() {
     let b = await bundle(__dirname + '/integration/node_require/main.js', {
       target: 'browser'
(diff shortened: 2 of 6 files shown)


DeMoorJasper:
This is not a standard yet, not sure if this is a good idea

arv:
Now I need to find out why it passes locally but not on the Travis/AppVeyor

DeMoorJasper:
@arv you still have to add it as a dependency if I'm not mistaken. That's probably the reason

arv:
@DeMoorJasper I need to figure out the test failures. Any ideas?

DeMoorJasper:
@arv Unfortunately I don't have any clue why it's failing

arv:
I can look at it tomorrow.

test/javascript.js
```diff
@@ -28,6 +28,21 @@ describe('javascript', function() {
     assert.equal(output.default(), 3);
   });
 
+  it('should produce a basic JS bundle with object rest spread support', async function() {
+    let b = await bundle(__dirname + '/integration/object-rest-spread/object-rest-spread.js');
```
DeMoorJasper: It wasn't bundling anything, got it it bundle correctly in the test now, not sure how this worked locally. Now it still throws an error though.

Hope you can resolve it :)

src/transforms/babel.js
```diff
@@ -265,6 +265,14 @@ async function getEnvPlugins(targets, useBuiltIns = false) {
     {},
     {targets, modules: false, useBuiltIns: useBuiltIns ? 'entry' : false}
   ).plugins;
+
+  // babel-preset-env version 6.x does not cover object-rest-spread so always
```
arv: This does not seem like the right place for this but I'm not sure where would be a better place to put this?

DeMoorJasper: It should be fine here, but than you should probably remove it from `JSAsset.js` as now it probably runs it twice...

DeMoorJasper: On sidenote though, this would also run over node_modules, not sure if that's intended

arv: > It should be fine here, but than you should probably remove it from JSAsset.js as now it probably runs it twice...

The one in `JSAsset.js` seems to only be used for parser options. Inspecting the plugins array we get from `getEnvPlugins` and it does not contain `babel-plugin-transform-object-rest-spread`

> On sidenote though, this would also run over node_modules, not sure if that's intended

That is what I want since npm modules are now starting to use this feature.

DeMoorJasper: Yes, but babylon runs before the other transforms to create the ast, so this would run twice, at different places, but it'll still run twice

arv: That makes sense. Thanks.

arv:
Thanks

danielo515:
Is this released ?

PR #1845, opened 2018-08-03
Fix glob import/require in Stylus assets

### LINKED ISSUE #1816: Can't import Stylus modules with file globbing
# 🐛 bug report

Although the Stylus binary compiles just fine files containing glob imports, it breaks when using with Parcel.

```
parcel serve frontend/index.html --no-autoinstall --no-hmr
```

## 🤔 Expected Behavior

Parcel should serve a CSS file compiled from Stylus, with or without glob imports.

## 😯 Current Behavior

```
ENOENT: no such file or directory, open '/app/frontend/styles/**/*
```

## 💻 Code Sample

```stylus
// Modules
@require '**/*'
```

## 🌍 Your Environment

| Parcel           | 1.9.7
| Node             | 10.5.0
| npm/Yarn         | 1.7.0
| Operating System | Docker (Debian)

prokopyl:
As said in #1816, Parcel currently overrides the Stylus resolver with its own, namely to support importing from `node_modules` and root imports. However, it currently fails to resolve glob imports, trying to resolve to files literally named `**/*` (which obviously don't exist, throwing `ENOENT`).

This PR works around the issue by detecting glob imports in the ahead-of-time Stylus `ImportVisitor`, applying `fast-glob` on the given path, and then associating multiple paths to the import instead of a single one.
During evaluation time, it will then replace the glob import statement with multiple ones (one for each resolved file) before passing them to Stylus, and then merging all the resulting node blocks into one.

This PR also adds a test case for Stylus glob import, to prevent future regressions.

This should fix #1816.

(Comments are obviously welcome, considering I don't really know the Parcel codebase there might be issues here and there.)

### DIFF
--- a/src/assets/StylusAsset.js
+++ b/src/assets/StylusAsset.js
@@ -3,7 +3,9 @@ const Asset = require('../Asset');
 const localRequire = require('../utils/localRequire');
 const Resolver = require('../Resolver');
 const fs = require('../utils/fs');
-const {dirname} = require('path');
+const {dirname, resolve, relative} = require('path');
+const isGlob = require('is-glob');
+const glob = require('fast-glob');
 
 const URL_RE = /^(?:url\s*\(\s*)?['"]?(?:[#/]|(?:https?:)?\/\/)/i;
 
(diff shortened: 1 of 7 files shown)


src/assets/StylusAsset.js
```diff
@@ -80,7 +82,25 @@ async function getDependencies(
       let path = imported.path.first.string;
 
       if (!deps.has(path)) {
-        deps.set(path, resolver.resolve(path, filepath));
+        if (isGlob(path)) {
+          deps.set(
```
DeMoorJasper: Probably better to do something like this:

``````

Comment (DeMoorJasper) [inline src/assets/StylusAsset.js]: Probably better to do something like this:

```Javascript
let globDeps = await glob(resolve(dirname(filepath), path), {
              onlyFiles: true
            }
```

DeMoorJasper:
Awesome, thanks for this!

PR #1848, opened 2018-08-03
Print user friendly error if no entrypoint can be found

### LINKED ISSUE #1797: Should default to index.html, or show an error

DeMoorJasper:
Print a user friendly error message if no entrypoint can be found.
Also prints a user friendly error if a user defined entrypoint has a resolve error

Closes #1797 Closes #1829

### DIFF
--- a/src/Bundler.js
+++ b/src/Bundler.js
@@ -144,9 +144,7 @@ class Bundler extends EventEmitter {
           ? options.contentHash
           : isProduction,
       throwErrors:
-        typeof options.throwErrors === 'boolean'
-          ? options.throwErrors
-          : process.env.NODE_ENV === 'test'
+        typeof options.throwErrors === 'boolean' ? options.throwErrors : true
     };
   }
 
@@ -252,7 +250,11 @@ class Bundler extends EventEmitter {
             this.buildQueue.add(asset);
             this.entryAssets.add(asset);
           } catch (err) {
-            throw new Error(`Failed to load entrypoint: ${entry}!`);
+            throw new Error(
+              `Failed to load entrypoint: "${entry}" in "${
+                this.options.rootDir
+              }"`
+            );
           }
         }
 
--- a/src/Bundler.js
+++ b/src/Bundler.js
@@ -251,7 +251,7 @@ class Bundler extends EventEmitter {
             this.entryAssets.add(asset);
           } catch (err) {
             throw new Error(
-              `Failed to load entrypoint: "${entry}" in "${
+              `Cannot resolve entry "${entry}" from "${
                 this.options.rootDir
               }"`
             );
@@ -259,7 +259,7 @@ class Bundler extends EventEmitter {
         }
 
         if (this.entryAssets.size === 0) {
-          throw new Error('No entrypoint(s) found.');
+          throw new Error('No entries found.');
         }
 
         initialised = true;
--- a/src/Bundler.js
+++ b/src/Bundler.js
@@ -329,10 +329,7 @@ class Bundler extends EventEmitter {
         this.hmr.emitError(err);
       }
 
-      if (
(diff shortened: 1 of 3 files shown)


src/Bundler.js
```diff
@@ -318,8 +329,9 @@ class Bundler extends EventEmitter {
         this.hmr.emitError(err);
       }
 
-      if (process.env.NODE_ENV === 'production') {
-        process.exitCode = 1;
+      if (process.env.NODE_ENV === 'production' || !initialised) {
+        await this.stop();
+        process.exit(1);
```
DeMoorJasper: Apparently `process.exitCode = 1;` didn't exit the process anymore for some reason, so changed this to be a lil more aggresive and added `await this.stop();` to make sure servers and workers shutdown properly

src/Bundler.js
```diff
@@ -242,10 +247,20 @@ class Bundler extends EventEmitter {
...
         for (let entry of this.entryFiles) {
-          let asset = await this.resolveAsset(entry);
-          this.buildQueue.add(asset);
-          this.entryAssets.add(asset);
+          try {
+            let asset = await this.resolveAsset(entry);
+            this.buildQueue.add(asset);
+            this.entryAssets.add(asset);
+          } catch (err) {
+            throw new Error(`Failed to load entrypoint: ${entry}!`);
```
fathyb: What do you think of a more precise error message, something like: `Failed to resolve entrypoint "${entry}" in "${rootDir}"`

Given that we infer `rootDir` I think it'd be better to give users more info so they can debug why we can't find their entrypoint.

src/Bundler.js
```diff
@@ -142,7 +142,11 @@ class Bundler extends EventEmitter {
       contentHash:
         typeof options.contentHash === 'boolean'
           ? options.contentHash
-          : isProduction
+          : isProduction,
+      throwErrors:
+        typeof options.throwErrors === 'boolean'
+          ? options.throwErrors
+          : process.env.NODE_ENV === 'test'
```
fathyb: I think we should default this to `false` in CLI and `true` when using the `Bundler` API.

src/Bundler.js
DeMoorJasper: Sure thing, I'll update it

src/Bundler.js
DeMoorJasper: Probably, I'll update it

DeMoorJasper: Would this be considered a breaking change?

DeMoorJasper: Just something I just thought about, parcel already throws errors in the api using the event handler.
So this might cause issues as the promise of bundle only returns on initial build. What do you think?

src/Bundler.js
```diff
@@ -242,10 +245,24 @@ class Bundler extends EventEmitter {
...
-          let asset = await this.resolveAsset(entry);
-          this.buildQueue.add(asset);
-          this.entryAssets.add(asset);
+          try {
+            let asset = await this.resolveAsset(entry);
+            this.buildQueue.add(asset);
+            this.entryAssets.add(asset);
+          } catch (err) {
+            throw new Error(
+              `Failed to load entrypoint: "${entry}" in "${
```
devongovett: I would say "Cannot resolve entry ${entry}". As of the error we haven't actually tried to load anything yet, just resolved.

PR #1851, opened 2018-08-03
Handle null mappings correctly in sourcemaps

### LINKED ISSUE #1825: Sourcemaps: null mappings not working as expected
# 🐛 bug report

## 🤔 Expected Behavior

Null mappings should be mapped

## 😯 Current Behavior

Null mappings are being ignored

## 💁 Possible Solution

Don't ignore null mappings

## 🔦 Context

Comments by @loganfsmyth in #1755

DeMoorJasper:
This allows null mappings in parcels sourcemaps.
Creating more accurate sourcemaps.
Also adds a couple extra tests for sourcemaps, using the SourceMap class directly, giving a more in depth coverage of SourceMap

Closes #1825

cc @loganfsmyth


src/SourceMap.js
```diff
@@ -33,13 +33,16 @@ class SourceMap {
     if (map instanceof SourceMapConsumer) {
       return map;
     }
+
     map = typeof map === 'string' ? JSON.parse(map) : map;
+    if (map.sourceRoot) delete map.sourceRoot;
```
loganfsmyth: Is the intention with this to normalize the file paths somehow?  It may well make sense for Parcel to normalize this stuff somehow, but dropping it entirely seems like it could lose useful information.

DeMoorJasper: The mapConsumer resolves paths if the sourcemap has a sourceRoot (turning them into absolute paths), so if parcel would be consuming a map of it's own this would duplicate the mappings (as internally parcel and sourcemaps use relative paths).

I'm not sure how else to solve this, I guess I can make every path in sourcemaps relative to sourceRoot, but that would slow it down quite a lot. I am pretty positive that this will rarely/never cause issues inside parcel as the source is defined where parcel inits the sourcemap (for example JSAsset).

devongovett:

PR #1861, opened 2018-08-07
Fix non-updating asset hashes

### LINKED ISSUE #1789: bug: HTML file is not updated on update; .cache has stale copy
# 🐛 bug report

**Repo**
1. Run `parcel index.html`
2. Change apple.html (add some numbers)

None, ran `parcel index.html`

## 🤔 Expected Behavior

Would expect the console.log to have the updated version of the HTML updated file

## 😯 Current Behavior

The console.log never updates with the latest version of `apple.html` despite the page knowing there was an update

## 💁 Possible Solution

The only workaround right now is to remove the `.cache` directory and rerun parcel.

## 🔦 Context

Trying out AngularJS experience with Parcel. AngularJS requires HTML templates; but even without AngularJS Parcel does not reload HTML files correctly.

## 💻 Code Sample

index.js
```js
const apple = require('./apple.html');
console.log(apple);
```

apple.html
```html
<h1>test224443331111112</h1>
```

index.html
```html
<!DOCTYPE html>
<html lang="en">
<head>
  <meta charset="UTF-8">
  <meta name="viewport" content="width=device-width, initial-scale=1.0">
  <meta http-equiv="X-UA-Compatible" content="ie=edge">
  <title>Document</title>
</head>
<body>
</body>
</html>
```

## 🌍 Your Environment

| Parcel           | using github repo  and link | 
| Node             |  8.3.0 | 
| npm/Yarn         | 1.5.1 | 
| Operating System | MacOS |

DeMoorJasper:
Include asset content in asset hash, as generated not always outputs the entire ast and makes it unreliable. (Only content would probably also be unreliable)

Should prevent some cases where assets don't get updated either from cache or HMR

Closes #1789


src/Asset.js
```diff
@@ -214,7 +214,13 @@ class Asset {
   }
 
   generateHash() {
-    return objectHash(this.generated);
+    let contentHash = '';
+    try {
+      contentHash = md5(this.contents);
+    } catch (err) {
```
DeMoorJasper: In case this.contents is not a string or undefined

devongovett:
Maybe the hash should be based on the contents after post processing instead of generate also? That seems to be related to this issue. Maybe only a content hash would be good enough though instead of hashing the whole result as well.

DeMoorJasper:
So moving hash generation to the Pipeline after postProcess?

devongovett:
Yeah I guess? I'd say just hash the original content but sometimes we don't actually read it (so you have the try..catch). Try just hashing the post-processed output instead of generate and see if that fixes the bug.

DeMoorJasper:
Changed it, should still fix the issue. As it's pretty much doing the same. Just on output instead of input. (which is probably more reliable for virtual assets like globs)

src/Asset.js
```diff
@@ -214,7 +213,7 @@ class Asset {
   }
 
   generateHash() {
```
devongovett: Is this method needed anymore?

DeMoorJasper: RawAsset uses it as it has no generated content

devongovett: Maybe we should keep the method then and just call it after post-processing is done?

devongovett:
Looks like the tests are failing because the yarn lockfile is pointing at adobe corp artifactory. Can you fix?

PR #1864, opened 2018-08-08
Detect and apply stripping flow types

### LINKED ISSUE #1863: Strip away flow types
# 🐛 bug report

Parcel is currently not able to consume flow typed modules

N/A

## 🤔 Expected Behavior

Flow types should be stripped (perhaps configurable through babel)

## 😯 Current Behavior

Flow types cause errors

## 💁 Possible Solution

- locally link a copy of the module that contains flow types
- Ask the maintainer to publish the module using `.flow` and `.js` files. `.flow` being the typed one and `.js` being processed by babel to strip away the typings
- Make parcel automagically detect and strip away flow types (in node modules only)

## 🔦 Context

Issue #1799

## 💻 Code Sample

```Javascript
import Typography from 'typography';
```

cc @chiefGui

DeMoorJasper:
Closes #1863


src/transforms/babel.js
```diff
@@ -312,7 +312,10 @@ async function getJSXConfig(asset, isSourceModule) {
  * Generates a babel config for stripping away Flow types.
  */
 function getFlowConfig(asset, isSourceModule) {
-  if (!isSourceModule && asset.contents.includes('@flow')) {
+  if (
+    !isSourceModule &&
+    asset.contents.substring(0, 20).indexOf('@flow') > -1
```
DeMoorJasper: https://jsperf.com/includes-vs-substring-includes-vs-indexof/1 Performance test to back up this choice

src/transforms/babel.js
```diff
@@ -301,3 +307,20 @@ async function getJSXConfig(asset, isSourceModule) {
     };
   }
 }
+
+/**
+ * Generates a babel config for stripping away Flow types.
+ */
+function getFlowConfig(asset, isSourceModule) {
+  if (
+    !isSourceModule &&
```
devongovett: Should this be the inverse? Wouldn't we want to enable flow on app code and locally linked modules? I'm not sure enabling it on published node_modules is a good idea. WDYT?

src/transforms/babel.js
```diff
@@ -301,3 +307,20 @@ async function getJSXConfig(asset, isSourceModule) {
...
+    asset.contents.substring(0, 20).indexOf('@flow') > -1
```
devongovett: Should we ensure that this is inside a comment? like look for `// @flow` or something?

src/transforms/babel.js
DeMoorJasper: The issue is published code, as they shouldn't ship with flow types, local code should follow babelrc.

Shipped code should probably use flow comments or seperate .flow files

src/transforms/babel.js
DeMoorJasper: The issue is that it can be `// @flow` or `/* @flow */` or any other variation of comment. So either we'd have to use a regex or do a babel walk to just detect this. Which would probably slow things down.

I can change it to a regex or babel walk just not sure about performance

src/transforms/babel.js
devongovett: I think it would be a nice feature to enable flow automatically on source code just like we do for JSX.

Shipped code should generally be pre-transpiled to remove the types I thought... Not sure, I guess it doesn't hurt to automatically do it for everything.

src/transforms/babel.js
devongovett: Yeah a regex would be ok (on the substring). Should be fast enough.

src/transforms/babel.js
DeMoorJasper: It should, but apparently not every module does it as you can see in the linked issue

src/transforms/babel.js
```diff
@@ -148,6 +149,11 @@ async function getBabelConfig(asset) {
     return jsxConfig;
   }
 
+  // If there is a Flow config, return that
+  if (flowConfig) {
```
devongovett: I guess we also need to merge the flow config with the generated env/jsx configs above right?

DeMoorJasper: Probably, fixed it

PR #1892, opened 2018-08-13
Import GraphQL files from other GraphQL files (closes #1477)

### LINKED ISSUE #1477: Import directive within GraphQL files

arianon:
An alternative to #1817 that doesn't rely on using `graphql-tag/loader`, which is a Webpack loader.

#### What this does:
1. Recursively traverse the GraphQL files.
2. Collect all the imports into a map.
3. Concatenate them.
4. Parse the result into an AST.

#### What this doesn't do:

This PR (as of time of writing) does not have feature parity with `graphql-tag/loader`, meaning that:
* It does not de-duplicate fragments that have the same name. _This is a behavior that I personally disapprove of since it can lead to surprising results instead of throwing an error if a user accidentally (however unlikely) reuses an already defined fragment name._
* It does not collect multiple defined queries/mutations and make them individually require-able from a JavaScript parent.

If the above behaviours are desired then I can implement them.

### DIFF
--- a/src/assets/GraphqlAsset.js
+++ b/src/assets/GraphqlAsset.js
@@ -24,7 +24,7 @@ class GraphqlAsset extends Asset {
 
     await Promise.all(
       code
-        .split(os.EOL)
+        .split(/\r\n?|\n/)
         .map(line => line.match(IMPORT_RE))
         .filter(match => !!match)
         .map(async ([, importName]) => {
--- a/src/assets/GraphqlAsset.js
+++ b/src/assets/GraphqlAsset.js
@@ -1,15 +1,62 @@
 const Asset = require('../Asset');
 const localRequire = require('../utils/localRequire');
+const Resolver = require('../Resolver');
+const fs = require('../utils/fs');
+const os = require('os');
+
+const IMPORT_RE = /^# *import +['"](.*)['"] *;? *$/;
 
 class GraphqlAsset extends Asset {
   constructor(name, options) {
     super(name, options);
     this.type = 'js';
+
+    this.gqlMap = new Map();
+    this.gqlResolver = new Resolver(
+      Object.assign({}, this.options, {
+        extensions: ['.gql', '.graphql']
+      })
+    );
+  }
+
+  async traverseImports(name, code) {
+    this.gqlMap.set(name, code);
+
+    await Promise.all(
+      code
+        .split(os.EOL)
+        .map(line => line.match(IMPORT_RE))
+        .filter(match => !!match)
+        .map(async ([, importName]) => {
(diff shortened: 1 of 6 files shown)


src/assets/GraphqlAsset.js
```diff
@@ -1,15 +1,62 @@
 const Asset = require('../Asset');
 const localRequire = require('../utils/localRequire');
+const Resolver = require('../Resolver');
+const fs = require('../utils/fs');
```
DeMoorJasper: Please use parcels fs util

src/assets/GraphqlAsset.js
```diff
@@ -1,15 +1,62 @@
...
+      })
+    );
+  }
+
+  async traverseImports(name, code) {
+    this.gqlMap.set(name, code);
+
+    await Promise.all(
+      code
+        .split(os.EOL)
```
DeMoorJasper: Will this work in every case?
Seems like this would not work if a user clones a repo that is being used developed various OSes

src/assets/GraphqlAsset.js
arianon: Pretty sure I am?

src/assets/GraphqlAsset.js
arianon: Good point, using a regex such as `/\r?\n/` is probably resilient.

arianon: Changed it to use `/\r\n?|\n` which is pretty much the same as [what the GraphQL specification defines as a line terminator](http://facebook.github.io/graphql/June2018/#sec-Line-Terminators)

src/assets/GraphqlAsset.js
DeMoorJasper: Guess I fucked up, looks good

PR #1909, opened 2018-08-19
Remove sass url loader, move url resolving to PostCSS

### LINKED ISSUE #316: Url() without quotes is not supported in sass

DeMoorJasper:
This PR moves url resolving to PostCSS instead of relying on SASS to work.

Figured it might be worth it as it has been a pretty stale issue on the SASS end for a lil while now.

Closes #316

### DIFF
--- a/src/assets/SASSAsset.js
+++ b/src/assets/SASSAsset.js
@@ -40,13 +40,6 @@ class SASSAsset extends Asset {
         ? opts.indentedSyntax
         : type === 'sass';
 
-    opts.functions = Object.assign({}, opts.functions, {
-      'url($url)': url => {
-        let filename = this.addURLDependency(url.getValue());
-        return new sass.types.String(`url(${JSON.stringify(filename)})`);
-      }
-    });
-
     opts.importer = opts.importer || [];
     opts.importer = Array.isArray(opts.importer)
       ? opts.importer
@@ -83,8 +76,7 @@ class SASSAsset extends Asset {
     return [
       {
         type: 'css',
-        value: this.ast ? this.ast.css.toString() : '',
-        hasDependencies: false
+        value: this.ast ? this.ast.css.toString() : ''
       }
     ];
   }
--- a/test/integration/scss-url/index.scss
+++ b/test/integration/scss-url/index.scss
@@ -6,3 +6,7 @@
 .index {
   background: url("http://google.com");
 }
+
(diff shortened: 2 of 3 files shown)


devongovett:
Won't this be slower since it will have to reparse the generated css?

DeMoorJasper:
@devongovett probably, but at least it works.
It doesn't seem like the sass maintainers are gonna solve this issue any time soon, we can always change it back once they do solve it.

DeMoorJasper:
@devongovett probably, but at least it works.
It doesn't seem like the sass maintainers are gonna solve this issue any time soon, we can always change it back once they do solve it.

devongovett:
I take it dart-sass has the same issue? Maybe we can just contribute?

DeMoorJasper:
Yes dart-sass has the same issue (as it's implementation tries to be identical to node-sass, it's just a little stricter and 100% JS, which makes it more stable in case of our workerfarm setup), not sure if it's a feature or an actual bug.
From the issue we have it seems to be a feature they haven't fully worked out yet from a design perspective although I'm not sure.

This seemed like the cleanest way to implement a hotfix for the meantime.

DeMoorJasper:
Found the issue where they're discussing the url loader design https://github.com/sass/sass/issues/2535

devongovett:
ok sure, but even if they are re-designing it seems like a bug that it works when quoted but not when unquoted...

I will merge this for now, but let's keep an eye on that sass issue and revert once they figure it out so we regain some performance.

PR #1924, opened 2018-08-23
Don't modify script nodes with text/html type

southpolesteve:
Knockout.js uses <script type="text/html> for named templates. Currently, the type attribute is stripped by parcel. This change prevents that from happening.

### DIFF
--- a/src/assets/HTMLAsset.js
+++ b/src/assets/HTMLAsset.js
@@ -64,7 +64,8 @@ const META = {
 const SCRIPT_TYPES = {
   'application/javascript': 'js',
   'text/javascript': 'js',
-  'application/json': false
+  'application/json': false,
+  'text/html': false
 };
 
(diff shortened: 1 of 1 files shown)


src/assets/HTMLAsset.js
```diff
@@ -64,7 +64,8 @@ const META = {
 const SCRIPT_TYPES = {
   'application/javascript': 'js',
   'text/javascript': 'js',
-  'application/json': false
+  'application/json': false,
+  'text/html': false
```
devongovett: Maybe we should change it to a whitelist instead of a blacklist? Basically, just `application/javascript` and  `text/javascript`, or if the type attribute is not specified. We could also enable typescript, coffeescript, etc.

southpolesteve: That makes sense. In the case of no `type` attribute you would like to keep existing behavior and default to JS? Wondering if this will be a breaking change for existing users.

PR #1934, opened 2018-08-24
feat:added no content hashing option to build cli

### LINKED ISSUE #1885: Make contentHash available in CLI
## 🤔 Expected Behavior

This is an important option.
To be available as a command line option.

## 😯 Current Behavior

In the Current version (1.97) it's only available as API.

EricTurf:
Based on request from #1885 

- **Added** `--no-content-hash` CLI option to `parcel build`
- **Changed** the exiting of the process to occur if `options.watch` is false instead of looking at `NODE_ENV`

CLOSES #1885


devongovett:
I'm curious why you would want to disable content hashing?

EricTurf:
I am working on a project where we are required to commit the bundle. Therefore, in PR we would get huge changes since sometimes the bundle was sometimes hashed and sometimes not. 

We had to move to the `API` way of doing things to avoid this but since we would like to stick to the CLI bundling approach we decided to open a PR to add that functionality.

src/cli.js
```diff
@@ -193,8 +194,11 @@ async function bundle(main, command) {
   }
 
   if (command.name() === 'build') {
-    command.production = true;
-    process.env.NODE_ENV = process.env.NODE_ENV || 'production';
+    command.watch = false;
+    process.env.NODE_ENV =
+      process.env.NODE_ENV || command.contentHash
+        ? 'production'
+        : 'development';
```
devongovett: I don't think we should change the behavior here.

kleinfreund:
What exactly does this do right now? As far as I can tell, specifying `--no-content-hash` for `parcel build` doesn’t change anything. The content hashes are still included in the file names.

numical:
The filename hashes are then based on the filename, not the file contents; aka static filenames

---- On Sun, 30 Sep 2018 07:47:12 +0000 notifications@github.com wrote ----

What exactly does this do right now? As far as I can tell, specifying --no-content-hash for parcel build doesn’t change anything. The content hashes are still included in the file names.

Andrew-Dyachenko:
> What exactly does this do right now? As far as I can tell, specifying `--no-content-hash` for `parcel build` doesn’t change anything. The content hashes are still included in the file names.

+1 the `--no-content-hash` flag using has no effect. The names of the building files still contain a hash in the file name. We're currently need static file names in a project

How to solve this?

img-2019-10-27-14-02-53

mischnic:
> doesn’t change anything

It does, by default, file hashes are calculated based on the file contents. With that flag, they are based only on the file path.

Andrew-Dyachenko:
Ohhhh!

I checked the launch of the assembly with flag `--no-content-hash` and without.
If you change something in the files and run files with the `--no-content-hash` flag, then the hash will not change. Since the path to the file remains the same. If run without a flag, the hash will change based on the changed content.

Looks like you're right! A little unseemly flag. It is necessary to supplement the official documentation with an extended explanation!

@mischnic Thank you!

Andrew-Dyachenko:
Nevertheless, in my opinion, we still need a flag that would permanently disable the hash in the file name. I believe that such a flag will be called: `--no-hash`

PR #747, opened 2018-02-05
Add support for HTML meta assets

dmgawel:
This PR introduces support for assets defined in HTML metadata, as specified by most common schemas:
* Open Graph protocol
* Twitter Cards
* schema.org
* Microsoft pinned site metadata

Resolves #220.

PR #1936, opened 2018-08-25
Added LD+JSON asset and test case

### LINKED ISSUE #1935: Missing support for dependency discovery in JSON-LD for schemas
https://github.com/parcel-bundler/parcel/blob/master/src/assets/HTMLAsset.js#L64 JSON is skipped. I know schema.org specs are quite extensive but there are a few basics: image / logo for assets.

# 🙋 feature request

Assets used in JSON-LD schemas aren't added to the dependency tree. And as a result are missing from builds.

## 🤔 Expected Behavior

A minimal subset of the attributes should be included in parcel, just like it was done with Open Graph, etc: https://github.com/parcel-bundler/parcel/pull/747

## 😯 Current Behavior

Nothing happens since `application/ld+json` scripts aren't recognized.

## 💁 Possible Solution

Parse JSON, fetch attributes that match a list (almost the same as https://github.com/parcel-bundler/parcel/pull/747) and finally add them as dependencies.

## 💻 Examples

```html
<script type="application/ld+json">
{
  "@context": "http://www.schema.org",
  "@type": "xxx",
  "name": "Parcel Bundler",
  "url": "https://parcel.url/",
  "image": "./images/header.jpeg",
  "logo": "./images/logo.png"
}
</script>
```

bre7:
Closes #1935 

## 📋 TODO

Review ✨

## 👌 DONE

- [x] 🚨 Resources MUST be valid URLs, not paths.
  - Uses `public-url` parameter. Avoids creating a new flag but spams the DOM
- [x] Deal with nested values for schema attributes
- [x] Added missing meta attribute `msapplication-config`. Complements #747
- [x] Dependencies are collected but the AST isn't saved correctly (still has old files)
- [x] Prevent Parcel from stripping script's type: `type="application/ld+json"`
- [x] ~~Minify JSON (`transform`)~~ Handled by htmlnano
    - JSON is minified by default when using `JSON.stringify()`

## 💁 Possible Solution for URLs

Ask for URL using a new parameter (`published-url` ?)
  - Probably the best solution but requires adding an extra parameter to the CLI

Assume the user will enter a full URL (warn otherwise) in the schema
  1. When parsing the schema, extract the path and treat it as any local path
  2. Resolve dependency
  3. Write back the URL + collected path
- Obvious **trade-off**: Parcel will assume all asset URLs in the schema target the site in development, when in fact they can be different domains.

Use `public-url`


src/Parser.js
```diff
@@ -20,6 +20,7 @@ class Parser {
     this.registerExtension('vue', './assets/VueAsset');
     this.registerExtension('json', './assets/JSONAsset');
     this.registerExtension('json5', './assets/JSONAsset');
+    this.registerExtension('json-ld', './assets/JSONLDAsset');
```
devongovett: let's make the extension `.jsonld` as specified here: https://json-ld.org/spec/latest/json-ld/#iana-considerations

src/assets/JSONLDAsset.js
```diff
@@ -0,0 +1,35 @@
+const Asset = require('../Asset');
+
+// A list of all attributes in a schema that may produce a dependency
+// Based on https://schema.org/Place
+const SCHEMA_ATTRS = [
+  "logo",
+  "photo",
+  "image"
+];
```
devongovett: Good start. I guess we should go through the specs and find all the possible attributes that are URLs

src/Parser.js
bre7: 👍

src/assets/JSONLDAsset.js
bre7: Not sure if they list every single property, but https://schema.org/ImageObject has a dependents list: "Instances of ImageObject may appear as values for the following properties"

bre7:
Devon, can I add a new cli parameter ? I can't really see another way to (properly) solve the issue with URLs/paths

src/assets/HTMLAsset.js
```diff
@@ -247,13 +249,15 @@ class HTMLAsset extends Asset {
       } else if (type === 'tag') {
         if (
           (rendition.type === 'js' && node.tag === 'script') ||
-          (rendition.type === 'css' && node.tag === 'style')
+          (rendition.type === 'css' && node.tag === 'style') ||
+          (rendition.type === 'jsonld' && node.tag === 'script')
```
bre7: Why not just check `isAstDirty` ?

src/assets/JSONLDAsset.js
```diff
@@ -0,0 +1,61 @@
...
+  }
+
+  parse(content) {
+    return JSON.parse(content.trim());
+  }
+
+  collectDependencies() {
+    for (let schemaKey in this.ast) {
+      // only check for single values, not nested data
+      // todo: check for nested data
```
devongovett: make a recursive function for this?

src/assets/JSONLDAsset.js
```diff
@@ -0,0 +1,61 @@
...
+      ) {
+        // paths aren't allowed, values must be urls
+        let assetPath = this.addURLDependency(this.ast[schemaKey]);
+        if (!isURL(assetPath)) {
+          assetPath = urlJoin(this.options.publicURL, assetPath);
+        }
+        this.ast[schemaKey] = assetPath;
+        this.isAstDirty = true;
+
+        if (this.options.publicURL === '/') {
```
devongovett: so I take it JSON LD requires full URLs not paths? If so, then this is probably fine to show a warning.

Maybe this check should check whether the public URL **starts with** / instead of fully equaling it?

src/assets/JSONLDAsset.js
```diff
@@ -0,0 +1,61 @@
...
+          console.warn(
```
devongovett: use the parcel logger for this

src/assets/JSONLDAsset.js
```diff
@@ -0,0 +1,61 @@
...
+// A list of all attributes in a schema that may produce a dependency
+// Based on https://schema.org/ImageObject
+// Section "Instances of ImageObject may appear as values for the following properties"
+const SCHEMA_ATTRS = [
+  'logo',
+  'photo',
+  'image',
+  'thumbnail',
+  'screenshot',
+  'primaryImageOfPage'
```
devongovett: You could look through https://schema.org/URL for more URL instances. Maybe Video/MediaObject and a few others of those would be good to support too.

bre7:
I still think `publicURL`, even if it works, is unsuitable for this specific use case since it ends up spamming the DOM (which imo, it shouldn't)

src/assets/HTMLAsset.js
```diff
@@ -247,13 +249,15 @@ class HTMLAsset extends Asset {
...
         ) {
           node.content = rendition.value;
         }
 
         // Delete "type" attribute, since CSS and JS are the defaults.
-        if (node.attrs) {
+        // Unless it's application/ld+json
+        if (node.attrs && node.attrs.type !== 'application/ld+json') {
```
bre7: Will be modified by https://github.com/parcel-bundler/parcel/pull/1924

bre7:
@devongovett needs a final review

PR #1939, opened 2018-08-26
Optimize simple CommonJS with scope hoisting

fathyb:
This PR optimizes simple CommonJS exports, like : 
```js
module.exports = ['foo', 'baz', 'bar']
```

This will improve minification as currently only Google Closure Compiler has a control-flow to handle this kind of case. (terser and babel-minify will bailout on reassignments)

#### Before

- `gen`
  ```js
  var $id$exports = {}
  $id$exports = ['foo', 'baz', 'bar']

  console.log('test', $id$exports[1])
  ```

- `minify`
  ```js
  var a = {}
  a = ['foo', 'baz', 'bar']

  console.log('test', a[1])
  ```
#### After

- `gen`
  ```js
  var $id$exports = ['foo', 'baz', 'bar']

  console.log('test', $id$exports[1])
  ```

- `minify`
  ```js
  console.log('test', 'baz')
  ```


src/scope-hoisting/hoist.js
```diff
@@ -112,6 +112,13 @@ module.exports = {
       });
 
       path.scope.setData('shouldWrap', shouldWrap);
+
+      if (!shouldWrap) {
```
fathyb: I referenced the binding here to not have to recrawl after the `scope.push` line 163

test/scope-hoisting.js
```diff
@@ -821,6 +821,22 @@ describe('scope hoisting', function() {
       assert(!contents.includes('bar'));
     });
 
+    // TODO(fathyb): reenable when minification is improved
+    it.skip('removes unused JSON fields', async function() {
```
fathyb: babel-minify doesn't support this yet, but I plan on fixing this soon

kzc:
> only Google Closure Compiler has a control-flow to handle this kind of case. (terser and babel-minify will bailout on reassignments)

@fathyb Not necessarily...

```
$ cat before.js 
var $id$exports = {}
$id$exports = ['foo', 'baz', 'bar']
console.log('test', $id$exports[1])
```
```
$ cat before.js | terser --toplevel -c unsafe
console.log("test","baz");
```
The uglify/terser `unsafe` compress option is actually safe for sane code - anything that does not modify builtins like polyfills. It's just being conservative by default.

Application code should specify `--toplevel` (or `--module`) to get rid of unused top level variable declarations.

PR #1834, opened 2018-08-02
Minimal verbose mode

DeMoorJasper:
Adds verbose and debug mode:

4: verbose, keep everything in log with timestamps, instead of using statusspinner
5: debug, save everything to a file with timestamps

Shouldn't slow down anything, verbose might even speed everything up as it uses native console.log instead of a spinner.

Also adds a log call after an asset has been processed to track which files are being successfully built

Closes #993

PR #1952, opened 2018-08-28
Parcel 2 RFC

jamiebuilds:
## [[View Rendered](https://github.com/parcel-bundler/parcel/blob/parcel-2-rfc/PARCEL_2_RFC.md)]

This RFC is written as documentation for the project, our documentation should
be detailed enough to work as a spec, and it's easier to write it once this way
than to approve a spec and start again on the documentation from scratch.

Everything in Parcel should be documented here, if it is not documented it will
not be part of Parcel 2.

---

**Please let us know if you have any suggestions by commenting inline on the diff of this draft.**

**All feedback, from anyone, is welcome.**


PARCEL_2_RFC.md
```diff
@@ -0,0 +1,1285 @@
...
+
+Aliases for `--mode test`.
+
+#### `--public-url <url>`
+
+[todo]
+
+#### `--log-level <level>`
+
+Set the log level, either "0" (no output), "1" (errors), "2" (warnings +
```
DeMoorJasper: Should there be a more detailed log mode? See #1834

PARCEL_2_RFC.md
```diff
@@ -0,0 +1,1285 @@
...
+multiple targets at once and have them live update as you make changes.
+
+Parcel is fast and predictable. It compiles all of your files in isolation in
+parallel inside workers, caching all of them as it goes along. Caches are
+stable across machines and are only affected by the files and configs within
+your project (unless you want to pass specific environment variables).
+
+## Getting Started
+
+Before we get started, you'll need to install Node and Yarn (or npm) and create
```
johnsonw: What is the minimum node version that is supported?

PARCEL_2_RFC.md
```diff
@@ -0,0 +1,1285 @@
...
+```
+
+You should also add some "scripts" to your `package.json` to run it easier.
+
+```json
+{
+  "name": "my-project",
+  "scripts": {
+    "build": "parcel --production",
+    "start": "parcel --serve"
```
johnsonw: Consider using a postversion script for production builds.

PARCEL_2_RFC.md
```diff
@@ -0,0 +1,1285 @@
...
+
+See [Targets](#targets)
+
+#### `package.json#browser`
+
+This is the "browser" target's entry point for the package.
+
+```json
+{
+  "browser": "distflinesof/browser/index.js"
```
DeMoorJasper: I think this is a typo. Shouldn't it be `dist/browser/index.js`?

PARCEL_2_RFC.md
```diff
@@ -0,0 +1,1285 @@
...
+
+Aliases for `--mode production`.
+
+##### `--test`
+
+Aliases for `--mode test`.
+
+#### `--public-url <url>`
+
+[todo]
```
johnsonw: Is this implemented yet?

PARCEL_2_RFC.md
```diff
@@ -0,0 +1,1285 @@
...
+{
+  "alias": {
+    "react": "preact-compat",
+    "react-dom": "preact-compat"
+  }
+}
+```
+
+See [Aliases](#aliases)
+
```
DeMoorJasper: Should we support externals? #144

PARCEL_2_RFC.md
johnsonw: It would read cleaner to list these out:

Sets the log level (Default 2):
1. `0`: No output
1. `1` : Errors
1. `2`: Warnings and Errors
1. `3`: All

PARCEL_2_RFC.md
```diff
@@ -0,0 +1,1285 @@
...
+When you do need to configure Parcel, it will be in one of 3 places.
+
+- If you need to configure the CLI, it will be a CLI flag
+- If you need to configure your package, it will be in the `package.json`
+- If you need to configure something with your files or the Parcel asset
+  pipeline, it will be in `.parcelrc`
+
+### `package.json`
+
+[todo]
```
johnsonw: What needs to be added here?

PARCEL_2_RFC.md
```diff
@@ -0,0 +1,1285 @@
...
+
+```json
+{
+  "name": "foo",
+  "main": "dist/main/index.js",
+  "module": "dist/module/index.js",
+  "browser": "dist/browser/index.js",
+  "browserslist": ["> 1%", "not dead"],
```
youknowriad: I know it's already "common" to define "browserslist" as a top level property like this but it creates an inconsistency with the "engines" property. Does it make sense for "browsers" to be just another "engine"?

jamiebuilds: If we change what people are already doing they will just end up adding and maintaining both

lbguilherme:
`--target node|browser|electron` is not there.

devongovett:
@lbguilherme see https://github.com/parcel-bundler/parcel/blob/parcel-2-rfc/PARCEL_2_RFC.md#targets

PARCEL_2_RFC.md
jamiebuilds: ...indeed, wtf happened?

PARCEL_2_RFC.md
jamiebuilds: It will always be whatever the lowest active LTS version of Node is.

PARCEL_2_RFC.md
jamiebuilds: I don't think that is necessary for this guide

PARCEL_2_RFC.md
jamiebuilds: I've been wanting to discuss this more with @devongovett cause it's sorta the odd flag out in the CLI

PARCEL_2_RFC.md
jamiebuilds: I'm confused about the difference between `verbose` and `debug`

Is it dramatically different from `parcel --verbose > parcel.log`?

PARCEL_2_RFC.md
jamiebuilds: dunno, just needs a description

azeznassar:
Could you add the basic CLI commands for getting started in both Yarn and NPM?

PARCEL_2_RFC.md
jamiebuilds: It's a good question, I think we need to talk about it a bit more.

One part of the problem I think belongs in aliases:

```json
"aliases": {
  "react": "https://unpkg.com/react@16.4.2/umd/react.production.min.js"
}
```

PARCEL_2_RFC.md
DeMoorJasper: Originally debug printed a bunch more things like bundler options for example.

With the current implementation it would probably be the same to do `parcel --verbose > parcel.log` except for the timestamp but that's just a slightly more complex command.

I still however think we should have both, but with Debug also having some more data like OS, Node Version, Parcel Version, Options, ...

jamiebuilds: Okay, I'd be happy to add another level above verbose, but I would argue that it shouldn't write to a file by default

DeMoorJasper: Great, I agree it wasn't the best idea to spam the user with logfiles

PARCEL_2_RFC.md
DeMoorJasper: Awesome, however I'm not sure how that would work in practice, as the interface on that is different than the node_modules version.

My main concern is probably that there is not really a standard for how to publish prebuilt packages.

PARCEL_2_RFC.md
```diff
@@ -0,0 +1,1285 @@
...
+than to approve a spec and start again on the documentation from scratch.
+
+Everything in Parcel should be documented here, if it is not documented it will
+not be part of Parcel 2.
+
+---
+
+## Introduction
+
+Parcel is a compiler for all your code, regardless of the language or toolchain.
```
sompylasar: "all your code" sounds too broad, consider "all your web application frontend code".

DeMoorJasper: Parcel can also be used for back-end code, mainly node

PARCEL_2_RFC.md
```diff
@@ -0,0 +1,1285 @@
...
+$ parcel [...entries] [...flags]
+```
+
+#### `[...entries]`
+
+Entry files to start bundling, these will be preserved as entry points in the
+output. Defaults to `package.json#source`, falling back to `src/index.*` or
+`index.*`. See [#Entries](#entries-sources-targets-environment).
+
+#### `--serve, -s`
```
sompylasar: Put the option-separating commas outside of the backticks consistently?
```diff
- #### `--serve, -s`
- #### `--open, -o [browser]`
+ #### `--serve`, `-s`
+ #### `--open [browser]`, `-o [browser]`
  #### `--cache <dir>`, `--no-cache`
  #### `--hot`, `--no-hot`
```

PARCEL_2_RFC.md
sompylasar: Have you considered the `debug` module where you'd use literal names as a combination of filters in the `DEBUG` envvar to introspect into individual modules and components? Ex., `DEBUG=parcel:*`.

PARCEL_2_RFC.md
```diff
@@ -0,0 +1,1285 @@
...
+Parcel has always and will always work out of the box for many projects with
+zero configuration. It should always be extremely simple to get started. But if
+you do want more control, we give you the tools to do so.
+
+### Configuring external tools
+
+A huge part of what Parcel does is run other tools over your code. Instead of
+pulling all that configuration into Parcel, we make use of their own
+configuration systems. So if you're using Babel, you should just use `.babelrc`
+files to configure it.
```
sompylasar: What if the tool does not support dynamic configuration files (that for example depend on the filepath that's being processed, and has a custom logic like substituting a parser based on some condition)? Parcel still uses the tools via their JS APIs, so it could potentially resolve and pass a generated configuration to them. I keep webpack mainly because it can do that and that I can generate a different babelrc for different parts of the app (node processes) within a single package, but Parcel requires the root of your package to be occupied by various tools' config files that are the same for the entire code base. Of course, in the ideal world, different parts would be separate packages, reusable parts in their own packages, etc., but it's a rare luxury to have a large real world app code structured like that. What do you think?

PARCEL_2_RFC.md
devongovett: We'd like to keep our options open. :) Perhaps something like react native could be packaged with Parcel. The architecture has nothing JavaScript or even Web specific in it, so it should be possible to extend to any platform.

DeMoorJasper:
@bjankord I think this might be a feature of `@parcel/transform-postcss` and shouldn't be part of the Parcel 2 Spec as this is specific to postcss and Parcel 2 itself is basically just a wrapper around tooling, so independant of any independant transform

mmun:
Could someone give an example of what the contents of a single bundle might look like?

jamiebuilds:
> What is a bundle exactly? Is it just a collection of arbitrary files (i.e. of different asset types)?

@mmun Bundles should consist only assets of the same asset type. Although transforms can convert one asset type to another.

PARCEL_2_RFC.md
```diff
@@ -0,0 +1,1285 @@
...
+
+## Plugin APIs
+
+There are several different types of plugins. They all look very similar, but
+are kept separate so we can have strict contracts one what each one is allowed
+to do.
+
+There are some rules that should be followed across every type of plugin:
+
+- **Stateless** — Avoid any kind of state, it will likely be the source of bugs
```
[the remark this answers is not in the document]
DeMoorJasper: cache will be handled by Parcel, as far as I know transforms won't have direct access to it. But every transform will be cached.

mmun:
@jamiebuilds Gotcha. How would you model optimizations across bundle? I'm thinking of something like inlining images into CSS.

jamiebuilds:
I imagine that would be best done in CSS compilers. I believe Sass does that already

PARCEL_2_RFC.md
```diff
@@ -0,0 +1,1285 @@
...
+
+1. Aliases will only be respected when specified outside of `node_modules`.
+2. Aliases specified outside of `node_modules` will affect assets inside of `node_modules`.
+3. Aliases cannot build off of other aliases.
+4. Only one alias will be applied at a time.
+5. Aliases must be valid npm package names.
+
+## Plugins
+
+### Resolvers
```
mohsen1: Just want to confirm a resolver plugin can do what TypeScript [module resolution spec](https://www.typescriptlang.org/docs/handbook/module-resolution.html) describes.

PARCEL_2_RFC.md
```diff
@@ -0,0 +1,1285 @@
...
+
+```json
+{
+  "reporters": ["@parcel/reporter-detailed"]
+}
+```
+
+See [Reporters](#reporters)
+
+## Parcel Architecture
```
mohsen1: An in-memory FS was discussed before. Will Parcel 2 maintain an in-memory FS? Will plugins get access to that FS?

I'm asking because some tools like GraphQL codegen write `.d.ts` TypeScript files to disk to make TypeScript happy about types. if the TypeScript compiler uses an in-memory FS and GraphQL writes to that you won't need to write to **source** directories to make the build pass

PARCEL_2_RFC.md
```diff
@@ -0,0 +1,1285 @@
...
+in your app or program. Every asset can have it's own dependencies on other
+assets which Parcel will pull in.
+
+The data structure that represents all of these assets and their dependencies
+on one another is known as "The Asset Graph".
+
+| Asset Name   | Dependencies        |
+| ------------ | ------------------- |
+| `index.html` | `app.css`, `app.js` |
+| `app.css`    | N/A                 |
```
tripodsan: what if the css references other dependencies, like background-images or fonts ?

PARCEL_2_RFC.md
```diff
@@ -0,0 +1,1285 @@
...
+### Transforms
+
+Transforms _transform_ single assets as they are discovered and added to the
+asset graph. They mostly call out to different compilers and preprocessors.
+
+```js
+import { Transform } from '@parcel/plugin';
+
+export default new Transform({
+  async config({ asset }) {
```
tripodsan: would be nice to add js-doc to all methods below.
also, what is the _type_ of the asset passed ?

PARCEL_2_RFC.md
```diff
@@ -0,0 +1,1285 @@
...
+
+**Official Transforms:**
+
+- `@parcel/transform-babel`
+- `@parcel/transform-coffeescript`
+- `@parcel/transform-graphql`
+- `@parcel/transform-json`
+- `@parcel/transform-json5`
+- `@parcel/transform-less`
+- `@parcel/transform-posthtml`
```
tripodsan: how can one extend/configure the official transformers, or make use in their own plugins? for example, I want to rewrite some HTML tags with the _posthtml_ transformer?

DeMoorJasper: Parcel takes in an array of transforms, so they're chained so you can manipulate the output (of the previous transform) as much as you want.

ASTs are being sent from transform to transform, so if it's compatible you can re-use it (for performance).

Not sure why you'd ever wanna extend a transform, you could possibly re-use internals of another transform although that is probably risky as that's usually outside the scope of the transforms public interface and therefore not protected from change in minor releases.

tripodsan: Take the example, where I want to rewrite some tags in the html files. I assume parcel will use the ootb posthtml transformer for finding the referenced assets (CSS,JS). so I would just add my own transformer, based on posthtml, or traverse the AST myself? And add this to the transformer chain for HTML via API or config?

DeMoorJasper: I have no clue what ootb means, but this is how it would probably work.

.parcelrc
```json
{
  "transforms": {
  	"*.html": ["@parcel/transform-posthtml", "parcel-transform-something"]
  }
}
```

Parcel will run through it's own stuff collecting deps,... than `parcel-transform-something` would run in the same thread on the same asset, with the output of the previous transform as it's input. You could traverse the ast to get some custom dependencies and I think you can also modify dependencies.

This is how plugins work posthtml as well, so it should be no different. Everything should be possible.

DeMoorJasper: The API will also search for the parcelrc file or default config as far as I know.

It will also take an options object similar to the one parcel uses now but slightly different as you can see some cli flags are different in this RFC.

Not sure if you'll be able to provide a hardcoded config.

tripodsan: > Not sure if you'll be able to provide a hardcoded parcelrc config.

We want to use parcel as packager within our CLI tool, and ensure that certain transformers are present. with the 1.x version, we just register the asset types directly via API. so it would be important to be able to do the same for the transformers as well. 

something along the lines of:

```
let tx = bundler.getTransformers('*.html');
tx.push(myCoolTransformer);
bundler.setTransformers('*.html');
```

(but maybe we should move this discussion into an issue)

DeMoorJasper: No need to create a new issue, this is one big RFC for parcel 2.
So it's already in the right place.

The API should probably also be documented in this spec.

PARCEL_2_RFC.md
KeineLimonade: This is already happening, but this references are not inline by default. They are, however, bundled and processed.

So I believe this usage is implied, though not explicitly stated.

tom-julux: That is already happening, but the references are not inline by default. They are, however, bundled and processed.

So I believe this usage is implied, though not explicitly stated.

PARCEL_2_RFC.md
Daniel15: @mohsen1 That looks great to me. Ideally Parcel could handle falling back to a local module if the external one fails to load, rather than having to manually write the code for that (see my comment in https://github.com/parcel-bundler/parcel/issues/144) but perhaps that's outside the scope of Parcel?

PARCEL_2_RFC.md
peanutbother: how about a configuration-check to test if the `.parcelrc` is a JSON format or otherwise try to evaluate it as JS file to dynamically parse the configuration, like in webpack?

example:

### JSON format
```
{
  "resolvers": ["@parcel/resolver-default", "./config/resolvers"],
  [...]
}
```

### JS format
```
const myCustomResolverConfig = require("./config/resolvers");

module.exports = {
  "resolvers": ["@parcel/resolver-default", ...myCustomResolverConfig ],
  [...]
}
```

if we do like that and parcel detects a JS-Config, it should default to parse config and override tool-specific configuration. then the tool itself gets the original config, overriden by parcel with the override-values of it.

so e.g. `.babelrc` contains a `transforms`-field but we explicitly override it in `.parcelrc`, then the final configuration would be the diff of both, where `.parcelrc` takes priority.

PARCEL_2_RFC.md
jamiebuilds: The apis are not finalized yet

PARCEL_2_RFC.md
jamiebuilds: this is an example

PARCEL_2_RFC.md
jamiebuilds: We will not be supporting a dynamic config format in JS or any other turing complete language. If tools do not support being configured the way you want, you'll have to bring that up with them.

peanutbother: Let's hope tool authors will get on the parcel train and start introducing better APIs when parcel v2 comes out.
But maybe it introduces new tools which target such scenarios when using parcel and that's what I'm hoping for personally.

mischnic:
Will there be a standardised way for plugins to have a config inside `.parcelrc`?

peanutbother:
> 
> 
> Will there be a standardised way for plugins to have a config inside `.parcelrc`?

I think this would be a counter productive choice since it would clutter up parcel specific configuration, since every tool should bring its own configuration.

Maybe you should add your comment as review / request

DeMoorJasper:
@mischnic this is kind of against what we are trying to do as far as I know. parcelrc just contains the plugins as described in this RFC, pretty sure it should not contain any plugin specific config.

You might try to convince otherwise though

mischnic:
> .parcelrc just contains the plugins as described in this RFC

Generally, I think this is a good decision, the used components and their configurations are cleanly separated (as opposed to the huge webpack config). But having loads of dotfiles (.babelrc, .postcssrc, .browserslistrc, cssnano.config.js, .posthtmlrc, .env, .mypluginrc) isn't a pretty solution either.
Anyway, I'll just use a separate config file.

jbreckmckye:
> Parcel has always and will always work out of the box for many projects with zero configuration. It should always be extremely simple to get started. But if you do want more control, we give you the tools to do so.

But why? Webpack already exists. I think Parcel has a perfect niche as a 'first choice' tool for simpler projects and smaller teams that don't want the hassle of configuring a build system.

I can imagine a few harmful effects of adding configuration files:

1. We have to reconcile new features with support for various config options. When conflicts arrive, we must either make breaking changes or accept compromises to our designs.
2. We take away the incentive to solve problems in a config-free way. 
   - Consider the way parcel follows hrefs to discover new html files to build. This is ingenious, highly intuitive, and makes the most of HTML's inbuilt semantics - but would we have discovered it if not for necessity? It would have been extremely easy to add an `entrypoints` array to `.parcelrc` instead, and thus re-invent a worse Webpack.
3. Parcel will try and beat Webpack on its own turf, and lose. The zero-config message will be diluted, and the project will no longer present a clear USP.

devongovett:
Merged the RFC doc into master. Further comments can continue here, and any changes should be submitted as PRs.

PARCEL_2_RFC.md
afridley: Will this work with out a url? In case you want to replace a module (@wordpress/blob) with a value (wp.blob) to be used to retrieve a global (wp.blob) variable. 

example from webpack
```js 
{ 
  externals: { 
    '@wordpress/blob': 'wp.blob'
   }
}
```

Or will this type of behavior not be supported by aliases?

PR #1955, opened 2018-08-29
Babel 7

### LINKED ISSUE #868: Babel 7 support
## Feature request

Support Babel 7

devongovett:
This updates to use Babel 7 internally. Babel 6 will also be supported for projects that haven't upgraded yet. Fixes #868.

To do:

- [x] Support both Babel 6 and 7 at the same time.
- [x] Automatically install babel-core v6 in a project if we detect babel 6 plugins.
- [ ] Load configs using Babel's helper methods
- [ ] Use babel-preset-env's new polyfill `useBuiltIns: "usage"` option.
- [x] Ensure all existing tests pass.
- [x] Add tests to ensure babel 6 is still supported.

If anyone wants to help with any of these items, please make PRs against the babel7 branch. Thanks! 🙏


src/visitors/matches-pattern.js
```diff
@@ -1,7 +1,7 @@
-const types = require('babel-types');
+const types = require('@babel/types');
 
-// from babel-types. remove when we upgrade to babel 7.
-// https://github.com/babel/babel/blob/0189b387026c35472dccf45d14d58312d249f799/packages/babel-types/src/index.js#L347
+// from @babel/types. remove when we upgrade to babel 7.
+// https://github.com/babel/babel/blob/0189b387026c35472dccf45d14d58312d249f799/packages/@babel/types/src/index.js#L347
```
devongovett: TODO: remove this file now that we are on babel 7 :)

devongovett:
I believe there may be AST changes, not 100% sure though. cc. @hzoo @loganfsmyth

I was going to use babel 7 for all our internal parsing needs. Many parcel projects don't need any custom babel plugins at all since we generate a config using preset-env and preset-react already. But for those that do have a .babelrc, I was going to try to detect the versions of the plugins (from package.json) and if they are v6, then install babel-core v6 locally in the project and use that to run the custom transforms. If the ASTs aren't compatible, we'll stringify with babel 7 and reparse with babel 6.

hzoo:
👋 There are a few AST changes but it's small https://babeljs.io/docs/en/v7-migration-api#ast-changes, mostly to match estree https://babeljs.io/docs/en/v7-migration-api#removal

devongovett:
The AST changes look pretty minimal. I wonder if it would be faster to traverse and convert the AST to a babel 6 compatible one and back than to generate + parse + generate + parse. Should be pretty easy to write a converter.

hzoo:
That would be essentially what babel-eslint does (parse with babel and then modify it to support espree/eslint), except it's for v6. https://github.com/babel/babel-eslint/blob/6aa8b6f02ff83cfb14eae2432f226f113929a50f/lib/parse.js#L90 Ideally we would just output a v6 compat mode so it would be faster but yeah not sure how there  would be a good way of doing this.

olstenlarck:
Yea, parse with Babel 7 always, then convert them back to v6, because changes are really minimal. Rest would be just the same.

I'm not very familiar with the codebase though, but just thinking. ;d ;]

package.json
```diff
@@ -15,6 +15,17 @@
...
+    "@babel/plugin-transform-flow-strip-types": "^7.0.0",
+    "@babel/plugin-transform-modules-commonjs": "^7.0.0",
+    "@babel/plugin-transform-react-jsx": "^7.0.0",
+    "@babel/preset-env": "^7.0.0",
+    "@babel/template": "^7.0.0",
+    "@babel/traverse": "^7.0.0",
+    "@babel/types": "^7.0.0",
     "ansi-to-html": "^0.6.4",
     "babel-code-frame": "^6.26.0",
     "babel-core": "^6.25.0",
```
DeMoorJasper: Shouldn't this be localRequired as we'll use Babel 7 for everything internal?

DeMoorJasper: WIP for this: #1958

olstenlarck:
Sweet. Feels like someone heard me ;p 

Last commits looks great :tada:

taylorjdawson:
How close is this to being merged? @devongovett

devongovett:
Some additional refactoring work is going on here: https://github.com/parcel-bundler/parcel/pull/1999.

One blocker: we need to figure out how to detect if a project is using babel 6 or 7. The initial approach was to check if any plugins contained the `@babel/` scope, but this won't work if a babelrc only contains third party plugins. Since most existing Parcel projects won't include `babel-core` in their package.json, we have no way of knowing the version they want... Any ideas here would be appreciated.

wongmjane:
@devongovett Re: Babel 6/7 detection

Are config that indicates the Babel version allowed?

package.json
```jsonc
{
  // ...
  "parcel": {
    "babel": 7
  }
}
```

olstenlarck:
Simple config option is the way to go. It is future-proof and most minimalist and elegant. Worth nothing and such configuration won't scare people and won't beat the "zero config" title.

Make it work with Babel 7 by default, and if users want 6 then say to them to add `babel: 6` option (would be good to allow through `--babel 6` cli flag or such). Explicitness is always a good thing. And in most cases, landed correctly, don't hurts.

devongovett:
If we have any config at all it will be in package.json as an explicit dependency on `babel-core` or `@babel/core`, not a CLI flag or a parcel specific config.

The problem we currently face is that we don't currently require this, so when existing apps upgrade to the new Parcel version, which babel do we use? I guess we use babel-core and make them manually upgrade to `@babel/core`? But what about new projects that want babel 7 out of the box?

Ideally we could look at the requested plugin's `peerDependencies` to determine their supported babel versions, but I doubt most plugins declare this dep.

olstenlarck:
I definitely don't understand what is the whole problem and why you are hitting your head in the walls for such things :laughing: :rocket: 

> If we have any config at all it will be in package.json 

Of course. That's the best way.

> as an explicit dependency on babel-core or @babel/core

That's good too. But if they don't have it as dep/devDep/peerDep then it should fallback to look for `parcel.babel` in package.json. And if there is nothing, then just use the Babel 7.

That's what i'm thinking and is based on what i've seen in that PR.
- Internally you switched to use Babel 7 by default, converting it to babel 6 AST, applies the magic, converts back to v7 and "stringify", right?
- Then what's te problem? News users are just using latest and greatest
- Old users that expect v6 will:
  + have v6 as dep/devDep/peerDep
  + or config option in package.json
  + or flag as last fallback possibility (least amount of migration and "on demand")


The strategy should be the core always to be with latest stable release and deal internally with the tweaks between the major Babel versions. And provide users with the least needed migration path with just saying to them to bump their babel version in the `parcel.babel` option.

Currently, switching from v6 to v7 is kinda "easy", because the package name changes. But in future all new versions will be in the `@babel` namespace, so.. it won't be so easy and for sure in one point in future you will need to change more in the internal logic than if you apply this simple strategy today.

It's would be absolutely future proof and most easier even for the v10 users. When Babel 8 come you will switch to the core to v8 and allow users to jump with `parcel.babel: 8`. When Babel 9 come then - same thing. And so on.

Definitely worth the cost of adding one config option to the "zero config" idea. Then you will just need to have strict policy to not add more "configurability" than that.

nuschk:
What's wrong with something like this?

```js
function getBabelVersion() {
  try {
    return require('@babel/core').version;
  } catch (e1) {
    try {
      return require('babel-core').version;
    } catch (e2) {
      console.error('No babel found');
      return undefined;
    }
  }
}
```

You could also think about failure modes, i.e., when multiple babel versions are installed (though that wouldn't be usable e.g. for 7.0 vs 8.0, so maybe it's really not an issue).

DeMoorJasper:
@nuschk It's currently not required to have babel as a local dependency when using parcel, therefore making this kind of a breaking change.

src/transforms/babel/babelrc.js
```diff
@@ -0,0 +1,219 @@
+const {buildRootChain} = require('@babel/core/lib/config/config-chain');
```
loganfsmyth: These are not public APIs. Doing this means your code could break at any time.

devongovett:
Merged the refactor here - splits up the babel transform into several files.

We now try to infer the version of babel to use. If you have `@babel/core` or `babel-core` in your package.json, we will use the version from there. If you had an older parcel app and didn't manually add `babel-core` to your app's package.json (relied on the internal parcel one), or you make a new app and add a `.babelrc` without also installing babel yourself, we try to infer the babel version from the plugins you use.

We look at the versions of dependencies/peerDependencies/devDependencies for each plugin and try to find the versions of core babel packages. If this fails, we default to babel 7 and show a warning. We then install the version of babel we inferred and save it to your package.json so we don't need to infer again the next time.

src/transforms/babel/babelrc.js
loganfsmyth: And that sucks, because when people see errors about this being a missing file, they'll come yell at us, not you.

devongovett: oops, I was experimenting. Those imports aren't used.

loganfsmyth: Ahh alright sorry, I've seen real code doing this in other projects and it's endlessly frustrating.

src/transforms/babel/babelrc.js
```diff
@@ -0,0 +1,219 @@
...
+  }
+
+  // Otherwise, don't load .babelrc for node_modules.
+  // See https://github.com/parcel-bundler/parcel/issues/13.
+  return null;
+}
+
+async function findBabelRc(asset) {
+  // TODO: support babelignore, etc.
+  return await asset.getConfig(['.babelrc', '.babelrc.js'], {
```
devongovett: @loganfsmyth would be nice if babel could expose an API for loading babelrc that returned it raw instead of normalizing. Unfortunately, using `loadPartialConfig` returns a result that isn't compatible with babel 6. So I will need to reimplement support for babelignore, etc. myself, which isn't great...

src/transforms/babel/babelrc.js
```diff
@@ -0,0 +1,219 @@
...
+const EXACT_RE = /^module:/;
+const BABEL_PLUGIN_PREFIX_RE = /^(?!@|module:|[^/]+\/|babel-plugin-)/;
+const BABEL_PRESET_PREFIX_RE = /^(?!@|module:|[^/]+\/|babel-preset-)/;
+const BABEL_PLUGIN_ORG_RE = /^(@babel\/)(?!plugin-|[^/]+\/)/;
+const BABEL_PRESET_ORG_RE = /^(@babel\/)(?!preset-|[^/]+\/)/;
+const OTHER_PLUGIN_ORG_RE = /^(@(?!babel\/)[^/]+\/)(?![^/]*babel-plugin(?:-|\/|$)|[^/]+\/)/;
+const OTHER_PRESET_ORG_RE = /^(@(?!babel\/)[^/]+\/)(?![^/]*babel-preset(?:-|\/|$)|[^/]+\/)/;
+const OTHER_ORG_DEFAULT_RE = /^(@(?!babel$)[^/]+)$/;
+
+function standardizeName(type, name) {
```
devongovett: would be cool if babel could expose this function as a util

Andarist:
> Babel 6 doesn't support .babelrc.js files, so just the mere presence of it indicates that Babel 7 is in use.

Not quite true, I've used successfully `.babelrc.js` with babel@6 by putting `  {"presets": ["./.babelrc.js"]}` in my `.babelrc` and I know I'm not the only one.

loganfsmyth:
That's fair. I hate that people do that, since a `.babelrc` has different rules, so calling it a `.babelrc` but loading it as a preset is misleading, but I guess it is what it is.

src/transforms/babel/babelrc.js
devongovett: This is used to infer the version of babel to use based on the plugins. I need to resolve each plugin's package.json to check its dependency on babel-core.

src/transforms/babel/babelrc.js
devongovett: > Can we fork config resolution based on the Babel version?

Maybe... but we do programmatically modify the config in some cases, so it would be nice not to need to have multiple versions of that code.

> Also, .babelrc.js files can export functions FYI

ok, we'd have to support that too.

> does this load .babelrc files relative to the file being compiled?

Yes...

src/transforms/babel/env.js
```diff
@@ -0,0 +1,65 @@
...
+    return null;
+  }
+
+  let key = JSON.stringify(targets);
+  if (envCache.has(key)) {
+    return envCache.get(key);
+  }
+
+  let plugins = presetEnv.default(
+    {assertVersion: () => true},
```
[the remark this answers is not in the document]
devongovett: Is there a helper function to run a preset function that I can use instead? I just need the list of plugins it produces.

devongovett:
> If there is a .babelrc, can we just search the text statically for "@babel" to see if it's a 7.x config?

We considered this, but it won't work if the config only contains third party plugins, not any core plugins.

> I'm still not sure I see anything, after the version is known, that would make loadPartialConfig unusable.

We do a bunch of config merging with the user specified babelrc and our internally generated configs. That code would need to work across multiple babel versions.

devongovett:
@loganfsmyth I opened an issue on babel to discuss adding a more granular API: https://github.com/babel/babel/issues/8715. I would like to move forward with basic babel 7 support in Parcel even without full support for all of the options (just `.babelrc`), and add more features later once we figure out a better config loading strategy.

devongovett:
Merged and published as a beta in v1.10.0-beta.0. Please help test this out and report any bugs you find! 🎉

olstenlarck:
Sweeet. So, in a new project it just works, right?

bkniffler:
I just switched to v1.10.0-beta.0 and except for updating the babelrc file to use the later @babel plugins, everything works good.

loganfsmyth:
@devongovett Cool, I'll respond to your issue when I get some time. I think some of they are reasonable requests. 

I think my fundamental issue with your approach here is that it assumes that it is even possible to autodetect the version of Babel in use. While that may be try between 6 and 7 in most cases, that doesn't seem like a thing we can rely on long-term when Babel releases further major versions. Along with that, there's no guarantee that, as users migrate to JS-based config files, that config files will be parsable for metadata in the way you're doing now. Especially if tools start adopting things like Yarn's new proposed resolution logic, I'd expect that all config files will transition to `require`ing plugins first, meaning that their aren't any paths available for you to resolve anyway, at runtime, so it seems like the vast amount of the work you'd need to do to detect the version would have to be performed as static analysis on the config files, without even running them, and that's not something Babel can help you with.

I'm 100% behind Parcel's desire to be zero-config, but it seems like you want to be both zero-config and un-opinionated, and those seem directly contradictory. I think you should just update to use Babel 7 as a new major bump and drop Babel 6, and bump your major any time any new dependency version comes out.

devongovett:
@bkniffler 

> except for updating the babelrc file to use the later @babel plugins, everything works good.

Does this mean something is broken using newer babel?

@loganfsmyth

> While that may be try between 6 and 7 in most cases, that doesn't seem like a thing we can rely on long-term when Babel releases further major versions.

One of the things mentioned as part of the babel 7 upgrade was that all of the plugins and presets now [have a peer dependency](https://babeljs.io/blog/2017/12/27/nearing-the-7.0-release.html#peer-dependencies-integrations) on the version of `@babel/core` that they need. This makes it pretty easy for us to auto detect the version of babel that Parcel needs to use. I think babel should enforce this requirement on third party plugins as well.

You're right that JS-based config files are problematic, and not really a great idea in general. Parcel will support them, but probably not quite as automatically as `.babelrc`. We cannot help auto install plugins, or detect the version of babel to use automatically in that case. I think if people want to use JS configs, then they are responsible for ensuring that all of their plugins are installed themselves, and that they have a dependency on babel core in their package.json prior to running Parcel.

src/transforms/babel/env.js
```diff
@@ -0,0 +1,65 @@
...
+    {
+      targets,
+      modules: false,
+      useBuiltIns: useBuiltIns ? 'entry' : false,
+      shippedProposals: true
+    }
+  ).plugins;
```
loganfsmyth: FYI, presets can have any of Babel's standard options, so you're throwing those away here. That means if we refactor this in the future to have other things, things will break in Parcel.

src/transforms/babel/config.js
```diff
@@ -0,0 +1,122 @@
...
+  if (!hasFlow) {
+    mergeConfigs(result, flowConfig);
+  }
+
+  return result;
+}
+
+module.exports = getBabelConfig;
+
+function mergeConfigs(result, config) {
```
loganfsmyth: This merge appears to just discard any other options that were specified in the Babel config.

src/transforms/babel/babelrc.js
```diff
@@ -0,0 +1,321 @@
...
+
+  if (typeof config === 'function') {
+    // We cannot support function configs since there is no exposed method in babel
+    // to create the API that is passed to them...
+    throw new Error(
+      'Parcel does not support function configs in .babelrc.js yet.'
+    );
+  }
+
+  for (let key of ['extends', 'overrides', 'test', 'include', 'exclude']) {
```
loganfsmyth: I think you'll want `'env'` in here too.

src/transforms/babel/babel7.js
```diff
@@ -0,0 +1,40 @@
...
+
+  config.code = false;
+  config.ast = true;
+  config.filename = asset.name;
+  config.cwd = pkg ? pkg.pkgdir : asset.options.rootDir;
+  config.babelrc = false;
+  config.configFile = false;
+  config.parserOpts = Object.assign({}, config.parserOpts, {
+    allowReturnOutsideFunction: true,
+    strictMode: false,
```
loganfsmyth: Assuming this is to allow parsing both scripts and modules, you'd likely be better off doing `sourceType: "unambiguous"`.

devongovett:
Also, @loganfsmyth @hzoo is the AST converter we wrote for babel 6 -> 7 and visa versa something that you think should exist in babel itself? I could see it being useful for other projects, and probably not something we should be maintaining in Parcel ideally.

src/transforms/babel/env.js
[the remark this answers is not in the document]
loganfsmyth: Could you point me to where you are processing these plugins? I'm having trouble finding it.

Part of my issue here is that the exact return value of a preset is not a public API, because a preset can also have other presets, and what we'd actually need is an API to fully flatten the configuration into a list of plugins, which is pretty much what `loadOptions` already provides. We wouldn't want to expose `makeAPI` because it exposes internal implementation details of Babel itself, and it's not obvious what anyone would actually be able to do with the result anyway, other than pass it right back to Babel again unprocessed.

loganfsmyth:
> One of the things mentioned as part of the babel 7 upgrade was that all of the plugins and presets now have a peer dependency on the version of @babel/core that they need. This makes it pretty easy for us to auto detect the version of babel that Parcel needs to use. I think babel should enforce this requirement on third party plugins as well.

I actually have very mixed feelings about our addition of peerDependencies and I think it yet to be seem whether they will be good or bad for us. Take even just the Parcel case, if the user installs Babel plugins in the project root and installs `@babel/core`, there's no guarantee that the version there will be the version that Parcel itself runs, since it might have it's own copy in `node_modules/parcel/node_modules/@babel/core`, meaning that the Babel version loaded by peerDependencies for the user's plugin would be different from the version that is actually calling Babel. That'd be different if Parcel itself wanted to require users to install `@babel/core` and itself had a peerDependency on `@babel/core`, but that's not happening. We've already got static assertions inside our own plugins either way, so in the end I'm not sure the peerDependencies in the plugins actually get us anything.

> You're right that JS-based config files are problematic, and I think they are not really a great idea in general. Parcel will support them, but probably not quite as automatically as .babelrc. We cannot help auto install plugins, or detect the version of babel to use automatically in that case. I think if people want to use JS configs, then they are responsible for ensuring that all of their plugins are installed themselves, and that they have a dependency on babel core in their package.json prior to running Parcel.

I understand where you're coming from, but I think for the average user, the expressive abilities of JS configs usually end up outweighing any benefits of configs being static files. If the user's config is representable as static JSON, you'll still get static JSON back from `loadPartialConfig`, so the benefit of static files is really only being able to process them without calling Babel itself, and I think that's just going to be a losing battle all around.

src/transforms/babel/config.js
devongovett: This is a very simple merge, yes. The first config to be passed is the user specified one. The others are all internally generated, and only have plugins and presets, so that shouldn't be a problem.

devongovett:
Parcel never calls user-specified plugins with its internally installed copy of `@babel/core`. It always uses the locally installed one specified by the app, which hopefully would match the peer dependencies of plugins. If one isn't installed locally in the app, we auto install one for the user. See https://github.com/parcel-bundler/parcel/blob/master/src/transforms/babel/babel7.js#L8-L10.

> the expressive abilities of JS configs usually end up outweighing any benefits of configs being static files

I disagree. I'm not sure what people are doing that they feel like they need to write a dynamic config with JS, but it seems like that will just be chaos similar to `webpack.config.js`, `gulpfile.js` and other past attempts.

That said, as long as there is a way to call into babel to resolve this JS-based config into a static config, it should be fine for Parcel. We won't be able to help the user with auto installation in that case, but that's fine I think. They already opted into advanced-mode by choosing JS config anyway.

src/transforms/babel/config.js
loganfsmyth: Fair point. I guess this more goes back to my other comment then. `preset-env` could at any time start using `overrides` or `env` or `test`/`include`/`exclude`, or add other options as top-level flags, and then it would cease to function on Parcel because Parcel is hard-coded to only carry over `plugins`.

loganfsmyth:
> Parcel never calls user-specified plugins with its internally installed copy of @babel/core

That's fair, but it only means that the reverse problem could exist with Parcel's `preset-env` since Parcel is the one that depends on `preset-env`, so the peerDependency will be Parcel's version, even though the Babel transform itself is being run by the user's copy. At the end of the day, it may not matter anyway, but I do think the peer dependency just adds to the confusion.

> I disagree. I'm not sure what people are doing that they feel like they need to write a dynamic config with JS, but it seems like that will just be chaos similar to webpack.config.js, gulpfile.js and other past attempts.

I'm sure we could talk about it for a while, but I'm not sure it's really worth it. From my standpoint, I don't consider those chaos.

> That said, as long as there is a way to call into babel to resolve this JS-based config into a static config, it should be fine for Parcel. We won't be able to help the user with auto installation in that case, but that's fine I think. They already opted into advanced-mode by choosing JS config anyway.

Yup, totally fair.

gaui:
v1.10.0-beta.1 works great - thank you!

defionscode:
@devongovett hmm for me it seems .babelrc and babel.config.js are completely ignored on `v1.10.0-beta.1`

loganfsmyth:
`babel.config.js` is definitely ignored at the moment, but `.babelrc` should work.

simonlc:
I don't understand if .babelrc is fully functional yet, but using the "env" format doesn't work, while .babelrc without it does work.

Doesn't work:
```
{
  "env": {
    "development": {
      "plugins": [
        "react-hot-loader/babel",
      ],
      "presets": [
        "@babel/env",
        "@babel/react"
      ]
    },
    "production": {
      "plugins": [
        ["transform-react-remove-prop-types", {
          "classNameMatchers": ["Component"],
          "removeImport": true
        }],
      ],
      "presets": [
        ["@babel/env", {
          "shippedProposals": true
        }],
        "@babel/react"
      ]
    }
  }
}
```

Works:
```
{
  "plugins": [
    ["transform-react-remove-prop-types", {
      "classNameMatchers": ["Component"],
      "removeImport": true
    }],
  ],
  "presets": [
    ["@babel/env", {
      "shippedProposals": true
    }],
    "@babel/react"
  ]
}
```

defionscode:
@gaui @loganfsmyth so in particular, it seems that if you supply `exclude` config option to a babel preset it doesn't actually exclude any plugins.  So something like 

```
      presets: [
        [
          '@babel/preset-env',
          {
            exclude: [
              'transform-classes'
            ]
          }
        ]
      ]
```

I will note that this is the same behavior in the current stable branch with babel < 7. Excludes are seemingly skipped. You can supply any random string to `exclude` and it will still build no problem when it really should be throwing a babel-related error IF it were reading that section in.

PR #1958, opened 2018-08-29
Various updates to Babel 7 branch

DeMoorJasper:
- Remove babel 6 dependencies from pkg.json (Some would be re-added to devDeps for babel6 compatibility tests in the future)
- Update internal Babel configs (for Node 6 compatibility) to use Babel 7
- Update hasPlugin to use arrays to check for both Babel 6 and 7 plugins


src/packagers/JSPackager.js
```diff
@@ -225,12 +225,11 @@ class JSPackager extends Packager {
       // Add source map url if a map bundle exists
       let mapBundle = this.bundle.siblingBundlesMap.get('map');
       if (mapBundle) {
-        await this.write(
```
DeMoorJasper: Discovered a potential bug with template strings in Babel 7. But I think it has already been reported. Feel free to check if anyone is up for it :)

It might have been an issue in the babel config I wrote, not sure.

package.json
```diff
@@ -27,17 +27,6 @@
...
-    "babel-generator": "^6.25.0",
-    "babel-plugin-transform-es2015-modules-commonjs": "^6.26.0",
-    "babel-plugin-transform-flow-strip-types": "^6.22.0",
-    "babel-plugin-transform-object-rest-spread": "^6.26.0",
-    "babel-plugin-transform-react-jsx": "^6.24.1",
-    "babel-preset-env": "^1.7.0",
-    "babel-template": "^6.26.0",
-    "babel-traverse": "^6.26.0",
-    "babel-types": "^6.26.0",
     "babylon": "^6.17.4",
```
devongovett: remove this too?

src/transforms/babel.js
```diff
@@ -203,21 +245,31 @@ async function getBabelRc(asset, isSource) {
 
     // If specified as an array, override the config with the one specified
     if (Array.isArray(babelify) && babelify[1]) {
-      return babelify[1];
+      return {
+        version: 6,
```
devongovett: I guess we should test this the same way as above - it's possible that there could be babel 7 plugins in here too.

src/transforms/babel.js
```diff
@@ -177,14 +192,40 @@ function mergeConfigs(a, b) {
...
+  );
 }
 
 function getPluginName(p) {
   return Array.isArray(p) ? p[0] : p;
 }
 
+function testPluginArrayForBabelScope(plugins) {
+  for (let plugin of plugins) {
+    if (getPluginName(plugin).startsWith('@babel/')) {
```
devongovett: Does babel 7 require the scope or will they automatically add it if you just write `env`? If so, then this check won't work.

src/transforms/babel.js
DeMoorJasper: Alright, just thought legacy, people won't use that for babel 7. But sure I'll make it check the plugins

src/transforms/babel.js
DeMoorJasper: Babel docs:

> You can still use the shorthand way of specifying a preset or plugin. However because of the switch to scoped packages, you still have to specify the @babel/ just like if you had your own preset to add to the config.

So yes it's required, although it's not 100% reliable as third party libraries will still have babel-....

package.json
DeMoorJasper: Forgot that one, thanks :)

src/transforms/babel.js
devongovett: ok

package.json
```diff
@@ -27,18 +27,6 @@
...
-    "babel-plugin-transform-es2015-modules-commonjs": "^6.26.0",
-    "babel-plugin-transform-flow-strip-types": "^6.22.0",
-    "babel-plugin-transform-object-rest-spread": "^6.26.0",
-    "babel-plugin-transform-react-jsx": "^6.24.1",
-    "babel-preset-env": "^1.7.0",
-    "babel-template": "^6.26.0",
-    "babel-traverse": "^6.26.0",
-    "babel-types": "^6.26.0",
-    "babylon": "^6.17.4",
     "babylon-walk": "^1.0.2",
```
DeMoorJasper: Should we remove this, or do a PR to their branch to update to `@babel/types` instead of `babel-types`

It might cause issues very rarely, but still might occur

devongovett: worth doing some perf testing. babylon-walk was WAY faster than babel-traverse for some cases where we don't need scope tracking and things.

DeMoorJasper: I'm pretty sure it will still be slower, haven't tested it. I did however open up a PR to babylon-walk https://github.com/pugjs/babylon-walk/pull/2

DeMoorJasper:
I'm not gonna be adding any more features to this PR, as it's becoming quite big and we should probably keep everything in one place. Feel free to merge whenever

src/assets/JSAsset.js
```diff
@@ -77,19 +78,38 @@ class JSAsset extends Asset {
...
     return options;
   }
 
   async parse(code) {
     const options = await this.getParserOptions();
-    // return babylon.parse(code, options);
-    return babelCore.parse(code, options);
+
+    if (options.babelVersion === 6) {
+      let babylon = await localRequire('babylon', this.name);
```
devongovett: Why do we still need babylon? I think we should always parse with babel 7 and convert the AST as needed. Our internal plugins will all be using babel 7, and some of them run before custom babel transforms.

src/transforms/babel.js
```diff
@@ -27,25 +27,34 @@ const JSX_PRAGMA = {
...
+    let babel6 = await localRequire('babel-core', asset.name);
+    let res = babel6.transformFromAst(asset.ast, asset.contents, config);
+    if (res.ast) {
+      asset.ast = res.ast;
+      asset.isAstDirty = true;
+    }
+
+    asset.ast = babelASTConverter(asset.ast, 6);
+
+    return babelTransform(asset, 7);
```
devongovett: Won't this result in transforming every file twice?

DeMoorJasper: Yes, but it will only run babel 6 transforms on the first run, convert to babel 7 and run the internal config. Thought this was the expected behaviour.
I might have made a mistake though, not entirely sure this is the proper way to do this.

src/assets/JSAsset.js
DeMoorJasper: I left it there for plugins that run on ast creation. Not sure if that's really necessary

src/transforms/babel.js
devongovett: oh I see. yeah that's how it should work.

PR #1968, opened 2018-08-31
Add support for Elm assets

benthepoet:
This PR adds support for assets written with Elm 0.19. Some notes are as follows.
- Elm assets require a full page reload when using HMR (updates can't be applied reliably). For this I added a `forceReload` flag to the `Asset` class so that any asset that requires a reload can be indicated as such. Aside from Elm assets, HTML assets also have this flag turned on.
- Minification is performed using a ~~two-pass~~ configuration as recommended by the official Elm docs (http://elm-lang.org/0.19.0/optimize)


kzc:
> Minification is performed using a two-pass configuration as recommended by the official Elm docs (http://elm-lang.org/0.19.0/optimize)

The two pass thing is due to a known unfixed regression in `uglify-js` in `pure_funcs` when the `rename` pass was introduced prior to the `compress` phase. Even so, mangle and compress can still work correctly with just a single uglify minify invocation by disabling this rename pass - see https://github.com/rtfeldman/elm-spa-example/pull/54

This `pure_funcs` `rename` bug is not present in [`terser`](https://github.com/fabiosantoscode/terser), as the `rename` pass is internally disabled - its flags are just no-ops. As result only a single minify invocation is needed.

benthepoet:
Thanks @kzc, I wasn't aware of that. I've updated it so it's just a single-pass now.

src/assets/ElmAsset.js
```diff
@@ -0,0 +1,93 @@
...
+            'A6',
+            'A7',
+            'A8',
+            'A9'
+          ],
+          pure_getters: true,
+          unsafe: true,
+          unsafe_comps: true
+        },
+        mangle: true
```
kzc: In the unlikely event that the Terser maintainer reintroduces the `rename` pass, please explicitly disable the option:
```js
         rename: false,
```

src/assets/ElmAsset.js
```diff
@@ -0,0 +1,93 @@
...
+      };
+
+      const {code} = minify(source, options);
+      return code;
```
kzc: The `minify` return `error` property should be examined. Terser will not throw upon `minify` error. Consult its API docs.

src/assets/ElmAsset.js
```diff
@@ -0,0 +1,94 @@
...
+          pure_getters: true,
+          unsafe: true,
+          unsafe_comps: true
+        },
+        mangle: true,
+        rename: false
+      };
+
+      const {code} = minify(source, options);
+      return code;
```
kzc: The `minify` return `error` property should be examined. Terser will not throw upon `minify` error. Consult its API docs.

kzc:
I noticed in the original docs for https://github.com/rtfeldman/elm-spa-example#step-2 that it specified the `compress` option `passes: 2` as an increased number of `compress` passes could potentially produce smaller minified output. It's up to you.

benthepoet:
@kzc I believe all the feedback should be accounted for now. I also set `compress` to perform 2 passes (probably helps to be consistent with the other example).

kzc:
Cool. I've never used Elm before - looks interesting. I should check it out.

test/integration/elm/elm.json
```diff
@@ -0,0 +1,24 @@
+{
```
DeMoorJasper: Parcel should probably be responsible for reading elm configs.

Or at least look for the configs so it can add it for invalidating cache.

benthepoet: That makes sense, I've updated the asset so that it will attempt to locate the `elm.json` file and add it as a dependency if found.

package.json
```diff
@@ -82,7 +82,9 @@
     "codecov": "^3.0.0",
     "coffeescript": "^2.0.3",
     "cross-env": "^5.1.1",
+    "elm": "^0.19.0-bugfix2",
```
DeMoorJasper: Perhaps change this to `^0.19.0` not sure why they're appending `-bugfix` instead of just bumping the patch version

src/assets/ElmAsset.js
```diff
@@ -0,0 +1,125 @@
...
+    this.hmrPageReload = true;
+  }
+
+  async collectDependencies() {
+    const {findAllDependencies} = await localRequire(
+      'find-elm-dependencies',
+      this.name
+    );
+    const dependencies = await findAllDependencies(this.name);
+    const packageFile = await findPackageFile(path.dirname(this.name));
```
DeMoorJasper: Change this to:
```Javascript
await this.getConfig(['elm.json'], {load: false});
```

This automatically recursively searches for the file and adds it to the deps, without reading it.

This way you can remove your helper function, cleans it up a little

src/assets/ElmAsset.js
```diff
@@ -0,0 +1,102 @@
...
+class ElmAsset extends Asset {
+  constructor(name, options) {
+    super(name, options);
+    this.type = 'js';
+    this.hmrPageReload = true;
+  }
+
+  async collectDependencies() {
+    const {findAllDependencies} = await localRequire(
+      'find-elm-dependencies',
```
devongovett: I think we can avoid installing this package directly if we use `elm.findAllDependencies` from the `node-elm-compiler` package, which already depends on `find-elm-dependencies`. This way only one package will end up getting installed in a user's app.

package.json
```diff
@@ -82,7 +82,9 @@
     "codecov": "^3.0.0",
     "coffeescript": "^2.0.3",
     "cross-env": "^5.1.1",
+    "elm": "^0.19.0",
```
devongovett: What is this dependency used for? I don't see it used anywhere in the code.

benthepoet: This provides the actual `elm` binaries that `node-elm-compiler` uses to compile the source files. The package isn't a dependency of `node-elm-compiler`, so unfortunately we have to pull it in.

devongovett: Ah I see. So since this is a dev dependency in parcel, so we might need to install it in a user's project as well, right?

benthepoet: Yes, that's correct.

src/assets/ElmAsset.js
```diff
@@ -0,0 +1,97 @@
...
+
+class ElmAsset extends Asset {
+  constructor(name, options) {
+    super(name, options);
+    this.type = 'js';
+    this.hmrPageReload = true;
+  }
+
+  async collectDependencies() {
+    await this.getConfig(['elm.json'], {load: false});
```
devongovett: doesn't seem like the result of this is used.

package.json
devongovett: ok, updated to auto install `elm` when the command isn't available globally.

devongovett:
Updated to auto install `elm` in the project if the command is not already installed globally. Also calls `elm init` to create `elm.json` if one isn't found.

PR #1976, opened 2018-09-03
Use babel-core bridge

codejamninja:
`babel-core@7.0.0-bridge.0` should be used instead of `babel-core@6.25.0` for babel7 compatibility.

https://babeljs.io/blog/2017/12/27/nearing-the-7.0-release#peer-dependencies-integrations

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
     "@babel/types": "^7.0.0",
     "ansi-to-html": "^0.6.4",
     "babel-code-frame": "^6.26.0",
-    "babel-core": "^6.25.0",
+    "babel-core": "^7.0.0-bridge.0",
     "babel-generator": "^6.25.0",
     "babel-plugin-transform-es2015-modules-commonjs": "^6.26.0",
     "babel-plugin-transform-flow-strip-types": "^6.22.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -28,7 +28,7 @@
     "@babel/types": "^7.0.0",
     "ansi-to-html": "^0.6.4",
     "babel-code-frame": "^6.26.0",
-    "babel-core": "^6.25.0",
+    "babel-core": "^7.0.0-bridge.0",
```
DeMoorJasper: This is basically `@babel/core` 7 right? So this contains breaking AST and types changes? That's kind of what we wanna prevent by converting ASTs or re-parsing the generated code.

codejamninja: No, this is not `@babel/core`. Please read the link I sent you. You are supposed to use it with `@babel/core` for backwards compatibility.

DeMoorJasper: It is, this is literally the entire code that package contains

```Javascript
module.exports = require("@babel/core");
```

devongovett:
We use `@babel/core` directly in the babel7 branch now.

codejamninja:
You're supposed to use both.

PR #1977, opened 2018-09-03
Use babel7 polyfill

codejamninja:
`@babel/polyfill` should be used instead of `babel-polyfill` for compatibility with babel7.

### DIFF
--- a/package.json
+++ b/package.json
@@ -84,10 +84,10 @@
     "ws": "^5.1.1"
   },
   "devDependencies": {
+    "@babel/polyfill": "^7.0.0",
     "@vue/component-compiler-utils": "^2.0.0",
     "babel-cli": "^6.26.0",
     "babel-plugin-transform-async-super": "^1.0.0",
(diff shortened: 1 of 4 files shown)


codejamninja:
This failed because there was a network error on the server. The tests are actually all passing 👍

DeMoorJasper:
Would be nicer if we would be using `useBuiltIns: "usage"` for preset env

codejamninja:
Does `useBuiltIns` work for browser applications? Also, until you support `useBuiltInts` can you use the correct polyfill. The old polyfill is breaking the code and the unit tests.

icopp:
`useBuiltIns` is intended for both node and the browser... See use of `targets` in the [docs](https://babeljs.io/docs/en/usage).

devongovett:
This is handled in the babel7 branch

sigerello:
@devongovett, any assumptions when `useBuiltIns: "usage"` will be supported on master?

PR #1997, opened 2018-09-09
Fix "serve" for relative public URL (Eg: ".", "./dist", "dist")

trongnd:


src/Server.js
```diff
@@ -9,6 +9,7 @@ const prettyError = require('./utils/prettyError');
 const AnsiToHtml = require('ansi-to-html');
 const logger = require('./Logger');
 const path = require('path');
+const fs = require('fs');
```
DeMoorJasper: Probably better to use the fs-utils that are part of parcel `./utils/fs`

DeMoorJasper:
Also please add an explanation in the PR with related issues, what it solves.
Possibly also how you solved it.

trongnd:
Yay. Sorry for missing details.
I'm trying to use relative public URL with parcel. The command would similar to:
`parcel serve src/index.html --public-url <relative-public-path>`
`<relative-public-path>` could be, e.g: `dist`, `./dist`, `.`
The built output is fine, but `serve` feature is not working correctly.

The currently logic of `serve` is:
https://github.com/parcel-bundler/parcel/pull/1997/files#diff-6f508a0628bff9c4c0d4a7bd1d5f40a5L55
```javascript
} else if (
  !pathname.startsWith(bundler.options.publicURL) ||
  path.extname(pathname) === ''
) {
   return sendIndex();
}
```
While `pathname` always starts with `/`, and `publicURL` doesn't start with `/`, so the condition is always `false`, that causes `serve` always sends `index` even on other file requests (`.js`, `.css`, etc...)
I updated so in case request URL has file extension, it would check for file existence, if the file exists, it send the file content otherwise send the index HTML.

Another issue I see that `sourceMappingURL` is using absolute path. That doesn't work with relative public URL in some cases. So I also updated to use only source map file name for `sourceMappingURL` in case public URL is relative.

src/Server.js
```diff
@@ -51,18 +52,27 @@ function middleware(bundler) {
...
-        req.url = pathname.slice(bundler.options.publicURL.length);
-        return serve(req, res, send404);
+        // convert related publicURL path to absolute path.
+        // eg: dist -> /dist, ./dist -> /dist, /dist -> /dist
+        let publicURL = `/${bundler.options.publicURL.replace(/^\.?\/?/, '')}`;
+        if (publicURL !== '/' && pathname.startsWith(publicURL)) {
+          pathname = pathname.slice(publicURL.length);
+        }
+        // check if file exists
+        fs.exists(path.join(bundler.options.outDir, pathname.slice(1))).then(
```
DeMoorJasper: Use await, it's cleaner

src/packagers/JSConcatPackager.js
```diff
@@ -530,10 +530,10 @@ class JSConcatPackager extends Packager {
       // Add source map url if a map bundle exists
       let mapBundle = this.bundle.siblingBundlesMap.get('map');
       if (mapBundle) {
-        output += `\n//# sourceMappingURL=${urlJoin(
-          this.options.publicURL,
-          path.basename(mapBundle.name)
-        )}`;
+        let sourceMappingURL = this.options.publicURL.startsWith('/')
```
DeMoorJasper: Why did you change this?

devongovett:

PR #1999, opened 2018-09-10
[WIP] Refactor babel support for babel 7

devongovett:
Splits up the babel transform into several files, and groups configs by babel version. WIP


src/transforms/babel/babelrc.js
```diff
@@ -0,0 +1,83 @@
...
+  ) {
+    return 7;
+  }
+
+  return 6;
+}
+
+function hasBabel7Plugin(plugins) {
+  return plugins.some(plugin => 
+    getPluginName(plugin).startsWith('@babel/')
```
devongovett: I don't think this way of detecting babel 7 is going to work. If a babelrc only contained third party babel 6 plugins, they wouldn't contain the `@babel/` scope. We need to find a better way of doing this...

src/assets/JSAsset.js
```diff
@@ -102,14 +100,15 @@ class JSAsset extends Asset {
...
-    const options = await this.getParserOptions();
-
-    if (options.babelVersion === 6) {
-      let babylon = await localRequire('babylon', this.name);
-      return babylon.parse(code, options);
-    } else {
-      return babelCore.parse(code, options);
+    // const options = await this.getParserOptions();
+    // return babelCore.parse(code, options);
+    await babel(this);
```
devongovett: Moved babel into parse. Its the first thing to need the AST, so makes sense for parsing to happen as part of babel's normal processing. If babel doesn't get applied to that file, then we fall back to parsing with `@babel/parser`.

src/transforms/babel/babelrc.js
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  // Otherwise, don't load .babelrc for node_modules.
+  // See https://github.com/parcel-bundler/parcel/issues/13.
+  return null;
+}
+
+async function findBabelRc(asset) {
+  // TODO: support babelignore, etc.
+  return await asset.getConfig(['.babelrc', '.babelrc.js'], {
```
devongovett: I really want to use babel's built in config loading helpers here, but the ones in 7 produce normalized configs that aren't compatible with babel 6 😢. There is no exposed function to load an unmodified config. So we'll end up needing to reimplement all of that ourselves.

src/transforms/babel/babelrc.js
vikr01: Not sure what this is for, but a good way of doing this is to check the actual versioning of `babel-core` or `@babel/core`.

Example:
```javascript
const {peerDependencies, dependencies, devDependencies} === require('@babel/preset-react/package.json') ;
let babelVersion;
let babelCore;
if(peerDependencies['@babel/core']) {
  babelCore = peerDependencies['@babel/core'];
}
[babelVersion] = babelCore.split('.');
```

devongovett: Yeah that would be nice, but I doubt most plugins actually declare this dependency in their package.json.

vikr01: For testing it, they would've needed it in their `dependencies` or `devDependencies` (unless it was a rare case where it was a subpackage of a monorepo with the `babel-core` or `@babel/core` version being added to the root). So that could be used as a backup.

PR #2002, opened 2018-09-11
bundle name hash-key generation is not environment independent #2001

### LINKED ISSUE #2001: bundle name hash-key generation is not environment independent
# 🐛 bug report

Parcel uses the full path of the files to create the hash-key. for projects that like to commit their generated files, this produces different outcomes depending on environment.

## 🤔 Expected Behavior

hash-key of bundles are independent of the directory the project is in.

## 😯 Current Behavior

hash-key of the bundles depend on the parent directory name.

## 💁 Possible Solution

compute hash to the relative path to the current directory.

## 🔦 Context

https://github.com/adobe/helix-cli/issues/180

tripodsan:
Closes #2001

### DIFF
--- a/src/Asset.js
+++ b/src/Asset.js
@@ -235,7 +235,7 @@ class Asset {
   generateBundleName() {
     // Generate a unique name. This will be replaced with a nicer
     // name later as part of content hashing.
-    return md5(this.name) + '.' + this.type;
+    return md5(this.relativeName) + '.' + this.type;
   }
 
   replaceBundleNames(bundleNameMap) {


DeMoorJasper:
What does this fix, please describe a description...

tripodsan:
I don't think that the failing test is related to my changes:

```

 440 passing (6m)
  1 failing
  1) wasm
       --target=node
         should load a wasm file in parallel with a dynamic JS import:
     Error: Cannot find module 'add.wasm'
```

DeMoorJasper:
@tripodsan It isn't, wasm tests are just instable

PR #2018, opened 2018-09-15
General documentation edits.

skyfriends:
---
name: 🙋 General documentation edits
---


skyfriends:
@devongovett 
The edits on `parcel` to `parcel-bundler` are a just a suggestion.  I have been acting as a teaching assistant in some web development classes, and have found that people tend to attempt adding packages to their application if it's represented as pre/code formatted text.  Just figured this might help with swooping up more folks to join the community 👍

devongovett:
Let's not change `parcel` to `parcel-bundler`. We actually already own the `parcel` name on npm as well, so that will work as well. We will switch over to `parcel` only in parcel 2.

skyfriends:
I didn't realize that -- awesome.  Can't wait for parcel 2 !

PR #2020, opened 2018-09-16
Cross-platform deterministic asset ids

Levertion:
Makes the deterministic asset ids introduced in #1694 not dependant on platform.

On windows, `path.relative` uses `\`, whereas on other systems it uses `/`. Because of this, the hashes were different.

## 💻 Examples

For a file at `<root>/src/index.js`, `Asset::relativeName` would be transformed into `src\index.js` on windows, giving the identifier `"2gen"`.

On other platforms, where the platform seperator is `/`, `Asset::relativeName` would be `src/index.js` as expected, with the identifier `"H99C"`.

This is causing unwanted build failures for my project when `dist` is committed to master, which I am doing to allow my typescript project to be used as an npm module directly from github. See Levertion/mcfunction-langserver#69

## ✔️ PR Todo

I haven't added any tests as #1694 did not change any tests and was included in a patch release => asset identifiers are not guaranteed to be backwards compatible. 

Additionally, I am not sure how I would actually test this behaviour, given that it uses the platform specific `path` module directly.

### DIFF
--- a/src/Asset.js
+++ b/src/Asset.js
@@ -21,7 +21,7 @@ class Asset {
     this.id = null;
     this.name = name;
     this.basename = path.basename(this.name);
-    this.relativeName = path.relative(options.rootDir, this.name);
+    this.relativeName = path.relative(options.rootDir, this.name).replace(/\\/g, "/");
     this.options = options;
     this.encoding = 'utf8';
     this.type = path.extname(this.name).slice(1);


DeMoorJasper:
Appears the tests are failing, could you fix that?

Levertion:
Sorry if my above comment wasn't clear - Do you want me to try and fix these tests which are broken on master?

DeMoorJasper:
@Levertion if you want to yes. Probably best to do it in a seperate pr

DeMoorJasper:
@Levertion it would be helpfull in reviewing if this breaks anything but it’s not a blocker no

PR #2032, opened 2018-09-19
🐛fixed detection of duplicate plugins for babel 7

anru:
---
name: 🐛 fixed detection of duplicate plugins for babel 7
---

## ↪️ Pull Request

This PR fixes bug in case you have following environment configured:

1. babel 7
2. any plugin from `babel-preset-env` list in your babelrc 

## 💻 Examples

Minimal example here is try to build project with `'@babel/proposal-object-rest-spread'` plugin included. I added test case for that example.

## ✔️ PR Todo

- [x] Looks like there is no related issues/PRs


src/transforms/babel/config.js
```diff
@@ -44,7 +44,12 @@ async function getBabelConfig(asset) {
     // Filter out plugins that are already applied by @babel/preset-env
     if (Array.isArray(babelrc.config.plugins)) {
       babelrc.config.plugins = babelrc.config.plugins.filter(plugin => {
-        return !ENV_PLUGINS[getPluginName(plugin)];
+        // we should test only packages starts with '@babel/' prefix here
+        const pluginName = getPluginName(plugin);
+        if (pluginName.indexOf('@babel/') === 0) {
```
devongovett: I don't think we want this if statement here. We also want to detect duplicates with the babel6 versions, right?

anru: Good point, I agree. But only way I see here is add `babel-preset-env` as dependency for importing different set of plugins for babel 6

src/transforms/babel/config.js
```diff
@@ -6,7 +6,14 @@ const path = require('path');
 const fs = require('../../utils/fs');
 
 const NODE_MODULES = `${path.sep}node_modules${path.sep}`;
-const ENV_PLUGINS = require('@babel/preset-env/data/plugins');
+
+const ENV_PLUGINS = {
+  // instead of installing huge `babel-preset-env` dependency we just copy 1 single file from that package
+  6: require('./data/babel-6-env-plugins')
```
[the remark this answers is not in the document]
anru: Oh, right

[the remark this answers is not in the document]
anru: @devongovett fixed, I've used `babel-upgrade`'s map for finding babel 7 plugins which matches corresponding plugins from babel 6. And that final plugin name is used for testing with intersection with `@babel/preset-env` which we are always use

test/javascript.js
```diff
@@ -85,6 +85,58 @@ describe('javascript', function() {
     );
   });
 
+  it('should handle duplicate plugins from preset env for babel v6', async function() {
```
devongovett: This test seems to pass, even without the code changes. Are you sure this PR is needed?

anru: This PR or this test ?

I know this test is passing without this PR (but not another one I've also added), but why we should exclude it, since it covers codebase for one more scenario ?

anru: second test `should handle duplicate plugins from preset env for babel v7` will not pass without this PR

anru: But I found this test is broken actually, also as getting corresponding babel 7 plugin name from babel 6. I'll fix it

anru: @devongovett fixed. Also I moved tests for `transforms/babel/config.js` to own test file `getBabelConfig.js`

anru: and tests are more accurate now

DeMoorJasper:
@anru babel-upgrade seems to be an unstable package. The tests are failing due to it

anru:
@DeMoorJasper ok, makes sense

anru:
@DeMoorJasper I removed all code related to removing babel 6's env plugins

devongovett:

PR #2036, opened 2018-09-19
migrate to monorepo

padmaia:
---
name: Migrate to monorepo
---

## ↪️ Pull Request

Planning for v2, parcel will be split into more modular packages that will live in the same repo. For now I have moved all of the v1 code into the `@parcel/core` package which is exposed by the `parcel-bundler` package. This will make it easy to work on the next version of core as a separate package in the `next` branch without messy conflicts when rebasing/merging from master. Things that can be reused in v2, like the watcher and logger, will be extracted and made into their own packages.

**Notes:**
* A common pattern for monorepos (babel, jest, lerna) is to run build tools from the root. I have done so with eslint and prettier, but when I tried to run mocha from the root, quite a few tests failed for reasons more complicated than broken paths due to code moving around. For now I have resorted to using `lerna run test`. One downside I have noticed is the output is not as nice. I was thinking that this work could be merged and I could then move onto getting tests to run from root (if that's what we want to do) as a separate task because rebasing a big change like this is not so fun, so I'm eager to get it in sooner than later. If it is deemed necessary to get done before merging, I can always add it to the PR todo list though.

* One thing I was not so sure about was the changelog. Should there just be one at the root, or should we use something like [lerna-changelog](https://github.com/lerna/lerna-changelog)?


devongovett:
Cool. Sorry for the conflicts. I'd like to get 1.10.0 out first, and then merge this.

devongovett:
@padmaia rebased again.

One thought: because of all these file moves, Github doesn't show history for any of the files anymore. 😢 I wonder if it would be worth rewriting the history instead so it looked like the files were always in the new locations? This way history on the file level would be preserved. WDYT?

devongovett:
also: the `monorepo` and `next` branches already have more packages set up, though they are a little behind master now...

padmaia:
@devongovett Hey sorry, just got back from a week of vacation. I will look into maintaining history, I guess I falsely assumed it was smart enough to do that when it sees the files are not new but moved. Also, I was aware of the monorepo and next branches. I thought it would be easier to do things more incrementally and this was the first step. Once this is merged I plan on rebasing the `graph-building-cancellation` which will pull out the watcher as it's own package and add the core-v2 package. Then we can start pulling out the other packages incrementally as we wire them up in v2. Let me know if that sounds good or if you have a better plan in mind.

padmaia:
@devongovett I looked a bit into rewriting history. It's possible, but not exactly easy. I found this article to be the most helpful explaining what would need to happen (https://medium.com/@slhenty/git-tips-for-subtree-splitting-and-moving-560f3743e19c). Personally I'd prefer to not rewrite history, but I could be convinced.

Here's an issue for Github to support the `--follow` flag if anyone wants to give a thumbs up (https://github.com/isaacs/github/issues/900)

Issue #1736, opened 2018-07-15
Inline file in bundle with fs.readFileSync not working with TypeScript


alexharri:
# 🐛 bug report

Parcel provides inline asset importing via statically analyzable `fs.readFileSync` statements. They work as expected in JavaScript but not Typescript.

### package.json

```js
{
  "scripts": {
    "start": "parcel index.html"
  },
  "dependencies": {
    "parcel": "^1.9.6"
  },
  "devDependencies": {
    "@types/node": "^10.5.2",
    "typescript": "^2.9.2"
  }
}
```

No custom `.babelrc` file

## 🤔 Expected Behavior

A statically analyzable `fs.readFileSync` statement should inline the file in the bundle, both in `.js` and `.ts` files.

## 😯 Current Behavior

`fs.readFileSync` statements are inlined in `.js` files but aren't in `.ts` files.

Output from console (see code sample)

![Console output](https://i.gyazo.com/f2ec6e173768f4eab7cd8631936786d4.png)

```
index.ts:5 Uncaught TypeError: fs_1.default.readFileSync is not a function
    at Object.parcelRequire.index.ts.fs (index.ts:5)
    at newRequire (2.0d1d60a5.js:48)
    at parcelRequire...\node_modules\parcel\src\builtins\_empty.js (2.0d1d60a5.js:80)
    at 2.0d1d60a5.js:106
parcelRequire.index.ts.fs @ index.ts:5
newRequire @ 2.0d1d60a5.js:48
parcelRequire...\node_modules\parcel\src\builtins\_empty.js @ 2.0d1d60a5.js:80
(anonymous) @ 2.0d1d60a5.js:106
```

## 💁 Possible Solution

Implement inline file imports in `.ts` files.

## 🔦 Context

I was working with WebGL and writing shaders. Writing them with template strings didn't give me any syntax highlighting so I wanted to write them as `.frag` and `.vert` files. I expected to be able to import them easily with Parcel but received the above error. After a bit of head-scratching I realized it was a Parcel + TypeScript problem and made a repro.

There's a fairly simple workaround this issue, which is just creating a `.js` file which imports the desired file, for example.

```js
// index.ts

import textFile from "./textFile.js";

// Do stuff with the text file.
```

```js
// textFile.js

import fs from "fs";

const textFile = fs.readFileSync(__dirname + "/textFile.txt", "utf8");

export default textFile;
```

It works but could create a lot of noise if using a lot of inline assets.

## 💻 Code Sample

[...]

PR #2046, opened 2018-09-23
Add test to fs.readFileSync in TypeScript files (#1736)

hasparus:
---
name: Add test to fs.readFileSync in TypeScript files
---
```
parcel-bundler version 1.9.7
```

## 💥 Problem

Inlining file in browser mode doesn't work in TypeScript files when `fs` is imported with star import or default import. The only way to use `readFileSync` is:
```typescript
import { readFileSync } from 'fs';
const raw = readFileSync(__dirname + '/raw.tsx', 'utf-8');
```

### Reproduction: 
Repo: https://github.com/hasparus/parcel-readfilesync-typescript-repro
ghPages: https://hasparus.github.io/parcel-readfilesync-typescript-repro/
CodeSandbox: https://codesandbox.io/s/github/hasparus/parcel-readfilesync-typescript-repro/tree/master/

Fun Fact: All ways to import `fs` work on CodeSandbox (parcel template).

## ↪️ Pull Request
Adds test to `fs.readFileSync` in TypeScript files and thus provides better workaround for #1736.

## ✔️ PR Todo
- [X] Link in #1736.
- [X] Post reproduction to CodeSandbox.
- [ ] Document how to use `readFileSync` in TypeScript?


DeMoorJasper:
It's good to have extra tests, but it would probably be better if we fixed the actual issue you had with it. At least if it's supported in normal JS

DeMoorJasper:
About codesandbox, it's probably using an outdated version of parcel. And it has also been modified a bit to work in the browser as far as I know.

PR #2048, opened 2018-09-23
Remove wasm-gc from RustAsset

DeMoorJasper:
---
name: 🙋 New Pull Request
---

## ↪️ Pull Request

`rustc` is now able to handle everything `wasm-gc` did according to the readme of `wasm-gc` therefore I made this PR removing wasm-gc.

## ✔️ PR Todo


devongovett:
readme also mentions a `--gc-sections` flag that the rust compiler supports. do we need to use that?

devongovett:
cc @alexcrichton

alexcrichton:
Looks good to me! It's true that `wasm-gc` is no longer needed, and the Rust compiler automatically by default executes the equivalent of `--gc-sections` when creating output, so no need to pass any extra flags either!

PR #2058, opened 2018-09-25
add <image> as a valid element type to bundle in <svg> tags

### LINKED ISSUE #2055: <svg><image xlink:href="myfile.svg" isn't detected...
Hello, and thanks for the amazing work!

In my `index.html` entry point, If i try to embed my external `.svg` like this, **it works**:

```html
<img src="myfile.svg" />
<!-- output: dist/myfile.123456789.svg" -->
```

but if i try to embed it [like this](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/image), **it doesn't**:
```html
<svg>
  <image xlink:href="myfile.svg" />  	
</svg>
<!-- no output... -->
```

How can i solve this?

Thank you!

makmm:
fixes #2055

---
name: 🙋 New Pull Request
---

## ↪️ Pull Request

fixes #2055 

## 💻 Examples

```
<svg>
  <image xlink:href="myfile.svg" />  	
</svg>
```
should work correctly now

### DIFF
--- a/src/assets/HTMLAsset.js
+++ b/src/assets/HTMLAsset.js
@@ -12,6 +12,7 @@ const ATTRS = {
   src: [
     'script',
     'img',
+    'image',
     'audio',
     'video',
     'source',


src/assets/HTMLAsset.js
```diff
@@ -12,6 +12,7 @@ const ATTRS = {
   src: [
     'script',
     'img',
+    'image',
```
DeMoorJasper: Should probably be part of the array of key: `'xlink:href'`

DeMoorJasper: I got that but this would only fix it for `<image src='' />` nor for <image xlink:href="myfile.svg" />

DeMoorJasper: Could you also add a test for this. To make sure it's fixed and doesn't cause anything else to fail

makmm:
@DeMoorJasper I have fixed what you've said and also added a test.

PR #2059, opened 2018-09-25
treat <script type="text/x-template"> as html

### LINKED ISSUE #1200: 🐛Assets in HTML template in script with type "x-template" won't get bundled

No configuration, I just `parcel index.html`

### 🤔 Expected Behavior
I think assets in "x-template" script tag should also be bundled. 
Code:
```html
  <script type="text/x-template" id="template">
<div>
  <img src="./static/img.png" />
</div>
</script>
```
### 😯 Current Behavior

`img.png` won't get bundled. 

### 💁 Possible Solution

Don't konw

### 🌍 Your Environment

| Parcel           |  v1.7.1
| Node             | v9.7.1
| npm/Yarn         | yarn
| Operating System | macOS High Sierra 10.13.3

makmm:
fixes #1200

---
name: 🙋 New Pull Request
---

## ↪️ Pull Request

fixes #1200

## 💻 Examples

```
<script type="text/x-template" id="template">
  <div>
    <img src="./static/img.png" />
  </div>
</script>
```
will properly import `./static/img.png`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to parcel will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and parcel adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [1.10.1] - 2018-09-26
+
+### Fixed
+
+- Fix error when node.id is undefined instead of null
+
 ## [1.10.0] - 2018-09-25
 
 ### Added
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "parcel-bundler",
-  "version": "1.10.0",
+  "version": "1.10.1",
   "description": "Blazing fast, zero configuration web application bundler",
   "main": "index.js",
   "license": "MIT",
--- a/src/visitors/dependencies.js
+++ b/src/visitors/dependencies.js
@@ -106,7 +106,7 @@ function hasBinding(node, name) {
     types.isArrowFunctionExpression(node)
   ) {
     return (
-      (node.id !== null && node.id.name === name) ||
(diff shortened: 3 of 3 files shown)


src/assets/HTMLAsset.js
```diff
@@ -67,7 +67,8 @@ const SCRIPT_TYPES = {
   'text/javascript': 'js',
   'application/json': false,
   'application/ld+json': 'jsonld',
-  'text/html': false
+  'text/html': false,
+  'text/x-template': 'html'
```
DeMoorJasper: Shouldn't this be just false if it's html? It's probably the same as `text/html` right?

DeMoorJasper: Should have looked at the issue first... Thanks for fixing this.

devongovett:
Is `text/x-template` an actual standard? Not sure if we should be inventing our own script types.

makmm:
@devongovett I don't really know. I just tried solving the issue, but didn't check for standards. Do you want me to properly add tests, even if it isn't standard?

Hoishin:
`x-template` isn't standard. I think it should be treated as plain text (as it says, **text**/x-template), rather than treating it as HTML.

Most of the time when you want to treat `x-template` as HTML, you are doing something wrong.

makmm:
@Hoishin The thing is, the x-template thing is used for stuff like Handlebars, and inside that template you DO want to bundle assets.

Hoishin:
`text/(x-)template` is expected to be read by template engine you use, and it really depends on the template engine. It might be HTML, Pug-like syntax, mastache syntax, or anything the template engine expects.
If there was standard, bundlers could process it as per the spec, but just treating it as HTML would introduce unexpected behaviour.

devongovett:
Closing. I don't think there is a good way to support this in a generic way.

Issue #2054, opened 2018-09-25
Browser field gets resolved incorrectly


thsorens:
# 🙋 feature request

When using superagent, using import Request from 'superagent';  Parcel seems to pick whatever is in the lib/node folder. 

## 🤔 Expected Behavior

Default behaviour for webpack and browserify is to detect what to import, and targets the client.js file in lib/client.js

## 😯 Current Behavior
Parcel targets the lib/node folder, and exposes internal node functions to the client. Some calls fails because of that. (.on is not a function or similar node internal functions)

## 💁 Possible Solution
Use the same logic as webpack/browserify to detect what to fetch.

Current workaround is importing superagent/lib/client instead of just superagent

DeMoorJasper:
The pkg.json would be a good starting point for us to debug this and a link to the pkg’s git repo would also help

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

thsorens:
https://github.com/visionmedia/superagent

```JS
   import Request from 'superagent';
    
  const submitFile = (file) =>{
      const req = Request.post(absoluteUrl)
        .timeout(50000)
        .set('Accept', 'application/json';

      req.attach('file', file, file.name);
      req.end(err, res => {

       })

    }

```
https://github.com/visionmedia/superagent/blob/master/package.json


Tried to add some console.logs inside the imported module in node_modules/superagent, the one parcel imports is lib/node/index.js, which is intended for server-side use only. Most of it still works in parcel, since it doesnt use node-internals, but when trying to post a file, it will fail. 

Targeting lib/client.js works just fine. As stated in the feature request, webpack seems to figure out that it should use the client.js file somehow, since this was never a problem with webpack. 

According to superagent docs: 
```
Browser and node versions
SuperAgent has two implementations: one for web browsers (using XHR) 
and one for Node.JS (using core http module). 
By default Browserify and WebPack will pick the browser version.
```

DeMoorJasper:
Tried to recreate the issue with https://github.com/parcel-bundler/parcel/pull/2067 but I failed to recreate it. Seems to work fine, you might be running an outdated Parcel version?

thsorens:
Im running parcel-bundler: 1.9.7. I can see that 1.10.1 just got available, did that fix it?

thsorens:
Also worth mentioning, it crashes runtime. It builds fine

DeMoorJasper:
[...]

PR #2067, opened 2018-09-27
Add extra test case for resolving browser aliases in subdirectories

DeMoorJasper:
---
name: 🙋 New Pull Request
---

## ↪️ Pull Request

A test case to try and recreate #2054

## ✔️ PR Todo


devongovett:
@DeMoorJasper this test appears to be failing on Windows. Looks like it's using the node version instead of the client version there. Perhaps that's the root of the original bug reported by @thsorens?

thsorens:
> @DeMoorJasper this test appears to be failing on Windows. Looks like it's using the node version instead of the client version there. Perhaps that's the root of the original bug reported by @thsorens?

Yes, this was when running Windows

src/Resolver.js
```diff
@@ -392,6 +392,11 @@ class Resolver {
   }
 
   lookupAlias(aliases, filename, dir) {
+    filename = path.normalize(filename);
+    Object.keys(aliases).forEach(
+      key => (aliases[path.normalize(key)] = aliases[key])
+    );
```
devongovett: Probably not a good idea to mutate the package.json... Maybe do the normalization + checking in the existing for loop below instead

DeMoorJasper: Probably a better idea, will change it

devongovett:

PR #2073, opened 2018-09-27
Fix env transforms in node_modules

### LINKED ISSUE #1887: production minify code doesn't work well(npm run build), but development is ok(npm start)
npm scripts pls read package.json below

# 🐛 bug report

babelrc
```js
{}
```
package.json
```js
{
  "name": "xxx",
  "version": "1.0.0",
  "main": "index.js",
  "license": "MIT",
  "scripts": {
    "start": "cross-env NODE_ENV=development rimraf dist/development & parcel index.html --open -d dist/development/",
    "build": "cross-env NODE_ENV=production rimraf dist/production & parcel build src/index.js --public-url ./dist/production -d dist/production/ --detailed-report"
  },
  "dependencies": {
    "axios": "^0.18.0",
    "babel-plugin-transform-runtime": "^6.23.0",
    "jsonp": "^0.2.1",
    "normalize.css": "^8.0.0",
    "query-string": "^6.1.0",
    "scroll-to-element": "^2.0.0",
    "vue": "^2.5.16",
    "vue-hot-reload-api": "^2.3.0",
    "vue-router": "^3.0.1"
  },
  "devDependencies": {
    "@vue/component-compiler-utils": "^2.1.0",
    "babel-core": "^6.26.3",
    "cross-env": "^5.2.0",
    "cssnano": "^4.0.5",
    "node-sass": "^4.9.2",
    "pug": "^2.0.3",
    "vue-template-compiler": "^2.5.16"
  },
  "engines": {
    "node": "8"
  }
}
```
## 🤔 Expected Behavior

production minify code work normally。(if i minify development code by other minify tool, work well too )

## 😯 Current Behavior

production minify code doesn't work well, and have no error in console(page can't render normal, it's blank);  development code work well, ,

## 💁 Possible Solution
production minify bug?

## 🌍 Your Environment
parcel with vue project

| Parcel           | 1.9.7
| Node             | 8.9.4
| npm/Yarn         |5.6.0
| Operating System | windows 7

DeMoorJasper:
---
name: 🙋 New Pull Request
---

## ↪️ Pull Request

I'm not sure if this is what should happen, as there where 2 unit tests stating that the modules should not be converted to match the global browser versions. Which seems kinda strange to me as that is the only reason preset-env is being ran over node_modules afaik?

This PR, removes the fallback to engines.node for target detection in case of targeting browsers as this is often an invalid indicator for browser specific modules.

Closes #1887

## ✔️ PR Todo

### DIFF
--- a/src/transforms/babel/env.js
+++ b/src/transforms/babel/env.js
@@ -18,13 +18,16 @@ async function getEnvConfig(asset, isSourceModule) {
   // Otherwise, load the source engines and generate a babel-present-env config.
   if (!isSourceModule) {
     let sourceEngines = await getTargetEngines(asset, false);
-    let sourceEnv = (await getEnvPlugins(sourceEngines, false)) || targetEnv;
+    let sourceEnv = await getEnvPlugins(sourceEngines, false);
 
-    // Do a diff of the returned plugins. We only need to process the remaining plugins to get to the app target.
-    let sourcePlugins = new Set(sourceEnv.map(p => p[0]));
-    targetEnv = targetEnv.filter(plugin => {
-      return !sourcePlugins.has(plugin[0]);
-    });
+    // Do a diff of the returned plugins if there is a known environment for the source.
+    // We only need to process the remaining plugins to get to the app target.
+    if (sourceEnv) {
+      let sourcePlugins = new Set(sourceEnv.map(p => p[0]));
+      targetEnv = targetEnv.filter(plugin => {
+        return !sourcePlugins.has(plugin[0]);
+      });
+    }
   }
 
   return {
--- a/src/utils/getTargetEngines.js
+++ b/src/utils/getTargetEngines.js
@@ -51,11 +51,6 @@ async function getTargetEngines(asset, isTargetApp) {
         }
       }
     }
-
-    // Fall back to package.engines.node for node_modules without any browser target info.
-    if (!isTargetApp && !targets.browsers && typeof nodeVersion === 'string') {
-      targets.node = nodeVersion;
-    }
   }
 
   // If we didn't find any targets, set some default engines for the target app.
--- a/test/javascript.js
+++ b/test/javascript.js
@@ -1149,7 +1149,7 @@ describe('javascript', function() {
     );
   });
 
-  it('should not compile node_modules by default', async function() {
+  it('should make node_modules browser compatible', async function() {
     await bundle(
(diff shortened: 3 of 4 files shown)


DeMoorJasper:
Apparently the linux test-runners are down and the treeshaking tests are using unstandardised JS in some tests so babel is complaining. Not sure how to fix this.

fathyb:
As far as scope-hoisting goes I think it'll be okay as soon as it doesn't add the commonjs transform.
For the rest I'm not really sure, it makes sense to me but I'm not sure of the consequences. The tests are coming from #559, pinging @devongovett for advice!

devongovett:
Hmm I don't know that we want to change this. We currently assume that node_modules is already compiled by default, unless a `browserslist` or `engines.node` value indicates otherwise. Without either of those keys in package.json, I'm not sure how we could assume anything about how the code was written other than that it is already compiled. Turning on babel for all node_modules will slow down Parcel by a ton I would guess.

DeMoorJasper:
@devongovett originally thought the logic was assuring the entire bundle matches a certain browserslist. This is the only way I could think of achieving that as defining a browserslist or engines field in a package is optional.

Perhaps we could allow some way to enable this, for people who really want full browser compatibility?

Not sure if this is a good idea or not, seems to break the moment anything has unstandardised JS.

devongovett:
node_modules shouldn't contain non-standard code for sure, but historically node_modules has been ES5 as well, meaning pre-babelified. This ensures the most interoperability between tools. Parcel takes this as the default, and assumes that if no other information is specified, the code is already pre-compiled to ES5. However, people are starting to target newer versions of node/browsers in their node_modules, so we added support for specifying that in `engines.node` and `browserslist` fields. When those are specified, and they target higher versions than the app, we run babel over the module. Otherwise, babel is not run over node_modules.

DeMoorJasper:
Alright so people should just change their package.json, got it.

n0v1:
I just stumbled upon [are-you-es5](https://github.com/obahareth/are-you-es5) which tries to [parse each dependency with acorn](https://github.com/obahareth/are-you-es5/blob/dde3eaf202392b72131d6ec3d8dfd67cb1877482/src/modules-checker.ts#L106) to tell if it is valid es5 or not. I guess Parcel could check dependencies before bundling or the generated bundles with this or a similar approach and transpile them in case they do not fit the target environment. Maybe behind a flag for performance reasons.

PR #2076, opened 2018-09-28
add missing --no-autoinstall flag to build command

jamiebuilds:
Was running `parcel build` and the auto installer started running, I went to disable it and found that the option was not available here


DeMoorJasper:
@devongovett thought changing the environment variable could overwrite this, but looking at the cli code it does not. Should have probably double checked first.
Not sure if build and serve should behave differently in such a way, it's a tad confusing.

PR #2118, opened 2018-10-10
Elm debug support

### LINKED ISSUE #2104: Elm --debug flag for time travelling debugger
# 🙋 feature request

I'd like to be able to see time travelling debugger when using Elm in combination with Parcel. In plain elm it's achieved by passing `--debug` flag to compiler. 

## 💁 Possible Solution

1. It can be enabled by default for all development build unless there is some use case where it's not needed.
2. Or one of Parcel flags can be used to enable this feature

solution 1 - I can do PR for. 
Solution 2 I need some help to figure out which flags can be used.

Maxim-Filimonov:
Fixes #2104


test/elm.js
```diff
@@ -2,7 +2,7 @@ const assert = require('assert');
 const fs = require('../src/utils/fs');
 const {bundle, assertBundleTree, run} = require('./utils');
 
-describe('elm', function() {
+describe.only('elm', function() {
```
DeMoorJasper: Remove this `.only`

src/assets/ElmAsset.js
```diff
@@ -40,10 +40,10 @@ class ElmAsset extends Asset {
       await this.getConfig(['elm.json'], {load: false});
     }
 
+    options.debug = !this.options.minify;
```
DeMoorJasper: Shouldn't this use `!options.production`

Maxim-Filimonov: agree 👍

test/elm.js
Maxim-Filimonov: Ooops. Good catch

Maxim-Filimonov:
@DeMoorJasper something else is required for merge ?

devongovett:

PR #2119, opened 2018-10-10
use npm for auto-install by default

### LINKED ISSUE #2117: Parcel uses yarn by default
# 🐛 bug report

Continuing with https://github.com/parcel-bundler/parcel/issues/1923#issuecomment-428101900 I want to report that in absence  of package-lock or yarn-lock parcel uses yarn by default.

```
yarn add v1.9.4
info No lockfile found.
[1/4] Resolving packages...
```

As you can see, yarn reports that no lockfile is found so, why is it using yarn ?
This is specially problematic when you use private packages and you have the credentials configured on `.npmrc`

## 🤔 Expected Behavior

By default parcel should use npm which is the more or less standard

## 🔦 Context

This is only a problem on CI environment

## 🌍 Your Environment

| Parcel           | 1.9.7 |
| Node             | 8.11 |
| npm/Yarn         | 6 |
| Operating System | linux |

Hoishin:
# ↪️ Pull Request

Changed the logic to determine which package manager to use during auto-installation.

The only situation Parcel should use yarn is when it satisfies
- There is `yarn` command available
AND
- There is `yarn.lock` file
AND
- There is NOT `package-lock.json`

Otherwise, it uses npm.

Fixes #2117 

## 🚨 Test instructions

Setup a testing directory with different setups (package-lock existence, yarn.lock existence, yarn command existence).
The only situation Parcel should use yarn is when
- There is `yarn` command available
AND
- There is `yarn.lock` file
AND
- There is NOT `package-lock.json`

## ✔️ PR Todo

~~- [ ] Added/updated unit tests for this change~~

## Side note

There is a situation where there are both package-lock.json and yarn.lock. Having npm as default package manager means that it should use npm in this situation.

That's why there is only one situation where Parcel should use yarn.

### DIFF
--- a/src/utils/installPackage.js
+++ b/src/utils/installPackage.js
@@ -82,14 +82,11 @@ async function determinePackageManager(filepath) {
   const hasYarn = await checkForYarnCommand();
 
   /**
-   * There is package-lock.json -> Use npm
-   * There is yarn.lock & yarn command -> Use Yarn
-   * Otherwise -> Use npm
+   * The only situation Parcel should use yarn is
+   * there is yarn command AND there is yarn.lock AND there is NO package-lock.json
+   * Otherwise, it uses npm
    */
-  if (npmLockFile) {
-    return 'npm';
-  }
-  if (hasYarn && yarnLockFile) {
+  if (hasYarn && yarnLockFile && !npmLockFile) {
     return 'yarn';
   }
   return 'npm';
--- a/src/utils/installPackage.js
+++ b/src/utils/installPackage.js
@@ -75,11 +75,11 @@ async function installPeerDependencies(filepath, name, options) {
 }
 
 async function determinePackageManager(filepath) {
-  const [npmLockFile, yarnLockFile] = await Promise.all([
+  const [npmLockFile, yarnLockFile, hasYarn] = await Promise.all([
     config.resolve(filepath, [PACKAGE_LOCK]),
-    config.resolve(filepath, [YARN_LOCK])
+    config.resolve(filepath, [YARN_LOCK]),
+    checkForYarnCommand()
   ]);
-  const hasYarn = await checkForYarnCommand();
 
   /**
    * The only situation Parcel should use yarn is
--- a/src/utils/installPackage.js
+++ b/src/utils/installPackage.js
@@ -75,10 +75,9 @@ async function installPeerDependencies(filepath, name, options) {
 }
 
 async function determinePackageManager(filepath) {
-  const [npmLockFile, yarnLockFile, hasYarn] = await Promise.all([
+  const [npmLockFile, yarnLockFile] = await Promise.all([
     config.resolve(filepath, [PACKAGE_LOCK]),
-    config.resolve(filepath, [YARN_LOCK]),
-    checkForYarnCommand()
+    config.resolve(filepath, [YARN_LOCK])
   ]);
 
   /**
@@ -86,9 +85,16 @@ async function determinePackageManager(filepath) {
    * there is yarn command AND there is yarn.lock AND there is NO package-lock.json
    * Otherwise, it uses npm
    */
-  if (hasYarn && yarnLockFile && !npmLockFile) {
+
+  if (!yarnLockFile || npmLockFile) {
+    return 'npm';
+  }
+
+  const hasYarn = await checkForYarnCommand();
+  if (hasYarn) {
     return 'yarn';
   }
+
   return 'npm';
 }
 
--- a/src/utils/installPackage.js
+++ b/src/utils/installPackage.js
@@ -10,7 +10,6 @@ const fs = require('./fs');
 const WorkerFarm = require('../workerfarm/WorkerFarm');
 
 const YARN_LOCK = 'yarn.lock';
-const PACKAGE_LOCK = 'package-lock.json';
 
 async function install(modules, filepath, options = {}) {
   let {installPeers = true, saveDev = true, packageManager} = options;
@@ -75,18 +74,13 @@ async function installPeerDependencies(filepath, name, options) {
 }
 
 async function determinePackageManager(filepath) {
-  const [npmLockFile, yarnLockFile] = await Promise.all([
-    config.resolve(filepath, [PACKAGE_LOCK]),
-    config.resolve(filepath, [YARN_LOCK])
-  ]);
+  const yarnLockFile = await config.resolve(filepath, [YARN_LOCK]);
 
   /**
-   * The only situation Parcel should use yarn is
-   * there is yarn command AND there is yarn.lock AND there is NO package-lock.json
-   * Otherwise, it uses npm
+   * no yarn.lock => use npm
+   * yarn.lock => Use yarn, fallback to npm
    */
-
-  if (!yarnLockFile || npmLockFile) {
+  if (!yarnLockFile) {
     return 'npm';
   }
 
--- a/src/utils/installPackage.js
+++ b/src/utils/installPackage.js
@@ -9,6 +9,9 @@ const path = require('path');
 const fs = require('./fs');
 const WorkerFarm = require('../workerfarm/WorkerFarm');
 
+const YARN_LOCK = 'yarn.lock';
+const PACKAGE_LOCK = 'package-lock.json';
+
 async function install(modules, filepath, options = {}) {
   let {installPeers = true, saveDev = true, packageManager} = options;
   if (typeof modules === 'string') {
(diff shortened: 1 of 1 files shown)


src/utils/installPackage.js
```diff
@@ -72,19 +75,24 @@ async function installPeerDependencies(filepath, name, options) {
...
+  const [npmLockFile, yarnLockFile] = await Promise.all([
+    config.resolve(filepath, [PACKAGE_LOCK]),
+    config.resolve(filepath, [YARN_LOCK])
   ]);
-  let hasYarn = await checkForYarnCommand();
-
-  // If Yarn isn't available, or there is a package-lock.json file, use npm.
-  let configName = configFile && path.basename(configFile);
-  if (!hasYarn || configName === 'package-lock.json') {
+  const hasYarn = await checkForYarnCommand();
```
DeMoorJasper: This is no longer necessary, npm should be a fallback if we're using yarn. This slows down autoinstall

DeMoorJasper: On second thought this would only degrade performance if there aren't a lot of installs

Hoishin: Doesn't it throw error if there is not yarn command and yarn.lock exists?

Hoishin: well actually I have to include this to `Promise.all` then it shouldn't affect that much either

DeMoorJasper: Yeah, but if you catch the error and try again with npm if that happens it would be faster as that's what commandExists does, spawning processes is pretty expensive. But as the result is cached it probably doesn't matter that much you could however combine them. `if yarnLock => try yarn => catch => use npm & set hasYarn to false`

src/utils/installPackage.js
```diff
@@ -72,19 +75,21 @@ async function installPeerDependencies(filepath, name, options) {
 }
 
 async function determinePackageManager(filepath) {
-  let configFile = await config.resolve(filepath, [
-    'yarn.lock',
-    'package-lock.json'
+  const [npmLockFile, yarnLockFile, hasYarn] = await Promise.all([
+    config.resolve(filepath, [PACKAGE_LOCK]),
+    config.resolve(filepath, [YARN_LOCK]),
+    checkForYarnCommand()
```
DeMoorJasper: As I mentioned in my previous comment this checkYarn function is a pretty expensive operation and will slow down things. It is unnecessary if there isn't a yarn.lock and therefore would use processing power and slow down parcel for no reason.

resolving configs on the other hand are very fast.

Hoishin:
Updated to lazily validate yarn command.

DeMoorJasper:
@Hoishin I got a little issue with the logic of when to pick yarn.

As yarn supports `package-lock.json` they can co-exist and from the moment there is a `yarn.lock` I'm pretty sure people expect to be using yarn.

The logic I had in mind was:

`yarn.lock => Use yarn, fallback to npm`
`no yarn.lock => use npm`

I might be wrong here, and this is part of the original discussion in the issue this resolves as far as I know.

This is also a bit of a guessing thing, as you can never know the exact command anyone expects, but we can try to make it as smooth as possible.

Hoishin:
I used `yarn` command in a directory with package-lock.json.

```
/tmp/test-dir
❯ ls
node_modules      package-lock.json package.json

/tmp/test-dir
❯ < package.json
{
        "private": true,
        "dependencies": {
                "lodash": "^4.17.11"
        }
}

/tmp/test-dir
❯ yarn
yarn install v1.10.1
info No lockfile found.
warning package-lock.json found. Your project contains lock files generated by tools other than Yarn. It is advised not to mix package managers in order to avoid resolution inconsistencies caused by unsynchronized lock files. To clear this warning, remove package-lock.json.
[1/4] 🔍  Resolving packages...
[2/4] 🚚  Fetching packages...
[3/4] 🔗  Linking dependencies...
[4/4] 📃  Building fresh packages...
success Saved lockfile.
✨  Done in 0.78s.
```

Even if there is yarn.lock
```
/tmp/test-dir
❯ ls
node_modules      package-lock.json package.json      yarn.lock

/tmp/test-dir
❯ < package.json
{
        "private": true,
        "dependencies": {
                "lodash": "^4.17.11"
        }
}

/tmp/test-dir
❯ yarn
yarn install v1.10.1
warning package-lock.json found. Your project contains lock files generated by tools other than Yarn. It is advised not to mix package managers in order to avoid resolution inconsistencies caused by unsynchronized lock files. To clear this warning, remove package-lock.json.
[1/4] 🔍  Resolving packages...
success Already up-to-date.
✨  Done in 0.06s.
```

It would confuse people more if Parcel uses yarn when there is package-lock.json.

DeMoorJasper:
@Hoishin as it says you should never have both so if it detects yarn.lock and package-lock somebody didn't know what they were doing and parcel shouldn't even consider this use-case to be any different. (this is a personal opinion)

About yarn supporting `package-lock.json` it's not the kinda support you might have hoped for and might be totally irrelevant for this (https://yarnpkg.com/blog/2018/06/04/yarn-import-package-lock/)

Hoishin:
I thought you meant something like yarn will import package-lock on install and try to sync. Sounds fair.

Does this look good now?

src/utils/installPackage.js
```diff
@@ -72,19 +74,22 @@ async function installPeerDependencies(filepath, name, options) {
...
-  // If Yarn isn't available, or there is a package-lock.json file, use npm.
-  let configName = configFile && path.basename(configFile);
-  if (!hasYarn || configName === 'package-lock.json') {
+  const yarnLockFile = await config.resolve(filepath, [YARN_LOCK]);
+
+  /**
+   * no yarn.lock => use npm
+   * yarn.lock => Use yarn, fallback to npm
+   */
+  if (!yarnLockFile) {
```
DeMoorJasper: It would've been shorter as 

```Javascript
if (yarnLockFile && (await checkForYarnCommand())) {
	return 'yarn';
}
return 'npm';
```

But I merged it anyway, it's just a minor thing anyway

DeMoorJasper:
@danielo515 the next one

DeMoorJasper:
@danielo515 depends on what is merged by the time it gets released we currently don’t really plan out releases. Sometimes there are milestones for patch releases, major is 2.0 and will not be for anytime soon

danielo515:
What are my options if I'm experiencing this problem right now?

Hoishin:
You can temporaliry resolve the problem if you can delete `yarn` in CI environment.

DeMoorJasper:
@danielo515 use `--no-autoinstall` or use the git version of parcel https://docs.npmjs.com/files/package.json#git-urls-as-dependencies

danielo515:
Thanks both
Another option could be to publish the master branch of parcel to my own npm repository (we have a private one)

Regards

devongovett:
Sorry for not reviewing earlier, but I don’t think we should do this. The default is yarn if it it is available. If you have a package-lock.json file, we already use npm. I think this is a reasonable default and I do not see a reason to change it.

Hoishin:
@devongovett The problem is the case of having no lock file at all. Yarn enforces you to make yarn.lock even if you gitignore, first time you use yarn the file is made.
On the other hand, it is entirely possible to have no lock file using npm, just setting lock file to off in .npmrc.

danielo515:
You are all talking about defaults like you know what people always need. Using yarn by default just denotes your personal preferences, not the actual reality, which is that most people just use npm.
I think you should put a bit more effort on providing the ability to choose rather than debating the best defaults.

This PR solves problems to me and my company. Can you say the opposite of reverting it ? Does that help anyone ?

devongovett:
If you don’t want to use yarn, you can just uninstall it and parcel will use npm. It seems to me like if you have it installed you’d probably like to use it.

Hoishin:
It's reported that it happens in CI, which I think is because yarn is installed by default. People also use both yarn and npm in the same machine. Some projects use yarn, some projects use npm.
(also don't get me wrong, I use yarn pretty much all the time unless npm is already used in a project)

devongovett:
The problem is new projects, not existing ones typically. In a new project, there will be no lockfile at all, so we choose yarn if it is available which will create the initial lockfile for you. In existing projects, there is typically already a lockfile available so we can choose npm or yarn appropriately.

I'm not really sure why it would matter whether npm or yarn is used if there is no lockfile at all. They both should be achieving the same outcome, and if not then that's a bug in npm or yarn.

In CI the autoinstaller should not be running at all anyway - it is automatically disabled when running `parcel build` for production. It is only intended for local development only. If you have dependencies in CI that aren't listed in package.json already, then you should ensure they are added and that `npm` or `yarn` runs prior to `parcel build`.

Hoishin:
Then I guess it really comes down to subjective matters and whether it wants to prioritize npm or yarn. If so I have no words since it really depends on maintainer's perspective.

danielo515:
I really appreciate the work you are doing with parcel, but I really think that you should provide the user with options to decide which behavior he/she prefers rather than putting so much effort trying to guess the best defaults, that you can not tweak at all.

Regards

PR #2125, opened 2018-10-12
Revert "use npm for auto-install by default"

devongovett:
Reverts parcel-bundler/parcel#2119


DeMoorJasper:
Related issue #2117

Issue #2081, opened 2018-09-29
Disabling autoinstall globally


mischnic:
# 🙋 feature request

Add the possibility to set `--no-autoinstall` globally using a env variable (e.g. `PARCEL_NO_AUTOINSTALL`).

## 💁 Possible Solution

Make it so. I can submit a PR if you would accept this feature.

## 🔦 Context

The autoinstall feature might be great for beginners, but today, it worked against me two times in a row:
- I switched over from React to Preact. First, I uninstalled react and installed the preact package to then start parcel and see where I need to change the imports. But instead, parcel reinstalled react.
- I imported "style.css" instead of "./style.css", and parcel installed the "style.css" package.
I don't want to add the `--no-autoinstall` flag to every project of mine,

cacheflow:
PR for this if anyone has a chance: https://github.com/parcel-bundler/parcel/pull/2130

chocolateboy:
My 2c: installing anything without the user's express knowledge and consent is a huge security risk which will end up compromising someone's system and bringing this project into disrepute.

While this workaround is well-meaning, and arguably better than the status quo, it's still a sticking plaster, which legitimizes the current (mis)behavior rather than fixing the underlying problem: *it should be opt in rather than opt out*.

DeMoorJasper:
@chocolateboy i agree about installing code dependencies, but not about the parsers and such parcel internally uses to process assets as it removes some of the complexity without the need of including them all as parcel dependencies

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

mischnic:
Possible example for the security risk: With `import React from 'reac';`, parcel installs that dependency (and runs it's postinstall scripts). Without autoinstall, I would have realised my typo while typing `yarn add reac`.

chocolateboy:
Related: https://github.com/zkat/npx/issues/198

PR #2130, opened 2018-10-14
Add disabling of autoinstall globally via environment variable

cacheflow:
# ↪️ Pull Request
-Allow disabling of autoinstall via a environment variable named PARCEL_NO_AUTOINSTALL.

## ✔️ PR Todo

- [x ] Added/updated unit tests for this change

Feature Request: https://github.com/parcel-bundler/parcel/issues/2081


mischnic:
Maybe this is a better place for disabling autoinstall:

https://github.com/parcel-bundler/parcel/blob/e86ddf45b9a46d9035dbbcbfe97a9ca1cc30ef03/src/Bundler.js#L137-L140

cacheflow:
Agreed @mischnic.

devongovett:
Hmm why do we need an environment variable and not just use the cli flag that already exists?

mischnic:
I don't want to add the --no-autoinstall flag to every project of mine and having this flag set implicitly (so not specified in the package.json) doesn't change the build output.

cacheflow:
I'd like the same in my projects @devongovett.

src/Bundler.js
```diff
@@ -135,7 +135,8 @@ class Bundler extends EventEmitter {
       detailedReport: options.detailedReport || false,
       global: options.global,
       autoinstall:
-        typeof options.autoinstall === 'boolean'
+        typeof options.autoinstall === 'boolean' &&
+        process.env.PARCEL_NO_AUTOINSTALL !== 'true'
```
devongovett: I think it would be better if the options were positive instead of negative, so `PARCEL_AUTOINSTALL=false` should disable it

cacheflow: That was my original thought. Brb updating.

devongovett:

PR #2149, opened 2018-10-16
fix(sourcemaps): Handle null mappings properly

### LINKED ISSUE #1825: Sourcemaps: null mappings not working as expected

DeMoorJasper:
# ↪️ Pull Request

This allows null mappings in parcels sourcemaps.
Creating more accurate sourcemaps.
Also adds a couple extra tests for sourcemaps, using the SourceMap class directly, giving a more in depth coverage of SourceMap

Closes #1825

Original PR: #1851

## 🚨 Test instructions

Run unit tests

## ✔️ PR Todo


packages/core/integration-tests/test/sourcemaps.js
```diff
@@ -2,6 +2,7 @@ const assert = require('assert');
 const fs = require('@parcel/fs');
 const path = require('path');
 const mapValidator = require('sourcemap-validator');
+const SourceMap = require('../src/SourceMap');
```
devongovett: Looks like the tests are failing because of this import. Maybe this test should be inside the `parcel-bundler` package rather than the integration tests?

DeMoorJasper: It probably should although maybe it would be a good idea to extract the sourcemap implementation in the future to allow plugins to use it?

devongovett: yeah eventually

devongovett:
seems like some html minification tests are failing? doesn't seem related though

PR #2152, opened 2018-10-16
feat(autoinstall): Add disabling of autoinstall globally via environm…

cacheflow:
# ↪️ Pull Request
-Allow disabling of autoinstall via a environment variable named PARCEL_AUTOINSTALL.

## ✔️ PR Todo

- [x ] Added/updated unit tests for this change

Feature Request: https://github.com/parcel-bundler/parcel/issues/2081


packages/core/parcel-bundler/src/Bundler.js
```diff
@@ -135,7 +135,8 @@ class Bundler extends EventEmitter {
       detailedReport: options.detailedReport || false,
       global: options.global,
       autoinstall:
-        typeof options.autoinstall === 'boolean'
+        typeof options.autoinstall === 'boolean' &&
+        process.env.PARCEL_AUTOINSTALL !== 'false'
```
devongovett: I think the local `autoinstall` option should override the environment variable. The environment variable should only apply if the option is not set.

packages/core/parcel-bundler/src/Bundler.js
```diff
@@ -103,6 +103,12 @@ class Bundler extends EventEmitter {
           : watch;
     const scopeHoist =
       options.scopeHoist !== undefined ? options.scopeHoist : false;
+    const autoInstall =
+      typeof options.autoinstall === 'boolean'
+        ? options.autoinstall
+        : typeof process.env.PARCEL_AUTOINSTALL === 'boolean'
```
devongovett: does that actually get converted to a boolean if you provide this environment variable from a terminal?

PR #2175, opened 2018-10-19
Graph building improvements

padmaia:
# ↪️ Pull Request

Here are some improvements to the graph building implementation based on comments @devongovett made on the last PR. I've started tracking "invalid" state in the AssetGraph class to signify that the graph may no longer be accurate based on the filesystem changing. This change allowed the `update` phase and the `complete` phase to behave similarly which cleaned up the code nicely.

(Sorry some Prettier changes made it in making the PR more cloudy)


packages/core/core-v2/test/PromiseQueue.test.js
```diff
@@ -0,0 +1,56 @@
+// @flow
+import assert from 'assert';
+import delay from 'delay';
+import randomInt from 'random-int';
+
+import PromiseQueue from '../src/PromiseQueue';
+
+describe.only('PromiseQueue', () => {
```
devongovett: remove .only

packages/core/core-v2/src/Parcel.js
```diff
@@ -112,72 +108,74 @@ export default class Parcel {
...
+      default:
+        throw new Error('Invalid Graph');
     }
   }
 
-  async resolve(dep: Dependency, { signal }: BuildOpts) {
-    // console.log('resolving dependency', dep);
-    let resolvedPath = await this.resolverRunner.resolve(dep);
+  resolve(dep: Dependency, {signal}: BuildOpts) {
+    return this.queue.add(async () => {
```
devongovett: I was a bit confused when reading this since the method changed from actually resolving to queuing a resolve. Can we either rename the method to something like `queueResolve` or keep `resolve` and do the queuing separately? Kinda like the latter, see next comment.

packages/core/core-v2/src/Parcel.js
```diff
@@ -112,72 +108,74 @@ export default class Parcel {
...
   }
 
-  async completeGraph({ signal }: BuildOpts) {
-    // AssetGraph keeps track of nodes that represent files that have not been transformed
-    // and dependencies that have not been resolved so that rebuilds start from the leaves
-    // of the graph
+  async completeGraph({signal}: BuildOpts) {
     for (let [id, node] of this.graph.incompleteNodes) {
-      this.mainQueue.add(() => this.processNode(node, { signal }));
+      this.processNode(node, {signal});
```
devongovett: I actually kinda liked having the actual queuing here instead of in the `resolve`/`transform` methods. It was easier to understand. At first I was going to comment that you missed an `await` here before reading the other methods and realizing that they changed to queue instead of actually resolve or transform.

packages/core/core-v2/src/Parcel.js
```diff
@@ -112,72 +108,74 @@ export default class Parcel {
     }
   }
 
-  async updateGraph() {
-    // Process all changes that may update the middle of the the graph
-    this.updateQueue.start();
-    await this.updateQueue.onIdle();
-    this.updateQueue.pause();
+  async updateGraph({signal}: BuildOpts) {
+    for (let [id, node] of this.graph.invalidNodes) {
```
devongovett: Where do nodes ever get removed from `invalidNodes`? I couldn't find it.

packages/core/core-v2/src/Parcel.js
```diff
@@ -112,72 +108,74 @@ export default class Parcel {
...
   }
 
-  async transform(file: File, { signal, shallow }: BuildOpts) {
-    // console.log('transforming file', file);
-    let { children: childAssets } = await this.transformerRunner.transform(file);
-    if (signal && !signal.aborted) {
-      let { prunedFiles, newDeps } = this.graph.updateFile(file, childAssets);
+  transform(file: File, {signal, shallow}: BuildOpts) {
+    return this.queue.add(async () => {
+      // console.log('transforming file', file);
```
devongovett: remove console

packages/core/core-v2/src/Parcel.js
```diff
@@ -112,72 +108,74 @@ export default class Parcel {
...
+        throw new Error('Invalid Graph');
     }
   }
 
-  async resolve(dep: Dependency, { signal }: BuildOpts) {
-    // console.log('resolving dependency', dep);
-    let resolvedPath = await this.resolverRunner.resolve(dep);
+  resolve(dep: Dependency, {signal}: BuildOpts) {
+    return this.queue.add(async () => {
+      // console.log('resolving dependency', dep);
```
devongovett: remove console

packages/core/core-v2/src/PromiseQueue.js
```diff
@@ -0,0 +1,78 @@
+// @flow
+
+type PromiseQueueOpts = {
+  maxConcurrent?: number
+};
+
+export default class PromiseQueue {
```
devongovett: Why do we need our own PromiseQueue, not just PQueue as you had before? (There is also one in Parcel 1, just curious the reason this time. 😉)

packages/core/core-v2/src/Parcel.js
padmaia: Woops, good catch

packages/core/core-v2/src/PromiseQueue.js
padmaia: Main motivation was better error handling. The primary way I am using the queue is to fill it up with some initial functions (that may add more functions as they run) and then run until the queue is empty. I wanted `queue.run()` to reject if any of the functions return a rejected promise. PQueue doesn't have `.run()`, but it does have `.pause()` and `.start()`, however `.start()` always resolves, so I decided it was time to implement my own fitting my needs :) That brings up a question I had. Should `.run()` reject in a `Promise.all()` way or a `Promise.settle()` way? Right now it behaves like `Promise.all()`.

packages/core/core-v2/src/Parcel.js
padmaia: Can do, I was mostly keeping these logs around because I use them frequently for debugging. What are your thoughts on having some debugging logs?

packages/core/core-v2/src/Parcel.js
padmaia: I think I interpreted [this comment](https://github.com/parcel-bundler/parcel/pull/2000#discussion_r223875907) you made on the last PR as a suggestion to only add to the queue in one place, but yeah, the name of the function should probably change to imply we are queuing if that's the case. I kind of like just having `processNode`, `resolve`, and `tranform` just add to the queue directly too. I can change it back.

Issue #1481, opened 2018-06-02
JavaScript file hash doesn't change when dependency filename changed


dobesv:
# 🐛 bug report

I noticed that a new JavaScript file was generated with the same name as the old one, even though its contents had changed.

## 🎛 Configuration (.babelrc, package.json, cli command)

I created a sample project in this repository:

https://github.com/dobesv/parcel-import-bug

## 🤔 Expected Behavior

If the contents of a generated JavaScript file change in any significant way, so should the hash in its filename.

## 😯 Current Behavior

As you can see in the below transcript, the contents of the generated JavaScript file are modified, but its hash doesn't change.

```
$ echo a>template5.html
$ parcel build index.html -d dist-a
✨  Built in 1.24s.

dist-a/parcel-import-bug.a9b3581e.js      299.6 KB    19.34s
dist-a/controller.1cfadc83.js             87.37 KB     220ms
dist-a/controller.343f9052.js              1.23 KB     173ms
dist-a/index.html                            245 B       5ms
dist-a/parcel-import-bug.bfb7fe78.urls        31 B     102ms
dist-a/test.ae9c52ee.css                      20 B      12ms
dist-a/template5.7deadfab.html                 2 B     193ms
dist-a/parcel-import-bug.bfb7fe78.json         0 B     103ms
dist-a/controller.343f9052.map                 0 B     209ms
dist-a/parcel-import-bug.e3817f33.map          0 B    19.26s
dist-a/controller.1cfadc83.map                 0 B     255ms
$ echo b>template5.html
$ parcel build index.html -d dist-b
✨  Built in 1.30s.

dist-b/parcel-import-bug.a9b3581e.js      299.6 KB    20.44s
dist-b/controller.1cfadc83.js             87.37 KB     227ms
dist-b/controller.343f9052.js              1.23 KB     182ms
dist-b/index.html                            245 B       3ms
dist-b/parcel-import-bug.bfb7fe78.urls        31 B     102ms
dist-b/test.ae9c52ee.css                      20 B      15ms
dist-b/template5.fc4c8ece.html                 2 B     190ms
dist-b/parcel-import-bug.bfb7fe78.json         0 B     101ms
dist-b/controller.343f9052.map                 0 B     217ms
dist-b/parcel-import-bug.e3817f33.map          0 B    20.34s
dist-b/controller.1cfadc83.map                 0 B     260ms
(env) 13:40:40 ~/parcel-import-bug 
[...]

PR #2179, opened 2018-10-20
Include indirect dependencies in bundle filename hash

dobesv:
# ↪️ Pull Request

Changes bundle hash calculation to include all dependencies.

## 💻 Examples

Previously a change to a file that a bundle depended on, but did not include, would not affect the hash in its filename.  This could result in a situation where the filename would not change even though references to the bundle's dependencies would be updated in the file.

See https://github.com/parcel-bundler/parcel/issues/1481

This PR includes a test case that fails if the changes are reverted, and also serves as an example of the issue.

### DIFF
--- a/packages/core/parcel-bundler/src/Bundle.js
+++ b/packages/core/parcel-bundler/src/Bundle.js
@@ -112,8 +112,8 @@ class Bundle {
     // Otherwise, use a hash of the filename so it remains consistent across builds.
     let ext = Path.extname(this.name);
     let hash = (contentHash
-      ? this.getHash()
-      : Path.basename(this.name, ext)
+        ? this.getHash()
+        : Path.basename(this.name, ext)
     ).slice(-8);
     let entryAsset = this.entryAsset || this.parentBundle.entryAsset;
     let name = Path.basename(entryAsset.name, Path.extname(entryAsset.name));
@@ -274,10 +274,15 @@ class Bundle {
 
   getHash() {
     let hash = crypto.createHash('md5');
-    for (let asset of this.assets) {
-      hash.update(asset.hash);
-    }
-
+    let visited = new Set();
+    let visit = asset => {
+      if (!visited.has(asset.id)) {
+        visited.add(asset.id);
+        hash.update(asset.hash);
+        Array.from(asset.depAssets.values()).forEach(visit);
+      }
+    };
+    this.assets.forEach(visit);
     return hash.digest('hex');
   }
 }
--- a/packages/core/parcel-bundler/test/contentHashing.js
+++ b/packages/core/parcel-bundler/test/contentHashing.js
@@ -71,4 +71,35 @@ describe('content hashing', function() {
 
     assert.notEqual(filename, newFilename);
   });
+
+  it('should update content hash when indirect dependency changes', async function () {
+    await ncp(
+      path.join(__dirname, '/integration/import-multilevel-html'),
+      path.join(__dirname, '/input')
+    );
+
+    await bundle(path.join(__dirname, '/input/index.html'), {
+      production: true
+    });
+
+    let html = await fs.readFile(path.join(__dirname, '/dist/index.html'), 'utf8');
+    let filename = html.match(/\/(input\.[0-9a-f]+\.js)/)[1];
+    assert(await fs.exists(path.join(__dirname, '/dist/', filename)));
+    let content = await fs.readFile(path.join(__dirname, '/dist/', filename));
+
+    await fs.writeFile(path.join(__dirname, '/input/test.txt'), 'hello world');
+
+    await bundle(path.join(__dirname, '/input/index.html'), {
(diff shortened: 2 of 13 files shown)


packages/core/parcel-bundler/src/Bundle.js
```diff
@@ -274,10 +274,11 @@ class Bundle {
 
   getHash() {
     let hash = crypto.createHash('md5');
-    for (let asset of this.assets) {
+    let visit = asset => {
       hash.update(asset.hash);
-    }
-
+      Array.from(asset.depAssets.values()).forEach(visit);
```
devongovett: I think this will be a problem if there are circular dependencies.

dobesv:
Hi Devon, when you depend on files outside the bundle the filename of the resource may be included in the bundle.  If the resource filename changes, the bundle filename must also change or clients might not download the new bundle. I think the example kind of demonstrates this pretty clearly.

packages/core/parcel-bundler/src/Bundle.js
dobesv: Thanks, I've added a test for that and a guard to prevent infinite recursion if there's a cycle.

dobesv:
It is true that this could cause more invalidations than needed, although I prefer a spurious invalidation to my app breaking because the browser has cached resources with references to files that no longer exist.

I think the incorrect invalidation would occur if you depend on a file that will not be bundled into the bundle and will not have a hash inserted into its name in the bundle.

I think this might happen if you depend on a root asset, like if you have a dependency on index.html and index.html was also provided on the command line.  Can you think of other cases?  It might be possible adapt the algorithm to (1) ignore assets that *will* be bundled into the bundle, since AFAIK they already affect the hash and (2) ignore assets which are not bundled but will not have a hash inserted into their name.

I don't mind attempting this but I'm not confident I can get it right, I'm not super intimate with Parcel so I might miss something.  I'd rather be conservative and have these extra invalidations, since a missing invalidation just adds a bit of load time, but a missing invalidation can break the app.

dobesv:
The example I used for the test is a bit contrived.  In my app it's more like having html -> js -> css -> images and html -> js -> html -> css -> images type dependencies, which make a bit more sense.

dobesv:
So is this PR rejected then?  Any other ideas on how to resolve the issue?

dobesv:
I created a plugin for this as a temporary workaround for anyone affected by this issue.  https://github.com/dobesv/parcel-plugin-bundle-deep-hash

ChALkeR:
@devongovett This is what I observe:
```
$ diff -qr dist.0 dist.1
Files dist.0/components.8aef4adc.js and dist.1/components.8aef4adc.js differ
Only in dist.0: filea.33d24a38.js
Only in dist.0: filea.33d24a38.map
Only in dist.1: filea.7fa09e67.js
Only in dist.1: filea.7fa09e67.map
Files dist.0/service-worker.js and dist.1/service-worker.js differ
```

That was generated by `parcel build --no-autoinstall frontend/index.html`.

Here, `components.js` requires `filea.js`, `filea.js` was updated and got its hash changed, `components.js` source was untouched, _but it's built code got updated because it includes an updated `filea.***.js` path.

Currently, Parcel **does not** update the hash of the file if the file contents of the _resulting_ file has been changed, it only updates the hash if the _source_ file has been changed.

This makes it unusable with [sw-precache](https://github.com/GoogleChromeLabs/sw-precache) (and probably any other persistent caching methods) -- an inconsistent state of the application with mismatching paths ends up in cache.

`filea.***.js` got replaced by the new version and the old one returns 404, but `components.8aef4adc.js` did not get updated because it has the same filename and exists in user cache already. But old `components.8aef4adc.js` tries to load old `filea.***.js` under the old name, which doesn't exist anymore, resulting in 404, so the whole application just breaks.

ChALkeR:
https://github.com/ChALkeR/parcel-cache-fail -- this is a demonstration with more explanation and details in readme.

ChALkeR:
@devongovett A trivial example with some description of what exactly fails in a real-world scenario: https://github.com/ChALkeR/parcel-cache-fail

This PR seems to resolve that issue for me.

dobesv:
Do the project owners want a different approach to this issue?  Any change requests on this PR?

DeMoorJasper:
Gonna close this as parcel 1 is in maintenance mode and this is pretty out of date anyway

wxsms:
@DeMoorJasper This issue (#1481 ) is still not resolved in the latest version of parcel.

dobesv:
@DeMoorJasper if I updated this to target the latest version of parcel, would it be merged?  Or would it be ignored again?

kpratik2015:
> I created a plugin for this as a temporary workaround for anyone affected by this issue. https://github.com/dobesv/parcel-plugin-bundle-deep-hash

Thank you, it helped 👍

PR #2197, opened 2018-10-25
Merge fs-watcher-child into Parcel's monorepo

DeMoorJasper:
# ↪️ Pull Request

This adds https://github.com/DeMoorJasper/fswatcher-child to the parcel monorepo, once this gets published `fswatcher-child` will become deprecated.

## ✔️ PR Todo


packages/core/parcel-bundler/test/integration/babel-plugin-autoinstall/package.json
```diff
@@ -1,3 +1,6 @@
 {
-  "name": "babel-plugin-autoinstall"
+  "name": "babel-plugin-autoinstall",
+  "devDependencies": {
+    "@babel/core": "^7.1.2"
+  }
```
devongovett: this change shouldn't be committed, it's part of the test to verify that it was added.

DeMoorJasper: Ow damn, I'll remove it and add it to a gitignore to prevent this in the future

packages/core/watcher/src/Watcher.js
```diff
@@ -1,27 +1,161 @@
...
+      for (let func of this.readyQueue) {
+        func();
+      }
+      this.readyQueue = [];
+    });
+
+    this.startchild();
+  }
+
+  startchild() {
```
devongovett: wonder if it would be worth it to run this inside the existing parcel worker farm at some point. may be worth experimenting with?

packages/core/utils/index.js
```diff
@@ -0,0 +1,10 @@
+// Node 8 supports native async functions - no need to use compiled code!
+exports.promisify =
+  parseInt(process.versions.node, 10) < 8
+    ? require('./lib/promisify')
+    : require('./src/promisify');
+
+exports.errorUtils =
```
devongovett: heh, we should try to find a way to make this easier if there are going to be a lot of utils 😉

packages/core/watcher/src/Watcher.js
DeMoorJasper: Yeah probably a good idea

PR #2200, opened 2018-10-25
Option to leave certain filenames untouched

LiamMartens:
# ↪️ Pull Request
adds functionality to leave filenames untouched for certain paths
this is useful for legacy projects where certain parts depend on predictable filenames instead of content hashes. This is an issue we ran into at the agency I am currently employed at.

Implementation wise, this adds an additional configuration option to the bundler called `copyPaths` which is an array of relative paths or regexes. If an asset matches the one of the `copyPaths` the filename will not receive a hash even if it isn't an entry file.

## 💻 Examples
In the example below all `.jpg` files will not receive a hash in the final filename.
```
const bundler = new Bundler(entryFiles, {
    copyPaths: [
        /\.jpg$/
    ]
});
```

## ✔️ PR Todo

- [ x] Added/updated unit tests for this change
- [ x] Filled out test instructions (In case there aren't any unit tests)

### DIFF
--- a/packages/core/parcel-bundler/src/Bundle.js
+++ b/packages/core/parcel-bundler/src/Bundle.js
@@ -93,29 +93,36 @@ class Bundle {
     return this.assets.size === 0;
   }
 
-  getBundleNameMap(contentHash, hashes = new Map()) {
+  getBundleNameMap(contentHash, copyPaths = [], hashes = new Map()) {
     if (this.name) {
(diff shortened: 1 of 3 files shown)


devongovett:
I don't want to add additional config options for this. If you really need to copy files, you could do so in a plugin, or as a separate build step. If an import is causing a particular issue, I'd rather solve that actual issue rather than working around it by copying the original file.

PR #2210, opened 2018-10-28
Add Kotlin asset support

### LINKED ISSUE #2177: Kotlin support
# 🙋 feature request

Support for building and including Kotlin (.kt) files. As Kotlin is getting increasingly popular, it makes sense for support to be built-in.

## 🤔 Expected Behavior

When a .kt file is found, it should be included in the output as a JavaScript file or a WASM file (would be configurable)

## 😯 Current Behavior

The .kt file is just copied into the output directory.

## 💁 Possible Solution

Node packages exist that should be able to build Kotlin files, however after testing `kotlin-node` could not be installed and `node-kotlin` would generate an empty zip inside a zip.

Instead kotlinc could be downloaded from JetBrains/kotlin and run with the correct arguments to generate .js files, which could then be inputted back into the system to create the output.

## 🔦 Context

I'm trying to use Kotlin instead of JavaScript or TypeScript because I prefer the language and how it works. As Kotlin can build to both JavaScript and WASM, it can be used as the single programming language in the project, as well as a server system.

## 💻 Examples

N/A

Zoweb:
# ↪️ Pull Request

This should close #2177 by adding support for Kotlin. Unfortunately, the only possible way to implement Kotlin support was to use the compiler, which can only generate files, so a temporary directory is created and deleted with those files, if there is nothing else in it.

Running Kotlin functions from other languages (e.g. importing a Kotlin file in a JavaScript file, then running a function defined in it) doesn't work as intended - the Kotlin compiler adds some extra text to the function names, which has to be added in the other language as well.

To get Kotlin code to run automatically, instead of just putting the code directly in the file, create a `fun main(args: Array<String>)`. This function will be called with an empty list of arguments.

Note that the Kotlin standard library is included in the result, and is huge. Building in production mode minifies the source but it is still over 600KB with the test described below. Implementing treeshaking may help, if this is not already done.

## 🚨 Test instructions

Create an `index.html` file and a `test.kt` file. Add a `<script>` tag with the src pointing to the `kt` file.

Inside the Kotlin file, add the following code:
```kotlin
fun main(args: Array<String>) {
    for (i in 0 until 100) {
        if (i % 20 == 0) print("FooBar")
        else if (i % 5 == 0) print("Bar")
        else if (i % 4 == 0) print("Foo")
        else print(".")

        println(" [$i]")
    }
}
```

This should print 100 lines, where every 4th is "Bingle [(line num)]", 5th is "Foo [(line num)]", 20th is "FooBar [(line num)]" and every other line is ". [(line num)]"

## ✔️ PR Todo


Zoweb:
Hmm, it looks like elm failed but just on Windows/Node 6

packages/core/parcel-bundler/src/assets/KotlinAsset.js
```diff
@@ -0,0 +1,65 @@
...
+    const kotlinCompiler = await localRequire(
+      '@jetbrains/kotlinc-js-api',
+      this.name
+    );
+    const fs = require('fs', this.name);
+
+    // remove extension and path
+    let slashyIndex = this.id.lastIndexOf('/'); // linux/mac
+    if (slashyIndex === -1) slashyIndex = this.id.lastIndexOf('\\'); // windows
+    const fileName = this.id.substring(slashyIndex, this.id.lastIndexOf('.'));
```
devongovett: Can you use the `path` library for this? It will take care of the normalization cross platform for you.

packages/core/parcel-bundler/src/assets/KotlinAsset.js
```diff
@@ -0,0 +1,65 @@
...
+
+    await kotlinCompiler.compile({
+      output: 'build/kt.temp/' + fileName + '.js',
```
devongovett: You should be explicit about exactly where this is going to go. Right now it will go into the current working directory. Instead, it would be better to go into a temporary directory.

devongovett:
This is awesome! It would be great if you could add some tests.

Zoweb:
Oh thanks for doing those changes, I only was able to see this today.

PR #2218, opened 2018-10-29
Pull in v2 work

padmaia:
# ↪️ Pull Request

Pulls in all of the v2 work from the `next` branch. All packages new v2 packages are written with ESM import/export syntax to work well with Flow. For now I have chosen to leave all packages used in v1 alone. I may have missed some things that could be cleaned up, but I wanted to open this before the hackathon starting tomorrow.


devongovett:
There are a bunch of util files that don't appear to be used yet, and there are two tests failing on windows for some reason. Not sure yet.

PR #2220, opened 2018-10-30
Add PackagerRunner

### LINKED ISSUE #2141: Parcel 2: PackagerRunner
The PackagerRunner object in `@parcel/core` is responsible for running Packager plugins in Parcel 2. Packager plugins write bundles to disk, combining their assets together into a single file.

The PackagerRunner should resolve Packager plugins from the Parcel config and execute them. A Packager plugin receives a `Bundle` as input, which is just a list of Assets, and returns a string or buffer containing the combined output. It will also need access to the Cache to read the contents of transformed assets, as the list of Assets provided as input only contains metadata. Packager plugins can also override the default file writing behavior, e.g. the Raw packager may just want to copy files using `fs.copy` instead of reading into memory and writing again.

fathyb:
Closes #2141

### DIFF
--- a/packages/core/utils/src/matchConfig.js
+++ b/packages/core/utils/src/matchConfig.js
@@ -1,0 +1,12 @@
+const {isMatch} = require('micromatch');
+const {basename} = require('path');
+
+module.exports = (config, file) => {
+  for (let pattern in config) {
+    if (isMatch(file, pattern) || isMatch(basename(file), pattern)) {
+      return config[pattern];
+    }
+  }
+
+  return null;
+};
--- a/packages/core/utils/index.js
+++ b/packages/core/utils/index.js
@@ -8,3 +8,8 @@ exports.errorUtils =
   parseInt(process.versions.node, 10) < 8
     ? require('./lib/errorUtils')
     : require('./src/errorUtils');
+
+exports.matchConfig =
+  parseInt(process.versions.node, 10) < 8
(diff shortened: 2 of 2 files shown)


packages/core/core/src/PackagerRunner.js
```diff
@@ -10,20 +11,23 @@ class PackagerRunner {
...
+    packager = require(packager);
 
     let modulesContents = await Promise.all(
       bundle.assets.map(async asset => {
-        // let fileContents = await packager.readFile({
-        //   filePath: asset.code,
-        // });
         let blobs = await this.cache.readBlobs(asset);
-
         let result = await packager.asset({blobs});
```
devongovett: I think I'd like to change the API to only have one method on a packager: `package`, which receives a list of assets with their output pre-read from the cache. Packagers also shouldn't have any filesystem related methods since the optimizers still need to run after the packagers.

packages/core/utils/src/matchConfig.js
```diff
@@ -0,0 +1,12 @@
+const {isMatch} = require('micromatch');
+const {basename} = require('path');
+
+module.exports = (config, file) => {
```
devongovett: I think thin should eventually go in the ConfigResolver: cc. @jamiebuilds

devongovett:
Closing. This was merged manually into the Config class.

PR #2228, opened 2018-10-31
Add ResolverRunner and default Resolver plugin

fathyb:


packages/core/builtins/src/bundle-loader.js
```diff
@@ -0,0 +1,82 @@
+var getBundleURL = require('./bundle-url').getBundleURL;
```
devongovett: I don't think we should include all of these builtins just yet. We may want to rewrite or move some of them around. I'd just include the node builtins in the resolver for now.

fathyb: Done ✨

PR #2229, opened 2018-10-31
cleanup utils

padmaia:
# ↪️ Pull Request

Only pulled out utils used so far in v2, and kept the rest in parcel-bundler. I did find some circular dependencies:

fs -> utils -> fs
logger -> workers -> utils -> logger

For the fs circular dependency I duplicated promisify in the fs package.
For the logger circular dependency, I commented out the logging in the installPackage and pipeSpawn utils.

Duplicating promisify is not ideal, but not the worst. I'll have to come up with an alternative to commenting out logging. Does it make sense to pull installPackage and pipeSpawn into their own packages?


padmaia:
Just noticed node 6 tests are failing for some reason, I'll look into that. All other failures are flow errors already present in `v2-work-so-far`.

devongovett:
Closing. Seems pretty outdated now.

PR #2239, opened 2018-11-02
Parcel 2: Simple working JS packager

devongovett:
Wraps all assets in functions.

### DIFF
--- a/packages/core/core/src/AssetGraph.js
+++ b/packages/core/core/src/AssetGraph.js
@@ -131,9 +131,10 @@ export default class AssetGraph extends Graph {
 
     for (let assetNode of assetNodes) {
       // TODO: dep should already have sourcePath
-      let depNodes = assetNode.value.dependencies.map(dep =>
-        nodeFromDep({...dep, sourcePath: file.filePath})
-      );
+      let depNodes = assetNode.value.dependencies.map(dep => {
+        dep.sourcePath = file.filePath;
+        return nodeFromDep(dep);
+      });
       let {removed, added} = this.updateDownStreamNodes(assetNode, depNodes);
       prunedFiles = prunedFiles.concat(getFilesFromGraph(removed));
       newDepNodes = newDepNodes.concat(getDepNodesFromGraph(added));
--- a/packages/core/core/src/Parcel.js
+++ b/packages/core/core/src/Parcel.js
@@ -174,6 +174,7 @@ export default class Parcel {
 
     let file = {filePath: resolvedPath};
     if (signal && !signal.aborted) {
+      dep.resolvedPath = resolvedPath;
       let {newFile} = this.graph.updateDependency(dep, file);
 
       if (newFile) {
--- a/packages/packagers/js/src/index.js
+++ b/packages/packagers/js/src/index.js
@@ -2,9 +2,45 @@
 'use strict';
 
 import {Packager} from '@parcel/plugin';
+import fs from 'fs';
+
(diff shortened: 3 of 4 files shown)


packages/core/core/src/AssetGraph.js
```diff
@@ -131,9 +131,10 @@ export default class AssetGraph extends Graph {
 
     for (let assetNode of assetNodes) {
       // TODO: dep should already have sourcePath
-      let depNodes = assetNode.value.dependencies.map(dep =>
-        nodeFromDep({...dep, sourcePath: file.filePath})
-      );
+      let depNodes = assetNode.value.dependencies.map(dep => {
+        dep.sourcePath = file.filePath;
```
devongovett: To discuss: do we even need this field on a dependency? An Asset node points to a Dependency in the graph, so we should be able to infer the file the dep came from based on its location in the graph.

packages/core/core/src/Parcel.js
```diff
@@ -174,6 +174,7 @@ export default class Parcel {
 
     let file = {filePath: resolvedPath};
     if (signal && !signal.aborted) {
+      dep.resolvedPath = resolvedPath;
```
devongovett: Not sure about mutating the dep here. Also we might not want to store this in the dep anyway - we could just figure it out based on where it is in the graph. That wouldn't work if we want to cache the resolution though, so perhaps we do want it in the end. Either way, it cannot be on the dep when it is initially created - it will need to be filled in later post-resolution.

packages/packagers/js/src/index.js
```diff
@@ -2,9 +2,45 @@
...
 
 export default new Packager({
   async package(bundle) {
-    return bundle.assets.map(asset => asset.output.code).join('\n\n');
+    let assets = bundle.assets
+      .map((asset, i) => {
+        let deps = {};
+
+        for (let dep of asset.dependencies) {
+          let resolvedAsset = bundle.assets.find(
```
devongovett: This isn't a great way to find the resolved ids for deps. It will be slower than necessary for sure. Perhaps another reason to have a bundle contain a Graph instead of a flat list of assets.

padmaia: For sure, I can start looking into that. We'll have to be able serialize/deserialize graphs since the packagers will be running in workers. Do you think that could possibly end up making it not worth it performance wise?

packages/core/core/src/Parcel.js
padmaia: If we are going to add this field to the dep, it might make more sense to do it in the AssetGraph's updateDependency function.

> That wouldn't work if we want to cache the resolution though

Are talking about the parcel cache or like the in memory "cache" on the v1 Resolver? Caching in memory is pretty much already taken care of by the way the graph is built. If a file is transformed a second time and finds some of the same dependencies it won't try to resolve them again. As for the parcel cache, I'm wondering if we want to cache resolved paths, at least to begin with. Seems like cache invalidation for that could get hairy.

packages/core/core/src/AssetGraph.js
devongovett: Yeah I saw that. But we could pass the source path as a second argument to `nodeFromDep` to create the id... Perhaps it is useful information to store on a dep. I'm not sure.

packages/core/core/src/Parcel.js
devongovett: I'm talking about the filesystem cache here. In Parcel 1 we do cache the resolved paths, and it increases performance very significantly.

PR #2243, opened 2018-11-02
Graph building improvements

padmaia:
Graph building improvements base on last PR feedback

### DIFF
--- a/packages/core/core/package.json
+++ b/packages/core/core/package.json
@@ -15,8 +15,8 @@
     "@parcel/cache": "^2.0.0",
     "@parcel/config-default": "^2.0.0",
     "@parcel/plugin": "^2.0.0",
-    "@parcel/utils": "^1.10.3",
     "@parcel/types": "^2.0.0",
+    "@parcel/utils": "^1.10.3",
     "abortcontroller-polyfill": "^1.1.9",
     "chalk": "^2.1.0",
     "commander": "^2.11.0",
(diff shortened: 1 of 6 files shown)


packages/core/core/src/Parcel.js
```diff
@@ -89,18 +87,10 @@ export default class Parcel {
...
-        } else {
-          // Otherwise cancel any work that may be in progress, add the change to the update queue,
-          // and start a new build.
+      this.watcher.on('change', filePath => {
+        if (this.graph.hasNode(filePath)) {
           controller.abort();
-          this.mainQueue.pause();
-          this.mainQueue.clear();
-          this.handleChange(event);
+          this.graph.invalidateNodeById(filePath);
```
devongovett: Do we need to clear the queue here too?

packages/core/core/src/PromiseQueue.js
```diff
@@ -0,0 +1,78 @@
+// @flow
+
+type PromiseQueueOpts = {
+  maxConcurrent?: number
+};
+
+export default class PromiseQueue {
```
devongovett: Looks like we also already have a PromiseQueue from parcel 1 in the utils package... Wonder if we can combine them?

packages/core/core/package.json
```diff
@@ -30,7 +30,9 @@
   },
   "devDependencies": {
     "@parcel/babel-register": "2.0.0",
-    "@parcel/eslint-config": "1.10.3"
+    "@parcel/eslint-config": "1.10.3",
+    "delay": "^4.1.0",
+    "random-int": "^1.0.0"
```
devongovett: Really do we need two more dependencies? These look like 1-liners to me...

PR #2247, opened 2018-11-04
Parcel 2:  Add connected file support to asset graph

devongovett:
This adds a new node type to the asset graph: `connected_file`. This node is like a cross between a normal `file` node and a `dependency` node. Connected files are returned by transformers like dependencies, but are fully resolved to absolute paths in the transformer, and do not get transformed like normal files. Connected files are watched, however, and trigger a rebuild of any file nodes that are connected to them.

Connected files are useful in two cases: 

1. As a part of assets for compilers that include external files as part of their builds. This is the case for stylus, sass, etc. which have their own `@import` syntax that they process themselves. The output of the compilation will include all imported files in the built asset, so we have no need to resolve and transform them. We do want to watch the included files though, so we return a connected file to represent that.

2. At the top-level, connected directly to a file to represent dependencies like config files (e.g. `.babelrc`). When a config file changes, we need to recompile all files that it affects.

In the graph, `connected_file` nodes are always connected to a `file` node, not an `asset`. This makes it easy to determine which files to invalidate when the connected file changes. However, in the cache we store a separate list of connected files on each asset with a hash on each one to determine when they changed. This allows us to only invalidate a single asset inside a file, e.g. stylus inside vue. If a the JS part of a vue file does not change but an included stylus file does, only the stylus part will be recompiled.

I added support for `fs.readFileSync` inlining to the parcel 2 JS transform, which uses connected files to ensure the files that it reads are watched. Updating the simple example to read a text file produces the following graph:

graph

### DIFF
--- a/packages/core/core/src/Graph.js
+++ b/packages/core/core/src/Graph.js
@@ -37,6 +37,10 @@ export default class Graph {
     return this.nodes.has(id);
   }
 
+  getNode(id: string) {
+    return this.nodes.get(id);
+  }
+
   addEdge(edge: Edge) {
     this.edges.add(edge);
     return edge;
@@ -50,6 +54,12 @@ export default class Graph {
     }
 
     return false;
+  }
+
+  getConnectedNodes(node: Node): Array<Node> {
+    let edges = Array.from(this.edges).filter(edge => edge.to === node.id);
+
+    return edges.map(edge => this.nodes.get(edge.from));
   }
 
   merge(graph: Graph) {
--- a/packages/core/core/src/Graph.js
+++ b/packages/core/core/src/Graph.js
@@ -56,10 +56,14 @@ export default class Graph {
     return false;
   }
 
-  getConnectedNodes(node: Node): Array<Node> {
+  getNodesConnectedTo(node: Node): Array<Node> {
     let edges = Array.from(this.edges).filter(edge => edge.to === node.id);
-
     return edges.map(edge => this.nodes.get(edge.from));
+  }
+
+  getNodesConnectedFrom(node: Node): Array<Node> {
+    let edges = Array.from(this.edges).filter(edge => edge.from === node.id);
+    return edges.map(edge => this.nodes.get(edge.to));
   }
 
   merge(graph: Graph) {
@@ -117,7 +121,7 @@ export default class Graph {
 
   // Update a node's downstream nodes making sure to prune any orphaned branches
   // Also keeps track of all added and removed edges and nodes
-  updateDownStreamNodes(fromNode: Node, toNodes: Array<Node>): GraphUpdates {
+  replaceNodesConnectedTo(fromNode: Node, toNodes: Array<Node>): GraphUpdates {
     let removed = new Graph();
     let added = new Graph();
 
--- a/packages/core/core/src/PackagerRunner.js
+++ b/packages/core/core/src/PackagerRunner.js
@@ -4,6 +4,7 @@ import Cache from '@parcel/cache';
 import {mkdirp, writeFile} from '@parcel/fs';
 import path from 'path';
 import type {Bundle, CLIOptions, Blob, FilePath} from '@parcel/types';
+import clone from 'clone';
 
 type Opts = {
   config: Config,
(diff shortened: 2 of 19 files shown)


packages/core/core/src/Graph.js
```diff
@@ -52,6 +56,12 @@ export default class Graph {
     return false;
   }
 
+  getConnectedNodes(node: Node): Array<Node> {
```
padmaia: Should we make a distinction that this only returns upstream nodes?

packages/core/types/index.js
```diff
@@ -47,7 +47,8 @@ export type Target = {
 
 export type Environment = {
   target: Target,
-  browserContext: 'browser' | 'worker' | 'serviceworker'
+  context: 'browser' | 'webWorker' | 'serviceWorker' | 'node' | 'electron',
+  includeNodeModules?: boolean
```
padmaia: How does includeNodeModules get set? Not so relevant to this PR, but I'm curious if this will be done differently once we support externals?

packages/core/core/src/Graph.js
devongovett: Probably. Have an idea for a name?

packages/core/core/src/PackagerRunner.js
```diff
@@ -45,6 +46,9 @@ export default class PackagerRunner {
   async package(bundle: Bundle): Promise<Blob> {
     let packager = await this.config.getPackager(bundle.filePath);
 
+    // Read the contents of each asset in the bundle from the cache.
+    // This mutates the assets, so clone the bundle first.
+    bundle = clone(bundle);
```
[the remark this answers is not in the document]
devongovett: Yeah I agree. But if the cache returned a new asset, we'd still need to mutate the bundle to assign its assets array...

packages/core/types/index.js
devongovett: Would be set as part of the targets configuration in package.json I guess.

packages/core/core/src/PackagerRunner.js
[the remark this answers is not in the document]
padmaia: I suppose so if we want the PackagerRunner to hand everything to the packager plugin all loaded up. Is there a reason why we wouldn't want the plugin to access the cache at all? If we are passing the packager a bundleGraph I could see something like `let allAssetCode = await bundleGraph.assets.map(asset => await asset.getContents())` working nicely. (That's very rough, the main point is that the packager runner could load asset content one by one.)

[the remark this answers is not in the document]
devongovett: Yeah we hadn't really considered having methods on assets I guess...

packages/core/core/src/Graph.js
devongovett: Upstream and downstream imply a direction relative to the graph rather than relative to the node. Maybe incoming/outgoing would be better? I'm fine either way I guess.

packages/core/core/src/PackagerRunner.js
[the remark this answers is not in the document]
devongovett: Let's come back to that idea. We may want methods for other things, e.g. loading config relative to an asset, getting an asset's package.json, etc. If not methods, then they will need to be util functions.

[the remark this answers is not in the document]
padmaia: Sure, it's not a blocker, just wanted to get the conversation started.

packages/core/core/src/Parcel.js
```diff
@@ -195,4 +200,24 @@ export default class Parcel {
   package(bundles: any[]) {
     return Promise.all(bundles.map(bundle => this.runPackage(bundle)));
   }
+
+  handleChange(filePath: string) {
```
padmaia: Can we move this logic into the `AssetGraph`? That way the `Parcel` class stays mostly unaware of the graph beyond telling it that a file has been transformed, a file has been resolved, or a file has changed.

packages/core/core/src/Graph.js
padmaia: Hmm, `incoming`/`outgoing` vs `upstream`/`downstream` seems pretty subjective to me. What if we went super explicit with `getNodesConnectedTo(node)`/`getNodesConnectedFrom(node)`/`replaceNodesConnectedTo(node)`? I know it's typical of me to throw out a long descriptive name, but if the feedback is that we're not quite sure what the method does by the name, then that's what I'm going to suggest.

devongovett: yeah I like that

PR #2248, opened 2018-11-04
Minify elm code with elm-minify

opvasger:
# ↪️ Pull Request
The goal of this PR is to have [elm-minify](https://github.com/opvasger/elm-minify) do minification on elm binaries when building with Parcel.

This is motived by the fact that the implementation Parcel is currently using relies on a hardcoded copy of a [terser/uglifyjs configuration from the official Elm guide](https://guide.elm-lang.org/optimization/asset_size.html). I would like to have future optimizations to my library be reflected when I build with Parcel

I *really* enjoy building with Parcel :sunny:

## 💻 Examples
I'm not sure how to provide examples.

## 🚨 Test instructions
I have a [PR for testing elm-minify internally](https://github.com/opvasger/elm-minify/pull/14). The project is is versioned according to SemVer 2.0

## ✔️ PR Todo


devongovett:
I'm not sure what the benefit of taking on a separate dependency is for this vs just using terser directly, which it looks like elm-minify also uses with the same configuration. Also elm-minify relies on a bunch of webpack dependencies that I'd rather not introduce into Parcel.

opvasger:
> I'm not sure what the benefit of taking on a
> separate dependency is for this vs just using
> terser directly, which it looks like elm-minify
> also uses with the same configuration

The config I'm using is slightly different and I'm not going to argue that a few bytes makes any real difference at this point.

But I do think more optimizations ([beyond what the Elm guide suggests](https://guide.elm-lang.org/optimization/asset_size.html)) will be added sooner or later. If so, I'd be nice to have them encapsulated for everyone in one place - this is the goal of `elm-minify` and the reason I'd like `parcel` to depend on it.

> elm-minify relies on a bunch of webpack
> dependencies that I'd rather not introduce into Parcel.

`elm-minify` directly depends on `webpack-sources` and `terser`. `webpack-sources` is a really small class library that I need for the `webpack` plugin `elm-minify` exposes. My development dependencies shouldn't be a problem for you.

Thanks for looking over the PR :sunny:

devongovett:
I think I'd rather keep the minification in house for now. That way we have more control over the output and can update parcel more easily. But thanks for making the PR and good luck with your lib.

opvasger:
I understand :sunny: Thanks!

PR #2150, opened 2018-10-16
fix(typescript): use tsconfig `extends` option and watch extended config

Hoishin:
original PR: #2109 

## ↪️ Pull Request

Resolve the "extends" property in tsconfig.json, and watch the parents config files.

Continuation of https://github.com/parcel-bundler/parcel/pull/646, but different approach.

## 💻 Examples

In `project/tsconfig.json`:
```json
{
  "compilerOptions": {
    "foo": "bar"
  }
}
```

In `project/src/tsconfig.json`:
```json
{
  "extends": "../tsconfig",
  "compilerOptions": {
    "some": "value"
  }
}
```

Previously Parcel could not resolve the `"extends"` property. Now it resolves and watch the parent config file(s)!

## ✔️ PR Todo

- [x] Added/updated unit tests for this change
- [x] Filled out test instructions
- [x] Included links to related issues/PRs

PR #2269, opened 2018-11-09
Yarn workspace - resolve symlink packages, and better tsconfig parser

aliai:
# ↪️ Pull Request

Parcel generally works nicely with monorepos out of the box (I am both super excited and impressed by how fast and well written it is 😍), however there are some quirks I noticed as we were integrating our project with Parcel.

I thought I would grab the shovel and dig into it myself. As it was tough to explain, with this PR I wanted to show what I am missing from Parcel. I also wrote some tests, but I am not sure if I did the tests the right way (I tried to mimic the other tests in the repo and made sure my tests are isolated from the others). Also, I'd be happy to hear from you and your opinion on this and whether or not you want me to change something :)

Here are the issues I found:

1. Parcel doesn't pick up on `tsconfig.json` file correctly. If `tsconfig.json` file has an `extends` field, for example, parcel does not read the referred file.
   - The fix is to use the internal `typescript` resolver to both find and parse the config file correctly.

2. Parcel doesn't find configuration files as it should in a yarn workspace projects.
   - When trying to see if the project is part of `yarn`, it fails to locate `yarn.lock` file inside a linked package (as it stops looking when it hits the `node_modules` folder, packages are treated as third party libraries).
   - This is the case with `.babelrc`, `tsconfig.json` or any other configurations that is located via `config.resolve` method.
   - The fix is simply look at the real path hierarchy of the file instead (I am not sure if this breaks any of assumption Parcel has of symlinked packages, I saw some `alias` logic along the way..)

## 💻 Examples

I have provided an example project in [`parcel-bundler/test/integration/yarn-workspace`](https://github.com/aliai/parcel/tree/yarn-workspace/packages/core/parcel-bundler/test/integration/yarn-workspace).

## 🚨 Test instructions

You can find the tests in [`parcel-bundler/test/yarn-workspace.js`](https://github.com/aliai/parcel/blob/yarn-workspace/packages/core/parcel-bundler/test/yarn-workspace.js)

## ✔️ PR Todo


aliai:
I am [spawning a process](https://github.com/aliai/parcel/blob/dcc8ec005f2f6545844ba4aa40c3e752ca1450d2/packages/core/parcel-bundler/test/utils.js#L276) to do `yarn install` on my test package during test. I suspect that's what's breaking the checks :( 

Do you know of a better way of doing it?

aliai:
I think I fixed the issues with checks, but they are not running anymore.

kra3:
+1 

we had a similar issue when we moved from a yarn mono repo to parcel. 
It will be great if the fix got in, so we don't have to keep patching it. 

Anything we could do to fast track the review process !!?

aliai:
Dear maintainers, are you working on this issue separately perhaps? or do you not see this as an issue at all? or maybe it is my implementation that sucks so much that you don't know where to begin explaining its terribleness? :P

Please let me know, I can take it!

packages/core/parcel-bundler/src/assets/TypeScriptAsset.js
```diff
@@ -24,13 +25,30 @@ class TypeScriptAsset extends Asset {
       fileName: this.relativeName
     };
 
-    let tsconfig = await this.getConfig(['tsconfig.json']);
+    // Resolve and parse tsconfig.json file
+    const tsconfigCompilerOptions = (() => {
+      const configFilename = typescript.findConfigFile(
+        this.name,
+        typescript.sys.fileExists
```
DeMoorJasper: Is this considered protected by semver by the typescript team?
Seems like an internal function to me.

Don't want to have bug reports about breaking changes on minor releases.

packages/core/parcel-bundler/src/utils/config.js
```diff
@@ -10,7 +10,7 @@ const PARSERS = {
 const existsCache = new Map();
 
 async function resolve(filepath, filenames, root = path.parse(filepath).root) {
-  filepath = path.dirname(filepath);
+  filepath = path.dirname(await fs.realpath(filepath));
```
DeMoorJasper: In what cases are these not realpaths? As far as I know assets are already resolved to their realpath.
I might be mistaken.

packages/core/parcel-bundler/test/integration/yarn-workspace/README.md
```diff
@@ -0,0 +1,12 @@
+# yarn-workspace
```
DeMoorJasper: Amazing to have some tests for yarn workspaces in :)

aliai:
Awesome @DeMoorJasper! Thanks a lot for your review! I know you are doing this on your free time, so I really appreciate it :)

packages/core/parcel-bundler/src/assets/TypeScriptAsset.js
aliai: Hmm, very good point. I did some research, and I have very little evidence that this could last in the future TypeScript releases. In fact, it is said that it might change [here](https://github.com/Microsoft/TypeScript/wiki/Using-the-Compiler-API). I will rewrite this part to reflect the same functionality without relying on TypeScript API.

packages/core/parcel-bundler/src/utils/config.js
aliai: Unfortunately not! There aren't many places where `realpath` is used in the codebase. I couldn't find a place where `realpath` is used for that purpose, and the yarn-workspace test shows that it actually does not resolve into the real path :)

I did a lot of tracing, and I figured this might be the best place to resolve them into real path. But, I may be wrong. Nevertheless, I don't think this would cause any performance issue.

DeMoorJasper: Yeah it looks good, don't expect any real performance issues as configs get cached anyways.

aliai:
I fixed the typescript thingy! But, it seems that the tests are randomly failing, could you please take a look? @DeMoorJasper :) How could you access the xml log file to see what's going on?

aliai:
When I run the tests locally I get the following issues:

1. Issue with `elm`, cause for some reason, installation failed for `elm` during the post installation and therefore `unpacked-bin` folder doesn't exist as it should. (clearing `node_modules` and reinstallation fixed this)
2. Some issues with `javascript` tests. It seems that the tests leave some stuff behind. This makes the results inconsistent when you run the tests multiple times.

Nevertheless, none of the above issues could be the reason behind why the tests are failing for this PR :( Hope we can merge https://github.com/parcel-bundler/parcel/pull/2361 through and get more info :)

I have pushed another [commit](3519fc42b9ab21763e2c8ba708f3b1dd07d19d3a) to address the above two issues I listed. A bit irrelevant to this PR though. Hope that's ok?

packages/core/parcel-bundler/test/javascript.js
```diff
@@ -115,6 +116,12 @@ describe('javascript', function() {
     assert(
       JSON.parse(pkg).devDependencies['@babel/plugin-proposal-class-properties']
     );
+
+    // cleanup node_modules, otherwise the next time we run the tests
+    // it doesn't add it to the "package.json" cause it exists in the node_modules when required
+    await rimraf(
```
DeMoorJasper: Not sure but perhaps we should move the babel autoinstall test to the input folder before running it like we do with other tests that create files

packages/core/parcel-bundler/src/assets/TypeScriptAsset.js
```diff
@@ -1,12 +1,53 @@
...
+      );
+      const mergedConfig = Object.assign({}, parentConfig, tsconfig, {
+        compilerOptions
+      });
+      delete mergedConfig.extends;
+      return mergedConfig;
+    }
+    return tsconfig;
+  }
+
```
devongovett: What do you think of the way #2150 approached this? i.e. using the typescript compiler infrastructure to do the config loading

packages/core/parcel-bundler/test/utils.js
```diff
@@ -272,6 +272,25 @@ function normaliseNewlines(text) {
   return text.replace(/(\r\n|\n|\r)/g, '\n');
 }
 
+function packageInstall(dir) {
```
devongovett: Why is this a test util? Parcel already has autoinstall

aliai: @devongovett I've tried without installing the package before starting the test. But, somehow it fails to recognise that it needs to do installation in the root directory rather than the package itself:

Here is the scenario:
- Yarn workspace example has three packages: `package-a`, `package-b` and `package-main`.
- I am running parcel on `package-main` package (given root directory set to yarn workspace root directory)

`package-main` has a dependency to `package-a`. Parcel recognises that `package-main` is part of `yarn` ecosystem (by traversing the tree down to `yarn.lock` file in the root). However, for some reason, it runs `yarn add package-a` inside `package-main`! Any ideas?

I will look further into this.

aliai: @devongovett Could you help me on this? :)

Do we expect the developers to run `npm install` or `yarn` when they start working on a brand new project they just pulled from a repository? Or do we expect them to run `parcel` right away, and that is supposed to work?

What's happening now is that it tries to install each dependency individually if they are not installed. Isn't that incredibly slow if you have hundreds of dependencies (in a large project code base, which is likely in a yarn workspace projects)?

I was thinking, if this is the path for Parcel, we can still make it happen. I just wanted to make sure we are on the same page. 

The solution I have in mind is somehow modify `resolvePackageAliases` where it checks not only `source`, `alias`, and `browser`, but also `workspaces`. The downside is that to resolve a single dependency, we need to read all `package.json` files defined in `package.json/workspaces` packages, resolve their names and see if the any matches the dependency. We can cache this action, but then we need to think about how we can bust the cache in case the user adds another package folder.

aliai:
Guys, anything I can do to help merge this feature?

packages/core/parcel-bundler/src/assets/TypeScriptAsset.js
devongovett: Yeah but also the config format could also change... I'm not sure we should be duplicating the extension mechanism for that reason.

aliai: Sure thing. I can revert back to what I had with Typescript implementation in this commit (https://github.com/parcel-bundler/parcel/pull/2269/commits/dcc8ec005f2f6545844ba4aa40c3e752ca1450d2). I will fix it tomorrow :) 

But, I have no clue why this PR is failing tests..

Teoxoy:
Any news on this? I would really like to see this merged as well.

PR #2270, opened 2018-11-09
Parcel 2: Multi-target builds

devongovett:
This adds support for multiple target builds to Parcel 2. This enables modern + legacy browser builds, browser + node builds, ES module + CommonJS builds, etc.

It changes the asset graph slightly to accommodate this:

1. Adds a new node type: `transformer_request`. This has a file path and an environment. Dependencies point to these nodes instead of files.
2. `transformer_request` nodes point to other dependencies, along with a `file` node that represents the resolved file and could be shared between multiple transformer requests. When a file is invalidated, we mark all the `transformer_request` nodes connected to that file as invalid.
3. Removes the `connected_file` node type since they are now the same as `file`.

Also adds a `TargetResolver` to core which handles resolving `Target` objects from package.json. There are now multiple dep nodes from the root node for each entry file, one for each target.

The cache has also been updated to store a separate cache entry for each environment a file has been compiled for.

Future work:

1. Update the default resolver plugin to support resolving based on environments, e.g. handling node builtins, `browser` field, etc.
2. Update the default bundler plugin to support creating separate bundles per target.

download-1


packages/core/cache/src/Cache.js
```diff
@@ -49,8 +50,8 @@ export default class Cache {
     this.dirExists = true;
   }
 
-  getCacheId(appendedData: string) {
-    return md5(this.optionsHash + appendedData);
+  getCacheId(appendedData: string, env: Environment) {
```
padmaia: I think it would be nice to have a generic implementation of the cache that is agnostic as to what is being cached, and some abstraction on top of it that is specific to what we are caching. Kind of like Graph and AssetGraph. That way we can have a well tested base that shouldn't change much, and the abstraction on top can be thinner and easier to understand when we change how things are cached. It's not something that needs to be done as part of this PR, just something I've been thinking about.

packages/core/core/src/AssetGraph.js
```diff
@@ -114,44 +126,45 @@ export default class AssetGraph extends Graph {
...
     }
 
     let prunedFiles = getFilesFromGraph(removed);
-    return {newFile, prunedFiles};
+    return {newRequest, prunedFiles};
   }
 
   // Once a file has been transformed, connect it to asset nodes representing the generated assets
-  updateFile(file: File, cacheEntry: CacheEntry): FileUpdates {
+  updateFile(req: TransformerRequest, cacheEntry: CacheEntry): FileUpdates {
```
padmaia: We should probably change the name of this method. I'd be okay with `updateTransformerRequest`, but I'm also thinking that it would be helpful to indicate that we're not just updating something, we are marking it as complete. So maybe something like `completeTranformerRequest`/`completeDependency[Request]` or `resolveTranformerRequest`/`resolveDependency[Request]`?

packages/core/core/src/AssetGraph.js
```diff
@@ -114,44 +126,45 @@ export default class AssetGraph extends Graph {
   }
 
   // Once a dependency is resolved, connect it to a node representing the file it was resolved to
```
padmaia: Comment is a bit outdated now. Update or delete.

packages/core/core/src/TargetResolver.js
```diff
@@ -0,0 +1,112 @@
...
+
+const DEFAULT_ENGINES = {
+  browsers: ['> 0.25%'],
+  node: '8'
+};
+
+const DEFAULT_ENV = {
+  context: 'browser',
+  includeNodeModules: true,
+  engines: DEFAULT_ENGINES
```
padmaia: Does it matter that the `DEFAULT_ENV.context` is 'browser', but `DEFAULT_ENGINES` includes a node version?

packages/core/core/src/TransformerRunner.js
```diff
@@ -70,26 +62,28 @@ class TransformerRunner {
...
-      filePath: file.filePath,
+      filePath: req.filePath,
       ast: null,
       code,
-      env
+      env: req.env
     };
 
     let context = {
+      type: path.extname(req.filePath).slice(1),
```
padmaia: Curious why we are slicing here?

packages/core/cache/src/Cache.js
devongovett: Yeah that's a nice idea. Would they both live in the `@parcel/cache` package though?

packages/core/core/src/TargetResolver.js
devongovett: Mmm, well the compiler should use the engine related to the context it is in so probably ok...

packages/core/core/src/TransformerRunner.js
devongovett: path.extname leaves the `.` at the start

packages/core/core/src/AssetGraph.js
```diff
@@ -185,20 +207,14 @@ export default class AssetGraph extends Graph {
...
-    }
-
-    if (node.type === 'connected_file') {
-      // Invalidate all file nodes connected to this node.
-      for (let connectedNode of this.getNodesConnectedTo(node)) {
-        if (connectedNode.type === 'file') {
-          this.invalidateNode(connectedNode);
-        }
+    // Invalidate all file nodes connected to this node.
+    for (let connectedNode of this.getNodesConnectedTo(node)) {
```
fathyb: Do we still need the File nodes? ie. just use Asset nodes for the invalidation?

devongovett: Well we use file nodes for connected files (like .babelrc) also, and it's useful to be able to easily find the transformer requests for a file when it changes without searching the entire graph.

PR #2271, opened 2018-11-09
Allow multiple extensions.

### LINKED ISSUE #1501: Multiple extension support
# 🙋 feature request
Multiple extension support (e.g tar.gz)

## 🤔 Expected Behavior
Resolve asset with full extension

## 😯 Current Behavior
Resolve asset with part of extension (e.g .gz)

## 💁 Possible Solution
Split extension from the first dot.
https://github.com/parcel-bundler/parcel/blob/24f28bc5d7dd4324841209379d2823654ab9c8e7/src/Parser.js#L69

## 🔦 Context
Create plugin with multiple dots extensions, or just load tar.gz file.

quadrupleslap:
# ↪️ Pull Request

It lets people add asset types that have multiple periods in the extension.

Closes #1501.

## 💻 Examples

```js
// A plugin might do something like this:

bundler.addAssetType('.entity.json', require.resolve('./EntityAsset'));

// Which would mean that the code could do something like this:

import RegularJSON from './regular.json';
import PlayerEntity from './player.entity.json';
```

## ✔️ PR Todo

- [ ] Check that the pull request isn't really stupid.
- [ ] Figure out whether this is backward-compatible with just using `extname`.
- [ ] Add some tests.

### DIFF
--- a/packages/core/parcel-bundler/src/Parser.js
+++ b/packages/core/parcel-bundler/src/Parser.js
@@ -70,13 +70,17 @@ class Parser {
       return GlobAsset;
     }
 
-    let extension = path.extname(filename).toLowerCase();
-    let parser = this.extensions[extension] || RawAsset;
-    if (typeof parser === 'string') {
-      parser = this.extensions[extension] = require(parser);
+    let ext = path.basename(filename).toLowerCase();
+    while ((ext = extension(ext))) {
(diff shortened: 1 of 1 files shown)


devongovett:
Would be great if you could add some tests for this.

quadrupleslap:
Actually, I see extname used in many more places than the one I changed, so I think some more stuff needs to be done. =\

PR #2286, opened 2018-11-16
refactor: declare variable in for to avoid pollution

clarkdo:
# ↪️ Pull Request

Add `var` for the variables in for-loop to avoid global variables pollution.

### DIFF
--- a/packages/core/parcel-bundler/src/builtins/hmr-runtime.js
+++ b/packages/core/parcel-bundler/src/builtins/hmr-runtime.js
@@ -106,8 +106,8 @@ function getParents(bundle, id) {
   var parents = [];
   var k, d, dep;
 
-  for (k in modules) {
-    for (d in modules[k][1]) {
+  for (var k in modules) {
+    for (var d in modules[k][1]) {
       dep = modules[k][1][d];
       if (dep === id || (Array.isArray(dep) && dep[dep.length - 1] === id)) {
         parents.push(k);


packages/core/parcel-bundler/src/builtins/hmr-runtime.js
```diff
@@ -106,8 +106,8 @@ function getParents(bundle, id) {
   var parents = [];
   var k, d, dep;
 
-  for (k in modules) {
-    for (d in modules[k][1]) {
+  for (var k in modules) {
+    for (var d in modules[k][1]) {
```
devongovett: Both of these variables are already declared above, so I'm not sure what you are trying to fix here. If you wanted to make `k` and `d` only accessible within the for loop, you'd need to declare them with `let` anyway in order to make them block scoped rather than function scoped. However, in this instance, we are supporting old browsers, so `let` is not allowed.

devongovett:
Going to close this. I don't think there was a bug here, as the variables were already declared properly.

packages/core/parcel-bundler/src/builtins/hmr-runtime.js
clarkdo: The `var` is for not polluting `k` and `d` into global scope 😄

devongovett: Those variables are declared already on line 107: https://github.com/parcel-bundler/parcel/pull/2286/files#diff-83d6bd1caa31efe06a39bffdd571526cL107

clarkdo: Opps, sorry that I didn't notice, thank you for the reminding 😄

PR #2288, opened 2018-11-16
Fix debug log files for certain locales

phlmn:
When using `--log-level 5` on a system with US locale the command failed because the name of the log file to be created contained slashes (eg. `2018/11/16`).

I replaced `Date.toLocaleString()` with `Date.toISOString()` to fix that and provide consistent results.


phlmn:
Sorry, I did forget to run prettier. 😳

Issue #2214, opened 2018-10-28
Would you consider not relying on deasync if it turns out that it's the only one with native dependencies?


xaduha:
I've grepped deasync and it doesn't seem terribly necessary. Only used here https://github.com/parcel-bundler/parcel/blob/master/packages/core/parcel-bundler/src/Asset.js#L122-L127

https://github.com/parcel-bundler/parcel/blob/75310e15f6fa3c8a0ce6593ada1bcca00240ea54/packages/core/parcel-bundler/src/utils/syncPromise.js

And it causes issues when installing on various platforms https://github.com/parcel-bundler/parcel/issues?utf8=✓&q=deasync

DeMoorJasper:
Well yeah it's deprecated. We could remove it and then the getter should throw an error instead of warn. It would however be a breaking change in the API and that's also the only reason it's there.
Not sure if any plugins still use the old API though. You could do some research and open a PR if you can't find any plugins that use it.

xaduha:
As long as it is going away at some point it's fine by me, v1.6.2 is good enough in the meantime.

DeMoorJasper:
It'll definitely not be part of `Parcel 2` as the plugin API for that is gonna be totally different.

I'll close this for now.

felixfbecker:
This is really painful when building Docker images. If you depend on node:alpine, it does not have Python installed. You need to install Python only for this dependency of Parcel.

PR #2289, opened 2018-11-16
Throw error when asset.package is used and remove deasync

nerdbeere:
Hello there,

this PR removes the `deasync` dependency as suggested in https://github.com/parcel-bundler/parcel/issues/2214

I understand that this introduces a breaking change but I needed it fixed for a project of mine so feel free to close the PR if you like.

For People who need that fix can either use `git+https://git@github.com/nerdbeere/parcel.git#1.x` in their `package.json` or use the temporary package `parcel-bundler-without-deasync`.


DeMoorJasper:
I don’t think any plugins still use the old field so this should be fine afaik

devongovett:
This is technically a breaking change which means we cannot release this until Parcel 2 (which won't need this anyway).

FYI looks like deasync was updated 10 days ago, so perhaps we just need to upgrade it in parcel?

nerdbeere:
@devongovett You are right, it is a breaking api change. So why not release a 2.0.0 containing this fix?

devongovett:
We don't do breaking changes without a really good reason, and this is not one. Can you fix the issue without removing deasync?

DeMoorJasper:
@devongovett doing a pr to this repo (https://github.com/abbr/deasync-bin) with the binaries for Node 11 should probably solve the issues with deasync, as it did when we had a similar issue with node 10 support.

PR #2295, opened 2018-11-18
Make content hash actually work

chaseWillden:
# ↪️ Pull Request

Currently contentHash: false doesn't work, it still adds the hash, this actually makes it work.

## 💻 Examples

contentHash was already implemented, but it just wasn't working

## ✔️ PR Todo


devongovett:
It was actually already working as designed. The option disables **content** hashing, but not the hash altogether. When content hashing is disabled, it uses a hash of the full file path instead. This way the hash doesn't change every time the file contents does. The hash is still necessary though since we output into a flattened directory, which could cause file naming conflicts.

PR #2298, opened 2018-11-18
Switch from toml to @iarna/toml

### LINKED ISSUE #2258: Switch from toml to @iarna/toml
# 💬 RFC

I noticed that y'all are using the [`toml`](https://github.com/BinaryMuse/toml-node) package for dealing with toml files, and that package seems abandoned. Last commit was August 31st, 2017, and it also only supports v0.4.0 of the spec, while the most recent version is v0.5.0.

## 🔦 Context

I've run into issues with this package such as trying to use dotted keys, thinking that they should work since the v0.5.0 spec is the one you get when you google toml. And I would imagine other people probably run into these kinds of problems too.

## 💻 Examples

I would suggest changing over to [`@iarna/toml`](https://github.com/iarna/iarna-toml), as that package supports the latest spec, as well as many other features (25x faster than `toml`, apparently). Plus, it also supports encoding as well as decoding, so you can also remove `tomlify-j0.4` and have less dependencies.

I would make a pull request, but I don't (currently) know how to work with Lerna.
(also, I'm not sure if an RFC was the right thing to open, I'm still a bit new to the open source community)

ABuffSeagull:
# ↪️ Pull Request

As toml is an abandoned package, switch to @iarna/toml
for implementation of the latest toml spec (v0.5.0).
Also remove tomlify-j0.4, as @iarna/toml supports encoding
and decoding.
Resolves #2258.

## 🚨 Test instructions

Reusing the same tests as before, since it's just a package swap.

## ✔️ PR Todo

### DIFF
--- a/packages/core/parcel-bundler/package.json
+++ b/packages/core/parcel-bundler/package.json
@@ -27,6 +27,12 @@
     "@babel/template": "^7.0.0",
     "@babel/traverse": "^7.0.0",
     "@babel/types": "^7.0.0",
+    "@iarna/toml": "^2.2.0",
+    "@parcel/fs": "^1.10.3",
+    "@parcel/logger": "^1.10.3",
+    "@parcel/utils": "^1.10.3",
+    "@parcel/watcher": "1.10.3",
+    "@parcel/workers": "^1.10.3",
     "ansi-to-html": "^0.6.4",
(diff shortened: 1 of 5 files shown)


batisteo:
:blush:

PR #2335, opened 2018-11-29
Parcel 2: Asset methods

devongovett:
This is an experiment with creating a mutable Asset class with some helper methods for transformers to use rather than only returning plain objects. The methods include things like `getConfig` and `getPackage` helpers, along with `addDependency`, `addConnectedFile, and `createChildAsset` mutation/factory methods. None of these are required - plain objects also work fine, and are converted on the fly by the transformer runner. I think it simplifies transformers quite a bit, and also simplifies the transformer runner.

### DIFF
--- a/packages/core/core/src/Asset.js
+++ b/packages/core/core/src/Asset.js
@@ -134,7 +134,7 @@ export default class Asset implements IAsset {
   }
 
   async getConfig(filePaths: Array<FilePath>) {
-    return config.load(filePaths);
+    return config.load(this.filePath, filePaths);
   }
 
   async getPackage() {
--- a/packages/core/core/src/Asset.js
+++ b/packages/core/core/src/Asset.js
@@ -1,0 +1,150 @@
+// @flow
+import type {
+  Asset as IAsset,
+  TransformerResult,
+  DependencyOptions,
+  Dependency,
+  FilePath,
+  File,
+  Environment,
+  JSONObject,
+  AST,
+  AssetOutput
+} from '@parcel/types';
+import md5 from '@parcel/utils/md5';
+import config from '@parcel/utils/config';
+import createDependency from './createDependency';
+
+type AssetOptions = {
+  id?: string,
+  hash?: string,
+  filePath: FilePath,
+  type: string,
+  code?: string,
+  ast?: ?AST,
+  dependencies?: Array<Dependency>,
+  connectedFiles?: Array<File>,
+  output?: AssetOutput,
+  env: Environment,
+  meta?: JSONObject
+};
+
+export default class Asset implements IAsset {
+  id: string;
+  hash: string;
+  filePath: FilePath;
+  type: string;
+  code: string;
+  ast: ?AST;
+  dependencies: Array<Dependency>;
+  connectedFiles: Array<File>;
+  output: AssetOutput;
+  env: Environment;
+  meta: JSONObject;
+
+  constructor(options: AssetOptions) {
+    this.id =
+      options.id ||
+      md5(options.filePath + options.type + JSON.stringify(options.env));
+    this.hash = options.hash || '';
+    this.filePath = options.filePath;
+    this.type = options.type;
+    this.code = options.code || (options.output ? options.output.code : '');
+    this.ast = options.ast || null;
+    this.dependencies = options.dependencies
+      ? options.dependencies.slice()
+      : [];
+    this.connectedFiles = options.connectedFiles
+      ? options.connectedFiles.slice()
+      : [];
+    this.output = options.output || {code: this.code};
+    this.env = options.env;
+    this.meta = options.meta || {};
+  }
+
+  toJSON(): AssetOptions {
+    // Exclude `code` and `ast` from cache
+    return {
+      id: this.id,
+      hash: this.hash,
+      filePath: this.filePath,
+      type: this.type,
+      dependencies: this.dependencies,
+      connectedFiles: this.connectedFiles,
+      output: this.output,
+      env: this.env,
+      meta: this.meta
+    };
+  }
+
+  addDependency(opts: DependencyOptions) {
+    let dep = createDependency(
+      {
+        ...opts,
+        env: mergeEnvironment(this.env, opts.env)
+      },
+      this.filePath
+    );
+
+    this.dependencies.push(dep);
+    return dep.id;
+  }
+
+  async addConnectedFile(file: File) {
+    if (!file.hash) {
+      file.hash = await md5.file(file.filePath);
+    }
+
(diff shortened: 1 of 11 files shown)


packages/core/core/src/TransformerRunner.js
```diff
@@ -136,28 +111,27 @@ class TransformerRunner {
...
           let cacheEntry = await this.runPipeline(
             nextInput,
             pipeline.slice(1),
             null,
-            getNextContext(context, result)
+            generate
           );
 
           assets = assets.concat(cacheEntry.assets);
           connectedFiles = connectedFiles.concat(cacheEntry.connectedFiles);
```
padmaia: I got a little confused here because I thought that cacheEntry was only passed in to the first call of `runPipeline`. I guess it has part of the shape of a CacheEntry, but it might help to not give it the same name as a function parameter. Maybe `nextPipelineResult`?

packages/core/core/src/Asset.js
```diff
@@ -0,0 +1,150 @@
...
+    }
+
+    return asset;
+  }
+
+  async getOutput() {
+    return this.output;
+  }
+
+  async getConfig(filePaths: Array<FilePath>) {
```
padmaia: Is this just so transformer plugins don't have to import the config loader util? If so, should we update the TerserTransformer to use it?

packages/core/types/index.js
```diff
@@ -85,8 +85,7 @@ export type SourceLocation = {
   end: {line: number, column: number}
 };
 
-export type Dependency = {
-  sourcePath: FilePath,
+export type DependencyOptions = {
```
padmaia: Kinda not feeling this name. What about DependencyDescription?

packages/core/core/src/TransformerRunner.js
```diff
@@ -95,36 +78,28 @@ class TransformerRunner {
...
+    let assets: Array<IAsset> = [];
     for (let result of results) {
-      let asset;
-
-      // If this is the first transformer, create a hash for the asset.
-      if (!context.hash) {
-        asset = await transformerResultToAsset(input, result, context);
-        context.hash = asset.hash;
-      }
+      let asset = input.createChildAsset(result);
```
padmaia: If the result is an Asset, could it just be used instead of creating a child asset? Not sure if it matters.

packages/core/types/index.js
```diff
@@ -98,6 +97,16 @@ export type Dependency = {
   meta?: JSONObject
 };
 
+export type Dependency = {
```
padmaia: Wondering if this should be DependencyRequest? Kinda makes sense since it is a combination of a dependency "description" (or whatever we decide to name it) and the site where the dependency was requested from.

packages/core/core/src/TransformerRunner.js
```diff
@@ -136,28 +111,27 @@ class TransformerRunner {
       }
 
       // If the generated asset has the same type as the input...
-      if (result.type === context.type) {
+      // TODO: this is incorrect since multiple file types could map to the same pipeline. need to compare the pipelines.
```
padmaia: Could you give an example of when this might happen?

packages/core/core/src/Asset.js
```diff
@@ -0,0 +1,150 @@
...
+    this.connectedFiles = options.connectedFiles
+      ? options.connectedFiles.slice()
+      : [];
+    this.output = options.output || {code: this.code};
+    this.env = options.env;
+    this.meta = options.meta || {};
+  }
+
+  toJSON(): AssetOptions {
+    // Exclude `code` and `ast` from cache
```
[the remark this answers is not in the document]
devongovett: Probably. That's true, although when run in the worker farm the assets are converted to JSON to pass over the IPC back to the main process.

packages/core/core/src/Asset.js
devongovett: Yes

packages/core/core/src/TransformerRunner.js
devongovett: i.e. if `result instanced Asset`? Yeah it could, but by copying the instance we ensure that future mutations don't break anything (i.e. if a transformer returned a constant for some reason, or had its own cache).

packages/core/core/src/TransformerRunner.js
devongovett: i.e. in the default config, we have `*.{js,mjs,es6}` mapping to a single transformer pipeline. If someone had an `.mjs` file as input, but the output was always `js` from the transformer, then we'd get into an infinite loop here. So we should be comparing the pipelines, not the extensions.

packages/core/core/src/TransformerRunner.js
devongovett: 👍 originally it was a full cache entry and then I changed it

packages/core/core/src/Asset.js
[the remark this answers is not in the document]
padmaia: Yeah I guess it wouldn't be every cache miss, just the cache misses before the workers warm up. Still probably better to guarantee it's removed once finalized though.

packages/core/core/src/TransformerRunner.js
padmaia: I guess I was thinking those would all still by type "js". Is type ever different than the extension?

packages/core/core/src/TransformerRunner.js
```diff
@@ -136,28 +111,27 @@ class TransformerRunner {
...
+            generate
           );
 
           assets = assets.concat(cacheEntry.assets);
           connectedFiles = connectedFiles.concat(cacheEntry.connectedFiles);
         }
       } else {
         // Jump to a different pipeline for the generated asset.
-        let nextInput = transformerResultToInput(input, result);
+        let nextInput = input.createChildAsset(result);
```
padmaia: Looks like this is already being done outside the if/else

packages/core/core/src/TransformerRunner.js
devongovett: Well we initialize `type` to the file extension at the start of the pipeline: 
https://github.com/parcel-bundler/parcel/pull/2335/files#diff-e00a1c679d921332a0033bb784c5d4b7R54 This means that the input type might be `mjs` but the output of the transformer would be `js`. Currently the code would recurse and run the JS transformer again with an input type of `js` rather than stopping. So, we need to compare the pipeline rather than the type.

packages/core/core/src/TransformerRunner.js
devongovett: Good catch!

devongovett:
Going to merge so I can use this in other work.

PR #2341, opened 2018-12-01
use RootDir for localrequire of compilers

### LINKED ISSUE #2318: Parcel cannot find dependencies when run on directory outside project
# 🐛 bug report

Setup

```bash
set -e

# put a sass file in a 'static' dir
mkdir -p static
echo 'p { color: red }' > static/foo.scss

# put a project with parcel in another 'parcel' dir
mkdir -p parcel
cd parcel
echo '{ "devDependencies": { "parcel-bundler": "^1.10.3", "sass": "^1.15.1" } }' > package.json
npm install

# run the local parcel build
./node_modules/.bin/parcel watch ../static/foo.scss
```

## 🤔 Expected Behavior

I would expect the file to build properly with parcel knowing where I called it from and to resolve the `sass` dependency from its local directory.

## 😯 Current Behavior

If run from `/tmp/parcel-folders-test/`, it says it’s installing `sass` and afterwards it gives the error:

> /tmp/parcel-folders-test/static/foo.scss: Cannot find module 'sass' from '/tmp/parcel-folders-test/static'

Furthermore, if you run it as an npm script:

```json
{
  "scripts": {
    "start": "parcel watch ../static/foo.scss"
  },
  "devDependencies": {
    "parcel-bundler": "^1.10.3",
    "sass": "^1.15.1"
  }
}
```

After running it, the package.json file updates to just be:

```json
{
  "devDependencies": {
    "sass": "^1.15.1"
  }
}
```

## 💁 Possible Solution

It seems like the parcel CLI is designed to work with a global install or a local install and it seems to resolve local dependencies by looking at the context of the file that is being bundled. It seems like it should also try to resolve dependencies based on the location of the parcel cli file too.

## 🔦 Context

I was experimenting with adding parcel bundling to a django project that was using docker and was seeing if I could have a separate container do the bundling and I wanted it to be able to work on other directories.

## 🌍 Your Environment

| Parcel           | 1.10.3 |
| Node             | v10.11..0 |
| npm/Yarn         | 6.4.1/1.6.0 |
| Operating System | Mac OSX 10.13.6 |

DeMoorJasper:
# ↪️ Pull Request

This PR ensures Parcel installls all compiler libraries for asset types in the same project (the root project).

This is for file structures where a project has multiple package.json's or a file outside of the project dir is required. (which would both resolve in re-installing the package)

This is a work in progress, still need to test this a little bit and make sure the current tests pass again

Fixes #2318

## ✔️ PR Todo

### DIFF
--- a/packages/core/parcel-bundler/src/assets/CoffeeScriptAsset.js
+++ b/packages/core/parcel-bundler/src/assets/CoffeeScriptAsset.js
@@ -1,5 +1,6 @@
 const Asset = require('../Asset');
 const localRequire = require('../utils/localRequire');
+const path = require('path');
 
 class CoffeeScriptAsset extends Asset {
   constructor(name, options) {
@@ -9,7 +10,10 @@ class CoffeeScriptAsset extends Asset {
 
   async generate() {
     // require coffeescript, installed locally in the app
-    let coffee = await localRequire('coffeescript', this.name);
+    let coffee = await localRequire(
+      'coffeescript',
+      path.join(this.options.rootDir, 'index')
+    );
 
     // Transpile Module using CoffeeScript and parse result as ast format through babylon
     let transpiled = coffee.compile(this.contents, {
--- a/packages/core/parcel-bundler/src/assets/ElmAsset.js
+++ b/packages/core/parcel-bundler/src/assets/ElmAsset.js
@@ -21,7 +21,7 @@ class ElmAsset extends Asset {
     try {
       await commandExists('elm');
     } catch (err) {
-      await localRequire('elm', this.name);
+      await localRequire('elm', path.join(this.options.rootDir, 'index'));
       options.pathToElm = path.join(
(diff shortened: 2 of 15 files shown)


packages/core/parcel-bundler/src/assets/CoffeeScriptAsset.js
```diff
@@ -9,7 +10,10 @@ class CoffeeScriptAsset extends Asset {
 
   async generate() {
     // require coffeescript, installed locally in the app
-    let coffee = await localRequire('coffeescript', this.name);
+    let coffee = await localRequire(
+      'coffeescript',
+      path.join(this.options.rootDir, 'index')
```
DeMoorJasper: Kind of wanna get rid of this `path.join` but not sure how to approach it.
I've seen it at various other places in the codebase as well.

DeMoorJasper:
Not sure why tests are failing. Gonna close this for now.

PR #2349, opened 2018-12-03
refactor: remove unnecessary not null check

clarkdo:
# ↪️ Pull Request

Remove `option` not null check since it won't be null

## :sparkles: PR Todo

### DIFF
--- a/packages/core/watcher/src/options.js
+++ b/packages/core/watcher/src/options.js
@@ -3,7 +3,7 @@ function type(options) {
 }
 
 function encode(options) {
-  if (options.ignored) {
+  if (options && options.ignored) {
     const ignoredType = type(options.ignored);
     if (ignoredType !== 'Array') {
       options.ignored = [options.ignored];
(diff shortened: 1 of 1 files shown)


clarkdo:
@DeMoorJasper Thank you for your reply 😄 

In my perspective, if `option` can be `null`, then we should also check https://github.com/parcel-bundler/parcel/blob/master/packages/core/watcher/src/options.js#L34 since it will lead to exception, otherwise `option` not null check can be removed.

But I can understand your suggestion. This is is not a critical issue or new feature, so you can feel free to close this ticket 😸

DeMoorJasper:
@clarkdo When I was reviewing this, I actually also realised https://github.com/parcel-bundler/parcel/blob/master/packages/core/watcher/src/options.js#L34 should be inside the check as well. If you'd like to fix that that would be awesome, as that would actually prevent issues if we ever allow the watcher to have null as options.

clarkdo:
@DeMoorJasper Updated

Issue #1128, opened 2018-04-03
Error while rendering map with mapbox-gl:  "_typeof is not defined"


dolbyzerr:
This a 🐛 bug report

### 🎛 Configuration (.babelrc, package.json, cli command)

```js
{
  "dependencies": {
    "mapbox-gl": "^0.44.1",
    "parcel-bundler": "^1.7.0"
  }
}
```

```bash
npx parcel index.html
```

### 🤔 Expected Behavior

A map should be rendered.

### 😯 Current Behavior

Grey area is rendered:
<img width="401" alt="screenshot 2018-04-03 12 11 57" src="https://user-images.githubusercontent.com/664335/38230471-3fe3d27e-3738-11e8-919f-89e867513392.png">

The error in console:
`Uncaught ReferenceError: _typeof is not defined`

<img width="534" alt="screenshot 2018-04-03 12 20 10" src="https://user-images.githubusercontent.com/664335/38230723-813c9cfa-3739-11e8-854b-0b2af4cfb96e.png">

### 💁 Possible Solution

As far as I understood from the error and this issue from mapbox https://github.com/mapbox/mapbox-gl-js/issues/3422 it happens because parcel transpiling mapbox file that already being transpiled. Is there any way to alter that behaviour?

### 💻 Code Sample

```html
<html>
<head>
  <link href='https://api.tiles.mapbox.com/mapbox-gl-js/v0.44.1/mapbox-gl.css' rel='stylesheet' />
</head>
<body>
  <div id='map' style='width: 400px; height: 300px;'></div>
  <script src="./index.js"></script>
</body>
</html>
```

```js
import mapboxgl from 'mapbox-gl'

mapboxgl.accessToken = 'pk.eyJ1IjoiZG9sYnl6ZXJyIiwiYSI6InhIS25oN0EifQ.QQrwwFUZu6trJNjGsrpTFQ'
var map = new mapboxgl.Map({
    container: 'map',
    style: 'mapbox://styles/mapbox/streets-v9'
});

```

### 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Parcel           | 1.7.0
| Node             | v8.6.0
| npm         | 5.6.0
| Operating System | macOS 10.13.4

dcposch:
i'm running into the same problem.

[...]

PR #2350, opened 2018-12-03
Parcel 2: Babel Transformer

devongovett:
This is the babel transformer plugin for Parcel 2.

## Changes to core

* The `getConfig` and `getPackage` Asset methods now mutate the asset by adding connected files directly rather than returning them alongside the config. Now they just return the config itself and update the asset to include those file references. This is much easier to use in practice, and the connected files are already propagated along the pipeline.
* The above means the cache entry doesn't need connected files directly anymore since they are all contained within assets instead.
* `getConfig` can be passed options including `packageKey`, which is used to get a key from package.json (e.g. `babel`), and `parse` which can be set to false to disable automatic parsing of the config file (used for `.browserslist` files).

## Differences from Parcel 1

* `.babelrc` config for `babel-preset-env` is no longer merged with our internally generated one. Instead, it overrides the internally generated config. This lets the user have more control when they need it.
* We no longer consider the `engines.node` field in package.json when compiling for a browser target. This was causing issues for people: https://github.com/parcel-bundler/parcel/issues/1128
* We use the `useBuiltIns: 'usage'` mode in babel-preset-env by default. This means that babel-polyfill will be included automatically based on usage of features in code.
* We no longer look in `.babelrc` files in `node_modules` to find browser/node target information.


jamiebuilds:
You're trying to support both Babel 6 and 7 simultaneously in the same transform?

devongovett:
Mmm yeah... that was copied from Parcel 1. I guess we could separate them out. There is a lot shared there though, like the config loading/generation.

jamiebuilds:
Didn't know that, it just seems a bit odd but probably fine

Issue #2137, opened 2018-10-15
Parcel 2: Cache


devongovett:
Extract the cache from Parcel 1, and move into its own package: `@parcel/cache` in `packages/core/cache`. It should also be modified to suit Parcel 2.

TODO: details

mischnic:
https://github.com/parcel-bundler/parcel/pull/2375#issuecomment-446067098:
> The current v2 work is in the v2-work-so-far branch, which does include a new cache already but not the server. 

Should Parcel 1 use that as well or is this resolved?

DeMoorJasper:
@mischnic the Parcel 1 and Parcel 2 cache are not compatible. So it's only for Parcel 2, the Parcel 1 cache should remain as is

devongovett:
Discussed with @padmaia:

* `@parcel/transformer-babel` should store versions of babel plugins in the cache and be able to check that they are the same when checking the cache.
* This will not work in all cases due to `.babelrc.js` and `babel.config.js` which might return functions directly. If those files are seen, don't cache at all.
* It may be problematic if a new `.babelrc` is created further down the tree than before so that files would resolve to a different config. Three options: 1) find all the babelrcs ahead of time. 2) check git to see what changed. 3) just resolve the config every time.
* Check versions of parcel plugins ahead of time to see what pipelines should be invalidated.

devongovett:
I was thinking about caching a bit more, and started wondering if perhaps we can serialize the asset graph to disk after a build and then run a fast comparison when parcel starts a rebuild to determine what nodes to invalidate. This way we can potentially avoid running resolvers and transformers at all for files that do not change, similar to how we do it for rebuilds using info from the watcher.

In order to make comparisons faster, we can also use `stat` information similar to how git does it. If the creation time, modification time, or file size changes, then we know the file is modified and there is no need to actually read the file to compare hashes. It is possible, however, for a file to be modified but for none of these pieces of metadata to change. This can happen if a file is modified more than once in a single second, for example, since many filesystems don't offer high precision modification times. There is some information about how git solves this problem [here](https://mirrors.edge.kernel.org/pub/software/scm/git/docs/technical/racy-git.txt) which we could potentially implement as well.

I don't think we can use git itself to detect changed files since git does not track some files (like node_modules, or anything else in `.gitignore`), and also some users may not be using git.

## Resolver cache

[...]

PR #2375, opened 2018-12-08
Extract Server and Cache into its own package (#2144)(#2137)

### LINKED ISSUE #2144: Parcel 2: Server

tom-julux:
# ↪️ Pull Request

The server and the cache from Parcel 1 was extracted into separate packages, namely `@parcel/server` and `@parcel/cache`. This should resolve #2144 and #2137.

## ✔️ PR Todo
For the server:

- [x] Extract server
- [x] Extract hmr-server
- [x] Find utils the server depends on
- [x] Move these utils
- [x] Make parcel use the new package

For the cache:
- [x] Fix broken tests
- [x] Extract Cache
- [x] Find utils the cache depends on
- [x] Move these utils
- [x] Make parcel use the new package
- [x] Fix broken tests

## Additional hints

I moved the `rimraf` and `ncp` dependencies into the `@parcel/test-utils` package, as `rimraf` was not used anywhere but in `tests/utils.js`. `ncp` was kept as a dependency in the main package, as it is used in a script.

I also moved a few utils into the `@parcel/utils` package.


mischnic:
Tests fail because there of `require("./FSCache")` here : https://github.com/tom-julux/parcel/blob/server-package/packages/core/parcel-bundler/src/Bundler.js#L8

devongovett:
Awesome, thanks for working on this! The current v2 work is in the v2-work-so-far branch, which does include a new cache already but not the server. If you’d like to rebase your work on that branch, and include just the server, that would be awesome. Sorry for the confusion. Once we get v2 into master it will be much easier.

tom-julux:
I will propably be quicker to redo the extraction from scratch, so I will close this pr for now.

PR #2378, opened 2018-12-08
Add support for HMR with elm-hot

ktosiek:
# ↪️ Pull Request

This PR adds integration with elm-hot - the HMR solution for Elm (at least I don't know about any other :-)). I've published this code before as a plugin, and after a suggestion from @domenkozar decided it would make more sense upstream.

## 💻 Examples

This will help when editing the app without changing the type of the app's model. Visual tweaks or testing the update logic should be much easier with working HMR.

## 🚨 Test instructions

Editing an Elm app (like changing a string) should cause the changes to be applied, without refreshing the page or changing the app's state.

## ✔️ PR Todo

- [x] Added/updated unit tests for this change
- [x] Filled out test instructions (In case there aren't any unit tests)
- [ ] Included links to related issues/PRs

PR #2388, opened 2018-12-10
Add support for HMR with elm-hot

ktosiek:
This is a copy of https://github.com/parcel-bundler/parcel/pull/2378, which is having some technical difficulties.

# ↪️ Pull Request

This PR adds integration with elm-hot - the HMR solution for Elm (at least I don't know about any other :-)). I've published this code before as a plugin, and after a suggestion from @domenkozar decided it would make more sense upstream.

## 💻 Examples

This will help when editing the app without changing the type of the app's model. Visual tweaks or testing the update logic should be much easier with working HMR.

## 🚨 Test instructions

Editing an Elm app (like changing a string) should cause the changes to be applied, without refreshing the page or changing the app's state.

## ✔️ PR Todo

### DIFF
--- a/packages/core/parcel-bundler/package.json
+++ b/packages/core/parcel-bundler/package.json
@@ -90,6 +90,7 @@
     "coffeescript": "^2.0.3",
     "cross-env": "^5.1.1",
     "elm": "^0.19.0",
+    "elm-hot": "^1.0.1",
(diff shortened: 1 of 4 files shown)


devongovett:
Awesome, thank you! 🎉

Issue #2396, opened 2018-12-11
Parcel 2: `@parcel/register` node require hook


devongovett:
The idea here is to enable Parcel transformers and resolver to be run in other places than just bundling via a package similar to `babel-register` or other require hooks in node. This way, you could run the same Parcel transformers that you run when bundling your app in your tests, which might run in node.

Currently it is very challenging to set up a real world app, even with Parcel, because you are still stuck configuring other tools like mocha, jest, babel, and others for your tests, not even your real app. If you could get the same zero configuration experience just by loading `@parcel/require` in node, that would be pretty great.

There are a bunch of Parcel features that would be great to be able to use in tests:

1. The `source` and `alias` fields in package.json. No manually configuring `babel-register` to include certain node_modules. Using the Parcel resolver, this would be possible.
2. Real Parcel transformers, no extra configuration. Normal `babel-register` might not be enough if you are using other compilers. We could even automatically handle things like CSS imports that won't work in node.
3. Take advantage of Parcel's cache for faster test runs - no need to recompile every time.

`@parcel/register` would only run the Parcel transformer pipeline, not any packagers or actual bundling. It would operate on a per-file level, and hook into the node module system to compile and resolve modules.

pspeter3:
Is this available? For example, if you're using Jest, is it possible to use Parcel to transform the TypeScript?

pete55104:
This seems important, and I'm interested as well.  In the process of converting a CRA react and node backend, and got as far as trying to run my Jest tests through parcel and ended up here.
I'm interested in working on this feature, although I'd need some guidance.  Let me see if I can get the existing @swc/Jest + babel tooling to work as a model.  

How do I tell which version of swc a given parcel version is using?  I'm on 2.2.1.

devongovett:
I had started on it here. Haven't had time to work on it recently. Feel free to take a look if you want! https://github.com/parcel-bundler/parcel/compare/parcel-node

josheppinette:
Without this, should there be a disclaimer that parcel doesn't support unit testing? One of the key features of parcel is the transformer pipeline, which is unsupported.

andymac4182:
@devongovett Any movement on this from you? Thoughts on starting with getting the new transform and resolve functions you have added to Parcel?

I can see the `@parcel/register` package that is available now. Curious if I am missing a jest one?

PR #2399, opened 2018-12-11
Parcel 2: @parcel/register

DeMoorJasper:
# ↪️ Pull Request

This PR implements `@parcel/register` (related: #2396)

Progress:

- [x] Be able to transform required assets on the fly using parcel core's transformer
- [x] Replace require with Parcel's resolver
- [x] Figure out how to ignore parcel specific modules


packages/core/register/src/register.js
```diff
@@ -0,0 +1,58 @@
+import Parcel from '@parcel/core';
+import {addHook} from 'pirates';
```
devongovett: lol 😆

packages/core/register/src/register.js
```diff
@@ -0,0 +1,58 @@
...
+function fileHandler(opts) {
+  let parcel = new Parcel({
+    entries: [path.join(process.cwd(), 'index.js')],
+    cliOpts: opts
+  });
+
+  return function(code, filename) {
+    // Not sure if this should work?
+    let result = syncPromise(
+      parcel.transform(
```
devongovett: You could use `parcel.runTransform` here to just run the transformer and not try to update the graph or watch any files.

DeMoorJasper: Awesome, that's probably the function I was looking for :)

However I'm having an issue with runtransform as the pipeline always ends up having `undefined` as values. I'll try to debug and fix this asap.

packages/core/register/.eslintrc.json
```diff
@@ -0,0 +1,3 @@
+{
+  "extends": "../../../.eslintrc.json"
```
padmaia: There's `@parcel/eslint-config` package you can extend so you can avoid using relative paths.

packages/core/register/package.json
```diff
@@ -0,0 +1,18 @@
+{
+  "name": "@parcel/register",
+  "version": "2.0.0",
+  "main": "./lib/register.js",
+  "license": "MIT",
+  "scripts": {
+    "test": "echo this package has no tests yet",
+    "test-ci": "yarn build && yarn test",
+    "format": "prettier --write \"./{src,bin,test}/**/*.{js,json,md}\"",
+    "lint": "eslint . && prettier \"./{src,bin,test}/**/*.{js,json,md}\" --list-different",
```
padmaia: eslint and prettier are run from the root by the way, so it's not totally necessary to define package scripts for them

DeMoorJasper: Thanks, I've already removed these in the next commit :)

packages/core/register/.eslintrc.json
DeMoorJasper: Perfect, I'll update it

packages/core/register/src/hook.js
```diff
@@ -0,0 +1,77 @@
...
+      );
+
+      if (result.assets && result.assets.length >= 1) {
+        let codePath =
+          (result.assets[0].output && result.assets[0].output.code) ||
+          result.assets[0].code ||
+          '';
+
+        // Read blobs, replace with cache.readBlobs in the future.
+        let codeContent = fs.readFileSync(
```
DeMoorJasper: How would register be able to access the cache store?

I've seen it getting added to the asset at bundling stage in the default-bundler PR, but afaik this isn't written at the runTransform stage.

packages/core/register/src/hook.js
```diff
@@ -0,0 +1,77 @@
...
+  return function(code, filename) {
+    let isInMonorepo =
+      filename.includes('parcel/packages') &&
+      !filename.includes('core/register');
+    if (filename.includes('@parcel') || isInMonorepo) {
+      return code;
+    }
+
+    try {
+      let result = syncPromise(
```
DeMoorJasper: Not sure if we should keep deasync in parcel

devongovett: haha well... idk how else you could do it if node requires hooks to be synchronous.

packages/core/register/src/hook.js
devongovett: Yeah there is an `asset.getOutput()` function being added in that pr.

DeMoorJasper: Awesome, will update it once that PR is merged

packages/core/register/src/hook.js
```diff
@@ -0,0 +1,77 @@
...
+  // Replace old hook, as this one likely contains options.
+  if (revert) {
+    revert();
+  }
+
+  // Register the hook
+  revert = addHook(fileHandler(opts), {
+    // Parcel should handle all the files?
+    matcher: () => true,
+    ignoreNodeModules: true
```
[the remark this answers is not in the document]
DeMoorJasper: Yeah not sure about this either, it was enabled before and was causing some issues with certain modules as it also overwrites the localRequired modules (so transforms and babel).
Gonna try and figure out how to ignore modules used by parcel.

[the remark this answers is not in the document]
devongovett: ohhhh yeah need to figure out how to ignore things from `localRequire` or parcel itself...

devongovett:
default bundler PR is merged, so you should be able to use `asset.getOutput` now. You might have to do some work to deserialize it from JSON to an actual Asset though. See https://github.com/parcel-bundler/parcel/blob/v2-work-so-far/packages/core/core/src/PackagerRunner.js#L36. I'd love to find a way to make that automatic in the future.

One other thing to work on for `@parcel/register` would be to use the Parcel resolver instead of the node one so we get `alias` and `source` field support, etc. Not sure how easy it is to override that in Node...

DeMoorJasper:
Everything seems to work now, from using Parcel's resolver to resolve requires to processing assets through Parcel.

However I still need to figure out how to optimise some of the stuff (like getting package.json's for checking if it's a parcel dep/plugin or not) and perhaps a way to get the application to exit once it's finished as currently Parcel keeps it from exiting (as the workerfarm still runs)

packages/core/register/src/hook.js
```diff
@@ -0,0 +1,115 @@
...
+
+  if (dir === '/') {
+    return null;
+  }
+
+  return readPkgUp(path.dirname(dir));
+}
+
+function isParcelDep(filename) {
+  let pkg = readPkgUp(path.dirname(filename));
```
devongovett: Maybe just use `loadConfig` from `@parcel/utils/config` to get package.json? It includes caching for free too. 😉

packages/core/register/src/hook.js
```diff
@@ -0,0 +1,115 @@
...
+  return pkg && (pkg.name.includes('@parcel') || pkg.name.includes('parcel'));
```
devongovett: I assume the second check is for third party plugins? I guess I'd look for `pkg.name.startsWith('@parcel') || pkg.name.startsWith('parcel-')` to be more safe.

packages/core/register/src/hook.js
```diff
@@ -0,0 +1,115 @@
...
+  };
+
+  syncPromise(parcel.init());
+
+  // As Parcel is pretty much fully asynchronous, create an async function and wrap it in a syncPromise later...
+  async function fileProcessor(code, filename) {
+    if (isParcelDep(filename)) {
+      return code;
+    }
+
```
devongovett: What about things loaded by parcel plugins, e.g. `babel` or `postcss`? We probably don't want those to be picked up by the hook either right? Maybe we need to track the parent modules somehow and ignore if they came from a parcel dep as well or does pirates already handle this?

packages/core/register/src/hook.js
```diff
@@ -0,0 +1,115 @@
...
+    try {
+      let result = await parcel.runTransform({
+        filePath: filename,
+        env: environment
+      });
+
+      if (result.assets && result.assets.length >= 1) {
+        let asset = new Asset({...result.assets[0], cache});
```
devongovett: Perhaps instead of just taking the first asset, we should search to find one with `type === 'js'` in case they come back in a different order for some reason.

Also, there could be no JS in the result as well. For example, if you required a css file from JS in your app, you probably would want to just ignore that in node e.g. for tests rather than have node throw a parse error. In that case, we could just automatically set the returned code to an empty string maybe. WDYT?

packages/core/register/src/hook.js
```diff
@@ -0,0 +1,115 @@
...
+        },
+        this.filename
+      );
+
+      filePath = syncPromise(parcel.resolverRunner.resolve(dep));
+    }
+
+    let fileExtension = path.extname(filePath);
+    if (!hooks[fileExtension]) {
+      hooks[fileExtension] = addHook(hookFunction, {
```
devongovett: I take it there is no way to just register a hook for all extensions then?

packages/core/register/src/hook.js
```diff
@@ -0,0 +1,115 @@
...
+        exts: [fileExtension]
```
devongovett: Looks like `ignoreNodeModules` is true by default. I think we will need to process them in some cases though...

devongovett:
Nice work! Sorry about the conflicts. The serializer had some refactors.

packages/core/register/src/hook.js
DeMoorJasper: There is not as far as I know.

packages/core/register/src/hook.js
DeMoorJasper: Yeah forgot that was the case, fixed it. Although we probably need to improve parcel dep detection a bit as well, before this would actually work properly

packages/core/register/src/hook.js
```diff
@@ -0,0 +1,142 @@
...
+import syncPromise from '@parcel/utils/lib/syncPromise';
+import {loadConfig} from '@parcel/utils/lib/config';
+
+const originalRequire = Module.prototype.require;
+const DEFAULT_CLI_OPTS = {
+  watch: false
+};
+
+let hooks = {};
+let parcelDeps = [];
```
DeMoorJasper: This array keeps track of every file the hook came across that is used by Parcel.

This should prevent Parcel from processing these, but it could get quite big which is kind of an issue. Perhaps this should be an array of nearest module path or something.

Like traverse up untill it finds a package.json and add that directory, than check if file that's being checked is part of that dir and if it is skip it ...

packages/core/core/src/createDependency.js
```diff
@@ -0,0 +1,18 @@
+// @flow
+import type {DependencyOptions, Dependency, FilePath} from '@parcel/types';
+import md5 from '@parcel/utils/lib/md5';
+
```
devongovett: I think this file got removed

packages/core/register/src/hook.js
```diff
@@ -0,0 +1,137 @@
...
+      let result = await parcel.runTransform({
+        filePath: filename,
+        env: environment
+      });
+
+      if (result.assets && result.assets.length >= 1) {
+        let output = '';
+        let asset = result.assets.find(a => a.type === 'js');
+        if (asset) {
+          asset = new Asset(asset);
```
devongovett: Should already be an asset instance here because of the serializer so no need to construct one

packages/core/register/src/hook.js
devongovett: I wonder if it actually matters if we just process everything, including parcel and its deps. How much slower is it? Most of the time we should just be passing the code straight through with no transforming for those. We could (and probably should) even turn off finding dependencies in the code since we never use them anyway. Does it actually break things to transform everything?

packages/core/core/src/createDependency.js
DeMoorJasper: Ow yes, thought the merge removed it automatically. Will remove it

packages/core/register/src/hook.js
DeMoorJasper: Awesome, will update :)

packages/core/register/src/hook.js
DeMoorJasper: Not skipping Parcel's code makes Parcel's resolver go into an infinite recursive loop. There is also some issues with the actual transforming failing, but can't remember what the error here is. But I doubt we'll be able to get it to transform.

Performance will also be worse but it probably won't be that bad

devongovett: Ah ok.

packages/core/register/src/hook.js
```diff
@@ -15,21 +15,50 @@ const DEFAULT_CLI_OPTS = {
...
+  let pkg = getPkgUp(filename);
+  if (!pkg || !pkg.files[0]) {
+    return null;
+  }
+
+  return path.dirname(pkg.files[0].filePath);
+}
+
+function isParcelModule(filename) {
+  return !!parcelModuleDirs.find(moduleDir => filename.startsWith(moduleDir));
```
DeMoorJasper: This could also be return 

```Javascript
return parcelModuleDirs.includes(getModuleDir(filename));
```

But that's probably not as good for performance as it uses the fs instead of doing simple string comparisons.

DeMoorJasper:
I think all my final concerns have been resolved, it's not as clean and small as I hoped it would be but I think it's pretty stable haven't been able to test it on a larger codebase though.

Would be great to test this with a larger application once Parcel 2 is stable enough to bundle a large application.

packages/core/core/src/Parcel.js
```diff
@@ -257,3 +261,7 @@ export default class Parcel {
     return Promise.all(promises);
   }
 }
+
+export {default as Asset} from './Asset';
+export {default as Dependency} from './Dependency';
+export {default as Environment} from './Environment';
```
devongovett: Probably should make an index file to export all of these and Parcel but ok for now

packages/core/register/src/hook.js
```diff
@@ -0,0 +1,161 @@
...
+
+    return '';
+  }
+
+  let hookFunction = (...args) => syncPromise(fileProcessor(...args));
+
+  function resolveFile(currFile, targetFile) {
+    let resolvedTargetFile;
+    try {
+      resolvedTargetFile = resolveFrom(path.dirname(currFile), targetFile);
```
devongovett: Do we need to do this additional resolution here? Could we just resolve it with parcel no matter what? Seems like resolving twice would decrease performance. Also, if we want to use `@parcel/register` for Parcel itself (in the tests, e.g. to make use of the `source` field), we'd want to use the parcel resolver no matter what.

DeMoorJasper: The additional resolution is just to make sure it's not a parcel dependency as that could cause issues.
At first I actually wanted to use Parcel's resolver on everything but that causes an infinite recursive loop. I could try to figure out a way to go around this, but I doubt it will be an easy fix.

devongovett: Well, just resolving using parcel doesn't mean we will actually compile right?

DeMoorJasper: That's right, but the resolving would cause the issue I mentioned. Sorry if it was a little confusing

devongovett:
Tried to simplify this slightly in https://github.com/parcel-bundler/parcel/pull/2544. Let me know what you think!

Issue #885, opened 2018-02-23
[RFC] Automatically extract modules shared across multiple bundles into their own bundle


jamiebuilds:
Given the following output from today:

```yaml
- bundle-1
  - module-1
  - module-2
  - module-3
- bundle-2
  - module-1
  - module-3
  - module-4
- bundle-3
  - module-4
  - module-5
```

Parcel should extract shared modules into their own bundles which should be side-linked with the bundle they are required in

```yaml
- bundle-1
  - (link shared-bundle-1 for module-1 and module-3)
  - module-2
- bundle-2
  - (link shared-bundle-1 for module-1 and module-3)
  - (link shared-bundle-2 for module-4)
- bundle-3
  - (link shared-bundle-2 for module-4)
  - module-5
- shared-bundle-1
  - module-1
  - module-3
- shared-bundle-2
  - module-4
```

Having a linked bundles should modify the corresponding async import sites to include both the imported bundle and the linked bundles.

```html
<script src="bundle-1"/>
```

Becomes

```html
<script src="shared-bundle-1"/>
<script src="bundle-1"/>
```

This can produce much more efficient bundles with zero duplication and potentially greatly improves cachability.

```
- bundle-1
  - module-2
- bundle-2 (empty)
- bundle-3
  - module-5
- shared-bundle-1
  - module-1
  - module-3
- shared-bundle-2
  - module-4
```

The initial implementation can come in two stages:

1. If a module is seen in more than one bundle, extract it into its own bundle
2. If two modules are always seen together, put them in the same extracted bundle

This should all happen automatically without creating a "commons chunk" or anything manually configured. I'm convinced there is an ideal solution here that won't need to be configured any further. But it will probably take us a little while to get it right.

---

**Note:** Implementing an equivalent to Webpack's CommonsChunkPlugin has come up a few times. I think we should avoid ever creating something like that.

Creating one giant "vendor" bundle is actually the worst possible solution to this problem. It makes for an extremely low cache hit rate across multiple builds and adds the maximum amount of overhead to every bundle that needs it.

jamiebuilds:
Here is some example output from what happens today.

<img width="1342" alt="screen shot 2018-02-22 at 5 50 36 pm copy" src="https://user-images.githubusercontent.com/952783/36574405-8d71eec0-17fa-11e8-8866-a9e35c650b52.png">

starkwang:
Big +1 on it. This feature is very useful for many multiple page applications.

[...]

PR #2401, opened 2018-12-12
Parcel 2: BundleGraph and Default Bundler plugin

### LINKED ISSUE #2139: Parcel 2: BundlerRunner and default bundler plugin
The BundlerRunner object in `@parcel/core` is responsible for running Bundler plugins in Parcel 2. Bundler plugins take in the asset graph and output a list of bundles, which are just a list of Asset ids. These bundles are then passed to packagers to write the output.

The BundlerRunner should resolve Bundler plugins from the Parcel config and execute them. Only one Bundler plugin is allowed at a time.

The default bundler plugin should support all of the features that Parcel 1 supports:

* Code splitting (async imports)
* Duplicate dependency hoisting
* Isolated bundles (e.g. workers)
* Single file bundles (e.g. html shouldn't be combined with other html)

Bundles will be not be named by Bundler plugins. There will be separate Namer plugins to do that.

devongovett:
This is an implementation of the default bundler plugin for Parcel 2, along with the infrastructure in core necessary to support it. It implements code splitting, splitting by asset type, asset deduping, and bundle splitting.

Closes #2139. Related to #885.

* Adds a `BundleGraph` to core, and changes the Bundler plugin API to return one. Bundler plugins accept the complete asset graph, along with an empty bundle graph instance to fill in. This way we don't need to expose the constructor to plugins. A `BundleGraph` contains nodes of type `bundle_group` and `bundle`.
* Each `Bundle` object now has an `AssetGraph` in it instead of just an array of assets. This gives packagers access to the dependencies and other information in the graph for that bundle.
* Adds a bunch of helper methods that will be useful for bundler plugins to `BundleGraph` and `AssetGraph`, along with some traversal methods to `Graph` itself.
* Changes existing JS packager to use the AssetGraph within the bundle to write its assets. This also included changing the way asset outputs were read from the cache to do it within the packager via `asset.getOutput()` rather than beforehand in the PackagerRunner.

The default bundler traverses the asset graph and creates a node of type `bundle_group` for every entry point and async dependency. Connected to these are `bundle` nodes, which each contain a subset of the full asset graph. The point of the `bundle_group` nodes is to group things that would be loaded in parallel, e.g. JS + CSS, or JS and other bundle split JS.

Child bundle groups can share dependencies with their parents, so assets that would be duplicated are removed from children. When a dependency with an isolated environment (e.g. a web worker) is hit where it would be impossible to access parent dependencies, it is connected to the root of the bundle graph instead. This way it will have no parents and thus not share anything with other bundles.

graph

Within each `bundle` node is a sub asset graph containing only the assets in that bundle. These asset graphs inside bundles start from one or more assets and point to the rest of the subgraph for those assets. When an async dependency is hit that creates a separate bundle, the dependency resolution is replaced with the `bundle_group` node from the bundle graph for this dependency, which gives packagers access to the referenced bundles that should be loaded.

graph copy

The above example shows three async dependencies, two of which depend on react and lodash. Since those are large libraries and they are shared, they are extracted out into their own separate bundle which is added to the original bundle groups to be loaded in parallel. This allows them to be cached independently of the original bundles, decreasing the overall application size.

Within those async bundles, the separated assets which are no longer in the original bundles are replaced with `asset_reference` nodes, which give packagers access to the original assets without including them or their dependencies in the bundle.

graph

The full rules for the default bundler are as follows:

1. If an async or entry dependency is seen, start a new bundle group.
2. If an asset is a different type than the current bundle, make a parallel bundle in the same bundle group.
3. If an asset is already in a parent bundle in the same entry point, exclude from child bundles.
4. If an asset is only in separate isolated entry points (e.g. workers, different HTML pages), duplicate it.
5. If the sub-graph from an asset is >= 30kb, and the number of parallel requests in the bundle group is < 5, create a new bundle containing the sub-graph.
6. If two assets are always seen together, put them in the same extracted bundle.
7. Consider larger split bundles before smaller ones.

Feedback on the API for AssetGraph and BundleGraph would be appreciated! One idea I had was to make bundler plugins composable, but I will explore that in a future PR. 😄

### DIFF
--- a/packages/core/core/src/AssetGraph.js
+++ b/packages/core/core/src/AssetGraph.js
@@ -98,7 +98,7 @@ export default class AssetGraph extends Graph {
 
   initializeGraph({entries, targets, rootDir}: AssetGraphOpts) {
     let rootNode = nodeFromRootDir(rootDir);
-    this.addNode(rootNode);
+    this.setRootNode(rootNode);
 
     let depNodes = [];
     for (let entry of entries) {
@@ -107,7 +107,8 @@ export default class AssetGraph extends Graph {
           createDependency(
             {
               moduleSpecifier: entry,
-              env: target.env
+              env: target.env,
+              isEntry: true
             },
             path.resolve(rootDir, 'index')
           )
@@ -220,6 +221,27 @@ export default class AssetGraph extends Graph {
     }
   }
 
+  getIncomingDependencies(asset: Asset): Array<Dependency> {
+    let assetNode = this.getNode(asset.id);
+    if (!assetNode) {
+      return [];
+    }
+
+    let transformerRequests = this.getNodesConnectedTo(assetNode);
+    let results = [];
+
+    for (let req of transformerRequests) {
+      let deps = this.getNodesConnectedTo(req);
+      for (let dep of deps) {
+        if (dep.type === 'dependency') {
+          results.push(dep.value);
+        }
+      }
+    }
+
+    return results;
+  }
+
   async dumpGraphViz() {
     let graphviz = require('graphviz');
     let tempy = require('tempy');
@@ -237,14 +259,6 @@ export default class AssetGraph extends Graph {
     };
 
     let nodes = Array.from(this.nodes.values());
-    let root;
-    for (let node of nodes) {
-      if (node.type === 'root') {
-        root = node;
-        break;
-      }
-    }
-    let rootPath = root ? root.value : '/';
 
     for (let node of nodes) {
       let n = g.addNode(node.id);
@@ -260,9 +274,8 @@ export default class AssetGraph extends Graph {
         let parts = [];
         if (node.value.isEntry) parts.push('entry');
         if (node.value.isAsync) parts.push('async');
-        if (node.value.isIncluded) parts.push('included');
         if (node.value.isOptional) parts.push('optional');
-        if (parts.length) label += '(' + parts.join(', ') + ')';
+        if (parts.length) label += ' (' + parts.join(', ') + ')';
         if (node.value.env) label += ` (${getEnvDescription(node.value.env)})`;
       } else if (node.type === 'asset') {
         label += path.basename(node.value.filePath) + '#' + node.value.type;
--- a/packages/core/core/src/AssetGraph.js
+++ b/packages/core/core/src/AssetGraph.js
@@ -9,11 +9,15 @@ import type {
   FilePath,
   TransformerRequest,
   Target,
-  Environment
+  Environment,
+  Bundle,
+  TraversalContext
 } from '@parcel/types';
 import path from 'path';
 import md5 from '@parcel/utils/md5';
 import createDependency from './createDependency';
+
+let BUNDLECOUNT = 0;
 
 export const nodeFromRootDir = (rootDir: string) => ({
   id: rootDir,
@@ -90,8 +94,8 @@ export default class AssetGraph extends Graph {
   incompleteNodes: Map<NodeId, Node>;
   invalidNodes: Map<NodeId, Node>;
 
-  constructor() {
-    super();
+  constructor(opts) {
+    super(opts);
     this.incompleteNodes = new Map();
     this.invalidNodes = new Map();
   }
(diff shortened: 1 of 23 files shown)


packages/core/core/src/AssetGraph.js
```diff
@@ -220,6 +226,96 @@ export default class AssetGraph extends Graph {
...
+      id: 'root',
+      value: null
+    });
+
+    graph.addEdge({from: 'root', to: assetNode.id});
+    return {
+      id: 'bundle:' + asset.id,
+      type: asset.type,
+      assetGraph: graph,
+      filePath: 'bundle.' + BUNDLECOUNT++ + '.js'
```
devongovett: Will be replaced by namer plugins

packages/core/core/src/AssetGraph.js
```diff
@@ -220,6 +226,96 @@ export default class AssetGraph extends Graph {
     }
   }
 
+  getDependencies(asset: Asset): Array<Dependency> {
```
devongovett: How do we feel about adding all of these helper methods meant for external plugin consumption to the existing AssetGraph class used in core?

packages/core/core/src/PackagerRunner.js
```diff
@@ -27,6 +29,14 @@ export default class PackagerRunner {
   }
 
   async writeBundle(bundle: Bundle) {
+    // deserialize asset graph from JSON
```
devongovett: Would be awesome to have a better way of serializing and deserializing complex objects from JSON like this...

packages/core/core/src/BundleGraph.js
```diff
@@ -0,0 +1,140 @@
...
+  Bundle,
+  BundleGroup,
+  GraphTraversalCallback
+} from '@parcel/types';
+import AssetGraph from './AssetGraph';
+
+const getBundleGroupId = (bundleGroup: BundleGroup) =>
+  'bundle_group:' + bundleGroup.dependency.id;
+
+export default class BundleGraph extends AssetGraph {
```
devongovett: Meant to make this extend from `Graph` instead of `AssetGraph` obviously, but `AssetGraph` has the nice `dumpGraphViz` function. Will move that to `Graph` later.

packages/core/core/src/AssetGraph.js
padmaia: I'd rather not expose the raw AssetGraph. Ideally we'd wrap it and only expose what we want the plugins to use.

devongovett: Yeah I agree with that. Let's get this PR in first and then discuss what the interface we want to expose is.

packages/core/core/src/Graph.js
```diff
@@ -154,4 +198,128 @@ export default class Graph {
...
+    return this.dfs({
+      visit,
+      startNode,
+      getChildren: this.getNodesConnectedFrom.bind(this)
+    });
+  }
+
+  traverseAncestors(
+    startNode: Node,
+    visit: (node: Node, context?: any, traversal: TraversalContext) => any
```
padmaia: Most of the other traversal methods have visit as the first arg, is there a reason why this one is different?

devongovett: well... in the other ones it is optional, whereas this one it is required. So I felt like it looked nicer to have the callback at the end in this case.

padmaia:
All in all, this looks really good! The graph visualizations really help explain what's going on. My one constructive criticism would be that it got a little confusing following how the bundle asset graphs are morphed by the bundler. I think the idea we talked about of exposing a wrapper or a different asset graph class to the bundler plugin would naturally help with this. It might help to strip out nodes that aren't really necessary for bundling, like files and transformer requests. Would be less cognitive load for bundler plugin writers and would be nicer to introspect.

packages/transformers/js/src/JSTransformer.js
```diff
@@ -89,9 +89,9 @@ export default new Transformer({
       }
 
       // Insert node globals
-      if (GLOBAL_RE.test(asset.code)) {
-        walk.ancestor(asset.ast.program, insertGlobals, asset);
-      }
+      // if (GLOBAL_RE.test(asset.code)) {
+      //   walk.ancestor(asset.ast.program, insertGlobals, asset);
+      // }
```
padmaia: Will this be uncommented eventually?

devongovett: yeah... that transform is currently broken

devongovett:
Yeah I was thinking about removing transformer requests and file nodes from the asset graphs inside bundles as well. Would definitely make things simpler. Need to figure out how to do the graph transforms for that when creating a bundle.

One idea I was also thinking about was how to make bundler plugins more composable so you could use the default one in addition to a third party plugin without the third party one needing to reimplement all of the features of the default. We could possibly do this by making the input to bundler plugins only a bundle graph instead of also the original asset graph. The initial bundle graph passed in would contain a single bundle containing all of the assets, and then the bundler plugins would take that bundle graph and modify it to include other bundles. This way you could chain multiple bundlers together by passing the bundle graph to each one to modify. Thoughts?

padmaia:
Hmm, do you have an example of when that would be useful? Seems like it would help if you want to break the bundles apart even more, but not very helpful if you want to rearrange what is in the bundles.

devongovett:
Yeah that's true. If we broke apart the default bundler slightly into its three phases so they could be composed together that might make it easier to keep parts and overrides parts. e.g. keep the dynamic import handling, but override the bundle splitting behavior. idk, I haven't fully thought it through yet.

devongovett:
Going to merge this to get some stuff in. Will go back and refactor some of this later.

Issue #2390, opened 2018-12-11
Is it possible to get parcel's dev server (parcel serve) to log HTTP requests?


rosensama:
# ❔ Question

Is there a log level I can set where parcel's dev server, invoked by `parcel serve` will log the browser requests it is serving?  

## 🔦 Context

When I'm debugging routing bugs in an SPA it's helpful to see the requests the UI made to the dev server or when index.html is served to it again unexpectedly.  

## 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Parcel           | 1.10.3 |
| Node             |11.2.0|
| npm/Yarn         | npm 6.4.1| 
| Operating System | OSX mojave|

I use parcel's dev server to deliver the UI and handle HMR.  The UI makes CORS requests to a separate REST API server, so requests to the dev server are usually of interest. 

Thanks,
Rob

DeMoorJasper:
Currently not but I guess this could be added to log-level 4 (verbose)

mischnic:
```
[7:48:28 PM]: Request: http://localhost:1234/style.e308ff8e.css
```
Requests are now logged with `parcel --log-level 4`. Not yet added to the documentation: https://github.com/parcel-bundler/website/issues/323

PR #2402, opened 2018-12-12
Log dev server access for log level verbose or more

lustoykov:
Attempts to address https://github.com/parcel-bundler/parcel/issues/2390


packages/core/parcel-bundler/test/server.js
```diff
@@ -1,9 +1,11 @@
 const assert = require('assert');
 const path = require('path');
 const fs = require('@parcel/fs');
+const logger = require('@parcel/Logger');
```
lustoykov: any guidance on how to import that properly? Seems to work locally, but not in CI.

DeMoorJasper: You have a typo it's lowercase `l`, `'@parcel/logger'`

packages/core/parcel-bundler/src/Server.js
```diff
@@ -107,6 +109,13 @@ function middleware(bundler) {
       res.writeHead(404);
       res.end();
     }
+
+    function logAccessIfVerbose() {
+      const protocol = req.connection.encrypted ? 'https' : 'http';
+      const fullUrl = `${protocol}://${req.headers.host}${req.url}`;
+
+      logger.verbose(fullUrl);
```
DeMoorJasper: Probably a good idea to add some more info here, besides from just a url, something like `Request: <url>`

lustoykov:
Thanks for the input! I think we're getting there 😆 

CI fails with one last error, but I'm clueless about it, see screenshot

screenshot 2018-12-13 at 16 52 55

Also, here's a screenshot of the final solution

screenshot 2018-12-13 at 16 12 17



**On a side note:** I've noticed the test `1) should auto install babel plugins` in `javascript.js` (besides some others vue-related) fails and this might be due to faulty .babelrc configuration left from a possible migration Babel 6 -> Babel 7.

I think in Babel 7 the **non-root .babelrc** is being ignored since not explicitly specified otherwise. Therefore `@babel/plugin-proposal-class-properties` is never being installed and the assertion fails. The other problem is that even if .babelrc is explicitly specified, test is fine only on first run where `@babel/plugin-proposal-class-properties` is actually installed and package.json modified respectively. However, in the test itself package.json is reverted to its previous state. In the next run, babel sees that `@babel/plugin-proposal-class-properties` is in `node_modules` and doesn't bother to modify `package.json` and the assertion fails again.

Issue #2391, opened 2018-12-11
Parcel 2: Setup integration tests


devongovett:
Get the existing integration tests from Parcel 1 running against Parcel 2. Most of the tests in Parcel 1 used test utils to run Parcel and assert things against the generated bundles. It should be mostly possible to create a compatible set of test utilities for Parcel 2 and run the same tests against both.

As discussed, one change would be to change the `assertBundleTree` method to just `assertBundles`, since the actual position of the bundle in a tree is not important, just that it exists and contains the right assets.

For bonus points, create a website with a graph tracking the number of tests remaining to fix in Parcel 2. 😉

PR #2410, opened 2018-12-14
pull out integration tests

padmaia:
# ↪️ Pull Request

See #2391 for details


packages/core/integration-tests/test/utils.js
```diff
@@ -0,0 +1,283 @@
+const Bundler = require('parcel-bundler');
```
padmaia: You might notice I have the utils duplicated here and in parcel-bundler. I did this mainly because they currently depend on existing in the same test folder where the tests that are using them live. Pulling them out would require refactoring pretty much all of the tests. Also there are still a couple integration tests that I felt were v1 specific that I kept in parcel-bundler, which created a circular dependency when I did try pulling them out. The simplest and safest solution seemed to be a tiny bit of duplication, but let me know if that's a deal breaker. I was thinking that we could pull out the integration utils in the v2 branch where we are going to need to refactor the tests to work against v2 anyway. This will also allow us to refactor in a more granular way as we add features and unskip the tests one by one instead of doing them all in one go and maybe missing something. Moving around and debugging a bunch of tests has not been the most fun task :P

devongovett:
CI seems to be failing for some reason...

packages/core/integration-tests/test/.eslintrc.json
```diff
@@ -0,0 +1,6 @@
+{
+  "extends": "../../../../.eslintrc.json",
```
devongovett: Can use `@parcel/eslint-config` here

packages/core/integration-tests/test/integration/.eslintrc.json
```diff
@@ -0,0 +1,13 @@
+{
+  "extends": "../.eslintrc.json",
```
devongovett: same

PR #2411, opened 2018-12-14
Track v2 progress

padmaia:
# ↪️ Pull Request

See #2391 for details.

~This PR depends on #2410 being merged into master, and then pulling those changes in v2-work-so-far.~

### DIFF
--- a/packages/core/is-v2-ready-yet/package.json
+++ b/packages/core/is-v2-ready-yet/package.json
@@ -1,6 +1,12 @@
 {
   "name": "@parcel/is-v2-ready-yet",
   "version": "2.0.0",
+  "private": true,
+  "license": "MIT",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/parcel-bundler/parcel.git"
+  },
   "scripts": {
     "start": "parcel index.html",
     "build": "parcel build index.html"


packages/core/integration-tests/ready-yet-reporter.js
```diff
@@ -0,0 +1,56 @@
...
+    commitDate = commitDate.substring(0, commitDate.lastIndexOf(' ')); // Get rid of timezone
+    let testHistory = JSON.parse(
+      fs.readFileSync('data/testHistory.json', 'utf8')
+    );
+    testHistory.push(`${commitHash}\t${commitDate}\t${ratio}`);
+    fs.writeFileSync(
+      'data/testHistory.json',
+      JSON.stringify(testHistory),
+      'utf8'
+    );
```
padmaia: @devongovett what's the best way to persist the test history file? I was thinking the simplest way would have the CI server commit and push based on test results for every push to the v2 branch.

padmaia: Could maybe even deploy the app after too, but not sure if you want to give the CI server that much responsibility.

devongovett: We could probably just run `now` in there to deploy it from CI?

devongovett:
This is awesome. So excited!

devongovett:
Looks great. Deployed at https://isparcel2readyyet.com! 🎉 

I guess we should probably give credit somewhere to isfiberreadyyet.com for the design?

Last thing is to figure out how to get this deployed automatically as part of CI? I deployed it manually with zeit now, but in order to do it automatically we'd want to only deploy it when the test results changed or the app itself changed, and we are on `v2-work-so-far`. That seemed like too much work, so unless you have a good idea, I'm fine with just manually running it from time to time. WDYT?

padmaia:
Yeah I agree, I think it's fine to deploy it manually. I do think we should capture every commit that affects the ratio though, so that should probably be automated on the CI server. I added a change to only write record test data when the ratio has changed. Do you think I could just add an `execSync` for `git add` and `git push` in that same block?

devongovett:
Yeah we could, or we could just run it prior to deploying manually.

The problem will be that the tests run many times across multiple build machines to test in multiple environments, but we'd only want to commit the results once.

padmaia:
Ah right. My only concern with running it manually is that we might lose out on some data points. It's pretty cool to be able to click on a point on the graph and be taken to the commit that turns those tests on and gets them passing. Not the end of the world if there are a few missing though I guess. Maybe anyone who contributes to getting some tests passing and really wants to make sure their effort is shown in the graph they can just commit the test history after a local test run.

devongovett:
yeah, I guess we could add an additional run in CI to capture this, or somehow only do it on one of the existing runs instead of all of them... but probably easier to just run it manually.

PR #2424, opened 2018-12-17
Parcel 2: Default bundle namer

### LINKED ISSUE #2140: Parcel 2: NamerRunner and default namer plugin
The NamerRunner object in `@parcel/core` is responsible for running Namer plugins in Parcel 2. Namer plugins name the output bundle files defined by Bundler plugins.

The NamerRunner should resolve Namer plugins from the Parcel config and execute them. Namer plugins can be chained. If one namer plugin does not return a name for a bundle, then the next one should be applied and so on.

The input to a Namer plugin is a Bundle object (which is just a list of Asset objects and some metadata about the reason the bundle was created by the Bundler plugin). The output is a list of Bundle objects with the name filled in. (cc. @jamiebuilds @padmaia do we want to have the bundle objects mutated, cloned, or return a map of bundle ids -> name or something? not sure here.)

The default namer plugin should follow the same naming rules as in Parcel 1:

* Entry bundles should retain the original filename, and relative path inside the `dist/` folder. This includes explicit entries from the CLI or API, as well as assets marked as entries by transformers, e.g. `<a href>` in HTML, and service workers in JS.
* Other bundles should get their original filename along with a content hash. In development mode, a hash of the relative path to the original file from the root directory is used instead so that it doesn't change on every build.
* Files with `index` as their basename (e.g. `index.js`), get their filename based on the parent directory name instead, for slightly nicer looking names. e.g. `foo/index.js` might get named `dist/foo.ad73fb.js`.

devongovett:
This implements bundle naming using namer plugins, and provides the default namer plugin for Parcel 2.

Closes #2240. Fixes #2140.

Summary of changes:

* Implements default namer plugin, which implements the rules set out in #2140 including content hashing.
* Marks bundles as entry bundles in the default bundler. Entry bundles are not content hashed.
* Propagates targets from the initial dependency created for the original entries to all bundles in the bundle graph.
* Uses user-defined filenames from targets for original entry bundles. Other bundles that are produced from this entry point are created in the same directory.
* Adjusts targets to make explicit output filenames optional. If not provided, parcel uses its default file naming scheme.
* Adds an `outputHash` property to `Asset` which is computed at the end of the transformer pipeline. The current `hash` property is computed after the first transform and is used for caching. The `outputHash` is used for content hashing.


packages/namers/default/src/DefaultNamer.js
```diff
@@ -0,0 +1,63 @@
...
+const COMMON_NAMES = new Set(['index', 'src', 'lib']);
+const DEFAULT_DIST_DIR = 'dist';
+
+export default new Namer({
+  name(bundle) {
+    // If the bundle has an explicit file path given (e.g. by a target), use that.
+    if (bundle.filePath) {
+      // TODO: what about multiple assets in the same dep?
+      // e.g. input is a Vue file, output is JS + CSS
+      // which is defined as a target in package.json?
```
devongovett: Any ideas how we could tell which one is the main one? Should we take the specified filename and just replace the extension with the bundle type?

packages/namers/default/src/DefaultNamer.js
```diff
@@ -0,0 +1,63 @@
...
+      return bundle.filePath;
+    }
+
+    // TODO: if split bundle, base name on original bundle names?
```
devongovett: Any ideas on how to get the information on the original bundles that the split bundle originated from here? I'd rather not add a reference to the original bundles to the bundle object but that may be the only way...

padmaia: Only other way I could think of would be to pass in the bundleGraph along with the bundle to be named.

devongovett: That could possibly work. The split bundle will be connected to both original bundle groups, but there is currently no way to know which bundle connected to a bundle group was the original one...

packages/namers/default/src/DefaultNamer.js
padmaia: Yeah, I think that works.

packages/namers/default/src/DefaultNamer.js
```diff
@@ -0,0 +1,63 @@
...
+    let entryAsset = bundle.assetGraph.getEntryAssets()[0];
+    let entryFilePath = entryAsset.filePath;
+    let name = path.basename(entryFilePath, path.extname(entryFilePath));
+
+    // If this is an entry bundle, exclude the hash and use the original relative path.
```
padmaia: What's the reasoning for this again?

devongovett: If you have `index.html` as an entry point, you typically want to keep that name rather than adding a content hash. Otherwise the link would change every time you deploy. Same thing for things like service workers, where the URL can never change.

Here is the original RFC: https://github.com/parcel-bundler/parcel/issues/872

Issue #1315, opened 2018-05-07
Index.html can not be omitted when serving multiple files


tamuraryoya:
# 🐛 bug report

When bundling and serve multiple entry files, `/ index.html` can not be accessed with `localhost:1234/`

However, you can access it with `localhost: 1234 / index.html`

## 🎛 Configuration (.babelrc, package.json, cli command)

```pug
//- src/index.pug
h1 hello world!
```

```pug
//- src/sub.pug
h1 sub page!
```

```
$ parcel 'src/**/*.pug'
```

## 🤔 Expected Behavior

When accessing `localhost:1234/`, "hello world!" is displayed

## 😯 Current Behavior

Accessing `localhost:1234/` will display empty HTML

```html
<html>
  <head></head>
  <body></body>
</html>
```

However, accessing `localhost: 1234 / index.html` will display the page correctly

## 💁 Possible Solution

I do not know, but when accessed with `/` we have to make sure to check the existence of `/ index.html`

## 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Parcel           |v1.8.1|
| Node             |v8.11.1|
| npm/Yarn         |v6.0.0/v1.5.1|
| Operating System |MacBook Pro (High Sierra)|

DeMoorJasper:
This is because it currently checks for an entrypoint and multiple entrypoints works with a html entrypoint thats being created in the background as a main asset to link all assets too

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

lucaseverett:
It would be helpful to have index.html served from / and any other entry points served without .html.

So, page1.html would be resolved by visiting /page1.

Is there a way to create URL rewrites for the server locally in a project?

MathieuLoutre:
I've got a working solution for this (as explained in #1778) and can PR if this is agreeable.

Essentially, it enhances the `respond` method by looking for indexes (with a method replacing `sendIndex` called `findIndex`), looking for the main bundle as it does now, failing that looking for an HTML file called `index.html`.

```js
 function respond() {
      let {pathname} = url.parse(req.url);
      if (bundler.error) {
        return send500(bundler.error);
      } else if (
        !pathname.startsWith(bundler.options.publicURL) ||
        path.extname(pathname) === ''
      ) {
        const index = findIndex(pathname)
        if (index) {
          req.url = `/${path.basename(index)}`;
          serve(req, res, send404);
        }
        else {
          send404();
        }
[...]

PR #2426, opened 2018-12-17
Serve index by default

LuudJanssen:
# ↪️ Pull Request

Solves #1315. When you have two entry points, say index.html and test.html, this code makes sure the server always serves index.html when hit with the folder both files are in, thus mimicking real world server which will also serve index.html by default. 

[This comment](https://github.com/parcel-bundler/parcel/issues/1315#issuecomment-441674707) from me on the issue explains a bit more about the implementation. 

## 💻 Examples

Given the directory structure after a build with two entry points (index.html and test.html):
dist/index.html
dist/test.html

The server should return index.html when localhost:1234/. Current behavior is returning neither index.html nor test.html and hence the request fails.

## 🚨 Test instructions

I wrote tests to test this in. It basically tests if the parcel server serves index.html if two entry points are given.

**However, I wasn't able to test this locally on my Windows machine due to [this issue](https://spectrum.chat/parcel/contributing/only-test-server-js-because-of-test-error~ff160383-9fc5-4a2d-a0d9-0228a578c763), hence the tests should be ran to test if they are correct and don't fail.**

## ✔️ PR Todo


danielwaltz:
Any updates on this? Running into this issue when using a glob as my entrypoint.

LuudJanssen:
@danielwaltz Update: Fixed most of the tests, but one failed test remains. Will be working on this in the coming days.

LuudJanssen:
@DeMoorJasper, I fixed some issues that were introduced while this PR was active, merged master and got the tests working. Should be okay now, can you take another look?

packages/core/parcel-bundler/src/Server.js
```diff
@@ -57,9 +57,7 @@ function middleware(bundler) {
         !pathname.startsWith(bundler.options.publicURL) ||
         path.extname(pathname) === ''
       ) {
-        // If the URL doesn't start with the public path, or the URL doesn't
-        // have a file extension, send the main HTML bundle.
-        return sendIndex();
+        sendIndex();
```
mrcoles: why did you drop the `return` here? It looks like it doesn’t matter, but seems like the code should be consistent with the other if-statement branches here.

```suggestion
        return sendIndex();
```

packages/core/parcel-bundler/src/Server.js
```diff
@@ -68,15 +66,33 @@ function middleware(bundler) {
...
+      if (bundler.mainBundle.type === 'html') {
+        // If the bundle is HTML, send that
+        return bundler.mainBundle.name;
+      } else if (pathname === bundler.options.publicURL) {
+        // if the pathname is empty, look for files named index.html in the bundle and send back the first one. If there's nothing, return null;
+        const indexes = Array.from(bundler.mainBundle.childBundles).filter(
+          asset => path.basename(asset.name) === 'index.html'
+        );
+        return indexes[0].name;
+      }
```
mrcoles: It seems like the logic here should be:

1. Is there a file that matches something like `${req.url}/index.html`, then serve that
2. If not, then is there an `index.html` at the root of the dist directory, then serve that
3. Otherwise 404

IDK if that’s kind of special-cased or not, but when you only specify one entry point, `parcel serve` always returns that file for *any* url. In my situations where I have had multiple entry points, I’ve wanted the root index file to be the catch-all and any other specific ones to be accessible directly. My use-case is having an index.html at the root running React Router and another file that lives outside of my react app that can be accessed directly.

packages/core/parcel-bundler/src/Server.js
LuudJanssen: Wasn't done on purpose, must've not noticed the inconsistency. I've applied the suggestion

packages/core/parcel-bundler/src/Server.js
LuudJanssen: Hey @mrcoles, thanks for your review! If I understand you correctly you want the order of resolving to be as you stated in your list (first specified file, then index.html, then 404). 

The thing I don't understand is how that's not the case right now? If I have two entry points, index.html and one.html and I point my browser to /one.html it'll serve one.html, otherwise it'll serve index.html (at least, it should). If you specify only one entry point, this code should

mrcoles: Hey @LuudJanssen it’s possible I’m misinterpreting what `bundler.options.publicURL` and `bundler.mainBundle.childBundles` are, but from looking at just this code it seems like the following edge cases could happen:

1. if there are multiple index.html entry points, it seems like this would randomly pick one of them (or more specifically whichever is the first one in `bundler.mainBundle.childBundles`—is that just how they’re ordered in the cli command?)

2. If there are *no* index.html entry points, but there are multiple entry points, then `indexes[0].name` will raise an error, because `indexes[0]` will be `undefined`. (Also for the sake of consistency, if none are found, it should probably return `null` to match the other case where `null` is returned)

Do you think it would make sense to add some additional tests for when there are multiple index.html files and when there are none?

mrcoles: Oh yeah, I am misinterpreting some of this (I was misunderstanding `pathname === bundler.options.publicURL`), but it looks like there are still some edge cases to handle?

devongovett:
It's still unclear to me how we would handle multiple `index.html` files as entry points, e.g. if the entries were `a/index.html and b/index.html`. In that case, neither should be served since there is no `index.html` at the root. You should have to go to `/a` or `/b`. Right now it looks like it would just choose the first one...

awendt:
@devongovett It seems like we've gone from "A much-requested feature is not working at all and requires work-arounds" to "We have a fix that covers all but a few edge cases."

Maybe we can still have this merged, even if it's not perfect?

LuudJanssen:
I'm sorry I didn't find the time yet to work on this issue. I'm hoping to soon find the time to finish the PR.

DeMoorJasper:
Parcel 1 is no longer accepting updates. Feel free to open a PR with improvements to the new dev-server of Parcel 2

ideoclickVanessa:
> Parcel 1 is no longer accepting updates. Feel free to open a PR with improvements to the new dev-server of Parcel 2

Can we expect Parcel 2 to be released within the next few days then?

DeMoorJasper:
@ideoclickVanessa the alpha has been out for a lil while, we’re working hard towards a stable version. Parcel 1 will only receive critical bugfixes in the meantime...

mrcoles:
I wrote a simple proxy script that I run alongside `parcel serve` and have it rewrite URLs to match what I want. Here’s a gist with a comment explaining how to use it: https://gist.github.com/mrcoles/ff979c09719a5e827eb2d122aa2c6332#gistcomment-3074317

I’ve been using it for a while now and it is covering all my needs—also HMR still works just fine too.

RayBB:
For anyone finding this a few years later, parcel v2 doesn't seem to have this problem so upgrading should fix it.

Issue #2155, opened 2018-10-17
Parcel 2: CSS Transform


devongovett:
Create the `@parcel/transform-css` package in `packages/transforms/css`. This package is a Parcel 2  transform plugin for CSS. It will be based on the `CSSAsset` from Parcel 1. The main difference is that it will not be responsible for transforming with PostCSS, that will be a separate `@parcel/transform-postcss` plugin.

The `@parcel/transform-css` package is responsible for the following things:

* Parsing with PostCSS if an AST is not already available from a previous transform.
* Collecting dependencies in the CSS - `@import`, `url()`, etc.
* Generating CSS output

devongovett:
Done. Need to update integration tests.

wbinnssmith:
I was able to unskip all but one or two integration tests. @kwelch is going to pass the remaining two 😃

wbinnssmith:
@kwelch is this good to close now?

kwelch:
Looking at https://github.com/parcel-bundler/parcel/blob/v2/packages/core/integration-tests/test/css.js, it appears all tests are running successfully!

PR #2440, opened 2018-12-20
Parcel 2: Add CSS transformer

padmaia:
CSS transformer plugin for Parcel 2.

See #2155 for details.

### DIFF
--- a/packages/resolvers/default/src/DefaultResolver.js
+++ b/packages/resolvers/default/src/DefaultResolver.js
@@ -12,7 +12,7 @@ import nodeBuiltins from 'node-libs-browser';
 export default new Resolver({
   async resolve(dep: Dependency, cli: CLIOptions, rootDir: string) {
     const resolved = await new NodeResolver({
-      extensions: ['js', 'json'],
+      extensions: ['js', 'json', 'css'],
       cli,
       rootDir
     }).resolve(dep);
--- a/packages/transformers/css/src/CSSTransformer.js
+++ b/packages/transformers/css/src/CSSTransformer.js
@@ -120,8 +120,13 @@ export default new Transformer({
   },
 
   generate(asset) {
-    let code = '';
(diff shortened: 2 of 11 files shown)


packages/transformers/css/src/CSSTransformer.js
```diff
@@ -0,0 +1,130 @@
...
+      }
+
+      // If this came from an inline <style> tag, don't inline the imported file. Replace with the correct URL instead.
+      // TODO: run CSSPackager on inline style tags.
+      // let inlineHTML =
+      //   this.options.rendition && this.options.rendition.inlineHTML;
+      // if (inlineHTML) {
+      //   name.value = addURLDependency(asset, dep, {loc: rule.source.start});
+      //   rule.params = params.toString();
+      // } else {
```
padmaia: Wasn't quite sure how rendition would be populated. Saw it was commented out in the babel transformer, so that's what I did here for now.

packages/transformers/css/src/CSSTransformer.js
```diff
@@ -0,0 +1,130 @@
...
+        }
+      }
+    });
+
+    return [asset];
+  },
+
+  generate(asset) {
+    let code = '';
+    postcss.stringify(asset.ast.program, c => (code += c));
```
devongovett: probably a good idea to check if the ast is dirty here and only re-stringify if needed.

packages/resolvers/default/src/DefaultResolver.js
```diff
@@ -12,7 +12,7 @@ import nodeBuiltins from 'node-libs-browser';
 export default new Resolver({
   async resolve(dep: Dependency, cli: CLIOptions, rootDir: string) {
     const resolved = await new NodeResolver({
-      extensions: ['js', 'json'],
+      extensions: ['js', 'json', 'css'],
```
devongovett: probably should generate this list from plugin config somehow... not in this pr

PR #2445, opened 2018-12-21
Resolve package.browser in subfolders (with backslashes)

### LINKED ISSUE #1883: Is "browser" field supported in node_modules dependencies?
# ❔ Question

Does Parcel handle `browser` fields defined in node_modules the same way browserify and webpack do?

## 🔦 Context

I try to use `ipfs` as a dependency and I found that browser field seem to be ignored by Parcel becaus it doesn't load the browser version of files and doesn't ignore files marked as false.
(for example: https://github.com/ipfs/js-ipfs/blob/master/package.json#L11)

I found that another user have a very similar issue (https://github.com/parcel-bundler/parcel/issues/200#issuecomment-379885444), which makes me thinks that Parcel does not handle this field the same way browserify and webpack do

And I found that another user had an issue with a missconfigured `browser` field in a node_modules dependency (https://github.com/parcel-bundler/parcel/issues/1631) which makes me think that Parcel does handle this field the same way browserify and webpack do

... and I've not found any documentation regarding this field so I'm asking the question

## 🌍 Your Environment

| Parcel           | 1.9.7 |
| Node             | 8.10.0 |
| npm/Yarn         |  yarn 1.9.4 |
| Operating System | Windows 10 64bits |

danmarshall:
# ↪️ Pull Request

Fixes #1883 

A file path in a subfolder on Windows may contain backslashes, which will not match aliases in the [browser field](https://github.com/defunctzombie/package-browser-field-spec) of package.json.


danmarshall:
Aw shucks, it's *mostly* good:
image

Not sure why it failed, other than this:
> WARNING: Scope hoisting tests are disabled on windows due to filesystem errors.

packages/core/parcel-bundler/src/Resolver.js
```diff
@@ -405,6 +405,10 @@ class Resolver {
           }
         }
       }
+      // Or try a lookup replacing backslash characters with forward slash
+      if (alias == null && ~filename.indexOf('\\')) {
+        alias = aliases[filename.replace(/\\/g, '/')];
```
devongovett: Should we use `path.normalize` for these operations?

danmarshall: Using `path.normalize` will remove `.` and `..` and collapse double slashes. That wouldn't fix the bug we're solving here, which is just converting backslashes to forward.

danmarshall:
@DeMoorJasper @devongovett thanks for the reviews. I added one more commit in hopes that it would trigger a new CI build, which it didn't. By adding a commit, GitHub states that I "dismissed" your review .🤔

danmarshall:
Woohoo! it passed 🥇

PR #2344, opened 2018-12-02
Defer throwing asset errors until after dependencies are handled.

MattCheely:
# ↪️ Pull Request

This allows compiled langauges like Elm that handle their own dependency
compilation to set up watchers for dependencies so that hot-reload
continues to work due to errors in new depdenencies. Fixes #2147.

There are also a few tweaks to `ElmAsset` to make sure compilation errors
are thrown after dependency collection

## 💻 Examples

Main.elm:
```elm
module Main exposing (main)

import Html exposing (text)


main =
    text "Hello World"
```

Broken.elm:
```elm
module Broken exposing (anError)

{- This module causes a compiler error -}


anError : String
anError =
    2
```

## 🚨 Test instructions

Using the files above, start the parcel server, and navigate to localhost:1234.

Add a line in `Main.elm` to import `Broken.elm`:
```
import Broken
```

Notice that the build fails

Update `Broken.elm` by changing the value `2` to `"foo"`.

In current parcel: no rebuild is triggered
In this branch: the error is resolved

## ✔️ PR Todo

If anybody knows of a good unit test strategy for this, I'm open to suggestions. 

- [ ] Added/updated unit tests for this change
- [X] Filled out test instructions (In case there aren't any unit tests)
- [X] Included links to related issues/PRs

PR #2468, opened 2018-12-26
Defer throwing asset errors until after dependencies are handled.

domenkozar:
Testing https://github.com/parcel-bundler/parcel/pull/2344 with following changes:

- removed whitespace
- rebased on top of master

(lost original authorship of the commit, hopefully we can fix this before merging)

### DIFF
--- a/packages/core/parcel-bundler/src/Bundler.js
+++ b/packages/core/parcel-bundler/src/Bundler.js
@@ -593,6 +593,13 @@ class Bundler extends EventEmitter {
       })
     );
 
+    // If there was a processing error, re-throw now that we've set up
+    // depdenency watchers. This keeps hot-reloading working if there is an
+    // error in a dependency not directly handled by Parcel.
+    if (processed.error !== null) {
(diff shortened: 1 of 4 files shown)


domenkozar:
@MattCheely I still get the same bug, why do you think elm-hot changed anything?

MattCheely:
@domenkozar :man_facepalming: Looks like you are right and it is still an issue. I was working on this at a meetup last night and during testing with a local up-to-date parcel install, it seemed like things were working, but reviewing things now, it must have made a mistake somewhere in that verification.

I can look at this some more tonight - it seems the latest changes in my branch are also no longer fixing the issue, but until I look closer, I'm not sure whether I introduced the problem when fixing test failures or it's related to the elm-hot changes.  Feel free to take a look in the meantime if you have the time and inclination.

MattCheely:
@domenkozar I think I have resolved all of my previous issues in  #2475. I think we can close this PR, if that's OK with you.

PR #2472, opened 2018-12-27
Generate source map files with long extentions (e.g. .js.map)

### LINKED ISSUE #2460: Parcel does not keep file extension when creating source maps
# 🐛 bug report

When using the following command, the source maps generated by parcel doesn't keep the file extension which makes debugging impossible using VSCode.

`parcel build src/index.ts --target node`

Parcel generates two file `index.js` and `index.map`

The reason why I know this is indeed the problem with my debugger is because if I rename `index.map > index.js.map` and change the last line of `index.js` to `//# sourceMappingURL=/index.js,map`, then the debugger works like a charm

```js
// ts config
{
  "compilerOptions": {
    "target": "es2015",
    "module": "commonjs",
    "moduleResolution": "node",
    "strict": true,
    "esModuleInterop": true,
    "skipLibCheck": true,
    "outDir": "dist",
    "lib": ["es2015"],
    "jsx": "react",
    "sourceMap": true
  },
  "include": ["**/*.ts", "**/*.tsx"],
  "exclude": ["**/*.test.ts", "**/*.test.tsx"]
}

// vscode launch.json
{
  // Use IntelliSense to learn about possible attributes.
  // Hover to view descriptions of existing attributes.
  // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
  "version": "0.2.0",
  "configurations": [
    {
      "type": "node",
      "request": "launch",
      "name": "Launch via NPM",
      "runtimeExecutable": "npm",
      "runtimeArgs": ["run-script", "debug"],
      "port": 9229,
      "sourceMaps": true,
      "outFiles": ["${workspaceFolder}/bin/**/*.js"]
    }
  ]
}

```

## 🤔 Expected Behavior

I should be able to use my launch config from VScode to be able to debug my NodeJS typescript program

## 😯 Current Behavior

VSCode is not able to connect to the typescript files break points because it probably uses a pattern to match only js file extension even for the source map.

## 💁 Possible Solution

Add a cli option to keep file extension when building source maps OR always include them since I don't see why they were removed in the first place 

## 🔦 Context

I'm building a project with apollo server + typescript + serverless with Parcel to have a blasing fast bundler and minimum payload

## 💻 Code Sample

n/a

## 🌍 Your Environment

| Parcel           | ^1.11.0
| Node             | 8.11.2
| npm/Yarn         | 6.5.0 (npm)
| Operating System | Mac OS high sierra

kirillrogovoy:
# ↪️ Pull Request

Closes #2460.

What I do here is preserve the original file extension before appending `.map` to the source map files.

Instead of `index.map` it would generate `index.js.map`. According to the [spec](https://sourcemaps.info/spec.html) it's the recommended convention. It also solves a real issue for debugging Javascript in VSCode.

While it solves the issue, I don't like the tech debt I introduce. Specifically, I don't like the conditionals about specific types of bundles in the otherwise abstract code.

@DeMoorJasper please, take a look and let me know if you can think of a more pleasing solution architecture-wise. This is my first peek at the internals of Parcel, so I don't have the whole context yet.

## ✔️ PR Todo

### DIFF
--- a/packages/core/integration-tests/test/css.js
+++ b/packages/core/integration-tests/test/css.js
@@ -12,7 +12,7 @@ describe('css', function() {
       assets: ['index.js', 'index.css', 'local.js', 'local.css'],
       childBundles: [
         {
-          name: 'index.map'
+          name: 'index.js.map'
         },
         {
           name: 'index.css',
@@ -88,7 +88,7 @@ describe('css', function() {
           childBundles: []
         },
         {
-          name: 'index.map',
+          name: 'index.js.map',
           type: 'map'
         }
       ]
--- a/packages/core/integration-tests/test/sourcemaps.js
+++ b/packages/core/integration-tests/test/sourcemaps.js
@@ -14,7 +14,7 @@ describe('sourcemaps', function() {
       assets: ['index.js'],
       childBundles: [
         {
-          name: 'index.map',
+          name: 'index.js.map',
           type: 'map'
         }
       ]
@@ -24,7 +24,7 @@ describe('sourcemaps', function() {
       path.join(__dirname, '/dist/index.js')
     )).toString();
     let map = (await fs.readFile(
-      path.join(__dirname, '/dist/index.map')
+      path.join(__dirname, '/dist/index.js.map')
     )).toString();
     mapValidator(raw, map);
     let mapObject = JSON.parse(map);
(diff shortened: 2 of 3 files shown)


DeMoorJasper:
I think there was no nice way of solving this. It works and it's not that bad, so I guess it's good enough

PR #2473, opened 2018-12-27
Update README.md

def14nt:
This offends my feminism

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ or with npm:
 npm install -g parcel-bundler
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -36,7 +36,7 @@ or with npm:
 npm install -g parcel-bundler
 ```
 
-2.  Parcel can take any type of file as an entry point, but an HTML or JavaScript file is a good place to start. If you link your main JavaScript file in the HTML using a relative path, Parcel will also process it for you, and replace the reference with a URL to the output file.
+2.  Parcel can take any type of file as an entry point, but a HTML or JavaScript file is a good place to start. If you link your main JavaScript file in the HTML using a relative path, Parcel will also process it for you, and replace the reference with a URL to the output file.
```
devongovett: This is incorrect. See https://english.stackexchange.com/questions/25742/is-it-supposed-to-be-a-html-or-an-html

PR #2475, opened 2018-12-28
Defer throwing asset errors until after dependencies are handled.

### LINKED ISSUE #2147: Elm: importing
# 🐛 bug report

In file B, importing a new file A that doesn't initially typecheck doesn't trigger reloads on subsequent reloads.

Barebones elm project.

## 🤔 Expected Behavior

When file A is saved, parcel reloads.

## 😯 Current Behavior

Nothing happens when saving file A, one needs to save file B.

## 🌍 Your Environment

| Parcel           | 1.10.3
| Operating System | Linux

MattCheely:
# ↪️ Pull Request

Previously submitted in #2468 and #2344 (sorry about the confusion)

This allows compiled langauges like Elm that handle their own dependency
compilation to set up watchers for dependencies so that hot-reload
continues to work due to errors in new depdenencies. Fixes #2147.

There are also a few tweaks to `ElmAsset` to make sure compilation errors
are thrown after dependency collection

## 💻 Examples

Main.elm:
```elm
module Main exposing (main)

import Html exposing (text)


main =
    text "Hello World"
```

Broken.elm:
```elm
module Broken exposing (anError)

{- This module causes a compiler error -}


anError : String
anError =
    2
```

## 🚨 Test instructions

Using the files above, start the parcel server, and navigate to localhost:1234.

Add a line in `Main.elm` to import `Broken.elm`:
```
import Broken
```

Notice that the build fails

Update `Broken.elm` by changing the value `2` to `"foo"`.

In current parcel: no rebuild is triggered
In this branch: the error is resolved

## ✔️ PR Todo

I will try to create a test for this in the hmr integration tests, now that I am more familiar with the tests.

### DIFF
--- a/packages/core/parcel-bundler/src/assets/ElmAsset.js
+++ b/packages/core/parcel-bundler/src/assets/ElmAsset.js
@@ -133,7 +133,16 @@ class ElmAsset extends Asset {
   }
 
   generateErrorMessage(err) {
-    return err.message;
+    // For some reason, if not converted to a plain object,
+    // the error message is lost somewhere between Pipeline.js
+    // and Bundler.js.
+
+    // The stack is not particularly useful, but other code may
+    // expect it and try to print it.
+    return {
+      message: err.message,
+      stack: ''
+    };
   }
 }
 
--- a/packages/core/parcel-bundler/src/assets/ElmAsset.js
+++ b/packages/core/parcel-bundler/src/assets/ElmAsset.js
@@ -44,12 +44,7 @@ class ElmAsset extends Asset {
       options.optimize = true;
     }
 
-    let compiled = await this.elm.compileToString(this.name, options);
-    this.contents = compiled.toString();
-    if (this.options.hmr) {
-      let {inject} = await localRequire('elm-hot', this.name);
-      this.contents = inject(this.contents);
-    }
+    this.elmOpts = options;
   }
 
   async collectDependencies() {
@@ -76,6 +71,13 @@ class ElmAsset extends Asset {
   }
 
   async generate() {
+    let compiled = await this.elm.compileToString(this.name, this.elmOpts);
+    this.contents = compiled.toString();
+    if (this.options.hmr) {
+      let {inject} = await localRequire('elm-hot', this.name);
+      this.contents = inject(this.contents);
+    }
+
     let output = this.contents;
 
     if (this.options.minify) {
@@ -128,6 +130,10 @@ class ElmAsset extends Asset {
 
       return result.code;
     }
+  }
+
+  generateErrorMessage(err) {
+    return err.message;
   }
 }
 
--- a/packages/core/parcel-bundler/src/Bundler.js
+++ b/packages/core/parcel-bundler/src/Bundler.js
@@ -593,6 +593,13 @@ class Bundler extends EventEmitter {
       })
     );
 
+    // If there was a processing error, re-throw now that we've set up
+    // depdenency watchers. This keeps reloading working if there is an
+    // error in a dependency not directly handled by Parcel.
+    if (processed.error !== null) {
+      throw processed.error;
+    }
+
     // Store resolved assets in their original order
     dependencies.forEach((dep, i) => {
       asset.dependencies.set(dep.name, dep);
--- a/packages/core/parcel-bundler/src/Pipeline.js
+++ b/packages/core/parcel-bundler/src/Pipeline.js
@@ -17,16 +17,23 @@ class Pipeline {
     }
 
     let asset = this.parser.getAsset(path, options);
-    let generated = await this.processAsset(asset);
+    let error = null;
     let generatedMap = {};
-    for (let rendition of generated) {
-      generatedMap[rendition.type] = rendition.value;
+    try {
+      let generated = await this.processAsset(asset);
+      for (let rendition of generated) {
+        generatedMap[rendition.type] = rendition.value;
+      }
+    } catch (err) {
+      error = err;
+      error.fileName = path;
     }
 
     return {
       id: asset.id,
       dependencies: Array.from(asset.dependencies.values()),
       generated: generatedMap,
+      error: error,
       hash: asset.hash,
       cacheData: asset.cacheData
     };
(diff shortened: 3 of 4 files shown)


packages/core/parcel-bundler/src/assets/ElmAsset.js
```diff
@@ -129,6 +131,19 @@ class ElmAsset extends Asset {
       return result.code;
     }
   }
+
+  generateErrorMessage(err) {
+    // For some reason, if not converted to a plain object,
+    // the error message is lost somewhere between Pipeline.js
+    // and Bundler.js.
```
MattCheely: This fixes an issue for Elm, but I probably also need to do something like this in `Pipeline.js` for other asset types that throw `Error`s.  I'm not certain why this is necessary, but I suspect that there's an object copy somewhere in `WorkerFarm` that causes data to be lost if `Pipeline.process` returns an obejct containing an `Error`

domenkozar:
@mattcheely tests still seem to fail (although they were fixed yesterday).

MattCheely:
@domenkozar Tests have been working for me locally for the most part, but I have noticed occasional timeouts on the hmr tests (mine and others), which seems to be part of the problem here. I don't have much insight into what's causing it, but there are a couple of things I can do to try simplifying the tests.

domenkozar:
@DeMoorJasper is that a known issue?

packages/core/parcel-bundler/src/assets/ElmAsset.js
MattCheely: This is fixed.

DeMoorJasper:
@domenkozar it only appears to be a consistent issue with this branch. But it’s a known issue that the tests are no longer stable.

MattCheely:
@DeMoorJasper I tidied up my hmr tests a bit and the last set of failures in the test run don't seem to be related to my changes. Let me know if there's something I should do to resolve that. 

FWIW, one thing I have noticed is that I see hmr tests timeout more frequently when running them in isolation via: `yarn run mocha --grep hmr`.  Usually it's the electron test that times out. When I run the full test suite, I don't think I've seen it fail, so there may be some cross-test interaction or subtle timing issues at play.

DeMoorJasper:
@MattCheely Looks like you fixed it if I look at the CI results, not sure why the babel autoinstall is failing but that's probably another issue unrelated to this PR

MattCheely:
@domenkozar to test in your project locally, run `yarn build` at the top of the parcel repo, then in `packages/core/parcel-bundler` run `yarn link`, which will make that folder available for linking in other projects.  Then in the project you want to test with, run `yarn link parcel-bundler`, which will create a symlink from `node_modules/parcel-bundler` in that project to your local checkout.

domenkozar:
Thanks @MattCheely - it does fix the issue. I wonder if error messages are fixed though, I still see two of them:

```
Server running at http://localhost:12345 
🚨  /home/ielectric/dev/monorepo/frontend/src/Main.elm: Compilation failed
[================                                  ] - 1 / 3-- PARSE ERROR ----------------------------------------------------- src/Foo.elm

Something went wrong while parsing bar's definition.

4| bar = 
5| 
   ^
I was expecting to see the rest of bar's definition. Maybe you forgot some code?
Or maybe the body of `bar` needs to be indented?
Detected errors in 1 module.                                         

[================                                  ] - 1 / 3-- PARSE ERROR ----------------------------------------------------- src/Foo.elm

Something went wrong while parsing bar's definition.

4| bar =
5|
   ^
I was expecting to see the rest of bar's definition. Maybe you forgot some code?
Or maybe the body of `bar` needs to be indented?
Detected errors in 1 module.

    at ChildProcess.<anonymous> (/home/ielectric/dev/monorepo/frontend/node_modules/node-elm-compiler/index.js:149:27)
    at emitTwo (events.js:126:13)
    at ChildProcess.emit (events.js:214:7)
    at maybeClose (internal/child_process.js:925:16)
    at Process.ChildProcess._handle.onexit (internal/child_process.js:209:5)
```

domenkozar:
My thinking is that due to https://github.com/rtfeldman/node-elm-compiler/blob/master/src/index.ts#L126 we can't get elm not to report the error, so we could only tell parcel to skip it.

domenkozar:
@MattCheely I think there was a typo, now it works correctly with:

```
diff --git a/packages/core/parcel-bundler/src/assets/ElmAsset.js b/packages/core/parcel-bundler/src/assets/ElmAsset.js
index 5a0f4cc..447e9a5 100644
--- a/packages/core/parcel-bundler/src/assets/ElmAsset.js
+++ b/packages/core/parcel-bundler/src/assets/ElmAsset.js
@@ -135,8 +135,8 @@ class ElmAsset extends Asset {
   generateErrorMessage(err) {
     // The generated stack is not useful, but other code may
     // expect it and try to print it, so make it an empty string.
-    err.stack == '';
-    return err;
+    err.stack = '';
+    return err
   }
 }
```

domenkozar:
Happy new year :tada: thanks for all of the work with above patch it works great!

MattCheely:
@domenkozar I will see if I can get that working again. It should be possible, at least for the parcel browser overlay. I probably accidentally reverted my previous fix when I was changing the way errors are handled in `Pipeline`. If I can get it fixed, I'll add a test for it so It doesn't accidentally regress again.

domenkozar:
Do tests run in parallel? There are some elm corruption errors that are known issue if you mess around with cache or run things in parallel. https://github.com/elm/compiler/issues/1845

Maybe doing `rm -rf elm-stuff` before each test could help (or separate working directories).

domenkozar:
Ah actually that's probably an old annotation. https://dev.azure.com/devongovett/devongovett/_build/results?buildId=627 shows only the known transient error.

@DeMoorJasper can we merge this?

DeMoorJasper:
@domenkozar unfortunately only Devon can merge PRs with failing tests

domenkozar:
seems to pass now :)

domenkozar:
@DeMoorJasper do you have an idea if Parcel bugfix release will happen anytime soon? So I either look into setting up using Parcel master or just wait for bugfix release. Thanks!

domenkozar:
I uploaded parcel-bundler-fork based on the current master to unblock my workflow. I plan to delete it as soon as 1.11.1 is out - thank you all again!

Issue #1562, opened 2018-06-16
Proxy support?


smithb1994:
I currently use [http-server](https://github.com/indexzero/http-server) to serve my SPA app on localhost:8080, but I set a proxy to localhost:3000 where I have my express server running with a REST API.  
  
With http-server the -P flag proxies all requests which can't be resolved locally to a given URL. 

For example, my serve command is:  
`http-server dist -c 1 -a localhost -P http://localhost:3000/`  
   
Is there an equivalent proxy flag for parcel serve?

KeineLimonade:
Currently not. But you can use parcel as middleware in your express application

zaynv:
But what if you're using Koa? I don't think it would work :(

jamiebuilds:
https://github.com/vkurchatkin/koa-connect

cdaringe:
https://www.npmjs.com/package/koa-parcel-middleware

ifrontend-xyz:
http://npm.taobao.org/package/parcel-proxy-server

Issue #55, opened 2017-12-06
Proxy


webular:
Is there a way to define a proxy?  In webpack we are doing something like this: 

```
devServer: {
  proxy: {
   '/api*' :{
      target: 'http://localhost:3001/graphql',
      secure: false,
    },
}
```

[More details about webpack's devserver proxy here.](https://webpack.js.org/configuration/dev-server/#devserver-proxy)

devongovett:
I don't really think this should be the responsibility of the bundler... Can you just enable CORS and hit localhost:3001 directly?

Alternatively, you could build your own small server with express and use parcel as a middleware.

```javascript
let Bundler = require('parcel-bundler');
let express = require('express');

let bundle = new Bundler('entry.html');
let app = express();

// define proxy routes here

app.use(bundle.middleware());
```

tb:
@devongovett I always use [create-react-app proxy](https://github.com/facebookincubator/create-react-app/blob/master/packages/react-scripts/template/README.md#proxying-api-requests-in-development) This allows me to use relative urls like `/api/endpoint` / `/graphql`. The same urls work then after deployment to production server.

webular:
@tb Oh - thats exactly what I want.  Thx for the tip!

jylin:
It's not the responsibility of the bundler, but it would be a very useful feature for the dev server. If your server is not in NodeJS (and you don't want to enable CORS), then without this you have to use some kind of post-processing to replace "locaholhost:PORT" with something else when deploying to production. It's much nicer to be able to use relative links and have the dev server have a config to proxy to the appropriate backend endpoint.

chasegiunta:
For the dev-server, I think this would be useful for non-spa sites as well (php applications, etc). Where you proxy the app (`myphpsite.test`), but watch the assets and/or templates directory for changes. Templates directory changes would trigger full reloads, while asset changes trigger HMR (in a perfect world)

samithaf:
Hey @webular , you can probably use **browser-sync** and **http-proxy-middleware** to achieve your end goal and it's relatively easy to set it up. E.g. 

 ```
import assign from 'lodash/assign';
import Bundler from 'parcel-bundler';
import serve from 'browser-sync';
import proxy from 'http-proxy-middleware';
import compress from 'compression';

const bundler = new Bundler('index.html');
 serve({
    port: process.env.PORT || 3000,
    open: false,
    server: { baseDir: 'dist', https: true },
    middleware: [
[...]

PR #2477, opened 2018-12-28
Allows proxying paths of your application to different host

arcturus:
# ↪️ Pull Request
This PR allows to pass extra parameters to the dev server to allow a specified path being proxied to a different host.

This is useful for development purposes, like proxy api calls (production, staging or testing) without having to enable CORS in a service that won't need to have CORS.

It looks for a key named 'proxy' in the package.json and assumes that is the hostname to proxy any calls to '/api/*'.

If instead of a string you provide an object with the keys 'path' and 'host' it will use those parameters to proxy.

## 💻 Examples
This feature has been requested before, #1562 or [spectrum](https://spectrum.chat/parcel?thread=b2982c28-7125-4c23-9eb2-35850f29fd67). Previously has been closed assuming that parcel can be used as a middleware for express. I end up writing this because my backend is in rust with [rocket.rs](https://rocket.rs/).

Also, planning to use in conjunction with [https://fakerql.com/](https://fakerql.com/) to build the frontend even before I write the backend supporting graphql.

## 🚨 Test instructions

The typical usage will be like:
```
   // package.json of the app using parcel
   "proxy": "https://<host to proxy requests>", // All requests to http://localhost:1234/api/* will be proxied
```

Using a public http echo [service like](http://scooterlabs.com):
```
 // package.json
 "proxy": "http://scooterlabs.com/echo",
```

Then you can access `http://localhost:1234/api/?echo=1`

```
 // package.json
 "proxy" : {
    "path": "service",
    "host": "http://scooterlabs.com/echo"
 },
```
Then you can access `http://localhost:1234/service/?echo=1`

## ✔️ PR Todo

Will need some help from maintainers to guide me on how to unit test this. Was planning to mock the module used for doing the http-proxy but didn't see anything like this in the unit tests.


devongovett:
What do you think about configuring this in package.json rather than a CLI flag similar to how create-react-app does it? https://facebook.github.io/create-react-app/docs/proxying-api-requests-in-development#docsNav

arcturus:
That's a fantastic idea, let me give it a try!

arcturus:
@devongovett I've seen the proposed solution and looks feasible and very clean.

I've realised that with the `proxy` option in the package.json forces you to proxy the `/api` path which is very useful. What about the following idea, if `proxy` is a string, that's the host that will proxy `/api/*`. If `proxy` option is an object, we can expect path and host as a way of fine tuning how do you want to proxy.

Sounds good to you?

domenkozar:
Would it be hard to support passing `--proxy-path` multiple times?

arcturus:
@devongovett I've completely reworked the PR following the philosophy of check the package.json.

arcturus:
> Would it be hard to support passing `--proxy-path` multiple times?

Just did this first version with 1 single path to simplify the landing, if there is an appetite for supporting multiple paths we can add them easily.

packages/core/parcel-bundler/src/utils/ProxyHelper.js
```diff
@@ -0,0 +1,62 @@
...
+ * 'proxyHost': the host to proxy the requests
+ * 'proxyPath': path to match requests to be proxied.
+ *
+ * @returns A proxy object if the configuration is correct, null
+ *          otherwise.
+ */
+function ProxyHelper() {
+  // Try to locate first the package.json of the running app
+  const appDirectory = fs.realpathSync(process.cwd());
+  const pkg = path.resolve(appDirectory, 'package.json');
```
mischnic: I think you should use `Resolver.getPackager` here instead (an instance variable on the bundler in `Server.js` `middleware()`)

https://github.com/parcel-bundler/parcel/blob/6cbf2a65cdcd66af135ff9eca2a3799bde9f8a5b/packages/core/parcel-bundler/src/Asset.js#L131

packages/core/parcel-bundler/src/utils/ProxyHelper.js
```diff
@@ -0,0 +1,62 @@
...
+ * Try to locate the package.json file of the app executing this.
+ * From that packaje.json reads an attribute 'proxy'.
+ *
+ * 'proxy' accepts a string or an object for configuring a proxy service.
+ *
+ * In case that a string is passed, will use it as a host to proxy
+ * and any request to '/api*' will be proxied.
+ *
+ * If an object is passed will lock for 2 keys:
+ * 'proxyHost': the host to proxy the requests
```
arcturus: nit to myself, the documentation is not accurate, the keys are `path` and `host`

timpur:
@arcturus, any progress on this PR ? This would be nice to use :)

arcturus:
> @arcturus, any progress on this PR ? This would be nice to use :)

Waiting for @mischnic feedback about going async or not.

packages/core/parcel-bundler/src/utils/ProxyHelper.js
mischnic: >  Another option is to configure the proxy on the `serve` function that is already async,  would you prefer to go that way?

That would mean that when you use Parcel's middleware via the API, the requests won't be proxied.
On the other hand, the current implementation using `process.cwd()` is certainly not correct, because the package file resolution algorithm used by the other code is: "go up in the folder structure until there is a package.json file", you'd want to do this for the entry point: `bundler.entryPoints[0]`.

The only solution I see currently is reimplementing this code to be synchronous (eventhough synchronous fs calls should really be avoided):
https://github.com/parcel-bundler/parcel/blob/c670decf1e8a039a80a47395bf916643d1955613/packages/core/parcel-bundler/src/Resolver.js#L422-L434

devinrhode2:
This would be awesome to have!

cdll:
merge failed?

packages/core/parcel-bundler/src/utils/ProxyHelper.js
arcturus: What about a very simple approach, using `process.cwd()` instead of using the bundler. We try to find the package.json in your project defining how the proxy should work. 
A simple sync path check will help to understand if we should proceed reading the package.json file.

I understand that `process.cwd()` will produce different results depending on where do you execute your process, but do you think is a good tradeoff to ask people to start their dev server from their working directory?

gamedevsam: No, I launch mine from `src/server`

jurginauskaite:
Comment

devcshort:
Still being worked on? Would love to have this functionality for current and future projects.

eekboom:
Please support multiple path/host entries. Unfortunately, we have multiple backend paths (like /api and /apiv2),

swillis12:
Any news on this? Looks like many people are hungry for this feature to get merged 💃

rametta:
Also curious about the progress of this 👀

telmotrooper:
Hope this gets done soon! 😀️

packages/core/parcel-bundler/src/utils/ProxyHelper.js
```diff
@@ -0,0 +1,62 @@
...
+
+class Proxy {
+  constructor(host, path) {
+    this.host = host;
+    this.path = path;
+    this._proxy = httpProxy.createProxyServer();
+  }
+
+  shouldProxyPath(path) {
+    return path.startsWith(this.path);
```
Kogia-sima: Should return false when `this.path[path.length] !== '/'`.

For example:

- this.path: 'service'
- path: 'service2/get'

=> should return false.

Refer the implementation of [connect](https://github.com/senchalabs/connect/blob/master/index.js#L164-L168)

dlight:
What's the preferred workaround until this gets merged? Is it using Express, as suggested in #55?

DeMoorJasper:
Wil be supported in v2

mmunte-impeo:
Very disappointing this gets closed with will be supported in V2.

fabien-h:
I'm using v2. Is it supported now ? (I cannot find it in the documentation)

DeMoorJasper:
@fabien-h yes it is supported in the current alpha of parcel 2

fabien-h:
Hum, sorry for polluting this thread. I'm on ^2.0.0-alpha.3.2

I created a .proxyrc file with:

```
{
  "/api": {
    "target": "https://localhost:3030",
    "secure": true
  }
}
```

And nothing is redirected. I've also tried a .proxyrc.js file, and the implementation in the package.json.

I placed the files in the /src folder and in the root folder (where my package.json is).

Is there a functionning example I can see ?

DeMoorJasper:
@fabien-h You can have a look at the PR, there should be tests which you can see as examples #3281

Kogia-sima:
@fabien-h How did you get the server certificate? If you are using non-verified certificate or key (e.g. ones generated on your local machine), you have to set `secure` option `false`.

xiemeilong:
I am using 2.0.0-alpha.3.2 too， the  proxy  is not working

mischnic:
@xiemeilong Does https://parcel2-docs.now.sh/recipes/api-proxy/ help?

xiemeilong:
@mischnic  It works when using  2.0.0-alpha.3.1, but 2.0.0-alpha.3.2 not.

mischnic:
@xiemeilong Works for me: https://github.com/mischnic/parcel-proxy-example

xiemeilong:
@mischnic  The proxy  works after I delete node_modules and reinstall. 
But no `[HPM] Proxy created: /api  ->  http://localhost:5000/` printed as your example.

PR #2484, opened 2018-12-31
Parcel 2: Get some JS integration tests passing

devongovett:
This gets some integration tests for the JS transformer pipeline passing against v2, including babel, fs inlining, node global insertion, environment variable substitution, etc. 

**Still skipped:**

* Dynamic imports (need to implement loaders)
* Production minification tests (need to normalize parcel options)
* Electron (need to figure out how it should really work)
* Autoinstall
* Non-JS imports (e.g. JSON, YAML, ...)

**Test changes:**

* Replace `assertBundleTree` with `assertBundles`, which asserts against a BundleGraph.
* Disables Node 6 in CI
* Move Babel tests to their own file
* Replace `--target` cli option with targets in package.json. This caused a few test fixtures to be duplicated in order to change the package.json target.
* Fix all flow and ESLint errors

**Core changes:**

* Use shared instance of WorkerFarm instead of starting up a new one each time. This makes running the tests 10x faster.
* Creates the cache directory on the main process before bundling starts rather than in each cache instance. Otherwise there are race conditions when multiple workers try to create it at the same time.
* Computes the root directory the same way as in Parcel 1: by looking at the common dir between the entries. This is necessary to get proper output filenames from the default namer plugin.
* Load environment variables from `.env` files.
* Support optional dependencies, e.g. `require` within a `try...catch` block. If they cannot be resolved, they are ignored rather than erroring.
* Load `.browserslistrc` files in TargetResolver along with package.json
* Add option to disable reading from the cache. This is necessary for the tests so we don't need to delete the `.parcel-cache` directory before each test. Unlike Parcel 1, we still write to the cache even with `--no-cache`, because we don't store the file contents of every asset in memory anymore (the packagers read from the cache).
* Fix PARCEL_WORKERS=0 environment variable to disable running any workers and only use the main process, typically for debugging.
* Add support for an `isURL` flag on dependencies, which disables resolution of node_modules for relative paths without a `./` prefix. This is typically seen outside of JS, e.g. CSS and HTML, along with some APIs in JS like web workers and service workers. Not totally sure about the naming here, but we can come back to it.


padmaia:
Any clue why the tests are failing on Windows? Other than that and the conflicts it LGTM.

mischnic:
> Any clue why the tests are failing on Windows? Other than that and the conflicts it LGTM.

```
EBUSY: resource busy or locked, rmdir 'D:\a\1\s\packages\core\parcel-bundler\test\input'
```

Maybe multiple tests run in parallel and one test's `beforeEach` tries to delete that folder while another is using it.

PR #2489, opened 2019-01-01
CSS/SASS/LESS sourcemaps

### LINKED ISSUE #903: Enable/disable node-sass sourcemaps

### LINKED ISSUE #1824: Implement sourcemaps for CSS
# 🙋‍♀️ Feature request

Hello, I have some problem with **parcel-bundler@1.9.7**, - there is no *sourcemap* emitted at all for generated CSS with *zero config*. Or, with *some* configuration for **postcss**, it's possible to make some  **sourcemaps** to be *inlined* in same generated CSS, but they contains *incorrect* path to source file, and can't be used in **Chrome Dev Tools**.

I've tried with no config at all, and with *all combination* of next flags in **package.json**
```js
  "postcss": {
    "map": {
      "inline": true,
      "sourcesContent": false,
      "from": "./src"
    },
    "plugins": {
      "autoprefixer": {}
    }
  },
  "stylus": {
    "sourcemap": {
      "comment": false,
      "inline": false,
      "basePath": "./src"
    }
  }
```

## 🤔 Expected Behavior

Expects that somehow user can get **sourcemaps** for CSS that will be fine used in Chrome Dev Tools to see what file contains what style

## 😯 Current Behavior

When I build a simple html file with linked *Stylus* file (at glance I think that for any preprocessor dialect) and have zero configuration, then **no** *sourcemaps* emitted.

With *some configuration* in **package.json** for **postcss** block, it's possible to make some *sourcemaps* to be *inlined* in CSS *(it's postcss stuff really, not a parcel one)*, but their contains incorrect path to source files

## 💁 Possible Solution

Let me think... )
## 🔦 Context

My project is *VueJS* with vue components, where for styling I use *Stylus* and for templates the *Pug* being used, with TypeScript, that majestically rises above all stuff...

But in simple configuration one html + one stylus file it's the same behavior
## 💻 Code Sample
https://gist.github.com/ocoka/918fad4442f49714cfd69a85f67e61f1

## 🌍 Your Environment

| Parcel           | 1.97
| Node             | 8.11.3
| npm/Yarn         |  6.2.0
| Operating System | Ubuntu 16.04

### LINKED ISSUE #2194: linked node_modules source maps are mixed w/ project
# 🐛 bug report

I am trying to include sourcemaps for a dependency that has been linked. `npm link mydep`

This is working with the caveat that the sourcemaps appear under the same root `src/` in chrome devtools. This is a problem b/c the linked node_module sourcemaps appear mixed with my applications sourcemaps and will be overwritten by my app if any conflicts exist.

possibly related to #1801?

```js
{
  "presets": [
    [
      "@babel/preset-env",
      {
        "modules": false,
        "targets": {
          "browsers": [
            "> 1%",
            "last 3 versions",
            "ie >= 9",
            "ios >= 8",
            "android >= 4.2"
          ]
        },
        "useBuiltIns": false
      }
    ]
  ],
  "plugins": [
    ["styled-components", { "displayName": true }],
  ]
}
```

## 🤔 Expected Behavior

I would expect any sourcemaps found for node_modules would be served at a different root. See next section for an example

## 😯 Current Behavior

currently they are served from the same `/src` root.

ex.

app
```
-- src
  -- index.js
```

mydep
```
-- src
  -- index.js
  -- cache
    -- index.js
```

The source maps in chrome devtools will appear as:

```
-- src
  -- index.js (this is from app)
  -- cache
    -- index.js (this is from mydep)
```

I would expect them to appear as:
The source maps in chrome devtools will appear as:

```
-- mydep (or even as a sub dir of node_modules)
  -- index.js
  -- cache
    -- index.js
-- src
  -- index.js
```
## 🔦 Context

Being able to correctly view sourcemaps for linked dependencies is useful when attempting to debug a package.

## 💻 Code Sample

just link any node_module w/ built source maps to your app

## 🌍 Your Environment

| Parcel           | 1.10.1
| Node             | v8.11.3
| npm              | 6.4.1
| Operating System | macos

### LINKED ISSUE #2692: `--no-source-maps` doesn't work if source maps are cached
# 🐛 bug report

Continuation of #2079 

> I am still seeing this issue on parcel-bundler 1.11.0.
> 
> 1. If I build with source maps, then `rm -r dist/` and build with `--no-source-maps`, then I still get source maps in `dist/`
> 2. If I then `rm -r .cache/ dist/` to remove both the cache and dist and build again with  `--no-source-maps`, then I get it without source maps
> 
> ```shell
> ./node_modules/.bin/parcel --version
> 1.11.0
> ```

by @mrcoles

```shell
parcel build index.js --no-source-maps
```

## 🤔 Expected Behavior

should not package sourcemaps

## 😯 Current Behavior

Packages cached sourcemaps

## 💁 Possible Solution

check if sourcemaps should be made in packaging

## 🌍 Your Environment

| Parcel           | 1.11
| Node             | 10
| npm/Yarn         | latest
| Operating System | OS X

mischnic:
# ↪️ Pull Request

Proper (non inline) sourcemaps for PostCSS. [Example](http://sokra.github.io/source-map-visualization/#base64,ZGl2e3dpZHRoOjEwMHB4O2hlaWdodDoxMDBweH1tYWlue2ZvbnQtZmFtaWx5Om1vbm9zcGFjZX1ib2R5e2JhY2tncm91bmQtY29sb3I6cmVkfQovKiMgc291cmNlTWFwcGluZ1VSTD0vc3R5bGUuY3NzLm1hcCov,eyJ2ZXJzaW9uIjozLCJzb3VyY2VzIjpbIm90aGVyLXN0eWxlLmNzcyIsImFub3RoZXItc3R5bGUuY3NzIiwic3R5bGUuY3NzIl0sIm5hbWVzIjpbXSwibWFwcGluZ3MiOiJBQUNBLElBQ0MsWUFDQSxhQ0hELEtBQ0Msc0JDRUQsS0FDQyIsImZpbGUiOiJzdHlsZS5jc3MubWFwIiwic291cmNlUm9vdCI6Ii4uL2ludGVncmF0aW9uL3NvdXJjZW1hcC1jc3MtaW1wb3J0Iiwic291cmNlc0NvbnRlbnQiOlsiXG5kaXYge1xuXHR3aWR0aDogMTAwcHg7XG5cdGhlaWdodDogMTAwcHg7XG59XG4iLCJtYWluIHtcblx0Zm9udC1mYW1pbHk6IG1vbm9zcGFjZTtcbn1cbiIsIlxuQGltcG9ydCBcIi4vb3RoZXItc3R5bGUuY3NzXCI7XG5cbmJvZHkge1xuXHRiYWNrZ3JvdW5kLWNvbG9yOiByZWQ7XG59XG5cbkBpbXBvcnQgXCIuL2Fub3RoZXItc3R5bGUuY3NzXCI7XG4iXX0=,CmRpdiB7Cgl3aWR0aDogMTAwcHg7CgloZWlnaHQ6IDEwMHB4Owp9Cg==,bWFpbiB7Cglmb250LWZhbWlseTogbW9ub3NwYWNlOwp9Cg==,CkBpbXBvcnQgIi4vb3RoZXItc3R5bGUuY3NzIjsKCmJvZHkgewoJYmFja2dyb3VuZC1jb2xvcjogcmVkOwp9CgpAaW1wb3J0ICIuL2Fub3RoZXItc3R5bGUuY3NzIjsK)

Closes #1824 Closes #903 Closes #2194 Closes #2692

## 💻 Examples

Anytime parcel minifies a css file

## ✔️ PR Todo

- [x] SASS ([demo](http://sokra.github.io/source-map-visualization/#base64,Ym9keXtmb250OjEwMCUgSGVsdmV0aWNhLHNhbnMtc2VyaWY7Y29sb3I6IzMzM31ib2R5IGRpdntiYWNrZ3JvdW5kLWNvbG9yOnJlZDt3aWR0aDoxMDBweDtoZWlnaHQ6MTAwcHh9Ci8qIyBzb3VyY2VNYXBwaW5nVVJMPS9zdHlsZS41NmY1ZGMzMS5jc3MubWFwKi8=,eyJ2ZXJzaW9uIjozLCJzb3VyY2VzIjpbInN0eWxlLnNjc3MiXSwibmFtZXMiOltdLCJtYXBwaW5ncyI6IkFBR0EsS0FDRSwrQkFDQSxXQUVBLFNBQ0MscUJBQ0EsWUFDQSIsImZpbGUiOiJzdHlsZS41NmY1ZGMzMS5jc3MubWFwIiwic291cmNlUm9vdCI6Ii4uIiwic291cmNlc0NvbnRlbnQiOlsiJGZvbnQtc3RhY2s6ICAgIEhlbHZldGljYSwgc2Fucy1zZXJpZjtcbiRwcmltYXJ5LWNvbG9yOiAjMzMzO1xuXG5ib2R5IHtcbiAgZm9udDogMTAwJSAkZm9udC1zdGFjaztcbiAgY29sb3I6ICRwcmltYXJ5LWNvbG9yO1xuXG4gIGRpdiB7XG4gIFx0YmFja2dyb3VuZC1jb2xvcjogcmVkO1xuICBcdHdpZHRoOiAxMDBweDtcbiAgXHRoZWlnaHQ6IDEwMHB4O1xuICB9XG59Il19,JGZvbnQtc3RhY2s6ICAgIEhlbHZldGljYSwgc2Fucy1zZXJpZjsKJHByaW1hcnktY29sb3I6ICMzMzM7Cgpib2R5IHsKICBmb250OiAxMDAlICRmb250LXN0YWNrOwogIGNvbG9yOiAkcHJpbWFyeS1jb2xvcjsKCiAgZGl2IHsKICAJYmFja2dyb3VuZC1jb2xvcjogcmVkOwogIAl3aWR0aDogMTAwcHg7CiAgCWhlaWdodDogMTAwcHg7CiAgfQp9))
- [x] SASS tests
- [x] LESS ([demo](http://sokra.github.io/source-map-visualization/#base64,Ym9keXtmb250LWZhbWlseTptb25vc3BhY2V9ZGl2e2JhY2tncm91bmQtY29sb3I6cmVkO3dpZHRoOjEwMHB4O2hlaWdodDoyMDBweH0KLyojIHNvdXJjZU1hcHBpbmdVUkw9L3N0eWxlLmY3OTZmYWFhLmNzcy5tYXAqLw==,eyJ2ZXJzaW9uIjozLCJzb3VyY2VzIjpbInN0eWxlLmxlc3MiXSwibmFtZXMiOltdLCJtYXBwaW5ncyI6IkFBR0EsS0FDQyxzQkFHRCxJQUNFLHFCQUNBLFlBQ0EiLCJmaWxlIjoic3R5bGUuZjc5NmZhYWEuY3NzLm1hcCIsInNvdXJjZVJvb3QiOiIuLiIsInNvdXJjZXNDb250ZW50IjpbIkB3aWR0aDogMTAwcHg7XG5AaGVpZ2h0OiBAd2lkdGggKiAyO1xuXG5ib2R5IHtcblx0Zm9udC1mYW1pbHk6IG1vbm9zcGFjZTtcbn1cblxuZGl2IHtcbiAgYmFja2dyb3VuZC1jb2xvcjogcmVkO1xuICB3aWR0aDogQHdpZHRoO1xuICBoZWlnaHQ6IEBoZWlnaHQ7XG59Il19,QHdpZHRoOiAxMDBweDsKQGhlaWdodDogQHdpZHRoICogMjsKCmJvZHkgewoJZm9udC1mYW1pbHk6IG1vbm9zcGFjZTsKfQoKZGl2IHsKICBiYWNrZ3JvdW5kLWNvbG9yOiByZWQ7CiAgd2lkdGg6IEB3aWR0aDsKICBoZWlnaHQ6IEBoZWlnaHQ7Cn0=))
- [x] LESS tests
- [x] importing css files with source map comment
- [x] extract CSSAsset's and JSAsset's `loadSourceMap` function
- [x] extendSourceMap doesn't really work with new SourceMap().generateEmptyMap
- [x] rewrite sourcemap sources of imported file (and SASS?) to be relative to entry root
- [x] cache invalidation? (At which stage (Asset/Packager) should sourcemaps be disabled to make `--no-sourcemaps` work)
- [x] should it be `app.js` -> `app.js.map` or ~~`app.map`~~


packages/core/parcel-bundler/src/assets/CSSAsset.js
```diff
@@ -125,6 +126,25 @@ class CSSAsset extends Asset {
         'module.exports = ' + JSON.stringify(this.cssModules, null, 2) + ';';
     }
 
+    let map;
+    if (this.options.sourceMaps && this.sourceMap)
+      map = new SourceMap(
+        this.sourceMap._mappings._array.map(v => ({
```
mischnic: Is there a function for this conversion somewhere already?

packages/core/integration-tests/test/sourcemaps.js
```diff
@@ -314,4 +323,159 @@ describe('sourcemaps', function() {
...
+    //   },
+    //   "map 'body'"
+    // );
+
+    assert.deepStrictEqual(
+      consumer.originalPositionFor(
+        indexToLineCol(raw, raw.indexOf('background-color'))
+      ),
+      {
+        source: '../integration/sourcemap-css-import/style.css',
```
lustoykov: I have some thoughts... maybe some of them not entirely correct, but would share them.

Surprisingly, this one fails for me with

```
 + expected - actual

 -  "source": "../integration/sourcemap-css-import/other-style.css"
 +  "source": "../integration/sourcemap-css-import/style.css"
```

Does it pass for you? Luckily, the source maps actually work! It's the tests that fail for some reason.

I think the reason is the assumption how `consumer.originalPositionFor` works and the implementation of `indexToLineCol`. As far as I was able to reproduce, `originalPositionFor` takes into consideration the contents of each `@import`. Meaning, if we look for `originalPosition` for a certain expression, the result returned would also count lines in all `@import`s in the file where the expression has been found. This is important, because it means that 

```
consumer.originalPositionFor(
  indexToLineCol(raw, raw.indexOf('background-color'))
)
```

would consider lines in `other-style.css` too, however we only match it against the contents of `style.css` 

```
line: indexToLineCol(style, style.indexOf('background-color')).line
```

As opposed to `style.css` concatenated with `other-style.css.` To verify: try adding lines / styles in `other-style.css` and see how `originalPositionFor` changes its result (lines increase). 

So I think a small adjustment to consider all imports lines when doing assertion would fix this.

For instance,

```
 line: indexToLineCol(style, style.indexOf('background-color')).line,
```

would be (more or less, we have to take into consideration extracting number of @import lines I guess)

```
 line: indexToLineCol(style.concat(otherStyle), style.concat(otherStyle).indexOf('background-color')).line,
```

Maybe the function `indexToLineCol` has to be then rewritten according to how `originalPositionFor works` and this would be it. 

I'm not entirely sure, because I don't have complete understanding of what's going on here, but I guess this would be my bet why tests fail.

packages/core/integration-tests/test/sourcemaps.js
```diff
@@ -314,4 +323,159 @@ describe('sourcemaps', function() {
...
+    //   consumer.originalPositionFor(indexToLineCol(raw, raw.indexOf('body'))),
+    //   {
+    //     source: '../integration/sourcemap-css-import/style.css',
+    //     name: null,
+    //     ...indexToLineCol(style, style.indexOf('body'))
+    //   },
+    //   "map 'body'"
+    // );
+
+    assert.deepStrictEqual(
```
lustoykov: I have some thoughts... maybe some of them not entirely correct, but would share them.

Surprisingly, this one fails for me with

```
 + expected - actual

 -  "source": "../integration/sourcemap-css-import/other-style.css"
 +  "source": "../integration/sourcemap-css-import/style.css"
```

Does it pass for you? Luckily, the source maps actually work! It's the tests that fail for some reason.

I think the reason is the assumption how `consumer.originalPositionFor` works and the implementation of `indexToLineCol`. As far as I was able to reproduce, `originalPositionFor` takes into consideration the contents of each `@import`. Meaning, if we look for `originalPosition` for a certain expression, the result returned would also count lines in all `@import`s in the file where the expression has been found. This is important because it means that 

```
consumer.originalPositionFor(
  indexToLineCol(raw, raw.indexOf('background-color'))
)
```

would consider lines in `other-style.css` too, however we only match it against the contents of `style.css` 

```
line: indexToLineCol(style, style.indexOf('background-color')).line
```

As opposed to `style.css` concatenated with `other-style.css.` To verify: try adding lines/styles in `other-style.css` and see how `originalPositionFor` changes its result (lines increase). 

So I think a small adjustment to consider all imports lines when doing assertion would fix this.

For instance,

```
 line: indexToLineCol(style, style.indexOf('background-color')).line,
```

would be (more or less, we have to take into consideration extracting the number of @import lines I guess)

```
 line: indexToLineCol(style.concat(otherStyle), style.concat(otherStyle).indexOf('background-color')).line,
```

Maybe the function `indexToLineCol` has to be then rewritten according to how `originalPositionFor works` and this would be it. 

I'm not entirely sure, because I don't have a complete understanding of what's going on here, but I guess this would be my bet why tests fail.

mischnic: Thank you for taking a look at this, but the tests were actually correct.

> Meaning, if we look for originalPosition for a certain expression, the result returned would also count lines in all @imports in the file where the expression has been found. This is important because it means that
> ```js
> consumer.originalPositionFor(
>   indexToLineCol(raw, raw.indexOf('background-color'))
> )
> ```
> would consider lines in other-style.css too, however we only match it against the contents of style.css

Maybe I can clear this up: the import statements get removed by parcel and the corresponding CSS files get added to the parent asset as a dependency. Every asset is processed by PostCSS and then the dependencies and the asset itself (in that order) are concatenated into a (output)  file. During the concatenation, we keep track of the the length of the previous segments so that the sourcemap indices can be adjusted accordingly.

packages/core/parcel-bundler/src/assets/CSSAsset.js
```diff
@@ -125,6 +126,30 @@ class CSSAsset extends Asset {
...
+            column: v.originalColumn
+          },
+          generated: {
+            line: v.generatedLine,
+            column: v.generatedColumn
+          }
+        }))
+        .filter(
+          ({original: {line, column}}) =>
+            line - 1 < source.length && column < source[line - 1].length
```
mischnic: PostCSS seems to generate sourcemaps with invalid source positions: [with this visualizer](http://sokra.github.io/source-map-visualization/#base64,ZGl2e3dpZHRoOjEwMHB4O2hlaWdodDoxMDBweH1tYWlue2ZvbnQtZmFtaWx5Om1vbm9zcGFjZX1ib2R5e2JhY2tncm91bmQtY29sb3I6cmVkfQovKiMgc291cmNlTWFwcGluZ1VSTD0vc3R5bGUuY3NzLm1hcCov,eyJ2ZXJzaW9uIjozLCJzb3VyY2VzIjpbIm90aGVyLXN0eWxlLmNzcyIsImFub3RoZXItc3R5bGUuY3NzIiwic3R5bGUuY3NzIl0sIm5hbWVzIjpbXSwibWFwcGluZ3MiOiJBQUNBLElBQ0MsWUFBYSxBQUNiLFlBQWMsQ0FDZCxBQ0pELEtBQ0MscUJBQXVCLENBQ3ZCLEFDQ0QsS0FDQyxvQkFBc0IsQ0FDdEIiLCJmaWxlIjoic3R5bGUuY3NzLm1hcCIsInNvdXJjZVJvb3QiOiIuLi9pbnRlZ3JhdGlvbi9zb3VyY2VtYXAtY3NzLWltcG9ydCIsInNvdXJjZXNDb250ZW50IjpbIlxuZGl2IHtcblx0d2lkdGg6IDEwMHB4O1xuXHRoZWlnaHQ6IDEwMHB4O1xufVxuIiwibWFpbiB7XG5cdGZvbnQtZmFtaWx5OiBtb25vc3BhY2U7XG59XG4iLCJcbkBpbXBvcnQgXCIuL290aGVyLXN0eWxlLmNzc1wiO1xuXG5ib2R5IHtcblx0YmFja2dyb3VuZC1jb2xvcjogcmVkO1xufVxuXG5AaW1wb3J0IFwiLi9hbm90aGVyLXN0eWxlLmNzc1wiO1xuIl19,CmRpdiB7Cgl3aWR0aDogMTAwcHg7CgloZWlnaHQ6IDEwMHB4Owp9Cg==,bWFpbiB7Cglmb250LWZhbWlseTogbW9ub3NwYWNlOwp9Cg==,CkBpbXBvcnQgIi4vb3RoZXItc3R5bGUuY3NzIjsKCmJvZHkgewoJYmFja2dyb3VuZC1jb2xvcjogcmVkOwp9CgpAaW1wb3J0ICIuL2Fub3RoZXItc3R5bGUuY3NzIjsK), these mappings are clearly visible (by hovering over the highlighted entries in the bottom section).
With the filtering, [this](http://sokra.github.io/source-map-visualization/#base64,ZGl2e3dpZHRoOjEwMHB4O2hlaWdodDoxMDBweH1tYWlue2ZvbnQtZmFtaWx5Om1vbm9zcGFjZX1ib2R5e2JhY2tncm91bmQtY29sb3I6cmVkfQovKiMgc291cmNlTWFwcGluZ1VSTD0vc3R5bGUuY3NzLm1hcCov,eyJ2ZXJzaW9uIjozLCJzb3VyY2VzIjpbIm90aGVyLXN0eWxlLmNzcyIsImFub3RoZXItc3R5bGUuY3NzIiwic3R5bGUuY3NzIl0sIm5hbWVzIjpbXSwibWFwcGluZ3MiOiJBQUNBLElBQ0MsWUFDQSxhQ0hELEtBQ0Msc0JDRUQsS0FDQyIsImZpbGUiOiJzdHlsZS5jc3MubWFwIiwic291cmNlUm9vdCI6Ii4uL2ludGVncmF0aW9uL3NvdXJjZW1hcC1jc3MtaW1wb3J0Iiwic291cmNlc0NvbnRlbnQiOlsiXG5kaXYge1xuXHR3aWR0aDogMTAwcHg7XG5cdGhlaWdodDogMTAwcHg7XG59XG4iLCJtYWluIHtcblx0Zm9udC1mYW1pbHk6IG1vbm9zcGFjZTtcbn1cbiIsIlxuQGltcG9ydCBcIi4vb3RoZXItc3R5bGUuY3NzXCI7XG5cbmJvZHkge1xuXHRiYWNrZ3JvdW5kLWNvbG9yOiByZWQ7XG59XG5cbkBpbXBvcnQgXCIuL2Fub3RoZXItc3R5bGUuY3NzXCI7XG4iXX0=,CmRpdiB7Cgl3aWR0aDogMTAwcHg7CgloZWlnaHQ6IDEwMHB4Owp9Cg==,bWFpbiB7Cglmb250LWZhbWlseTogbW9ub3NwYWNlOwp9Cg==,CkBpbXBvcnQgIi4vb3RoZXItc3R5bGUuY3NzIjsKCmJvZHkgewoJYmFja2dyb3VuZC1jb2xvcjogcmVkOwp9CgpAaW1wb3J0ICIuL2Fub3RoZXItc3R5bGUuY3NzIjsK) is the output.

Chrome works with both cases, but the `source-map` package used in the tests doesn't behave as expected with the duplicate & incorrect mapppings.

mischnic:
Some vue tests fail because the order of the `childBundles` seems to change, the tests pass on my machine.

devongovett:
Awesome work, really excited about this! 🎉 

Will take a look soon.

packages/core/integration-tests/test/sourcemaps.js
```diff
@@ -2,8 +2,65 @@ const assert = require('assert');
 const fs = require('@parcel/fs');
 const path = require('path');
 const mapValidator = require('sourcemap-validator');
+const SourceMap = require('parcel-bundler/src/SourceMap');
```
mischnic: Is importing an "internal" file from a different package acceptable?

packages/core/parcel-bundler/src/assets/CSSAsset.js
```diff
@@ -1,17 +1,26 @@
...
 const valueParser = require('postcss-value-parser');
 const postcssTransform = require('../transforms/postcss');
 const CssSyntaxError = require('postcss/lib/css-syntax-error');
+const SourceMap = require('../SourceMap');
+const logger = require('@parcel/logger');
 
 const URL_RE = /url\s*\("?(?![a-z]+:)/;
 const IMPORT_RE = /@import/;
 const PROTOCOL_RE = /^[a-z]+:/;
+const SOURCEMAP_RE = /\/\*\s*[@#]\s*sourceMappingURL\s*=\s*([^\s]+)\s*\*\//;
```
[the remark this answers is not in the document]
mischnic: Well, not quite:
CSS: `/* #sourceMappingURL= */`
JS: `// #sourceMappingURL= `

So this would be possible:
`const SOURCEMAP_RE = new RegExp("\/\*"+ common-part + "\s*\*\/")`

packages/core/parcel-bundler/src/assets/CSSAsset.js
```diff
@@ -23,7 +32,9 @@ class CSSAsset extends Asset {
   }
 
   parse(code) {
-    let root = postcss.parse(code, {from: this.name, to: this.name});
+    let root = postcss.parse(code, {
```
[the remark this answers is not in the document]
mischnic: The postcss docs say for `parse`:
bildschirmfoto 2019-01-05 um 18 10 40
So `to` was never used in that call

mischnic:
Last remaining issue: `extendSourceMap` doesn't really work with `new SourceMap().generateEmptyMap`
```js
     } else {
        map = new SourceMap().generateEmptyMap(this.relativeName, css);
      }

      if (this.previousSourceMap) {
        map = await new SourceMap().extendSourceMap(
          this.previousSourceMap,
          map
        );
      }

      if (this.contentsSourceMap) {
        map = await new SourceMap().extendSourceMap(
          this.contentsSourceMap,
          map
        );
      }
```

packages/core/parcel-bundler/src/assets/CSSAsset.js
```diff
@@ -125,6 +148,48 @@ class CSSAsset extends Asset {
...
+            typeof this.sourceMap === 'string'
+              ? JSON.parse(this.sourceMap)
+              : this.sourceMap;
+
+          map = await new SourceMap().addMap(map);
+
+          if (this.sourceMap.toJSON) {
+            // this.sourceMap instanceof SourceMapGenerator
+            let sourceLines = this.contents.split('\n');
+            map.mappings = map.mappings.filter(
```
mischnic: PostCSS seems to generate sourcemaps with invalid source positions: [with this visualizer](http://sokra.github.io/source-map-visualization/#base64,ZGl2e3dpZHRoOjEwMHB4O2hlaWdodDoxMDBweH1tYWlue2ZvbnQtZmFtaWx5Om1vbm9zcGFjZX1ib2R5e2JhY2tncm91bmQtY29sb3I6cmVkfQovKiMgc291cmNlTWFwcGluZ1VSTD0vc3R5bGUuY3NzLm1hcCov,eyJ2ZXJzaW9uIjozLCJzb3VyY2VzIjpbIm90aGVyLXN0eWxlLmNzcyIsImFub3RoZXItc3R5bGUuY3NzIiwic3R5bGUuY3NzIl0sIm5hbWVzIjpbXSwibWFwcGluZ3MiOiJBQUNBLElBQ0MsWUFBYSxBQUNiLFlBQWMsQ0FDZCxBQ0pELEtBQ0MscUJBQXVCLENBQ3ZCLEFDQ0QsS0FDQyxvQkFBc0IsQ0FDdEIiLCJmaWxlIjoic3R5bGUuY3NzLm1hcCIsInNvdXJjZVJvb3QiOiIuLi9pbnRlZ3JhdGlvbi9zb3VyY2VtYXAtY3NzLWltcG9ydCIsInNvdXJjZXNDb250ZW50IjpbIlxuZGl2IHtcblx0d2lkdGg6IDEwMHB4O1xuXHRoZWlnaHQ6IDEwMHB4O1xufVxuIiwibWFpbiB7XG5cdGZvbnQtZmFtaWx5OiBtb25vc3BhY2U7XG59XG4iLCJcbkBpbXBvcnQgXCIuL290aGVyLXN0eWxlLmNzc1wiO1xuXG5ib2R5IHtcblx0YmFja2dyb3VuZC1jb2xvcjogcmVkO1xufVxuXG5AaW1wb3J0IFwiLi9hbm90aGVyLXN0eWxlLmNzc1wiO1xuIl19,CmRpdiB7Cgl3aWR0aDogMTAwcHg7CgloZWlnaHQ6IDEwMHB4Owp9Cg==,bWFpbiB7Cglmb250LWZhbWlseTogbW9ub3NwYWNlOwp9Cg==,CkBpbXBvcnQgIi4vb3RoZXItc3R5bGUuY3NzIjsKCmJvZHkgewoJYmFja2dyb3VuZC1jb2xvcjogcmVkOwp9CgpAaW1wb3J0ICIuL2Fub3RoZXItc3R5bGUuY3NzIjsK), these mappings are clearly visible (by hovering over the highlighted entries in the bottom section).
With the filtering, [this](http://sokra.github.io/source-map-visualization/#base64,ZGl2e3dpZHRoOjEwMHB4O2hlaWdodDoxMDBweH1tYWlue2ZvbnQtZmFtaWx5Om1vbm9zcGFjZX1ib2R5e2JhY2tncm91bmQtY29sb3I6cmVkfQovKiMgc291cmNlTWFwcGluZ1VSTD0vc3R5bGUuY3NzLm1hcCov,eyJ2ZXJzaW9uIjozLCJzb3VyY2VzIjpbIm90aGVyLXN0eWxlLmNzcyIsImFub3RoZXItc3R5bGUuY3NzIiwic3R5bGUuY3NzIl0sIm5hbWVzIjpbXSwibWFwcGluZ3MiOiJBQUNBLElBQ0MsWUFDQSxhQ0hELEtBQ0Msc0JDRUQsS0FDQyIsImZpbGUiOiJzdHlsZS5jc3MubWFwIiwic291cmNlUm9vdCI6Ii4uL2ludGVncmF0aW9uL3NvdXJjZW1hcC1jc3MtaW1wb3J0Iiwic291cmNlc0NvbnRlbnQiOlsiXG5kaXYge1xuXHR3aWR0aDogMTAwcHg7XG5cdGhlaWdodDogMTAwcHg7XG59XG4iLCJtYWluIHtcblx0Zm9udC1mYW1pbHk6IG1vbm9zcGFjZTtcbn1cbiIsIlxuQGltcG9ydCBcIi4vb3RoZXItc3R5bGUuY3NzXCI7XG5cbmJvZHkge1xuXHRiYWNrZ3JvdW5kLWNvbG9yOiByZWQ7XG59XG5cbkBpbXBvcnQgXCIuL2Fub3RoZXItc3R5bGUuY3NzXCI7XG4iXX0=,CmRpdiB7Cgl3aWR0aDogMTAwcHg7CgloZWlnaHQ6IDEwMHB4Owp9Cg==,bWFpbiB7Cglmb250LWZhbWlseTogbW9ub3NwYWNlOwp9Cg==,CkBpbXBvcnQgIi4vb3RoZXItc3R5bGUuY3NzIjsKCmJvZHkgewoJYmFja2dyb3VuZC1jb2xvcjogcmVkOwp9CgpAaW1wb3J0ICIuL2Fub3RoZXItc3R5bGUuY3NzIjsK) is the output.

Chrome works with both cases, but the `source-map` package used in the tests doesn't behave as expected with the duplicate & incorrect mapppings.

packages/core/parcel-bundler/src/SourceMap.js
```diff
@@ -38,7 +38,13 @@ class SourceMap {
   }
 
   async addMap(map, lineOffset = 0, columnOffset = 0) {
+    if (map.toJSON) {
```
DeMoorJasper: When is this the case?
Parcel never creates a sourcemapGenerator (as the SourceMap class is both a Generator and Consumer and faster) or does any css asset return a sourceMapGenerator instead of a sourceMap?

packages/core/parcel-bundler/src/SourceMap.js
```diff
@@ -57,6 +63,7 @@ class SourceMap {
       }
     } else {
       if (!map.eachMapping) {
+        // not a SourceMapConsumer
```
DeMoorJasper: This check is for a SourceMap instance that has been converted to JSON (to transfer it from worker=>Master).
So this comment is slightly incorrect?

packages/core/parcel-bundler/src/assets/CSSAsset.js
```diff
@@ -125,6 +148,48 @@ class CSSAsset extends Asset {
         'module.exports = ' + JSON.stringify(this.cssModules, null, 2) + ';';
     }
 
+    let map;
+    if (this.options.sourceMaps) {
+      if (this.sourceMap) {
+        if (this.sourceMap instanceof SourceMap) {
+          map = this.sourceMap;
+        } else {
+          map =
```
DeMoorJasper: This line isn't necessary, `addMap` should already be able to handle a sourcemap both in string and object form

packages/core/parcel-bundler/src/SourceMap.js
```diff
@@ -38,7 +38,13 @@ class SourceMap {
   }
 
   async addMap(map, lineOffset = 0, columnOffset = 0) {
+    if (map.toJSON) {
+      // SourceMapGenerator
+      map = map.toJSON();
+    }
+
     if (!(map instanceof SourceMap) && map.version) {
```
DeMoorJasper: Probably should be changed to

```Javascript
if ((!(map instanceof SourceMap) && map.version) || typeof map === 'string') {
```

To be able to handle stringified maps (as mentioned in one of my later comments), I probably forgot to add this to this function (it's already part of getConsumer)

packages/core/parcel-bundler/src/assets/CSSAsset.js
DeMoorJasper: Shouldn't these mappings be updated before converting them to a `SourceMap` instance.

mischnic: `this.sourceMap` might be a SourceMapGenerator, which doesn't expose the mappings. The SourceMap class is "private", so I can just use the mappings there.

packages/core/parcel-bundler/src/SourceMap.js
mischnic: > Parcel never creates a sourcemapGenerator

PostCSS does...

packages/core/parcel-bundler/src/assets/CSSAsset.js
DeMoorJasper: When? What creates a SourceMapGenerator? `private` doesn't exist in Javascript so not sure what you mean by that.

packages/core/parcel-bundler/src/assets/CSSAsset.js
```diff
@@ -93,21 +100,37 @@ class CSSAsset extends Asset {
...
   }
 
-  generate() {
-    let css = this.ast ? this.ast.render() : this.contents;
+  async generate() {
+    let css;
+    if (this.ast) {
+      let result = this.ast.render(this.name);
+      css = result.css;
+      if (result.map) this.sourceMap = result.map;
```
DeMoorJasper: perhaps change this to `result.map.toJSON` with the lineupdates you do later.
So this is outputs a standardised sourcemap?

packages/core/parcel-bundler/src/SourceMap.js
DeMoorJasper: Ow alright didn't expect that.
Perhaps better to move that transforming to a standard sourcemap to where you get the result from PostCSS? (See later comment)

mischnic: This need to happen in `CSSAst.render` and transforms/postcss. This way, I tried to prevent duplicating the filter part.

packages/core/parcel-bundler/src/assets/CSSAsset.js
DeMoorJasper: Owww damn thought the fix for null mappings was already in master, anyways this PR should fix this: #2149 afaik

DeMoorJasper: Looking at the example you put it the original comment it looks like null mappings. I might be wrong

mischnic: > If that PR doesn't fix it, I'm not sure why it's in those sourcemaps, but it looks to me like it should be fixed on SourceMap level instead of filtering this out. As it's probably a valid sourcemap.

The thing is: the sourcemaps work just fine in Chrome. Testing them is the issue. I could only do the filtering in the tests and leave the invalid mappings in the sourcemap?

DeMoorJasper: Yeah that's probably a better idea

mischnic: Please take a look at [this](http://sokra.github.io/source-map-visualization/#base64,Ym9keXtmb250OjEwMCUgSGVsdmV0aWNhLHNhbnMtc2VyaWY7Y29sb3I6IzMzM31ib2R5IGRpdntiYWNrZ3JvdW5kLWNvbG9yOnJlZDt3aWR0aDoxMDBweDtoZWlnaHQ6MTAwcHh9bWFpbntkaXNwbGF5Om5vbmV9Ci8qIyBzb3VyY2VNYXBwaW5nVVJMPS9zdHlsZS5jc3MubWFwICov,eyJ2ZXJzaW9uIjozLCJzb3VyY2VzIjpbImxpYnJhcnkuc2NzcyIsInN0eWxlLmNzcyJdLCJuYW1lcyI6W10sIm1hcHBpbmdzIjoiQUFHQSxLQUNFLCtCQUNBLFVBQUEsQ0FFQSxTQUNDLHFCQUNBLFlBQ0EsWUFBQSxDQUFBLEFDUkgsS0FDQyxZQUFjLENBQ2QiLCJmaWxlIjoic3R5bGUuY3NzLm1hcCIsInNvdXJjZVJvb3QiOiIuLi9pbnRlZ3JhdGlvbi9zb3VyY2VtYXAtY3NzLWV4aXN0aW5nIiwic291cmNlc0NvbnRlbnQiOlsiJGZvbnQtc3RhY2s6ICAgIEhlbHZldGljYSwgc2Fucy1zZXJpZjtcbiRwcmltYXJ5LWNvbG9yOiAjMzMzO1xuXG5ib2R5IHtcbiAgZm9udDogMTAwJSAkZm9udC1zdGFjaztcbiAgY29sb3I6ICRwcmltYXJ5LWNvbG9yO1xuXG4gIGRpdiB7XG4gIFx0YmFja2dyb3VuZC1jb2xvcjogcmVkO1xuICBcdHdpZHRoOiAxMDBweDtcbiAgXHRoZWlnaHQ6IDEwMHB4O1xuICB9XG59IiwiQGltcG9ydCBcIi4vbGlicmFyeS5jc3NcIjtcblxubWFpbiB7XG5cdGRpc3BsYXk6IG5vbmU7XG59XG4iXX0=,JGZvbnQtc3RhY2s6ICAgIEhlbHZldGljYSwgc2Fucy1zZXJpZjsKJHByaW1hcnktY29sb3I6ICMzMzM7Cgpib2R5IHsKICBmb250OiAxMDAlICRmb250LXN0YWNrOwogIGNvbG9yOiAkcHJpbWFyeS1jb2xvcjsKCiAgZGl2IHsKICAJYmFja2dyb3VuZC1jb2xvcjogcmVkOwogIAl3aWR0aDogMTAwcHg7CiAgCWhlaWdodDogMTAwcHg7CiAgfQp9,QGltcG9ydCAiLi9saWJyYXJ5LmNzcyI7CgptYWluIHsKCWRpc3BsYXk6IG5vbmU7Cn0K) (from the `should load existing sourcemaps for CSS files` test with minification enabled).

The generated column 102 maps to two valid positions at the same time. 

Filtering it after concatenating doesn't seem to work...

packages/core/integration-tests/test/sourcemaps.js
mischnic: I'm using it to test the sourcemaps:
```js
let index = map.findClosestGenerated(
  generatedPosition.line,
  generatedPosition.column
);
```

DeMoorJasper: In that case it's probably fine

mischnic:
Running with HMR also adds the sourcemaps of bundle-url.js and css-loader.js into the CSS sourcemap. But these two are packaged into a seperate JS bundle which should also have it's own sourcemap bundle? (This happens without this PR as well).

DeMoorJasper:
@mischnic well, before there were no css sourcemaps so this wasn't really an issue.

But it's definitely something we'll need to figure out before getting this in

packages/core/parcel-bundler/src/assets/CSSAsset.js
```diff
@@ -172,13 +250,24 @@ class CSSAst {
...
-      this.css = '';
-      postcss.stringify(this.root, c => (this.css += c));
+      let {css, map} = this.root.toResult({
+        to: name,
+        map: {inline: false, annotation: false, sourcesContent: true}
+      });
+
+      this.css = css;
+
+      return {
```
DeMoorJasper: Perhaps we should refactor this to:
```Javascript
return [
  {
    type: 'css',
    sourceMap: map
  }
];
```

and create (or extend existing by looking at childBundles) the actual sourcemap bundles here: https://github.com/parcel-bundler/parcel/blob/master/packages/core/parcel-bundler/src/Bundle.js#L39

Probably a good idea to get some input from Devon as well on this as Pipeline is his design

mischnic: JSAsset does something like that already:
```js
return {
    js: code,
    map: this.sourceMap
};
```
But I struggle to find where is this actually used (and how to make it work for css)

DeMoorJasper: That's the same as:

```
return [
  {
    type: 'map',
    value: this.sourceMap
  },
  {
    type: 'js',
    value: code
  }
];
```

This was just for backwards compatibility

packages/core/parcel-bundler/src/assets/CSSAsset.js
mischnic: Nearly fixed in PostCSS.

packages/core/parcel-bundler/src/assets/CSSAsset.js
mischnic: @devongovett ?

mischnic: At first, I thought `SourceMapPackager.shouldAddAsset` could be used for this, but it doesn't get called at all.

devongovett: Yeah I think we will need to add a `map` property to renditions themselves rather than having a the map be a separate rendition since only one rendition of the same type is supported (and there would be no way to associate them). Then, we'll need to special case source maps in the Bundler to create the `map` bundles. Perhaps in `Bundle#addAsset` we could add the map to a sibling `map` bundle? https://github.com/parcel-bundler/parcel/blob/master/packages/core/parcel-bundler/src/Bundle.js#L39

mischnic: In `Bundle#addAsset`, the map isn't really available anymore (on `asset.generated`), because it got "filtered" here: https://github.com/parcel-bundler/parcel/blob/d7dc51f9eacd80aadd1e41888771033e447758f1/packages/core/parcel-bundler/src/Pipeline.js#L25-L27

devongovett: Oh right... hmm

mischnic: >  For now, maybe you could move this conversion from array to object into the Bundler after adding assets to the bundles (createBundleTree). We just need to make sure it is converted before running Packagers for backwards compatibility

1. Assets are added by createBundleTree via Bundle#addAsset
2. Transform asset.generated to generatedMap
3. Run packagers: JSAsset calls Bundle#addAsset for bundle loaders 🙈

mischnic: I've chosen a different approach (loading sourcemaps from previous asset stages doesn't work yet)

mischnic: How can/why should a bundle be a child bundle and a sibling bundle at the same time?

mischnic: @devongovett ?

mischnic: > I am not against adding hacks, but I'm also fine with waiting for this feature until Parcel 2 and perhaps starting on the implementation there instead. What do you think?

I've solved the previous issue, regarding something more cosmetic:
The first two files go together and the next two have differing names (but would work fine):
```
test.9ac3ef99.css
test.9ac3ef99.css.map
test.f2b3894d.js
test.23936829.js.map
```

A possible fix?
```diff
diff --git a/packages/core/parcel-bundler/src/Bundle.js b/packages/core/parcel-bundler/src/Bundle.js
index 16d44aee..94e08697 100644
--- a/packages/core/parcel-bundler/src/Bundle.js
+++ b/packages/core/parcel-bundler/src/Bundle.js
@@ -120,6 +120,8 @@ class Bundle {
   }

   getHashedBundleName(contentHash) {
+    if (this.type == 'map')
+      return this.parentBundle.getHashedBundleName(contentHash) + '.map';
     // If content hashing is enabled, generate a hash from all assets in the bundle.
     // Otherwise, use a hash of the filename so it remains consistent across builds.
     let basename = Path.basename(this.name);
```

packages/core/parcel-bundler/src/assets/CSSAsset.js
```diff
@@ -125,11 +149,40 @@ class CSSAsset extends Asset {
...
+          css
+        );
+      }
+    }
+
     return [
       {
         type: 'css',
         value: css,
-        cssModules: this.cssModules
```
mischnic: In VueAsset, it's called `modules: `. Should this be `modules: ` as well or isn't it used at all?

mischnic:
Many tests fail because `assertBundleTree`'s order on childbundles is inconsistent?

mischnic:
One last thing and this should be finished:
The `--no-source-maps` flag isn't being cached properly, I think there are two ways to handle this:
- Invalidate the cache based on `--no-source-maps` (so basically re-run the asset)
- Always output sourcemaps in the Asset (performance?) and don't output a SourceMapBundle if the flag is passed

```bash
$ parcel build index.html --no-source-maps

dist/test.7f0ab984.js     1.14 KB    1.40s
dist/index.html             113 B    587ms
dist/test.8d042a79.css       27 B    1.09s

$ parcel build index.html

dist/test.7f0ab984.js     1.14 KB    66ms
dist/index.html             113 B    15ms
dist/test.8d042a79.css       27 B    14ms
```

mischnic:
Ready for review 🚀

packages/core/parcel-bundler/src/Asset.js
```diff
@@ -219,6 +219,19 @@ class Asset {
   }
 
   async postProcess(generated) {
+    let hasMap = false;
```
devongovett: Can we move this code to a separate method, or into the Pipeline class? If we add it here, you'll need to ensure to call `super.postProcess` from all assets that implement a `postProcess` method, as you've done for a couple already. That would be a potentially breaking change for plugins, so if we can move this outside that would be better.

mischnic: Done!

PR #2490, opened 2019-01-01
can config pug pretty mode

miewx:
pug will add space between inline block element  ( see https://github.com/pugjs/pug/issues/2682 ) where use pretty:true, then there will be space in page .

so , I think should let user can config pretty:false for the  visual effect development


miewx:
> You need to run `npx prettier --write packages/core/parcel-bundler/src/assets/PugAsset.js` to make the tests pass

done

devongovett:
Perhaps we should just remove the `pretty` option altogether (unless in the config)? From the issue you linked:

> Do not pass {pretty: true}, it breaks lots of things and as such we generally view it as unsupported.

mischnic:
> Perhaps we should just remove the pretty option altogether

https://pugjs.org/api/reference.html#options:
>pretty: `boolean | string`
>[Deprecated.] Adds whitespace to the resulting HTML to make it easier for a human to read using `'  '` as indentation. If a string is specified, that will be used as indentation instead (e.g. `'\t'`). We strongly recommend against using this option. Too often, it creates subtle bugs in your templates because of the way it alters the interpretation and rendering of whitespace, and so this feature is going to be removed. Defaults to `false`.

devongovett:
Closed by #2582.

PR #2494, opened 2019-01-02
Security fix 1794. WS server missing Origin validation.

carlosgeos:
# ↪️ Pull Request

I believe this PR solves #1783 and amends #1794. I think the merged fix was doing nothing basically (this one --> https://github.com/parcel-bundler/parcel/pull/1794/commits/a1a542278bd9ea08876afaffac7d17d3e6359109)

It was giving Websocket.Server an `origin` option which is not even accepted by the constructor of Websocket.Server: https://github.com/websockets/ws/blob/master/doc/ws.md#new-websocketserveroptions-callback

This would mean that [this](https://www.npmjs.com/advisories/721) (apparently high) security vulnerability is still present up to the latest version as of now (1.11.0)

I used the proper `verifyClient` hook to check the origin.
I also included a default `hmrHostname` on `cli.js` since there was no `hmrHostname` by default !

## 💻 Examples

- Navigating to `localhost:1234` works the same as before. Everything works.
- Some site running malicious JS can't connect to the websocket server to sniff the code changes, since the origin will be `example.com` and not localhost.

## 🚨 Test instructions

I adapted about half the HMR tests and they pass.

The other half use the Node `vm` library and I couldn't find a way for this isolated V8 environment to send an origin header with the WebSocket connection. Therefore they fail. I need some help here :stuck_out_tongue: 

I used the proof of concept found in this blog to test the fix (before and after): https://blog.cal1.cn/post/Sniffing%20Codes%20in%20Hot%20Module%20Reloading%20Messages. I basically set up a remote server, serving some JS that fetches `http://localhost:1234` and gets the random port number of the websocket server, then connects to it. All the details are in the blog post.

Of course the issue with the static assets (where the random port is found, actually) and CORS is still there, and I will look into it next.

## ✔️ PR Todo

- [ ] Added/updated unit tests for this change (PARTIALLY)


packages/core/parcel-bundler/src/cli.js
```diff
@@ -214,6 +214,7 @@ async function bundle(main, command) {
 
   command.throwErrors = false;
   command.scopeHoist = command.experimentalScopeHoisting || false;
+  command.hmrHostname = command.hmrHostname || 'localhost';
```
DeMoorJasper: This is a breaking change, so not sure if this should have a default

DeMoorJasper: That would still be a breaking change, but yes. If you really wanna do it than that would probably be a cleaner solution

carlosgeos: The problem is: 

https://github.com/parcel-bundler/parcel/blob/c3d99dee452640cd0ffb0e17682e482e8c3abd85/packages/core/parcel-bundler/src/cli.js#L25-L28

It says it defaults to window.location.hostname but this is not true. It is only true, for some `var hostname` later in the code, that is used in the client side code for connecting to the websocket server:

https://github.com/parcel-bundler/parcel/blob/c3d99dee452640cd0ffb0e17682e482e8c3abd85/packages/core/parcel-bundler/src/builtins/hmr-runtime.js#L26

This default does not reach HMRServer.js (obviously). When the time to check the origin comes, hmrHostname is empty, and the check never takes place. And the check, even if it happened, does nothing, as explained before.

hmrHostname needs to have a default somewhere and `normalizeOptions` is the best place. I don't see why it's a breaking change: people using a specified hmrHostname (because they are using some complex setup) will still keep that.

Another option: deprecate hmrHostname like you mentioned in https://github.com/parcel-bundler/parcel/pull/1794#issue-203903881 and https://github.com/parcel-bundler/parcel/pull/1482#issuecomment-395091843. The flags `host` or `hostname` can be used to make the origin check, and as fallback to window.location.hostname in the client.

What do you think ?

DeMoorJasper:
Closing this as it hasn't been updated in a while and the fix is included in Parcel 2's HMR

Issue #2499, opened 2019-01-02
Parcel 2: Allow transformer plugins to control cache invalidation


devongovett:
Transformer plugins may need to control when the cache is invalidated. For example, a plugin that inlines environment variables would need to invalidate the cache when those environment variables change.

We could consider simply invalidating the cache whenever ANY environment variable changes, but that would cause too many unneeded invalidations since there are typically lots of environment variables on people's systems that change frequently. It would also make the cache less portable. There will also be other cases when plugins may need to do something custom.

In Parcel 1, there is a `shouldInvalidate` method on transformers for this use case. It will be slightly more complicated in Parcel 2 since there can be multiple transformers in a pipeline that might need to handle this separately.

Proposed implementation:

* Add an optional `shouldInvalidateCache` method to transformers, which is passed a reference to the cached asset's `meta` property. This property lets transformers store custom data on an asset, which is included in the cache. The `shouldInvalidateCache` method could do some comparisons using that data, and return a boolean indicating whether the cache entry should be invalidated.
* The TransformerRunner would call each transformer's `shouldInvalidateCache` method in the pipeline, and if any of them return `true`, the cache entry would be ignored and the pipeline would be fully executed as if there were no cache entry.

abdullahtariq1171:
Thank you @padmaia for writing this comprehensively, It will definitely be helpful, I listed down points from you comment that we need to check for cache as checking only `shouldInvalidateCache` doesn’t suffice 

1. Check if transformer pipeline is changed for a asset (this includes order of transformers too)
2. Check if transformer/plugin `version` of is changed 
2. Check if transformer config is changed, We need cache `transformer.getConfig` results for this
3. Run `shouldInvalidateCache` for stuff like inlines environment variables

--
I totally agree and support idea of agnostic cache, I am not familiar with cache module though I will look into it see what I can do. Will ping if I needed any help.

--
**HELP NEEDED: Questions:** 
- As we need transformer/plugin meta (version, name) and In our `Config` module, we are using `localRequire` which doesn't return the `package.json` data although [resolve](https://github.com/browserify/resolve#resolveid-opts-cb) returns pacakge.json in cb data but we are not using it at the moment (as far as i know). Please confirm and I will create a utility function

- I couldn't figure out how are we testing/running on v2 branch? I see there is 
https://github.com/parcel-bundler/parcel/blob/v2-work-so-far/packages/core/core/bin.js but it throws error for me.

devongovett:
[...]

PR #2502, opened 2019-01-03
Parcel 2: Allow transformer plugins to control cache invalidation

abdullahtariq1171:
# ↪️ Pull Request

Parcel 2: Allow transformer plugins to control cache invalidation

See https://github.com/parcel-bundler/parcel/issues/2499 for detail
## ✔️ PR Todo

### DIFF
--- a/packages/core/core/src/TransformerRunner.js
+++ b/packages/core/core/src/TransformerRunner.js
@@ -39,12 +39,15 @@ class TransformerRunner {
     let code = await fs.readFile(req.filePath, 'utf8');
     let hash = md5(code);
 
+    let pipeline = await this.config.getTransformers(req.filePath);
+
     // If a cache entry matches, no need to transform.
     let cacheEntry = await this.cache.read(req.filePath, req.env);
     if (
       cacheEntry &&
       cacheEntry.hash === hash &&
-      (await checkCachedAssets(cacheEntry.assets))
+      (await checkCachedAssets(cacheEntry.assets)) &&
+      checkPipelineCacheInvalidation(pipeline, cacheEntry)
     ) {
       return cacheEntry;
     }
@@ -57,7 +60,6 @@ class TransformerRunner {
       env: req.env
     });
 
-    let pipeline = await this.config.getTransformers(req.filePath);
     let {assets, initialAssets} = await this.runPipeline(
       input,
       pipeline,
@@ -69,6 +71,7 @@ class TransformerRunner {
       env: req.env,
       hash,
       assets,
+      meta: input.meta,
       initialAssets
     };
 
@@ -227,6 +230,17 @@ async function finalize(asset: Asset, generate: GenerateFunc): Promise<Asset> {
   asset.outputHash = md5(asset.output.code);
 
   return asset;
+}
+
+async function checkPipelineCacheInvalidation(
+  pipeline: Array<Transformer>,
+  cacheEntry: CacheEntry
+): boolean {
+  return pipeline.every(
+    transformer =>
+      !transformer.shouldInvalidateCache ||
+      !transformer.shouldInvalidateCache(cacheEntry.meta)
+  );
 }
 
 async function checkCachedAssets(assets: Array<IAsset>): Promise<boolean> {
--- a/packages/core/types/index.js
+++ b/packages/core/types/index.js
@@ -201,7 +201,8 @@ export type Transformer = {
     assets: Array<Asset>,
     config: ?Config,
     opts: CLIOptions
-  ) => Async<Array<TransformerResult>>
+  ) => Async<Array<TransformerResult>>,
+  shouldInvalidateCache?: (meta: JSONObject) => boolean
 };
 
 export type CacheEntry = {


packages/core/core/src/TransformerRunner.js
```diff
@@ -229,6 +232,17 @@ async function finalize(asset: Asset, generate: GenerateFunc): Promise<Asset> {
...
 }
 
+async function checkPipelineCacheInvalidation(
+  pipeline: Array<Transformer>,
+  cacheEntry: CacheEntry
+): boolean {
+  return pipeline.every(
+    transformer =>
+      !transformer.shouldInvalidateCache ||
+      !transformer.shouldInvalidateCache(cacheEntry.meta)
```
devongovett: The `meta` property would be on an individual asset rather than the whole cache entry. The cache entry can potentially hold multiple assets. For example, in a Vue file, there is typically a JS section, a template, and some CSS. This is the output of a single transformer for one file, so it is stored in one cache entry for that file.

We'll need to do this check for each asset in the cache entry.

packages/core/core/src/TransformerRunner.js
```diff
@@ -229,6 +232,17 @@ async function finalize(asset: Asset, generate: GenerateFunc): Promise<Asset> {
   return asset;
 }
 
+async function checkPipelineCacheInvalidation(
+  pipeline: Array<Transformer>,
+  cacheEntry: CacheEntry
+): boolean {
+  return pipeline.every(
```
devongovett: There is one potential issue here, which is that an asset might switch pipelines in the middle. For example, the following Parcel config could exist:

```json
{
  "transformers": {
    "*.ts": ["@parcel/transformer-typescript"],
    "*.js": ["@parcel/transformer-babel", "@parcel/transformer-js"]
  }
}
```

In this case, a `.ts` file would first be compiled to javascript by the typescript compiler, and then run through the pipeline for `.js` files. We would need to follow all of these transformers through as well in order to ensure that `shouldInvalidateCache` is called on all of the transformers that were originally run to produce the cached asset.

The problem we have is that we don't currently store all of the intermediary steps in the cache, just the final asset at the end of all of the transformers. We also don't store which transformers ran to produce that final asset, so we currently have no way of knowing which transformers to call to check a cache entry. We may need to store this information in the cache. We should probably also store the version of the transformer plugin that generated the asset in the cache as well in order to invalidate it if you upgrade to a newer version.

padmaia:
Hey, love the excitement for caching! I've left a comment on the parent issue which adds some more detail on the requirements. https://github.com/parcel-bundler/parcel/issues/2499#issuecomment-451559499

packages/core/core/src/TransformerRunner.js
```diff
@@ -229,6 +232,17 @@ async function finalize(asset: Asset, generate: GenerateFunc): Promise<Asset> {
   return asset;
 }
 
+async function checkPipelineCacheInvalidation(
+  pipeline: Array<Transformer>,
+  cacheEntry: CacheEntry
```
padmaia: @devongovett Will we need to pass the parcel cli options here too?

packages/core/core/src/TransformerRunner.js
abdullahtariq1171: We can keep array of `appliedTransformers` in asset. Every time we run `transformer.transform` for an asset, we push `asset.appliedTransformers.push(transformer.name)`. We will save this in cache for each asset and load only these transformers while checking cache invalidation.

padmaia: I think we are going to need more information than just an array. Let's say we run a build with the following config

```
"transformers": {
  "*.ts": ["@parcel/transformer-typescript"],
  "*.js": ["@parcel/transformer-babel", "@parcel/transformer-js"]
}
```

and then on the next run we have the this config

```
"transformers": {
  "*.ts": ["@parcel/transformer-typescript"],
  "*.ts.js": ["@parcel/transformer-js"],
  "*.js": ["@parcel/transformer-babel", "@parcel/transformer-js"]
}
```

because maybe it's not necessary to run babel on typescript generated assets. In the first case, the `appliedTransformers` array would be `["@parcel/transformer-typescript", "@parcel/transformer-babel", "@parcel/transformer-js"]`. If we are just going down the line, and none of the transformers have changed, we would think the cache is valid, but it's not because it doesn't have the right applied transformers and the result may be different.

I chose to use an object where the key is the asset filepath and the value is an array of "TransformerMeta" objects which I was using to store name and version and a hash for the transformers config. So the first run would look like:

```
{
  "whatever.ts": [
    {"name": "@parcel/transformer-typescript", "version": "1.0.0", validHash: "ABC123"}
  ],
  "whatever.ts.js": [
    {"name": "@parcel/transformer-babel", "version": "1.0.0", validHash: "ABC123"},
    {"name": "@parcel/transformer-js", "version": "1.0.0", validHash: "ABC123"}
  ]
}
```

When we are checking validity on the second run, if `@parcel/transformer-typescript` has not changed we know it will produce `whatever.ts.js` again, but when we resolve the transformers for the js asset we see that the transformers have changed and the cache is invalid.

I also chose not to store this info directly on the asset because assets stick around in memory throughout the whole process and we really only need that info during the transformation phase. I suppose we could keep it on the asset as long as it is removed before we add it to the graph, but I find it simpler to just have the cache entry have meta about the transformation and the resulting assets separate. That way there's no mental hurdles of "what phases does asset have this info attached?"

padmaia:
Hey there, I'm going to close this as I've started work on the more complex cache invalidation scenarios. If you are still interested you can take a look at a recent spike ( #2886 ) I pushed up to gather feedback. Let me know if you have any thoughts :)

PR #2518, opened 2019-01-06
addURLDependency: use always relative path

### LINKED ISSUE #1801: Wrong relative path to URL assets when bundle lives in subdirectory

mischnic:
# ↪️ Pull Request

Fixes #1801
Previously rejected implementation: #1809

## 💻 Examples

```
assets
├── a
│   └── bundle.css
├── b
│   └── bundle.css
└── img
    └── foo.png
```
Running parcel with the two bundles.css as entry points, in the output
```
dist
├── a
│   └── bundle.css
├── b
│   └── bundle.css
└── foo.d7b28491.png
```
a/bundle.css would contain `url("foo.d7b28491.png")` (should be "../foo")

## ✔️ PR Todo


DeMoorJasper:
@mischnic just fyi the previous PR you linked just got closed because master had been force-pushed and Github automatically closed all open PRs, it had not been rejected.

mischnic:
>just fyi the previous PR you linked just got closed because master had been force-pushed and Github automatically closed all open PRs, it had not been rejected.

Didn't compare the dates 😄 . But @devongovett wrote:
> I think it already gives a relative path. This makes the output more portable depending on where you move it to. Why does this not work?

The previous PR added the public url, this makes the relative path correct.

rhurstdialpad:
This has broken CSS dependencies for me. For example, my fonts are stored in a /src/fonts away from my CSS which is stored in /src/css/fonts.css. Parcel correctly copies my fonts into the dist directory along with the generated CSS file (/dist/fonts.xxxxxx.css & /dist/my_font.xxxxxx.woff), but the final CSS file tries to load the fonts from a URL that includes the relative path from the original CSS file to the fonts directory (url('../fonts/my_font.xxxxxx.woff')), rather than the name of the file in the dist directory (url('my_font.xxxxxx.woff')). As such the parcel dev server cannot resolve the font files.

rhurstdialpad:
This broke our build.

We have `--public-url` set to `/static/compiled/single`
Our fonts are in `/fonts` and our css is in `css/single`
Currently our css files that reference font files are ending up with a url to each font file that includes the public url and a path that seems to be the resolved path from the root dir and the relative path from the css file to the font file. The path from the css directory to the fonts directory is `../../fonts`. So in our case the font urls all look something like `/static/compliled/single/[dir outside project]/[dir outside project]/[font file].xxxxx.woff`. Obviously this is totally wrong, and as a the dev server 404s when trying to load the fonts.

mischnic:
> We have `--public-url` set to `/static/compiled/single`.
Our fonts are in `/fonts` and our CSS is in `css/single`.

Is that second line referring to the url or your folder structure during development?
Could you please give an overview how both look like? Is this correct?
File system:
```
─ MyProject
  ├─ fonts
  │  └─ *.woff
  └─ css
     └─ single
        └─ *.css
```

Wanted output structure:
```
─ /static/compiled/single/
  ├─ *.woff
  └─ *.css
```
What does `dir outside project` mean exactly?


Otherwise, a very minimal reproduction would be helpful.

rhurstdialpad:
That looks about right, with the exception of the output structure. `/static/compiled/single/` is the path I want the dev server to serve the assets on. My dist folder is the default.

When I say folder outside the project I mean just that. If my project is at /Users/Robert/Developer/MyProject, then the path is /static/compliled/single/Robert/Developer/[font file].xxxxx.woff.

Issue #2519, opened 2019-01-07
Autoinstalling react: Unexpected token


mischnic:
# 🐛 bug report

Autoinstalling react doesn't work in a "completely clean" project.

## 🎛 Configuration (.babelrc, package.json, cli command)

No babelrc, no package.json

## 🤔 Expected Behavior

Parcel should install react, react-dom and then bundle.

## 😯 Current Behavior

Parcel doesn't recognize the JSX syntax?
```
🚨  ...../index.js:5:1: Unexpected token (5:1)
  3 |
  4 | const MyComponent = () => (
> 5 | 	<button>Hello</button>
    | ^
  6 | );
  7 |
  8 | ReactDOM.render(<MyComponent />, document.getElementById("root"));
```

## 💁 Possible Solution

Workaround: install `react` as a dependency. Running `parcel index.html` will then corretly install `react-dom`and build successfully.

## 💻 Code Sample

index.js
```jsx
import React from "react";
import ReactDOM from "react-dom";

const MyComponent = () => (
	<button>Hello</button>
);

ReactDOM.render(<MyComponent />, document.getElementById("root"));
```
Command:
```js
parcel index.js
```

## 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Parcel           |38df513
| Node             |11.6.0
| npm/Yarn         |
| Operating System |macOS 10.13.6

lustoykov:
I think if it was `index.jsx` as opposed to `index.js` it should work. Having that in mind - looks ok to me?

lustoykov:
For the protocol, using a normal `.js` extension, without JSX syntax, works fine too.

```js
import React from 'react';
import ReactDOM from 'react-dom';

const MyComponent = () => (
  React.createElement('div', null, 'Hello')
);

ReactDOM.render(
  React.createElement(MyComponent, null, null), 
  document.getElementById('root')
);
```

lustoykov:
I think babel is not even able to generate `AST` currently, because the condition to add `@babel/plugin-transform-react-jsx` fails.

https://github.com/parcel-bundler/parcel/blob/6566e160affa161daa6d5486139e6059f685ba3a/packages/core/parcel-bundler/src/transforms/babel/jsx.js#L40

So one possible (but pretty ugly) solution would be to add additional check if the contents of the asset include `from react/preact/etc` or `require('react/preact/etc)` to the condition above.

[...]

PR #2530, opened 2019-01-11
Enables jsx plugin in case jsx syntax is used in js files

lustoykov:
# ↪️ Pull Request

Enables usage of JSX syntax in `.js` files. There are two other alternative solution proposals apart from the current one (see https://github.com/parcel-bundler/parcel/issues/2519).

**1. Always enable JSX.** 
Opinion: It's a clean solution, especially in comparison to current one, but I think it would add overhead in terms of speed where jsx is not needed.
**2. Catch the error when babel tries to parse, enable `jsx` and re-build.**
Opinion: Babel fails with error code `'BABEL_PARSE_ERROR'` that is too generic to identify if it's a jsx problem or not. It's a possible alternative, but also feels a bit hacky.

Feedback appreciated, I intend to add tests if we agree on a solution. I'm not sure if this is a Good First Issue since the solutions we've come up with are not ideal.

## 💻 Examples

```js
import React from "react";
import ReactDOM from "react-dom";

const MyComponent = () => (
	<button>Hello</button>
);

ReactDOM.render(<MyComponent />, document.getElementById("root"));
```

JSX-syntax works now even in files with extension `.js`. For instance, `index.js`.

## 🚨 Test instructions

Create `index.js` file with above contents and run `parcel index.js`

## ✔️ PR Todo


mischnic:
> 🚨 Test instructions
> Create index.js file with above contents and run parcel index.js

More specifically: in an empty folder (no package.json), create a index.js ...

packages/core/parcel-bundler/src/transforms/babel/jsx.js
```diff
@@ -12,6 +12,26 @@ const JSX_PRAGMA = {
   hyperapp: 'h'
 };
 
+/**
+ * Solves a use case where people use JSX in .js files
+ */
+function isUsingJSXinJS(asset) {
+  // matches import * as React from 'react' and alike
+  const es6Candidate = /from\s+[`"'](react|preact|nervjs|hyperapp)[`"']/;
```
DeMoorJasper: Would this make sense to be just one regex?

That would simplify this code a lot

packages/core/parcel-bundler/src/transforms/babel/jsx.js
```diff
@@ -37,7 +57,11 @@ async function getJSXConfig(asset, isSourceModule) {
     }
   }
 
-  if (pragma || JSX_EXTENSIONS[path.extname(asset.name)]) {
+  if (
+    pragma ||
+    JSX_EXTENSIONS[path.extname(asset.name)] ||
+    isUsingJSXinJS(asset)
```
devongovett: Should we also use the regex to determine what pragma to use? Presumably this case is to detect if you haven't already installed one of those dependencies yet.

lustoykov: This ia a valid concern, I was thinking about it as well. Luckily, the pipeline is built in a way that it would, after creating ast in pretransform, re-collect dependencies, install them and re-iterate over that function which results in setting correct pragma.

On second thought, I've decided to implement your suggestion as it should add robustness.

packages/core/integration-tests/test/integration/jsx-preact-no-dep/index.js
```diff
@@ -0,0 +1,3 @@
+const Preact = require('preact');
```
lustoykov: not using import here for the sake of testing require syntax as well.

packages/core/parcel-bundler/src/transforms/babel/jsx.js
```diff
@@ -12,6 +12,26 @@ const JSX_PRAGMA = {
...
+  return new RegExp(
+    `from\\s+[\`"']${dependency}[\`"']|require\\([\`"']${dependency}[\`"']\\)`
+  );
+}
+
+/**
+ * Solves a use case where JSX is used in .js files or where
+ * package.json is empty yet and pragma can not be determined based
+ * on pkg.dependencies / pkg.devDependencies
+ */
```
lustoykov: the comment should be more clear, for instance

```
 * Solves a use case where JSX is used in .js files, but
 * package.json is empty or missing yet and pragma cannot be determined based
 * on pkg.dependencies / pkg.devDependencies
```

packages/core/parcel-bundler/src/transforms/babel/jsx.js
```diff
@@ -12,6 +12,26 @@ const JSX_PRAGMA = {
...
+}
+
+/**
+ * Solves a use case when JSX is used in .js files, but
+ * package.json is empty or missing yet and therefore pragma cannot
+ * be determined based on pkg.dependencies / pkg.devDependencies
+ */
+function maybeCreateFallbackPragma(asset) {
+  for (const dep in JSX_PRAGMA) {
+    if (asset.contents.match(createJSXRegexFor(dep))) {
```
devongovett: Maybe we should create these regexes ahead of time rather than recreating them for every file

lustoykov: +1 thanks for the hint, will fix tomorrow morning

packages/core/parcel-bundler/src/transforms/babel/jsx.js
```diff
@@ -24,9 +24,17 @@ function createJSXRegexFor(dependency) {
  * package.json is empty or missing yet and therefore pragma cannot
  * be determined based on pkg.dependencies / pkg.devDependencies
  */
+const cacheJsxRegexFor = {};
```
lustoykov: @devongovett, hope that goes in direction of your proposal. Regexes are lazily created in case needed and then cached (similarly to ahead of time creation).

PR #2506, opened 2019-01-04
upgrade htmlnano dep

devmao:
# ↪️ Pull Request

See issue #2505 

## ✔️ PR Todo

- [ ] Added/updated unit tests for this change
- [ ] Filled out test instructions (In case there aren't any unit tests)
- [ ] Included links to related issues/PRs

PR #2532, opened 2019-01-12
Fix htmlnano tests

mischnic:
# ↪️ Pull Request

The version upgrade in https://github.com/parcel-bundler/parcel/pull/2506 seems to have changed the output regarding `collapseWhitespace`. Setting that htmlnano option to `all`  would make current tests pass, but is not always safe ([see here](https://github.com/posthtml/htmlnano#collapsewhitespace))

## ✔️ PR Todo

### DIFF
--- a/packages/core/integration-tests/test/html.js
+++ b/packages/core/integration-tests/test/html.js
@@ -289,7 +289,7 @@ describe('html', function() {
       'utf8'
     );
     assert(html.includes('Other page'));
-    assert(!html.includes('\n'));
+    assert.equal(html.split('\n').length, 2);
   });
 
   it('should read .htmlnanorc and minify HTML in production mode', async function() {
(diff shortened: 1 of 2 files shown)


mischnic:
> Appears this doesn't fix the tests?

Strange, it does locally.
Only some commits on master after the mentioned PR fail due to the additional newline, but not all.
Also, that PR shouldn't have changed anything regarding testing, the lockfile already contained that version. 🤔

devongovett:
Yeah seems like it's still failing on windows

mischnic:
> Yeah seems like it's still failing on windows

Strange:
`<!doctype html><html><head><title>Test</title></head><body> <h1>Other page</h1> </body></html>`

DanielRuf:
Windows uses `\r\n` afaik.

millette:
Maybe use https://nodejs.org/dist/latest-v10.x/docs/api/os.html#os_os_eol in that case?

DanielRuf:
> Maybe use [nodejs.org/dist/latest-v10.x/docs/api/os.html#os_os_eol](https://nodejs.org/dist/latest-v10.x/docs/api/os.html#os_os_eol) in that case?

Definitely better and something what I would recommend. But this could fail in some cases but let's try it.

DanielRuf:
Will do it probably later this week, currently checking some projects and working on a few other CI fixes.

Issue #2274, opened 2018-11-10
Markdown Support


LandonSchropp:
# 🙋 Feature Request

I'd love to see markdown added to Parcel.

## 🤔 Expected Behavior

Ideally, I could create files with a `.md` extension, run `parcel serve source/posts/*.md` and I'd be done.

I'd also expect to be able to specify some kind of layout file for my Markdown. Most static site generators I've seen use front matter for this:

```
---
layout: ../layouts/layout.html
tags: awesome, blog
date: 2018-11-10
---

# My Awesome Blog Post

...
```

## 😯 Current Behavior

Currently, this doesn't seem to be supported in Parcel core. There are two plugins, [parcel-plugin-markdown](https://github.com/gongpeione/parcel-plugin-markdown) and [parcel-plugin-md](https://github.com/0326/parcel-plugin-md), but they don't really work how I'd expect them to.

## 💁 Possible Solution

I think it makes a ton of sense to add markdown support to Parcel directly.

## 🔦 Context

I really want to build a small static blog with Parcel. However, without Markdown support, this is extremely hard to do. Creating blogs that can be hosted on places like GitHub pages is very handy, and will be even more so once GitHub Actions launches to the public.

## 💻 Examples

The [Parcel website](https://github.com/parcel-bundler/website) repository uses a different generator to handle its Markdown.

kirillrogovoy:
Hi,

I think it's a cool idea for Parcel to treat Markdown files as entry points.

It could compile Markdown into HTML and then do whatever it does to HTML. I think it's not going to be hard to implement.

However, you mentioned layouts and front matter. In my opinion, Parcel shouldn't invent its own templating system for Markdown, so we'd have to use an existing solution.

@LandonSchropp Can you think of an existing tool that already does that — wraps a compiled Markdown file into an HTML template and forwards the attributes from the front matter to that template?

@DeMoorJasper If such a tool as described above existed and seemed to be actively maintained, so you think it's a good idea to build it into Parcel?

kirillrogovoy:
I couldn't find something like that by myself. 🤷‍♂️

Maybe we could indeed use a templating language like Pug or Mustache. With Mustache, a layout would look like this:
```html
<html>
  <head>
    <title>{{title}}</title>
  </head>
  <body>
    <div class="blog-post">
      {{{markdown_content}}}
    </div>
  </body>
</html>
```

Markdown file:
```md
---
title: "How to be"
layout: "./layout.html"
---

[...]

PR #2538, opened 2019-01-13
Add Markdown support as an entry point (issue #2274)

Benjmhart:
# ↪️ Pull Request
- added support for markdown entry points with dynamic installation of the marked package to parse the markdown code into html.
- added initial test for markdown parsing
- this is a partial fix toward ticket #2274 
https://github.com/parcel-bundler/parcel/issues/2274

## 💻 Examples

user can now run parcel build index.md, and get an index.html in the /dist folder

## 🚨 Test instructions

unit test is provided in the integration tests.   it confirms that the markdown file is parsed into html, and that image assets are registered/handled correctly.

## ✔️ PR Todo


packages/core/parcel-bundler/src/assets/MarkdownAsset.js
```diff
@@ -0,0 +1,11 @@
+const localRequire = require('../utils/localRequire');
+const HTMLAsset = require('./HTMLAsset');
+
+class MarkdownAsset extends HTMLAsset {
```
devongovett: I wouldn't recommend extending from another asset type. Instead, we can rely on the pipeline to pass through the generated HTML to the HTMLAsset.

Extend from `Asset` here, and return the html from the `generate` method. For an example, see how e.g. JSONAsset does it.

mischnic: Rather the PugAsset:
https://github.com/parcel-bundler/parcel/blob/d7dc51f9eacd80aadd1e41888771033e447758f1/packages/core/parcel-bundler/src/assets/PugAsset.js#L8

https://github.com/parcel-bundler/parcel/blob/d7dc51f9eacd80aadd1e41888771033e447758f1/packages/core/parcel-bundler/src/assets/PugAsset.js#L36

Benjmhart:
Hi there,   

So we've adjusted the MarkdownAsset to avoid extending the HTML Asset.  We did need to add a flag to make the pipeline behave properly.  Thank you for your useful examples in the JS and pug assets @devongovett  and @mischnic 

Please let us know if this is more in line with what you were looking for. 

Also, we wanted to discuss front matter support.  While the marked library appears to include support for basic front matter (mainly styling). the original feature request included mustache/templating specs.  We held off on this aspect of the feature because it makes parcel a bit opinionated on exactly how it should be done, and we wanted input from the core team before moving ahead there.

Thanks

packages/core/parcel-bundler/src/Pipeline.js
```diff
@@ -117,7 +116,9 @@ class Pipeline {
         type,
         value: asset.generated[type],
         // for scope hoisting, we need to post process all JS
-        final: !(type === 'js' && this.options.scopeHoist)
+        final: asset.hasOwnProperty('needsPipelineProcessing')
+          ? !asset.needsPipelineProcessing
```
devongovett: Why is this new property needed? Normally, if the `type` of the asset changes (e.g. from `.md` to `.html`), the pipeline will automatically pass it through to the next asset type (in this case `HTMLAsset`).

packages/core/parcel-bundler/src/assets/MarkdownAsset.js
```diff
@@ -0,0 +1,17 @@
...
+class MarkdownAsset extends Asset {
+  constructor(name, options) {
+    super(name, options);
+    this.type = 'html';
+    this.needsPipelineProcessing = true;
+    this.hmrPageReload = true;
+  }
+  async parse(code) {
+    let marked = await localRequire('marked', this.name);
+    this.contents = marked(code);
```
devongovett: Rather than reassigning `this.contents`, you could leave it as the original code and just return the result of `marked` from the `generate` method instead of `parse`.

packages/core/parcel-bundler/src/Pipeline.js
kirillrogovoy: Thanks for your comment!

> Normally, if the type of the asset changes (e.g. from .md to .html), the pipeline will automatically pass it through to the next asset type

Could you, please point out where that happens exactly? 

From what we discovered, the only way for an asset to undergo one more processing by another asset type is for this `final` property to be `false`.

https://github.com/parcel-bundler/parcel/blob/bdc044acfad84ba25bdb6631fbd3dd955424a227/packages/core/parcel-bundler/src/Pipeline.js#L55-L58

As you can see in the diff, the original code only allows `final` to be `false` for JS.

Maybe the right way to go would be to remove those 4 lines above altogether. 😱 But I suppose they are there for a reason.

Please, let me know if I've got it wrong.

kirillrogovoy: Ohhh, I've got it!

packages/core/parcel-bundler/src/assets/MarkdownAsset.js
kirillrogovoy: Thanks!

Please, see the latest version.

kirillrogovoy:
Hey, @devongovett!

Just a quick update.

Our plan:
1. Wait for you to take another look, and hopefully approve this PR.
2. Get it merged to the upstream. (Wow, we have basic but useless support for Markdown!)
3. Get your feedback on how the templating should work (https://github.com/parcel-bundler/parcel/pull/2538#issuecomment-455893028).
4. Implement whatever we agree on in a separate PR.

Sounds good?

kirillrogovoy:
Shamelessly pinging you guys! 🙂 
@devongovett @mischnic

tomfinney:
Is there a way to opt out of this or something? I was importing markdown files with a glob import like 

```
import foo from "/markdown/*.md";
// {
//   'file-1': '/file-1.8e73c985.md',
//   'file-2': '/file-1.8e73c985.md'
// }
```

Now, instead file paths for each markdown file, I get the entire contents of the markdown file parsed as HTML which is cool but the meta data for each md file is now in the HTML body.

kirillrogovoy:
Hey @tomfinney,

Thanks for reporting this.

I'll try to look into this soon.

1. Could you, please describe the sort of meta data you used and which is now inaccessible?
2. The old and new behaviors aside, what would be the *perfect* interface for your use case?

tomfinney:
Hi @kirillrogovoy 

I was using the YAML front matter style to prepend markdown files with some meta data based on: https://jekyllrb.com/docs/front-matter/ 

I have an example of one of my files here: https://github.com/tomfinney/website/blob/master/src/markdown/blogs/01_new_website.md

I'm not really sure if using the YAML style for storing meta data in a file is standard so I guess it would make sense that it would be parsed as HTML. It's just now when I import a markdown file, I have no way of splitting the metadata from the contents unless I add some special delimiter characters around them. 

Originally, I was fetching the contents of the markdown file and the splitting on the `---` and using a YAML parser to convert the meta data to an object.

So I guess the perfect interface for my use case would be that if I were to use a glob import, it would still return the keyed object of the file name to the file paths so I can choose how to process the files.

kirillrogovoy:
I see.

Would it work for you if `import` resulted in a keyed object containing the parsed metadata and the parsed HTML?

I'm planning on adding the front-matter support out of the box.

tomfinney:
I was just writing another comment about how that would actually be the best solution. I'm essentially just replicating front matter at the moment so if that was supported out of the box, it'd be perfect.

PR #2547, opened 2019-01-16
Add autorun for NodeJS builds

Ayc0:
# ↪️ Pull Request

Currently, if you want to create a plain NodeJS build, parcel is only building the JS files, not running them.

This PR adds the equivalent of `serve` but for NodeJS env.

And the script will be restarted at every build if `watch` is enabled.

## 💻 Examples

`parcel index.js --target=node` will also run index.js

## 🚨 Test instructions

I don't know how to test that the build is indeed run after the building phase.

## ✔️ PR Todo

### DIFF
--- a/packages/core/parcel-bundler/src/AutoRun.js
+++ b/packages/core/parcel-bundler/src/AutoRun.js
@@ -31,6 +31,9 @@ class AutoRun {
   }
 
   run(bundle) {
+    if (!bundle.name) {
+      return;
+    }
     this.prog = spawn('node', [bundle.name], {
       stdio: [process.stdin, process.stdout, process.stderr],
       detached: false
--- a/packages/core/parcel-bundler/src/AutoRun.js
+++ b/packages/core/parcel-bundler/src/AutoRun.js
@@ -1,50 +1,24 @@
-const {spawn} = require('child_process');
+const {fork} = require('child_process');
 
 class AutoRun {
   constructor(bundler) {
     this.bundler = bundler;
-    this.prog = null;
-
-    this.selfStoppedPIDs = {};
 
     bundler.on('bundled', async bundle => {
-      await this.clearPrevious();
+      if (this.previousRun) {
+        this.previousRun.kill();
+      }
       this.run(bundle);
     });
 
     bundler.bundle();
   }
 
-  async clearPrevious() {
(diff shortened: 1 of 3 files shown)


devongovett:
Very cool! Thanks for your contribution.

I am wondering about the auto-restarting. Do we want the program to always start again from the beginning every time you make a change? Perhaps we could support hot module replacement like we do in the browser for long running programs. If the program is a script which exits itself at the end, then it makes sense to run it every time, but for e.g. servers, you might be able to hot replace the modules while it continues to run. That would be more consistent with browser behavior, but not sure what you'd expect in all cases. Thoughts?

Ayc0:
I don't know how to enable hmr within node but I can check this out
I was already using parcel with this bit of code for a project that needed to be stopped so we should support both cases (maybe the if the hmr tag enabled or not)

devongovett:
I think we probably should support both options even for the browser (e.g. window.reload instead of HMR). Probably could be the same option.

mischnic:
> I think we probably should support both options even for the browser (e.g. window.reload instead of HMR). Probably could be the same option.

-> #289 


I've opened an issue for node HMR: #2549

Ayc0:
@devongovett at the beginning, I tried to clear entries from require.cache and then re-require them.

But as parcel is creating 1 bundle with every files inside, you have to require this file.

And let say you are listening to a port in your bundle, if we clear the previous file from the cache and re-require it, the port will still be used when the bundle is re-required. And so you'll have this error `Error: listen EADDRINUSE: address already in use`

I tried to use parcel's cache with `parcelRequire.cache`

Maybe there's something that we could do with `child_process.fork` but I don't know what

DeMoorJasper:
Feel free to reopen this for v2, we no longer update Parcel 1

Ayc0:
I started to work on an external plugin to add this behavior in parcel 2:
https://www.npmjs.com/package/parcel-reporter-node-runner

For now I added the initial basic support for `parcel watch`, but it doesn't support HMR for now

I'm gonna see if I can add support for it

Ayc0:
Btw, I tried to do `parcel serve ./src/index.js --target main`, but when using `serve` with a target that only builds for node, I have this error: `@parcel/core: Only browser targets are supported in serve mode`
image

Issue #1725, opened 2018-07-13
How can I import global scss file in Vue?


Jancat:
# ❔ Question
I'm building a Vue.js application and installed `node-sass` devDependency already.

I want to import global scss file(main.scss) in `main.ts` and all `.vue` component can use the global style directly. But it not works in the below code.

The **main.ts** file:
```
import Vue from 'vue'
import './styles/main.scss'   // import global app style file
import router from '@/router'
import App from '@/App.vue'

new Vue({
  router,
  render: h => h(App),
}).$mount('#app')
```

**App.vue**
```vue
<template>
  <div id="app">
    <router-view/>
  </div>
</template>

<style lang="scss">
// I had to import main scss file in every .vue component. Otherwise the `$vue-green` is undefined when bunddling.
@import './styles/main';

// the #app class hadn't apply to the root div element?
#app {
  font-family: '-apple-system', 'Avenir', Helvetica, Arial, sans-serif;
  -webkit-font-smoothing: antialiased;
  -moz-osx-font-smoothing: grayscale;
  text-align: center;
  background-color: $vue-green;
}
</style>
````

**/styles/main.scss**
```scss
@import 'variables';   // ./variables.scss defined the $vue-green
```

**problem:**
1. The **main.scss** hadn't been injected to the `.vue` component.
1. The **#app** style hadn't been applied to `<div id="app">` root element.

## 🔦 Context
How to import scss into Vue application properly?

## 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Parcel           |  1.9.4
| Node             | 10.5.0
| npm/Yarn         | Yarn
| Operating System | Mac OS

DeMoorJasper:
That will not inject it, it’ll create a new css bundle that would be added to an index.html if provided

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

DeMoorJasper:
@Jancat If I understand correctly than the sass-ressources-loader is basically just sass-node but for webpack and we support every sass feature so I guess we support it...

Not sure why that isn't working probably a good idea to dive into the output folder and see for yourself why it isn't working, some debugging from your side would definitely help. I can't guess what's going wrong

Jancat:
@DeMoorJasper I checked the  **main.css** in **dist** directory, and the `#app` name was converted to `#_app_e6llq_1`, failing to match `<div id="app">` in output **dist/index.html** . Bug?

[...]

PR #2558, opened 2019-01-20
Inject content at beginning of stylus files

shreeve:
# ↪️ Pull Request

Allows content to be injected at the beginning of `stylus` files.

This is especially helpful if you have global variables (commonly used in themes or branded versions of the same site), that you need to define for your stylus files. Instead of having to type `@import "variables"` (or similar) at the beginning of each `*.styl` file, you can simple provide the content to inject in your `package.json` file.

It seems that this is a common and long-standing request (see #1725) that has never been addressed. Granted, this solution is pretty basic of just injecting to all `stylus` files, but if more elaborate use cases are needed, they could easy be added based on the same idea. Given that this will inject into every `stylus` file, it's best to use it to just define variables or mixins that don't have side effects.

## 💻 Examples

```
{ ... this is your package.json file ...

  "stylus": {
    "inject": "@import '~/src/styles/app.styl'"
  }
}
```

## 🚨 Test instructions

Add an `inject` key to your top-level `stylus` key in `package.json`, and verify that it's working.

## ✔️ PR Todo

### DIFF
--- a/packages/core/parcel-bundler/src/assets/StylusAsset.js
+++ b/packages/core/parcel-bundler/src/assets/StylusAsset.js
@@ -5,6 +5,7 @@ const Resolver = require('../Resolver');
 const fs = require('@parcel/fs');
 const {dirname, resolve, relative} = require('path');
 const {isGlob, glob} = require('../utils/glob');
+const {EOL} = require('os');
 
 const URL_RE = /^(?:url\s*\(\s*)?['"]?(?:[#/]|(?:https?:)?\/\/)/i;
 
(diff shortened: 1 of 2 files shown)


DeMoorJasper:
@shreeve has this been standardised somewhere?
We should try to keep non standard config to a bare minimum and if this isn't a standard for less configs it might actually not be a good idea?

shreeve:
Looking at how webpack does it, for example, I can see that:

vue config js app 2019-01-21 14-51-17

There really is a difference between a list of _files_ to import, versus allowing the injection of actual _content_ at the beginning of each file. As a result, a key name of `inject` seems to convey more clearly that this particular content _is_ actually being injected at the beginning of the file.

In parcel, the `SASSAsset.js` file does something similar as this PR, but it uses the key name of `data`, which seems ambiguous.

In the end, I don't see a standardized way of doing this, but using a special `inject` key seems like a simple and understandable method.

shreeve:
I'm going to work on a set of changes, of which the `inject` is just one. As a result, closing this PR for now.

Issue #1126, opened 2018-04-02
serve --port default to $PORT?


jamiebuilds:
`$PORT` is a really common environment variable for specifying the port to use for web servers. 

We may want to consider defaulting to the value of `process.env.PORT || 1234`

https://github.com/parcel-bundler/parcel/blob/fc041d03c36bf41ea73c398e8d177de1a456e796/src/cli.js#L182

ilkerceng:
Hi @jamiebuilds 
I am getting this error when i try to deploy on heroku: 

> Error R10 (Boot timeout) -> Web process failed to bind to $PORT within 60 seconds of launch. 

Is it related with that issue? I couldn't fix it.

lizzzp1:
Could I take this issue?

foxt:
> $PORT is a really common environment variable for specifying the port to use for web servers.

Yes, it infact is - and you get very confused why parcel keeps trying to steal the $PORT you set for your application server  🙈

PR #2559, opened 2019-01-20
use env port

lizzzp1:
#1126 
~~Replaces `command.port` with `process.env.PORT`~~
Adds `process.env.PORT` as additional default

I _think_ it's ok to just replace? 😅 

## ✔️ PR Todo

Don't think test are applicable here.

### DIFF
--- a/packages/core/parcel-bundler/src/cli.js
+++ b/packages/core/parcel-bundler/src/cli.js
@@ -219,7 +219,7 @@ async function bundle(main, command) {
   command.target = command.target || 'browser';
   if (command.name() === 'serve' && command.target === 'browser') {
     const server = await bundler.serve(
-      command.port || 1234,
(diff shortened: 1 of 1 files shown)


mischnic:
> I *think* it's ok to just replace? 😅

It should be `command.port || process.env.PORT || 1234`, that way the cli flag will take precedence over the env variable

lizzzp1:
@mischnic thank you! -- Tests are failing. 🤔  Could it be a yarn/node version issue?

mischnic:
>  Tests are failing. 🤔 Could it be a yarn/node version issue?

These tests have been failing for a while, they are not related to your changes (I've tried to fix them in #2532 but Windows isn't working yet)

devongovett:
Thank you! 🙏

PR #2568, opened 2019-01-22
Parcel 2: ConfigResolver

devongovett:
This is the resolver for `.parcelrc` files for Parcel 2. It handles finding, validating, and merging configs, along with the `Config` class which handles loading and validating plugins.

* Resolves `.parcelrc` files from the project root directory.
* Validates all config properties.
* Resolves and merges config `extends` property supporting relative paths and npm package names.
* Checks for and validates `engines.parcel` field in package.json.

`@parcel/core` will not have a dependency on `@parcel/config-default`, which will keep it very small. Core supports either a `config` or `defaultConfig` option to pass an explicit config to override `.parcelrc`, or a default in case a `.parcelrc` isn't found. The CLI will pass `@parcel/config-default` as the default config, so if you are using the Parcel default CLI you will have a dependency on plugins, but not if other tools want to use `@parcel/core` with a custom config directly.

Closes #2568.

## Pipeline spread elements

The implementation follows the Parcel 2 RFC with one addition: support for spread elements (`...`) in pipeline configs. These allow you to extend an existing pipeline without needing to specify the whole thing again.

```json
{
  "transforms": {
    "*.js": ["parcel-transform-foo", "..."]
  }
}
```

The `"..."` element in a pipeline means to continue to the next matching pipeline rather than stopping. In this case, that means the glob matching "*.js" from `@parcel/config-default`. You can put a `...` anywhere in the pipeline to control the order, e.g. if you want the transform you are adding to go last, then `["...", "parcel-transform-foo"]`.

Also, the globs may not match exactly. You might want to have a glob that matches many file types, and override one of those to add an additional transform.

```json
{
  "transforms": {
    "*.jsx": ["parcel-transform-jsx", "..."],
    "*.{js,jsx}": ["@parcel/transform-js"]
  }
}
```

In this case, `parcel-transform-jsx` would get applied only for `.jsx` files, but `@parcel/transform-js` would still get applied due to the `...`.

To implement this, `ConfigResolver` merges pipelines where the globs match exactly (e.g. `*.js` specified in both configs), and `Config` merges pipelines where the globs do not match exactly (e.g. `*.jsx` and `*.{js,jsx}`).


padmaia:
I could see the `...` syntax being helpful, but I'm trying to think of ways it could backfire. Like what if you had wanted your custom plugin to run after a pipeline that has one plugin, but then the default pipeline adds another plugin. Would running your custom plugin after the new second default plugin break things or cause performance problems (maybe not being able to reuse the AST anymore)? Would we then consider any addition or removal of plugins to the default config a major bump? I think it may need to be thought out a bit more and we should make sure it's solving a legitimate pain point.

devongovett:
I think being able to extend a config to *add* one plugin without re-specifying an entire pipeline is going to be very important, and probably the most common use case for projects to have a `.parcelrc` file at all. I think it will be very rare for people to need to define an entire Parcel config from scratch. Therefore, I think we need to make that as easy as possible.

> Would running your custom plugin after the new second default plugin break things or cause performance problems (maybe not being able to reuse the AST anymore)?

It definitely shouldn't break anything, as plugins are expected to work independently of one another, in isolation, etc. They don't know about the other plugins that will run before or after, so order should be irrelevant. Yes, it is possible for the AST not be be reused anymore, but the API is designed to be resilient to that. It might cause perf issues, but at least it will still work properly. Reusing ASTs is an optimization, not a requirement.

> Would we then consider any addition or removal of plugins to the default config a major bump?

I don't think so. The `...` syntax means "run whatever the config I extended defined", not run a specific set of transformers. Adding features is not a breaking change according to semver. It would be like if a library you call in your app added a feature. As long as those changes wouldn't break things, then it should be a minor. If someone wants to avoid that problem, they can always pin to a specific minor version of `@parcel/config-default`.

padmaia:
> I think being able to extend a config to add one plugin without re-specifying an entire pipeline is going to be very important, and probably the most common use case for projects to have a .parcelrc file at all.

Yeah, but they should probably be inspecting the default config to figure out if it would be better to put it at the end or the beginning of the pipeline and at that point are you really saving that much work typing `...` instead of copy pasting? One downside to copy pasting would be missing out on improvements to a pipeline you customized, so that's not really perfect either. I think my issue with `...` is just that it makes it easier to unknowingly configure Parcel in a non optimal way. Pit of despair vs pit of success. If we provided some type of warning that you parsed the same type of AST twice, that might take care of my concerns. Might make sense to do that even if we didn't decide to go with `...` for some reason. So yeah, not 100% against it, just feel like it needs to be thought out more.

Another option might be to give a warning when an update of the default config changes on of the pipelines you extend, which would hopefully prompt you to take a look and make sure your plugin still works well in the pipeline. I feel like people might ignore that though or they might think they're fine when they're not. Providing warnings for unoptimized pipelines seems like the best safety net.

devongovett:
I'd say 99% of the time they will want their new plugin to go at the start of the pipeline to override the default behavior. If you need more nuanced behavior, you can always look at the default config to see how it works, but you'd need to do that whether or not we add the `...` syntax. I would like to optimize for the common case, which is `["my-fancy-plugin", "..."]`. More advanced usecases are still possible.

Also, ASTs are probably NOT going to be shared between most plugins. Most of the time they will compile to strings and re-parse. AST sharing was an optimization for our internal plugins mostly, which happen to use Babel ASTs for example. If you add another JS compiler like SweetJS, it is most likely going to have a different AST format than Babel. Even Terser has a different one, and that's a builtin plugin. So, warning when ASTs are not shared would cause almost every plugin to throw a warning, which wouldn't be a great experience.

padmaia:
Two more questions real quick? If the answer to both is yes, then I'm good on the PR.

1. We've talked about transformers a lot, but are we sure that all of the other array fields operate the same way (a plugin should never break from another plugin being added or removed)?
2. Are we sure that we will never want to have an array field where order matters? (Honestly that seems like a pretty good restraint)

devongovett:
> I meant more like detecting that the AST went from a type "foo" to type "bar" and then back to "foo".

Ah I see. That's a good idea. 👍 

> You seem really confident that this is the best way to go

Well... I think there should be a way to do what I'm describing, I'm not sure it's the best syntax. 😉 

 > We've talked about transformers a lot, but are we sure that all of the other array fields operate the same way (a plugin should never break from another plugin being added or removed)?

I think so? Resolvers override other resolvers, optimizers chain like transformers, runtimes should work independently, ...

> Are we sure that we will never want to have an array field where order matters? (Honestly that seems like a pretty good restraint)

Order matters for things that override, e.g. resolvers. When a resolver returns, it aborts the rest of the resolvers in the array. I think the `...` is pretty clear about order though since you can control which position you place it in. Yeah, it doesn't cover all possible orders, but it at least solves the simple cases.

PR #2578, opened 2019-01-25
Allow externals -- POC for #144

moos:
# ↪️ Pull Request

Does not bundles modules in package.json's `peerDependencies`.   Possible solution for #144.  (Note commit message erroneously refs 114!!)

## 💻 Examples
package.json:
```
  "peerDependencies": {
    "prop-types": "^15.6.2",
    "react": "^16.7.0",
    "react-dom": "^16.7.0",
    "react-intl": "^2.8.0"
  },
```
Then:
```
$ npm run build  # parcel build

dist/foo-bar.js                         38.47 KB    96ms
├── node_modules/react-idle-timer/dist/index.es.js         6.46 KB     1ms
├── src/sessionTimeout/index.js        3.68 KB     3ms
...
```
Notice no `react`, `react-dom`, etc. in the final bundle.

`parcel start` (HMR) will still pull those modules in.

## ✔️ PR Todo


mischnic:
This should also enable parcel to bundle libraries which shouldn't contain React for example.

packages/core/parcel-bundler/src/Bundler.js
```diff
@@ -566,6 +570,14 @@ class Bundler extends EventEmitter {
 
     // Call the delegate to get implicit dependencies
     let dependencies = processed.dependencies;
+
+    // exlucde externals
+    if (this.externals.length && !this.options.hmr) {
```
mischnic:
```suggestion
    if (this.externals && this.externals.length && !this.options.hmr) {
```

To fix a test: `Cannot read property 'length' of undefined`

devongovett:
I don't think peerDependencies can be used as externals. peerDependencies are used to prevent multiple versions of a library from being installed by npm. We shouldn't extend its meaning. If we add this feature, it should be a separate configuration in package.json.

That said, I'd like to hold off on this until Parcel 2. We need to focus our development energies there, so we won't be adding any new features to Parcel 1 at this point. If you'd like to discuss how externals could be implemented in Parcel 2, please open an RFC!

PR #2579, opened 2019-01-26
WIP: Parcel 2 server

### LINKED ISSUE #2144: Parcel 2: Server

DeMoorJasper:
# ↪️ Pull Request

I did some fiddling around with the server and hmr socket for v2
The Server was fairly straightforward and should work pretty much entirely now.

HMR on the other hand is probably gonna need a lot of tweaking due to how assets work in Parcel 2?

I could probably use some guidance on how to do this with the asset graph as it probably needs some kind of diff between previous state and new state, which is not available atm afaik?
I currently implemented some kind of state on hmrServer which gets pushed to the client on `buildEnd` which is just a new event I added to AssetGraphBuilder. It doesn't seem to me like the proper way to do this.

I also added a helper function for normalising cli options, I think this will simplify using these options in multiple places a little?

Closes #2144

## ✔️ PR Todo


packages/core/core/src/AssetGraphBuilder.js
```diff
@@ -166,11 +179,27 @@ export default class AssetGraphBuilder extends EventEmitter {
       }
     }
 
+    if (this.hmrServer) {
+      this.hmrServer.updateAsset(cacheEntry);
```
devongovett: A cache entry can contain multiple assets, so probably need to split them out here?

packages/core/core/src/AssetGraphBuilder.js
```diff
@@ -71,6 +73,10 @@ export default class AssetGraphBuilder extends EventEmitter {
         }
       });
     }
+
+    if (opts.cliOpts.hot) {
+      this.hmrServer = new HMRServer(this, opts.cliOpts);
```
devongovett: I am wondering about potentially keeping the HMRServer completely separate from the AssetGraphBuilder, and keep that just for building the graph. Possibly it could live in Parcel instead, which is the glue code between all the pieces?

packages/core/core/src/Parcel.js
```diff
@@ -25,20 +29,39 @@ type ParcelOpts = {
...
     this.options = options;
     this.entries = Array.isArray(entries) ? entries : [entries];
     this.rootDir = getRootDir(this.entries);
+
+    this.normaliseCliOptions();
+  }
+
+  normaliseCliOptions() {
+    this.options.cliOpts.cacheDir =
+      this.options.cliOpts.cacheDir || DEFAULT_CACHE_DIR;
```
devongovett: I kinda liked doing this normalization in the individual places rather than all in one place... What do you think? Why did you move them here?

DeMoorJasper:
Gonna write a new PR for this as this has gotten very out of sync with v2

PR #2610, opened 2019-01-31
Ensure css files are written before hmr is triggered

garthenweb:
# ↪️ Pull Request

The project I am building is generating a big CSS file (~20k lines of code/ 500kb file).
When generating this file in development the hot reload is triggered before the file is completely generated, it will only load the first ~2000 lines of the code. Triggering the update again by hand (finding the link tag within the dom and updating the number after the `?` parameter) is working.

Moving the emit of the hot reload after `this.mainBundle.package` call solves the issue.

## 💻 Examples

I cannot really show an example because I don't have a public project with that many lines of code. I guess it also depends on the filesystem speed to reproduce this (my filesystem is actually slow because of Windows/ WSL).

## 🚨 Test instructions

My setup uses SCSS, CSS Modules and Typescript. Let me know if there are more things you are interested in.

### DIFF
--- a/packages/core/parcel-bundler/src/Bundler.js
+++ b/packages/core/parcel-bundler/src/Bundler.js
@@ -298,11 +298,6 @@ class Bundler extends EventEmitter {
         asset.replaceBundleNames(this.bundleNameMap);
       }
 
-      // Emit an HMR update if this is not the initial bundle.
-      if (this.hmr && !isInitialBundle) {
-        this.hmr.emitUpdate(changedAssets);
-      }
-
       logger.progress(`Packaging...`);
 
       // Package everything up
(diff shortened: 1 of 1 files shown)


garthenweb:
@DeMoorJasper Thank you for the feedback, had actually a hard time understanding the logs in the pipeline and was not sure if its me or the "normal" issues. Will try to check and fix those on the weekend.

garthenweb:
@DeMoorJasper Just checked why the tests failed: The code in the bundler after `this.hmr.emitUpdate` runs completely synchronously now, therefore the event loop schedules the `assert.deepEqual` calls before `outputs.push(o)`.
Splitting the call stack by inserting a `setTimeout` allows to execute the update first and then executed the assertion.

PR #2626, opened 2019-02-10
temp disable debug option - until elm compiler issue is fixed

Maxim-Filimonov:
See https://github.com/parcel-bundler/parcel/issues/2104#issuecomment-461707960
Waiting for https://github.com/elm/compiler/issues/1851 to be resolved.

### DIFF
--- a/packages/core/integration-tests/test/elm.js
+++ b/packages/core/integration-tests/test/elm.js
@@ -14,7 +14,8 @@ describe('elm', function() {
     let output = await run(b);
     assert.equal(typeof output().Elm.Main.init, 'function');
   });
-  it('should produce a elm bundle with debugger', async function() {
+  // Waiting for the elm compiler bug to be fixed https://github.com/elm/compiler/issues/1851
+  xit('should produce a elm bundle with debugger', async function() {
     let b = await bundle(__dirname + '/integration/elm/index.js');
 
     await run(b);
--- a/packages/core/parcel-bundler/src/assets/ElmAsset.js
+++ b/packages/core/parcel-bundler/src/assets/ElmAsset.js
@@ -39,7 +39,9 @@ class ElmAsset extends Asset {
       await this.getConfig(['elm.json'], {load: false});
     }
 
-    options.debug = !this.options.production;
+    // Disabling for now due to a bug in the Elm compiler:
+    // https://github.com/elm/compiler/issues/1851
+    // options.debug = !this.options.production;
     if (this.options.minify) {
       options.optimize = true;
     }


domenkozar:
@kyasu1 do you have  `Attribute` inside a `Msg`? Or something else like `Array`?

I don't think it makes sense to disable a feature due to a bug in compiler in a very edge case. 

I do think we should allow such users to opt-out, maybe with additional flag/setting to disable debugger?

domenkozar:
There is also a possible fix in https://github.com/elm/compiler/pull/1850 that needs testing.

kyasu1:
@domenkozar I myself not using such Msg types but the dependency packages might be. I think this bug arises pretty common when projects are getting complex and large. 

I can't test the compiler fix you mentioned, out of my skill , but it may take long time until fully tested and merged. For that period we are no able to use `parceljs` for developing if we experience that bug.

So I think disabling this option temporally is much easier and feasible at the moment. And I more like debugging with `Debug.log` than the Time travel debugger...

domenkozar:
Note that you can downgrade to 1.10.3 for now, which didn't have debugger on. You'll need to use https://www.npmjs.com/package/parcel-plugin-elm-hot for HMR, but that's easy to wire up. Does that resolve your problem?

kyasu1:
Thank you for telling me about the workaround, it works perfectly ! Actually this is not the issue of `parceljs`, so I do not insist to apply this patch. Hopefully the bug in `elm/compiler` will be fixed soon...

sklirg:
Would it be possible to add a feature flag for this instead? So users can opt in/out if this at will. E.g. an env variable `PARCEL_ELM_DEBUG=bool` or a CLI flag like `--disable-debug`.

Maxim-Filimonov:
@sklirg what's the use case of disabling it ? Beyond the compiler issue.

sklirg:
That's exactly the use case. I'd like my code to compile without having to downgrade parcel.

Currently I monkey-patch it locally by editing the file in node modules...

lzrski:
The env variable solution suggested by @sklirg (https://github.com/parcel-bundler/parcel/pull/2626) sounds very reasonable. There is no need to extend the CLI or even document it anywhere. It can be retired as soon as Elm compiler is fixed (which can take a long time). 

On the other hand there may be legitimate reasons to disable debugger anyway - e.g. I'm working on an app that subscribes to animation frames all the time and I'm getting hundreds of thousands of messages that take up memory and are not very useful anyway. In this case I would be happy to disable debugger even with the compiler fixed.

Punie:
The way the elm debugger works at the moment is kinda annoying in some cases (even without accounting for the compiler bug).
It injects a `div` in the page that can cause some layout issues that can be a pain to debug.

I would be strongly in favour of a cli flag for disabling the debugger, as that is something that some people might want to do event after the compiler bug is resolved.

sklirg:
I opened #2818 with my fix since some of you guys seemed interested in it. 👍

Maxim-Filimonov:
@domenkozar Just thinking out loud about this. 
Would you be open to considering the possibility of allowing to pass arguments through parcel to a compiler used? 
[Similar](https://github.com/wking-io/elm-live#adding-the-debugger-during-dev) to elm-live approach to passing extra arguments to elm compiler.

This way precise customization of tool can still be possible via additional parameters without need to change parcel's code.

DeMoorJasper:
Closing in favor of #2818

Issue #2039, opened 2018-09-20
Tree shaking react-router-dom


hannupekka:
# 🐛 bug report

If I enable `--experimental-scope-hoisting` on my build, seems that when I import with:

```js
import { BrowserRouter as Router, Switch, Route } from 'react-router-dom';
```

everything resolves as `undefined`.

Example repo here: https://github.com/hannupekka/photogallery-example - build with `yarn build`

## 🎛 Configuration (.babelrc, package.json, cli command)

https://github.com/hannupekka/photogallery-example/blob/master/.babelrc
https://github.com/hannupekka/photogallery-example/blob/master/package.json

CLI command:
```js
parcel build src/index.html --experimental-scope-hoisting
```

## 🤔 Expected Behavior

Application should work

## 😯 Current Behavior

Everything from `react-router-dom` is `undefined`, resulting in https://reactjs.org/docs/error-decoder.html/?invariant=130&args[]=undefined&args[]=

## 💻 Code Sample

https://github.com/hannupekka/photogallery-example - build with `yarn build`

## 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Parcel           | 1.9.7
| Node             | 9.0.0
| npm/Yarn         | 1.3.2
| Operating System | macOS 10.13.6

mischnic:
I only get: "$WkSz$var$react_pose_1.default.div is not a function"

mischnic:
Quite a strange bug you've discovered:

This
```js
import posed, { PoseGroup } from 'react-pose';

console.log(posed.div);
console.log(PoseGroup);
```
will log two functions when compiled without tree shaking but `posed.div` will be undefined if treeshaken. This only happens in Typescript. Rename `index.ts` to `index.js` and it works fine:
Minimum example: https://github.com/mischnic/photogallery-example

mischnic:
@hannupekka Removing the module option in your tsconfig solves the issue
```diff
diff --git a/tsconfig.json b/tsconfig.json
index 68257ce..3a75a22 100644
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,7 +1,6 @@
 {
   "compilerOptions": {
     "target": "es5",
-    "module": "commonjs",
     "strict": true,
     "jsx": "react",
     "noImplicitAny": true,
```


[...]

Issue #1378, opened 2018-05-16
Better TypeScript support


mohsen1:
# 💬 RFC Better TypeScript support 

TypeScript is working out of the box for the most part but there are a few pain points:
 ### Pain points
* [TypeScript module resolution algorithm](https://www.typescriptlang.org/docs/handbook/module-resolution.html) is not implemented in Parcel. #202 This covers these configs:
  * `baseUrl`
  * `paths`
  * `allowSyntheticDefaultImports`
* Compiler errors are not printed in the console #465
  * `parcel-plugin-typescript` adds it but console is not cleared after issue is resolved 

### Improvements
* Automatically create and configure `tsconfig`
  * Once a `ts` or `tsx` file was imported but `tsconfig` is not present, create one
* Automatically install `@types` packages #1161
* Support for Custom Typescript transformers #699

## 🔦 Context
TypeScript is pretty popular and having first class support for it in Parcel will attract a wider range of users. 

## 💻 Examples
I tried cloning [react-parcel-example](https://github.com/jaredpalmer/react-parcel-example) and making repo states that demonstrate those pain points:

1. JSX is not working in default mode (preserve): 
    * https://github.com/mohsen1/react-parcel-example-typescript/tree/90f9f72b2f3fcfcbd5410b097e5c20cc10d8a325
2. `baseUrl` config is not working 
    * https://github.com/mohsen1/react-parcel-example-typescript/tree/3b74951e47d77d7ba79dbd56ef25621d312ca394

fathyb:
Duplicates : 
- module resolution #202
- type checking #465 
- auto install `@types` #1161

---

`parcel-plugin-typescript` implements the TypeScript module resolution (`baseUrl`, `paths`). `allowSyntheticDefaultImports` is not part of the module resolution algorithm, it's used for the module *symbol* resolution by the type-checker. If you are using TypeScript 2.8+ you should be able to import synthetic defaults.

The plugin is originally a spin-off of a PR implementing type-checking, but the feature wasn't well received by the community, so I guess it's better to leave the choice to the end user.

The TypeScript plugin will probably be integrated as a core plugin once we settle on a plugin RFC though (aka `2.0`). The long term goal would be to integrate TypeScript just like we do with Babel, and let developers use a pure TypeScript toolchain without Babel.

szagi3891:
@DeMoorJasper - When could this happen?

uoziod:
+1

DeMoorJasper:
[...]

PR #2627, opened 2019-02-10
Parcel 2: TypeScript Transformer

DeMoorJasper:
# ↪️ Pull Request

TS Transformer ported from Parcel 1 to Parcel 2

Not sure if we actually wanna implement it this way though, as Babel now supports TS and it would probably be a bit faster to create and transform the ast using babel/ts

Perhaps we should have both?

## ✔️ PR Todo


mischnic:
> TS Transformer ported from Parcel 1 to Parcel 2
> 
> Not sure if we actually wanna implement it this way though, as Babel now supports TS and it would probably be a bit faster to create and transform the ast using babel/ts

One major misconception in the past was that Parcel doesn't do typechecking, I guess this won't be changed?

DeMoorJasper:
@mischnic Parcel doesn't do typechecking. However it would be cool to have this feature in Parcel 2, that's the only reason I can think of to use ts-compiler

devongovett:
@fathyb what do you think? I know people have been asking for [better typescript support](https://github.com/parcel-bundler/parcel/issues/1378), and one of those asks is support for actual type checking. I kinda think that type checking and linting (e.g. eslint) could be handled in a similar way: we compile synchronously, but type check/lint in the background. That way you don't have to wait as long for builds when you're just hacking, but have the type checking/linting information available if you want it. Thoughts?

mischnic:
Coming from #2039, Parcel 1 basically ignored the `target` and `module` option in tsconfig.json.
I think it would be better for performance and DX to not run Babel after Typescript (by default), so that these option actually work (at least `target`, no sure about `module`).

timfish:
Just be aware that if you use Babel with separate type checking, there is currently no support for:
- [`const enum`](https://github.com/babel/babel/issues/8741)
- `emitDecoratorMetadata` not supported out of the box

PR #2632, opened 2019-02-11
Parcel 2: CLI and reporter plugins

devongovett:
This PR implements the default Parcel 2 CLI, along with support for reporter plugins and the default CLI reporter. 

## CLI

The CLI is mostly backward compatible with Parcel 1, minus a few options which have been removed. I unified the options object that the CLI passes with the `ParcelOptions` type accepted by the `Parcel` API. The CLI normalizes options before passing them to the API. `@parcel/core` no longer has a dependency on `@parcel/config-default`. The CLI passes the default config as an option to the Parcel API. This keeps the core quite small.

## Reporters

The bulk of this PR deals with reporters and the default reporter plugin. Reporters receive events from Parcel as they happen and can decide what to do with them. The events currently available are:

* `buildStart` - emitted when the build is about to start.
* `buildProgress` - emitted periodically throughout the build. There are several types of progress events, which correspond to the build phases of Parcel: `resolving`, `transforming`, `bundling`, `packaging`, and `optimizing`. Each of these events has an event object which provides some details about what is happening for the reporter to use as needed. The default reporter displays a message with a spinner in the CLI.
* `buildSuccess` - emitted when the build is successful. Includes the asset graph and bundle graph that were generated, along with the build time. The default reporter displays a bundle report with the largest bundles and assets that were generated in production mode.
* `buildFailure` - emitted when the build fails. Includes the error that occurred which the default reporter displays in the CLI.
* `log` - emitted by the `@parcel/logger` package which is now just a shared event emitter. Includes the logged message and log level. The default reporter decides whether to display the logs by their level and the `--log-level` CLI option.

The default reporter is written using [ink](https://github.com/vadimdemedes/ink/tree/next), which is a react renderer for the CLI. It makes writing CLI UIs that update in place rather than just appending newlines quite simple. As each event comes into the reporter, a reducer runs to compute the next state of the UI, which react/ink takes care of updating.

## Build Stats

In order to implement the bundle report for production builds, I needed to add some stats to asset and bundle objects for their build time and output size. This brought up the question of how we measure these stats.

* Currently the "build time" of an asset is the full time it took to transform the file it came from. This was the easiest way to measure it, and similar to Parcel 1, however not that accurate necessarily. Since there could be many assets in a file (e.g. `.vue`), do we try to measure this per asset rather than per file or is per file good enough?
* The "build size" of an asset is the output size at the end of the transform phase for the asset. Unfortunately, this is prior to packagers and optimizers running, so when displayed as part of a bundle report, it tends to make assets appear quite large and not "add up" correctly to the full bundle size. In Parcel 1, we tried to measure the real bundled size of an asset based on the size of the bundle before and after adding the asset to the packager. In Parcel 2, this isn't really possible since we pass all assets at once, and also optimizers run after packagers. The original size of the file may have to be good enough...
* The "build time" of a bundle is the amount of time it took to package and optimize the bundle. This is different than in Parcel 1, which tried to guesstimate the total build time of a bundle based on the build time of the assets it contained. This was never really correct since an asset could appear in multiple bundles, so adding up the bundle build times didn't match the full build time.

These stats are kinda hard to measure as you can see, and everything depends on how you count. I'm not sure what makes the most sense here. If anyone has ideas, please let me know!


packages/core/parcel/package.json
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "parcel",
+  "version": "2.0.0",
+  "license": "MIT",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/parcel-bundler/parcel.git"
+  },
+  "bin": {
+    "parcel2": "src/bin.js"
```
devongovett: Just for now, until we release. That way you can have both Parcel 1 and 2 on your system at the same time.

packages/core/parcel/src/bin.js
```diff
@@ -0,0 +1,2 @@
+#!/usr/bin/env node -r @parcel/babel-register
```
devongovett: TODO: this package should be pre-compiled

packages/core/parcel/src/cli.js
```diff
@@ -0,0 +1,180 @@
...
+import type {ParcelConfig, ParcelOptions} from '@parcel/types';
+const chalk = require('chalk');
+const program = require('commander');
+const version = require('../package.json').version;
+const path = require('path');
+
+program.version(version);
+
+// --no-cache, --cache-dir, --no-source-maps, --no-autoinstall, --global?, --public-url, --log-level
+// --no-content-hash, --experimental-scope-hoisting, --detailed-report
```
devongovett: These options were in Parcel 1, but idk what to do with them

padmaia:
Sorry it has taken me so long to review this. A couple of comments:

> Reporters receive events as they happen and can either use the Parcel logger to output to stdout/stderr or they can return assets to be generated on the file system.

1. This implementation seems to deviate from the Parcel 2 RFC. I like your current implementation because there is more control over the output, which would be useful for users of the programmatic API. The logger becomes more of a collector of logs, which the reporters can decide what to do with. Seems like a good model to me, but I just wanted to make sure we are intentionally deviating from the spec and call that out. Also wasn't 100% sure what was meant by reporters returning assets, but I'm guessing that just means they might output files, correct? Would the creation of those files be handled by the reporters themselves or does returning assets imply the creation would be done by the ReporterRunner or something else in core?

2. Back to the logger collecting log events, it wasn't super clear to me why we would also need the bus. It seemed like you were using report action which uses the bus for specific build events, but it also seemed like something that could still be handled by the logger, and maybe was in the past with the "progress" method? Is there a compelling reason for the bus that I am missing? Also, we had talked about the logger not depending on the WorkerFarm. A big reason for that was to deal with a circular dependency (WorkerFarm -> utils -> logger -> WorkerFarm) which we are now avoiding by using deep imports of the utils (works for now), but I think it still helps to untangle things. If we removed the WorkerFarm dependency from the logger, then it could have a single responsibility of collecting logs. The worker file could be responsible for detecting if it is in a worker process and sending captured logs to the main process logger when necessary.

3. #nitpick of the day - I noticed some inconsistencies naming test files, some have extension `.test.js` while others have just `.js`. I don't care what we choose, but we should be consistent.

padmaia:
Oh yeah, also tests are broken :P

devongovett:
1. Yes, it does deviate slightly from the RFC, but the RFC wasn't very well defined in the first place, and the change was intentional. I also wasn't sure what returning assets would do - I guess generate files. It seems fine to me to let reporters just do this directly with normal fs APIs rather than delegate back to Parcel core to do it.

2. Ahhh the bus. Well, basically the bus exists to ferry events from worker processes to the main process so that reporters can access them. You are right that the logger also does this, but I wanted to keep the logger API for public consumption by plugins, and emitting reporter events other than logs should only be done in parcel core. It would be nice to remove the WorkerFarm dependency from the logger, but we need it in order to send the events back to the main process. Perhaps since the logger is just an event emitter now we could listen for events in the worker itself and send them over the bus in core instead.

3. Good catch. 😉

wbinnssmith:
Hi All,

Just merged v2 into this branch and resolved things :)

The tests now pass as well. I made some changes to get things going:

* Since this contains a major architectural change to Logger, I bumped the logger package to 2.0 and fixed `parcel-bundler` on `1.10`. It's a lot of work to rearchitect everything that relied on it to use the new logger and reporter infra.
	* This could introduce some cognitive overhead understanding that old packages use the older logger, but aside from the test suite I don't think anything in the v2 code path uses these?
* Moved logger to *have* an `EventEmitter` rather than *be* one. This makes for easier flow typing for consumers (`on` would otherwise yield `any`), and keeps our api surface for it minimal in the case it ever gets exposed to plugins.
* Splits `ParcelConfig` into another type -- `ParcelConfigFile` which represents the contents we're going to read from disk (doesn't have `filePath`, etc.) This should catch situations where we assumed we'd have a filePath but we didn't.
* Makes the patching of `console.log` and friends opt-in. This messed up a few things like mocha.
* Creates a proxy to `stdout` for ink. While running `yarn test` from the root of the repo with lerna, `process.stdout` does not have a `columns` property, which throws off ink. This makes sure one is there.
	* We might want to use a different reporter entirely in test. Having ink constantly clear lines (sometimes the whole screen) is pretty annoying. I assume it respects CI not being a tty though.

Going to leave final approval to @padmaia and @DeMoorJasper.

devongovett:
Thanks @wbinnssmith! LGTM. 🎉

Issue #2592, opened 2019-01-29
Use of CSS Modules "composes:" results with duplicate classes exports


wajerrr:
# 🐛 bug report

Using 'composes' with CSSmodules results with duplicate styles exported:

## 🎛 Configuration (.babelrc, package.json, cli command)

package.json
```js
{
  "name": "parcel-cssmodules-example",
  "version": "1.0.0",
  "description": "",
  "main": "index.js",
  "scripts": {
    "start": "parcel --no-cache index.html",
    "build": "parcel build --no-cache index.html"
  }
  ,"postcss": {
    "modules": true
  },
  "author": "",
  "license": "ISC",
  "devDependencies": {
    "parcel-bundler": "^1.11.0",
    "postcss-modules": "^1.4.1"
  }
}

```
## 😯 Current Behavior // Example 

mixin.css
```
.test {
    height: 100px;
    width: 100px;
    ...
}
```

composes-1.css
```
 .main {
    border: 3px solid orange;
    composes: test from './mixins.css';
}
```

composes-2.css
```
 .main {
    border: 3px solid red;
    composes: test from './mixins.css';
}
```

composes-3.css
```
 .main {
    border: 3px solid brown;
    composes: test from './mixins.css';
}
```

example.js 
```
import * as composes1 from './composes-1.css';
import * as composes2 from './composes-2.css';
import * as  composes3 from './composes-3.css';

window.onload = () => document.body.innerHTML = `
    <div class="${composes1.main}"></div>
    <div class="${composes2.main}"></div>
    <div class="${composes3.main}"></div>
`;
```

will  result with `output.css` having `.test (._test_giiwg_1)` class exported 3 times

```
._test_giiwg_1{
     /* styles from .test in mixins.css */
}
._main_1la6g_1{
    border:3px solid orange
}
._test_giiwg_1{
 /* styles from .test in mixins.css */
}
._main_1655g_3{
    border:3px solid red
}
._test_giiwg_1{
     /* styles from .test in mixins.css */
}
._main_hdlwj_1{
    border:3px solid brown
}
```
## 🤔 Expected Behavior

`.test  (._test_giiwg_1)` class exported only once

output.css
```
._test_giiwg_1{
     /* styles from .test in mixins.css */
}
._main_1la6g_1{
    border:3px solid orange
}
._main_1655g_3{
    border:3px solid red
}
._main_hdlwj_1{
    border:3px solid brown
}
```

## 💻 Code Sample

[...]

PR #2642, opened 2019-02-13
Fix postcss modules composes imports

### LINKED ISSUE #2501: CSS Modules not compatible using composes imports with SASS
# 🐛 bug report

When using css modules with `composes` from another scss file, the imported file is not transpiled using sass. Everything is working as expected when using `composes` inside the same file.

``` scss
// inline.scss
$primary-color: green;

.test {
    height: 100px;
    width: 100px;
    background: $primary-color;
}

// composes.scss
.main {
    composes: test from './inline.scss';
}
```

``` js
// example.js
import * as composes from './composes.scss';

window.onload = () => document.body.innerHTML = `
    <div class="${composes.main}"></div>
`;
```
``` json
// .postcssrc
{
  "modules": true
}
```

``` json
// .sassrc
{
  "includePaths": [
    "node_modules"
  ]
}
```

``` bash
parcel build example.js
```

Please see https://github.com/garthenweb/parcel-css-modules-example for more information.

## 🤔 Expected Behavior

It should resolve the sass variable `$primary-color` to css. Expected css output:

``` css
._test_dek43_5{height:100px;width:100px;background:green}
```

## 😯 Current Behavior

The variable `$primary-color` is not transpiled. Current Output:

``` css
$primary-color:green;._test_dek43_5{height:100px;width:100px;background:$primary-color}
```

There is no error from Parcel, but it returns invalid CSS, therefore the browser will not render a rectangle.

## 💁 Possible Solution

I am not 100% sure, but I guess that parcel is first running sass and afterward runs Postcss including CSS Modules. But Postcss does not transpile imported styles from sass to CSS. I assume it could work when we run sass after the Postcss transform.

I already tried to use the [`@csstools/postcss-sass`](https://www.npmjs.com/package/@csstools/postcss-sass) plugin to do so but it threw an error.

Maybe we could use the syntax parameter form the Postcss config to identify that a transform should be followed after Postcss like when the [`postcss-scss`](https://www.npmjs.com/package/postcss-scss) syntax is applied. But there might be edge cases, not sure for now...

I would love to help solving this, but I need some guidance on where to start. It would be nice to receive some tipps from the core team where to start.

## 🔦 Context

I try to migrate an existing project from Webpack to Parcel where the composes pattern is used quite a lot. The Webpack config is doing the transform by adding multiple loaders, I didn't look too much into details there but in case its helpful to solve this issue I can create an example.

## 💻 Code Sample

I created an example project to demonstrate the behavior:
https://github.com/garthenweb/parcel-css-modules-example

## 🌍 Your Environment

| Parcel           | 1.11.0
| Node             | 10.14.2
| npm       | 6.4.1
| Operating System | Windows 10 using WSL/ Mac OSX

garthenweb:
# ↪️ Pull Request

The PR fixes different issues with postcss composes when using the import syntax, e.g. `composes: my-class from './another-file.css';`:

* importing the same file in different dependencies lead to duplicate CSS rules (#2592)
* importing other dependencies as CSS, e.g. sass files, the asset was not transpiled (#2501)
* importing files using the parcel paths, e.g. `composes: my-class from '~another-file.css`  throws an error because the file was not found (no ticket available)

Sorry for the big PR, I started fixing #2592 and found out that the issues are really closely related and it would not make sense to solve them separately (I drifted into a wrong direction when only thinking about one issue). Anyhow, all fixes are within separate commits (see the first 3).

Fixes: #2592, fixes #2501

## 💻 Examples

See the example in #2592 and #2501.

## 🚨 Test instructions

See the example in #2592 and #2501. I also tried to cover everything in integration tests.

## ✔️ PR Todo

- [x] Test in bigger projects (with more edge cases, I just checked the example in #2592 for now) 
- [x] Test if it's working in combination with other Postcss transforms (CSS classes generated with a hash might be different in the dependency after parsed by parcel)


garthenweb:
@DeMoorJasper @devongovett I consider this one ready now, I am happy to receive feedback!

There is still one test failing, as far as I can see it is not related to my changes and also fails in master at the moment, therefore, I ignored it for now.

Some further notes I would like to share:
I tried to work around the custom Postcss Loader and the default class name in the first place, but had a hard time for multiple reasons.
* By default the Loader adds the imports on top of the asset. This introduced the duplicate CSS rules as it is not aware of other files.
* The imported CSS will not be transpiled in case its e.g. sass or less file. Even if we would run sass again afterward, the CSS classes are already defined and this by default based on a hash of its content. Therefore, when we add this dependency again by JavaScript, the hash will be different.
* Even if we disable to add the loaded CSS (see `finalSource`) it would create the class names based on the content, which is not working for transpiled files.
* As we already know about which classes are required for the other file in `collectDependencies` we could store them directly in the `cssModules` object and do not wait for Postcss at all, but this is only working when composes is used within one level (see test `should support deep nested postcss composes imports`).

The only chance I see to archive this without a custom loader would be to extend the Bundle to share information between assets as its implemented for `Assets.replaceBundleNames` already, but this would increase the overall complexity a lot. I started working on this in https://github.com/garthenweb/parcel/tree/bugfix/css-modules-composes-no-duplicate-classes-alternative but to many edge cases raised up.

I guess this is a good compromise between complexity and performance and when using css only it should still be possible to change the `generateScopedName` functions.

garthenweb:
Ahhhh, finally I understand the CI, never found the "test" tab and always tried to check the logs... Sorry for the circumstances!
My tests are a little naive implemented 🙈  Should be fixable, I will have a look.

garthenweb:
Okay, the pipeline is now finally green.

Took a bit longer as one of the tests displayed an issue with the default file system loader on windows, which was most likely there before this PR: See https://github.com/parcel-bundler/parcel/pull/2642/commits/7e47b8d1b08e8babe0214bd2a9f8033d5425dcd7 for the fix as well as https://github.com/css-modules/css-modules-loader-core/issues/230 for the original issue.

garthenweb:
I just made a small clean up for the Postcss loader and merge the master branch.

Maybe one more note: My team is using this for one week now without any issues and we also shipped it to production. It reduces the CSS bundle size by ~25% for us as we don't have duplicate classes within the bundle anymore, further it solved all issues we had with those duplicate CSS rules (caused by the changed specificity).

Issue #2607, opened 2019-01-31
Sourcemap url with multiple entrypoints


mischnic:
# 🐛 bug report

The sourcemap url in Javascript bundles isn't respecting the subfolder of the corresponding bundle.

## 🎛 Configuration (.babelrc, package.json, cli command)

`parcel build src/a/index.js src/b/index.js`

## 🤔 Expected Behavior

The two output files should contains these sourcemap comments:
a/index.js: `//# sourceMappingURL=/a/index.js.map`
b/index.js: `//# sourceMappingURL=/b/index.js.map`

## 😯 Current Behavior

The two output files contain these sourcemap comments (the map files are emitted correctly though) :
a/index.js: `//# sourceMappingURL=/index.js.map`
b/index.js: `//# sourceMappingURL=/index.js.map`

## 💁 Possible Solution

This is the related code:
https://github.com/parcel-bundler/parcel/blob/791b4b80462fee453d82f4c711b86553daa3427e/packages/core/parcel-bundler/src/packagers/JSPackager.js#L238-L245

## 💻 Code Sample

```js
// src/a/index.js
console.log("asd");

// src/b/index.js
console.log("asd");
```
## 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Parcel           |master
| Node             |11
| npm/Yarn         |
| Operating System |

lustoykov:
@mischnic you mind me taking this one or you've already started working on it?

On a first thought

`path.basename(mapBundle.name) -> path.relative(this.options.outDir, mapBundle.name)` 

might do it?

mischnic:
Go ahead. I would say the publicUrl needs to be prepended as well?

lustoykov:
yes, I had in mind:

```
let mapUrl = urlJoin( 
     this.options.publicURL, 
     path.basename(mapBundle.name) -> path.relative(this.options.outDir, mapBundle.name)
   );
```

will give it a try

lustoykov:
Can we close, @mischnic?

PR #2645, opened 2019-02-14
Fix sourceMappingURL for bundles with multiple entry points

lustoykov:
# ↪️ Pull Request

sourceMappingURL is not correct for bundles with multiple entry points

## 💻 Examples

Taken from #2607 

```
parcel build src/a/index.js src/b/index.js
```

The two output files should contains these sourcemap comments:

```
a/index.js: `//# sourceMappingURL=/a/index.js.map`
b/index.js: `//# sourceMappingURL=/b/index.js.map`
```

but in reality `/a/` and `/b/` are not there.

## 🚨 Test instructions

See #2607 for detailed instructions. For tldr, see Examples 👆 

## ✔️ PR Todo

### DIFF
--- a/packages/core/parcel-bundler/src/packagers/JSPackager.js
+++ b/packages/core/parcel-bundler/src/packagers/JSPackager.js
@@ -239,7 +239,7 @@ class JSPackager extends Packager {
       if (mapBundle) {
         let mapUrl = urlJoin(
           this.options.publicURL,
-          path.basename(mapBundle.name)
+          path.relative(this.options.outDir, mapBundle.name)
         );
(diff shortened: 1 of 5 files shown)


packages/core/parcel-bundler/src/packagers/JSPackager.js
```diff
@@ -239,7 +239,7 @@ class JSPackager extends Packager {
       if (mapBundle) {
         let mapUrl = urlJoin(
           this.options.publicURL,
-          path.basename(mapBundle.name)
+          path.relative(this.options.outDir, mapBundle.name)
```
lustoykov: only change is here, rest are tests

PR #2650, opened 2019-02-17
Eliminate dead dependencies from bundle

garthenweb:
# ↪️ Pull Request

I recently restructured my app a bit to have easier handling between server and client code. Afterward, I saw that my bundle size increased a lot because a lot of server dependencies were added to the client bundle.

For environment variables and `process.browser` we allow dead code elimination but terser runs at the end, therefore all dependencies are already collected and added to the code base. Cleaning up directly in preprocess solves the issue.

## 💻 Examples

Please see the test case.

## 🚨 Test instructions

Please see the test case.

## ✔️ PR Todo


packages/core/parcel-bundler/src/assets/JSAsset.js
```diff
@@ -153,17 +153,41 @@ class JSAsset extends Asset {
...
+      this.options.minify &&
+      this.options.target === 'browser' &&
+      (hasProcessBrowser || hasProcessEnv)
+    ) {
+      await babel7(this, {
+        internal: true,
+        config: {
+          plugins: [
+            [
+              require('babel-plugin-minify-dead-code-elimination'),
```
garthenweb: I am actually not sure about the library. As we already have terser to eliminate dead code, we might better use it here as well, but as it's not easy to make it compatible with the babel ast this was the easiest solution I found. Maybe you have some thoughts on this and can show me how to use terser here.

devongovett: Terser already runs later so I'm not sure why you would need this.

garthenweb: If you disable this lines of code, the test will fail. The issue is, that collectDependencies runs after the pretransform. Therefore the dependencies are in the bundle. The code within the falsy if case will be removed by terser in the end, but the collected dependencies will stay. Maybe something else is wrong here, but with this lines of code I saved ~200kb of JavaScript for my application. Maybe you have a better idea how to solve this?
I guess this will be solved by scope hoisting as well, but as it's experimental for know it would be nice to solve this for now.

mischnic: @devongovett If unused branches (after the env and process.browser visitors) contain requires, these requires will actually be part of the bundle.

@garthenweb A cleaner (and faster) solution is to change the env and process visitor to automatically remove unreachable branches if possible. A start:
```diff
diff --git a/packages/core/parcel-bundler/src/visitors/env.js b/packages/core/parcel-bundler/src/visitors/env.js
index 295c654c..468f2d21 100644
--- a/packages/core/parcel-bundler/src/visitors/env.js
+++ b/packages/core/parcel-bundler/src/visitors/env.js
@@ -1,7 +1,8 @@
 const types = require('@babel/types');
 
 module.exports = {
-  MemberExpression(node, asset) {
+  MemberExpression(path, asset) {
+    const {node} = path;
     // Inline environment variables accessed on process.env
     if (types.matchesPattern(node.object, 'process.env')) {
       let key = types.toComputedKey(node);
@@ -12,6 +13,24 @@ module.exports = {
           morph(node, value);
           asset.isAstDirty = true;
           asset.cacheData.env[key.value] = process.env[key.value];
+          if (types.isUnaryExpression(path.parent, {operator: '!'})) {
+            path = path.parentPath;
+          }
+          if (
+            types.isConditionalExpression(path.parent) ||
+            types.isIfStatement(path.parent)
+          ) {
+            const result = path.evaluate();
+            if (result.confident) {
+              if (result.value && path.parent.consequent) {
+                path.parentPath.replaceWith(path.parent.consequent);
+              } else if (!result.value && path.parent.alternate) {
+                path.parentPath.replaceWith(path.parent.alternate);
+              } else {
+                path.parentPath.remove();
+              }
+            }
+          }
         }
       }
     }

diff --git a/packages/core/parcel-bundler/src/visitors/process.js b/packages/core/parcel-bundler/src/visitors/process.js
index 4a3c209b..8e102bb0 100644
--- a/packages/core/parcel-bundler/src/visitors/process.js
+++ b/packages/core/parcel-bundler/src/visitors/process.js
@@ -5,12 +5,27 @@ module.exports = {
     // Inline process.browser
     const isProcess = path.node.object.name === 'process';
     const isBrowser = path.node.property.name === 'browser';
-    const isAssignment = path.parentPath.type === 'AssignmentExpression';
     if (isProcess && isBrowser) {
-      if (isAssignment) {
+      if (t.isAssignmentExpression(path.parentPath)) {
         path.parentPath.remove();
       } else {
-        path.replaceWith(t.booleanLiteral(true));
+        if (t.isIfStatement(path.parentPath)) {
+          path.parentPath.replaceWith(path.parent.consequent);
+        } else if (
+          t.isUnaryExpression(path.parentPath) &&
+          path.parent.operator === '!' &&
+          t.isIfStatement(path.parentPath.parentPath)
+        ) {
+          if (path.parentPath.parent.alternate) {
+            path.parentPath.parentPath.replaceWith(
+              path.parentPath.parent.alternate
+            );
+          } else {
+            path.parentPath.parentPath.remove();
+          }
+        } else {
+          path.replaceWith(t.booleanLiteral(true));
+        }
       }
     }
   }     
diff --git a/packages/core/parcel-bundler/src/assets/JSAsset.js b/packages/core/parcel-bundler/src/assets/JSAsset.js
index 5d42aca0..48024ae3 100644
--- a/packages/core/parcel-bundler/src/assets/JSAsset.js
+++ b/packages/core/parcel-bundler/src/assets/JSAsset.js
     // Inline environment variables
     if (this.options.target === 'browser' && ENV_RE.test(this.contents)) {
       await this.parseIfNeeded();
-      this.traverseFast(envVisitor);
+      this.traverse(envVisitor);
     }
```

garthenweb: Thank you for the advice, I will check as soon as I find time for it.

garthenweb: @mischnic I implemented your feedback now and build a stand-alone version to remove the dead code.
The failing tests don't look related to my changes, an earlier build succeeded, see https://dev.azure.com/devongovett/devongovett/_build/results?buildId=1363. I only pushed small cleanup changes afterward.
I had the feeling that it makes more sense to move the `process.js` and `env.js` file together as they are doing more or less the same. Let me know if you have any concerns about this.

As always, I am happy to receive feedback!

garthenweb:
A short update on this one:
The first version of this, using `babel-plugin-minify-dead-code-elimination` did not catch a case with multiple `or` operators in an `if` statement as `react-hot-loader` is using it to remove dev dependencies: https://github.com/gaearon/react-hot-loader/blob/master/index.js#L5

Therefore, I extended the implementation to cover cases like this as well and shipped to production some days ago. We didn't saw any issues so far.

Overall, this feature saved about ~400kb/ ~10% JavaScript in production (~3.9 MB before; ~3.5 MB after). Please note that we don't use experimental scope hoisting yet. Also, the numbers are for sure heavily dependent on your application, it might vary a lot for other apps.

mischnic:
I ran into an issue while trying to test this:
```js
function init(options) {
  process.env.HMR_PORT = options.hmrPort;
  process.env.HMR_HOSTNAME = options.hmrHostname;
}
```
becomes
```js
function init(options) {
  "0" = options.hmrPort;
  "" = options.hmrHostname;
}
```
We include a process polyfill if the target is browser, the left side of an assignment should (/can) not be changed

garthenweb:
Hmm, alright. I will have a look if I can make the tests run again!
Did you run the tests locally or how did you figure this out? For me, the tests succeeded.

mischnic:
I didn't run the Parcel tests, I tried it with a very large app of mine.
The CI fails are unrelated to your changes (some Babel issue which was fixed in a recent PR)

garthenweb:
Thank you for the hint, your change already did the trick. I merged the master and updated a test case, as well as the fix. Let me know if you have more feedback :)

While implementing the test cases, I found that an assignment in the browser does not really make sense, as we will not consider it at all. I wonder whether we should remove env assignments in the browser completely to save some more bytes and/ or throw a warning in the console. Any opinion on that?

mischnic:
👍 

> Let me know if you have more feedback :)

I think `src/visitors/dead-code-elimination.js` should be renamed to `env-dce.js` because it not only does DCE mainly the env replacement.

> While implementing the test cases, I found that an assignment in the browser does not really make sense, as we will not consider it at all. I wonder whether we should remove env assignments in the browser completely to save some more bytes and/ or throw a warning in the console. Any opinion on that?

Bundling npm packages which were never meant for the browser do use them and blindly removing some statements can cause compatibility issues.

garthenweb:
> I think src/visitors/dead-code-elimination.js should be renamed to env-dce.js because it not only does DCE mainly the env replacement.

Makes sense. As it also covers `process.browser`, would `process-dce.js` also work for you?

> Bundling npm packages which were never meant for the browser do use them and blindly removing some statements can cause compatibility issues.

Alright, let's keep it as it is.

mischnic:
> Makes sense. As it also covers process.browser, would process-dce.js also work for you?

Sure

devongovett:
I'm having a hard time understanding why this is necessary. Don't we already ignore dependencies inside falsey branches? https://github.com/parcel-bundler/parcel/blob/master/packages/core/parcel-bundler/src/visitors/dependencies.js#L43

That runs after the env/process.browser transforms, so should prevent dependencies from being added unnecessarily.

What additional code is this removing that terser wouldn't already?

garthenweb:
@devongovett Ah, I was not aware that this exists. Maybe it would have been a better starting point to solve this. Checking the source code of `isInFalsyBranch`, this will only handle simple if-else statements.

What I have implemented will also cover the [ternary operator](isConditionalExpression) ([example](https://github.com/parcel-bundler/parcel/pull/2650/files#diff-a968c3aaf6942fea21f81b3710394528R2)) and [early returns](https://github.com/parcel-bundler/parcel/pull/2650/files#diff-09e74c100737bfa0eff50c1f1dffaae6R77) ([example](https://github.com/parcel-bundler/parcel/pull/2650/files#diff-ac9b378b953ccb841e8f2c325b5dc0c2R2)).

Also, for the example of [react-hot-loader](https://github.com/gaearon/react-hot-loader/blob/master/index.js#L5), the triple or-statement is not considered confident by babel, therefore I implemented a [workaround](https://github.com/parcel-bundler/parcel/pull/2650/files#diff-09e74c100737bfa0eff50c1f1dffaae6R98). I think some may argue that this is a bug in Babel. As there is a [similar issue](https://github.com/babel/babel/issues/8017) which is open for more than one year I decided to implement it for parcel first to have a solution faster (I would maybe like to investigate into a PR for Babel later).

> What additional code is this removing that terser wouldn't already?

Terser is pretty accurate in removing all the code in the end, so this is not an issue. But the dependencies collected by parcel stay and therefore the bundle size grows.

Does this help with your confusion? If so, what would be the next steps from your point of view? Moving everything into the `isInFalsyBranch`?

devongovett:
Ah I see. Yeah adding support for ternaries etc. seems like an ok idea. If it's too hard to fit the support for that into `isInFalsyBranch`, then we could keep this and remove that function completely as it will have already been handled.

I think babel is not confident because it references `module.hot` which is not defined. The workaround you've done doesn't seem very safe to me since it just ignores things that it isn't confident about. That could easily lead to removing code that should be kept.

I am somewhat worried about build performance though. This switches from using `traverseFast` to babel's traverse, which is much slower due to the scope tracking that it does. Perhaps we could only do this in production mode? But then again in that case, we would have scope hoisting to take care of much of this.

DeMoorJasper:
Gonna close this as parcel 1 is in maintenance mode and this is a pretty significant PR.

I think most of this has already been addressed in Parcel 2, feel free to try out Parcel 2 and create a new issue with more improvements or contribute to Parcel 2's treeshaking

garthenweb:
@DeMoorJasper Just to let you know: Parts of this is still an issue in Parcel v2. Using early return statements will include both versions in the bundle:

``` js
// will at the moment include both versions in parcel v2
const test = () => {
  if (process.browser) {
    require('./browser')()
    return
  }
  require('./server')()
}
// only includes on version in parcel v2
const test = () => {
  if (process.browser) {
    require('./browser')()
  } else {
    require('./server')()
  }
}
```

I can create a more detailed issue later, maybe we can port parts of this PR to Parcel v2.
On the positive side, as far as I remember, the second option was also not working in parcel v1, so it is already an improvement :)

DeMoorJasper:
@garthenweb ow alright, a detailed issue describing further improvements seem like a good idea

PR #2663, opened 2019-02-21
[v2] Fix import paths in ConfigResolver

wbinnssmith:
Every other use of the modules `config` and `localRequire` in
`@parcel/utils` imports it from `@parcel/utils/lib/config`, which is a real
file on disk after build. Without this, both mocha and flow fail right
away.

Please let me know if this is intentional for some reason.

Test Plan: `mocha` in `packages/core` and `flow` in the repo root. Verify
error messages related to the missing module are no longer there.

### DIFF
--- a/packages/core/core/src/ConfigResolver.js
+++ b/packages/core/core/src/ConfigResolver.js
@@ -1,11 +1,11 @@
 // @flow
 import type {FilePath, ParcelConfig, PackageName} from '@parcel/types';
-import {resolveConfig} from '@parcel/utils/lib/config';
+import {resolveConfig} from '@parcel/utils/src/config';
 import Config from './Config';
 import fs from '@parcel/fs';
 import {parse} from 'json5';
 import path from 'path';
-import localRequire from '@parcel/utils/lib/localRequire';
+import localRequire from '@parcel/utils/src/localRequire';
 import assert from 'assert';
 
 type Pipeline = Array<PackageName>;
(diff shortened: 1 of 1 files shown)


wbinnssmith:
cc @padmaia @devongovett

packages/core/core/src/ConfigResolver.js
```diff
@@ -1,11 +1,11 @@
 // @flow
 import type {FilePath, ParcelConfig, PackageName} from '@parcel/types';
-import {resolveConfig} from '@parcel/utils/config';
+import {resolveConfig} from '@parcel/utils/lib/config';
```
padmaia: Could you change `lib` to `src` for both? We've been preferring to use babel's register instead of building for now.

Some context: there are probably other spots where we're using lib, but it's just something we missed in a PR, and there's a larger PR (#2632) that changes them all back. We would need to change it back when we're ready to release, but we also might not be using deep imports for utils by then. It helps right now because when we pulled out the utils into their own package and used a index file to export everything, there were circular dependencies, but eventually it would be good to untangle things. We are also working on a `@parcel/register` which could look up other fields besides "main" in package.json, so we could configure "main" to point at "lib" and "source" to point at "src" and when we're using `@parcel/register` in development we don't need to worry about building before trying something out. That wouldn't really work for deep imports though, so that's one motivation to not use them.

wbinnssmith: Makes sense to me. Thanks for the details :)

PR #2666, opened 2019-02-22
add banner option for bundler and use in core packagers

### LINKED ISSUE #2459: Add comment header to bundle
# ❔ Question

Hi. How can I provide a comment header in the final bundle output?

## 🔦 Context

 I would like to point users to where they can see the original GitHub project.

This is also a legal requirement to include proper license headers on external modules. I tried using "@preserve" JSDoc, and while terser does seem to include it sometimes, it's tucked away in a module (which is in random order in the final output), and it doesn't seem to consistently be kept in the output, bizarrely enough.

## 💻 Code Sample

None.

## 🌍 Your Environment

| Parcel           | 1.10.0
| Node             | v10.0.0
| npm/Yarn         | npm v6.5.0

outofambit:
# ↪️ Pull Request

closes #2459

Adds a `banner` option to `Bundler` and adds utility method in `Packager` to write that banner with the proper markup to the first line of the package.

Implements this pattern (with tests) for the `JSPackager`, `CSSPackager`, and `HTMLPackager`.

(If this approach looks good, I'm happy to add documentation for this new option to this PR.)

## 💻 Examples

The primary use case here is for library maintainers to write the version of their library (or some other information) to the top of the package file. An example string would be "The Best HTTP library ever v0.1" which would be written to the top of each package file with the appropriate comment markup. For example, the JS package would have this at the top of the file:

```js
/** The Best HTTP library ever v0.1 */

```

## 🚨 Test instructions

See included integration tests.

## ✔️ PR Todo

### DIFF
--- a/packages/core/integration-tests/test/css.js
+++ b/packages/core/integration-tests/test/css.js
@@ -44,7 +44,7 @@ describe('css', function() {
       path.join(__dirname, '/dist/index.css'),
       'utf8'
     );
-    assert(file.startsWith(`// ${bannerString} \n`));
+    assert(file.startsWith(`/* ${bannerString} */\n`));
   });
 
   it('should support loading a CSS bundle along side dynamic imports', async function() {
--- a/packages/core/integration-tests/test/css.js
+++ b/packages/core/integration-tests/test/css.js
@@ -33,6 +33,20 @@ describe('css', function() {
     assert.equal(output(), 3);
   });
 
+  it('should include a banner when supplied', async function() {
+    const bannerString = 'Welcome to my bundle, v0';
+    let b = await bundle(path.join(__dirname, '/integration/css/index.js'), {
+      banner: bannerString
+    });
+
+    await run(b);
+    let file = await fs.readFile(
+      path.join(__dirname, '/dist/index.css'),
+      'utf8'
+    );
+    assert(file.startsWith(`// ${bannerString} \n`));
+  });
+
   it('should support loading a CSS bundle along side dynamic imports', async function() {
     let b = await bundle(
       path.join(__dirname, '/integration/dynamic-css/index.js')
--- a/packages/core/integration-tests/test/html.js
+++ b/packages/core/integration-tests/test/html.js
@@ -54,6 +54,20 @@ describe('html', function() {
     }
   });
 
+  it('should include a banner when supplied', async function() {
+    const bannerString = 'Welcome to my bundle, v0';
+    let b = await bundle(path.join(__dirname, '/integration/html/index.html'), {
+      banner: bannerString
+    });
+
+    await run(b);
+    let file = await fs.readFile(
+      path.join(__dirname, '/dist/index.html'),
+      'utf8'
+    );
+    assert(file.startsWith(`<!-- ${bannerString} -->\n`));
+  });
+
   it('should find href attr when not first', async function() {
     let b = await bundle(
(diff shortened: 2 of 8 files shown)


packages/core/integration-tests/test/css.js
```diff
@@ -33,6 +33,20 @@ describe('css', function() {
...
+    let b = await bundle(path.join(__dirname, '/integration/css/index.js'), {
+      banner: bannerString
+    });
+
+    await run(b);
+    let file = await fs.readFile(
+      path.join(__dirname, '/dist/index.css'),
+      'utf8'
+    );
+    assert(file.startsWith(`// ${bannerString} \n`));
```
DeMoorJasper: css should probably use: 
```css
/* comment */
```

outofambit: thanks, that's a good point! actually now that i'm thinking about it some more, i think all the packagers should use the multiline comment markup. If the developer passes a multiline banner string, it would break the current implementation for JS and CSS.

outofambit: fixed up in c3c9bf4. @DeMoorJasper let met know of any more feedback/next steps!

devongovett:
Hmm. I'm worried this is too specific and not customizable enough to include in core, and might be better handled in plugins or a post-build script. What do you think?

outofambit:
Hi @devongovett! I'm mostly going off of similar functionality [in rollup](https://rollupjs.org/guide/en#output-banner-output-footer), though it's true that webpack and browserify don't include this in their core.

This would be relatively easy to implement for a single file type via a packager plugin, but the part that would be much trickier is making the banner customizable. Since `bundler` scrubs the `options` given to it and plugins can't add additional options to the `options` object shape, I think I would have to create another config/API. But perhaps there's another way of going about that?

I'd also be interested to hear if there's additional customization you're thinking of that might make this more compelling for core? Thanks!

devongovett:
I'd like to hold off on this until Parcel 2. It will be much easier for plugins to implement this there, and even if it's done in core, Parcel 1 is pretty much feature frozen at this point so we can focus on Parcel 2.

machsix:
FYI, if you use Parcel 2 with `@parcel/optimizer-terser`. You can add a banner in the `js` file easily by adding a file `.terserrc.js` in the root of your project and add content as the following:
```js
module.exports = {
  output: {
     preamble: METADATA
  }
}
```
replace `METADATA` with the banner you want

bnjmnrsh:
This is all that is required to generate a dynamic banner for your js files. I've not quite figured out how to do the same with css yet with the addition [Parcel CSS](https://parceljs.org/blog/parcel-css/) . 

Inside  `.terserrc.js`

```
const pkg = require('./package.json')

const banner = `/*! ${pkg.name ? pkg.name : pkg.name} v${ pkg.version} | (c) ${new Date().getFullYear()} ${pkg.author} | ${pkg.license} License | ${pkg.homepage} */
`

module.exports = {
  output: {
    preamble: banner
  }
}
```

bnjmnrsh:
@devongovett et al,
As this functionality is already available for js via terser, should discussion of an equivalent `preamble` string be raised in Lightning CSS or would this functionality be best handled as a custom optimizer plugin as suggested by @mischnic in [327](https://github.com/parcel-bundler/lightningcss/issues/327#issuecomment-1302257369)?

At present LigtningCSS doesn't have `.lightingcssrc.js` equivalent,  as all flags are passed via `package.json`.

PR #2671, opened 2019-02-22
fix simple example

padmaia:
# ↪️ Pull Request

Get's the simple example working again.

### DIFF
--- a/packages/core/core/src/Config.js
+++ b/packages/core/core/src/Config.js
@@ -13,7 +13,7 @@ import type {
   Packager,
   Optimizer
 } from '@parcel/types';
-import localRequire from '@parcel/utils/lib/localRequire';
+import localRequire from '@parcel/utils/src/localRequire';
 import {isMatch} from 'micromatch';
 import {basename} from 'path';
 import {CONFIG} from '@parcel/plugin';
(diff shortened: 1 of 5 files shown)


packages/core/workers/src/Worker.js
```diff
@@ -38,9 +38,10 @@ class Worker extends EventEmitter {
 
     this.child = childProcess.fork(childModule, process.argv, options);
 
-    // Unref the child and IPC channel so that the workers don't prevent the main process from exiting
-    this.child.unref();
-    this.child.channel.unref();
+    // TODO: This was causing odd behavior, need to fix
```
padmaia: @DeMoorJasper Was seeing some odd behavior related to this, still don't totally understand, but I'll try to explain. An error was getting thrown because the deserialization of the Config was losing the configPath, which seemed like it was causing the main process to exit immediately without any indication of anything going wrong. Weird thing is once I fixed the deserialization bug, it still sometimes exits early with these lines uncommented, but builds fine when they are removed. One instance is when either the .parcel-cache or dist folder is not created, the main process will exit before completing the build. When I comment out these lines I don't run into that issue. Not sure why these lines would affect creating the cache and dist folders. Haven't spent much time looking into it, but wanted to give you a heads up?

Is it safe to comment out these lines for now? Could you provide some more context why they were needed?

packages/core/core/src/Config.js
```diff
@@ -37,16 +37,11 @@ export default class Config {
...
-    this.configPath = filePath;
-    this.resolvers = config.resolvers || [];
-    this.transforms = config.transforms || {};
-    this.runtimes = config.runtimes || {};
-    this.bundler = config.bundler || '';
-    this.namers = config.namers || [];
-    this.packagers = config.packagers || {};
-    this.optimizers = config.optimizers || {};
-    this.reporters = config.reporters || [];
+  constructor(config: ParcelConfig) {
```
padmaia: Deserialization broke things because the filePath was not getting passed into the constructor by the serializer

padmaia:
Closing this for now as I separated out the config serialization into its own PR (#2779 ). Still need to figure out why `unref` causes such weird behavior, but I'll leave those lines be in case it encourages someone to look into it before I do.

PR #2676, opened 2019-02-23
Unhandled HMR updates should cause a page reload

### LINKED ISSUE #289: 🙋 Regular Browser Reload on File Save

### LINKED ISSUE #2455: Hot Module Reload incompatible with some widgets libraries
# 🐛 bug report

I'm trying to use the excellent [Webix components library](https://stackoverflow.com/questions/200284/what-are-alternatives-to-extjs) with Parcel. The problem is that Parcel apparently doesn't fully unload the nodes created by Webix, so after HMR triggers, Webix complains about "Non unique view id"s. Full repro on CodeSandbox.

[Parcel + webix](https://codesandbox.io/s/6l5nr5r4zn)

### LINKED ISSUE #2509: HMR with reacts appends a new app instance instead of replace the current
I have a parcel + react + css modules project. Whenever I save a file, the entire application gets re-rendered and appended to the #root container instead of replacing the whole thing.

This is what my index.js file looks like:

```javascript
import React, { StrictMode } from 'react';
import ReactDOM from 'react-dom';
import { BrowserRouter as Router } from 'react-router-dom';
import { ApolloProvider } from 'react-apollo';
import './index.css';
import App from './app';
import client from './apollo-client';

ReactDOM.createRoot(document.getElementById('root')).render(
  <StrictMode>
    <Router>
      <ApolloProvider client={client}>
        <App />
      </ApolloProvider>
    </Router>
  </StrictMode>,
);

```

This is what it looks like when I save the same file three times.

screen shot 2019-01-04 at 12 10 09

What am I doing wrong?

### LINKED ISSUE #2606: Hot reload with setInterval
# 🐛 bug report

1. When I start up a parcel server;
2. In index.js i write this line of code:
`setInterval(console.log, 1000, 'hello')`
3. Then i change this line with:
`setInterval(console.log, 1000, 'how are you')`
4. Then i save changes.
5. After reloading i get both intervals going at the same time.

package.json
```json
{
  "name": "parcel-test",
  "version": "1.0.0",
  "description": "parcel-test",
  "main": "index.js",
  "scripts": {
  	"dev": "parcel serve src/index.html",
  	"build": "parcel build src/index.html",
  	"help": "parcel help",
    "test": "test"
  },
  "author": "",
  "license": "ISC",
  "devDependencies": {
    "parcel-bundler": "^1.11.0"
  }
}
```

## 🤔 Expected Behavior

Previous interval should be cleared.

## 😯 Current Behavior

Previous interval is not cleared.

## 🌍 Your Environment

| Parcel           | 1.11.0
| Node             | 8.11.2
| npm/Yarn         | 5.6.0

### LINKED ISSUE #2675: Page reload doesn't work with lit-element
# 🐛 bug report

Im working on a simple single page application build with lit-element. I bundle everything together using parcel and that works pretty good. But in live development, whenever I change something inside of a custom element, the page won't refresh and I geht this issue:
"Uncaught DOMException: Failed to execute 'define' on 'CustomElementRegistry': this name has already been used with this registry"

```js
{
  "scripts": {
    "dev": "parcel watch src/index.html",
    "start": "parcel src/index.html",
    "build": "parcel build src/index.html"
  },
  "dependencies": {
    "lit-element": "^2.0.1",
    "lit-element-router": "^1.0.0"
  },
  "devDependencies": {
    "parcel-bundler": "^1.11.0"
  },
  "browserslist": [
    "Chrome 72"
  ]
}
```

## 🤔 Expected Behavior

The custom element should be reloaded and I should just see the changes I made.

## 😯 Current Behavior

I get this error message instead:
"Uncaught DOMException: Failed to execute 'define' on 'CustomElementRegistry': this name has already been used with this registry"
And there is no reload. If I refresh the page myself, everything works just fine... until I change something else.

## 💁 Possible Solution

No idea :(

## 🔦 Context

I also tried out my project on webpack and everything works there, except I don't like how webpack manages assets etc.
So it seems to be a parcel issue.

## 🌍 Your Environment

| Parcel           | 1.11.0
| Node             | 11.10.0
| npm        | 6.7.0

mischnic:
# ↪️ Pull Request

As mentioned in https://github.com/parcel-bundler/parcel/issues/289#issuecomment-379529297

Closes #289, closes #2675, closes #2606, closes #2455, closes #2509

## 💻 Examples

Would case a full reload:
```js
document.body.innerHTML += "Test21213asdadsd<br>";
```

Would not:
```js
document.body.innerHTML += "Test21213asdadsd<br>";

if(module.hot){
	module.hot.accept(()=>{
		console.log("Hello");
	})
}
```

## 🚨 Test instructions

See example

## ✔️ PR Todo

### DIFF
--- a/packages/core/parcel-bundler/src/builtins/hmr-runtime.js
+++ b/packages/core/parcel-bundler/src/builtins/hmr-runtime.js
@@ -33,23 +33,27 @@ if ((!parent || !parent.isParcelRequire) && typeof WebSocket !== 'undefined') {
     var data = JSON.parse(event.data);
 
     if (data.type === 'update') {
-      console.clear();
-
-      data.assets.forEach(function (asset) {
-        hmrApply(global.parcelRequire, asset);
-      });
-
       var handled = false;
-      data.assets.forEach(function (asset) {
-        if (!asset.isNew) {
-          var didAccept = hmrAccept(global.parcelRequire, asset.id);
-          if (didAccept) {
-            handled = true;
-          }
-        }
+      data.assets.forEach(function(asset){
+        handled = handled ||
+          Boolean(global.parcelRequire.cache[asset.id].hot._acceptCallbacks.length) ||
+          Boolean(global.parcelRequire.cache[asset.id].hot._disposeCallbacks.length);
       });
 
-      if(!handled) {
+      if(handled){
+        console.clear();
+
+        data.assets.forEach(function (asset) {
+          hmrApply(global.parcelRequire, asset);
+        });
+
+        var handled = false;
+        data.assets.forEach(function (asset) {
+          if (!asset.isNew) {
+            hmrAccept(global.parcelRequire, asset.id);
+          }
+        });
+      } else {
         window.location.reload();
       }
     }
--- a/packages/core/parcel-bundler/src/builtins/hmr-runtime.js
+++ b/packages/core/parcel-bundler/src/builtins/hmr-runtime.js
@@ -29,7 +29,7 @@ if ((!parent || !parent.isParcelRequire) && typeof WebSocket !== 'undefined') {
   var ws = new WebSocket(protocol + '://' + hostname + ':' + process.env.HMR_PORT + '/');
   ws.onmessage = function(event) {
     updatedAssets = {};
-    
+
     var data = JSON.parse(event.data);
 
     if (data.type === 'update') {
@@ -42,7 +42,10 @@ if ((!parent || !parent.isParcelRequire) && typeof WebSocket !== 'undefined') {
       var handled = false;
       data.assets.forEach(function (asset) {
         if (!asset.isNew) {
-          handled = handled || hmrAccept(global.parcelRequire, asset.id);
+          var didAccept = hmrAccept(global.parcelRequire, asset.id);
+          if (didAccept) {
+            handled = true;
+          }
         }
       });
 
--- a/packages/core/parcel-bundler/src/builtins/hmr-runtime.js
+++ b/packages/core/parcel-bundler/src/builtins/hmr-runtime.js
@@ -34,10 +34,14 @@ if ((!parent || !parent.isParcelRequire) && typeof WebSocket !== 'undefined') {
 
     if (data.type === 'update') {
       var handled = false;
-      data.assets.forEach(function(asset){
-        handled = handled ||
-          Boolean(global.parcelRequire.cache[asset.id].hot._acceptCallbacks.length) ||
-          Boolean(global.parcelRequire.cache[asset.id].hot._disposeCallbacks.length);
+      data.assets.forEach(function(asset) {
+        handled =
+          handled ||
+          (global.parcelRequire.cache[asset.id] &&
+            (Boolean(global.parcelRequire.cache[asset.id].hot._acceptCallbacks.length) ||
+             Boolean(global.parcelRequire.cache[asset.id].hot._disposeCallbacks.length)
+             )
+          );
       });
 
       if(handled){
--- a/packages/core/parcel-bundler/src/builtins/hmr-runtime.js
+++ b/packages/core/parcel-bundler/src/builtins/hmr-runtime.js
@@ -20,7 +20,7 @@ function Module(moduleName) {
 }
 
 module.bundle.Module = Module;
-var updatedAssets;
+var checkedAssets, assetsToAccept;
 
 var parent = module.bundle.parent;
 if ((!parent || !parent.isParcelRequire) && typeof WebSocket !== 'undefined') {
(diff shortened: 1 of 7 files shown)


mischnic:
I just noticed a major flaw: this breaks the "React HMR just works" feature, because re-executing all Javascript is fine with React. But it doesn't work with setInterval (#2606) or HTML custom elements (#2675).

One possible solution: disabling this change when react is used, or only adding the reload flag.

DeMoorJasper:
If this breaks react it will likely break a lot more libraries/frameworks.

I think the reload flag would probably be our best bet in this case

mischnic:
Closing as this change would break the other half of libraries (current React et al. work and setTimeout etc. don't).

packages/core/parcel-bundler/src/builtins/hmr-runtime.js
```diff
@@ -36,11 +36,16 @@ if ((!parent || !parent.isParcelRequire) && typeof WebSocket !== 'undefined') {
         hmrApply(global.parcelRequire, asset);
       });
 
+      var handled = false;
       data.assets.forEach(function (asset) {
         if (!asset.isNew) {
-          hmrAccept(global.parcelRequire, asset.id);
+          handled = handled || hmrAccept(global.parcelRequire, asset.id);
```
devongovett: I think this will cause updates to stop after the first asset accepts. Then handled will be true and hmrAccept won’t get called on other assets. I think we need to ensure that hmrAccept is still called for all assets, and just keep track of handled separately.

mischnic: Fixed,
another way to fix this would have been:
```js
handled = hmrAccept(global.parcelRequire, asset.id) || handled;
```
But I think it doesn't really convey the intention as well (at first glance).

mischnic:
A downside: everything runs twice, before the reload and after the reload. Might still be an issue in some infinite loop situations.

bildschirmfoto 2019-03-03 um 21 48 39

devongovett:
mmm, so really what we should do is do two passes: 1 to check if any module will accept, and another to actually re-execute the modules if so, or reload the page if not.

mischnic:
> to check if any module will accept

= does any `accept` or `dispose` callback exist. I *think* that's correct

packages/core/parcel-bundler/src/builtins/hmr-runtime.js
```diff
@@ -29,21 +29,36 @@ if ((!parent || !parent.isParcelRequire) && typeof WebSocket !== 'undefined') {
...
-      data.assets.forEach(function (asset) {
-        hmrApply(global.parcelRequire, asset);
+      var handled = false;
+      data.assets.forEach(function(asset) {
+        handled =
+          handled ||
+          (global.parcelRequire.cache[asset.id] &&
+            (Boolean(global.parcelRequire.cache[asset.id].hot._acceptCallbacks.length) ||
+             Boolean(global.parcelRequire.cache[asset.id].hot._disposeCallbacks.length)
+             )
```
devongovett: This isn't quite enough. You need to traverse upward from the asset to the root asset and check if any of the parents accept the update as well as the asset itself. Basically, we need to split `hmrAccept` up into two functions. One that traverses upward and checks if an update should be accepted, but without actually re-executing them. It can also keep track of all of the assets that need to be re-executed in a list. Then, if an update is accepted, go through the list we tracked earlier, and re-execute them.

mischnic: Done: extracted the actual execution into `hmrAcceptRun` and kept track of them in `assetsToAccept`

louh:
Hey, I wanted to comment here because when we updated our `parcel-bundler` to 1.12.3 we found that HMR is broken. (issue: https://github.com/streetmix/streetmix/pull/1357#issuecomment-505055425) Only by reverting to an earlier version we would get it back, but that's what let me down this rabbit hole to try and find what might have broken in 1.12.x.

When we adopted Parcel initially, we did have issues with the HMR (as many people have) but we gradually resolved each of our issues and now HMR works quite well with our app. As a result, it was upsetting to discover that default HMR functionality completely broke. Usually, when an API changes in a library, I expect to see a large, bold "BREAKING CHANGE" notice in the Changelog, which would have given us context and an ability to properly handle the change. Ideally, this would have also been a major version bump as this functionality was introduced without being backwards-compatible. (I think, as a minor version bump, the ideal solution would have been to disable HMR with a flag, instead of changing the default.)

I now know what to do, but I think the fix shouldn't have been hidden in a comment in another issue (https://github.com/parcel-bundler/parcel/issues/289#issuecomment-469536807).

I'll be looking into making a pull request to update your documentation and Changelog for people who might be running into similar issues. Thanks for your work on Parcel!

wyozi:
#2509 still a problem in 2.0.0-alpha.3.2, so not sure this PR fixed that

PR #2677, opened 2019-02-23
Add --reload flag

### LINKED ISSUE #289: 🙋 Regular Browser Reload on File Save

mischnic:
# ↪️ Pull Request

Add a flag to always fully reload the page.

Closes #289

After merging:
- Add a comment about this flag and close: #2675, #2606, #2455, as workaround for #2509
- Add to docs

## 🚨 Test instructions

`parcel index.html --reload` will fully reload the page if some JS is changed (instead of merely re-executing it).

## ✔️ PR Todo


devongovett:
Is this an alternative to #2676?

mischnic:
> Is this an alternative to #2676?

I'm not sure if this is actually useful with that other one merged.
I somewhat wrote about that in the issue: https://github.com/parcel-bundler/parcel/issues/289#issuecomment-466706957

mischnic:
> Is this an alternative to #2676?

Actually, that one doesn't work in 100% of cases.

devongovett:
I do think page reloading should probably be the default unless you accept the HMR update, as proposed by #2676. HMR *not* working seems to be more common than working. You need to write your code in a specific way in order for HMR to work most of the time. I think if you don't opt in then we should just reload the page by default.

mischnic:
> You need to write your code in a specific way in order for HMR to work most of the time.

With current React/Preact code, it works perfectly as it is right now (re-executing JS). What would I have to do the reenable that?
```js
if(module.hot){
	module.hot.accept(() => "dummy");
}

devongovett:
> With current React/Preact code, it works perfectly as it is right now

Sometimes, in very simple apps, yes. But if you do anything more complicated, like have global state in redux, etc., then it won't always work properly.

Yeah, you'd need to do what you said above. You can just `module.hot.accept()` as well - the function isn't required.

I believe this is also required by webpack, or it will just reload the page, so we'd be matching their behavior.

mischnic:
> I believe this is also required by webpack, or it will just reload the page, so we'd be matching their behavior.

👍  -> #2676

PR #2689, opened 2019-02-26
[v2] Parcel.js: Move worker creation into `init`

wbinnssmith:
This means that we don't need the config around in `run` itself. Referencing this nonexistent local caused a flow (and I assume runtime) error.

Test Plan: ¯\\\_(ツ)\_/¯

### DIFF
--- a/packages/core/core/src/Parcel.js
+++ b/packages/core/core/src/Parcel.js
@@ -75,23 +75,6 @@ export default class Parcel {
       throw new Error('Could not find a .parcelrc');
     }
 
-    this.farm = await WorkerFarm.getShared(
-      {
-        config,
-        cliOpts: this.options.cliOpts,
-        env: this.options.env
-      },
-      {
-        workerPath: require.resolve('./worker')
-      }
-    );
-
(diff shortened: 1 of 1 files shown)


wbinnssmith:
More seriously, as a test plan, `flow` passes. I did not run the code, however, as the simple example is broken. See #2671.

packages/core/core/src/Parcel.js
```diff
@@ -75,23 +75,6 @@ export default class Parcel {
...
-      }
-    );
-
-    this.runPackage = this.farm.mkhandle('runPackage');
-  }
-
-  async run() {
-    await this.init();
-
     this.bundlerRunner = new BundlerRunner({
```
DeMoorJasper: Not sure if bundlerRunner should be part of init, as it's not being used in the parcel/register hook

padmaia: Ah, I suggested he fix it this way since I noticed that's what @devongovett was doing in his reporters branch (https://github.com/parcel-bundler/parcel/pull/2632/files#diff-bf9c3feeacb4c7e393721871916cf9f2R65)

PR #2690, opened 2019-02-27
[v2] Make types exact

wbinnssmith:
This makes many of the defined types [exact object types](https://flow.org/en/docs/types/objects/#toc-exact-object-types). This [is going to be the default in Flow going forward](https://medium.com/flow-type/on-the-roadmap-exact-objects-by-default-16b72933c5cf) and anecdotally catches many issues like mistyped property names from silently introducing errors. Flow will provide codemods to get us back to the same type notation while maintaining these semantics and behavior.

These are the more trivial cases. There are more exact objects to come.

Also replaces our custom `Signal` type with the DOM `AbortSignal`, since Flow doesn't differentiate between environments. Seems okay for now.

Test Plan: `flow`

### DIFF
--- a/packages/bundlers/default/src/DefaultBundler.js
+++ b/packages/bundlers/default/src/DefaultBundler.js
@@ -8,10 +8,10 @@ const OPTIONS = {
   maxParallelRequests: 5
 };
 
-type Context = {
+type Context = {|
   bundleGroup: BundleGroup,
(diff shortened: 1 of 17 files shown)


wbinnssmith:
Looks like a clean diff is dependent on #2687 merging 😢

PR #2691, opened 2019-02-27
Add helpful plugin errors

rhurstdialpad:
# ↪️ Pull Request

This PR adds helpful errors when Parcel plugins throw exceptions. Instead of silently failing, then producing a missing dependency error, a warning is printed explaining what error was thrown by the plugin or one of its parsers.

One question still remains. Should this be a warning as I have implemented here, or should parcel's main process re-throw the error from the plugin. The latter would involve capturing the error in the parcel worker and stopping all other work, passing it up to the main process, then finally throwing.

This makes plugin development much easier as prior to this change it was impossible to see plugin errors.

The following error occured because I deleted parcel-bundler from the plugin's node_modules folder:
screen shot 2019-02-26 at 6 40 07 pm


DeMoorJasper:
Probably an error as parcel uses errors as well for failed parsing or requires

PR #2694, opened 2019-02-27
Build the bundle manually with 'rs' in CLI

### LINKED ISSUE #2342: Manual recompile
I would like to have a manual restart "button" like nodemon supports. If you send it a `SIGUSR 20` process it will restart the child process.

I have a Docker Compose container starting parcel with:
```
npx parcel serve \
  --port $WEB_PORT \
  --hmr-port $WEB_HMR_PORT \
  --log-level 3 \
  src/app/index.html
```

I also have a `Docker Compose Volume` putting stuff into `src/`. It also puts stuff into `common/`. Parcel will rebuild on changes to files in `src/` but not in `common/` so I need a way to poke it and rebuild after saving files into `common/`. Right now I have to restart the complete container every time I change common code which is a lot of time wasted.

kirillrogovoy:
# ↪️ Pull Request

Closes #2342.

Hey,

Adding the same feature Nodemon has — https://github.com/remy/nodemon#manual-restarting — manually starting the build from the CLI.

I suppose this is also what's asked for in #2342.

## 💻 Examples

``` bash
>  parcel index.html
✨ Built in 76ms.
[type rs and press enter]
✨ Built in 2ms.
```

## 🚨 Test instructions

(AFAIK we don't currently have tests involving CLI)

How to test it:
1. Create a dummy HTML file with any HTML contents
2. Run your dev version of `parcel` on it
3. Wait for it to get built
4. Type "rs" and press enter
5. Make sure it gets built again: check the update time of the compiled file in `dist/`

## ✔️ PR Todo

### DIFF
--- a/packages/core/parcel-bundler/src/cli.js
+++ b/packages/core/parcel-bundler/src/cli.js
@@ -1,6 +1,7 @@
 require('v8-compile-cache');
 const chalk = require('chalk');
 const program = require('commander');
+const os = require('os');
 const version = require('../package.json').version;
 
 program.version(version);
@@ -228,6 +229,14 @@ async function bundle(main, command) {
         command.open
       );
     }
+
+    process.stdin.on('data', data => {
+      const inputString = data.toString();
+
+      if (inputString === 'rs' + os.EOL) {
+        bundler.bundle();
+      }
+    });
   } else {
     bundler.bundle();
   }


kirillrogovoy:
Also updated the docs: https://github.com/parcel-bundler/website/pull/389

mischnic:
As far as I can see, this is only for `parcel serve`. Wouldn't it make sense to also enable this for `parcel watch`?

devongovett:
Why is this feature necessary?  What is the use case not solved by the watcher?

kirillrogovoy:
>  As far as I can see, this is only for `parcel serve`. Wouldn't it make sense to also enable this for `parcel watch`?

You're right. Will do!

> Why is this feature necessary? What is the use case not solved by the watcher?

See #2342.

> I also have a `Docker Compose Volume` putting stuff into `src/`. It also puts stuff into `common/`. Parcel will rebuild on changes to files in `src/` but not in `common/` so I need a way to poke it and rebuild after saving files into `common/`. Right now I have to restart the complete container every time I change common code which is a lot of time wasted.

devongovett:
> Parcel will rebuild on changes to files in src/ but not in common/

Why is this? Is your code importing things from common? Perhaps it is a bug in the watcher?

kirillrogovoy:
That was the citation from #2342 — You should probably be asking @Industrial. 🙂

I acted on that issue because it was assigned the "Feature" label, so I thought it meant that the core team had already decided that this is what Parcel needs.

devongovett:
Ah ok, sorry for the misunderstanding. The feature label just means it would be a feature if implemented, not that it is planned to be implemented. I think that issue is more of an issue with the particular setup, rather than a feature we need to add to parcel.

PR #2695, opened 2019-02-27
Don't override JSX pragmas set by presets

### LINKED ISSUE #2470: Allow presets to override built-in plugins
# 🐛 bug report

I was attempting to use parcel with a react project using Emotion v10, a CSS-in-JS library. Emotion recommends using a preset which configures several plugins to enable a custom jsx pragma, per https://emotion.sh/docs/css-prop#babel-preset. However, when using this with parcel, parcel's default configuration for compiling jsx files doesn't seem to respect the configuration inside the preset. I ended up copy/pasting the plugin configuration from the preset into my `plugins` configuration in .babelrc:
```js
  "plugins": [
    // Copy/pasta the config in @emotion/babel-preset-css-prop to override parcel's default.
    // Without this, parcel will default to using the one in transforms/babel/jsx
    ["@babel/plugin-transform-react-jsx", {
      "pragma": pragmaName, // default pragma is React.createElement
      "pragmaFrag": "React.Fragment", // default is React.Fragment
    }]
  ]
```
which did work.

full config:

```js
// copy/pasta from @emotion/babel-preset-css-prop
var pragmaName = '___EmotionJSX';
module.exports = {
  "presets": ["@emotion/babel-preset-css-prop"],
  "plugins": [
    "transform-inline-environment-variables",
    // Copy/pasta the config in @emotion/babel-preset-css-prop to override parcel's default.
    // Without this, parcel will default to using the one in transforms/babel/jsx
    ["@babel/plugin-transform-react-jsx", {
      "pragma": pragmaName, // default pragma is React.createElement
      "pragmaFrag": "React.Fragment", // default is React.Fragment
    }]
  ]
}
```

## 🤔 Expected Behavior

I shouldn't need to override the plugin; it should be able to be overridden from the preset.

## 😯 Current Behavior

I do need to override it.

## 🌍 Your Environment

| Parcel           |  1.11.0
| Node             | 10.14.2
| npm         | 6.4.1
| Operating System | osx

mischnic:
# ↪️ Pull Request

We only checked if specified plugins overwrote the jsx pragma, not if some preset included a jsx plugin (recursively).

Fixes #2470

## 💻 Examples

`@emotion/babel-preset-css-prop` sets the `jsxPragma` to a custom value. Previously Parcel would have overwritten that.

## ✔️ PR Todo

### DIFF
--- a/packages/core/integration-tests/package.json
+++ b/packages/core/integration-tests/package.json
@@ -13,6 +13,7 @@
   },
   "devDependencies": {
     "@babel/core": "^7.2.0",
+    "@emotion/babel-preset-css-prop": "^10.0.7",
     "@jetbrains/kotlinc-js-api": "^1.2.12",
     "@parcel/fs": "^1.11.0",
     "@parcel/test-utils": "^1.11.0",
--- a/packages/core/integration-tests/test/integration/jsx-dont-overwrite-preset/.babelrc
+++ b/packages/core/integration-tests/test/integration/jsx-dont-overwrite-preset/.babelrc
@@ -1,0 +1,3 @@
+{
+  "presets": ["@emotion/babel-preset-css-prop"]
+}
--- a/packages/core/integration-tests/test/integration/jsx-dont-overwrite-preset/index.jsx
+++ b/packages/core/integration-tests/test/integration/jsx-dont-overwrite-preset/index.jsx
@@ -1,0 +1,1 @@
+module.exports = <div />;
--- a/packages/core/integration-tests/test/integration/jsx-dont-overwrite-preset/package.json
+++ b/packages/core/integration-tests/test/integration/jsx-dont-overwrite-preset/package.json
@@ -1,0 +1,6 @@
+{
+  "alias": {
+    "@emotion/core": "./stub",
+    "react": "./stub"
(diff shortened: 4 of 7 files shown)


mischnic:
How should I test this?

Local Babel plugins/presets aren't supported (is this a bug or intended):
```json
{
  "presets": ["./test/integration/jsx-dont-overwrite-preset/preset.js"]
}
```
and the only preset that I could find that does this overwriting is `@emotion/babel-preset-css-prop`, should I add that as a devDependency to the integration-tests package?

DeMoorJasper:
@mischnic I guess adding it as an integration test would be the only way to test this. Possibly with the emotion babel plugin. Doesn't really matter that much now that integration tests are a seperate pkg

devongovett:
This seems kinda dangerous and complicated to support. I think it would be easier if we just disabled adding the JSX preset if any babel config is defied, i.e. we should never merge our autogenerated config if a user provides an explicit config. We are already doing this in parcel 2. I'm not sure if it would be considered a breaking change for Parcel 1.

devongovett:
I'd like to wait until Parcel 2 to make the breaking change listed above.

PR #2702, opened 2019-02-28
Unify build abort errors

wbinnssmith:
This unifies the concept of a `BuildAbortError`, which is often thrown in response to a build's abort controller being aborted. It also tests that it is properly thrown in `AssetGraphBuilder`'s `resolve` method if the signal has been aborted.

Previously, `Parcel.js` checked identity against an error that was never thrown, which would never have logged anything.

### DIFF
--- a/packages/core/core/src/Parcel.js
+++ b/packages/core/core/src/Parcel.js
@@ -14,10 +14,8 @@ import getRootDir from '@parcel/utils/getRootDir';
 import loadEnv from './loadEnv';
 import path from 'path';
 import Cache from '@parcel/cache';
-import AssetGraphBuilder from './AssetGraphBuilder';
+import AssetGraphBuilder, {BuildAbortError} from './AssetGraphBuilder';
 import ConfigResolver from './ConfigResolver';
-
-const abortError = new Error('Build aborted');
 
 type ParcelOpts = {|
   entries: string | Array<string>,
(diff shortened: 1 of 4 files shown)


packages/core/core/test/AssetGraphBuilder.test.js
```diff
@@ -0,0 +1,79 @@
+// @flow strict-local
+
+import invariant from 'assert';
```
wbinnssmith: `invariant` has a hardcoded meaning in flow, letting it refine nullable types. In its callable form, it has identical behavior to `assert`, so I'm just aliasing it here.

packages/core/core/src/Parcel.js
```diff
@@ -131,7 +129,7 @@ export default class Parcel {
       // console.log('Finished build'); // eslint-disable-line no-console
       return bundleGraph;
     } catch (e) {
-      if (e !== abortError) {
+      if (e.code !== 'BUILD_ABORTED') {
```
wbinnssmith: I'd rather import the class and do an `instanceof` check here since it's more type-checker friendly, but we've been using `code` elsewhere. Anyone have a preference?

padmaia: Yeah, seems like using an `instanceof` check would be less likely to break things if the code is changed. @devongovett, what was the reasoning for checking against e.code in other places.

wbinnssmith: Just grepped through things and it looks like we only reference `err.code` to switch on Node's own error objects, which use `code` to differentiate themselves. I think it makes sense that our own errors subclass Error instead.

PR #2706, opened 2019-03-01
Flowify `@parcel/fs`

wbinnssmith:
This:
* adds typings for `mkdirp`
* migrates `@parcel/fs` to es modules
* adds flow typings to `@parcel/fs` using `FSPromise` types built into Flow (this is the api `require('fs').promises` uses)
* updates its import uses across parcel
* removes the Node 6-compatible `fs.js` in favor of using the babel-register hook for now

Test Plan: `yarn && yarn flow && yarn test` from the root.


packages/core/cache/src/Cache.js
```diff
@@ -126,7 +126,7 @@ export class Cache {
   async readBlob(blobKey: FilePath) {
     let extension = Path.extname(blobKey);
     let data = await fs.readFile(Path.resolve(this.dir, blobKey), {
-      encoding: extension === '.bin' ? null : 'utf8'
+      encoding: extension === '.bin' ? undefined : 'utf8'
```
wbinnssmith: the flow typings say that `encoding` is nullable rather than optional ¯\\\_(ツ)\_/¯

packages/core/core/src/PackagerRunner.js
```diff
@@ -25,14 +27,14 @@ export default class PackagerRunner {
     let contents = await this.package(bundle);
     contents = await this.optimize(bundle, contents);
 
-    // $FlowFixMe - filePath should already be filled in at this point
-    let dir = path.dirname(bundle.filePath);
+    let filePath = nullthrows(bundle.filePath);
```
wbinnssmith: using nullthrows to refine `filePath`

PR #2714, opened 2019-03-05
Clean existing lints and enable recommended flowtype lints

wbinnssmith:
Clean existing lints and enable recommended flowtype lints

This:
* Resolves an unused import lint
* Suppresses lint warnings for two reasonable cases of console.log, though one should probably use `@parcel/logger`
* Enables recommended flowtype lints. As far as I can tell, we had the flow eslint plugin installed and registered but never opted into any of its rules. This does that, though we could take a more minimal approach and just enable the `flowtype/define-flow-type` rule, which we need for eslint to ignore special global flow types and functions.

Test Plan: `yarn lint` from the root with no errors. `yarn flow` from the root with no errors.

### DIFF
--- a/packages/core/register/src/hook.js
+++ b/packages/core/register/src/hook.js
@@ -58,8 +58,10 @@ export default function register(opts = DEFAULT_CLI_OPTS) {
         return output;
       }
     } catch (e) {
+      /* eslint-disable no-console */
       console.error('@parcel/register failed to process: ', filename);
       console.error(e);
+      /* eslint-enable */
     } finally {
       isProcessing = false;
     }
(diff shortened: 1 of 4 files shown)


packages/core/register/src/hook.js
```diff
@@ -58,8 +58,10 @@ export default function register(opts = DEFAULT_CLI_OPTS) {
         return output;
       }
     } catch (e) {
+      /* eslint-disable no-console */
       console.error('@parcel/register failed to process: ', filename);
```
wbinnssmith: This should probably use `@parcel/logger`, but I don't have a way of manually verifying it.

PR #2722, opened 2019-03-05
Fix remaining v2 tests

wbinnssmith:
This addresses the remaining failing tests in v2.

* Moves `parcel-bundler` to use `@parcel/babel-register` in its mocha options as it relies on `fs`, which is now flowified
* Lessens reliance on async `sleep` in watcher tests by waiting for the watcher to be ready and specific events to be emitted. I assume these events are public API?

Test Plan: `yarn && yarn flow && yarn lint && yarn test`

### DIFF
--- a/packages/core/watcher/src/child.js
+++ b/packages/core/watcher/src/child.js
@@ -19,6 +19,14 @@ function init(options) {
   watcher = new FSWatcher(options);
   watcher.on('all', sendEvent);
   sendEvent('ready');
+
+  // only used for testing
+  watcher.once('ready', async () => {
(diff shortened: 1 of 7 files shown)


packages/core/watcher/src/child.js
```diff
@@ -18,6 +18,8 @@ function init(options) {
   options = optionsTransfer.decode(options);
   watcher = new FSWatcher(options);
   watcher.on('all', sendEvent);
+  // only used for testing
+  watcher.on('ready', () => sendEvent('_chokidarReady'));
```
padmaia: Might be a good idea to put this comment on the same line if the line below is needed outside of testing.

PR #2723, opened 2019-03-05
Normalize input path if possible

QuackMartins:
# ↪️ Pull Request

This pull request adds a simple normalization which tries to transform an absolute path (eg. /project/todo/node_modules/library/file) to a module path (library/file), this fixed a problem I was having when using the preset "react-app"

For some reason, previously it was throwing an error with this message
```
🚨 Cannot resolve dependency 
'/projects/my-project/node_modules/babel-preset-react-app/node_modules/@babel/runtime/helpers/esm/extends' 
at '/projects/my-project/node_modules/babel-preset-react-app/node_modules/@babel/runtime/helpers/esm/extends'
```

I debugged and I saw the resolve function sometimes received the input **@babel/runtime/helpers...** which is fine, but other times it received the full path **/projects/my-project/node_modules/@babel/runtime/helpers...** 

I am not sure why, but I suspect it's related of how **@babel/plugin-transform-runtime** works

Despite parcel could run with **preset-env, preset-react** and **@babel/plugin-transform-runtime**, my bundle was increasing in 10KB compared to **create-react-app**
With this change, my bundle decreased 10KB :tada::tada::tada::tada: 

## 🚨 Test instructions

It is reproducible by creating a brand new project with **create-react-app** and **ejecting** it
Write a .baberc like
```
{
  "presets": [
    "react-app"
  ],
  "plugins": [
  ]
}

```
Run another npm install to assure all dependencies are ok (I needed to)... 
Add a tag script in `public/index.html` to reference `../src/index.js`
Adjust `public/index.html` to remove their dynamic % PUBLIC_URL %f
Run `npm install parcel`... then
```
npx parcel build public/index.html
```

## ✔️ PR Todo

### DIFF
--- a/packages/core/parcel-bundler/src/Resolver.js
+++ b/packages/core/parcel-bundler/src/Resolver.js
@@ -30,6 +30,11 @@ class Resolver {
   }
 
   async resolve(input, parent) {
+    if (this.rootPackage) {
+      let rootNodeModules = this.rootPackage.pkgdir + "/node_modules/"
+      input = input.replace(rootNodeModules, '');
+    }
+    
     let filename = input;
 
     // Check the cache first
--- a/packages/core/parcel-bundler/src/Resolver.js
+++ b/packages/core/parcel-bundler/src/Resolver.js
@@ -31,9 +31,11 @@ class Resolver {
 
   async resolve(input, parent) {
     if (this.rootPackage) {
-      let rootNodeModules = this.rootPackage.pkgdir + "/node_modules/"
-      input = input.replace(rootNodeModules, '');
+      let baseModulesPath = this.rootPackage.pkgdir + '/node_modules/';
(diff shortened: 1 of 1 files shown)


QuackMartins:
I found the reason why the error happens and its indeed related on how preset **react-app** works, it uses absolute path to resolve **@babel/runtime**

It's possible to add an option this to the preset to prevent absolute path like:
```
{
  "presets": [
    ["react-app", { "absoluteRuntime": false }]
  ],
  "plugins": [
  ]
}
```

Despite that... I think this normalization is still useful

devongovett:
This seems dangerous. The resolver should always resolve to an absolute path so I'm not sure why pre-normalizing it would help. Also, it would probably cause issues if a specific version or path to a package was needed.

QuackMartins:
I saw that the **input** for the resolve method commonly is **module**/**path**
Maybe we can dig deeper to find why an absolute path didn't work

I tried debugging it but found no clue :disappointed:

Issue #2619, opened 2019-02-05
Two argument version of `new Worker()` not recognized


chriswa:
# 🐛 bug report

`new Worker('myWorker.js', { type: 'module' })` doesn't get transpiled (I'm guessing because visitors/dependencies.js 's isWebWorker test involves `args.length === 1`.) Removing the second argument to the Worker ctor results in Parcel working flawlessly. It seems like the functionality of `{ type: 'module' }` is already supported, but the parser disallows it.

## 🎛 Configuration (.babelrc, package.json, cli command)

Zero configuration! 😀 

## 🤔 Expected Behavior

Parcel should process `new Worker` with a second argument (ignoring it.)

## 😯 Current Behavior

The `new Worker` code remains untranspiled and the worker source code is not processed into a new bundle.

## 💁 Possible Solution

Two workarounds:

1. omit the second argument, which results in Parcel working fine but that breaks the untranspiled code, since the browser won't resolve modules in the worker (this is currently behind a flag in Chrome, but they do have a bug to get that fixed!)

2. ```
	<script>
		// hack to support Parcel, which doesn't seem to like two arguments getting passed to the Worker ctor
		const OrigWorker = window.Worker
		window.Worker = function(src) {
			return new OrigWorker(src, { type: 'module' })
		}
	</script>

## 🔦 Context

I'm trying to keep my source code runnable without a build step in at least one modern browser (yay native es6 module support!) while providing a build step to run things in all/most modern browsers.

Coming from Webpack, Parser has already exceeded my wildest dreams. <3

## 💻 Code Sample

main.js
```new Worker('myWorker.js', { type: 'module' })```

myWorker.js
```console.log('Hello world!')```

## 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Parcel           | 1.11.0
| Node             | 10.3.0
| npm/Yarn         | 6.1.0
| Operating System | Win10

chriswa:
P.S. I love you. This project is amazing! ❤️

mischnic:
As already noted, this is the relevant code:
https://github.com/parcel-bundler/parcel/blob/eb759a44fdad1f158ff315f653ade89dafa843ef/packages/core/parcel-bundler/src/visitors/dependencies.js#L83-L92

lizzzp1:
Could I give this a shot?

QNimbus:
[...]

PR #2732, opened 2019-03-07
Two argument version of new Worker() not recognized #2619

fogost:
# ↪️ Pull Request

Allow second argument in web worker  #2619

## 💻 Examples

main.js
    new Worker('myWorker.js', { type: 'module' })

myWorker.js
    console.log('Hello world!')

## ✔️ PR Todo


packages/core/parcel-bundler/src/assets/JSAsset.js
```diff
@@ -76,6 +77,10 @@ class JSAsset extends Asset {
   }
 
   async pretransform() {
+    if (WORKER_RE.test(this.contents) && WORKERMODULE_RE.test(this.contents)) {
+      this.contents = this.contents.replace(WORKERMODULE_RE, "type: 'classic'");
```
devongovett: can you do this inside the AST visitor rather than as a string replacement? it will be much more reliable that way

DeMoorJasper:
This has been added to v2

PR #2735, opened 2019-03-07
Update transformer entries to use `src`

wbinnssmith:
Similar to #2728 and #2730, this updates the final references from `lib` to `src`. This allows us to leverage flow properly and run the app and tests through `@babel/register` before we prepare a build process.

This refines the `Meta` type to include a `globals` key which is used to map global names to Assets.

### DIFF
--- a/packages/core/core/src/Asset.js
+++ b/packages/core/core/src/Asset.js
@@ -2,17 +2,17 @@
 
 import type {
   Asset as IAsset,
-  TransformerResult,
-  DependencyOptions,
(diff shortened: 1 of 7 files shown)


packages/core/core/src/Asset.js
```diff
@@ -2,17 +2,17 @@
 
 import type {
   Asset as IAsset,
-  TransformerResult,
```
wbinnssmith: a-z, pull in `Meta` and remove `JSONObject`

PR #2737, opened 2019-03-07
Flow strictify `Asset.js`

wbinnssmith:
This moves `Asset.js` to use `flow strict`. It addresses what Flow calls sketchy null checks -- cases where we're using boolean logic to test whether something is null where falsy values could go unnoticed.

This also adds some basic test coverage to constructing `Asset`s to provide some confidence.

Note that this **does** change the logic in these places, but I believe this better implements intended behavior. For example, previously empty strings of `code` would not be considered, and now they are, as are `outputSize`s of 0, which also would have been ignored.

### DIFF
--- a/packages/core/core/package.json
+++ b/packages/core/core/package.json
@@ -34,7 +34,8 @@
     "@parcel/babel-register": "2.0.0",
     "@parcel/eslint-config": "1.10.3",
     "delay": "^4.1.0",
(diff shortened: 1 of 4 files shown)


wbinnssmith:
This also adds a dependency on a a flow-typed implementation of `shallowequal`.

Test Plan: `yarn test` with the added tests.

PR #2751, opened 2019-03-08
Separate graphviz debugging and have BundleGraph subclass Graph directly

wbinnssmith:
Separate graphviz debugging and have BundleGraph subclass Graph directly

Since `BundleGraph` was subclassing `AssetGraph` only to share the graphviz debugging, extract that into a utility and rebase it directly on `Graph`.

This also moves this debugging to use an environment variable rather than toggling code comments. I'm open to switching this back but I think it's quite nice.

This also simplifies the `AssetGraph` constructor, removing its unused argument and moving property initialization to the near-standardized property initialization proposal. Open to revert this as well.

Test Plan: Run flow, lint, and test. Run `PARCEL_DUMP_GRAPH=1 yarn test` and verify graphviz images are dumped.

### DIFF
--- a/packages/bundlers/default/src/DefaultBundler.js
+++ b/packages/bundlers/default/src/DefaultBundler.js
@@ -162,10 +162,14 @@ export default new Bundler({
       }
     }
 
-    // bundleGraph.dumpGraphViz();
-
-    // bundleGraph.traverseBundles(bundle => {
-    //   bundle.assetGraph.dumpGraphViz();
-    // });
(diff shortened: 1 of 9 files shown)


devongovett:
One thing I was thinking was possibly moving this into a reporter plugin, once #2632 lands.

wbinnssmith:
that sounds exactly like where this should be. Is it all right if I merge this now to unblock some flowtyping I'm doing? If not, I can probably work around that some other way.

PR #2761, opened 2019-03-10
Add backslash to case in Resolver.js

pocesar:
# ↪️ Pull Request

might catch some undefined windows behaviors

## ✔️ PR Todo


devongovett:
Do you have a bug related to this? The input to a resolver should be coming from source code (e.g. `require`), which requires absolute paths to begin with `/`.

pocesar:
@devongovett I don't actually, I was annoyed by the inconsistency in Windows paths (and broken CSS relative paths in 1.12) then I went reviewing the whole code for everything that assumes unix path forward slashes, and in regexes

PR #2740, opened 2019-03-07
Correctly build dependency URLs (for CSS)

mischnic:
# ↪️ Pull Request

Fixes #2736
Fix underlying issue and undo workaround of #2518

## 💻 Examples

Dependency URLs for bundles generated in subfolders would contain something like this `` (not properly resolved).

## ✔️ PR Todo

- [x] Added/updated unit tests for this change
- [x] Filled out test instructions (In case there aren't any unit tests)
- [x] Included links to related issues/PRs

PR #2770, opened 2019-03-11
Fix relative paths in CSS (take 2)

### LINKED ISSUE #2736: Incorrect svg filepath
# 🐛 bug report

After updating to the latest version of Parcel (1.12.0) svg images no longer load properly due to the wrong paths being generated.

```js
{
  "presets": ["@babel/env", "@babel/react"],
  "plugins": ["@babel/plugin-proposal-class-properties","react-hot-loader/babel","@babel/plugin-transform-runtime"]
}
```

## 🤔 Expected Behavior

Generated links should look like:
```html
<img alt="logo" src="./logo_color.7e674228.svg" class="Component-logo-64">
```

## 😯 Current Behavior
Links generated by the newest version of Parcel:
```html
<img alt="logo" src="/....logo_color.7e674228.svg" class="Component-logo-64">
```

## 💻 Code Sample

```js
import logo from "./Images/logo_color.svg";
...
<img alt={"logo"} src={logo} className={classes.logo} />
```

## 🌍 Your Environment

| Parcel           |
| Node             |
| npm/Yarn         |
| Operating System |

devongovett:
This is take 2 of #2740. It changes `addURLDependency` to return a relative path, fixes the test to resolve the asset relative to the css file, and adds a test for importing a raw asset in JS in a subdirectory.

Fixes #2736. Closes #2740.

### DIFF
--- a/packages/core/parcel-bundler/src/Asset.js
+++ b/packages/core/parcel-bundler/src/Asset.js
@@ -257,8 +257,7 @@ class Asset {
         // Replace temporary bundle names in the output with the final content-hashed names.
         let newValue = value;
         for (let [name, map] of bundleNameMap) {
-          let mapRelative = path.relative(path.dirname(this.relativeName), map);
-          newValue = newValue.split(name).join(mapRelative);
+          newValue = newValue.split(name).join(map);
         }
 
         // Copy `this.generated` on write so we don't end up writing the final names to the cache.
--- a/packages/core/parcel-bundler/src/Asset.js
+++ b/packages/core/parcel-bundler/src/Asset.js
@@ -119,11 +119,12 @@ class Asset {
 
     this.addDependency(depName, Object.assign({dynamic: true, resolved}, opts));
 
-    const parsed = URL.parse(url);
-    parsed.pathname = this.options.parser
+    let parsed = URL.parse(url);
+    let name = this.options.parser
       .getAsset(resolved, this.options)
       .generateBundleName();
 
+    parsed.pathname = path.relative(path.dirname(this.relativeName), name);
     return URL.format(parsed);
   }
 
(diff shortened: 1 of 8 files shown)


packages/core/parcel-bundler/src/Asset.js
```diff
@@ -119,11 +119,12 @@ class Asset {
...
     this.addDependency(depName, Object.assign({dynamic: true, resolved}, opts));
 
-    const parsed = URL.parse(url);
-    parsed.pathname = this.options.parser
+    let parsed = URL.parse(url);
+    let name = this.options.parser
       .getAsset(resolved, this.options)
       .generateBundleName();
 
+    parsed.pathname = path.relative(path.dirname(this.relativeName), name);
```
mischnic: `name` is always just a hash+extension, so this `relative` call doesn't really do anything useful.

Here's a repo where it still isn't working (the error resurfaced I tried to fix the original relative PR): https://github.com/mischnic/2736-svg-import
In the dist root, a style file contains `url('../logo_color.78fe8740.svg');`.

(I think this should be `path.posix.relative` to not generate backslashes on Windows.)

devongovett: the test for CSS that I updated fails without this line...

mischnic: In the repo I linked above, this is the core problem:

Relative name of asset: `a/style.css`. A URL Dependency is in root, so `../logo_color.78fe8740.svg` seems correct.

Actually, `a/style.css` gets put into the root as well (does the Asset know this?) and now the `../` part is wrong.

devongovett: ahhh you're right. I guess we cannot correctly generate the relative path.

PR #2788, opened 2019-03-13
Remove unnecessary await in returns from async functions

wbinnssmith:
See #2705 and #2707 for rationale.


wbinnssmith:
This also cherry-picks enabling the lint rule from master.

PR #2800, opened 2019-03-14
Flowify @parcel/logger

wbinnssmith:
This:
* Adds typings for dependencies of logger. The API surface of `ora` is quite large so a lot of it is untested. At some point I'll contribute it to `flow-typed` and we can pull it back in. Same with the others
* Flowifies modules in `@parcel/logger` as strict, including Logger and prettyError
* Removes the unused `prettyError` in utils. Do we want to move it there eventually?
* Updates uses of logger to use `default` on module exports now that it's an es module.

Test Plan: run the simple example and verify logger functions.


DeMoorJasper:
Isn't the logger going to be changed into an eventemitter by this PR #2632 ?

PR #2802, opened 2019-03-14
Flowify @parcel/workers

wbinnssmith:
This:
* Makes `@parcel/workers` use es modules
* Flowifies `@parcel/workers`
* Flowifies `serializer`, a dependency of workers, and makes it an es module
* Updates references in tests and other CJS modules to use the export named `default`
* Conservative changes in the runtime code:
	* [1] Moves property initialization to property initializer syntax
	* [2] Inlines some calls to avoid reassigning to argument bindings (which flow doesn't like)
	* [3] Uses `nullthrows` to refine types that were previously expected to be non-null
	* [4] Makes `WorkerFarm` passes a minimal subset of its options to `Worker` rather than all of its own options
	* [5] Clears a worker's call queue rather than nulling it out
* Less-conservative changes:
	* [6] Refactors `WorkerFarm#processRequest` to use factory functions closing over values rather than building objects over time
		* [7] Does the same with `Child#handleRequest`
	* [8] `getShared` throws immediately if there isn't a shared instance and no options are passed

I'm not 100% certain about the shapes of the types, nor if different components use different shapes of `WorkerRequest`, for example. A thorough review from @DeMoorJasper would be much appreciated.

Test Plan: flow, lint, test. Run the simple example with 1 local worker, and using default number of workers. Verify output.


wbinnssmith:
cc @padmaia @devongovett

packages/core/core/src/loadEnv.js
```diff
@@ -1,23 +1,27 @@
...
+    process.env.NODE_ENV == null ? 'development' : process.env.NODE_ENV;
+
   const dotenvFiles = [
     `.env.${NODE_ENV}.local`,
     `.env.${NODE_ENV}`,
     // Don't include `.env.local` for `test` environment
     // since normally you expect tests to produce the same
     // results for everyone
-    NODE_ENV !== 'test' && '.env.local',
+    NODE_ENV === 'test' ? null : '.env.local',
```
wbinnssmith: avoiding a sketchy null check

packages/core/core/src/loadEnv.js
```diff
@@ -1,23 +1,27 @@
...
     '.env'
   ].filter(Boolean);
 
   await Promise.all(
     dotenvFiles.map(async dotenvFile => {
-      const envPath = await resolveConfig(filepath, [dotenvFile]);
-      if (envPath) {
+      const envPath = await resolveConfig(filePath, [dotenvFile]);
+      if (envPath != null) {
```
wbinnssmith: avoiding a sketchy null check. Behavior change: we no longer take this path if `envPath` is an empty string.

packages/core/workers/src/WorkerFarm.js
```diff
@@ -78,22 +108,23 @@ class WorkerFarm extends EventEmitter {
...
       return this.stopWorker(worker);
     }
   }
 
   startChild() {
-    let worker = new Worker(this.options);
+    let worker = new Worker({forcedKillTime: this.options.forcedKillTime});
 
-    worker.fork(this.options.workerPath, this.bundlerOptions);
+    worker.fork(nullthrows(this.options.workerPath), this.bundlerOptions);
```
wbinnssmith: [3]

packages/core/workers/src/child.js
```diff
@@ -44,38 +59,56 @@ class Child {
...
   }
 
-  async handleResponse(data) {
-    let idx = data.idx;
+  async handleResponse(data: WorkerResponse): Promise<void> {
+    let idx = nullthrows(data.idx);
     let contentType = data.contentType;
     let content = data.content;
-    let call = this.responseQueue.get(idx);
+    let call = nullthrows(this.responseQueue.get(idx));
```
wbinnssmith: [3]

packages/core/workers/src/child.js
```diff
@@ -90,11 +123,19 @@ class Child {
...
-    call.awaitResponse = awaitResponse;
+  async addCall(
+    request: CallRequest,
+    awaitResponse: boolean = true
+  ): Promise<mixed> {
+    // $FlowFixMe
+    let call: ChildCall = {
+      ...request,
+      type: 'request',
+      child: nullthrows(this.childId),
```
wbinnssmith: [3]

packages/core/workers/src/child.js
```diff
@@ -1,39 +1,54 @@
...
+      return this.handleRequest(message);
     }
   }
 
-  async send(data) {
-    data = serialize(data);
-    process.send(data, err => {
+  async send(data: WorkerMessage): Promise<void> {
+    let processSend = nullthrows(process.send).bind(process);
+    processSend(serialize(data), err => {
```
wbinnssmith: [2] inlining this since it was only used once and we reassigned to an argument.

packages/core/workers/src/Worker.js
```diff
@@ -1,31 +1,49 @@
...
-    this.isStopping = false;
+let WORKER_ID = 0;
+export default class Worker extends EventEmitter {
+  +options: WorkerOpts;
+  child: ChildProcess;
+  id: number = WORKER_ID++;
+  processQueue: boolean = true;
+  sendQueue: Array<any> = [];
+
+  calls: Map<number, WorkerCall> = new Map();
```
wbinnssmith: [1]

packages/core/workers/src/WorkerFarm.js
```diff
@@ -78,22 +108,23 @@ class WorkerFarm extends EventEmitter {
...
+  onError(error: ErrorWithCode, worker: Worker) {
     // Handle ipc errors
     if (error.code === 'ERR_IPC_CHANNEL_CLOSED') {
       return this.stopWorker(worker);
     }
   }
 
   startChild() {
-    let worker = new Worker(this.options);
+    let worker = new Worker({forcedKillTime: this.options.forcedKillTime});
```
wbinnssmith: [4]

packages/core/workers/src/WorkerFarm.js
```diff
@@ -119,7 +150,7 @@ class WorkerFarm extends EventEmitter {
         }
       }
 
-      worker.calls = null;
+      worker.calls.clear();
```
wbinnssmith: [5]

packages/core/workers/src/WorkerFarm.js
```diff
@@ -150,32 +181,46 @@ class WorkerFarm extends EventEmitter {
...
+      location: FilePath
+    |} & $Shape<WorkerRequest>,
+    worker?: Worker
+  ): Promise<?WorkerResponse> {
+    let {method, args, location, awaitResponse, idx} = data;
     if (!location) {
       throw new Error('Unknown request');
     }
 
+    const responseFromContent = (content: any): WorkerDataResponse => ({
```
wbinnssmith: [6]

packages/core/workers/src/child.js
```diff
@@ -44,38 +59,56 @@ class Child {
...
-    let args = data.args;
-
-    let result = {idx, child, type: 'response'};
-    try {
-      result.contentType = 'data';
-      if (method === 'childInit') {
-        result.content = this.childInit(...args, child);
-      } else {
-        result.content = await this.module[method](...args);
+  async handleRequest(data: WorkerRequest): Promise<void> {
```
wbinnssmith: [7] Refactored to use these factory functions which close over the request values to create responses. These could be pulled outside of Worker and accept the request instead.

packages/core/workers/src/WorkerFarm.js
```diff
@@ -243,15 +288,22 @@ class WorkerFarm extends EventEmitter {
...
         this.options.maxConcurrentWorkers > 0)
     );
   }
 
-  static async getShared(options, farmOptions) {
+  static async getShared(
+    options?: BundlerOptions,
+    farmOptions?: $Shape<FarmOptions>
+  ): Promise<WorkerFarm> {
+    if (!shared && !options) {
```
wbinnssmith: [8]

packages/core/workers/src/child.js
DeMoorJasper: Not sure what the added benefit of this is? Kinda just wanna know out of interest. PR looks good to me.

Issue #1401, opened 2018-05-20
ES Module type=module cause "parcelRequire is not defined" error


tarepan:
# 🐛 bug report
Using ES Module in main HTML cause "parcelRequire is not defined" error.  
If we use **type="module"**, minimal code in [Getting Started (Parcel official)](https://parceljs.org/getting_started.html) make this error.  

## 🎛 Configuration (.babelrc, package.json, cli command)
zero configuration.  

## 🤔 Expected Behavior
If we write super simple code with ES Module like below,  

```html
<!-- index.html -->
<html>
<body>
<script src="./index.js" type="module"></script>
</body>
</html>
```
```javascript
// index.js
console.log("hello parcel");
```
I hope it is bundled properly and get "hello parcel" in console without error.

## 😯 Current Behavior
In console,  
```
hello parcel.   index.js 1
Uncaught ReferenceError: parcelRequire is not defined at test.904c6489.js:10
```
## 💁 Possible Solution
Apparently ES Module cause this error.  
When I removed **type="module"** from script element, no error observed.  

It suggest that we can avoid error by avoiding **type="module"** for parcel.  
But I think this approach is not good.  
ES Module (type="module") is generally available (now 2018-05, all popular browsers supported!!), so ES module is straight forward way to import module.  

## 🔦 Context
I do not know what problem is caused by this error.  

## 🌍 Your Environment
| Software         | Version(s) |
| ---------------- | ---------- |
| Parcel           |  parcel-bundler@1.8.1 |
| Node             | (problem happen both) 8.9.3 & 9.11.1 |
| npm/Yarn         | |
| Operating System | Windows10 Pro, build 17134 |

##
Thank you good library, I hope this project become better and better.

itshizhan:
since you use type="module",why use parcel?

tarepan:
Mainly for **module resolution.**  
And for old browser users (ES module not supported, need Bundle.)

Schubidu:
A good point to use Parcel in this case is the minification and files with hashes

emirotin:
I have the same problem in different context. I'm trying to consume parcel-bundled module from my webpack-bundled app. Since ES6 modules (my app) run in strict context an attempt to consume the module results in assignment to undefined variable and crashes my app.

alexeyraspopov:
Seems like this is no longer reproducible in v1.9.4.

alexeyraspopov:
[...]

PR #2809, opened 2019-03-16
Add var declaration to parcelRequire definition in prelude.js

twome:
# ↪️ Pull Request

This addresses the ReferenceError when running the final output bundles from issue #1401. The browser/node environment doesn't have `parcelRequire` defined before referencing it, and there was no `var/let/const` declaration.

## 💻 Examples

All JS modules that have the contents of **prelude.js** or **prelude2.js** prepended to them, once referenced as a finished bundle, give the error `Uncaught ReferenceError: parcelDefine is not defined` when run in a browser.

## 🚨 Test instructions

This PR passes all unit tests from running `yarn test` within /packages/core/parcel-bundler/ on macOS 10.14.2. I'm not familiar enough with unit testing browser output to write one.

1. Include a plain JavaScript dependency, in whatever module format (ESM, CommonJS etc), which itself includes dependencies on other JavaScript files, in your entry bundle.
2. Run this bundle in a browser via a <script> reference.
3. Ensure on page load that the browser dev tools show no ReferenceErrors relating to `parcelRequire`.

## ✔️ PR Todo

### DIFF
--- a/packages/core/parcel-bundler/src/builtins/prelude.js
+++ b/packages/core/parcel-bundler/src/builtins/prelude.js
@@ -5,7 +5,7 @@
 //
 // anything defined in a previous bundle is accessed via the
 // orig method which is the require for previous bundles
-parcelRequire = (function (modules, cache, entry, globalName) {
+var parcelRequire = (function (modules, cache, entry, globalName) {
   // Save the require from previous bundle to this closure if any
   var previousRequire = typeof parcelRequire === 'function' && parcelRequire;
   var nodeRequire = typeof require === 'function' && require;
--- a/packages/core/parcel-bundler/src/builtins/prelude.js
+++ b/packages/core/parcel-bundler/src/builtins/prelude.js
@@ -5,7 +5,7 @@
 //
 // anything defined in a previous bundle is accessed via the
 // orig method which is the require for previous bundles
-var parcelRequire = (function (modules, cache, entry, globalName) {
+(typeof window !== 'undefined' ? window : global).parcelRequire = (function (modules, cache, entry, globalName) {
   // Save the require from previous bundle to this closure if any
   var previousRequire = typeof parcelRequire === 'function' && parcelRequire;
   var nodeRequire = typeof require === 'function' && require;
--- a/packages/core/parcel-bundler/src/builtins/prelude.js
+++ b/packages/core/parcel-bundler/src/builtins/prelude.js
@@ -5,9 +5,12 @@
 //
 // anything defined in a previous bundle is accessed via the
 // orig method which is the require for previous bundles
-(typeof window !== 'undefined' ? window : global).parcelRequire = (function (modules, cache, entry, globalName) {
+function getGlobalObject(){
+  return typeof window !== 'undefined' ? window : global;
+}
+getGlobalObject().parcelRequire = (function (modules, cache, entry, globalName) {
   // Save the require from previous bundle to this closure if any
-  var previousRequire = typeof parcelRequire === 'function' && parcelRequire;
+  var previousRequire = typeof getGlobalObject().parcelRequire === 'function' && getGlobalObject().parcelRequire;
   var nodeRequire = typeof require === 'function' && require;
 
   function newRequire(name, jumped) {
@@ -16,7 +19,7 @@
         // if we cannot find the module within our internal map or
         // cache jump to the current global require ie. the last bundle
         // that was added to the page.
-        var currentRequire = typeof parcelRequire === 'function' && parcelRequire;
+        var currentRequire = typeof getGlobalObject().parcelRequire === 'function' && getGlobalObject().parcelRequire;
         if (!jumped && currentRequire) {
           return currentRequire(name, true);
         }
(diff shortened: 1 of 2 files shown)


mischnic:
This test is failing on all platforms:
`javascript should split bundles when a dynamic import is used with --target=node`

```
Error: Cannot find module 'local.js'
    at require (/Users/vsts/agent/2.148.1/work/1/s/packages/core/test-utils/lib/utils.js:279:18)
    at newRequire (evalmachine.<anonymous>:34:18)
    at newRequire (evalmachine.<anonymous>:21:18)
    at localRequire (evalmachine.<anonymous>:53:14)
    at evalmachine.<anonymous>:168:18
```

twome:
If you only run the tests specifically related to `parcel-bundler`, as I described, then it passes all 104 tests in that subdirectory (on macOS 10.14.2). If you look at the code I changed, you can see it's an absolutely tiny change that shouldn't impact anything at all except how the code executes in browsers. The Azure pipelines are *already* failing on the official master; I'm quite sure that my PR is completely unrelated to that.

mischnic:
> If you only run the tests specifically related to `parcel-bundler`, as I described, then it passes all 104 tests in that subdirectory

The most important tests are in `integration-tests`. 

> absolutely tiny change that shouldn't impact anything at all except how the code executes in browsers.

This is why we have a large number of integration tests: there are many edge cases.
And you can't tell me it's a flakey test if all (3 Platforms, 3 Node versions) are all failing

twome:
I'm not saying it's a flakey test, I'm saying the actual change required is literally adding a single `var` declaration: debugging this unit test failure would take me *hours* to figure out how the entirety of this very complex project works, but probably only a few minutes for someone like you to fix. This bug #1401 makes using Parcel on JS projects - which is probably safe to say is the vast majority of Parcel use-cases - impossible without extremely hacky workarounds. It's already cost me about 4 hours of monkey-patching and this bug has been affecting people for almost an entire year. I'm *begging* you to please look at it and fix it.

mischnic:
The main issue here seems to be that Parcel#s async requiring depends on `parcelRequire` being global so that async bundles (which are just loaded by appending a script tag to the head) then have access to the previous `parcelRequire` and can hook in.
https://github.com/parcel-bundler/parcel/blob/32f332782e830036c775dc7ad2631b5801719dc3/packages/core/parcel-bundler/src/builtins/prelude.js#L8-L10
https://github.com/parcel-bundler/parcel/blob/32f332782e830036c775dc7ad2631b5801719dc3/packages/core/parcel-bundler/src/builtins/prelude.js#L24-L30
Making `parcelRequire` "script-local" using var breaks this.

adjenks:
Perhaps something more isomorphic, like:
```
(typeof window !== 'undefined' ? window : global).parcelRequire = (function (modules, cache, entry, globalName) {
//etc
}
```

adjenks:
All the errors appear to be similar. Integration test failure because caniuse-lite is outdated.
One such error:
```
lerna ERR! npm run test-ci stderr:
Browserslist: caniuse-lite is outdated. Please run next command `yarn upgrade caniuse-lite browserslist`
```

twome:
Yep. What's the usual etiquette for this kind of thing:

- submit a separate PR with that dependency update?
- add a new commit and let the maintainers cherry-pick if they want to separate them?
- let the maintainers ignore the CI errors and update that dep on their own terms?

adjenks:
@twome Cool. All checks passed, here's to teamwork 🍻.
@mischnic Please take a peak, we did a thing.

packages/core/integration-tests/test/integration/babel-plugin-autoinstall/package.json
```diff
@@ -1,3 +1,6 @@
 {
-  "name": "babel-plugin-autoinstall"
+  "name": "babel-plugin-autoinstall",
```
mischnic: Don't commit this change. This is an artifact of the autoinstall test

lazylazyllama:
Actually in my case this is not enough. I stell get the the parcelRequire is undefined error. 
To completely get rid of this error I needed not only to add window or global to the parcelRequire definition, but also whenever it is used.

Right now what you do is this:
```js
(typeof window !== 'undefined' ? window : global).parcelRequire = (function (modules, cache, entry, globalName) {
```

I think it would be better to do it like this:
```js
(typeof window !== 'undefined' ? window : global).parcelRequire = (function (modules, cache, entry, globalName) {
  var previousRequire = typeof (typeof window !== 'undefined' ? window : global).parcelRequire === 'function' && (typeof window !== 'undefined' ? window : global).parcelRequire;
```

So it might be better to use a function for this.
I already did a very similar thing in PR #2839 

So if you guys agree, might there be a way to merge my pr into yours or something like this, so that we don't write things twice and prevent conflicts and so on?

twome:
@lazylazyllama Probably best to just incorporate that <GLOBAL_OBJECT>.parcelRequire technique everywhere, and then whichever PR gets merged last can resolve conflicts (which probably won't be necessary and would have to be done at some point anyway).

devongovett:
Don't use `<script type="module">` with parcel. Parcel expects its generated scripts not to run in strict mode so it can assign to the global object and such. I don't see a reason to use `<script type="module">` with Parcel since Parcel already compiles modules to ES5.

adjenks:
@twome Does that solve the problem you were trying to remedy with your pull request?

PR #2814, opened 2019-03-18
Flowify @parcel/utils

wbinnssmith:
This:
* Disables sketchy null checks (which only applies to flow strict modules) for booleans. This would likely only catch the case where you'd forget to check a null or undefined boolean and intended it to have distinct behavior from a `false` boolean, which is unlikely. This would make you awkwardly use `Boolean(maybeNullOrBoolean)` or `!!maybeNullOrBoolean` to cast it explicitly. Leaves on sketchy null checks for other types, which seem far more error prone.
* Adds types for our glob module
* Adds (minimal) types for `resolve` from npm
* Creates our own `resolve` in utils which is pre-promisified and flowtyped.
* Defines a clear `JSONError` representation of an error in `errorUtils`
* Adds return types to methods in `PromiseQueue`
* Adds types for the `is-url` module from npm as well as our internal one.
* Adds types to `bundle-url`
* Splits `localRequire` and `localResolve` into a default and named export, respectively, and adds types to them.

Each of these points is separated in its own commit. 


packages/core/utils/src/index.js
```diff
@@ -0,0 +1 @@
+// @flow
```
wbinnssmith: This should all come back when have a single entrypoint to `utils` to export things.

packages/core/utils/src/installPackage.js
```diff
@@ -47,10 +58,14 @@ async function install(modules, filepath, options = {}) {
...
-async function installPeerDependencies(filepath, name, options) {
+async function installPeerDependencies(
+  filepath: FilePath,
+  name: string,
+  options
+) {
   let basedir = path.dirname(filepath);
-  const resolved = await resolve(name, {basedir});
-  const pkg = await config.loadConfig(resolved, ['package.json']).config;
+  const [resolved] = await resolve(name, {basedir});
```
wbinnssmith: `util.promisify` looks like it only resolves with the first non-error value in the callback, whereas ours will pack non-error values into an array. Some places used `util.promisify`, and some used our own `promisify`. This unifies them, though in this case the resolved path needs to be unpacked.

PR #2815, opened 2019-03-18
Flowify Watcher

wbinnssmith:
This adds types for chokidar, the package the watcher uses under the hood, as well as for the watcher itself.

There are a few meaningful changes to the runtime code, and are noted inline.

### DIFF
--- a/packages/core/watcher/src/Watcher.js
+++ b/packages/core/watcher/src/Watcher.js
@@ -1,8 +1,15 @@
-const fork = require('child_process').fork;
-const optionsTransfer = require('./options');
-const Path = require('path');
-const {EventEmitter} = require('events');
-const {jsonToError} = require('@parcel/utils/src/errorUtils');
+// @flow strict-local
+
+import type {FilePath} from '@parcel/types';
+import type {FSWatcherOptions} from 'chokidar';
+
+import {jsonToError} from '@parcel/utils/src/errorUtils';
+import {fork, type ChildProcess} from 'child_process';
+import EventEmitter from 'events';
+import nullthrows from 'nullthrows';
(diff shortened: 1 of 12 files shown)


packages/core/watcher/src/Watcher.js
```diff
@@ -55,24 +65,28 @@ class Watcher extends EventEmitter {
       cwd: process.cwd()
     };
 
-    this.child = fork(Path.join(__dirname, 'child'), process.argv, options);
+    let child = (this.child = fork(
```
wbinnssmith: Gave this a local binding since flow assumed it could change below.

packages/core/watcher/src/Watcher.js
```diff
@@ -83,34 +97,38 @@ class Watcher extends EventEmitter {
     this.emit('childDead');
   }
 
-  handleEmit(event, data) {
+  // $FlowFixMe
+  handleEmit(event: string, data: any): void {
     if (event === 'watcherError') {
-      data = jsonToError(data);
+      this.emit(event, jsonToError(data));
```
wbinnssmith: Returning early. Flow doesn't like reassigning to arguments.

packages/core/watcher/src/Watcher.js
```diff
@@ -149,10 +167,11 @@ class Watcher extends EventEmitter {
     });
   }
 
-  getWatched() {
+  getWatched(): {[string]: []} {
```
wbinnssmith: This api is kind of weird. It returns an object mapping paths to empty arrays. Are these intended to be mutated later or something? @DeMoorJasper?

PR #2817, opened 2019-03-19
Parcel 2: V2 Server

### LINKED ISSUE #2144: Parcel 2: Server

DeMoorJasper:
# ↪️ Pull Request

I'll add some comments to the code with questions to the rest of the team.

Based on the work done in #2579

Closes #2144

Also updates the error pages:

500:

Screenshot 2019-03-30 at 15 46 33

404:

Screenshot 2019-03-30 at 16 00 21

## 🚨 Test instructions

Run the hmr-react example or look at the passing integration tests.

## ✔️ PR Todo

- [x] Get Dev Server to work
- [x] Get HMR Server to work
- [x] Get HMR Runtime to be added to the bundle
- [x] Update Dev Server tests and get them to pass
- [x] Update HMR Tests
- [x] Get HMR Runtime to work properly
- [x] Get HMR Tests to Pass
- [x] Filter out non browser assets (HMRServer)
- [x] Filter out assets that apply to a different env (Runtime)


packages/core/core/src/Parcel.js
```diff
@@ -104,6 +119,20 @@ export default class Parcel {
...
 
     this.runPackage = this.farm.mkhandle('runPackage');
+
+    if (this.options.serve) {
+      // Not sure if the server should even be mentioned in the core?
+      // Perhaps it should be part of the cli?
+      this.server = await serve(this, this.options.serve);
+    }
+
+    if (this.options.hot) {
```
DeMoorJasper: Not sure if this should be part of the core?

packages/core/core/src/Parcel.js
```diff
@@ -104,6 +119,20 @@ export default class Parcel {
     );
 
     this.runPackage = this.farm.mkhandle('runPackage');
+
+    if (this.options.serve) {
```
DeMoorJasper: Not sure if this should be in core, seems like it would be better fit in cli as people who use the API will probably not want the server I guess

DeMoorJasper: Perhaps we could even make the server a reporter of some sort?

packages/core/core/src/Parcel.js
```diff
@@ -119,6 +148,8 @@ export default class Parcel {
   async build(): Promise<BundleGraph> {
     try {
       // console.log('Starting build'); // eslint-disable-line no-console
+      this.pending = true;
```
DeMoorJasper: This should probably go away...

packages/core/core/src/Parcel.js
DeMoorJasper: Same as server, perhaps this could be some kind of reporter or use some events the reporter PR implements instead of the ones I added.

Tried to keep this PR to as little edits to core as possible

packages/core/core/src/Parcel.js
```diff
@@ -104,6 +119,20 @@ export default class Parcel {
...
+      // Not sure if the server should even be mentioned in the core?
+      // Perhaps it should be part of the cli?
+      this.server = await serve(this, this.options.serve);
+    }
+
+    if (this.options.hot) {
+      this.hmrServer = await serve(this, this.options.hot);
+
+      this.assetGraphBuilder.on('transformed', cacheEntry => {
+        this.hmrServer.addChangedAsset(cacheEntry);
```
DeMoorJasper: Not sure if this is a good way to get changedAssets?

packages/core/core/src/AssetGraphBuilder.js
```diff
@@ -155,6 +156,12 @@ export default class AssetGraphBuilder extends EventEmitter {
     let cacheEntry = await this.runTransform(req);
     let time = Date.now() - start;
 
+    report({
```
DeMoorJasper: Not sure if this is the right place and probably not a proper phase name either

packages/examples/simple/package.json
```diff
@@ -4,7 +4,7 @@
   "license": "MIT",
   "private": true,
   "scripts": {
-    "demo": "node -r @parcel/babel-register ./node_modules/.bin/parcel src/index.js"
+    "demo": "node -r @parcel/babel-register ./node_modules/.bin/parcel src/index.js -w"
```
DeMoorJasper: watching isn't working apparently

packages/reporters/hmr-server/src/HMRServer.js
```diff
@@ -0,0 +1,152 @@
...
+  server: Server;
+  wss: WebSocket.Server;
+  unresolvedError: HMRMessage | null = null;
+  changedAssets: Array<CacheEntry> = [];
+
+  async start(options: ServerOptions) {
+    await new Promise(async resolve => {
+      if (!options.https) {
+        this.server = http.createServer();
+      } else if (typeof options.https === 'boolean') {
```
wbinnssmith: Nit: by the first condition not having been taken, this basically means `else if (options.https === true)`, which is a bit hard to read. This is totally personal preference, but what about:

```js
  if (options.https == null || options.https === false) {
    this.server = http.createServer();
  } else if (options.https === true) {
    this.server = https.createServer(generateCertificate(options));
  } else {
    // options.https is an object
    this.server = https.createServer(await getCertificate(options.https));
  }
```

packages/reporters/hmr-server/src/HMRReporter.js
```diff
@@ -0,0 +1,38 @@
...
+
+let hmrServer: HMRServer | null = null;
+export default new Reporter({
+  async report(event, options) {
+    if (options.hot) {
+      if (!hmrServer) {
+        hmrServer = new HMRServer();
+
+        if (options.hot === true) {
+          options.hot = {
```
wbinnssmith: this will rewrite the same `ParcelOptions` object that's passed around the internals of Parcel and out to other plugins. Maybe just create a local and modify that?

I wonder if we be handing a deep copy of parcel options to plugins 🤔

packages/reporters/dev-server/src/ServerReporter.js
```diff
@@ -0,0 +1,42 @@
...
+import {Reporter} from '@parcel/plugin';
+import Server from './Server';
+
+let server: Server | null = null;
+export default new Reporter({
+  async report(event, options) {
+    if (options.serve) {
+      if (!server) {
+        if (options.serve === true) {
+          options.serve = {
```
wbinnssmith: this also mutates the shared `ParcelOptions`

packages/runtimes/hmr/src/HMRRuntime.js
```diff
@@ -0,0 +1,52 @@
...
+
+import type {BundleGroupNode} from '@parcel/types';
+
+import {Runtime} from '@parcel/plugin';
+import {readFile} from '@parcel/fs';
+
+const HMR_RUNTIME = './loaders/hmr-runtime';
+
+let hmrRuntimeCode = null;
+export default new Runtime({
```
wbinnssmith: Unrelated: this is pretty close to the JSRuntime 😄  I wonder if all runtime will end up looking like this. Maybe the runtime api could just wrap this pattern 🤔

cc @devongovett

packages/core/server-utils/src/getCertificate.js
```diff
@@ -0,0 +1,16 @@
+// @flow
+import {readFile} from '@parcel/fs';
+
+export default async function getCertificate(options: {
```
wbinnssmith: `export default async function getCertificate(options: HTTPSOptions) {`

packages/examples/simple/package.json
wbinnssmith: heh, this also just moved to the new cli interface in #2860 

I think it would be `parcel2 watch src/index.js` there?

packages/reporters/dev-server/src/Server.js
```diff
@@ -0,0 +1,207 @@
+// @flow
+import type {Request, Response, DevServerOptions} from './types.js.flow';
+import type {BundleGraph} from '@parcel/types';
+import type {PrintableError} from '@parcel/reporter-cli/src/prettyError';
```
wbinnssmith: I guess we have to move this (back?) to utils

packages/reporters/dev-server/src/ServerReporter.js
```diff
@@ -0,0 +1,42 @@
+// @flow
+import {Reporter} from '@parcel/plugin';
+import Server from './Server';
+
+let server: Server | null = null;
+export default new Reporter({
+  async report(event, options) {
+    if (options.serve) {
```
wbinnssmith: Nit: an early return here would reduce the amount of indentation

packages/reporters/hmr-server/src/HMRReporter.js
```diff
@@ -0,0 +1,38 @@
+// @flow
+import {Reporter} from '@parcel/plugin';
+import HMRServer from './HMRServer';
+
+let hmrServer: HMRServer | null = null;
+export default new Reporter({
+  async report(event, options) {
+    if (options.hot) {
```
wbinnssmith: Nit: an early return here would reduce the amount of indentation

packages/reporters/hmr-server/src/HMRReporter.js
DeMoorJasper: To be safe we probably would want to clone config before passing.

packages/reporters/dev-server/src/Server.js
DeMoorJasper: Yeah I think so

packages/examples/simple/package.json
DeMoorJasper: Yeah saw that, will update once that PR is merged

packages/reporters/hmr-server/src/HMRReporter.js
DeMoorJasper: One issue with this though, HMRServer should be able to pass a port to HMRRuntime in some way, not sure how to pass that.

DeMoorJasper: I was thinking about letting core normalise some options like serve and hot, but this would be adding extra complexity to parcel core that can probably be worked around. Also this wouldn't event resolve the port issue as that one's currently just wathever port is available.

packages/reporters/cli/src/index.js
```diff
@@ -3,6 +3,7 @@
 // $FlowFixMe This is fine.
 const isTTY = process.stdout.isTTY;
 
-export default (isTTY
+export default require('./SimpleCLIReporter')
```
DeMoorJasper: For some reason the CLIReporter isn't working anymore in this branch.

Probably has to do with my added transformEvent

packages/core/core/src/AssetGraphBuilder.js
devongovett: The other reporting about transforming happens inside the TransformerRunner. Should this one go there too?

packages/core/parcel/src/cli.js
```diff
@@ -152,17 +152,19 @@ function normalizeOptions(command): ParcelOptions {
   if (command.name() === 'serve') {
     serve = {
       https,
-      port: command.port,
-      host: command.host
+      port: command.port || 1234,
+      host: command.host,
+      certificateDir: '.parcel-cert'
```
devongovett: Would we not just put it in .parcel-cache?

packages/core/server-utils/package.json
```diff
@@ -0,0 +1,30 @@
+{
+  "name": "@parcel/server-utils",
```
devongovett: Should this be a separate package or just part of @parcel/utils?

packages/core/server-utils/src/generateCertificate.js
```diff
@@ -0,0 +1,129 @@
+// @flow
+import forge from 'node-forge';
+import fs from 'fs';
+import mkdirp from 'mkdirp';
+import path from 'path';
+import logger from '@parcel/logger';
+
+const DEFAULT_CERTIFICATE_DIR = '.parcel-cert';
```
devongovett: Same

packages/reporters/dev-server/src/Server.js
```diff
@@ -0,0 +1,204 @@
...
+    }
+  }
+
+  async sendIndex(req: Request, res: Response) {
+    if (this.bundleGraph) {
+      // If the main asset is an HTML file, serve it
+      let htmlBundle = null;
+      this.bundleGraph.traverseBundles(bundle => {
+        if (bundle.type === 'html' && bundle.isEntry) {
+          htmlBundle = bundle;
```
devongovett: The third argument to the traversal function is a traversal context object which has a `stop` method. Should be slightly faster to stop once you’ve found the bundle you need.

devongovett: Also, what should we do about multiple entry points here? We don’t currently handle that case in parcel 1 but if a user has multiple html files as entries, we should pick the one named `index.html` by default. If there is no index.html, then take the first entry point. There is an issue for that somewhere...

packages/reporters/dev-server/src/Server.js
```diff
@@ -0,0 +1,204 @@
...
+
+    if (this.error) {
+      const {message, stack} = prettyError(this.error, {color: true});
+      let stackHTML = ansiHtml(stack);
+      let template500 = (await readFile(
+        path.join(__dirname, 'templates/500.html')
+      )).toString('utf8');
+
+      res.end(
+        template500
```
devongovett: Maybe... use a real templatng language? 😉

packages/reporters/hmr-server/src/HMRReporter.js
```diff
@@ -0,0 +1,35 @@
...
+          port: 0,
+          certificateDir: '.parcel-cert'
+        };
+      }
+
+      await hmrServer.start(options.hot);
+    }
+
+    if (event.type === 'buildProgress' && event.phase === 'transformFinished') {
+      hmrServer.addChangedAsset(event.cacheEntry);
```
devongovett: I wonder if instead of doing this as a reporter event, we should just mark the changed assets in the graph in core and provide a list of them to the buildSuccess event instead? One fewer event, and less internal types to expose (CacheEntry).

packages/reporters/hmr-server/src/HMRServer.js
```diff
@@ -0,0 +1,150 @@
...
+      this.broadcast({
+        type: 'error-resolved'
+      });
+
+      return;
+    }
+
+    if (reload) {
+      this.broadcast({
+        type: 'reload'
```
devongovett: I think the logic for when to reload the page should be done in loader runtimes. There are a few cases, but mostly it’s based on asset type. HTML should cause a reload, but only if you’re on that page that changed (can’t be determined server side). JS causes a reload unless you called module.hot.accept. CSS always HMR accepts without reloading. If you add a new bundle, page should reload because a script/style tag might have been added to the html page. Probably more too, but the logic for all of that should happen in the loader plugins with custom logic depending on the type of file. Doesn’t have to be in this PR, I’m just writing things down as they come to me.

packages/runtimes/hmr/src/HMRRuntime.js
devongovett: What is the generalizable pattern? Some logic, then conditionally call addRuntimeAsset?

packages/reporters/dev-server/src/Server.js
DeMoorJasper: Originally was gonna use pug but wasn't sure if I should add another dep. But will do makes sense to do it...

packages/reporters/hmr-server/src/HMRReporter.js
DeMoorJasper: I think the many events actually causes the cli to crash for me so that would probably help also would be a lot faster

packages/reporters/hmr-server/src/HMRServer.js
DeMoorJasper: You're probably right already removed a bunch of the reload stuff anyway. Will look for more of that and clean it up

packages/core/server-utils/package.json
DeMoorJasper: I don't mind adding it to the utils

packages/core/parcel/src/cli.js
DeMoorJasper: In Parcel 1 it's pretty frustrating that the self signed certificate changes everytime you clear the cache. But in Parcel 2 cache should be stable so not sure if this should be seperate?

packages/core/core/src/AssetGraphBuilder.js
DeMoorJasper: I guess that would work although I'm kinda more leaning towards the other approach of keeping the map of changedassets in core and just sending it to all reporters on buildSuccess

packages/reporters/dev-server/src/Server.js
mischnic: > There is an issue for that somewhere...

https://github.com/parcel-bundler/parcel/issues/1315

packages/core/parcel/src/cli.js
devongovett: Yeah... well hopefully it should be better. I think it would be annoying to need to ignore two separate parcel dirs in `.gitignore`.

packages/core/integration-tests/test/server.js
```diff
@@ -2,24 +2,18 @@ const assert = require('assert');
...
-  let server;
-  afterEach(function() {
-    if (server) {
-      server.close();
-      server = null;
-    }
-  });
-
-  function get(file, client = http) {
+describe.only('server', function() {
```
DeMoorJasper: We should probably figure out a way to stop everything after a test.

Currently I'm not sure how to stop all the servers. Perhaps we need to add an onStop event to the reporters?
It will probably be usefull for more than just tests as people will probably want to stop the bundler through the api

packages/reporters/dev-server/src/ServerReporter.js
```diff
@@ -0,0 +1,34 @@
+// @flow
+import {Reporter} from '@parcel/plugin';
+import Server from './Server';
+
+const DEFAULT_CACHE_DIR = '.parcel-cache';
+
+let servers: Map<number, Server> = new Map();
```
DeMoorJasper: Added support for running multiple servers, was originally a workaround for the tests but we should probably keep this as there are probably use cases where people would wanna open multiple instances of parcel.
Might even be useful for a centralised dev server of some kind with multiple devs/projects?

DeMoorJasper:
All the hmr and dev-server tests pass now, except for some minor things like emitting too much assets. Gotta figure out how to get only the truly changed assets and filter out non-browser targets.

Also the hmr tests don't really wanna run if they all run after one another, perhaps there should be a way to stop the watcher, server, ... in the tests after each test?

packages/core/core/src/Parcel.js
```diff
@@ -67,14 +67,15 @@ export default class Parcel {
       rootDir: this.rootDir
     });
 
+    let targetResolver = new TargetResolver();
+    let targets = await targetResolver.resolve(this.rootDir);
+
     this.reporterRunner = new ReporterRunner({
       config,
+      targets,
```
DeMoorJasper: This should probably be merged with the config?

packages/core/core/src/Parcel.js
```diff
@@ -128,10 +129,14 @@ export default class Parcel {
       this.reporterRunner.report({
         type: 'buildSuccess',
         assetGraph,
+        changedAssets: new Map(this.assetGraphBuilder.changedAssets),
         bundleGraph,
         buildTime: Date.now() - startTime
       });
 
+      // TODO: Probably get a cleaner way to reset the map?
+      this.assetGraphBuilder.changedAssets = new Map();
```
padmaia: AssetGraphBuilder could probably reset the Map at the start of the `build` method?

DeMoorJasper: Yeah that would work, will change it

.flowconfig
```diff
@@ -27,4 +27,4 @@ untyped-import
 untyped-type-import
 
 [version]
-0.97.0
\ No newline at end of file
+>=0.97.0
```
wbinnssmith: Flow doesn't follow semver, so I don't think we should do this. What happened that led to this?

packages/bundlers/default/package.json
```diff
@@ -12,6 +12,6 @@
   },
   "dependencies": {
     "@parcel/plugin": "^2.0.0",
-    "@parcel/utils": "^1.10.3"
+    "@parcel/utils": "^2.0.0"
```
wbinnssmith: 😍thanks for doing this!

packages/reporters/cli/src/index.js
```diff
@@ -1,8 +1,10 @@
 // @flow strict-local
 
 // $FlowFixMe This is fine.
-const isTTY = process.stdout.isTTY;
+// const isTTY = process.stdout.isTTY;
```
wbinnssmith: Can this be reverted back?

packages/core/core/src/public/Bundle.js
```diff
@@ -116,7 +116,7 @@ export class MutableBundle extends Bundle implements IMutableBundle {
     this.#bundle.stats = stats;
   }
 
-  removeAsset(asset: Asset): void {
+  removeAsset(asset: Asset): ?NodeId {
```
wbinnssmith: so far we don't expose nodes or node ids through the public api. Mind if I change this to not return the value of the inner `removeAsset`?

.flowconfig
DeMoorJasper: I'm running 0.98.0, so I changed it.
Is it likely to have breaking changes? As I need the latest flow version for some react projects I have.

packages/core/core/src/public/Bundle.js
DeMoorJasper: Sure go ahead. I just did it to fix types

packages/reporters/cli/src/index.js
DeMoorJasper: Kind of not really. It breaks with the servers for some reason. So that would need to get fixed, but probably not in this PR

.flowconfig
DeMoorJasper: Perhaps we could just update it to the latest version and pin it to that one?

PR #2797, opened 2019-03-14
Control Elm debug option via environment variable

lzrski:
# ↪️ Pull Request

A counter proposal for https://github.com/parcel-bundler/parcel/pull/2626.
Work around for https://github.com/elm/compiler/issues/1851

Instead of disabling debug completely this commit introduces an env variable (PARCEL_ELM_NO_DEBUG) as a temporary (?) mean of disabling Elm debugger. Once upstream issue is fixed this change may be reverted, however as I mentioned here https://github.com/parcel-bundler/parcel/pull/2626#issuecomment-472934834, it may be a good idea to keep it around.

## 💻 Examples

As discussed in linked issue and PR.

## 🚨 Test instructions

Take any SSCCE from https://github.com/elm/compiler/issues/1851 and try serving them with:

```sh
npx parcel src/Main.elm
```

and 

```sh
PARCEL_ELM_NO_DEBUG=1 npx parcel src/Main.elm
```

## ✔️ PR Todo

- [ ] Added/updated unit tests for this change
- [x] Filled out test instructions (In case there aren't any unit tests)
- [x] Included links to related issues/PRs

PR #2818, opened 2019-03-19
Allow overriding elm compiler debugger with environmental variables

sklirg:
# ↪️ Pull Request

Adds an environmental variable to disable the Elm compiler because of bugs that can happen compile time with the debugger enabled (https://github.com/elm/compiler/issues/1851). 

This fix will make the build process work exactly as before unless you set the `PARCEL_ELM_DISABLE_DEBUG` environment variable to `1` or `true` (case-insensitive), in which case it will pass the debug flag as `false` through parcel to the build/compiler.

See #2626 and https://github.com/parcel-bundler/parcel/issues/2104#issuecomment-461707960 for more context.

## 💻 Examples

One of my Elm projects gets this error 100% of the time when starting the dev server (`parcel serve`) on cold cache, or 100% of the time when initiating a rebuild after starting it from cache.

To remedy this I currently monkey-patch the `ElmAsssets.js` file in my local node_modules to disable the debugger. Since more people are affected by this I figured I'd open a PR with my suggested fix. This supersedes and closes #2626 by adding a fix which does not break the workflow for current users and users of the debugger.

## 🚨 Test instructions

I've added a test case for the environmental override, and the previous tests pass. To test this fix, please see the examples in https://github.com/elm/compiler/issues/1851 and try them both with and without this fix.

## ✔️ PR Todo


sklirg:
I'm not sure why the tests fail only in node10 on Windows. They pass with node{6,8} on Windows, as well as node{6,8,10} on MacOS and Linux. These changes should not behave differently on specific OS-es or node versions...

lzrski:
Seems like a duplicate of my  #2797. Your solution is more elegant and has test cases, so I favor it over mine. 👍

n1k0:
Is there still any plan to land this? I would be nice being able to decide if we want to work with the Elm debugger or not, not only because of elm/compiler#1851 but also because some people don't use it at all while it gets in the way visually. Having an option to enable/disable it would be nice.

adit-hotstar:
Please allow users to disable the Elm debugger. I'm currently developing an app for a device with a small screen, and the Elm debugger gets in the way of development. This is a serious issue for me.

This is a terrible design decision. I shouldn't be forced to use the Elm debugger if I don't want to. You should really allow users to choose whether or not they'd like to use a certain feature.

devongovett:
Seems like this is something that should be in elm.json, or some other elm config file, not an environment variable.

alex-tan:
I think this already works `PARCEL_ELM_DISABLE_DEBUG=true`

alex-tan:
Weird I thought this PR was already merged in. Nevermind

n1k0:
Any news with this? Do we plan to land this someday, or has the idea been entirely discarded? Would be nice to have some status update.

DeMoorJasper:
This should be fixed in the new transformer for Parcel 2, gonna close this for now

n1k0:
@DeMoorJasper why has this issue been closed? As far as I understand, the new transformer has not landed yet (#3351). Or maybe I missed something? If not, I suggest to reopen this issue for easier status tracking.

DeMoorJasper:
I closed it because we were tracking elm improvements in multiple issues. I'll reopen it if you really want to

n1k0:
Ah, I realize this is a PR, not an issue. So it's probably just fine to leave it closed. Sorry for the noise :)

PR #2824, opened 2019-03-21
Create `events` package and `ValueEmitter`

wbinnssmith:
This creates the `@parcel/events` package, which will store all our event-oriented code like emitters, disposables, etc.

We have some bad habits:

```
~/s/parcel $ rg --type js "setTimeout\(" | wc -l
       6
~/s/parcel $ rg --type js "clearTimeout\(" | wc -l
       2
```

```
~/s/parcel $ rg --type js "\w+\.(on|addListener)\(" | wc -l
      41
~/s/parcel $ rg --type js "\w+\.(off|removeListener)\(" | wc -l
       1
```

(though `commander` uses `on` and doesn't create any subscriptions, but you get the point)

In short runs of Parcel, retaining references unnecessarily won't have much of an impact since the process exits quickly, but in scenarios like watch mode it's pretty easy for memory to balloon if these aren't cleaned up. I'd like to clean these up, as well as reduce our dependence on EventEmitter`.

Using `EventEmitter`s makes flow typing far more difficult by always needing to `any`-type unknown event names, adds a ton of room for error by misspelling event names, etc. We also limit our ability to have a more useful superclass when we subclass `EventEmitter`.

`ValueEmitter` represents a stream of a single type of value (of course a union type satisfies this). I think it's best if we use these internally and then expose a proper typed api through named methods. Other than that, it implements similar semantics as `EventEmitter` with a smaller api.

Instead of having a separate `removeListener`/`off` method, `ValueEmitter` returns an object with a `dispose`, which is a lot less awkward than passing back the original callback. These disposables compose nicely too, and I'd like to make classes that currently extend `EventEmitter` both have and be disposables.

Both Atom's [`event-kit`](https://github.com/atom/event-kit) and Nuclide's [`UniversalDisposable`](https://github.com/facebookarchive/nuclide/blob/master/modules/nuclide-commons/UniversalDisposable.js) are good examples of composable disposables and their concepts have influenced this PR (though an implementation of something like this is yet to land here).

Thanks to @matthewwithanm for looking this over and suggesting freeing the emitter reference and enforcing disposal, as seen in 0245d8f4.

Alternate implementation: We can continue extending `EventEmitter` and get some benefit from typing handlers for specific event types [like so](https://stackoverflow.com/questions/49411515/limiting-the-type-of-eventname-in-a-class-that-extends-eventemitter-with-flow), but we'd still have to allow a string fallback unless we overwrite `on` in our subclass to only accept certain event names, which is just as clumsy as creating individual emitters like we'd do here. Plus we'd still have to subclass `EventEmitter`.

Test Plan: flow, lint, test.


packages/core/logger/src/Logger.js
```diff
@@ -2,27 +2,18 @@
...
+import {ValueEmitter} from '@parcel/events';
 import {inspect} from 'util';
 
 class Logger {
-  // TODO: This can't be explicitly annotated as an EventEmitter since
-  // declared private properties with type annotations break eslint's
-  // no-unused-var rule (even with babel-eslint). Annotate this when
-  // things aren't broken: https://github.com/babel/babel-eslint/issues/688
-  #emitter = new EventEmitter();
+  _logEmitter: ValueEmitter<LogEvent> = new ValueEmitter();
```
wbinnssmith: moved this to an underscored property since this should be annotated, but the bug above prevents using a private property.

wbinnssmith:
Yep. @devongovett, @DeMoorJasper, have any thoughts?

packages/core/logger/src/Logger.js
```diff
@@ -2,27 +2,18 @@
...
-      dispose: () => {
-        this.#emitter.removeListener('log', cb);
-      }
-    };
+    return this._logEmitter.addListener(cb);
   }
 
   verbose(message: string): void {
-    this.#emitter.emit('log', {
+    this._logEmitter.emit({
```
DeMoorJasper: not sure if we should deviate from the standard eventemitter model that's part of Node.js.

Perhaps this can still remain as `emit(type, value)` ?

It would probably also make some stuff easier and possible faster for event listening? (if the listener only cares about 1 type)

wbinnssmith: > not sure if we should deviate from the standard eventemitter model that's part of Node.js.

I hear you. This implements the semantics of EventEmitter nearly identically, but with a slightly different API to help us better type things. The model is the same, except instead of registering arbitrary misspellable strings for `any`-typed callbacks there's a concrete, typed API and internal structure.

> Perhaps this can still remain as emit(type, value) ?

> It would probably also make some stuff easier and possible faster for event listening? (if the listener only cares about 1 type)

I should have provided a better example than the logger :P It's not that there's only one emitter with a single list of handlers, pushing the responsibility of filtering events to the caller.

Instead, the intent is that you create one value emitter for each type of "event" you'd use an EventEmitter for, along with a public method for registering listeners. 

This way, the "type" is encoded as part API and itself can have documentation, type annotation, etc.

Let's say in addition to wanting to observe logs, we wanted to observe "foo" events (like you would with `logger.on('foo', ...)`:

```js
class Logger {
  _logEmitter: ValueEmitter<LogEvent> = new ValueEmitter();
  _fooEmitter: ValueEmitter<Foo> = new ValueEmitter();

  onLog(cb: (event: LogEvent) => mixed): IDisposable {
    return this._logEmitter.addListener(cb);
  }
  
  onFoo(cb: (event: Foo) => mixed): IDisposable {
    return this._fooEmitter.addListener(cb);
  }
```

... and then you could emit Foos by calling `this._fooEmitter.emit(myFoo)` within the logger.

Re: easier -- I can totally relate to this too. Creating `n` emitters along with `n` public api methods for `n` "events" seems like a lot. But this way these events are very clearly part of the public api and can be easily documented and type checked. If we wanted to type an event emitter like in the [alternate suggestion](https://stackoverflow.com/questions/49411515/limiting-the-type-of-eventname-in-a-class-that-extends-eventemitter-with-flow), it would probably look something like this (I've never actually done/tested this; just riffing on that SO post):

```js
class Logger extends EventEmitter {
	on('log', (log: Log) => mixed)): this;
	on('foo', (foo: Foo) => mixed)): this;
	on(event: string, ...args: Array<any>): this;
	emit('log', log: Log): this;
	emit('foo', foo: Foo): this;
	emit(event: string, ...args: Array<any>): this;
}
```

...which not only loses the typesafety in the misspelled-event-name case, since it must at minimum satisfy the `EventEmitter` interface, but IMHO is just as tedious and verbose.

wbinnssmith:
Some extra examples:
  * A very common pattern in Atom, where an EventEmitter is created internally (rather than extending EventEmitter) [but  public methods are created for each event](https://github.com/atom/atom/blob/master/src/package-manager.js#L104-L106).
  * [RxJS Subjects](http://reactivex.io/documentation/subject.html), which these are similar to in that they implement both consumers and producers of a single type of values over time. Rx is much larger in scope, though.
    * [An example in Nuclide of creating internal Subjects and exposing a typed API](https://github.com/facebookarchive/nuclide/blob/01f18e6e4b7126fde5684396daf67038d3a8402f/modules/nuclide-debugger-common/VsDebugSession.js#L70-L83)

wbinnssmith:
I played around with typing `EventEmitter` as an intersection type:

```js
class Logger {
  emitter: {|
    on(event: 'log', handler: (LogEvent) => mixed): void,
    addListener(event: 'log', handler: (LogEvent) => mixed): void,
    emit(event: 'log', n: LogEvent): void;
    on(event: 'foo', handler: (Foo) => mixed): void,
    addListener(event: 'foo', handler: (Foo) => mixed): void,
    emit(event: 'foo', n: Foo): void
  |} & EventEmitter;
}
```

Unfortunately this still means we lose some type-safety as a handler for foo receives `Foo | any`. Afaik, there's no way around this since we have to implement the any-typed EventEmitter interfaces. @jamiebuilds -- I'd love to know if it's do better than this.

The other option is leave all EventEmitters with their standard `any`-typed interfaces and only type the public api. This would would provide type-safety as far as the public interface goes but not internally.

Thoughts?

wbinnssmith:
Does anyone feel strongly against this being merged?

PR #2826, opened 2019-03-21
[RFC] Centralize babel and eslint configuration

wbinnssmith:
This centralizes configuration for babel and eslint as much as possible. It sets the stage for unifying our build, lint, etc. into single processes that takes place at the root of the repo. I intend for this PR to double as an RFC so please comment below if you have comments or alternate suggestions.

In cases where configuration needed to deviate, I left behind rc files in those directories. Overrides for these exceptions *could* be included in the centralized configs but I think by leaving them close to the code the effect is nicer.

The exception to this is large patterns of files that affect all packages. For example, any file matching `**/test/**/*.js` or `**/test.js` will automatically use the `mocha` environment in eslint. This cut back the number of custom eslint configs drastically.

In the case of babel, we don't actually have a build process in place for v2 yet, so it's sufficient to just create a [`babel.config.js`](https://babeljs.io/docs/en/config-files#project-wide-configuration) as [Babel suggests for monorepos](https://babeljs.io/docs/en/config-files#monorepos) and tell Babel register about it.

Test Plan: flow, lint, test-ci


wbinnssmith:
Closing since Parcel itself does not support `babel.config.js` but can bundle itself.

We'll explore a unified build prior to Parcel 2 Alpha, and will either work around this, or potentially @padmaia will have addressed the cacheability issues with `babel.config.js`.

The eslint config changes have been extracted into #2827

PR #2827, opened 2019-03-21
Centralize eslint configuration

wbinnssmith:
Extracted from #2826. Check out that PR for context.

Test Plan: flow, lint, test


packages/reporters/cli/.eslintrc
```diff
@@ -1,9 +1,6 @@
 {
   "extends": "@parcel/eslint-config",
   "plugins": ["react"],
-  "env": {
-    "browser": true
```
wbinnssmith: this isn't a browser environment, despite using react

PR #2832, opened 2019-03-22
Merge react eslint rules into main config

wbinnssmith:
Since we're now using react in the cli reporters, it makes sense to globally apply the recommended eslint-react rules across the project. They shouldn't interfere with our code that doesn't use react.

Adjusts a few things in the is-v2-ready-yet site to appease these rules.

Test Plan: Run the simple example and verify reporter functions correctly. Build the is-v2-ready-yet site and test the heat map and footer links.


packages/core/is-v2-ready-yet/.eslintrc
```diff
@@ -1,11 +1,9 @@
 {
   "extends": "@parcel/eslint-config",
-  "plugins": ["react"],
   "env": {
     "browser": true
   },
   "rules": {
-    "react/jsx-uses-vars": "error",
-    "react/jsx-uses-react": "error"
+    "react/prop-types": "off" // TODO: Enable this when this package is flow-typed
```
wbinnssmith: eslint parses configs with json5, so this comment is okay. That said, some other tools don't like it (VSCode complains right away when opening the file). Alternative: could move this to a js or yaml config.

Issue #2833, opened 2019-03-22
TypeError: Cannot read property 'TYPED_ARRAY_SUPPORT' of undefined #803


lazylazyllama:
# 🐛 bug report

When I import oidc-client and write "new UserManager({...})" in a parcel bundled project, I get this issue:
"Uncaught TypeError: Cannot read property 'TYPED_ARRAY_SUPPORT' of undefined"
I tried it without parcel and it worked. So the issue should be caused by parcel.

## 🎛 Configuration (.babelrc, package.json, cli command)

```json
{
  "name": "my-project",
  "version": "0.1.0",
  "license": "UNLICENSED",
  "author": "Me",
  "scripts": {
    "start": "parcel src/index.html --no-cache --port 8080",
    "start:build": "http-server dist/ -o",
    "build": "parcel build src/index.html --no-cache"
  },
  "dependencies": {
    "lit-element": "^2.1.0",
    "oidc-client": "^1.7.0"
  },
  "devDependencies": {
    "http-server": "^0.11.1",
    "parcel-bundler": "^1.12.3",
    "parcel-plugin-webcomponents": "^1.2.0"
  },
  "browserslist": [
    "last 2 chrome version",
    "last 2 firefox version",
    "last 2 edge version",
    "last 2 safari version",
    "not dead"
  ]
}

```

## 🤔 Expected Behavior

oidc-client should just work.

## 😯 Current Behavior

![image](https://user-images.githubusercontent.com/10547444/54849928-f13c6680-4ce5-11e9-930c-5ea7792e3891.png)

## 💁 Possible Solution

No idea

## 🔦 Context

I just tried to use oidc-client in a parcel bundled project. Everything worked under webpack.

## 💻 Code Sample

```js
import { UserManager } from 'oidc-client'

const userManager = new UserManager({})
```

## 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Parcel           | 1.12.3
| Node             | 11.11.0
| npm              | 6.7.0
| Operating System | Windows 10

mischnic:
This part works fine for me:
```js
import { UserManager } from 'oidc-client'

const userManager = new UserManager({});
```
Does that error happen during the import or during a function call later?

Somehow, `Buffer` is undefined when the Buffer polyfill code runs...

lazylazyllama:
Oh that's my bad. I didn't realize things would work in a normal script.
So I didn't provide enough information here.
[...]

Issue #1747, opened 2018-07-17
Convert `global` to window for browser targets.


gdborton:
# 🐛 bug report
I'm working on prototyping our build in parcel to measure performance. One of the things that I noticed different between webpack and parcel through my work is that webpack will wrap calls to `global` in a function that exposes the `global` variable as `window`.  Parcel bundles just fail when run in the browser with `global is undefined`.

## 🤔 Expected Behavior
In a compiled bundle targeting the browser, `console.log(global);` should behave the same as `console.log(window);`

## 😯 Current Behavior
`console.log(global);` throws with `global is undefined`

## 💁 Possible Solution
One solution is to follow webpack's solution. For any file that includes a reference to `global` wrap the contents of the file with:

```js
(
  function (global) {
    // body of original module.
  }.call(this, require('parcels-global-shim')
)
```

gdborton:
Closing this for now as I couldn't repro in a smaller repository. It may be something specific to our build. I'll reopen if I find more on this.

gdborton:
Re-opened as I was able to track this down to using `type="module"` in my script tags. We don't actually need this, so we can work around, but figured I'd leave this open for tracking purposes.

lazylazyllama:
#2833 

I have the same problem with global by trying to build a website based on web components in which case ``ype="module` is necessary.

mischnic:
Ideally, something babel based could polyfill `globalThis`. But strangely enough `@babel/preset-env` doesn't do this.

ljharb:
Hi! If Parcel is trying to be a bundler for _node_ modules, why is it not correctly providing all browser-compatible node globals, in this case, `global`?

mischnic:
With Parcel 2, `console.log(global)` is turned into 
```js
var $parcel$global =
	typeof globalThis !== "undefined"
		? globalThis
		: typeof self !== "undefined"
		? self
		: typeof window !== "undefined"
		? window
		: typeof global !== "undefined"
		? global
		: {};
console.log($parcel$global);
```
for both Node and Browser targets (though it's not really necessary for Node...)

ljharb:
@mischnic it's necessary for packages that support node/browser versions that predate `globalThis`. Sounds like this issue may be fixed in Parcel 2, then?

JakeChampion:
It's not fixed in parcel 2 as that is what I was using when I came across this issue

mischnic:
Could someone please give a code sample that doesn't work as expected?

[...]

PR #2839, opened 2019-03-24
Fix global "this" in es modules

lazylazyllama:
# ↪️ Pull Request

This PR fixes the problem that `this` is undefined in es modules.
So the value of `global` should not be `undefined` anymore in node modules like buffer etc.

#1747 
#2833 

## 💻 Examples

See the issues i referenced before.

## ✔️ PR Todo

### DIFF
--- a/packages/core/parcel-bundler/src/builtins/prelude.js
+++ b/packages/core/parcel-bundler/src/builtins/prelude.js
@@ -5,6 +5,14 @@
 //
 // anything defined in a previous bundle is accessed via the
 // orig method which is the require for previous bundles
+
+var getGlobal = function () {
+  if (typeof self !== 'undefined') { return self; }
+  if (typeof window !== 'undefined') { return window; }
+  if (typeof global !== 'undefined') { return global; }
+  throw new Error('unable to locate global object');
+};
+
 parcelRequire = (function (modules, cache, entry, globalName) {
   // Save the require from previous bundle to this closure if any
   var previousRequire = typeof parcelRequire === 'function' && parcelRequire;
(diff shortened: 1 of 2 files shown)


lazylazyllama:
I'm very new to parcel and don't know much about the difference between prelude.js and prelude2.js so I'm not sure if there needs work to be done for preload2.js as well.

devongovett:
I don't understand why you would need to load a parcel generated script with `<script type="module">`. Parcel does not generate ES module output so this shouldn't be needed.

lazylazyllama:
I need that because I'm working with web components which are within es-modules.

And the output of parcel is:
```html
<script type="module" src="/app.0a187044.js"></script>
```

So it will be loaded as a es module and it works perfectly fine with all the parcel code inside.

If you ask yourself: "Why use parcel together with es-modules in the first place?"
Here are some answers: #1401

devongovett:
Wouldn't your code also work fine without the `type="module"` bit? Parcel will compile your ES modules down to ES5 anyway, so there is no need to expose them as actual ES modules to the browser.

lazylazyllama:
True, but I still don't see the reason why not to support ES Modules?

Especially in a few years we propably won't compile to es5 anymore. Also it would be nicer if we don't have to adapt our es6 code to parcel just because it can't handle es modules.

The main thing I don't like about this, is that our es6 code makes no sense anymore before we compiled it to parcel. (Importing a es6 module without type="module")

DanielJoyce:
Yep, don't pull a Nodejs, support real modules...

daniellacosse:
It seems that this is crucial for supporting testing via parcel. Jest and AVA both expect ESModules and run in strict mode. If people start having to maintain a separate build pipeline to run their tests, they're just going to switch to that pipeline.

Issue #2678, opened 2019-02-24
.babelrc is not picking up NODE_ENV?


tiborsaas:
# ❔ Question / Bug?

Should this `.babelrc` work? Parcel doesn't seem to pick up the plugin I've specified.

I simply run it as

`parcel ./index.html --open`

If I place the `plugins` key to the root of the JSON, then it starts working, but I want to avoid it because for Jest I don't want that plugin to run.

Doc says it should work: https://new.babeljs.io/docs/en/next/babelrc.html

```json
{
    "env": {
        "production": {
            "plugins": [
                [
                    "@babel/plugin-transform-runtime",
                    {
                        "corejs": 2
                    }
                ]
            ]
        },
        "development": {
            "plugins": [
                [
                    "@babel/plugin-transform-runtime",
                    {
                        "corejs": 2
                    }
                ]
            ]
        },
        "test": {
            "plugins": [
            ]
        }
    },
}
```

## 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Parcel           | 1.11.0
| Node             | 10.5
| npm/Yarn         | 6.1.0
| Operating System | Win10

tiborsaas:
The problem was that I probably had conflicting babel packages installed, so I removed all babel reference from `package.json`, deleted node_modules and added support for only the latest browsers.

https://github.com/parcel-bundler/parcel/issues/1762

dusty:
I ran into this issue as well and the plugins were not run in the specified environment.  In my case I had only one plugin and it was isolated to the production environment.  

Looks like the cause was found in PR #2840.  As soon as I added an additional plugin to run in all environments (outside of the env property) it started working.  As it appears that parcel is skipping running the plugins if no global plugins are defined or they are empty, but its not checking to see if there are plugins related to env.

Issue #1605, opened 2018-06-23
BABEL_ENV does not change unless you use `no-cache`


rodoabad:
# 🐛 bug report

I'm trying to build a CJS and ES compatible package by having CJS point to `main` and ES point to `module`.

I'm going to point to a specific version because I'm currently active in this repository.

https://github.com/rodoabad/man-in-the-mirror/tree/v1.0.26

If you build without `--no-cache`, you will get the same output for both CJS and ES.

```
    "build-cjs": "cross-env BABEL_ENV=cjs parcel build ./src/ManInTheMirror.js --out-dir ./dist/cjs --no-cache",
    "build-es": "cross-env BABEL_ENV=es parcel build ./src/ManInTheMirror.js --out-dir ./dist/es --no-cache",
```

## 🎛 Configuration (.babelrc, package.json, cli command)

`.babelrc`
```json
{
  "plugins": [
    "transform-object-rest-spread"
  ],
  "presets": [
    "react",
    "env"
  ],
  "env": {
    "cjs": {
      "presets": [
        [
          "env",
          {
            "modules": "commonjs"
          }
        ]
      ]
    },
    "es": {
      "presets": [
        [
          "env",
          {
            "modules": false
          }
        ]
      ]
    }
  }
}

```

## 🤔 Expected Behavior

Using http://www.mergely.com/editor to compare diffs.

I should get two different output since one is tailored for CommonJS while the other one is for ES modules.

![gt](https://user-images.githubusercontent.com/1240667/41804286-7e49e17e-7659-11e8-9962-ab125fbe85b7.PNG)

## 😯 Current Behavior

You get the same output.

## 💁 Possible Solution

Use `--no-cache` like how I'm doing it.

## 💻 Code Sample

https://github.com/rodoabad/man-in-the-mirror/tree/v1.0.26

## 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Parcel           | 1.9.2
| Node             | 8
| npm/Yarn         | npm
| Operating System | Windows 10

PR #2840, opened 2019-03-24
Keep babel env option while transforming

lukyth:
# ↪️ Pull Request

According to https://new.babeljs.io/docs/en/next/babelrc.html#env-environment-option, we can specify a plugin to run in a certain environment. But right now the `env` option will be ignored while parcel is doing the babel transformation.

This's probably related to https://github.com/parcel-bundler/parcel/issues/1605 and https://github.com/parcel-bundler/parcel/issues/2678.

I haven't write any test for this yet. But I think it would be better to check beforehand whether this behavior is desired or not. So I open this PR to confirm it first.

I tested this change manually in my project by directly modifying `parcel-bundler` in `node_modules` and this fix solves the issue 😅 

## 💻 Examples

Example:
```js
// .babelrc
{
  "env": {
    "development": {
      "plugins": ["babel-plugin-styled-components"]
    }
  }
}
```

Parcel will ignore the `env` option in the following `.babelrc` so it's gonna treat that `.babelrc` as an empty one and do nothing.

## 🚨 Test instructions

I'm still figuring out how to write a test for this. Here's the manual test instruction for the time being.

My project use TypeScript, Parcel, and Styled Components

```js
// .babelrc.js
module.exports = {
  "env": {
    "development": {
      "plugins": ["babel-plugin-styled-components"]
    }
  }
}
```
```json
// tsconfig.json
{
  "compilerOptions": {
    "target": "es5",
    "module": "esnext",
    "lib": [
      "es2017",
      "dom"
    ],
    "jsx": "react",
    "sourceMap": true,
    "removeComments": true,
    "noEmit": true,
    "strict": true,
    "noUnusedLocals": true,
    "noUnusedParameters": true,
    "noImplicitReturns": true,
    "noFallthroughCasesInSwitch": true,
    "moduleResolution": "node",
    "esModuleInterop": true,
    "experimentalDecorators": true,
    "forceConsistentCasingInFileNames": true
  }
}
```
The `babel-plugin-styled-components` won't be applied. But when I change `.babelrc.js to the one below, the plugin works.
```js
// .babelrc.js
module.exports = {
  "plugins": ["babel-plugin-styled-components"]
}
```

## ✔️ PR Todo

### DIFF
--- a/packages/core/parcel-bundler/src/transforms/babel/babelrc.js
+++ b/packages/core/parcel-bundler/src/transforms/babel/babelrc.js
@@ -15,7 +15,8 @@ async function getBabelConfig(asset, isSource) {
   // Ignore if the config is empty.
   if (
     (!config.plugins || config.plugins.length === 0) &&
-    (!config.presets || config.presets.length === 0)
+    (!config.presets || config.presets.length === 0) &&
+    !config.env
   ) {
     return null;
   }
--- a/packages/core/parcel-bundler/src/transforms/babel/config.js
+++ b/packages/core/parcel-bundler/src/transforms/babel/config.js
@@ -93,7 +93,8 @@ function mergeConfigs(result, config) {
   if (
     !config ||
     ((!config.config.presets || config.config.presets.length === 0) &&
-      (!config.config.plugins || config.config.plugins.length === 0))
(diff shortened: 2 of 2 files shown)


DeMoorJasper:
Will be fixed in Parcel 2

PR #2842, opened 2019-03-25
Fix failed tests

lukyth:
# ↪️ Pull Request

Right now the `master` branch is having this error

Screen Shot 2019-03-25 at 11 07 50

Many recent PRs are failing the CI because of this.

I also took this chance to fix the code format as well. There're some files that aren't formatted correctly by `yarn format`. 6603999

## 💻 Examples

https://dev.azure.com/devongovett/devongovett/_build/results?buildId=1436 (from https://github.com/parcel-bundler/parcel/pull/2836)
https://dev.azure.com/devongovett/devongovett/_build/results?buildId=1451 (from https://github.com/parcel-bundler/parcel/pull/2840)
https://dev.azure.com/devongovett/devongovett/_build/results?buildId=1450 (from https://github.com/parcel-bundler/parcel/pull/2839)

## 🚨 Test instructions

- This test shouldn't fail.
```
parcel-bundler:   1) babel
parcel-bundler:        "before all" hook:
```
- `yarn format` shouldn't produce any changes.

## ✔️ PR Todo


lukyth:
I also found this failing test
Screen Shot 2019-03-25 at 12 28 17
Turns out the error is not coming from Parcel anymore (https://github.com/parcel-bundler/parcel/pull/2693), but from Babel instead.
Screen Shot 2019-03-25 at 12 28 56
Since the error message from Babel seems meaningful enough, should we still keep this test and the logic from https://github.com/parcel-bundler/parcel/pull/2693?

packages/core/integration-tests/test/integration/fs-destructure/index.js
```diff
@@ -1,2 +1,2 @@
-var {readFileSync} = require('fs');
+var readFileSync = require('fs').readFileSync;
```
mischnic: Other changes look fine, but this file was correct. If the test fails, then because Parcel isn't working correctly.

lukyth: Figured that. I wanted to see if the CI is gonna pass when I fix every failed test I found on my local machine. Turns out there's a lint error on Windows but I don't have the OS to debug that 😓 

That aside, I didn't know where to look at in order to fix this. Can you give me a hint? 🙏

lukyth:
Now there's only a lint error on Windows left. I don't have a Windows machine to check what cause the error, so it'd be great if someone else can help me with this.

twome:
The Windows part of `azure-pipelines.yaml` has a 3-space indent in the extremely unlikely case that affects anything. [Patch to fix some tab characters I found, too](https://github.com/parcel-bundler/parcel/commit/e2cb481f7c30fd1d1f0ffdf345db246221b335f0.patch).

mischnic:
This would also bring back the corejs warning:
```
WARNING: We noticed you're using the `useBuiltIns` option without declaring a core-js version. Currently, we assume version 2.x when no version is passed. Since this default version will likely change in future versions of Babel, we recommend explicitly setting the core-js version you are using via the `corejs` option.

You should also be sure that the version you pass to the `corejs` option matches the version specified in your `package.json`'s `dependencies` section. If it doesn't, you need to run one of the following commands:

  npm install --save core-js@2    npm install --save core-js@3
  yarn add core-js@2              yarn add core-js@3
```

lukyth:
I assume we will stick with `core-js@2` for now since migrating to `v3` seems like a breaking change. https://babeljs.io/blog/2019/03/19/7.4.0#migration-from-core-js-2

Please check 5e15771 out if that's good enough as the fix for the warning.

DeMoorJasper:
Closing in favor of #2990 which is based on this PR

PR #2844, opened 2019-03-25
README.md: Make autoinstall more prominent

eliask:
The npm autoinstall feature is a surprising default and it should be very visible anywhere Parcel's features are mentioned.

# ↪️ Pull Request

The autoinstall feature should be very visible in all documentation. So I'm taking the first step with the README.

I find the feature to be a potential security risk since I'm only a typo away from accidentally installing some malicious package.

Now the only mention of the autoinstall feature seems to be hidden away here:
https://parceljs.org/cli.html#disable-autoinstall

An aside: Ideally the default should be disabled, and there should also be a warning whenever autoinstall is in use.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,7 @@
 - ✂️ Zero configuration **code splitting** using dynamic `import()` statements.
 - 🔥 Built in support for **hot module replacement**
 - 🚨 Friendly error logging experience - syntax highlighted code frames help pinpoint the problem.
+- Automatically installs missing packages **from npm** by default (use `--no-autoinstall` to disable)
 
 ## Getting started
 


devongovett:
You are also a typo away when just running `yarn` or `npm` on your CLI. I don't see this as being any different.

eliask:
But the auto-install feature only requires saving a typo for a second or so for it to have an effect, even if an install was not intended.
But the general case for this is that the feature will probably come as a surprise to most users, so they wouldn't know to guard against this eventuality.

But failing that, don't you think it's a surprising enough and/or magical enough feature that it should be documented very prominently?
Should it be left only to a small corner in the docs as a passing mention?

icopp:
Along with security concerns, autoinstall can lead to extremely aggravating things with already-comitted lockfiles changing whenever changing git branches when a local dev server is already running.

PR #2849, opened 2019-03-26
Implement a simple CLI reporter for non-TTYs

wbinnssmith:
This implements `SimpleCLIReporter`, an alternate reporter to export in non-TTY environments. Instead of using ink, which freely rewrites the screen to make updates, simply log messages to stdout/stderr.

This does, however, use ink to write out the bundle report, in favor of reusing that component code. Since only write once and never update the `<BundleReport>`, the screen is never rewritten, so it outputs nicely. We should continue to do this selectively if the main CLI expands.

Test Plan: 
* Disable unrefing of worker processes. This causes the main process to exit too early. This is a bug outside the scope of this PR that needs to be fixed.
* Invoke the simple example and pipe its output to a file. 

Verify:
```
...
Packaging async.1760b8d1.js...
Packaging react.e190d87e.js...
Packaging async2.694a57a4.js...
Built in 1.48s

dist/legacy/react.e190d87e
    603.24      23ms
├── ../../../node_modules/lodash/lodas
   527.24 K     23ms
├── ../../../node_modules/react/cjs/react.developmen
    65.93 K     18ms
├── ../../../node_modules/react/node_modules/prop-types/checkPropType
     3.51 K     10ms
├── ../../../node_modules/object-assign/inde
     2.06 K      7ms
├── ../../../node_modules/react/node_modules/prop-types/lib/ReactPropTypesSecr
      314 B      8ms
└── ../../../node_modules/react/inde
       178       6ms
...
```

### DIFF
--- a/packages/reporters/cli/src/SimpleCLIReporter.js
+++ b/packages/reporters/cli/src/SimpleCLIReporter.js
@@ -1,0 +1,103 @@
+// @flow strict-local
+
+import type {ReporterEvent, ParcelOptions} from '@parcel/types';
+
+import type {Writable} from 'stream';
+
+import {render} from 'ink';
+import {Reporter} from '@parcel/plugin';
+import * as React from 'react';
+import prettifyTime from '@parcel/utils/src/prettifyTime';
+
+import BundleReport from './BundleReport';
+import prettyError from './prettyError';
+import {getProgressMessage} from './utils';
+import logLevels from './logLevels';
+
+export default new Reporter({
+  report(event: ReporterEvent, options: ParcelOptions) {
+    _report(event, options);
+  }
+});
+
+let stdout = process.stdout;
+let stderr = process.stderr;
+
+// Exported only for test
+export function _setStdio(stdoutLike: Writable, stderrLike: Writable) {
+  stdout = stdoutLike;
+  stderr = stderrLike;
+}
+
+// Exported only for test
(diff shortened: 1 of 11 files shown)


wbinnssmith:
An alternate implementation proposal:

We continue to use ink in the non-TTY case. Since it uses `chalk`, coloring sequences respect non-TTYs, and the only issue is ink sending the rewrite sequence all the time.

Instead, we carefully understand the state updates that cause ink to rewrite the screen (basically any non-append-only case) and explicitly avoid them in non-tty ("simple") mode.

Advantages:
* Single reporter implementation that handles both
* More consistency in output between tty/non-tty as we're not reimplementing things

Disadvantages:
* We need to be aware of when ink will redraw in response to non-append-only state changes and place the equivalent of `if (isTTY)` throughout rendering

packages/reporters/cli/src/SimpleCLIReporter.js
```diff
@@ -0,0 +1,103 @@
...
+        default:
+          throw new Error('Unknown log level ' + event.level);
+      }
+    }
+  }
+}
+
+function writeOut(message: ?string): void {
+  if (message != null) {
+    stdout.write(message + '\n');
```
wbinnssmith: using `console.log`/`console.error` rather than stdio would make this more portable, but we're already kind of coupled to stdio with ink.

Issue #2362, opened 2018-12-05
Modify content of index.html dynamically (templating)


rodryquintero:
# 🙋 feature request
To be able to modify `index.html` with variables or a template language. See Webpack's HTML plugin.
https://github.com/jantimon/html-webpack-plugin

## 🤔 Expected Behavior
For example, I can include the version number of the application on the `<title>` tag of the file by using a feature like this.

## 😯 Current Behavior
I do not see a way you can modify content of the `index.html` file dynamically.

## 💁 Possible Solution
Something like the Webpack HTML plugin would work.
https://github.com/jantimon/html-webpack-plugin

## 🔦 Context
Information like dates, timestamps, versions can be included in the `index.html` file. You can pull some information directly from the `package.json` file, in some use cases.

## 💻 Examples
Inserting the version number on the `<title>` could be done like this.

```
<html>
<head>
<title>`${VERSION}`</title>
<body>
<h2>Header</h2>
<p>Text ... ... .. .. ..</p>
</body>
</head>
</html>

**Comment by mischnic** [comment]
Parcel supports pug out of the box, this *might* work:
```pug
// index.pug
p #{name}'s Pug source code!
```

```js
// .pugrc.js
module.export = {
	locals: {
		name: "something" /* or `require("./package.json").version` */
	}
}
```

rodryquintero:
Thanks @mischnic. I have created the `.pug` file and its `.pugrc.js file`. However, it appears that it is not been considered by the pub compiler since i cannot see the variable values

Am I missing something?

DeMoorJasper:
You could probably also use a posthtml plugin for this, this one seems to do something similar as what is requested here: https://github.com/posthtml/posthtml-expressions

rodryquintero:
Thanks @DeMoorJasper and @mischnic. Will look into your recommendations. I consider this issue closed.

mateatslc:
for the latecomers: check out https://github.com/krotovic/parcel-plugin-interpolate-html too

wclr:
What would be the best way to make `parcel` consume JS/TS source (which exports rendered HTML text) as an entry template?

PR #2852, opened 2019-03-26
Add JSON Reporter

wbinnssmith:
This adds a JSON reporter to Parcel 2. While refactoring how we invoke yarn and log its output, I found its JSON reporter really useful in the absence of a programmatic api. Similarly,  outputting feedback in newline-delimited json allows Parcel to be used in novel ways outside of JavaScript programs that could otherwise use it as a library.

This:
* Removes the old cli interface that was replaced in #2362
* Removes the cli reporter from the base parcel config. This replaced with an option for "environmental reporters", reporters that are automatically added when specified as an option in `ParcelOptions`. 
	* This is to avoid coupling reporters specific to how Parcel was invoked (via the cli with or without `--json`, as a programmatic api, etc.) to the core config object.
* Patches `console` in workers. This allows them to send messages to the main process via the logger and also be written (as newline-delimited json in this case)
* Extracts generation of the summary of the bundle graph to be re-used across the JSON reporter, as well as the simple and rich cli reporters.

I'm open to not including the JSON reporter itself in core, but if we don't we should provide the facility for something like the environmental reporters implemented here.

Test Plan:
* Remove the process `unref` in Worker. This once again caused the main process to exit too early. This is a bug outside the scope of this PR that needs to be fixed.
* Invoke `yarn demo --json` in the simple example and verify json output (shortened with [...]):

```
~/s/p/p/e/simple $ yarn demo --json                                                                                                                                                                                                                                                             
yarn run v1.15.2
$ node -r @parcel/babel-register ./node_modules/.bin/parcel2 build src/index.js --json
{"type":"buildStart"}
{"type":"buildProgress","phase":"transforming","filePath":"/Users/wbinnssmith/src/parcel/packages/examples/simple/src/index.js"}
[...]
{"type":"buildProgress","phase":"transforming","filePath":"/Users/wbinnssmith/src/parcel/node_modules/prop-types/lib/ReactPropTypesSecret.js"}
{"type":"buildProgress","phase":"bundling"}
{"type":"buildProgress","phase":"transforming","filePath":"/Users/wbinnssmith/src/parcel/packages/runtimes/js/src/JSRuntime.js"}
{"type":"log","level":"warn","message":"'Browserslist: caniuse-lite is outdated. Please run next command `yarn upgrade caniuse-lite browserslist`'"}
[...]
{"type":"buildProgress","phase":"transforming","filePath":"/Users/wbinnssmith/src/parcel/packages/runtimes/js/src/bundle-url.js"}
{"type":"buildProgress","phase":"packaging","bundleFilePath":"dist/legacy/index.js"}
[...]
{"type":"buildProgress","phase":"packaging","bundleFilePath":"dist/modern/worker.5074a895.js"}
```

There's also a final `buildSuccess` event emitted that I'll shorten and indent below, but this is also newline-delimited JSON:

```
{
  "type": "buildSuccess",
  "buildTime": 1831,
  "bundles": [
    {
      "filePath": "dist/legacy/react.804b5e33.js",
      "size": 617494,
      "time": 57,
      "largestAssets": [
        {
          "filePath": "/Users/wbinnssmith/src/parcel/node_modules/lodash/lodash.js",
          "size": 539894,
          "time": 52
        },
        {
          "filePath": "/Users/wbinnssmith/src/parcel/node_modules/react/cjs/react.development.js",
          "size": 67515,
          "time": 63
        },
        {
          "filePath": "/Users/wbinnssmith/src/parcel/node_modules/prop-types/checkPropTypes.js",
          "size": 3368,
          "time": 26
        },
        {
          "filePath": "/Users/wbinnssmith/src/parcel/node_modules/object-assign/index.js",
          "size": 2108,
          "time": 8
        },
        {
          "filePath": "/Users/wbinnssmith/src/parcel/node_modules/prop-types/lib/ReactPropTypesSecret.js",
          "size": 314,
          "time": 30
        },
        {
          "filePath": "/Users/wbinnssmith/src/parcel/node_modules/react/index.js",
          "size": 182,
          "time": 11
        }
      ],
      "totalAssets": 6
    },
    [...]
    {
      "filePath": "dist/modern/worker.5074a895.js",
      "size": 3473,
      "time": 25,
      "largestAssets": [
        {
          "filePath": "/Users/wbinnssmith/src/parcel/packages/examples/simple/src/worker.js",
          "size": 24,
          "time": 17
        }
      ],
      "totalAssets": 1
    }
  ]
}
```

[Here are the bundle report types](https://github.com/parcel-bundler/parcel/blob/56522b78865646ab7730dc98892a508600419fcb/packages/core/utils/src/generateBundleReport.js#L6) and [here are the json message types](https://github.com/parcel-bundler/parcel/blob/56522b78865646ab7730dc98892a508600419fcb/packages/reporters/json/src/JSONReporter.js#L163)

### DIFF
--- a/packages/core/utils/src/generateBundleReport.js
+++ b/packages/core/utils/src/generateBundleReport.js
@@ -4,13 +4,13 @@ import type {Asset, BundleGraph} from '@parcel/types';
 import nullthrows from 'nullthrows';
 
 export type BundleReport = {|
-  bundles: Array<{
+  bundles: Array<{|
     filePath: string,
     size: number,
     time: number,
     largestAssets: Array<{filePath: string, size: number, time: number}>,
     totalAssets: number
-  }>
+  |}>
 |};
 
 export default function generateBundleReport(
--- a/packages/core/utils/src/generateBundleReport.js
+++ b/packages/core/utils/src/generateBundleReport.js
@@ -1,0 +1,47 @@
+// @flow strict-local
+
+import type {Asset, BundleGraph} from '@parcel/types';
+import nullthrows from 'nullthrows';
+
+export type BundleReport = {|
+  bundles: Array<{
+    filePath: string,
+    size: number,
+    time: number,
+    largestAssets: Array<{filePath: string, size: number, time: number}>,
+    totalAssets: number
+  }>
+|};
+
+export default function generateBundleReport(
+  bundleGraph: BundleGraph,
+  largestAssetCount: number = 10
+): BundleReport {
+  let bundles = [];
+  bundleGraph.traverseBundles(bundle => {
+    bundles.push(bundle);
+  });
+  bundles.sort((a, b) => b.stats.size - a.stats.size);
+
+  return {
+    bundles: bundles.map(bundle => {
+      let assets: Array<Asset> = [];
+      bundle.assetGraph.traverseAssets(asset => {
+        assets.push(asset);
+      });
+      assets.sort((a, b) => b.stats.size - a.stats.size);
+
+      return {
+        filePath: nullthrows(bundle.filePath),
+        size: bundle.stats.size,
+        time: bundle.stats.time,
+        largestAssets: assets.slice(0, largestAssetCount).map(asset => ({
+          filePath: asset.filePath,
+          size: asset.stats.size,
+          time: asset.stats.time
+        })),
+        totalAssets: assets.length
+      };
+    })
+  };
+}
--- a/packages/reporters/json/src/JSONReporter.js
+++ b/packages/reporters/json/src/JSONReporter.js
@@ -1,0 +1,183 @@
+// @flow strict-local
+
+import type {
+  BuildProgressEvent,
+  LogEvent,
+  ParcelOptions,
+  ReporterEvent
+} from '@parcel/types';
+
+import {Reporter} from '@parcel/plugin';
+import generateBundleReport, {
+  type BundleReport
+} from '@parcel/utils/src/generateBundleReport';
+
+/* eslint-disable no-console */
+const writeToStdout = makeWriter(console.log);
+const writeToStderr = makeWriter(console.error);
+/* eslint-enable no-console */
+
+const LOG_LEVELS = {
+  none: 0,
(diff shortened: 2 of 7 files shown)


packages/reporters/json/src/JSONReporter.js
```diff
@@ -0,0 +1,189 @@
...
+    case 'packaging':
+      return {
+        type: 'buildProgress',
+        phase: progressEvent.phase,
+        bundleFilePath: progressEvent.bundle.filePath
+      };
+  }
+}
+
+type JSONReportEvent =
```
wbinnssmith: These are the shapes of the messages we end up writing out

packages/core/utils/src/generateBundleReport.js
```diff
@@ -0,0 +1,47 @@
+// @flow strict-local
+
+import type {Asset, BundleGraph} from '@parcel/types';
+import nullthrows from 'nullthrows';
+
+export type BundleReport = {|
+  bundles: Array<{
```
wbinnssmith: Here's the shape of the bundles that get written in the `buildSuccess` event

DeMoorJasper:
I don't think this should be behind a cli flag or be part of core.
Seems like something that should be part of a custom config or detected by installed deps if no config is present.
Not entirely sure why environmental reporters are important? Looks overcomplicated to me.

wbinnssmith:
I'm thinking of closing this out for now. I'm on board with not including this in core, but I'm concerned it won't be used by much until a use-case presents itself. What do you think?

DeMoorJasper:
I think it's totally fine to add this reporter into the parcel 2 branch just not as part of the default config. Might be cool for people who wanna build stuff around parcel.
Like Github CI bots and such

wbinnssmith:
Okay, I've removed all the environmental reporters and cli stuff. This is now just another reporter that can be included as part of the config. If it is, the cli reporter will need to be removed in those configurations.

wbinnssmith:
This is good for another review. I added a test to make sure this doesn't break since it's not exercised as part of the default config.

PR #2866, opened 2019-03-28
Add a test case for relative urls

mischnic:
# ↪️ Pull Request

As per https://github.com/parcel-bundler/parcel/pull/2740#issuecomment-471371666
To help prevent another disaster....

Running `yarn` in the root always removes `parcel-bundler` from the lockfile, how did it get there?


mischnic:
https://github.com/parcel-bundler/parcel/pull/2842 describes why the tests fail

DeMoorJasper:
Maybe this should be ported to v2?

Issue #2851, opened 2019-03-26
Wrong source map path for CSS files when multiple entry points


hauleth:
# 🐛 bug report

When building application with multiple entrypoints the source map for CSS file is invalid.

## 🎛 Setup project

```
yarn init .
yarn add -D parcel-bundler
mkdir js css
touch js/index.js css/main.css
yarn parcel build js/index.js css/main.css
```

## 😯 Current Behavior

When I check last line of `dist/css/main.css` I get:

```css
/*# sourceMappingURL=/main.css.map */
```

## 🤔 Expected Behavior

CSS should end with proper path for source map:

```css
/*# sourceMappingURL=/css/main.css.map */
```

## 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Parcel           | 1.12.3
| Node             | v9.11.1
| npm/Yarn         | yarn 1.15.2
| Operating System | macOS 10.14.3 (18D109)

PR #2867, opened 2019-03-28
Fix sourceMappingURL for bundles with multiple entry points [CSS]

rynpsc:
# ↪️ Pull Request

Fixes incorrect sourcemap URL's when you have multiple entry points.

Related issue https://github.com/parcel-bundler/parcel/issues/2851

## 💻 Examples

Given the followong CLI command

`parcel build src/a/style.css src/b/style.css`

The following files are produced

```
dist/a/style.css /*# sourceMappingURL=/style.css.map */
dist/b/style.css /*# sourceMappingURL=/style.css.map */
```

The sourcemap URLs are incorrect they should be `/a/style.css.map` and `/b/style.css.map`.

With the fix in place we instead get:

```
dist/a/style.css /*# sourceMappingURL=/a/style.css.map */
dist/b/style.css /*# sourceMappingURL=/b/style.css.map */
```

The same issue was previously fixed with JavaScript sourcemaps https://github.com/parcel-bundler/parcel/pull/2645

## ✔️ PR Todo


rynpsc:
Have updated desciptions and added tests. Currently failing due to https://github.com/parcel-bundler/parcel/pull/2842.

mischnic:
Now consistent with the logic in `JSPackager.js` 👍

rynpsc:
@mischnic Noticed the check for the sourcemap comment was using JS syntax rather than CSS.

mischnic:
> Noticed the check for the sourcemap comment was using JS syntax rather than CSS.

We really need to get CI working again.

Very picky comment: could you rename the vars in that test? (from `jsOutput`)

sammckay10:
Has this been merged?!?

mmcc:
I'm running into this exact issue and it appears this was approved, but closed instead of merged (unless I'm mistaken). Was that an accident?

PR #2871, opened 2019-03-30
Strip `type="module"` attribute from output <script> tags

twome:
# ↪️ Pull Request

This addresses the feature request in  #1401. Because Parcel assumes a non-strict non-module execution context for its output ES5, running the output from a `type="module"` script tag will hit a ReferenceError when `parcelRequire` tries to implicitly define itself as a global variable using undeclared assignment. With this, users won't need to know/remember to only use vanilla <script> tags in their source HTML. 

## 💻 Examples

See #2809's use-case.

## 🚨 Test instructions

Same repro as as #2809's. Only errors on macOS 10.14.2 are `should produce a basic kotlin bundle` (no local Java).

## ✔️ PR Todo


twome:
Please ignore this PR! GitHub automatically closed this, so I made a new copy of it: #2872

PR #2873, opened 2019-03-30
Add `@parcel/eslint-plugin` and a rule to prevent self-package imports

wbinnssmith:
This adds a parcel eslint *plugin* in addition to our config. This allows us to write our own eslint rules to enforce patterns and prevent issues specific to parcel.

The first rule is `no-self-package-imports`, which prevents requiring/importing modules by package name within the same package. e.g. `require('@parcel/core/foo')` while in `@parcel/core/bar` should be `require('./foo')`.

This can easily happen accidentally while refactoring across the monorepo since yarn links all modules in its root `node_modules`, so while these requires resolve fine in the monorepo, they'll break in published modules.
 
Test Plan: `yarn test` within `eslint-plugin`. Also violated the rule and fixed it via `eslint --fix` and VSCode.


packages/dev/eslint-plugin/lib/utils.js
```diff
@@ -0,0 +1,102 @@
...
+      path.dirname(origin),
+      request.replace(new RegExp('^' + pkgName), pkgRoot)
+    )
+    // `require` expects unix-style '/' separators, but `path.relative` will respect
+    // `path.sep`.
+    .split(path.sep)
+    .join('/');
+
+  if (relative.endsWith('.')) {
+    // prefer `require('../')` over `require('..')`
```
wbinnssmith: we do, right? I tested a bit and node doesn't seem to care.

PR #2876, opened 2019-04-01
Fix HMR dependencies

bertfreudenberg:
# ↪️ Pull Request

Currently, when an asset is accepted, only its direct chain of parents is accepted along with it.

But all parents of all accepted assets need to be accepted, otherwise they will hold onto old imports.

This patch introduces a topological sort for dependencies ensuring that all imports are accepted before all their parents.

This order is equivalent to the order happening on initial load.

## 💻 Examples

a.js:

    export default "X";

b.js:

    import A from "./a.js";
    export default function() { return A; }

c.js:

    import A from "./a.js";
    export default function() { return A; }

index.js:

    import B from "./b.js";
    import C from "./c.js";
    console.log(B(), C());
    module.hot.accept();   

On initial load, this prints `X X`.

When changing `"X"` to `"Y"` in a.js, after hot-reloading it prints `X Y`.

With my patch, it prints `Y Y` because both b.js and c.js get reloaded.


packages/core/parcel-bundler/src/builtins/hmr-runtime.js
```diff
@@ -20,29 +20,32 @@ function Module(moduleName) {
...
-          var didAccept = hmrAcceptCheck(global.parcelRequire, asset.id);
-          if (didAccept) {
-            handled = true;
-          }
-        }
+        if (!asset.isNew) assetsToCheck.push(asset.id);
       });
+      var handled = false;
+      var id;
+      while (id = assetsToCheck.shift()) {
```
DeMoorJasper: Could you change this to something like

```JavaScript
while (id) {
  ...
  id = assetsToCheck.shift()
}
```

Looks a little less complex.
Everything else looks fine.

bertfreudenberg: Good point - fixed.

bertfreudenberg:
@DeMoorJasper think this is good to go now? Who can approve it?

DeMoorJasper:
@bertfreudenberg yes, just gotta test if tests still pass and have a more in depth review

DeMoorJasper:
This is not gonna end up in v1, due to v1 no longer receiving new features. Only critical fixes.

Gonna close this

PR #2884, opened 2019-04-02
Only patch `console` in workers. Ignore their stdout.

wbinnssmith:
This removes the Logger's ability to patch `console.log` and friends. This was used in reporters that have sensitive stdio (such as the ink and JSON reporters) to prevent things from disrupting terminal output. This happened when plugins wrote to their stdout/stderr, which we have no control over. This currently only happens when [Browserslist writes a warning to stderr when its database is out of date](https://github.com/browserslist/browserslist/issues/361).

Instead, don't inherit subprocess stdio from the main process -- create a local pipe that transforms the stdio from the subprocess into lines sent to the logger. This will allow us to tag these log messages in the future as originating from a particular worker.

Note: it's still possible that plugins which run in the main process (Bundlers, Reporters, etc) could write to stdio and disrupt things.

Test Plan: 
  * Wrote integration tests in the workerfarm module to cover this behavior (`yarn test`).
  * Temporarily added `console.log`, `process.stio.write`, and `Logger` calls in the babel transform and verified multiple sets of output came through the main process's stdout/stderr in the simple example.

### DIFF
--- a/packages/core/workers/src/Worker.js
+++ b/packages/core/workers/src/Worker.js
@@ -48,13 +48,11 @@ export default class Worker extends EventEmitter {
       v => !/^--(debug|inspect)/.test(v)
     );
 
-    let options = {
+    this.child = childProcess.fork(childModule, process.argv, {
       execArgv: filteredArgs,
       env: process.env,
       cwd: process.cwd()
-    };
-
-    this.child = childProcess.fork(childModule, process.argv, options);
+    });
 
     // Unref the child and IPC channel so that the workers don't prevent the main process from exiting
     this.child.unref();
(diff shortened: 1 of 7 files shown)


wbinnssmith:
cc @devongovett re: CLIReporter

packages/core/workers/src/Worker.js
```diff
@@ -48,13 +49,23 @@ export default class Worker extends EventEmitter {
...
       execArgv: filteredArgs,
       env: process.env,
-      cwd: process.cwd()
-    };
+      cwd: process.cwd(),
+      stdio: ['ignore', 'pipe', 'pipe', 'ipc']
+    });
 
-    this.child = childProcess.fork(childModule, process.argv, options);
+    this.child.stdout.on('data', data => {
```
devongovett: Are these chunked into lines automatically by node or could the chunks be shorter/longer than a line?

wbinnssmith: Good question. Afaict, Node doesn't chunk anything; newlines exist in the stream because they're written there by console.log (which puts them in automatically) or writing to stdout directly with a newline at the end.

Splitting by newline definitely assumes that new lines map to new log messages. I'm not 100% about this since e.g. `console.log('foo\nbar')` would get treated as two separate messages.

Conversely, Node buffers writable streams and is inconsistent about when it flushes `data` events, which is why this splitting exists in the first place -- before splitting, tests would fail inconsistently because the stream would emit a `data` event containing more than one logical message.

I can look a bit more into this.

wbinnssmith:
I don’t think intercepting stdout/stderr in the workers is going to work terribly well. We don’t have enough control over when Node flushes the writable streams to discern individual messages. Plus, writing arbitrary bytes to a stream (stdout/stderr) doesn't map cleanly to log messages.

VSCode has a nice extension model wherein extensions run in their own process (the extension host) and `console.log`-ing within that process appends to a log specific to that extension host.

I think we should go back to patching `console`, but this time only inside the worker processes. Within the worker process, they can be adapted to invoke the logger, which will get forwarded to the main process.

This means that direct writes to `stdout` in the worker process will be ignored. VSCode appears to do the same. Plugins shouldn't expect to do anything with the raw `stdout` stream anyway, and it wouldn't be portable in other environments Parcel runs/will run in (like the browser). We should probably document this behavior.

Going to update this PR to reflect this since implementing this should be pretty straightforward. Let me know your thoughts @padmaia and @devongovett.

wbinnssmith:
Okay, I've updated the PR to reflect the proposal above.

packages/core/workers/src/Worker.js
```diff
@@ -151,30 +149,31 @@ export default class Worker extends EventEmitter {
...
       let call = this.calls.get(idx);
       if (!call) {
         // Return for unknown calls, these might accur if a third party process uses workers
         return;
       }
 
-      if (contentType === 'error') {
-        call.reject(jsonToError(content));
+      if (message.contentType === 'error') {
+        message;
```
devongovett: ?

PR #2892, opened 2019-04-05
First iteration of Parcel 2 Plugin API

wbinnssmith:
This implements a first iteration of a Parcel 2 Plugin API. The best way to get an understanding of how to use the API is to look at the existing Parcel plugins that were adapted to it. I'll point them out inline.

A number of things have changed:

* A new directory in core called "public", which stores all modules that implement public API.
* We now give plugins facades for our most complex structures. All of our prior structures still exist, though some of their APIs only used by plugins were moved to their public wrappers:
    * `Bundle` (read-only)
    * `MutableBundle`
    * ~~`RuntimeBundle` (specific to runtimes -- a read-only bundle with a `addRuntimeAsset` method)~~
    * `FulfilledBundle` - A bundle with an immutable, but guaranteed `filePath`
    * `MainAssetGraph` (read-only)
    * `BundleGraph`
        * currently writable, but there should be a read-only version for reporters
        * The name `BundleGraph` doesn't really fit with this public API anymore, as none of the APIs we expose are graph-like. We should change it.
* GraphViz support currently removed -- ~~will be restored in a following PR~~ restored in #2905 as a reporter using undocumented apis on the public interfaces.

Notes:
* I believe we should take care to only wrap structures in public APIs right before they are handed to plugins:
    * Working with the public api wrappers within core itself makes it difficult to understand which version of a structure you should expect. 
    * If we used the public versions across workers within core, we'd have to teach them to de/serialize, which involves exposing a method that plugins might attempt to use. By wrapping at the last moment, we avoid this.
* This grew out of exploring options in @devongovett's `wip/graph-api` branch, and incorporates some of its code.

Concerns:
* Naming is hard. Modules that call out to plugins often need to reference types/concrete classes for both the public and private structures. Maybe we should adopt a convention to lead with `Public`/`Private`, etc?
* The public `BundleGraph` currently needs to access the internal `Bundle` structure from public bundles it receives. This is currently implemented with an exposed `WeakMap` that every bundle registers itself with. I'm not sure of a cleaner way of accomplishing this. 

Test Plan: flow, test. Verify simple example output is exactly the same as on the v2 branch.


packages/bundlers/default/src/DefaultBundler.js
```diff
@@ -1,5 +1,5 @@
 // @flow
```
wbinnssmith: Calling out `DefaultBundler` as an example of a bundler using the api.

packages/packagers/js/src/JSPackager.js
```diff
@@ -11,19 +11,19 @@ const PRELUDE = fs
 export default new Packager({
```
wbinnssmith: Calling out `JSPackager` as an example of a packager using the api.

packages/runtimes/js/src/JSRuntime.js
```diff
@@ -1,8 +1,5 @@
 // @flow
```
wbinnssmith: Calling out `JSRuntime` as an example of a runtime using the api. This one cleaned up quite a bit.

packages/core/types/index.js
```diff
@@ -294,129 +293,78 @@ export type GraphTraversalCallback<TNode, TContext> = (
...
-  from: NodeId,
-  to: NodeId
-|};
-
-export type GraphUpdates<TNode> = {|
-  added: Graph<TNode>,
-  removed: Graph<TNode>
-|};
-
-export interface Graph<TNode: Node> {
```
wbinnssmith: These were removed as plugins no longer rely on these interfaces. `@parcel/types` is now only types belonging to the public api or types shared among parcel's own packages (which should overlap significantly)

Modules internal to core can still reference types from the concrete versions through e.g. `import type Graph from './Graph';`

wbinnssmith:
A question for reviewers: what other structures internal to Parcel that are currently exposed to plugins are you uncomfortable with as public API? `Asset` comes to mind as another we should adapt.

packages/runtimes/js/src/JSRuntime.js
```diff
@@ -74,14 +57,15 @@ export default new Runtime({
...
-        env: bundle.env,
         code: `module.exports = require('./bundle-loader')([${loaderModules.join(
           ', '
-        )}, ${JSON.stringify(bundleGroup.value.entryAssetId)}]);`
+        )}, ${JSON.stringify(bundleGroup.entryAssetId)}]);`,
+        bundleGroup
       });
     }
+
+    return assets;
```
wbinnssmith: moved runtime plugins to a declarative api, rather than imperatively calling `addRuntimeAsset`.

Issue #2887, opened 2019-04-04
--no-autoinstall not working because of capitalization typo in source


SirUppyPancakes:
# 🐛 bug report

In `cli.js` an option called `--no-autoinstall` produces a value in the final `options` object called `autoinstall`, while in `Bundler.js` the `normalizeOptions` function expects a value called `autoInstall`. Since `autoInstall` is never declared, the auto install feature cannot be disabled via the flag as expected.

## 🤔 Expected Behavior

`--no-autoinstall` should work

## 😯 Current Behavior

`--no-autoinstall` is ignored

## 💁 Possible Solution

Rewrite Parcel in TypeScript and catch bugs like these at compile-time 😉 

## 🔦 Context

Jokes aside, I'd like to use this opportunity to emphasize that, while I am personally a big advocate of "powerful aggressive defaults", I think that auto-installing arbitrary packages (especially where developer typos are guaranteed) is more harmful than helpful, at least not by a sheer import. Perhaps the feature could somehow check if any of the API was actually used after the import... but even then, that would be quite the complicated feature.

## 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Parcel           | 1.12.3
| Node             | v11.4.0
| Yarn         | v1.12.3
| Operating System | Windows 10 x64

ciaranashton:
I would like to have a look into a solution for this problem if no one else is?

ciaranashton:
I know there is a bigger question here around defaults, however, I have got a simple fix PR up so people can continue to use the feature in the time being.

#2898

ideoclickVanessa:
Is there a current work around for this issue until the new release?

My local dev environment is currently broken due to `parcel watch --no-autoinstall` attempting and failing to download my custom module from npm; thus, refusing to actually build the project.

SirUppyPancakes:
I had been editing the source in `node_modules` which isn't great since any `npm` or `yarn` commands overwrite your changes, but you might have some luck with environment variables, since it looks like `Bundler.js` sets the `autoInstall` feature based on some environment stuff as another level of configuration: https://github.com/parcel-bundler/parcel/blob/636bb50baeae15730c3b91cd3b58b457c31f7200/packages/core/parcel-bundler/src/Bundler.js#L147

krishnagopinath:
In case someone else wanted to get this to work, I set it up this way and everything worked as expected!

```
[...]

PR #2898, opened 2019-04-06
Fix capitalization typo in source for --no-autoinstall

ciaranashton:
# ↪️ Pull Request

Fixes bug #2887 

See description of the bug above.

When using the CLI to serve or watch with the `--no-autoinstall` it will now correctly check for the flag.

The changes on `108`, `109`, `148` and `149` are changed due to the `npm run -s precommit` command.

## 🚨 Test instructions

Without unit tests for the cli instructions this isn't covered. There is an option, however, for adding end to end tests. Full flow from cli instruction to desired build. I would be interested in looking into adding this if requested.

You can manually test this by adding the flag to a simple project and seeing that the dependencies will no longer bu automatically installed.

## ✔️ PR Todo


packages/core/parcel-bundler/src/Bundler.js
```diff
@@ -142,11 +142,11 @@ class Bundler extends EventEmitter {
       detailedReport: options.detailedReport || false,
       global: options.global,
       autoinstall:
-        typeof options.autoInstall === 'boolean'
-          ? options.autoInstall
+        typeof options.autoinstall === 'boolean'
```
DeMoorJasper: This would be a breaking change.
We should check both `options.autoInstall` and `options.autoinstall` or set `options.autoInstall = options.autoinstall` in the cli

ciaranashton: Thanks for you fast response @DeMoorJasper. 
I prefer the latter, this way we don't need to change the logic of the actual Bundler. I have made the changes.
I am however having issues with a test that isn't related to the work I have been doing, hence the failing tests. Is this a known issue?

mischnic:
Strangely, `autoInstall` was never added to the API docs: https://parceljs.org/api.html

ciaranashton:
@mischnic I'm happy to get an update in for that as well. Also, could you point me in the direction of an explanation around the current issue with the failing test blocking merges?

mischnic:
@ciaranashton That was more of a general finding, not a request 😄 

As for the CI: this PR took a stab at it #2842

DeMoorJasper:
Gonna close this as v1 isn't really receiving more updates

PR #2899, opened 2019-04-07
Include the copyright and BSD license for the dotenv types

wbinnssmith:
The types and comments are taken from the source for dotenv, so abide by its license and include the full BSD text.

PR #2900, opened 2019-04-07
Upgrade to dotenv 7 and use its types

wbinnssmith:
This upgrades to dotenv 7, which is written with Flow, so no libdefs are needed.

Supersedes #2899

Test Plan: Add a `.env` in the simple example, add a `console.log` in `worker.js`, run the simple example, and verify values from `.env` are logged.


wbinnssmith:
Here's the CHANGELOG for `dotenv`: https://github.com/motdotla/dotenv/blob/03a891554c49915fe919c649f51b3adcae662a84/CHANGELOG.md

There don't seem to be any notable breaking changes, except for maybe motdotla/dotenv#376, which would only happen with unbalanced quotes anyway.

PR #2904, opened 2019-04-08
Make connectedFiles and dependencies Maps

wbinnssmith:
This prevents them from containing multiple references to the same file/dependency. Currently in v2 assets contain many duplicate `connectedFiles`

This also exposes a public api for retrieving them as Arrays. I'm thinking this, along with `addConnectedFile`/`addDependency`, should be the only public api to these structures when we wrap `Asset`.

In each case, the stored value is overwritten in when the new value is added.

Test Plan: wrote unit tests for `Asset` (`yarn test`)

### DIFF
--- a/packages/core/core/src/Dependency.js
+++ b/packages/core/core/src/Dependency.js
@@ -13,7 +13,6 @@ import {md5FromString} from '@parcel/utils/src/md5';
 
 type DependencyOpts = {|
   ...DependencyOptions,
-  moduleSpecifier: ModuleSpecifier,
   env: IEnvironment,
   id?: string,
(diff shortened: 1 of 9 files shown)


packages/core/core/src/Dependency.js
```diff
@@ -13,7 +13,6 @@ import {md5FromString} from '@parcel/utils/src/md5';
 
 type DependencyOpts = {|
   ...DependencyOptions,
-  moduleSpecifier: ModuleSpecifier,
```
wbinnssmith: this is already a part of `DependencyOptions`

PR #2905, opened 2019-04-08
Reintroduce graphviz dumping as a development-only option

wbinnssmith:
~~This implements graphviz dumping as a reporter. Right now, it's included in the default config, which feels wrong. It only reports when the `PARCEL_DUMP_GRAPH` environment variable is present.~~

~~It also needs access to the internals of each graph to provide useful debugging information, so undocumented `_dumpToGraphViz` methods are exposed on each graph, and are invoked in the reporter. This does not leak any information or control to plugins as it only has a side effect of writing the image to disk and returning a path to it.~~

This is now a development-only option. `graphviz` is now a `devDependency`. We now also early return in a way that should be able to be optimized out in production builds of parcel.

### DIFF
--- a/packages/core/core/src/dumpGraphToGraphViz.js
+++ b/packages/core/core/src/dumpGraphToGraphViz.js
@@ -1,0 +1,116 @@
+// @flow
+
+import type {Environment} from '@parcel/types';
+
+import type Graph from './Graph';
+import type {AssetGraphNode, BundleGraphNode} from './types';
+
+import invariant from 'assert';
+import nullthrows from 'nullthrows';
+import path from 'path';
+import BundleGraph from './BundleGraph';
+
+const COLORS = {
+  root: 'gray',
+  asset: 'green',
+  dependency: 'orange',
+  transformer_request: 'cyan',
+  file: 'gray',
+  default: 'white'
+};
+
+export default async function dumpGraphToGraphViz(
+  // $FlowFixMe
+  graph: Graph<AssetGraphNode> | Graph<BundleGraphNode>,
+  name: string
+): Promise<void> {
+  if (
+    process.env.PARCEL_BUILD_ENV === 'production' ||
+    process.env.PARCEL_DUMP_GRAPHVIZ == null
+  ) {
+    return;
+  }
+
+  const graphviz = require('graphviz');
+  const tempy = require('tempy');
+
+  let g = graphviz.digraph('G');
+
+  let nodes = Array.from(graph.nodes.values());
+  for (let node of nodes) {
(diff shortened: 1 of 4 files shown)


packages/core/core/src/AssetGraph.js
```diff
@@ -317,4 +318,8 @@ export default class AssetGraph extends Graph<AssetGraphNode> {
       value: asset
     });
   }
+
+  _dumpToGraphViz(name: string): Promise<string> {
```
DeMoorJasper: Shouldn't this become part of the reporter instead of core?

wbinnssmith: If it were part of the reporter, we'd have to expose the raw graph as part of public API, which I don't think we want to do. I agree this isn't optimal. I'm open to suggestions 😄

DeMoorJasper: Ow thought we were gonna sent a minimal graph to the reporter that was standardised. My bad.

wbinnssmith: We do exactly that 😄 : https://github.com/parcel-bundler/parcel/blob/d7e740e46f1a8de564a7706159772ff9d4e00640/packages/core/core/src/Parcel.js#L128

However, this reporter needs access to the raw graph 😞

DeMoorJasper: Ow alright just thought the public api graph was more complete then.
Guess I'll have to look a little bit more into how minimal it actually is.

packages/core/core/src/dumpGraphToGraphViz.js
```diff
@@ -0,0 +1,116 @@
...
+  default: 'white'
+};
+
+export default async function dumpGraphToGraphViz(
+  // $FlowFixMe
+  graph: Graph<AssetGraphNode> | Graph<BundleGraphNode>,
+  name: string
+): Promise<void> {
+  if (
+    process.env.PARCEL_BUILD_ENV === 'production' ||
```
wbinnssmith: We could use `NODE_ENV` here, but `parcel build` itself sets this to `production`, making this not an option in places like the simple example.

packages/core/core/src/dumpGraphToGraphViz.js
```diff
@@ -0,0 +1,116 @@
...
+    process.env.PARCEL_DUMP_GRAPHVIZ == null
+  ) {
+    return;
```
wbinnssmith: Is DCE good enough to compile out code following an early return? Might need wrap the following code in an `if` with inverted logic.

PR #2917, opened 2019-04-11
Cherry-pick CSS Sourcemaps

wbinnssmith:
This forward-ports @mischnic's work on css sourcemaps into the v2 branch.

Note that this is a cherry-pick of #2489 as well as two other commits necessary for existing tests to pass. The css integration tests continue to be skipped, and it's possible we're missing more code that we don't exercise yet. A full merge of master back into v2 should resolve that.

### DIFF
--- a/packages/core/integration-tests/test/css.js
+++ b/packages/core/integration-tests/test/css.js
@@ -17,7 +17,11 @@ describe.skip('css', function() {
         {
           name: 'index.css',
           assets: ['index.css', 'local.css'],
-          childBundles: []
+          childBundles: [
+            {
+              type: 'map'
+            }
+          ]
         }
       ]
     });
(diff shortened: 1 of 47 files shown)


DeMoorJasper:
Perhaps it would be a better idea to try and merge master into v2 instead?
I don't really see a reason to cherypick merge this instead as Sourcemaps in v2 should probably be less hacky.

wbinnssmith:
> I don't really see a reason to cherypick merge this instead as Sourcemaps in v2 should probably be less hacky.

I mostly want the tests, but I suppose everything could just be copied over.

Merging master into v2 is a pretty big undertaking that I probably won't have much context on 😄 I'd definitely be on board with that but I'm not the best to do so.

PR #2918, opened 2019-04-11
Implement Basic CSS Packager

wbinnssmith:
This implements a basic css packager for Parcel 2:
* The css packager currently just concatenates the bundle's css together . No support for source maps yet.
* The default bundler now removes non-matching assets when assembling bundles (e.g. it does not include css text in a js bundle). It leaves behind asset references in their place.
* `AssetGraph` now exposes `traverseAssetsWithReferences` which iterates both "real" assets as well as references.
* In the case of asset references, the JS Packager no longer includes content, and instead leaves behind an empty no-op function.

A no-op will not always be the desired behavior for importing non-matching assets: for example CSS Modules export a mapping, and raw assets export a path. @padmaia and I believe that it might be best in those cases for transformers to return an additional JS Asset containing code with that behavior. Let me know if this is in line with what you might have had in mind, @devongovett.

Test Plan: Run the simple example with the new html docs. Verify headings and paragraphs are colored.

### DIFF
--- a/packages/bundlers/default/src/DefaultBundler.js
+++ b/packages/bundlers/default/src/DefaultBundler.js
@@ -1,5 +1,5 @@
 // @flow
-import type {BundleGroup, Bundle, MutableBundle} from '@parcel/types';
+import type {BundleGroup, MutableBundle} from '@parcel/types';
 import {Bundler} from '@parcel/plugin';
 
 const OPTIONS = {
@@ -10,7 +10,7 @@ const OPTIONS = {
 
 type BundleContext = {|
   bundleGroup: BundleGroup,
-  bundle?: Bundle
+  bundle?: MutableBundle
 |};
 
 export default new Bundler({
@@ -61,13 +61,22 @@ export default new Bundler({
 
             // Mark bundle as an entry, and set explicit file path from target if the dependency has one
             bundle.isEntry = !!dep.isEntry;
-            if (dep.target && dep.target.distPath) {
-              bundle.filePath = dep.target.distPath;
+            let target = dep.target;
+            if (
+              target &&
+              target.distPath &&
+              target.distPathType === bundle.type
+            ) {
+              bundle.filePath = target.distPath;
             }
 
             // If there is a current bundle, but this asset is of a different type,
             // separate it out into a parallel bundle in the same bundle group.
             if (context.bundle) {
+              // Remove this asset from the current bundle since it's of a different type.
+              // `removeAsset` leaves behind an asset reference in its place.
+              context.bundle.removeAsset(node.value);
+
               let bundles = bundleGraph.getBundlesInBundleGroup(
                 context.bundleGroup
               );
--- a/packages/core/integration-tests/test/css.js
+++ b/packages/core/integration-tests/test/css.js
@@ -30,8 +30,7 @@ describe('css in v2', () => {
     assert.equal(output(), 3);
   });
 
-  // Currently failing as Windows produces different hashes
-  it.skip('should support loading a CSS bundle along side dynamic imports', async () => {
+  it('should support loading a CSS bundle along side dynamic imports', async () => {
     let b = await bundle(
       path.join(__dirname, '/integration/dynamic-css/index.js')
     );
@@ -48,8 +47,8 @@ describe('css in v2', () => {
           'JSRuntime.js'
         ]
       },
-      {name: 'local.0feb842b.js', assets: ['local.js']},
-      {name: 'local.6f071801.css', assets: ['local.css']},
+      {name: /local\.[0-9a-f]{8}\.js/, assets: ['local.js']},
+      {name: /local\.[0-9a-f]{8}\.css/, assets: ['local.css']},
       {name: 'index.css', assets: ['index.css']}
     ]);
 
--- a/packages/core/integration-tests/test/css.js
+++ b/packages/core/integration-tests/test/css.js
@@ -20,7 +20,7 @@ describe('css in v2', () => {
         assets: ['index.js', 'local.js']
       },
       {
-        name: 'local.css',
+        name: 'index.css',
         assets: ['index.css', 'local.css']
       }
     ]);
--- a/packages/core/integration-tests/test/css.js
+++ b/packages/core/integration-tests/test/css.js
@@ -1,91 +1,65 @@
 const assert = require('assert');
 const path = require('path');
 const fs = require('@parcel/fs');
-const {bundle, run, assertBundleTree, rimraf, ncp} = require('./utils');
+const {
+  bundle,
+  run,
+  assertBundles,
+  assertBundleTree,
+  rimraf,
+  ncp
+} = require('./utils');
 
-describe.skip('css', function() {
-  it('should produce two bundles when importing a CSS file', async function() {
+describe('css in v2', () => {
+  it('should produce two bundles when importing a CSS file', async () => {
     let b = await bundle(path.join(__dirname, '/integration/css/index.js'));
 
-    await assertBundleTree(b, {
-      name: 'index.js',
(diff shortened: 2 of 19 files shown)


packages/bundlers/default/src/DefaultBundler.js
```diff
@@ -61,13 +62,22 @@ export default new Bundler({
 
             // Mark bundle as an entry, and set explicit file path from target if the dependency has one
             bundle.isEntry = !!dep.isEntry;
-            if (dep.target && dep.target.distPath) {
-              bundle.filePath = dep.target.distPath;
+            let target = dep.target;
+            if (
+              target &&
+              target.distPath &&
+              target.distPathType === bundle.type
```
wbinnssmith: Previously all entries would have the name `distPath` (css was written to `index.js` for example). I added `distPathType` to identify the specific entry this was intended for. Not sure about the case with multiple entries of this type.

packages/examples/simple/legacy.html
```diff
@@ -0,0 +1,24 @@
+<!DOCTYPE html>
```
wbinnssmith: I found these useful for a manual end-to-end test of the simple example, so I'm checking them in.

packages/packagers/js/src/JSPackager.js
```diff
@@ -18,29 +18,41 @@ export default new Packager({
...
+      }
 
-      let dependencies = bundle.assetGraph.getDependencies(asset);
-      for (let dep of dependencies) {
-        let resolved = bundle.assetGraph.getDependencyResolution(dep);
-        if (resolved) {
-          deps[dep.moduleSpecifier] = resolved.id;
+      let wrapped = first ? '' : ',';
+      if (type === 'asset_reference') {
+        wrapped += JSON.stringify(asset.id) + ':[0,{}]';
```
wbinnssmith: Instead of writing in a no-op function each time, pass `0` as a sentinel indicating this is a stub. Alternatives:

* declare a `noop` function and reference it here. This will require polluting `globalThis` or wrapping the prelude and this code in an iifee. This means the prelude doesn't need to change either, and keeps the shapes of these tuples predictable.
* passing `null` or `undefined` rather than `0`, which adds more bytes to the output source text and is somewhat more reasonable.

I decided to go with `0`, at the cost of complicating the shapes of these tuples, to save bytes and minimize polluting `globalThis`. Open to suggestions.

packages/packagers/js/src/prelude.js
```diff
@@ -47,13 +47,16 @@ parcelRequire = function(modules, cache, entry, globalName) {
...
 
-      modules[name][0].call(
-        module.exports,
-        localRequire,
-        module,
-        module.exports,
-        this
-      );
+      var moduleFn = modules[name][0];
+      if (moduleFn) {
```
wbinnssmith: This handles the case the value here is `0`.

packages/packagers/js/src/JSPackager.js
devongovett: I wouldn't worry too much about optimizing it. For prod builds, the tree shaking implementation will take care of most of this.

packages/core/core/src/AssetGraph.js
```diff
@@ -281,6 +285,23 @@ export default class AssetGraph extends Graph<AssetGraphNode>
     }, startNode);
   }
 
+  traverseAssetsWithReferences(
```
devongovett: hmmm... wondering about adding a whole new method for this. Any reason to not just use `traverse` directly?

packages/packagers/js/src/JSPackager.js
wbinnssmith: How would it? 

Another thought: If these were `null` rather than `0` I could imagine terser creating a common reference, which would probably make this pretty negligible.

packages/core/core/src/AssetGraph.js
wbinnssmith: This is used in the packager, which we don't want access to the raw graph (see the public api discussion).

I could maybe see an option in `traverseAssets`, but it would overload the signature too much imo.

wbinnssmith: One thing we could do is expose the more limited `traverse`[0] we have on the main graph on the bundles themselves, and add `asset_reference`s as well. That would address this nicely.

[0] https://github.com/parcel-bundler/parcel/pull/2892/files#diff-2ae149d0ae6e33d140f738e72ea8e8b1R54

packages/packagers/js/src/JSPackager.js
devongovett: Well with the tree shaker, references to modules that don't exist are just replaced with `{}`. No need for a function at all.

packages/core/integration-tests/test/css.js
```diff
@@ -1,91 +1,65 @@
...
+      }
+    ]);
 
     let output = await run(b);
     assert.equal(typeof output, 'function');
     assert.equal(output(), 3);
   });
 
-  it('should support loading a CSS bundle along side dynamic imports', async function() {
+  // Currently failing as Windows produces different hashes
```
wbinnssmith: should we try to address this and achieve consistent hashes across build platforms? Tools might behave differently and make this unpredictable so I'm not sure it's realistic.

Probably for these tests we should disable content digests or teach `assertBundle` to ignore them?

packages/namers/default/src/DefaultNamer.js
```diff
@@ -1,60 +1,82 @@
...
+    }
 
-      // Get a content hash for this bundle, for long term caching.
-      let hash = getHash(bundle);
-      name = `${name}.${hash.slice(-8)}.${bundle.type}`;
+    let firstBundleInGroup = bundleGroupBundles[0];
+
+    // Base split bundle names on the first bundle in their group.
+    // e.g. if `index.js` imports `foo.css`, the css bundle should be called
+    //      `index.css`.
```
wbinnssmith: after this change, the simple example now outputs the following tree:

```
dist/
├── legacy
│   ├── async.5dfad0ea.css
│   ├── async.8feed3b6.js
│   ├── async.db321f18.js
│   ├── async2.2655b727.js
│   ├── index.css
│   ├── index.js
│   └── worker.5074a895.js
└── modern
    ├── async.5dfad0ea.css
    ├── async.bf6ddf52.js
    ├── async.db321f18.js
    ├── async2.2655b727.js
    ├── index.css
    ├── index.js
    └── worker.5074a895.js
```

packages/namers/default/src/DefaultNamer.js
```diff
@@ -1,60 +1,82 @@
...
+import assert from 'assert';
 import crypto from 'crypto';
 import path from 'path';
 
 const COMMON_NAMES = new Set(['index', 'src', 'lib']);
 const DEFAULT_DIST_DIR = 'dist';
 
 export default new Namer({
-  name(bundle, opts) {
+  name(bundle: Bundle, bundleGraph: BundleGraph, opts: NamerOptions): FilePath {
```
jamiekyle-eb: btw one of the reasons I originally wanted to use `namer({ ... })` or `new Namer({ ... })` was so that type systems could automatically infer these types for you without having to add type annotations to them. 

If you're needing to manually type them, you might want to look into the type definition for `Namer` so that it does the work for you

packages/core/integration-tests/test/css.js
```diff
@@ -1,91 +1,64 @@
...
+      {name: 'index.css', assets: ['index.css']}
+    ]);
 
     let output = await run(b);
     assert.equal(typeof output, 'function');
     assert.equal(await output(), 3);
   });
+});
 
+describe.skip('css', function() {
```
devongovett: should the old tests be removed/moved?

wbinnssmith: I'll be getting more of them passing soon in following PRs 😃

PR #2928, opened 2019-04-16
Upgrade flow to 0.97

wbinnssmith:
I'm a little concerned about facebook/flow#7545, so we'll keep an eye on things and revert if necessary, but so far based on the test plan things seem fine.

Test Plan: 
  * `yarn flow`. 
  * Change the `version` field in `.flowconfig` and verify `yarn flow` does not start without matching versions
  * Exercise `flow-for-vscode`'s autocomplete, go-to-definition, diagnostics, etc.

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -25,3 +25,6 @@ unclear-type
 unsafe-getters-setters
 untyped-import
 untyped-type-import
+
+[version]
+0.97.0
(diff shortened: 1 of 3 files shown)


wbinnssmith:
Note: This also changes `flow-bin` to use an exact version. Flow doesn't follow semver and it's pretty common for things to break in minor versions. It also adds the same restriction to `.flowconfig` to make sure a non-local version of flow doesn't accidentally run on the project.

Issue #2635, opened 2019-02-12
parcel: not found with yarn (pnp)


dhoko:
# 🐛 bug report

I can't use parcel (_npm scripts_) if I have yarn with pnp active.

## 🤔 Expected Behavior

```
$ yarn start
yarn run v1.13.0
$ parcel index.html
Server running at http://localhost:1234 
✨  Built in 458ms.
```

As it is if pnp is disable.

## 😯 Current Behavior

`/bin/sh: 1: parcel: not found `

![screenshot from 2019-02-12 15-04-59](https://user-images.githubusercontent.com/713283/52640781-98ec9a80-2ed7-11e9-9ab1-4f91a2bc5ad7.png)

## 🔦 Context

I wanted to try using yarn + pnp

## 💻 Code Sample

https://github.com/dhoko/demo-parcel

## 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| nvm              | 0.26.1
| Node             | v11.6.0 
| Yarn             | 1.13.0
| Operating System | Ubuntu 18.04.1 LTS

mischnic:
I'm quite sure this is the same issue as https://github.com/yarnpkg/yarn/issues/6917

quasicomputational:
This is definitely caused proximally by that Yarn bug, but the reason Parcel is ending up unplugged is because it has a postinstall script, which exists for non-technical purposes. If you install it with `--ignore-scripts`, then you can run the `parcel` command.

But it very likely won't actually work, because Parcel doesn't know about PnP resolution yet. I had a brief look at `Resolver.js`, but it's not an entirely trivial change and would need tests, so, yeah.

frunkad:
Anytime when this will be fixed?

DeMoorJasper:
@frunkad it'll be fixed in Parcel 2, see #3582

mischnic:
Closing as a non-parcel-bug

PR #2942, opened 2019-04-22
Adds support for PnP

arcanis:
# ↪️ Pull Request

This PR adds support for PnP to the Parcel resolver.

Related: #2635 

## 💻 Examples

Parcel running under [Berry](https://github.com/yarnpkg/berry) now works 💃

## 🚨 Test instructions

Unit tests usually are a bit impractical since it requires to run under a different environment (since it relies on `process.versions.pnp`). Any idea? I tested it locally by following the steps in the getting started (absolutely great experience btw!) and adding a `require('lodash')` in `index.js`.

To reproduce:

- Make a new project with Yarn and cd' into it
- Run `yarn policies set-version berry`
- Run `yarn --version` to confirm you're running with the v2
- Run `yarn add lodash` and add the `require('lodash')` to `index.js`
- Run `yarn parcel index.html`

## ✔️ PR Todo


DeMoorJasper:
Is there a way you could add a test for this?

arcanis:
Just pushed a commit with a test. I cheat a bit by modifying (then restoring) the environment but it should be fine as long as the tests don't run in parallel 🤞

I also had to commit the result of `yarn install && yarn format`, as they were changing other files than the ones affected by my PR.

packages/core/parcel-bundler/src/Resolver.js
```diff
@@ -60,7 +60,17 @@ class Resolver {
     let module = await this.resolveModule(filename, parent);
```
devongovett: We may want to adjust this slightly so `resolveModule` is not called when using pnp, since that will still look for a node_modules directory (it calls `findNodeModulePath`). We'll need to split out the rest of that method so that things like aliases and builtins are still applied, but it doesn't search the filesystem.

devongovett: Alternatively, `resolveModule` could also check for pnp and skip looking for node_modules if it is available.

devongovett:
Does this also support pnp in yarn v1? Should it?

packages/core/parcel-bundler/src/Resolver.js
arcanis: Good idea, done 👍

arcanis:
@devongovett I haven't tested but it should, yes. The `pnpapi` package is exactly the same from the consumer perspective - only the internal implementation changes a bit.

Of course the v2 is so much better anyway 🤡

arcanis:
Any blocker?

arcanis:
Ping ? 😊

mischnic:
@arcanis There is a lint error:
```
/Users/vsts/agent/2.150.3/work/1/s/packages/core/integration-tests/test/pnp.js
  2:7  error  'fs' is assigned a value but never used  no-unused-vars
```

texastoland:
I'm actively waiting on this to support some courses on egghead.io. Is there are part of the change that's too drastic?

codyaverett:
This would be nice to have now.  Are you all holding off on this for v2?

DeMoorJasper:
Yes, this PR should be implemented for v2. Alpha's gonna come out soon and Parcel 1 should only receive bugfixes at this point

DeMoorJasper:
closing in favor of #3582

PR #2958, opened 2019-04-25
Replace bundles with bundle references when merging bundle graph into bundles

wbinnssmith:
When adding new bundles to the bundle graph, we merge in a subgraph of the bundle graph into the bundle's own asset graph. This allows it to trace dependencies to other bundles and bundle groups.

However, merging this subgraph has a recursive effect: ancestor bundles will actually contain all the bundles of their children and every node, which makes serialization and IPC expensive, and sometimes impossible in the case a child bundle refers to its parent (see the `html-root` integration test).

Instead, convert bundle nodes to bundle reference nodes, which point to objects that don't expose the inner assetgraphs of bundles. Since they don't expose assetgraphs, they can't be recursive.

Test Plan: Run with graphviz visualization and verify `bundle_reference` nodes in bundle assetgraphs.

### DIFF
--- a/packages/core/core/src/public/BundleGraph.js
+++ b/packages/core/core/src/public/BundleGraph.js
@@ -35,16 +35,32 @@ class BaseBundleGraph {
     this.#graph = graph;
   }
 
-  getBundleGroups(bundle: IBundle): Array<BundleGroup> {
-    let node = this.#graph.getNode(bundle.id);
-    if (!node) {
-      return [];
-    }
+  getBundleGroupsContainingBundle(bundle: IBundle): Array<BundleGroup> {
+    let node = nullthrows(
+      this.#graph.getNode(bundle.id),
+      'Bundle graph must contain bundle'
+    );
 
     return this.#graph.getNodesConnectedTo(node).map(node => {
       invariant(node.type === 'bundle_group');
       return node.value;
     });
+  }
+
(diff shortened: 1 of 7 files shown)


packages/core/core/src/public/BundleGraph.js
```diff
@@ -210,3 +221,66 @@ function findBundlesWithAsset(
...
+    }
+  }
+
+  for (let edge of bundleGraph.edges) {
+    bundleAssetGraph.addEdge(edge);
+  }
+}
+
+class BundleReference {
+  #bundle;
```
wbinnssmith: We need to wrap instead of copy as bundles can be mutated after they're created (bundler, namers, etc.)

packages/core/types/index.js
```diff
@@ -350,8 +348,9 @@ export type BundleGroup = {
 
 export interface BundleGraph {
   findBundlesWithAsset(asset: Asset): Array<Bundle>;
-  getBundleGroups(bundle: Bundle): Array<BundleGroup>;
-  getBundles(bundleGroup: BundleGroup): Array<Bundle>;
+  getBundleGroupsContainingBundle(bundle: Bundle): Array<BundleGroup>;
```
wbinnssmith: There are now two methods for getting bundle groups for a bundle: one retrieving the bundle groups that contain the bundle, and the other for retrieving groups within the bundle (that the bundle depends on).

Too verbose? Thoughts on better names?

PR #2962, opened 2019-04-26
Raw file support

wbinnssmith:
This depends on #2918. A review there would be much appreciated 😄

This adds support for raw files to Parcel 2. Unlike in Parcel 1, this is not implemented as a transform outputting a js asset. Instead, it's implemented in the JS Runtime:

* If an asset reference to a non-js asset is found, and there is no loader for it, fall back to injecting a js asset returning a relative url to the bundle that contains it.
* The runtime will no longer throw if it encounters a bundle without a loader (it assumes that point 1 has taken place).

This also adds the concept of an "isolated" asset via the `isIsolated` property. Isolated assets are always bundled on their own in their own dedicated bundle and bundle group. Relevant transformers should flag this property to `true` for this behavior to take place in the bundler. This happens now in the `RawTransformer` (which otherwise just returns the asset it is given), and will likely happen in the HTML transformer and others where those assets should not be bundled together.

Test Plan: Enabled the existing raw asset integration tests 😃


packages/core/core/src/AssetGraph.js
```diff
@@ -265,7 +265,13 @@ export default class AssetGraph extends Graph<AssetGraphNode> {
 
     let res: ?Asset = null;
     this.traverse((node, ctx, traversal) => {
-      if (node.type === 'asset' || node.type === 'asset_reference') {
+      // Prefer real assets when resolving dependencies, but use the first
```
wbinnssmith: This allows the runtime to inject the js asset which exports the url, and have it properly resolve in the packager. Otherwise, the asset reference to the "real" file would be resolved.

An alternative would be to reorder the edges so that the one pointing to the js asset would come first, but that seems clumsy.

packages/core/core/src/BundlerRunner.js
```diff
@@ -163,8 +162,8 @@ export default class BundlerRunner {
       bundle.assetGraph.merge(subGraph);
 
       bundle.assetGraph.addEdge({
-        from: bundleGroup
-          ? getBundleGroupId(bundleGroup)
+        from: dependency
```
wbinnssmith: Changing the runtimes api (again). This also means that runtime assets will be a peer to the bundle group, not a child, but they'll still resolve fine. It also unifies things with the raw use-case.

packages/runtimes/js/src/JSRuntime.js
```diff
@@ -33,11 +35,38 @@ export default new Runtime({
...
+          // "raw asset"-style fallback
+          // if this dependency doesn't resolve to a js asset, it's an asset reference
+          // of a different type. If there isn't a loader for it, replace it with
+          // a js asset that exports a relative url (using publicURL) to the bundle
+          // it's located in.
+          let assetBundle = bundleGraph.findBundlesWithAsset(resolvedAsset)[0];
+          let hasLoader = loaders && loaders[assetBundle.type];
+          if (!hasLoader) {
+            assets.push({
+              filePath: resolvedAsset.filePath + '.js',
```
wbinnssmith: These "files" need to end in `.js` so they go through the appropriate pipeline, just like loaders. So `test.txt` ends up with a js asset called `test.txt.js` ¯\_(ツ)_/¯

packages/core/integration-tests/test/javascript.js
```diff
@@ -586,30 +584,26 @@ describe('javascript', function() {
...
+        type: 'txt',
+        assets: ['test.txt']
+      }
+    ]);
 
     let output = await run(b);
     assert.equal(typeof output, 'function');
-    assert(/^\/test\.[0-9a-f]+\.txt$/.test(output()));
-    assert(await fs.exists(path.join(__dirname, 'dist/', output())));
+    assert(/^\/dist\/test\.[0-9a-f]+\.txt$/.test(output()));
```
wbinnssmith: we should probably support `outDir` as a parcel option again (and default it to "dist"), and not hardcode these things into the namers. When we write in these relative urls, we don't know any prefix the namer has applied. I imagine in most cases that prefix will not be part of the `publicUrl`.

wbinnssmith:
Just pushed an update with support for binary buffers and streams. Would appreciate reviews @devongovett @DeMoorJasper @padmaia

packages/core/core/src/Asset.js
```diff
@@ -46,63 +59,175 @@ type SerializedOptions = {|
...
+      throw new Error(
+        'Stream has already been read. This may happen if a plugin reads from a stream and does not replace it.'
+      );
+    }
+
+    let size = 0;
+    let hash = crypto.createHash('md5');
+
+    // Since we can only read from the stream once, compute the content length
+    // and hash while it's being written to the cache.
```
wbinnssmith: I was thinking that we could work with functions that *return* readable streams, but this way we enforce that we never transform or read gigantic files unnecessarily. It does make for gymnastics like this though. But they're great for perf :)

packages/core/core/src/Asset.js
```diff
@@ -46,63 +59,175 @@ type SerializedOptions = {|
...
       dependencies: Array.from(this.dependencies),
       connectedFiles: Array.from(this.connectedFiles),
-      output: this.output,
+      isIsolated: this.isIsolated,
       outputHash: this.outputHash,
       env: this.env,
       meta: this.meta,
-      stats: this.stats
+      stats: this.stats,
+      contentKey: this.contentKey
```
wbinnssmith: these keys replace cache references, and are their own distinct properties.

packages/core/core/src/Asset.js
```diff
@@ -46,63 +59,175 @@ type SerializedOptions = {|
...
+  async getCode(): Promise<string> {
+    if (this.contentKey) {
+      this.content = Cache.getStream(this.contentKey);
+    }
+
+    if (typeof this.content === 'string' || this.content instanceof Buffer) {
+      return this.content.toString();
+    }
+
+    this.content = (await bufferStream(this.content)).toString();
```
wbinnssmith: maybe we should just work with binary buffers or streams internally, and always convert to a string whenever `code` is requested? Would simplify some of this.

packages/core/core/src/Asset.js
```diff
@@ -131,22 +256,38 @@ export default class Asset implements IAsset {
...
+    let size;
+    if (content === this.content) {
+      hash = this.hash;
+      size = this.stats.size;
+    } else if (typeof content === 'string' || content instanceof Buffer) {
+      hash = md5FromString(content);
+      size = content.length;
+    } else {
+      hash = null;
+      size = NaN;
```
wbinnssmith: Unfortunately we can't know the length of a stream (nor its hash) until we read from it.

packages/core/core/src/TransformerRunner.js
```diff
@@ -114,7 +132,7 @@ export default class TransformerRunner {
       if (cacheEntry) {
         let cachedAssets = (
           cacheEntry.initialAssets || cacheEntry.assets
-        ).filter(child => child.hash === asset.hash);
+        ).filter(child => child.hash && child.hash === asset.hash);
```
wbinnssmith: `hash` can be falsy (it could before as well), so make sure we're working with a real value.

packages/core/core/src/Asset.js
```diff
@@ -46,63 +59,175 @@ type SerializedOptions = {|
...
+      stats: this.stats,
+      contentKey: this.contentKey
     };
   }
 
+  /*
+   * Prepares the asset for being serialized to the cache by commiting its
+   * content and map of the asset to the cache.
+   */
+  async commit(): Promise<void> {
```
devongovett: I assume we don't want to make this method public. Are you planning on adding a public API wrapper for `Asset`?

packages/core/core/src/Asset.js
```diff
@@ -46,63 +59,175 @@ type SerializedOptions = {|
...
       );
-    this.hash = options.hash || '';
+    this.hash = options.hash;
     this.filePath = options.filePath;
+    this.isIsolated = options.isIsolated == null ? false : options.isIsolated;
     this.type = options.type;
-    this.code = options.code || (options.output ? options.output.code : '');
+    this.content = options.content || '';
+    this.contentKey = options.contentKey;
     this.ast = options.ast || null;
```
devongovett: Should we make ASTs accessed in the same way (via a method)? If we start caching them we'll want to read them from packagers on demand.

packages/core/core/src/Asset.js
devongovett: I wonder about the perf implications of always reading from a stream vs just using `readFile`. Node might optimize `readFile` internally, not sure.

PR #2967, opened 2019-04-30
V2 scope hoisting

devongovett:
This adds scope hoisting/tree shaking to Parcel 2. The code still needs a bit of cleanup, and I may need some help with a few things there.

This introduces a few new concepts to Parcel core.

* **Symbols** - assets and dependencies can now have `symbols` associated with them. This is a map of external names to internal names. For assets, this is the exports, e.g. `export {Foo}`. For dependencies, this is the imports, e.g. `import {Foo} from 'bar'`. This metadata is used to link together imports and exports from different modules, and determine which exports of a module are used.
* **Side Effects** - an asset can now be marked as side effect free by the resolver. Resolvers now return `TransformerRequest`s, which include a `sideEffects` property, derived from the field in package.json. This is propagated to the asset as well.
* **Weak Dependencies** - Dependencies can be marked with the `isWeak` flag to indicate that they are not necessarily required. This occurs specifically for ES module re-exports (e.g. `export * from 'foo'`). When a weak dependency is seen, and the asset has no side effects, we can avoid even parsing the re-exported module unless a symbol from that module is actually used.
* **Deferred Dependencies** - When a weak dependency is seen that has no used symbols, the TransformerRequest is added to a list of deferred nodes in the AssetGraph, and is not processed. If a symbol from that dependency is later needed, the transformation is enqueued at that time.

The actual scope hoisting implementation is fairly similar to Parcel 1, and passes all integration tests. Mostly things were generalized a bit with the symbols concept rather than being specific to JS. I would like to see tree shaking implementations for e.g. CSS modules that use the metadata for used symbols to remove unused classes. Same for e.g. WASM.

The other main difference from Parcel 1 is the optimization using weak dependencies, so large libraries using `sideEffects: false` like lodash-es build much faster since we skip parsing most of it altogether.

Things I need help with:

1. APIs. What should we expose as API? What is the best way to store the symbols in the graph?
2. Code clean up. There are a lot of flow errors.
3. See inline comments


packages/core/core/src/AssetGraph.js
```diff
@@ -163,9 +172,32 @@ export default class AssetGraph extends Graph<AssetGraphNode> {
     let requestNode = nodeFromTransformerRequest(req);
     let {added, removed} = this.replaceNodesConnectedTo(depNode, [requestNode]);
 
+    // Defer transforming this dependency if it is marked as weak, there are no side effects,
+    // and no re-exported symbols are used by ancestor dependencies.
+    // This helps with performance building large libraries like `lodash-es`, which re-exports
+    // a huge number of functions since we can avoid even transforming the files that aren't used.
+    let defer = false;
+    if (dep.isWeak && req.sideEffects === false) {
```
devongovett: This code is confusing and probably a little specific to go in the AssetGraph itself. Would like to come up with a better abstraction for storing symbols in the graph.

packages/core/core/src/AssetGraph.js
```diff
@@ -311,10 +374,44 @@ export default class AssetGraph extends Graph<AssetGraphNode> {
...
 
+    asset.meta.isReferenced = true; // FIXME
     this.replaceNode(assetNode, {
       type: 'asset_reference',
       id: 'asset_reference:' + assetNode.id,
       value: asset
     });
   }
+
+  resolveSymbol(asset: Asset, symbol: Symbol) {
```
devongovett: Same here.

packages/core/core/src/BundleGraph.js
```diff
@@ -51,4 +51,20 @@ export default class BundleGraph extends Graph<BundleGraphNode> {
       }
     });
   }
+
+  isAssetReferenced(asset: Asset) {
```
devongovett: Currently unused. There is a hack later on to set a property when an asset is removed from a bundle since packagers currently don't have access to the BundleGraph. We need a way to know if an asset is used by another bundle...

packages/core/core/src/TransformerRunner.js
```diff
@@ -75,11 +77,11 @@ export default class TransformerRunner {
 
     // If the transformer request passed code rather than a filename,
     // use a hash as the id to ensure it is unique.
-    if (req.code) {
```
devongovett: This was causing bugs since all assets in the output had the same id...

packages/core/core/src/public/Bundle.js
```diff
@@ -73,14 +75,28 @@ export class Bundle implements IBundle {
...
+
+  traverseAssets<TContext>(visit: GraphVisitor<Asset, TContext>): ?TContext {
+    return this.#bundle.assetGraph.traverseAssets(visit);
+  }
+
+  traverseAncestors(startNode: TNode, visit: GraphVisitor<TNode, TContext>) {
+    return this.#bundle.assetGraph.traverseAncestors(startNode, visit);
+  }
+
+  hasChildBundles() {
```
devongovett: Probably should be in the BundleGraph instead, but again the packager doesn't have access to the BundleGraph...

packages/core/core/src/public/Bundle.js
```diff
@@ -73,14 +75,28 @@ export class Bundle implements IBundle {
     return this.#bundle.assetGraph.getTotalSize(asset);
   }
 
-  traverseAssets<TContext>(
-    visit: GraphTraversalCallback<Asset, TContext>
-  ): ?TContext {
-    return this.#bundle.assetGraph.traverse((node, ...args) => {
-      if (node.type === 'asset') {
-        return visit(node.value, ...args);
+  resolveSymbol(asset: Asset, symbol: Symbol) {
```
devongovett: Doesn't seem to make sense in a Bundle... Need a better place to put this.

packages/core/core/src/public/Bundle.js
```diff
@@ -73,14 +75,28 @@ export class Bundle implements IBundle {
...
+    return this.#bundle.assetGraph.resolveSymbol(asset, symbol);
+  }
+
+  traverseAssets<TContext>(visit: GraphVisitor<Asset, TContext>): ?TContext {
+    return this.#bundle.assetGraph.traverseAssets(visit);
+  }
+
+  traverseAncestors(startNode: TNode, visit: GraphVisitor<TNode, TContext>) {
```
devongovett: Don't really like exposing this

packages/packagers/js/src/JSPackager.js
```diff
@@ -2,56 +2,61 @@
...
-      ', ' +
-      'null' +
-      ')'
-    );
+  async package(bundle, options) {
+    let ast = await concat(bundle, options);
+    ast = link(bundle, ast, options);
+    return generate(bundle, ast, options);
+
+    // let promises = [];
```
devongovett: Need to figure out when/how to enable scope hoisting. Probably production only for performance

packages/shared/scope-hoisting/src/hoist.js
```diff
@@ -0,0 +1,620 @@
...
+
+    let {left, right} = path.node;
+    if (t.isIdentifier(left) && left.name === 'exports') {
+      path.get('left').replaceWith(getExportsIdentifier(asset, path.scope));
+      asset.meta.isCommonJS = true;
+    }
+
+    // If this assignment is the first reference to module.exports, replace it with a variable declaration.
+    // This way we avoid the extra assignment to an empty exports object.
+    // if (t.matchesPattern(left, 'module.exports')) {
```
devongovett: todo

packages/shared/scope-hoisting/src/hoist.js
```diff
@@ -0,0 +1,620 @@
...
+    let ignore =
+      args.length !== 1 ||
+      !t.isStringLiteral(args[0]) ||
+      path.scope.hasBinding('require');
+
+    if (ignore) {
+      if (isRequire) {
+        callee.name = 'parcelRequire';
+      }
+      // path.get('callee').replaceWith(t.identifier('parcelRequire'));
```
devongovett: todo

packages/transformers/js/src/JSTransformer.js
```diff
@@ -104,20 +105,22 @@ export default new Transformer({
       }
     }
 
+    hoist(asset);
```
devongovett: todo: only enable in prod

packages/shared/scope-hoisting/package.json
```diff
@@ -0,0 +1,33 @@
...
+    "node": ">= 8.0.0"
+  },
+  "scripts": {
+    "pretest": "yarn build",
+    "test": "cross-env NODE_ENV=test mocha",
+    "test-ci": "yarn build && yarn test",
+    "build": "babel src -d lib",
+    "prepublish": "yarn build"
+  },
+  "devDependencies": {
```
wbinnssmith: we shouldn't need these `devDependencies` since they're listed in the root package.json

packages/shared/scope-hoisting/package.json
```diff
@@ -0,0 +1,33 @@
...
+    "@parcel/eslint-config": "1.10.3",
+    "mocha": "^5.2.0"
+  },
+  "dependencies": {
+    "@babel/generator": "^7.3.3",
+    "@babel/template": "^7.2.2",
+    "@babel/traverse": "^7.2.3",
+    "@babel/types": "^7.3.3",
+    "@parcel/types": "^2.0.0",
```
wbinnssmith: Is this a real runtime dependency? Otherwise could probably just remove it as it's also not necessary as a devdependency.

devongovett:
Cleaned up the code a bit to get it ready for review. The remaining questions are mostly around the public API. Would like to discuss and hear ideas about what we should expose.

devongovett:
A read-only BundleGraph is now sent to packagers along with the current bundle to package. This is so that the JSPackager can determine whether an asset is referenced by another bundle. The code could perhaps be optimized later so that we aren't serializing and sending the entire bundle graph to the workers for every bundle to package...

packages/core/core/src/AssetGraph.js
```diff
@@ -280,15 +321,29 @@ export default class AssetGraph extends Graph<AssetGraphNode> {
...
-  traverseAssets(
-    visit: GraphTraversalCallback<Asset, AssetGraphNode>,
-    startNode: ?AssetGraphNode
-  ): ?AssetGraphNode {
-    return this.traverse((node, ...args) => {
-      if (node.type === 'asset') {
-        return visit(node.value, ...args);
+  getAncestorDependencies(asset: Asset): Array<IDependency> {
+    let node = this.getNode(asset.id);
+    if (!node) {
```
wbinnssmith: should we be throwing in situations like these?

packages/core/core/src/AssetGraph.js
```diff
@@ -326,4 +381,37 @@ export default class AssetGraph extends Graph<AssetGraphNode> {
 
     return referenceId;
   }
+
+  resolveSymbol(asset: Asset, symbol: Symbol): SymbolResolution {
```
wbinnssmith: this also is very specific to symbols and I'm not sure it should live directly on the `AssetGraph`. Maybe make this a function that takes an AssetGraph? Or maybe directly on the public Bundle object?

packages/core/core/src/public/Bundle.js
wbinnssmith: now it does :)

packages/core/core/src/Graph.js
```diff
@@ -326,6 +365,54 @@ export default class Graph<TNode: Node> {
...
+      if (fn(node)) {
+        res.push(node);
+        traversal.skipChildren();
+      }
+    }, node);
+    return res;
+  }
+
+  findNode(predicate: TNode => boolean): ?TNode {
+    return Array.from(this.nodes.values()).find(predicate);
```
wbinnssmith: we should probably be consistent with `[...iterable]` vs `Array.from(iterable)` 😃

packages/core/core/src/Dependency.js
```diff
@@ -25,27 +26,54 @@ export default class Dependency implements IDependency {
...
   constructor(opts: DependencyOpts) {
     this.moduleSpecifier = opts.moduleSpecifier;
     this.isAsync = opts.isAsync;
     this.isEntry = opts.isEntry;
     this.isOptional = opts.isOptional;
     this.isURL = opts.isURL;
+    this.isWeak = opts.isWeak;
     this.loc = opts.loc;
-    this.meta = opts.meta;
+    this.meta = opts.meta || {};
```
wbinnssmith: ❤️

packages/core/core/src/AssetGraph.js
```diff
@@ -163,9 +170,37 @@ export default class AssetGraph extends Graph<AssetGraphNode> {
...
+    let defer = false;
+    if (dep.isWeak && req.sideEffects === false) {
+      let assets = this.getNodesConnectedTo(depNode);
+      let symbols = invertMap(dep.symbols);
+      invariant(
+        assets[0].type === 'asset' || assets[0].type === 'asset_reference'
+      );
+      let resolvedAsset = assets[0].value;
+      let deps = this.getAncestorDependencies(resolvedAsset);
+      defer = deps.every(
```
wbinnssmith: maybe a comment explaining what all this does? or is that taken care of above?

packages/core/core/src/Asset.js
```diff
@@ -47,7 +49,9 @@ type AssetOptions = {|
   outputHash?: string,
   env: Environment,
   meta?: Meta,
-  stats: Stats
+  stats: Stats,
+  symbols?: Map<Symbol, Symbol> | Array<[Symbol, Symbol]>,
```
wbinnssmith: `symbols?: Iterable<[Symbol, Symbol]>` would cover both of these I think. I don't mind being explicit either.

packages/core/core/src/AssetGraph.js
devongovett: maybe, but in this case if you asked for ancestor dependencies of an entry I think it's better to return an empty array than throw.

packages/core/core/src/AssetGraph.js
devongovett: One option I explored was to have symbol nodes in the asset graph, but it got complicated very quickly. Maybe a separate symbol graph? Either way, I'm not happy with the current approach of storing symbols in maps on assets/dependencies so any suggestions to make it more general/better would be helpful.

packages/core/core/src/Asset.js
devongovett: Nice

PR #2969, opened 2019-05-01
Parcel config invalidation spike mk2

padmaia:
**Redo of #2886 based on PR feedback**

# Notable differences from last PR

* I've added the concept of reverse handles to the WorkerFarm, which allow workers to call functions in the main process. (I also incidentally found some issues with local workers not behaving as expected, which should probably be fixed with a LocalWorker class that has the same methods as Worker or something, but for now I just fixed them hackily). The first use case for a reverse handle is for loading config. `loadConfig` is a method defined on the `AssetGraphBuilder` so we can use the graph to avoid unnecessary loading of config that has already been loaded. By making it a reverse handle we can still handle transformations that jump pipelines in the worker processes. In the last PR this did not work well because we only knew what the entry asset and initial pipeline configs were. Now we can load more config in the middle of a transformation when we jump pipelines.

* Config requests will now be based on the plugin making the request instead of a config type, so there is no longer a need for a new config loader plugin type.

* Added a `Config` class similar to the `Asset` class to make it easier for plugins to tell Parcel when a config should be invalidated and reloaded.

# Still to do
- [x] Load parcel config at the root to get correct bundler/packagers/namers/reporters
- [x] Turn back on bundling/packaging and make sure everything still works
- [ ] Passing tests & Flow coverage
- [ ] Clean up unused code

# Things that still need to be explored, but probably not as part of this PR

* Still need to figure out how to allow tools/frameworks built on Parcel to provide default config or override config. One possible solution is that tools/frameworks simply author their own plugins which may be simple extensions of Parcel's default plugins. The concern with this is that parcel plugins are required to be fully fledged packages, so these tools/frameworks would have to maintain many packages. If they are not already using a multi package repo, this would be annoying. Another option would be to add hooks before and/or after `plugin.loadConfig` to provide a default if nothing is found or manipulate what is returned.

* Plugins may end up requesting more than one config. For example the babel transformer uses babel config and also information from the package.json. I'm not sure how this should be represented yet. Would they be separate requests? Ideally, but I'm not sure how to tell `loadConfig` that it only needs to do partial work. Would these configs all live in the `Config` class or would there be a collection of multiple `Config` instances? Not sure.

* There are times where we won't be able to reliably tell if something should be invalidated or not. For instance, babel plugins like `babel-preset-env` may end up resolving additional configuration like `.browserslistrc` and don't have the information to tell us that it used that file and should be invalidated when it changes. We need a way to tell Parcel to at least invalidate on initial start up. Unfortunately this may need to be the case whenever there is custom babel config :/ I think our use of `babel-preset-env` should still be okay since we are resolving browserslist ourselves at start up everytime so we should be able to invalidate only when targets change.

* When dev deps change, we may not be able to rebuild without restarting Parcel because of node's require cache. We should either alert the user that they should restart or do the restarting ourselves. We could also try clearing out the require cache, but that has the potential to get hairy.

# Try it out

```
cd packages/examples/real-simple
yarn demo start
```

In another terminal:

```
cd packages/examples
cp simple/.parcelrc real-simple/.parcelrc

# parcel should rebuild the config, but since the content ends up being the same...
# ...it should not run any transformations again

rm simple/.parcelrc
```

It should also handle rerunning the transformation if the config changes. It should only run the transformers if the content of the config actually changed.


padmaia:
*Update:*

Getting the tests to pass has been a bit of a slog. At the moment, something about bundling is misbehaving. I've been comparing my branch graphviz against v2's graphviz and it's been kind of frustrating with all of the extra nodes, especially once the graph starts getting more complicated than the real simple example. I'm also not totally sure how I feel about combining the AssetGraph and AssetGraphBuilder together. I've been brainstorming about the idea of having two separate graphs, one being the AssetGraph which would have only have assets and dependencies and the other graph would be responsible to maintaining state of in progress, invalid, and finished requests (RequestGraph?). That way we could just build the skinny AssetGraph incrementally instead of pruning and/or skipping unimportant nodes from the entire graph every time we bundle and it would be much easier to follow what's going on during bundling. This approach also has the potential to work nicely with scope hoisting since that didn't work perfectly with the Asset Graph. Since we will be adding more specialized Graphs, we could also build a SymbolGraph (context: https://github.com/parcel-bundler/parcel/pull/2967#discussion_r279622254). I'm going to take a step back and split out the pieces of my PR that are easier to land and start landing things more incrementally. I'll start by extracting tasks that are easier to land like moving the watcher out of AssetGraphBuilder to Parcel. Then I will explore the separate RequestGraph approach, and once we're happy with that it should be fairly straight forward to add config requests and dev dep requests.

padmaia:
Going to close this, since I will be opening incremental PRs based on the work I've done here. Feel free to still leave feedback, so that I can address it in future PRs.

Issue #1345, opened 2018-05-11
Doing less work in dev


jamiebuilds:
# 💬 RFC

I'm wondering if there's more work that we can carve out of Parcel while running in dev mode by assuming they are running in the most modern browsers.

- we can stop compiling node_modules using Babel
- we can skip any transforms that only exist for older browsers (Babel, autoprefixer, etc)

## 🔦 Context

Most of the time we're compiling things unnecessarily and it slows dev loops down

DeMoorJasper:
I think this would be safe to do as long as we add a cli option to have the ability to re-enable it to debug bugs in IE

jamiebuilds:
We do when you provide it, but dev mode could still do less work

mohsen1:
Assuming browser is not safe. I am still dealing with iOS 7 Safari in my day job :( 

Dev server can sniff browser UA header to decide how to compile. It doesn't have to be too fancy, just knowing browser name/version is enough. I'm sure there are nice packages out there doing this.

budarin:
It should be under a flag.
What if I want to develop under one of those older browser or want to have a look?

mischnic:
I think this is already done in Parcel 2?

Jogai:
I think this is a bad idea. I just had a bug because of this (nightly 221). Jquery worked fine during dev, but failed on build. You don't want your output to be that different. Build should do basically the same as dev, but in an optimized fashion.

PR #2972, opened 2019-05-03
Naming and targets refactor

wbinnssmith:
This heavily refactors and reworks naming in Parcel 2:

* Parcel options are now pre-resolved during initialization and passed through parcel from there. A distinction between the option shape passed to the `Parcel` constructor and the shape of the resolved options is made in types.
* Namers are expected to only name the basenames of bundles within their dist directories. This basename can include path separators, and they will be appended to both dist paths and public urls for the bundles
* `distPath` is now split into `distDir` and `distEntry` on targets. This is basically the same splitting that occurred in the default namer but formalizes it, giving plugins consistent understanding of where files are written.
* `publicUrl` is now part of targets as well, and can be specified in the target entry in `package.json`, defaulting to '/'. It no longer exists as a top-level Parcel option
* `targets`, when passed to the Parcel constructor, can now be a list of strings instead of target objects (or a mix). If a string, we look up the resolved target name and include that. 
	* e.g. if you invoke `parcel build --target browser`, we find the browser target and only build that. Multiple targets are allowed in the form `parcel build --target browser --target browserModern`
* When in serving mode (server or hmr), only one target is supported, and it must be a browser target. 
* **In all development modes**, if a target is not explicitly provided via `--target` or similar, one targeting only the latest version of modern browsers is used. This is intended to reduce the amount of work we do in development (#1345)

Test Plan: 
* Existing integration tests were changed to reflect changes in `publicUrl` and continue to pass. 
* Use the cli to pass multiple targets to the server and verify it fails.
* Add a target for node and use the cli to pass this target by name to the server. Verify it fails
* Manual test of the simple example. Alter the publicUrl of a target and verify the url entered via raw runtime includes it.


wbinnssmith:
cc @jamiekyle @jamiekyle-eb this is probably of interest to you

PR #2975, opened 2019-05-04
Output shebang for node entry files

### LINKED ISSUE #2381: When entry file has a shebang, output it in the first line of the bundle
# 🙋 feature request

In UNIX systems, the first line of a script can be a [shebang](https://en.wikipedia.org/wiki/Shebang_(Unix)) with the path to the interpreter. When the main entry file has this shebang (and optionally when `target: node`), this shebang line could be taken _out of the bundled modules_ and outputted in the first line of the bundle.

## 🤔 Expected Behavior
`hello-world.js`:
```js
#!/usr/bin/env node

console.log("Hello, world!");
```

```sh
$ parcel build -t node hello-world.js
```

`dist/hello-world.js`:
```js
#!/usr/bin/env node

// parcel bundle prologue ...function(require,module,exports) {
console.log("Hello, world!");
// }... parcel bundle epilogue
```

## 😯 Current Behavior
`dist/hello-world.js`:
```js
// parcel bundle prologue ...function(require,module,exports) {
#!/usr/bin/env node
console.log("Hello, world!");
// }... parcel bundle epilogue
```

(which is actually a `SyntaxError`)

## 🔦 Context

I run into this when trying to make executable scripts for node which have all their files already bundled.

cprecioso:
# ↪️ Pull Request

If target is node, check if the main asset is an entry asset and if it
has a shebang. In that case, strip it from the asset and put it at the
top of the output.

Fixes #2381 

## 💻 Examples

See the expected behaviour in #2381 

## 🚨 Test instructions

`index.js`:
```js
#!/usr/bin/env node
console.log("Hola")
```
When this PR is applied, it will not give a syntax error, and will run as expected when used as an executable script (`chmod +x dist/index.js` and `./dist/index.js`).

## ✔️ PR Todo

### DIFF
--- a/packages/core/parcel-bundler/src/packagers/JSPackager.js
+++ b/packages/core/parcel-bundler/src/packagers/JSPackager.js
@@ -26,6 +26,21 @@ class JSPackager extends Packager {
           this.options.hmrHostname
         )};` + preludeCode;
     }
+
+    // If target is node, check if the main asset is an entry asset and if it has a shebang.
+    // In that case, strip it from the asset and put it at the top of the output
+    if (this.bundler.options.target === 'node') {
+      const entryAsset = this.bundle.entryAsset;
+      if (
+        this.bundler.entryAssets.has(entryAsset) &&
+        entryAsset.generated.js.trimStart().startsWith('#!')
+      ) {
+        const firstNewline = entryAsset.generated.js.indexOf('\n') + 1;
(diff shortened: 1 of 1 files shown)


devongovett:
We're feature frozen for Parcel 1 at this point. #3110 should cover this for Parcel 2.

PR #2979, opened 2019-05-06
Shake exports with pure property assignments

### LINKED ISSUE #2909: Why the production bundle size is significantly larger that create-react-app
I did a very simple experiment with the following settings

index.html (used for `parcel-bundler` only as `create-react-app` creates one automatically)
```html
<html>
<body>
  <div id="root"></div>
</body>
</html>
```

index.js (also used in the project initialized with `create-react-app`) 
```jsx
import React from 'react'
import ReactDOM from 'react-dom'
import { FaBeer } from 'react-icons/fa'

const App = () => (
  <h1>
    Let's have some <FaBeer />!
  </h1>
)

ReactDOM.render(<App />, document.getElementById('root'))
```

`./node_modules/.bin/parcel --version`
1.12.3
`./node_modules/.bin/parcel build index.html`
parcel-build

While in a separate project initialized with create-react-app (2.1.8), the bundle size is way much smaller
cra-build-after-gzip

Without gzip:
cra-build

Nodejs version: 8.14.1
OS: Mac OSX Mojave

Can someone tell me why the bundle size difference is so significant? Are there any options I can use to improve/reduce the bundle size using parcel?

devongovett:
Fixes #2909.

For example, assigning a `displayName` to a function is pretty common in react code, and does not cause side effects.

### DIFF
--- a/packages/core/parcel-bundler/src/scope-hoisting/shake.js
+++ b/packages/core/parcel-bundler/src/scope-hoisting/shake.js
@@ -1,7 +1,5 @@
 const t = require('@babel/types');
 
-const EXPORTS_RE = /^\$([^$]+)\$exports$/;
-
 /**
  * This is a small small implementation of dead code removal specialized to handle
  * removing unused exports. All other dead code removal happens in workers on each
@@ -44,12 +42,9 @@ function getUnusedBinding(path, name) {
     return null;
   }
 
-  if (isPure(binding)) {
+  let pure = isPure(binding);
+  if (!binding.referenced && pure) {
     return binding;
-  }
-
-  if (!EXPORTS_RE.test(name)) {
-    return null;
   }
 
   // Is there any references which aren't simple assignments?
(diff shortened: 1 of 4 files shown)


packages/core/parcel-bundler/src/scope-hoisting/shake.js
```diff
@@ -85,7 +76,8 @@ function isExportAssignment(path) {
     // match "path.any = any;"
     path.parentPath.isMemberExpression() &&
     path.parentPath.parentPath.isAssignmentExpression() &&
-    path.parentPath.parentPath.node.left === path.parentPath.node
+    path.parentPath.parentPath.node.left === path.parentPath.node &&
+    path.parentPath.parentPath.get('right').isPure()
```
mischnic: Should rather be called `isExportAssignmentPure` now?

packages/core/parcel-bundler/src/scope-hoisting/shake.js
```diff
@@ -122,6 +114,15 @@ function remove(path) {
   } else if (isUnusedWildcard(path)) {
     remove(path.parentPath);
   } else if (!path.removed) {
-    path.remove();
+    if (
```
mischnic: The code duplication isn't ideal but I don't see a cleaner way without doing some extra if checks

packages/core/parcel-bundler/src/scope-hoisting/shake.js
devongovett: actually... I think this check is unnecessary now

PR #2982, opened 2019-05-06
Attempt at fixing the tests in master

DeMoorJasper:
# ↪️ Pull Request

Closes #2842 closes #2935

## 🚨 Test instructions

`yarn test` shouldn't throw errors


mischnic:
Related: #2842

packages/core/integration-tests/test/fs.js
```diff
@@ -52,7 +52,7 @@ describe('fs', function() {
       assert.equal(output, 'hello');
     });
 
-    it('should inline a file with fs require destructure', async function() {
+    it.skip('should inline a file with fs require destructure', async function() {
```
DeMoorJasper: This is related to babel issues afaik, skipped it for now as babel was locked to <7.4 due to bugs.

mischnic: > locked to <7.4 due to bugs.

How long will it stay it that way?

DeMoorJasper: Not sure Devon locked it, perhaps the bugs have already been fixed

cc @devongovett

mischnic: One issue atm with the old babel: `This experimental syntax requires enabling the parser plugin: 'objectRestSpread'` if the deps contain rest spread (which isn't a proposal anymore)

DeMoorJasper:
Everything passes now 🎉

Ready for another review

mischnic:
The tests pass on my machine, but the output is strange:
```

  FSCache
    ✓ should create directory on ensureDirExists (76ms)
(node:9100) UnhandledPromiseRejectionWarning: #<Object>
(node:9100) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
(node:9100) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
(node:9100) UnhandledPromiseRejectionWarning: #<Object>
(node:9100) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 2)
    ✓ should cache resources (81ms)
    ✓ should return null for invalidated resources
    ✓ should remove file on delete (72ms)
    ✓ should remove from invalidated on write (63ms)
```

DeMoorJasper:
@mischnic that's very strange but probably not caused by this PR

packages/core/integration-tests/test/fs.js
devongovett: I believe babel is fixed now, so we can upgrade it again.

DeMoorJasper:
Opened a new PR based on #2842 with the more up to date babel version

PR #2986, opened 2019-05-07
Clear scope cache before crawling to fix scope hoisting classes

### LINKED ISSUE #2981: --experimental-scope-hoisting not working with rollup build
# 🐛 bug report

in the following [example](https://github.com/IdoJS/experimental-scope-hoisting) you can see a problem with --experimental-scope-hoisting when using rollup build cjs

## 🤔 Expected Behavior
``` gherkin
Given a project with lerna and multiple packages
```

|package| build | script|
|--------|-------|------|
|hoistexample-rollup| rollup | rollup src/index.js --file lib/index.js --format cjs
|hoistexample-ts | ts |  tsc --module commonjs --target es5 --outDir lib

```gherkin
When a main package imports the packages 
And bundle it all with 
 # parcel build -o build index.js  --experimental-scope-hoisting --no-minify 
Then  parcel change the variables name of the rollup build correctly as it does for ts build
```

## 😯 Current Behavior

``` gherkin
Given a project with lerna and multiple packages
```

|package| build | script|
|--------|-------|------|
|hoistexample-rollup| rollup | rollup src/index.js --file lib/index.js --format cjs
|hoistexample-ts | ts |  tsc --module commonjs --target es5 --outDir lib

```gherkin
When a main package imports the packages 
And bundle it all with 
 # parcel build -o build index.js  --experimental-scope-hoisting --no-minify 
Then parcel change the variables name of the rollup build incorrectly 
And runtime throw exception of ReferenceError
```

## 💁 Possible Solution

you can see from the [example](https://github.com/IdoJS/experimental-scope-hoisting)
that the problem occurs when doing the export in one line.
 ```
export class failurtestHositingRollup2 {
  static create() {
    return new failurtestHositingRollup2();
  }
};

export default class failurtestHositingRollup {
  static create() {
    return new failurtestHositingRollup();
  }
}

```
## 💻 Code Sample

**parcel bundle with ts import build:**
```js
var $sbuw$var$testHositing =
/** @class */
function () {
  function testHositing() {}

  testHositing.create = function () {
    return new testHositing();
  };

  return testHositing;
}();
```

**parcel bundle with rollup import build:**
```js

class $gcK$var$failurtestHositingRollup {
  static create() {
    return new failurtestHositingRollup();
  }

}
```
## 🌍 Your Environment

| parcel-bundler           |1.6.1
| Node             | v10.15.0
| Yarn         | 1.12.3
| Operating System | MAC

devongovett:
Fixes #2981.

This works around a babel issue where re-crawling the scope doesn't update all references.


packages/core/parcel-bundler/src/scope-hoisting/hoist.js
```diff
@@ -51,6 +52,7 @@ function hasSideEffects(asset, {sideEffects} = asset._package) {
 module.exports = {
   Program: {
     enter(path, asset) {
+      traverse.cache.clearScope();
```
mischnic: Does this have (performance) side effects for every import of `@babel/traverse` ?

devongovett: Possibly, but it’s the only way. I believe Babel’s behavior may actually be correct. I don’t think `scope.crawl` was meant to be called manually. It is called automatically by Babel when setting up a scope for a path, and only re-crawls that scope, not child scopes. So we need to clear the cache so that Babel will automatically crawl child scopes again when we traverse them.

Ideally we wouldn’t need to re-crawl at all, but some Babel plugins are badly behaved and don’t update the scope properly when they mutate the AST.

PR #2992, opened 2019-05-07
Remove circular package dependencies

wbinnssmith:
V2 currently has circular package dependencies, mainly because `@parcel/utils` (which has a wide number of uses) uses `@parcel/workers` for package installation. 

* Remove these circular dependencies by extracting both localRequire and installPackage into their own packages.
* Add eslint configuration so that `@parcel/utils` does not import `@parcel/workers` again
* Remove unused dependencies in `@parcel/logger` (from before it was a simple emitter), which also introduced circular package dependencies

Test Plan: 

* `yarn test` and verify no messages from yarn about circular dependencies.
* `yarn lint` before changes and verify failure from `@parcel/utils` using workers. Lint afterward and verify clean.


wbinnssmith:
> Could local-require and install-package be combined into one package or do they have to be separate?

They could. What would you call it? `@parcel/utils-that-require-workers`? 🙃I figured keeping the responsibilities separate with packages creates a clearer boundary, but I definitely see the downsides to more packages than necessary.

devongovett:
> What would you call it?

Lol, no idea :D

mischnic:
merge both into `local-require` ? `install-package` is also "local" as in `localRequire`

padmaia:
I don't see an issue with keeping them as separate packages. I understand we don't want to maintain a million packages, but so far these are the only two examples of smaller packages. I don't really see it spiraling out of control. Just my two cents.

PR #3001, opened 2019-05-08
Create public Asset and MutableAsset

wbinnssmith:
* Creates public `Asset` and `MutableAsset` interfaces and implementations, which prevent access to internal-only methods like `commit`, direct access to the `content` blob, the `Map`s used for dependency and connected files, etc.
	* Currently `MutableAsset` is only needed in `Transformer#transform`!
* Makes `CacheEntry` an internal-only type
* Removes unused `TransformFinishedEvent` (cc @DeMoorJasper ?) which exposed a `CacheEntry`
* Fixes types in `TransformerRunner` and normalizes all returned types to `TransformerResults`

Test Plan: flow, lint, test. Manual test with the simple example.

### DIFF
--- a/packages/core/core/src/BundleGraph.js
+++ b/packages/core/core/src/BundleGraph.js
@@ -1,6 +1,7 @@
 // @flow strict-local
 
-import type {Asset, GraphTraversalCallback} from '@parcel/types';
+import type {GraphTraversalCallback} from '@parcel/types';
+import type Asset from './Asset';
 import type {Bundle, BundleGraphNode} from './types';
 
 import Graph from './Graph';
--- a/packages/core/core/src/public/Asset.js
+++ b/packages/core/core/src/public/Asset.js
@@ -1,0 +1,145 @@
+// @flow strict-local
+// flowlint unsafe-getters-setters:off
+
+import type {Readable} from 'stream';
+
+import type {
+  Asset as IAsset,
+  AST,
+  Config,
+  Dependency,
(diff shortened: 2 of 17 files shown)


packages/core/core/src/BundleGraph.js
```diff
@@ -15,7 +15,7 @@ export default class BundleGraph extends Graph<BundleGraphNode> {
     });
   }
 
-  isAssetInAncestorBundle(bundle: Bundle, asset: Asset): boolean {
+  isAssetInAncestorBundle(bundle: Bundle, assetId: string): boolean {
```
devongovett: Why are we switching to ids for this?

packages/core/core/src/public/Asset.js
```diff
@@ -0,0 +1,167 @@
...
+
+  get filePath(): FilePath {
+    return this.#asset.filePath;
+  }
+
+  get meta(): Meta {
+    return this.#asset.meta;
+  }
+
+  get outputHash(): string {
```
devongovett: Should this be internal?

wbinnssmith: It's used here:

https://github.com/parcel-bundler/parcel/blob/b96d50c4311ed7bec2859150233991648068044d/packages/namers/default/src/DefaultNamer.js#L83

Maybe we separate Asset into `TransformerAsset` and `CompletedAsset` (or similar), where only the latter has `outputHash`?

PR #3009, opened 2019-05-09
HTML Transformer & Packager

wbinnssmith:
This implements a first iteration of an HTML transformer & packager in Parcel 2.

* Adds `addURLDependency` to the public `MutableAsset` as an api for adding a dependency on a url-like string and injecting a dependency id into the transformed ast/code.
* Builds a mapping from dependency ids to relative bundle paths and injects this mapping into assets before they are packaged. When assets are asked for code (currently only the string version), if this mapping is present, replace references to the dependency ids to the relative path to the relevant bundle
* Integrates @devongovett's work in the `html-transformer` branch
* Adds minimal types to `posthtml`, `posthtml-parser`, and `posthtml-render`

Test Plan: `yarn test`. A basic html example was also added to the examples directory.


packages/core/core/src/public/BundleGraph.js
```diff
@@ -133,11 +134,24 @@ export class MutableBundleGraph extends BaseBundleGraph
   }
 
   addBundleGroup(parentBundle: ?IBundle, bundleGroup: BundleGroup) {
-    let node = {
-      id: getBundleGroupId(bundleGroup),
-      type: 'bundle_group',
-      value: bundleGroup
-    };
+    let bundleGroupId = getBundleGroupId(bundleGroup);
```
wbinnssmith: This implements bundle group "sharing": if a node for this group already exists, we reference it instead. However, the `dependency` on the group will continue to reference the first bundle that depends on it. We should probably decouple dependencies from bundle groups, and store those connections in the bundle graph instead.

packages/examples/html/src/index.html
```diff
@@ -0,0 +1,26 @@
+<!DOCTYPE html>
```
wbinnssmith: @DeMoorJasper should we just combine this with the hmr example now that the hmr example serves an html file?

packages/core/core/src/PackagerRunner.js
```diff
@@ -94,3 +103,41 @@ export default class PackagerRunner {
...
+
+    let [entryBundleNode] = bundle.assetGraph.getNodesConnectedFrom(
+      bundleGroupNode
+    );
+    invariant(entryBundleNode && entryBundleNode.type === 'bundle_reference');
+
+    let entryBundle = entryBundleNode.value;
+    depToBundlePath.set(
+      dep.id,
+      urlRelative(nullthrows(bundle.name), nullthrows(entryBundle.name))
```
wbinnssmith: Part of me wants to avoid relative paths all together. I know they're portable, and we should know exactly where they are relative to this bundle, but something feels safer about using a path from the root all the time. For example, do all asset types support relative paths?

packages/core/core/src/AssetGraphBuilder.js
```diff
@@ -159,7 +160,7 @@ export default class AssetGraphBuilder extends EventEmitter {
 
     for (let asset of cacheEntry.assets) {
       asset.stats.time = time;
-      this.changedAssets.set(asset.id, asset);
+      this.changedAssets.set(asset.id, new Asset(asset));
```
wbinnssmith: This is breaking HMR all together. I guess we missed this @DeMoorJasper

packages/core/core/src/PackagerRunner.js
```diff
@@ -67,6 +68,14 @@ export default class PackagerRunner {
       bundle
     });
 
+    let depToBundlePath = generateDepToBundlePath(internalBundle);
+
+    // Inject the dependency-to-relative-path mapping into all the bundle's
+    // assets so that substitution can occur when the asset's output is generated.
+    internalBundle.assetGraph.traverseAssets(asset => {
+      asset.depToBundlePath = depToBundlePath;
```
devongovett: Wonder if we should just do the replacement after the packager runs instead of adding this to individual assets?

packages/core/core/src/PackagerRunner.js
```diff
@@ -94,3 +103,43 @@ export default class PackagerRunner {
...
+    if (node.type !== 'dependency') {
+      return;
+    }
+
+    let dep = node.value;
+    if (!dep.isURL || !dep.isAsync) {
+      return;
+    }
+
+    let [bundleGroupNode] = bundle.assetGraph.getNodesConnectedFrom(node);
```
devongovett: Should BundleGraph have a `getEntryBundles` helper method? Seems relatively useful...

packages/core/core/src/TransformerRunner.js
```diff
@@ -209,7 +209,7 @@ export default class TransformerRunner {
         !transformer.canReuseAST(input.ast, this.options)) &&
       previousGenerate
     ) {
-      let output = await previousGenerate(input);
+      let output = await previousGenerate(new MutableAsset(input));
```
devongovett: How were the types not catching these before?

packages/core/core/src/public/BundleGraph.js
devongovett: What do we use the `dependency` for?

packages/core/utils/src/urlRelative.js
```diff
@@ -0,0 +1,7 @@
+// @flow strict-local
+
+import path from 'path';
+
+export default function urlRelative(from: string, to: string): string {
```
devongovett: Parcel 1 had `urlJoin` for this, which handled a few more edge cases (e.g. query strings and hashes).

packages/namers/default/src/DefaultNamer.js
```diff
@@ -9,6 +9,8 @@ import path from 'path';
 
 const COMMON_NAMES = new Set(['index', 'src', 'lib']);
 
+const namedTargets: Set<Target> = new Set();
```
devongovett: Hmm this seems... interesting. Why?

packages/transformers/html/src/HTMLTransformer.js
```diff
@@ -0,0 +1,38 @@
...
+
+export default new Transformer({
+  canReuseAST(ast) {
+    return ast.type === 'posthtml' && ast.version === '0.0.0';
+  },
+
+  async parse(asset) {
+    return {
+      type: 'posthtml',
+      version: '0.0.0',
```
devongovett: probably should get a real version from posthtml-parser's package.json

packages/transformers/html/src/HTMLTransformer.js
```diff
@@ -0,0 +1,38 @@
...
+
+import parse from 'posthtml-parser';
+import render from 'posthtml-render';
+import nullthrows from 'nullthrows';
+import collectDependencies from './dependencies';
+import extractInlineAssets from './inline';
+
+export default new Transformer({
+  canReuseAST(ast) {
+    return ast.type === 'posthtml' && ast.version === '0.0.0';
```
devongovett: use semver.satisfies. Maybe we should make a helper function for this too... e.g. `asset.astSatisfies('posthtml', '^1.2.3')`. (naming tbd)

packages/namers/default/src/DefaultNamer.js
wbinnssmith: A bit of a hack to avoid naming more than one bundle as the dist entry. The case here was an html entry file linking to another html file, which also got tagged as an entry in the same bundle group to avoid being tagged with a digest.

Arguably in this case, that other html file shouldn't be part of the same bundle group?

packages/transformers/html/src/HTMLTransformer.js
wbinnssmith: Yeah, I did this originally. It seems like posthtml doesn't really version anything (or break things significantly so far, from what I can tell). I'd be on board with this.

packages/transformers/html/src/HTMLTransformer.js
wbinnssmith: 👍I know :) 

I used equality here because of the versioning ambiguity in posthtml, but I'll make the change below and use semver.

re: `astSatisfied`, I'm fine with letting plugins do this on their own, personally.

packages/core/utils/src/urlRelative.js
wbinnssmith: This isn't exactly the same, no? Maybe `urlJoin(path.dirname(from), to)` would be, but I'll have to check with hashes and querystrings.

packages/core/core/src/PackagerRunner.js
wbinnssmith: I agree -- this feels cleaner. We'll still have to tackle the sourcemaps issue post-packager, but I don't think it'll be any easier/more difficult.

packages/core/core/src/PackagerRunner.js
wbinnssmith: 👍

packages/namers/default/src/DefaultNamer.js
devongovett: Hmm yes. How did multiple html files get added to the same bundle group? Aren't all HTML dependencies async?

packages/core/core/src/TransformerRunner.js
wbinnssmith: not entirely certain, but it looks like their interfaces were compatible before. With `addURLDependency` that's no longer the case.

packages/core/core/src/PackagerRunner.js
wbinnssmith: thought right now PackagerRunner doesn't have access to the bundle graph. I might punt on this for now since it doesn't affect the public API here, but I agree it would be useful.

packages/core/core/src/public/BundleGraph.js
wbinnssmith: We use them

* When making new connections when adding a bundle to a group ([0] and [1])
* in the JS Runtime to determine if the bundle group is brought in via a URL or not (whether to add a runtime or not) [2]

[0] https://github.com/parcel-bundler/parcel/blob/v2/packages/core/core/src/public/BundleGraph.js#L146
[1] https://github.com/parcel-bundler/parcel/blob/v2/packages/core/core/src/public/BundleGraph.js#L186
[2] https://github.com/parcel-bundler/parcel/blob/v2/packages/runtimes/js/src/JSRuntime.js#L83

packages/namers/default/src/DefaultNamer.js
wbinnssmith: resolved :) `addURLDependency` makes sure of this now.

PR #3013, opened 2019-05-10
Make Some Core Parcel API Private

wbinnssmith:
* Moves all the current properties on the core `Parcel` object to private properties. None are currently used externally, and most (if not all) look that they shouldn't be.
* Removes the `.bundle()` method, which wasn't used publicly, simply called the bundler runner, and returned an internal-only `BundleGraph`. We now call out to the bundler runner directly inside Parcel.
* Removes `.package()` method, instead using it as an unexposed  function.

Currently `@parcel/register` isn't tested or flow-typed, so the test plan passes. However, `@parcel/register` broke a while back, and we should expose more narrow APIs on `Parcel` or elsewhere to get it functioning again.

Test Plan: flow, lint, test

### DIFF
--- a/packages/core/integration-tests/test/server.js
+++ b/packages/core/integration-tests/test/server.js
@@ -2,7 +2,7 @@ const assert = require('assert');
 const path = require('path');
 const fs = require('@parcel/fs');
 const logger = require('@parcel/logger');
-const {bundler} = require('@parcel/test-utils');
+const {bundler, sleep} = require('@parcel/test-utils');
 const http = require('http');
 const https = require('https');
 const getPort = require('get-port');
@@ -107,7 +107,11 @@ describe('server', function() {
 
   it('should serve a 500 if the bundler errored', async function() {
     let port = await getPort();
-    let b = bundler(path.join(__dirname, '/integration/commonjs/index.js'), {
+    let inputDir = path.join(__dirname, '/input/server-500');
+    await fs.ncp(path.join(__dirname, '/integration/commonjs'), inputDir);
+    let entry = path.join(inputDir, 'index.js');
+
+    let b = bundler(entry, {
       serve: {
         https: false,
         port: port,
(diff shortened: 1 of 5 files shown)


packages/core/integration-tests/test/server.js
```diff
@@ -118,17 +119,22 @@ describe('server', function() {
...
-    b.reporterRunner.report({
-      type: 'buildFailure',
-      error: new Error('This is a server test error')
-    });
+    let priorContent = await fs.readFile(entry);
+    await fs.writeFile(entry, 'syntax\\error');
+
+    // TODO: Programatically wait for an error to occur, or time out. Right now,
+    //       Parcel does not expose this without creating a reporter that must
+    //       be loaded from disk.
```
wbinnssmith: One solution to this would be to accept reporter objects as part of `new Parcel()`, or to implement emitter(s) that report information like in Parcel 1.

packages/core/integration-tests/test/server.js
```diff
@@ -118,17 +119,22 @@ describe('server', function() {
 
     await b.run();
 
-    b.reporterRunner.report({
```
wbinnssmith: Since this is no longer directly exposed, we have to test this in a more end-to-end way without using internals, which is probably for the better.

packages/core/integration-tests/test/server.js
```diff
@@ -107,7 +107,11 @@ describe('server', function() {
 
   it('should serve a 500 if the bundler errored', async function() {
     let port = await getPort();
-    let b = bundler(path.join(__dirname, '/integration/commonjs/index.js'), {
+    let inputDir = path.join(__dirname, '/input/server-500');
+    await fs.ncp(path.join(__dirname, '/integration/commonjs'), inputDir);
```
padmaia: Looks like a promisified ncp already exists in test utils? Do you think we'll need it outside of tests? Should probably remove from test utils if so, but not a blocker or anything.

PR #3015, opened 2019-05-10
Exit with a code of 1 for failures outside watch

wbinnssmith:
In the CLI, currently all exceptions are logged but are caught and allow the process to exit gracefully, even outside watch mode.

Instead:
* Catch all errors in watch mode **within Parcel**. With reporter infrastructure, it should be clear what caused an error, without disrupting 
	* Should certain classes of error allow Parcel to be terminated in watch mode? We should let things like e.g. transformer parse errors to keep the watcher going, but should more critical errors take down the watcher?
* In non-watch modes (not watch or server -- basically just the `build` command right now), throw exceptions from within `Parcel#run` (reject the promise it returns) and expect callers to handle them. This will ultimately be tools wrapping Parcel, but right now amounts to just the CLI.
* If the CLI catches an exception from Parcel, assume it was reported via a reporter and exit the process with code 1 immediately without logging or rethrowing.
	* Before this patch, errors were shown twice: once in a reporter, and once in the cli `console.error`.
* Catch all unhandled promise rejections in the CLI. We shouldn't do this within core as Parcel can be required as a library and we don't want to change the environment.
* Resolve a default `logLevel` of `info` at the start of Parcel.
* Don't show error stack traces unless `logLevel` is verbose.

Test Plan:
* Throw an error in the BundlerRunner (main process), and Transformer Runner (worker) and verify both cause Parcel to exit outside watch mode. 
	* Verify they let Parcel continue to run within watch mode, but that their messages are logged.
* Introduce a syntax error in the simple example and 
	* Verify that the long babel backtrace is not shown unless `logLevel` is verbose
	* Verify that the babel code frames continue to be shown. 
	* Verify that running with `logLevel` of verbose shows the babel stacktraces 
* flow, lint, test

### DIFF
--- a/packages/core/core/src/Parcel.js
+++ b/packages/core/core/src/Parcel.js
@@ -94,16 +94,32 @@ export default class Parcel {
     this._initialized = true;
   }
 
-  async run(): Promise<BundleGraph> {
+  // `run()` returns `Promise<?BundleGraph>` because in watch mode it does not
+  // return a bundle graph, but outside of watch mode it always will.
+  async run(): Promise<?BundleGraph> {
     if (!this._initialized) {
       await this.init();
     }
 
-    this.assetGraphBuilder.on('invalidate', () => {
-      this.build();
-    });
+    let resolvedOptions = nullthrows(this.resolvedOptions);
+    try {
+      this.assetGraphBuilder.on('invalidate', () => {
+        this.build().catch(e => {
+          if (!resolvedOptions.watch) {
+            throw e;
+          }
+        });
+      });
 
-    return this.build();
+      let graph = await this.build();
+      if (!resolvedOptions.watch) {
+        return graph;
+      }
+    } catch (e) {
+      if (!resolvedOptions.watch) {
+        throw e;
+      }
+    }
   }
 
   async build(): Promise<BundleGraph> {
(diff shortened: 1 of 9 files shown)


packages/reporters/cli/src/SimpleCLIReporter.js
```diff
@@ -88,11 +88,11 @@ function writeOut(message: string): void {
   stdout.write(message + '\n');
 }
 
-function writeErr(message: string | Error): void {
+function writeErr(message: string | Error, level: LogLevel): void {
   let error = prettyError(message, {color: false});
   // prefix with parcel: to clarify the source of errors
   writeErrLine('parcel: ' + error.message);
-  if (error.stack != null) {
+  if (error.stack != null && logLevels[level] >= logLevels.verbose) {
```
wbinnssmith: Maybe we should always log stacktraces in the SimpleCLIReporter since it's more utilitarian, and suppress them in non-verbose modes only in the main, user-friendly cli?

packages/core/core/src/Parcel.js
```diff
@@ -94,19 +94,34 @@ export default class Parcel {
     this._initialized = true;
   }
 
-  async run(): Promise<BundleGraph> {
+  // `run()` returns `Promise<?BundleGraph>` because in watch mode it does not
```
wbinnssmith: From a public api standpoint, what if watching were something you told Parcel to do, rather than a configuration option? conflating them both in `run` here is getting awkward.

It would be neat if `Parcel.watch()` could return something like an observable of `BundleGraph`s (or a similar summary structure containing it), since reporters have to be packages/modules on disk.

Issue #3007, opened 2019-05-09
Parcel 2: Source Maps


devongovett:
This issue is to track the general source map infrastructure in Parcel. Source map support for plugins should use that infrastructure to implement support for source maps.

Transformer plugins should be able to return source maps in addition to code. Those individual source maps will need to be combined together by a packager. In Parcel 1, this is a collaboration between the individual packager (e.g. JS Packager) and a general SourceMapPackager. The JSPackager provides offsets of the assets it writes, and the SourceMapPackager uses them to combine the source maps together.

For Parcel 2, I think we should do away with the SourceMapPackager, and instead provide a utility package for plugins to do their own source map building. I don't think offsets are enough information in many cases (e.g. tree shaking or other optimization), so it should be up to the packager/optimizer to do their own source map building. That said, a `@parcel/source-map` package could help with that.

In addition, ASTs from transformers should be stored in the cache so that packagers can optionally read them and combine them together at the AST level rather than the string code level. For advanced optimizations like tree shaking, this is necessary to retain high quality source mapping information without re-parsing both the code and source map.

The Packager API should change to return `{code, map}` instead of just code. Possibly packagers could return an AST as well, to send to optimizers? That will require more thought.

PR #3028, opened 2019-05-14
V2 Source Maps

DeMoorJasper:
# ↪️ Pull Request

This PR implements @parcel/source-map

This is based on Parcel 1 sourcemaps with a bunch of bugfixes, improvements, flow types and unit tests.

Related: #3007

- [x] update source-map package as we're dropping Node 6 support 🎉
- [x] Extract SourceMap into @parcel/source-map
- [x] Write unit tests for @parcel/source-map
- [x] Fix all known bugs for @parcel/source-map
- [x] Start using this package inside transformers
- [x] Figure out how packagers will create sourcemaps
- [x] read file content on stringify


packages/core/source-map/src/SourceMap.js
```diff
@@ -0,0 +1,441 @@
...
+    } else {
+      this.sources = new Map();
+    }
+
+    this.lineCount = 0;
+  }
+
+  // This should never be necessary as long as sourcemaps are valid
+  // It's still here in case some asset needs it but should be deprecated over time...
+  filterMappings(mappings: Array<MappingType>): Array<MappingType> {
```
DeMoorJasper: Not sure if we should remove this, it's probably no longer necessary, we used to use it in the constructor.
But in theory all mappings should be valid now that we support null mappings properly.

packages/core/source-map/src/SourceMap.js
```diff
@@ -0,0 +1,441 @@
...
+      let isValidGenerated =
+        typeof mapping.generated.line === 'number' &&
+        mapping.generated.line > 0 &&
+        typeof mapping.generated.column === 'number';
+
+      return isValidOriginal && isValidGenerated;
+    });
+  }
+
+  _validateMappings(mappings: Array<MappingType>) {
```
DeMoorJasper: This should be removed eventually this is just for testing some stuff and seeing if there are invalid mappings somewhere.

packages/core/source-map/src/SourceMap.js
```diff
@@ -0,0 +1,441 @@
...
+          line: line,
+          column: 0
+        }
+      });
+    }
+
+    return this;
+  }
+
+  async extendSourceMap(extension: SourceMap | RawMapInputType) {
```
DeMoorJasper: I've changed this from `async extendSourceMap(original, extension) {` into `async extendSourceMap(extension) {` as we already use it kind of this way in Parcel 1.
In Parcel 2 we will probably pass through and cache instances of the parcel sourcemap so this seemed like it might be faster as well.

packages/core/source-map/src/SourceMap.js
```diff
@@ -328,26 +328,35 @@ export default class SourceMap {
...
+      ) {
+        if (line < this.mappings[middleIndex][key].line) {
+          stopIndex = middleIndex - 1;
+        } else if (line > this.mappings[middleIndex][key].line) {
+          startIndex = middleIndex + 1;
+        }
+        middleIndex = Math.floor((stopIndex + startIndex) / 2);
+      }
+    } else {
+      middleIndex = this.mappings.findIndex(
```
DeMoorJasper: Unfortunately I'm not sure how to binary search efficiently with null values, so I added a search if we're searching for original positions.
Luckily this actually never happens inside Parcel. (we really never used this inside Parcel 1)
Perhaps we could even remove `generatedPositionFor(...)`?

packages/core/source-map/src/SourceMap.js
```diff
@@ -0,0 +1,424 @@
...
+    let mapping = this.mappings[index];
+    return {
+      source: mapping.source,
+      name: mapping.name,
+      line: mapping.original ? mapping.original.line : null,
+      column: mapping.original ? mapping.original.column : null
+    };
+  }
+
+  generatedPositionFor(originalPosition: PositionType) {
```
DeMoorJasper: Perhaps we should get rid of this, as we'll never use it inside Parcel afaik.

packages/core/source-map/src/SourceMap.js
```diff
@@ -0,0 +1,385 @@
+// @flow
+import {SourceMapConsumer, SourceMapGenerator} from 'source-map';
+import lineCounter from '../../utils/src/lineCounter';
```
wbinnssmith: nit: can we rename this to `countLines` or a verb of some kind? lineCounter sounds like an object or a class.

also this should be exported at the top level of `@parcel/utils` and imported like `import {countLines} from '@parcel/utils`.

we should probably have a lint rule for requiring outside of packages with `../` since that will break on publish

packages/core/source-map/src/SourceMap.js
```diff
@@ -0,0 +1,385 @@
...
+
+type PositionType = {
```
wbinnssmith: Can these be exact types? i.e. `{| ... |}`

packages/core/source-map/src/SourceMap.js
```diff
@@ -0,0 +1,385 @@
...
+  line: number,
+  column: number
+};
+
+type MappingType = {
+  generated: PositionType,
+  original: PositionType | null,
```
wbinnssmith: `?Position` is fine

packages/core/source-map/src/SourceMap.js
wbinnssmith: maybe just `type Position` rather than `PositionType`?

packages/core/source-map/src/SourceMap.js
```diff
@@ -0,0 +1,385 @@
...
+  name: string
+};
+
+export default class SourceMap {
+  mappings: Array<MappingType>;
+  sources: Map<string, string | null>;
+
+  constructor(
+    mappings?: Array<MappingType> = [],
+    sources?: Map<string, string> | {[key: string]: string}
```
wbinnssmith: we do really need to accept either or can we settle on one?

packages/core/source-map/src/SourceMap.js
wbinnssmith: Since this doesn't rely on `this`, maybe it can be extracted into its own module and imported here instead?

packages/core/source-map/src/SourceMap.js
```diff
@@ -0,0 +1,385 @@
...
+        throw new Error('generated mapping is undefined');
+      }
+
+      if (!mapping.source) {
+        throw new Error('source should be defined');
+      }
+
+      let isValidOriginal =
+        mapping.original === null ||
+        (typeof mapping.original.line === 'number' &&
```
wbinnssmith: Since the typing guarantees that if `original` is present, doesn't that mean that `mapping.original.line` must be a number?

packages/core/source-map/src/SourceMap.js
```diff
@@ -0,0 +1,385 @@
...
+          mapping.original.line > 0 &&
+          typeof mapping.original.column === 'number' &&
+          mapping.source);
+
+      if (!isValidOriginal) {
+        throw new Error('Invalid original mapping');
+      }
+
+      let isValidGenerated =
+        typeof mapping.generated.line === 'number' &&
```
wbinnssmith: same as above: this can be

`mapping.generated != null && mapping.generated.line > 0`, yeah?

packages/core/source-map/src/SourceMap.js
```diff
@@ -0,0 +1,385 @@
...
+    return new SourceMapConsumer(sourcemap);
+  }
+
+  async _addSourceMap(
+    map: SourceMap,
+    lineOffset: number = 0,
+    columnOffset: number = 0
+  ) {
+    if (lineOffset === 0 && columnOffset === 0) {
+      Array.prototype.push.apply(this.mappings, map.mappings);
```
wbinnssmith: `this.mappings.push(...map.mappings)`? 😃

packages/core/source-map/src/SourceMap.js
```diff
@@ -0,0 +1,385 @@
...
+  }
+
+  addConsumerMapping(
+    mapping: ConsumerMappingItemType,
+    lineOffset: number = 0,
+    columnOffset: number = 0
+  ) {
+    let original = null;
+    if (
+      typeof mapping.originalLine === 'number' &&
```
wbinnssmith: same as above

packages/core/source-map/src/SourceMap.js
```diff
@@ -0,0 +1,385 @@
...
+
+    if (column < 0) {
+      throw new Error('Column numbers must be >= 0');
+    }
+
+    if (this.mappings.length < 1) {
+      return null;
+    }
+
+    var startIndex = 0;
```
wbinnssmith: `let`? 😄 (I know this is from Parcel 1, but we could update here while we're at it)

packages/core/source-map/src/SourceMap.js
```diff
@@ -0,0 +1,385 @@
...
+    return {
+      source: mapping.source,
+      name: mapping.name,
+      line: mapping.original ? mapping.original.line : null,
+      column: mapping.original ? mapping.original.column : null
+    };
+  }
+
+  sourceContentFor(fileName: string): string | null {
+    return this.sources.get(fileName) || null;
```
wbinnssmith: `return this.sources.get(fileName) ?? null;`, in the case the source is `''`?

wbinnssmith:
Looking awesome 😃

packages/core/source-map/src/SourceMap.js
wbinnssmith: oh, I see this is all validation stuff and probably should err on the side of being safe.

packages/core/source-map/src/SourceMap.js
wbinnssmith: update:

> we should probably have a lint rule for requiring outside of packages with ../ since that will break on publish

did this in #3038 😃

packages/core/source-map/src/SourceMap.js
DeMoorJasper: sure, would make sense

packages/core/source-map/src/SourceMap.js
DeMoorJasper: Not sure if that matters, there were performance issues with let in v1 version

packages/core/source-map/src/SourceMap.js
DeMoorJasper: Sure

packages/core/source-map/src/SourceMap.js
DeMoorJasper: Would like to benchmark these changes once I've got my hands on a bunch of sourcemaps I can run through this

packages/core/source-map/src/SourceMap.js
DeMoorJasper: Can't update it as it throws eslint errors. That's why I solved it like this...`eslint(no-restricted-imports)`

PR #3032, opened 2019-05-15
Fix sourcemap loading if name contains spaces

DeMoorJasper:
# ↪️ Pull Request

If a sourcemap had any spaces in it's filename sourcemapLoader would fail and create syntax errors by only removing a part of the sourcemap

## 💻 Examples

This would fail:

```JS
//# sourceMappingURL=this is a sourcemap.js.map
```

Now it works :)

## ✔️ PR Todo


mischnic:
The sourcemap **URL** should never contain literal spaces (URL encoded with `%20`)
But if some npm package does indeed have such an URL....

DeMoorJasper:
@mischnic there are packages that use spaces unfortunately... Also thought it was against the spec but app they did it anyway

PR #3035, opened 2019-05-15
Implement Graph with adjacency lists rather than edge lists

wbinnssmith:
* The existing edge list implementation of `Graph` very frequently iterates the entire list of edges in the graph to find outbound and inbound edges for specific uses.
* Instead, create two adjacency lists[0]: one for outbound edges from nodes and one for inbound edges. This means that getting a node's inbound or outbound edges no longer grows with the entire set of edges in the graph, and instead only grows with the node's own edges.
* `graph.hasEdge` and `graph.removeEdge` are now O(1) rather than O(E).
* `graph.getNodesConnectedTo` and `graph.getNodesConnectedFrom` are now bounded by the number of edges attached to the node rather than O(E).
* Retains an interfaces for retrieving all the edges in the graph. Like `graph.edges`, except it returns an array rather than a set (edge membership can still be tested with `graph.hasEdge`).
    * This is used to serialize the graph, as well as for graph copying.

This also implements `DefaultMap`, like python's `defaultdict`, for easy setting of default mappings if they don't exist for a node.

Test Plan: `yarn test`, and manual test the simple example.

[0] https://en.wikipedia.org/wiki/Adjacency_list

### DIFF
--- a/packages/core/core/src/BundlerRunner.js
+++ b/packages/core/core/src/BundlerRunner.js
@@ -173,12 +173,12 @@ export default class BundlerRunner {
       // the node to it.
       bundle.assetGraph.merge(subGraph);
 
-      bundle.assetGraph.addEdge({
-        from: dependency
+      bundle.assetGraph.addEdge(
+        dependency
           ? dependency.id
           : nullthrows(bundle.assetGraph.getRootNode()).id,
-        to: entryId
-      });
+        entryId
+      );
     }
   }
 }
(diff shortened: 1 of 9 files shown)


wbinnssmith:
One thing we could do away with is the representation of Edges as objects. This would make the AdjacencyList `Map<NodeId, Set<Node, Id>>` which currently is sufficient to describe edges. If we added edge types, this could change.

wbinnssmith:
> One thing we could do away with is the representation of Edges as objects, at least internally (we'd still need a way to describe edges at the interface boundaries).

Just pushed an update that implements this, as well as removing edge objects from most public APIs. They're still used for serializing, deserializing, and copying graphs.

PR #3037, opened 2019-05-16
PostCSS and CSS Modules support

wbinnssmith:
This implements support for postcss and css modules in Parcel 2.

* Inspired by create-react-app [0], files now opt-into css modules support by including `.module.` in their name, like `foo.module.css`. Support for the non-standard `"modules": true`in postcss config has been removed. PostCSS configs that use the `postcss-modules` plugin continue to be supported (though they are adapted to work with Parcel).
* Based on code and contributions to Parcel 1's PostCSS support.
* Integrates some of @padmaia's code in the `wip-css-support` branch, especially `loadExternalPlugins.js`
* Pass a `resolve` function to transformer's `getConfig` function. This allows us to override PostCSS's internal resolution to use Parcel's configured resolvers.
* Transformer returns both a css asset as well as a js asset, which provides the export 
    * Should this be moved to the JS runtime like raw assets? Raw asset urls are built into Parcel, but this doesn't seem like it should be. 
    * Should there be a way to override what the JS runtime returns when a given dependency is requested?
* Adds Parcel dependencies for anything matching the `composes: foo from './bar'` pattern

**One significant change:** we now prune assets that don't match the bundle type when cloning the asset's subtree in the main asset graph. Prior to this change, it was possible for bundles to contain assets of different types in some circumstances.

Test Plan: Flow, lint, test. Separated and enabled the postcss integration tests from v1.

[0] https://facebook.github.io/create-react-app/docs/adding-a-css-modules-stylesheet


packages/core/core/src/public/MainAssetGraph.js
```diff
@@ -38,6 +38,12 @@ export default class MainAssetGraph implements IMainAssetGraph {
       to: assetNode.id
     });
 
+    graph.traverseAssets(currentAsset => {
```
wbinnssmith: We now prune assets that don't match the bundle type when cloning the asset's subtree in the main asset graph.

This could be done in the creation of the subgraph itself, but given @padmaia's work on new node types which might include the creation of a leaner graph, I didn't want to create a bunch of unnecessary work. This can and should probably change down the road.

packages/transformers/postcss/src/PostCSSTransformer.js
```diff
@@ -0,0 +1,195 @@
+// @flow
+
+import type {FilePath, MutableAsset} from '@parcel/types';
```
kwelch: Does this need to be in the dep list?

wbinnssmith: Yarn workspaces links everything at the repo root so flow can find this fine. And since it's not a real runtime dependency (it's `import type`), this won't cause any issues when it comes to using published versions.

It certainly works for now, and I think if we decide one way or another, we should enforce it via lint. Right now I'm trying not to list any `@parcel` packages in `devDependencies` since it's not strictly necessary. Though as @padmaia mentioned this does couple the package to our current monorepo (specifically yarn workspace) structure and it would be ideal if packages can be trivially removed if necessary.

wbinnssmith: I'm sure @jamiebuilds has thoughts on this 😃

kwelch: I am very on the fence with this, the wind could easily take me to either side. As for this I just wanted to ensure it was intentional.

PR #3039, opened 2019-05-16
Add source-map types from flow-typed and integrate them

wbinnssmith:
@DeMoorJasper, feel free to merge or do what you want with this.

### DIFF
--- a/packages/core/source-map/src/SourceMap.js
+++ b/packages/core/source-map/src/SourceMap.js
@@ -1,46 +1,25 @@
-// @flow
+// @flow strict-local
+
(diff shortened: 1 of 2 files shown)


packages/core/source-map/src/SourceMap.js
```diff
@@ -68,12 +47,12 @@ export default class SourceMap {
...
       }
 
-      if (!mapping.source) {
+      if (mapping.source == null) {
         throw new Error('source should be defined');
       }
 
       let isValidOriginal =
-        mapping.original === null ||
+        mapping.original == null ||
```
DeMoorJasper: Why did you change it to `== null`?

PR #3044, opened 2019-05-16
Restore webp to raw globs in default config

wbinnssmith:
This restores *.webp to the raw transformer and packager globs. Looks like a bad merge left this out.

### DIFF
--- a/packages/configs/default/index.json
+++ b/packages/configs/default/index.json
@@ -8,7 +8,7 @@
     ],
     "*.css": ["@parcel/transformer-css"],
     "*.{htm,html}": ["@parcel/transformer-html"],
-    "*.{jpg,jpeg,png,gif,svg,webm,pdf,txt}": ["@parcel/transformer-raw"]
+    "*.{jpg,jpeg,png,gif,svg,webm,webp,pdf,txt}": ["@parcel/transformer-raw"]
   },
   "namers": ["@parcel/namer-default"],
   "runtimes": {
@@ -19,7 +19,7 @@
     "*.html": "@parcel/packager-html",
     "*.css": "@parcel/packager-css",
     "*.js": "@parcel/packager-js",
-    "*.{jpg,jpeg,png,gif,svg,webm,pdf,txt}": "@parcel/packager-raw"
+    "*.{jpg,jpeg,png,gif,svg,webm,webp,pdf,txt}": "@parcel/packager-raw"
   },
   "resolvers": ["@parcel/resolver-default"],
   "reporters": [


mischnic:
> Would be cool if there would be a way to handle all unknown extensions with raw, but it's not in the RFC so I guess that's not gonna happen.

`*.*` (or `*`) is a valid glob, so this should be fine?
```
  "packagers": {
    "*.html": "@parcel/packager-html",
    "*.*": "@parcel/packager-raw"
  }
```

wbinnssmith:
> *.* (or *) is a valid glob, so this should be fine?

Yeah, let's do that.

My only concerns is that JSON (and JavaScript objects) don't have predictable key orders by spec, so how would we determine precedence?

mischnic:
> My only concern is that JSON (and JavaScript) objects don't have predictable key orders by spec, so how would we determine precedence?

Correct. But the RFC says:

> While objects in JSON are technically unordered, Parcel
does use the order to give globs priority when a file name is being tested
against them

😐

wbinnssmith:
> Just waiting for Node JSON.parse to somehow have unstable Object order and having to use a custom JSON parser....

V8 has respected insertion order for a long time, so I'd be surprised if it changed. That said, what about environments like browsers or future js environments Parcel will run in?

wbinnssmith:
cc @jamiebuilds @jamiekyle-eb @devongovett for the RFC

mischnic:
Firefox and Safari don't use V8.

jamiebuilds:
We can add our own meaning on top of JSON, it's not a terribly big deal. We really only need to be concerned with V8, but just for the record both Safari and Firefox give you the properties in a stable order too:

Screen Shot 2019-05-16 at 3 54 17 PM
Screen Shot 2019-05-16 at 3 55 06 PM


If all else fails, we can easily write JSON parser that gives you the properties in a stable order.

The only time any of the JavaScript engines won't give you keys back in the order they were defined in is if they were numbers:

```js
Object.keys(JSON.parse('{ "2": 1, "1": 2 }')) // ["1", "2"]
```

But I'm not terribly worried about someone naming a file 42 and then writing a glob "42" and then it somehow also matching something else.

jamiebuilds:
Btw we also could have gone with something like:

```json
{
  "resolvers": ["@parcel/resolver-default"],
  "transforms": [
    { "match": "*.vue", "use": ["@parcel/transform-vue"] },
    { "match": "*.scss", "use": ["@parcel/transform-sass"] },
    { "match": "*.js", "use": ["@parcel/transform-babel"] },
    { "match": "*.css", "use": ["@parcel/transform-postcss"] },
    { "match": "*.html", "use": ["@parcel/transform-posthtml"] },
  ],
  "packagers": [
    { "match": "*.js", "use": "@parcel/packager-js" },
    { "match": "*.css", "use": "@parcel/packager-css" },
    { "match":  "*.html", "use": "@parcel/packager-html" },
    { "match": "*.wasm", "use": "@parcel/packager-wasm" },
    { "match": "*.raw", "use": "@parcel/packager-raw" }
  ],
  "optimizers": [
    { "match": "*.js", "use": ["@parcel/optimizer-uglify"] },
    { "match": "*.css", "use": ["@parcel/optimizer-cssnano"] },
    { "match": "*.html", "use": ["@parcel/optimizer-htmlnano"] },
    { "match": "*.{png,jpg,jpeg,svg,...}", "use": ["@parcel/optimizer-imagemin"] }
  ],
  "loaders": [
    { "match": "*.js", "use": "@parcel/loader-js" },
    { "match": "*.wasm", "use": "@parcel/loader-wasm" }
  ],
  "reporters": ["@parcel/reporter-detailed"]
}
```

But we wanted the config to have a very lightweight feel to it, with as little typing as reasonable.

```json
{
  "resolvers": ["@parcel/resolver-default"],
  "transforms": {
    "*.vue": ["@parcel/transform-vue"],
    "*.scss": ["@parcel/transform-sass"],
    "*.js": ["@parcel/transform-babel"],
    "*.css": ["@parcel/transform-postcss"],
    "*.html": ["@parcel/transform-posthtml"],
  },
  "packagers": {
    "*.html": "@parcel/packager-html",
    "*.wasm": "@parcel/packager-wasm",
    "*.raw": "@parcel/packager-raw"
  },
  "optimizers": {
    "*.js": ["@parcel/optimizer-uglify"],
    "*.css": ["@parcel/optimizer-cssnano"],
    "*.html": ["@parcel/optimizer-htmlnano"],
    "*.{png,jpg,jpeg,svg,...}": ["@parcel/optimizer-imagemin"]
  },
  "loaders": {
    "*.js": "@parcel/loader-js",
    "*.wasm": "@parcel/loader-wasm"
  },
  "reporters": ["@parcel/reporter-detailed"]
}
```

I also didn't want people to draw comparisons between this format and Webpack's on a syntactic similarity basis, because the important details of Parcel are besides this config format.

PR #3047, opened 2019-05-16
Resolve option file paths as absolute

wbinnssmith:
This resolves all the file paths present in parcel options as absolute paths. Following `path.resolve(filePath)`, behavior, if a path is not absolute:

* Target `distDir`s resolved from package.json entries are resolved from the directory of the package.json file
    * Currently (before and after this patch), if targets are specified in `package.json` and the default is used, its distDir is actually resolved from `cwd`, not from the package.json file. Should we change this to be resolved from the package.json's directory instead?
* Target `distDir`s passed in the programmatic api (`new Parcel({targets: [{...}, ...})`  are resolved from cwd.
* `cacheDir` is always resolved from cwd.

Test Plan:
* Add a log after options are resolved and run the simple example. Verify every path listed is absolute.
* Add a log after bundle file paths are added and verify bundle file paths are absolute.
* `yarn test`

### DIFF
--- a/packages/core/core/src/TargetResolver.js
+++ b/packages/core/core/src/TargetResolver.js
@@ -26,11 +26,13 @@ const DEVELOPMENT_BROWSERS = [
   'last 1 Edge version'
 ];
 
-const DEFAULT_DIST_DIR = 'dist';
+const DEFAULT_DIST_DIRNAME = 'dist';
+const COMMON_TARGETS = ['main', 'module', 'browser'];
 
 export default class TargetResolver {
   async resolve(
     rootDir: FilePath,
+    cacheDir: FilePath,
     initialOptions: InitialParcelOptions
   ): Promise<Array<Target>> {
     let packageTargets = await this.resolvePackageTargets(rootDir);
@@ -51,7 +53,12 @@ export default class TargetResolver {
           return matchingTarget;
         }
 
-        return target;
+        return {
+          ...target,
+          // In the case the supplied target was a relative path, resolve it from
+          // cwd just like Node does for other file paths.
+          distDir: path.resolve(target.distDir)
+        };
       });
 
       if (serveOptions) {
@@ -74,7 +81,10 @@ export default class TargetResolver {
         targets = [
           {
             name: 'default',
-            distDir: 'dist',
+            // For serve, write the `dist` to inside the parcel cache, which is
+            // temporary, likely in a .gitignore or similar, but still readily
+            // available for introspection by the user if necessary.
+            distDir: path.resolve(cacheDir, DEFAULT_DIST_DIRNAME),
             publicUrl: serveOptions?.publicUrl ?? '/',
             env: new Environment({
               context: 'browser',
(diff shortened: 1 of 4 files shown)


packages/core/core/src/BundlerRunner.js
```diff
@@ -61,11 +61,16 @@ export default class BundlerRunner {
...
+    let bundleNames = bundles.map(
+      b =>
+        nullthrows(b.target).name +
+        ':' +
+        normalizeSeparators(nullthrows(b.name))
+    );
     assert.deepEqual(
-      bundlePaths,
-      unique(bundlePaths),
-      'Bundles must have unique filePaths'
```
wbinnssmith: Actually, I think filePath might be a better sanity check. Since it's possible to generate a name with slashes, and distDirs are independent, it's possible that more than one {target,name} combination could pick the same file path. I don't think the converse is possible.

Or maybe we do both? @devongovett @padmaia what do you think?

packages/core/core/src/resolveOptions.js
```diff
@@ -49,14 +49,18 @@ export default async function resolveOptions(
     ])) || path.join(process.cwd(), 'index')
   );
 
+  let cacheDir =
+    // If a cacheDir is provided, resolve it relative to cwd. Otherwise,
+    // use a default directory resolved relative to the project root.
+    initialOptions.cacheDir != null
```
wbinnssmith: This is a breaking change, but I think it makes sense. At least for the default, the cache dir shouldn't depend on where one invokes `parcel` from, but instead should be centralized at the project root.

If one is provided though, we should following Node's typical behavior of resolving relative filePaths to the pwd. If callers want to be certain, they should provide an absolute path.

This does change a few things: for example, when running integration tests, now the cache directory is at the root of the repo workspace rather than in the integration tests package.

cc @devongovett @padmaia

wbinnssmith:
Ah yeah, forgot about that 😄 Updated.

packages/core/core/src/TargetResolver.js
```diff
@@ -74,7 +80,7 @@ export default class TargetResolver {
         targets = [
           {
             name: 'default',
-            distDir: 'dist',
+            distDir: path.resolve(DEFAULT_DIST_DIRNAME),
```
wbinnssmith: hmm, should this be relative to the package.json file instead?

@padmaia @devongovett

padmaia: Probably. `cwd` doesn't feel quite right.

packages/core/core/src/resolveOptions.js
padmaia: Do we clean up the cache used for tests? Should we? If so, it would probably be easiest to nuke it after each test run which would be bad if we want to keep a `.parcel-cache` around for building Parcel itself. We keep saying it, but using a in memory filesystem for tests would solve this nicely.

packages/core/core/src/TargetResolver.js
wbinnssmith: For `serve`, what do you think about using a temp directory for dist? I'd rather not write to the user's project unexpectedly, especially if they weren't expecting a directory named `dist`.

cc @devongovett

packages/core/core/src/resolveOptions.js
wbinnssmith: Nope, we don't even clean up dist directories in most cases.

I think ideally each test gets its own dist directory and cache, each in temp directories, which are cleared after each test.

Would be neat if we could use something like an overlay fs [0] (popularized by docker) to have reads of existing files go through the real filesystem, and have the remaining writes and reads be in memory.

[0] https://en.wikipedia.org/wiki/OverlayFS

packages/core/core/src/TargetResolver.js
devongovett: It's nice to be able to easily inspect the files that parcel generates. Also, you'd get a `dist` directory when building for prod, so you'd need it in your `.gitignore` anyway so probably not a big deal if it shows up in dev as well.

devongovett: and yeah... I think it should be relative to package.json if it was declared there.

wbinnssmith: It's possible the user has configured a different output directory than `dist` for their other targets. This is for one we come up with on the fly for serve. Should we just use whatever they specify for another target? 😬

PR #3048, opened 2019-05-16
Watching refactor

padmaia:
# ↪️ Watching refactor

As part of the caching plan, we will need to not only know if files that exist in the asset graph have changed, but also if certain files that do not exist in the asset graph have been added (e.g. a babel config file being added).  The watcher is now setup in `Parcel.js` and will watch all files in the project root. I also swapped in the new watcher so that we will be able to query which files have been changed/updated/deleted since the last run of Parcel.

## 🚨 Test instructions

Run `yarn demo` in simple example and make sure it is rebuilding the files you change


packages/core/core/src/resolveOptions.js
```diff
@@ -38,6 +39,16 @@ export default async function resolveOptions(
...
 
+  let projectRoot = path.dirname(
+    (await resolveConfig(path.join(process.cwd(), 'index'), [
+      'yarn.lock',
+      'package-lock.json',
+      'pnpm-lock.yaml',
+      '.git',
+      '.hg'
+    ])) || path.join(process.cwd(), 'index')
+  );
```
padmaia: Are we concerned about this affecting start up time? Would not be great for people who are not building from a package or git repo, but I'm not sure how often that happens.

packages/core/core/src/Parcel.js
```diff
@@ -92,6 +95,41 @@ export default class Parcel {
...
+    if (this.#resolvedOptions.watch) {
+      // TODO: ideally these should all be absolute paths already set up on #resolvedOptions
+      let targetDirs = this.#resolvedOptions.targets.map(target =>
+        path.resolve(process.cwd(), target.distDir)
+      );
+      let cacheDir = path.resolve(
+        process.cwd(),
+        this.#resolvedOptions.cacheDir
+      );
+      let vcsDirs = ['.git', '.hg'].map(dir => path.join(projectRoot, dir));
```
padmaia: This would not handle the case where the vcs folder is nested deeply inside of the project root, but that does not seem common. It could be outside of the project root, but we won't be watching outside of project root, so it shouldn't matter.

packages/core/core/src/resolveOptions.js
padmaia: Also this would maybe not work well if people were using a monorepo of git sub modules and building from within one of those sub modules. Is that something we care to support?

flow-libs/parcel-watcher.js.flow
```diff
@@ -0,0 +1,79 @@
+// @flow
+
+// Derived from the README and source of ncp located at
```
wbinnssmith: this isn't relevant for this one 😃

flow-libs/parcel-watcher.js.flow
```diff
@@ -0,0 +1,79 @@
...
+    ): Promise<mixed>,
+    writeSnapshot(
+      dir: FilePath,
+      snapshot: FilePath,
+      opts: Options
+    ): Promise<FilePath>
+  };
+}
+
+// const binding = require('bindings')('watcher.node');
```
wbinnssmith: can get rid of this too

packages/core/core/package.json
```diff
@@ -21,7 +21,7 @@
     "@parcel/plugin": "^2.0.0",
     "@parcel/types": "^2.0.0",
     "@parcel/utils": "^2.0.0",
-    "@parcel/watcher": "^2.0.0",
+    "@parcel/watcher": "^2.0.0-alpha.3",
```
wbinnssmith: I wonder if we should be versioning the other packages in core as something like this so we don't publish 2.0 before it's ready

packages/core/core/src/AssetGraphBuilder.js
```diff
@@ -187,6 +160,20 @@ export default class AssetGraphBuilder extends EventEmitter {
       }
     }
   }
+
+  isInvalid() {
+    return !!this.graph.invalidNodes.size;
```
wbinnssmith: Just preference, but maybe `return this.graph.invalidNodes.size > 0;`?

packages/core/core/src/AssetGraphBuilder.js
```diff
@@ -187,6 +160,20 @@ export default class AssetGraphBuilder extends EventEmitter {
...
+  }
+
+  respondToFSEvents(events: Array<Event>) {
+    for (let {type, path} of events) {
+      // TODO: eventually handle all types of events
+      if (type === 'update' && this.graph.hasNode(path)) {
+        this.controller.abort();
```
wbinnssmith: does aborting this signal synchronously cause an error to throw? if it does, it would prevent other files in this event from getting invalidated

packages/core/core/src/Parcel.js
```diff
@@ -92,6 +95,41 @@ export default class Parcel {
 
     this.#runPackage = this.#farm.mkhandle('runPackage');
 
+    await this.initializeWatcher();
+  }
+
+  async initializeWatcher() {
+    let projectRoot = this.#resolvedOptions.projectRoot;
+    if (this.#resolvedOptions.watch) {
+      // TODO: ideally these should all be absolute paths already set up on #resolvedOptions
```
wbinnssmith: If you review and approve #3047 this will come true 😅

packages/core/core/src/Parcel.js
```diff
@@ -92,6 +95,41 @@ export default class Parcel {
...
+        process.cwd(),
+        this.#resolvedOptions.cacheDir
+      );
+      let vcsDirs = ['.git', '.hg'].map(dir => path.join(projectRoot, dir));
+      let ignore = [cacheDir, ...targetDirs, ...vcsDirs];
+      this.#watcher = await watcher.subscribe(
+        projectRoot,
+        (err, events) => {
+          if (err) {
+            throw err;
```
wbinnssmith: right now this will throw the error in the same tick that this callback fires, and will be uncaught. Maybe this could return something that signals the error. Observables would be really nice here...

packages/core/core/src/Parcel.js
```diff
@@ -92,6 +95,41 @@ export default class Parcel {
...
+          }
+
+          this.#assetGraphBuilder.respondToFSEvents(events);
+          if (this.#assetGraphBuilder.isInvalid()) {
+            this.build().catch(() => {
+              // Do nothing, in watch mode reporters should alert the user something is broken, which
+              // allows Parcel to gracefully continue once the user makes the correct changes
```
wbinnssmith: maybe do the same `instanceof BuildError` check here?

packages/core/core/src/resolveOptions.js
wbinnssmith: > Also this would maybe not work well if people were using a monorepo of git sub modules and building from within one of those sub modules. Is that something we care to support?

This is probably going come come up at some point.

> Are we concerned about this affecting start up time? 

Since this doesn't depend on stuff from earlier in this function, and is async, we could probably run this concurrently with the other async results above with `await Promise.all(...)`

packages/core/core/test/AssetGraphBuilder.test.js
```diff
@@ -35,7 +35,8 @@ const DEFAULT_OPTIONS = {
   entries: [],
   logLevel: 'none',
   rootDir: FIXTURES_DIR,
-  targets: []
+  targets: [],
```
wbinnssmith: I wish we configured prettier to do trailing commas. That would get rid of a lot of these diffs 😢

packages/core/parcel/src/cli.js
```diff
@@ -144,9 +145,11 @@ async function run(entries: Array<string>, command: any) {
...
     await parcel.run();
   } catch (e) {
-    // Simply exit if the run fails. If an exception is thrown during the run,
-    // it is given to reporters in a buildFailure event. In watch mode,
-    // exceptions won't be thrown beyond Parcel.
+    // If an exception is thrown during Parcel.build, it is given to reporters in a
+    // buildFailure event. In watch mode, build exceptions won't be thrown beyond Parcel.
+    // If an exception is thrown outside of Parcel.build, it won't go through reporters
+    // and should be logged.
+    if (!(e instanceof BuildError)) console.error(e);
```
wbinnssmith: 👏

packages/core/core/test/AssetGraphBuilder.test.js
wbinnssmith: cc @devongovett @DeMoorJasper to bikeshed this

I'd love to make this happen once PRs and Parcel 1 activity calms down

flow-libs/parcel-watcher.js.flow
padmaia: Woops, was so happy to get flow passing, I forgot to clean up :)

packages/core/core/src/AssetGraphBuilder.js
wbinnssmith: Looks like that won't be an issue. It's kind of weird the signal can be aborted multiple times, but that's probably fine too. Could move it out of the loop if you want.

packages/core/core/src/Parcel.js
padmaia: Wouldn't it only ever be an instance of `BuildError`?  Since `build()` catches all errors and either swallows them if their a `BuildAbortError` or wraps them in a `BuildError` and throws.

packages/core/core/src/AssetGraphBuilder.js
padmaia: No, it doesn't throw anything it just sets the `signal.aborted` property to true. I don't think it would hurt anything, but it is kind of weird that it might be called multiple times now that we are receiving multiple events at once. I decided to move it to the beginning of the build method, where I think it makes a little more sense.

PR #3061, opened 2019-05-20
Implement explicit `watch` function and remove option

wbinnssmith:
Rather than implement watch as an option to the `Parcel` constructor, allow for callers to subscribe to results from a `.watch()` function instead.

Here's the interface:

```
  async watch(
    cb?: (err: ?Error, buildEvent?: BuildEvent) => mixed
  ): Promise<AsyncSubscription>;
```

`watch` accepts a callback that receives either an error object or a `BuildEvent`. `BuildEvent` is either a success or failure event identical to what we give reporters. 

Setting up a watch subscription itself is an asynchronous process, so a promise to an `AsyncSubscription` is returned. An `AsyncSubscription` is an object with an asynchronous `unsubscribe` method.

`parcel.watch()` supports multiple subscribers by returning a subscription object that allows callers to unsubscribe when they are no longer interested in events. On the first watch subscription, Parcel begins subscribing to the underlying `@parcel/watcher`, and when the last subscriber unsubscribes, the inner watcher subscription is unsubscribed and removed.

This means that `parcel.run()` has predictable behavior (always returning a promise of a single result), and that callers interested in build successes and failures over time (such as the server tests) can receive them.

Since `watch` is no longer a parcel option, reporters like the server reporter need to know when parcel has been invoked with `watch` as opposed to a single `run`. This change adds `watchStart` and `watchEnd` events to be broadcast to reporters, allowing for reporters to respond like starting a server. The `watchEnd` event is also awaited, so things like servers can be shut down.

Test Plan: Reworked server tests to use the watch api and new `watchStart` and `watchEnd` events.

### DIFF
--- a/packages/core/parcel/src/cli.js
+++ b/packages/core/parcel/src/cli.js
@@ -142,15 +142,21 @@ async function run(entries: Array<string>, command: any) {
     ...(await normalizeOptions(command))
   });
 
-  try {
-    await parcel.run();
-  } catch (e) {
-    // If an exception is thrown during Parcel.build, it is given to reporters in a
-    // buildFailure event. In watch mode, build exceptions won't be thrown beyond Parcel.
-    // If an exception is thrown outside of Parcel.build, it won't go through reporters
-    // and should be logged.
-    if (!(e instanceof BuildError)) console.error(e);
-    process.exit(1);
+  if (command.name() === 'watch' || command.name() === 'serve') {
+    await parcel.watch(err => {
+      if (err) {
+        throw err;
+      }
+    });
+  } else {
+    try {
+      await parcel.run();
+    } catch (e) {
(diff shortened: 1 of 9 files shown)


wbinnssmith:
cc @DeMoorJasper for changes to the server reporter

packages/core/parcel/src/cli.js
```diff
@@ -142,15 +142,23 @@ async function run(entries: Array<string>, command: any) {
...
+      if (err) {
+        throw err;
+      }
+    });
+  } else {
+    try {
+      await parcel.run();
+    } catch (e) {
+      // If an exception is thrown during Parcel.build, it is given to reporters in a
+      // buildFailure event. In watch mode, build exceptions won't be thrown beyond Parcel.
```
padmaia: Don't really need the "In watch mode[...]" sentence anymore

PR #2966, opened 2019-04-30
Remove legacy posthtml-include support (addDependencyTo)

christianscott:
# ↪️ Pull Request

Reopening #2166 on behalf of @LandonSchropp ([I asked if it was OK 🙂](https://twitter.com/LandonSchropp/status/1122638522240733185)).

Only difference is I didn't remove the test case, which seems to pass just fine.

## 💻 Examples

See #2166.

## 🚨 Test instructions

See #2166.

## ✔️ PR Todo

- [x] Added/updated unit tests for this change
- [x] Filled out test instructions (In case there aren't any unit tests)
- [x] Included links to related issues/PRs

PR #3063, opened 2019-05-21
Parcel v2:  PostHTML transformer

### LINKED ISSUE #2159: Parcel 2: PostHTML Transform
Create the `@parcel/transform-posthtml` package in `packages/transforms/posthtml`. This package is a Parcel 2 transform plugin for PostHTML. It will be based on the `posthtml` transform in `HTMLAsset` from Parcel 1, split into its own package.

The `@parcel/transform-posthtml` package is responsible for the following things:

* Loading `.posthtmlrc` files
* Parsing and transforming if needed

kwelch:
# ↪️ Pull Request
Migration of the PostHTML transformer into v2.

Closes #2159.

## 🚨 Test instructions
yarn test

## ✔️ PR Todo

### DIFF
--- a/flow-libs/posthtml.js.flow
+++ b/flow-libs/posthtml.js.flow
@@ -14,11 +14,22 @@ declare module 'posthtml' {
     content?: Array<string>
   };
 
+  declare type PostHTMLTree = Array<PostHTMLNode>;
+  declare type PostHTMLOptions = {
+    sync: Boolean,
+    skipParse: Boolean
+  };
+
   declare var walk: (fn: (node: PostHTMLNode) => PostHTMLNode) => void;
+  declare var process: (
+    tree: PostHTMLTree | string,
+    options: ?PostHTMLOptions
+  ) => Promise<{html: string, tree: PostHTMLTree}>;
 
   declare class Api {
     static walk: typeof walk;
     walk: typeof walk;
+    process: typeof process;
   }
 
   declare class PostHTML extends Api {
@@ -26,5 +37,5 @@ declare module 'posthtml' {
     version: number;
   }
 
-  declare module.exports: () => PostHTML;
+  declare module.exports: (plugins: Array<Function>) => PostHTML;
 }
--- a/flow-libs/posthtml.js.flow
+++ b/flow-libs/posthtml.js.flow
@@ -19,12 +19,20 @@ declare module 'posthtml' {
     sync: Boolean,
     skipParse: Boolean
   };
+  declare type PostHTMLMessage = {
+    type: 'dependency',
+    file: string
+  };
 
   declare var walk: (fn: (node: PostHTMLNode) => PostHTMLNode) => void;
   declare var process: (
     tree: PostHTMLTree | string,
     options: ?PostHTMLOptions
-  ) => Promise<{html: string, tree: PostHTMLTree}>;
+  ) => Promise<{
+    html: string,
+    tree: PostHTMLTree,
+    messages: Array<PostHTMLMessage>
+  }>;
 
   declare class Api {
     static walk: typeof walk;
--- a/flow-libs/posthtml.js.flow
+++ b/flow-libs/posthtml.js.flow
@@ -37,5 +37,5 @@ declare module 'posthtml' {
     version: number;
   }
 
-  declare module.exports: (plugins: Array<Function>) => PostHTML;
+  declare module.exports: (plugins: ?Array<Function>) => PostHTML;
 }
--- a/packages/configs/default/index.json
+++ b/packages/configs/default/index.json
@@ -7,7 +7,10 @@
       "@parcel/transformer-terser"
     ],
     "*.css": ["@parcel/transformer-css", "@parcel/transformer-postcss"],
-    "*.{htm,html}": ["@parcel/transformer-html"],
+    "*.{htm,html}": [
+      "@parcel/transformer-html",
+      "@parcel/transformer-posthtml"
+    ],
     "*": ["@parcel/transformer-raw"]
   },
   "namers": ["@parcel/namer-default"],
--- a/packages/configs/default/index.json
+++ b/packages/configs/default/index.json
@@ -8,8 +8,8 @@
     ],
     "*.css": ["@parcel/transformer-css", "@parcel/transformer-postcss"],
     "*.{htm,html}": [
-      "@parcel/transformer-html",
-      "@parcel/transformer-posthtml"
+      "@parcel/transformer-posthtml",
+      "@parcel/transformer-html"
     ],
     "*": ["@parcel/transformer-raw"]
   },
--- a/packages/configs/default/index.json
+++ b/packages/configs/default/index.json
@@ -6,7 +6,7 @@
       "@parcel/transformer-js",
       "@parcel/transformer-terser"
     ],
-    "*.css": ["@parcel/transformer-postcss", "@parcel/transformer-css"],
+    "*.css": ["@parcel/transformer-css", "@parcel/transformer-postcss"],
     "*.{htm,html}": [
       "@parcel/transformer-posthtml",
       "@parcel/transformer-html"
--- a/packages/configs/default/index.json
+++ b/packages/configs/default/index.json
@@ -6,7 +6,7 @@
       "@parcel/transformer-js",
       "@parcel/transformer-terser"
     ],
-    "*.css": ["@parcel/transformer-css", "@parcel/transformer-postcss"],
+    "*.css": ["@parcel/transformer-postcss", "@parcel/transformer-css"],
     "*.{htm,html}": [
       "@parcel/transformer-posthtml",
       "@parcel/transformer-html"
--- a/packages/configs/default/package.json
+++ b/packages/configs/default/package.json
@@ -12,6 +12,7 @@
   "dependencies": {
     "@parcel/bundler-default": "^2.0.0-alpha.0",
     "@parcel/namer-default": "^2.0.0-alpha.0",
+    "@parcel/optimizer-cssnano": "^2.0.0-alpha.0",
     "@parcel/packager-css": "^2.0.0-alpha.0",
     "@parcel/packager-html": "^2.0.0-alpha.0",
     "@parcel/packager-js": "^2.0.0-alpha.0",
--- a/packages/configs/default/package.json
+++ b/packages/configs/default/package.json
@@ -24,6 +24,7 @@
     "@parcel/transformer-html": "^2.0.0-alpha.0",
     "@parcel/transformer-js": "^2.0.0-alpha.0",
     "@parcel/transformer-postcss": "^2.0.0-alpha.0",
+    "@parcel/transformer-posthtml": "^2.0.0-alpha.0",
     "@parcel/transformer-raw": "^2.0.0-alpha.0",
     "@parcel/transformer-terser": "^2.0.0-alpha.0",
     "@parcel/reporter-dev-server": "^2.0.0-alpha.0",
--- a/packages/core/integration-tests/test/posthtml.js
+++ b/packages/core/integration-tests/test/posthtml.js
@@ -3,7 +3,6 @@ const fs = require('@parcel/fs');
 const {
   bundle,
   assertBundles,
-  assertBundleTree,
   removeDistDirectory,
   distDir
 } = require('@parcel/test-utils');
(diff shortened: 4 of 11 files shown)


wbinnssmith:
This is so awesome!

packages/transformers/posthtml/.eslintrc.json
```diff
@@ -0,0 +1,6 @@
+{
```
wbinnssmith: I don't think we need this file. Was this just copied from another transformer? We should probably remove it there too at some point if so.

packages/transformers/posthtml/src/PostHTMLTransformer.js
```diff
@@ -0,0 +1,74 @@
...
+    const depConfig = {
+      addDependencyTo: {
+        addDependency: name =>
+          asset.addDependency(name, {includedInParent: true})
+      }
+    };
+    Object.keys(plugins).forEach(p => Object.assign(plugins[p], depConfig));
+  }
+  // TODO: find a way to load plugins
+  // config.plugins = await loadPlugins(plugins, asset.name);
```
wbinnssmith: hmm not sure if we should try to make https://github.com/parcel-bundler/parcel/blob/425885992156b6b3e3ca02b3ffd1693fff6cbba5/packages/transformers/postcss/src/loadPlugins.js#L6 more generic or not

packages/transformers/posthtml/src/PostHTMLTransformer.js
```diff
@@ -0,0 +1,74 @@
...
+import semver from 'semver';
+
+const getPostHTMLConfig = async asset => {
+  let config = await asset.getConfig(
+    ['.posthtmlrc', '.posthtmlrc.js', 'posthtml.config.js'],
+    {
+      packageKey: 'posthtml'
+    }
+  );
+  if (!config && !asset.options.minify) {
```
wbinnssmith: We should move minifying into an optimizer (this can definitely be another PR), so can just remove these lines here.

An aside: is `options` a thing on asset? I think we should use the parcel `options` passed into all the transformer methods. But again minify stuff should be in an optimizer.

wbinnssmith:
Thanks so much for this 😃I just merged #3062 a few hours ago which will probably create conflicts for this. Resolving should be pretty straightforward but I'd be happy to take care of that if you'd like.

packages/transformers/posthtml/.eslintrc.json
kwelch: Yes, it was before I knew what I was doing 😂. I will remove it.

packages/transformers/posthtml/src/PostHTMLTransformer.js
kwelch: They were shared before but I was hesitate and didn't see postcss was converted yet, so I decided to hold off on it. However I also realized this is probably why the tests are failing. 😅

packages/transformers/posthtml/src/PostHTMLTransformer.js
kwelch: I will remove, good call. This was directly ported and I must have glazed over this part.

packages/transformers/html/.eslintrc.json
```diff
@@ -1,6 +0,0 @@
-{
```
wbinnssmith: thanks 😄

packages/transformers/posthtml/src/PostHTMLTransformer.js
```diff
@@ -0,0 +1,74 @@
...
+    return;
+  }
+
+  config = config || {};
+  const plugins = config.plugins;
+  if (typeof plugins === 'object') {
+    // This is deprecated in favor of result messages but kept for compatibility
```
wbinnssmith: I could be wrong, but it looks like we won't need this anymore as `messages` covers the same thing, see #2966 and its changed files: https://github.com/parcel-bundler/parcel/pull/2966/files#diff-2dcba7b85fdbaf96d8c8a1e09d2c940eL39

cc @DeMoorJasper

DeMoorJasper: Yeah would be great to get this deprecated feature out in Parcel 2

kwelch: I will work to find a way to remove this. Wasn't care enough in to understand it when I grabbed that learning more and more about this now.

kwelch: Looks like the solution, is to just remove it 😅simple enough.

kwelch:
Been playing with this for a while I think it is close, but for some odd reason the plugin include is not finding a match in the tree. Dinner time so I am going to do the whole walk away and ✨everything works when I look again. 🤞

packages/transformers/posthtml/src/PostHTMLTransformer.js
```diff
@@ -0,0 +1,63 @@
...
+  async getConfig({asset}) {
+    return getPostHTMLConfig(asset);
+  },
+
+  canReuseAST({ast}) {
+    return ast.type === 'posthtml' && semver.satisfies(ast.version, '^0.4.0');
+  },
+
+  async parse({asset, config}) {
+    return {
```
devongovett: we should skip parsing altogether if there is no config.

packages/transformers/posthtml/src/PostHTMLTransformer.js
```diff
@@ -0,0 +1,63 @@
...
+
+  async transform({asset, config}) {
+    if (!config) {
+      return [asset];
+    }
+
+    let res = await posthtml(config.plugins).process(asset.ast, config);
+
+    asset.ast = res.tree;
+    asset.isAstDirty = true;
```
devongovett: I don't think you need this. Was an optimization when PostHTML was part of the HTML transform, but as a separate transformer it will already call generate when needed.

packages/configs/default/index.json
```diff
@@ -7,7 +7,10 @@
       "@parcel/transformer-terser"
     ],
     "*.css": ["@parcel/transformer-css"],
-    "*.{htm,html}": ["@parcel/transformer-html"],
+    "*.{htm,html}": [
+      "@parcel/transformer-posthtml",
```
kwelch: I think the more aggressive thing should be first, let me know if that is not right

packages/transformers/html/src/HTMLTransformer.js
```diff
@@ -11,13 +11,13 @@ import extractInlineAssets from './inline';
 
 export default new Transformer({
   canReuseAST({ast}) {
-    return ast.type === 'posthtml' && semver.satisfies(ast.version, '^0.4.0');
+    return ast.type === 'posthtml' && semver.satisfies(ast.version, '^0.11.3');
```
kwelch: The version used before was the `parser` version, but when we parse it gives us back the version of `posthtml`, so I am updating these to this.

packages/transformers/posthtml/src/PostHTMLTransformer.js
kwelch: Agreed!

packages/core/integration-tests/test/posthtml.js
```diff
@@ -55,7 +59,7 @@ describe.skip('posthtml', function() {
     assert(asset.dependencies.get(other).includedInParent);
```
kwelch: I wasn't happy to skip these, but I wasn't sure the new way to do this.

packages/transformers/posthtml/src/loadPlugins.js
```diff
@@ -0,0 +1,46 @@
...
+    return [];
+  }
+}
+
+async function loadPlugin(
+  pluginArg: string | Function,
+  relative: FilePath,
+  options: mixed = {}
+): mixed {
+  if (typeof pluginArg !== 'string') {
```
kwelch: I took this whole file from #3037, just had to add this line in case the config had a require statement. As @wbinnssmith mentioned, we could find a common place for this to live. Until then I duplicated it.

flow-libs/posthtml.js.flow
```diff
@@ -14,17 +14,28 @@ declare module 'posthtml' {
     content?: Array<string>
   };
 
+  declare type PostHTMLTree = Array<PostHTMLNode>;
```
wbinnssmith: 🤩

packages/configs/default/index.json
wbinnssmith: Yeah, I think I got this backwards in the PostCSS transform

packages/transformers/html/src/HTMLTransformer.js
wbinnssmith: does the parser result actually include version information? Last I remember the posthtml ast is basically unversioned (I don't think they've made any breaking changes so far either).

Personally it doesn't matter to me as long as we commit to something and make sure third party plugins follow it.

kwelch: I looked into it and I didn't see a version returned from `posthtml` or the `parser` so I reverted back to using the parser version, as it felt more right.

packages/configs/default/index.json
devongovett: Yeah posthtml/postcss should be first, in case they introduce new dependencies.

packages/configs/default/package.json
```diff
@@ -22,6 +22,7 @@
     "@parcel/transformer-babel": "^2.0.0-alpha.0",
     "@parcel/transformer-css": "^2.0.0-alpha.0",
     "@parcel/transformer-html": "^2.0.0-alpha.0",
+    "@parcel/transformer-posthtml": "^2.0.0",
```
devongovett: Can you make this use an alpha version?

packages/transformers/posthtml/src/PostHTMLTransformer.js
```diff
@@ -0,0 +1,74 @@
...
+import {Transformer} from '@parcel/plugin';
+
+import posthtml from 'posthtml';
+import parse from 'posthtml-parser';
+import render from 'posthtml-render';
+import nullthrows from 'nullthrows';
+import semver from 'semver';
+import loadPlugins from './loadPlugins';
+
+const getPostHTMLConfig = async asset => {
```
devongovett: any reason why this function isn't inlined below?

packages/configs/default/package.json
kwelch: Is there a "fancy" way of doing that, or should I just update version, manually?

packages/configs/default/index.json
kwelch: Should I update postcss in this or separate?

packages/core/integration-tests/test/posthtml.js
devongovett: Yeah I think you'd have to look for a `connected_file` node in the asset graph, but that is not exposed by Parcel. cc. @wbinnssmith any ideas?

packages/configs/default/package.json
devongovett: I think just manually here and in the posthtml package.json

packages/transformers/posthtml/src/PostHTMLTransformer.js
kwelch: I had considered putting it in a `config.js`. I agree input would be better.

kwelch: Consolidated this.

packages/configs/default/index.json
kwelch: I moved `postcss` to be consistent.

kwelch: I ended up reverting the move of postcss, because it cause test failures. If it is suppose to be postcss first, we may have to look into why those tests were failing.

packages/core/integration-tests/test/posthtml.js
wbinnssmith: There hasn't been a real usecase for it yet, but I'm okay exposing `getConnectedFiles()` on the public Asset object.

kwelch: Nice! I will look into exposing that in a different PR. It will be good exposure into areas outside of the transformer.

kwelch: I did update the posthtml transformer to include dependencies so that `getConnectedFile` can be eventually used.

packages/transformers/posthtml/src/PostHTMLTransformer.js
```diff
@@ -0,0 +1,81 @@
...
+    if (!config) {
+      return [asset];
+    }
+
+    const ast = nullthrows(asset.ast);
+
+    let res = await posthtml(config.plugins).process(ast.program, config);
+
+    if (res.messages) {
+      await Promise.all(
```
devongovett: doesn't look like this needs to be async right?

kwelch: `asset.addConnectedFile` returns a promise, so I thought they needed to be awaited.

devongovett: oh you're right. I forgot about that.

PR #3064, opened 2019-05-21
Restore rich CLI

wbinnssmith:
This restores the rich ink-based CLI to parcel 2. It was disabled following issues with the devserver, but those don't seem to be present anymore.

There is one behavioral change: we now only print the bundle report in production mode.

Additionally, an unrelated error where the server would continue to serve 500 pages after build errors were resolved has been fixed.

Test Plan: 
* Run the kitchen sink, simple, and react-hmr examples and verify output and relevant exit code.
* Start the hmr server in the react-hmr example and introduce a syntax error. Verify that hmr reflects the error without reload. Verify reloading shows a 500 page. Fix the error and reload without a 500 page.

### DIFF
--- a/packages/reporters/cli/src/CLIReporter.js
+++ b/packages/reporters/cli/src/CLIReporter.js
@@ -1,19 +1,149 @@
 // @flow strict-local
 
-import nullthrows from 'nullthrows';
+import type {
+  BundleGraph,
+  LogEvent,
+  ParcelOptions,
+  ProgressLogEvent,
+  ReporterEvent
+} from '@parcel/types';
+import type {RerenderFunc} from 'ink';
+
+import {prettifyTime} from '@parcel/utils';
+import {render} from 'ink';
 import {Reporter} from '@parcel/plugin';
 import React from 'react';
-import {render} from 'ink';
+
+import {getProgressMessage} from './utils';
+import logLevels from './logLevels';
 import UI from './UI';
 
-let ui: ?UI;
+type State = {|
+  progress: ?ProgressLogEvent,
+  logs: Array<LogEvent>,
+  bundleGraph: ?BundleGraph
+|};
+
+let state: State = {
+  progress: null,
+  logs: [],
+  bundleGraph: null
+};
+let rerender: RerenderFunc;
 
 export default new Reporter({
   report(event, options) {
-    if (!ui) {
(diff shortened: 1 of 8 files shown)


wbinnssmith:
Just pushed an update which lifts state outside of the `UI` component so that it is truly more ui-centric. This removes the need for the imperative `report` api on the UI component, relying on `ref` getting called synchronously during render, and moves it to more of a redux-style global state + reducer.

Let me know what you think @devongovett

packages/reporters/cli/src/CLIReporter.js
```diff
@@ -1,16 +1,142 @@
...
 
-let ui: ?UI;
-render(<UI ref={u => (ui = u)} />);
+type State = {|
+  progress: ?ProgressLogEvent,
+  logs: Array<LogEvent>,
+  bundleGraph: ?BundleGraph
+|};
+
+let state: State = {
```
wbinnssmith: this exposes the true global nature of this state (it was before through the global `UI` object`).

Maybe we should allow plugins to export factory functions that return disposables instead? Or even just a class reference? cc @jamiebuilds @jamiekyle-eb on the RFC

packages/reporters/cli/src/CLIReporter.js
```diff
@@ -1,16 +1,149 @@
...
+};
+let rerender: RerenderFunc;
 
 export default new Reporter({
   report(event, options) {
-    nullthrows(ui).report(event, options);
+    let newState = reducer(state, event, options);
+    let uiElement = <UI {...newState} options={options} />;
+    if (rerender) {
+      rerender(uiElement);
```
wbinnssmith: maybe we should just use redux to avoid all this? Instead of re-rendering at the top level each time, we'd render once and dispatch actions with these updates.

devongovett: seems like so much for one component...

packages/reporters/cli/src/CLIReporter.js
```diff
@@ -1,16 +1,149 @@
...
+    } else {
+      render(uiElement);
+    }
   }
 });
+
+function reducer(
```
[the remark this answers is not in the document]
wbinnssmith: hmm, even though `dispatch` should be stable, I'm not sure it's a good idea to expose it outside the render function

PR #3065, opened 2019-05-21
New AssetGraph

padmaia:
# ↪️ Pull Request

While working on adding config loading to the graph, which necessitated adding 4 more types to the graph, all of which no longer matter once we hit the bundling phase, I came up with the idea of building multiple graphs at the same time. One graph will still be called the `AssetGraph`, but it will only have nodes needed for Bundling onward (Assets and Dependencies). The other graph will be named the `RequestGraph` which keeps track of all of the requests that give us results that make up the `AssetGraph`. Eventually they will be built along side each other (in progress), but for now I am just converting the` OldAssetGraph` to the new `AssetGraph`, so that we can start working with the new `AssetGraph` as soon as possible, while I'm working on the multi graph approach.

## 🚨 Test instructions

Run examples with `PARCEL_DUMP_GRAPHVIZ=1` and make sure graphs make sense

### DIFF
--- a/packages/core/core/src/AssetGraphBuilder.js
+++ b/packages/core/core/src/AssetGraphBuilder.js
@@ -16,10 +16,13 @@ import {
   type AbortSignal
 } from 'abortcontroller-polyfill/dist/cjs-ponyfill';
 import {PromiseQueue} from '@parcel/utils';
+import OldAssetGraph from './OldAssetGraph';
 import AssetGraph from './AssetGraph';
 import ResolverRunner from './ResolverRunner';
 import WorkerFarm from '@parcel/workers';
 import type {Event} from '@parcel/watcher';
+
+import dumpToGraphViz from './dumpGraphToGraphViz';
 
 type BuildOpts = {|
   signal: AbortSignal,
@@ -35,7 +38,7 @@ type Opts = {|
 |};
 
 export default class AssetGraphBuilder extends EventEmitter {
-  graph: AssetGraph;
+  graph: OldAssetGraph;
   queue: PromiseQueue;
   resolverRunner: ResolverRunner;
   controller: AbortController;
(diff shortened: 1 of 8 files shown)


packages/core/core/src/dumpGraphToGraphViz.js
```diff
@@ -75,28 +69,22 @@ export default async function dumpGraphToGraphViz(
...
     } else {
       // label += node.id;
       label = node.type;
     }
-
     n.set('label', label);
   }
-
-  for (let edge of graph.edges) {
+  for (let edge of graph.getAllEdges()) {
```
wbinnssmith: thanks for doing this 😃

packages/core/types/index.js
```diff
@@ -227,6 +227,11 @@ export type TransformerRequest = {
   code?: string
 };
 
+export type AssetGroup = {
```
wbinnssmith: can this be an exact type?

packages/core/core/src/Graph.js
```diff
@@ -115,17 +115,17 @@ export default class Graph<TNode: Node> {
   removeNode(node: TNode): this {
     let removed = new this.constructor();
 
+    for (let from of this.inboundEdges.get(node.id)) {
```
wbinnssmith: yeah this makes a lot more sense. we should probably move up removing outbound edges too, since those are also edges to a node that's no longer in the graph.

packages/core/core/src/AssetGraphBuilder.js
```diff
@@ -35,7 +38,7 @@ type Opts = {|
 |};
 
 export default class AssetGraphBuilder extends EventEmitter {
-  graph: AssetGraph;
+  graph: OldAssetGraph;
```
wbinnssmith: Will `OldAssetGraph` eventually be deleted and `AssetGraphBuilder` will build two newer graphs -- one of them called `AssetGraph` but will be the newer, smaller one?

padmaia: Yes, that's the plan.

packages/core/types/index.js
padmaia: Yeah, good catch.

PR #3068, opened 2019-05-22
Added new info command 

jamesgeorge007:
# ↪️ Pull Request

Added a new `info` command which prints out information about the local environment (proves to be useful while submitting bug reports).

## 💻 Examples

`parcel info`

```bash 
Environment Info:

  System:
    OS: Linux 4.18 Ubuntu 18.10 (Cosmic Cuttlefish)
    CPU: (4) x64 Intel(R) Core(TM) i5-7200U CPU @ 2.50GHz
  Binaries:
    Node: 8.11.4 - /usr/bin/node
    Yarn: 1.17.0-0 - /usr/local/bin/yarn
    npm: 6.9.0 - /usr/local/bin/npm
  Browsers:
    Chrome: 74.0.3729.157
    Firefox: 66.0.5
  npmGlobalPackages:
    parcel-bundler: 1.12.3
```

## ✔️ PR Todo

### DIFF
--- a/packages/core/parcel-bundler/package.json
+++ b/packages/core/parcel-bundler/package.json
@@ -47,6 +47,7 @@
     "deasync": "^0.1.14",
     "dotenv": "^5.0.0",
     "dotenv-expand": "^4.2.0",
+    "envinfo": "^7.3.1",
     "fast-glob": "^2.2.2",
     "filesize": "^3.6.0",
     "get-port": "^3.2.0",
--- a/packages/core/parcel-bundler/src/cli.js
+++ b/packages/core/parcel-bundler/src/cli.js
@@ -1,5 +1,6 @@
 require('v8-compile-cache');
 const chalk = require('chalk');
+const envinfo = require('envinfo');
 const program = require('commander');
 const version = require('../package.json').version;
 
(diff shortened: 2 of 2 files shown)


jamesgeorge007:
@DeMoorJasper Thoughts :thinking:

mischnic:
One interesting info would be if that parcel instance is actually the globally installed one

jamesgeorge007:
> One interesting info would be if that parcel instance is actually the globally installed one

```bash
npmGlobalPackages:
    parcel-bundler: 1.12.3
```
It returns the globally installed `parcel-bundler` version.

DeMoorJasper:
@jamesgeorge007 but it does not return whether parcel is being used globally or locally or does it?

Global is globally installed local is executed from package.json

jamesgeorge007:
@DeMoorJasper  It does return the globally installed version as per now.

DeMoorJasper:
@jamesgeorge007 parcel can also be ran locally. Sometimes running it globally causes issues, so it would be helpfull to know whether the parcel instance is the global or local one.

Local being in npm_modules and global being global...

You currently do not print information about the local Parcel version nor whether it's actually using the global one.

For example if you use the scripts section in pkg.json for running parcel it will use the local version regardless of whether there is a global version installed. (So your output would give a false image as the local one could be a different version)

jamesgeorge007:
@DeMoorJasper What causes the CI build to fail :thinking:

jamesgeorge007:
@DeMoorJasper Made it such that both the local and global instances show up. Makes sense :thinking:

DeMoorJasper:
@jamesgeorge007 CI is just unstable, sometimes it's Azure DevOps being unstable, but usually it's just some command failing. We're not sure why though

PR #3070, opened 2019-05-22
Pass more css integration tests

wbinnssmith:
This passes all but two css tests, one which requires a source map comment, and another which requires `@import` nesting.

For consistency with existing tests and Parcel 1, this also makes bundle url imports relative to the publicUrl rather than relative to the current bundle. While either is probably fine, and the former is arguably more portable independent of the publicUrl, I think the former is more common on the web and expected by developers.

As discussed in #3044, and to unblock some tests, this also changes raw assets to use a fallback `*` glob, rather than enumerating a bunch of common extensions.

Test Plan: yarn test

### DIFF
--- a/packages/core/core/src/PackagerRunner.js
+++ b/packages/core/core/src/PackagerRunner.js
@@ -8,7 +8,7 @@ import type InternalBundleGraph from './BundleGraph';
 
 import invariant from 'assert';
 import {mkdirp, writeFile, writeFileStream} from '@parcel/fs';
-import {TapStream, urlRelative} from '@parcel/utils';
+import {TapStream, urlJoin} from '@parcel/utils';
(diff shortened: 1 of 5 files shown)


packages/core/core/src/PackagerRunner.js
```diff
@@ -150,7 +150,10 @@ function generateDepToBundlePath(
     let entryBundle = entryBundleNode.value;
     depToBundlePath.set(
       dep.id,
-      urlRelative(nullthrows(bundle.name), nullthrows(entryBundle.name))
+      urlJoin(
```
wbinnssmith: one thing we could try is using a url relative to the current bundle only in the case that `publicUrl` wasn't provided by the user, and if it was, use a url relative to that. What do you think?

PR #3071, opened 2019-05-23
Restore @parcel/register

wbinnssmith:
In preparation for the release of Parcel 2 and its future development, this updates @parcel/register to work with the latest changes.

Since we want to keep the core Parcel api as minimal as possible, but need access to internals to run single transformations and single resolutions, we take the following approach:

* Implement the bulk of parcel-register in `@parcel/core`. This allows us to freely import symbols which allow us to invoke internal methods on the core `Parcel` object. Once Parcel 2 packages are built into single files, these exports will not leave the flat file, but can be freely required within it. This is exposed as a named `register` export from core.
* Implement a package named `@parcel/register` which imports `register` from core and invokes it with a config.

Other notes: 
* The core `register` function returns a disposable which undoes the effects of `register`. It returns resolution back to Node directly, and stops transformation. In the future, this should also shut down the workers.
* Moved from monkeypatching `require` to monkey patching `Module._resolveFilename` [0]. At the risk of overriding internals which have not changed in the past seven years [0], we gain the ability to use this same logic to power `require.resolve`
* Adds integration tests to cover resolving, transforming, and disposing behavior. These work with a pre-built parcel but are disabled until there is a published version.
* `@parcel/register` should ultimately depend on an already-published version of Parcel in order to run it over Parcel itself. Since it depends on Parcel, one must by definition exist to use it. See the whole compiler bootstrapping problem [1]

Test Plan: Since `@parcel/require` uses parcel to transform and resolve, it requires a pre-built version of Parcel and cannot be tested without this (I developed one to test this locally). I'll be publishing a branch with a simple build process in order to get an initial version of Parcel 2 published so we can use it in `@parcel/require` to develop Parcel 2.

[0] https://github.com/nodejs/node-v0.x-archive/issues/1125#issuecomment-10748203
[1] https://en.wikipedia.org/wiki/Bootstrapping_(compilers)


mischnic:
> @parcel/register should ultimately depend on an already-published version of Parcel in order to run it over Parcel itself. Since it depends on Parcel, one must by definition exist to use it. See the whole compiler bootstrapping problem [1]

This always feels like building a house of cards to me. I think it should always be possible to compile Parcel with Parcel running via @babel/register.

wbinnssmith:
> I think it should always be possible to compile with Parcel running via @babel/register.

I agree -- though we had discussed moving to Parcel's `source`field in package.json to point to our sources (which we'd want babel to run over), and leaving `main` for compiled code run directly in node. @devongovett I'd be interested to hear your thoughts on this.

mischnic:
> I agree -- though we had discussed moving to Parcel's `source` field in package.json to point to our sources (which we'd want babel to run over), and leaving `main` for compiled code run directly in node.

Yes, we do need a compile step

> We also make assumptions that Parcel packages will be built into single files via Parcel itself -- though those assumptions are mostly re: not exposing internals. If the versions we actually publish are built through Parcel then we'll be fine.

I was thinking:
In the packages we run `"build": "node -r @parcel/babel-register ../parcel/cli.js src/index.js -o lib.js"` (or something similar to essentially run parcel via an interpreter to build itself - to stay in the compiler-bootstrapping terminology).
This seems so easy to do that creating a boostrap compiler which only includes the JS asset/packager to then compile the rest is unnecessary (I can't think of a reason).

This would also prevent a situation where the old version of Parcel can compile Parcel but the version that is currently being built won't be able to do that anymore when it becomes the old version.

Side note: Is there anything special about these packages that requires them to be built with Parcel?

packages/core/integration-tests/test/parcel-register.js
```diff
@@ -0,0 +1,109 @@
+// @flow strict-local
+
+import {execSync} from 'child_process';
+import assert from 'assert';
+import path from 'path';
+
+describe.skip('@parcel/register', () => {
```
devongovett: why are they skipped?

wbinnssmith: From the test plan:

> Since @parcel/require uses parcel to transform and resolve, it requires a pre-built version of Parcel and cannot be tested without this (I developed one to test this locally).

Parcel register requires parcel, and I wasn't daring enough to try to combine `@babel/register` and `@parcel/register`.

PR #3072, opened 2019-05-23
Rename package versions to 2.0.0-alpha.0

wbinnssmith:
Updates:
* The version field in parcel 2 `package.json`s
* The required semver fields in `dependencies` and `devDependencies`
* The semver `engines` field for `parcel`

This is so we can begin publishing very early builds of Parcel to get the bootstrapping process started.

I'm definitely open to changing the prerelease name, ~~but we should avoid `alpha` since we plan on using that for more official releases~~ Let's try going with `alpha.0` (before our more public `alpha.1`), and we can publish canary releases with Lerna as `alpha.0+$SHA`.

Test Plan: 
* Run `yarn` and verify the top-level and package-level `node_modules/@parcel` have only symlinks
* Flow, lint, test

### DIFF
--- a/packages/bundlers/default/package.json
+++ b/packages/bundlers/default/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@parcel/bundler-default",
-  "version": "2.0.0-prealpha0",
+  "version": "2.0.0-alpha.0",
   "license": "MIT",
   "repository": {
     "type": "git",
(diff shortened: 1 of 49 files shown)


wbinnssmith:
Lerna has a neat canary [0] feature which we could leverage for nightly or regular releases. We could also adopt its convention of -alpha0+$SHA or more simply -alpha0+0, -alpha0+1, etc.

[0] https://github.com/lerna/lerna/tree/master/commands/publish#--canary

PR #3075, opened 2019-05-23
Adds Documentation about v2 Transformers

kwelch:
# ↪️ Pull Request
Notes from my learnings on how the transformers work and their lifecycle. 

## ✔️ PR Todo

### DIFF
--- a/docs/Transformers.md
+++ b/docs/Transformers.md
@@ -8,6 +8,8 @@ Transformers are ran in order by file type. It is recommended to have more speci
 
 ## Transformer Life Cycle
 
+Below is the list of the life cycle methods in the order they are called.
+
 ### `getConfig`
 
 Loads configuration specifically for this transformer.
--- a/docs/Transformers.md
+++ b/docs/Transformers.md
@@ -1,8 +1,38 @@
 # Transformers
 
-A transformer allows you to access and perform transformative operations on an asset.
+Transformers have three roles: to transform an assets content from one form to another, to create additional assets, and to register dependencies and connected files.
 
-Transformers are defined in: `./packages/configs/default/index.json`
+### Transform
+
+TODO: Put a blurb here
+
+**Common Examples**
+
+- Babel: Transforms latest JavaScript to browser compatible JavaScript
+- SASS: Compiles sass files into css files
+
+### Create Assets
+
+TODO: Put a blurb here
+
+**Common Examples**
+
+- Babel: Transforms latest JavaScript to browser compatible JavaScript
+- SASS: Compiles sass files into css files
+
+### Register Dependencies
+
+TODO: Put a blurb here
+
+**Common Examples**
+
+- HTML: Finds and adds linked files into the tree
+- JS: Reads imports and require statements to find dependencies
+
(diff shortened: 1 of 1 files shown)


docs/Transformers.md
```diff
@@ -0,0 +1,77 @@
...
+Below is the list of the life cycle methods in the order they are called.
+
+### `getConfig`
+
+Loads configuration specifically for this transformer.
+
+**API**
+
+```
+getConfig ({
```
wbinnssmith: In the long run we should probably be extracting API docs like this from the source

docs/Transformers.md
```diff
@@ -0,0 +1,77 @@
+# Transformers
+
+A transformer allows you to access and perform transformative operations on an asset.
```
wbinnssmith: Just a thought. Maybe:

> Transformers have three roles: to transform an assets content from one form to another, to create additional assets, and to register dependencies and connected files.

And then describe each of those uses.

docs/Transformers.md
```diff
@@ -0,0 +1,77 @@
...
+
+Transformers are defined in: `./packages/configs/default/index.json`
```
wbinnssmith: > Transformers are defined in the "transformers" key of configuration. For the default configuration, see 
 `./packages/configs/default/index.json`

docs/Transformers.md
```diff
@@ -0,0 +1,77 @@
...
+
+Transformers are ran in order by file type. It is recommended to have more specialized transformers first, and have them skip assets that don't match their case.
```
kwelch: I am not sure how true this is. I believe it is true based on my limited experience, but I am not confident in this.

docs/Transformers.md
```diff
@@ -0,0 +1,108 @@
+# Transformers
+
+Transformers have three roles: to transform an assets content from one form to another, to create additional assets, and to register dependencies and connected files.
+
+### Transform / Compile
+
+Covertes the assets content from one structure to another.
```
jamiekyle-eb: `Covertes`

docs/Transformers.md
```diff
@@ -0,0 +1,108 @@
...
+
+### Create Assets
+
+Dynamic create new assets based on the assests provided.
+
+**Common Examples**
+
+- Babel: Transforms latest JavaScript to browser compatible JavaScript
+- SASS: Compiles sass files into css files
+- CSSModules: Produces JavaScript files to be imported
```
jamiekyle-eb: I'm not entirely sure how this is meaningfully different than "Transform / Compile"

docs/Transformers.md
kwelch: Yea, this is not my strong spot.

docs/Transformers.md
kwelch: I agree this is an example the cross over into both. I put it here because it produced a new file in addition to transpiling. 

I am open to changed on this. Just went with that I was aware of.

DeMoorJasper:
Gonna close this as it's been sitting idle for quite some time, feel free to reopen

PR #3079, opened 2019-05-24
Introduce RequestGraph to help build skinny AssetGraph

padmaia:
# ↪️ Pull Request

The AssetGraphBuilder now builds two graphs in parallel, the "skinny" `AssetGraph` that now only has Dependencies and Assets (and AssetGroups), and the new `RequestGraph` that keeps track of the state of async requests and invalidates them when necessary. This is beneficial because the `AssetGraph` will only have nodes that the Bundling and Packaging phases care about. We can build it incrementally instead of stripping away unimportant nodes after each build, or keeping unimportant nodes and skipping over them while still sending them over IPC. It also makes the `AssetGraph` much easier to read, especially once I start adding more request nodes for config invalidation.   

## 🚨 Test instructions

## ✔️ PR Todo

- [x] Clean up `RequestGraph` nodes, when nodes are removed from `AssetGraph`


packages/core/core/test/AssetGraphBuilder.test.js
```diff
@@ -55,35 +54,12 @@ describe('AssetGraphBuilder', () => {
...
+      builder.assetGraph.nodes.has(
         new Dependency({
           moduleSpecifier: './module-b',
           env: DEFAULT_ENV
         }).id
       )
     );
   });
-
-  it('throws a BuildAbortError when resolving if signal aborts', async () => {
```
padmaia: I've removed the `AbortController` code from `AssetGraphBuilder` because we have plans to move it up to Parcel. I plan on adding that back in a separate PR.

padmaia:
Tests were passing locally 😬 I'll try to take a look when I get home

packages/core/core/src/AssetGraph.js
```diff
@@ -4,60 +4,158 @@ import invariant from 'assert';
...
-//   value: asset
-// });
+import Dependency from './Dependency';
+import Graph, {type GraphOpts} from './Graph';
+import type {CacheEntry, DependencyNode, AssetGraphNode, NodeId} from './types';
+
+type AssetGraphOpts = {|
+  onNodeAdded?: (node: AssetGraphNode) => mixed,
+  ...GraphOpts<AssetGraphNode>
+|};
```
wbinnssmith: nit: newline

packages/core/core/src/AssetGraph.js
```diff
@@ -4,60 +4,158 @@ import invariant from 'assert';
...
-//   type: 'asset',
-//   value: asset
-// });
+import Dependency from './Dependency';
+import Graph, {type GraphOpts} from './Graph';
+import type {CacheEntry, DependencyNode, AssetGraphNode, NodeId} from './types';
+
+type AssetGraphOpts = {|
+  onNodeAdded?: (node: AssetGraphNode) => mixed,
+  ...GraphOpts<AssetGraphNode>
```
wbinnssmith: nit:  `...` is usually first in these cases for overriding properties. but it behaves the same in this case ¯\\\_(ツ)\_/¯

packages/core/core/src/AssetGraph.js
```diff
@@ -72,7 +170,7 @@ export default class AssetGraph extends Graph<AssetGraphNode> {
   }
 
   getDependencyResolution(dep: IDependency): ?Asset {
-    let depNode = this.getNode(dep.id);
+    let depNode = this.nodes.get(dep.id);
```
wbinnssmith: why this change?

packages/core/core/src/AssetGraphBuilder.js
```diff
@@ -1,202 +1,113 @@
...
+
+    this.assetGraph = new AssetGraph({
+      onNodeAdded: node => this.handleNodeAddedToAssetGraph(node)
+    });
+    this.requestGraph = new RequestGraph({
       config,
-      options
+      options,
+      onAssetRequestComplete: (request, result) =>
+        this.handleCompletedAssetRequest(request, result),
```
wbinnssmith: since they're both single lines, can each of these functions just be inlined directly here?

packages/core/core/src/AssetGraphBuilder.js
```diff
@@ -1,202 +1,113 @@
...
+  handleCompletedDepPathRequest(
+    requestNode: DepPathRequestNode,
+    result: AssetRequest | null
+  ) {
+    this.assetGraph.resolveDependency(requestNode.value, result);
   }
 
   isInvalid() {
-    return this.graph.invalidNodes.size > 0;
+    return this.requestGraph.isInvalid();
```
wbinnssmith: 😃

packages/core/core/src/Parcel.js
```diff
@@ -242,7 +242,7 @@ export default class Parcel {
           this.#watchEvents.emit({error: err});
           return;
         }
-
+        // console.log('DETECTED CHANGES', events);
```
wbinnssmith: debug :)

packages/core/core/src/RequestGraph.js
```diff
@@ -0,0 +1,227 @@
...
+      return cacheEntry;
+    } catch (e) {
+      // TODO: add connectedFiles even if it failed so we can try a rebuild if those files change
+      throw e;
+    }
+  }
+
+  async resolvePath(dep: Dependency) {
+    try {
+      // console.log('RESOLVING', request);
```
wbinnssmith: debug :)

packages/core/core/src/types.js
```diff
@@ -45,26 +44,22 @@ export type DependencyNode = {|
...
-export type TransformerRequestNode = {|
-  id: string,
-  +type: 'transformer_request',
-  value: TransformerRequest
-|};
 export type AssetGroupNode = {|
   id: string,
   +type: 'asset_group',
-  value: AssetGroup
+  value: AssetRequest
```
wbinnssmith: the value for an `AssetGroupNode` is an `AssetRequest`? If so, maybe a different name or a comment explaining why?

packages/core/core/src/types.js
```diff
@@ -45,26 +44,22 @@ export type DependencyNode = {|
...
-  | FileNode
-  | RootNode
-  | TransformerRequestNode
-  | BundleGroupNode
-  | BundleReferenceNode;
+export type DepPathRequestNode = {|
+  id: string,
+  +type: 'dep_path_request',
+  value: Dependency
+|};
```
wbinnssmith: nit: newline

packages/core/integration-tests/test/json-reporter.js
```diff
@@ -19,6 +19,7 @@ describe('json reporter', () => {
     let i = 0;
     // $FlowFixMe
     console.log = function log(msg) {
+      console.log('MSG', msg);
```
wbinnssmith: debug :)

packages/core/integration-tests/test/server.js
```diff
@@ -152,7 +152,7 @@ describe('server', function() {
     subscription = await b.watch();
     await getNextBuild(b);
 
-    await sleep(10000);
+    await sleep(4000);
```
wbinnssmith: can this be removed entirely? not sure why I didn't before since awaiting `getNextBuild` should take care of this

packages/examples/simple/src/foo.js
```diff
@@ -1,3 +1,5 @@
 export default function() {
   return 'hello!';
 }
+
+//syntax\\error
```
wbinnssmith: debug :)

packages/core/core/src/AssetGraph.js
padmaia: Something was broken with it, I think because I was prepending the node type to the id, but decided to just stop doing that. Thought I put it back to the way it was, good catch.

packages/core/integration-tests/test/server.js
padmaia: looks like it :)

packages/core/core/src/RequestGraph.js
```diff
@@ -0,0 +1,227 @@
...
+      config,
+      options
+    });
+  }
+
+  async initFarm() {
+    // This expects the worker farm to already be initialized by Parcel prior to calling
+    // AssetGraphBuilder, which avoids needing to pass the options through here.
+    this.farm = await WorkerFarm.getShared();
+    this.runTransform = this.farm.mkhandle('runTransform');
```
devongovett: I wonder about having the graph being the one doing the actual resolving and transforming, vs being a data structure...

packages/core/core/src/types.js
```diff
@@ -45,26 +44,26 @@ export type DependencyNode = {|
...
-export type OldAssetGraphNode =
-  | AssetNode
-  | AssetReferenceNode
-  | DependencyNode
-  | FileNode
-  | RootNode
-  | TransformerRequestNode
-  | BundleGroupNode
-  | BundleReferenceNode;
+export type DepPathRequestNode = {|
```
devongovett: why `DepPathRequestNode` and not just `DependencyRequestNode`? what is the path part? maybe I missed something

packages/core/core/test/AssetGraph.test.js
```diff
@@ -27,41 +27,40 @@ const stats = {size: 0, time: 0};
...
+      entries: ['/path/to/index1', '/path/to/index2']
     });
 
-    assert(graph.nodes.has('/'));
+    assert(graph.nodes.has('@@root'));
     assert(
       graph.nodes.has(
         new Dependency({
-          moduleSpecifier: './index1',
+          moduleSpecifier: '/path/to/index1',
```
devongovett: why did these change to absolute?

packages/core/utils/src/PromiseQueue.js
```diff
@@ -21,8 +21,19 @@ export default class PromiseQueue {
     this._maxConcurrent = opts.maxConcurrent;
   }
 
-  add(fn: () => Promise<mixed>): void {
-    this._queue.push(fn);
+  add<T>(fn: () => Promise<T>): Promise<T> {
+    return new Promise((resolve, reject) => {
```
devongovett: what's this wrapper for?

padmaia: So that you have access to the results of the promise returned from the function added to the queue. I'm keeping track of in progress request promises so that if that if a an identical request comes in while one is already in progress, it can use the in progress promise. It's not totally needed now as identical dep and asset requests shouldn't fire at the same time, but config request and dev dep requests will bring that possibility. In my config invalidation spike branch `incompleteNodes` turned into `inProgress` promises so I just decided to bring it in now instead of changing it later.

packages/core/core/src/types.js
padmaia: I thought it helped to be more specific about what we are requesting. Instead of `DependencyRequest` and `DevDepRequest` which makes it unclear how they are different, we would have `DepPathRequest` and `DepVersionRequest` which should be a little more intuitive. It may not be the best name at the moment because right now we are getting more than just the path from the resolvers (`AssetRequestRequest`? 😛 ), but I have some thoughts on that. I think ideally the resolvers shouldn't have to know what an `AssetRequest` is. Their job should only be to return a path. The AssetGraphBuilder can take care of adding the dep's env to the follow up `AssetRequest`. As for the side effects, I was thinking that getting the `package.json` for a file about to be transformed would work well as a `ConfigRequest` once I have that piece implemented. So I changed the name in anticipation of those changes.

packages/core/core/src/RequestGraph.js
padmaia: @wbinnssmith and I were talking about making both the `RequestGraph` and `AssetGraph` have instances of a `Graph`. Mainly we were talking about not exposing low level graph methods to users of `AssetGraph`, which is less of a concern for the `RequestGraph` since it is only used by `AssetGraphBuilder`, but that would take care of any concern of having the data structure do async stuff. I like the idea, but I think we can address that in a future PR.

packages/core/core/test/AssetGraph.test.js
padmaia: I noticed that in the initialization of the `AssetGraph` we are not even instantiating the entry deps with `sourcePath`, seemingly because the entries are all resolved to absolute paths before hand so we don't need the `sourcePath` to know where to resolve from. I guess resolving entries to absolute paths is happening in the CLI though. Should we require users of the programmatic API to pass entries as absolute paths?

wbinnssmith: > Should we require users of the programmatic API to pass entries as absolute paths?

I think we should follow what we're doing with `cacheDir` and target dist information passed through the public api: just do what Node does and resolve things relative to pwd.

Of course, just as you'd do with `fs.readFile`, etc., you're far better off being certain and passing an absolute path.

wbinnssmith: FWIW, while exploring options overriding node module resolution in #3071, I found that when you run `node` with an entrypoint, it invokes the resolver with a `null` sourcePath equivalent and an absolute path for the `moduleSpecifier` equivalent.

PR #3085, opened 2019-05-24
Scopehoist optimization v2

mischnic:
**With this change, packaging yarnpkg/yarn wasn't any faster. Treeshaking was *signifcantly* faster, but the mangling's `getBinding` took longer, negating any improvement.
Investigating...**

The optimization I talked about a few days ago.

- Keep track of all changes to the global program during linking so that treeshaking doesn't need to crawl.

I someone wants to test the performance, this PR targeted at V1 is also here: https://github.com/parcel-bundler/parcel/tree/scopehoist-optimization
(I noticed a 3x packaging performance increase for a 10MB bundle).


mischnic:
One remaining issue:
Running `treeShake` over this snippet (directly after parsing it) doesn't remove the comment.
```
// ASSET: .../parcel/packages/core/integration-tests/test/integration/scope-hoisting/es6/re-export-all/c.js
var $c92d2fcff5d8673ce5ee0086d7c8ef5e$exports = {};
var $c92d2fcff5d8673ce5ee0086d7c8ef5e$export$foo = 2,
    $c92d2fcff5d8673ce5ee0086d7c8ef5e$export$bar = 3;
output = $c92d2fcff5d8673ce5ee0086d7c8ef5e$export$foo + $c92d2fcff5d8673ce5ee0086d7c8ef5e$export$bar;
```

In the test `supports re-exporting all exports from another module` is it removed though.

mischnic:
Babel itself seems to only keep comment on node removal if that node is the first or last node of the parent container (https://github.com/babel/babel/blob/master/packages/babel-traverse/src/path/comments.js)

Not yet sure how to solve this....

devongovett:
@mischnic what's the status of this? I see a WIP label.

mischnic:
I want to redo some benchmarks now that terser is an optimizer and as I wrote in the description, in one case, the shaking itself was faster but then mangling was slower (no idea why some calls to babel took longer because of my changes)

PR #3089, opened 2019-05-26
Implement basic json transformer

wbinnssmith:
This implements a basic JSON transformer. It currently turns JSON the asset into a JavaScript one, which could become problematic should other assets need to import JSON. However, this is what Parcel 1 does, and it addresses our need to build Parcel 2 with Parcel 2.

Test Plan: Restored one integration test.

### DIFF
--- a/packages/transformers/json/src/JSONTransformer.js
+++ b/packages/transformers/json/src/JSONTransformer.js
@@ -1,0 +1,11 @@
+// @flow
+
+import {Transformer} from '@parcel/plugin';
+
+export default new Transformer({
+  async transform({asset}) {
+    asset.type = 'js';
+    asset.setCode(`module.exports = ${await asset.getCode()};`);
+    return [asset];
+  }
+});
--- a/packages/transformers/json/src/JSONTransformer.js
+++ b/packages/transformers/json/src/JSONTransformer.js
@@ -1,11 +1,18 @@
 // @flow
 
 import {Transformer} from '@parcel/plugin';
+import json5 from 'json5';
 
 export default new Transformer({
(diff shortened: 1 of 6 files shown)


packages/transformers/json/src/JSONTransformer.js
```diff
@@ -0,0 +1,11 @@
+// @flow
+
+import {Transformer} from '@parcel/plugin';
+
+export default new Transformer({
+  async transform({asset}) {
+    asset.type = 'js';
+    asset.setCode(`module.exports = ${await asset.getCode()};`);
+    return [asset];
```
devongovett: Ideally we’d also support parsing and re-stringifying json5 as well like parcel 1, but that could be added later.

wbinnssmith: Should json5 be part of the default JSON transformer? Would be convenient, but it would be different than, for example, how Node loads json (which would throw if it encountered anything nonstandard).

If so, it should be a pretty straightforward here :)

devongovett: JSON5 is pretty widely adopted by various tools at this point...

wbinnssmith: Sorry, what I meant was "should we load all json files through json5?" A lot of tools these days at the very least ignore comments, etc. in JSON despite them being named `.json`. This would be different behavior from Parcel 1, though.

devongovett: Yeah that's what json5 is right? Allowing comments etc. In Parcel 1 we do allow them in `.json` files

PR #3092, opened 2019-05-27
Use fsStream.bytesWritten

mischnic:
Use https://nodejs.org/api/fs.html#fs_writestream_byteswritten instead of homebrew TapStream in `writeFileStream`

The equivalent part in TransfomerRunner can't be converted so easily:
https://github.com/parcel-bundler/parcel/blob/ef9b9772eea6504520ba8b3fbf3fbee1536c123d/packages/core/core/src/TransformerRunner.js#L341-L354

### DIFF
--- a/packages/core/fs/src/index.js
+++ b/packages/core/fs/src/index.js
@@ -71,6 +71,7 @@ export function writeFileStream(
     let fsStream = fs.createWriteStream(filePath);
     stream
       .pipe(fsStream)
+      // $FlowFixMe
       .on('finish', () => resolve(fsStream.bytesWritten))
       .on('error', reject);
   });
--- a/packages/core/fs/src/index.js
+++ b/packages/core/fs/src/index.js
@@ -66,11 +66,12 @@ export const ncp: (
 export function writeFileStream(
   filePath: FilePath,
   stream: Readable
-): Promise<void> {
+): Promise<number> {
   return new Promise((resolve, reject) => {
+    let fsStream = fs.createWriteStream(filePath);
     stream
-      .pipe(fs.createWriteStream(filePath))
-      .on('finish', resolve)
+      .pipe(fsStream)
+      .on('finish', () => resolve(fsStream.bytesWritten))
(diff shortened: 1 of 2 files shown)


mischnic:
```
Cannot get `fsStream.bytesWritten` because property `bytesWritten` is missing in `WriteStream` [1].

   packages/core/fs/src/index.js:74:35
     74|       .on('finish', () => resolve(fsStream.bytesWritten))
                                           ^^^^^^^^^^^^^^^^^^^^^

References:
   /private/tmp/flow/flowlib_3ee1e628/node.js:1113:71
   1113|   declare function createWriteStream(path: string, options?: Object): WriteStream;
                                                                               ^^^^^^^^^^^ [1]
```

Ah... great, it's missing in the flow node.js:

https://github.com/facebook/flow/blob/1cff4574d1c499c574ba0f261b6b5c053e4ae4f9/lib/node.js#L864-L866

wbinnssmith:
> Ah... great, it's missing in the flow node.js:

A FlowFixMe is fine for now

packages/core/fs/src/index.js
```diff
@@ -66,11 +66,12 @@ export const ncp: (
 export function writeFileStream(
   filePath: FilePath,
   stream: Readable
-): Promise<void> {
+): Promise<number> {
```
wbinnssmith: does this really need to return a number? since we have a reference to the stream in the `PackagerRunner` (or any other case where we'd call this), we could just read this property directly after this resolves.

This is fine too, though maybe we could return something more meaningful in the future.

mischnic: > since we have a reference to the stream in the PackagerRunner (or any other case where we'd call this), we could just read this property directly after this resolves.

No, `bytesWritten` exists only on the FS.WriteStream, not on the "normal" Writeable stream. And `fs.createWriteStream(filePath))` is called in this function

wbinnssmith: ah yeah. Sounds good.

PR #3093, opened 2019-05-27
Make Cache no longer a singleton

wbinnssmith:
This makes the `Cache` object (indirectly) constructible, allowing for multiple cache objects to exist in memory at once. By not exposing the `Cache` class directly and only exposing `getCacheByDir`, we can return the same cache instance (and invalidations) for requests to the same cache directory.

This is part of an effort to decouple Workers from bundler options. Workers now no longer rely on the side effect of initializing the cache to do work. Instead, the work they do will initialize the cache if necessary.

This also fixes an unrelated test failure from cssnano being missing from the default config's `package.json`

Test Plan: yarn test


packages/core/cache/src/Cache.js
```diff
@@ -105,4 +76,35 @@ export class Cache {
...
 
-export default new Cache();
+// Cache for whether a cache dir exists
+const existsCache: Set<FilePath> = new Set();
+const cacheByDir: DefaultMap<FilePath, Cache> = new DefaultMap(dir => {
+  return new Cache(dir);
+});
+
+export async function createCacheDir(dir: FilePath): Promise<void> {
+  if (existsCache.has(dir)) {
```
wbinnssmith: This existed before, and seems questionable. 

The directory could easily be removed from underneath Parcel between runs. Since `mkdirp` is safe to rerun, and it is now parallelized below, I wonder what the perf cost would be to just run this every time Parcel is about to perform a build.

packages/core/cache/src/Cache.js
```diff
@@ -105,4 +76,35 @@ export class Cache {
   }
 }
 
-export default new Cache();
+// Cache for whether a cache dir exists
+const existsCache: Set<FilePath> = new Set();
+const cacheByDir: DefaultMap<FilePath, Cache> = new DefaultMap(dir => {
+  return new Cache(dir);
```
wbinnssmith: this `DefaultMap` has the potential to grow unbounded. I considered exposing a way of freeing references to Caches, but this could be problematic with multiple concurrent parcels working with the same cache, as they would share the same `invalidations` state. Maybe we could count references, and free the cache when users dispose of them?

On the other hand, these objects aren't large and we're unlikely to have a large number of parcels working on large numbers of cache dirs.

wbinnssmith:
The only thing I don't like about this is now Assets need to know details about the cache (the location, and the fact that it uses the filesystem at all). For multiple cache backends, and composable caches, this won't scale. Some alternatives:

* We construct the Cache with a `cacheUri`, which like the various database uri schemes out there, can describe necessary arguments to construct the cache.
* We teach the Cache how to serialize/deserialize (which would involve moving the Cache into core to work with the flatfile serialization @devongovett and I discussed), and pass cache objects to Asset and serialize/deserialize as part of its options.

packages/core/cache/src/Cache.js
DeMoorJasper: It wouldn't make any significant performance impact, just wrote a bunch of test scripts that test this. (all using fs-extra as it has a bunch of other fancy ways to do this)

cached
```JS
const fs = require("fs-extra");
const path = require("path");
const { TARGET_DIR, FILE_PATH, TEST_STRING, RUNS } = require("./constants");

let existsCache = new Set();
async function createDir() {
  if (existsCache.has(TARGET_DIR)) return;
  await fs.mkdirp(TARGET_DIR);
  existsCache.add(TARGET_DIR);
}

async function run() {
  for (let i = 0; i < RUNS; i++) {
    await createDir();
    await fs.writeFile(FILE_PATH, TEST_STRING);
  }
}

run();
```

mkdirp
```JS
const fs = require("fs-extra");
const path = require("path");
const { TARGET_DIR, FILE_PATH, TEST_STRING, RUNS } = require("./constants");

async function run() {
  for (let i = 0; i < RUNS; i++) {
    await fs.mkdirp(TARGET_DIR);
    await fs.writeFile(FILE_PATH, TEST_STRING);
  }
}

run();
```

outputFile
```JS
const fs = require("fs-extra");
const path = require("path");
const { TARGET_DIR, FILE_PATH, TEST_STRING, RUNS } = require("./constants");

async function run() {
  for (let i = 0; i < RUNS; i++) {
    await fs.outputFile(FILE_PATH, TEST_STRING);
  }
}

run();
```

constants.js
```JS
const path = require("path");

const TARGET_DIR = path.join(__dirname, "test");
const FILE_PATH = path.join(TARGET_DIR, "test.js");
const TEST_STRING = "test".repeat(100000);
const RUNS = 7500;

module.exports = { TARGET_DIR, FILE_PATH, TEST_STRING, RUNS };
```

Screenshot 2019-05-31 at 20 23 17

wbinnssmith: Thanks so much for looking into this @DeMoorJasper. Just removed the `existsCache` 😄

wbinnssmith:
Anyone have any remaining thoughts here? This is blocking #3100, #3101, and #3121.

packages/core/cache/src/Cache.js
padmaia: I don't think we need to keep the invalidations state in Cache. That is left over from when the Cache was more coupled to how Parcel transformations were cached.

PR #3101, opened 2019-05-28
Decouple bundler options from WorkerFarm

wbinnssmith:
Depends on #3100 and #3093

This frees WorkerFarm from needing to know about bundler options, making a farm instance reusable across multiple instances of Parcel.

The worker.js file contents is now just this:

```js
// ...
export function init() {}

export function runTransform({
  request,
  config,
  options
}: {
  request: AssetRequest,
  config: Config,
  options: ParcelOptions
}) {
  return new TransformerRunner({
    config,
    options
  }).transform(request);
}

export function runPackage({
  bundle,
  bundleGraph,
  config,
  options
}: {
  bundle: Bundle,
  bundleGraph: BundleGraph,
  config: Config,
  options: ParcelOptions
}) {
  return new PackagerRunner({
    config,
    options
  }).writeBundle(bundle, bundleGraph);
}
```

...meaning that there are no longer any side effects in `init` (which is now empty!). Any instance of Parcel can pass its own options and config along with its transformation or package request without relying on side effects and global state.

Test Plan: `yarn test`, manual tests of kitchen sink example.


packages/core/core/src/worker.js
```diff
@@ -1,45 +1,43 @@
...
 
-type Options = {|
-  config: Config,
-  options: ParcelOptions,
-  env: JSONObject
-|};
-
-let transformerRunner: TransformerRunner | null = null;
-let packagerRunner: PackagerRunner | null = null;
+export function init() {}
```
wbinnssmith: What do you think about getting rid of `init` entirely? Right now `WorkerFarm` only has one usecase and that's this worker file, and it doesn't use `init`.

packages/core/core/src/worker.js
```diff
@@ -1,45 +1,43 @@
...
+export function runTransform({
+  request,
+  config,
+  options
+}: {
+  request: AssetRequest,
+  config: Config,
+  options: ParcelOptions
+}) {
+  return new TransformerRunner({
```
devongovett: wonder how costly constructing these objects is. maybe we could cache them somehow if it becomes a problem...

wbinnssmith: Pretty cheap as far as I can tell. Just creating new objects and simple property assignment in the constructors. Nothing else really happens between that and performing the work.

wbinnssmith:
If you're concerned about de/serialization cost, here's an example of a transformer request from the kitchen sink example:

```
{"type":"request","idx":3,"child":1,"method":"runTransform","args":[{"request":{"filePath":"/Users/wbinnssmith/src/parcel/packages/examples/kitchen-sink/src/index.js","env":{"$$type":"@parcel/core/src/Environment.js","value":{"context":"browser","engines":{"browsers":["last 1 Chrome version"]},"includeNodeModules":true}}},"config":{"$$type":"@parcel/core/src/Config.js","value":{"filePath":"/Users/wbinnssmith/src/parcel/packages/examples/kitchen-sink/.parcelrc","resolvers":["@parcel/resolver-default"],"transforms":{"*.js":["@parcel/transformer-babel","..."],"*.{js,mjs,jsm,jsx,es6}":["@parcel/transformer-babel","@parcel/transformer-js","@parcel/transformer-terser"],"*.css":["@parcel/transformer-css","@parcel/transformer-postcss"],"*.{htm,html}":["@parcel/transformer-html"],"*":["@parcel/transformer-raw"]},"runtimes":{"browser":["@parcel/runtime-js","@parcel/runtime-browser-hmr"],"node":["@parcel/runtime-js"]},"bundler":"@parcel/bundler-default","namers":["@parcel/namer-default"],"packagers":{"*.html":"@parcel/packager-html","*.css":"@parcel/packager-css","*.js":"@parcel/packager-js","*":"@parcel/packager-raw"},"optimizers":{"*.css":["@parcel/optimizer-cssnano"]},"reporters":["@parcel/reporter-cli","@parcel/reporter-dev-server","@parcel/reporter-hmr-server"]}},"options":{"env":{},"entries":["/Users/wbinnssmith/src/parcel/packages/examples/kitchen-sink/src/index.js"],"defaultConfig":{"bundler":"@parcel/bundler-default","transforms":{"*.{js,mjs,jsm,jsx,es6}":["@parcel/transformer-babel","@parcel/transformer-js","@parcel/transformer-terser"],"*.css":["@parcel/transformer-css","@parcel/transformer-postcss"],"*.{htm,html}":["@parcel/transformer-html"],"*":["@parcel/transformer-raw"]},"namers":["@parcel/namer-default"],"runtimes":{"browser":["@parcel/runtime-js","@parcel/runtime-browser-hmr"],"node":["@parcel/runtime-js"]},"optimizers":{"*.css":["@parcel/optimizer-cssnano"]},"packagers":{"*.html":"@parcel/packager-html","*.css":"@parcel/packager-css","*.js":"@parcel/packager-js","*":"@parcel/packager-raw"},"resolvers":["@parcel/resolver-default"],"reporters":["@parcel/reporter-cli","@parcel/reporter-dev-server","@parcel/reporter-hmr-server"],"filePath":"/Users/wbinnssmith/src/parcel/packages/configs/default/index.json"},"cache":true,"cacheDir":".parcel-cache","mode":"production","minify":true,"sourceMaps":true,"hot":false,"serve":false,"targets":[{"name":"browser","distEntry":"index.js","distDir":"dist/legacy","publicUrl":"/","env":{"$$type":"@parcel/core/src/Environment.js","value":{"context":"browser","engines":{"browsers":["> 0.25%"]},"includeNodeModules":true}}},{"name":"browserModern","distDir":"dist/modern","distEntry":"index.js","publicUrl":"/","env":{"$$type":"@parcel/core/src/Environment.js","value":{"context":"browser","engines":{"browsers":["last 1 Chrome version"]},"includeNodeModules":true}}}],"autoinstall":true,"logLevel":"info","rootDir":"/Users/wbinnssmith/src/parcel/packages/examples/kitchen-sink/src","scopeHoist":true,"projectRoot":"/Users/wbinnssmith/src/parcel"}},true]}
```

packages/core/core/src/worker.js
devongovett: I'd be ok with it.

devongovett:
Seems like the options could be slimmed a little... e.g. defaultConfig is redundant, maybe targets too since they are copied into assets/dependencies. Only things that need to be passed to plugins rather than the initial options are needed.

But fine for now.

PR #3102, opened 2019-05-28
Add reverse handle functionality to WorkerFarm

padmaia:
# ↪️ Pull Request

Reverse handles are functions defined in the main process that can be called in worker processes. We will need this functionality for config requests because the main process will be in charge of loading configuration for plugins, so that the graph can be used to avoid rework.

## 🚨 Test instructions

### DIFF
--- a/packages/core/workers/src/WorkerFarm.js
+++ b/packages/core/workers/src/WorkerFarm.js
@@ -5,7 +5,6 @@ import type {
   BundlerOptions,
   CallRequest,
   WorkerRequest,
-  WorkerResponse,
   WorkerDataResponse,
   WorkerErrorResponse
 } from './types';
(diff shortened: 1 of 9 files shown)


packages/core/workers/src/WorkerFarm.js
```diff
@@ -302,6 +316,12 @@ export default class WorkerFarm extends EventEmitter {
     );
   }
 
+  createReverseHandle(fn: () => mixed) {
+    let handle = new Handle();
+    this.handles.set(handle.id, fn);
+    return handle;
```
wbinnssmith: can we return a disposable here in addition to the handle to run `this.handles.delete(handle.id)` when Parcel no longer needs it?

padmaia: Discussed. Will cover this in a separate PR.

PR #3106, opened 2019-05-29
Updates V2 CSSPackager to include media 

### LINKED ISSUE #2524: Parcel 2: CSS Packager
Create the `@parcel/packager-css` package in `packages/packagers/css`. This package is a Parcel 2 packager plugin for CSS.

kwelch:
# ↪️ Pull Request
Specifically handles the case where the media was prefixed to an import. This will wrap the imported asset in a media block. 

fixes #2524

## ✔️ PR Todo

### DIFF
--- a/packages/packagers/css/src/CSSPackager.js
+++ b/packages/packagers/css/src/CSSPackager.js
@@ -6,8 +6,24 @@ export default new Packager({
   async package({bundle}) {
     let promises = [];
     bundle.traverseAssets({
-      exit: asset => {
-        promises.push(asset.getCode());
+      exit: async asset => {
+        let parentDep;
+        bundle.traverseAncestors(asset, (node, _, {stop}) => {
+          if (node.type === 'dependency') {
+            parentDep = node.value;
+            stop();
+          }
+        });
+
+        promises.push(
+          asset.getCode().then((css: string) => {
+            if (parentDep && parentDep.meta && parentDep.meta.media) {
+              return `@media ${parentDep.meta.media} {\n${css}\n}\n`;
+            }
+
+            return css;
+          })
+        );
       }
     });
     let outputs = await Promise.all(promises);
--- a/packages/packagers/css/src/CSSPackager.js
+++ b/packages/packagers/css/src/CSSPackager.js
@@ -7,11 +7,15 @@ export default new Packager({
     let promises = [];
     bundle.traverseAssets({
       exit: asset => {
+        // Figure out which media types this asset was imported with.
+        // We only want to import the asset once, so group them all together.
         let media = [];
         bundle.traverseAncestors(asset, (node, _, {skipChildren}) => {
           if (node.type === 'dependency') {
             let dep = node.value;
             if (!dep.meta || !dep.meta.media) {
+              // Asset was imported without a media type. Don't wrap in @media.
+              media.length = 0;
               skipChildren();
               return;
             }
@@ -30,8 +34,8 @@ export default new Packager({
         );
       }
     });
-    let outputs = await Promise.all(promises);
 
+    let outputs = await Promise.all(promises);
     return {contents: await outputs.map(output => output).join('\n')};
   }
 });
--- a/packages/packagers/css/src/CSSPackager.js
+++ b/packages/packagers/css/src/CSSPackager.js
@@ -7,18 +7,22 @@ export default new Packager({
     let promises = [];
     bundle.traverseAssets({
       exit: asset => {
-        let parentDep;
-        bundle.traverseAncestors(asset, (node, _, {stop}) => {
+        let media = [];
+        bundle.traverseAncestors(asset, (node, _, {skipChildren}) => {
           if (node.type === 'dependency') {
-            parentDep = node.value;
-            stop();
+            let dep = node.value;
+            if (!dep.meta || !dep.meta.media) {
+              skipChildren();
+              return;
+            }
+            media.push(dep.meta.media);
           }
         });
 
         promises.push(
           asset.getCode().then((css: string) => {
-            if (parentDep && parentDep.meta && parentDep.meta.media) {
-              return `@media ${parentDep.meta.media} {\n${css}\n}\n`;
+            if (media.length) {
+              return `@media ${media.join(', ')} {\n${css.trim()}\n}\n`;
             }
 
             return css;
--- a/packages/packagers/css/src/CSSPackager.js
+++ b/packages/packagers/css/src/CSSPackager.js
@@ -6,7 +6,7 @@ export default new Packager({
   async package({bundle}) {
     let promises = [];
     bundle.traverseAssets({
-      exit: async asset => {
+      exit: asset => {
         let parentDep;
         bundle.traverseAncestors(asset, (node, _, {stop}) => {
           if (node.type === 'dependency') {
--- a/packages/core/integration-tests/test/css.js
+++ b/packages/core/integration-tests/test/css.js
@@ -5,7 +5,6 @@ const {
   bundle,
   run,
   assertBundles,
-  assertBundleTree,
   distDir,
   removeDistDirectory
 } = require('@parcel/test-utils');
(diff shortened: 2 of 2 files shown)


kwelch:
This has drastically increased the time it takes to package css files. However, the asset doesn't have the meta passed from the dep and pulling the dep from the asset puts it on the right rule.

packages/packagers/css/src/CSSPackager.js
```diff
@@ -6,8 +6,24 @@ export default new Packager({
   async package({bundle}) {
     let promises = [];
     bundle.traverseAssets({
-      exit: asset => {
-        promises.push(asset.getCode());
+      exit: async asset => {
+        let parentDep;
+        bundle.traverseAncestors(asset, (node, _, {stop}) => {
```
wbinnssmith: This packager and scope hoisting both use `traverseAncestors`, which leaks the internal nodes and internal assets from core. We should figure out from their use an appropriate public api.

Maybe for this one it's `bundle.resolveAssetDependency` or something like that.

wbinnssmith:
> This has drastically increased the time it takes to package css files. 

I wonder what about this is slow. Finding the ancestor dependency should basically be constant time with the inbound adjacency list we added to `Graph` 🤔

packages/packagers/css/src/CSSPackager.js
```diff
@@ -6,8 +6,24 @@ export default new Packager({
   async package({bundle}) {
     let promises = [];
     bundle.traverseAssets({
-      exit: asset => {
-        promises.push(asset.getCode());
+      exit: async asset => {
```
devongovett: async seems not needed since there is no await in this function

packages/packagers/css/src/CSSPackager.js
devongovett: I think a `getIncomingDependencies` or something could work. Not sure on the naming.

devongovett:
Did you try profiling? Easiest way is to disable workers so everything runs in one process by setting `PARCEL_WORKERS=0` in the environment. Then you can use the node `--inspect-brk` and open Chrome debugger.

kwelch:
> Did you try profiling?
No but I can try that later. My note about speed was 100% based on the output from the test runner that the tests were taking longer. 😅

I was going to link to the build output, but it appears that the integration-tests are not actually working in Azure. 
https://dev.azure.com/devongovett/devongovett/_build/results?buildId=2621&view=logs&jobId=c9e7f81a-e301-5408-62aa-d2082bd3a1e5&taskId=8a28f03a-85a3-57ec-157d-3f669a62b681&lineStart=405&lineEnd=406&colStart=1&colEnd=1

kwelch:
Here is the timing I meant:

<details>
<summary>Before Media Changes (Done in 19.38s)</summary>
```
yarn run v1.13.0
$ cross-env NODE_ENV=test mocha --grep css


  css

WARNING: We noticed you're using the `useBuiltIns` option without declaring a core-js version. Currently, we assume version 2.x when no version is passed. Since this default version will likely change in future versions of Babel, we recommend explicitly setting the core-js version you are using via the `corejs` option.

You should also be sure that the version you pass to the `corejs` option matches the version specified in your `package.json`'s `dependencies` section. If it doesn't, you need to run one of the following commands:

  npm install --save core-js@2    npm install --save core-js@3
  yarn add core-js@2              yarn add core-js@3

    ✓ should produce two bundles when importing a CSS file (2476ms)
    ✓ should bundle css dependencies in the correct, postorder traversal order (61ms)
    ✓ should support loading a CSS bundle along side dynamic imports (195ms)
    - should support importing CSS from a CSS file
    ✓ should support linking to assets with url() from CSS (76ms)
Without `from` option PostCSS could generate wrong source map and will not find Browserslist config. Set it to CSS file path or to `undefined` to prevent this warning.
    ✓ should support linking to assets with url() from CSS in production (764ms)
    ✓ should support linking to assets in parent folders with url() from CSS (93ms)
    ✓ should minify CSS when minify is set (88ms)

  html
    - should handle inline css with @imports

  less
    - should support transforming less with postcss

  postcss
    ✓ should support transforming css modules with postcss (89ms)
    ✓ should support transforming with postcss twice with the same result (105ms)
(node:31538) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 reporterEvent listeners added. Use emitter.setMaxListeners() to increase limit
    ✓ should support postcss composes imports (117ms)
    ✓ should not include css twice for postcss composes imports (101ms)
    - should support postcss composes imports for sass
    ✓ should support postcss composes imports with custom path names (74ms)
    ✓ should support deep nested postcss composes imports (101ms)
    ✓ should support postcss composes imports for multiple selectors (67ms)
    ✓ should automatically install postcss plugins with npm if needed (9915ms)
    ✓ should automatically install postcss plugins with yarn if needed (441ms)

  sass
    - should support requiring scss files
    - should support scss imports
    - should support requiring empty scss files
    - should support linking to assets with url() from scss
    - should support transforming scss with postcss

  stylus
    - should support transforming stylus with postcss


  16 passing (15s)
  10 pending

✨  Done in 19.38s.
```
</details>

<details>
<summary>After Media Changes (Done in 16.34s)</summary>
```
yarn run v1.13.0
$ cross-env NODE_ENV=test mocha --grep css


  css
Browserslist: caniuse-lite is outdated. Please run next command `yarn upgrade caniuse-lite browserslist`
    ✓ should produce two bundles when importing a CSS file (3052ms)
    ✓ should bundle css dependencies in the correct, postorder traversal order (58ms)
    ✓ should support loading a CSS bundle along side dynamic imports (219ms)
    ✓ should support importing CSS from a CSS file (72ms)
    ✓ should support linking to assets with url() from CSS (87ms)
Without `from` option PostCSS could generate wrong source map and will not find Browserslist config. Set it to CSS file path or to `undefined` to prevent this warning.
    ✓ should support linking to assets with url() from CSS in production (493ms)
    ✓ should support linking to assets in parent folders with url() from CSS (70ms)
    ✓ should minify CSS when minify is set (90ms)

  html
    - should handle inline css with @imports

  less
    - should support transforming less with postcss

  postcss
    ✓ should support transforming css modules with postcss (106ms)
(node:31404) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 reporterEvent listeners added. Use emitter.setMaxListeners() to increase limit
    ✓ should support transforming with postcss twice with the same result (107ms)
    ✓ should support postcss composes imports (119ms)
    ✓ should not include css twice for postcss composes imports (103ms)
    - should support postcss composes imports for sass
    ✓ should support postcss composes imports with custom path names (74ms)
    ✓ should support deep nested postcss composes imports (150ms)
    ✓ should support postcss composes imports for multiple selectors (77ms)
    ✓ should automatically install postcss plugins with npm if needed (6691ms)
    ✓ should automatically install postcss plugins with yarn if needed (455ms)

  sass
    - should support requiring scss files
    - should support scss imports
    - should support requiring empty scss files
    - should support linking to assets with url() from scss
    - should support transforming scss with postcss

  stylus
    - should support transforming stylus with postcss


  17 passing (12s)
  9 pending

✨  Done in 16.34s.
```
</details>

Looks like the point is moot as the speed is actually faster and with an additional test. 

I think what caught me off guard was the time stamps in red next to the tests which in jest I only see if they are running long.

packages/packagers/css/src/CSSPackager.js
kwelch: Looks like scope hoisting is also catching `assets`, not sure there is a solid abstraction with only these 2 places using it at this point.

kwelch: Based on v1, it looks like we are looking at all the parentDeps, should I not stop the traversal?

https://github.com/parcel-bundler/parcel/blob/576dba26cbc5f52d8637a4824d76d3f25fba6185/packages/core/parcel-bundler/src/packagers/CSSPackager.js#L15-L31

packages/packagers/css/src/CSSPackager.js
```diff
@@ -7,7 +7,23 @@ export default new Packager({
     let promises = [];
     bundle.traverseAssets({
       exit: asset => {
-        promises.push(asset.getCode());
+        let parentDep;
+        bundle.traverseAncestors(asset, (node, _, {stop}) => {
+          if (node.type === 'dependency') {
+            parentDep = node.value;
+            stop();
```
devongovett: The algorithm here appears a little different from what Parcel 1 was doing. Parcel 1 collected all media values from all incoming dependencies and joined them together, whereas this stops after the first dependency. If a dep was seen that didn't have a media setting at all, then it is meant for all media, and so the wrapper is not added. I think perhaps you could use `skipChildren` here instead, and just push the media type to an array.

kwelch: Yea, I was thinking the same thing. 

https://github.com/parcel-bundler/parcel/pull/3106#discussion_r289544286

Since, we are going upward does `skipChildren` make sense?

devongovett: Yeah... I guess the API is a bit confusing. "children" in this case refer to the next items we would traverse (in this case they are parents). Maybe we should reconsider that name, but it should do what you want.

kwelch: Updated! This should be properly going up the chain and stopping at the first none wrapped dependancy

PR #3107, opened 2019-05-29
Add V2 CoffeeScript Transformer

### LINKED ISSUE #2160: Parcel 2: CoffeeScript transform
Create the `@parcel/transform-coffeescript` package in `packages/transforms/coffeescript`. This package is a Parcel 2 transform plugin for CoffeeScript. It will be based on the `CoffeeScriptAsset` from Parcel 1.

The `@parcel/transform-coffeescript` package is responsible for the following things:

* Transforming CoffeeScript to JavaScript
* Generating source maps.

kwelch:
# ↪️ Pull Request
Used the v1 transformer to add the v2 transformer.

fixes #2160

## ✔️ PR Todo

### DIFF
--- a/packages/core/integration-tests/test/javascript.js
+++ b/packages/core/integration-tests/test/javascript.js
@@ -14,7 +14,7 @@ const {makeDeferredWithPromise} = require('@parcel/utils');
 
 describe('javascript', function() {
   afterEach(async () => {
-    await removeDistDirectory();
+    // await removeDistDirectory();
   });
 
   it('should produce a basic JS bundle with CommonJS requires', async function() {
@@ -762,18 +762,15 @@ describe('javascript', function() {
     assert.equal(output(), 3);
   });
 
-  it.skip('should support requiring CoffeeScript files', async function() {
+  it('should support requiring CoffeeScript files', async function() {
     let b = await bundle(path.join(__dirname, '/integration/coffee/index.js'));
 
-    await assertBundles(b, {
-      name: 'index.js',
-      assets: ['index.js', 'local.coffee'],
-      childBundles: [
-        {
-          type: 'map'
-        }
-      ]
-    });
+    await assertBundles(b, [
+      {
+        name: 'index.js',
+        assets: ['index.js', 'local.coffee']
+      }
+    ]);
 
     let output = await run(b);
     assert.equal(typeof output, 'function');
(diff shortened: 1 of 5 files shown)


packages/core/integration-tests/test/javascript.js
```diff
@@ -14,7 +14,7 @@ const {makeDeferredWithPromise} = require('@parcel/utils');
 
 describe('javascript', function() {
   afterEach(async () => {
-    await removeDistDirectory();
+    // await removeDistDirectory();
```
wbinnssmith: debug 😄

kwelch: Ah. You caught me. 😂

packages/transformers/coffeescript/src/CoffeeScriptTransformer.js
```diff
@@ -0,0 +1,33 @@
...
+  },
+
+  async transform({asset}) {
+    asset.type = 'js';
+
+    return [asset];
+  },
+
+  async generate({asset}) {
+    const transpiled = coffee.compile(await asset.getCode());
```
devongovett: Should we compile based on the AST? Otherwise there's no point in parsing above since it will parse again here.

kwelch: Couldn't find it at first. Got it wired up now.

mischnic:
As far as I can see, sourcemaps can only be emitted by `coffescript.compile`, not via `nodes`. Is there even any benefit to having the AST as an object as it is now?

kwelch:
>Is there even any benefit to having the AST as an object as it is now?

I think the only benefit there would be is if there was another coffee transformer. In its current state, I think we are find removing the parse and transform and just using the generate to create the code and sourcemap.

kwelch:
@mischnic I tried to reduce this down to remove the parse and use compile but it fails, since it no longer calls the generate function in the coffee transformer.

mischnic:
>  since it no longer calls the generate function in the coffee transformer.

I guess it would then have to be in `transform` then.
What is cached for transformers?  The AST and/or the result of transform?

kwelch:
Sweet found an example in JSONTransformer of how to do it via the transform. That make so much sense. Thanks!

devongovett:
That's fine. FYI you can also just return a new object instead of mutating. 🤷‍♂ 

```javascript
return {
  type: 'js',
  code: output
};
```

kwelch:
> That's fine. FYI you can also just return a new object instead of mutating. 🤷‍♂
> 
> ```js
> return {
>   type: 'js',
>   code: output
> };
> ```

That makes sense, I think I matched this based on the `JSONTransformer`

PR #3109, opened 2019-05-29
Register classes with serializer rather than dynamic require

wbinnssmith:
This reworks how de/serialization happens in Parcel 2.

Rather than transforming sources to include a path to require constructors, manually register name/constructor pairs with the serializer.

Core registers its own classes with the serializer by using a key of its own version number along with a simple hardcoded name of the class.

This allows the serializer to function when Parcel 2 is built with Parcel 2 into a flat file, as well as will make it easier to run Parcel 2 in a browser.

Test Plan: yarn test, manual test of kitchen sink and react hmr examples.


packages/core/core/src/registerCoreWithSerializer.js
```diff
@@ -0,0 +1,31 @@
...
+  throw new Error('Expected package version to be a string');
+}
+
+let registered;
+export default function registerCoreWithSerializer() {
+  if (registered) {
+    return;
+  }
+
+  register('Asset', Asset);
```
wbinnssmith: I created this list based on logging requests to the serializer for object values. Apparently this is all that was needed?

Not sure how to make sure that it's clear that if you add a new class that must be serialized, it doesn't slip through the cracks, since we don't throw on unexpected constructors.

Maybe we should?

packages/core/utils/src/serializer.js
```diff
@@ -1,5 +1,21 @@
 // @flow
 
+const nameToCtor: Map<string, Class<*>> = new Map();
+const ctorToName: Map<Class<*>, string> = new Map();
```
wbinnssmith: Currently these cannot be cleared. Since they don't grow over time, and once constructors and require cache entries are loaded into memory, they can't be unloaded, I don't think being able to dispose of this would be very impactful.

packages/core/core/src/registerCoreWithSerializer.js
devongovett: what do you think about having these register calls inside the files where the types are declared rather than maintaining a list here?

devongovett: I guess that would require us to actually `import` all of those types before tying to serialize anything though... maybe this is clearer then. 🤷‍♂

packages/core/utils/src/serializer.js
```diff
@@ -1,5 +1,21 @@
...
+
+export function registerSerializableClass(name: string, ctor: Class<*>) {
```
devongovett: probably could get the name from `constructor.name`?

packages/core/utils/src/serializer.js
devongovett: doesn't seem like a big deal. there are only a few of them

packages/core/core/src/registerCoreWithSerializer.js
wbinnssmith: This was my exact thought process 😅

packages/core/utils/src/serializer.js
wbinnssmith: I considered this, however, I always get concerned relying on constructor names in any permanent sense because of minifies.

I guess this would really not impact us here unless we had different builds of the same version of core registering themselves, in which case we'd have other problems.

This would also lock us into needing to use constructed classes for everything, when in the future we could accept another type name property instead. Again maybe not an issue...

wbinnssmith: The other thing is that we need to incorporate other differentiating properties like the version of the constructor.

PR #3110, opened 2019-05-30
Preserve hashbangs and file modes from entry assets to bundles

wbinnssmith:
This is necessary for Parcel 2 to build node cli executables like Parcel's own cli.

For hashbangs in JavaScript, we extract the `interpreter` value from the Babel AST, and store it in `asset.meta.interpreter`. We then read from this during packaging and, if present, prepend it to the bundle output. We do the same in scope hoisting.

For file modes (unix-style numeric permission values), a `fileMode` property is attached to each asset. After packaging and optimizing, before the file is written, we use the `fileMode` associated with the entry asset as the file mode for the file written to disk.

Test Plan: Added integration tests for hashbangs, hashbangs in scope hoisting, and preserving file modes.

### DIFF
--- a/packages/core/core/src/PackagerRunner.js
+++ b/packages/core/core/src/PackagerRunner.js
@@ -50,10 +50,14 @@ export default class PackagerRunner {
       this.distExists.add(dir);
     }
 
-    // Use the file mode from the entry asset as the file mode for the bundle
-    let options = {
-      mode: (await fs.stat(bundle.assetGraph.getEntryAssets()[0].filePath)).mode
-    };
+    // Use the file mode from the entry asset as the file mode for the bundle.
+    // Don't do this for browser builds, as the executable bit in particular is unnecessary.
+    let options = nullthrows(bundle.target).env.isBrowser()
+      ? undefined
+      : {
+          mode: (await fs.stat(bundle.assetGraph.getEntryAssets()[0].filePath))
+            .mode
+        };
 
     let size;
     if (contents instanceof Readable) {
--- a/packages/core/core/src/PackagerRunner.js
+++ b/packages/core/core/src/PackagerRunner.js
@@ -8,7 +8,7 @@ import type InternalBundleGraph from './BundleGraph';
 
 import {Readable} from 'stream';
 import invariant from 'assert';
-import {mkdirp, writeFile, writeFileStream} from '@parcel/fs';
+import * as fs from '@parcel/fs';
 import {urlJoin} from '@parcel/utils';
 import {NamedBundle} from './public/Bundle';
 import nullthrows from 'nullthrows';
@@ -46,15 +46,20 @@ export default class PackagerRunner {
     let filePath = nullthrows(bundle.filePath);
     let dir = path.dirname(filePath);
     if (!this.distExists.has(dir)) {
-      await mkdirp(dir);
+      await fs.mkdirp(dir);
       this.distExists.add(dir);
     }
 
+    // Use the file mode from the entry asset as the file mode for the bundle
+    let options = {
+      mode: (await fs.stat(bundle.assetGraph.getEntryAssets()[0].filePath)).mode
+    };
+
     let size;
     if (contents instanceof Readable) {
-      size = await writeFileStream(filePath, contents);
+      size = await fs.writeFileStream(filePath, contents, options);
     } else {
-      await writeFile(filePath, contents);
+      await fs.writeFile(filePath, contents, options);
       size = contents.length;
     }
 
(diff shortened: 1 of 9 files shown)


packages/transformers/js/src/JSTransformer.js
```diff
@@ -82,7 +82,15 @@ export default new Transformer({
       walk.ancestor(ast.program, collectDependencies, asset);
     }
 
-    if (!asset.env.isNode()) {
+    if (asset.env.isNode()) {
```
wbinnssmith: @devongovett does `isNode` cover electron too? Maybe safer to just always do this?

devongovett: yeah it does cover electron

packages/core/core/src/TransformerRunner.js
```diff
@@ -337,29 +341,37 @@ async function summarizeRequest(
...
-    );
+          })
+        )
+      ),
+      // `mode` is a decimal representation of the more common octal format of
+      // unix file permissions. Assign it to the asset so it can be used to determine
+      // the permissions of the output bundle. Node's documentation on this is pretty
+      // sparse as it's probably just the result of a `stat` syscall. More info:
+      // https://stackoverflow.com/questions/11775884/nodejs-file-permissions
+      stat(req.filePath).then(stats => stats.mode)
```
devongovett: hmm an extra stat call for every asset won't be cheap. wonder if we could just stat the entry file in the packager runner rather than propagating `fileMode` everywhere...

wbinnssmith: Awesome idea. Just implemented it 😄

mischnic:
The hashbang should be removed (not added in the Packager) with target: browser

mischnic:
Related: https://github.com/parcel-bundler/parcel/issues/2381, #2975

wbinnssmith:
> Related: #2381, #2975

Awesome 😃

> The hashbang should be removed (not added in the Packager) with target: browser

I considered this too. I thought it would be a weird circumstance that an entry file with a hashbang would be built for a browser, and that if we just removed it, it might hide a real problem.

Maybe there's a case for sharing an entry amongst node and browser targets, and having the node one be executable?

mischnic:
> Maybe there's a case for sharing an entry amongst node and browser targets, and having the node one be executable?

I think the shebang isn't actually part of the JS syntax eventhough Chrome and Firefox accept them.
Safari however: `SyntaxError: Invalid character: '#'`

wbinnssmith:
Given there's a proposal for browser environments to ignore hashbangs, I'm on board with removing them and not restoring them for browser targets. @devongovett, what do you think?

packages/core/core/src/PackagerRunner.js
```diff
@@ -40,15 +40,20 @@ export default class PackagerRunner {
...
     let dir = path.dirname(filePath);
     if (!this.distExists.has(dir)) {
-      await mkdirp(dir);
+      await fs.mkdirp(dir);
       this.distExists.add(dir);
     }
 
+    // Use the file mode from the entry asset as the file mode for the bundle
+    let options = {
+      mode: (await fs.stat(bundle.assetGraph.getEntryAssets()[0].filePath)).mode
```
mischnic: I don't think this should be done for browser targets. Having a file with the executable bit set might have unexpected security issues on the server.

Also, shouldn't this rather copy only the executable bit and leave read/write on default?

wbinnssmith:
What's preventing this from landing?

packages/shared/scope-hoisting/src/generate.js
```diff
@@ -13,5 +13,12 @@ export function generate(bundle: Bundle, ast: AST, options: ParcelOptions) {
     code = `\n${code}\n`;
   }
 
-  return {contents: `(function(){${code}})();`};
+  let entryAsset = bundle.getEntryAssets()[0];
+  // $FlowFixMe
+  let interpreter: ?string = entryAsset.meta.interpreter;
```
mischnic: The isBrowser check is needed here as well

packages/core/core/src/PackagerRunner.js
mischnic: @wbinnssmith ?

wbinnssmith:
This is good for another review @mischnic @devongovett

packages/core/core/src/PackagerRunner.js
mischnic: > This is good for another review

...

wbinnssmith: Good call. I had assumed you were referring to hashbangs before :)

I agree the executable bit isn't necessary and only creates concerns with webservers. Not sure about the other bits of permissions, but you could probably make the argument they do the same (why should the permissions of source code in a build environment reflect the permissions of the build target run on a webserver?)

PR #3118, opened 2019-05-31
Parcel 2: Stylus Transformer

devongovett:
This implements a Stylus transformer for Parcel 2, directly ported from Parcel 1, and passes all integration tests.

In addition, it fixes a long standing TODO in the TransformerRunner where an asset could change `type`, but not change pipelines (multiple file extensions could match the same pipeline). This occurs e.g. with `.htm` -> `.html`. Relatedly, assets were not jumping pipelines properly if they were mutated by a transformer to change the type (e.g. stylus -> css).


packages/transformers/css/src/CSSTransformer.js
```diff
@@ -21,6 +21,10 @@ export default new Transformer({
   },
 
   async parse({asset}) {
+    if (asset.meta.hasDependencies === false) {
```
devongovett: This is set by the Stylus transformer to indicate that it has already processed all dependencies, and that the CSS transformer can skip this asset completely. This is required because when stylus processes e.g. `url()` it replaces them with a dependency id to be filled in later. When the CSS transformer runs, it would pick that up and try to resolve a dependency for the id which obviously doesn't exist. Also, it's faster to do it this way since the resulting CSS doesn't need to be re-parsed.

packages/resolvers/default/src/DefaultResolver.js
```diff
@@ -18,7 +18,7 @@ import builtins from './builtins';
 export default new Resolver({
   async resolve({dependency, options}) {
     const resolved = await new NodeResolver({
-      extensions: ['js', 'json', 'css'],
+      extensions: ['js', 'json', 'css', 'styl'],
```
devongovett: Really need to figure this out...

packages/transformers/stylus/package.json
```diff
@@ -0,0 +1,24 @@
...
+  "engines": {
+    "node": ">= 8.0.0",
+    "parcel": "^2.0.0-alpha.0"
+  },
+  "dependencies": {
+    "@parcel/fs": "^2.0.0-alpha.0",
+    "@parcel/local-require": "^2.0.0-alpha.0",
+    "@parcel/plugin": "^2.0.0-alpha.0",
+    "@parcel/utils": "^2.0.0-alpha.0",
+    "semver": "^5.4.1"
```
mischnic: `semver` is not needed

packages/core/core/src/TransformerRunner.js
```diff
@@ -97,6 +97,7 @@ export default class TransformerRunner {
     let {assets, initialAssets} = await this.runPipeline(
       input,
       pipeline,
+      pipeline,
```
wbinnssmith: may be time to convert `runPipeline` to a params object given there are four args and the second and third are both `pipeline` here 😄

packages/core/types/index.js
```diff
@@ -300,31 +300,36 @@ export interface TransformerResult {
...
     asset: MutableAsset,
-    resolve: (from: FilePath, to: string) => Promise<FilePath>,
+    resolve: ResolveFn,
     options: ParcelOptions
   }) => Async<Config | void>,
   canReuseAST?: ({ast: AST, options: ParcelOptions}) => boolean,
   parse?: ({
     asset: MutableAsset,
     config: ?Config,
+    resolve: ResolveFn,
```
wbinnssmith: @padmaia is not going to like this 😝

packages/transformers/css/src/CSSTransformer.js
wbinnssmith: Can you put this in a comment above this line?

packages/core/core/src/TransformerRunner.js
devongovett: haha yeah

Issue #2310, opened 2018-11-21
Import single value (for security reasons)


doomedramen:
My project (a node web app) has a file (`config.js`) sitting at its root. It contains passwords, preferences etc. I would like to pull one of these into my front end js file that is being built with parcel but when it is built the entire config.js file copied to the output.

For example:
```
import {apdexT} from '../../../config.js';
```
in my code I expect to see the val `apdexT` but I see the entire contents of `../../../config.js`
Is there a way for the rest of the fie (that is not used in the file built with parcel) to be ignored?

Thank you in advance.

wookoouk:
in my built code there a like that reads:
```
module.exports={port:3001,db:"uptime",secret:"Change Me",showSitesToUnauth:!0,graphsOnIndex:!0,allowRegistration:!1,apdexT:.2,monitoring:{localRunner:!0,interval:5},slack:{enabled:!0,webhookURL:"..."},airdale:{enabled:!1,apiKey:""}};
```
This is the entire contents of the config.js file, only `apdexT` should have been imported.

mischnic:
You could use an envfile

mischnic:
You could create a `.envfile` containing
```
API_KEY=0123456
```
and then use it like this: `process.env.API_KEY`.

(Inspired by https://github.com/parcel-bundler/parcel/issues/2299)

macklinu:
My use case feels similar to this issue (tree shaking for JS assets) but not quite since my example is importing a JSON asset. Should I track JSON asset tree-shaking in a separate ticket (don't see anything open for this at the moment)?

DeMoorJasper:
@macklinu JSON files get transformed to JS and JS gets transformed by babel and gathers information for treeshaking so in theory it should already do this.

Feel free to experiment with it using `parcel build index.html --experimental-scope-hoisting`

mischnic:
After treeshaking: 

```js
(function() {
	var a = {};
	a = {
		version: "1.2.3",
		scripts: {
			build: "parcel build --target node --experimental-scope-hoisting index.js",
			"build-ts": "parcel build --target node index.js"
		},
		devDependencies: {
			"@babel/core": "^7.2.0",
			"@babel/preset-env": "^7.2.0"
		}
	};
	console.log(a.version);
})();

```

mischnic:
If I recall correctly, parcel's scope hoisting visitor removes all unreferenced variables. It doesn't go into the "object" level.

mischnic:
[...]

PR #3120, opened 2019-05-31
Convert Terser to optimizer

wbinnssmith:
This converts the terser transformer to an optimizer. This:
* Optimizes the prelude, which never went through the transformer
* Upgrades terser to 4.0
* Continues to honor the user's uglify/terser config
* Uses terser's default for mangling (top-level and eval are false) in the fallback config

Test Plan: `yarn test`, and use the kitchen sink example with scope hoisting on and off to verify executable output. Verify in the non-scopehoisted case that `parcelRequire` isn't mangled.


mischnic:
Previous discussion on whether terser should run on Asset or Packager level: https://github.com/parcel-bundler/parcel/issues/2310

This is *much* better for code size, but might have longer built time (and memory usage?).

wbinnssmith:
Re: build time, given terser (and most optimizers in general) are likely only enabled with `minify: true` which isn't the default at development, I'm not really concerned unless it's really significant.

mischnic:
> top-level false

Could we enable this?
> set to true if you wish to enable top level variable and function name mangling and **to drop unused variables and functions.**

wbinnssmith:
> Could we enable this?

I think we can, given Parcel controls the top-level. Is it possible that after scope hoisting and tree shaking there are unused variables?

devongovett:
I think that option is for mangling, not dead code elimination

packages/optimizers/terser/src/TerserOptimizer.js
```diff
@@ -0,0 +1,40 @@
...
+
+export default new Optimizer({
+  async optimize({contents, bundle, options}) {
+    if (!options.minify) {
+      return contents;
+    }
+
+    if (typeof contents !== 'string') {
+      throw new Error(
+        'TerserOptimizer: Only string contents are currently supported'
```
devongovett: ahh this is interesting. maybe once we introduce ast sharing for optimizers, we could end up with a `getCode`, `getBuffer`, `getAST` etc. on a `Bundle` similar to the ones in `Asset`

PR #3121, opened 2019-06-01
Account for core version and impactful options when reusing cacheEntries

wbinnssmith:
Depends on #3093

When we made the changes to cache not aware of Parcel specifics, we didn't update the ability to ignore cache entries based on changes in parcel options like minify and scope hoisting.

This restores them, and also stores the version of core on the entry as well, in the case changes to core could impact the transform output.

Open question:
* How should we invalidate the cache based on changes to parcel plugin versions and plugin configs?

Test Plan: Run parcel with scope hoisting. Run parcel without scope hoisting. Verify the bundle output does not include hoisted contents.

### DIFF
--- a/packages/core/core/src/TransformerRunner.js
+++ b/packages/core/core/src/TransformerRunner.js
@@ -15,13 +15,16 @@ import type {CacheEntry} from './types';
 
 import path from 'path';
 import clone from 'clone';
+import {createReadStream} from 'fs';
+import shallowEqual from 'shallowequal';
 import {
   md5FromFilePath,
   md5FromReadableStream,
   md5FromString
 } from '@parcel/utils';
 import {TapStream, unique} from '@parcel/utils';
-import {createReadStream} from 'fs';
+import {getCacheByDir} from '@parcel/cache';
+import {pick, objectSortedEntriesDeep} from '@parcel/utils';
 
 import Dependency from './Dependency';
 import Config from './Config';
@@ -29,7 +32,7 @@ import ResolverRunner from './ResolverRunner';
 import {report} from './ReporterRunner';
 import {MutableAsset, assetToInternalAsset} from './public/Asset';
 import InternalAsset from './Asset';
-import {getCacheByDir} from '@parcel/cache';
+import corePackage from '../package.json';
 
 type Opts = {|
   config: Config,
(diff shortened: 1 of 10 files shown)


wbinnssmith:
now using the `shallowEqual` module instead of our own: https://github.com/parcel-bundler/parcel/compare/c3507b0469d917566b21e0b6011c0abcdfbdfe5b..c2315b5a0bee1f72345ef991a2d892f39a9225e9

packages/core/core/src/TransformerRunner.js
```diff
@@ -38,8 +41,12 @@ type Opts = {|
 
 type GenerateFunc = (input: IMutableAsset) => Promise<GenerateOutput>;
 
+const CORE_VERSION = corePackage.version;
 const BUFFER_LIMIT = 5000000; // 5mb
 
+// These keys can affect asset output, so if they differ, we should always transform
+const IMPACTFUL_OPTIONS = ['minify', 'hmr', 'scopeHoist'];
```
mischnic: Does scopehoisting bust the complete cache? Because with a typescript project, the typescript-transform could be reused (and css/html/... as well).

padmaia:
I guess updating parcel core versions shouldn't happen too often, but it does seem unfortunate that the entire cache would have to be evicted as a result of possibly unrelated changes. Not sure how we could ensure that only relevant changes evict the cache though. A specialized package could work, but not sure that's the direction we want to go in.

packages/core/core/src/TransformerRunner.js
wbinnssmith: I think this probably falls into letting transformers decide if they'd reuse cached results. @padmaia has been thinking about how to get transformers participating in cache invalidation. 

For now I think leaning on being as conservative as possible is probably the way to go. But right now we still don't invalidate if a plugin updates, which seems like a big oversight. I think @padmaia's working on that too.

wbinnssmith:
> I guess updating parcel core versions shouldn't happen too often, but it does seem unfortunate that the entire cache would have to be evicted as a result of possibly unrelated changes.

I'm again erring on the side of being conservative with cache invalidation. To me, updating parcel is significant to warrant a change in cache. Maybe we do some semver thing here and promise to never break the cache entry format within a given range?

wbinnssmith:
Closing in favor of #3127.

Issue #2883, opened 2019-04-02
Critical dependency: the request of a dependency is an expression


mattrothenberg:
# 🐛 bug report

Howdy! I'm the author of [`vue-storybook`](https://github.com/mattrothenberg/vue-storybook/tree/v2), a library that allows Vue developers to add `<story>` blocks to their Vue single-file components and have them automatically translated into [Storybook](https://storybook.js.org/) stories.

Parcel is my go-to bundler, though I'm experiencing some strange behavior when bundling/publishing my library code and subsequently using it in a freshly scaffolded Vue CLI project. Without fail, when running the Vue development server, I'm getting the following warning:

```
WARNING in ./node_modules/vue-storybook/dist/index.js 1:292-296
Critical dependency: the request of a dependency is an expression
 @ ./config/storybook/config.js
 @ multi ./node_modules/@storybook/core/dist/server/common/polyfills.js ./node_modules/@storybook/core/dist/server/preview/globals.js ./config/storybook/config.js ./node_modules/webpack-hot-middleware/client.js?reload=true
```

I'm confused, because I _don't_ see this message when I bundle my library code with Webpack or Rollup. I'm doubly confused because I'm not using any special configuration with Parcel, just a good ole' `build` command. It's just a warning, so I'm not _too_ concerned. But I've gotten user feedback that this is distracting. I'd like the warning to go away.

## 🎛 Configuration (.babelrc, package.json, cli command)
There's no Babel configuration for `vue-storybook`, though freshly scaffolded Vue CLI projects have the following `babel.config.js`

```js
module.exports = {
  presets: [
    '@vue/app'
  ]
}
```

## 🤔 Expected Behavior

When using `vue-storybook` with a freshly scaffolded Vue CLI project, there should be no such warning about `the request of a dependency is an expression`

## 😯 Current Behavior
When using `vue-storybook` with a freshly scaffolded Vue CLI project, there is a warning about `the request of a dependency is an expression`

## 💁 Possible Solution
No ideas. I'm fairly certain this is user error somewhere on my part.

## 🔦 Context
I'm trying to publish a new version of my library (bundled with Parcel), and I'd like consumers _not_ to see this warning message when using it.

## 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Parcel           | 1.12.3
| Node             | 10.15.1
| npm/Yarn         |1.13.0
| Operating System | OSX 10.14.3

mischnic:
[...]

PR #3129, opened 2019-06-04
Use global.require to silence webpack warning

mischnic:
# ↪️ Pull Request

I'm still not quite sure whether module.require and require are completely equivalent. Doing `toString` in node show this:

```js
module.require = function(id) {
  validateString(id, 'id');
  if (id === '') {
    throw new ERR_INVALID_ARG_VALUE('id', id, 'must be a non-empty string');
  }
  return Module._load(id, this, /* isMain */ false);
};

global.require = function(path) {
  try {
    exports.requireDepth += 1;
    return mod.require(path);
  } finally {
    exports.requireDepth -= 1;
  }
};
```

Related: #2883


mischnic:
Is `typeof global.require === "function"` needed as well?

devongovett:
I don’t think require is actually a global in node right? It wraps modules in a closure with a require argument, just like parcel does. So checking global.require probably won’t work.

mischnic:
You're right...
But `module.require` and `require` are slightly different.

devongovett:
Should we check that both module.require and require exist, but actually call require?

mischnic:
> but actually call require?

No, because the original issue is that with `typeof require === 'function' && require`, Webpack can infer that `nodeRequire` is `require` and tries to bundle that call.

packages/shared/scope-hoisting/src/prelude.js
```diff
@@ -6,8 +6,8 @@ parcelRequire = function(name) {
   }
 
   // Try the node require function if it exists.
-  if (typeof require === 'function') {
-    return require(name);
+  if (typeof global !== 'undefined' && typeof global.require === 'function') {
```
devongovett: Probably should update this one as well 😉

packages/packagers/js/src/prelude.js
```diff
@@ -10,7 +10,10 @@
 parcelRequire = function(modules, cache, entry, globalName) {
   // Save the require from previous bundle to this closure if any
   var previousRequire = typeof parcelRequire === 'function' && parcelRequire;
-  var nodeRequire = typeof require === 'function' && require;
+  var nodeRequire =
+    typeof module !== 'undefined' &&
+    typeof module.require === 'function' &&
+    module.require.bind(module);
```
devongovett: Can you put a comment about why we don't just use `require` so someone doesn't try to change this later?

PR #3134, opened 2019-06-06
Add a basic build system for Parcel 2

wbinnssmith:
While we're striving to build Parcel 2 with Parcel 2, we've decided for early alphas to build Parcel simply with babel.

This implements an in-place build system that will allow us to continue to use `@babel/register` during development, avoid using a watcher and `src`/`lib` copies, but still be able to publish compiled output in `lib`.

This is achieved by a gulpfile which performs the following:
* first, compiling relevant files with babel and placing them in `lib`
* copying remaining files to `lib`
* adjusting `package.json` `main` and `bin` entries to refer to `lib` counterparts instead of `src` files.
* removing `-r @babel/register` from node interpreter hash bangs

* `gulp` is now the default `build` script in the monorepo package.
* build steps using the babel cli have been removed in favor of the single build script
* the `postinstall` step invoking build has been removed as it is no longer necessary to develop parcel (nor has it been for a while), and it would leave the working tree in a dirty state.

Note this leaves the git directory in a dirty state, as it modifies `package.json` in place. During tests and local development, this can easily be reversed by running `git checkout packages/` or `git reset --hard` if necessary.

Test Plan: Run `yarn build`, `yarn`, and run the examples.

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -1,0 +1,3 @@
+module.exports = {
+  presets: ['@parcel/babel-preset']
+};
--- a/package.json
+++ b/package.json
@@ -11,25 +11,29 @@
     "packages/*/*"
   ],
   "scripts": {
-    "build": "lerna run build --stream",
-    "test": "lerna run test --stream",
-    "test-ci": "lerna run test-ci --stream",
-    "lint": "eslint . && prettier \"./packages/*/*/{src,bin,test}/**/*.{js,json,md}\" --list-different",
-    "format": "prettier --write \"./packages/*/*/{src,bin,test}/**/*.{js,json,md}\"",
-    "precommit": "lint-staged",
+    "build": "gulp",
     "clean-test": "rimraf packages/core/integration-tests/.parcel-cache && rimraf packages/core/integration-tests/dist",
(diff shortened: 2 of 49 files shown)


package.json
```diff
@@ -11,25 +11,29 @@
...
-    "build": "lerna run build --stream",
-    "test": "lerna run test --stream",
-    "test-ci": "lerna run test-ci --stream",
-    "lint": "eslint . && prettier \"./packages/*/*/{src,bin,test}/**/*.{js,json,md}\" --list-different",
-    "format": "prettier --write \"./packages/*/*/{src,bin,test}/**/*.{js,json,md}\"",
-    "precommit": "lint-staged",
+    "build": "gulp",
     "clean-test": "rimraf packages/core/integration-tests/.parcel-cache && rimraf packages/core/integration-tests/dist",
     "clean": "yarn clean-test && lerna clean --yes && lerna exec -- rm -rf ./lib && yarn",
-    "postinstall": "yarn build"
```
padmaia: yay!

babel.config.js
```diff
@@ -0,0 +1,3 @@
+module.exports = {
+  presets: ['@parcel/babel-preset']
```
padmaia: Does this mean you can get rid of the `babelrc` files in workspace packages?

PR #3139, opened 2019-06-07
Strict mode prelude

mischnic:
Make prelude work in strict mode (needed for `<script type="module">` and when consuming a Parcel bundle in Rollup).

Technically, `self` and `global` should be enough to emulate `globalThis`. But I couldn't any exact data on the browser support of `self`.

#1401


packages/packagers/js/src/prelude.js
```diff
@@ -87,6 +86,12 @@ parcelRequire = function(modules, cache, entry, globalName) {
     ];
   };
 
+  /* eslint-disable no-undef */
+  if (typeof window !== 'undefined') window.parcelRequire = newRequire;
```
wbinnssmith: Here's an interesting take on this from @mathiasbynens: https://mathiasbynens.be/notes/globalthis#naive-polyfill

I wonder if we could get away with using something as hacky as that polyfill to detect the `globalThis` more accurately. Downside is it requires getters (es5), is hacky (temporarily mutates the object prototype), and we'd have to take extra precaution for ie10.

I think we should certainly use `globalThis` is it's available, and consider trying `this` in the case we're in a non-module context. We're unlikely to be in strict mode since we control the top-level of the script so it's likely accurate in any non-module context (which would implicitly be strict mode and have an empty `this`).

What worries me about using `this` is that from a quick check in node, the top-level `this` is an empty object in strict mode, not something nullish...

mischnic: Yeah, I know that post.

> And there you have it: the most horrifying polyfill you’ve ever seen! It completely defies the prevalent best practice to not modify objects you don’t own.

...

> But alas, this doesn’t work in strict mode functions or within JavaScript modules in non-browser environments (except those with globalThis support).

- without using `this`, it should work in strict mode ?
- `within JavaScript modules in non-browser contexts`: Node is supported via `global`, we don't need to support the V8 oder JSC shell.

packages/packagers/js/src/prelude.js
```diff
@@ -6,8 +6,7 @@
 // anything defined in a previous bundle is accessed via the
 // orig method which is the require for previous bundles
 
-// eslint-disable-next-line no-global-assign
-parcelRequire = function(modules, cache, entry, globalName) {
+(function(modules, cache, entry, globalName) {
   // Save the require from previous bundle to this closure if any
   var previousRequire = typeof parcelRequire === 'function' && parcelRequire;
```
mischnic: Whoops, the global object would be needed here as well (`global.parcelRequire`)

packages/packagers/js/src/prelude.js
wbinnssmith: In either case, someone might be using parcel to build a library that should be able to be loaded in a variety of environments.

Certainly for the application usecase, we don't need to worry since we don't need to support every environment, plus `this` will reliably be the global this in non-strict/non-module cases. But the building a library with Parcel case breaks both these assumptions -- `this` could be a CJS-style module if, e.g. webpack were bundling a parcel-built library, *and* we'd need to support whichever environment the library is running in.

One way of making this work is just making clear that parcel-built libraries can only run in a subset of environments, and not using `this` in our global this detection.

tl;dr I think we're good with what you're proposed, but we should still definitely use `globalThis` if it's there.

packages/packagers/js/src/prelude.js
wbinnssmith: yeah. Probably best just to make a local within our closure and use that throughout. Will clean up some of the repetition later too.

packages/packagers/js/src/prelude.js
mischnic: >  plus this will reliably be the global this in non-strict/non-module cases. But the building a library with Parcel case breaks both these assumptions 

But we can't use `this` inside of the IIFE wrapper either way.
We can't use this inside of an arrow function like in Mathias' post, because no code in the prelude is global (at least in this configuration).

> webpack were bundling a parcel-built library

I don't understand how this changes "the environment", the only thing we're concerned with is setting `globalThis.parcelRequire`.

> we should still definitely use globalThis if it's there.

👍

mathiasbynens: > We can't use this inside of an arrow function like in Mathias' post, because no code in the prelude is global (at least in this configuration).

Exactly. Sounds like you're hitting the "not running in the global scope" case described in my write-up. You'd need the full polyfill.

mischnic: In which situation would 
```js
const getGlobalThis = () => {
	if (typeof globalThis !== 'undefined') return globalThis;
	if (typeof self !== 'undefined') return self;
	if (typeof window !== 'undefined') return window;
	if (typeof global !== 'undefined') return global;
        // ?
};
```
not return the global object? `globalThis` if available, `self/window` for Worker/Browser, `global` for Node.
Are there any environments other than the JS engine shells that you are referring to in this line?
> Assume an environment where you cannot rely on the value of `globalThis`, `window`, `self`, `global`, ~or this;~

mathiasbynens: It's not just JS engine shells, it's also things like Rhino/RingoJS/Narwhal or [Deno](https://github.com/denoland/deno).

packages/packagers/js/src/prelude.js
```diff
@@ -6,10 +6,24 @@
...
+  var globalObject =
+    typeof globalThis !== 'undefined'
+      ? globalThis
+      : typeof self !== 'undefined'
+      ? self
+      : typeof window !== 'undefined'
+      ? window
+      : typeof global !== 'undefined'
+      ? global
+      : {};
```
devongovett: 😭

mischnic: Runnning out of options for a global object here, async loading wouldn't be supported in this case 🙈

devongovett: lol strict mode is annoying. Ideally we'd get rid of the global variable entirely... not sure how though.

devongovett: we could do it based on target maybe. i.e. browser target has `window`, node has `global`, etc.

mischnic: I don't see a way of getting the modules/state from an earlier `parcelRequire` call without globals.
Unless, loading an async module could somehow happen inside the closure of the initial `parcelRequire`, passing it over. But with the `<script>` injection that doesn't seem possible.

Issue #2813, opened 2019-03-18
Simple tilde resolution not working for me


Kurren123:
# 🐛 bug report

Simple tilde paths as described under [getting started](https://parceljs.org/module_resolution.html#~-tilde-paths) don't seem to be resolving.

## 🤔 Expected Behavior

A [repo of the error](https://github.com/Kurren123/parcel-js-bug). My folder structure:

```
node_modules
  xhook
    dist
      xhook.js
folder1
  index.html
index.html
package.json
```

And both `index.html` files have:

```html
<html>
  <body>
    <script src="~/node_modules/xhook/dist/xhook.js"></script>
  </body>
</html>
```

When I run `parcel folder1/index.html` it should build.

## 😯 Current Behavior

`parcel index.html` works fine but  `parcel folder1/index.html` returns:

```
D:\dev\parcel-test\folder1\node_modules\xhook\dist\xhook.js: ENOENT: no such file or directory, open 'D:\dev\parcel-test\folder1\node_modules\xhook\dist\xhook.js'
```

It seems to be treating the tilde as a relative path rather than looking for the folder containing `node_modules`.

## 💻 Code Sample

[Repository of the bug](https://github.com/Kurren123/parcel-js-bug).

## 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Parcel           | 1.12.2 |
| Node             | 11.12.0 |
| npm/Yarn         | npm 6.7.0 |
| Operating System | Windows 10 |

lustoykov:
What is the expected behaviour here? Do we want `~` to work same as in webpack? Means that a lookup is performed against node_modules to resolve the path? If yes, then the path should have been 
`<script src="~xhook/dist/xhook.js"></script>`. However, even in this case, parcel does not work as expected, but I could try to fix that if the specification is to be compliant with other bundlers? 

Think the relevant code is here 

https://github.com/parcel-bundler/parcel/blob/32f332782e830036c775dc7ad2631b5801719dc3/packages/core/parcel-bundler/src/Asset.js#L95-L105

mischnic:
`~/node_modules/...` is actually correct: https://parceljs.org/module_resolution.html#~-tilde-paths

> that if the specification is to be compliant with other bundlers?

That difference isn't ideal though

lustoykov:
I've looked a bit more into it and looks like the behavior is expected?

[...]

PR #3141, opened 2019-06-08
Make tilde resolving more intuitive (and implement RFC)

mischnic:
Related: #2813

Monorepo example where this caused issues (entrypoint: `client/src/index.js`):
```
project
|-package.json
|-node_modules
|-yarn.lock
|- client
   |- src
      |- index.js
   |- other.js
   |- package.json
```
`options.rootDir`: `project/client/src`
`options.projectRoot`: `project`

---

Resolving `~/other.js` from `project/client/src/index.js` would:
- start in the directory: `project/client/src` (dirname of dependent)
- stop because `rootDir` is reached.

*Result*: `project/client/src/other.js`. Doesn't exist

**Clearly, the resolving should stop in the projectRoot. (Change 1, this would resolve many of the problems in the linked issue already)**

---

Now, resolving `~/other.js` from `project/client/src/index.js` would:
- start in the directory: `project/client/src` (dirname of dependent)
- `project/client/node_modules` doesn't exist
- `project/node_modules` does exist (and `projectRoot` is reached)

*Result*: `project/other.js`. Doesn't exist

**Resolving should stop when the directory contains a package.json file (because of monorepos). (Change 2)**


wbinnssmith:
I realize this would be a breaking change, but can we make this always resolve to the projectRoot? The fact that it behaves unpredictably depending on where you use it across a monorepo is pretty unintuitive imo.

mischnic:
"The closest package root". The fact that `package.json` and `node_modules` don't always stay together in a monorepo is really just an implementation detail of lerna/yarn/...

> The fact that it behaves unpredictably depending on where you use it across a monorepo is pretty unintuitive imo.

If you want to get something from another package in a monorepo, you would use the package name, right? Not `~/packages/utils/....`.

Quoting the RFC:
> ### Tilde Paths
>```
>~/src/file.js
>```
>Only when used outside of node_modules directories, the ~ is replaced by an absolute path to the closest package root:
>```
>/path/to/app #(/package.json)
>```
>To form a path that looks like:
>```
>/path/to/app/src/file.js
>```

savovs:
I made a reproduction repo for a situation that might be helpful: https://github.com/savovs/parcel-tilde-bug

packages/resolvers/default/src/DefaultResolver.js
```diff
@@ -175,18 +175,19 @@ class NodeResolver {
...
       case '~':
         // Tilde path. Resolve relative to nearest node_modules directory,
-        // or the project root - whichever comes first.
+        // the nearest directory with package.json or the project root - whichever comes first.
         while (
-          dir !== this.options.rootDir &&
-          path.basename(path.dirname(dir)) !== 'node_modules'
+          dir !== this.options.projectRoot &&
+          path.basename(path.dirname(dir)) !== 'node_modules' &&
+          !(await fs.exists(path.join(dir, 'package.json')))
```
devongovett: Why would there not be a `package.json` file inside a `node_modules` directory? Seems like this would add a bunch of unnecessary checks.

mischnic: Do you mean that this change isn't needed at all or that this check should be simplified (not check for the package.json) if `dir` contains `node_modules`?

devongovett: Perhaps it could be optimized. For example, it the path contains `node_modules` at all, then just walk up to that point. Otherwise, fall back to looking for a `package.json`. This way, we don't check for `package.json` files at all if inside `node_modules`.

PR #3142, opened 2019-06-09
V2: Typescript Typechecking

DeMoorJasper:
# ↪️ Pull Request

This PR implements the TypeScript TypeChecking Validator

## ✔️ PR Todo

- [x] Write validator plugin type
- [x] Extract typechecking from `parcel-plugin-ts` into a validator plugin
- [x] Add examples


devongovett:
This is very cool. I agree. We should add a new plugin type for this. Then we can run the typechecker in the background so that it doesn’t block the build in development. As you said, this will be useful for things like linters as well.

DeMoorJasper:
@devongovett awesome, about config though do you think it would be a good idea or even possible to return a path as well? I'm not entirely sure TypeScript needs it but it's how it's stated in TS docs, so I guess it's supposed to be that.

Also I'm not entirely sure how to properly emit errors so the reporter only reports it once

devongovett:
Not sure what you mean by return a path...

For the errors, can you somehow filter the errors in the checker to only throw type errors rather than other types of error (e.g. syntax errors)? Only type errors are not already handled by the transformer.

Some other questions:

* What about resolving? How do we make sure the way the TS checker resolves and the way the TS transformer resolves are the same? I guess we'll be adding a TS resolver plugin, but we'll need the TS checker plugin to also use that.
* From looking at the code, @fathyb's plugin does a lot more than what is here. I'm not totally sure what exactly, but do we need to carry any of that over?

DeMoorJasper:
As far as I've been able to look through fathyb's code, it does some resolving (mainly for transform) and some enhancements to improve performance of typechecking (as it's insanely slow, but we'll add a custom plugin type for this so we can optimise it in core).
I think it should be sufficient to get the path of the tsconfig it's using to let ts do all the resolving by itself relative to this config.

About resolving I think typescript should be able to handle it. It's using the langService which should support all the advanced TS resolving already as far as I understand. Once transformers support all tsconfig resolving things it should be identical.

devongovett:
But parcel specific things like aliases won't work though right?

DeMoorJasper:
You need to configure those in tsconfig anyway, to get things like IDE integration working. But it currently won't work without the tsconfig config

devongovett:
IDE integration would be problematic with other languages like plain JS though too, and we still support aliases there...

devongovett:
Seems like we could probably convince vs code to support aliases defined in package.json. Tilde paths may be more contentious.

DeMoorJasper:
That would be awesome.

In case of typescript all these things (aliasing, absolute tilde) can be configured in tsconfig.

However it would be a big timesaver if vscode would support these things out of the box without the need of extra config, especially helpful for JS as you mentioned

DeMoorJasper:
Should be ready for review

packages/validators/ts-typechecker/src/typecheckValidator.js
```diff
@@ -0,0 +1,45 @@
...
+import LanguageServiceHost from './LanguageServiceHost';
+
+export default new Validator({
+  async validate({asset, options}) {
+    let ts = await localRequire('typescript', asset.filePath);
+    let config = {
+      config: await asset.getConfig(['tsconfig.json']),
+      baseDir:
+        path.dirname(
+          await findUp('tsconfig.json', {
```
mischnic: findUp returns undefined if it can't find a tsconfig.

DeMoorJasper: TypeScript already handles empty configs, but not sure if we should provide a custom default

DeMoorJasper: I also think findUp should just be a temporary solution and there should be some util that's shared with config for resolving config paths.
Fixed the potential issue anyway

mischnic: Well, the tests fail (even with that fix):
If the config above (from `getConfig`) is null, typescript throws an error.

DeMoorJasper: ah damn, will update it to default to `{}`

DeMoorJasper: Should all be fixed now and no longer even use findUp...

packages/core/core/src/RequestGraph.js
```diff
@@ -150,6 +153,9 @@ export default class RequestGraph extends Graph<RequestGraphNode> {
   async transform(request: AssetRequest) {
     try {
       let start = Date.now();
+      if (!(await realpath(request.filePath)).includes('node_modules')) {
+        await this.runValidate(request);
```
devongovett: Do we actually need to await this here or could we run this in the background? For development mode, it might be nice to not block the build based on type checking.

packages/core/core/src/RequestGraph.js
```diff
@@ -150,6 +153,9 @@ export default class RequestGraph extends Graph<RequestGraphNode> {
   async transform(request: AssetRequest) {
     try {
       let start = Date.now();
+      if (!(await realpath(request.filePath)).includes('node_modules')) {
```
devongovett: Maybe do this check inside the validator runner?

packages/core/core/src/ValidatorRunner.js
```diff
@@ -0,0 +1,176 @@
...
+import InternalAsset from './Asset';
+
+type Opts = {|
+  config: Config,
+  options: ParcelOptions
+|};
+
+const BUFFER_LIMIT = 5000000; // 5mb
+
+export default class validatorRunner {
```
devongovett: hmm a bunch of stuff in here seems duplicated from transformers... wonder what we can reuse.

packages/core/core/src/RequestGraph.js
DeMoorJasper: depends, if it throws errors it should probably be awaited as we'd otherwise have uncaught exception errors.

Although I think after the slack discussion it's safe to say we probably never wanna fail a build and just throw warnings.

packages/core/core/src/RequestGraph.js
DeMoorJasper: Sure will update it

packages/core/core/src/ValidatorRunner.js
DeMoorJasper: Most of it's copy paste from transformerrunner

flow-typed/npm/typescript_v3.3.x.js
```diff
@@ -0,0 +1,8762 @@
+declare module 'typescript' {
```
wbinnssmith: where did this come from? doesn't have the usual flow-typed digest on top 🤔

DeMoorJasper: https://github.com/flow-typed/flow-typed/blob/master/definitions/npm/typescript_v3.3.x/flow_v0.42.x-/typescript_v3.3.x.js

packages/core/core/src/RequestGraph.js
devongovett: Maybe we could just queue the validations here and then run them after the build is complete?

packages/validators/ts-typechecker/src/typecheckValidator.js
```diff
@@ -0,0 +1,41 @@
...
+export default new Validator({
+  async validate({asset, options}) {
+    let ts = await localRequire('typescript', asset.filePath);
+
+    let configNames = ['tsconfig.json'];
+    let configPath = await resolveConfig(asset.filePath, configNames);
+    let tsconfig = (await asset.getConfig(configNames)) || {};
+    let baseDir = configPath ? path.dirname(configPath) : options.projectRoot;
+
+    let parsedCommandLine = ts.parseJsonConfigFileContent(
```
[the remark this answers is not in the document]
DeMoorJasper: Yes unfortunately it will

packages/validators/ts-typechecker/src/LanguageServiceHost.js
```diff
@@ -0,0 +1,86 @@
...
+  getScriptVersion(fileName: string) {
+    return this.files[fileName] && this.files[fileName].version.toString();
+  }
+
+  getScriptSnapshot(fileName: string) {
+    if (!this.ts.sys.fileExists(fileName)) {
+      return;
+    }
+
+    const content = this.ts.sys.readFile(fileName);
```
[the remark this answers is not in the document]
DeMoorJasper: It’s more than just the current asset as it also does imports afaik so not sure how to do that.

packages/core/core/src/RequestGraph.js
DeMoorJasper: Not sure what the benefit of that would be unless we'd spawn a dedicated worker to run all typechecking. as that'll allow ts to keep a state and not reparse configs, recheck files and whatnot

devongovett: The benefit is faster builds in development. When changing code in development, we want to compile as quickly as possible and trigger HMR. TS can happen in the background after that. This means not using the workers for TS checking until after the build is complete.

devongovett: The idea of keeping state is interesting too. We could have a dedicated "validator" worker to run TS and other linters in maybe... I think that's how [fork-ts-checker-webpack-plugin](https://github.com/Realytics/fork-ts-checker-webpack-plugin) works, and fathy's plugin looks like it has its own worker too...

packages/validators/ts-typechecker/src/LanguageServiceHost.js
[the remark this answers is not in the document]
devongovett: Ah yeah ok

packages/core/core/src/RequestGraph.js
DeMoorJasper: Yes most plugins do it that way. I think it's the most performant way of doing this. Haven't experimented with running all workers with a TypeChecking state but I guess it wouldn't be much faster or even slower and just waste resources. I'll experiment with both once the TS PR is in and some SourceMap bugs are fixed.
This validator PR is not really high priority anyway and it was just a bit of an experiment to see if and how we'd support this.

PR #3146, opened 2019-06-12
Config invalidation

padmaia:
# ↪️ Pull Request

## Done

* Added config requests to the `RequestGraph` so that we can rebuild when either Parcel config or third party config changes.
* Transformations are invalidated when parcel config changes.
* Added support for nested parcel configs which allows you to have a specific configuration for a directory. This will only apply for transformers and resolvers as they operate on a file by file basis, whereas bundlers/packagers/optimizers/reporters  etc. work on a the project level. ( I have only implemented this behavior for transformations for now. If we still want it for resolvers, that will come in a future PR.)

## TODOs for future PRs

* Nested config support for resolving.
* Refactor existing plugins to use the new config loading model.
* Invalidation when plugins change versions.
* Cache garbage collection. One thing to note is that all of the things that may invalidate a cache key are now included in the cache key, which means there is less garbage collection than there was previously. I think it should be pretty straightforward to deal with this by returning all of the used cache keys with the transformed assets and the asset request can hold on to them and clean the entries up when it is removed from the graph. I need to think about how this would play with a distributed cache though, so I've held off on implementing it.

## 🚨 Test instructions

I've manually tested this with the simple example, but I need to add some integration tests.

## ✔️ PR Todo

- [x] Add some integration tests


packages/core/core/src/Config.js
```diff
@@ -1,224 +1,117 @@
 // @flow
```
padmaia: @wbinnssmith Since this is a public API, should it go in `src/public`? Also haven't thought through if we need to have a public and an internal version.

wbinnssmith: I think we should probably adhere to this more, and this would be a good first step :)

packages/core/core/src/loadParcelConfig.js
```diff
@@ -0,0 +1,315 @@
...
+import * as fs from '@parcel/fs';
+import {parse} from 'json5';
+import path from 'path';
+import {localResolve} from '@parcel/local-require';
+import assert from 'assert';
+
+type Pipeline = Array<PackageName>;
+type ConfigMap<K, V> = {[K]: V};
+
+export default async function loadParcelConfig(
```
wbinnssmith: This is so awesome. The prior class had mostly methods that didn't use `this`.

packages/core/utils/src/config.js
```diff
@@ -36,9 +37,10 @@ export async function resolveConfig(
 
   for (const filename of filenames) {
     let file = path.join(filepath, filename);
-    let exists = existsCache.has(file)
-      ? existsCache.get(file)
-      : await fs.exists(file);
+    let exists =
+      existsCache.has(file) && (!opts || !opts.noCache)
```
wbinnssmith: can this grow indefinitely? can't it be invalidated by files being removed?

wbinnssmith: `existsCache.has(file) && !opts?.noCache` maybe?

packages/core/core/src/AssetGraphBuilder.js
```diff
@@ -7,7 +7,7 @@ import type {Event} from '@parcel/watcher';
 
 import type Asset from './Asset';
 import AssetGraph from './AssetGraph';
-import type Config from './ParcelConfig';
+import type ParcelConfig from './ParcelConfig';
```
wbinnssmith: ❤️

packages/core/core/src/loadParcelConfig.js
```diff
@@ -42,48 +43,50 @@ export default async function loadParcelConfig(
   return parcelConfig;
 }
 
-export async function resolve(filePath: FilePath): Promise<?Config> {
+export async function resolve(filePath: FilePath) {
```
wbinnssmith: This is a personal thing and we should probably come to an understanding across Parcel, but I prefer to explicitly type function return types for exported functions.

packages/core/core/src/loadParcelConfig.js
```diff
@@ -42,48 +43,50 @@ export default async function loadParcelConfig(
...
+  configFile: ParcelConfigFile,
+  filePath: FilePath
+) {
+  let resolvedFile: ResolvedParcelConfigFile = {filePath, ...configFile};
+  let config = new ParcelConfig(resolvedFile);
+  let relativePath = path.relative(process.cwd(), filePath);
+  validateConfigFile(configFile, relativePath);
 
-  let extendedFiles: Array<FilePath> = [];
+  //let extendedFiles: Array<FilePath> = [];
```
wbinnssmith: delete?

packages/core/core/src/loadParcelConfig.js
```diff
@@ -42,48 +43,50 @@ export default async function loadParcelConfig(
...
+      ? configFile.extends
+      : [configFile.extends];
     for (let ext of exts) {
-      let resolved = await resolveExtends(ext, config.filePath);
-      extendedFiles.push(resolved);
-      let [baseConfig, moreExtendedFiles] = await readAndProcess(resolved);
-      extendedFiles = extendedFiles.concat(moreExtendedFiles);
-      config = mergeConfigs(baseConfig, config);
+      let resolved = await resolveExtends(ext, filePath);
+      //extendedFiles.push(resolved);
```
wbinnssmith: delete?

packages/core/core/src/Config.js
```diff
@@ -0,0 +1,124 @@
+// @flow
+import nullthrows from 'nullthrows';
+import path from 'path';
+
+import type {FilePath, PackageName} from '@parcel/types';
+
+type ConfigOpts = {|
+  searchPath: FilePath,
+  resolvedPath?: FilePath,
+  result?: any,
```
wbinnssmith: can we avoid this `any`?

packages/core/core/src/ConfigLoader.js
```diff
@@ -0,0 +1,81 @@
...
+        break;
+    }
+    devDeps.forEach(devDep => config.setDevDep(devDep));
+
+    config.addGlobWatchPattern('**/.parcelrc');
+
+    // let reliesOnLockFile = false;
+    for (let extendedFile of extendedFiles) {
+      // // ? Does this work for Windows
+      // if (extendedFile.includes('/node_modules/')) {
```
wbinnssmith: delete?

packages/core/core/src/RequestGraph.js
```diff
@@ -194,6 +233,62 @@ export default class RequestGraph extends Graph<RequestGraphNode> {
...
+    //   })
+    // );
+
+    return config;
+  }
+
+  async runConfigRequest(configRequest: ConfigRequest) {
+    let result = await this.configLoader.load(configRequest);
+    configRequest.result = result;
+    //this.addConfigResultToGraph(requestNode, result);
```
wbinnssmith: delete?

packages/core/core/src/loadParcelPlugin.js
```diff
@@ -0,0 +1,40 @@
...
+    throw new Error(
+      `The plugin "${pluginName}" is not compatible with the current version of Parcel. Requires "${parcelVersionRange}" but the current version is "${PARCEL_VERSION}".`
+    );
+  }
+
+  // $FlowFixMe
+  let plugin = require(resolved);
+  plugin = plugin.default ? plugin.default : plugin;
+  plugin = plugin[CONFIG];
+  // this.pluginCache.set(pluginName, plugin);
```
wbinnssmith: delete?

packages/core/core/src/RequestGraph.js
```diff
@@ -1,12 +1,15 @@
 // @flow strict-local
 import invariant from 'assert';
+//$FlowFixMe
+import {isMatch} from 'micromatch';
```
wbinnssmith: I can't believe there isn't a flow-typed definition for this 😲

packages/core/core/src/ConfigLoader.js
padmaia: Was keeping this because it will come back around when I get to the `// TODO:` on line 55, but I've got it in another branch so I'll clean it up

packages/core/core/src/Config.js
padmaia: Offline discussion for posterity:

Seems like we would have to add a Generic to Transformer and return `{result, devDeps, includedFiles, watchGlob}` and what not instead of passing in a config object that gets mutated. Something to explore, but not a blocker for this PR since `getConfig` currently returns `any` so there's no regression.

packages/core/utils/src/config.js
padmaia: Ideally anything that needs to be invalidated would be invalidated in the same way at the same time, which at this point is the `RequestIGraph`. I'm not really sure we should keep the `existsCache` at all. The only other things using it right now are `loadEnv` which only happens once (right?) and `installPackage`, which probably doesn't happen often enough to notice a performance improvement.

packages/core/core/src/Transformation.js
```diff
@@ -0,0 +1,499 @@
...
+          initialAsset.filePath,
+          asset.type,
+          pipeline
+        );
+      }
+
+      if (nextPipeline) {
+        let nextPipelineAssets = await this.runPipeline(nextPipeline, asset);
+        await Promise.all(
+          nextPipelineAssets.map(nextAsset => nextAsset.getCode())
```
wbinnssmith: Why do we do this? This is going to force reading entire streams that could be filled with large binary data.

padmaia: I chose to cache the at the pipeline layer because I felt it was easier to follow how intermediate assets can get cached. Also now that the cache keys have all of the invalidation information, it means we don't have can still invalidate individual pipeline results. I get that it's not ideal to commit it and then read it right away again. We should probably have a way we can commit the result of a pipeline without the content on the assets while still keeping the content on the in memory assets.

devongovett: I still don't understand what effect this has. Doesn't look like you're using the results of getting the code right?

Also, this won't work with binary assets for example. We have separate methods for those: `getCode`, `getBuffer`, `getMap`, etc. and only the transformer plugin will know which to call.

padmaia: > I still don't understand what effect this has. Doesn't look like you're using the results of getting the code right?

I guess I was thinking that `Asset.commit()` cleared code/content from the asset. Is it only the serialization that does that?

Issue #2158, opened 2018-10-17
Parcel 2: HTML Transform


devongovett:
Create the `@parcel/transform-html` package in `packages/transforms/html`. This package is a Parcel 2 transform plugin for HTML. It will be based on the `HTMLAsset` from Parcel 1. The main difference is that it will not be responsible for transforming with PostHTML, that will be a separate `@parcel/transform-posthtml` plugin.

The `@parcel/transform-html` package is responsible for the following things:

* Parsing with PostHTML if an AST is not already available from a previous transform.
* Collecting dependencies - e.g. images, scripts, css, etc.
* Creating sub-assets for inline scripts and styles to be processed by the pipeline, and re-inserting them back into the HTML
* Generating HTML if needed.

reebr:
Can we have any flag on dependencies collection to handle cases like #2637?

mischnic:
> Can we have any flag on dependencies collection to handle cases like #2637?

Just for future reference: excluding dependencies (e.g. from `src`) has these primaries use-cases:
- file will exist on a relative url in production
- preprocessing HTML for a server-used template.

reebr:
The same thing occurs on `<a>` tags. 😢

RobertWHurst:
@reebr The trouble with that is the pug does not escape that HTML after your pipe. It simply becomes HTML. If parcel ignores it then it's not treating the HTML output uniformly, which I actually think is more confusing. If something in pug is supposed to become HTML, as is the case here, I want parcel to process it as such.

wbinnssmith:
- [ ] Missing inline asset support

kwelch:
I have this extract and creating the proper assets for this, I just need to update the nodes within html asset. I have been swamped on a few thing and haven't had a chance to look into this since before the holiday. Hoping to look at it again this week.

devongovett:
Going to close this issue for alpha 1 and move inline asset support to #2158 in alpha 2.

PR #3148, opened 2019-06-14
[WIP] V2: Extract html inline assets

kwelch:
# ↪️ Pull Request
This adds extraction of inline assets from html files. 

fixes: #2158

## ✔️ PR Todo


packages/core/integration-tests/test/html.js
```diff
@@ -576,99 +576,110 @@ describe('html', function() {
...
     );
 
-    const bundleContent = (await outputFS.readFile(b.filePath)).toString();
-    assert(!bundleContent.includes('someArgument'));
+    assertBundles(b, [
+      {name: 'index.html', assets: ['index.html']},
+      {type: 'js', isInline: true, assets: ['index.html']},
+      {type: 'js', isInline: true, assets: ['index.html']},
+      {type: 'js', isInline: true, assets: ['index.html']},
+      {type: 'js', isInline: true, assets: ['index.html']}
```
kwelch: I find it odd that the bundles are here, but they do not output as expected.

packages/core/integration-tests/test/html.js
```diff
@@ -576,99 +576,110 @@ describe('html', function() {
...
+    ]);
+
+    let html = await outputFS.readFile(path.join(distDir, 'index.html'));
+    console.log(await outputFS.readdir(distDir));
+    console.log(html.toString());
```
kwelch: Need to remove these before merging, this is the best way to debug for now

packages/packagers/html/src/HTMLPackager.js
```diff
@@ -47,13 +47,56 @@ export default new Packager({
     }, []);
 
     let {html} = await posthtml([
-      insertBundleReferences.bind(this, bundles)
+      insertBundleReferences.bind(this, bundles),
+      replaceInlineAssetContent.bind(
+        this,
+        getAssetContent.bind(this, bundleGraph, () => {})
+      )
```
kwelch: This second plugin is a bit messy, but still determining the API I need.

packages/packagers/html/src/HTMLPackager.js
```diff
@@ -47,13 +47,56 @@ export default new Packager({
...
     ]).process(code);
 
     return {contents: html};
   }
 });
 
+async function getAssetContent(bundleGraph, getBundleResult, assetId) {
+  let inlineBundle: ?Bundle;
+  bundleGraph.traverseBundles((bundle, context, {stop}) => {
+    if (bundle.id.includes(assetId)) {
```
kwelch: I would appreciate any recommendations on a better way to determine find the bundle based on the `assetId`.

packages/packagers/html/src/HTMLPackager.js
```diff
@@ -47,13 +47,56 @@ export default new Packager({
...
+    return node;
+  });
+
+  for (let node of inlineNodes) {
+    node.attrs['is-inline'] = true;
+    node.attrs['new-content'] = await getAssetContent(
+      node.attrs['data-parcelId']
+    );
+    // remove attr from output
+    delete node.attrs['data-parcelId'];
```
kwelch: I am assuming we don't want to leave this.

packages/packagers/html/src/HTMLPackager.js
```diff
@@ -47,13 +47,56 @@ export default new Packager({
...
+  tree.walk(node => {
+    if (node.attrs && node.attrs['data-parcelId']) {
+      inlineNodes.push(node);
+    }
+    return node;
+  });
+
+  for (let node of inlineNodes) {
+    node.attrs['is-inline'] = true;
+    node.attrs['new-content'] = await getAssetContent(
```
kwelch: Adding debug attributes.

kwelch:
So, this has all the necessary parts. However, it is still inconsistent. I would appreciate some reviews on this and recommendations on how to make it better.

packages/packagers/html/src/HTMLPackager.js
devongovett: I’d look for bundle.getMainEntry().id.

devongovett: We could probably optimize this a bit too. Rather than traversing the whole graph to find each asset, you could use bundleGraph. findBundlesWithAsset and filter the result which should be quite a bit faster.

kwelch: Nice, I saw that and it looked like it needed an asset. I make an attempt to go that direction.

kwelch: I tried going to a `findBundlesWithAsset` approach, but the public `BundleGraph` doesn't have that function. 

`bundle.getMainEntry` should work though. However, it appears something recently changed on the v2 branch as I am getting some errors trying to find the inline bundles in the bundle graph.

kwelch:
I got it working consistently for inline JS. I need to run though it a few more times for the other html tests. 

I also need to add some core tests for the changes I made internally.

packages/core/core/src/PackagerRunner.js
```diff
@@ -130,7 +148,13 @@ export default class PackagerRunner {
       bundle,
       bundleGraph: new BundleGraph(bundleGraph, this.options),
       sourceMapPath: path.basename(bundle.filePath) + '.map',
-      options: this.pluginOptions
+      options: this.pluginOptions,
+      getBundleResult: (bundle: BundleType, bundleGraph: BundleGraphType) => {
```
devongovett: Not sure about this name. I guess `packageBundle` isn't clear that it also includes optimizing... `packageAndOptimizeBundle(bundle)`? `getInlineBundleContents`?

I also think we should only allow inline bundles here, and should throw an error if not. So maybe a name specific to that would be good?

packages/packagers/html/src/HTMLPackager.js
```diff
@@ -47,13 +49,73 @@ export default new Packager({
...
   }
 });
 
+async function getAssetContent(
+  bundleGraph: BundleGraph,
+  getBundleResult,
+  assetId
+): Promise<?Blob> {
+  let inlineBundle: ?Bundle;
+  bundleGraph.traverseBundles((bundle, context, {stop}) => {
```
devongovett: This could probably be optimized to traverse the graph once at the start of packaging, and make a map of inline asset keys to bundles. This way you'd avoid re-traversing the graph for each one. Probably not a big deal, but might be nice as a future improvement.

packages/packagers/html/src/HTMLPackager.js
```diff
@@ -47,13 +49,73 @@ export default new Packager({
...
+    let newContent = await getAssetContent(
+      bundleGraph,
+      getBundleResult,
+      node.attrs['data-parcel-key']
+    );
+
+    if (newContent != null) {
+      node.content = newContent;
+
+      if (node.attrs['parcel-should-delete-type'] === 'true') {
```
devongovett: Could this be done in the transformer directly?

packages/transformers/html/src/inline.js
```diff
@@ -49,12 +52,37 @@ export default function extractInlineAssets(
...
         parts.push({
           type,
           code: value,
-          inlineHTML: true,
+          uniqueKey: parcelKey,
+          isIsolated: true,
+          isInline: true,
           meta: {
             type: 'tag',
+            inlineKey: parcelKey,
```
devongovett: Do we need the key in two places? Maybe we could only use `uniqueKey` everywhere?

packages/transformers/html/src/inline.js
```diff
@@ -49,12 +52,37 @@ export default function extractInlineAssets(
...
+        }
+
+        // Inform packager to remove type, since CSS and JS are the defaults.
+        // Unless it's application/ld+json
+        if (
+          node.attrs &&
+          (node.tag === 'style' ||
+            (node.attrs.type && SCRIPT_TYPES[node.attrs.type] === 'js'))
+        ) {
+          node.attrs['parcel-should-delete-type'] = 'true';
```
devongovett: Do we need the type later? Or could you just delete it here instead?

packages/transformers/sass/src/SassTransformer.js
```diff
@@ -40,7 +40,11 @@ export default new Transformer({
       config = {};
     }
 
-    config.file = asset.filePath;
+    if (asset.isInline) {
```
devongovett: Not sure I fully understand this change. Could you elaborate?

packages/core/core/src/PackagerRunner.js
```diff
@@ -43,37 +49,49 @@ export default class PackagerRunner {
...
-    let {contents, map} = result;
+    let packaged = await this.package(bundle, bundleGraph);
+    let res = await this.optimize(
+      bundle,
+      bundleGraph,
+      packaged.contents,
+      packaged.map
+    );
+
+    let map = res.map ? await this.generateSourceMap(bundle, res.map) : null;
```
devongovett: Perhaps we should force inline source maps for inline bundles? Looks like right now the source map just gets thrown away by the HTMLPackager, but if it were generated inline it could work. I guess we don't support inline source maps at all right now, so we could add that later.

packages/packagers/html/src/HTMLPackager.js
kwelch: I had considered that. There is an odd edge where if the compiling messes us it changed the type and not the code. 

My logic for here was that the transformer did the gathering and the packager is doing mutations to the inline assets, but this line felt odd.

packages/transformers/html/src/inline.js
kwelch: The one in meta was a backup in case we found a way to determine uniqueness outside the transformer

packages/transformers/sass/src/SassTransformer.js
kwelch: To compile sass we cannot give it a file because the inline data is not a file, we just have the content to conform with their API, we need to pass it as data is we have a string of content instead of file. 

The indent further down could be done better but is based on the type and that we don't have a config for inline content. The expectation is that they are setting the type of the style to sass because they want indent syntax instead of scss with block syntax.

packages/core/core/src/PackagerRunner.js
kwelch: My thought was that we would add inline later once we have support for it. 

How best can we track that? Create an issue? Throw in a `TODO`

packages/core/core/src/PackagerRunner.js
kwelch: I like that idea, updated.

packages/packagers/html/src/HTMLPackager.js
kwelch: I am good with adding this. However, it is possible we have inline bundles that are a part of a different html we are parsing, right? 

That could still be a better performance hit. 🤔

packages/transformers/html/src/inline.js
kwelch: yea, I put the delete here for clarity.

packages/core/core/src/PackagerRunner.js
devongovett: An issue would be great.

kwelch: I created #3490, feel free to beef it up with any more context that I may be missing.

packages/transformers/sass/src/SassTransformer.js
devongovett: Huh, didn't realize we weren't passing the code in all the time. Seems like it would be problematic for even normal assets for SASS to be responsible for reading the file because it won't use our virtual FS. Can revisit that later.

devongovett:
@kwelch looks good! One test is failing on windows for some reason...

Pytal:
Hey what is the purpose of uniqueKey?

kwelch:
@Pytal 
uniqueKey was added to help identify the difference between the different assets that live in the same file. Prior to adding this the 4 content within the HTML page were being targeted as the same asset since fileName and type were the only used parameters to create the asset id. The asset id was used to create the content cache and when the assets were created they were all sent to the same buffer creating improper output.

Pytal:
How would I go about implementing this in a transformer when extracting css from js?

kwelch:
There are two parts to it really the packager is the part that is loading it back in, but the extraction takes places in the [HTMLTransformer](https://github.com/kwelch/parcel/blob/22a3fb600ff968f2c6fb244e0ab758ccd47ff32c/packages/transformers/html/src/inline.js#L18). The advantage of HTML is that you have attributes that you can use to store the asset data for retrieval in the package. I would have to think about it further, but I am not sure of a good way to accomplish that the same way in JS.

Pytal:
Would it be feasible to add an addDependencyOnPackage() method?

kwelch:
My concern is more around connecting the asset to the node in JS. With html I was able to write the ID to the HTML AST during transformer and then use it to retrieve and replace in the packager. There is no direct communication between the packager and transformer, thus you need to not only flag the asset to parcel, but also find a way to identify and connect the node on the AST to the asset that lives in parcel.

Pytal:
I feel that there may be a more

kwelch:
Cool, I am excited to see what that would look like. 

To make it similar to the HTML extraction that would be the next step to accomplish something in a similar way.

PR #3149, opened 2019-06-16
Fix terser config and reenable some tests

mischnic:
...


packages/optimizers/terser/src/TerserOptimizer.js
```diff
@@ -19,6 +19,7 @@ export default new Optimizer({
       );
     }
 
+    //TODO bundle.filePath should be path to entry asset
```
mischnic: The optimizer needs access to the entry asset to load the config

mischnic: @wbinnssmith ?

packages/optimizers/terser/src/TerserOptimizer.js
```diff
@@ -19,23 +18,24 @@ export default new Optimizer({
...
     }
 
-    let userConfig = await loadConfig(bundle.filePath, [
-      '.terserrc',
-      '.uglifyrc',
-      '.uglifyrc.js',
-      '.terserrc.js'
-    ]);
+    let userConfig = await bundle
+      .getEntryAssets()[0]
```
mischnic: Is this safe to do? When does a (javascript) bundle have multiple entry assets?

wbinnssmith: Why the change?

packages/optimizers/terser/src/TerserOptimizer.js
wbinnssmith: Oh, seeing this now.

As far as I know, `loadConfig` looks upward the filesystem from the path, so it's not exclusively the dist directory. That seems fine to me.

Why does the optimizer need the entry asset?

mischnic: > As far as I know, loadConfig looks upward the filesystem from the path, so it's not exclusively the dist directory. That seems fine to me.

What about this case?
```
project
 - dist
 - src
   - .terserrc
   - index.js
```

And the dist directory could also be outside of the project dir

> Why does the optimizer need the entry asset?

The search for the config needs to start from *some* source asset. The entry asset seems reasonable.

wbinnssmith: Yeah, I kind of like having the terser config relative to the dist directory. That way, you could even have multiple terser configs across targets.

I go back on forth on whether it makes sense to store terser config with the sources like that, but I can see it being surprising if we don't pick it up. @devongovett?

mischnic: > I kind of like having the terser config relative to the dist directory

This doesn't really work with the the split input/output fs PR...

mischnic: @wbinnssmith The config is loaded via `options.inputFS.readFile`.
"relative to the dist directory" isn't possible anymore

mischnic:
Where should the search for `.terserrc` start? The first entry asset of the bundle?

PR #3163, opened 2019-06-24
CHANGELOG: call out HMR breaking change in 1.12.x

louh:
Updating the HMR default behavior is a BREAKING change and should be more visible.

See also: https://github.com/parcel-bundler/parcel/pull/2676#issuecomment-505059049


devongovett:
I don't consider this a breaking change as it does not impact production builds, which is why I was ok with merging it. We should, however, document this on the website.

PR #3167, opened 2019-06-25
`parcelRequire is not defined`

### LINKED ISSUE #3290: parcelRequire.register incorrectly added
# 🐛 bug report

If the entry HTML file has a single JS file as a script (and no dynamic imports) (like the react-hmr example), a `parcelRequire.register("...",{})` is inserted incorrectly (and this calls throws an error as well because the prelude defining `parcelRequire` wasn't inserted).

This is because `bundleGraph.isAssetReferenced(asset)` is true (referenced by the HTML file), but the module only needs to be registered if it's (dynamically) imported by a JS bundle. So the check should really be something like an imaginary `bundleGraph.isAssetReferencedByType(asset, "js")`.

https://github.com/parcel-bundler/parcel/blob/baf40f7a331024bbfb3bb0394242fbd8d2928330/packages/shared/scope-hoisting/src/concat.js#L98-L103

```
Uncaught ReferenceError: parcelRequire is not defined:
	parcelRequire.register("1d046a184a4b101007cf746265aa38be",{})
```

CC @wbinnssmith I don't know if there's a way to get this information from the public bundleGraph?

mischnic:
Fixes #3290

## 😯 Current Behavior

Including a JS file from a html file (see test) (so the html is the parent of the JS) causes this test to fail (concat.js):
https://github.com/parcel-bundler/parcel/blob/41f106f3936b890d032f655f0fdb380d873352c5/packages/shared/scope-hoisting/src/concat.js#L83-L88

This works with only JS assets because some parent asset would include the `parcelRequire` prelude.

`Uncaught ReferenceError: parcelRequire is not defined`, because `parcelRequire.register()` is inserted (unnecessarily) and the `parcelRequire` isn't inserted.

## 💁 Possible Solution
`bundleGraph.isAssetReferenced(asset)` -> (imaginary) `bundleGraph.isAssetReferencedByJS(asset)`


packages/core/test-utils/src/utils.js
```diff
@@ -101,16 +101,21 @@ export async function bundle(
...
+  bundleOrGraph: BundleGraph | Bundle,
   globals: mixed,
   opts: {require?: boolean} = {}
 ): Promise<mixed> {
-  let bundles = [];
-  bundleGraph.traverseBundles(bundle => {
-    bundles.push(bundle);
-  });
+  let bundle: Bundle;
+  if (bundleOrGraph.traverseBundles) {
```
mischnic: Flow doesn't like this, it should really be `bundleOrGraph instanceof BundleGraph` but that is obviously not possible.

This change was needed because the `bundleGraph`'s entry bundle can be a html file but I want to execute the included JS script.

packages/core/core/src/BundleGraph.js
```diff
@@ -185,6 +185,29 @@ export default class BundleGraph {
...
+  isAssetReferencedByType(asset: Asset, type: string): boolean {
+    return this._graph
+      .getNodesConnectedTo(
+        nullthrows(this._graph.getNode(asset.id)),
+        'references'
+      )
+      .some(v => v.type === type);
+  }
+
+  hasParentBundleOfType(bundle: Bundle, type: string): boolean {
```
mischnic: There aws another issue in the function: it used`bundle.isEntry`, but it should really check: does `bundle` have any JS parent bundles (only the top level JS bundle should insert the prelude), not sure if this is the right way

mischnic:
The added methods in BundleGraph will likely might need to be cleaned up and/or modified to be more general?

`isAssetReferencedByAssetType` exists because only asynchronously dependent-on js assets need to be registered (is this always true?) and they're connected via an `references` edge (is this always true?):
BundleGraph
![BundleGraph](https://user-images.githubusercontent.com/4586894/62708519-e0219700-b9f3-11e9-818d-412595b72716.png

PR #3204, opened 2019-07-09
TypeError: Cannot read property 'TYPED_ARRAY_SUPPORT' of undefined #2833

idanilt:
# ↪️ Pull Request

Parcel uses node-libs-browser, this package is not maintained and causing those issue #2833
The fix is simply use the buffer and process up to date packages without the un-maintained container package
## 💻 Examples

Maintenance not a feature
In the future, it will be much better if this functionality will be extracted to babel plugin

## 🚨 Test instructions

Check for regression

## ✔️ PR Todo

- [ ] Added/updated unit tests for this change // little help do I need to do anything with those?
- [ ] Filled out test instructions (In case there aren't any unit tests) // little help do I need to do anything with those?


devongovett:
There are a lot more builtins than that. If we're going to move away from node-libs-browser, then we should support all of its features.

idanilt:
Thanks, I have notice, you can see all the builtins I added in the PR

I am failing on sum env, I couldn't find why (there is no summery) do you know why?

P.S.
I am very curios to why it's inside parcel and not inside a babel plugin (with default babel configuration)?

idanilt:
Thank you guys, your project was very refreshing
But the PR's not going to anywhere, so it waste of time

Hope you will sort your stuff soon, wish you best of luck

PR #3215, opened 2019-07-09
Plugin version invalidation

padmaia:
# ↪️ Pull Request
This PR adds functionality to invalidate transformations when previously used plugins change versions.

### DIFF
--- a/packages/core/core/src/Graph.js
+++ b/packages/core/core/src/Graph.js
@@ -85,6 +85,20 @@ export default class Graph<TNode: Node> {
     return this.outboundEdges.get(from).has(to);
   }
 
+  ensureConnection(fromNode: TNode, toNode: TNode) {
+    if (!this.hasNode(fromNode.id)) {
+      this.addNode(fromNode);
+    }
+
+    if (!this.hasNode(toNode.id)) {
+      this.addNode(toNode);
+    }
+
+    if (!this.hasEdge(fromNode.id, toNode.id)) {
+      this.addEdge(fromNode.id, toNode.id);
+    }
+  }
+
   getNodesConnectedTo(node: TNode): Array<TNode> {
     return Array.from(this.inboundEdges.get(node.id).values()).map(from =>
       nullthrows(this.nodes.get(from))
@@ -159,8 +173,19 @@ export default class Graph<TNode: Node> {
 
   // Update a node's downstream nodes making sure to prune any orphaned branches
   // Also keeps track of all added and removed edges and nodes
-  replaceNodesConnectedTo(fromNode: TNode, toNodes: Array<TNode>): void {
-    let childrenToRemove = new Set(this.outboundEdges.get(fromNode.id));
+  replaceNodesConnectedTo(
+    fromNode: TNode,
+    toNodes: Array<TNode>,
+    replaceFilter?: Function
+  ): void {
+    let outboundEdges = this.outboundEdges.get(fromNode.id);
+    let childrenToRemove = new Set(
+      replaceFilter
+        ? [...outboundEdges].filter(toNodeId =>
+            replaceFilter(nullthrows(this.nodes.get(toNodeId)))
+          )
+        : outboundEdges
+    );
     for (let toNode of toNodes) {
       let existingNode = this.getNode(toNode.id);
       if (!existingNode) {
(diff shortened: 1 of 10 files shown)


packages/core/core/src/Graph.js
```diff
@@ -85,6 +85,20 @@ export default class Graph<TNode: Node> {
     return this.outboundEdges.get(from).has(to);
   }
 
+  ensureConnection(fromNode: TNode, toNode: TNode) {
+    if (!this.hasNode(fromNode.id)) {
```
wbinnssmith: if this branch is taken, won't this result in an island in the graph?

packages/core/core/src/Graph.js
```diff
@@ -85,6 +85,20 @@ export default class Graph<TNode: Node> {
     return this.outboundEdges.get(from).has(to);
   }
 
+  ensureConnection(fromNode: TNode, toNode: TNode) {
```
wbinnssmith: I don't know how I feel about this method in general. It feels like using it is a sign that we're not sure the graph is in a good state.

Also, re: the name, `ensureConnection` feels like it should just make sure that the two nodes are connected in the graph (that behavior is pretty much what `addEdge` does now, as currently it's okay to re-add an edge). Something with this name actually adding missing nodes to the graph feels unexpected.

Can this behavior be moved into the call sites that are actually uncertain about the graph state? I feel like keeping this around might encourage bad behavior.

packages/core/core/src/resolveOptions.js
```diff
@@ -32,15 +32,25 @@ export default async function resolveOptions(
...
       '.git',
       '.hg'
-    ])) || path.join(process.cwd(), 'index')
-  );
+    ])) || path.join(process.cwd(), 'index');
+
+  let lockFile = null;
+  let rootFileName = path.basename(projectRootFile);
+  if (
+    rootFileName === 'yarn.lock' ||
```
wbinnssmith: maybe extract these into a constant list to be shared with the lines above?

packages/core/types/index.js
```diff
@@ -173,7 +173,8 @@ export type ParcelOptions = {|
   logLevel: LogLevel,
   rootDir: FilePath,
   targets: Array<Target>,
-  projectRoot: FilePath
+  projectRoot: FilePath,
+  lockFile?: FilePath
```
wbinnssmith: This should probably be `lockFile: ?FilePath`, as we produce this object and can guarantee the key.

packages/core/core/src/Transformation.js
```diff
@@ -201,12 +210,15 @@ export default class Transformation {
     );
 
     // TODO: sort
-    let configHashesId = JSON.stringify(
-      [...configs].map(([, config]) => config.resultHash)
+    let configsId = JSON.stringify(
+      [...configs].map(([, {resultHash, devDeps}]) => ({
```
wbinnssmith: nit: maybe a list of tuples rather than a list of objects which lack guaranteed key order

packages/core/core/src/Graph.js
```diff
@@ -159,8 +173,19 @@ export default class Graph<TNode: Node> {
 
   // Update a node's downstream nodes making sure to prune any orphaned branches
   // Also keeps track of all added and removed edges and nodes
-  replaceNodesConnectedTo(fromNode: TNode, toNodes: Array<TNode>): void {
-    let childrenToRemove = new Set(this.outboundEdges.get(fromNode.id));
+  replaceNodesConnectedTo(
+    fromNode: TNode,
+    toNodes: Array<TNode>,
+    replaceFilter?: Function
```
wbinnssmith: `replaceFilter?: (TNode) => boolean`

packages/core/core/src/Graph.js
padmaia: >  It feels like using it is a sign that we're not sure the graph is in a good state.
I'm using it for the case "These two nodes may already be connected, but if they're not, connect them". I wouldn't call that a bad state. `replaceNodesConnectedTo` does this as well, but it also does some stuff I don't want like overwriting the value of the node that already existed and replacing all of the other nodes unless I provide a function. I guess `addEdge` ensures against duplicate edges now (it didn't before, when I originally wrote this code). I could remove the abstraction and change the call sites back to:

```
if (!this.hasNode(toNode)) {
  this.addNode(toNode)
}

this.addEdge(fromNodeId, toNodeId)
```

but I don't totally understand what makes the abstraction dangerous.

packages/core/core/src/Transformation.js
padmaia: I think I should probably just build one big object and call `md5FromObject` which guarantees a sorted key order.

PR #3220, opened 2019-07-11
start server when --target=electron

zeevl:
This starts the web server when `--target=electron|browser`.   It was originally the case, but broken with [this pr](https://github.com/parcel-bundler/parcel/commit/9064b3b6b34cba08d6c33e5d88298485b1ee87f7).

Tests included. 

Unfortunately, I still haven't quite been able to get HMR working with my electron + typescript + react project.   It fails with not finding `tslib`, which AFIKT is an issue with the way parcel overwrites `require`.. but ran out of time to diagnose. 

Hopefully this helps someone, nonetheless. :)

### DIFF
--- a/packages/core/integration-tests/test/server.js
+++ b/packages/core/integration-tests/test/server.js
@@ -264,4 +264,21 @@ describe('server', function() {
     logger._log.restore();
     logger.setOptions({logLevel: 3});
   });
+
+  it('should start the dev server when --target=electron', async function() {
+    let b = await bundler(
+      path.join(__dirname, '/integration/html/index.html'),
+      {
+        target: 'node'
+      }
+    );
+
+    server = await b.serve(0);
+
+    let data = await get('/');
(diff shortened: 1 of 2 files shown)


jakoblaegdsmand:
just my two cents here ... AFAIK you don't need to start the development server to get HMR support. The HMR header is included in the javascript files, so in electron you can load the files using a file:// url and HMR will work. I do it like this in a project of mine.

zeevl:
@jakoblaegdsmand I don't believe that works.   
If you have parcel build the js directly, electron doesn't pick up the changes without a full reload.   

If parcel builds the html and it's loaded as a file://, updated js files are referenced using `src='/app123.js'`, which resolves to `file:///app123.js`.

See here - https://github.com/zeevl/parcel-react-electron

DeMoorJasper:
Gonna close this as it's a breaking change and Parcel 1 no longer receives new features

PR #3223, opened 2019-07-12
Support loading js bundles in web worker

### LINKED ISSUE #1654: Support code splitting in web workers
We currently don't support code splitting in web workers via `import()`, because the bundle loader for JS relies on the DOM. We should extend this to also support web workers.

This will allow moving common dependencies between workers, or even between workers and the main process to a separate bundle so it only gets loaded once. See https://twitter.com/DasSurma/status/1013489346090012672. cc. @surma.

In order to be consistent about how `require()` works, we won't make it automatically create a split point when called in a worker, as that would cause the code to potentially be removed from another bundle. But if you use `import()` to load the shared module in both the main bundle and a worker, the module would be placed into its own bundle and loaded in both places.

Related: #758

trentmwillis:
# ↪️ Pull Request

Updates the browser js-loader to provide a code path for loading in Web Workers. Fixes #1654 (partially, see below).

However, does _not_ change any of the bundling behavior. For the use case in the app I work on, the current bundling behavior works sufficiently. Might need follow-up PRs to address the other concerns in #1654.

Also, it's worth noting we could theoretically use the new code path for bundles in the main thread as well, but the `<script>` approach provides a better experience in the DevTools (and seems more "correct" to me).

## 💻 Examples

None.

## 🚨 Test instructions

Tested locally by linking into a project. Could not find where unit/integration tests for this would go in the repo, but would be happy to add them with a bit of guidance on that front!

## ✔️ PR Todo


wbinnssmith:
hi @trentmwillis! Thanks for sending this by :)

We're spending most of our effort on Parcel 2 these days and are generally putting a hold on new features in Parcel 1. It's up to @devongovett and @DeMoorJasper, but maybe we could merge this to unblock your work and forward port any changes to v2.

Neat to hear that you're using Parcel at Netflix. Would be awesome to hear more about what you need from it and how we can help in Parcel 2 😃

Re: the implementation: using the function constructor probably won't work in a lot of CSP environments (eval-like things often being disallowed). Is there another way of dynamically loading code in a worker?

devongovett:
Thanks for contributing @trentmwillis! 🙏 

The one thing I'm worried about with using XmlHttpRequest is that it will require CORS headers on scripts. Maybe for workers we could consider using `importScripts` instead which wouldn't have this limitation. What do you think?

trentmwillis:
Hi @wbinnssmith thanks for the info. For context, this is _not_ actually blocking our work. We'd like to be able to code-split + dynamic import in our Web Workers, but we can also import them statically. It's primarily a perf enhancement.

So far Parcel has been great! It's "just worked" for everything we needed up to this issue.

I had not considered the CSP constraint. Other possible approaches, that I know of, would be:
1. Use `eval`. But that is likely going to be in the same boat as `new Function`.
2. Use `importScripts`. That _should_ work, now that I think about it. I had not tried it originally because I was unfamiliar with exactly how the bundle needed to be evaluated once loaded and, thus, I more-or-less based the implementation off the Node loader.

I'll wait for @devongovett or @DeMoorJasper to chime in with if I should continue pursuing this. FWIW, I'd be happy to help land this in V1 as well as V2.

trentmwillis:
Updated to `importScripts`. Tested in our app and it works!

trentmwillis:
Let me know if you'd like me to port this to Parcel 2 :)

DeMoorJasper:
@trentmwillis ow yeah, please port it to Parcel 2.

Pretty sure we're no longer planning to update Parcel 1

trentmwillis:
Okay, I'll see about porting this sometime soon and will close this PR.

PR #3228, opened 2019-07-14
Use v8 serializer instead of JSON

devongovett:
This updates the serializer to use the native v8 serializer instead of JSON for the cache and worker IPC. This has several benefits:

* Many more native JS types are supported including Maps, Sets, Dates, etc.
* Reference integrity is maintained. Multiple references pointing to the same object will only be serialized once, and the reference will still point to the same object when deserialized.
* Cyclic objects are supported.
* It should be significantly smaller in many cases due to duplicate objects only being serialized once and the binary format instead of plain text.
* A bunch of unnecessary copying was removed in various places because we can serialize maps and sets directly now instead of converting to and from arrays. Might be a bit faster.

One downside is that the current worker farm has to convert the serialized buffers to base64 and back since node only supports sending strings over IPC (it's JSON based). When we swap to using node's new worker threads instead of processes this won't be an issue (we'll actually be able to do zero-copy transfers).

### DIFF
--- a/packages/core/cache/src/Cache.js
+++ b/packages/core/cache/src/Cache.js
@@ -6,7 +6,6 @@ import type {FilePath} from '@parcel/types';
 
 import * as fs from '@parcel/fs';
 import {createReadStream, createWriteStream} from 'fs';
-import invariant from 'assert';
 import path from 'path';
 import logger from '@parcel/logger';
 import {serialize, deserialize, registerSerializableClass} from '@parcel/utils';
@@ -20,17 +19,7 @@ export default class Cache {
     this.dir = cacheDir;
   }
 
-  static deserialize(opts: {cacheDir: FilePath}) {
-    return new Cache(opts.cacheDir);
-  }
-
-  serialize() {
-    return {
-      cacheDir: this.dir
-    };
-  }
-
-  _getCachePath(cacheId: string, extension: string = '.json'): FilePath {
+  _getCachePath(cacheId: string, extension: string = '.v8'): FilePath {
     return path.join(
       this.dir,
       cacheId.slice(0, 2),
@@ -53,9 +42,7 @@ export default class Cache {
 
   async get(key: string) {
     try {
-      let data = await fs.readFile(this._getCachePath(key), {encoding: 'utf8'});
-
-      invariant(typeof data === 'string');
+      let data = await fs.readFile(this._getCachePath(key));
       return deserialize(data);
     } catch (err) {
       if (err.code === 'ENOENT') {
--- a/packages/core/core/src/Asset.js
+++ b/packages/core/core/src/Asset.js
@@ -42,23 +42,15 @@ type AssetOptions = {|
   ast?: ?AST,
   map?: ?SourceMap,
   mapKey?: ?string,
-  dependencies?: Iterable<[string, Dependency]>,
-  connectedFiles?: Iterable<[FilePath, File]>,
+  dependencies?: Map<string, Dependency>,
+  connectedFiles?: Map<FilePath, File>,
   isIsolated?: boolean,
   outputHash?: string,
   env: Environment,
   meta?: Meta,
   stats: Stats,
-  symbols?: Map<Symbol, Symbol> | Array<[Symbol, Symbol]>,
+  symbols?: Map<Symbol, Symbol>,
   sideEffects?: boolean
-|};
-
-type SerializedOptions = {|
-  ...AssetOptions,
-  ...{|
-    connectedFiles: Array<[FilePath, File]>,
-    dependencies: Array<[string, Dependency]>
-  |}
 |};
 
 export default class Asset {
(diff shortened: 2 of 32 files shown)


packages/core/cache/src/Cache.js
```diff
@@ -20,17 +19,7 @@ export default class Cache {
...
-  }
-
-  serialize() {
-    return {
-      cacheDir: this.dir
-    };
-  }
-
-  _getCachePath(cacheId: string, extension: string = '.json'): FilePath {
+  _getCachePath(cacheId: string, extension: string = '.v8'): FilePath {
```
wbinnssmith: one potential downside of serializing this way when it comes to the on-disk cache is that it's not as easily introspectable for debugging without some kind of tool, as it has been as json text.

Maybe at some point we could build something that reads these in, deserializes with v8, and displays some kind text representation of it (or maybe something interactable like browser devtools 😃)

packages/core/core/src/Asset.js
```diff
@@ -101,38 +93,38 @@ export default class Asset {
...
     this.env = options.env;
     this.meta = options.meta || {};
     this.stats = options.stats;
-    this.symbols = new Map(options.symbols || []);
+    this.symbols = options.symbols || new Map();
     this.sideEffects = options.sideEffects != null ? options.sideEffects : true;
   }
 
-  serialize(): SerializedOptions {
+  static deserialize(opts: AssetOptions) {
```
wbinnssmith: is this still necessary?

packages/core/core/src/Environment.js
```diff
@@ -70,6 +70,11 @@ export default class Environment implements IEnvironment {
   }
 
   merge(env: ?EnvironmentOpts) {
+    // If merging the same object, avoid copying.
```
wbinnssmith: Good catch. This has probably been the common case?

packages/core/utils/src/serializer.js
```diff
@@ -16,37 +17,160 @@ export function registerSerializableClass(name: string, ctor: Class<*>) {
...
+        processKey(key, object[key]);
       }
     }
 
-    return serialized;
-  });
+    return result;
+  };
+
+  let mapped = memoizedFn(object);
```
wbinnssmith: could just inline this below

packages/core/utils/src/serializer.js
```diff
@@ -16,37 +17,160 @@ export function registerSerializableClass(name: string, ctor: Class<*>) {
...
+    return Object.create(
+      Object.getPrototypeOf(object),
+      Object.getOwnPropertyDescriptors(object)
+    );
+  }
+
+  return object;
+}
+
+function mapObject(object: any, fn: (val: any) => any, preOrder = false) {
```
wbinnssmith: `function mapObject<T, U>(object: T, fn: (val: T) => U, preOrder = false) {` maybe?

the generic types without a constraint should behave like `mixed`, which is probably all you need, and can eliminate the `any`s here.

packages/core/core/src/Asset.js
devongovett: unfortunately yes if we want to exclude `code`, `map`, and `ast` from the cache

packages/core/core/src/Environment.js
devongovett: Yep! And it's resulted in a lot of duplicate objects everywhere

packages/core/cache/src/Cache.js
devongovett: That's true. I think a tool to inspect the cache might be useful for debugging anyway. Just the file structure of it is somewhat confusing already. Would be better to have a tool that allowed inspecting the cache based on original asset paths or something.

packages/core/utils/src/serializer.js
```diff
@@ -16,37 +17,160 @@ export function registerSerializableClass(name: string, ctor: Class<*>) {
...
+  let mapped = memoizedFn(object);
+  return walk(mapped, mapped === object);
+}
+
+export function serialize(object: any): Buffer {
+  let mapped = mapObject(
+    object,
+    value => {
+      let serialized = value;
+      if (value && typeof value.serialize === 'function') {
```
chocolateboy: If this is ever used to serialize user-defined objects/classes, it would be safer to use a symbol e.g.:

```javascript
const SERIALIZE = Symbol()

// or (for use outside of Parcel)

const SERIALIZE = Symbol.for('parceljs.serialize')

// ...

if (value && typeof value[SERIALIZE] === 'function') { ... }
``` 
Even if the hook is only used internally, it might still better communicate the intention and avert future surprises.

## See Also

- https://www.keithcirkel.co.uk/metaprogramming-in-es6-symbols/#3-giving-developers-ability-to-add-hooks-to-their-objects-through-your-api
- https://nodejs.org/api/util.html#util_util_inspect_custom

devongovett: That's a good suggestion. We could also only allow a `serialize` method for constructors that are registered with the serializer.

packages/core/utils/src/serializer.js
devongovett: the result is used twice in the next line...

packages/core/utils/src/serializer.js
devongovett: I experimented but couldn't get it to work... :/ Not really sure the value of it anyway - it really could be anything.

PR #3229, opened 2019-07-14
Implement node worker_threads backend

devongovett:
This builds on #3228 and implements a worker farm backend based on the node 10 [worker_threads](https://nodejs.org/api/worker_threads.html) module. It will still continue to fall back to child processes on older versions of node.

Most of the implementation is the same, it just abstracts away the details into backends that each implement the same interface. The worker threads implementation should be faster in theory (I haven't fully tested yet) since it is much faster to spin up threads than processes, and the memory space is shared which means that messages are not copied between threads as they are with processes. Instead, [transferrable objects](https://developer.mozilla.org/en-US/docs/Web/API/Transferable) are used to implement zero-copy array buffer sharing.

Some additional testing is still needed to determine the performance characteristics of threads vs processes on a larger app. We also should determine whether to adjust some options depending on if threads are available, such as using the local worker or not, and warming up the workers or not. I'm not sure if workers share a require cache.

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -6,7 +6,8 @@
 <PROJECT_ROOT>/.nyc_output/.*
 .*/tmp/**/.*
 .*/.parcel-cache/.*
-.*/dist/.*
+<PROJECT_ROOT>/packages/core/integration-tests/dist/**
+<PROJECT_ROOT>/packages/core/integration-tests/test/input/**
 
 [include]
 
--- a/.flowconfig
+++ b/.flowconfig
@@ -6,6 +6,7 @@
 <PROJECT_ROOT>/.nyc_output/.*
 .*/tmp/**/.*
 .*/.parcel-cache/.*
+.*/dist/.*
 
 [include]
 
--- a/flow-libs/worker_threads.js.flow
+++ b/flow-libs/worker_threads.js.flow
@@ -1,0 +1,143 @@
+// @flow
+
+// Derived from the TypeScript definitions https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/worker_threads.d.ts
+// TODO: contribute this to flow core
+
+declare module "worker_threads" {
+  declare var isMainThread: boolean;
+  declare var parentPort: null | MessagePort;
+  declare var threadId: number;
+  declare var workerData: any;
+
+  declare class MessageChannel {
+    +port1: MessagePort;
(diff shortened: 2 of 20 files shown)


packages/core/integration-tests/test/html.js
```diff
@@ -414,7 +414,7 @@ describe('html', function() {
     ]);
   });
 
-  it('should support bundling HTM', async function() {
+  it.skip('should support bundling HTM', async function() {
```
devongovett: Skipping this test for now. It seems like the change to fix it from #3118 somehow got overwritten in a merge...

.flowconfig
```diff
@@ -6,6 +6,7 @@
 <PROJECT_ROOT>/.nyc_output/.*
 .*/tmp/**/.*
 .*/.parcel-cache/.*
+.*/dist/.*
```
wbinnssmith: won't this capture any dist files in `node_modules`? There are a couple of situations where we rely on the published flow sources for their typings, rather than using definition files.

flow-libs/worker_threads.js.flow
```diff
@@ -0,0 +1,143 @@
+// @flow
```
wbinnssmith: 😍

packages/core/workers/src/WorkerFarm.js
```diff
@@ -350,7 +341,13 @@ export default class WorkerFarm extends EventEmitter {
   }
 
   static isWorker() {
-    return process.send && require.main.filename === require.resolve('./child');
+    try {
```
wbinnssmith: this is leaking a bit about the backend into the WorkerFarm.

I wonder if `Worker` could accept a backend as an argument and it could be determined in `WorkerFarm` instead. That way, we could also use something like `WorkerImpl.isWorker()` here. This would allow us to easily scale beyond these two backends too.

devongovett: Yeah I was thinking about that too. Would allow for additional backends like web workers as well.

.flowconfig
devongovett: hmm yeah I guess so. This was because I was tired of flow going crazy when I ran the tests and it created a lot of files.

devongovett:
Updated to make the backend an option to the WorkerFarm and refactor things a bit to clean up how the backends are loaded.

PR #3232, opened 2019-07-14
V2: Typescript Typechecking

DeMoorJasper:
# ↪️ Pull Request

## ✔️ PR Todo


packages/validators/ts-typechecker/src/typecheckValidator.js
```diff
@@ -0,0 +1,43 @@
...
+
+let langServiceCache = {};
+
+export default new Validator({
+  async validate({asset, options}) {
+    let ts = await localRequire('typescript', asset.filePath);
+
+    let configNames = ['tsconfig.json'];
+    // $FlowFixMe
+    let configPath = await resolveConfig(asset.filePath, configNames);
```
devongovett: todo: another plugin type to add a `loadConfig` method to. cc. @padmaia

packages/validators/ts-typechecker/src/typecheckValidator.js
```diff
@@ -0,0 +1,43 @@
...
+
+    if (!langServiceCache[configPath]) {
```
devongovett: Does this cache ever need to be invalidated? i.e. when the config changes? Can be part of the overall config redesign later.

packages/core/core/src/Validation.js
```diff
@@ -0,0 +1,214 @@
...
+      },
+      sideEffects: sideEffects
+    });
+  }
+
+  async loadPipeline(filePath: FilePath): Promise<Pipeline> {
+    let configRequest = {
+      filePath,
+      meta: {
+        actionType: 'validation'
```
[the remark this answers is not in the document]
DeMoorJasper: Should probably get moved into run

packages/core/core/src/Validation.js
```diff
@@ -0,0 +1,214 @@
...
+  options: ParcelOptions
+|};
+
+class Pipeline {
+  validators: Array<Validator>;
+  configs: ConfigMap;
+  options: ParcelOptions;
+  resolverRunner: ResolverRunner;
+  generate: GenerateFunc;
+  postProcess: ?PostProcessFunc;
```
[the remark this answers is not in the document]
DeMoorJasper: I'll simplify it

packages/core/core/src/Validation.js
```diff
@@ -0,0 +1,214 @@
...
+          sourcePath: from
+        })
+      )).filePath;
+    };
+
+    for (let validator of this.validators) {
+      await validator.validate({
+        asset: new MutableAsset(asset),
+        options: this.options,
+        resolve
```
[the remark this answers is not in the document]
DeMoorJasper: Not sure, Would be nice to overwrite the real TS resolver with whatever resolver Parcel uses. Although not sure if it's a good idea because that might give unexpected behaviour compared to running tsc

packages/core/core/src/Validation.js
```diff
@@ -0,0 +1,214 @@
...
+      });
+    }
+  }
+}
+
+async function summarizeRequest(
```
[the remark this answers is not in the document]
DeMoorJasper: Sure

packages/validators/ts-typechecker/src/typecheckValidator.js
DeMoorJasper: If it needs to consider the changes while running it definitely needs a way to invalidate this

packages/core/core/src/Validation.js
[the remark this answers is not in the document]
DeMoorJasper: I guess so, don't completely understand how the config works at the moment

packages/core/core/src/RequestGraph.js
```diff
@@ -248,6 +264,19 @@ export default class RequestGraph extends Graph<RequestGraphNode> {
     }
   }
 
+  async validate(requestNode: AssetRequestNode) {
+    try {
+      await this.runValidate({
+        request: requestNode.value,
+        loadConfig: this.loadConfigHandle,
+        parentNodeId: requestNode.id,
```
padmaia: So the parentNodeId is provided so the config requests can attach themselves to the asset request node, so that if any of the config requests are invalidated, the asset request will be invalidated. Seems like if the typescript config changes, it would trigger the transformation to run again. Something like typescript config changing might warrant the transformation running again, but not something like eslintrc. Not sure if it's a huge issue because if the relevant config for a transformation doesn't change we'll just pick up from the cache instead of running the transformers. Would be better if we could figure out a way to not trigger all of that though.

packages/core/core/src/Validation.js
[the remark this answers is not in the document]
devongovett: cc. @padmaia I think it needs to return the names of the validator plugins maybe? https://github.com/parcel-bundler/parcel/blob/v2/packages/core/core/src/ConfigLoader.js#L42-L49

packages/core/core/src/Validation.js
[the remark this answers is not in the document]
devongovett: Good question...

packages/core/core/src/RequestGraph.js
DeMoorJasper: Not sure I completely understand? Currently validation isn't really cacheable and should run in each subsequent build no matter if there is cache available. Unless we'd figure out a way to return the errors/warnings instead of letting the validator throw them?

I just copied the code from transforms, wasn't sure what most of it was used for but it's all being used somewhere afaik

packages/validators/ts-typechecker/src/typecheckValidator.js
DeMoorJasper: Added an object hash so this should work now

packages/core/core/src/Validation.js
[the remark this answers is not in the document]
DeMoorJasper: Removed the resolve function for now

packages/core/core/src/Validation.js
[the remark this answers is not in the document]
DeMoorJasper: I updated it to return names of validators as it's like that for transforms.

@padmaia is this how it's supposed to be?

packages/core/core/src/RequestGraph.js
padmaia: The `loadConfig` function is passed in from the main process so that the results can be stored in the graph for both a faster retrieval when asking for that same config again and also so that the transformation can be invalidated when the config is invalidated. I was thinking that validator plugins would be using `loadConfig` as well since it was being passed in, but it looks like it is only being used for loading parcel config right now. If validator plugins were using it the way it is now, then a change in their config would invalidate the transformation causing it to run again. The configs actually used in the transformation wouldn't change though, so it would just end up using what it found in the cache.

Ideally validator and other plugins besides transformers should be using `loadConfig`, so that we can avoid rework of resolving configs between processes and so that we don't have to quit the process when we change configs. It would be a little weird if you only had to restart parcel when you change certain types of configs.

How I would see this working for validation would probably be something like the `RequestGraph` having an additional node type for a `validation_request`. It would be invalidated if the original file changes, the validator plugins change, or any config the plugins used change. This way changes to transformer plugins/config wouldn't trigger validations and vice versa.

packages/core/core/src/Validation.js
```diff
@@ -0,0 +1,99 @@
...
+      }
+    };
+
+    let config = await this.loadConfig(configRequest);
+    let parcelConfig = nullthrows(config.result);
+
+    let validators = await parcelConfig.getValidators(this.request.filePath);
+    for (let validator of validators) {
+      await validator.validate({
+        asset: new MutableAsset(asset),
```
devongovett: probably shouldn't be mutable for validators

packages/core/core/src/Validation.js
```diff
@@ -0,0 +1,99 @@
...
+    this.configRequests = [];
+    this.loadConfig = configRequest => {
+      this.configRequests.push(configRequest);
+      return loadConfig(configRequest, parentNodeId);
+    };
+    this.options = options;
+  }
+
+  async run(): Promise<void> {
+    if (this.request.filePath.includes('node_modules')) return;
```
devongovett: Should we just not queue files in node_modules at all? Would avoid crossing the IPC boundary just to do nothing...

packages/core/types/index.js
```diff
@@ -328,6 +331,19 @@ export interface TransformerResult {
...
 type ResolveFn = (from: FilePath, to: string) => Promise<FilePath>;
+
+type ResolveConfigFn = (
+  configNames: Array<FilePath>
+) => Promise<FilePath | null>;
+
+export type Validator = {|
+  validate({
+    asset: MutableAsset,
+    resolveConfig: ResolveConfigFn, // This is a temporary function and should be replaced with something cacheable
```
devongovett: sounds like a good followup task

packages/core/core/src/Validation.js
DeMoorJasper: Ow right, this was just copy-paste will change it to Asset

packages/core/core/src/Validation.js
DeMoorJasper: This was how it was originally but you commented on it to move it to the validatorrunner.

Will change it back. Well it wasn't a queue back then...

DeMoorJasper:
Everything should be updated now. Ready for another review. Will do config things in a follow-up PR

packages/core/core/src/Validation.js
devongovett: Oh.. haha sorry!

devongovett: feel free to update in a followup

PR #3234, opened 2019-07-14
Fix pipeline id identity in workers

devongovett:
This fixes a small bug in the transformation pipeline when run in workers where the pipeline ids (which are arrays) do not have equal object identity. This is due to the reverse handle being used to get the Parcel config, which means the config is copied across IPC each time and therefore has different object identity. This is fixed by joining the pipeline into a string and using that as an id instead.

A future optimization could be to not go across IPC every time we switch pipelines at all. We only need to re-resolve the parcel config if the directory changes, which will never happen for an asset. So we really only need to resolve the config once at the start of a transformation rather than for each individual pipeline.

Left a few other comments inline for some other things I noticed:

* Are we reading/writing from the cache every time we jump a pipeline? This seems probably unnecessary. I think we might only need to store the first and last pipelines.
* There might be a further bug related to this where the asset type changes but it still resolves to the same pipeline. It should finish that pipeline rather than stopping. That would require some further refactoring, and I wasn't sure how @padmaia wanted to do it.

### DIFF
--- a/packages/core/core/src/Transformation.js
+++ b/packages/core/core/src/Transformation.js
@@ -122,6 +122,7 @@ export default class Transformation {
     initialAsset: InternalAsset
   ): Promise<Array<InternalAsset>> {
     let initialType = initialAsset.type;
+    // TODO: is this reading/writing from the cache every time we jump a pipeline? Seems possibly unnecessary...
     let initialCacheEntry = await this.readFromCache(
       [initialAsset],
       pipeline.configs
@@ -250,7 +251,7 @@ export default class Transformation {
     }
 
     let pipeline = new Pipeline({
-      id: parcelConfig.getTransformerNames(filePath),
+      id: parcelConfig.getTransformerNames(filePath).join(':'),
       transformers: await parcelConfig.getTransformers(filePath),
       configs,
       options: this.options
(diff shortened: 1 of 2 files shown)


padmaia:
> Are we reading/writing from the cache every time we jump a pipeline? This seems probably unnecessary. I think we might only need to store the first and last pipelines.

One reason for this is more granular cache invalidation. If one of the plugins changed in the 2nd of 3 pipelines, we can still use the cache for the 1st pipeline. It also works well with a distributed cache. Let's say we have a transformation which will jump pipelines more than once. "I think we might only need to store the first and last pipelines." That means that the cache key for the first entry would have to be made up from meta about all of the pipelines that ran before the last one, but we only know which pipelines it will jump into after the transformation is run. If I've never run the transformation before, but someone else has, there's no way for me to take advantage of that because my cache key to see if I can skip transforming can only contain information for the first pipeline which would be a cache miss.

Is the concern that it would take up too much space in the cache?

> There might be a further bug related to this where the asset type changes but it still resolves to the same pipeline. It should finish that pipeline rather than stopping. That would require some further refactoring, and I wasn't sure how @padmaia wanted to do it.

Uh yeah, looks like you're right. I meant to account for it, but I'm checking the pipeline ids after it's already been kicked out of the pipeline. `Pipeline` has the parcel config, so it could probably just use it to check if the next pipeline id would be the same, and only bail out if it would be different.

packages/core/core/src/Transformation.js
```diff
@@ -250,7 +251,7 @@ export default class Transformation {
     }
 
     let pipeline = new Pipeline({
-      id: parcelConfig.getTransformerNames(filePath),
+      id: parcelConfig.getTransformerNames(filePath).join(':'),
```
padmaia: Woops, yeah I noticed this in my current branch as well.

devongovett:
Yeah I was just thinking about size and perf of writing/reading from the cache for every pipeline. But invalidating based on plugin versions makes sense. I had forgotten about that.

Issue #1251, opened 2018-04-26
Custom tsconfig.json name 🙋


laurence-myers:
This a 🙋 feature request.

### 🎛 Configuration (.babelrc, package.json, cli command)

File structure:

```
project
+- tsconfig.json
+- tsconfig_webview.json
+- src/
+-- main.ts
+- webview/
+-- webview.html
+-- webview.ts
```

### 🤔 Expected Behavior

Parcel allows specifying the tsconfig.json file to use.

### 😯 Current Behavior

Parcel will only look for a file called "tsconfig.json" in the directory where Parcel is executed.

### 💁 Possible Solution

I've found a workaround, which is to add a file "tsconfig.json" to the "webview" directory, and run parcel from that directory instead of the project root.

Ideally, there should be a command line option to specify the name and path of the tsconfig.json file.

### 🔦 Context

- Compiling and bundling WebView code for use in a React Native project, both using TypeScript, but requiring different compilation targets (ES6 for React Native, ES5 for WebView).

### 🌍 Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Parcel           | 1.7.0
| Node             | 8.9.4
| Yarn         | 1.3.2
| Operating System | Windows 10

DeMoorJasper:
This isn't standard in the tsconfig spec right?
Unless u are using extends?

fathyb:
I'm not sure to understand what's the problem with putting `tsconfig.json` in the `webview` directory? It's the standard way and will work natively with VSCode config resolution. However, you should not need to run `parcel` *in* the `webview` directory. Take this structure : 


```tree
project
+- tsconfig.json
+- src/
+-- main.ts
+- webview/
+-- tsconfig.json // you can use "extends" to inherit from ../tsconfig.json
+-- webview.html
+-- webview.ts
```

This should work: 
- `parcel webview/webview.html` <== build using `project/webview/tsconfig.json`
- `parcel src/main.ts` <== build using `project/tsconfig.json`

laurence-myers:
I'm closing this because:
1. There's a workaround.
1. More sophisticated support can probably be provided by a third-party loader.
1. I ended up switching back to webpack due to #1252, .ts file changes not getting picked up in "watch" mode, and other unrelated bundling requirements.

antongolub:
@fathyb faced with a similar issue, but in a slightly different context.

[...]

PR #3238, opened 2019-07-15
Add tsconfig path customization

antongolub:
# ↪️ Pull Request 
Adds custom `tsconfig` path through CLI. Refers to  #1251

## 💻 Examples

```
NODE_ENV=production parcel build --tsconfig tsconfig.rhino.json
```

## ✔️ PR Todo

### DIFF
--- a/packages/core/integration-tests/test/integration/typescript-custom-tsconfig/custom.tsconfig.json
+++ b/packages/core/integration-tests/test/integration/typescript-custom-tsconfig/custom.tsconfig.json
@@ -1,0 +1,5 @@
+{
+  "compilerOptions": {
+    "removeComments": true
+  }
+}
--- a/packages/core/integration-tests/test/integration/typescript-custom-tsconfig/index.ts
+++ b/packages/core/integration-tests/test/integration/typescript-custom-tsconfig/index.ts
@@ -1,0 +1,2 @@
+/* test comment */
+module.exports = 2;
--- a/packages/core/integration-tests/test/typescript.js
+++ b/packages/core/integration-tests/test/typescript.js
@@ -99,9 +99,24 @@ describe('typescript', function() {
     assert(!js.includes('local.a'));
   });
 
-  it('should support loading tsconfig.json', async function() {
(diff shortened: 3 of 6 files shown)


devongovett:
I don't think we should add typescript specific options to Parcel's CLI. What problem are you trying to solve? Normally tsconfig is resolved based on the location of your typescript files. Why doesn't this work for you?

antongolub:
In our case we use different build strategies for a single `src/main` to generate several runtime-specific bundles ("nodejs10", "rhino", "browser1", "browser2", ...). The current CLI API requires additional directories, but we'd like to keep [maven dir contract](https://maven.apache.org/guides/introduction/introduction-to-the-standard-directory-layout.html) as much as possible.

sushruth:
@devongovett I think this is more of a bundler problem than typescript problem. Just having a way to specify which `tsconfig.RANDOM_NAME.json` to use for which target could be a great feature and looking at  `packages/utils/ts-utils/src/loadTSConfig.js`, looks like an easy feature to provide without breaking many things. This is mostly to tell the bundler simply what file to use.

mmunte-impeo:
This is badly needed. Those who use typescript on large and complex projects will sure know why it is important to be able to have multiple tsconfig files. 

I really hope this PR is completed. I have evaluated parcel for a simple project and like it a lot. But my larger projects would require this PRs feature.

DeMoorJasper:
Parcel 1 no longer accepts new features, therefore I'll close this. Also not sure if this is the way to do this, maybe we can find a better way to configure this for Parcel 2?

elzii:
I'll chime in with a common use case:

Often in developing plugins for applications that provide a webview, I will want:

1. A `tsconfig.webview.json`config
2. A `tsconfig.plugin.json` config
3. A no-frills dev bundler for the webview, to not be beholden to constantly restarting the app (think Sketch for example), for which Parcel is a good choice.

Relying on it just to detect `tsconfig.json` leaves you at the mercy of the other bundlers...

Issue #2143, opened 2018-10-15
Parcel 2: Abstract filesystem to support more environments


devongovett:
Create a `@parcel/fs` package in `packages/core/fs` which contains promisified versions of all necessary fs functions. This is currently in the utils of Parcel 1. Update all references in Parcel 1 and Parcel 2 to use this package for all filesystem access. Plugins will also be able to use it if they need direct access to the filesystem.

Parcel 2 will support running in environments other than node, e.g. browsers. This could be accomplished by building Parcel using an alias for `@parcel/fs` to some other filesystem package.

humphd:
This is interesting.  We've got a drop-in replacement for `fs` for browsers in https://github.com/filerjs/filer.  It would be neat to be able to run parcel in the browser.

Doing a mix of browser based fs like this, and combining it with Service Workers to cache URLs, works really well.

cc @modeswitch

mischnic:
An in-memory fs would also be useful for the parcel test suite.

mischnic:
> Create a @parcel/fs package in packages/core/fs which contains promisified versions of all necessary fs functions.

Isn't this already resolved?

devongovett:
Yeah that part is done. I wanted to possibly experiment with passing an fs instance around rather than importing directly from `@parcel/fs` through. For the tests, ideally we'd want to read from the filesystem directly, but write the output into a virtual in-memory fs. For that, we'd need `inputFS` and `outputFS` options to parcel that would get passed around everywhere.

mischnic:
Maybe an alternative to passing fs objects around:

Configuring the `@parcel/fs` package once and then using prefixes to specify input/output
```js
import fs from "@parcel/fs";
fs.setInput(...);
fs.setOutput(...);

// ...

fs.writeFile("out://dist/index.html");
fs.readdir("in://.");

devongovett:
This is mostly done in #3244. Remaining work is to abstract the watcher so watch events work when using the MemoryFS.

Issue #1253, opened 2018-04-26
[RFC] Run Parcel in the browser 💥


devongovett:
I chatted with @CompuIves (author of Code Sandbox) at React Amsterdam about potentially running Parcel fully inside the browser. They do this in Code Sandbox using their own bundler currently, but it would be cool if we could enable them to use more pieces of Parcel. 

In addition, it would be super cool to have an online demo showing how Parcel works, similar to the one Rollup has. Maybe we could even just use Code Sandbox for that. 😜

There are a few things that we currently depend on Node for that won't work in the browser.

* Workers - We could use web workers in the browser instead of processes. With the new worker farm implementation that just landed this should be easier.
* Filesystem - We could abstract the filesystem so there could be a node implementation, and a virtual filesystem implementation for the browser.
* Server/HMR - can't run an HTTP server in the browser, but maybe we could abstract this too so you could plug in a different HMR implementation via events or something.
* Cache - use an in memory cache instead of a filesystem? Maybe something like indexeddb if we wanted to persist?
* Native modules - there are some like the SASS compiler that won't work, but could probably be compiled to webassembly.

These should all be solvable if we make Parcel sufficiently pluggable so that these pieces could be replaced with alternative implementations, even if they aren't all provided by Parcel itself. For example, there might be many implementations of a filesystem, not just a single browser one that we provide. We'd want to allow e.g. Code Sandbox to implement their own and plug it into Parcel easily.

This is kind of a meta issue to discuss the above items. We can break out into separate issues to discuss each one as we get closer to implementation. For now, let me know your thoughts!

CompuIves:
This sounds great, I'd love to help wherever I can!

I agree with the given points. The solution to some native modules would be to create browser specific transpilers (eg. Sass.js instead of node-sass), that wouldn't be too hard to implement I reckon.

I'm looking forward to more discussion on this, I'd really love to have Parcel working inside CodeSandbox 😄

fathyb:
Awesome idea!

- Filesystem : Ping @DeMoorJasper. We've discussed a bit about implementing a virtual fs in Parcel, perhaps this may help in this issue? IIRC CodeSandbox already has its vfs.

- Server/HMR - As we only serve static files we could totally polyfill this only using Object URLs (`URL.createObjectURL`) and use a internal messaging system instead of WebSockets.

A concrete `index.html` example :
```html
<!DOCTYPE html>
<html>
  <head>
    <title>Code Sandbox + Parcel</title>
[...]

PR #3244, opened 2019-07-17
Abstract filesystem and implement memory fs

devongovett:
Related: #1253 #2143

This PR abstracts the filesystem into an interface defined in `@parcel/fs`, and adds 2 implementations of that interface:

* `NodeFS` implements real filesystem access using the node `'fs'` module
* `MemoryFS` implements an in-memory filesystem using `SharedArrayBuffer` and IPC to share files across workers without copying (when using worker threads).

Both implementations are serializable so they can be sent across IPC and used by plugins.

Two options are added to Parcel to make use of these filesystem implementations: `inputFS` is used when reading from the user's project, and `outputFS` is used when writing to the dist and also to read/write from the cache. These options must both be instances of the `FileSystem` interface, but may be different filesystem objects. This allows flexibility to e.g. read from a real file system but write to an in memory implementation, or any other combination. This is in fact what we do in the tests, for performance and stability reasons.

All plugins should make use of these options instead of using the `fs` module. Also `process.cwd()` call should be replaced with `fs.cwd()` calls, since the cwd can be different depending on the filesystem being used and may not match the system cwd.

Some numbers, from the tests:

* tests on v2 branch: 1m 18s
* with worker threads: 58s
* running against memory fs: 20s

There are a few places where we still write to the real filesystem in the tests, but not many. Mostly they are due to actually testing for things we don't emulate in the memory fs, like the watcher.

This will open up the ability to run Parcel in the browser more easily. The `MemoryFS` should already work in the browser but I've not tested that. Once we get the worker farm working in web workers, we'll be pretty close I think.


devongovett:
hmm - doesn't look like the tests actually ran. lerna just stopped after the first package... 🤔

packages/core/fs/src/MemoryFS.js
```diff
@@ -0,0 +1,378 @@
...
+      }
+    }
+
+    return new ReadStream();
+  }
+
+  createWriteStream(filePath: FilePath) {
+    let fs = this;
+    let buffer = Buffer.alloc(0);
+    class WriteStream extends Writable {
```
mischnic: - The `bytesWritten` property is missing:
https://github.com/parcel-bundler/parcel/blob/5a9b703a3f42c8fa3ea2df78100a85d5b6375237/packages/core/core/src/PackagerRunner.js#L181-L188
- It might be faster to not create a new class on every `createWriteStream` call?

packages/core/fs/src/MemoryFS.js
```diff
@@ -0,0 +1,378 @@
...
+    size,
+    blksize: 0,
+    blocks: 0,
+    atimeMs: 0,
+    mtimeMs: 0,
+    ctimeMs: 0,
+    birthtimeMs: 0,
+    atime: new Date(0),
+    mtime: new Date(0),
+    ctime: new Date(0),
```
mischnic: How does the cache behave in this case (with MemoryFS as input fs)? The input files wouldn't have any change time info.

devongovett: Yeah none of that info was needed by Parcel itself, so I didn't implement it for now. Could be added later though.

devongovett:
@mischnic addressed your comments. Implemented full (or as much as possible) stat information, and separated the stream classes out.

packages/core/core/src/Parcel.js
```diff
@@ -71,10 +70,10 @@ export default class Parcel {
       this.#initialOptions
     );
     this.#resolvedOptions = resolvedOptions;
-    await createCacheDir(resolvedOptions.cacheDir);
+    await createCacheDir(resolvedOptions.outputFS, resolvedOptions.cacheDir);
 
     let {config} = await loadParcelConfig(
-      path.join(process.cwd(), 'index'),
+      path.join(resolvedOptions.inputFS.cwd(), 'index'),
```
wbinnssmith: what's the rationale for moving `cwd` to the filesystem interface? Any time we use a different fs backend we'll also want to implement `cwd` differently?

devongovett: `process.cwd()` is implicitly referencing the native OS filesystem. There is no guarantee that this directory will exist in other filesystems.

devongovett:
> We should definitely follow this with a lint rule that errors on using Node's native fs in most packages 

great idea!

packages/core/fs/src/MemoryFS.js
```diff
@@ -0,0 +1,493 @@
...
+    options: ?FileOptions
+  ) {
+    filePath = this._normalizePath(filePath);
+    if (this.dirs.has(filePath)) {
+      throw new Error(`EISDIR: ${filePath} is a directory`);
+    }
+
+    let dir = path.dirname(filePath);
+    if (!this.dirs.has(dir)) {
+      throw new Error(`ENOENT: ${dir} does not exist`);
```
mischnic: Node errors have a `code` prop (`ENOENT` in this case). From the `fs.stat` docs: "In case of an error, the `err.code` will be one of Common System Errors." Do we want to stay compatible with NodeFS in that regard?

packages/core/fs/src/MemoryFS.js
```diff
@@ -0,0 +1,493 @@
...
+    encoding: any,
+    callback: (error?: Error) => void
+  ) {
+    let c = Buffer.isBuffer(chunk) ? chunk : Buffer.from(chunk, encoding);
+    this.buffer = Buffer.concat([this.buffer, c]);
+    callback();
+  }
+
+  _final(callback: (error?: Error) => void) {
+    this.fs.writeFile(this.filePath, this.buffer, this.options).then(callback);
```
mischnic: Should rather be something like:
```js
.then(() => callback(), (err) => callback(err))
```
otherwise errors from `writeFile` would be silently swallowed

devongovett: Good catch

packages/core/fs/src/MemoryFS.js
devongovett: Updated

devongovett:
Added the --exit flag to mocha to work around a segfault due to our use of `unref()` on the worker and `SharedArrayBuffer`. Filed a bug with node: https://github.com/nodejs/node/issues/28777

PR #3254, opened 2019-07-21
Cache asset graph, bundle graph, and packager/optimizer results

devongovett:
Currently, the Parcel cache only stores individual assets, but there are now a few other things that can be cached. This PR adds caching for the asset graph as a whole, the bundle graph, and the results of packaging and optimizing individual bundles.

## Asset Graph

When Parcel exits, it now writes the full asset graph and request graph into the cache. When it starts up again, it can simply load them from disk rather than rebuild them from scratch. Even with individual asset transformations cached, it was still slow to build the entire graph for large applications due to resolving and checking the cache individually for each asset.

To make this work, we also use the new watcher to store a snapshot of the current filesystem into the cache along side the graphs. When Parcel starts up, it loads the graphs, and checks with the watcher to see what files have changed since the snapshot. Then, it applies those changes to the graph just like it does in watch mode. This means that if nothing changes between parcel runs, no work will occur. If only one file changes, then only that file will be re-transformed.

Since we store the entire graph in the cache now, and that includes the individual asset nodes (though not their contents), it raises the question of whether we need to store the individual assets at all. I guess the intermediary assets between pipelines might make that useful still, but there is some duplication now.

## Bundle Graph

We also now cache the bundle graph. If the asset graph does not change, then the bundle graph will also not change, so we can just reuse the same bundle graph from the cache. This is currently really only a performance gain if you restart parcel without changing anything. Perhaps it won't be necessary if the new bundler is sufficiently fast as well. In the future, I could see us caching the bundle graph and somehow only rebuilding bundles that changed, but that requires further thought.

## Packagers and Optimizers

We now cache the result of packaging and optimizing individual bundles. This is based on a hash of the bundle's asset graph, along with the parcel plugins + versions that were used to create it. Eventually, this should also include any third party configs used by those plugins, similar to what we're doing for transformers.

## Performance

I tested this on a relatively small app that just imported something from material-ui. Output size after minifying was ~86kb.

* Uncached (v2): 5.96s
* Cached (v2): 3.78s
* Cached (cache-graph): 402ms

About 200ms of the remaining time is actually just requiring plugins (specifically the reporter). Perhaps that will be reduced once we start bundling plugins for publishing.

## Follow up

* **Garbage collection**: this is needed in general, but this PR adds a bunch more things to the cache which will need to be garbage collected. Would be good to discuss this. One way to avoid it would be to not make the hash part of the cache key, and instead move it inside the cache entry and check it after retrieving. This would mean the cache key would not change every time. But perhaps there is a better way.
* **Third party config**: we need third party config support like we have to transformers in all of our other plugins as well, and we need to track when that changes and invalidate bundling/packaging/optimizing.
* **Abstract watcher**: we should probably abstract the watcher as part of the filesystem implementations so it works e.g. in the browser.

### DIFF
--- a/packages/core/core/src/AssetGraphBuilder.js
+++ b/packages/core/core/src/AssetGraphBuilder.js
@@ -2,8 +2,8 @@
 import EventEmitter from 'events';
 
 import type {AssetRequest, ParcelOptions, Target} from '@parcel/types';
-import {PromiseQueue} from '@parcel/utils';
-import type {Event} from '@parcel/watcher';
+import {PromiseQueue, md5FromObject, md5FromString} from '@parcel/utils';
+import watcher, {type Event} from '@parcel/watcher';
 
 import type Asset from './Asset';
 import AssetGraph from './AssetGraph';
@@ -16,6 +16,8 @@ import type {
 } from './types';
 
 import dumpToGraphViz from './dumpGraphToGraphViz';
+import Cache from '@parcel/cache';
+import path from 'path';
 
 type Opts = {|
   options: ParcelOptions,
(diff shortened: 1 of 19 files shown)


packages/core/core/src/AssetGraphBuilder.js
```diff
@@ -30,43 +32,64 @@ export default class AssetGraphBuilder extends EventEmitter {
...
+  cacheKey: string;
+  cache: Cache;
+
+  async init({config, options, entries, targets, assetRequest}: Opts) {
+    this.options = options;
+    let {minify, hot, scopeHoist} = options;
+    this.cacheKey = md5FromObject({
+      options: {minify, hot, scopeHoist},
+      entries,
+      targets
```
padmaia: Would it make sense to eventually have different graphs per target? That way we wouldn't lose the entire graph when we add or remove a target? Seems like we still have to keep entries together in one graph since we might end up with shared bundles between entries, but targets seem like they could maybe be separated cleanly?

Also, I've been thinking that options could maybe be treated similarly to files. Requests could specify which options they care about like they specify which files they care about right now, and when they change we can invalidate only the nodes that are connected to those options. That way we don't have to resolve everything all over when the minify option is changed.

devongovett: Yeah that's an interesting idea. Currently, we would still get some caching benefit since the individual assets are still cached in addition to the graph, so only the new target would actually be compiled. Not sure how often targets will be added/removed, but it's something to consider.

PR #3258, opened 2019-07-23
Process Pug output with PostHTML

### LINKED ISSUE #3252: Process Pug output with PostHTML?
# ❔ Question

Using Parcel, is it possible to process the HTML output of Pug assets with PostHTML?

## 🔦 Context

I have a [static site](https://github.com/cmessinides/cmessinides.com) that I'm building with Parcel. I'm writing the markup in Pug files and Parcel is emitting the HTML output, but I want to use some PostHTML plugins to further transform the resulting HTML. However, after adding a `posthtml.config.js` file to the root directory and installing and configuring a plugin (`posthtml-img-autosize`), the HTML output did not appear to have been transformed by PostHTML.

## 💻 Code Sample

Here's a minimal example of what I'm trying to accomplish:

- [Repo](https://github.com/cmessinides/parcel-pug-posthtml-example)
- [Live demo](https://parcel-pug-posthtml-example.netlify.com/)

## 🌍 Your Environment

| Parcel           | 1.12.3
| Node             | 11.2.0
| npm/Yarn         |  Yarn 1.16.0
| Operating System | WSL Ubuntu 18.04 on Windows 10 Build 17134.885

CMessinides:
# ↪️ Pull request

This processes the output of Pug files with PostHTML. It does this by using `HTMLAsset` as the base class of `PugAsset`, instead of `Asset`. The `parse()` method on the refactored `PugAsset` class simply compiles the Pug files into HTML before passing them off to `HTMLAsset.parse()` to convert the result into an AST. The existing functionality of `HTMLAsset` takes care of the rest, as if the Pug files had been HTML files all along.

This fixes #3252.

## 💻 Examples

Consider the following project directory:

```
📂 example
├ 📄 .posthtmlrc.js
├ 📄 included.html
├ 📄 index.pug
```
```js
// .posthtmlrc.js
module.exports = {
  plugins: [
    require('posthtml-include')({
      root: __dirname
    })
  ]
}
```

```html
<!-- included.html -->
<div>I'm included!</div>
```

```pug
//- index.pug
doctype html
html
  head
    title Pug with PostHTML
  body
    | <include src="included.html"></include>
```

With this change, Parcel will now output an `index.html` file with the contents of `included.html`, instead of leaving `<include src="included.html"></include>` in place. (This is a trivial example, considering Pug has a built-in includes system, but this will allow all PostHTML plugins to process Pug output.)

## 🚨 Test instructions

I have added an integration test for this change, and all other integration tests for `PugAsset` still pass as well. As far as I know, there are no other tests affected by this change.

## ✔️ PR Todo

### DIFF
--- a/packages/core/integration-tests/test/integration/pug-posthtml/.posthtmlrc.js
+++ b/packages/core/integration-tests/test/integration/pug-posthtml/.posthtmlrc.js
@@ -1,0 +1,10 @@
+module.exports = {
+	plugins: [
+		// Admittedly, using PostHTML for includes in Pug is unnecessary,
+		// but posthtml-include is already a development dependency of Parcel,
+		// so it's an easy pick for this example.
+    require('posthtml-include')({
+      root: __dirname
+    })
+  ]
+}
--- a/packages/core/integration-tests/test/integration/pug-posthtml/included.html
+++ b/packages/core/integration-tests/test/integration/pug-posthtml/included.html
@@ -1,0 +1,1 @@
+<div>I'm included!</div>
--- a/packages/core/integration-tests/test/integration/pug-posthtml/index.pug
+++ b/packages/core/integration-tests/test/integration/pug-posthtml/index.pug
@@ -1,0 +1,6 @@
+doctype html
+html
+  head
+    title Pug with PostHTML
+  body
+    | <include src="included.html"></include>
--- a/packages/core/integration-tests/test/pug.js
+++ b/packages/core/integration-tests/test/pug.js
@@ -123,6 +123,24 @@ describe('pug', function() {
     assert(html.includes('FILTERED: Hello!'));
   });
 
+  it('should support post-processing with PostHTML', async function() {
+    const b = await bundle(
+      path.join(__dirname, '/integration/pug-posthtml/index.pug')
+    );
+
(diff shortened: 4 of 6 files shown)


devongovett:
I don't think this should be necessary. Parcel should already be processing the output of pug (or anything that generates HTML) with the HTML asset. It works in a pipeline. If the type of the asset changes, it will be processed by the asset of that new type as well. Perhaps you could add some logs into HTMLAsset and see where it gets?

CMessinides:
That's interesting. That's how I assumed it would work, but in the issue I linked above (#3252), I have a minimal example of a Pug and PostHTML project where the Pug output does not get processed by PostHTML.

For what it's worth, I wrote the integration test first for this PR, and it failed before I made this change. So in both cases it does not appear to be working as intended.

I'm away from my computer atm, so I can't test HTMLAsset with logs right now, but I'll give it a shot and let you know what I find.

CMessinides:
So I've done some logging in HTMLAsset and rerun the integration test without the change, and you're correct @devongovett: The change isn't necessary and Parcel works as is. There must be some config error or error in a PostHTML plugin on my end.

I'll close this PR, but if you think it would be worthwhile to keep the integration test I wrote, I'm happy to open another PR just for that.

PR #3261, opened 2019-07-24
Babel config invalidation

padmaia:
# ↪️ Pull Request

This PR updates `@parcel/transformer-babel` to use the new config invalidation pattern. It also uses `@babel/core` to resolve babel configs, so `babel.config.js` is now supported.  There are certain newly supported scenarios that are non optimal (such as having a `babel.config.js` file in the first place) in which case a warning will be administered. One more notable change is that it is now using `core-js` over the deprecated `@babel/polyfill`.

## ✔️ PR Todo

- [x] Implement `config.shouldInvalidateOnStartup()` now that graphs are being cached
- [x] Figure out how to surface performance warnings (can we deduplicate?)
- [x] Integration tests
- [x] Add a README that describes how the plugin works and steps to take to optimize for caching
- [x] Separate Config into public and internal classes


packages/transformers/babel/src/config.js
```diff
@@ -1,141 +1,191 @@
...
+async function definePluginDependencies(config) {
+  let babelConfig = config.result.config;
+  if (babelConfig == null) {
+    return;
+  }
+
+  let configItems = [...babelConfig.presets, ...babelConfig.plugins];
+  await Promise.all(
+    configItems.map(async configItem => {
+      let pkg = nullthrows(
```
devongovett: Does the request graph already handle resolving these package versions?

packages/transformers/babel/src/config.js
```diff
@@ -1,141 +1,191 @@
...
+export async function load(config: IConfig) {
+  let partialConfig = loadPartialConfig({filename: config.searchPath});
+  if (partialConfig && partialConfig.hasFilesystemConfig()) {
+    let {babelrc, config: configjs} = partialConfig;
+    let {canBeRehydrated, dependsOnRelative, dependsOnLocal} = getStats(
+      partialConfig.options
+    );
+
+    // babel.config.js files get required by @babel/core so there's no use in including it for watch mode invalidation
+    if (babelrc != null && configjs == null) {
```
devongovett: what about `.babelrc.js`? Does that apply to the `babelrc` or `config` field?

packages/transformers/babel/src/config.js
```diff
@@ -1,141 +1,191 @@
...
+      config.setResolvedPath(babelrc);
+    } else if (configjs) {
+      // TODO: warn and invalidate on startup
+    }
+
+    if (babelrc && (await isExtended(/*babelrc*/))) {
+      // TODO: warn and invalidate on startup
```
devongovett: why do we need to invalidate if you used `extends`? Would we not just add all the extended files as included files?

packages/transformers/babel/src/config.js
```diff
@@ -1,141 +1,191 @@
...
+      // TODO: maybe add feature to babel that gives details on extended files?
+    }
+
+    if (dependsOnRelative || dependsOnLocal) {
+      // TODO: warn that you may not see changes to relative or local plugins/presets until their
+      // containing package changes versions. You can run with --no-cache to see changes until ready
+      // to bump version.
```
devongovett: what if we added them as included files that invalidated the config when they changed? I guess the problem is node's `require.cache`?

packages/transformers/babel/src/config.js
```diff
@@ -1,141 +1,191 @@
...
+      config.setResult({
+        internal: false,
+        config: partialConfig.options
       });
+
+      await definePluginDependencies(config);
+      config.setResultHash(md5FromObject(partialConfig.options));
+    } else {
+      config.shouldReload();
+      config.setResultHash(JSON.stringify(Date.now()));
```
devongovett: is a hash necessary in this case?

packages/core/types/index.js
```diff
@@ -334,29 +373,29 @@ export type Transformer = {
     asset: MutableAsset,
     resolve: ResolveFn,
     options: ParcelOptions
-  }) => Async<Config | void>,
+  }) => Async<ThirdPartyConfig | void>,
```
devongovett: Can we add `loadConfig` and `rehydrateConfig` to the `Transformer` type here?

packages/transformers/babel/src/BabelTransformer.js
```diff
@@ -3,29 +3,27 @@ import {Transformer} from '@parcel/plugin';
...
-import babel6 from './babel6';
 import babel7 from './babel7';
-import getBabelConfig from './config';
 import {relativeUrl} from '@parcel/utils';
+import {load, rehydrate} from './config';
 
 export default new Transformer({
-  async getConfig({asset}) {
-    return getBabelConfig(asset);
+  async loadConfig(config) {
```
devongovett: Can we make these method signatures match the others with an options object? That way we can add other params in the future more easily.

Also, do you think these methods should have access to the Asset being transformed, and the Parcel options object?

packages/core/types/index.js
```diff
@@ -299,6 +302,42 @@ export interface Asset extends BaseAsset {
...
+  watchGlob: ?Glob;
+  devDeps: Map<PackageName, ?string>;
+  pkg: ?PackageJSON;
+
+  setResolvedPath(filePath: FilePath): void;
+  // $FlowFixMe could be anything
+  setResult(result: any): void; // TODO: fix
+  setResultHash(resultHash: string): void;
+  addIncludedFile(filePath: FilePath): void;
+  setDevDep(name: PackageName, version?: Semver): void;
```
devongovett: Can we match these with the methods we have on Asset? `addConnectedFile` and `addDependency`?

packages/core/types/index.js
```diff
@@ -299,6 +302,42 @@ export interface Asset extends BaseAsset {
...
+  resultHash: ?string;
+  includedFiles: Set<FilePath>;
+  watchGlob: ?Glob;
+  devDeps: Map<PackageName, ?string>;
+  pkg: ?PackageJSON;
+
+  setResolvedPath(filePath: FilePath): void;
+  // $FlowFixMe could be anything
+  setResult(result: any): void; // TODO: fix
+  setResultHash(resultHash: string): void;
```
devongovett: Did you consider an API where you the result directly from the plugin, instead of mutating the passed object? We could also automatically compute the hash from the result in core, unless there's a case where leaving that to the plugin is required.

```javascript
loadConfig({config}) {
  return {
    filePath: '...', // instead of setResolvedPath
    config: {...}    // instead of setResult
  };
}
```

I'm ok with either, just wanted to ensure we considered it.

packages/core/core/src/Transformation.js
```diff
@@ -250,13 +251,19 @@ export default class Transformation {
           moduleName,
           parcelConfig.resolvedPath
         );
+        if (thirdPartyConfig.rehydrate) {
+          await plugin.rehydrateConfig(thirdPartyConfig);
+        } else if (thirdPartyConfig.reload) {
+          await plugin.load(thirdPartyConfig);
```
devongovett: should this be `loadConfig`?

packages/core/core/src/Transformation.js
```diff
@@ -250,13 +251,19 @@ export default class Transformation {
           moduleName,
           parcelConfig.resolvedPath
         );
```
devongovett: So, if I understand correctly, all config for all plugins is loaded in the main thread. Then it's sent back to the transformer thread, and rehydrated. But, if it needs to be reloaded, it is immediately reloaded again on the transformer thread. So we end up needing to load it twice in that case?

packages/core/types/index.js
```diff
@@ -299,6 +302,42 @@ export interface Asset extends BaseAsset {
   +stats: Stats;
 }
 
+// TODO: figure out naming
+export interface IConfig {
```
devongovett: I'm not sure all of this should be a public type. With our other types, we have an internal version with fields like the below, and a public one with the public interface. It think we should aim to match for config as well.

As for the interface name, we've generally named it without the prefix, so just `Config` instead of `IConfig`.

packages/core/types/index.js
```diff
@@ -299,6 +302,42 @@ export interface Asset extends BaseAsset {
...
+    options: ?{packageKey?: string, parse?: boolean}
+  ): Promise<ThirdPartyConfig | null>;
+  getConfig(
+    filePaths: Array<FilePath>,
+    options: ?{packageKey?: string, parse?: boolean}
+  ): Promise<ThirdPartyConfig | null>;
+  getPackage(): Promise<PackageJSON | null>;
+  isSource(): Promise<boolean>;
+  shouldRehydrate(): void;
+  shouldReload(): void;
```
devongovett: maybe a getter/setter for these instead of methods?

packages/core/core/src/RequestGraph.js
```diff
@@ -307,22 +306,25 @@ export default class RequestGraph extends Graph<RequestGraphNode> {
...
         this.addNode(depVersionRequestNode);
       }
       this.addEdge(configRequestNode.id, depVersionRequestNode.id);
       depVersionRequestNodes.push(
         nullthrows(this.getNode(depVersionRequestNode.id))
       );
 
-      let version = await this.getSubTaskResult(depVersionRequestNode);
-      config.setDevDep(depVersionRequest.moduleSpecifier, version);
+      if (version != null) {
```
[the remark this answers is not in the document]
padmaia: good catch

packages/core/core/src/public/Config.js
```diff
@@ -1,44 +1,78 @@
...
   resolvedPath: ?FilePath;
+  // $FlowFixMe
   result: ?any;
   resultHash: ?string;
   includedFiles: Set<FilePath>;
   watchGlob: ?Glob;
   devDeps: Map<PackageName, ?string>;
+  pkg: ?PackageJSON;
+  rehydrate: ?boolean;
+  reload: ?boolean;
```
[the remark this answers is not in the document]
padmaia: Yeah, I honestly didn't spend too much time thinking about what should be public/private, just wanted to get everything working first. Agreed though. I'll do another run through, there's probably things from previous PRs that should also be more thoughtful about private/public API.

packages/core/core/src/Transformation.js
padmaia: Yup 😬

packages/core/types/index.js
padmaia: Yeah, that's something I've discussed with @wbinnssmith a bit. One reason I was leaning this way is that it is more consistent with how we pass `Asset` to the other methods. It also alleviates some work with methods like `getConfig` that both help with resolving/parsing config and also adding whichever file was resolve to included files.

> We could also automatically compute the hash from the result in core, unless there's a case where leaving that to the plugin is required.

Yeah, my thinking was it would be nice to be flexible because it's possible a resolved config might contain more information than is relevant for invalidation. One example where we're actually using it right now is the parcel config. For invalidation of a pipeline transformation, we only care about one pipeline, not the entire config. We could accomplish the same thing in other ways for Parcel, but it seemed like it might be useful for plugins. For instance, plugins might be using information from package.json, but they only care a single field. (That reminds me the babel transformer should probably invalidate if the `source` field changes)

packages/core/core/src/Transformation.js
padmaia: Yeah, for babel that will happen if you are using a `babel.config.js` file and you have something like:
```
module.exports = {plugins: [require('babel-plugin-whatever')]}
```
because at that point all we have is a function which can not be rehydrated. And unfortunately, we don't know if the user is using this type of configuration until we try to load it. Not ideal obviously, but I thought it would be better to just support it and surface a warning that this type of configuration has performance impacts. I was thinking it would also be a good idea to document how to optimize for caching in a README for this package. I'll add that as a TODO.

packages/core/types/index.js
padmaia: Yeah, again was kind of neglecting internal/public differences.

> As for the interface name, we've generally named it without the prefix, so just Config instead of IConfig.

Ah yeah, thought we had something like IDependency in the types package before, but it looks like I either have a bad memory or we changed it. I'll update the name :)

packages/core/types/index.js
padmaia: Those are the functions that the plugin uses to tell core that the config should rehydrated or reloaded. The functions just set a variable on the class that core then reads later. That's why the variables weren't named in a way that makes it clear that they are booleans. I wasn't sure what else to name them that wouldn't conflict with the method names.

packages/transformers/babel/src/config.js
```diff
@@ -1,141 +1,191 @@
...
+  return false;
+}
+
+async function isLocal(/* configItemPath */) {
+  // TODO: check if realpath is different, need access to fs in case of memory filesystem
+  return false;
+}
+
+function prepForReyhdration(options) {
+  options.presets = (options.presets || []).map(configItem => ({
```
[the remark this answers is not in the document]
padmaia: Omitting something. With the new serializer it was choking on functions so I had to intentionally exclude them here. I can add a comment.

packages/transformers/babel/src/flow.js
```diff
@@ -1,19 +1,19 @@
...
-import type {MutableAsset} from '@parcel/types';
+
+import {IConfig} from '@parcel/types';
+import type {BabelConfig} from './types';
 
 /**
  * Generates a babel config for stripping away Flow types.
  */
-export default async function getFlowConfig(asset: MutableAsset) {
-  if (/^(\/{2}|\/\*+) *@flow/.test((await asset.getCode()).substring(0, 20))) {
```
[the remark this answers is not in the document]
padmaia: Hmm okay, that's something I can look into. What plugin do you mean by the "it still has to add the parser plugin"?

packages/transformers/babel/src/config.js
padmaia: The request graph now skips resolving dev dep versions if it sees that they were already resolved by the plugin.

packages/transformers/babel/src/env.js
```diff
@@ -1,52 +1,44 @@
...
   // If this is the app module, the source and target will be the same, so just compile everything.
   // Otherwise, load the source engines and generate a babel-present-env config.
-  if (!isSourceModule) {
-    let sourceEngines = await getTargetEngines(asset);
-    let sourceEnv = (await getEnvPlugins(sourceEngines, false)) || targetEnv;
+  if (!(await config.isSource())) {
+    let sourceEngines = await getTargetEngines(config);
+    if (!sourceEngines) return null;
 
-    // Do a diff of the returned plugins. We only need to process the remaining plugins to get to the app target.
```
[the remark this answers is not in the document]
padmaia: I'm not sure how relevant this comment is anymore. I still diff the plugins to see if we need to transform with the env preset, but the diffed plugins are all just functions right? I have no way of knowing *which* plugins need to run in a cacheable and rehydratable way. Unfortunately this means that more plugins than necessary will run over these linked packages, but I feel like it's a problem that could possibly be solved in the preset itself. If the preset accepted another option like `startingTarget`, that would be perfect. I thought I added a comment suggesting this, but maybe not because I can't find it now. I suppose it would still be helpful to leave a comment describing why we diffing the plugins though.

packages/transformers/babel/src/config.js
padmaia: Mmm, that's a good question. I was thinking that babelrc was specifically the old json style and config was for the new js style. I'll double check though.

packages/transformers/babel/src/config.js
padmaia: I guess I was thinking that there could be many chained extensions and I didn't want to keep following through to all of them when `@babel/core` already did that. I'd rather `@babel/core` just tell us. But yeah, might be something worth doing.

packages/transformers/babel/src/config.js
padmaia: We'd have to resolve all of the files that the resolved file depends on in case those changed to, which seemed like it would add a lot of complexity. Do you think that would be worth it?

packages/transformers/babel/src/config.js
padmaia: Well the hash needs to be different each time because we have no way of knowing if the config is the same as last time or not. Are you suggesting something like core doing this automatically is no hash is set?

packages/core/types/index.js
padmaia: Oh yeah, meant to do that.

packages/core/types/index.js
padmaia: Yeah `addConnectedFile` and `addIncludedFile` should definitely be consistent. I think I was maybe the one who suggested `connected` vs `included` in the first place. I don't remember perfectly, but I think that was when we were still thinking things like configs would be connected to the asset, where "connected" makes more sense because they are not actually included in the resulting asset. Now that config is connected to config request nodes, "included" seemed like better naming in both cases. I don't think it matters too much, I might be overthinking it. Thoughts?

As for `addDevDep` and `addDependency` I was thinking they are slightly different so it would make sense to have different names. With dependencies we end up resolving the path to that dependency whereas with devDeps we just resolve the version. I don't feel too strongly that they should definitely be named different things though. What do you think?

packages/transformers/babel/src/flow.js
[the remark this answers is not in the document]
devongovett: The flow plugin actually adds two plugins: the parser plugin to enable parsing flow syntax, and the babel plugin to strip the types off. See https://github.com/babel/babel/blob/master/packages/babel-plugin-transform-flow-strip-types/src/index.js#L14. Last time I tried, enabling the flow syntax parsing was quite a bit slower than not so keeping it off when not needed might be good for performance. Also, I believe the flow parser might introduce some ambiguities with how JS is parsed, so it would be more spec compliant to keep it off unless it is enabled with the comment.

packages/transformers/babel/src/config.js
devongovett: Sure, but if it's already being handled by the graph do we need to do it here?

packages/transformers/babel/src/config.js
devongovett: `.babelrc.js` is definitely still supported: https://babeljs.io/docs/en/configuration#babelrcjs

packages/transformers/babel/src/config.js
devongovett: Oh yeah for sure. We should make a PR to babel for that.

packages/transformers/babel/src/config.js
devongovett: Yeah I was kinda thinking of not having plugins set hashes at all, and we'd just do it in core...

packages/core/types/index.js
devongovett: I'm fine with `addConnectedFile` or `addIncludedFile` as long as it's consistent. For dependencies, I can see where the difference is - a dependency actually refers to a `Dependency` type. Maybe we could just spell it out fully: `addDevDependency` (like the package.json field).

packages/core/types/index.js
devongovett: With transformers we kinda actually support both: there is that `TransformerResult` type which has fields like `code` and `map` that you can return instead of mutating the asset. We found this to be challenging to deal with when eg adding dependencies, and passing assets into util functions, and mutation was easier there. The difference here is that Assets start out fully hydrated (they have input code etc.), whereas config starts out as an empty object. It just felt a little strange. I'm not totally against what's here, just curious what it would look like another way.

I think your usecase with hashes is a good one to allow control to plugins, but I think we should make it optional. i.e. if no hash is set by the plugin, we could set one automatically based on the result. Perhaps the `shouldReload` setting could also set the hash to something random?

packages/core/core/src/Transformation.js
devongovett: Do configs need to be loaded on the main thread at all? I guess to check the graph? What if we just didn't store anything in the graph at all for configs that are uncacheable?

packages/core/types/index.js
devongovett: Yeah I was thinking we could have both `get shouldReload()` and `set shouldReload()`. The internal variable would become private.

packages/transformers/babel/src/config.js
padmaia: Yeah, I was just thinking that the `babelrc` property returned from `loadPartialConfig` was for json and the property`config` was for js. Looks like it uses the `babelrc` field for `.babelrc.js` though 😕 Ugh, is the distinction just that `babel.config.js` behaves differently because it's expected to be at the root? Guess I'll have to figure out a different way to determine if a js file was used.

packages/core/core/src/public/Config.js
```diff
@@ -1,44 +1,78 @@
-// @flow
-import type {FilePath, PackageName, Glob} from '@parcel/types';
+// @flow strict-local
+import path from 'path';
+
+import type {
```
wbinnssmith: nit: type imports first

packages/transformers/babel/src/config.js
padmaia: Ah yeah. I was thinking that Babel already did the work of resolving the package (and then the entry file for that package) so it would be less work to resolve the `package.json` from within the package rather than resolving all over again.

packages/transformers/babel/src/BabelTransformer.js
padmaia: Options I'll definitely start passing now. The asset seems like it could be useful especially if the content of the asset might influence the config, so we don't have to read the file more than once. It's a little tricky right now because the loading of the config is taking place in the main process which doesn't have access to the asset and I'm not sure it would be a good idea to pass it across IPC. I've been thinking of ways that we could maybe get rid of the reverse handles, so if we figure that out that would enable us to provide the asset. One downside of not using reverse handles anymore would be when we do need to reload config, internal tool caches would be spread out across all of the processes. Probably still worth it, though.

For now plugins at least have access to the assets path. Currently it's named `searchPath`, but a better name for that may be `assetPath`.

packages/transformers/babel/src/config.js
padmaia: Oh yeah, also the `require.cache` problem. I'll think about this a little more, but I think the best thing to do for now is warn.

packages/transformers/babel/src/env.js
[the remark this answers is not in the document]
padmaia: Added a more accurate comment

packages/transformers/babel/src/flow.js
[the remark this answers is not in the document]
padmaia: Okay yeah, probably still worth it to do then. Problem right now is that we don't have access to asset in the main process right now. I mentioned in another [comment](https://github.com/parcel-bundler/parcel/pull/3261#discussion_r310237543) that I'm thinking of ways to solve for that. Is it cool if we leave this problem for a followup?

packages/core/types/index.js
padmaia: Updated `Asset` to use `addIncludedFile` instead of `addConnectedFile` and changed `setDevDep` to `addDevDependency`

padmaia:
@devongovett I've separated Config into a public and internal class. Going to resolve conversations related that so it's easier for me to tell what's still left to do, but let me know what you think.

packages/core/types/index.js
padmaia: I haven't had the time to explore this yet. Is this a blocker for the PR or could we follow up on it?

packages/core/core/src/Transformation.js
padmaia: We probably need some placeholder so we can at least invalidate on startup. Might not have to store an entire Config class though.

packages/transformers/babel/src/config.js
padmaia: I am now checking the ext of babelrc to see if it is a js file.

PR #3262, opened 2019-07-25
P2B-31 mdx transformer

qburst-aparna:
This is the initial commit for a simple mdx transformer implementation for brisk. `yarn demo` in the kitchen-sink example gives ` Unexpected token (9:9)` error from `/parcel/node_modules/@babel/parser/lib/index.js:10157:28)`. If we replace the render.js to app.js in the script in kitchen-sink it logs the MDX component.


packages/configs/default/index.json
```diff
@@ -1,6 +1,7 @@
 {
   "bundler": "@parcel/bundler-default",
   "transforms": {
+    "*.{md,mdx}": ["@parcel/transformer-mdx"],
```
DeMoorJasper: Not sure if this should also parse standard markdown by default as normal markdown could be expected to result in html instead of JS

DeMoorJasper:
Closing in favor of #3858

PR #3266, opened 2019-07-25
Unified test suite

gorakong:
# ↪️ Pull Request

Unified monorepo test suite

## 💻 Examples

Running 'yarn test' used to create one mocha instance per package. This change ensures that only one instance of mocha is run.

## 🚨 Test instructions

Old behavior: Running 'yarn test' produced test logs that were interleaved due to the presence of multiple running mocha instances.
New behavior: Running 'yarn test' outputs the results of all tests in all packages in one go.

## ✔️ PR Todo

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,8 +18,11 @@
     "lint": "eslint . && prettier \"./packages/*/*/{src,bin,test}/**/*.{js,json,md}\" --list-different",
     "precommit": "lint-staged",
     "prepublishOnly": "yarn build",
-    "test-ci": "lerna run test-ci --stream",
-    "test": "cross-env NODE_ENV=test mocha"
+    "test:unit": "cross-env NODE_ENV=test mocha --config unit.mocharc.json",
+    "test:integration": "(cd packages/core/integration-tests && yarn test) && mocha --config integration.mocharc.json",
+    "test:integration-ci": "(cd packages/core/integration-tests && yarn test-ci) && mocha --config integration.mocharc.json",
+    "test": "yarn test:unit && yarn test:integration",
+    "test-ci": "yarn test:unit && yarn test:integration-ci"
   },
   "devDependencies": {
     "eslint": "^5.16.0",
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
     "prepublishOnly": "yarn build",
     "test:unit": "cross-env NODE_ENV=test mocha --config unit.mocharc.json",
     "test:integration": "mocha --config integration.mocharc.json && (pushd packages/core/integration-tests && yarn test && popd)",
-    "test:integration-ci": "mocha --config integration.mocharc.json && (pushd packages/core/integration-tests && yarn test-ci && popd)",
+    "test:integration-ci": "mocha --config integration.mocharc.json && (cd packages/core/integration-tests && yarn test-ci)",
     "test": "yarn test:unit && yarn test:integration"
   },
   "devDependencies": {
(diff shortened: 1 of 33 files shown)


package.json
```diff
@@ -48,5 +48,9 @@
   "collective": {
     "type": "opencollective",
     "url": "https://opencollective.com/parcel"
+  },
+  "dependencies": {
+    "@babel/core": "^7.5.5",
```
wbinnssmith: I think if we use the `@parcel/babel-register` package we shouldn't need these here.

package.json
```diff
@@ -19,7 +19,7 @@
     "precommit": "lint-staged",
     "prepublishOnly": "yarn build",
     "test-ci": "lerna run test-ci --stream",
-    "test": "lerna run test --stream"
+    "test": "mocha packages/**/*.test.js --require @parcel/babel-register"
```
wbinnssmith: Can this be moved into a mocharc file? https://mochajs.org/#priorities

wbinnssmith:
I don't think most of the integration tests have a long enough timeout with mocha's default settings, so we'll either need to increase the timeout across all tests (in a mocharc or mocha opts file), or create multiple suites.

I think one suite that covers fast unit tests, and one that runs longer integration tests might be ideal. The latter could have a custom timeout. We could also run the two suites in parallel. What do you think @padmaia @devongovett?

After that we should be able to remove the `mocha.opts` files across the repo:

```
~/s/parcel $ rg --files . | grep mocha.opts
./packages/core/test-utils/mocha.opts
./packages/core/core/test/mocha.opts
./packages/core/logger/test/mocha.opts
./packages/core/source-map/test/mocha.opts
./packages/core/parcel-bundler/test/mocha.opts
./packages/core/workers/test/mocha.opts
./packages/utils/events/test/mocha.opts
./packages/core/utils/test/mocha.opts
./packages/configs/default/test/mocha.opts
./packages/core/integration-tests/test/mocha.opts
./packages/reporters/cli/test/mocha.opts
./packages/reporters/json/test/mocha.opts
```

and replace them with just the two.

yarn.lock
```diff
@@ -26,6 +26,13 @@
   dependencies:
     "@babel/highlight" "^7.0.0"
 
+"@babel/code-frame@^7.5.5":
```
padmaia: Why has the yarn.lock file changed if none of the dependencies have changed?

wbinnssmith: I think this is from the prior commit. Just needs to be updated with another run of yarn.

package.json
padmaia: We've started using the extension for `.test.js` for new unit tests, but the integration tests do not have that extension yet. One option would be to change the glob to just look for any files inside of the `test` folder at the root of all packages, but this would also be a good opportunity to consistently name our tests with the `.test.js` extension.

padmaia:
Could you also clean up the `test` scripts in the packages, now that they are no longer being used?

padmaia:
It looks like some of the tests are expecting `NODE_ENV` to be set to `test`. I'm not sure if this will cause any issues with other tests that are not necessarily expecting that, but go ahead and set it and see how it goes.

gorakong:
@padmaia @wbinnssmith thanks for the reviews! I've updated the Test Instructions of the PR description, got rid of the test scripts from individual packages, and set the `NODE_ENV` to `test`. Getting rid of the test scripts seems to have caused CI to fail though

package.json
```diff
@@ -19,7 +19,7 @@
     "precommit": "lint-staged",
     "prepublishOnly": "yarn build",
     "test-ci": "lerna run test-ci --stream",
```
wbinnssmith: looks like this is still using lerna. `test-ci` is what ci runs. We should also remove these entries from the other packages.

PR #3272, opened 2019-07-26
Fix #3156

### LINKED ISSUE #3156: PostHTML Plugins: ast.walk is not a function
# 🐛 bug report

I was using Parcel, PostHTML and several PostHTML plugins to create my documents. However the build fails on certain conditions.

Here is my `.posthtmlrc`:

```.posthtmlrc
{
  "plugins": {
    "posthtml-include-md": true,
  }
}
```

`test.html` (plugins above do not have any effect on this HTML since this doesn't include any markdown paragraphs):
```html
<!DOCTYPE html>
<html>
	<head>
		<meta charset="utf-8">
	</head>
	<body>
		<div class="baz">foobar</div>
	</body>
</html>
```

CLI Command: `$ parcel test.html`

## 🤔 Expected Behavior

```bash
$ parcel test.html
✨  Built in 6ms.
```

## 😯 Current Behavior

```bash
$ parcel test.html
🚨  /path/to/my/document/test.html: ast.walk is not a function
    at HTMLAsset.collectDependencies (/path/to/parcel/packages/core/parcel-bundler/src/assets/HTMLAsset.js:138:9)
    at HTMLAsset.getDependencies (/path/to/parcel/packages/core/parcel-bundler/src/Asset.js:81:18)
```

## 💁 Possible Solution

Following the stack trace and the source code, I figured out that the problem came from where Parcel loaded plugins. `ast` in the error log above was not an object that expresses AST but was a expoted function that returns an actual plugin function (that takes a tree as an argument). As PostHTML `process()` function didn't have type checks, it simply returned an actual plugin function which clearly didn't have a `walk` property.

I made a small modification to `/packages/core/parcel-bundler/src/utils/loadPlugins.js` then the build succeeded. I removed a length check for `options` so that `plugin` functions were always executed even with empty `options`. 

```js
async function loadPlugin(plugin, relative, options) {
 ...
   if (typeof options !== 'object') {
      options = {};
    }
   
    if (Object.keys(options).length > 0) { // removed this 
      plugin = plugin(options);
    }
...
```

Giving the plugin options (like `{"a":"b"}` instead of `true`) in `.posthtmlrc` also made the build successful.

In addition, [README.md of posthtml/posthtml](https://github.com/posthtml/posthtml) also shows that we have to call the exported functions (`require('...')()`) of the plugin before calling `process(ast, options)` since they are functions that return functions.

```js
posthtml(
  [
    require('posthtml-to-svg-tags')(),
    require('posthtml-extend-attrs')({
      attrsTree: {
        '.wow' : {
          id: 'wow_id',
          fill: '#4A83B4',
          'fill-rule': 'evenodd',
          'font-family': 'Verdana'
        }
      }
    })
  ])
  .process(html/*, options */)
  .then((result) =>  console.log(result.html))
```

As the changes I made to `loadPlugins.js` might cause problems of other transformers such as PostCSS, I tried to modifying `parcel-bundler/src/transforms/posthtml.js` instead but I wasn't able to figure out how to pass the plugins' options to them. For Parcel v2 this should be straightforward thanks to the non-shared `loadPlugins.js`.

## 🌍 Your Environment

| Parcel           | 1.12.3 |
| Node             | v11.13.0 |
| npm/Yarn         | 1.16.0 (yarn) |
| Operating System | Arch Linux (kernel: 5.0.5-arch1-1) |

samuelgozi:
`.posthtmlrc` plugins can now have `true` instead of an empty object when no configuration is needed.
This conforms to the documentation: https://parceljs.org/html.html#posthtml.

# ↪️ Pull Request

This resolves #3156 (That is already closed, but not fixed).

So when using posthtml plugins, sometimes no configuration is required, and according to [parcel's official documentation](https://parceljs.org/html.html#posthtml)
> If there are no options for a plugin, just set it to `true` instead.

The problem was that for some reason(I don't know why) the object for the settings of each plugin is modified by using `Object.assign(pluginsSettings, settingsToAdd)`.
However, when `Object.assign` is passed a boolean as the first argument it fails silently(technically the operation is performed, but its still a boolean, so there is no access to those props anymore).

You can see the code for the fix, but basically, I did it that way in order to avoid an `if` check.

## 🚨 Test instructions

To be honest im not familiar enough with the internals of Parcel to do a good test.

## ✔️ PR Todo

### DIFF
--- a/packages/core/parcel-bundler/src/transforms/posthtml.js
+++ b/packages/core/parcel-bundler/src/transforms/posthtml.js
@@ -45,7 +45,10 @@ async function getConfig(asset) {
           asset.addDependency(name, {includedInParent: true})
       }
     };
-    Object.keys(plugins).forEach(p => Object.assign(plugins[p], depConfig));
+    
+    Object.keys(plugins).forEach(p => {
+        plugins[p] = Object.assign({}, plugins[p], depConfig);
+    });
   }
   config.plugins = await loadPlugins(plugins, asset.name);
   config.skipParse = true;


samuelgozi:
For some reason, I'm just not able to run tests on my machine without it failing(on master), so if someone could give me a hand making the tests work on my machine I'll be grateful.

DeMoorJasper:
gonna lose this as parcel 1 is no longer receiving updates.

Feel free to check and implement this for v2

PR #3278, opened 2019-07-26
Support loading runtimes from multiple inbound dependencies

wbinnssmith:
This corrects a bug when multiple bundles attempted to async import a module, only one would receive the necessary runtime code to load it.

As a concrete example:

`index.js`:
```js
import('./a');
import('./b');
```

`a.js`:
```js
import('./c');
```

`b.js`:
```js
import('./c');
```

(And some `c.js`, but contents don't matter)

Fails to run, as one of `a` or `b` does not get the necessary runtime code.

Both bundles would have dependencies pointing to the bundle group containing the bundled async module, but because bundle groups had only one reference to an inbound dependency, only the dependency referenced there would have runtime code inserted.

This is fixed by using the relationship between dependencies and bundle groups encoded in the bundlegraph, rather than using a property on bundle groups. The `dependency` property on bundle groups has been removed.

Test Plan: Run the added unit test without this patch, and verify it fails. Add this patch and verify it passes.

### DIFF
--- a/packages/core/core/src/BundleGraph.js
+++ b/packages/core/core/src/BundleGraph.js
@@ -9,7 +9,13 @@ import type {
   TraversalActions
 } from '@parcel/types';
 
-import type {AssetNode, Bundle, BundleGraphNode, DependencyNode} from './types';
+import type {
+  AssetNode,
+  Bundle,
+  BundleGraphNode,
+  BundleGroupNode,
+  DependencyNode
+} from './types';
 import type Asset from './Asset';
 import type Graph from './Graph';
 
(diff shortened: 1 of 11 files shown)


packages/core/core/src/BundleGraph.js
```diff
@@ -294,6 +300,58 @@ export default class BundleGraph {
...
+              .find(
+                node =>
+                  node.type === 'dependency' &&
+                  this._graph.hasEdge(bundle.id, node.id, 'contains')
+              )
+          };
+        })
+        // TODO: Remove this filter when bundle groups have the correct bundles
+        // pointing to them
+        .filter(
```
wbinnssmith: This filter works around a bug where bundles up the tree don't get prior connections removed during the bundling process. I think this was pointed out here: https://github.com/parcel-bundler/parcel/pull/3259/files#r306637515

packages/core/core/src/BundleGraph.js
```diff
@@ -294,6 +300,61 @@ export default class BundleGraph {
       });
   }
 
+  getBundleGroupsReferencedByBundle(
```
devongovett: Kinda weird name since it actually returns both bundle groups and dependencies... but idk what else it could be

PR #3281, opened 2019-07-29
v2: Proxy table support

Kogia-sima:
# ↪️ Pull Request

This PR implements the proxy table handler for v2. For the similar feature support in v1, see #2477.

Many people use Parcel dev-server as main development server, but some users want to redirect the requests to the specified URL (e.g. '/api/*') to another host. In order to achieve this, users have to implement a new reporter and replace the default dev-server with it.

With this feature, you can define the proxy table in the `package.json` at or above the project root. Also you can pass any option allowed in [http-proxy-middleware](https://github.com/chimurai/http-proxy-middleware), which is powerful and generic middleware package. In addition, this implementation safely handles the proxy to another host (using the same algorithm as [connect](https://github.com/senchalabs/connect)).

## 💻 Examples

In 'package.json', you can defined the proxy table.

```
{
  "proxy": {
    "/api": {
      "target": "http://localhost:8000"
    },
    "/authorization/login": {
      "target": "https://example.com/login",
      "secure": true
    }
  }
}
```

Each key is interpreted as context path (currently glob/regex is not supported), and value as the options passed to [http-proxy-middleware](https://github.com/chimurai/http-proxy-middleware). If you want to redirect all requests to another host, then you can simply write

```
{
  "proxy": "http://localhost:8000"
}
```

## 🚨 Test instructions

1. If `package.json` exists at or above the project root, then 'ProxyHandler' should load the proxy table from `package.json`, then append them to stack.
2. If `package.json` doesn't exist, then stack should remain empty.
3. `ProxyHandler` should call exactly one middleware per request.
4. `ProxyHandler` should ignore case of context path (proxy table keys).
5. Middleware mounted to '/api' should not match with the request to '/api-get'.
6. dev-server should properly redirect the requests as defined in proxy table.

Since `ProxyHandler` uses the same algorithm with [connect](https://github.com/senchalabs/connect), 3., 4., and 5. are actually not needed.

## ✔️ PR Todo


Kogia-sima:
It seems that `loadConfig()` API was changed in the unmerged commits. I'll fix it.

Kogia-sima:
I cannot figure out why integration-tests fail with timeout error. I followed the changes but could not find any infinite loop.

Kogia-sima:
Finally found that it is my fault. Should be good now.

devongovett:
Hey @Kogia-sima! Thanks so much for contributing this feature to Parcel 2. 🎉 

I know I'm contradicting myself compared to what I said on the previous PR, but I am starting to think that the config for this shouldn't be in package.json. I had previously based that on what create-react-app does, but it seems they have moved away from that type of configuration in CRA v2. Instead, they now expose a JavaScript API in a separate file.

I'm thinking we could do something similar in Parcel: a `.proxy` file (JSON) with the config options you've implemented so far, and a `.proxyrc.js` file (JavaScript) which works like CRA and exports a function that accepts an HTTP server to attach to.

Also I see in your PR that you have needed to implement a bunch of routing stuff yourself since we aren't currently using express/connect. I think we could probably avoid that by just using express/connect directly instead. The `app` we expose to the `.proxyrc.js` function would be this express/connect instance. I say express/connect because I'm not sure we need the full express, but maybe. Let me know what you think!

Kogia-sima:
I found that CRA provide the following proxy [API](https://facebook.github.io/create-react-app/docs/proxying-api-requests-in-development).

```javascript
// src/setupProxy.js

const proxy = require('http-proxy-middleware');

module.exports = function(app) {
  app.use(proxy('/api', { target: 'http://localhost:5000/' }));
};
```

This API is very powerful and I like this. However, I think the configuration in JS and JSON file should be similar, so what about this?

```javascript
// .proxyrc.js

const userDefinedMiddleware = (req, res, next) => { ... };

module.exports = {
  "/api": {
    target: "http://localhost:8000",
    pathRewrite: 
  },
  "/authorization/login": userDefinedMiddleware
};
```

These configuration doesn't need the `http-proxy-middleware` be installed manually.

Also, JSON configuration file name should be `.proxyrc`, not `proxy` because [cosmiconfig](https://github.com/davidtheclark/cosmiconfig) package doesn't search the file named `.proxy`.

devongovett:
Yeah I meant `.proxyrc`, sorry. I don't think it's terrible that they are different APIs in different files. I guess we could accept either an exported object or a function though.

Kogia-sima:
I got it. Let me try.

Kogia-sima:
Fixed.

packages/reporters/dev-server/src/Server.js
```diff
@@ -179,8 +181,51 @@ export default class Server extends EventEmitter {
...
   }
 
+  /**
+   * Load proxy table from package.json and apply them.
+   */
+  async applyProxyTable(app: any) {
+    // avoid skipping project root
+    const fileInRoot: string = path.join(this.options.projectRoot, '_');
+
+    const pkg = await loadConfig(this.options.inputFS, fileInRoot, [
```
DeMoorJasper: I'm not sure if the proxy config file should be loaded from the server. Probably better to make it part of the Parcel config? So load it in core or the CLI.

cc @devongovett

Kogia-sima: @DemoorJasper Still not sure why configuration file should be read from Parcel core packages, but if it is needed, how should I define the new option? Should I hold the exported function directly, create a `connect` app, or define a wrapper class?

DeMoorJasper: Ow it's a middleware function. Might be strange to load that in core idk

Kogia-sima: Middlerware for proxies only? Or transfer the all implementation about HTTP request handling into Parcel core?

Kogia-sima: Bundler of v1 has middleware function, so should v2 provide the same function (which handles the proxies as well)?

DeMoorJasper:
@Kogia-sima would be nice to have some tests for this in a follow-up PR

gaorubin1990:
Cross-domain cookies are not available, how can i do ?

PR #3282, opened 2019-07-29
Parcel 2: CoffeeScript sourcemaps

### LINKED ISSUE #2160: Parcel 2: CoffeeScript transform

kwelch:
# ↪️ Pull Request
Adds sourcemaps to v2 coffescript transformer

resolves #2160 

## ✔️ PR Todo

### DIFF
--- a/packages/transformers/coffeescript/src/CoffeeScriptTransformer.js
+++ b/packages/transformers/coffeescript/src/CoffeeScriptTransformer.js
@@ -20,11 +20,7 @@ export default new Transformer({
     // return from compile is based on sourceMaps option
     if (options.sourceMaps) {
       asset.setCode(output.js);
-      asset.setMap(
-        new SourceMap(output.sourceMap, {
-          [sourceFileName]: null
-        })
-      );
+      asset.setMap(SourceMap.fromRawSourceMap(output.v3SourceMap));
     } else {
       asset.setCode(output);
     }
--- a/packages/transformers/coffeescript/src/CoffeeScriptTransformer.js
+++ b/packages/transformers/coffeescript/src/CoffeeScriptTransformer.js
@@ -1,11 +1,34 @@
+// @flow
 import {Transformer} from '@parcel/plugin';
+import SourceMap from '@parcel/source-map';
 import coffee from 'coffeescript';
+import {relativeUrl} from '@parcel/utils';
 
 export default new Transformer({
-  async transform({asset}) {
+  async transform({asset, options}) {
+    let sourceFileName: string = relativeUrl(
+      options.projectRoot,
+      asset.filePath
+    );
+
     asset.type = 'js';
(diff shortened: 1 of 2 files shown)


packages/transformers/coffeescript/src/CoffeeScriptTransformer.js
```diff
@@ -1,11 +1,34 @@
...
+    let output = coffee.compile(await asset.getCode(), {
+      filename: sourceFileName,
+      sourceMap: options.sourceMaps
+    });
+
+    // return from compile is based on sourceMaps option
+    if (options.sourceMaps) {
+      asset.setCode(output.js);
+      asset.setMap(
+        new SourceMap(output.sourceMap, {
```
DeMoorJasper: Shouldn't this be: `SourceMap.fromRawSourceMap(output.sourceMap)` ?

new SourceMap takes in an array of mappings. Not sure if `output.sourceMap` is an array of mappings or an actual sourceMap in case it is it should be as listed above.

kwelch: Oh good to know. I was basing it off of BabelTransformer, and didn't have much examples go to off.

DeMoorJasper: Not sure if there's any sourcemap tests for coffeescript. You can write some and base it of the JS sourcemap tests

kwelch: I mainly want to ensure they are being produced. I found the SourceMaps tests, but nothing that tested the maps file was generated or inlined.

kwelch: Basically, I have no faith in myself and unless I see a test that tells me it worked I won't believe I did it right.

DeMoorJasper: We should probably add one. For now you could just log whatever `output.sourceMap` is and based on that figure out what you should use. We can do the tests in a follow-up PR if that's easier

kwelch: Sounds good! Thanks

kwelch: Logging it showed me that it didn't match the contract. I knew it CS was outputting another map, so I checked that and it matched up. This should be all updated and set to go.

PR #3298, opened 2019-08-01
Use parentNode.removeChild instead of remove to support IE11

### LINKED ISSUE #3291: ChildNode.remove not supported in IE11
# 🐛 bug report

css-loader calls a ChildNode method that is not supported in any version of Internet Explorer. This breaks HMR 

https://github.com/parcel-bundler/parcel/blob/75310e15f6fa3c8a0ce6593ada1bcca00240ea54/packages/core/parcel-bundler/src/builtins/css-loader.js#L6

No Babel or PostCSS config. Our browserlistrc only and explicitly targets ie11

## 🤔 Expected Behavior

Parcel should not allow an exception to be thrown in IE11

## 😯 Current Behavior

IE11 throws an exception, you can skip over it and the reload works, but you have to jump over it in the debugger every time.

IE11 Stack trace:
```
SCRIPT438: Object doesn't support property or method 'remove'
File: Function code (6), Line: 8, Column: 5
```

## 💁 Possible Solution

A simple solution could be to use the `removeChild` method instead:

```javascript
function updateLink(link) {
  var newLink = link.cloneNode();
  newLink.onload = function () {
    if (link.parentNode !== null) {
        link.parentNode.removeChild(link);
    } 
    
  };
  newLink.href = link.href.split('?')[0] + '?' + Date.now();
  link.parentNode.insertBefore(newLink, link.nextSibling);
}
```

## 🔦 Context

An exception being thrown on every CSS change is causing churn.

## 🌍 Your Environment

| Parcel           |1.12.3
| Node             |v12.2.0
| npm         |6.9.0
| Yarn         |n/a
| Operating System |macOS 10.14.6

Taym95:
# ↪️ Pull Request
Try to fix #3291 by using `parentNode.removeChild` instead of `remove` because it's not supported by IE11.

## ✔️ PR Todo

### DIFF
--- a/packages/core/parcel-bundler/src/builtins/css-loader.js
+++ b/packages/core/parcel-bundler/src/builtins/css-loader.js
@@ -3,7 +3,9 @@ var bundle = require('./bundle-url');
 function updateLink(link) {
   var newLink = link.cloneNode();
   newLink.onload = function () {
-    link.remove();
+    if (link.parentNode !== null) {
+        link.parentNode.removeChild(link);
+    }
   };
   newLink.href = link.href.split('?')[0] + '?' + Date.now();
   link.parentNode.insertBefore(newLink, link.nextSibling);


Taym95:
Weird why tests are failing?

DeMoorJasper:
Looks good to me, maybe we should port this over to Parcel 2?

Taym95:
Yes sure, just created #3985.

PR #3317, opened 2019-08-05
Use plain objects instead of deeply nested serialized classes

devongovett:
This refactors Parcel 2 to use plain objects for its core data structures instead of classes, which improves overall performance by ~25% due to less serialization. Classes in the public interface wrap these internal objects on demand. For assets, there is also `InternalAsset`, which wraps these plain objects for internal use.

In addition to the ones we have before, these objects now have public wrapper classes:

* Dependency
* Environment
* Target
* PluginOptions

`PluginOptions` replaces the resolved `ParcelOptions` type, and wraps that type to expose only what we want to expose to plugins. `ParcelOptions` is moved to the internal core types. In addition, the cache object is now passed as part of the options, which might be nice to allow for future alternative cache implementations.

This PR also fixes the parallelism of `PromiseQueue` so that we actually build assets in parallel, which improves performance significantly (to Parcel 1 levels), and also improves various serialization perf (esp with large source maps). The serializer will now treat objects return from a `serialize` method as final, and will not traverse further into them, which improves perf for large objects like the bundle graph and source maps.

The overall perf improvement from this PR is ~54% on a large project I'm testing. 💥


packages/core/types/index.js
```diff
@@ -143,19 +143,20 @@ export type PackageJSON = {
...
   entries?: FilePath | Array<FilePath>,
   rootDir?: FilePath,
   config?: ResolvedParcelConfigFile,
   defaultConfig?: ResolvedParcelConfigFile,
-  env?: {[string]: ?string},
+  env?: {[string]: string},
   targets?: ?(Array<string> | {+[string]: TargetDescriptor}),
 
-  cache?: boolean,
+  disableCache?: boolean,
```
wbinnssmith: For reasoning about this internally and with plugins, it might make more sense for this to be something like `enableCache`, which defaults to true, so we can avoid the double negative. This is okay though.

packages/core/types/index.js
```diff
@@ -173,18 +174,23 @@ export type InitialParcelOptions = {|
...
-  entries: Array<FilePath>,
-  logLevel: LogLevel,
-  rootDir: FilePath,
-  targets: Array<Target>,
-  projectRoot: FilePath,
-  lockFile: ?FilePath,
-  inputFS: FileSystem,
-  outputFS: FileSystem
-|};
+export interface PluginOptions {
```
wbinnssmith: For a situation like this I kind of prefer the exact object type over the interface, so that if someone typos a property name it causes a type error. We can still accomplish readonly properties like this:

```
export type ParcelOptions = {|
  +cacheDir: FilePath,
  +entries: Array<FilePath>,
  +logLevel: LogLevel,
  ...
|};
```

or even with `$ReadOnly`[0]:

```
export type ParcelOptions = $ReadOnly<{|
  cacheDir: FilePath,
  entries: Array<FilePath>,
  logLevel: LogLevel,
  ...
|}>;
```

What was the rationale behind using an interface?

[0] https://flow.org/en/docs/types/utilities/#toc-readonly

PR #3320, opened 2019-08-06
Remove WorkerFarm singleton

wbinnssmith:
This removes the "shared" WorkerFarm, allowing callers to `new Parcel` to pass and control their own `WorkerFarm`. `killWorkers` has been removed from Parcel options, replaced by the option to pass a `WorkerFarm` that can be ended through `workerFarm.end()`.

This also means that multiple WorkerFarms can coexist at the same time: creating a WorkerFarm with different options means that a prior one does not need to be shut down.

Shared singleton state remains in the `childState`, however this is limited only to worker scripts, which always belong to one worker farm. The main process, however, can have many.

Since workers need to be able to call the main process, `WorkerApi` has been separated and is passed to every worker function.

Since `localRequire` depended on the ability to call into the main thread and did so with the shared workerfarm, it is now bound to `WorkerApi` in the worker script and is passed into plugins rather than directly required.

Test Plan: `yarn test`


packages/core/fs/src/MemoryFS.js
```diff
@@ -37,8 +40,8 @@ export class MemoryFS implements FileSystem {
 
   serialize(): SerializedMemoryFS {
     if (!this.handle) {
-      this.handle = WorkerFarm.createReverseHandle(
-        async (fn: string, args: any[]) => {
+      this.handle = this.farm.createReverseHandle(
```
devongovett: FYI I changed this in my other PR: #3317

packages/core/local-require/src/localRequire.js
```diff
@@ -1,29 +1,39 @@
...
+  installPackageFromWorker
+} from '@parcel/install-package';
 import {dirname} from 'path';
 
 import {resolve} from '@parcel/utils';
 
 const cache: Map<string, [string, ?PackageJSON]> = new Map();
 
-export default async function localRequire(
+export async function localRequireFromWorker(
```
devongovett: heh, this is interesting. I guess it's fine for our internal use, but what about plugins? Should they know if they are running in a worker or not?

packages/core/parcel/src/cli.js
```diff
@@ -219,6 +223,7 @@ async function normalizeOptions(command): Promise<InitialParcelOptions> {
     serve,
     targets: command.target.length > 0 ? command.target : null,
     autoinstall: command.autoinstall !== false,
-    logLevel: command.logLevel
+    logLevel: command.logLevel,
+    workerFarm
```
devongovett: isn't there a default worker farm? does the CLI need to make one?

packages/core/types/index.js
```diff
@@ -339,16 +348,25 @@ type ResolveConfigFn = (
...
+  // $FlowFixMe
+) => Promise<any>;
+
 export type Transformer = {
   getConfig?: ({
     asset: MutableAsset,
     resolve: ResolveFn,
-    options: ParcelOptions
+    options: ParcelOptions,
+    localRequire: LocalRequire
```
devongovett: ah interesting. I was working on moving this into a package manager that would get passed around. seems like an ok temporary solution though

packages/core/workers/src/Handle.js
```diff
@@ -1,17 +1,43 @@
...
-      return WorkerFarm.callMaster({handle: opts.id, args}, true);
+  dispose() {
+    handleIdToWorkerApi.delete(this.id);
+  }
+
+  static deserialize(opts: {|id: number|}) {
+    return function(...args: Array<mixed>) {
+      let workerApi = child
+        ? child.workerApi
+        : nullthrows(handleIdToWorkerApi.get(opts.id));
```
devongovett: when does this case happen, i.e. when is there no child state?

packages/core/workers/src/WorkerFarm.js
```diff
@@ -299,69 +323,24 @@ export default class WorkerFarm extends EventEmitter {
     );
   }
 
-  createReverseHandle(fn: () => mixed) {
-    let handle = new Handle();
-    this.handles.set(handle.id, fn);
+  createReverseHandle(fn: (fn: string, args: Array<mixed>) => mixed) {
+    let handle = new Handle({workerApi: this.workerApi});
+    this.handleFns.set(handle.id, fn);
+    this.handles.push(handle);
```
devongovett: maybe the handle functions should just be part of the Handle object itself rather than two maps

packages/core/parcel/src/cli.js
wbinnssmith: it doesn't -- we create one if it's not passed. I'll remove it here.

packages/core/local-require/src/localRequire.js
wbinnssmith: They shouldn't, no. That's why we pass `localRequire` to certain apis (`getConfig` and `transform` in transformers for now) for plugins. I don't think they should be requiring it directly.

Another option would be to expose `childState` or something like it from the worker package, but that wouldn't be compatible with different versions of workerfarm being present.

packages/core/workers/src/Handle.js
wbinnssmith: This can happen any time the handle is executed in the main process — like in the local worker.

PR #3324, opened 2019-08-07
Adds --hmr-proxy-url CLI argument.

nisarhassan12:
# ↪️ Pull Request

We're trying to run parcel in [Gitpod](https://gitpod.io) and HMR was not working.
In Gitpod, web servers are exposed via a proxy, and something running on ```0.0.0.0:1337``` needs to be accessed via a special URL like ```1337-<Workspace ID>.<Region>.gitpod.io:443/``` . We made it work by connecting to it via special proxy like ```1337-abcd-1234.ws-eu0.gitpod.io:443``` 

Unfortunately ```parcel ./src/index.html --hmr-proxy-url="<our proxy url>"``` didnt worked  but  ```HMR_PROXY_URL="<our proxy url>" parcel ./src/index.html``` (command prefixed with an enviroment variable) worked we're not sure Why?

### DIFF
--- a/packages/core/parcel-bundler/src/builtins/hmr-runtime.js
+++ b/packages/core/parcel-bundler/src/builtins/hmr-runtime.js
@@ -24,9 +24,15 @@ var checkedAssets, assetsToAccept;
 
 var parent = module.bundle.parent;
 if ((!parent || !parent.isParcelRequire) && typeof WebSocket !== 'undefined') {
-  var hostname = process.env.HMR_HOSTNAME || location.hostname;
-  var protocol = location.protocol === 'https:' ? 'wss' : 'ws';
-  var ws = new WebSocket(protocol + '://' + hostname + ':' + process.env.HMR_PORT + '/');
+  var hmrUrl = '';
+  if (process.env.HMR_PROXY_URL) {
+    hmrUrl =  process.env.HMR_PROXY_URL;
+  } else {
+    var hostname = process.env.HMR_HOSTNAME || location.hostname;
+    var protocol = location.protocol === 'https:' ? 'wss' : 'ws';
+    hmrUrl = protocol + '://' + hostname + ':' + process.env.HMR_PORT + '/';
+  }
+  var ws = new WebSocket(hmrUrl);
   ws.onmessage = function(event) {
     checkedAssets = {};
     assetsToAccept = [];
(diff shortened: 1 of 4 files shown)


mischnic:
Why aren't the `--hmr-hostname ` and `--hmr-port` options not enough in your case? 
Because the client and server port differ?

nisarhassan12:
@mischnic Yes!  in Gitpod ```--hmr-port``` needs to be greater than 1024 because we don't have root privileges, and ports like 1337 are always exposed to the frontend with a URL like 1337-abc-123.ws-eu0.gitpod.io:443 because of the web app proxy (the actual port here will always be 443, only the hostname changes)

mischnic:
In this case, I would say a `--hmr-client-port` (along those lines, defaults to `--hmr-port`) would be better because `hmr-port/hostname` and `hmr-proxy-url` are essentially doing the same thing and will cause confusion.

You opened this PR against master, but we won't be adding new features to Parcel 1 to concentrate efforts on Parcel 2 (in the `v2` branch).

> Unfortunately parcel ./src/index.html --hmr-proxy-url="<our proxy url>" didnt worked but HMR_PROXY_URL="<our proxy url>" parcel ./src/index.html (command prefixed with an enviroment variable) worked we're not sure Why?

It should work, because of this line
```js
process.env.HMR_PROXY_URL = options.hmrProxyUrl;
```

TPReal:
Any chances for this change in v2? I would love that.

nisarhassan12:
> Any chances for this change in v2? I would love that.
@TPReal  Thanks! we will look into this for sure soon.

DeMoorJasper:
@TPReal proxy support is already in Parcel v2

TPReal:
@DeMoorJasper I don't see the proxy option, neither in the docs nor in the code. The code looks functionally similar to v1 (I'm looking at https://github.com/parcel-bundler/parcel/blob/v2/packages/runtimes/hmr/src/loaders/hmr-runtime.js#L30). Did I miss something? Can you point me to the source or docs related to this functionality?

DeMoorJasper:
@TPReal ow sorry, I thought this was for the dev server not the hmr server, shouldn't have been that quick. There's only proxy support for the dev server in v2 atm.

I'll reopen this, not entirely sure if a cli flag is the way to go though

DeMoorJasper:
Gonna close this due to all tests failing and inactivity.

If anyone could port this to v2 that would be awesome :)

PR #3329, opened 2019-08-07
Don't deduplicate or share bundles in isolated environments

### LINKED ISSUE #3327: Assets with isolated environments should not be deduplicated
E.g. worker scripts cannot reuse assets present in a page script

wbinnssmith:
Resolves #3327

This opts bundles with isolated environments out of optimizations in the bundler, as they are unable to access assets in parent bundles or load shared bundles.

Test Plan: Added integration test

### DIFF
--- a/packages/bundlers/default/src/DefaultBundler.js
+++ b/packages/bundlers/default/src/DefaultBundler.js
@@ -112,6 +112,12 @@ export default new Bundler({
     // Step 2: remove assets that are duplicated in a parent bundle
     bundleGraph.traverseBundles({
       exit(bundle) {
+        if (bundle.env.isIsolated()) {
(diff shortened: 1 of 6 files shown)


packages/bundlers/default/src/DefaultBundler.js
```diff
@@ -112,6 +112,12 @@ export default new Bundler({
     // Step 2: remove assets that are duplicated in a parent bundle
     bundleGraph.traverseBundles({
       exit(bundle) {
+        if (bundle.env.isIsolated()) {
```
wbinnssmith: This could be done on an asset level, but checking the bundle allows us to skip this entire traversal.

Issue #3301, opened 2019-08-02
Parcel 2: HTML inline <script> and <style> tags


devongovett:
This issue is for inline <script> and <style> support in Parcel 2's HTML transformer.

*details here*

Scrum:
@devongovett and where are the details? maybe I could help something?

dy:
Do I get it right that with this feature it will be possible to use import directly from inline scripts?

```js
<script type="module">
import $ from 'spect'

$`.login`.on('submit', e => {
})
</script>
```

kwelch:
Yes, the hope is that we will be able to resolve modules.

devongovett:
Complete.

PR #3331, opened 2019-08-08
Insert sibling bundle references into HTML

devongovett:
This ports the logic to insert sibling bundles into HTML in the packager from Parcel 1, and re-enables the tests. For example, a <script> tag in the original HTML may import CSS files. This will result in a sibling bundle in the same bundle group as the JS, which will be inserted as a <link> element into the HTML by the packager.

This also disables inline assets in HTML for now, as they were resulting in separate bundles being generated. Will be re-enabled once the work on #3301 is done.

### DIFF
--- a/packages/core/integration-tests/test/html.js
+++ b/packages/core/integration-tests/test/html.js
@@ -72,7 +72,7 @@ describe('html', function() {
     ]);
   });
 
-  it.skip('should insert sibling CSS bundles for JS files in the HEAD', async function() {
+  it('should insert sibling CSS bundles for JS files in the HEAD', async function() {
     let b = await bundle(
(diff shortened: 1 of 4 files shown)


packages/core/integration-tests/test/html.js
```diff
@@ -146,73 +136,53 @@ describe('html', function() {
...
+      {
+        name: 'index.html',
+        assets: ['index.html']
+      },
+      {
+        type: 'css',
+        assets: ['index.css']
+      },
+      {
+        type: 'js',
```
devongovett: This test doesn't quite work yet since we don't insert an HMR runtime into the page if there is no JS bundle. Before, we'd insert a JS bundle to handle updating the CSS when it changed. Might need to revisit that separately.

Issue #2237, opened 2018-11-02
"process" not defined for babel-plugins (emotion)


sebald:
# 🐛 bug report

I was trying out `emotion` @ version `10.0.0-beta.7` with the babel-plugin.
With the default config:

```
{
  "plugins": ["emotion"]
}
```

the plugin seems to try to access `process` to determine if it should run in production or development mode (to add labels or minify), but fails since it can not access the variable.

```
Uncaught ReferenceError: process is not defined
    at Object.parcelRequire.src/ui/SomeComponent/SomeComponent.tsx.@emotion/styled-base
```

*Note: I am not 100% sure that this is an issue wie Parcel, could also be that the plugins does something it shouldn't do?*

## 🎛 Configuration (.babelrc, package.json, cli command)

```
// .babelrc
{
  "plugins": ["emotion"]
}
```

```
// package.json
{
  "main": "src/index.ts",
  "dependencies": {
    "@emotion/core": "^10.0.0-beta.7",
    "@emotion/styled": "^10.0.0-beta.7",
    "react": "^16.7.0-alpha.0",
    "react-dom": "^16.7.0-alpha.0"
  },
  "devDependencies": {
    "@types/react": "^16.4.18",
    "@types/react-dom": "^16.0.9",
    "parcel-bundler": "^1.10.3",
    "typescript": "^3.1.6"
  },
  "scripts": {
    "start": "parcel index.html"
  }
}
```

## 🤔 Expected Behavior

`emotion` can not access required ENV vars:

```
Uncaught ReferenceError: process is not defined
    at Object.parcelRequire.src/ui/SomeComponent/SomeComponent.tsx.@emotion/styled-base
```

## 😯 Current Behavior

`emotion` can not access the `process` var and throws.

## 🔦 Context

Trying the next version of `emotion` (v10).

## 💻 Code Sample

```tsx
SomeComponent.tsx

import * as React from 'react';
import styled from '@emotion/styled';

export const SomeComponent = styled.div`
  color: hotpink;
`;
```

## 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Parcel           | 1.10.3
| Node             | v10.8.0
| Yarn             | 1.9.4
| Operating System | MacOS Mojave

XGHeaven:
Same error here

alflennik:
There is a PR here that will resolve this issue (not merged yet) https://github.com/parcel-bundler/parcel/pull/2695

[...]

PR #3334, opened 2019-08-08
Inline environment variables from prior transforms

wbinnssmith:
Previously, we only used a quick regex check over the original source code to see if we should inline environment variables. Now, if the AST is dirty, we always traverse it to locate referenced environment variables.

This also fixes a bug where in development, @babel/register would run over relative babel plugins, which would lead to a plugin depending on itself.

This also adds an integration test to detect that we properly insert dependencies from a prior transform.

We should follow up with similar fixes for inlining `fs` calls and node globals.

Test Plan: Added integration tests.


mischnic:
A related issue: https://github.com/parcel-bundler/parcel/issues/2237 (I think)

PR #2653, opened 2019-02-18
Support less import paths from .lessrc

johanberonius:
# ↪️ Pull Request

Respect the "paths" setting from `.lessrc` to set additional import paths. This was not recognized by the resolver in LESSAsset.

## 💻 Examples

If `.lessrc` contains

```
{
    "paths": [
        "../node_modules/foo/bar/"
    ]
}
```
And file at `./src/index.less` contains:
```
@import "baz";
```
It will find the file at `./src/baz.less`, but if that doesn't exist it will also look for `./node_modules/foo/bar/baz.less`

## 🚨 Test instructions

Added one test to `test/less.js`.

## ✔️ PR Todo

- [X] Added/updated unit tests for this change
- [X] Filled out test instructions (In case there aren't any unit tests)
- [X] Included links to related issues/PRs

## Related issues/PRs

- Open issue #1818 
- Previous closed PR #1823

PR #3347, opened 2019-08-11
V2: Less Transformer

arianon:
This PR ports the Less Asset from Parcel 1 to a Transformer for Parcel 2, and makes the integration tests pass.

The only semantic change as far as I can tell is that the new tilde resolution algorithm (namely, the `~/foo.less` case) in V2 seems to require the existence of a `package.json` in the root, otherwise it should work just the same.


packages/resolvers/default/src/DefaultResolver.js
```diff
@@ -17,7 +17,7 @@ import builtins from './builtins';
 export default new Resolver({
   async resolve({dependency, options}) {
     const resolved = await new NodeResolver({
-      extensions: ['ts', 'tsx', 'js', 'json', 'css', 'styl'],
+      extensions: ['ts', 'tsx', 'js', 'json', 'css', 'styl', 'less'],
```
kwelch: I think we talked about not adding to this default extensions list.

packages/core/utils/src/index.js
```diff
@@ -12,6 +12,7 @@ export {default as getCertificate} from './getCertificate';
 export {default as getRootDir} from './getRootDir';
 export {default as isURL} from './is-url';
 export {default as objectHash} from './objectHash';
+export {default as parseCSSImport} from './parseCSSImport';
```
kwelch: I don't think we need to put this in utils yet. We should be able to put this in with the Transformer.

arianon: I put it in utils so it could be reused for the Sass Transformer in #3353

packages/resolvers/default/src/DefaultResolver.js
arianon: I agree that this is a bit icky, but not doing so would break the case for 

```less
// vars.less
@red: #00dd00;

// main.less
@import "a";
body {
  color: @red;
}
```

But I may be able to make the resolvePathPlugin a bit smarter.

DeMoorJasper: We should probably fallback to the default less resolver

packages/core/utils/src/index.js
DeMoorJasper: Not sure if we should even keep the special css import syntax, it seems like there's not really a standard for it.

The `~node_module/something` import syntax got introduced by the webpack loader as far as I can tell.

I think the best way forward with this is to put the resolve in a try catch and in catch try to use the default less resolver and if that fails test for special webpack specific import syntax like the `~node_module/something` and throw a clear error for it.

packages/resolvers/default/src/DefaultResolver.js
arianon: This change actually does nothing semantically, because `NodeResolver` [not only prioritizes the parents' expression, it adds it even if it's not there](https://github.com/parcel-bundler/parcel/blob/v2/packages/resolvers/default/src/DefaultResolver.js#L114-L118), so cases like `@import "~library/styles";` will still work.

DeMoorJasper: Ow great didn't know about that haven't read through the resolver in a lil while

packages/core/utils/src/index.js
arianon: Thinking about it for a bit I agree with taking this out, it was never supported for the CSS Asset nor the Stylus Asset, plus the similarity between `~module/style` and `~/module/style` can lead to confusing errors due to typos.

And of course, `~/node_modules/module/style` does the trick just fine.

packages/transformers/less/src/LessTransformer.js
```diff
@@ -0,0 +1,92 @@
...
+        }
+
+        supportsSync() {
+          return false;
+        }
+
+        async loadFile(rawFilename, ...args) {
+          let filename = rawFilename;
+          if (filename[0] == '/' || filename[0] == '~') {
+            filename = await resolve(asset.filePath, filename);
```
DeMoorJasper: This should probably be in a try ... catch ... and check for the ~node_module syntax in the catch throwing a user friendly error.
Other than that this looks good.

devongovett: We should probably also support the `paths` feature of less config (#2653). Rather than reimplementing ourselves though, it would be good to fall back to super there instead. I think that would have to happen first, or perhaps instead of node_modules resolution. Webpack also does it that way: if there is a `paths` option, use the default resolver, otherwise do what you have here.

DeMoorJasper: This PR only uses the Parcel resolver if it starts with `/` or `~` and uses super.loadFile in all other cases which shouldn't break paths afaik

devongovett: oh I see... hmm what about node_modules resolution though? those wouldn't start with either of those chars.

DeMoorJasper: Good question, not sure how to solve that properly with paths

arianon: Injecting the path of the `node_modules` folder to `paths` could work, but to do that we'll probably need access to the `NodeResolver` class, particularly its [`findNodeModulePath` method](https://github.com/arianon/parcel/blob/3d477d3fa0a37b89ff984dbf020b4b5cad110c70/packages/resolvers/default/src/DefaultResolver.js#L234-L271)

arianon: Also @DeMoorJasper , I'm not entirely sure what error do you want caught in `loadFile`, care to elaborate?

DeMoorJasper: @arianon originally I thought you were gonna try to run it through Parcel's resolver first and than it would error if it can't find a file, but that wouldn't work for paths.
So your solution seems better. Although now node_modules is kinda an issue.

packages/transformers/less/src/LessTransformer.js
```diff
@@ -0,0 +1,103 @@
...
+
+        supportsSync() {
+          return false;
+        }
+
+        async loadFile(rawFilename, ...args) {
+          let filename = rawFilename;
+
+          if (WEBPACK_ALIAS_RE.test(filename)) {
+            let correctPath = filename.replace(/^~/, '~/node_modules/');
```
DeMoorJasper: Not sure if this should be `~/node_modules/some-module/something.less` or just `some-module/something.less`.
We should support both, we might wanna recommend the easier and less Parcel specific one although not sure as it might clash with relative paths. We could also mention both but that seems a bit much?

packages/transformers/less/src/LessTransformer.js
arianon: Using your suggestion, but flipping it so the default Less resolver runs first, and then use the Parcel resolver when it fails, does the trick for both `paths` and `node_modules`

packages/transformers/less/src/LessTransformer.js
arianon: I don't think that case would be common, assuming someone has `normalize.css` in their `/styles` folder, and then does `@import "normalize.css"` in their `index.less` file, I believe the user will expect their possibly-custom `normalize.css` file to be imported, not the one installed from npm.

PR #3353, opened 2019-08-11
V2: Sass Transformer

arianon:
Ports the V1 Sass Asset to a V2 Transformer, and passes the corresponding integration tests.


packages/transformers/sass/src/SassTransformer.js
```diff
@@ -0,0 +1,76 @@
...
+    ];
+
+    return config;
+  },
+
+  async transform({asset, config}) {
+    let sass;
+    // Use Node Sass if installed, but prefer Dart Sass.
+    try {
+      sass = await localRequire('node-sass', asset.filePath, true);
```
DeMoorJasper: We should probably drop support for `node-sass` it caused issues in Parcel 1

packages/transformers/sass/src/SassTransformer.js
```diff
@@ -0,0 +1,76 @@
...
+    asset.type = 'css';
+    asset.setCode(css);
+    return [asset];
+  }
+});
+
+function resolvePathImporter({asset, resolve}) {
+  return function(rawUrl, prev, done) {
+    let url = rawUrl.replace(/^file:\/\//, '');
+    url = parseCSSImport(url);
```
DeMoorJasper: We should probably drop this, fallback to default sass resolver and check for webpack specific import syntax and throw errors on that. Similar to less

packages/transformers/sass/src/SassTransformer.js
```diff
@@ -0,0 +1,82 @@
...
+    if (WEBPACK_ALIAS_RE.test(url)) {
+      const correctPath = url.replace(/^~/, '~/node_modules/');
+      const error = new Error(
+        `The @import path "${url}" is using webpack specific syntax, which isn't supported by Parcel.\n\nTo @import files from node_modules, use "${correctPath}"`
+      );
+      done(error);
+      return;
+    }
+
+    if (url[0] === '/' || url[0] === '~') {
```
DeMoorJasper: We kinda need to figure out node_modules for this. As `some-module/something.scss` should work besides `~/node_modules/some-module/something.scss`

packages/transformers/sass/src/SassTransformer.js
```diff
@@ -0,0 +1,78 @@
...
+      done(error);
+      return;
+    }
+
+    resolve(prev, url)
+      .then(resolvedPath => {
+        done({file: resolvedPath});
+      })
+      .catch(() => {
+        done(null);
```
arianon: We return `null` instead of an error so that [Sass' resolution algorithm](https://sass-lang.com/documentation/js-api#importer) can continue.

>Imports are resolved by trying, in order:
>  * Loading a file relative to the file in which the `@import` appeared.
>  * Each custom importer.
>  * Loading a file relative to the current working directory.
>  * Each load path in `includePaths`
>  * Each load path specified in the `SASS_PATH` environment variable, which should be semicolon-separated on Windows and colon-separated elsewhere.

packages/transformers/sass/src/SassTransformer.js
arianon: Let's warn about `node-sass` not being supported by Parcel if it's installed then.

packages/transformers/sass/src/SassTransformer.js
```diff
@@ -0,0 +1,96 @@
+// @flow
+
+import {Transformer} from '@parcel/plugin';
+import localRequire, {localResolve} from '@parcel/local-require';
```
wbinnssmith: Unfortunately this will break with the latest changes introduced by #3320.

Starting with #3320, we pass `localRequire` as an option bound to the current workerfarm instance (we need it since we have to centralize auto-installs in the main process). We'll need to do something similar with `localResolve`.

arianon: The node-sass detection doesn't need the ability to auto install, or to require even, only resolution, maybe we can use `resolve` from `@parcel/utils`?

PR #3369, opened 2019-08-12
debounce concurrent builds of the same parcel

Noviny:
# ↪️ Pull Request

Was working on task P2B-3 "Prevent multiple concurrent builds of the same Parcel" with @wbinnssmith and @padmaia - this is an initial implementation to see if this is the right direction for this. 😁

NB: The goal was to 'prevent multiple concurrent builds'. I've implemented this with a debouncing of the builds, which will stop downstream tasks failing, but instead working of the same data as the current active build.

Alternatives could be to have a queue of requested builds to ensure each gets its own results, or outright erroring/returning an error event if someone tries to trigger a second build while one is in progress.

Finally, testing. I *think* the test I wrote is clever and will always Be Fine, since build A and build B will always be requested as a single pass of the js lifecycle, but relying on this feels too clever, and I would love better tests here, though I don't know how to write them. The using build time is also potentially flakey - return the `startTime` as part of the event would solve this but wasn't sure if it was good to add that just for testing.

## ✔️ PR Todo

### DIFF
--- a/packages/core/core/src/Parcel.js
+++ b/packages/core/core/src/Parcel.js
@@ -54,6 +54,7 @@ export default class Parcel {
   >();
   #watcherSubscription; // AsyncSubscription
   #watcherCount = 0; // number
+  #currentBuild = null; // ?Promise<IBundleGraph>
 
   constructor(options: InitialParcelOptions) {
     this.#initialOptions = options;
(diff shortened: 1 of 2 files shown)


padmaia:
This was resolved in https://github.com/parcel-bundler/parcel/pull/3650

PR #3374, opened 2019-08-12
V2 README

padmaia:
## Notable Changes
* Merged Parcel 2 RFC into main README
* Pulled `Getting Started` from RFC and moved it to the top. Moved rest of RFC underneath `Documentation` and added a generated Table of Contents.
* Added a script to lint that fails if the README's toc need to be updated and committed
* Moved the README to the parcel package and linked it in the root of the repo 

## Test Plans
* Updated README in a way that would change Table of Contents and verified script failed
* Updated README in a way that would not change Table of Contents and verified script passed


scripts/validate-readme-toc.js
```diff
@@ -0,0 +1,19 @@
+// TODO: remove deep import once API is exposed (It is merged to master but not published yet)
+const transform = require('doctoc/lib/transform');
+const {readFileSync} = require('fs');
+const path = require('path');
+
+const README_PATH = path.join(__dirname, '../README.md');
```
kwelch: In the script you passed the file path in, is the plan to expand this to allow for using that instead of the constant?

packages/core/parcel/README.md
```diff
@@ -1286,38 +1341,73 @@ asset graph. They mostly call out to different compilers and preprocessors.
 import {Transform} from '@parcel/plugin';
```
kwelch: This part of the docs seem to match up with the PR I created, #3075. Maybe what I wrote is not necessary and we can close that PR.

PR #3383, opened 2019-08-14
Inline and DCE process.env.PARCEL_BUILD_ENV and fix Linux dev env

wbinnssmith:
Currently developing Parcel 2 on Linux is broken as the entire hashbang line is treated as a filepath [0]

To resolve this, in development we can conditionally require `@parcel/babel-register`. Instead of a brittle regex replacement like we do with the hashbang, we can inline a special environment variable value and run dead code elimination to remove it entirely during builds.

This was actually the intention behind using `process.env.PARCEL_BUILD_ENV` (which currently is never set) in `dumpGraphToGraphViz`. [1]

Now it's possible by running two babel plugins during production builds: 
* `babel-plugin-transform-inline-environment-variables` configured to only inline `PARCEL_BUILD_ENV`. This was chosen over `NODE_ENV` as that must remain a runtime check when the user invokes Parcel.
* `babel-plugin-minify-dead-code-elimination`, which actually performs the dead code elimination. Alternatively we could run everything through Terser.

Alternatives:
* We could also transform something like `__DEV__` to make this more concise and typo-proof.
* We could simply inline a static value for the expression, avoiding actually needing to set the environment variable during the build.

Test Plan: Run `yarn build` from the root.
* Examine the output of `parcel/lib/cli.js`. Verify it does not contain `require('@parcel/babel-register')`.
* Examine the output of `dumpGraphToGraphViz.js` and verify the function is now entirely empty.
* Run the `parcel` binary in development mode on Linux and verify it runs whereas before it did not.

[0] https://unix.stackexchange.com/questions/399690/multiple-arguments-in-shebang
[1] https://github.com/parcel-bundler/parcel/blob/cc69cb7ad83c8a1f09f126a7cf61cf1109401bc6/packages/core/core/src/dumpGraphToGraphViz.js#L30-L35

### DIFF
--- a/packages/core/parcel/src/bin.js
+++ b/packages/core/parcel/src/bin.js
@@ -1,2 +1,9 @@
-#!/usr/bin/env node -r @parcel/babel-register
+#!/usr/bin/env node
+
+'use strict';
+
+if (process.env.PARCEL_BUILD_ENV !== 'production') {
+  require('@parcel/babel-register');
+}
+
 require('./cli');
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -6,9 +6,6 @@ const path = require('path');
 const rimraf = require('rimraf');
 const babelConfig = require('./babel.config.js');
 
-const BABEL_REGISTER_INTERPRETER =
-  '#!/usr/bin/env node -r @parcel/babel-register';
-
 const IGNORED_PACKAGES = [
   '!packages/examples/**',
   '!packages/core/integration-tests/**',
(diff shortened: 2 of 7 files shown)


packages/core/parcel/src/bin.js
```diff
@@ -1,2 +1,9 @@
-#!/usr/bin/env node -r @parcel/babel-register
+#!/usr/bin/env node
+
+'use strict';
```
wbinnssmith: We configure Babel to compile everything as if it were an ES module, so this is set for consistency between the source and the compiled file.

PR #3386, opened 2019-08-14
Add toml transformer and make test pass 

### LINKED ISSUE #3361: Parcel 2: TOML transformer
Create the `@parcel/transformer-toml` package in `packages/transforms/toml`. This package is a Parcel 2 transform plugin for [TOML](https://github.com/toml-lang/toml). It should be based on the `TOMLAsset` from Parcel 1.

The `@parcel/transformer-toml` package is responsible for the following things:

* Parsing TOML
* Generating JavaScript

Tom-Julux:
This PR ports the TOML Asset from Parcel 1 to a Transformer for Parcel 2, and makes the integration tests pass. This resolves #3361

## Test Instruction

There was this other test, that checked if an imported TOML asset is minified for production. I did not port this test, as this is not really somthing specific to TOML. TOML is transformed to JS, so this test should just be about whether js is minified. But I believe there already is a test for that.

https://github.com/Tom-Julux/parcel/blob/e6aa36116bfe132581e3e98d5e23bcf5cd14b0c8/packages/core/integration-tests/test/javascript.js#L1025-L1037


packages/transformers/toml/src/TOMLTransformer.js
```diff
@@ -0,0 +1,18 @@
+// @flow
+
+import {Transformer} from '@parcel/plugin';
+import toml from '@iarna/toml';
```
DeMoorJasper: Could you turn this into a localRequire?

packages/transformers/toml/src/TOMLTransformer.js
```diff
@@ -0,0 +1,18 @@
...
+
+export default new Transformer({
+  async transform({asset}) {
```
DeMoorJasper:
```JS
async transform({asset, localRequire}) {
    const toml = await localRequire('@iarna/toml', asset.filePath);
```

packages/transformers/toml/package.json
```diff
@@ -0,0 +1,17 @@
...
+    "type": "git",
+    "url": "https://github.com/parcel-bundler/parcel.git"
+  },
+  "main": "src/TOMLTransformer",
+  "engines": {
+    "parcel": "^2.0.0-alpha.1.1"
+  },
+  "dependencies": {
+    "@parcel/plugin": "^2.0.0-alpha.1.1",
+    "@iarna/toml": "^2.2.3"
```
DeMoorJasper: should be a peerDependency

Tom-Julux:
Sure thing

DeMoorJasper:
Switched some stuff around, it should work. Hope you don't mind

PR #3404, opened 2019-08-14
Add PWAManifestTransformer

### LINKED ISSUE #235: 🐛 Problem with loading manifest.json
### 🐛 Bug report
While building or running in dev mode with parcel, if I include manifest.json in html, then parcel return's me .js file path.

### 🤔 Expected Behavior

```html
<link rel="manifest" href="/dist/manifest.json">
```

### 😯 Current Behavior

```html
<link rel="manifest" href="/dist/05bfa35095e6bfacdb8b334d1023872b.js">
```

### 🌍 Your Environment

| Parcel           |  1.2.0
| Node             | node 8.9.2
| Operating System |  macOS Sierra 10.12.6

### LINKED ISSUE #301: 🙋 PWA support

### LINKED ISSUE #3362: Parcel 2: Web Manifest transformer
Create the `@parcel/transformer-webmanifest` package in `packages/transforms/webmanifest`. This package is a Parcel 2 transform plugin for web manifests. It should be based on the `WebManifestAsset` from Parcel 1.

The `@parcel/transformer-webmanifest` package is responsible for the following things:

* Parsing JSON
* Collecting dependencies
* Rewriting urls
* Generating JSON

We currently only handle the `.webmanifest` extension, but many people want to be able to use `.json` as well (#235, #1039). In order to support this, we need a way for the HTML transformer to specify that the dependency should be treated as a `webmanifest` file regardless of its actual extension.

Drapegnik:
# ↪️ Pull Request

Resolves #3362, Resolves #235, Resolves #301

## ✔️ PR Todo

- [x] ported `WebManifestAsset` -> `PWAManifestTransformer`
- [x] support `.json` webmanifest


packages/core/integration-tests/test/webmanifest.js
```diff
@@ -0,0 +1,46 @@
...
+        assets: ['screenshot.png']
+      },
+      {
+        type: 'js',
+        assets: ['serviceworker.js']
+      }
+    ]);
+  });
+
+  it("should treat webmanifest as an entry module so it doesn't get content hashed", async function() {
```
Drapegnik: This test passes, but when I check `dist` dir there is no `index.html`.

btw, how can I get bundling result after particular test case?

Drapegnik:
Hi, @devongovett! Could you check this please?
If everything is ok, I'll add the rest.

Should I add something for rewriting urls?

packages/core/integration-tests/test/webmanifest.js
devongovett: By default, the tests write to an in-memory filesystem rather than the real file system. If you want, you can change the `outputFS` to a `NodeFS` instance temporarily if you need to test something. https://github.com/parcel-bundler/parcel/blob/v2/packages/core/test-utils/src/utils.js#L27

packages/transformers/webmanifest/src/dependencies.js
```diff
@@ -0,0 +1,36 @@
+// @flow
```
devongovett: you could probably just inline this file into the transformer file. might be a bit simpler to read

devongovett:
Looking good so far @Drapegnik! 😄

packages/transformers/webmanifest/src/dependencies.js
```diff
@@ -0,0 +1,36 @@
...
+};
+
+export default function collectDependencies(
+  asset: MutableAsset,
+  json: {[string]: any}
+) {
+  nullthrows(json);
+  Object.keys(HANDLERS).forEach(key => {
+    if (json[key]) {
+      HANDLERS[key](asset, json[key]);
```
devongovett: The url rewriting should happen here: if you return the result of `asset.addURLDependency`, and replace `json[key]` with it, it will be replaced with a URL in a later build step.

packages/transformers/pwa-manifest/src/PWAManifestTransformer.js
```diff
@@ -0,0 +1,57 @@
...
+    env: {context: 'service-worker'}
+  })
+};
+
+const collectDependencies = (asset: MutableAsset) => {
+  const ast = nullthrows(asset.ast);
+
+  Object.keys(DEPS).forEach(key => {
+    // $FlowFixMe
+    let node = ast[key];
```
Drapegnik:
```
Cannot get ast[key] because:
 • property icons is missing in AST [1].
 • property screenshots is missing in AST [1].
 • property serviceworker is missing in AST [1].
```

is it right way to save `json` to `asset.ast`? (as in v1)

packages/transformers/pwa-manifest/src/PWAManifestTransformer.js
```diff
@@ -0,0 +1,57 @@
...
+  }
+  list.forEach(getSrcHandler().bind(null, asset));
+};
+
+const DEPS = {
+  icons: handleArray,
+  screenshots: handleArray,
+  serviceworker: getSrcHandler({
+    isEntry: true,
+    env: {context: 'service-worker'}
```
Drapegnik: 1. is it right to use `{context: 'service-worker'}` here?
2. could you clarify when we should use `isEntry: true`? can't figure it out from the code.

Drapegnik:
@devongovett,

I decided to divide `WebManifestTransformer` to [PWA](https://developer.mozilla.org/en-US/docs/Web/Manifest) & [WebExtension](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json) transformers, because manifests have different structure, so:

- everything under `rel="manifest` marked as `env: {context: 'pwa-manifest'}`
- `manifest.json` without `pwa-manifest` context is `webext-manifest`

packages/core/integration-tests/test/pwa-manifest.js
```diff
@@ -0,0 +1,67 @@
...
+    );
+
+    await assertBundles(b, [
+      {
+        name: 'index.html',
+        assets: ['index.html']
+      },
+      {
+        type: 'webmanifest',
+        assets: ['manifest.json']
```
Drapegnik: in `BundleGraph` we have `manifest.json`, but then in `dist` - `manifest. webmanifest` - is that a problem?

packages/transformers/pwa-manifest/src/PWAManifestTransformer.js
```diff
@@ -0,0 +1,57 @@
+// @flow
+// https://developer.mozilla.org/en-US/docs/Web/Manifest
+
+import {Transformer} from '@parcel/plugin';
+import type {MutableAsset} from '@parcel/types';
+
+import nullthrows from 'nullthrows';
+
+const getSrcHandler = (opt = {}) => (asset, dep) => {
+  dep.src = asset.addURLDependency(dep.src, opt);
```
Drapegnik: I add this for urls rewriting and get following from [this manifest](https://github.com/Drapegnik/parcel/blob/transformer-webmanifest/packages/core/integration-tests/test/integration/pwa-manifest/manifest.webmanifest):

```json
{
  "name": "pwa-webmanifest-example",
  "icons": [
    {
      "src": "1eda49598769bd3a3cd2b1e31d3cfaff",
      "sizes": "100x100",
      "type": "image/png"
    }
  ],
  "screenshots": [
    {
      "src": "f432622196df181e7734bdfd95fa62d0",
      "sizes": "100x100",
      "type": "image/png"
    },
    { "src": "" }
  ],
  "serviceworker": { "src": "d771eacd0fc5e521bf134585524d4727" }
}
```

devongovett: Yep, those ids will be replaced later with urls once the final bundle names are known.

packages/transformers/pwa-manifest/src/PWAManifestTransformer.js
devongovett: The `AST` type has a wrapper around it.

I think it doesn't really make sense as an ast. I'd just combine the three methods below into only `transform` and not use `asset.ast` at all.

packages/configs/default/index.json
```diff
@@ -5,6 +5,7 @@
       "@parcel/transformer-babel",
       "@parcel/transformer-js"
     ],
+    "manifest.json": ["@parcel/transformer-webext-manifest"],
```
devongovett: This seems dangerous to me. Many things could be called `manifest.json` in theory. I'd say, keep the web extension transformer as an official plugin, but don't include in the default config. That way, if you're writing a web extension, you can opt-in to the processing of `manifest.json`, and it won't affect other apps.

packages/transformers/pwa-manifest/src/PWAManifestTransformer.js
devongovett: `isEntry` is used to prevent the filename from changing based on the content. Service workers in particular need to have a consistent filename between builds, or browsers will treat them as different applications.

`context: 'service-worker'` should be correct there, yes.

packages/configs/default/index.json
Drapegnik: So, should I add this logic to `json-transformer`?
```
    if (asset.env.context === 'pwa-manifest') {
      asset.type = 'webmanifest';
    }
```

packages/transformers/pwa-manifest/src/PWAManifestTransformer.js
Drapegnik: How can I test it?
```bash
yarn build
parceldev build --no-cache packages/core/integration-tests/test/integration/pwa-manifest/inde
x.html
```

produces the same file (with ids)

Drapegnik: @devongovett @DeMoorJasper @padmaia @wbinnssmith,

Could you tell me where is this final replacement (ids -> urls) occurs?
Looks like `.webmanifest` is ignored.

Drapegnik: never mind, I found - `generateDepToBundlePath`)

Drapegnik:
Ready to final review.
I'd like to add `WebextManifestTransformer` in next PR

packages/transformers/pwa-manifest/src/PWAManifestTransformer.js
Drapegnik: the problem was because `.webmanifest` was being processed by `RawPackager`, which returns stream, so `typeof contents !== 'string'` and `generateDepToBundlePath` was skipped

I fixed this by adding `StringPackager` - not sure if this is the best way

Drapegnik:
@devongovett, hi!
Could you please review this pr?

Drapegnik:
@devongovett, hey?)

gnarmedia:
@Drapegnik looks like it has conflicts now, but I'd love to see @devongovett merge this. This will allow me to move many projects over to parcel and greatly reduce my complexity for my builds

Drapegnik:
@gnarmedia, I'm also really looking forward to some response from @devongovett.

packages/transformers/html/src/dependencies.js
```diff
@@ -129,7 +129,8 @@ export default function collectDependencies(
 
     if (tag === 'link' && attrs.rel === 'manifest' && attrs.href) {
       attrs.href = asset.addURLDependency(attrs.href, {
-        isEntry: true
+        isEntry: true,
+        env: {context: 'pwa-manifest'}
```
Drapegnik: I did rebase and now get:
```
Error: Received URL without a pathname.
```

something with `moduleSpecifier` (new api?)

How do I properly add `manifest` dep here?

DeMoorJasper:
@Drapegnik could you rebase this on v2 and fix the failing tests? Lemme know if you need any help with fixing tests

Drapegnik:
@DeMoorJasper, hi! after rebase I've got:
```
  pwa-manifest
    1) should support .webmanifest
    ✓ should support .json webmanifest (87ms)
    2) should treat .webmanifest as an entry module so it doesn't get content hashed
    3) should rename webmanifest *.json to *.webmanifest
```

with `Error: Received URL without a pathname .` could you help me with it?

maybe it's due `. webmanifest` extension? (`json` manifest somehow passed)

DeMoorJasper:
@Drapegnik apologies for not reaching out sooner and giving feedback on this plugin.

I've tried to get the tests to pass a couple times, but also couldn't figure out why they were failing and eventually moved on to other issues.

@mischnic apparently is working on fixing the issues with your PR, not entirely sure if we should close this one and move progress over to that PR?

Drapegnik:
@DeMoorJasper, ok, but what about https://github.com/parcel-bundler/parcel/pull/3439, which depends on this one?

DeMoorJasper:
@Drapegnik I assume that one is not included in @mischnic 's PR

PR #3411, opened 2019-08-15
Fix autoinstall not using proper package names

### LINKED ISSUE #3396: Parcel 2: Install-Package installs incorrect packages
# 🐛 bug report

The module name passed to [`install`](https://github.com/parcel-bundler/parcel/blob/v2/packages/core/local-require/src/localRequire.js#L46) is the full module name instead of the bare identifier.

To give you an example, I noticed this because Parcel tried to run the following into my repo:

```
yarn add --json @parcel/transformer-posthtml/package.json -D
```

## 💁 Possible Solution

Some sanitisation should be done to:

- Ensure that it isn't an absolute or relative path
- Remove the nested file specifier from the request

Meaning that Parcel would only run `yarn add lodash`, not `yarn add lodash/get`.

DeMoorJasper:
# ↪️ Pull Request

This PR adds a utility in install-package that validates and fixes package names before they get installed.

See the issue this fixes for context.

Fixes #3396 

## ✔️ PR Todo

### DIFF
--- a/packages/core/install-package/src/validateModuleSpecifiers.js
+++ b/packages/core/install-package/src/validateModuleSpecifiers.js
@@ -1,12 +1,16 @@
 // @flow
 
-const MODULE_REGEX = /((@[^/\s]+\/){0,1}([^/@\s]+)){1}(@[^/\s]+){0,1}/;
+const MODULE_REGEX = /^((@[^/\s]+\/){0,1}([^/\s.~]+[^/\s]*)){1}(@[^/\s]+){0,1}/;
 
 export default function validateModuleSpecifiers(
   modules: Array<string>
 ): Array<string> {
   return modules.map(module => {
-    // $FlowFixMe, not sure why this happens
-    return MODULE_REGEX.exec(module)[0] || '';
+    let matches = MODULE_REGEX.exec(module);
+    if (matches) {
+      return matches[0];
+    }
+
+    return '';
   });
 }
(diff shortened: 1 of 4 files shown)


packages/core/install-package/src/validateModuleSpecifiers.js
```diff
@@ -0,0 +1,12 @@
+// @flow
+
+const MODULE_REGEX = /((@[^/\s]+\/){0,1}([^/@\s]+)){1}(@[^/\s]+){0,1}/;
+
+export default function validateModuleSpecifiers(
+  modules: Array<string>
+): Array<string> {
+  return modules.map(module => {
+    // $FlowFixMe, not sure why this happens
+    return MODULE_REGEX.exec(module)[0] || '';
```
DeMoorJasper: Should this fallback to an empty string or throw an error, seems like this should always return a package name so this will probably never happen unless there's a bug somewhere.

packages/core/install-package/src/validateModuleSpecifiers.js
```diff
@@ -0,0 +1,16 @@
+// @flow
+
+const MODULE_REGEX = /^((@[^/\s]+\/){0,1}([^/\s.~]+[^/\s]*)){1}(@[^/\s]+){0,1}/;
```
kwelch: You have `~`, but no matching case in the test. I have never seen a `~` in a module, is that a special character to allow for `-`?

DeMoorJasper: I'll add a test, but it's mainly to disallow Parcel from trying to install local packages

DeMoorJasper: Ow there's already a test for it

kwelch: Oh I missed that. Cool!

PR #3414, opened 2019-08-16
Reflect child bundle contents in own bundle hash

### LINKED ISSUE #3393: JS filename not updated in script tag after hash change
`parcel index.html`:

```html
<script type="text/javascript" src="./index.js"></script>
```

```js
console.log("hey");
```

Changing the script file does make HMR reload the page, but the old script is still executed (also when doing a manual a manual refresh)
Restarting Parcel has no effect, deleting the cache updates the served script.

Running `parcel build index.html --no-scope-hoist --no-minify`, making a change and rerunning the build creates a new JS file (with a different hash and reflecting the changes) but doesn't update the JS filename in `index.html`.

2.0.0-alpha.1.1

### LINKED ISSUE #3409: Parcel 2: Cache bug
# 🐛 bug report
Parcel 2 returns the cached version, even after cli command re-run:

[Image from Gyazo](https://gyazo.com/d302b03ac07e7cda3c3673378f298e11)

Screen Shot 2019-08-15 at 18 00 28

## 💻 Code Sample

https://github.com/Drapegnik/parcel2-cache-bug

## 🌍 Your Environment

| Parcel           |`2.0.0-alpha.1.1`
| Node             | `v10.14.1`
| Yarn         | `1.17.3`
| Operating System | `macOS`

wbinnssmith:
Resolves #3393 and Resolves #3409

Since we include a bundle's hash in filenames, anything that could possibly alter the content must be factored into what a bundle's hash is, including that a bundle can reference another bundle with a changing digest.

This extracts the current hashing strategy into `_getContentHash`, which returns a hash of the bundles own content. In `getHash`, recursively retrieve child bundle's content hashes and reflect them in the current bundle's hash. Using `traverseBundles`, which now has the option to start at a given bundle, will prevent cyclic bundle dependencies from being considered.

Test Plan: 
* Enabled the `contentHashing` integration test. 
* Temporarily removed `--no-cache` from the react-hmr example and demonstrate #3393 before applying this patch. Apply the patch and verify #3393 is no longer reproducible.

### DIFF
--- a/packages/core/core/src/BundleGraph.js
+++ b/packages/core/core/src/BundleGraph.js
@@ -40,6 +40,11 @@ type BundleGraphEdgeTypes =
   | 'references';
 
 export default class BundleGraph {
+  // A cache of bundle content hashes. Currently, a new BundleGraph is created in response
+  // to any asset change, so this doesn't need much invalidation. However, currently namers run
+  // before runtimes, and can access `getHash` despite runtimes altering bundle content later.
+  // TODO: Implement invalidation since runtimes can alter bundle contents?
+  _bundleContentHashes: Map<string, string> = new Map();
   _graph: Graph<BundleGraphNode, BundleGraphEdgeTypes>;
 
   constructor(graph: Graph<BundleGraphNode, BundleGraphEdgeTypes>) {
@@ -282,11 +287,14 @@ export default class BundleGraph {
     return this._graph.getNodesConnectedFrom(bundleNode, 'bundle').length > 0;
   }
 
-  traverseBundles<TContext>(visit: GraphVisitor<Bundle, TContext>): ?TContext {
+  traverseBundles<TContext>(
+    visit: GraphVisitor<Bundle, TContext>,
+    startBundle?: Bundle
+  ): ?TContext {
     return this._graph.filteredTraverse(
       node => (node.type === 'bundle' ? node.value : null),
       visit,
-      null,
+      startBundle ? nullthrows(this._graph.getNode(startBundle.id)) : null,
       'bundle'
     );
   }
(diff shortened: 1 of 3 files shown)


packages/core/core/src/BundleGraph.js
```diff
@@ -450,13 +454,29 @@ export default class BundleGraph {
...
+    }
+
     let hash = crypto.createHash('md5');
     // TODO: sort??
     this.traverseAssets(bundle, asset => {
       hash.update(asset.outputHash);
     });
 
+    let hashHex = hash.digest('hex');
+    this._bundleContentHashes.set(bundle.id, hashHex);
```
wbinnssmith: Since computing the hash for a bundle how involves computing content hashes for its entire bundle subgraph (in cases where there's only one entry, this is every bundle), cache them.

wbinnssmith:
@mischnic, @magcius, @drapegnik: can you confirm this fixes the issues you found?

mischnic:
> can you confirm this fixes the issues you found?

Yes!

Drapegnik:
@wbinnssmith, works for me!

devongovett:
Hmm won’t this mean that any time a child bundle is invalidated, all of the ancestors will also be invalidated? Kinda negates the point of content hashing for cacheability no?

mischnic:
> cacheability

Aren't they rather for cache busting?
Just to make sure: this invalidation doesn't propogate across entry points (e.g. a html file including a service worker shouldn't invalidate the html file)?

magcius:
The other option here would be to provide a system for "this asset has a cached item reference as a dependency". That would allow for CSS files that reference image assets, etc. to get invalidated when the image asset changes, without invalidating the entire bundle at once.

wbinnssmith:
> Hmm won’t this mean that any time a child bundle is invalidated, all of the ancestors will also be invalidated? Kinda negates the point of content hashing for cacheability no?

Any bundle could end up referencing any of its child bundles via its name/url, and recursively so. 

> Just to make sure: this invalidation doesn't propagate across entry points (e.g. a html file including a service worker shouldn't invalidate the html file)?

It currently does not, as the fact that entry bundles do not contain content digests is a detail of the `DefaultNamer`, so I figured we should be conservative in core.

This is a conservative approach for sure, and could possibly miss situations where we reference another bundle's name that factors in something other than content. I'm open to suggestions.

packages/core/test-utils/src/utils.js
```diff
@@ -42,7 +42,6 @@ export async function ncp(source: FilePath, destination: FilePath) {
           .on('error', reject);
       });
     } else if (stats.isDirectory()) {
-      outputFS.mkdirp(destPath);
```
wbinnssmith: This wasn't awaited, and it also happens anyway in the next recursive call to `ncp`.

PR #3415, opened 2019-08-16
Upgrade to Flow 105 and use explicit inexact object type syntax

wbinnssmith:
Flow is moving to exact objects by default [0], and introducing an explicit syntax for specifying that an object is inexact (`{ foo: string, ...}`). At this point it appears that Step 2 of the steps in that post is complete.

To ease the transition, the Flow team is providing:

* A lint rule to enforce that object types explicitly specify whether they are exact or inexact (`implicit-inexact-object`). This has been enabled in our codebase in the `.flowconfig`.
* A codemod to add explicit inexact object syntax to the currently implicit object type. I have run this over our codebase, including our own custom library definitions in `flow-libs`, in 0acda441bfa89daa67d295fb88655ec964ce8675
    * Run as `flow-upgrade 0.80.0 0.84.0` since this codemod was introduced in flow 0.84.0. I used 0.80 as an arbitrary earlier version.
* Updates to flow-typed libdefs to use the explicit syntax. I've removed old libdefs and run `flow-typed install` to get the latest versions with this syntax.

I also used `flow/tool`[1] to automatically add suppressions to new violations introduced by upgrading our `source-map` libdef. cc @DeMoorJasper: 

`path/to/flow/tool add-comments --all --bin node_modules/.bin/flow --comment "\$FlowFixMe Added with libdef upgrade in #3415" .`

[0] https://medium.com/flow-type/on-the-roadmap-exact-objects-by-default-16b72933c5cf
[1] https://github.com/facebook/flow/blob/6344cf9f4aa83c4eeda5ac9bb159457d01ea08e6/tool

### DIFF
--- a/flow-typed/npm/source-map_v0.7.x.js
+++ b/flow-typed/npm/source-map_v0.7.x.js
@@ -1,12 +1,12 @@
-// flow-typed signature: 5dc4832be8a4e3b97e16fecc54691aac
-// flow-typed version: b6551bfd76/source-map_v0.7.x/flow_>=v0.25.x
+// flow-typed signature: df79a314b7d0c7ec7dd9f01a8fcd37cf
+// flow-typed version: 4f76a29260/source-map_v0.7.x/flow_>=v0.104.x
 
 declare module 'source-map' {
   declare export type SourceMapUrl = string;
   declare export type StartOfSourceMap = {|
     +file?: string,
     +sourceRoot?: string,
-    +skipValidation?: boolean
+    +skipValidation?: boolean,
   |};
   declare export type RawSourceMap = {|
     +version: number,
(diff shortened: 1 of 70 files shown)


flow-typed/npm/source-map_v0.7.x.js
```diff
@@ -1,12 +1,12 @@
-// flow-typed signature: 5dc4832be8a4e3b97e16fecc54691aac
```
DeMoorJasper: Do we really have to update this? The new mappings this PR introduces aren't correct?

wbinnssmith: Agreed. I'm happy to wait to land this until we can upstream some changes to the libdef.

Did we modify the prior one or do you know if there were bugs introduced with this newer libdef?

DeMoorJasper: We modified the prior one

I apparently never opened a PR to `flow-typed` will make one today

DeMoorJasper: It got merged 🎉

Issue #3359, opened 2019-08-12
Parcel 2: Reason transformer


devongovett:
Create the `@parcel/transformer-reason` package in `packages/transforms/reason`. This package is a Parcel 2 transform plugin for Reason. It should be based on the `ReasonAsset` from Parcel 1.

The `@parcel/transformer-reason` package is responsible for the following things:

* Compiling Reason to JavaScript with `bsb-js`
* Collecting dependencies as included files so Parcel watches them (Parcel 1 does not currently do this properly)

devongovett:
Going to close this. I don't have the bandwidth to implement and maintain this. It would be great as a third party plugin if someone wants to build it!

PR #3417, opened 2019-08-16
add @parcel/transformer-reason

abdullahtariq1171:
# ↪️ Pull Request

This PR tries to add reason plugin for transforms. Fixes https://github.com/parcel-bundler/parcel/issues/3359

Todo:
- [x] Compiling Reason to JavaScript with bsb-js
- [ ] Collecting dependencies as included files so Parcel watches them (Parcel 1 does not currently do this properly)

## ✔️ PR Todo

### DIFF
--- a/packages/configs/default/index.json
+++ b/packages/configs/default/index.json
@@ -5,6 +5,7 @@
       "@parcel/transformer-babel",
       "@parcel/transformer-js"
     ],
+    "*.{ml,re}": ["@parcel/transformer-reason"],
     "*.{json,json5}": ["@parcel/transformer-json"],
     "*.toml": ["@parcel/transformer-toml"],
     "*.yaml": ["@parcel/transformer-yaml"],
--- a/packages/configs/default/package.json
+++ b/packages/configs/default/package.json
@@ -39,6 +39,7 @@
     "@parcel/transformer-stylus": "^2.0.0-alpha.1.1",
     "@parcel/transformer-sugarss": "^2.0.0-alpha.1.1",
     "@parcel/transformer-toml": "^2.0.0-alpha.1.1",
-    "@parcel/transformer-yaml": "^2.0.0-alpha.1.1"
+    "@parcel/transformer-yaml": "^2.0.0-alpha.1.1",
+    "@parcel/transformer-reason": "^2.0.0-alpha.1.1"
   }
 }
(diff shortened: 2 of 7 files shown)


abdullahtariq1171:
I'm not exactly sure how gonna collect dependencies as included files, any help would be very useful.

mischnic:
I have no experience with bucklescript, but the `bsb-js` was last updated one year ago.
The official docs talk about `bs-platform` (https://bucklescript.github.io/docs/en/installation) and to not use an integration with a bundler 😕 (https://bucklescript.github.io/docs/en/interop-with-js-build-systems#popular-js-build-systems)

abdullahtariq1171:
yes, `bsb-js` is just a wrapper on top of `bs-platform`.

I've thinking about this, we can detect the dependencies using regex but since we use `bsb-js`, and it doesn't return any information. So how we going to actually collect these dependencies. 

cc: @devongovett 

---

Also here is previous discussion https://github.com/parcel-bundler/parcel/pull/342, might be helpful some questions.

gugahoa:
Hi all, maybe this PR could take inspiration from the new react-hooks template for bsb found here: https://github.com/BuckleScript/bucklescript/blob/3fe3e854f8a0d31691d396e5a41e728f58f2f6cd/jscomp/bsb/templates/react-hooks/watcher.js

It shows how to do a simply watcher.

DeMoorJasper:
@gugahoa it's more than just watching we need to know all the files that are being compiled/bundled for cache invalidation as well. Unless we simply disable cache for reason, but that's not really optimal.

gugahoa:
@DeMoorJasper What would be the impact of disabling cache for reason?

DeMoorJasper:
@gugahoa it would be a lot slower to build reason

mischnic:
bsb seems to generate a file listing all dependencies of a source file: https://bucklescript.github.io/docs/en/interop-with-js-build-systems#getting-projects-dependencies

But you need to explicitly specify source folders in the `bsconfig.json` file. bsb really doesn't want to be integrated into another build system.

gugahoa:
@mischnic @DeMoorJasper 
So in simple terms, a good enough reason plugin would:
- Run bsb in watch mode with websocket
- Listen to websocket to be notified on change
- On notifiy, <somehow> discover the changed file
- Find the .mlast.d of that file and invalidate those files

Is that it?

DeMoorJasper:
@abdullahtariq1171 could you update this PR to be more complete as requested?

abdullahtariq1171:
hey guys I'm sorry I won't be able to work on this. Anyone want's to grab this from here would be welcomed.

PR #3421, opened 2019-08-16
Create @parcel/babel-preset-env to support custom Babel configs

wbinnssmith:
This creates a new package, `@parcel/babel-preset-env`, for when the user has provided their own babel config. Previously, we offered multi target support by generating custom targets and sending them to `@babel/preset-env` directly when we generate its options.

In cases the user has their own babel config, we no longer generate a default config of our own and lose the ability to specify targets.

Instead, create a minimal babel preset which wraps `@babel/preset-env`, and accepts a stringified target description via babel's `caller` option.

In the case the user's babel config is not used by Parcel, it falls back to `@babel/preset-env`'s behavior. Additionally, any `targets` supplied via a babel option in the user's config will override what Parcel gives it.

This PR also changes our default config to use `@parcel/babel-preset-env`, sending targets via the caller api.

Test Plan: Existing integration tests continue to pass.


packages/transformers/babel/src/babel7.js
```diff
@@ -1,37 +1,49 @@
...
+      targets: JSON.stringify(options.targets)
+    }
+  };
 
   let code = await asset.getCode();
 
   let res;
   if (asset.ast) {
-    res = babel.transformFromAst(asset.ast.program, code, config);
+    res = babel.transformFromAstSync(asset.ast.program, code, config);
```
wbinnssmith: This behaves async if a callback is given, and may be removed at some point.

https://babeljs.io/docs/en/babel-core#transformfromast

packages/transformers/babel/src/env.js
```diff
@@ -1,93 +1,68 @@
...
 /**
  * Generates a @babel/preset-env config for an asset.
  * This is done by finding the source module's target engines, and the app's
  * target engines, and doing a diff to include only the necessary plugins.
  */
 export default async function getEnvOptions(config: Config) {
   // Load the target engines for the app and generate a @babel/preset-env config
-  let targetEngines = config.env.engines;
-  let envOptions = await getPresetOptions(targetEngines, true);
+  let appBabelTargets = enginesToBabelTargets(config.env.engines);
```
padmaia: way better name 🙂

packages/transformers/babel/src/getBabelTargets.js
```diff
@@ -0,0 +1,104 @@
...
+  // Dont compile if we couldn't find any targets
+  if (Object.keys(targets).length === 0) {
+    return null;
+  }
+
+  return targets;
+}
+
+async function loadBrowserslist(config): Promise<?BrowserslistConfig> {
+  let browserslistConfig = await config.getConfig(
```
padmaia: We probably need to invalidate this config when browserslist changes (if we find it) right? Don't think we were doing that before, just bringing it up.

PR #3424, opened 2019-08-17
Remove async/await from Transformation#getCacheKey

wbinnssmith:
This function doesn't do anything async. We should consider https://eslint.org/docs/rules/require-await

Test Plan: flow and test


wbinnssmith:
> We should consider https://eslint.org/docs/rules/require-await

An implementation of this is in #3425

Issue #3356, opened 2019-08-11
Parcel 2: Kotlin transformer


devongovett:
Create the `@parcel/transformer-kotlin` package in `packages/transforms/kotlin`. This package is a Parcel 2 transform plugin for Kotlin. It should be based on the `KotlinAsset` from Parcel 1.

The `@parcel/transformer-kotlin` package is responsible for the following things:

* Compiling Kotlin to JS with `@jetbrains/kotlinc-js-api`
* Collecting all files referenced in the code so that Parcel watches them (not currently supported in Parcel 1)
* Generating source maps

DeMoorJasper:
Do we really want this?
Seems like there's no way to know what assets to watch, the webpack plugin just uses globs and does not even use the official webpack watcher as far as I can tell. Seems pretty hacky for something that has a very small userbase.

Maybe not something we want to support as an official plugin

max-kammerer:
Hi @devongovett, @DeMoorJasper! I'm not expert in JS world and not everything is clear from this and related discussions and prs. Could you clarify what is required from Kotlin/JS side to unblock integration with Parcel?

mischnic:
For a Parcel transformer, we need some compiler function that takes as an input
- the Kotlin file that should be translated into JS 

and returns 

- the output JS file
- all files in the project that influenced how the translation worked (config files, other Kotlin files that were imported by the "entry" Kotlin file from above), so that Parcel known when the Kotlin compiler API has to be called again. `@jetbrains/kotlinc-js-api` didn't return this


<br>


Then this would be possible (assuming that Kotlin/JS has this concept of exporting functions from Kotlin to JS) with Parcel:
```js
import {someExportedKotlinFunction} from "./utils.kt";

console.log(someExportedKotlinFunction());
```
or (only run `fun main` and use some fully-contained web framework)
```html
<div id="app-root"></div>
<script src="main.kt" type="module"></script>
```

max-kammerer:
@mischnic Thank you for clarification!

As I know in current approach Kotlin/JS compiles all Kotlin files in same module into single JS file. 

>all files in the project that influenced how the translation worked (config files, other Kotlin files that were imported by the "entry" Kotlin file from above), so that Parcel known when the Kotlin compiler API has to be called again. @jetbrains/kotlinc-js-api didn't return this

[...]

PR #3444, opened 2019-08-20
Add kotlin transformer

abdullahtariq1171:
# ↪️ Pull Request

Closes https://github.com/parcel-bundler/parcel/issues/3356

## ✔️ PR Todo


packages/transformers/kotlin/src/KotlinTransformer.js
```diff
@@ -1,11 +1,57 @@
 // @flow
 
 import {Transformer} from '@parcel/plugin';
+import fs from '@parcel/fs';
```
Taym95: I think you don't need `fs` from `@parcel/fs` you can use `asset.fs`.

packages/transformers/kotlin/src/KotlinTransformer.js
```diff
@@ -1,11 +1,57 @@
...
+      asset.filePath
+    );
+
+    let id = Math.random()
+      .toString(36)
+      .slice(3);
+    let dir = path.join(os.tmpdir(), id);
+    let filename = path.join(dir, id + '.js');
+
+    await fs.mkdirp(dir);
```
Taym95: You can use this `asset.fs.mkdirp(dir)`.

packages/transformers/kotlin/src/KotlinTransformer.js
```diff
@@ -1,11 +1,57 @@
...
+      output: filename,
+      sources: [asset.filePath],
+      moduleKind: 'commonjs',
+      noStdlib: false,
+      metaInfo: true,
+      sourceMaps: options.sourceMaps
+    });
+
+    let code = await fs.readFile(filename, 'utf8');
+
```
Taym95: Here also you can use this `asset.fs.readFile(filename, 'utf8')`.

packages/transformers/kotlin/src/KotlinTransformer.js
```diff
@@ -1,11 +1,57 @@
...
+    if (options.sourceMaps) {
+      let sourceMap = await fs.readFile(filename + '.map', 'utf8');
+
```
Taym95: Same here you can use `asset.fs`.

packages/transformers/kotlin/src/KotlinTransformer.js
```diff
@@ -1,11 +1,57 @@
...
+      asset.setMap(sourceMap);
+
+      // remove source map url
+      code = code.substring(0, code.lastIndexOf('//# sourceMappingURL'));
+    }
+
+    asset.setCode(code);
+
+    // delete temp directory
+    await fs.rimraf(dir);
```
Taym95: Same here you can use `asset.fs`.

packages/transformers/kotlin/src/KotlinTransformer.js
mischnic: (Correct, you *have* to use `asset.fs`)

packages/core/integration-tests/test/kotlin.js
```diff
@@ -14,7 +14,7 @@ describe.skip('kotlin', function() {
   it('should produce a basic kotlin bundle', async function() {
     let b = await bundle(__dirname + '/integration/kotlin/index.js');
 
-    await assertBundleTree(b, {
+    await assertBundles(b, {
```
abdullahtariq1171: test fails here with `TypeError: expectedBundles is not iterable`.

abdullahtariq1171:
**Question:**

> Collecting all files referenced in the code so that Parcel watches them (not currently supported in Parcel 1)

How can we collect dependencies here? I see in some transformers we get ast which we visit and extract imports but here we don't parse the code instead we get generated output code and maps from `@jetbrains/kotlinc-js-api` (`kotlin-compiler` wrapper)

packages/core/integration-tests/test/kotlin.js
mischnic: the second argument has to be an array, e.g.:
```js
    assertBundles(b, [
      {
        type: 'js',
        assets: ['index.js']
      }
    ]);
```

PR #3445, opened 2019-08-20
Fix contributing alias work flow

padmaia:
# ↪️ Pull Request
Someone mentioned in the spectrum contributing channel that they were following the alias instructions in the `CONTRIBUTING.md` file and running into issues. [[Spectrum post](https://spectrum.chat/parcel/contributing/first-time-contributor-issues-with-running-local-version~966dc9fe-32ca-4498-b28e-c268b0c7689f)]

The first issue was that we are patching console before our cli reporter is set up, so nothing was getting logged when an error was being thrown during initialization. The error that was getting thrown during initialization was related to `@babel/register` trying to resolve plugins relative to the directory where parcel was called instead of the parcel repo.

## 🚨 Test instructions

Followed alias instructions in `CONTRIBUTING.md` and ran `parceldev build index.js` in a simple project outside of the parcel repo and confirmed it built as expected.

### DIFF
--- a/packages/core/core/src/ReporterRunner.js
+++ b/packages/core/core/src/ReporterRunner.js
@@ -40,7 +40,9 @@ export default class ReporterRunner {
       }
     });
 
-    patchConsole();
+    if (this.options.patchConsole) {
+      patchConsole();
+    }
   }
 
   async report(event: ReporterEvent) {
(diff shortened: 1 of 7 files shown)


mischnic:
For me, there is no terminal output when running outside of the monorepo (but Node is using 100% CPU, so it's doing *something*)

PR #3451, opened 2019-08-21
 Fixes #3133 by upgrading serialize-to-js from 1.1.1 to 3.0.0 

### LINKED ISSUE #3133: DoS vulnerability in dependency serialize-to-js@1.1.1
# 🐛 bug report

The dependency [serialize-to-js](https://github.com/parcel-bundler/parcel/blob/master/packages/core/parcel-bundler/package.json#L71) is exposing a DoS vulnerability in version 1.1.1. See [this issue for more details](https://github.com/commenthol/serialize-to-js/issues/7).

I stumbled upon this issue indirectly because Github sent me a different security alert for safer-eval@1.3.2, which is a transitive dependency of serialize-to-js. However, upgrading safer-eval is not sufficient due to the above referenced vulnerability.

## 💁 Possible Solution

Upgrade to serialize-to-js@^3.0.0 if no functionality is lost (the method `deserialized` was dropped due to the issue in version 2.0.0)

## 🌍 Your Environment

| Parcel           | 1.12.3
| Node             | 12.4.0
| npm/Yarn         | 1.16.0
| Operating System |

jonsth131:
# ↪️ Pull Request

Fixes issue #3133 by upgrading serialize-to-js from 1.1.1 to 3.0.0

Changed import in serializeObject.js to work with 3.0.0.

## 🚨 Test instructions

Create YAML import by following [https://parceljs.org/yaml.html](https://parceljs.org/yaml.html) and run.

Create TOML import by following [https://parceljs.org/toml.html](https://parceljs.org/toml.html) and run.

## ✔️ PR Todo

- [X] Filled out test instructions

### DIFF
--- a/packages/core/parcel-bundler/package.json
+++ b/packages/core/parcel-bundler/package.json
@@ -1,6 +1,6 @@
 {
   "name": "parcel-bundler",
-  "version": "1.12.4",
+  "version": "1.12.3",
   "description": "Blazing fast, zero configuration web application bundler",
   "main": "index.js",
   "license": "MIT",
--- a/packages/core/parcel-bundler/package.json
+++ b/packages/core/parcel-bundler/package.json
@@ -1,6 +1,6 @@
 {
   "name": "parcel-bundler",
-  "version": "1.12.3",
+  "version": "1.12.4",
   "description": "Blazing fast, zero configuration web application bundler",
   "main": "index.js",
   "license": "MIT",
@@ -68,7 +68,7 @@
     "posthtml-render": "^1.1.3",
     "resolve": "^1.4.0",
     "semver": "^5.4.1",
-    "serialize-to-js": "^1.1.1",
+    "serialize-to-js": "^3.0.0",
     "serve-static": "^1.12.4",
     "source-map": "0.6.1",
     "terser": "^3.7.3",
--- a/packages/core/parcel-bundler/src/utils/serializeObject.js
+++ b/packages/core/parcel-bundler/src/utils/serializeObject.js
@@ -1,5 +1,5 @@
 const {minify} = require('terser');
-const {serialize} = require('serialize-to-js');
+const serialize = require('serialize-to-js');
 
 function serializeObject(obj, shouldMinify = false) {
   let code = `module.exports = ${serialize(obj)};`;
--- a/yarn.lock
+++ b/yarn.lock
@@ -9398,13 +9398,10 @@ send@0.16.2:
     range-parser "~1.2.0"
     statuses "~1.4.0"
 
-serialize-to-js@^1.1.1:
-  version "1.2.2"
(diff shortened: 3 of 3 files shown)


suchig:
Is there a reason this change is not merged? This shows as a high vulnerability and needs to be fixed.

DeMoorJasper:
@suchig merged it. The npm audit thing is always pretty extreme, it gives high risk to something that's barely medium, don't blindly trust these warnings.

A lot of people make a fuss about these warnings in tools and most of these vulnerabilities are about user input which we don't really have except for people using them and why would someone dos their own computer...

There's probably not a single npm module that abuses this vulnerability and it seems unlikely that someone would trigger this DoS vulnerability in their own code...

Haven't read through the entire doc about the vulnerability but I doubt it even affected Parcel.

DeMoorJasper:
@suchig looked further into this to validate my thoughts on why this probably didn't affect Parcel.

According to the issue linked in the original issue here. The vulnerability does not apply to Parcel in any way.

According to the disclosed audit report I'd even say this is just how JavaScript's eval works and isn't even a vulnerability more as common sense by a dev of any program to just not allow users to write code that goes straight into eval...

See https://github.com/commenthol/serialize-to-js/issues/7#issuecomment-492076598

moeffju:
Hi all, I would like to install the fixed version but due to the monorepo I cannot find a way to properly reference the commit id from `package.json`. Adding the dependency `"parcel-bundler/parcel#dc393bf"` tries to install the entire monorepo, which is obviously not what I want. 

Before I build this version locally and commit it to the “downstream” repo for the time being, is there a plan to publish a new version ASAP, or do you know of a way to make `npm` install the specific commit but build the “sub-package”?

DeMoorJasper:
You can use https://yarnpkg.com/lang/en/docs/selective-version-resolutions/ to fix a version although it contains a breaking change so it will break Parcel.

Locally linking from monorepo is pretty hard unless you clone the entire repo somewhere and link it that way.

Anyway I'm not gonna go in depth here as this isn't even a security issue. Npm audits is pretty much a joke

devongovett:
We will not be releasing further updates to Parcel 1. As Jasper said, this is not a real issue. Please just ignore the warning and keep using the existing published version.

moeffju:
I don’t disagree that it is exceedingly unlikely for the reported vulnerability to be an issue in parcel, however as long as 1.12.3 is the latest version on the npm registry, every project using parcel in their devDependencies will get this very serious-sounding warning. It might be easiest to just release a new minor version. If parcel requires the `deserialize` function and you believe there is no security risk in doing so, you could just copy it into parcel from https://github.com/commenthol/serialize-to-js/blob/a73605a0098233f12794a2b4ee2c8d7e2594a332/lib/deserialize.js#L31.

I totally understand how annoying this must feel, but it seems that the alternatives are more people asking about this and/or desensitizing developers to `npm audit` warnings, both of which seem undesirable.

devongovett:
Just published v1.12.4 with the update.

PR #3467, opened 2019-08-26
Fix for scopehoisting reexport optimization

### LINKED ISSUE #3437: Scope Hoisting with Sentry fails
# 🐛 bug report

When including the `@sentry/browser` dependency into the bundle the build fails.
Watch mode works as expected.

I am using a bare repository without any config files (but had the same error in a project with more configurations as well).

## 🤔 Expected Behavior

Should bundle the JavaScript file.

## 😯 Current Behavior

Failing with error

```
× Got unexpected null
    at nullthrows (<path>\npm-cache\_npx\12688\node_modules\parcel\node_modules\nullthrows\nullthrows.js:7:15)
    at BundleGraph.resolveSymbol (<path>\npm-cache\_npx\12688\node_modules\parcel\node_modules\@parcel\core\lib\BundleGraph.js:324:48)
    at BundleGraph.resolveSymbol (<path>\npm-cache\_npx\12688\node_modules\parcel\node_modules\@parcel\core\lib\public\BundleGraph.js:103:51)
    at markUsed (<path>\npm-cache\_npx\12688\node_modules\parcel\node_modules\@parcel\scope-hoisting\lib\concat.js:219:32)
    at bundle.traverseAssets.asset (<path>\npm-cache\_npx\12688\node_modules\parcel\node_modules\@parcel\scope-hoisting\lib\concat.js:213:9)
    at enter (<path>\npm-cache\_npx\12688\node_modules\parcel\node_modules\@parcel\core\lib\Graph.js:442:16)
    at enter (<path>\npm-cache\_npx\12688\node_modules\parcel\node_modules\@parcel\core\lib\Graph.js:442:16)
    at enter (<path>\npm-cache\_npx\12688\node_modules\parcel\node_modules\@parcel\core\lib\Graph.js:442:16)
    at walk (<path>\npm-cache\_npx\12688\node_modules\parcel\node_modules\@parcel\core\lib\Graph.js:282:26)
    at walk (<path>\npm-cache\_npx\12688\node_modules\parcel\node_modules\@parcel\core\lib\Graph.js:303:22)
```

## 💁 Possible Solution

It only fails when using scope hoisting, in watch mode or with `--no-scope-hoist` it is working as expected.

## 💻 Code Sample

```
// src/index.js
import * as Sentry from '@sentry/browser';
```

`npx parcel@2.0.0-alpha.1.1 build .\src\index.js`.

## 🌍 Your Environment

| Parcel           | 2.0.0-alpha.1.1
| Node             | 10.15.1
| npm         | 4.6.1

I already saw issues when enabling scope hoisting in parcel v1 so I assume this is not a new bug.

mischnic:
# ↪️ Pull Request

1. Fixes the issue described in https://github.com/parcel-bundler/parcel/issues/3437#issuecomment-524283523 by @fredj (which is different from the issue itself)

This function optimized away assets/dependencies that are reexported but not used (with `sideEffects: false`)
https://github.com/parcel-bundler/parcel/blob/65b4c36de684f22f860c68f2f1e784380b9e6c17/packages/core/core/src/AssetGraph.js#L123-L154

but concat then tried to `nullthrows(asset)` eventhough that value would never be used.

The elements of the return value are checked against `falsy`, so I suspect the `nullthrows` was added accidentally: https://github.com/parcel-bundler/parcel/blob/65b4c36de684f22f860c68f2f1e784380b9e6c17/packages/shared/scope-hoisting/src/concat.js#L84-L85

2. Another bug related to the same deferring:
```
ReferenceError: $c05725fa5f4c434a9b59e8c06b8e9707$import$Foo is not defined
```

```js
import Foo from "./d.js";
export { default as Other } from "./d.js";
export function foo() {
	return Foo;
}
```

would result in a `Dependency` with symbols: { default: "$abd289...$import$Other" }, and the identifier for the first import was never replaced in the link phase (essentially because `symbols` is a 1:1 map).

Now, if there's already a dependency with the same imported symbol, the existing local identifier is used.

3. Fixes #3437 : 
Importing and reexporting the same symbol would create a strong (non-weak) dependency for the import but then `addDependency({isWeak: true})` would overwrite that strong dependency.
```js
import Foo from './foo.js';
export { default as Foo } from './foo.js';

export default function() {
    return Foo;
}
```
This means that the normal import is "removed" as well when the reexport isn't used resulting in stray & unlinked `$c057...707$import$Foo` references (as in https://github.com/parcel-bundler/parcel/issues/3437#issuecomment-524283523) *or* `getDependency` returning `undefined` (as in #3437).

Fixes #3437


mischnic:
Hm, that specific bug was fixed, but there is still an issue with the deferred assets: there are some import variables left in the bundle, e.g. `$cd3c54a92d551c6ffe8ee51e784f9a3d$import$coordinatesEqual` (with the example repo at https://github.com/fredj/ol-parcel)

mischnic:
Ready for review, going to fix #3437 in a followup PR

fredj:
Thanks a lot for your work on this @mischnic

packages/shared/scope-hoisting/src/hoist.js
```diff
@@ -455,23 +464,32 @@ const VISITOR = {
...
+            id.name = existing;
+          }
           // this will merge with the existing dependency
           asset.addDependency({
             moduleSpecifier: dep.moduleSpecifier,
             symbols: new Map([[imported, id.name]]),
-            isWeak: true
+            // `symbols` is empty if this dependency was added by JSTransformer/collectDependencies
+            // in that case, override `isWeak`
+            isWeak: dep.symbols.size === 0 ? true : dep.isWeak
```
mischnic: Not sure what the best way to solve this, the behaviour here needs to be:
- If `isWeak` is not set (after the dependency was added by the dependencies collector in JSTransformer), override it with `true`. But `isWeak` is a `boolean`, not `?boolean`.
- otherwise, only set `isWeak` if it already is weak.

I see two possible solutions:
- change `isWeak` to also allow `undefined/null`
- what I did here with the symbols seems fragile

packages/transformers/js/src/visitors/dependencies.js
```diff
@@ -15,13 +15,13 @@ export default {
   ExportNamedDeclaration(node, asset) {
     asset.meta.isES6Module = true;
     if (node.source) {
-      addDependency(asset, node.source);
+      addDependency(asset, node.source, {isWeak: true});
```
mischnic: Can this change affect bundling without scope hoisting?

devongovett: I think this might cause unexpected results when not using scope hoisting. The logic to defer transformation is based on the symbol metadata, which is only collected in the scope hoisting visitor. So with node symbol metadata it would look like nothing is being used, and you could get to the end and some code would be excluded that shouldn't be. Might be worth testing against `lodash-es` or something.

mischnic: I've made `isWeak` have 3 values now: true, false, unknown (null/undefined); that's cleaner at any rate.

PR #3480, opened 2019-08-30
Use `useBuiltins: usage` in default Babel preset-env config

wbinnssmith:
This switches `useBuiltins` to be `usage` in the default babel preset-env config. This has the effect of inserting `require`s to `core-js` in babel-compiled code when environments make use of certain language features requiring polyfills.

With the existing default targets (preset-env's own, essentially es5 environments), this impacts our test suite and our users. To reduce the number of included dependencies, **we are restoring Parcel 1's default engines of > 0.25% (now >= 0.25%), and node 8**.

Parcel now accepts `defaultEngines` in options which will be used in the case there aren't any targets located. Currently this only applies as a fallback for targets found on the filesystem (`package.json`), but it could apply to programmatic usage and in other situations where engines aren't found.

The test suite uses `defaultEngines` to set `last 1 Chrome version` so that core-js does not insert any new polyfills in cases like async imports. Currently core-js will insert Promise polyfills even when processing for modern environments and `modules: false` is set in preset-env. Addressing this should allow us to use our default browser engines.

Test Plan: flow, lint, test


wbinnssmith:
We should probably also use `defaultEngines` [here](https://github.com/parcel-bundler/parcel/pull/3480/files#diff-fe6d044e6bae9c502bb1592b36d7d1bfR160) if there's no browserlist?

PR #3494, opened 2019-09-04
CLI option for setting public path to use in serve mode

gorakong:
This option allows for the specification of the base path of the assets to be served.

Examples for use:

`parcel2 src/vendor.ts --public-url="/static" --port=3000`


wbinnssmith:
@devongovett @DeMoorJasper any concerns with this? My only ones would be that it's only available in serve mode and the name might be misleading.

devongovett:
I was curious about this. What was the reason for it? It feels weird as a CLI option...

green-arrow:
@devongovett - You can currently set the `--public-url` option with parcel v1, which is very important for us since our UI apps sit behind an API gateway and are routed via subpath.

wbinnssmith:
@green-arrow that's not what this PR does: here it only applies to the devserver. Changing public urls can now be accomplished through a field on the target declared in `package.json`.

Unrelated: @devongovett requiring the public url to be declared in the target descriptor in the package json does limit how dynamic it can be. If one wanted to produce builds that only differed in public url, they'd have to either use the programmatic api to pass the target or modify the package.json file between runs...

parcel-benchmark:
## Benchmark Results

#### Timings

| --- | --- | --- |
| Cold | 23.55s | +109.00ms |
| Cached | 13.87s | +246.00ms |

#### Cold Bundles

| dist/legacy/parcel.ff215454.webp | 102.94kb | +0.00b | 448.00ms | -70.00ms 🚀 |
| dist/modern/parcel.76ee4591.webp | 102.94kb | +0.00b | 489.00ms | -47.00ms 🚀 |

#### Cached Bundles

| dist/legacy/parcel.ff215454.webp | 102.94kb | +0.00b | 448.00ms | -70.00ms 🚀 |
| dist/modern/parcel.76ee4591.webp | 102.94kb | +0.00b | 489.00ms | -47.00ms 🚀 |

</p></details>

**[Click here to view a detailed benchmark overview.](https://benchmark.parceljs.org/benchmark/245316387090727438)**

wbinnssmith:
@devongovett would you be open to something like `--server-public-url`?

devongovett:
We discussed this yesterday, and I believe came to the conclusion that we'd remove public urls per target and go back to a CLI option. We could not come up with a usecase for different public urls for each target.

Also discussed possibly moving everything to relative paths instead of absolute, but public url would still be needed for the dev server.

Issue #3503, opened 2019-09-07
Babel compilation gives error "[BABEL]: FILE: .value is not a valid Plugin property"


mkg20001:
# 🐛 bug report

## 🎛 Configuration (.babelrc, package.json, cli command)

Command: `parcel build public/index.pug`

(Full command: `git clean -dxf && yarn install && yarn link parcel && yarn build`)

Repo: https://github.com/mercode-org/www.mercode.org/tree/feat/parcel-v2

## 🤔 Expected Behavior

Parcel should compile the files as-is

## 😯 Current Behavior

```
yarn run v1.17.3
warning ../package.json: No license field
$ rm -rf dist && parcel build public/index.pug && sed 's|"/|"https://mercode.org/|g' -i dist/portal.html
🚨 [BABEL] /home/maciej/Projekte/www.mercode.org/public/js/main.js: .value is not a valid Plugin property
    at Object.keys.forEach.key (/home/maciej/Projekte/www.mercode.org/node_modules/@babel/core/lib/config/validation/plugins.js:52:56)
    at Array.forEach (<anonymous>)
    at validatePluginObject (/home/maciej/Projekte/www.mercode.org/node_modules/@babel/core/lib/config/validation/plugins.js:50:20)
    at instantiatePlugin (/home/maciej/Projekte/www.mercode.org/node_modules/@babel/core/lib/config/full.js:209:55)
    at cachedFunction (/home/maciej/Projekte/www.mercode.org/node_modules/@babel/core/lib/config/caching.js:33:19)
    at loadPluginDescriptor (/home/maciej/Projekte/www.mercode.org/node_modules/@babel/core/lib/config/full.js:200:10)
    at config.plugins.reduce (/home/maciej/Projekte/www.mercode.org/node_modules/@babel/core/lib/config/full.js:69:20)
    at Array.reduce (<anonymous>)
    at recurseDescriptors (/home/maciej/Projekte/www.mercode.org/node_modules/@babel/core/lib/config/full.js:67:38)
    at loadFullConfig (/home/maciej/Projekte/www.mercode.org/node_modules/@babel/core/lib/config/full.js:108:6)
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.

...
```

## 💻 Code Sample

Repo: https://github.com/mercode-org/www.mercode.org/tree/feat/parcel-v2 (note the branch)

## 🌍 Your Environment

| Software         | Version(s) |
[...]

PR #3507, opened 2019-09-09
Fix babel config creation

padmaia:
# ↪️ Pull Request

This PR fixes some more issues that were discovered while trying to run Parcel globally on a project that has custom Babel configuration. One issue was the instance of `@babel/core` used to create the babel config was a different than the instance that ran transform. In the monorepo, they ended up being the same version, so we didn't catch it in our tests. There was also a case where we needed to use `localRequire` over `require`. Last thing was that the reload case was broken. I've added tests cases for running Parcel globally both with the default Babel config and with custom Babel config, as well as the reload case.


mischnic:
This seems to fix #3503, is that a "coincidence" or does it make sense?

padmaia:
@mischnic Yes, this should resolve that issue.

packages/core/integration-tests/test/babel.js
```diff
@@ -313,6 +324,50 @@ describe('babel', function() {
     await outputFS.rimraf(path.join(fixtureDir, 'dist'));
   });
 
+  it('should support building with default babel config when running parcel globally', async function() {
+    let tmpDir = tempy.directory();
+    let distDir = path.join(tmpDir, 'dist');
+    await fs.ncp(
```
devongovett: should we maybe copy this into the memory fs instead to avoid writing to disk during the tests?

PR #3515, opened 2019-09-13
[T-74] Inline sourcemaps

### LINKED ISSUE #2733: JS Inline Sourcemaps
# 🙋 feature request

I would like to have sourcemaps inline at the bottom of my JS files.

## 🤔 Expected Behavior

Instead of building out a map file it would put the source mapping into the source file at the bottom of it/inline.

## 😯 Current Behavior

Currently all map files are generated alongside the built file(s).

## 💁 Possible Solution

Utilize babel config options if parcel is using babel.

## 🔦 Context

I've tried adding a `.babelrc` with the following for inline JS but it didn't do anything. It could be nice to see what the resulting babel config is that parcel uses for debugging purposes. I'm assuming parceljs builds out its own babel config and uses babel.

```
{
  "presets": [
    ["@babel/preset-env", {
      "sourceMaps": "inline"
    }]
  ]
}
```

DeMoorJasper:
# ↪️ Pull Request

Fixes #2733

## ✔️ PR Todo


packages/core/core/src/PackagerRunner.js
```diff
@@ -147,15 +147,20 @@ export default class PackagerRunner {
     let packaged = await packager.package({
       bundle,
       bundleGraph: new BundleGraph(bundleGraph, this.options),
-      sourceMapPath: path.basename(bundle.filePath) + '.map',
+      getSourceMapReference: map => {
+        return bundle.isInline ||
+          (bundle.target.sourceMap && bundle.target.sourceMap.inline)
```
kwelch: Nice! Is there a way to test this path?

DeMoorJasper: Not entirely sure what you mean? I've added a unit test that tests inline sourceMaps

DeMoorJasper: Ow the tests for target are in `packages/core/core/test/TargetResolver.test.js` haven't really modified that since I introduced sourcemaps. Should be covered by the inline test I wrote anyway

JasonGoemaat:
What are these links from the height bot?  Just seems to want me to login

DeMoorJasper:
@JasonGoemaat it's our internal progress tracker/project management application that we use to keep track of everything. Github issues is a pretty big mess at this scale.

Most issues do link to existing Github issues though

PR #3533, opened 2019-09-20
Parcel 2: Update TSCTransformer to use loadConfig

kwelch:
Found the easiest case. Since it only uses `json` it can just load and sort the config.


kwelch:
Picked wrong base

PR #3542, opened 2019-09-23
[v2] Use Symbol.for to support developing locally linked plugins

qw-in:
# ↪️ Pull Request

Spent a good while trying to test out creating plugins for parcel v2. There did not seem to be any good way that I could find to just manually add a plugin so ended up linking one using 

```
npm / yarn link
```

This loaded the plugin, however it would always be imported as `undefined`. Turns out it was because they were using separate instances of `@parcel/plugin` which both defined the unique plugin `CONFIG` key as `Symbol('config')`. This meant that once the plugin was imported it would ALWAYS be `undefined` due to

https://github.com/parcel-bundler/parcel/blob/535b3dcb8744fe0a3db109ffd9b33855a4f1f763/packages/core/core/src/loadParcelPlugin.js#L40

Using `Symbol.for('config')` should hopefully solve this local development issue along with a lot of potential headache if package managers ever keep more than one version of `@parcel/plugin` around

## 🚨 Test instructions

1. Create minimal plugin with `package.json` and `Namer.js` or similar
1. `yarn link` it to where you are using parcel
1. Include in `.parcelrc`
1. Run parcel, `cannot read property <foo> of undefined` no longer seen

## ✔️ PR Todo

- [X] Added/updated unit tests for this change (Assuming not applicable)


packages/core/plugin/src/PluginAPI.js
```diff
@@ -12,7 +12,7 @@ import type {
   Validator as ValidatorOpts
 } from '@parcel/types';
 
-export const CONFIG = Symbol('config');
+export const CONFIG = Symbol.for('parcel-plugin-config');
```
DeMoorJasper: I've updated this to `Symbol.for('parcel-plugin-config');` so we definitely won't have any conflicts as using a global key `config` seems too likely to already exist to me

Issue #3011, opened 2019-05-10
Parcel 2: Differential Bundling


devongovett:
Parcel 2 already supports compiling to multiple targets when your input is a JavaScript file, but in order to make this easy to use, we need support for inserting the outputs of these targets into an HTML bundle automatically. Today, differential script loading in browsers can be done like this:

```html
<script type="module" src="module.js"></script>
<script nomodule src="nomodule.js"></script>
```

While this is useful, it is not sustainable as new features continue to be added to the ECMAScript standard. New proposals like [this one](https://github.com/whatwg/html/issues/4432) may help with that, and we should work with browser vendors to make that happen.

That said, the syntax may get somewhat complicated to implement in your HTML file manually, and the number of targets may change over time as the browsers you support implement new features. Application authors shouldn't need to care about this syntax, or the exact list of targets they need to bundle for.

Ideally, users would specify their browser targets (similar to `browserslist`), and Parcel would automatically generate a number of bundles as needed to support those targets. Users should not need to specify the exact features they want to generate bundles for, as this will be inferred from their browserslist. Initially, that would be `module` and `nomodule`, but over time may increase as new syntax for differential script loading becomes available in browsers.

When processing an HTML file, Parcel will discover `<script>` tags as it does today. When it finds one, it will create a number of dependencies for the referenced file, each for a different environment. It will then replace the script tag with multiple script tags, or whatever syntax is needed for browsers to load the correct targets based on feature support.

Input:

```html
<script src="myscript.js"></script>
```

Output:

```html
<script type="module" src="/dist/myscript.module.js"></script>
<script nomodule src="/dist/myscript.nomodule.js"></script>
```

Parcel will also detect if you do this manually in your HTML file, e.g. by pointing to module and nomodule script tags yourself, and it will produce ES module or script output based on that target.

rtsao:
As I mentioned in https://github.com/parcel-bundler/parcel/issues/1168#issuecomment-482753640, the use of this approach adds a constraint that all dependencies must be valid ES modules (i.e. no non-strict mode code).

It's still pretty common for npm dependencies to be unable to be parsed as ES modules. That said, I think this could be a purely build-time error without any false positives/negatives. So assuming this were some option/flag, Parcel could throw if any incompatible dependencies are used.

green-arrow:
[...]

Issue #1168, opened 2018-04-10
Efficient IE11 support


jpeg729:
IE 11 will continue to be supported by Microsoft for several years, and people tell me that I have to support IE 11. However, I write code in ES2015, so I am going to have to transpile it to es5 and bundle it with all the necessary polyfills for older browsers. But I don't want modern browsers to have to load the huge bundle. I don't think I am alone in wanting that.

I would like <script src="myscript.js"> to be transformed into something like this
```
<script type="module" src="modern-bundle.js"></script>
<script nomodules src="legacy-bundle.js"></script>
```
with a little boilerplate to prevent browsers like safari 10.1 from running the legacy bundle if they have already run the modern bundle.
IE 11 downloads both bundles, but only runs the legacy bundle. 

Answers to objections
- Edge, Chrome and Opera already support type="module".
- The next version of Firefox will support type="module".
- IE 11 downloads both bundles, but IE 11 is generally used on a desktop computer with a solid internet connection.
- The other desktop browsers that download both bundles are becoming less and less used.
- The number of mobile browsers in use that will download both bundles is diminishing.
and yet some of us still need to support IE 11.

Of course, I would like parcel to support this by default, but I would settle for a plugin. How hard would it be to write a plugin that enables this? Where should I start?

fathyb:
I think this is more of an RFC and doesn't have much to do with IE11. If I understand correctly what you want is to target both old browsers and new ones supporting ES6 modules?

How should we handle imports to non-js modules, like CSS? Should we generate intermediary modules for it? How should handle CommonJS (`require`, `exports`)? Example :

- `main.js`
```js
import {foo} from './bar'

console.log('Foo', foo)
```

- `foo.jsx`
```jsx
import React from 'react'
import {fooClass} from './index.css'

module.exports.foo = <h2 className={fooClass}bar</h2>
```

- `foo.css`
```css
.fooClass {
  background: url('./asset.png')
}
```

That said I think this is an extremely cool feature to have.

devongovett:
Sounds like #41?

jpeg729:
I was aiming for something a little more specific than #41.
What I wanted to suggest here is that parcel could by default produce legacy/modern bundles.

rtsao:
One problem I've experienced with this strategy is handling non-strict mode dependencies. Normally when bundling, strict mode can be individually applied to dependencies.

[...]

PR #3545, opened 2019-09-23
Add support for ES Module and CommonJS output, and <script type="module">

devongovett:
Related: #3011 #1168.

This is a refactor of the scope hoisting compiler to support multiple module output formats, including ES modules, CommonJS, and browser script output. This allows much improved support for building libraries with Parcel, and also enables native browser ES module output with `<script type="module">`. 🎉 

The babel transformer has been adjusted to take ES module targets into account when building, and automatically use a higher browser target for ES module supporting browsers when that target is seen. Also our loader infrastructure will automatically include an `import()` polyfill when needed, as not all esmodule supporting browsers support dynamic imports. But, if you only target dynamic import supporting browsers, then the polyfill is not included. Without any loader runtime code or prelude, bundle sizes should be significantly smaller with native esmodules.

In addition, the behavior of the TargetResolver has been changed to treat the package.json `main`, `module`, and `browser` fields as library targets rather than app targets. This means that you will get CommonJS or ES Module output by default with those fields instead of browser script output. Also, loaders for async resources (e.g. css) will be excluded and native imports will be left in place instead, which can be picked up by another bundler later.

This required introducing two new fields to the `Environment` object: `outputFormat`, and `isLibrary`. The TargetResolver sets these fields automatically based on package.json information, but they can also be configured manually. In addition, `outputFormat` is automatically set to `esmodule` by the HTML transformer when a `<script type="module">` tag is seen, and turned off when `<script nomodule>` is seen. This makes it really easy to create multiple targets if you use an HTML entry point.

Limitations: these new output formats are only supported with scope hoisting enabled, so won't work in development mode. Features like HMR just aren't possible with ES modules, and we also don't currently track enough import/export metadata without the scope hoisting compiler. This is probably fine though as this is meant as an optimization, and will result in faster compilation during development (only need to compile the `nomodule` version).

Other changes:

* Quite a few scope hoisting bugs are fixed in this PR, including working with TypeScript, and dynamic imports.
* The `includeNodeModules` flag now supports taking either a boolean or a whitelist of module names to include for more granular control.
* `includeNodeModules` is now processed by the resolver instead of in the transformer. A dependency is always created now, but sometimes isn't resolved to anything. This allows us to keep the metadata about what symbols were imported etc., but not cause any transforming to occur. An `isExcluded` flag can be set by the resolver to cause the ResolverRunner to return `null` instead of throwing an error.
* Babel preset env's `useBuiltins: 'usage'` option has been reverted to `useBuiltins: 'entry'`. Usage is far too aggressive about including polyfills, and seemed to result in much larger bundle sizes unnecessarily. If you need polyfills, you can include them yourself.
* The PostCSS transformer now merges dependencies into CSS module objects.

Future work:

* Compile a single script into multiple with `module`/`nomodule` automatically
* Better support for relative URLs in loaders instead of making everything absolute.
* Include regenerator polyfill automatically (possibly babel-transform-runtime?)
* External module support for browser script targets?
* Module prefetching
* More tests


packages/core/core/src/BundleGraph.js
```diff
@@ -490,6 +511,7 @@ export default class BundleGraph {
       hash.update(this.getContentHash(childBundle));
     }, bundle);
 
+    hash.update(JSON.stringify(bundle.env));
```
wbinnssmith: `JSON.stringify(objectSortedEntriesDeep(bundle.env))`?

packages/runtimes/js/src/cache.js
```diff
@@ -0,0 +1,14 @@
+let cachedBundles = {};
```
wbinnssmith: maybe `cacheLoader.js` for the name of the file?

packages/shared/scope-hoisting/src/formats/commonjs.js
```diff
@@ -0,0 +1,213 @@
+// @flow
```
wbinnssmith: can these files be `// @flow strict-local`?

mischnic: Seems like they can't:
```
Importing from an untyped module makes it `any` and is not safe! Did you mean to add `// @flow` to the top of `@babel/types`? (`untyped-import`)
```

packages/shared/scope-hoisting/src/link.js
```diff
@@ -37,15 +46,45 @@ export function link({
...
+          importedFiles.set(b.filePath, {
+            bundle: b,
+            assets: new Set()
+          });
+        }
+      }
+    }
+  }
+
+  function assertString(v): string {
```
mischnic: Should be defined outside of the `link` function

packages/shared/scope-hoisting/src/formats/commonjs.js
devongovett: yeah... I really wish we had flow types for Babel

packages/shared/scope-hoisting/src/prelude.js
```diff
@@ -12,6 +13,12 @@ var globalObject =
     : {};
 
 globalObject.parcelRequire = function(name) {
+  // Execute the bundle wrapper function if there is one registered.
```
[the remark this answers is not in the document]
devongovett: haha oh nice! didn't even see that :D

packages/core/core/src/Environment.js
```diff
@@ -59,10 +61,25 @@ export function createEnvironment({
     }
   }
 
+  if (outputFormat == null) {
```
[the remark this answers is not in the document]
devongovett: yeah I thought "module" was confusing - commonjs files are also modules. 🤷‍♂

packages/shared/scope-hoisting/src/formats/commonjs.js
```diff
@@ -0,0 +1,213 @@
...
+    return t.objectProperty(id, id, false, true);
+  });
+
+  let statement = REQUIRE_TEMPLATE({
+    BUNDLE: t.stringLiteral(relativeBundlePath(from, bundle))
+  });
+
+  if (specifiers.length > 0) {
+    statement = ASSIGN_TEMPLATE({
+      SPECIFIERS: t.objectPattern(specifiers),
```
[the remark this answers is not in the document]
devongovett: that's true... I guess I was thinking if you are outputting commonjs you are either targeting node (of which all of the supported versions support destructuring). But I guess you could be building a library and a future bundler might not transpile it...

packages/shared/scope-hoisting/src/generate.js
```diff
@@ -1,26 +1,45 @@
...
+  // at the right time (after other bundle dependencies are loaded).
+  let contents = '';
+  if (isAsync && bundle.env.outputFormat === 'global') {
+    contents = `${hashBang}parcelRequire.registerBundle(${JSON.stringify(
+      nullthrows(entry).id
+    )},function(){${code}});`;
+  } else {
+    contents =
+      bundle.env.outputFormat !== 'global'
+        ? hashBang + code
```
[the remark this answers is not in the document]
devongovett: Oh that's interesting! Good catch.

packages/core/core/src/Environment.js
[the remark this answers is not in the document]
devongovett: I guess we probably need to verify the other environment values as well, e.g. `context`. Gonna make a separate task for this.

packages/shared/scope-hoisting/src/formats/commonjs.js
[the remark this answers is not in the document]
devongovett: Updated to generate multiple variable declarations with member expressions if engines that don't support destructuring are used.

packages/shared/scope-hoisting/src/generate.js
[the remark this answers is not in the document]
devongovett: Updated the terser optimizer as suggested

packages/shared/scope-hoisting/src/link.js
```diff
@@ -161,6 +209,77 @@ export function link({
...
+      }
+
+      specifiers.set(imported, renamed);
+    }
+
+    return specifiers.get('*');
+  }
+
+  function addBundleImport(mod, path) {
+    // Find the first bundle containing this assed, and create an import for it if needed.
```
mischnic:
```suggestion
    // Find the first bundle containing this asset, and create an import for it if needed.
```

packages/core/utils/src/relativeBundlePath.js
```diff
@@ -0,0 +1,16 @@
+// @flow
+import type {Bundle} from '@parcel/types';
+import path from 'path';
+import nullthrows from 'nullthrows';
+
+export function relativeBundlePath(from: Bundle, to: Bundle) {
+  let p = path.relative(
+    path.dirname(nullthrows(from.filePath)),
```
mischnic:
```suggestion
    path.posix.dirname(nullthrows(from.filePath)),
```
also here

packages/resolvers/default/src/DefaultResolver.js
```diff
@@ -179,6 +194,20 @@ class NodeResolver {
...
   }
 
+  shouldIncludeNodeModule(env, name) {
+    if (env.includeNodeModules === false) {
+      return false;
+    }
+
+    if (Array.isArray(env.includeNodeModules)) {
+      let parts = this.getModuleParts(name);
+      // $FlowFixMe - flow is dumb
```
mischnic: Because the call to `getModuleParts` invalidates the check, I suggest:
```js
  shouldIncludeNodeModule({includeNodeModules}, name) {
    if (includeNodeModules === false) {
      return false;
    }

    if (Array.isArray(includeNodeModules)) {
      let parts = this.getModuleParts(name);
      return includeNodeModules.includes(parts[0]);
    }

    return true;
  }
```

devongovett:
> I don't think the current design for includeNodeModules is very useful, if you write a React library, you might want to bundle everything except react.

Yeah this is the opposite usecase: when building a library you usually don't want to include any node_modules, but sometimes want to include one or two, for example, modules that are in the same monorepo. I have a repo which includes a bunch of react components that will be published as libraries, and I want to include the CSS along with it parts of which come from another package. We will eventually support a blacklist as well for externals, which are more useful when building applications.

mischnic:
We can do this in a followup, but I found a case that could be better optimized:
```html
<script src="./main.js"></script>
```
```js
import("./lib.js").then(v => {
	console.log(v);
});
```
Browserslist: Chrome 70 (so dynamic import is supported)

Using dynamic import works also with the `global` output format (of `main.js` in this case), currently both are `global` and therefore the dynamic import polyfill is used.
The `lib.js` bundle should be `esmodule` and get imported with the native `import()`

devongovett:
Oh `import()` works even without `<script type="module">`?

mischnic:
> Oh import() works even without <script type="module">?

Yes!

packages/core/utils/src/relativeBundlePath.js
```diff
@@ -0,0 +1,15 @@
+// @flow
+import type {Bundle} from '@parcel/types';
+import path from 'path';
+import nullthrows from 'nullthrows';
+
+export function relativeBundlePath(from: Bundle, to: Bundle) {
```
mischnic: This is really an URL rather than a path (because it returns a URL with forward slashes also on Windows).

devongovett:
> Yes!

Awesome, had no idea. Gonna leave that as a followup task.

PR #3551, opened 2019-09-24
Parcel 2: Fix inline sources in sourcemaps

DeMoorJasper:
# ↪️ Pull Request

This PR fixes an issue with inlining sources in sourcemaps as well as adds a test to prevent issues like this in the future.

## ✔️ PR Todo


packages/core/logger/src/Logger.js
```diff
@@ -67,7 +67,7 @@ let consolePatched = false;
 // TODO: Implement the rest of the console api as needed.
 // TODO: Does this need to be disposable/reversible?
 export function patchConsole() {
-  if (consolePatched) {
+  if (consolePatched || process.env.NODE_ENV === 'test') {
```
DeMoorJasper: How would I achieve this without breaking the tests?

Setting the option `patchConsole: false` does not seem to really do anything

DeMoorJasper: I'm willing to work on this if this would be something anyone wants.

devongovett: Yeah we should probably make the `patchConsole` option also apply in workers... @wbinnssmith wdyt?

PR #3555, opened 2019-09-24
Create bundles during post-traversal

wbinnssmith:
This implements @devongovett's idea of creating bundles during post-traversal [0], removing the need for `summarizeBundle`.

`summarizeBundle` existed because bundle <=> asset/dependency relationships were actually established after plugin hooks had run, as it could have been too costly to add and remove subgraphs as additional bundles were created. This way, adding subgraphs can stop when a bundle border from a bundle lower in the graph is encountered.

Some notes:

* Connections from bundles to bundle groups along the `bundle` edge type are now created in `addAssetGraphToBundle` and removed (if needed) in `removeAssetGraphFromBundle`.
    * This addresses https://github.com/parcel-bundler/parcel/pull/3259#discussion_r306637515
      > Also, as it is right now, could this miss connecting bundles to this bundle group that are created later?
    * `removeAssetGraphFromBundle` safely removing the connection is untested (tests continue to pass after removing this code). Any thoughts on a good test case?
* Child bundles are now traversed in opposite order, as they are now created in the opposite order. 
    * This affected `assertBundles`, which depends on ordering. This was mitigated in some cases by additionally sorting by the `name` property.
    * Naming is now based on the bundle containing the bundle group's main asset, rather than the first bundle to appear in the group, as this has reversed.
* Changes the edge type from dependencies to runtime assets from `references` to an untyped edge. The removal of the contains edge in `removeAssetGraphFromBundle` is sufficient to stop traversal here.

[0] https://github.com/parcel-bundler/parcel/pull/3259#discussion_r307003889


wbinnssmith:
Just updated this so it now supersedes #3556. The changes to `removeAssetGraphFromBundle` conflicted.

PR #3558, opened 2019-09-24
Fix for 0-byte raw asset bundles

gorakong:
Bug:

After raw assets were written, the stream would be exhausted. Reading from the same stream would then return 0 bytes.

Fix:

After raw assets are cached, we request a new readable stream from the cache.


packages/core/integration-tests/test/javascript.js
```diff
@@ -662,7 +662,8 @@ describe('javascript', function() {
 
   it('should support importing a URL to a raw asset', async function() {
     let b = await bundle(
-      path.join(__dirname, '/integration/import-raw/index.js')
+      path.join(__dirname, '/integration/import-raw/index.js'),
+      {disableCache: false}
```
devongovett: hmm, not sure about changing this for the tests. why did you need this change?

gorakong: @devongovett It's to turn on caching to simulate the asset being written to cache, in order to get the number of bytes written to disk (to prove that it's not 0)

PR #3564, opened 2019-09-25
Fix some type imports

padmaia:
# ↪️ Pull Request
Noticed when running a prod build of Parcel that there were a few places where we were importing types incorrectly. Running locally with `@babel/register` doesn't present the issue because it transpiles away the types when they are imported.

## 🚨 Test instructions
Built a project with a prod build of Parcel and made sure it ran with no errors.


devongovett:
Amazing, just ran into this too.

PR #3565, opened 2019-09-26
[T-345] Parcel 2: `loadConfig` in PostHTML

kwelch:
This deprecates `getConfig` in `PostHTMLTransformer`, in favor of `loadConfig`.

- [x] Adding new config tests
- [x] Update to handle static rc files via `loadConfig`
- [x] Update to handle js config without `require` statements
- [x] Update to handle js config with `require` statements
- [x] should call `config.shouldInvalidateOnStartup()` for `js` file configs
- [x] Add perf warning in verbose mode
- [ ] watch mode integration tests
- [x] should call `setResolvedPath` to invalidate on changes

### DIFF
--- a/packages/transformers/posthtml/src/PostHTMLTransformer.js
+++ b/packages/transformers/posthtml/src/PostHTMLTransformer.js
@@ -9,6 +9,20 @@ import nullthrows from 'nullthrows';
 import semver from 'semver';
 import loadPlugins from './loadPlugins';
 
+const canSerializeConfig = config => {
+  if (!config || !config.plugins) {
+    return true;
+  }
+  if (Array.isArray(config.plugins)) {
+    return config.plugins.every(plugin => typeof plugin === 'string');
+  } else if (typeof config.plugins === 'object') {
+    return Object.keys(config.plugins).every(
+      plugin => typeof plugin === 'string'
+    );
+  }
+  return true;
+};
+
 export default new Transformer({
   async loadConfig({config, options}) {
     let configResult = await config.getConfig(
@@ -19,19 +33,25 @@ export default new Transformer({
     configResult = configResult || {};
     configResult.skipParse = true;
 
-    // if (canSerializeConfig(config)) {
-    //   config.shouldReload();
-    //   config.setResult({
-    //     internal: false
-    //   });
-    //   config.setResultHash(JSON.stringify(Date.now()));
-    // }
+    // isReload case
+    if (config.result != null) {
+      configResult.plugins = await loadPlugins(
+        configResult.plugins,
+        config.searchPath,
+        options
+      );
+      config.setResult(configResult);
+    } else if (!canSerializeConfig(configResult)) {
+      config.shouldReload();
+      config.setResult({isReload: true});
+      config.setResultHash(JSON.stringify(Date.now()));
+    } else {
+      if (configResult.plugins) {
+        config.shouldRehydrate();
+      }
 
-    if (configResult.plugins) {
-      config.shouldRehydrate();
+      config.setResult(configResult);
     }
-
-    config.setResult(configResult);
   },
 
   async rehydrateConfig({config, options}) {
--- a/packages/transformers/posthtml/src/PostHTMLTransformer.js
+++ b/packages/transformers/posthtml/src/PostHTMLTransformer.js
@@ -102,7 +102,7 @@ export default new Transformer({
 
     const ast = nullthrows(asset.ast);
 
-    let res = await posthtml(config.plugins).process(
+    let res = await posthtml(config.fullyLoaded.plugins).process(
       ast.program,
       config.fullyLoaded,
     );
--- a/packages/transformers/posthtml/src/PostHTMLTransformer.js
+++ b/packages/transformers/posthtml/src/PostHTMLTransformer.js
@@ -10,22 +10,39 @@ import semver from 'semver';
 import loadPlugins from './loadPlugins';
 
 export default new Transformer({
-  async getConfig({asset, options}) {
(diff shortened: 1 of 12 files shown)


parcel-benchmark:
## Benchmark Results

#### Timings

| --- | --- | --- |

#### Cold Bundles

#### Cached Bundles

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 15.06s | -8.00ms |
| Cached | 9.57s | +30.00ms |

#### Cold Bundles

| dist/main/index.js | 34.53kb | +0.00b | 2.36s | -122.00ms 🚀 |
| dist/module/index.js | 34.33kb | +0.00b | 2.36s | -122.00ms 🚀 |

#### Cached Bundles

| dist/main/index.js | 34.53kb | +0.00b | 14.00ms | -1.00ms 🚀 |

</p></details>

**[Click here to view a detailed benchmark overview.](https://benchmark.parceljs.org/benchmark/247525515700732426)**

padmaia:
Nice work! Just a few comments:

* Should `config.shouldInvalidateOnStartup()` be called when `.js` config files are found? That's something we do in the Babel case, because we can't be certain that the config file won't return different results even if the file has changed, so we at least invalidate it when starting Parcel. We also leave a perf warning in verbose mode to suggest using the static config formats if possible.

* Also, I noticed in the posthtml docs that there is a  `posthtml-load-config` package for loading config. Is there a reason we are not using it? Makes sense if it's not doing anything more special than Parcel's generic config loading, but just wanted to make sure we're not losing out on anything. If we're intentionally not using it, it might help to leave a comment so this question doesn't get raised again.

* It would be nice to see some watch mode integration tests to make sure changes to the config files trigger a rebuild. There should be some examples of this in the babel integration tests, though it was a hassle to get working with the in memory filesystem. I don't completely recall what the issue was, I think it may have been that we were using `@babel/core` to load the config which is not aware of the in memory filesystem, so you may be able to avoid that. Example: https://github.com/parcel-bundler/parcel/blob/v2/packages/core/integration-tests/test/babel.js#L377-L404

kwelch:
Thanks for the review! Here are some comments and questions I have. 

> * Should `config.shouldInvalidateOnStartup()` be called when `.js` config files are found?

I saw this and I wanted to replicate that functionality. Unfortunately, the `configResult` that is returned does not include the file path, which `loadPartialConfig` returned allowing for that. What do you think about extending the default config loader to pass back the resolve config file path to allow for this? 

>We also leave a perf warning in verbose mode to suggest using the static config formats if possible.
I will go back and add these. I always forget to go back and add these sort of messaging. Thanks for the callout. 

> * Also, I noticed in the posthtml docs that there is a  `posthtml-load-config` package for loading config. Is there a reason we are not using it?

I also saw that and honestly I could not understand what it was doing. It appeared that it expected you to have the config file loaded already, so I didn't understand the value add. 

> * It would be nice to see some watch mode integration tests to make sure changes to the config files trigger a rebuild.

Awesome! I tried to follow babel as much as possible, but I did not see this. I will totally take a look and get these in place for this.

padmaia:
Hey @kwelch, here's a branch that I think should mostly work once you rebase:

https://github.com/parcel-bundler/parcel/compare/kwelch/v2-loadConfig-posthtml...padmaia/v2-loadConfig-posthtml 

I did it quickly just get you headed in the right direction, but it could probably use some cleanup. Let me know if you have any questions.

padmaia:
I also noticed the resolved path isn't getting set anywhere, which means the config request won't be invalidated when the file changes. Can we get that path somehow?

kwelch:
> Hey @kwelch, here's a branch that I think should mostly work once you rebase:
> 
> [kwelch/v2-loadConfig-posthtml...padmaia/v2-loadConfig-posthtml](https://github.com/parcel-bundler/parcel/compare/kwelch/v2-loadConfig-posthtml...padmaia/v2-loadConfig-posthtml)
> 
> I did it quickly just get you headed in the right direction, but it could probably use some cleanup. Let me know if you have any questions.

Thanks I got the tests passing. 

> Can we get that path somehow?

I will look into updating the `config.getConfig` to include the path details. This will also mean we can call `config.shouldInvalidateOnStartup()` for `js` configs. 

I am also going to update the summary to match some of the requests from the last review.

padmaia:
Oh yeah, I forgot `config.getConfig` sets the `resolvedPath` automatically. Nevermind then! It would be cool for it to also set `shouldInvalidateOnStartup` automatically when it runs into js config files as well. Good idea :)

packages/transformers/posthtml/src/PostHTMLTransformer.js
```diff
@@ -9,23 +9,71 @@ import nullthrows from 'nullthrows';
...
+    loaded.skipParse = true;
+
+    if (!canSerializeConfig(loaded)) {
+      config.setResult({fullyLoaded: loaded});
+      config.shouldInvalidateOnStartup();
+      config.setResultHash(JSON.stringify(Date.now()));
+    } else {
+      let fullyLoaded = {
+        ...loaded,
+        plugins: await loadPlugins(loaded.plugins, config.searchPath, options),
```
kwelch: @padmaia, I am sure we want to run this during `loadConfig`. Since this will make the config unserializable. 

I think the [`postDeserializeConfig`](https://github.com/parcel-bundler/parcel/pull/3565/files#diff-e170e4a4898565038389b704114bb0acR74), is covering this.

kwelch:
WooHoo! its green again!

I still need to get some of the watch mode tests together. I looked over the babel ones, but my kids got up from their nap and it was play time 😅. 

I will try to get on this more in the evenings.

Issue #3442, opened 2019-08-20
Parcel 2: Bundles must have unique filePaths when using duplicate content in files


garthenweb:
# 🐛 bug report

When including two files with the same content and name but in different folders, the build fails with `Bundles must have unique filePaths`.

## 🎛 Configuration (.babelrc, package.json, cli command)

None, see example here: https://github.com/garthenweb/parcel2-unique-filePaths-example

## 🤔 Expected Behavior

In a perfect world parcel should detect that both assets are equal and could just include one in the bundle. But I can imagine that this might have side effects.
I would be fine if the build would not fail. A warning would be nice as well as I was not aware that I have duplicate files in my bundle before running parcel 2.

## 😯 Current Behavior

Errors with
```
🚨 Bundles must have unique filePaths
    at BundlerRunner.nameBundles (<path>/node_modules/@parcel/core/lib/BundlerRunner.js:121:21)
    at async BundlerRunner.bundle (<path>/node_modules/@parcel/core/lib/BundlerRunner.js:92:5)
    at async Parcel.build (<path>/node_modules/@parcel/core/lib/Parcel.js:289:25)
    at async Parcel.run (<path>/node_modules/@parcel/core/lib/Parcel.js:196:18)
    at async Command.run (<path>/node_modules/parcel/lib/cli.js:132:7)
```

## 💁 Possible Solution

Including the file path into the hash generation would maybe help.

## 🔦 Context

I was running parcel 2 on a bigger code base and had round about 10 files of type scss, svg and js in the bundle that caused this error. I detected the files by adding the following into the BundleRunner.js file which throw the error:

```
console.log(bundlePaths.filter((val, index,list) => {
  return list.indexOf(val) !== index;
}))
```

While creating the example project I found out that css files worked perfectly fine but for scss files it was failing.

## 💻 Code Sample

See example here: https://github.com/garthenweb/parcel2-unique-filePaths-example

## 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Parcel           | 2.0.0-alpha.1.1
| Node             | v12.8.1
| npm         | 6.10.2
| Operating System | Max OSX

mischnic:
> In a perfect world parcel should detect that both assets are equal and could just include one in the bundle. But I can imagine that this might have side effects.

Exactly, that's unfortunately not possible: https://github.com/parcel-bundler/parcel/issues/3368

mischnic:
[...]

PR #3567, opened 2019-09-26
Factor asset filePaths when computing bundle hashes

wbinnssmith:
Currently it's possible for bundles to produce identical hashes if the output contents of their assets are identical. Factor in asset filePaths as well.

In the future we may be able to deduplicate bundles entirely based on their packaged and optimized content as well as source map content.

Is there anything else we should include in the bundle hash? #3545 also includes the bundle's environment.

Test Plan: Added an integration test. Verified it failed before applying the patch.


garthenweb:
We had a bug in parcel v1 where a dependency was updated in the main bundle which was also referenced in a child bundle, so the id changed in both bundles but the hash of the child bundle was not updated as the real content did not change. As we cached those JS files some users had a "module not found" error.

I am not sure if this is relevant for v2 (especially with tree shaking) and it also didn't happen always (didn't found out the exact cause so far). But maybe it is worth to add the IDs of the dependencies into the hash?

I can check on the weekend if this is still an issue with the latest v2 version (this commit) and create a bug report/ example project if needed!

garthenweb:
BTW: I think this PR might fix https://github.com/parcel-bundler/parcel/issues/3442 .

packages/core/core/src/BundleGraph.js
```diff
@@ -518,7 +518,7 @@ export default class BundleGraph {
     let hash = crypto.createHash('md5');
     // TODO: sort??
     this.traverseAssets(bundle, asset => {
-      hash.update(asset.outputHash);
+      hash.update([asset.outputHash, asset.filePath].join(':'));
```
DeMoorJasper: Won't this break sharing cache between machines as this is an absolute path?

Not sure if this is something we wanna support.

mischnic: > Not sure if this is something we wanna support.

Will was working on a HTTP remote cache backend, so I would say yes.

garthenweb:
I didn't manage to set up an example which bundles node modules with the v2 branch, but here is an example using parcel v1 which illustrates the issue mentioned above: https://github.com/garthenweb/parcel-bundle-names-example Maybe it is useful, let me know if there are questions :)

packages/core/core/src/BundleGraph.js
wbinnssmith: Yeah, we'll definitely need filepaths on the assets themselves to be relative to something like the project root.

PR #3570, opened 2019-09-27
Workerfarm shared references and first use case

padmaia:
# ↪️ Pull Request
This PR adds functionality to `@parcel/workers` that allows the main process to create shared references in all workers. This way we can serialize a data structure once and send it to all workers just once, and then we can just send a much smaller reference over IPC whenever we need to use it. The first use case for this is for packaging. Currently we are serializing and sending the entire bundle graph over IPC for every bundle. This becomes more of a burden the more bundles you have. In a stress test that contains 80 bundles, this improved the packaging phase speed by 50%. This may also be useful for things like Parcel options and targets which right now we are serializing and sending over IPC over and over again.

## 🚨 Test instructions

- [x] Unit tests
- [x] Manually run against stress test and confirm faster packaging phase


packages/core/core/src/PackagerRunner.js
```diff
@@ -66,16 +68,43 @@ export default class PackagerRunner {
         };
   }
 
-  async writeBundle(bundle: InternalBundle, bundleGraph: InternalBundleGraph) {
+  async writeBundles(bundleGraph: InternalBundleGraph) {
+    let farm = nullthrows(this.farm);
+    // TODO: maybe avoid doing this if we don't have to?
+    let {ref, dispose} = await farm.createSharedReference(bundleGraph);
```
padmaia: For cases where there are less bundles than the amount of workers, we are sending the bundle graph across IPC slightly more than we need to. Not sure if this is that much of a concern since in those cases the BundleGraph would reasonably small?

padmaia:
"Shared References" was the first thing that came to mind while brainstorming. Does this seem like a good naming choice?

packages/core/core/src/PackagerRunner.js
devongovett: could be one really big bundle or something... seems ok though since it will only get serialized once

mischnic:
```
packages/core/core/src/Parcel.js
  13:13  error  'InternalBundleGraph' is defined but never used  no-unused-vars
  14:13  error  'ParcelConfig' is defined but never used         no-unused-vars
```

Issue #3289, opened 2019-07-31
Changing output path and caching


mischnic:
# 🐛 bug report (Parcel 2)

Changing the output bundle path (e.g. `package.json#browser` in the react-hmr example when running `yarn parcel2 build src/index.html`) doesn't work without clearing the cache.

CC @padmaia

mischnic:
This was fixed by #3583

PR #3583, opened 2019-09-29
[T-93] Support building all packages in a monorepo, and invalidate entries and targets

devongovett:
This PR adds support for very easily building all packages in a monorepo. Just run `parcel build packages/*`, for example, and Parcel will resolve entries and targets for each package and build them all in parallel. Each package should define a `source` field in package.json, along with its targets. The following package.json will produce a CommonJS and an ES module version of the entry TypeScript file, for example:

```json
{
  "name": "my-package",
  "source": "src/index.ts",
  "main": "dist/index.js",
  "module": "dist/index.module.js"
}
```

This also moves entry and target resolution into the RequestGraph so that they can be invalidated by the watcher. Entries can now be globs, as well as point to directories containing a package.json with a `source` field. When a new file matches the glob, or the `source` field changes, the entry resolution will run again. After entries are resolved to files, the target resolver is run against each to produce one or more targets, which are also added to the graph. When package.json updates to change the targets, they are invalidated and re-resolved as well.


devongovett:
Had to do some ugly workarounds to the graph in this PR in order to get invalidations on package.json to work without crashing. 

Basically, the graph looks like this:
RequestGraph

When package.json changes, 4 nodes are invalidated: `entry_request`, `target_request`, `config_request`, and `asset_request`. Then, they all start processing in parallel. That's where things get tricky. As the requests are resolved, they mutate the graph. If the entry or target changed, it will replace the entire subgraph from the initial dependency downward. If a config or asset request resolves after that, it will have been deleted from the graph already.

Ideally, we would not perform any resolution on the config or asset request until we know that the entries haven't changed. Same for the asset request: we shouldn't invalidate that until the config request has been invalidated. package.json may have changed, but that doesn't mean the individual config extracted from there has. The user may have edited a completely different part. If so, we can avoid re-transforming. I think we should discuss these as future optimizations, and possibly come up with a different algorithm for invalidating and processing the graph...

mischnic:
Does this fix https://github.com/parcel-bundler/parcel/issues/3289?

devongovett:
Yes, I believe it should!

packages/core/core/src/EntryResolver.js
```diff
@@ -0,0 +1,104 @@
...
+    let stat;
+    try {
+      stat = await this.fs.stat(entry);
+    } catch (err) {
+      throw new Error(`Entry ${entry} does not exist`);
+    }
+
+    if (stat.isDirectory()) {
+      let pkg = await this.readPackage(entry);
+      if (pkg && typeof pkg.source === 'string') {
```
kwelch: Internally, we are currently declaring a source in the `pkg` as well but under a custom property. Do we expect to make this property standard, or should we allow configuration on the name of this key?

devongovett: `source` is already supported by Parcel 1 and is in the RFC for v2 as a "standardish" field: https://github.com/parcel-bundler/parcel/blob/master/PARCEL_2_RFC.md#sources

parcel-benchmark:
## Benchmark Results

#### Timings

| --- | --- | --- |
| Cold | 11.09s | -799.00ms 🚀 |
| Cached | 8.89s | +722.00ms ⚠️ |

#### Cold Bundles

| dist/legacy/parcel.ff215454.webp | 102.94kb | +0.00b | 216.00ms | +20.00ms ⚠️ |
| dist/modern/parcel.76ee4591.webp | 102.94kb | +0.00b | 215.00ms | +20.00ms ⚠️ |
| dist/modern/index.css | 29.00b | +0.00b | 1.30s | +105.00ms ⚠️ |

#### Cached Bundles

| dist/legacy/parcel.ff215454.webp | 102.94kb | +0.00b | 216.00ms | +20.00ms ⚠️ |
| dist/modern/parcel.76ee4591.webp | 102.94kb | +0.00b | 215.00ms | +20.00ms ⚠️ |
| dist/modern/index.css | 29.00b | +0.00b | 1.30s | +105.00ms ⚠️ |

</p></details>

**[Click here to view a detailed benchmark overview.](https://benchmark.parceljs.org/benchmark/245340990039130634)**

tommedema:
Does this mean that typescript project references are supported?

PR #3584, opened 2019-09-29
Add support for Workers with type: 'module'

### LINKED ISSUE #2619: Two argument version of `new Worker()` not recognized

mischnic:
Closes #2619
Closes #2732

This adds the primitive for module workers (and enables transforming them when a second argument is specified).
I don't think there there are any practical advantages to setting `type: module` (yet).

### DIFF
--- a/packages/transformers/js/src/visitors/dependencies.js
+++ b/packages/transformers/js/src/visitors/dependencies.js
@@ -1,12 +1,17 @@
+// @flow
+
+import type {MutableAsset} from '@parcel/types';
+
 import * as types from '@babel/types';
 import traverse from '@babel/traverse';
 import {isURL} from '@parcel/utils';
 import nodeBuiltins from 'node-libs-browser';
 import {hasBinding} from './utils';
+import invariant from 'assert';
 
 const serviceWorkerPattern = ['navigator', 'serviceWorker', 'register'];
 
-export default {
+export default ({
   ImportDeclaration(node, asset) {
     asset.meta.isES6Module = true;
     addDependency(asset, node.source);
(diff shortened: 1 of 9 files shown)


packages/transformers/js/src/visitors/dependencies.js
```diff
@@ -79,21 +87,44 @@ export default {
...
+          types.isIdentifier(v.key, {name: 'type'})
+        );
+        if (prop && types.isStringLiteral(prop.value))
+          isModule = prop.value.value === 'module';
+      }
+
+      addURLDependency(asset, args[0], {
+        env: {
+          context: 'web-worker',
+          outputFormat: isModule ? 'esmodule' : undefined
```
devongovett: Probably should ensure that scope-hoisting is enabled here. The packager will throw if you try to output es modules without scope hoisting enabled. The HTML transformer checks:

https://github.com/parcel-bundler/parcel/blob/14043b250a6c63cc0a80960eb0b97821c1539bf5/packages/transformers/html/src/dependencies.js#L76

mischnic: Done

Issue #3586, opened 2019-09-30
Dynamically importing CommonJS produces empty bundle


mischnic:
Dynamically importing a commonjs module from a `esmodule` asset produces an empty bundle.

```html
<script type="module" src="./index.js"></script>
```
```js
// index.js
import("./commonjs.js").then(v => console.log(v));
```
```js
// commonjs.js
module.exports = 2;
```

Output (why is that even called `...exports`?):
```js
var $c0ee4c22c4f80f14efb8476eb9807b50$exports = {};
$c0ee4c22c4f80f14efb8476eb9807b50$exports = import('' + './commonjs.e4b0f555.js');
$c0ee4c22c4f80f14efb8476eb9807b50$exports.then(v => console.log(v));
```

`./commonjs.e4b0f555.js` is empty, without `treeShake`, the output is
```js
// ...
var $dfc1ea661454a5b57fd7e754dd151b32$exports = {};
$dfc1ea661454a5b57fd7e754dd151b32$exports = 2;
```

We check `b.env.outputFormat === 'esmodule'`, but this value is just inherited from that parent asset.

The root cause is of course that we don't/can't translate commonjs to esmodules, but this is a case where this might be unexpected.

cc @devongovett

mischnic:
I guess we should for now throw an error if the entry asset of the dynamic bundle has CommonJS exports?

wangkewei1993:
hi

devongovett:
`import()` returns the entire module namespace. Should we just map the common JS `module.exports` object to that namespace if we detect it is CommonJS?

mischnic:
> Should we just map the common JS module.exports object to that namespace if we detect it is CommonJS?

```js
const $module$exports = {
	function : () => {},
	constant: 2
};

export $module$exports; // !!
```

Exporting a JS object as a namespace isn't possible, either `export {prop: val, ...}` (but we don't know the properties at built time, that is the whole point of this) or `export default $module$exports`.

mischnic:
Fixed by #3800

PR #3587, opened 2019-09-30
[T-94] Always generate esmodule bundles for dynamic import when possible

mischnic:
Always generate esmodule bundle for dynamic import when possible.

Merging this would make #3586 even more apparent, because it would also happen without `<script type=module>`. That should be figured out beforehand.


mischnic:
I don't know how to test this cleanly because Node doesn't support dynamic imports

mischnic:
Depending on the implementation, this would either
- make builds slower or
- make bundles bigger.

PR #3599, opened 2019-10-03
Parcel 2: Add Benchmark Action

DeMoorJasper:
# ↪️ Pull Request

This PR adds a workflow file to trigger the PR Benchmark action, code for this action can be found here: https://github.com/DeMoorJasper/parcel-benchmark-action

The action is still lacking some features and timings aren't really reliable on the Github action infrastructure. However updates to the action will not affect the workflow file in any way.

The action uses the metrics reporter, so please merge that one first (#3596) otherwise the action will always fail...

## 💻 Examples

Screenshot 2019-10-03 at 13 15 08

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -10,4 +10,4 @@ jobs:
       - name: PR Benchmarks
         uses: DeMoorJasper/parcel-benchmark-action@master
         env:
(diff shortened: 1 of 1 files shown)


Daeshanae:
ok

PR #3601, opened 2019-10-03
infra: preparation for building Parcel with Parcel 

mischnic:
~**WIP: Waiting for flow integration: https://github.com/facebook/flow/pull/8128**~

With the eventual move from gulp to using Parcel to build Parcel itself (which we could start experimenting with after #3583 is merged, we can also delay this PR until then), I changed all package.json files to use the correct pattern for that, e.g.:
```json5
{
  // ...
  "main": "lib/BabelTransformer.js",
  "source": "src/BabelTransformer.js",
}
```

To still be able to use babel-register, a babel plugin was added that replaces `require`s and `import`s with the resolved path while using `package.json#source` instead of `package.json#main`.

Flow is also configured to resolve imports this way (`flow-resolver.js`). **In my testing this didn't have any negative effect on flow's performace.** (I'm saying this because that file is actually a kind of server that flow sends dependency requests to via stdin).

Parcel's own `resolve` utilites are patched in the same way when `PARCEL_BUILD_ENV` is set to development (`@parcel/babel-register` is conditionally required in the same way).

The only module that didn't work well so well was `@parcel/babel-preset-env` because it is resolved by babel itself, I therefore changed that single JS file to use CommonJS instead of ESM. An alternative solutions to this would be passing `require.resolve('@parcel/babel-preset-env/src/index.js)')` to babel in `packages/transformers/babel/src/env.js` conditionally on `PARCEL_BUILD_ENV`.


parcel-benchmark:
## Benchmark Results
**[Click here to view a detailed benchmark overview.](https://benchmark.parceljs.org/benchmark/245324051098108426)**

mischnic:
NOOOOO 😞 

flow commit from 2 days ago:
> Delete module.resolver support
> We experimented with this feature but it's dead internally (and hopefully externally).
> https://github.com/facebook/flow/commit/87820b8e21e626126e2a8ab7e3a70f253a24bafd

devongovett:
Haha oh no. 😢 

One way is to make an `index.js.flow` in the root of every package that re-exports `src`. Flow will use that over package.json#main.

Otherwise, we could try contributing `source` field support to flow. I think @wbinnssmith knows people on the team?

mischnic:
> One way is to make an index.js.flow in the root of every package that re-exports src. Flow will use that over package.json#main.

I thought I had tried that already and it didn't seem to work. Will try that again.

(Yarn PnP has a bigger problem here, they used the same custom resolver technique (that feature was added by arcanis). Not sure what they'll do)

> Otherwise, we could try contributing source field support to flow. I think @wbinnssmith knows people on the team?

Heh, @jamiebuilds has a feature request exactly for this `main`/`source` case with a `module.system.node.main_field` config option 😄 
https://github.com/facebook/flow/issues/5725#issuecomment-361325590

mischnic:
Modules that currently fail to build with Parcel:
```
contains fs.readFileSync (not bundled):
    reporters/dev-server/src/templates/*
    runtimes/hmr
    packagers/js

contains require.resolve (link.js throws):
    resolvers/default
    runtimes/js
    core/core
    core/parcel
    core/register
    core/workers


aborts silently?:
    core/fs

`Invalid left-hand side in assignment expression`?:
    core/parcel

```

mischnic:
> I think @wbinnssmith knows people on the team?

Here's the PR: https://github.com/facebook/flow/pull/8128, they don't seem to be very interested in merging PRs in general (even older ones)...

PR #3610, opened 2019-10-06
CONTRIBUTING.md: add a snippet re: how to run the dev version

kirillrogovoy:
# ↪️ Pull Request

Hey. Just a quick documentation addon, because the day before, it could save me an hour of searching and asking stupid questions on Spectrum chats. 😅 

## ✔️ PR Todo


mischnic:
Actually, there's already a paragraph about this further down in that file:

> - Another way, which we highly recommend if you plan to switch between different parcel sources often, is to create some terminal aliases. This way you can have different aliases for each of the versions of parcel you wish to maintain on your machine.......

kirillrogovoy:
Do you think they are mutually exclusive?

I've read the file back and forth a couple of times and it didn't help me understand how to simply run the development version with no aliases or any other attachments.

Besides, that paragraph uses `alias parcelfork="node ~/my/path/to/fork/bin/cli.js"` as the example, whereas the correct path would rather be `~/my/path/to/fork/packages/core/parcel/src/bin.js` in general case.

That `packages/core/parcel` part isn't mentioned anywhere, and even if it was, I believe the path to the dev executable should be as much on the surface as possible because that's what you would want most of the time when starting contributing.

mischnic:
You're right, this makes it easier to understand either way.

kirillrogovoy:
So, anything you'd like me to change?

CONTRIBUTING.md
```diff
@@ -31,22 +31,28 @@ To get started with the project:
...
 
-```js
+```sh
 yarn test
 ```
 
+Run the development version of Parcel:
+
+```sh
+./packages/core/parcel/src/bin.js -V
```
mischnic:
```suggestion
./packages/core/parcel/src/bin.js build index.html
```

Maybe this makes it more obvious that this is the `parcel` binary?

devongovett:
You can also use `yarn link` for this:

```sh
cd packages/core/parcel
yarn link
```

Now you can access the `parcel` command globally from anywhere.

Should we add that too?

kirillrogovoy:
I don't mind, but then we should specify how to reverse that linkage so that the user can use the global Parcel again if they want to.

What's the correct way to reverse it? I guess, simply running `yarn unlink` should suffice?

devongovett:
Ah actually that's already here https://github.com/parcel-bundler/parcel/blob/50dcbff9d5d1fe6965e4816728e5d64990569b40/CONTRIBUTING.md#link-to-local-parcel-build.

kirillrogovoy:
Then, I suppose, it's good to go?

kirillrogovoy:
Up

kirillrogovoy:
Up @mischnic

lbogdan:
Why not just run `yarn parcel build index.html` from the parcel workdir, as `parcel` is installed as a `node_modules` executable?

https://github.com/parcel-bundler/parcel/blob/ddb253e6ce00a1709131519c349f892322b53526/packages/core/parcel/package.json#L12

```sh
~/work/parcel$ ls -al node_modules/.bin/parcel
lrwxrwxrwx 1 node node 37 Jan  4 16:26 node_modules/.bin/parcel -> ../../packages/core/parcel/lib/bin.js
```

Issue #3477, opened 2019-08-29
[RFC] Import query params


padmaia:
## Problem Statement

Currently in Parcel v2, dependencies are resolved to files and the resulting file paths are matched against the configured globs in the `.parcelrc` file to see which pipeline the file should be transformed by. This works for most cases, but sometimes you might want to force a dependency down a specific pipeline. For instance, let's say you are building documentation for a React component. By default, importing the React component will bundle the component as a JavaScript module, but what if you additionaly want the component's source as plain text? Maybe you also want to run some transformer over the component that extracts what prop types it expects and bundle that as well.

A very naive solution that would work for any and all tooling would be to just create files for the source text and prop types in a prebuild step and then you can import three individual files. The downside to this approach is that you would then have two build steps, which means you have two build commands that you have to string together and two separate watchers if you want to build in watch mode. Imagine in this scenario you updated `MyComponent.js`. Both the prebuild and the bundle watchers would pick up the change and start rebuilding. Then when the prebuild step finishes, the derived files are updated, which causes the bundler to rebuild yet again. It makes for a much better developer experience if this can all be handled in one process.

With webpack, this is usually solved by using [inline loaders]([https://webpack.js.org/concepts/loaders/#inline](https://webpack.js.org/concepts/loaders/#inline)).
```js
import MyComponent from './MyComponent';
import MyComponentSrc from '!!raw-loader!./MyComponent';
import MyComponentPropTypes from `!!extract-react-types-loader!./MyComponent`;
```

One downside to this approach is that the detail that you are using webpack leaks into your source code. Ideally this could be done in a more bundler agnostic way.

## Proposed Solution: Import Query Params

The ECMAScript modules standard (ESM for short) is now supported across all modern browsers and will soon be supported by Node.js as well. One neat thing about ESM is that it essentially treats all module specifiers as URLs, which means adding query parameters is perfectly valid syntax and not something bespoke to a specific bundler. This seems like a great way to specify how you would like a dependency loaded:
```js
import MyComponent from './MyComponent.js';
import MyComponentSrc from './MyComponent.js?as=raw-text';
import MyComponentPropTypes from './MyComponent.js?as=react-types';
```

So what would it take for Parcel to support this? The most straightforward approach requires no changes to Parcel's configuration so it's got that going for it. All you would have to do to force a file to be transformed by a different pipeline is add a query param to the module specifier and configure a pipeline in your `.parcelrc` with a glob that matches it.
```json
{
  "transforms": {
[...]

PR #3613, opened 2019-10-07
Add support for generating TypeScript declarations, and named pipelines

devongovett:
This adds support for generating TypeScript declaration files as a library target, along with the beginnings of named pipeline support. This allows the same input file to be sent through multiple different pipelines to produce different outputs (a `.js` file and a `.d.ts` file in this case). A `pipeline` parameter has been added to dependencies, asset requests, and config requests to support this in core. Related: #3477.

Eventually, resolvers should be able to return pipeline names parsed from import specifiers, so you could reference pipelines from code. For now, the pipeline is set on entries based on the target name (e.g. `types` from package.json). So, you can have a package.json like this:

```json
{
  "name": "my-package",
  "main": "dist/index.js",
  "types": "dist/index.d.ts"
}
```

And a `.parcelrc` like this (actually, this is in the default config):

```json
{
  "transforms": {
    "types:*.{ts,tsx}": "@parcel/transformer-typescript-types",
  }
}
```

Parcel will resolve the targets in package.json, and apply the appropriate pipeline based on the target name. If no named pipeline is found for that target (e.g. no `main` pipeline), then it will fall back to the non-named pipeline that matches.

In terms of actually generating TS declarations, we use the TypeScript compiler and a custom TypeScript transformer to do that. TypeScript normally generates separate `.d.ts` file for each file that you import in your code, but if you specify the `outFile` option it will generate a "bundle" file instead with all modules inlined. Unfortunately, this still includes all types, not only ones that are exported by the package. 

So, we have a custom TS transformer which uses the AST to tree shake and flatten these modules into one top-level `.d.ts` file containing only exported types. It works in three passes, similar to the JS tree shaking implementation: first collecting information about the imported, exported, and local declarations in each module, then propagating exports from the entry module through to the children and marking used nodes, and finally removing unused nodes and flattening into the top-level scope, renaming declarations as needed.


parcel-benchmark:
## Benchmark Results

#### Timings

| --- | --- | --- |
| Cold | 11.20s | -473.00ms |
| Cached | 8.98s | -89.00ms |

#### Cold Bundles

| dist/modern/index.js | 794.00b | +0.00b | 541.00ms | -28.00ms 🚀 |
| dist/modern/index.css | 29.00b | +0.00b | 1.40s | +74.00ms ⚠️ |

#### Cached Bundles

| dist/legacy/index.js | 252.00b | +0.00b | 16.00ms | -2.00ms 🚀 |
| dist/modern/index.css | 29.00b | +0.00b | 11.00ms | -1.00ms 🚀 |

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 15.99s | +188.00ms |

#### Cold Bundles

*No bundle changes detected.*

#### Cached Bundles

</p></details>

**[Click here to view a detailed benchmark overview.](https://benchmark.parceljs.org/benchmark/245584963556082186)**

packages/shared/scope-hoisting/src/link.js
```diff
@@ -545,11 +545,10 @@ export function link({
...
 
-        let paths = path.unshiftContainer('body', imports);
-        for (let path of paths) {
-          if (path.isDeclaration()) {
-            path.scope.registerDeclaration(path);
-          }
+        if (imports.length > 0) {
+          // Add import statements and update scope to collect references
+          path.unshiftContainer('body', imports);
+          path.scope.crawl();
```
mischnic: Can you a test case for that? (So it doesn't break again if we come back to optimize this)

devongovett: done!

packages/core/core/src/Dependency.js
```diff
@@ -31,7 +32,8 @@ export function createDependency(opts: DependencyOpts): Dependency {
     md5FromString(
```
wbinnssmith: there are a lot of things in here now. what about using `md5FromObject`? something like

```
md5FromObject({
  sourceAssetId: opts.sourceAssetId,
  moduleSpecifier: opts.moduleSpecifier,
  env: opts.env,
  target: opts.target,
  pipeline: opts.pipeline
})

File: packages/packagers/ts/src/TSPackager.js [hunk 3]
```diff
@@ -0,0 +1,23 @@
+// @flow strict-local
```
wbinnssmith: Is this only for packaging type definitions? Maybe the name of the file could include "types"?

packages/configs/default/index.json
```diff
@@ -1,6 +1,7 @@
 {
   "bundler": "@parcel/bundler-default",
   "transforms": {
+    "types:*.{ts,tsx}": ["@parcel/transformer-typescript-types"],
```
jamiebuilds: Is this supposed to be the syntax for named pipelines?

devongovett: Yes.

PR #3615, opened 2019-10-07
RFC: Support cache between versions

### LINKED ISSUE #3598: Parcel 2: versions and caches
# ❔ Question

When a  new Parcel 2 version will be released(when it'll be stable), since it's a possibility that the graph change, would it automatically invalidates the outdated cache or should we always delete the cache ourselves when upgrading the parcel version, if so, could you make it automatic instead ?

wbinnssmith:
Resolves #3598. Thanks @banou26 for mentioning this.

**NOTE**: Right now, structures like AssetGraph, RequestGraph, and BundleGraph depend on many structures that could change subtly. I'm thinking of an alternate approach in which we always factor in e.g. up to the minor version of core and not version independent structures. This seems too error prone to do perfectly and manually, and maybe we could just reserve all breaking changes to any structure for semver minor versions.

This supports caching structures between versions of Parcel by:

* Allowing individual classes to include versions, beyond versioning in the `package.json`
* Including these versions in the cache keys of these structures

Without these changes, upgrading Parcel with a new version in `@parcel/core`'s package.json would cause Parcel to continue to look in the same location for a cache entry, only for it to fail to deserialize as the versions conflict. Instead, factor important versions into the cache keys.

Test Plan:
* Run the react-hmr example. Increment the version in `@parcel/core`'s package.json and re-run. Verify a deserializer error.
* Apply patches. Re-run the example, increment again, and run once more. Verify no deserializer errors.
* Increment `VERSION` on AssetGraph and re-run. Verify building takes place between runs.

### DIFF
--- a/packages/core/core/src/AssetGraph.js
+++ b/packages/core/core/src/AssetGraph.js
@@ -69,6 +69,8 @@ const nodeFromEntryFile = (entry: string) => ({
 });
 
 export default class AssetGraph extends Graph<AssetGraphNode> {
+  static VERSION = Graph.VERSION + ':1';
+
   onNodeAdded: ?(node: AssetGraphNode) => mixed;
   onNodeRemoved: ?(node: AssetGraphNode) => mixed;
   hash: ?string;
--- a/packages/core/core/src/AssetGraphBuilder.js
+++ b/packages/core/core/src/AssetGraphBuilder.js
@@ -183,8 +183,12 @@ export default class AssetGraphBuilder extends EventEmitter {
   }
 
   getCacheKeys() {
-    let assetGraphKey = md5FromString(`${this.cacheKey}:assetGraph`);
-    let requestGraphKey = md5FromString(`${this.cacheKey}:requestGraph`);
+    let assetGraphKey = md5FromString(
+      `${this.cacheKey}:assetGraph:${AssetGraph.VERSION}`
+    );
+    let requestGraphKey = md5FromString(
+      `${this.cacheKey}:requestGraph:${RequestGraph.VERSION}`
+    );
     let snapshotKey = md5FromString(`${this.cacheKey}:snapshot`);
     return {assetGraphKey, requestGraphKey, snapshotKey};
   }
--- a/packages/core/core/src/BundleGraph.js
+++ b/packages/core/core/src/BundleGraph.js
@@ -41,6 +41,8 @@ type BundleGraphEdgeTypes =
   | 'references';
 
 export default class BundleGraph {
+  static VERSION = Graph.VERSION + ':1';
+
   // TODO: These hashes are being invalidated in mutative methods, but this._graph is not a private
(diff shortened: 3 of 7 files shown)


parcel-benchmark:
## Benchmark Results

#### Timings

| --- | --- | --- |
| Cold | 11.79s | +213.00ms |
| Cached | 9.33s | +115.00ms |

#### Cold Bundles

| dist/legacy/parcel.ff215454.webp | 102.94kb | +0.00b | 235.00ms | -23.00ms 🚀 |
| dist/modern/parcel.76ee4591.webp | 102.94kb | +0.00b | 234.00ms | -27.00ms 🚀 |
| dist/modern/index.js | 794.00b | +0.00b | 571.00ms | -32.00ms 🚀 |
| dist/legacy/index.js | 252.00b | +0.00b | 573.00ms | -31.00ms 🚀 |

#### Cached Bundles

| dist/legacy/parcel.ff215454.webp | 102.94kb | +0.00b | 14.00ms | -1.00ms 🚀 |

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 16.08s | -495.00ms |

#### Cold Bundles

*No bundle changes detected.*

#### Cached Bundles

</p></details>

**[Click here to view a detailed benchmark overview.](https://benchmark.parceljs.org/benchmark/245612758528164362)**

packages/core/core/src/AssetGraph.js
```diff
@@ -69,6 +69,8 @@ const nodeFromEntryFile = (entry: string) => ({
 });
 
 export default class AssetGraph extends Graph<AssetGraphNode> {
+  static VERSION = Graph.VERSION + ':1';
```
padmaia: Yeah, manually updating this definitely seems error prone. I'd be more comfortable if we moved these structures to their own package(s) so that things like `lerna version` could automatically bump versions when things change. Seems like right now, it would be okay to just have one package for these structures.

devongovett: For now, let's just use the core version from package.json. Seems easiest to me. then we can determine a more granular way to share caches between parcel versions where possible.

jamiebuilds:
I would not be bothered by the entire cache being invalidated when switching versions of Parcel, even if they are minor or patch versions

padmaia:
Yeah, after giving this some thought I find myself agreeing with @jamiebuilds. While I don’t like the idea of evicting the entire cache, this does seem like the one scenario that probably warrants it. I don’t really like the idea of only factoring in major and minor versions because then that still allows some human error that we made some bug fix change that would affect the cache, but didn’t think to bump the minor version. Originally I thought that it might work to have a package that could be independently versioned from core in order to evict the cache less often, but the more I thought what would need to be included in that package, it started looking more and more like most of core. And then you still have the human error factor in not including the right things in that package. Evicting the entire cache doesn’t feel great, but if we have features like a shared multilayer cache then it’s less of a big deal. So yeah, I’m on board with just using `@parcel/core` version in cache keys, including the patch version.

Issue #3591, opened 2019-10-01
Flag to delete output folder before building


mkg20001:
# 🙋 Feature request

A flag/option to delete the output folder before building.

It's common place to do that in order to avoid mixing dev asssets and production assets.

## 🤔 Expected Behavior

Simply remove the output destination recursively before building via a flag
Should be enabled by default.

## 😯 Current Behavior

It has to be scripted manually.

## 💁 Possible Solution

Per-target flag/option `deleteBeforeBuild` which is enabled by default in production build mode. Simply removes the output folder.

## 🔦 Context

Most people add `rm -rf dist;` before the parcel command

That's bad for windows users since rm doesn't work that way there

Having it in parcel would make things a lot easier and save the hassle of manually adding the recursive rm command

## 💻 Examples

```console
$ npx parcel build index.html -o dist --delete-before-build
```

```json
{
  "target": {
    "browser": {
      "main": "index.html",
      "deleteBeforeBuild": true
    }
  }
}
```

fagnervalente:
Hi @mkg20001 
After read a bunch of articles about responsibilities of this kind of tools, I advise you to use Gulp + Parcel. Parcel is a bundles as a webpack do, so what you are waiting for is a task better executed by a task manager as a Gulp. In this case, Parcel will run as one of your task list. It is very easy to config and be happy. I wish helped you. Best regards.

Userfull links
- https://www.npmjs.com/package/gulp-parcel
- https://gulpjs.com

mischnic:
I would say using gulp is overkill here, running `rm -rf build && parcel build ...` works just fine.

But I still think that this should be somehow integrated into Parcel.

kirillrogovoy:
I'm sorry if it comes around as a stupid question, but is there really a reason to **not** remove the output directory before building?

What would be the use case for that?

I feel like removing that directory beforehand is what you would want by default 99% of the time, so just making it the default behavior with no flags involved might make sense, and that other 1% of cases would be handled with manual scripting.

mkg20001:
@kirillrogovoy While I tend to agree with that, it's still something _someone_ might want to disable. Although we could also add it and then wait for somebody to request that (disabling auto-deletion) as a feature

kirillrogovoy:
I'm sure there'll be cases where auto-deletion is undesirable, but I'd be curious to study them. 

Ultimately, just as for manual deletion, preservation might be done via scripting around Parcel for that 1 case out of 100. :) 

[...]

PR #3616, opened 2019-10-08
Purge dist dirs before bundling

kirillrogovoy:
# ↪️ Pull Request

Closes https://github.com/parcel-bundler/parcel/issues/3591.

## 🚨 Test instructions

```bash
parcel build index.html
touch dist/extra

parcel build index.html

ls dist # no ./extra is there
```

## ✔️ PR Todo


kirillrogovoy:
One caveat: the `should support transforming with postcss twice with the same result` test is broken, because it builds two bundles concurrently into the same dist directory.

So it runs into a race condition where the second bundle can't be written because the first bundle has removed the dist directory.

I think that's exactly that 1-out-of-100 case where removing the directory by default doesn't work. 

My intuitive thinking is that Parcel shouldn't probably support concurrent building into the same directory by default because it can get unpredictable in other ways too.

mischnic:
As said in https://github.com/parcel-bundler/parcel/issues/3591, there  definitely needs to be a flag to enable/disabled this (not sure what the default should be).

> But something to consider too is that there might be big performance gains to be had using caching and not cleaning by default

I think Parcel will always (over)write these files, reading them and comparing against Parcel's output would be slower than just writing the to disk.

DeMoorJasper:
I don't think this should be the default, probably best to make it an opt-in using a flag as mischnic mentioned

DeMoorJasper:
Closing this due to inactivity

Issue #3337, opened 2019-08-09
RFC: Rich, actionable errors in Parcel 2


gorakong:
# 💬 RFC

## 🔦 Context

Currently Parcel 2 has a few shortcomings in its error reporting:

* Babel errors are repeated, include a parser stacktrace, and aren't colored correctly. They also aren't clickable to the exact row and column the error occurred.
* Resolver errors don't include a code frame like they did in Parcel 1.
* We have the context and information to provide resolver errors like Rome does in https://twitter.com/sebmck/status/1150878549441572864

## 💻 Examples

Rome's resolver errors:

![](https://user-images.githubusercontent.com/15333808/62805094-d48da900-baa3-11e9-8f31-017a534fb027.png)

[VSCode's Diagnostic structure](https://code.visualstudio.com/api/references/vscode-api#Diagnostic)

## Proposal

Let's create a single standard error type that can describe any type of code-related error that can occur in Parcel so it can be shown in a rich, actionable way:

```js
type Diagnostic = {|
  message: string | Array<
    | string
    | {|type: 'filePath', value: FilePath|}
    | {|type: 'message', value: string, style?: 'emphasis'|}
  >
  severity: 'error' | 'warning' | 'info',
  codeFrame?: CodeFrame
|}

class ParcelError extends Error {
  source: string, // Name of the plugin or component that caused the error
  diagnostics: Array<Diagnostic>

  constructor({diagnostics: Array<Diagnostic>, source: string});
}
```

Here's an example:

```js
throw new ParcelError({
  source: 'resolver',
  diagnostics: [{
    severity: 'error',
    message: [
      'failed to resolve', 
      {type: message, value: "'./foo'", style: 'emphasis'}, 
      'from', 
      {type: 'filePath', value: '/Users/gora/parcel/bar.js'},
    ],
    codeFrame: // Code frame showing the failing import
  }]
});
```

This way, the appropriate UI interface can display this error as it would like. Since reporters receive these, they can be mapped to terminal uis, web pages, IDEs, etc.

Alternatives: Instead of providing a babel-style code frame (or maybe optionally instead of providing a code frame), we could accept a filePath and row/column tuple and create the codeframe at the view layer.

Once we decide on a direction, I'm happy to take on an implementation.

cc @devongovett @wbinnssmith @padmaia @mischnic @DeMoorJasper @sebmck @jamiebuilds

jamiebuilds:
[...]

PR #3648, opened 2019-10-16
[T-16] Diagnostics

DeMoorJasper:
# ↪️ Pull Request

This PR implements a generic Diagnostic type that allows Parcel to do more streamlined and detailed error reporting. The diagnostic object adds the ability to create more advanced codeframes, hints and metadata without implementing any render behaviour which allows complete freedom for any environment to setup their own rendering behaviour whether that would be ansi, html or even highlighting the errors straight in your IDE

This is just a starting point towards making Parcel even more user friendly.

Future work includes but is not limited to adding more diagnostics in existing code with/without hints, catching and properly converting existing errors into diagnostics and ensuring all errors are being logged properly and the enduser always receives all the information we can provide.

- [x] Create type for diagnostic
- [x] Move ts validator to diagnostics
- [x] Move all old logs to diagnostics
- [x] Create a ThrowableDiagnostic class so we can throw diagnostics instead of always calling `@parcel/logger` (these will always be level error?)
- [x] Properly convert Error Objects into a diagnostic for better debugging
- [x] Fix all Flow issues
- [x] Implement custom codeframe pkg for rendering multiple highlights
- [x] Markdown => ansi
- [x] Make validators return an array of diagnostics
- [x] Add an easy util that replaces prettyError to format diagnostics using Parcel's default ansi markdown and code-frame utils

Related #3337

## 💻 Examples

Examples of rendered diagnostics

Screenshot 2019-10-28 at 13 53 58
Screenshot 2019-10-28 at 13 54 15


parcel-benchmark:
## Benchmark Results

#### Timings

| --- | --- | --- |
| Cold | 10.41s | -106.00ms |
| Cached | 8.47s | +10.00ms |

#### Cold Bundles

| dist/legacy/parcel.ff215454.webp | 102.94kb | +0.00b | 205.00ms | -11.00ms 🚀 |
| dist/modern/parcel.76ee4591.webp | 102.94kb | +0.00b | 196.00ms | -19.00ms 🚀 |
| dist/modern/index.css | 29.00b | +0.00b | 1.31s | +68.00ms ⚠️ |

#### Cached Bundles

| dist/modern/parcel.76ee4591.webp | 102.94kb | +0.00b | 12.00ms | +1.00ms ⚠️ |
| dist/modern/index.js | 794.00b | +0.00b | 12.00ms | -1.00ms 🚀 |
| dist/legacy/index.css | 29.00b | +0.00b | 12.00ms | +1.00ms ⚠️ |
| dist/modern/index.css | 29.00b | +0.00b | 10.00ms | -1.00ms 🚀 |

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 15.81s | +1.11s ⚠️ |

#### Cold Bundles

| dist/main/index.js | 34.53kb | +0.00b | 2.37s | +121.00ms ⚠️ |
| dist/module/index.js | 34.33kb | +0.00b | 2.36s | +125.00ms ⚠️ |

#### Cached Bundles

</p></details>

**[Click here to view a detailed benchmark overview.](https://benchmark.parceljs.org/benchmark/246374640278045194)**

packages/core/cache/src/Cache.js
```diff
@@ -75,7 +75,10 @@ export default class Cache {
       await this.fs.writeFile(blobPath, data);
       return key;
     } catch (err) {
-      logger.error(`Error writing to cache: ${err.message}`);
+      logger.error({
+        origin: '@parcel/cache',
+        message: `Error writing to cache: ${err.message}`
```
wbinnssmith: could we make a factory function that returns loggers with pre-bound names, kind of like Python's stdlib `logging` module?

https://docs.python.org/3.7/library/logging.html#logging.getLogger

DeMoorJasper: Sure, although I'm not really sure how you'd figure out the origin for it. Or if we even need an origin at all

packages/core/logger/src/Logger.js
```diff
@@ -74,29 +75,46 @@ export function patchConsole() {
...
   consolePatched = true;
 }
 
+function messagesToDiagnostic(messages: Array<mixed>): Diagnostic {
+  if (messages.length === 1 && messages[0] instanceof Error) {
+    let error: Error = messages[0];
+
+    return {
+      message: error.message,
+      origin: '@parcel/logger',
```
DeMoorJasper: should the default origin be `@parcel/logger`. Might be confusing as the error did not originate from the logger.

Perhaps it should be `unknown` or just an empty string?

packages/core/diagnostic/src/diagnostic.js
```diff
@@ -0,0 +1,99 @@
...
+  loc?: {
+    column: number,
+    line: number,
+    ...
+  },
+  source?: string,
+  ...
+};
+
+export function errorToDiagnostic(error: PrintableError): Diagnostic {
```
DeMoorJasper: Should this fetch source from FileSystem if `error.fileName` isn't undefined and `error.source` is undefined.

packages/core/diagnostic/src/diagnostic.js
```diff
@@ -0,0 +1,99 @@
...
+        end: {
+          line: error.loc.line,
+          column: error.loc.column
+        }
+      }
+    };
+  }
+
+  return {
+    origin: 'Error',
```
DeMoorJasper: Should this go up the tree if filename is defined to find the nearest pkg.json and use the `name` of that package as origin?

packages/core/logger/src/Logger.js
DeMoorJasper: Changed it to `console` and `Error`

packages/core/core/src/Validation.js
```diff
@@ -71,7 +71,11 @@ export default class Validation {
     };
 
     let config = await this.loadConfig(configRequest);
-    let parcelConfig: ParcelConfig = nullthrows(config.result);
+    nullthrows(config.result);
```
DeMoorJasper: This was actually a regression bug by the packageManager...

packages/reporters/cli/src/formatCodeFrame.js
```diff
@@ -0,0 +1,31 @@
...
+    return 'Could not create codeframe, no highlights defined.';
+  }
+
+  // TODO: Support multiple code highlights
+  let highlight = Array.isArray(codeframe.codeHighlights)
+    ? codeframe.codeHighlights[0]
+    : codeframe.codeHighlights;
+
+  let result = codeFrameColumns(
+    codeframe.code,
```
devongovett: what would happen if this were not JS - like CSS or something?

devongovett:
One idea would be to pass the origin to the logger when we're creating it. Instead of a singleton, the logger would be passed as an option to all plugins, and before calling that plugin we'd construct a logger with the origin already set. We always know what npm module the plugin came from, so this would avoid the plugins themselves from needing to set an origin.

packages/reporters/cli/src/formatCodeFrame.js
DeMoorJasper: than codeframe.code should be css. In theory a codeframe isn't language specific.

The syntax highlighting of babel/code-frame is JS specific though

DeMoorJasper:
@devongovett sure I can pass some kind of logger instance to the plugins if you think that's a good idea

packages/core/core/src/Validation.js
wbinnssmith: Flow should be able to better understand `invariant(config.result != null)` as it doesn't know that `nullthrows` throws, only that it returns something non-null.

packages/core/cache/src/Cache.js
wbinnssmith: The origin and the name of the logger would be the same, yeah? something like `createLogger(origin: string): Logger`

DeMoorJasper: Actually after looking through the code pluginOptions is initialised before any plugin is even loaded. So that one can't be used, so it would have to be another param or we'd have to create a pluginOptions instance for every plugin.

Not sure if this is something we should be doing

packages/core/diagnostic/src/diagnostic.js
```diff
@@ -0,0 +1,125 @@
...
+|};
+
+// A Diagnostic is a style agnostic way of emitting errors, warnings and info
+// The reporter's are responsible for rendering the message, codeframes, hints, ...
+export type Diagnostic = {|
+  message: string,
+  origin: string, // Name of plugin or file that threw this error
+
+  // Asset metadata
+  filename?: string,
```
devongovett: nit: `filePath: FilePath`

packages/core/diagnostic/src/diagnostic.js
```diff
@@ -0,0 +1,125 @@
...
+  language?: string,
+
+  // Codeframe data
+  codeframe?: DiagnosticCodeFrame,
```
devongovett: nit: `codeFrame`

packages/core/diagnostic/src/diagnostic.js
```diff
@@ -0,0 +1,125 @@
...
+  // Stacktrace for error, not really needed if there's a codeframe...
+  stack?: string,
+
+  // Hints to resolve issues faster
+  hints?: Array<string>
+|};
+
+// This type should represent all error formats Parcel can encounter...
+export type PrintableError = Error & {
+  fileName?: string,
```
devongovett: nit: `filePath: FilePath`

packages/core/diagnostic/src/diagnostic.js
```diff
@@ -0,0 +1,99 @@
...
+    ...
+  },
+  source?: string,
+  ...
+};
+
+export function errorToDiagnostic(error: PrintableError): Diagnostic {
+  let codeframe: DiagnosticCodeFrame | void = undefined;
+
+  if (error.loc && error.source) {
```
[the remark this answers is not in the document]
devongovett: I suppose the reporter could do that itself?

packages/core/diagnostic/src/diagnostic.js
devongovett: Is this meant to be the parcel plugin that created the error? If we put a try..catch around all the places we call into plugins in core and wrapped their errors, we could populate this automatically.

packages/core/diagnostic/src/diagnostic.js
```diff
@@ -0,0 +1,125 @@
...
+
+  constructor(opts: ThrowableDiagnosticOpts) {
+    // Make it kinda compatible with default Node Error
+    super(opts.diagnostic.message);
+    this.stack = opts.diagnostic.stack || super.stack;
+
+    this.#diagnostic = opts.diagnostic;
+  }
+
+  toObject() {
```
devongovett: hmm - diagnostic is private, but exposed here? Just so it's read only?

packages/core/logger/src/Logger.js
```diff
@@ -60,6 +86,44 @@ class Logger {
 const logger = new Logger();
 export default logger;
 
+export type PluginLoggerOpts = {|
+  origin: string
+|};
+
+export class PluginLogger {
```
devongovett: if we're not going to expose a default `@parcel/logger` anymore, then maybe we should get rid of that package, move it into core, and put `PluginLogger` in the public folder like the other public objects?

packages/core/markdown-ansi/src/markdown-ansi.js
```diff
@@ -0,0 +1,23 @@
...
+
+// double char markdown matchers
+const BOLD_REGEX = /\*{2}([^*]+)\*{2}/g;
+const UNDERLINE_REGEX = /_{2}([^_]+)_{2}/g;
+const STRIKETHROUGH_REGEX = /~{2}([^~]+)~{2}/g;
+
+// single char markdown matchers
+const ITALIC_REGEX = /\*([^*]+)\*|_([^_]+)_/g;
+
+export default function markdownParser(input: string): string {
```
devongovett: whoa 😲

packages/core/utils/src/prettyDiagnostic.js
```diff
@@ -0,0 +1,55 @@
...
+import mdAnsi from '@parcel/markdown-ansi';
+
+type AnsiDiagnosticResult = {|
+  message: string,
+  stack: string,
+  codeframe: string,
+  hints: Array<string>
+|};
+
+export default function prettyDiagnostic(
```
devongovett: Should this be in `@parcel/reporter-cli`? Is it used in any other places?

packages/reporters/cli/src/Log.js
```diff
@@ -38,55 +57,83 @@ export function Log({event}: StringOrErrorLogProps) {
   throw new Error('Unknown log event type');
 }
 
-function InfoLog({event}: StringLogProps) {
-  return <Text>{event.message}</Text>;
+function Hints({hints}: {hints: Array<string>, ...}) {
+  return (
+    <div>
+      <div>Hints:</div>
```
devongovett: the heading is probably not needed? it's probably clear from context that the below hints are hints. :)

packages/core/diagnostic/src/diagnostic.js
DeMoorJasper: Yes, this would be the origin of the error so the plugin/parcel pkg that triggers it

packages/core/core/src/Validation.js
```diff
@@ -91,11 +97,30 @@ export default class Validation {
...
+          logger.error(error);
+        }
+
+        for (let warning of warnings) {
+          logger.warn(warning);
+        }
+
+        // Throw to fail build...
+        if (errors.length > 0) {
+          throw new ThrowableDiagnostic({
```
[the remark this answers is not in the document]
DeMoorJasper: Yeah... it would print twice was gonna refactor buildFailure and error handling a bit more in a follow-up, not entirely sure yet how to do it.

Arrays seem like a good idea

packages/core/diagnostic/src/diagnostic.js
DeMoorJasper: Yeah... probably doesn’t matter I’ll rename #diagnostic to diagnostic and remove toObject... seems obsolete

packages/core/utils/src/prettyDiagnostic.js
DeMoorJasper: It should probably be used by hmr and devserver to format the errors

packages/reporters/cli/src/Log.js
DeMoorJasper: Sure makes sense I’ll update accordingly :)

packages/core/logger/src/Logger.js
DeMoorJasper: Currently we do, but eventually it would be nice to pass a pluginLogger instance to each plugin call. Not entirely sure how to do it yet though. Creating PluginOptions per call seems expensive?

packages/core/diagnostic/src/diagnostic.js
DeMoorJasper: This is part of the mozilla Error spec https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/fileName

wbinnssmith: I think this is reason for us to choose something else like `filePath`. It looks like that `fileName` is the script file that raised the error itself [0], but this should be e.g. where a parse error occurred during parsing.


[0] https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/fileName

DeMoorJasper: PrintableError is actually the type for errors not created by parcel

PR #3654, opened 2019-10-17
React Refresh

### LINKED ISSUE #3499: Support Fast Refresh for React
Fast Refresh is a new hot reloading implementation for React that's first-class and doesn't have weird semantic quirks present in userland solutions. We first implemented it for RN, but integrating it on the web needs some work in the bundlers.

If you're interested in adopting it for Parcel, I wrote down first thoughts on how it could be done: https://github.com/facebook/react/issues/16604#issuecomment-528663101. This should be something you can reasonably expect to prototype in a few hours. There's a much more in-depth discussion on nuances to happen afterwards but I'd be happy to help steer that work and answer questions. Thanks for consideration!

mischnic:
# ↪️ Pull Request

This adds support for React Refresh, the successor of react-hot-loader.
Closes #3499

(AFAICT, create-react-app doesn't have this by default and there are currently multiple competing webpack plugins. I'm not aware of any web bundler implementation that correctly implements this like Metro)

A syntax error, and errors thrown during module initialization or render cause only the parent to rerender

Add charset to the dev server's headers to prevent the emoji in the HMRRuntime (`🚨`) from becoming `ðŸš¨`.

## 🚨 Test instructions

Running the `example/react-refresh` example, changing some component reloads that component while the random numbers at the top (local state) don't change.
This also works after a syntax error was created and resolved.

## 🗣  "Slightly controversial decisions"

- What babel parser options should a transformer use if it doesn't want to deal with the users babel config? Maybe we should create a utility function for that.
   - asset.meta.babelPlugins ?


parcel-benchmark:
## Benchmark Results

#### Timings

| --- | --- | --- |
| Cold | 10.17s | -140.00ms |
| Cached | 8.34s | -9.00ms |

#### Cold Bundles

| dist/legacy/parcel.ff215454.webp | 102.94kb | +0.00b | 215.00ms | -52.00ms 🚀 |
| dist/modern/parcel.76ee4591.webp | 102.94kb | +0.00b | 214.00ms | -52.00ms 🚀 |
| dist/modern/index.js | 794.00b | +0.00b | 484.00ms | -123.00ms 🚀 |
| dist/legacy/index.js | 252.00b | +0.00b | 486.00ms | -122.00ms 🚀 |
| dist/legacy/index.css | 29.00b | +0.00b | 1.44s | -122.00ms 🚀 |

#### Cached Bundles

| dist/modern/parcel.76ee4591.webp | 102.94kb | +0.00b | 11.00ms | -1.00ms 🚀 |
| dist/modern/index.js | 794.00b | +0.00b | 12.00ms | -1.00ms 🚀 |
| dist/legacy/index.js | 252.00b | +0.00b | 15.00ms | -1.00ms 🚀 |
| dist/legacy/index.css | 29.00b | +0.00b | 11.00ms | -1.00ms 🚀 |
| dist/modern/index.css | 29.00b | +0.00b | 10.00ms | -1.00ms 🚀 |

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 14.56s | -18.00ms |

#### Cold Bundles

*No bundle changes detected.*

#### Cached Bundles

</p></details>

**[Click here to view a detailed benchmark overview.](https://benchmark.parceljs.org/benchmark/246509433519079946)**

packages/runtimes/hmr/src/loaders/hmr-runtime.js
```diff
@@ -81,6 +81,9 @@ if ((!parent || !parent.isParcelRequire) && typeof WebSocket !== 'undefined') {
       document.body.appendChild(overlay);
     }
   };
+  ws.onclose = function(e) {
+    console.warn('[parcel] 🚨 Connection to the HMR server was lost');
```
mischnic: I added this because the page needs to be refresh after parcel was restarted (otherwise HMR silently stops)

packages/transformers/babel/src/jsx.js
```diff
@@ -49,7 +52,14 @@ export default async function getJSXOptions(config: Config) {
 
   if (pragma || JSX_EXTENSIONS[path.extname(config.searchPath)]) {
     return {
-      plugins: [['@babel/plugin-transform-react-jsx', {pragma, pragmaFrag}]]
+      plugins: [
+        ['@babel/plugin-transform-react-jsx', {pragma, pragmaFrag}],
+        reactLib === 'react' &&
+          (await options.packageManager.resolve(
```
mischnic: There need to be check here if the `react-refresh-transformer` is running or this needs to be executed separately in that transformer.

packages/transformers/react-refresh/src/ReactRefreshTransformer.js
```diff
@@ -0,0 +1,153 @@
...
+    });
+
+    asset.addDependency({
+      moduleSpecifier: wrapperPath
+    });
+
+    return [asset];
+  },
+
+  async generate({asset, options}) {
```
mischnic: Does this need to be duplicated from the JSTransformer ?

packages/transformers/react-refresh/src/ReactRefreshTransformer.js
```diff
@@ -0,0 +1,169 @@
...
+        asset.filePath,
+        false
+      );
+      reactRefreshBabelPlugin = result.resolved;
+    } catch (_) {
+      if (!didWarnAboutInstallingReactRefresh) {
+        logger.warn(
+          'Please install the `react-refresh` package for React hot reloading'
+        );
+        didWarnAboutInstallingReactRefresh = true;
```
mischnic: This warns once per Worker. Is there a better solution?

gaearon:
Can you check if converting from function component to a class and back works and doesn't get stuck? This would verify the "refresh boundary" logic is correct?

mischnic:
Thank you for suggesting this stress test.
It doesn't work 😞 (nothing is updated), but my wrapper calls `Refresh.register` every time (on both the function and the class). 

```js
Refresh.register(function FunctionDefault() {
  return _react.default.createElement("h1", null, "Default Export Function");
}, 'e7c468969341ee7cfcd0f5a483ddf139 exports%default')
// <refresh>
Refresh.register(class FunctionDefault extends _react.default.Component {
  render() {
    return _react.default.createElement("h1", null, "Default Class");
  }
}, 'e7c468969341ee7cfcd0f5a483ddf139 exports%default' )
```

https://github.com/parcel-bundler/parcel/blob/5f55a4f43ccca404e20ccdca5f976d3752c3b2e1/packages/transformers/react-refresh/src/helpers/helpers.js#L83

So when changing from a function component to a class and then to a function back nothing changes in the process, but after that, updating the contents of the function component works as expected.

What should I do differently here? From my perspective it looks like this is rather caused by react-refresh itself...

mischnic:
@gaearon https://github.com/pmmmwh/react-refresh-webpack-plugin doesn't work either in this situation. Are you sure this isn't caused by `react-refresh` itself?

mischnic:
@gaearon Disabling the `react-fresh/babel` fixes the issue (function components work, classes work, switching between the two works - though I'm sure the babel plugin doesn't something state related as well?). Are you sure this isn't caused by `react-refresh` itself?

The structure of the wrapped assets is:
```js
var helpers = require("....");

var prevRefreshReg = window.$RefreshReg$;
var prevRefreshSig = window.$RefreshSig$;
helpers.prelude(module);

try {
  var _react = _interopRequireDefault(require("react"));


  class FunctionDefault extends _react.default.Component {
    render() {
      return _react.default.createElement("h1", null, "Default Class (XXX)");
    }
  }
  exports.default = FunctionDefault;
} finally {
  window.$RefreshReg$ = prevRefreshReg;
  window.$RefreshSig$ = prevRefreshSig;
}

if (isReactRefreshBoundary(module)) { // all exports are likely components
  registerExportsForReactRefresh(module); // Refresh.refresh() for every component export

  module.hot.accept();
  if (Refresh.hasUnrecoverableErrors()) {
    window.location.reload();
  }
  enqueueUpdate(Refresh);
}
```

mischnic:
@gaearon I don't want to annoy you, but could you please tell me if I'm doing something wrong her or if there's a problem with `react-refresh`?

mischnic:
I'm starting to think that this isn't possible with just wrapping assets with a prelude and postlude. Metro has quite a deep integration in their prelude (a module's parent needs to be rerun if it's signature changed....). We'll have to extend our hmr-runtime

mischnic:
Everything seems to work now, but the `react-refresh/babel` babel plugin needs to run before preset-env or anything else that transpiles ES6 imports. Running the plugin in `@parcel/transformer-babel` isn't ideal...

packages/transformers/react-refresh-babel/src/ReactRefreshBabelTransformer.js
```diff
@@ -0,0 +1,117 @@
...
+      program: parse(code, {
+        filename: this.name,
+        allowReturnOutsideFunction: true,
+        strictMode: false,
+        sourceType: 'module',
+        plugins: [
+          'exportDefaultFrom',
+          'exportNamespaceFrom',
+          'dynamicImport',
+          'jsx'
```
mischnic: What babel parser options should a transformer use if it doesn't want to deal with the users babel config? Maybe we should create a utility function for that.

packages/transformers/react-refresh-babel/src/ReactRefreshBabelTransformer.js
```diff
@@ -0,0 +1,117 @@
...
+import {transformFromAst} from '@babel/core';
+import generate from '@babel/generator';
+import {parse} from '@babel/parser';
+
+function shouldExclude(asset, options) {
+  return (
+    !asset.env.isBrowser() ||
+    !options.hot ||
+    !asset.isSource ||
+    asset.filePath.includes('packages/runtimes')
```
mischnic: There should be a way to distinguish "user" assets (`isSource` isn't enough) from node_modules / generated ones.

packages/runtimes/react-refresh/src/ReactRefreshRuntime.js
```diff
@@ -0,0 +1,40 @@
...
+    let code;
+    if (mainEntry && mainEntry.meta.reactRefreshRuntimePath !== undefined) {
+      let runtime = mainEntry.meta.reactRefreshRuntimePath;
+      invariant(typeof runtime === 'string');
+
+      code = `
+var Refresh = require('${path.relative(__dirname, runtime)}');
+
+Refresh.injectIntoGlobalHook(window);
+window.$RefreshReg$ = function() {};
```
devongovett: Are we sure these functions aren't supposed to do something?

packages/transformers/react-refresh-babel/src/ReactRefreshBabelTransformer.js
devongovett: Won't this not work outside the monorepo?

packages/transformers/react-refresh-babel/src/ReactRefreshBabelTransformer.js
```diff
@@ -0,0 +1,117 @@
...
+    )).resolved;
+
+    let code = await asset.getCode();
+    let transformResult = transformFromAst(ast.program, code, {
+      code: false,
+      ast: true,
+      filename: asset.filePath,
+      babelrc: false,
+      configFile: false,
+      plugins: [reactRefreshBabelPlugin]
```
devongovett: Should we just add this plugin to the babel transformer like we have for JSX, TS, Flow, etc? Would we one less transform pass and a bunch less duplicated code.

packages/transformers/react-refresh-wrap/src/ReactRefreshWrapTransformer.js
```diff
@@ -0,0 +1,133 @@
...
+      );
+
+      res.code = generated.code;
+      // $FlowFixMe...
+      res.map = new SourceMap(generated.rawMappings, {
+        [sourceFileName]: null
+      });
+    }
+
+    if (asset.meta.globals && asset.meta.globals.size > 0) {
```
devongovett: shouldn't be needed - JSTransformer already handles that.

packages/transformers/react-refresh-wrap/src/ReactRefreshWrapTransformer.js
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    asset.meta.reactRefreshRuntimePath = (await options.packageManager.resolve(
+      'react-refresh/runtime',
+      __filename
+    )).resolved;
+
+    let wrapperPath = path.relative(path.dirname(asset.filePath), WRAPPER);
+
+    ast.program.program.body = wrapper({
```
devongovett: Would this be simpler as a string transform instead of an AST one? Could possibly avoid most of the code in this transformer for parsing/generating.

packages/transformers/react-refresh-babel/src/ReactRefreshBabelTransformer.js
mischnic: Yes, this is a hack for testing until I find a better way.

packages/transformers/react-refresh-wrap/src/ReactRefreshWrapTransformer.js
mischnic: But then I'd have to patch the sourcemap?

packages/runtimes/react-refresh/src/ReactRefreshRuntime.js
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    let mainEntry = bundle.getMainEntry();
+    let code;
+    if (mainEntry && mainEntry.meta.reactRefreshRuntimePath !== undefined) {
+      let runtime = mainEntry.meta.reactRefreshRuntimePath;
+      invariant(typeof runtime === 'string');
+
+      code = `
+var Refresh = require('${path.relative(__dirname, runtime)}');
```
devongovett: Could you just put `require('react-refresh/runtime')` here? Then, below set `filePath` to be `path.join(path.dirname(mainEntry), 'react-refresh-entry.js')`. That way, it will resolve the `react-refresh` module relative to the same directory as the main entry. That would avoid the need to resolve it in the transformer, and also the hack to check if it's in the monorepo (path would contain node_modules, so `isSource` would be correct).

packages/transformers/react-refresh-wrap/src/ReactRefreshWrapTransformer.js
devongovett: ugh, sourcemaps! 🙈

packages/runtimes/react-refresh/src/ReactRefreshRuntime.js
mischnic: Then react-refresh has to be a dependency of the runtime as well and the react-refresh instance in the transformer and in the runtime **has** to be the same

packages/runtimes/react-refresh/src/ReactRefreshRuntime.js
mischnic: Yes, every asset has a prelude, where there hooks are set to an actual implementation, and a postlude, where they are set to their previous values (like a stack). So the function at the very bottom has to be defined.
Metro does this and Dan also put this into his implementers guide.

packages/runtimes/react-refresh/src/ReactRefreshRuntime.js
devongovett: oh, well if it's a dep of both `@parcel/transformer-react-refresh-wrap` and `@parcel/runtime-react-refresh`, then you could set `filePath` to `__filename` like you have it and it would resolve relative to the runtime.

mischnic: > well if it's a dep of both

that doesn't work, it wouldn't guarantee that both use the same `react-refresh` asset/module in the browser. Just like React doesn't really like it if there are two instances of it running on one page.

mischnic: > Would only be different if we mess it up

We would rely on the package manager that is installing parcel to hoist properly.

devongovett: oh - are there npm-alikes that don't do that?

mischnic: Theoretically, the version ranges could be resolved to different versions (not sure if this could happen when updating), you could craft a lockfile that forces different versions for the two packages (unless we pin the version).
Not sure if a package manager still does that. But I think this is an implementation detail that was never part of the node_modes "spec".

mischnic: I have made them peerDependencies (and only one plugin with a "hard" dependency), now I'd say it isn't our fault 🤷‍♂

packages/transformers/react-refresh-babel/src/ReactRefreshBabelTransformer.js
```diff
@@ -0,0 +1,38 @@
...
+    !options.hot ||
+    !asset.isSource ||
+    asset.filePath.includes('packages/runtimes')
+  );
+}
+
+export default new Transformer({
+  async transform({asset, options}) {
+    // TODO tests break with this?
+    //    asset.type = 'js';
```
mischnic: 🤷‍♂

packages/transformers/react-refresh-wrap/src/ReactRefreshWrapTransformer.js
mischnic: But if the AST from the JSTransformer was reused, then JSTransformer#generate is never called?

packages/transformers/react-refresh-babel/src/ReactRefreshBabelTransformer.js
devongovett: makes sense. the JS transformer should handle this

packages/configs/default/index.json
```diff
@@ -3,8 +3,10 @@
   "transforms": {
     "types:*.{ts,tsx}": ["@parcel/transformer-typescript-types"],
     "*.{js,mjs,jsm,jsx,es6,ts,tsx}": [
+      "@parcel/transformer-react-refresh-babel",
```
devongovett: What’s the performance impact of including these extra transformers in the default config? If I’m building a non-React app, I don’t need these technically. If it’s minimal, then we should keep them here. Otherwise, we should investigate making it faster somehow or make a `@parcel/config-react` preset.

packages/runtimes/react-refresh/src/ReactRefreshRuntime.js
```diff
@@ -0,0 +1,31 @@
...
+window.$RefreshReg$ = function() {};
+window.$RefreshSig$ = function() {
+  return function(type) {
+    return type;
+  };
+};`,
+        isEntry: true
+      };
+    } else {
+      // We don't need that in a child bundle (beacuse the main bundle will have set the global)
```
devongovett: The above check will only skip split bundles, not normal async bundles - those still have entry assets. Do we need to skip those too? If so, you could check `bundle.isEntry` as well.

packages/transformers/js/src/JSTransformer.js
```diff
@@ -66,7 +66,6 @@ export default new Transformer({
   },
 
   async transform({asset, options}) {
-    asset.type = 'js';
```
devongovett: 🤔

packages/transformers/react-refresh-babel/src/ReactRefreshBabelTransformer.js
```diff
@@ -0,0 +1,27 @@
...
+import {Transformer} from '@parcel/plugin';
+import invariant from 'assert';
+
+function shouldExclude(asset, options) {
+  return !asset.env.isBrowser() || !options.hot || !asset.isSource;
+}
+
+export default new Transformer({
+  async transform({asset, options}) {
+    if (!shouldExclude(asset, options)) {
```
devongovett: What if it’s not a react app? Do we need to detect if `react` is in package.json or something?

packages/transformers/babel/src/BabelTransformer.js
```diff
@@ -24,6 +24,10 @@ export default new Transformer({
   async transform({asset, config, options}) {
     // TODO: come up with a better name
     if (config?.config) {
+      if (asset.meta.babelPlugins && Array.isArray(asset.meta.babelPlugins)) {
+        // $FlowFixMe
+        config.config.plugins.push(...asset.meta.babelPlugins);
```
devongovett: I wonder about mutating the config... Does this affect the cache?

packages/transformers/react-refresh-wrap/src/ReactRefreshWrapTransformer.js
devongovett: Mmm, yeah. This is annoying. We should possibly rethink how parsing/generating works. Maybe a separate “parser” plugin type would be useful to centralize this.

packages/transformers/js/src/JSTransformer.js
mischnic: I'm not using `.jsx` anymore in the tests and have added this bug to the Todo.

packages/runtimes/react-refresh/src/ReactRefreshRuntime.js
mischnic: Well:
```
.parcel-cache/dist/index.html isEntry: true
.parcel-cache/dist/react.7ae39257.js isEntry: false
.parcel-cache/dist/LazyComponent.b9708ad0.js isEntry: false
```

(I have added the react in package.json logic here as well)

packages/transformers/react-refresh-babel/src/ReactRefreshBabelTransformer.js
mischnic: Done (and also safe if the babel transform has some bug)

packages/configs/default/index.json
mischnic: Do you mean that the transformers themselves (the overhead) or the action that they perform?

packages/core/integration-tests/test/react-refresh.js
```diff
@@ -0,0 +1,144 @@
...
+const inputDir = path.join(__dirname, '/input');
+
+describe('react-refresh', function() {
+  let b,
+    root,
+    window,
+    subscription,
+    randoms = {};
+
+  before(async () => {
```
devongovett: I'd recommend making this a `beforeEach` so we don't reuse state between test runs.

packages/core/integration-tests/test/react-refresh.js
```diff
@@ -0,0 +1,144 @@
...
+    await sleep(100);
+    await fs.rimraf(inputDir);
+    await sleep(100);
+    await fs.ncp(path.join(__dirname, '/integration/react-refresh'), inputDir);
+    let port = await getPort();
+    b = bundler(path.join(inputDir, 'index.js'), {
+      outputFS: fs,
```
devongovett: Maybe use the memory fs for input instead. Should be more reliable on Windows if possible.

packages/core/fs/src/OverlayFS.js
```diff
@@ -68,7 +68,19 @@ export class OverlayFS implements FileSystem {
 
   readFile = read('readFile');
   writeFile = write('writeFile');
-  copyFile = write('copyFile');
+  async copyFile(source: FilePath, destination: FilePath) {
```
mischnic: copyFile didn't work across the writable and readable fs

packages/configs/default/index.json
devongovett: The overall cost

PR #3662, opened 2019-10-21
Support shared bundles in workers

devongovett:
This adds support for shared bundles between web workers, and between the main thread and web workers. For example, you might import lodash in two different workers, or in both the main thread and a worker. Since lodash is large, it will now get placed in its own bundle and loaded separately in the worker with `importScripts()`.

This also fixes some bugs with scope hoisting and shared bundles in general, in particular the prelude not being included where needed sometimes.


packages/core/core/src/Environment.js
```diff
@@ -97,3 +98,14 @@ export function mergeEnvironments(
     ...b
   });
 }
+
+export function getEnvironmentHash(env: Environment) {
+  // context is excluded from hash so that assets can be shared between e.g. workers and browser.
+  // Different engines should be sufficient to distinguish multi-target builds.
```
devongovett: I'm not 100% sure about doing this. It's theoretically possible for a transformer to return different content if `env.context === 'web-worker'` for example, and this effectively prevents that. The benefits of sharing assets between web workers and main thread are pretty compelling though. I guess the only other option is to remove workers as separate contexts from browser...

packages/core/core/src/BundleGraph.js
```diff
@@ -278,17 +278,11 @@ export default class BundleGraph {
...
-        )
-        .map(node => this._graph.getNodesConnectedTo(node, 'bundle'))
-        .reduce((acc, v) => acc.concat(v), [])
-        .filter(node => node.type === 'bundle' && node.value.type === type)
-        .length > 0
-    );
+    return this._graph
+      .getNodesConnectedTo(nullthrows(this._graph.getNode(bundle.id)), 'bundle')
+      .map(node => this._graph.getNodesConnectedTo(node, 'bundle'))
+      .reduce((acc, v) => acc.concat(v), [])
```
wbinnssmith: `flatMap` 🙂

mischnic: Supported in Node 11+ 😞

wbinnssmith: sorry, should have clarified: `import {flatMap} from '@parcel/utils';`. We have our own 😄

PR #3673, opened 2019-10-24
Parcel 2: HTMLNano Optimizer

### LINKED ISSUE #3366: Parcel 2: htmlnano optimizer
Create the `@parcel/optimizer-htmlnano` package in `packages/optimizers/htmlnano`. This package is a Parcel 2 optimizer plugin for HTML. It should be based on the htmlnano transform from Parcel 1.

The `@parcel/optimizer-htmlnano` package is responsible for the following things:

* Loading htmlnano config from `.htmlnanorc` and `.htmlnanorc.js`
* Applying PostHTML with the htmlnano plugin to minify HTML

gonzarodriguezt:
### ↪️ Pull Request

Fixes #3366

**Obs.:** There's a super weird bug that makes the svg attribute `baseProfile` lowercase, because of that the original test was failing, that's the reason behind [this change](https://github.com/parcel-bundler/parcel/compare/v2...gonzarodriguezt:bugfix/optimizer-htmlnano?expand=1#diff-728d2fc29ee35fbd3dee59dfdc80d85dR260), i'm aware that is not the best solution, but the change was useful to explain it, because i couldn’t find the origin of that bug.

I tried to reproduce it with `htmlnano: ^2.2` in a local repo but i didn’t succeeded. Then i disabled the `minify: true` option and the attribute was lowercased anyway, so maybe that's made in a previous transformer or something.

### ✔️ PR Todo


packages/optimizers/htmlnano/src/HTMLNanoOptimizer.js
```diff
@@ -0,0 +1,38 @@
...
+
+    if (typeof contents !== 'string') {
+      throw new Error(
+        'HTMLNanoOptimizer: Only string contents are currently supported'
+      );
+    }
+
+    let userConfig = await loadConfig(
+      options.inputFS,
+      nullthrows(bundle.getMainEntry()).filePath,
```
mischnic: Not all bundles have a main entry, please use `path.join(options.projectRoot, 'index')` for now.

Issue #3305, opened 2019-08-02
Support "external" modules in Parcel 2


wbinnssmith:
Parcel could allow listing of "external" modules which won't be bundled.

mischnic:
Related?: #144

maximal:
+1 here

This keeps me from switching to Parcel on about 5 projects.

devongovett:
@maximal what are you trying to accomplish with externals? trying to gather usecases to help inform this feature for v2.

devongovett:
Seems like maybe there's two separate but related features here:

1. Excluding a module from a bundle, but leaving the import or require there. Use case would be for e.g. libraries with peer dependencies, builtin modules in node/electron, etc. Maybe this could be solved by excluding `peerDependencies`.
2. Mapping an import to a global variable that is available somehow. Use case would be for e.g. libraries from a cdn. This might need separate configuration, e.g. with aliases.

charlielow:
> @maximal what are you trying to accomplish with externals? trying to gather usecases to help inform this feature for v2.

For us the need is to exclude some 3rd party modules from the bundle to avoid duplicate loading on pages where there may be more than one independent React app. 

Rather than include React / ReactDOM, let both apps share a globally available version of each of these two modules.

This also allows for browser caching of these large assets whereas if they were packaged in multiple different bundles, caching of these individual files would not be possible.

Ideally Parcel would have an option similar to Rollup's `external`

`
external: [
  'react',
  'react-dom',
  'jquery'
]
`

ericsoco:
I think this is what I'm running into in #4003.

I'm trying to bundle a web application and it looks like the issue may be dependencies that conditionally `require` other deps, so one more vote for conditional `requires` per https://github.com/parcel-bundler/parcel/issues/3305#issuecomment-542772700.

FreeFrags:
I noticed that #3753 was merged and closed doesnt it cover this issue or am i missing something here?

mischnic:
https://github.com/parcel-bundler/parcel/pull/3753 isn't merged yet

DeMoorJasper:
Can this be closed since #4072 has been merged?

panlina:
@DeMoorJasper I'm new to Parcel. From #4072 I didn't see how `externals` is implemented. It lets you set `"react": false` but where do I specify the global name (`React`) it resolves to?

DeMoorJasper:
[...]

PR #3674, opened 2019-10-25
Added support for externals through peer dependencies

subhero24:
Allows parcel to not bundle the peerDependencies of a project and leave them as external requires. This makes it convenient to use parcel for bundling libraries or packages linked locally.

#144 
#3305 

It works by skipping the package in the node resolve algorithm if it was specified as a peer dependency   (as if it was not there in the first place). The resolve algorithm can thus find the package higher up the tree where it was not specified as a peer dependency. In case it did not resolve (as all instances of the package in the tree are specified as a peer dependency), it will be ignored as a dependency and leave the require statement in place.

I find this useful for consuming local packages that have for example a "react" peer dependency. React should not be bundled multiple times as this makes React complain. Also useful to leave "@babel/runtime" requires in place for packages/libraries that want to depend on the babel runtime.

### DIFF
--- a/packages/core/parcel-bundler/src/Bundler.js
+++ b/packages/core/parcel-bundler/src/Bundler.js
@@ -18,6 +18,7 @@ const PromiseQueue = require('./utils/PromiseQueue');
 const installPackage = require('./utils/installPackage');
 const bundleReport = require('./utils/bundleReport');
 const prettifyTime = require('./utils/prettifyTime');
+const getModuleParts = require('./utils/getModuleParts');
 const getRootDir = require('./utils/getRootDir');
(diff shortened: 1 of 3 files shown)


devongovett:
Would you like to take a shot at implementing this in v2? The resolver is here: https://github.com/parcel-bundler/parcel/blob/v2/packages/resolvers/default/src/DefaultResolver.js#L143. You can return `isExcluded: true` to prevent a module from being bundled, so it should be even simpler than in v1. I don't think we need a CLI flag for this.

Going to close this PR. We aren't adding any new features to v1 at this point.

subhero24:
Ok, I will give this a go for v2

PR #3683, opened 2019-10-26
postcss-modules: make getJSON customizable

Epimodev:
# ↪️ Pull Request

In `postcss-modules` plugin, there is `getJSON` option which give the possibility to run some instructions with exported classes.
In my case, I want to use this feature to generate TypeScript definition files when I change my css/scss files. And we can imagine other use cases of this feature.

But when we define `getJSON` in `postcss.config.js` we overwrite the `getJSON` of parcel which set classNames in css assets, so css modules doesn't work anymore in the application.

The goal of this PR is to avoid overwrite the behavior of parcel when we define `getJSON` in `postcss.config.js`.

## 💻 Examples

posrtcss.config.js
```js
module.exports = {
  modules: true,
  plugins: {
    autoprefixer: {},
    'postcss-modules': {
      generateScopedName: '[local]___[hash:base64:5]',
      getJSON: (filename, json) => {
        const classNames = Object.keys(json);
        return writeTypings(filename, classNames);
      },
    },
  },
};
```

css file `style.css`:
```css
.button {
  padding: 10px;
  background: green;
  color: white;
  border-radius: 3px;
}
```

js File:
```js
import style from './style.css';

console.log(style); // expect `{ button: "button___1MgdH" }` but display `{}`
```

## 🚨 Test instructions

- `git clone git@github.com:Epimodev/react-parcel-boilerplate.git` it will avoid set up a parcel project from scratch
- `git checkout feat/write-css-types` checkout the branch where I configure postcss to generate TS definition of css files with `getJSON` option
- `yarn` to install dependencies
- `yarn link "parcel-bundler"` to use local version of parcel
- `yarn start --no-cache` will start a dev server on port 1234
- once the app is started you can open it in your browser
  - if you use current version of parcel, buttons shouldn't have any style and the console should display `STYLE {}`
  - if you use my forked version, buttons should be green and the console should display `STYLE {button: "button___1MgdH"}`

## ✔️ PR Todo

- [ ] Added/updated unit tests for this change (I didn't find any unit test about postcss)
- [ ] Included links to related issues/PRs (I didn't find related issues/PRs)

### DIFF
--- a/packages/core/parcel-bundler/src/transforms/postcss.js
+++ b/packages/core/parcel-bundler/src/transforms/postcss.js
@@ -43,20 +43,25 @@ async function getConfig(asset) {
     throw new Error('PostCSS config should be an object.');
   }
 
-  let postcssModulesConfig = {
-    getJSON: (filename, json) => (asset.cssModules = json),
-    Loader: createLoader(asset),
-    generateScopedName: (name, filename) =>
-      `_${name}_${md5(filename).substr(0, 5)}`
-  };
-
+  let userPostcssModulesConfig = {};
   if (config.plugins && config.plugins['postcss-modules']) {
-    postcssModulesConfig = Object.assign(
-      postcssModulesConfig,
-      config.plugins['postcss-modules']
-    );
+    userPostcssModulesConfig = config.plugins['postcss-modules'];
+
     delete config.plugins['postcss-modules'];
   }
+
+  let postcssModulesConfig = {
+    Loader: createLoader(asset),
+    generateScopedName: (name, filename) =>
+      `_${name}_${md5(filename).substr(0, 5)}`,
+    ...userPostcssModulesConfig,
+    getJSON: (filename, json, outputFilename) => {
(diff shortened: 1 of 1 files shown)


garthenweb:
@Epimodev Just in case you didn't saw it before: As I had the exact same use case I build a plugin that should fulfill your need: https://www.npmjs.com/package/parcel-plugin-css-modules-type-generator Maybe it is useful for you as well :)

Epimodev:
@garthenweb I didn't know this plugin exists, thanks a lot 👍   
I've just tried it and it seems working perfectly. So I'll use it.

I still think we should be able to use `getJSON` option without overwrite css asset behavior.

DeMoorJasper:
@Epimodev Could you port this PR over to the v2 branch, as there's a plugin for this in Parcel 1 I don't think we should add this functionality as Parcel 1 is in maintenance mode.

Epimodev:
No problem to port this over v2.
I will try during next days.

PR #3687, opened 2019-10-27
[T-88] Verify target descriptors

mischnic:
# ↪️ Pull Request

Verifies target descriptors from package.json and when passed via the API.

Open questions:
- Should we allow engine properties apart from `npm`, `node`, `parcel`, `electron`, `browser`?  Asking because `browser` vs `browsers` hat a lot of potential for typos
- Is `isLibrary` a valid target descriptor option?

Bildschirmfoto 2019-11-01 um 14 09 38


parcel-benchmark:
## Benchmark Results

#### Timings

| --- | --- | --- |

#### Cold Bundles

#### Cached Bundles

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 14.72s | +254.00ms |
| Cached | 9.12s | +25.00ms |

#### Cold Bundles

*No bundle changes detected.*

#### Cached Bundles

| dist/main/index.js | 34.53kb | +0.00b | 15.00ms | +3.00ms ⚠️ |
| dist/module/index.js | 34.33kb | +0.00b | 11.00ms | +2.00ms ⚠️ |

</p></details>

**[Click here to view a detailed benchmark overview.](https://benchmark.parceljs.org/benchmark/247372635527184905)**

packages/core/core/src/TargetResolver.js
```diff
@@ -37,6 +41,228 @@ const DEFAULT_PRODUCTION_ENGINES = {
 const DEFAULT_DIST_DIRNAME = 'dist';
 const COMMON_TARGETS = ['main', 'module', 'browser', 'types'];
 
+function parseEngines(
+  targetName: ?string,
+  engines: mixed
+): Engines | typeof undefined {
+  if (engines === undefined) {
+    return;
+  } else if (engines && typeof engines === 'object') {
```
wbinnssmith: by not differentiating between null and undefined, this can just be:

```
): ?Engines {
  if (engines == null) {
    return;
  } else if (typeof engines === 'object') {
```

packages/core/core/src/TargetResolver.js
```diff
@@ -37,6 +41,228 @@ const DEFAULT_PRODUCTION_ENGINES = {
...
+      }: ${String(engines)}`
+    );
+  }
+}
+
+function parseDescriptor(
+  targetName: string,
+  descriptor: mixed
+): {|
+  context?: EnvironmentContext | typeof undefined,
```
wbinnssmith: So far we haven't been differentiating between null and undefined, using `?EnvironmentContext` in a case like this

packages/core/core/test/TargetResolver.test.js
```diff
@@ -308,4 +316,124 @@ describe('TargetResolver', () => {
...
+    await rejects(
+      () => targetResolverUnknown.resolve(COMMON_TARGETS_FIXTURE_PATH),
+      {
+        message:
+          'Unexpected properties in descriptor for target "main": "somethingElse"'
+      }
+    );
+  });
+
+  it.only('rejects invalid or unknown fields', async () => {
```
wbinnssmith: debug 😄

packages/core/core/test/TargetResolver.test.js
```diff
@@ -8,6 +8,9 @@ import {DEFAULT_OPTIONS} from './utils';
 
 import TargetResolver from '../src/TargetResolver';
 
+//$FlowFixMe
+const rejects = assert.rejects;
```
wbinnssmith: why not `assert.rejects` below?

packages/core/core/src/TargetResolver.js
mischnic: Because most properties in `TargetDescriptor` are optional and have a non-nullish type, for example:

This fails:
```js
type Opts = {
  context?: string
};

let value: ?string = null;

let opts: Opts = {
  context: value
}
```

This works:
```js
type Opts = {
  context?: string
};

let value: string | typeof undefined = ”xyz";

let opts: Opts = {
  context: value
}
```

packages/core/core/test/TargetResolver.test.js
mischnic: 🙈

packages/core/core/test/TargetResolver.test.js
mischnic: I wanted to avoid a `//$FlowFixMe` comment on every assert, and two `it` test cases use the function. Should I completely remove this declaration or make two local variables in `it`?

packages/core/core/src/TargetResolver.js
devongovett: I think the `Opts` type needs to be this:

```js
type Opts = {
  context?: ?string
}
```

packages/core/core/src/TargetResolver.js
```diff
@@ -37,6 +43,362 @@ const DEFAULT_PRODUCTION_ENGINES = {
...
+    publicUrl,
+    sourceMap: _sourceMap,
+    ...rest
+  } = descriptor;
+
+  if (Object.keys(rest).length !== 0) {
+    let keys = Object.keys(rest)
+      .map(k => {
+        let hint;
+        for (let l of LIST_OF_DESCRIPTOR_KEYS) {
```
devongovett: Should we use some fuzzy string matching algorithm to get the closest hint? 😜

packages/core/core/src/TargetResolver.js
```diff
@@ -37,6 +43,362 @@ const DEFAULT_PRODUCTION_ENGINES = {
...
+  if (engines === undefined) {
+    return;
+  } else if (engines && typeof engines === 'object') {
+    let result = {};
+    for (let key in engines) {
+      let value: mixed = engines[key];
+      if (key === 'browsers' && Array.isArray(value)) {
+        value = value.map(v => {
+          if (typeof v !== 'string') {
+            throw new Error(
```
devongovett: Should these be diagnostics too?

devongovett:
Awesome stuff! Love the pretty error messages! 💯 

Did you think about maybe using a schema-based validation library for this? e.g. something like json-schema? Wonder if it might reduce the amount of manual validation code... Maybe not with the custom errors. Just a thought.

packages/core/core/src/TargetResolver.js
mischnic: Correct, the push yesterday was more of a POC  😄

packages/core/core/src/TargetResolver.js
mischnic: I'm now using all possible values sorted by their levenshtein distance that require less than `actualValue.length / 2` modifications. Otherwise all possible values are printed (and this distinction is also reflected in the hints).

Bildschirmfoto 2019-11-01 um 14 09 38

Performance-wise, in the screenshot these calculations take 0.879ms, 0.179ms, 0.063ms, 0.518ms. (Clearly correlated with the number of items).

This could probably be optimized with a levenshtein automaton instead (where the calculation aborts after the max number of steps is reached rather than filtering them out afterwards), but that would probably be overkill?

mischnic:
> Did you think about maybe using a schema-based validation library for this? e.g. something like json-schema? Wonder if it might reduce the amount of manual validation code... Maybe not with the custom errors. Just a thought

In the beginning, I opted for manual checking because it made flow happy. But adding a value to an enum wouldn't display an error in that code and using a schema is clearly a cleaner approach.

~library~ 🙈 
I tried `ajv`, but with `oneOf` it displayed all possible down the tree and not the most likely (the branch where the top-level type is correct and where the error occurs further down).

packages/core/core/src/TargetResolver.js
mischnic: Using a schema instead now

devongovett:
Awesome! This looks great. I love the schema validation library with support for diagnostics. Will be very useful for other validations we want to do. Maybe we could validate `.parcelrc` the same way. 🎉

PR #3688, opened 2019-10-27
V2: Inline process.browser for better code elimination

garthenweb:
# ↪️ Pull Request

Migrates #2583 to v2. This feature is heavily used in some projects and can save a lot of dead code to be eliminated.

I made two changes compared to v1:
* The `process.browser = true` assignment that might be somewhere in the code is not replaced anymore but always set to `true`. Removing this line could lead to breaking the code in rare situations, but not changing it too `true` would be inconsequent as well. Let me know if you have any thoughts about that.
* The replacement should also be done for Node and Electron. I think there is no need to keep any reference as those should always be static. `process.browser` is different from `process.env` which can change after build time on the server. Tests for Node and Electron as skipped for now, as all those tests were skipped so far. I expect this to be implemented later for v2.

## 💻 Examples

See  #2583.

## 🚨 Test instructions

Place `process.browser` somewhere in your code and see it is replaced with `true` when build for the target browser.

### DIFF
--- a/packages/transformers/js/src/JSTransformer.js
+++ b/packages/transformers/js/src/JSTransformer.js
@@ -4,7 +4,7 @@ import semver from 'semver';
 import generate from '@babel/generator';
 import {Transformer} from '@parcel/plugin';
 import collectDependencies from './visitors/dependencies';
-import envVisitor from './visitors/env';
+import processVisitor from './visitors/process';
 import fsVisitor from './visitors/fs';
 import insertGlobals from './visitors/globals';
 import {parse} from '@babel/parser';
@@ -17,6 +17,7 @@ import SourceMap from '@parcel/source-map';
 
 const IMPORT_RE = /\b(?:import\b|export\b|require\s*\()/;
 const ENV_RE = /\b(?:process\.env)\b/;
+const BROWSER_RE = /\b(?:process\.browser)\b/;
 const GLOBAL_RE = /\b(?:process|__dirname|__filename|global|Buffer|define)\b/;
 const FS_RE = /\breadFileSync\b/;
 const SW_RE = /\bnavigator\s*\.\s*serviceWorker\s*\.\s*register\s*\(/;
@@ -46,6 +47,7 @@ export default new Transformer({
       !options.scopeHoist &&
       !canHaveDependencies(code) &&
       !ENV_RE.test(code) &&
+      !BROWSER_RE.test(code) &&
       !FS_RE.test(code)
     ) {
       return null;
@@ -74,9 +76,17 @@ export default new Transformer({
     let ast = asset.ast;
     let code = await asset.getCode();
 
-    // Inline environment variables
-    if (!asset.env.isNode() && (ast.isDirty || ENV_RE.test(code))) {
-      walk.simple(ast.program, envVisitor, {asset, env: options.env});
+    // Inline process/ environment variables
+    if (
+      (!asset.env.isNode() && (ast.isDirty || ENV_RE.test(code))) ||
+      (asset.env.isBrowser() && (ast.isDirty || BROWSER_RE.test(code)))
+    ) {
+      walk.ancestor(ast.program, processVisitor, {
+        asset,
+        env: options.env,
+        isNode: asset.env.isNode(),
+        isBrowser: asset.env.isBrowser(),
+      });
     }
 
     // Collect dependencies
--- a/packages/transformers/js/src/visitors/process.js
+++ b/packages/transformers/js/src/visitors/process.js
@@ -2,9 +2,9 @@ import * as types from '@babel/types';
 import {morph} from './utils';
 
 export default {
-  MemberExpression(node, {asset, env}) {
+  MemberExpression(node, {asset, env, isBrowser, isNode}, ancestors) {
     // Inline environment variables accessed on process.env
-    if (types.matchesPattern(node.object, 'process.env')) {
+    if (!isNode && types.matchesPattern(node.object, 'process.env')) {
       let key = types.toComputedKey(node);
       if (types.isStringLiteral(key)) {
         // Try using the value from the passed env (either from new Parcel
@@ -16,6 +16,19 @@ export default {
           asset.ast.isDirty = true;
           // asset.meta.env[key.value] = process.env[key.value];
         }
+      }
+      // Inline process.browser
+    } else if (isBrowser && types.matchesPattern(node, 'process.browser')) {
+      // the last ancestor is the node itself, the one before may be it's parent
+      const parent = ancestors[ancestors.length - 2];
+      const isAssignmentExpression = Boolean(
+        parent && types.isAssignmentExpression(parent) && parent.left === node,
+      );
+
+      if (isAssignmentExpression) {
+        parent.right = types.booleanLiteral(true);
+      } else {
+        morph(node, types.booleanLiteral(true));
       }
     }
   },
--- a/packages/core/integration-tests/test/javascript.js
+++ b/packages/core/integration-tests/test/javascript.js
@@ -1005,9 +1005,29 @@ describe('javascript', function() {
     assert.equal(output(), 'test:test');
   });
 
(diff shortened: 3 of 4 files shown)


packages/transformers/js/src/JSTransformer.js
```diff
@@ -74,9 +76,16 @@ export default new Transformer({
...
 
-    // Inline environment variables
-    if (!asset.env.isNode() && (ast.isDirty || ENV_RE.test(code))) {
-      walk.simple(ast.program, envVisitor, {asset, env: options.env});
+    // Inline process/ environment variables
+    if (
+      !asset.env.isNode() &&
+      (ast.isDirty || ENV_RE.test(code) || BROWSER_RE.test(code))
+    ) {
+      walk.ancestor(ast.program, processVisitor, {
```
garthenweb: Not sure how the performance is impacted by that change compared to `simple`, but this was the only way I found to do the `isAssignmentExpression` check.

packages/transformers/js/src/visitors/process.js
```diff
@@ -16,6 +16,17 @@ export default {
           // asset.meta.env[key.value] = process.env[key.value];
         }
       }
+    } else if (types.matchesPattern(node, 'process.browser')) {
+      // Inline process.browser
+      const assignmentAncestor = ancestors.find(ancestor => {
```
garthenweb: Is there a better way of doing this? It looks a bit wired to me, but I didn't find a better solution.

packages/transformers/js/src/JSTransformer.js
mischnic: The only difference is that it updates an array with `ancestors.push(...)` and `.pop()` before/after calling the visitor.

packages/transformers/js/src/visitors/process.js
mischnic: The list of ancestors should be sorted, so the last element could be the assignment expression.

garthenweb: So, the following should work as well?

```js
const lastAncestor = ancestors[ancestors.length - 1];
const isAssignmentExpression = Boolean(
  lastAncestor &&
    types.isAssignmentExpression(lastAncestor) &&
    lastAncestor.left === node
);
```

Unfortunately, when doing this the tests are failing. Can it be that it is sorted within the program, but not relative to the MemberExpression?

garthenweb: So, the following should work as well?

```js
const lastAncestor = ancestors[ancestors.length - 1];
const isAssignmentExpression = Boolean(
  lastAncestor &&
    types.isAssignmentExpression(lastAncestor) &&
    lastAncestor.left === node
);
```

Unfortunately, when doing this the tests are failing. Can it be that it is sorted within the program, but not relative to the MemberExpression?

packages/transformers/js/src/JSTransformer.js
garthenweb: Cool, so I guess this shouldn't have a big impact... :)

packages/transformers/js/src/visitors/process.js
mischnic: `ancestors` seems to include the current node (try [here](https://astexplorer.net/#/gist/65c6dd4d9da0841ec7b79bafa28c3f92/4071d6f7523de5f27df15c51b9fd5e07d7e8205c), open the console), so you need `- 2`

garthenweb: Ah, thanks for checking, that makes sense! I updated the code...

mischnic:
> The process.browser = true assignment

Why would you set that? If a bundler doesn't support it and/or without a bundler?

garthenweb:
> > The process.browser = true assignment
> 
> Why would you set that? If a bundler doesn't support it and/or without a bundler?

When not in a browser, I would not expect this line to be in the code at all, as it is only inserted when polyfilling node process: https://github.com/defunctzombie/node-process/blob/master/browser.js#L156
By default in other environments, this should be undefined. But you could just add it by hand. In case we want to transform all usages, this change would be the most consistent.

I can bearly find use cases for this, so it is really hard to decide on that. Maybe this could harm in a test environment where you want to mock browser behavior?
Maybe it is better to not touch it at all in none browser environments, what do you think?

mischnic:
Not sure if we should replace it on node as well. It isn't a "standardized" property on `process` and purely  added by bundlers / the polyfill you linked.

> But you could just add it by hand.

Is that common (where the value differs from `node-process/browser.js`)? I hope not.

<br>

`process.browser` should be true for `context: electron-renderer` and false for `context: electron-main`. Here the definitions for `env.isNode()` and `env.isBrowser()` https://github.com/parcel-bundler/parcel/blob/7cf6a3ba6ba9764d4881fee30632c305c9be7724/packages/core/core/src/public/Environment.js#L15-L24

garthenweb:
Okay, I agree with not touching it on NodeJS and updated the PR.
Is there any way I can activate the tests for NodeJS? At the moment `asset.env.isNode()` always returns `false`. I tried to configure the environment but without success.

I guess electron is handled with `asset.env.isNode()` as well? I am a bit unsure now how to configure the tests so that they make sense and pass 😕  Would be happy for an advice!

mischnic:
> I am a bit unsure now how to configure the tests

Either adding a `integration-tests/test/integration/process/package.json` with:
```
{
	"main": "dist/node.js",
	"browser": "dist/browser.js",
	"electronMain": "dist/electron.js",
	"targets": {
		"electronMain": {
			"context": "electron-main"
		},
		"electronRenderer": {
			"context": "electron-renderer"
		}
	}
}
```
or passing this as the second argument to `bundle` (though not 100% sure if that works):
`{ targets: { main: { context: "node"}}` (or `default` instead of `main`)

garthenweb:
Alright, tests are in place now. Unfortunately, `electron-main` and `electron-renderer` are no supported context targets, it throws here: https://github.com/parcel-bundler/parcel/blob/7cf6a3ba6ba9764d4881fee30632c305c9be7724/packages/core/test-utils/src/utils.js#L186

Is this expected or a bug? Based on the types there is a mismatch as far as I can see: https://github.com/parcel-bundler/parcel/blob/7cf6a3ba6ba9764d4881fee30632c305c9be7724/packages/core/types/index.js#L76

mischnic:
Yeah, that was never updated from Parcel 1. It should be:

```js
    case 'browser': 
		...
    case 'node':
	case 'electron-main':
		....
    case 'electron-renderer':
		...
    default:
      throw new Error('Unknown target ' + target);
```

Could you please update that as well?

packages/transformers/js/src/JSTransformer.js
```diff
@@ -74,9 +76,16 @@ export default new Transformer({
     let ast = asset.ast;
     let code = await asset.getCode();
 
-    // Inline environment variables
-    if (!asset.env.isNode() && (ast.isDirty || ENV_RE.test(code))) {
-      walk.simple(ast.program, envVisitor, {asset, env: options.env});
+    // Inline process/ environment variables
+    if (
+      !asset.env.isNode() &&
```
mischnic: It still wouldn't run on `electron-renderer` because of this `isNode`.

garthenweb: Good catch! The tests also complained after the splitting 👍

garthenweb:
Okay, I hope that everything works now.
I fixed the test utils as advised and also reactivated the environment tests for electron. I assume the same behavior we see for the `process.browser` applies to `process.env`, let me know if I am wrong with this.

Thank you for your support on this :)

mischnic:
> same behavior we see for the process.browser applies to process.env

For `electron-renderer`, `process.env` should not be replaced though.

DeMoorJasper:
@garthenweb could you rebase this PR?

garthenweb:
Sure, will have a look on the weekend!

garthenweb:
@DeMoorJasper The branch is rebased and up to date now. Tests are passing. I squashed the commits as this was the easiest way to rebase.

Let me know if something is missing!

PR #3695, opened 2019-10-28
Cache asset ASTs and reuse in packagers, scope hoisting source maps

### LINKED ISSUE #1564: Add support for source maps with tree shaking
Source maps are currently unsupported with tree shaking. Because of the way we concatenate the ASTs, and the ordering required, the existing source map infrastructure was insufficient to implement this easily for the first release.

There are a couple ways we could make this work:

1. We could use the source map generated by babel in the concatenate transform and map back to the original source maps generated for each file.
2. We could attempt to share the original ASTs of each file into the cache so we wouldn't need to parse it again in the packager. These ASTs have the original source map info embedded in them as well, so when we concatenate the ASTs together we can generate a combined source map automatically as well via babel. This option will require some performance testing to see how slow it is to pass the whole AST back over the IPC from workers vs passing generated code.

devongovett:
This stores ASTs for each asset in the cache and reuses them when needed in packager plugins. Previously, the scope hoisting packager needed to re-parse every asset and re-generate at the end, and we didn't support source maps because it was hard to preserve and combine the source maps of many assets correctly.

Now the final `generate` step in transformation happens lazily: when `getCode()` is called in the packager. If `getAST()` is called, in case of scope hoisting, then generating only happens on the final bundle. Because we are combining ASTs in scope hoisting now rather than re-parsing, we get source maps "for free".

Closes #1564.

To do:

- [x] Cache the results of `getCode` when it generates from an AST
- [x] Figure out where to load the generate plugin relative from
- [x] Figure out how to compute asset size when we don't generate code
- [x] Expose dirty status from AST to plugins?
- [x] Clear code when AST is set?


parcel-benchmark:
## Benchmark Results

#### Timings

| --- | --- | --- |

#### Cold Bundles

#### Cached Bundles

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 15.24s | -921.00ms 🚀 |
| Cached | 9.57s | -69.00ms |

#### Cold Bundles

| dist/main/index.js | 34.53kb | +0.00b | 2.61s | +134.00ms ⚠️ |
| dist/module/index.js | 34.33kb | +0.00b | 2.60s | +133.00ms ⚠️ |

#### Cached Bundles

| dist/module/index.js | 34.33kb | +0.00b | 11.00ms | +1.00ms ⚠️ |

</p></details>

**[Click here to view a detailed benchmark overview.](https://benchmark.parceljs.org/benchmark/247426908818506249)**

packages/core/fs/src/MemoryFS.js
```diff
@@ -166,7 +166,12 @@ export class MemoryFS implements FileSystem {
   }
 
   readFile(filePath: FilePath, encoding?: buffer$Encoding): Promise<any> {
-    return Promise.resolve(this.readFileSync(filePath, encoding));
+    try {
+      let res = this.readFileSync(filePath, encoding);
+      return Promise.resolve(res);
+    } catch (err) {
+      return Promise.reject(err);
+    }
```
mischnic: Or simply:
```js
async readFile(...){
	return this.readFileSync(...);
}
```

packages/core/source-map/src/SourceMap.js
```diff
@@ -393,19 +393,23 @@ export default class SourceMap {
...
+      for (let sourceName in sources) {
         let sourceContent = this.sourceContentFor(sourceName);
-        if (sourceContent !== null) {
+        if (sourceContent != null) {
           generator.setSourceContent(sourceName, sourceContent);
         } else {
           try {
             let content = await nullthrows(fs).readFile(
-              path.join(rootDir || '', sourceName),
+              path.resolve(rootDir || '', sourceName),
```
mischnic: The problem with `path.resolve` is that is uses the native node fs for the cwd.

packages/shared/scope-hoisting/src/generate.js
```diff
@@ -1,44 +1,78 @@
...
+  ExpressionStatement,
+>('(function () { STATEMENTS; })()');
 
+export function generate(
+  bundleGraph: BundleGraph,
+  bundle: Bundle,
+  ast: File,
+  options: PluginOptions,
+) {
   // $FlowFixMe
```
mischnic: We can actually remove the fixme by wrapping `...meta.interpreter` in `assertString` (from `./utils`)

packages/shared/scope-hoisting/src/generate.js
```diff
@@ -1,44 +1,78 @@
...
+      }),
+    ];
+  } else if (bundle.env.outputFormat === 'global') {
+    statements = [WRAPPER_TEMPLATE({STATEMENTS: statements})];
   }
 
-  return {contents};
+  ast = t.file(
+    t.program(
+      // $FlowFixMe Statement is incompatible with BabelNodeStatement
```
mischnic: That fixme doesn't actually seem to be needed?

packages/transformers/js/src/visitors/fs.js
```diff
@@ -101,9 +83,29 @@ export default ({
           throw err;
         }
       }
+
```
mischnic: I think adding `return` after `logger.warn(errorToDiagnostic(err));` would fix CI.

packages/shared/scope-hoisting/src/generate.js
wbinnssmith: It is 😞

mischnic: Well, I can remove the fixme and flow doesn't complain.
(Statement and BabelNodeStatement are two `type`s to the same sum type).

wbinnssmith: Strange — you're right. I remember this being an error before. @padmaia can attest to this I think 😅

mischnic: I also remember flow warning about unnecessary fixme's,  but not anymore...

wbinnssmith:
This is now ready for review 🎉 

cc @devongovett @mischnic

mischnic:
@wbinnssmith Have you tested/improved the performance of this? Is it actually faster now?

wbinnssmith:
It's not faster, but it's not slower. It was about 30% slower in the stress test until I added conditional generation at the end of transformation if it's not js and not scopehoisting.

packages/transformers/js/src/JSTransformer.js
```diff
@@ -124,76 +129,82 @@ export default new Transformer({
...
+      code: generated.code,
+      map:
+        generated.rawMappings != null
+          ? new SourceMap(generated.rawMappings, {
+              [sourceFileName]: null,
+            })
+          : null,
+    };
+
+    res.code = generateGlobals(asset) + res.code;
```
mischnic: So the globals are only inserted when code is generated? I guess this should rather be done at the end of transform?

packages/transformers/react-refresh-wrap/src/ReactRefreshWrapTransformer.js
```diff
@@ -94,45 +93,44 @@ export default new Transformer({
     return [asset];
   },
 
-  async generate({asset, options}) {
-    let code = await asset.getCode();
+  generate({asset, ast, options}) {
```
mischnic: ~As I understand, the code in `Transformation.js#runTransformer` uses the last transformer with a generate function, so this is actually not needed?~

But that would require this transformer to be run after a transformer that does contain the parse/generate code for Babel.
I guess the only solution would be a shared `@parcel/ast-babel` as an abstraction...

packages/packagers/js/src/JSPackager.js
```diff
@@ -60,16 +61,17 @@ export default new Packager({
...
-    let codeQueue = new PromiseQueue({maxConcurrent: 32});
-    let mapQueue = new PromiseQueue({maxConcurrent: 32});
+    let queue = new PromiseQueue({maxConcurrent: 32});
     bundle.traverse(node => {
       if (node.type === 'asset') {
-        codeQueue.add(() => node.value.getCode());
-        mapQueue.add(() => node.value.getMap());
+        queue.add(async () => ({
+          code: await node.value.getCode(),
+          map: await node.value.getMap(),
```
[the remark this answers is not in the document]
wbinnssmith: Yeah this stuck out to me too, but if either one doesn't exist, it causes a generate from the AST, which I imagine would happen in both. cc @devongovett

packages/transformers/js/src/JSTransformer.js
wbinnssmith: I think this may be to avoid dirtying the AST to save on performance if we're not scope hoisting — we can just reuse the existing code and map. @devongovett?

packages/transformers/react-refresh-wrap/src/ReactRefreshWrapTransformer.js
wbinnssmith: Yeah I think Babel parse+generate code is duplicated here and across the JS and Babel transformers. A shared library would make sense.

packages/shared/scope-hoisting/src/concat.js
```diff
@@ -123,8 +123,14 @@ export async function concat(bundle: Bundle, bundleGraph: BundleGraph) {
...
 
 async function processAsset(bundle: Bundle, asset: Asset) {
-  let code = await asset.getCode();
-  let statements: Array<Statement> = parse(code, asset.filePath);
+  let statements: Array<Statement>;
+  if (asset.astGenerator && asset.astGenerator.type === 'babel') {
+    let ast = await asset.getAST();
+    statements = nullthrows(ast).program.program.body;
+  } else {
+    let code = await asset.getCode();
```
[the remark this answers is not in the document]
wbinnssmith: I took a look at `@babel/parser` and there doesn't seem to be an option to pass along an existing sourcemap when parsing. Does Babel read and use a sourcemap comment when parsing?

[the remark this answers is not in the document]
mischnic: Apparently only `require("@babel/core").transform*` parses the comment (which wouldn't really work here either).
If we indeed use swc for env, this would be rather important.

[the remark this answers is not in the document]
devongovett: For now, I'm ok with merging as is. In order to do this properly, we'd need to parse the source map and code, traverse all AST nodes, and map their source location through the source map. At the moment, this case should never happen unless you replace the JSTransformer or put a transformer after it that returns a string instead of an AST. I think that should be fairly rare, so we can address this later if it comes up.

packages/transformers/js/src/JSTransformer.js
devongovett: Hmm, does this currently mean that if you have globals but the AST is never dirtied, they are never inserted?

wbinnssmith:
I'm going to mark this final to-do as complete as it seems we've decided to use sourcemap information to report on asset size when bundling completes. 

We also use asset size as a heuristic for extracting common bundles during bundling. For tree-shaken builds, this seems difficult to do accurately. See #4368.

packages/transformers/js/src/JSTransformer.js
wbinnssmith: Good point. Maybe we should attach the globals as AST nodes during transformation if we're scope hoisting (although I'd prefer not to rely on that), and otherwise continue to add them here in `generate`?

As I understand, we do this to avoid dirtying the AST to prevent generation. The simplest solution would just to prepend the globals during transform as @mischnic suggests. I don't expect most modules to require node globals, so maybe this is worth the perf cost. What do you think @devongovett?

wbinnssmith: Inserting the globals as a string here also probably offsets the sourcemap.

packages/core/core/src/InternalAsset.js
```diff
@@ -123,114 +146,204 @@ export default class InternalAsset {
...
-        contentStream.pipe(
-          new TapStream(buf => {
-            size += buf.length;
-            hash.update(buf);
-          }),
-        ),
-      ),
-      this.mapBuffer == null
+    await Promise.all([
+      contentKey == null
```
mischnic: `Promise.all([contentKey !== null && this.options.cache..., ...])`

packages/transformers/babel/src/BabelTransformer.js
```diff
@@ -49,31 +44,7 @@ export default new Transformer({
...
-
-      return {
-        code: generated.code,
-        map,
-      };
-    } catch (e) {
-      throw await babelErrorEnhancer(e, asset);
-    }
+  generate({asset, ast, options}) {
+    return generate({asset, ast, options});
```
mischnic: `class BabelTransformer extends Transformer mixin BabelAST` 😄

packages/transformers/babel/src/babel7.js
```diff
@@ -34,6 +37,7 @@ export default async function babel7(
     configFile: false,
     parserOpts: {
       ...babelOptions.config.parserOpts,
+      sourceFilename,
```
mischnic: Nit: inline variable?

packages/core/core/src/InternalAsset.js
```diff
@@ -241,6 +354,33 @@ export default class InternalAsset {
     }
   }
 
+  async getAST(): Promise<?AST> {
+    if (this.value.astKey != null) {
```
mischnic: So if you call getAST multiple times, this would repeatedly deserialize it from cache?

packages/transformers/js/src/JSTransformer.js
```diff
@@ -75,29 +68,31 @@ export default new Transformer({
     }
 
     asset.type = 'js';
-    if (!asset.ast) {
+    let ast = await asset.getAST();
+    if (!ast) {
```
mischnic: - Generally, when would getAST return a nullish value? Should this be an assertion?
- We should probably add the checks that are in `parse` here well?

packages/core/core/src/InternalAsset.js
wbinnssmith: Yeah, this is true of `getMap` as well. We should address this.

packages/transformers/js/src/JSTransformer.js
wbinnssmith: The Babel Transformer itself doesn't parse at all, instead using the ast if it's available from a prior transformer and sending the code as a string through babel [0]

[0] https://github.com/parcel-bundler/parcel/blob/e75bad2ff98eea4a02290bd74d71d1f77131e68e/packages/transformers/babel/src/babel7.js#L57

Issue #3708, opened 2019-10-30
Included files don't invalidate the cache (Parcel 2)


victorandree:
**Note:** Updated the issue with the latest alpha release (3.2). Included files are now "watched" but still don't cause cache invalidation – demonstrated by the reproduction repository (unless it's a problem with the pug transformer specifically).

# 🐛 bug report

In Parcel 2 (2.0.0-alpha.3.2), it doesn't seem like files included by a transform as an asset's `includedFiles` are considered for cache invalidation ~~or rebuilding (when watching).~~

This can be demonstrated by e.g. using the `include` feature of Pug, where the v2 transformer will loop over includes and do [`asset.addIncludedFile({filePath})`](https://github.com/parcel-bundler/parcel/blob/v2/packages/transformers/pug/src/PugTransformer.js#L36)

I initially reported this as a comment on the call to test Parcel 2, but decided to create a separate issue because I can't see it tracked already. Apologies if I've missed it.

## 🎛 Configuration (.babelrc, package.json, cli command)

I have a reproduction repository at https://github.com/victorandree/test-parcel2-includes where master is using v2, and a branch [parcel@v1](https://github.com/victorandree/test-parcel2-includes/tree/parcel%40v1) uses Parcel 1.

## 🤔 Expected Behavior

Since included files are merged into the asset, changes made to them should probably invalidate the asset cache ~~and trigger rebuilds~~.

## 😯 Current Behavior

~~In Parcel 2, when running `npx parcel watch`, changes made to an included file don't result in the assets being rebuilt.~~ When running `npx parcel build`, changes made to an included file aren't reflected (because of caching).

## 💁 Possible Solution

Treat included files as dependencies for the above purposes.

## 🔦 Context

I've been trying to write a transform for Parcel 2, where it's not appropriate to treat includes as dependencies – because the includes aren't assets on their own. I noticed that changes made to these includes had no effect on my main asset.

## 💻 Code Sample

I have a reproduction repository at https://github.com/victorandree/test-parcel2-includes where master is using v2, and a branch [parcel@v1](https://github.com/victorandree/test-parcel2-includes/tree/parcel%40v1) uses Parcel 1.

## 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Parcel           | 2.0.0-alpha.3.2
| Node             | 12.13.0
| npm/Yarn         | 6.12.0
| Operating System | macOS 10.14.16

devongovett:
[...]

PR #3717, opened 2019-10-31
[T-184] Refactor RequestGraph and add RequestTracker class

padmaia:
Happy Halloween! 🎃 To celebrate, I'm opening a big spooky refactor PR 👻

This PR is an overhaul of the `RequestGraph`. It used to be very coupled to how the `AssetGraph` is built, but I've made it more generic so that it will be easier to add more types of requests in the future. This is not quite finished with regards to my long term vision, but should be complete enough to merge once reviewed and approved. I will do my best to detail what's changed and what should be done in follow up PRs.

### What's changed?

- New `RequestTracker`

  In addition to the `RequestGraph` there is now a `RequestTracker` class. All requests that we wish to be "tracked" will run through the `RequestTracker`, which will take care of adding corresponding request nodes to the `RequestGraph`, executing the request, and adding the result and invalidations to the graph. If it is an existing request that has not been invalidated, then the result stored in the graph will be returned. The `RequestTracker` is constructed with a map of request "runners", which are responsible for actually running the requests and adding the correct invalidations. Each runner handles a specific type of request and has two methods, `run` and `onComplete`. The `RequestTracker` accepts an `AbortSignal` when running requests, so first it calls `runner.run()`, then checks to see if the signal is aborted, and only continues to `runner.onComplete()` if it isn't. This makes it so we won't have to thread the abort signal thoughout all of our async requests (side note: this is what got me started on this path in the first place). Having this central place that all requests travel through makes it easy to add middleware-ish behavior like this, which has already come in handy for debugging. I also find the new request runner pattern organizes the code nicely and will hopefully make it easier to add useful unit tests (TODO).

- No more `loadConfig` reverse handle

  Originally I was thinking that it would be helpful to use the `RequestGraph` for config loading, which would mean that it would have to take place in the main process, but since we can't assume that filepaths won't influence config, there ends up being a config request for every file. There's not much benefit to having access to the graph and what we've found is that workers end up sitting idle, waiting for the main thread to finish loading config when they could be loading those configs in parallel. There's also the cost of `require`ing all of the babel plugins during `loadPartialConfig` that won't even be doing any work in the main process. I expect this will help the benchmark, but I haven't run it locally yet. This required a change to the Transformer plugin API, which may end up applying to other plugins. I came up with `preSerializeConfig` and `postSerializeConfig`, but I'm definitely open to suggestions.

- No more queue

  The RequestGraph/RequestTracker are now really only responsible for running and invalidating requests and keeping track of the state and relationships of those requests. They do not control when requests are run. I thought about moving the queue to `AssetGraphBuilder`, but I wanted to experiment with just keeping track of incomplete nodes in the asset graph and using a while loop to keep processing until the graph is complete. I'm not sure I like it better. It seems easier to wind up in infinite loops and it would probably result in less parallel work. I was thinking that working with the queue gets kind of tricky when you want to work in distinct phases, but I'm not sure it has to be. For instance, I originally wanted to process all invalid nodes before moving on to incomplete nodes because they might end up getting rid of those incomplete nodes. So if an invalid request resulted in a new (incomplete) request being added, I didn't want it immediately added to the queue. I found the logic around whether or not it gets added to the queue not very clear and self evident. Thinking about it more it seems like it might be okay to just add new requests found during the invalidation phase straight to the queue, which simplifies things. I will probably end up experimenting with the queue again, but didn't think it was super important for the PR.

* Prioritization of processing invalid Nodes

  Invalid nodes are now processed in batches. Entry requests are rerun first as they may remove target requests. Target requests are next, then dependency path requests, and then asset requests. Dependency path requests and asset requests could maybe go at the same time though, so if it's deemed important we could make that optimization.

### Future Work

- `runner.onError()` method

  I could see value in adding another method to runners when `runner.run()` rejects (possibly `onError` or `onReject`) so that runners could decide what to do in that situation. A good example is that an asset request may fail while processing an included file. The transformation failed, but we know about the included file, so it would be nice to be able to add a file invalidation for it and rerun once it changes. Another example is when dependency path requests fail, the request should be invalidated whenever a file is added (maybe with a more specific glob).

- `runner.onRemove()` method

  Another runner method I see potential in is `onRemove` when a request no longer needs to be tracked. This could be really helpful in cleaning up no longer referenced cache entries like asset content and package results.

- `RequestTrackerAPI`

  Right now the `RequestGraph` gets passed to the runners in `onComplete`, but I've also been thinking that it would be nicer to send in higher level api instead. It would allow you to set up invalidations, but it would also give you a `runRequest` method that automatically connects calls to it as sub requests. `Transformation` has some logic that is aggregating all of the config requests it makes, which then gets sent back to the `AssetRequestRunner.onComplete` method, which is long because it's manually handling the sub request nodes. It would be really nice to just call `requestTracker.runRequest('config_request', {...})` inside of the `AssetRequestRunner` and have the `RequestTracker` take care of the rest.

- More Request Runners

  Once the `RequestTrackerAPI` is in place, it would be worth it to implement a `ConfigRequestRunner` and a `DepVersionRequestRunner`. I'm also thinking the `AssetGraphBuilder` could turn into `AssetGraphRequestRunner`. It's kind of weird right now that the `AssetGraphBuilder` is still responsible for persisting both the `RequestGraph` and `AssetGraph` separately. Eventually I dont think the `RequestGraph` will be exposed outside of the `RequestTracker` at all. The `AssetGraph` would just be a result on `RequestGraph` node. This would help with the concern of atomicity while persisting the coupled `AssetGraph` and `RequestGraph`, though I'm not sure if writing them separately was done intentionally for performance reasons or something. Then request runners can expand beyond building out the `AssetGraph`.

- Send valid sub requests along with `runTransformation`

  One minor regression that was introduced with these changes is that if a transformation gets invalidated for any reason, all of the subrequests will run from scratch again. Previously only invalidated subrequests would run again and the valid subrequests would return the result from the graph. This happened because we are no longer using the reverse handle to load config in the main process so we don't have access to the graph. My plan is to send the transformation request subgraph to the worker so that it can be used as if it were in the main process and then we can replace the subgraph when it is returned from the transformation. It would also be worth investigating whether this is even worth it. It's possible it's just as fast to just run all of the subrequests again, but we'll see.

- Maybe add timings for performance measurement to `RequestTracker`?

### TODO for current PR

- [x] Make flow happy
- [x] Reenable validation
- [x] Add file invalidations for transformed files (https://github.com/parcel-bundler/parcel/issues/3708)


packages/core/core/src/AssetGraphBuilder.js
```diff
@@ -58,30 +74,43 @@ export default class AssetGraphBuilder extends EventEmitter {
       entries
     });
 
+    this.handle = workerFarm.createReverseHandle(() => {
```
padmaia: I discovered a bug in the WorkerFarm. When Handles that are created in a worker are deserialized in the main process, what gets returned is actually a handle in the main process (https://github.com/parcel-bundler/parcel/blob/v2/packages/core/workers/src/Handle.js#L53). This problem was hidden before because there were 2 handles in the main process and 2 handles in the worker process. When I removed the `loadConfig` reverse handle I started getting the error `Error: Corresponding Handle was not found. It may have been disposed.` even though it was not being disposed. It was just being looked up where it didn't exist.

parcel-benchmark:
## Benchmark Results

#### Timings

| --- | --- | --- |

#### Cold Bundles

#### Cached Bundles

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 14.95s | -649.00ms |
| Cached | 11.93s | +2.49s ⚠️ |

#### Cold Bundles

*No bundle changes detected.*

#### Cached Bundles

| dist/main/index.js | 34.53kb | +0.00b | 18.00ms | +4.00ms ⚠️ |
| dist/module/index.js | 34.33kb | +0.00b | 15.00ms | +5.00ms ⚠️ |

</p></details>

**[Click here to view a detailed benchmark overview.](https://benchmark.parceljs.org/benchmark/247765923176382986)**

packages/core/core/src/AssetGraphBuilder.js
```diff
@@ -58,30 +74,43 @@ export default class AssetGraphBuilder extends EventEmitter {
...
+      ['target_request', new TargetRequestRunner({options, assetGraph})],
+      [
+        'asset_request',
+        new AssetRequestRunner({options, workerFarm, assetGraph})
+      ],
+      [
+        'dep_path_request',
+        new DepPathRequestRunner({options, config, assetGraph})
+      ]
+      // ['config_request', new ConfigRequestRunner({options})],
```
wbinnssmith: debug 😄

packages/core/core/src/Parcel.js
```diff
@@ -238,6 +238,7 @@ export default class Parcel {
     signal?: AbortSignal,
     startTime?: number
   |}): Promise<BuildEvent> {
+    //console.log('BUILDING');
```
wbinnssmith: debug 😄

packages/core/core/src/ReporterRunner.js
```diff
@@ -41,7 +41,7 @@ export default class ReporterRunner {
     });
 
     if (this.options.patchConsole) {
-      patchConsole();
+      //patchConsole();
```
wbinnssmith: we need to fix this 😢

packages/core/core/src/AssetGraphBuilder.js
```diff
@@ -90,11 +119,45 @@ export default class AssetGraphBuilder extends EventEmitter {
...
-  async build(): Promise<{|
+  async build(
+    signal?: AbortSignal
+  ): Promise<{|
     assetGraph: AssetGraph,
     changedAssets: Map<string, Asset>
   |}> {
-    await this.requestGraph.completeRequests();
+    // TODO: optimize prioritized running of invalid nodes
+    let requestPriority = [
```
wbinnssmith: could this be a module-level `const`?

packages/core/core/src/AssetGraphBuilder.js
```diff
@@ -90,11 +119,45 @@ export default class AssetGraphBuilder extends EventEmitter {
...
+          );
+        }
+      }
+      await Promise.all(promises);
+    }
+
+    while (this.assetGraph.hasIncompleteNodes()) {
+      let promises = [];
+      for (let id of this.assetGraph.incompleteNodeIds) {
+        let node = nullthrows(this.assetGraph.getNode(id));
```
wbinnssmith: this could be inlined below

packages/core/core/src/ReporterRunner.js
DeMoorJasper: Isn't the option patchConsole: false working?

packages/core/core/src/Transformation.js
```diff
@@ -8,23 +8,22 @@ import type {
...
   ParcelOptions
 } from './types';
-import type {WorkerApi} from '@parcel/workers';
 
 import invariant from 'assert';
 import path from 'path';
 import nullthrows from 'nullthrows';
 import {md5FromObject} from '@parcel/utils';
 
+import ConfigLoader from './ConfigLoader';
```
kwelch: I like this name a log better than `Config`

packages/core/core/src/Transformation.js
```diff
@@ -98,6 +97,15 @@ export default class Transformation {
     let results = await this.runPipeline(pipeline, asset);
     let assets = results.map(a => a.value);
 
+    for (let {request, result} of this.configRequests) {
+      //console.log('RESULT BEFORE', result);
+      let plugin =
+        request.plugin != null &&
+        (await this.parcelConfig.loadPlugin(request.plugin));
```
kwelch: To clarify, we are going to continue with `loadPlugin`, but we are moving away from rehydrate and reload, in a sense. Because we are going to serialize and deserialize it.

packages/transformers/babel/src/config.js
```diff
@@ -196,18 +217,23 @@ function isLocal(/* configItemPath */) {
   return false;
 }
 
-function prepForReyhdration(options) {
+export function preSerialize(config: Config) {
```
kwelch: Nice! I really like this, it feel cleaner.

packages/core/core/src/Transformation.js
padmaia: Well... `Config` still exists. `ConfigLoader` is responsible for running plugins' `loadConfig` which takes a `Config` and uses it to store the result and all of the files that should invalidate that `loadConfig` call. I'm not super happy with the naming, but you know what they say.

packages/core/core/src/Transformation.js
padmaia: Yeah, so before the config was being loaded first in the main process and then sent over to the worker, so plugins `loadConfig` would return a serialized result. This wouldn't be efficient for plugins that are run in the main process and it's not efficient or necessary now that we are are able to call `loadConfig` in workers. So now `loadConfig` returns a result that can directly be used by the plugin and Parcel will call `preSerializeConfig` and `postSerializeConfig` when appropriate.

packages/core/core/src/Transformation.js
```diff
@@ -98,6 +97,15 @@ export default class Transformation {
     let results = await this.runPipeline(pipeline, asset);
     let assets = results.map(a => a.value);
 
+    for (let {request, result} of this.configRequests) {
+      //console.log('RESULT BEFORE', result);
```
devongovett: Debug

packages/core/core/src/RequestTracker.js
```diff
@@ -0,0 +1,427 @@
...
+    throw new Error(
+      `RequestRunner for type ${this.type} did not implement a onComplete()`
+    );
+  }
+
+  generateRequestId(request: TRequest) {
+    return md5FromObject({type: this.type, request});
+  }
+
+  createAPI(requestId: string): RequestRunnerAPI {
```
devongovett: Is the idea here to hide the tracker from these runner implementations? Right now the runners are subclasses, so they’d technically still have access to the tracker directly.

packages/core/core/src/requests.js
```diff
@@ -0,0 +1,377 @@
...
+|};
+
+type DepPathRequest = {|
+  id: string,
+  +type: 'dep_path_request',
+  request: Dependency,
+  result?: AssetRequestDesc
+|};
+
+export class EntryRequestRunner extends RequestRunner<FilePath, EntryResult> {
```
devongovett: We could probably extract these into their own files at some point...

packages/core/core/src/RequestTracker.js
```diff
@@ -0,0 +1,427 @@
...
+  value: glob
+});
+
+const nodeFromRequest = (request: Request) => ({
+  id: request.id,
+  type: 'request',
+  value: request
+});
+
+export class RequestGraph extends Graph<
```
devongovett: Should this be in its own file?

packages/core/core/src/RequestTracker.js
```diff
@@ -0,0 +1,427 @@
...
+  invalidateOnFileUpdate: FilePath => void,
+  invalidateOnStartup: () => void,
+  replaceSubrequests: (Array<RequestGraphNode>) => void
+|};
+
+export function generateRequestId(type: string, request: mixed) {
+  return md5FromObject({type, request});
+}
+
+export class RequestRunner<TRequest, TResult> {
```
devongovett: Should this be in a separate file?

packages/core/core/src/requests.js
padmaia: Yeah, I'm imagining the runners will get their own files, it was just easier to put all of this logic in one file while I was trying out different things. I can move them into different files once we start taking on more use cases than building the AssetGraph.

packages/core/core/src/RequestTracker.js
padmaia: Again, I was mostly keeping them together while I was trying things out. The `RequestTracker/RequestGraph/RequestRunner` are all pretty coupled though, so I could see value in keeping them together to understand quickly how they work together? I don't feel super strongly about that though.

packages/core/core/src/RequestTracker.js
padmaia: This API was a little more useful in my initial implementation, where the runners didn't have access to the tracker. It was also more useful when there was a map of runners passed into the Tracker, because then the API could have a `runRequest` method that hard codes the request id as the parent id for the sub request. That's not really an option now that you create  the runners with the tracker and just call them directly, but I thought the runner map idea was less composable, so I'm okay on losing out on that behavior. Right now the API's only usefulness is that you don't have to pass in the request id to the invalidation functions. Since we will need to pass the request ID in anyway to use as the parent ID for calling sub requests, I may end up getting rid of it.

PR #3722, opened 2019-11-01
[T-185] Codeframe tabs support

DeMoorJasper:
# ↪️ Pull Request

Adds tabs support in codeframes.

Also starts the catching of errors in plugins to append the correct origin

Closes T-185

### DIFF
--- a/packages/core/codeframe/src/codeframe.js
+++ b/packages/core/codeframe/src/codeframe.js
@@ -46,6 +46,9 @@ export default function codeFrame(
 ): string {
   if (highlights.length < 1) return '';
 
+  // Replace all tabs with a space to simplify charcount
+  code = code.replace(/\t/g, ' ');
+
   let opts: CodeFrameOptions = {
     useColor: !!inputOpts.useColor,
     syntaxHighlighting: !!inputOpts.syntaxHighlighting,
--- a/packages/core/codeframe/src/codeframe.js
+++ b/packages/core/codeframe/src/codeframe.js
@@ -25,6 +25,7 @@ type CodeFrameOptions = {|
 |};
 
 const NEWLINE = /\r\n|[\n\r\u2028\u2029]/;
+const TAB_REPLACE_REGEX = /\t/g;
 const TAB_REPLACEMENT = '  ';
 
 const highlightSyntax = (txt: string, lang?: string): string => {
@@ -120,7 +121,7 @@ export default function codeFrame(
     ? highlightSyntax(code, opts.language)
     : code
   )
-    .replace(/\t/g, TAB_REPLACEMENT)
+    .replace(TAB_REPLACE_REGEX, TAB_REPLACEMENT)
     .split(NEWLINE);
   for (let i = startLine; i < lines.length; i++) {
     if (i > endLine) break;
@@ -154,7 +155,9 @@ export default function codeFrame(
         // If there's a whole line highlight
         // don't even bother creating seperate highlight
         highlightLine += highlighter(
-          '^'.repeat(originalLine.replace(/\t/g, TAB_REPLACEMENT).length)
+          '^'.repeat(
+            originalLine.replace(TAB_REPLACE_REGEX, TAB_REPLACEMENT).length
+          )
         );
       } else {
         let sortedColumns =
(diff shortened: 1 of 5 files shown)


parcel-benchmark:
## Benchmark Results

#### Timings

| --- | --- | --- |

#### Cold Bundles

#### Cached Bundles

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 15.02s | +531.00ms |
| Cached | 9.71s | +15.00ms |

#### Cold Bundles

| dist/main/index.js | 34.53kb | +0.00b | 2.34s | +154.00ms ⚠️ |
| dist/module/index.js | 34.33kb | +0.00b | 2.34s | +155.00ms ⚠️ |

#### Cached Bundles

| dist/main/index.js | 34.53kb | +0.00b | 13.00ms | +2.00ms ⚠️ |
| dist/module/index.js | 34.33kb | +0.00b | 10.00ms | +1.00ms ⚠️ |

</p></details>

**[Click here to view a detailed benchmark overview.](https://benchmark.parceljs.org/benchmark/247840714812031497)**

packages/core/codeframe/src/codeframe.js
```diff
@@ -46,6 +46,9 @@ export default function codeFrame(
 ): string {
   if (highlights.length < 1) return '';
 
+  // Replace all tabs with a space to simplify charcount
+  code = code.replace(/\t/g, ' ');
```
mischnic: Could you add a comment here that tabs should be treated as a single space
 https://github.com/parcel-bundler/parcel/blob/5e38099af6d4cf53d5d6e411b11bbd706ffc7434/packages/core/diagnostic/src/diagnostic.js#L10-L15

Are you sure this is what babel does (ignore the wrong codehighligh locations for now)? The indentation is barely noticeable:

Bildschirmfoto 2019-11-01 um 15 25 41

DeMoorJasper: I'm not sure this is what babel does, except for the tab space at the start they don't really seem to do anything with tabs

DeMoorJasper: In any case we should count a tab as 1 character length as it can probably be different from machine to machine

mischnic: `@babel/cli` displays tabs with the width of two spaces, but it really is a tab character.
Printing a tab with  `printf "x\tx"` has a width of 8. 🤷‍♂

mischnic: > In any case we should count a tab as 1 character length as it can probably be different from machine to machine

But then the highlight column would have to be adjusted as well?

DeMoorJasper: I'll make the tab width variable (through a param or const in the codeframe-util no fancy editorconfig [yet])

But it will still be just length 1 for the highlights. I think it will be the most reliable way to do this as tabs don't really have a fixed length and are just 1 character after all

DeMoorJasper: I could also just keep tabs and not replace them with spaces if that seems better?

mischnic: > But a tab will still just count as 1 character for the highlights. I think it will be the most reliable way to do this as tabs don't really have a fixed length and are just 1 character after all

So the one who prints the codeframe is responsible for expanding these counts as necessary? 👍 

(I don't really care about the exact tab width, just that the source line and highlight arrows align 😄 )

DeMoorJasper: > But then the highlight column would have to be adjusted as well?

how come, in that test the `\t` still counts as a single character, they apparently just don't replace it with spaces

DeMoorJasper: > So the one who prints the codeframe is responsible for expanding these counts as necessary? 👍

Exactly, the codeframe util will have to figure it out, or whatever new formatters we come up with in the future

mischnic: I think this would work: count highlight column in chars (tab = 1), then when it is printed, then for the highlight line:
```js
sourceLine.splice(0, column).replace(/\s/g, " ")/*either keep tabs here or replace with 2 spaces*/
   + "^^^^^" + highlight.message
```

DeMoorJasper: Yeah seems like that's the only real way of doing it, ours is slightly more complicated though due to supporting multiple highlights

packages/core/codeframe/src/codeframe.js
```diff
@@ -170,13 +179,24 @@ export default function codeFrame(
 
           let whitespaceLength = startCol - lastCol;
```
mischnic: `whitespaceLength ` could be inlined into the if condition

packages/core/codeframe/src/codeframe.js
```diff
@@ -170,13 +179,24 @@ export default function codeFrame(
...
           if (whitespaceLength > 0) {
-            highlightLine += ' '.repeat(whitespaceLength);
+            let whitespace = originalLine
+              .substring(lastCol, startCol)
```
mischnic: Any particular reason why `lastCol` and `startCol` are swapped here (substring works fine, but it still looks kinda strange)?

PR #3726, opened 2019-11-01
Encoded inline bundles

### LINKED ISSUE #3718: RFC: Encoded Inline Bundles
## Problem

In Parcel, we currently have multiple cases where one bundle can either reference or include the content of another bundle:

* Inlining a child bundle's contents as text in the parent bundle (support for compiling inline `<script>` tags in html documents)
* Referencing a url (relative or from the target's `publicUrl`) of a child bundle, when importing e.g. css from js or using `new Worker('./workerEntry')`

There are cases, however, when it's desirable to inline contents of another bundle (perhaps also as a url that itself contains the contents of the child bundle, rather than a relative http reference):

* `import`ing small images as base64 `data:` uris in JS, css, html
* Passing `new Worker()` a base64 or blob-encoded uri instead of a relative http reference

## Proposal

Using named pipelines [0], create several named pipelines that correspond to how the bundle should be inlined, and use these when depending on the asset that creates this bundle, e.g.:

* `import encodedImageUrl from 'url-base64:./image.png'` (result will be a `data:$mime;$base64content` url)
* `new Worker('url-text:worker.js')` (text would be in the form of a blob url)

This allows us to implement this functionality outside of core, with the following changes to the default config:

```
{
  "transforms": {
    "url-base64:*": ["@parcel/transformer-inline", "..."],
    "url-text:*": ["@parcel/transformer-inline", "..."]
    // ...
  },
  "optimizers": {
    "url-base64:*": ["@parcel/optimizer-url-base64", "..."],
    "url-text:*": ["@parcel/optimizer-url-text", "..."]
    // ...
  }
}
```

**NOTE:** This will require us to support declaring optimizers for a named pipeline. This is currently only implemented for transformers. Currently only dependencies and asset requests contain the requested pipeline. This will require us to persist the name of the pipeline on the produced asset, as well as carry it over to any bundles it forms, allowing us to influence the optimizers that run.

The role of the `"@parcel/transformer-inline"` transform will be to simply set `asset.isInline` to `true`. 

The role of the optimizers will be to actually perform the translation to e.g. base64. These optimizers _will not_ produce output particular to any environment. For example, the `url-text` optimizer will not produce the JavaScript code necessary to construct a blob url of the text. Instead, this will be the responsibility of the including asset's transformer or the parent bundle's packager.

### Questions

* Should this all just be part of core? Support for inline assets already is (for the HTML use case) and I'm not immediately sure if support for inline assets can be entirely removed from core. Either way, it would be nice to either include complete support for inline assets in core, or not include it at all.
* `asset.isInline` could be an enum type of `'base64' | 'blob'`, but that would require a change to core, and this information is already encoded in the pipeline name.
* It's possible that the packager for `url-base64` could produce `data:$mime;$base64content` since like http urls, it's universal. That said, particular environments might require escaping of strings, etc. So maybe it's just best to let the packager do this.

[0] https://github.com/parcel-bundler/parcel/pull/3613/files#r333689194

wbinnssmith:
Resolves #3718.

This implements the ability to inline the content of one bundle into another by using protocol-style imports in the form of `data-url:./img.png` to receive a data url or `bundle-text:./worker.js` to receive the literal text of the compiled bundle content.

This is accomplished by adding `transformer-inline`, which sets `isInline` to true on assets, attaching the requested pipeline name to assets and bundles, and adding an optimizer to transform bundle contents into data urls. References to the inline bundle are left behind in the form of bundle ids (this may change with css), and the packaged inline bundles are then replaced during packaging by invoking the inline packaging api.

For data urls, the mimetype is automatically determined based on the bundle filepath, and the content is encoded either as text for non-binary content, or as base64 for binary content (content with a 0x00 byte present).

Question: Is there a more robust heuristic for determining binary content than checking for the presence of a 0x00 byte?

Remaining tasks:

- [x] Add support for url dependencies
- [x] Add support in the css packager
- [ ] Adjust source maps to reflect inline content

Test Plan: Added integration tests.


wbinnssmith:
Re: detecting binary files, https://www.npmjs.com/package/isbinaryfile has an extra step counts characters it considers "weird". We could use that module here instead.

parcel-benchmark:
## Benchmark Results

#### Timings

| --- | --- | --- |

#### Cold Bundles

#### Cached Bundles

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 15.14s | -235.00ms |
| Cached | 9.81s | +90.00ms |

#### Cold Bundles

*No bundle changes detected.*

#### Cached Bundles

| dist/main/index.js | 34.53kb | +0.00b | 15.00ms | +3.00ms ⚠️ |
| dist/module/index.js | 34.33kb | +0.00b | 10.00ms | +1.00ms ⚠️ |

</p></details>

**[Click here to view a detailed benchmark overview.](https://benchmark.parceljs.org/benchmark/247872448875201038)**

packages/core/core/src/ResolverRunner.js
```diff
@@ -33,8 +35,49 @@ export default class ResolverRunner {
 
     let resolvers = await this.config.getResolvers();
 
+    let pipeline;
+    let filePath;
+    if (dependency.moduleSpecifier.includes(':')) {
+      [pipeline, filePath] = dependency.moduleSpecifier.split(':');
+      let transformsWithPipelines = {};
+      for (let key of Object.keys(this.config.transforms)) {
```
devongovett: Should we also include optimizers here? What if you had a named pipeline that only existed in the optimizer config?

packages/core/utils/src/string.js
```diff
@@ -0,0 +1,9 @@
+// @flow strict-local
+
+export function escapeJSStringLiteral(str: string): string {
```
devongovett: Maybe just use `JSON.stringify`?

packages/core/utils/src/replaceBundleReferences.js
```diff
@@ -0,0 +1,165 @@
...
+      let packagedBundle = await getInlineBundleContents(
+        entryBundle,
+        bundleGraph
+      );
+      let packagedContents = (packagedBundle.contents instanceof Readable
+        ? await bufferStream(packagedBundle.contents)
+        : packagedBundle.contents
+      ).toString();
+
+      let inlineType = nullthrows(entryBundle.getMainEntry()).meta.inlineType;
```
devongovett: Where do we define what `inlineType` values are allowed? I'm not sure it should be enforced in this util in case new ones need to be defined by plugins.

wbinnssmith: It's not really enforced. `inlineType` isn't a formal concept in core and is just something that transformers and packagers agree on.

packages/core/core/src/ResolverRunner.js
wbinnssmith: Good point. Think we should be using the config matching method or just extract the pipelines ourselves here and match against them?

packages/core/utils/src/replaceBundleReferences.js
devongovett: it's enforced by the line below asserting the values...

wbinnssmith: This function isn't run in core though, it's run in packagers. Perhaps this util shouldn't live in `core/utils`?

devongovett: Yeah true, but the util is still kinda “shared” code that shouldn’t make assumptions if it’s meant to be informal.

packages/core/core/src/ResolverRunner.js
devongovett: Perhaps we could extract the named pipelines once and cache them? Maybe in the ParcelConfig class?

PR #3730, opened 2019-11-03
Fix code splitting with interop shared bundle

### LINKED ISSUE #3686: v2: code splitting with `process` in shared module throws in child bundle
# 🐛 bug report

When using styled-components within a parent and child bundle, the application throws at runtime with `TypeError: c(...).div is not a function`.
It looks like, within the child bundle, styled-components refers to the named imports instead of the default import.

```json
{
  "scripts": {
    "start": "rm -rf .parcel-cache dist && parcel build index.html && serve dist"
  },
  "dependencies": {
    "react": "^16.11.0",
    "react-dom": "^16.11.0",
    "styled-components": "^4.4.0"
  },
  "devDependencies": {
    "parcel": "^2.0.0-alpha.2.1",
    "serve": "^11.2.0"
  }
}
```

## 🤔 Expected Behavior

It should allow importing styled-components in a parent and child bundle with the same syntax.

## 😯 Current Behavior

Currently, it fails. It would work when using `styled.default.div` in the child bundle.

## 💻 Code Sample

EDIT: See a smaller sample [here](https://github.com/parcel-bundler/parcel/issues/3686#issuecomment-549055131).

index.html

```html
<!DOCTYPE html>
<main></main>
```
index.js

```jsx
import React, { Suspense } from "react";
import styled from "styled-components";

const Page = React.lazy(() => import("./Page"));

const Section = styled.section`
  width: 100%;
`;

render(
  <Suspense fallback={null}>
    <Section />
    <Page />
  </Suspense>,
  document.querySelector("main")
);
```
page.js

```jsx
import React from "react";
import styled from "styled-components";

const Div = styled.div`
  width: 100%;
`;

const Page = () => <Div />;

export default Page;
```

## 🌍 Your Environment

| Parcel           | 2.0.0-alpha.2.1
| Node             | v12.8.1

garthenweb:
# ↪️ Pull Request

When a shared module, that is defined as a common js module, is used in a parent and child bundle, it is not resolved the same way in both bundles.
The `__esModule` is removed/ not set at all, therefore the interop check is wrong and reference errors are the result.

Fixes #3686.

When a `require` call is used, the asset is marked as `isCommonJS`: [hoist.js#L339](https://github.com/parcel-bundler/parcel/blob/v2/packages/shared/scope-hoisting/src/hoist.js#L339)
As far as I understand, due to scope hoisting, the `require` call is removed and therefore the import is detected as an unused value: [link.js#L341](https://github.com/parcel-bundler/parcel/blob/v2/packages/shared/scope-hoisting/src/link.js#L341)
Therefore the `__esModule` flag is not injected: [link.js#L361](https://github.com/parcel-bundler/parcel/blob/v2/packages/shared/scope-hoisting/src/link.js#L361)

Inserting the `__esModule` flag always, in case the `require `call was removed and the `isCommonJS` flag is not set, solves the issue for me.

## 💻 Examples

See https://github.com/parcel-bundler/parcel/issues/3686, https://github.com/parcel-bundler/parcel/issues/3686#issuecomment-549055131 or the tests added in the pr.

## ✔️ PR Todo

- [x] Fix second integration test


packages/shared/scope-hoisting/src/link.js
```diff
@@ -337,32 +337,28 @@ export function link({
...
+                if (binding.path.node.init) {
+                  binding.path
+                    .getStatementParent()
+                    .insertAfter(ESMODULE_TEMPLATE({EXPORTS: name}));
+                }
 
-              // Insert __esModule interop flag if the required module is an ES6 module with a default export.
-              // This ensures that code generated by Babel and other tools works properly.
-              if (
-                asset.meta.isCommonJS &&
```
garthenweb: Was there a reason why `isCommonJS` was checked here? From my point of view, it does not matter as long as it is an es6 module.

packages/shared/scope-hoisting/src/link.js
```diff
@@ -337,32 +337,28 @@ export function link({
           }
         } else {
           if (mod.meta.id && assets.get(assertString(mod.meta.id))) {
+            let name = assertString(mod.meta.exportsIdentifier);
             // Replace with nothing if the require call's result is not used.
             if (!isUnusedValue(path)) {
-              let name = assertString(mod.meta.exportsIdentifier);
               node = t.identifier(replacements.get(name) || name);
+            }
```
garthenweb: That we only run the code to insert `__esModule` if the `require` call is removed did not look related to me. Was there any reason for it?

garthenweb:
I found a solution for both test cases. It is also working within my project.

I would love to get feedback on it, especially as I am not super familiar with the code I touched and am not sure if the changes might have unwanted side effects.

packages/shared/scope-hoisting/src/link.js
mischnic: It was only inserted if the call was *not* removed (`mod` is truthy) and if the call is `!isUnused(...)`?

garthenweb: Exactly. Now it is always inserted in case `mod` is truthy, independent of what `!isUnused(...)` returns.

mischnic: For that, we should use `!isUnusedValue(path) || bundleGraph.isAssetReferencedByAssetType(mod, 'js')`, since the root issue is that `isUnusedValue` doesn't work across bundles. (`isAssetReferencedByAssetType` should really be renamed `isAssetReferencedByAnotherBundleOfType`).

garthenweb: Thanks for the feedback. I update the code as suggested, just running it for JS assets does make sense to me.

I can do the renaming in this PR as well if you like to :)  Just let me know.

garthenweb: I just pushed the renaming as all tests passed locally. Feel free to revert in case you don't want to have it in here or is not what you expected.

garthenweb:
Are there any plans to get this merged? Or is anything missing/ does it need further investigation?

mischnic:
> Are there any plans to get this merged?

Yes

> does it need further investigation?

I wanted to look into why `asset.meta.isCommonJS && mod.meta.isES6Module` isn't correct in every situation.

mischnic:
Finally got around to take a look
The problem is that the check in `link.js` didn't factor in that an asset can be exported and used in another bundle, doing this instead produces less code:


```diff
diff --git packages/shared/scope-hoisting/src/link.js packages/shared/scope-hoisting/src/link.js
index 0a5d295f..09bfa977 100644
--- packages/shared/scope-hoisting/src/link.js
+++ packages/shared/scope-hoisting/src/link.js
@@ -365,16 +365,21 @@ export function link({
         } else {
           if (mod.meta.id && assets.get(assertString(mod.meta.id))) {
             // Replace with nothing if the require call's result is not used.
-            if (
-              !isUnusedValue(path) ||
-              bundleGraph.isAssetReferencedByAnotherBundleOfType(mod, 'js')
-            ) {
+            let isReferenced = bundleGraph.isAssetReferencedByAnotherBundleOfType(
+              mod,
+              'js',
+            );
+            if (!isUnusedValue(path) || isReferenced) {
               let name = assertString(mod.meta.exportsIdentifier);
               node = t.identifier(replacements.get(name) || name);
 
               // Insert __esModule interop flag if the required module is an ES6 module with a default export.
               // This ensures that code generated by Babel and other tools works properly.
-              if (mod.meta.isES6Module && mod.symbols.has('default')) {
+              if (
+                (asset.meta.isCommonJS || isReferenced) &&
+                mod.meta.isES6Module &&
+                mod.symbols.has('default')
+              ) {
                 let binding = path.scope.getBinding(name);
                 if (binding && !binding.path.getData('hasESModuleFlag')) {
                   if (binding.path.node.init) {
```

garthenweb:
@mischnic Thanks for having a look! Your suggestion makes sense to me. I tried to implement them, did some small adjustments to make it more descriptive and added a todo comment. I hope that works for you!

PS: I would really love to see a working parcel v2 release ❤️

mischnic:
> I would really love to see a working parcel v2 release

Me too 😉

PR #3753, opened 2019-11-07
Support for peer dependencies in v2

subhero24:
As requested in #3674 I implemented support for peer dependencies in v2 with integration tests.
This changes the module resolution to discard packages that where specified as peer dependencies. It also excludes files from the bundle if all references to a package where specified as a peer dependency. 

This solves issues: 
#144 
#3305


packages/resolvers/default/src/DefaultResolver.js
```diff
@@ -179,6 +179,33 @@ class NodeResolver {
...
+    // Finds the first package that does not specify filename as a peer dependency
+    // and resolve from there. If there was no package, or all packages specified
+    // filename as a peer dependency, the module is marked as external
+    let pkg = await this.findPackage(dir);
+    let externalPeerDependency = false;
+    while (pkg) {
+      let isPeerDependency = false;
+      let hasPeerDependencies = pkg.peerDependencies != undefined;
+      if (hasPeerDependencies) {
+        isPeerDependency = Object.keys(pkg.peerDependencies).includes(filename);
```
mischnic:
```suggestion
        isPeerDependency = filename in pkg.peerDependencies
```
would probably be faster?

And `hasPeerDependencies` can be inlined above.
`externalPeerDependency` is always  set to `isPeerDependency`, so `isPeerDependency` isn't really need either.

packages/resolvers/default/src/DefaultResolver.js
```diff
@@ -179,6 +179,33 @@ class NodeResolver {
       return builtin;
     }
 
+    // Resolve peer dependencies
+    // Finds the first package that does not specify filename as a peer dependency
+    // and resolve from there. If there was no package, or all packages specified
+    // filename as a peer dependency, the module is marked as external
+    let pkg = await this.findPackage(dir);
```
mischnic: We should probably have a cache in front of `findPackage`?

subhero24: There is a cache in findPackage > readPackage, or do you want to prevent traveling up the directory hierarchy?

packages/resolvers/default/src/DefaultResolver.js
subhero24: Yes, that would indeed be faster :-) will update ...

mischnic:
So if I have `react-dom` as a dependency and have not installed `react`, this would throw something like `Cannot resolve 'react' from .../my-project`. And I wouldn't know that react-dom caused this.

We should print a more helpful error (and maybe hint at how to solve this: adding as a peerDependency or a dependency)

FreeFrags:
Whats the status of this issue? Seems like there are many issues which are all on the same subject. This looks like the latest conversation about it. it would be amazing if this gets supported.

lots0logs:
@subhero24 @mischnic 
bump

mischnic:
@subhero24 Can you/do you want to update this?

subhero24:
> So if I have `react-dom` as a dependency and have not installed `react`, this would throw something like `Cannot resolve 'react' from .../my-project`. And I wouldn't know that react-dom caused this.

Currently the error for the unfound peer dependency is `Cannot find module 'react' from '../my-project/node_modules/react-dom'` so the user knows it was caused by react-dom. Is that enough? or would you still want the  more helpful hint about adding it as a (peer)dependency? (isn't this hint about adding it as a (peer)dependency applicable to all `Cannot find module ...` errors though?)

DeMoorJasper:
#4072 implemented another way of doing this.

Gonna close this for now as it's probably no longer needed, feel free to re-open discussion on more excluding of modules improvements

PR #3796, opened 2019-11-17
[T-208] Resolver error tracking

DeMoorJasper:
# ↪️ Pull Request

Add error codeframes to resolver.

Screenshot 2019-11-23 at 18 45 35


packages/core/core/src/ResolverRunner.js
```diff
@@ -28,6 +29,35 @@ export default class ResolverRunner {
...
+  async getThrowableDiagnostic(dependency: Dependency, message: string) {
+    let diagnostic: Diagnostic = {
+      message,
+      origin: '@parcel/resolver-default'
+    };
+
+    if (dependency.loc && dependency.sourcePath) {
+      diagnostic.filePath = dependency.sourcePath;
+      diagnostic.codeFrame = {
+        code: await this.options.inputFS.readFile(
```
DeMoorJasper: How would I go about getting the asset source in the resolver?

Reading it from fileSystem would result in incorrect codeFrames... It has to be a snapshot from the moment the location data was extracted. Which is not possible atm afaik...

packages/core/core/src/ResolverRunner.js
```diff
@@ -105,11 +135,14 @@ export default class ResolverRunner {
     let dir = dependency.sourcePath
       ? path.dirname(dependency.sourcePath)
       : '<none>';
-    let err = new Error(
+
+    let err: any = await this.getThrowableDiagnostic(
```
DeMoorJasper: Can this error actually be generalised in the ResolverRunner, but what if it isn't?
Should the resolver return/throw a diagnostic with information on why it didn't resolve? This would enable advanced diagnostics like Rome does but how would we do this if there are multiple resolver, return the last one? If the last one is a simple resolver for a project specific syntax the chances are big that one has far less advanced diagnostics to the Parcel one.

parcel-benchmark:
## Benchmark Results

#### Timings

| --- | --- | --- |
| Cold | 10.48s | +23.00ms |
| Cached | 9.19s | +116.00ms |

#### Cold Bundles

| dist/legacy/parcel.ff215454.webp | 102.94kb | +0.00b | 204.00ms | +21.00ms ⚠️ |
| dist/modern/parcel.76ee4591.webp | 102.94kb | +0.00b | 201.00ms | +19.00ms ⚠️ |

#### Cached Bundles

| dist/modern/parcel.76ee4591.webp | 102.94kb | +0.00b | 12.00ms | +1.00ms ⚠️ |
| dist/legacy/index.js | 252.00b | +0.00b | 21.00ms | +4.00ms ⚠️ |

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 16.01s | +210.00ms |
| Cached | 13.24s | -124.00ms |

#### Cold Bundles

*No bundle changes detected.*

#### Cached Bundles

*No bundle changes detected.*

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 3.82m | -1.29s |
| Cached | 11.53s | -103.00ms |

#### Cold Bundles

| dist/index.js | 2.31mb | -312.00b 🚀 | 2.16m | -1.65s |
| dist/media-viewer.12bd9010.js | 86.76kb | +0.00b | 10.83s | +992.00ms ⚠️ |
| dist/card.80c210d3.js | 76.35kb | +0.00b | 10.83s | +992.00ms ⚠️ |
| dist/popup.37645d0d.js | 44.40kb | +0.00b | 42.66s | -3.41s 🚀 |
| dist/DatePicker.6844209a.js | 27.57kb | +0.00b | 50.50s | -2.65s 🚀 |
| dist/EmojiPickerComponent.bb15ebef.js | 26.66kb | +0.00b | 50.50s | -2.65s 🚀 |
| dist/EmojiUploadComponent.0460cc59.js | 2.91kb | +0.00b | 50.50s | -2.65s 🚀 |
| dist/ResourcedEmojiComponent.2d80adbd.js | 2.25kb | +0.00b | 10.70s | +966.00ms ⚠️ |
| dist/esm.976debcc.js | 178.00b | +0.00b | 10.58s | +969.00ms ⚠️ |
| dist/dropzone.976debcc.js | 177.00b | +1.00b ⚠️ | 2.16m | +1.39m ⚠️ |
| dist/clipboard.976debcc.js | 177.00b | +1.00b ⚠️ | 46.05s | -1.65s |
| dist/browser.976debcc.js | 177.00b | +1.00b ⚠️ | 1.42m | +37.50s ⚠️ |
| dist/editorView.976debcc.js | 177.00b | -1.00b 🚀 | 45.03s | -23.46s 🚀 |
| dist/media-card-analytics-error-boundary.976debcc.js | 176.00b | -5.00b 🚀 | 46.88s | -1.40m 🚀 |
| dist/16.976debcc.js | 174.00b | -5.00b 🚀 | 46.05s | -2.38s 🚀 |

#### Cached Bundles

| dist/index.js | 2.32mb | -312.00b 🚀 | 398.00ms | -1.00ms |
| dist/esm.976debcc.js | 176.00b | +2.00b ⚠️ | 168.00ms | -4.00ms |
| dist/media-card-analytics-error-boundary.976debcc.js | 176.00b | +2.00b ⚠️ | 165.00ms | -4.00ms |
| dist/dropzone.976debcc.js | 176.00b | +2.00b ⚠️ | 161.00ms | -4.00ms |
| dist/clipboard.976debcc.js | 176.00b | +2.00b ⚠️ | 150.00ms | -5.00ms |
| dist/browser.976debcc.js | 176.00b | +2.00b ⚠️ | 150.00ms | -4.00ms |
| dist/16.976debcc.js | 176.00b | +2.00b ⚠️ | 141.00ms | -4.00ms |
| dist/editorView.976debcc.js | 176.00b | +2.00b ⚠️ | 125.00ms | -3.00ms |

</p></details>

**[Click here to view a detailed benchmark overview.](https://benchmark.parceljs.org/benchmark/249288652804325897)**

packages/core/core/src/ResolverRunner.js
devongovett: Hmm, well ideally it *would* be based on the original source code. That's what the user wrote, so it makes sense to show them errors based on that (not generated code). The `loc` should be based on the original source code too. If only AST transforms happen (in the case of babel), then this should be true, otherwise we should be passing an input source map through as part of the transformation (e.g. TSC -> JS - doesn't seem to be the case currently though 😞).

packages/core/core/src/ResolverRunner.js
devongovett: Well, we can only show as much information as the resolver gave us... Normally, we return null from resolvers when they don't resolve anything and go to the next one, and we should keep doing that. However, I think the DefaultResolver should throw an error in a couple specific cases:

1. User specified `main`, `module`, etc. in package.json that points to a file that doesn't exist (after trying all extensions).
2. User specifies an `alias` that points to a file that doesn't exist.
3. package.json is not parsable as JSON.
4. Fields we read in package.json are not valid (e.g. `main` is a boolean or something).

In these cases, the errors are caused by user code being incorrect. We should surface those so they can be fixed. These could be additional code frames in addition to the one showing where the import came from in the original code.

packages/core/core/src/ResolverRunner.js
DeMoorJasper: So there's no way we could fix this? I tried it on Babel btw so it should've returned the correct location? Which it didn't

devongovett: Hmm I’m confused. Why would the error location be incorrect?

DeMoorJasper: This is what I get when using the dependency location on the source

Screenshot 2019-11-17 at 21 57 49

DeMoorJasper: Nevermind, it was 0-based, works now. I thought every location in Babel was 1-based as there codeframes where as well...

Screenshot 2019-11-17 at 22 03 50

mischnic: https://github.com/babel/babel/blob/master/packages/babel-parser/ast/spec.md#node-objects:
> Each Position object consists of a line number (1-indexed) and a column number (0-indexed):

DeMoorJasper: Ow got it will not add +1 on columns

packages/core/core/src/ResolverRunner.js
DeMoorJasper: Sounds good, will do the advanced diagnostics with hints in a follow-up. Gonna focus on the basic diagnostics the resolverrunner can create in this PR

packages/transformers/js/src/visitors/dependencies.js
```diff
@@ -240,15 +240,15 @@ function addDependency(
 ) {
   asset.addDependency({
     moduleSpecifier: node.value,
-    loc: node.loc && node.loc.start,
+    loc: node.loc && createDependencyLocation(node.loc.start, node.value, 1, 1),
```
devongovett: I think these are off by one line. I'm getting this:

image

Changing `lineOffset` to `0` makes it work...

devongovett:
We should figure out how to test these error messages so we know if the underlying tools change the offsets or something...

DeMoorJasper:
@devongovett so I should write an integration test for babel and postcss resolver errors to check if the offsets are correct or should I figure out a way to do this without the resolver in between? I doubt it would ever change though, seems like a pretty significant breaking change

devongovett:
Yeah, seems more likely that someone would inadvertently change our implementation and not realize it broke the error handling rather than theirs. We could get this in and you could write some tests separately if you want.

packages/transformers/js/src/visitors/dependencies.js
DeMoorJasper: strange it worked on my test, I'll change it and maybe add a couple tests later today or tomorrow

Issue #3377, opened 2019-08-13
🚀 Help test the Parcel 2 prerelease


mischnic:
We have just released Parcel 2 ~alpha 1~ ~alpha 2~ alpha 3 (`yarn add -D parcel@next`)! 

For a more bleeding edge version, use `yarn add -D parcel@nightly` (published every midnight UTC).

Please try it out and report any issues you find. If you'd like to contribute, we need help migrating over all of the remaining functionality from Parcel 1 (for example on [these issues](https://github.com/parcel-bundler/parcel/issues?q=is%3Aopen+is%3Aissue+label%3A%22%3Apray%3A+Help+Wanted%22+label%3A%22%F0%9F%94%8C+Parcel+2+Plugins%22)).

You can read more details in the announcement post (https://medium.com/@devongovett/parcel-2-0-0-alpha-1-is-here-8b160c6e3f7e) and the V2 Readme (https://github.com/parcel-bundler/parcel/blob/v2/packages/core/parcel/README.md).

To concentrate on Parcel 2, Parcel `1.x.x` will only receive bugfixes (and likely only critical ones).

*Please don't report bugs or feature requests here in the comments but open new issues instead.*

DOSputin:
Awesome!

Looking over the medium post and other docs and open issues, there are lot's of mentions of the deprecated `.babelrc` configuration method.  Didn't see any mention or confirmation of added support for babel 7's `babel.config.js`?

devongovett:
I don't think `.babelrc` is deprecated, but yes, `babel.config.js` will be supported. See #3261.

hacknug:
Is `parcel serve` supposed to use `~/package.json#source` automatically? All I get is a `No entries found` if I don't pass my entry file explicitly to the CLI.

arcanis:
One early feedback: lots of exceptions seem to be eaten somewhere, which makes writing plugins a bit harder as you need to wrap them in try/catch statements to print them manually.

I also noticed that the promise queue never seems to throw when one of the promises it contains fails, causing the program to exit as soon as the event loop is empty - and never waking up the promise queue.

Banou26:
You guys should probably create issues for those observations.

JounQin:
Why `--out-dir` is removed?

devongovett:
[...]

PR #3804, opened 2019-11-18
chore(README.md): replace transform with transformer name

Stradivario:
# ↪️ [Pull Request](https://github.com/parcel-bundler/parcel/issues/3377#issuecomment-554996003)
## ✔️ By specification `transform` is renamed to `transformer` thus we need to update readme.md according.

- [x] Rename `transform` with `transformer` inside README.md

### DIFF
--- a/packages/core/parcel/README.md
+++ b/packages/core/parcel/README.md
@@ -626,7 +626,7 @@ See [Resolvers](#resolvers)
 ```json
 {
   "transforms": {
-    "*.js": ["@parcel/transform-babel"]
+    "*.js": ["@parcel/transformer-babel"]
   }
 }
 ```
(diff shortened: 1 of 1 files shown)


packages/core/parcel/README.md
```diff
@@ -626,7 +626,7 @@ See [Resolvers](#resolvers)
 ```json
 {
   "transforms": {
```
devongovett: We should probably change this in parcelrc to match as well...

Stradivario: Sounds reasonable but i am not aware of the changes and where are they applied and this is why i didn't changed to not break something.

Stradivario: If it is "transforms" the values inside should be "transform"
If it is "transformers" it sounds good to be "transformer" 

Basically i will change them according to what you wrote and we can discuss more further.

Issue #3803, opened 2019-11-18
`--open` flag not working in Parcel 2


pnfcre:
# 🐛 Bug Report

I tried passing `--open` to the `parcel src/index.html` command, but nothing happened. This feature works perfectly in the latest stable `parcel-bundler` package.

## 🎛 Configuration (package.json)

```json
{
  "scripts": {
    "start": "parcel src/index.html --open"
  }
}
```

## 🤔 Expected Behavior

Parcel should open `http://localhost:1234` in my default web browser.

## 😯 Current Behavior

Nothing happens.

## 🔦 Context

I tried to use the `--open` flag as I'm used to do in `parcel-bundler`.

## 💻 Code Sample

src/index.html:
```html
hello world
```

package.json:
```json
{
  "scripts": {
    "start": "parcel src/index.html --open"
  }
}
```

```sh
$ yarn start
```

## 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Parcel           | 2.0.0-alpha.2.1
| Node             | v13.0.1
| Yarn             | 1.19.1
| Operating System | Windows 10 v1809 (x64)

Taym95:
I will take this issue!

mackness:
Seem to work with parcel@1.12.4

pnfcre:
Yes, this issue is specific to Parcel 2 @mackness.

Taym95:
I think this one can be closed! #4596

DeMoorJasper:
Ow yes it can, thanks for the heads up

PR #3825, opened 2019-11-21
Add browser open flag

Taym95:
# ↪️ Pull Request
Implement #3803 

## 🚨 Test instructions
Use `--open` flag to open browser:

`parcel index.html --open`

## ✔️ PR Todo

### DIFF
--- a/packages/core/parcel/src/cli.js
+++ b/packages/core/parcel/src/cli.js
@@ -195,10 +195,11 @@ async function run(entries: Array<string>, command: any) {
     });
 
     if (command.open && command.name() === 'serve') {
-      const port = command.port || process.env.PORT || 1234;
+      const {
+        serve: {port, https, host}
+      } = parcel.getOptions();
       await openInBrowser(
-        `${command.https ? 'https' : 'http'}://${command.host ||
-          'localhost'}:${port}`,
+        `${https ? 'https' : 'http'}://${host || 'localhost'}:${port}`,
         command.open
       );
     }
--- a/packages/core/parcel/src/cli.js
+++ b/packages/core/parcel/src/cli.js
@@ -194,8 +194,7 @@ async function run(entries: Array<string>, command: any) {
       }
     });
 
-    command.target = command.target.length > 0 ? command.target : 'browser';
-    if (command.open && command.target === 'browser') {
+    if (command.open && command.name() === 'serve') {
       const port = command.port || process.env.PORT || 1234;
       await openInBrowser(
         `${command.https ? 'https' : 'http'}://${command.host ||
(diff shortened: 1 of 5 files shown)


packages/core/utils/src/openInBrowser.js
```diff
@@ -0,0 +1,58 @@
...
+
+function getAppName(appName: string): string {
+  if (['google', 'chrome'].includes(appName)) {
+    return getChromeAppName();
+  } else if (['brave', 'Brave'].includes(appName)) {
+    return 'Brave Browser';
+  } else return appName;
+}
+
+export default async function openInBrowser(url: string, browser: string) {
```
Taym95: The best solution will be to use `'react-dev-utils/openBrowser';` but I didn't want to add external dependency just for an opening a browser so I used existing function and I added some changes, but you someone want to use `'react-dev-utils/openBrowser';`` I can add it!

packages/core/parcel/src/cli.js
```diff
@@ -163,6 +164,16 @@ async function run(entries: Array<string>, command: any) {
       }
     });
 
+    command.target = command.target.length > 0 ? command.target : 'browser';
+    if (command.open && command.target === 'browser') {
```
mischnic: `command.target` doesn't do the same thing as in Parcel 1, it build only the list of targets that you specify, not all in `package.json#targets`.

Also, the browser should only open in serve mode, not with `parcel watch`.

The port is already determined in `normalizeOptions` below, `command.port` without any fallbacks is sufficient.

Taym95: `command.port` it's giving `undefined` that why I used: 
`const port = command.port || process.env.PORT || 1234;`

mischnic: The actual port is in the return value of  `normalizeOptions` (in `.serve.port`):

https://github.com/parcel-bundler/parcel/blob/cf3e84cbfc5d3f4a18da9fdcdacaae96c75b5550/packages/core/parcel/src/cli.js#L156

https://github.com/parcel-bundler/parcel/blob/cf3e84cbfc5d3f4a18da9fdcdacaae96c75b5550/packages/core/parcel/src/cli.js#L230-L245

Taym95: Because `options` are private in `Parcel` class, I added `getOptions` so I can access to `port, https, host`, if you have better solution please provide it so we can improve it!

DeMoorJasper:
Could you rebase this PR and ensure it uses proper formatting so the diff isn't that large?

mischnic:
@Taym95 ?

Taym95:
> @Taym95 ?

Hey, sorry for inactivity!

DeMoorJasper:
Closing due to inactivity, gonna create a new PR with a fixed version of this...

PR #3850, opened 2019-11-25
Import inspector only if it's available

### LINKED ISSUE #3819: V2: detect inspector and fallback is necessary
# 🙋 feature request

https://github.com/parcel-bundler/parcel/blob/d24e09a9797af41a9d2c6c733b1c6cf6671d60c7/packages/core/workers/src/Profiler.js#L2

The inspector module not always available, it depends on build param `--without-inspector`

## 🤔 Expected Behavior

Compatible with `--without-inspector` mode.

## 😯 Current Behavior

Direct import

## 💁 Possible Solution

Not pretty sure how to implement a fallback

## 🔦 Context

In some production environments, built node with `--without-inspector`, because v8_inspector use `prime_rehash_policy` symbol that requires GLIBCXX_3.4.18 on some systems.

Taym95:
# ↪️ Pull Request
Try to fix #3819 

## ✔️ PR Todo


packages/core/workers/src/Profiler.js
```diff
@@ -39,17 +39,23 @@ type PositionTickInfo = {|
...
+      return Promise.all([
+        this.sendCommand('Profiler.setSamplingInterval', {
+          interval: 100
+        }),
+        this.sendCommand('Profiler.enable'),
+        this.sendCommand('Profiler.start')
+      ]);
+    } catch (err) {
+      /* eslint-disable no-console */
+      console.error('Profiling not available', err);
```
mischnic: This way, in `sendCommand` `invariant(this.session != null);` would throw an unhelpful error. So we should throw an error here instead of `console.error`.

packages/core/workers/src/Profiler.js
```diff
@@ -39,17 +39,22 @@ type PositionTickInfo = {|
 export default class Profiler {
   session: Session;
 
-  startProfiling() {
-    this.session = new Session();
-    this.session.connect();
+  async startProfiling() {
+    try {
+      const inspector: any = await require('inspector');
```
devongovett: `require` is synchronous. No need to `await`

devongovett: Also, `inspector` is still being imported at the top of the file. Was that intentional?

Taym95: Yes exactly, but we will have type issue in this line: 

https://github.com/parcel-bundler/parcel/blob/42f3276c86a8da0a8cb1a1f462e8589eb82ff68e/packages/core/workers/src/Profiler.js#L40

mischnic: Use `import type {Session} from ...` or `import typeof {Session} from ...`

Taym95: Thanks, this is what I was looking for!

packages/core/workers/src/Profiler.js
mischnic: Well, `try {...} catch(e){ throw e; }` is redundant. The error should actually tell the user that the Parcel flag `--profile` flag is causing this and errors should only be caught for `require('inspector')`, not for the commands after that
Also, the `any` in `const inspector: any = ...` isn't needed

packages/core/workers/src/Profiler.js
```diff
@@ -40,16 +40,21 @@ export default class Profiler {
...
-    ]);
+      return Promise.all([
+        this.sendCommand('Profiler.setSamplingInterval', {
+          interval: 100
+        }),
+        this.sendCommand('Profiler.enable'),
+        this.sendCommand('Profiler.start')
+      ]);
+    } catch (err) {
+      throw err;
```
devongovett: why do you re-throw the same error here? Is the try..catch needed at all?

mischnic: -> https://github.com/parcel-bundler/parcel/pull/3850#discussion_r355049882

devongovett: Should we wrap in a diagnostic?

Taym95: Hey @devongovett, what do you mean by `diagnostic `?

packages/core/workers/src/Profiler.js
```diff
@@ -40,16 +41,23 @@ export default class Profiler {
...
+      return Promise.all([
+        this.sendCommand('Profiler.setSamplingInterval', {
+          interval: 100
+        }),
+        this.sendCommand('Profiler.enable'),
+        this.sendCommand('Profiler.start')
+      ]);
+    } catch (err) {
+      throw new ThrowableDiagnostic({
+        diagnostic: errorToDiagnostic(err)
```
mischnic: This would probably result in this error message (which isn't very helpful):
```
🚨  Error: Cannot find module 'inspector'
Require stack:
- .../parcel/packages/core/workers/src/Profiler.js
- ...
```

So instead (to also only catch require errors and not something related to the Session itself): 

```js
let inspector;
try {
    inspector = require('inspectors');
} catch (err) {
        diagnostic: {
            message: `The inspector module isn't available`,
            origin: '@parcel/workers',
            hints: ['Disable build profiling']
        }
    });
}

this.session = new inspector.Session();
...
```

Taym95:
Hey guys, why the lint is still complaining?

DeMoorJasper:
@Taym95 trailing commas got added to the prettier rules, fixed it. Once the tests pass I'll merge it

PR #3860, opened 2019-11-29
Add flag to listen on STDIN close

### LINKED ISSUE #2710: Add option to/watch for stdin close
# 🙋 feature request

I want to use Parcel together with my [Phoenix](https://phoenixframework.com) project. This projects allows you to define assets builder for you by simply stating how to start it, however it will not send signal to the watching process when it exists, it will just close `stdin` and will hope that process notice that and will close itself.

## 💁 Possible Solution

- listen on `stdin` and close itself when those close;
- or add flag similar to Webpack `--watch-stdin` that will enable such behaviour

## Current behaviour

```
true | parcel watch index.js
```

Will keep running

## Expected behaviour

```
true | parcel watch --watch-stdin index.js
```

Will end immediately.

hauleth:
# ↪️ Pull Request

This allows using Parcel as asset builder in Phoenix pipelines which requires
builder to close when STDIN closes.

## 💻 Examples

Just run:

```sh
echo | parcel watch --watch-for-stdin file.js
```

## 🚨 Test instructions

The same as in example. The application should just stop and exit.

## ✔️ PR Todo

Close #2710


packages/core/parcel/src/cli.js
```diff
@@ -88,6 +88,7 @@ applyOptions(serve, commonOptions);
 let watch = program
   .command('watch [input...]')
   .description('starts the bundler in watch mode')
+  .option('--watch-for-stdin', 'exit when stid closes')
```
wbinnssmith: nit: stid => stdin

hauleth: Fixed

packages/core/parcel/src/cli.js
```diff
@@ -88,6 +88,7 @@ applyOptions(serve, commonOptions);
 let watch = program
   .command('watch [input...]')
   .description('starts the bundler in watch mode')
+  .option('--watch-for-stdin', 'exit when stdin closes')
```
devongovett: Should we also add this to the `serve` command?

hauleth: I do not use `serve` command so I didn't know if this would be useful for anyone, but it can be added.

hauleth:
Added that option to `parcel serve` as well.

PR #3897, opened 2019-12-10
Have NODE_ENV reflect Parcel's environment, not what it's building

wbinnssmith:
Previously, `NODE_ENV` varied based on what Parcel was building, which prevented libraries like React, which is used through `ink` in Parcel's CLI, from behaving in production mode in Parcel itself. 

Now, we capture any original `NODE_ENV`  as a build option and pass it through in `env`, and then set Parcel's own  `NODE_ENV` to `PARCEL_BUILD_ENV` if it's set ("production" in builds). Any passed `env` is now merged with the results of loading `dotenv`.

~This required adding a top-level Parcel option for `nodeEnv` as a build option, distinct from the `env` option. `NODE_ENV` could be added to an `env` passed by the cli, but we rely on the absence of `env` to determine whether to load the environment from `dotenv`.~

This also reverses the priority of `dotenv` files, making more specific `dotenv` files override less specific ones. This grew out of the integration test for `dotenv` with this behavior. It seems more intuitive to me. cc @demoorjasper

Test Plan:

Added integration test for overriding `NODE_ENV` in cases with and without a `dotenv` file present. Additionally:
* Verify that `process.env.NODE_ENV` remains `undefined` in Node when running Parcel from source, and that it loads the development build of React
* Verify that production React in Parcel itself is loaded when using a built version of Parcel
* Verify that a development version of React is loaded when using `parcel serve` in the react-hmr example
* Verify that a development version of React is *not* loaded when using `parcel server` with `NODE_ENV` set to "production" when invoking Parcel (Note: This required clearing cache between development and production builds as environment variables don't currently seem to invalidate cache.)

### DIFF
--- a/packages/core/parcel/src/cli.js
+++ b/packages/core/parcel/src/cli.js
@@ -35,6 +35,16 @@ const path = require('path');
 const getPort = require('get-port');
 const version = require('../package.json').version;
 
+// Capture the NODE_ENV this process was launched with, so that it can be
+// used in Parcel (such as in process.env inlining).
+const initialNodeEnv = process.env.NODE_ENV;
+// Then, override NODE_ENV to be PARCEL_BUILD_ENV (replaced with `production` in builds)
+// so that dependencies of Parcel like React (which renders the cli through `ink`)
+// run in the appropriate mode.
+if (typeof process.env.PARCEL_BUILD_ENV === 'string') {
+  process.env.NODE_ENV = process.env.PARCEL_BUILD_ENV;
+}
+
 program.version(version);
 
 // --no-cache, --cache-dir, --no-source-maps, --no-autoinstall, --global?, --public-url, --log-level
@@ -231,9 +241,9 @@ async function run(entries: Array<string>, command: any) {
 async function normalizeOptions(command): Promise<InitialParcelOptions> {
   let nodeEnv;
   if (command.name() === 'build') {
-    nodeEnv = process.env.NODE_ENV || 'production';
+    nodeEnv = initialNodeEnv || 'production';
   } else {
-    nodeEnv = process.env.NODE_ENV || 'development';
+    nodeEnv = initialNodeEnv || 'development';
   }
 
   let https = !!command.https;
@@ -294,6 +304,8 @@ async function normalizeOptions(command): Promise<InitialParcelOptions> {
     autoinstall: command.autoinstall ?? true,
     logLevel: command.logLevel,
     profile: command.profile,
-    nodeEnv,
+    env: {
+      NODE_ENV: nodeEnv,
+    },
   };
 }
--- a/packages/core/parcel/src/cli.js
+++ b/packages/core/parcel/src/cli.js
@@ -229,10 +229,11 @@ async function run(entries: Array<string>, command: any) {
 }
 
 async function normalizeOptions(command): Promise<InitialParcelOptions> {
+  let nodeEnv;
   if (command.name() === 'build') {
-    process.env.NODE_ENV = process.env.NODE_ENV || 'production';
+    nodeEnv = process.env.NODE_ENV || 'production';
   } else {
-    process.env.NODE_ENV = process.env.NODE_ENV || 'development';
+    nodeEnv = process.env.NODE_ENV || 'development';
   }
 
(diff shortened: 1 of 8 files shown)


parcel-benchmark:
## Benchmark Results

#### Timings

| --- | --- | --- |
| Cold | 6.17s | -3.24s 🚀 |
| Cached | 4.30s | -2.64s 🚀 |

#### Cold Bundles

| dist/legacy/parcel.ff215454.webp | 102.94kb | +0.00b | 192.00ms | +45.00ms ⚠️ |
| dist/modern/parcel.76ee4591.webp | 102.94kb | +0.00b | 191.00ms | +45.00ms ⚠️ |
| dist/legacy/index.js | 243.00b | +0.00b | 418.00ms | +78.00ms ⚠️ |
| dist/modern/index.js | 215.00b | +0.00b | 413.00ms | +59.00ms ⚠️ |
| dist/legacy/index.css | 29.00b | +0.00b | 1.26s | +109.00ms ⚠️ |

#### Cached Bundles

| dist/modern/parcel.76ee4591.webp | 102.94kb | +0.00b | 10.00ms | +1.00ms ⚠️ |
| dist/legacy/index.css | 29.00b | +0.00b | 10.00ms | +1.00ms ⚠️ |
| dist/modern/index.css | 29.00b | +0.00b | 10.00ms | +1.00ms ⚠️ |

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 9.55s | -3.83s 🚀 |
| Cached | 6.11s | -2.98s 🚀 |

#### Cold Bundles

| dist/module/index.js | 34.02kb | +2.00b ⚠️ | 1.79s | -51.00ms |
| dist/main/index.js | 34.00kb | +2.00b ⚠️ | 1.79s | -52.00ms |

#### Cached Bundles

| dist/module/index.js | 34.03kb | +2.00b ⚠️ | 8.00ms | -2.00ms 🚀 |
| dist/main/index.js | 34.00kb | +2.00b ⚠️ | 11.00ms | -5.00ms 🚀 |

</p></details>

#### Timings

| --- | --- | --- |

#### Cold Bundles

#### Cached Bundles

</p></details>

**[Click here to view a detailed benchmark overview.](https://benchmark.parceljs.org/benchmark/251399343586673161)**

devongovett:
Would prefer to avoid having a separate option for this from env. I think other environment variables should probably work the same way too (ie merge with .env not be replaced by it). Could we always load .env and not have that depend on the env option existing or not?

DeMoorJasper:
Not sure why I was cced into this issue, looks good to me only concern I have is similar to Devon's comment.
Can't we infer the NODE_ENV in any way other than passing an extra option specifically for it?

wbinnssmith:
>  I think other environment variables should probably work the same way too (ie merge with .env not be replaced by it). Could we always load .env and not have that depend on the env option existing or not?

This makes sense to me. The only thing is that `dotenv` depends on knowing the `NODE_ENV` to determine which `.env` files to load. I suppose it could get this from the passed `env` itself. I wonder if this should be happening in the CLI instead. I don't know if it would be surprising for external tools using Parcel as a library to have `dotenv` transparently loaded.

> Not sure why I was cced in this PR

This was because of the `git blame` on the dotenv code 😅, though you may have only ported it from Parcel 1.

wbinnssmith:
Just pushed some changes implementing this 😄

packages/core/parcel/src/cli.js
```diff
@@ -35,6 +35,16 @@ const path = require('path');
 const getPort = require('get-port');
 const version = require('../package.json').version;
 
+// Capture the NODE_ENV this process was launched with, so that it can be
+// used in Parcel (such as in process.env inlining).
+const initialNodeEnv = process.env.NODE_ENV;
+// Then, override NODE_ENV to be PARCEL_BUILD_ENV (replaced with `production` in builds)
+// so that dependencies of Parcel like React (which renders the cli through `ink`)
```
kwelch: >through `ink`
Was that intentional? was it possibly suppose to be link

mischnic: it is correct: https://www.npmjs.com/package/ink

PR #3918, opened 2019-12-15
Change HMR websocket to use the dev server

lbogdan:
# ↪️ Pull Request

As discussed with @devongovett on Spectrum, this is a proof of concept for changing the HMR socket from using a separate HTTP server on a secondary port to using the dev server.

The main motivation for this change is to be able to run Parcel on CodeSandbox inside a container sandbox out-of-the-box, and probably other remote development environments / tools that only forward / proxy a single HTTP port.

As the title implies, this is a PoC, and you should probably take into account that:
- I've never used Flow, the type changes are based solely on my TS experience;
- I don't have enough context to figure out if the changes accommodate all use cases
- as v2 is still alpha, I removed all backwards compatibility - cli arguments & old behavior.

That being said, please take a look at this and let me know what your thoughts are. Thanks!

## ✔️ PR Todo


packages/reporters/dev-server/src/ServerReporter.js
```diff
@@ -1,11 +1,15 @@
...
+import type {HTTPServer} from './types.js.flow';
+export type {HTTPServer};
+
 import {Reporter} from '@parcel/plugin';
 import invariant from 'assert';
 import Server from './Server';
 import path from 'path';
 
 let servers: Map<number, Server> = new Map();
+export let devServer: HTTPServer | null = null;
```
DeMoorJasper: Not sure if this is a good idea, maybe we should merge the hmr and dev server into one reporter so we don't have to share state between them.

Also won't this make the servers map unusable? This allowed parcel to run multiple servers at once, mainly when using the api this is possible and shouldn't break.

lbogdan: > Not sure if this is a good idea, maybe we should merge the hmr and dev server into one reporter so we don't have to share state between them.

I agree that merging them would make the most sense (and I guess also address your other concert?), I just didn't want to start with that big a refactor. 🙂 

> Also won't this make the servers map unusable? This allowed parcel to run multiple servers at once, mainly when using the api this is possible and shouldn't break.

Hm, interesting, can you point me to a documentation section / article / project using such a setup?

DeMoorJasper: I can't point to a setup that does this, it was just to prevent future issues if anyone would ever create more than one parcel instance using the api that starts up a devserver.

Originally I thought it would maybe get used in a setup where someone wants to run multiple applications for example in a monorepo while sharing the workers and cache for better performance

lbogdan: Oh, OK, I think I understand the use case, and it would most probably not work in the current form. So should I fix it, or refactor the whole thing and merge the reporters?

DeMoorJasper: I think so, merging the reporters seems like the best solution to fix this imo

lbogdan: Out of curiosity, was there a reason to not do this (HMR websocket using the dev server) from the get-go?

DeMoorJasper: HMR and Dev server had a different port on Parcel 1 so I kept it that way in Parcel 2 and therefore also created 2 reporters

Not entirely sure why Devon made it like this originally

lbogdan:
@DeMoorJasper As discussed, I moved `HMRServer` to the `dev-server` reporter. Please take a look and let me know what you think.

lbogdan:
> was there anything else you wanted to add to this PR?

@DeMoorJasper Not really, it should be good to go! Do you think it would make sense to do the same for Parcel 1, maybe under a command argument, to not break backwards compatibility?

DeMoorJasper:
awesome, I don't think we should do this for Parcel 1 as it's kind of in maintenance mode.

Would have to ask @devongovett

lbogdan:
Oh, OK, I didn't know that, was thinking that it's still actively maintained. But let me know nevertheless, I did a PoC for it, too.

PR #3920, opened 2019-12-15
[T-343] Use loadConfig for postcss

DeMoorJasper:
# ↪️ Pull Request

This PR updates the postcss config loading to use the new config api.

## ✔️ PR Todo


parcel-benchmark:
## Benchmark Results

#### Timings

| --- | --- | --- |
| Cold | 11.51s | +78.00ms |
| Cached | 7.88s | -63.00ms |

#### Cold Bundles

| dist/legacy/parcel.ff215454.webp | 102.94kb | +0.00b | 195.00ms | -23.00ms 🚀 |
| dist/modern/parcel.76ee4591.webp | 102.94kb | +0.00b | 193.00ms | -23.00ms 🚀 |
| dist/modern/index.css | 29.00b | +0.00b | 1.21s | -169.00ms 🚀 |

#### Cached Bundles

| dist/modern/parcel.76ee4591.webp | 102.94kb | +0.00b | 13.00ms | -1.00ms 🚀 |
| dist/legacy/index.js | 243.00b | +0.00b | 20.00ms | -4.00ms 🚀 |
| dist/modern/index.js | 215.00b | +0.00b | 15.00ms | +1.00ms ⚠️ |
| dist/legacy/index.css | 29.00b | +0.00b | 13.00ms | -1.00ms 🚀 |

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 15.85s | +71.00ms |
| Cached | 7.95s | +77.00ms |

#### Cold Bundles

*No bundle changes detected.*

#### Cached Bundles

*No bundle changes detected.*

</p></details>

#### Timings

| --- | --- | --- |

#### Cold Bundles

#### Cached Bundles

</p></details>

**[Click here to view a detailed benchmark overview.](https://benchmark.parceljs.org/benchmark/251843832882135562)**

packages/transformers/postcss/src/loadConfig.js
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+  return configHydrator(configFile, config, options);
+}
+
+export function preSerialize(config: Config) {
+  if (!config.result) return;
+
+  // This is a very weird bug
+  /*config.setResult({
```
wbinnssmith: What's the bug? 😄

DeMoorJasper: not sure it just didn't work with setResult but just overwriting result as I did works fine. The types said `TODO: Fix bug` so I assume it was a known issue

devongovett:
Ah! We are actually using `postcss.config.js` at work. 😢 We use it to extend from another postcss config from a package in node_modules... I don't believe there is a way to do that in JSON at the moment. Thoughts?

DeMoorJasper:
@devongovett you could probably extract it into a package as it’s a monorepo? I can also add support for it in here but than css assets will never really be truly cached when using the `postcss.config.js` resulting in slower builds similar to the babel one... Or maybe I can add support for extends? That might be cool

talentlessguy:
@devegovett I have a `.postcssrc` file in one my PostCSS project and it works fine :think:

packages/transformers/postcss/src/loadConfig.js
```diff
@@ -0,0 +1,149 @@
...
+      ) {
+        throw new Error('PostCSS config must have plugins');
+      }
+
+      let configFilePlugins = Array.isArray(configFile.plugins)
+        ? configFile.plugins
+        : Object.keys(configFile.plugins);
+      for (let p of configFilePlugins) {
+        // JavaScript configs can use an array of functions... ugh
+        if (typeof p === 'function') {
```
DeMoorJasper: Not entirely sure if there is a way to tell the config object something is not serialisable?

cc @devongovett @padmaia

packages/core/integration-tests/test/postcss.js
```diff
@@ -291,7 +291,7 @@ describe('postcss', () => {
         name: 'style.css',
         assets: ['style.css'],
         includedFiles: {
-          'style.css': ['.postcssrc', 'config.css', 'package.json'],
```
DeMoorJasper: Shouldn't this still be a dependency?

Just wanna check to ensure there's nothing wrong with the caching on this implementation

packages/transformers/postcss/src/loadConfig.js
```diff
@@ -0,0 +1,149 @@
...
+  options,
+  logger,
+}: {|
+  config: Config,
+  options: PluginOptions,
+  logger: PluginLogger,
+|}) {
+  let configFile: any = await config.getConfig(
+    ['.postcssrc', '.postcssrc.json', 'postcss.config.js', '.postcssrc.js'],
+    {packageKey: 'postcss'},
```
DeMoorJasper: `package.json` key config seems to be broken from the test I wrote, not entirely sure if this is a known issue or if I just use it incorrectly

packages/core/integration-tests/test/postcss.js
padmaia: These will be included files on the Config instead of the Asset now. Can't really test it the same way since you don't have access to the config files. For Babel I added some tests that build, change config, build again and make sure the output changed in the way that I expected.

DeMoorJasper: Ow alright, will leave it as is for now then. I'll add additional tests if requested or if issues arise

packages/transformers/postcss/src/loadConfig.js
padmaia: You mean something generic that wouldn't require knowledge of the config format? That would be interesting.

packages/core/integration-tests/test/postcss.js
mischnic: Can the `assertBundles` be extended to assert configs (and these files) as well?

packages/transformers/postcss/src/loadConfig.js
DeMoorJasper: I mean tell the config that there's no way to serialise this as it contains functions.
Unless we'd eval or new Function them afterwards although that sounds like a terrible idea

packages/transformers/postcss/src/loadConfig.js
```diff
@@ -0,0 +1,149 @@
...
+        ? configFile.plugins
+        : Object.keys(configFile.plugins);
+      for (let p of configFilePlugins) {
+        // JavaScript configs can use an array of functions... ugh
+        if (typeof p === 'function') {
+          configFile.__contains_functions = true;
+
+          // Just do all the things, tbh have no clue what this does...
+          config.shouldInvalidateOnStartup();
+          config.shouldReload();
```
padmaia: `shouldInvalidateOnStartup()` is used in the case where the config is programming language file as opposed to a static format file. The file may not have changed since the last run, but something that the file could potentially reach out to might have so we have to load the config again to make sure the results are the same.

`shouldReload` is a hold over from when we were loading config in the main process and passing it to worker processes, but it's not really necessary anymore. I cleaned up the `BabelTransformer`'s use of it when I made that change, but I forgot to remove the capability. Extra credit if you wanna do that in this PR 😉

packages/transformers/postcss/src/loadConfig.js
padmaia: Oh I see. Yeah right now it wouldn't matter because anytime we rerun a transformation we reloading all of it's configs. I think we were, but there was a bit of a regression when refactoring everything to run in the worker thread. Pretty sure the `shouldReload` helped with that now that I'm thinking about it. Maybe we should keep the function and use it when we start using cached configs for requests that weren't invalidated again. (Maybe that's why I kept it around)

I was thinking of a common util that would walk over object properties and array values recursively or something to determine if it could not be serialized for some reason, but I'm not sure how useful that would actually be.

packages/core/integration-tests/test/postcss.js
padmaia: Well the config objects are not returned as part of the `BundleGraph` at the moment. We were a bit limited before when core was loading configs in the main process because we didn't have access to the Asset yet, but now we could attach that information to the Asset if we think it would be valuable.

packages/transformers/postcss/src/loadConfig.js
padmaia: > but it's not really necessary anymore.

Scratch that, it may end up being useful. See [this comment](https://github.com/parcel-bundler/parcel/pull/3920/files#r379677001).

packages/transformers/postcss/src/loadConfig.js
```diff
@@ -0,0 +1,149 @@
...
+  // Load the custom config...
+  let modules;
+  let configPlugins = configFile.plugins;
+  if (
+    configPlugins != null &&
+    typeof configPlugins === 'object' &&
+    configPlugins['postcss-modules'] != null
+  ) {
+    modules = configPlugins['postcss-modules'];
+    // $FlowFixMe
```
padmaia: I don't think this is necessary anymore, when I removed it I didn't get any Flow errors.

packages/transformers/postcss/src/loadConfig.js
padmaia: I don't think I noticed that. I just copied `getConfig` from the Asset class, maybe something broke along the way.

packages/transformers/postcss/src/loadConfig.js
```diff
@@ -0,0 +1,149 @@
...
+  }
+
+  return configHydrator(configFile, config, options);
+}
+
+export function preSerialize(config: Config) {
+  if (!config.result) return;
+
+  if (config.result.raw.__contains_functions) {
+    // $FlowFixMe
```
padmaia: I think the flow errors you are suppressing here are actually useful. The variable `config` here is an instance of a public class, which has an underlying config that is not actually being mutated here.

packages/transformers/postcss/src/loadConfig.js
```diff
@@ -0,0 +1,149 @@
...
+    config.result = {};
+  } else {
+    // $FlowFixMe
```
padmaia: Here too

packages/transformers/postcss/src/loadConfig.js
DeMoorJasper: The util seems useful to me, would be nice if serialisation would just be wrapped in a `try {} catch {}` or something and on error (aka cannot be serialised) calls the `shouldReload()` automatically.

It would also significantly simplify this part of the code and be a lot easier to setup on even complexer config types.

packages/transformers/postcss/src/loadConfig.js
DeMoorJasper: Alright, thanks for clarifying

packages/transformers/postcss/src/loadConfig.js
DeMoorJasper: config.setResult didn't work, what is the correct way to do this?

PR #3925, opened 2019-12-17
Disregard order in dev deps when creating cache keys

padmaia:
# ↪️ Pull Request
I noticed some transformation cache misses on a repo that had not changed and it turned out to be  that Babel's `loadPartialConfig().options.presets` and `loadPartialConfig().options.plugins` does not always return the same order of presets/plugins for the same config. While plugin/preset order does matter in the config, this should be taken care of with the `resultHash` and the order in which `devDeps` are defined should not matter. Since `devDeps` is a Map which is converted into an array for `md5FromObject`, the arrays and therefore the hash are sometimes different. This PR converts the map into an object instead and since the keys and values are the same, `md5FromObject` returns the same hash.

## 🚨 Test instructions
Before change, noted transformation cache misses. After change, noted zero transformation cache misses.


parcel-benchmark:
## Benchmark Results

#### Timings

| --- | --- | --- |
| Cold | 10.78s | -1.14s 🚀 |
| Cached | 7.96s | -9.00ms |

#### Cold Bundles

| dist/legacy/index.js | 243.00b | +0.00b | 532.00ms | -63.00ms 🚀 |
| dist/modern/index.js | 215.00b | +0.00b | 526.00ms | -62.00ms 🚀 |
| dist/modern/index.css | 29.00b | +0.00b | 1.60s | +189.00ms ⚠️ |

#### Cached Bundles

| dist/legacy/parcel.ff215454.webp | 102.94kb | +0.00b | 15.00ms | -2.00ms 🚀 |
| dist/modern/parcel.76ee4591.webp | 102.94kb | +0.00b | 14.00ms | -1.00ms 🚀 |
| dist/modern/index.js | 215.00b | +0.00b | 15.00ms | -2.00ms 🚀 |
| dist/legacy/index.css | 29.00b | +0.00b | 14.00ms | -1.00ms 🚀 |

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 16.16s | -284.00ms |
| Cached | 8.13s | -88.00ms |

#### Cold Bundles

*No bundle changes detected.*

#### Cached Bundles

*No bundle changes detected.*

</p></details>

#### Timings

| --- | --- | --- |

#### Cold Bundles

#### Cached Bundles

</p></details>

**[Click here to view a detailed benchmark overview.](https://benchmark.parceljs.org/benchmark/252032247835656713)**

packages/core/core/src/Transformation.js
```diff
@@ -607,8 +607,20 @@ function normalizeAssets(
...
+  let impactfulConfigInfo = {};
+
+  for (let [configType, {devDeps, resultHash}] of configs) {
+    let devDepsObject = {};
+
+    for (let [moduleName, version] of devDeps) {
+      devDepsObject[moduleName] = version;
+    }
+
+    impactfulConfigInfo[configType] = {
```
mischnic: `configType` is the plugin name here. Could a single transformer run multiple times in a pipeline (meaning something could get overwritten)?

padmaia: We load config at the beginning of loading a pipeline, so even if a transformer were defined twice, the inputs would be the same, so I don't think this is something we need to worry about.

PR #2882, opened 2019-04-02
Add Markdown support — v2

kirillrogovoy:
# ↪️ Markdown support — v2

Implements what originally was asked in https://github.com/parcel-bundler/parcel/issues/2274.

New features:
- [Front matter](https://www.npmjs.com/package/front-matter) support
- Templating support via [Mustache](https://github.com/janl/mustache.js/) templates
- Specifying options for Marked via `marked.config.js`
- Specifying options for Marked via Front matter attributes

**See integration tests for examples.**

I'll be happy to write the docs for Markdown support once we settle on these changes.

## ✔️ PR Todo

- [x] Added/updated unit tests for this change
- [x] Filled out test instructions (In case there aren't any unit tests)
- [x] Included links to related issues/PRs

PR #3384, opened 2019-08-14
WIP: Add Markdown Plugin

JosephMart:
# ↪️ Pull Request

This PR tries to add markdown plugin for transforms. Fixes #3357

## ✔️ PR Todo

- [ ] Added/updated unit tests for this change
- [ ] Filled out test instructions (In case there aren't any unit tests)
- [ ] Included links to related issues/PRs

PR #3936, opened 2019-12-22
[T-290] Markdown transformer

### LINKED ISSUE #3357: Parcel 2: Markdown transformer
Create the `@parcel/transformer-markdown` package in `packages/transforms/markdown`. This package is a Parcel 2 transform plugin for markdown. It should be based on the `MarkdownAsset` from Parcel 1.

The `@parcel/transformer-markdown` package is responsible for the following things:

* Compiling markdown to HTML using `marked`

DeMoorJasper:
# ↪️ Pull Request

This PR adds a markdown transformer for Parcel 2, the aim of this PR is to get parcel to become a fully functional static site generator from markdown files, by allowing specification of a layout file and metadata through frontmatter.

We can probably use this to generate the Parcel 2 docs site?

For code highlighting this uses highlight.js and for markdown parsing it uses marked.

Loosely based on #2882 

Closes #3357

Close T-290

## ✔️ PR Todo


parcel-benchmark:
## Benchmark Results

#### Timings

| --- | --- | --- |
| Cold | 7.42s | +409.00ms ⚠️ |
| Cached | 3.93s | -3.00ms |

#### Cold Bundles

| dist/legacy/parcel.ff215454.webp | 102.94kb | +0.00b | 141.00ms | -71.00ms 🚀 |
| dist/modern/parcel.76ee4591.webp | 102.94kb | +0.00b | 188.00ms | -20.00ms 🚀 |
| dist/legacy/index.js | 252.00b | +0.00b | 436.00ms | -50.00ms 🚀 |

#### Cached Bundles

| dist/legacy/parcel.ff215454.webp | 102.94kb | +0.00b | 381.00ms | +367.00ms ⚠️ |
| dist/modern/parcel.76ee4591.webp | 102.94kb | +0.00b | 380.00ms | +368.00ms ⚠️ |
| dist/legacy/index.js | 252.00b | +0.00b | 385.00ms | +367.00ms ⚠️ |
| dist/modern/index.js | 215.00b | +0.00b | 381.00ms | +368.00ms ⚠️ |
| dist/legacy/index.css | 29.00b | +0.00b | 380.00ms | +368.00ms ⚠️ |
| dist/modern/index.css | 29.00b | +0.00b | 378.00ms | +367.00ms ⚠️ |

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 11.53s | -0.00ms |
| Cached | 3.93s | -162.00ms |

#### Cold Bundles

*No bundle changes detected.*

#### Cached Bundles

| dist/module/index.js | 34.29kb | +0.00b | 10.00ms | -377.00ms 🚀 |
| dist/main/index.js | 34.27kb | +0.00b | 15.00ms | -378.00ms 🚀 |

</p></details>

#### Timings

| --- | --- | --- |

#### Cold Bundles

#### Cached Bundles

</p></details>

**[Click here to view a detailed benchmark overview.](https://benchmark.parceljs.org/benchmark/252468048336781825)**

packages/transformers/markdown/src/MarkedTransformer.js
```diff
@@ -0,0 +1,60 @@
...
+      breaks: false,
+      sanitize: false,
+      smartLists: true,
+      smartypants: false,
+      xhtml: false,
+    });
+
+    let template = DEFAULT_TEMPLATE;
+    if (attributes.template) {
+      let templateLocation = await resolve(asset.filePath, attributes.template);
```
DeMoorJasper: This should probably also change imports to be relative to the markdown file location as otherwise imports will break

packages/transformers/markdown/src/MarkedTransformer.js
```diff
@@ -0,0 +1,60 @@
...
+      template = await options.inputFS.readFile(templateLocation, 'utf-8');
+      asset.addIncludedFile({
+        filePath: templateLocation,
+      });
+    }
+
+    let res = Mustache.render(template, attributes);
```
DeMoorJasper: This should probably also change imports to be relative to the markdown file location as otherwise imports will break

packages/transformers/markdown/src/MarkedTransformer.js
```diff
@@ -0,0 +1,60 @@
...
+import hljs from 'highlight.js';
+import fm from 'front-matter';
+import Mustache from 'mustache';
+
+const DEFAULT_TEMPLATE = `<!DOCTYPE html>
+<html lang="en">
+<head>
+  <meta charset="UTF-8">
+  <meta name="viewport" content="width=device-width, initial-scale=1.0">
+  <meta http-equiv="X-UA-Compatible" content="ie=edge">
```
DeMoorJasper: Is there a way to include css from node_modules into html in Parcel 2?

Either wanna import a css file that has some basic styling for the markdown or just import highlight.js's github theme. For both I'd need that functionality, I can always hack it by inlining it but would prefer to do it properly

mischnic:
Related: https://github.com/parcel-bundler/parcel/pull/3384, (https://github.com/parcel-bundler/parcel/pull/3262)

packages/examples/markdown/src/template.html
```diff
@@ -0,0 +1,12 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <meta charset="UTF-8">
+  <meta name="viewport" content="width=device-width, initial-scale=1.0">
+  <meta http-equiv="X-UA-Compatible" content="ie=edge">
+  <title>CUSTOM TEMPLATE: {{title}}</title>
```
mischnic: Nit: the hightlight.js stylesheet needs to be loaded

DeMoorJasper: Do we still support loading css from node_modules in html? Couldn't get it to work

mischnic: Well `~/node_modules/..` should work. But that wouldn't work anymore with PnP

DeMoorJasper: That's unfortunate we should probably support node modules and update the example after that. Maybe like css `~node-module-name/...`

packages/transformers/markdown/src/MarkedTransformer.js
mischnic: The only (worse?) alternative would be to process the two assets separately and only then merge them in a packager...

DeMoorJasper: That would perhaps be faster but not entirely sure how that would work as the packager is the htmlpackager and not sure if we should add markdown stuff to it? I haven't written any packager stuff for parcel 2 yet so I might be wrong here

mischnic: I would do it like this:
The markdown transformer adds the html template as a dependency and then keeps the asset type at `md`.
Then a md packager takes the html dependency of the markdown asset (from the BundleGraph) and merges them.
This way, the template is also only transformed once.

(Maybe the bundler plugin needs to know about the md-html relationship as well.)
(The HTML transformer needs to be configure to not destroy the template tags.)

DeMoorJasper: sounds doable I'll look into it tomorrow

packages/transformers/markdown/src/MarkedTransformer.js
```diff
@@ -0,0 +1,49 @@
+// @flow
+import {Transformer} from '@parcel/plugin';
+import {promisify} from '@parcel/utils';
+import marked from 'marked';
```
devongovett: We might want to look into using something like [remark](https://github.com/remarkjs/remark) for markdown since it supports all kinds of plugins for different things. For example, syntax highlighting could be done by a plugin for remark however people like without needing Parcel to build it into the transformer. Thoughts?

DeMoorJasper: Not sure, wouldn't that require a javasrcript based config?

packages/examples/markdown/src/template.html
```diff
@@ -0,0 +1,13 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <meta http-equiv="X-UA-Compatible" content="ie=edge" />
+    <title>CUSTOM TEMPLATE: {{title}}</title>
+    <link rel="stylesheet" href="style.css" />
```
DeMoorJasper: Found a workaround to import node_modules styles 🎉

packages/packagers/markdown-html/src/MarkdownHtmlTransformer.js
```diff
@@ -0,0 +1,39 @@
+// @flow
+import assert from 'assert';
+import {Packager} from '@parcel/plugin';
+import {replaceURLReferences} from '@parcel/utils';
+import Mustache from 'mustache';
+
+export default new Packager({
```
DeMoorJasper: @mischnic I'm not entirely sure how you saw this? How would this ever output html with the html extension?

DeMoorJasper: Also there's no way I'm aware of the get the result of another packager from within a packager. As html assets get replaced in packaging, mainly inline assets?

mischnic: I somehow mixed up transformer and packager and thought that a packager could change the type 🙈 

Maybe the named pipeline system should be extended to packagers as well?

DeMoorJasper: Not sure @devongovett any ideas?

bradenhs:
The ability to use parcel 2 as a static site generator with code highlighting built in would be awesome! What's the status of this PR?

DeMoorJasper:
@bradenhs it's ready for a merge just awaiting a review

mischnic:
- Using a language (i.e. `tsx` that is supported with GitHub) fails at buildtime. I think we should rather fallback to `plaintext` like in the marked docs: https://marked.js.org/#/USING_ADVANCED.md

- Should the dependencies (especially highlight.js and marked) be installed on demand instead (`packageManager.require`)? They are quite large given that most users wouldn't process markdown anyway.

- Relative paths in the template don't work (`Failed to resolve 'style.css' from './markdown/src/sub/other.md'`)
This makes it somewhat unusable for a SSG usecase.
```
src/
├── index.md       `template: ./template.html`
├── style.css
├── sub
│   └── other.md   `template: ../template.html`
└── template.html  `<link rel="stylesheet" href="style.css" />`
```


- (More of a future roadmap: The only thing that is blocking me from using Parcel instead of eleventy is an auto-generated navigation/sitemap...)

DeMoorJasper:
@mischnic I haven't really found a way around relative paths except running posthtml over the template and rewriting all relative imports to be relative to markdown instead of the template.

I agree with dynamically using the markdown and highlight libraries.

Will refactor everything and hack in the relative support once Devon has reviewed this. My main question about remark vs marked is there no point in changing except for the config but that might require a js config? Which we are actively trying to prevent everywhere.

mischnic:
>  My main question about remark vs marked is there no point in changing except for the config but that might require a js config? Which we are actively trying to prevent everywhere.

True, but we could do use a `.remarkrc` just like `.posthtmlrc` (remark and PostHTML actually have the exact same node API). It seems like remark don't have an official config file either.

DeMoorJasper:
@mischnic with no point in changing that it would/should output the same as marked anyway if there's no config. So the only reason I see was the config as it's main selling point is the plugins

devongovett:
I think we need to come up with something that's extensible, or people will just run into issues pretty quickly. Remark has a great plugin ecosystem around it that people can use to extend it. We just need a way of letting people specify those. I think things like syntax highlighting should also be done in plugins too so it can be customizable. We just need a way to run remark and load plugins inside parcel.

I'm also thinking that maybe part of this should be done in a packager too. As @mischnic said, we'll need a way of generating a list of links between pages and such. If we ran all of the pages through a packager at the end, that could inject the markdown into a template or something to generate the full HTML page with the list of links, which could come from the bundle graph. We're doing something similar at work for our docs website, except using MDX and React.

I think we'd want a way to specify layouts at the page level along with other metadata like title, description, etc. This could be done using frontmatter perhaps, or however plugins wanted, and stored in `asset.meta`. Then the packager could read the metadata for bundle in the bundle graph while its generating the list of pages.

I'm not saying all that has to happen in this PR, but I think we should move in this direction. In my experience so far it's been awesome to use Parcel as a static site generator, so I'd like for it to be as extensible as possible.

DeMoorJasper:
Might recreate this PR at some point in the future, for now gonna close it

PR #3964, opened 2020-01-05
Delay dispatching of buildSuccess / buildFailure events

### LINKED ISSUE #3961: Parcel 2: CLI is sometimes inconsistent with build state
# 🐛 bug report

CLI is sometimes inconsistent with build state.

## 🤔 Expected Behavior

CLI is consistent with build state at all times.

## 😯 Current Behavior

Sometimes (rarely on first build, but very often on rebuilds) the CLI output is inconsistent with the build status, permanently showing a spinner with "Optimizing file..." after the build succeeds:

image

## 💁 Possible Solution

Investigating this, I found the root cause to be the throttled (delayed) dispatch of `buildProgress` events

https://github.com/parcel-bundler/parcel/blob/ddb253e6ce00a1709131519c349f892322b53526/packages/reporters/cli/src/UI.js#L44-L53

causing events to be dispatched out-of-order and get into an inconsistent state, because `buildProgress` events will be dispatched **after** the `buildSuccess` one.

Is there any particular reason for that throttling?

If no, I tested removing it - fixes the issue.

If yes, I tested delaying the dispatch of `buildSuccess` / `buildFailure` events with 200ms (> the throttling interval of 100ms) - fixes the issue.

## 🌍 Your Environment

| Parcel           | v2 branch |

lbogdan:
# ↪️ Pull Request

Fixes #3961 , see that for details.

## ✔️ PR Todo


lbogdan:
What do you mean by *normal logs*?

DeMoorJasper:
The `log` event, maybe it should also be delayed to maintain the correct order, although it will be less noticeable. Maybe just delay everything idk

lbogdan:
The issue with `buildSuccess/Failure` events is that if they're out-of-order, this eventually leads to an inconsistent state. Out of order log events won't, and probably not be noticeable.

PR #3887, opened 2019-12-07
Ignore: `url(#...)` (`behavior` property)

mischnic:
# ↪️ Pull Request

Some libraries still include these statements in their stylesheets (see issue).

Closes #3873

**Should `url(#foo)` only be allowed in a `behavior` statement or everywhere?**

## 💻 Examples

`behavior` is non standard (IE)

```css
.lvml {
	behavior: url(#default#VML);
	display: inline-block;
	position: absolute;
}
```

PR #3968, opened 2020-01-07
fix: Improve error message for URLs without pathname

pke:
Prints out the `filePath` that is missing the `pathname` part/component.

# ↪️ Pull Request

I was getting this error in the output.
I also now know where it was coming: leaflet.css and the error itself is fixed in #3887
However I still believe the error reporting could be improved more. We should also print out the name of the faulty asset (leaflet.css) and where it was referenced in that case.
What do you think?

```
🚨 Error: Received URL without a pathname.
Error: Received URL without a pathname.
    at ResolverRunner.resolve (/var/node_modules/@parcel/core/lib/ResolverRunner.js:80:15)
    at async DepPathRequestRunner.runRequest (/var/node_modules/@parcel/core/lib/RequestTracker.js:356:43)
    at async Promise.all (index 19)
```

Which was not very helpful, cause it omitted the URL.

## ✔️ PR Todo

~~Since I don't know what causes the error, I can't provide any regression tests. Maybe someone can give me some hints.~~

What would be a good test for this?


pke:
Any ideas why its failing on Windows?

pke:
@DeMoorJasper did you find out why the tests failed on Windows 10?

DeMoorJasper:
@pke tests are just flaky sometimes

PR #3990, opened 2020-01-12
Fix bundle imports of reexports

### LINKED ISSUE #3832: Linker optimization invalidates bundle graph and breaks bundle imports
# 🐛 bug report

## 😯 Current Behavior

`The requested module './3816-async-import-hoist.81448d5a.js' does not provide an export named '$c6824872617571f6713b4c469c53c9$exports'`

## 🖊 Reason

There is some common asset used by the main and async bundle that is just a reexport:
`(i18n) --reexport--> T.jsx`.

The async bundle uses `T.jsx` instead of the reexporting asset `i18n` (pink in the graphic). However, this isn't reflected in the bundle graph which determines which asset will be imported from the other bundle (this is still `i18n` instead of `T.jsx`).

Bildschirmfoto 2019-11-22 um 11 03 29

https://github.com/parcel-bundler/parcel/blob/cf3e84cbfc5d3f4a18da9fdcdacaae96c75b5550/packages/shared/scope-hoisting/src/link.js#L117-L128

## 💻 Code Sample

https://github.com/mischnic/3816-async-import-hoist

## 🌍 Your Environment

| Parcel           |v2

mischnic:
# ↪️ Pull Request

Bundle imports (to other bundles) were previously not importing the asset that is exported in the other bundle, but the asset that does to actual export (so reexporting assets were incorrectly skipped for this). This meant that import and exports didn't match up.

Fixes #3832, see for more details.


packages/core/integration-tests/test/output-formats.js
```diff
@@ -695,7 +695,33 @@ describe('output formats', function() {
       );
     });
 
-    it('should generating ESM from CommonJS', async function() {
+    it.only('should create correct bundle import for reexports', async function() {
```
garthenweb: Should we remove the `only` to execute all tests again?

mischnic: 😬

PR #3996, opened 2020-01-13
Add --serve-index option for v2

hanabi1224:
Fix https://github.com/parcel-bundler/parcel/issues/1315 for v2


packages/core/parcel-bundler/src/Bundler.js
```diff
@@ -154,6 +154,8 @@ class Bundler extends EventEmitter {
           : isProduction,
       throwErrors:
         typeof options.throwErrors === 'boolean' ? options.throwErrors : true,
+      serveIndex:
```
DeMoorJasper: I don't think this should be an option, seems logical to me that it would return the `index.html` file if there is one.

hanabi1224: Resolved

PR #4055, opened 2020-01-29
Remove default url importing behavior for JS and add named pipeline

devongovett:
In Parcel 1, we allowed importing a URL to any file in your JavaScript by default. This meant that it was possible to import e.g. images and use the returned url at runtime. This was useful, and similar to what webpack does via `file-loader`, but caused some problems. Most importantly, it made it impossible for us to ever add support for a new file format without technically causing a breaking change. For example, if we wanted to add an SVG transformer that returned JSX, we could not without breaking existing apps that relied on it returning a URL.

In Parcel 2, we will require the use of a `url:` protocol in the import specifier to import a type of file that is not explicitly supported in `.parcelrc` (or the default config).

```js
import logoUrl from 'url:./logo.png';
```

This will allow us to add support for new file types in the future without causing a breaking change. Note that this only applies to JS (and compile-to-js langs), not languages where you are explicitly referencing a URL (e.g HTML, CSS, etc.). So HTML `` and CSS `url()` will continue to work as they do today, without the `url:` prefix.

This is a pretty big breaking change from Parcel 1, and might be controversial. However, I think it will be more maintainable long term and Parcel 2 is an opportunity to do it.


AshotN:
Wouldn't have also worked out if instead of explicitly asking for a URL, you explicitly ask for JSX instead. 

It's just rather ugly to write `url:` constantly.

PR #4098, opened 2020-02-07
Resolve targets from project root unless entries point at package directories

devongovett:
Parcel currently has support for building multiple packages in a monorepo. For each entry, we resolve the targets in the nearest package.json. This works great, but sometimes there are cases where we might want to perform a single build across packages instead, rather than building packages separately. For example, you might have documentation for each package that you wish to compile into a single website rather than separate ones for each package.

This changes the target resolution algorithm to use the targets from the monorepo root unless the entry is a package directory. So, if you did `parcel build packages/*` it would resolve entries and targets from the package.json files in each package directory as before, but if you did `parcel build packages/*/docs/*.md` it would resolve targets from the project root and create a single build across packages instead.

For non-monorepos, this should have no effect.


devongovett:
The added `yarn.lock` files in the tests are to simulate a project root, so that parcel doesn't search outside the test for targets.

packages/core/utils/src/getRootDir.js
```diff
@@ -31,3 +33,19 @@ export default function getRootDir(files: Array<FilePath>): FilePath {
 
   return cur ? cur.dir : process.cwd();
 }
+
+// Transforms a path like `packages/*/src/index.js` to the root of the glob, `packages/`
+function findGlobRoot(dir: FilePath) {
+  let parts = dir.split(path.sep);
+  let i = parts.length - 1;
+  let last = parts.length;
+  while (i >= 0) {
```
wbinnssmith: `for(let i = parts.length - 1; i >= 0; i--) {`?

Issue #4105, opened 2020-02-10
Issue with Parcel 2 alpha and custom transformer


green-arrow:
# 🐛 bug report

## 🎛 Configuration (.babelrc, package.json, cli command)

Basic `.parcelrc` with a custom transformer fails. I'm working on creating a transformer for SVG (similar to the svgr parcel plugin for Parcel v1) in order to fully test out the alpha.

`.parcelrc`

```json
{
  "extends": ["@parcel/config-default"],
  "transforms": {
    "*.svg": ["parcel-transformer-react-svg"]
  }
}
```

`package.json`

```json
  "alias": {
    "parcel-transformer-svg": "./parcel-transformer-react-svg.js"
  },
```

## 🤔 Expected Behavior

Parcel runs the custom transformer

## 😯 Current Behavior

I get an error message that seems to originate from the default bundler:

```
🚨 @parcel/bundler-default: The expression evaluated to a falsy value:

  (0, _assert.default)(context != null)

AssertionError [ERR_ASSERTION]: The expression evaluated to a falsy value:

  (0, _assert.default)(context != null)

    at enter (/Users/andrewwa/Developer/workspace/nova/node_modules/@parcel/bundler-default/lib/DefaultBundler.js:84:29)
    at enter (/Users/andrewwa/Developer/workspace/nova/node_modules/parcel/node_modules/@parcel/core/lib/Graph.js:503:16)
    at walk (/Users/andrewwa/Developer/workspace/nova/node_modules/parcel/node_modules/@parcel/core/lib/Graph.js:338:26)
    at walk (/Users/andrewwa/Developer/workspace/nova/node_modules/parcel/node_modules/@parcel/core/lib/Graph.js:359:22)
    at walk (/Users/andrewwa/Developer/workspace/nova/node_modules/parcel/node_modules/@parcel/core/lib/Graph.js:359:22)
    at walk (/Users/andrewwa/Developer/workspace/nova/node_modules/parcel/node_modules/@parcel/core/lib/Graph.js:359:22)
    at Graph.dfs (/Users/andrewwa/Developer/workspace/nova/node_modules/parcel/node_modules/@parcel/core/lib/Graph.js:383:12)
    at Graph.traverse (/Users/andrewwa/Developer/workspace/nova/node_modules/parcel/node_modules/@parcel/core/lib/Graph.js:287:17)
[...]

Issue #4104, opened 2020-02-10
Error messages are useless without info and stacktrace


ranisalt:
# 🐛 bug report

Currently, the error messages (at least the ones I met) provide no helpful information on fixing the problem. Example:

```
Error: Asset not found
    at BundleGraph.getDependencies (/path/to/node_modules/@parcel/core/lib/BundleGraph.js:187:13)
    at BundleGraph.getDependencies (/path/to/node_modules/@parcel/core/lib/public/BundleGraph.js:104:48)
    at /path/to/node_modules/@parcel/reporter-dev-server/lib/HMRServer.js:79:44
    at Array.map (<anonymous>)
    at HMRServer.emitUpdate (/path/to/node_modules/@parcel/reporter-dev-server/lib/HMRServer.js:78:50)
    at Object.report (/path/to/node_modules/@parcel/reporter-dev-server/lib/ServerReporter.js:94:21)
    at ReporterRunner.report (/path/to/node_modules/@parcel/core/lib/ReporterRunner.js:69:31)
error Command failed with exit code 1.
```
With this error message, it is impossible to know what asset was not found and what is the import path to this asset error so that I can go there and fix it.

## 🎛 Configuration (.babelrc, package.json, cli command)
Using out of the box, run `parcel serve index.html`

## 🤔 Expected Behavior

It should at least be helpful showing the path where the issue happened and other related information, in the above case I expect at least the trace up until the import point of the missing file and what file is missing

## 😯 Current Behavior

It shows nothing and to the user it looks like a bug inside parcel instead

## 💻 Code Sample

I can't provide a sample as I don't know where the issue is happening.

## 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Parcel           | 2.0.0-alpha.3.2 |
| Node             | 13.8.0 |
| npm/Yarn         | 1.22.0 |
| Operating System | Arch Linux |

ranisalt:
Here's another example that's unhelpful:
```
🚨 Build failed.
@parcel/packager-js: External modules are not supported when building for browser
Error: External modules are not supported when building for browser
    at Object.generateExternalImport (/path/to/node_modules/@parcel/scope-hoisting/lib/formats/global.js:50:9)
    at exit (/path/to/node_modules/@parcel/scope-hoisting/lib/link.js:559:23)
[...]

PR #4108, opened 2020-02-11
Parcel v2: SVGO + svg to react transformers

green-arrow:
# ↪️ Pull Request

In Parcel v1, I used `parcel-plugin-svgr` to generate react components out of `.svg` files. I thought it might be nice to provide a native transformer for this in Parcel v2.

Related issue: #4105 

## 💻 Examples

Add the relevant transforms to your `.parcelrc`:

```json
{
  "transformers": {
    "*.svg": ["@parcel/transformer-svgo", "@parcel/transformer-svg-react", "@parcel/transformer-js"]
  }
}
```

## 🚨 Test instructions

In a react application, add the transformer to the `.parcelrc`, import an SVG file, and render it as a component:

```jsx
import MyLogo from './Logo.svg';

export default function() {
  return (<header>
    <MyLogo title="Any react SVG props can be passed" />
  </header>);
}
```

## ✔️ PR Todo


green-arrow:
I'm more than happy to write any tests needed if someone can point me in the right direction. My thought was to just put a test under `core/integration/svg-react` and make sure a basic SVG-to-react use case works.

Also, I'm not able to completely test the output of the final bundle on my app because I get this error at the end:

```
 @parcel/runtime-js: bundleGraph.getBundleGroupsReferencedByBundle(...) is not a function or its retur
n value is not iterable
TypeError: bundleGraph.getBundleGroupsReferencedByBundle(...) is not a function or its return value is n
ot iterable
    at Object.apply (/Users/andrewwa/Developer/workspace/nova/node_modules/@parcel/runtime-js/lib/JSRunt
ime.js:76:22)
    at applyRuntimes (/Users/andrewwa/Code/parcel/packages/core/core/lib/applyRuntimes.js:46:44)
```

Any help with that would also be super appreciated 😄

packages/transformers/svg-react/src/SVGRTransformer.js
```diff
@@ -0,0 +1,77 @@
...
+  let pascalCaseFileName = camelcase(name, {
+    pascalCase: true,
+  });
+
+  return `Svg${pascalCaseFileName}`;
+}
+
+const [parcelPluginSymbol] = Object.getOwnPropertySymbols(JSTransformer);
+
+JSTransformer[parcelPluginSymbol].getConfig = async function({asset}) {
```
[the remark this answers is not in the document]
green-arrow: @DeMoorJasper - My goal was to re-use the logic in `JSTransformer`, since ideally the only difference is this transformer does a "pre-transform" to get it in a react component state. Other than that, it should go through the normal JS transformation. The reason I wanted to do this was to preserve the babel transforms, since I know differential bundling is a key feature of parcel 2.

Is what I said above accurate? If so, is the right approach to somehow extend `JSTransformer`? If not, should I just duplicate the necessary logic from `JSTransformer` in this one?

packages/transformers/svg-react/src/SVGRTransformer.js
```diff
@@ -0,0 +1,77 @@
...
+    ],
+    {
+      packageKey: 'svgo',
+    },
+  );
+
+  svgoConfig = svgoConfig || {};
+  svgoConfig = {...defaultConfig, ...svgoConfig};
+
+  this.svgo = new SVGO(svgoConfig);
```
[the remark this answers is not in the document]
green-arrow: @DeMoorJasper - that's a great point. I'll pull this out into it's own `@parcel/transformer-svgo` if that works

packages/transformers/svg-react/src/SVGRTransformer.js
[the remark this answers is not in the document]
DeMoorJasper: In Parcel 2 we have a concept called pipelines so in theory there would be no need to extend any transformer as you can let another transformer run before or after this transformer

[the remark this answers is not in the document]
green-arrow: @DeMoorJasper - I updated the transformer to just generate the react code and not run any JS transforms. In my application, I can then update my `.parcelrc` to run it through both transformers:

```json
{
  "transforms": {
    "*.svg": ["@parcel/transformer-svgo", "@parcel/transformer-svg-react", "@parcel/transformer-js"]
  }
}
```

Is there any way to specify a "follow-up" transformer? For SVG to react, it seems to make sense to always run the result of this transformer through the one that handles JS, so it is transformed for the target environment properly. Thoughts?

green-arrow:
@DeMoorJasper - Recent commit addresses your PR comments.

I'll look into writing the integration test today.

It looks like the failing test is also failing on `v2`, so doesn't appear related to these changes. I can definitely look into it though! Would you want a separate PR for a fix for the test or should I include it here?

packages/transformers/svg-react/src/SvgReactTransformer.js
```diff
@@ -0,0 +1,58 @@
...
+  let pascalCaseFileName = camelcase(name, {
+    pascalCase: true,
+  });
+
+  return `Svg${pascalCaseFileName}`;
+}
+
+export default new Transformer({
+  async parse({asset}) {
+    let code = await asset.getCode();
```
DeMoorJasper: If you move the logic from parse to transform you don't have to return an ast and can just use asset.setcode and return the asset wihtout needing babel at all.

That's what we've done in a couple transformers that dont' utilise ast's

packages/transformers/svgo/src/SVGOTransformer.js
```diff
@@ -0,0 +1,45 @@
...
+
+    return svgoConfig;
+  },
+
+  transform({asset}) {
+    return [asset];
+  },
+
+  async generate({asset}) {
+    let code = await asset.getCode();
```
DeMoorJasper: Maybe move this to transform and not have a generate?

packages/transformers/svgo/src/SVGOTransformer.js
```diff
@@ -0,0 +1,45 @@
...
+      ],
+      {
+        packageKey: 'svgo',
+      },
+    );
+
+    svgoConfig = svgoConfig || {};
+    svgoConfig = {...defaultConfig, ...svgoConfig};
+
+    this.svgo = new SVGO(svgoConfig);
```
DeMoorJasper: Not entirely sure if this is a good idea as it's not guaranteed that this state will be kept in other functions. It currently works but we could shift around a bit in transformrunner and break this.

Probably best to initialise svgo where you actually use it and keep the config to the purest JSON form possible. This will also help us improve this plugin for better caching in the future :)

DeMoorJasper:
@green-arrow v2 appears to be stable not sure why it is failing in this PR and not in v2.

packages/transformers/svgo/src/SVGOTransformer.js
green-arrow: For some reason I thought `generate` was required 😛

green-arrow:
@DeMoorJasper - Addressed your feedback in the last commit

Maybe my fork `v2` branch is not up to date (it failed when I ran that integration test locally). I'll give it another go.

green-arrow:
@DeMoorJasper - I synced my fork with the parcel `v2` branch, and still receive the same error for that integration test when I run it locally.

DeMoorJasper:
Looks great. I'll have a look at the test tomorrow, strange that it fails in here and nowhere else.

green-arrow:
@DeMoorJasper - As far as I can tell, something weird is happening with that `matchesEngine` call and I'm not sure exactly what. Something that seems weird, though, is that the `index.html` file used for the test contains the script entry as follows:

```html
<script type="module" src="./index.js"></script>
```

In this case, it seems like the target should be browsers that support modules, which would mean no polyfill necessary. So it seems like the output might be correct, but the test is just using the wrong input file?

I guess the question is, given this type of entry point (HTML with type="module" in the script), along with a target browser query of an older browser (like IE 11), what should happen? I'm not sure I know what the correct answer to that question is, but I think it's the root cause of why this particular test is broken.

DeMoorJasper:
@green-arrow seems rebuilding the yarn lockfile and using the current version of parcel in transforms solved the issue, not entirely sure why but it works so no need to dig deeper into it for now.

green-arrow:
@DeMoorJasper - Latest commit contains an integration test for `svgo` and `svg-react` transforms. Let me know if you'd like me to add anything else!

packages/core/integration-tests/test/svg.js
```diff
@@ -0,0 +1,39 @@
...
+import path from 'path';
+
+const config = {
+  ...defaultConfigContents,
+  transforms: {
+    ...defaultConfigContents.transforms,
+    '*.svg': [
+      '@parcel/transformer-svgo',
+      '@parcel/transformer-svg-react',
+      '@parcel/transformer-js',
```
DeMoorJasper: replace this with `...` than it will run the Javascript pipeline instead :)

DeMoorJasper: I just did it myself, hopefully everything works. So we can ship this 🚀

green-arrow: @DeMoorJasper - haha whoops I did it too and didn't realize.

Just curious, what is that magic of `...`?

DeMoorJasper: It's based on the spread operator, it tells parcel to look for any additional pipelines if there is any. and place them at the position of the `...` in this case after svgo and svg-react.

I think this is actually no necessary and will automatically run through the js pipeline without the `...`.

See the PR https://github.com/parcel-bundler/parcel/pull/2568 for more info

mischnic: `...` means "continue with the pipeline declared in the `extend`ed config".

In this case, doing `asset.type = "js";` will switch to the `js` pipeline either way, so `...` would be ignored anyway

green-arrow:
Hi @DeMoorJasper! I was just wondering when the next alpha release will be that contains this change. When I have it linked and running locally my build passes, but fails at the bundling stage with the same error from [this comment](https://github.com/parcel-bundler/parcel/issues/4104#issuecomment-584340054). However, the fix there of deleting the cache and re-running does not work for me. I can consistently `rm -rf .parcel-cache && yarn start` and reproduce the error.

I was hoping to get an alpha build to rule out anything weird from my local linked install. Thoughts?

DeMoorJasper:
@green-arrow you can use the nightly builds of parcel `yarn add parcel@nightly`

Not entirely sure when we'll release a new alpha, we're probably gonna try and get some caching stability and other things before moving to a new alpha (or beta)

CrashyBang:
Out of curiosity are all 3 `["@parcel/transformer-svgo", "@parcel/transformer-svg-react", "@parcel/transformer-js"]` packages required?

Cheers.

mischnic:
- svgo: compress (optional)
- svg-react: transform svg into JS exporting a react component (required)
- js: collect dependencies (required)

Your config should look like this:
```json
{
  extends: ['@parcel/config-default'],
  transforms: {
    '*.svg': ['@parcel/transformer-svgo', '@parcel/transformer-svg-react']
  }
}
```
The js transformer will automatically be called.

CrashyBang:
Awsome, so in theory:

```
{
  extends: ['@parcel/config-default'],
  transforms: {
    '*.svg': ['@parcel/transformer-svg-react']
  }
}
```

Would work also?

green-arrow:
@CrashyBang yes, the SVGO plugin is completely optional.

fregante:
I wasn't able to make the easiest version of this work. Have a look: https://github.com/parcel-bundler/parcel/issues/5206

PR #4114, opened 2020-02-12
Bundle hashing improvements

padmaia:
# ↪️ Pull Request

Currently namers are using hashes that are only determined by the content of source assets of of the bundle and all of its child bundles.

This is not ideal for two reasons:
  1. The hashes do not change when runtime, packager, or optimizer plugins change and therefore change the output of the bundle.
  2. Hashes based on the content of all child bundles causes cascading invalidation.

**This PR introduces the idea of hash references for finalized bundle content which solves both of these problems.**

Now namers add a hash reference to the bundle name, which will later be replaced with the real bundle hash by core. The hash is not generated until all of the bundles have been run through packagers and optimizers, which solves the first problem. The bundles are still processed in parallel workers, but they are only written to cache (with the references still in place) and not to dist. While writing to cache, the content of the bundle is hashed and hash references to other bundles are recorded. Once all of the bundles have been processed, core figures out the final hashes based on the content hash of the bundle as well as the content hash of any referenced child bundles. It then streams the cached content/map to dist, replacing the hash references along the way. This allows runtime plugins to avoid the cascading invalidation problem by creating assets that contain a mapping of stable bundle ids to bundle names with hash references. These assets will only exist in one entry bundle and all other bundles will not need to update their hashes when child bundles change because they only reference the stable ids.

## ✔️ PR Todo

- [x] Stream directly from cache to dist 
- [x] Implement `getBundleInfoFromCache`
- [x] Address questions called out in comments
- [ ] Maybe fix sourcemaps (I hear they are somewhat broken already, so maybe this can wait until later)


packages/core/core/src/BundleGraph.js
```diff
@@ -574,6 +574,8 @@ export default class BundleGraph {
     return hashHex;
   }
 
+  // ? Should this be removed? Or at least renamed to signify that it incorporates content of child bundles,
+  //  - even though that may not affect the content of the output bundle
```
padmaia: Thoughts?

packages/core/core/src/PackagerRunner.js
```diff
@@ -73,121 +84,84 @@ export default class PackagerRunner {
...
-        continue;
-      }
-
-      promises.push(
-        this.writeBundle(bundle, bundleGraph, ref).then(stats => {
-          bundle.stats = stats;
+    let bundleInfoMap = {};
+    // ? hashRef should maybe just be first 8 digits of id, but that would require that id is an md5 hash already
+    //  - only reason I'm not doing that right now is that it feels like we're doing a bunch of hashes of hashes
+    //  - not sure if that's a bad thing or not
```
padmaia: Thoughts?

packages/core/core/src/PackagerRunner.js
```diff
@@ -73,121 +84,84 @@ export default class PackagerRunner {
...
-      cacheKey,
-    );
-    let {size} = await this.writeToDist({
-      bundle,
-      bundleGraph,
-      contents,
-      map,
-    });
+    let info = {};
+    // ? Why would cacheKey be null
```
padmaia: Was this maybe for inline bundles?

parcel-benchmark:
## Benchmark Results

#### Timings

| --- | --- | --- |
| Cold | 3.51s | +16.00ms |
| Cached | 3.86s | +1.05s ⚠️ |

#### Cold Bundles

| dist/parcel.a2ee7a0b.webp | 102.94kb | -5.25kb 🚀 | 97.00ms | +4.00ms |
| dist/index.js | 1.07kb | +5.00b ⚠️ | 208.00ms | -10.00ms |

#### Cached Bundles

| dist/parcel.a2ee7a0b.webp | 102.94kb | -5.25kb 🚀 | 960.00ms | +949.00ms ⚠️ |
| dist/index.js | 1.07kb | +5.00b ⚠️ | 1.08s | +1.07s ⚠️ |

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 7.26s | -216.00ms |
| Cached | 4.79s | +1.53s ⚠️ |

#### Cold Bundles

| dist/index.js | 64.04kb | +397.00b ⚠️ | 1.03s | -81.00ms 🚀 |

#### Cached Bundles

| dist/index.js | 64.04kb | +393.00b ⚠️ | 2.01s | +1.66s ⚠️ |

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 3.53s | -195.00ms 🚀 |
| Cached | 3.79s | +959.00ms ⚠️ |

#### Cold Bundles

| dist/index.js | 743.00b | +7.00b ⚠️ | 165.00ms | -7.00ms |

#### Cached Bundles

| dist/index.js | 743.00b | +7.00b ⚠️ | 1.07s | +1.05s ⚠️ |

</p></details>

**[Click here to view a detailed benchmark overview.](https://benchmark.parceljs.org/benchmark/257132276276527626)**

packages/core/core/src/PackagerRunner.js
```diff
@@ -430,52 +405,108 @@ export default class PackagerRunner {
...
 ): Promise<number> {
-  return new Promise((resolve, reject) => {
-    let fsStream = fs.createWriteStream(filePath, options);
-    stream
-      .pipe(fsStream)
-      // $FlowFixMe
-      .on('finish', () => resolve(fsStream.bytesWritten))
-      .on('error', reject);
+  // TODO: stream directly to filesystem instead of to string first
+  //  - need to figure out how to replace hash references in a stream
```
padmaia: TODO: Skip stream replacement if we know there are no references to replace. We could probably start streaming them to dist immediately instead of waiting for all of the bundles to be processed to create the `hashRefToNameHash` mapping as well.

packages/core/core/src/BundleGraph.js
kwelch: Minimum renamed. If it no longer serves a purpose, I vote remove.

packages/core/core/src/PackagerRunner.js
padmaia: > Skip stream replacement if we know there are no references to replace.

Done.

> We could probably start streaming them to dist immediately instead of waiting for all of the bundles to be processed to create the hashRefToNameHash mapping as well.

Not done yet.

packages/core/core/src/PackagerRunner.js
padmaia: It was, not relevant anymore now that `getBundleInfo` is writing to cache instead of `getBundleResult`.

packages/core/core/src/PackagerRunner.js
padmaia: > We could probably start streaming them to dist immediately instead of waiting for all of the bundles to be processed to create the hashRefToNameHash mapping as well.

Done.

packages/core/core/src/BundleGraph.js
padmaia: I decided to remove `getHash` from the public bundle API as plugins should only be using hash references. Since right now only `PackagerRunner` is using it, I changed it to only consider inline bundles.

parcel-benchmark:
## Benchmark Results

#### Timings

| --- | --- | --- |
| Cold | 3.30s | -1.16s 🚀 |
| Cached | 3.27s | +485.00ms ⚠️ |

#### Cold Bundles

| dist/parcel.a2ee7a0b.webp | 102.94kb | +0.00b | 2.00ms | -118.00ms 🚀 |
| dist/index.js | 1.07kb | +5.00b ⚠️ | 6.00ms | -221.00ms 🚀 |

#### Cached Bundles

| dist/parcel.a2ee7a0b.webp | 102.94kb | +0.00b | 3.00ms | -8.00ms 🚀 |
| dist/index.js | 1.07kb | +5.00b ⚠️ | 7.00ms | -9.00ms 🚀 |

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 6.15s | -1.37s 🚀 |
| Cached | 3.10s | +28.00ms |

#### Cold Bundles

| dist/index.js | 64.04kb | +401.00b ⚠️ | 5.00ms | -1.10s 🚀 |

#### Cached Bundles

| dist/index.js | 64.04kb | +397.00b ⚠️ | 280.00ms | +267.00ms ⚠️ |

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 3.43s | -250.00ms 🚀 |
| Cached | 3.18s | +389.00ms ⚠️ |

#### Cold Bundles

| dist/index.js | 743.00b | +7.00b ⚠️ | 4.00ms | -178.00ms 🚀 |

#### Cached Bundles

| dist/index.js | 743.00b | +7.00b ⚠️ | 5.00ms | -8.00ms 🚀 |

</p></details>

**[Click here to view a detailed benchmark overview.](https://benchmark.parceljs.org/benchmark/257287872383549962)**

packages/core/core/src/PackagerRunner.js
```diff
@@ -1,12 +1,11 @@
-// @flow
+// @flow strict-local
```
wbinnssmith: 🎉

packages/core/core/src/PackagerRunner.js
```diff
@@ -37,6 +42,20 @@ type Opts = {|
   report: ReportFn,
 |};
 
+type BundleInfo = {|
+  hash: string,
```
wbinnssmith: can these properties be read-only?

packages/core/core/src/PackagerRunner.js
```diff
@@ -430,58 +423,164 @@ export default class PackagerRunner {
...
+  bundleId,
+  bundleInfoMap,
+  hashRefToId,
+  included = new Set(),
+) {
+  included.add(bundleId);
+  for (let hashRef of bundleInfoMap[bundleId].hashReferences) {
+    let referencedId = hashRefToId[hashRef];
+    if (!included.has(referencedId)) {
+      for (let ref in getBundlesIncludedInHash(
```
wbinnssmith: nit: `for (let ref of Object.keys(getBundlesIncludedInHash(...`

packages/core/core/src/PackagerRunner.js
padmaia: Implemented offline feedback in [cc6fd15](https://github.com/parcel-bundler/parcel/pull/4114/commits/cc6fd158443fa3c63e8be7bf8baf0f533a9831f4)

packages/core/core/src/PackagerRunner.js
padmaia: woops, this should be a `for...of` loop

padmaia: It doesn't really need to be a `Set` anymore since I'm checking to see if the referenced id is already in the collection before recursing. I'll change it to an array and instead do `included.push(...getBundlesIncludedInHash())`

padmaia:
@DeMoorJasper The only todo item left is "maybe fix sourcemaps". Could you help me understand how this change might break sourcemaps and what work would be necessary to fix them? Do you consider it a blocker for this PR?

DeMoorJasper:
@padmaia it depends, it will mainly break for minified builds as column indexes will become invalid.
We'd have to offset the columns by the difference in filename size.

I don't think this has to be done in this PR there's probably other issues related to this currently, we should probably figure out an easy abstracted way to do this without impacting performance too much. Which is probably a relatively large change. I mainly worry about performance with this change. As minification is one line every entry needs to be offset...

devongovett:
What if we explicitly designed it so that the columns don't change. If the bundle reference is always the same length as the final bundle url it won't break source maps. One idea would be to insert a full url but with the bundle id where the hash normally would be, and replace that hash with the final hash later. That way the length wouldn't change, just the contents of the hash.

padmaia:
@devongovett That would be nice. Two things to consider with that.

1. The bundle id is a stable id, which would be useful when creating a runtime asset with a mapping of stable ids to final file paths that helps with the the cascading invalidation problem. If core ends up replacing the bundle id, the runtime asset would have to use `md5FromString('stable_id:' + bundleId)` or something. Core could probably take care of this by replacing `md5FromString('hash_reference:' + bundleId)` instead. That is unless we use a prefix. 

2. Core is looking for hash references while writing to cache and currently there is a hash reference prefix that can be searched for with the regex `\HASH_REF_\w{32}\g`. Without the prefix we would have to search for at least every child bundle's unique 32 character id considering there will potentially be an entry bundle that contains a mapping of all of the bundles in the graph. I imagine that would have a performance impact, but it's not something I've tested out. It might be okay to keep the prefix and the id used in the reference would be less than 32 characters, which increases the chance of collision, but would still be unlikely and maybe something we could detect? I'm only using the first 8 characters in the final replacement because the rest of the file name makes collision less likely. Would a 32 character hash in the name be too much? If we decide it's okay to use a prefix, at which point we're all ready losing characters from the bundle id, we could probably lose a few more, but where is the sweet spot?

Maybe we could look into these options as a follow up task?

Issue #3917, opened 2019-12-15
Parcel 2: @parcel/transformer-babel: Support for the experimental syntax 'optionalChaining' isn't currently enabled:


Banou26:
# 🐛 bug report

When trying to use a babel proposal, e.g optional chaining, Parcel 2 throws an error

## 🎛 Configuration (.babelrc, package.json, cli command)
`.babelrc`
```js
{
  "plugins": [
    "@babel/plugin-proposal-optional-chaining"
  ]
}
```

```js
{
  "private": true,
  "name": "web",
  "version": "1.0.0",
  "main": "index.js",
  "devDependencies": {
    "@babel/plugin-proposal-optional-chaining": "^7.7.5",
    "parcel": "^2.0.0-alpha.3.2"
  },
  "scripts": {
    "dev": "parcel src/index.html"
  },
  "browserslist": [
    "last 1 Chrome versions"
  ]
}
```

## 🤔 Expected Behavior

Parcel should use the current babelrc config

## 😯 Current Behavior

Parcel throws
```
× @parcel/transformer-babel: Support for the experimental syntax 'optionalChaining' isn't currently enabled
```
## 💻 Code Sample

```js
const foo = undefined?.bar
```

mischnic:
Works for me with the package.json you posted and doing a fresh `yarn install`.

```json
{
  "scripts": {
    "build": "parcel build index.js"
  },
  "devDependencies": {
    "@babel/plugin-proposal-optional-chaining": "^7.7.5",
    "parcel": "^2.0.0-alpha.3.2"
  }
}

```

Maybe some dependency in your lockfile is outdated?

---
Unrelated:

> "main": "index.js",
> "dev": "parcel src/index.html"

The input and output file type don't match up.

Banou26:
It seems the problem only occurs in my lerna monorepo, if i `yarn install && yarn dev` outside of it, everything works but once in it, it throws that error

mischnic:
Indeed

mischnic:
```
3917-babel-proposal
├── package.json // {"workspaces": ["test"]}
├── test
│   ├── .babelrc // {"plugins": ["@babel/plugin-proposal-optional-chaining"]}
│   ├── index.js
│   └── package.json
└── yarn.lock
```

The babelrc isn't used because this to babel 
[...]

PR #4132, opened 2020-02-14
Resolve .babelrc files in a monorepo.

### LINKED ISSUE #4120: Parcel2 ignores .babelrc files in monorepos.
# 🐛 bug report

If you have a yarn-workspaces-based monorepo with custom `.babelrc` configuration, parcel ignores them.

This is related to #3917 - I'm filing a separate issue because there are other ways to resolve the problem reported there (for example, by adding `@babel/plugin-proposal-optional-chaining` to the default babel config - which IMO would be a good idea regardless of how this issue is treated).

I have a project configured like this:
```
├── package.json // {"workspaces": ["app1"]}
├── app1
│   ├── .babelrc
                  /* {
                      "plugins": ["@babel/plugin-proposal-class-properties"],
                       "presets": [[ "@babel/preset-typescript", { "isTsx": true }]]
                  } */
│   ├── index.ts // uses class properties
│   └── package.json
└── yarn.lock
```

## 🤔 Expected Behavior

The `.babelrc` file in the `app1` project should be respected, and `parcel build` should succeed.

## 😯 Current Behavior

`parcel build` fails with this error:

```
🚨 Build failed.
@parcel/transformer-babel: /Users/Andrew/Projects/parcel2-monorepo-babel-bug/app1/src/index.ts: Missing class properties transform.
  1 | class MyClassWithProperties {
> 2 |   private message = "Hello from a class!";
    |   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  3 |   sayHello() {
  4 |     console.log(this.message);
  5 |   }
SyntaxError: /Users/Andrew/Projects/parcel2-monorepo-babel-bug/app1/src/index.ts: Missing class properties transform.
  1 | class MyClassWithProperties {
> 2 |   private message = "Hello from a class!";
    |   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  3 |   sayHello() {
  4 |     console.log(this.message);
  5 |   }
    at File.buildCodeFrameError (/Users/Andrew/Projects/parcel2-monorepo-babel-bug/node_modules/@babel/core/lib/transformation/file/file.js:267:12)
    at NodePath.buildCodeFrameError (/Users/Andrew/Projects/parcel2-monorepo-babel-bug/node_modules/@babel/traverse/lib/path/index.js:144:21)
    at pushBody (/Users/Andrew/Projects/parcel2-monorepo-babel-bug/node_modules/@babel/plugin-transform-classes/lib/transformClass.js:156:20)
    at buildBody (/Users/Andrew/Projects/parcel2-monorepo-babel-bug/node_modules/@babel/plugin-transform-classes/lib/transformClass.js:130:5)
    at classTransformer (/Users/Andrew/Projects/parcel2-monorepo-babel-bug/node_modules/@babel/plugin-transform-classes/lib/transformClass.js:527:5)
    at transformClass (/Users/Andrew/Projects/parcel2-monorepo-babel-bug/node_modules/@babel/plugin-transform-classes/lib/transformClass.js:563:10)
    at PluginPass.ClassExpression (/Users/Andrew/Projects/parcel2-monorepo-babel-bug/node_modules/@babel/plugin-transform-classes/lib/index.js:63:54)
    at newFn (/Users/Andrew/Projects/parcel2-monorepo-babel-bug/node_modules/@babel/traverse/lib/visitors.js:179:21)
    at NodePath._call (/Users/Andrew/Projects/parcel2-monorepo-babel-bug/node_modules/@babel/traverse/lib/path/context.js:55:20)
    at NodePath.call (/Users/Andrew/Projects/parcel2-monorepo-babel-bug/node_modules/@babel/traverse/lib/path/context.js:42:17)
```

Interestingly, I get the same error, even when I move the `.babelrc` file to the root of the monorepo.

## 💁 Possible Solution

As @mischnic pointed out [here](https://github.com/parcel-bundler/parcel/issues/3917#issuecomment-565828384), these lines return a `partialConfig` _without_ a `.babelrc

https://github.com/parcel-bundler/parcel/blob/92fd22f262eb6f301bfd00769066a8f6fb2ef1e4/packages/transformers/babel/src/config.js#L36-L40

A possible solution would be to leverage the [`babelrcRoots`](https://babeljs.io/docs/en/options#babelrcroots) option to pass in the package directory (in addition to the monorepo root that is already specified in the `root` parameter)

## 🔦 Context

I'm trying to set up a monorepo where the packages are built with parcel2, and I need to customize the babel configuration for some of the packages individually.

## 💻 Code Sample

You can reproduce the issue in `master` branch of [this repository](https://github.com/astegmaier/parcel2-monorepo-babel-bug).

There is a `working-example` branch that shows the same app (with the same `.babelrc`) building correctly in the non-monorepo case.

## 🌍 Your Environment

| Parcel           | 2.0.0-nightly.99
| Node             | 10.16.3
| Yarn         | 1.21.1
| Operating System | Mac OSX 10.15.3

astegmaier:
# ↪️ Pull Request

This PR contains a possible fix to #4120 (might also address one of the causes of #3917). The goal is to allow parcel2 to pick up configuration specified in `.babelrc` files in a monorepo.

I wanted to make sure that the problem I experienced was considered holistically (taking into account all the likely ways that babel users are used to specifying configuration). As it turned out, that was kind of complicated, so I made a detailed explanation here:

https://astegmaier.github.io/parcel2-monorepo-babel-bug/parcel-goals

I'm putting this out there for early feedback. Specifically:
1. ***Goals*** - does my [scenario list](https://astegmaier.github.io/parcel2-monorepo-babel-bug/parcel-goals#solution-scenarios) make sense?
2. ***Performance*** - in order to implement the desired behavior, I needed to determine the sub-project's directory in the process of processing each compiled file, so I can pass it to the [`babelrcRoots`](https://babeljs.io/docs/en/options#babelrcroots) property of the babel api. The code I wrote makes calls to the file system each time a file's config is desired. I wonder if this information might be cached (maybe it already is, and I didn't realize how to access it)?
3. ~~***Unit Tests*** - I would love some pointers about how/where to do this, patterns to follow, and how much of the [scenario list](https://astegmaier.github.io/parcel2-monorepo-babel-bug/parcel-goals#solution-scenarios) you think is important to unit test.~~ ***Edit (2/24): This is done.***
4. ***Documentation*** - There's a lot of nuance here, and it would be great if the final parcel2 release came with clear, complete documentation about how to get this working. Is there a place where I could put a draft (to be incorporated into the final site down the road)?

## 💻 Examples

The main goal is to make sure that Parcel2 will pick up `.babelrc` files in sub-projects of a monorepo. For example, if you have a project like this...

```
├── package.json // {"workspaces": ["app1"]}
├── app1
│   ├── .babelrc // Provides plugins/presets to allow index.js to be compiled.
│   ├── index.js // uses some advanced JS feature that Babel won't compile by default
│   └── package.json
└── yarn.lock
```
...things will work.

## 🚨 Test instructions

See [this test repo](https://github.com/astegmaier/parcel2-monorepo-babel-bug) for an environment where you can see the issue and test out the changes.

## ✔️ PR Todo

Closes  #4120


DeMoorJasper:
You goals make total sense to me, I agree that sub-projects babelrc files should be respected like the babel-cli does.

astegmaier:
I added a unit test for the most complicated scenario (i.e. the one where you have to merge a `.babelrc` from a sub-project with `babel.config.json` from the root). Although this is not 100% comprehensive, it seems likely that anything that would break this scenario would also break the less-complicated scenarios (e.g. you _only_ have `.babelrc` in the sub-project without any `babel.config.json` in the root), and it was valuable to keep the integration tests from ballooning.

@DeMoorJasper - let me know if you'd like to see any changes before the merge, or if there's any place where I can help with documentation.

packages/transformers/babel/src/config.js
```diff
@@ -29,6 +29,20 @@ export async function load(
...
+  let babelrcRoots = [options.projectRoot];
+  let packageJSONPath = await resolveConfig(
+    options.inputFS,
+    config.searchPath,
+    ['package.json'],
+  );
+  if (packageJSONPath) {
+    let packageRoot = path.dirname(packageJSONPath);
+    if (packageRoot && packageRoot !== options.projectRoot) {
+      babelrcRoots.push(packageRoot);
```
devongovett: We probably also need to track the package.json paths we searched as part of the config so it will be invalidated in the cache properly. Maybe a glob? cc. @padmaia

padmaia: Oh yeah, I guess that never got set. Easy to add though, it's just a call to `config.setWatchGlob()`.

astegmaier: I'm not sure I fully understand how the cache invalidation works - could you elaborate on what you expect to happen (but isn't already), and how I might be able to test it?

In case it's relevant, one thing to note is that the sub-project babel config might be specified in a `.babelrc`/`babelrc.json` file in the sub-package directory, or as part of the `babel` property of the `package.json` file. Presumably, we'd want to make sure that the bundle will get regenerated if any of these change (and do whatever is necessary to keep the cache clean). Given this, are there any other tweaks you think might be necessary?

PR #4141, opened 2020-02-16
Fix static class properties with scope hoisting

### LINKED ISSUE #3664: V2: Scope Hoisting Static Class Property not Working
# 🐛 bug report

When using a static class property within the class itself or outside/ in another file, the reference to its class is not renamed properly.

.babelrc

```json
{
  "plugins": [
    ["@babel/plugin-proposal-class-properties", { "loose": true }]
  ]
}
```

package.json

```json
{
  ...
  "browserslist": ["last 2 Chrome versions", "last 2 Firefox versions"],
  ...
}
```

## 🤔 Expected Behavior

Should rename static properties within classes properly.

## 😯 Current Behavior

Does not rename static property which results in `Uncaught ReferenceError: $c4ee6a119cd77881a886bf5cd2629a0$var$A is not defined` at runtime.

## 💻 Code Sample

index.js
```js
class A {
  static getTest = () => 'test';

  render() {
    return A.getTest;
  }
}

const a = new A();
a.render();
```

```
parcel build index.js
```

dist/index.js
```
!function(){class e{render(){return $c4ee6a119cd77881a886bf5cd2629a0$var$A.getTest}}e.getTest=()=>"test",(new e).render()}();
```

## 🌍 Your Environment

| Parcel           | v2 (eedc965377940034cf78a026d581f12098df724e)
| Node             | v12.8.1
| npm         | v6.11.2
| Operating System | Windows 10/ WSL

garthenweb:
# ↪️ Pull Request

Closes #3664

Using static class properties with scope hoisting is currently failing at runtime.

Even though the fix was implemented in babel, it can still happen, especially when using yarn, that an old version of the @babel/traverse is installed. Therefore I suggest updating the package to the current version which contains the fix.

I thought about adding a test case but decided against it, as it would only test the babel fix and not a feature of Parcel.

### DIFF
--- a/packages/shared/scope-hoisting/package.json
+++ b/packages/shared/scope-hoisting/package.json
@@ -19,7 +19,7 @@
     "@babel/generator": "^7.3.3",
     "@babel/parser": "^7.0.0",
     "@babel/template": "^7.2.2",
-    "@babel/traverse": "^7.2.3",
+    "@babel/traverse": "^7.8.4",
     "@babel/types": "^7.3.3",
     "@parcel/diagnostic": "^2.0.0-alpha.3.1",
     "@parcel/utils": "^2.0.0-alpha.3.1",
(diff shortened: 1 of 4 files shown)


mischnic:
@devongovett @wbinnssmith Not sure if we should rather update all Babel dependencies?

DeMoorJasper:
Gonna close this for now, we should probably get a fix for this in though

Issue #4185, opened 2020-02-19
@parcel/validator-typescript does not re-analyze files when they change.


astegmaier:
# 🐛 bug report

When you configure to use `@parcel/validator-typescript` it will work correctly and detect type errors the _first_ time you run `parcel myFile.ts`.  If you fix the type error and save the typescript file, parcel will re-build the bundle, but the type checker will still show the old error.

## 🎛 Configuration (.babelrc, package.json, cli command)
***.parcelrc:***
```js
{
    "extends": "@parcel/config-default",
    "validators": {
        "*.{ts,tsx}": ["@parcel/validator-typescript"]
    }
}
```
***package.json:***
```
{
  "name": "parcel2-typescript-validator-bug",
  "version": "1.0.0",
  "description": "A demonstration of a bug when attempting to use @parcel/validator-typescript",
  "main": "index.html",
  "repository": "https://github.com/astegmaier/parcel2-typescript-validator-bug.git",
  "author": "Andrew Stegmaier <ansteg@microsoft.com>",
  "license": "MIT",
  "private": true,
  "scripts": {
    "build": "parcel build src/index.html",
    "start": "parcel src/index.html"
  },
  "devDependencies": {
    "@parcel/validator-typescript": "2.0.0-nightly.117",
    "@types/react": "^16.9.20",
    "@types/react-dom": "^16.9.5",
    "parcel": "2.0.0-nightly.115",
    "typescript": "3.7.5"
  },
  "dependencies": {
    "react": "^16.12.0",
    "react-dom": "^16.12.0"
  }
}
```
***src/index.html:***
```html
<!DOCTYPE html>
<html lang="en">
  <head>
    <meta charset="utf-8" />
    <title>Parcel Typescript Validator Bug</title>
  </head>
  <body>
    <noscript>You need to enable JavaScript to run this app.</noscript>
    <div id="root"></div>
    <script src="./index.tsx"></script>
  </body>
</html>
```

***src/index.tsx:***
```ts
import React from "react";
import ReactDOM from "react-dom";
import App from "./App";

const message: number = "A new Message!"; // This is a type error!

ReactDOM.render(<App message={message} />, document.getElementById("root"));
```

***src/App.tsx***
```ts
import React from "react";

const App = ({ message }: { message: string }) => <div>{message}</div>;

export default App;

```

## 🤔 Expected Behavior

[...]

PR #4187, opened 2020-02-19
Fix typescript rechecking after files change.

astegmaier:
# ↪️ Pull Request

This PR contains a possible fix to #4185. The goal is to enable `@parcel/validator-typescript` to correctly re-check files when they change, instead of remaining stuck with the errors it found the first time it ran.

I'm putting it out there for early feedback. ~~Unit tests are not done yet, but I'd love any pointers to tests that need to validate similar things (e.g. that something re-ran when a file changed), or general advice.~~ ***Edit (2/24): I successfully added a unit test, and made changes to make sure it works on Windows (it initially did not).***

~~One other small nit was: I wasn't super happy with the way that `lint-staged`+`prettier` formatted the lines I added. For example, it broke apart [lines 70-72](https://github.com/parcel-bundler/parcel/pull/4187/files#diff-82caeb66e0bad40d35067e3f80f3338cR70-R72
) in an unhelpful way. I want to conform to whatever code styling guides that are defined project-wide, but I was wondering if there is some easy way to tweak this (if not, not worries - the code works).~~ ***Edit (2/24): Unfortunately it doesn't look like there's an easy way to tweak this, so I moved on.***

## 💻 Examples

First, you need to configure `@parcel/validator-typescript to run with the following `.parcelrc`
```json
{
    "validators": {
    }
}
```
Then, if you have a file that contains a type error, for example:

***index.ts***
```ts
const message: number = "Hello World!" // This is a type error!
```
...and you run `parcel index.ts`, you'll see the error correctly reported. However when you fix the error and save the file, the validator will re-run, but the old error will still be reported. With this change, however, you'll see the error go away in the terminal.

## 🚨 Test instructions

See the `old errors` branch of [this repo](https://github.com/astegmaier/parcel2-typescript-validator-bug/tree/new-issue) for a repo where the original problem can be seen. If you clone this branch, build it, and `yarn link @parcel/validator-typescript` to that test repository, you can see the problem go away.

## ✔️ PR Todo


packages/validators/typescript/src/TypeScriptValidator.js
```diff
@@ -48,18 +51,25 @@ export default new Validator({
...
 
     if (!langServiceCache[configHash]) return;
 
-    const diagnostics = langServiceCache[configHash].getSemanticDiagnostics(
-      asset.filePath,
-    );
+    // Make sure that when the typescript language service asks us for this file, we let it know that there is a new version.
+    langServiceCache[configHash].host.invalidate(asset.filePath);
+
+    const diagnostics = langServiceCache[
```
DeMoorJasper: Prettier does some not so pretty things from time to time not sure if we can even do anything about it.

packages/validators/typescript/src/TypeScriptValidator.js
```diff
@@ -48,18 +51,25 @@ export default new Validator({
...
+      };
     }
 
     if (!langServiceCache[configHash]) return;
 
-    const diagnostics = langServiceCache[configHash].getSemanticDiagnostics(
-      asset.filePath,
-    );
+    // Make sure that when the typescript language service asks us for this file, we let it know that there is a new version.
+    langServiceCache[configHash].host.invalidate(asset.filePath);
```
DeMoorJasper: Wow, I thought typescript automatically did this. My bad

padmaia: Does Typescript consider more than one file at a time while type checking? Is it possible that multiple files may have changed, but we are only invalidating the one, which could lead to a false error?

DeMoorJasper: Yes, I do think that is possible, hopefully their invalidate function reruns validations over all imports as well. Not sure, would have to look up if there’s any documentation on it

astegmaier: Based on my reading of the code (which may be wrong - please correct me if so), the `invalidate` function is not a typescript API - it's part of _our_ implementation of the `LanguageServiceHost` - see here:

https://github.com/parcel-bundler/parcel/blob/706a7e647f90c65724f6b3560376c9ffab80eb5e/packages/utils/ts-utils/src/LanguageServiceHost.js#L18-L28

 It has the effect of ensuring that the next time the Typescript Language Service calls getScriptVersion (which _is_ part of [the interface that a `LanguageServiceHost` must implement](https://github.com/microsoft/TypeScript/blob/7f02927b7803384b87c50c884b07cd872dbed1af/src/services/types.ts#L213)), it returns an incremented version number:

https://github.com/parcel-bundler/parcel/blob/706a7e647f90c65724f6b3560376c9ffab80eb5e/packages/utils/ts-utils/src/LanguageServiceHost.js#L34-L36

By setting breakpoints in my simple test app, I can see that that when you ask the Language Service for diagnostics about any given file (through `getSemanticDiagnostics`) it will make calls to `getScriptVersion` for every file that is a dependency of the file you are checking. Presumably, if we return an incremented version number, it can tell that it needs to re-analyze the dependency as well. 

I could see a possible bug here related to order (maybe that's what you were talking about above): if you have a project with `a.ts` that depends on `b.ts`, and you change both of them, we invalidate and check `a.ts` before we have invalidated `b.ts`, so you get a potentially false error on `a.ts` (which was analyzed assuming the old contents of `b.ts`. I'll test this scenario to confirm and think about ways to address it.

packages/validators/typescript/src/TypeScriptValidator.js
```diff
@@ -45,21 +48,28 @@ export default new Validator({
     if (tsconfig && !langServiceCache[configHash]) {
       let parsedCommandLine = ts.parseJsonConfigFileContent(
         tsconfig,
-        ts.sys,
+        new ParseConfigHost(options.inputFS, ts),
```
astegmaier: Note: this is a new change that I had to make in order to get unit tests working. Without it, typescript will not look to the `inputFS` specified in parcel configuration (which is injected in unit tests so we can modify it programmatically).

packages/validators/typescript/src/TypeScriptValidator.js
astegmaier: I tested things out, and this is indeed a problem - I filed #4204 to track it.

Since it was a pre-existing problem (this PR merely unblocks you from discovering it), I'd propose merging this, and doing a separate PR to tackle that one.

astegmaier:
@padmaia and/or @DeMoorJasper - I _think_ this is ready to merge as far as functionality and unit tests go. I know @DeMoorJasper already approved it, but since it's my first merge, and some things have changed since the initial approval, I wanted to get at least one of you to give me the go-head out of an abundance of caution.

I'm a bit puzzled by the test suite failures in Azure Pipelines - I don't get them on my Windows or Mac machines locally, and I saw that the last merge also had similar failures. Is this a known issue? If yes, is it okay to merge with the outstanding errors?

packages/utils/ts-utils/src/LanguageServiceHost.js
```diff
@@ -16,12 +16,15 @@ export class LanguageServiceHost extends FSHost {
   }
 
   invalidate(fileName: FilePath) {
-    const entry = this.files[fileName];
+    // When the typescript language server calls "getScriptVersion", it will normalize paths for cross-platform (e.g. C:\myFile.ts on Windows becomes C:/myFile.ts). We need to do the same thing.
+    // $FlowFixMe getNormalizedAbsolutePath is missing from the flow-typed definition.
+    const normalizedFileName = this.ts.getNormalizedAbsolutePath(fileName);
+    const entry = this.files[normalizedFileName];
```
astegmaier: This is something I had to add since you looked at it last in order to get things working on Windows.

packages/validators/typescript/src/TypeScriptValidator.js
DeMoorJasper: Awesome, it should have been like that all along, thanks for fixing this bug

packages/validators/typescript/src/TypeScriptValidator.js
DeMoorJasper: That's unfortunate, sounds good to fix this separate from this PR

packages/utils/ts-utils/src/LanguageServiceHost.js
DeMoorJasper: Looks good

astegmaier:
I fixed the merge issues (sorry for that!), so I think it's good-to-go.

(The problem was that I ran `rebase` more than once and picked up changes from the main `v2` repo each time, which then got re-played on top of already-committed stuff. Lesson learned!)

PR #4188, opened 2020-02-20
Use stable ids in names of inline bundles

padmaia:
# ↪️ Pull Request
Fixes #4145

The issue was caused by a bundle that contained a hash reference to an inline parent bundle, which is something I didn't anticipate, but is apparently needed for scope hoisting. @devongovett @mischnic Is this expected?

See below for a visual of the bundle graph. In this case the blue bundle is referencing the pink bundle.

bundle_graph

To fix this, I changed the default namer to not use content hash references in inline bundle names. It still uses some characters of the bundle id to ensure uniqueness. This seems like a safe option since inline bundles are not served by themselves so their names do not need to be long term cacheable.

## 🚨 Test instructions
Reproduced the failure in the linked issue and tested that the changes were able to build the code successfully. We should probably add some integration tests, but I wanted to make sure that this scenario is expected.

## ✔️ PR Todo

- [ ] Probably add some integration tests

PR #4189, opened 2020-02-20
Incorporate child bundle ids in BundleGraph.getHash(bundle)

padmaia:
# ↪️ Pull Request
Before the addition of bundle content hash references (#4114), `BundleGraph.getHash(bundle)` incorporated the content of all descendant bundles (See #3414 for reasoning). With bundles maintaining stable hash references through packaging and optimizing, this was overkill, so I updated the method. Unfortunately I updated it to not consider they may end up referencing different bundles 😔.

Now all descendant bundle ids are considered in the `BundleGraph.getHash(bundle)` function. This is still a little overkill because bundles usually only reference direct child bundles, but in the future when we add mappings of stable ids to final bundle names for all bundles in entry bundles, we would need to consider every child. Maybe the runtime assets with the mappings could add dependencies to all bundles it references, in which case checking only child bundles would be okay. But still, we kind of have a problem that packagers have access to the entire `BundleGraph` and we aren't totally sure what they are going to do with it. If we are only going to use direct child bundle ids in the hash, we need to make sure that's all packagers have access to. That or add another method to packagers that can narrow down what should go in the cache key before packaging.

Another thing I just discovered was that scope hoisting brings in the possibility of bundles referencing parent bundles. They reference not just their names, but their exports too, which means we probably also have to consider parent bundle content in the hash? This is based on my findings working on #4188.

## ✔️ PR Todo

- [ ] Add some tests
- [ ] Maybe consider parent bundle's content based on scope hoisting findings

### DIFF
--- a/packages/core/core/src/BundleGraph.js
+++ b/packages/core/core/src/BundleGraph.js
@@ -576,14 +576,15 @@ export default class BundleGraph {
 
   getHash(bundle: Bundle): string {
     let hash = crypto.createHash('md5');
-    this.traverseBundles((childBundle, ctx) => {
+    this.traverseBundles((childBundle, ctx, traversal) => {
       if (
         childBundle.id === bundle.id ||
         (ctx?.parentBundle === bundle.id && childBundle.isInline)
       ) {
         hash.update(this.getContentHash(childBundle));
-      } else if (!childBundle.isInline) {
+      } else {
         hash.update(childBundle.id);
+        traversal.skipChildren();
       }
       return {parentBundle: childBundle.id};
     }, bundle);
--- a/packages/core/core/src/BundleGraph.js
+++ b/packages/core/core/src/BundleGraph.js
@@ -582,7 +582,7 @@ export default class BundleGraph {
         (ctx?.parentBundle === bundle.id && childBundle.isInline)
       ) {
         hash.update(this.getContentHash(childBundle));
-      } else {
+      } else if (!childBundle.isInline) {
         hash.update(childBundle.id);
       }
       return {parentBundle: childBundle.id};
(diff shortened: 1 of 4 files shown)


parcel-benchmark:
## Benchmark Results

#### Timings

| --- | --- | --- |
| Cold | 4.25s | -1.39s 🚀 |
| Cached | 3.70s | -87.00ms |

#### Cold Bundles

| dist/parcel.d5807e82.webp | 102.94kb | +0.00b | 98.00ms | -6.00ms 🚀 |
| dist/index.js | 1.07kb | +5.00b ⚠️ | 218.00ms | -9.00ms |

#### Cached Bundles

| dist/parcel.d5807e82.webp | 102.94kb | +0.00b | 3.00ms | +1.00ms ⚠️ |
| dist/index.js | 1.07kb | +5.00b ⚠️ | 8.00ms | +1.00ms ⚠️ |

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 9.25s | +512.00ms ⚠️ |
| Cached | 3.81s | +30.00ms |

#### Cold Bundles

| dist/index.js | 64.04kb | +397.00b ⚠️ | 1.28s | +11.00ms |

#### Cached Bundles

| dist/index.js | 64.04kb | +397.00b ⚠️ | 442.00ms | +141.00ms ⚠️ |

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 4.36s | -87.00ms |
| Cached | 3.71s | +287.00ms ⚠️ |

#### Cold Bundles

| dist/index.js | 743.00b | +7.00b ⚠️ | 211.00ms | +7.00ms |

#### Cached Bundles

| dist/index.js | 743.00b | +7.00b ⚠️ | 316.00ms | +310.00ms ⚠️ |

</p></details>

**[Click here to view a detailed benchmark overview.](https://benchmark.parceljs.org/benchmark/257855504137585161)**

PR #4209, opened 2020-02-23
Image transformer

### LINKED ISSUE #3477: [RFC] Import query params

### LINKED ISSUE #3736: Parcel 2: WebP Transformer
Create the `@parcel/transformer-webp` package. This should include a Transformer plugin that converts images to [WebP](https://developers.google.com/speed/webp) format. Ideally, this should wrap a C++ module to call [libwebp](https://developers.google.com/speed/webp/docs/api) rather than relying on writing temporary files and calling command line tools. Perhaps such a module already exists, otherwise we could create one. If this proves too difficult, shelling out to a command line tool could be a good starting point.

## Configuration

Optional configuration should be done through a `.webprc` file. We should include good default settings, but it should be possible to override them if needed. See [here](https://developers.google.com/speed/webp/docs/cwebp) for a full list of options.

We could also support passing options via query params for image-specific configuration. (??)

## Default Parcel Config

This transformer should be included in `@parcel/config-default` under a `webp:` named pipeline:

```json
{
  "transforms": {
    "webp:*.{jpg,jpeg,png,tiff,webp}": "@parcel/transformer-webp"
  }
}
```

This allows it to be used like the following to convert images in source code:

```html
<picture>
  <source type="image/webp" srcset="webp:./image.jpg">
  <source type="image/jpeg" srcset="./image.jpg">
  <img src="./image.jpg">
</picture>
```

The same original JPEG image is referenced in both `<source>` tags, but the first is converted to webp by Parcel via the named pipeline. This won't only work in HTML, but also in other places. For example, you could reference an image from CSS like this:

```css
.header {
  background-image: url("webp:./header.jpg");
}
```

### LINKED ISSUE #3737: Parcel 2: Image Resizing Transformer
Create the `@parcel/transformer-image-resize` package. This should include a Transformer plugin that resizes images using the [sharp](https://github.com/lovell/sharp) library. It should accept options via query parameters to determine what size to resize the image to.

For example, you could reference an image like this to resize it to 200px wide:

```html
<img src="./image.jpg?width=200" width="200">
```

This would also work anywhere images can be referenced, e.g. CSS or JS.

This transformer should be included in `@parcel/config-default` in a pipeline for `*.{png,jpg,jpeg}`. However, in order to improve development build performance, it should only run in production mode. This should be safe as long as users are using explicit sizes on their image tags in addition to their URLs. Also, if no query params are provided to the image url, the plugin should do nothing.

DeMoorJasper:
# ↪️ Pull Request

Allow resizing of images using query parameters and sharp

Closes #3737
Closes #3736
Closes #3477

Closes T-6

## 💻 Examples

```
import image from 'url:./image.jpg?height=200';
import image from 'url:./image.webp?height=200&width=300';
import image from 'url:./image.png?width=300';
```

## 🚨 Test instructions

Run kitchen sink example

## ✔️ PR Todo


parcel-benchmark:
## Benchmark Results

#### Timings

| --- | --- | --- |
| Cold | 3.50s | -971.00ms 🚀 |
| Cached | 3.16s | +185.00ms ⚠️ |

#### Cold Bundles

| dist/parcel.d5807e82.webp | 102.94kb | +0.00b | 110.00ms | +9.00ms ⚠️ |
| dist/index.js | 1.07kb | +5.00b ⚠️ | 222.00ms | +8.00ms |

#### Cached Bundles

| dist/parcel.d5807e82.webp | 102.94kb | +0.00b | 270.00ms | +268.00ms ⚠️ |
| dist/index.js | 1.07kb | +5.00b ⚠️ | 275.00ms | +269.00ms ⚠️ |

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 7.40s | +174.00ms |
| Cached | 3.13s | -87.00ms |

#### Cold Bundles

| dist/index.js | 64.04kb | +397.00b ⚠️ | 1.05s | -8.00ms |

#### Cached Bundles

| dist/index.js | 64.04kb | +397.00b ⚠️ | 329.00ms | -8.00ms |

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 3.58s | -683.00ms 🚀 |
| Cached | 3.08s | +174.00ms ⚠️ |

#### Cold Bundles

| dist/index.js | 743.00b | +7.00b ⚠️ | 157.00ms | -711.00ms 🚀 |

#### Cached Bundles

| dist/index.js | 743.00b | +7.00b ⚠️ | 5.00ms | -0.00ms |

</p></details>

**[Click here to view a detailed benchmark overview.](https://benchmark.parceljs.org/benchmark/258176667132887553)**

packages/core/core/src/public/Asset.js
```diff
@@ -87,6 +88,10 @@ class BaseAsset {
     return this.#asset.value.filePath;
   }
 
+  get query(): QueryParameters {
```
DeMoorJasper: Should transformers be able to add additional data to the query?

Only example I can think of is html-transformer that would add width and height from srcset to the image asset, but that's probably something with addDependency anyway

packages/utils/node-resolver-core/src/NodeResolver.js
```diff
@@ -32,6 +35,26 @@ type Module = {|
   filePath?: FilePath,
 |};
 
+const parseFilePath = (
```
DeMoorJasper: Not entirely sure if this is the correct place for this.

I do believe we should pass queryparams to the resolver, in case it should resolve to some remote asset like `deno` does for example... But maybe we should make the resolverrunner handle this and pass it as an argument to each resolver so they can opt in to caring about it or not?

packages/utils/node-resolver-core/src/NodeResolver.js
```diff
@@ -5,14 +5,17 @@ import type {
...
 } from '@parcel/types';
 import path from 'path';
 import {isGlob} from '@parcel/utils';
 import micromatch from 'micromatch';
 import builtins from './builtins';
 import nullthrows from 'nullthrows';
+import querystring from 'querystring';
 
 const EMPTY_SHIM = require.resolve('./_empty');
+const QUERY_PARAMS_REGEX = /^([^\s?]*)(\?.*)?/;
```
DeMoorJasper: Note to self: update this regex to support spaces, we allow this in html

packages/examples/kitchen-sink/src/index.js
```diff
@@ -1,6 +1,7 @@
 import styles from './styles.css';
 import parcel from 'url:./parcel.webp';
 import {message} from './message';
+import warehouse from 'url:./image.jpg?height=200&as=webp';
```
DeMoorJasper: Throws `Error: Unknown pipeline webp` is this intended behaviour or a bug in named pipelines?
cc @devongovett

devongovett: caching bug?

packages/transformers/image/src/ImageResizeTransformer.js
```diff
@@ -0,0 +1,31 @@
+// @flow
+import {Transformer} from '@parcel/plugin';
+import sharp from 'sharp';
+
+export default new Transformer({
+  async transform({asset}) {
+    let inputBuffer = await asset.getBuffer();
```
devongovett: Looks like sharp supports streaming. Might save some memory? https://www.npmjs.com/package/sharp#stream

packages/examples/kitchen-sink/src/index.js
DeMoorJasper: you're right, adding `--no-cache` fixes it

packages/transformers/image/src/ImageResizeTransformer.js
DeMoorJasper: Sure, didn't know parcel asset has a setStream function.

packages/core/core/src/BundleGraph.js
```diff
@@ -566,7 +567,13 @@ export default class BundleGraph {
     let hash = crypto.createHash('md5');
     // TODO: sort??
     this.traverseAssets(bundle, asset => {
-      hash.update([asset.outputHash, asset.filePath].join(':'));
+      hash.update(
+        [
+          asset.outputHash,
+          asset.filePath,
+          querystring.stringify(asset.query),
```
DeMoorJasper: Is this how this is done?

packages/core/integration-tests/test/image.js
```diff
@@ -0,0 +1,28 @@
...
+import path from 'path';
+
+describe('image', function() {
+  it('Should be able to resize images', async () => {
+    await bundle(path.join(__dirname, '/integration/image/resized.js'));
+
+    let dirContent = await outputFS.readdir(distDir);
+    console.log(dirContent);
+
+    assert.equal(dirContent.length, 8);
```
DeMoorJasper: This should output an image for each format, it just outputs one.
I tried updating the key but it does not seem to work. Not entirely sure how I should fix this. I thought it might be a bundling thing but the assets should be isolated?

mischnic:
I'm not entirely sure if this is already the case, but I think query params should generally only be allowed when explicitly specifying a pipeline (for the same reason that we went with a protocol-like syntax for pipelines instead of query params).

DeMoorJasper:
@mischnic Maybe I misunderstood but I don't think it should be explicitly specified as a pipeline (in parcelrc). That would result in a lot of pipeline config (I think)?

packages/transformers/image/src/ImageResizeTransformer.js
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    if (format) {
+      if (!FORMATS.has(format)) {
+        throw new Error(`Sharp does not support ${format} images.`);
+      }
+
+      asset.type = format;
+
+      imagePipeline[FORMATS.get(format)]();
```
brettinternet: Nearly all formats (except raw) have a common `quality` attribute in the "options" argument that's passed into the format methods. Perhaps `?quality={number}` would be a good candidate for query param, and could be passed in here as a parameter as `{ quality }`?

See: https://sharp.pixelplumbing.com/api-output

DeMoorJasper: Seems fair enough, will add it in later today

DeMoorJasper:
Gonna rework this... Too much conflicts at this point

PR #4212, opened 2020-02-24
Allow passing config to posthtml-render to control HTML rendering.

kevincox:
# ↪️ Pull Request

Allow passing config to posthtml-render to control HTML rendering.

## 💻 Examples

```
"posthtml-render": { quoteAllAttributes: false },
```

This option would allow your HTML to be smaller and easier to read. However this change is generic and provides access to any posthtml-render options.

## 🚨 Test instructions

- Add a "posthtml-render" property to `package.json` and see the effect.
- Create a `posthtml-render.config.js` file and see the effect.

## ✔️ PR Todo

### DIFF
--- a/packages/core/parcel-bundler/src/transforms/posthtml.js
+++ b/packages/core/parcel-bundler/src/transforms/posthtml.js
@@ -35,6 +35,7 @@ async function getConfig(asset) {
   }
 
   config = config || {};
+  delete config.render;
   const plugins = config.plugins;
   if (typeof plugins === 'object') {
     // This is deprecated in favor of result messages but kept for compatibility
--- a/packages/packagers/html/src/HTMLPackager.js
+++ b/packages/packagers/html/src/HTMLPackager.js
@@ -53,6 +53,14 @@ export default (new Packager({
       .getSiblingBundles(bundle)
       .filter(b => !b.isInline && !referenced.has(b.id));
 
+    let posthtmlConfig = await asset.getConfig(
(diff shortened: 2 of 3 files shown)


mischnic:
Could we use reuse `.posthtmlrc` (and derivates) ?

https://github.com/parcel-bundler/parcel/blob/e94f0814b5649b48a5368478cb94aa87bbca77ec/packages/transformers/posthtml/src/PostHTMLTransformer.js#L14-L19

kevincox:
On Tue., Feb. 25, 2020, 18:41 Niklas Mischkulnig, <notifications@github.com>
wrote:

> Could we use reuse .posthtmlrc (and derivates) ?
>
I looked at this but it appears that the set of options here are completely
disjoint from the set of options used by posthtml-render. This leaves us
with two options if we use the existing file:

1. Add a `render` or similar attribute to the config, pass all but that to
posthtml and pass that attribute as the config for posthtml-render.
2. Pass the same config to both and hope that they don't collide (and
ignore unknown options).

I'm not a huge fan of either but they could work.

>

DeMoorJasper:
@kevincox I think option 1 is the one we should go with, creating an extra config file seems like a worse idea

kevincox:
Easy enough to implement. I'll update the PR if that is the only blocking
issue.

DeMoorJasper:
@kevincox great, looking forward to it

kevincox:
Done

PR #4215, opened 2020-02-24
Fix main build script to work on Windows.

astegmaier:
# ↪️ Pull Request

The current `build` script in the main `package.json` will choke on Windows, because it tries to define an environment variable in the POSIX style. This is easy to fix using [cross-env](https://www.npmjs.com/package/cross-env) (similar to the existing `test:unit` script).

## 💻 Examples

Currently (without this change), running `yarn test` on windows will show this error:
```
'NODE_ENV' is not recognized as an internal or external command,
operable program or batch file.
```
With this change, it will succeed.

## 🚨 Test instructions

Just run `yarn build`/`npm test`.

## ✔️ PR Todo

- It didn't seem necessary/possible to add a unit test for this change (let me know if you disagree)


mischnic:
Could you please add cross-env to the dependencies in the root package.json?

astegmaier:
> Could you please add cross-env to the dependencies in the root package.json?

Done. Two other possible cleanup tasks that occurred to me when I searched the project for `cross-env` were:
1. Remove the reference to `cross-env` from the `package.json` of `parcel-bundler` (which doesn't appear to use it):
https://github.com/parcel-bundler/parcel/blob/04f2e662b571a9caba58363aa0ce3303a282e281/packages/core/parcel-bundler/package.json#L90
2. Bump the version to the latest `cross-env`  (`5.2.1`) in the two places that _do_ use it (main `package.json` and `packages/core/integration-tests/package.json`). `Yarn.lock` seems to say that this is the version that gets installed anyways.

Do both of those changes make sense to you?

DeMoorJasper:
@astegmaier `parcel-bundler` is parcel 1 which we keep in the monorepo so we could merge this with the master branch (which eventually never happened). It's still good for reference.

Anyway feel free to update cross-env to the latest version and remove it from `parcel-bundler` package

astegmaier:
> Anyway feel free to update cross-env to the latest version and remove it from `parcel-bundler` package

Done. I used the latest version (`7.0.0`). It drops support for Node 8, but it appears that we require node 10, so that seemed okay.

https://github.com/parcel-bundler/parcel/blob/e94f0814b5649b48a5368478cb94aa87bbca77ec/package.json#L46

PR #4076, opened 2020-02-03
Ensure that shared sibling bundles are added to all relevant bundle groups

devongovett:
This fixes cases where two entries depend on a shared asset which has siblings (e.g. a shared JS bundle with a sibling CSS bundle). Due to DFS, the subtree of the shared asset is only processed once, meaning any sibling bundles created due to type changes would only be connected to the first bundle group. To work around this, we store a list of sibling bundles for each asset in the graph, and when we re-visit a shared asset, we connect them all to the current bundle group as well.

PR #4231, opened 2020-02-27
Fixes for shared bundles

devongovett:
This PR includes fixes for two bundling issues related to shared bundles.

1. As a result of #4076, unrelated sibling bundles were being connected to bundle groups incorrectly in cases like the below graph. In this case, DatePicker and TextField both share a dependency on textfield.css. DatePicker also depends on datepicker.css. However, the TextField JS bundle ends up with an import for the datepicker.css also since it is tracked as a sibling when we visit textfield.css the first time. This is fixed by tracking siblings whenever a bundle is created rather than only when a new bundle group is created.
image

2. In a monorepo, if two packages depended on the same shared CSS file, only a single bundle would be produced in one of the packages, and the other package would end up depending on that first package. So, imports like `import "../foo/dist/test.css"` would be produced, which wouldn't work when the package was published. This was due to the bundle ids clashing and being deduped in the graph. To fix this, we now include the target's dist directory in the id of the bundle so that two targets in different output directories never share bundles.


packages/core/core/src/public/MutableBundleGraph.js
```diff
@@ -118,8 +118,11 @@ export default class MutableBundleGraph implements IMutableBundleGraph {
       ? assetToInternalAsset(opts.entryAsset)
       : null;
 
+    let target = targetToInternalTarget(opts.target);
     let bundleId = md5FromString(
-      'bundle:' + (opts.uniqueKey ?? nullthrows(entryAsset?.value.id)),
+      'bundle:' +
+        (opts.uniqueKey ?? nullthrows(entryAsset?.value.id)) +
+        target.distDir,
```
wbinnssmith: is there a reason this isn't just a hash of the (ordered) target object?

packages/bundlers/default/src/DefaultBundler.js
```diff
@@ -66,45 +64,53 @@ export default new Bundler({
...
-            for (let bundle of siblings) {
-              bundleGraph.addBundleToBundleGroup(bundle, bundleGroup);
-            }
-          } else {
-            siblingBundlesByAsset.set(asset.id, siblingBundles);
-          }
 
-          let parentAsset = context.parentNode.value;
           if (parentAsset.type === asset.type) {
+            // If any sibling bundles were created for this asset or its subtree previously,
```
wbinnssmith: should this be moved into the consequent of the first `if` below and updated to reflect the change?

parcel-benchmark:
## Benchmark Results

#### Timings

| --- | --- | --- |
| Cold | 3.81s | -39.00ms |
| Cached | 3.09s | -307.00ms 🚀 |

#### Cold Bundles

| dist/index.js | 1.07kb | +8.00b ⚠️ | 214.00ms | -5.00ms |

#### Cached Bundles

| dist/parcel.d5807e82.webp | 102.94kb | +0.00b | 2.00ms | +1.00ms ⚠️ |
| dist/index.js | 1.07kb | +8.00b ⚠️ | 6.00ms | -1.00ms 🚀 |

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 9.86s | +690.00ms ⚠️ |
| Cached | 3.48s | -50.00ms |

#### Cold Bundles

| dist/index.js | 67.47kb | +429.00b ⚠️ | 1.18s | -24.00ms |

#### Cached Bundles

| dist/index.js | 67.47kb | +429.00b ⚠️ | 7.00ms | +1.00ms ⚠️ |

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 4.19s | -15.00ms |
| Cached | 3.20s | -298.00ms 🚀 |

#### Cold Bundles

| dist/index.js | 743.00b | +7.00b ⚠️ | 185.00ms | +11.00ms ⚠️ |

#### Cached Bundles

| dist/index.js | 743.00b | +7.00b ⚠️ | 5.00ms | -0.00ms |

</p></details>

**[Click here to view a detailed benchmark overview.](https://benchmark.parceljs.org/benchmark/258564385464648201)**

packages/core/core/src/public/MutableBundleGraph.js
devongovett: well... I wasn't sure if we wanted to include everything in there, e.g. environment, etc. Two targets in the same dist dir with different environments wouldn't share bundles then maybe?

packages/bundlers/default/src/DefaultBundler.js
devongovett: you mean move the comment?

PR #4237, opened 2020-02-28
Fix typescript rechecking when many files change

### LINKED ISSUE #4204: @parcel/validator-typescript can report incorrect errors if you change multiple files at once
# 🐛 bug report

In testing PR #4187 (which unblocks the ability of `@parcel/validator-typescript` to re-analyze files after they are saved), I discovered a pre-existing problem with the way that the validator handles re-analyzing projects in cases where multiple typescript files change at the same time. 

Thanks to @padmaia for asking [the right question](https://github.com/parcel-bundler/parcel/pull/4187#discussion_r382167489) to help uncover this.

***.parcelrc:***
```js
{
    "validators": {
    }
}
```
***package.json:***
```
{
  "version": "1.0.0",
  "main": "index.html",
  "private": true,
  "scripts": {
  },
  "devDependencies": {
    "@parcel/validator-typescript": "2.0.0-nightly.117", // In order to see this problem, you need to use a build created after #4187 is merged.
    "parcel": "2.0.0-nightly.115",
    "typescript": "3.7.5"
  }
}
```

***src/index.ts:***
```ts
import { showMessage } from "./showMessage";
const message: string= "A new Message!";
showMessage(message);
```

***src/myFunction.ts***
```ts
export function showMessage(message: string) {
    console.log(message);
}
```
## 🤔 Expected Behavior

When you start the above project with parcel (by running `parcel src/index.ts`), it should build with no warnings (because the initial types of both `index.ts` and `App.ts` are valid).

Now, change _both_ files (taking care to save them both at the same time) to:

***src/index.ts:***
```ts
import { showMessage } from "./showMessage";
const message: number = 123; // note the new type of 'number'
showMessage(message);
```

***src/myFunction.ts***
```ts
export function showMessage(message: number) { // we've also changed the type to 'number' here.
    console.log(message);
}
```
Parcel should rebuild the project successfully and not show any type errors (because there are none).

## 😯 Current Behavior

`@parcel/typescript-validator` will show:
```
🚨 Build failed.
@parcel/validator-typescript: Argument of type 'number' is not assignable to parameter of type 'string'.
/Users/Andrew/Projects/parcel2-typescript-validator-bug/src/index.ts
/Users/Andrew/Projects/parcel2-typescript-validator-bug/src/index.ts:3:13
  2 | const message: number = 123;
> 3 | showMessage(message);
>   |             ^^^^^^^^ Argument of type 'number' is not assignable to parameter of type 'string'.
  4 | 
```

## 💁 Possible Solution

[`TypeScriptValidator`](https://github.com/parcel-bundler/parcel/blob/v2/packages/validators/typescript/src/TypeScriptValidator.js) runs the `validate` method once for each file that changes. In order to make sure that the TypeScript `LanguageService` is updated about the changes, it needs to make an `invalidate()` call to the `LanguageServiceHost`, which ensures that when the `LanguageService` asks for that file, the it will inform the service about the changes. But because the `invalidate()` method for the _second_ file change changes is only run _after_ the first file is already analyzed, the first file's analysis is wrong (because it is based off the old version of the second file).

One idea of how to fix it would be: create some sort of queue that only calls the the `LanguageService` _after_ all the files that changed have been invalidated.

## 🔦 Context

## 💻 Code Sample

See the `multiple-files-change-issue` branch of [this repo](https://github.com/astegmaier/parcel2-typescript-validator-bug/tree/multiple-files-change-issue) for a demonstration.

## 🌍 Your Environment

| Parcel           | 2.0.0-nightly.115
| Node             | 10.6.3
| Yarn         | 1.21.1

astegmaier:
# ↪️ Pull Request

***Update (3/20):*** This PR is ready-to-review given the conversation below. See [this comment](https://github.com/parcel-bundler/parcel/pull/4237#issuecomment-601989475) for details.

This is a ~~work-in-progress~~ PR to facilitate a conversation about how best to fix #4204. The problem it is trying to solve is: when multiple typescript files that depend on each other change at the same time, the typechecker can (sometimes) report incorrect errors.
 
~~The current changes demonstrate only _one_ way of solving the problem (and IMO not really the best long-term way) - see the discussion below for additional ideas. Also, there are `ANDREW_TODO` comments in various places to call out lines that I would need to clean up before finalizing.~~

## 💻 Examples

Start with a simple, two-file project with `@parcel/validator-typescript` enabled:

***.parcelrc:***
```js
{
    "validators": {
    }
}
```

***src/index.ts:***
```ts
import { showMessage } from "./showMessage";
const message: string= "A new Message!";
showMessage(message);
```

***src/myFunction.ts***
```ts
export function showMessage(message: string) {
    console.log(message);
}
```
Then, _simultaneously_ edit both files to this:

***src/index.ts:***
```ts
import { showMessage } from "./showMessage";
const message: number = 123; // note the new type of 'number'
showMessage(message);
```

***src/myFunction.ts***
```ts
export function showMessage(message: number) { // we've also changed the type to 'number' here.
    console.log(message);
}
```
Parcel will sometimes report an incorrect error: `Argument of type 'number' is not assignable to parameter of type 'string'` on `index.ts` (i.e. `index.ts` in being analyzed without being aware of the update to `showMessage.ts`).

The reason for the problem is that parcel will call the [`validate`](https://github.com/parcel-bundler/parcel/blob/48982e22998eb47a45e254c74ce1fcacd0c4956f/packages/validators/typescript/src/TypeScriptValidator.js#L41) method of `TypeScriptValidator` twice ([in parallel](https://github.com/parcel-bundler/parcel/blob/870013cab1003cb422fd425af50c934407df9ecb/packages/core/core/src/AssetGraphBuilder.js#L211-L217), potentially on two separate workers) -- one time for each of the two changed files. Without this PR, each call will [update a cached instance of `LanguageServiceHost`](https://github.com/parcel-bundler/parcel/blob/48982e22998eb47a45e254c74ce1fcacd0c4956f/packages/validators/typescript/src/TypeScriptValidator.js#L68) with a new the version _of the file being analyzed_, but will do nothing to update version of the other file. Then, the Typescript `LanguageService` then will call into the `LanguageServiceHost` [to check if there is a new version](https://github.com/parcel-bundler/parcel/blob/48982e22998eb47a45e254c74ce1fcacd0c4956f/packages/utils/ts-utils/src/LanguageServiceHost.js#L37-L39) of the file being analyzed _and all it's dependencies_.

If we're lucky, and the files are analyzed in the right order (`myFunction.ts` > `index.ts`), everything works. If we're unlucky, and the order is reversed (or if the analysis happens on separate `worker` threads), then when we ask `LanguageService` to analyze the new version of `index.ts`, it will check to see if `myFunction.ts` has a new version - and `LanguageServiceHost` will answer "no". Thus, it will assume that the `message` parameter of `myFunction` is still of type `string`, and we'll see an (incorrect) error.

## 💡Possible Solutions

In [a conversation](https://github.com/parcel-bundler/parcel/issues/4204#issuecomment-591978444) with @DeMoorJasper, we came up with 4 ideas for a solution.
1. **Correct But Slow** - when running `validate` on a single file, always assume that all files in the project may have changed. 
     - This guarantees that the result will always be correct.
     - It could be significantly slower, especially when analyzing a large project.
2. **`Validate` is handed multiple files** - Change the contract with `Validators` so that the `validate` method always passes an array of all the assets that have changed. Then, we know which ones to update on the `LanguageServiceHost`.
     - This could still return incorrect results if project validations happened on multiple different threads (e.g. Thread 1 validates and updates the version number for file `X` in it's in-memory copy of `LanguageServiceHost`. Then the user makes additional changes, and Thread 2 validates and updates the version number for file `Y`. If `Y` depended on `X`, we'd still get an incorrect error).
3. **Single-Threaded Guarantee**  ***(Implemented in this PR as of 3/20)*** - In addition to the change described in number 2 above, _also_ enable a validator to opt-in to a guarantee that it will only ever execute on a single thread.
     - How would a given `Validator` specify whether it needed to only be on a single thread (like typescript), or could support being parallelized and run on each asset independently (like eslint)? 
     - What is the right way to update the `WorkerFarm` to support this?
4. **willReEvaluate** - Add a `willReEvalidate` method to the `Validator` contract. Parcel would promise to call this API with the full array of files that are about to be re-validated _on every existing thread_ where a validator was initialized and at least one call to `validate` had been made.
      - What is the right way to update the `WorkerFarm` to support this?

## 🚨 Test instructions

I added two unit tests to this PR that you can run in watch mode with this command:
```
yarn test:integration --grep ts-validator -w --watch-files ts-validation
```
You can also use the `multiple-files-change-issue` branch of [this repo](https://github.com/astegmaier/parcel2-typescript-validator-bug/tree/multiple-files-change-issue) for a demonstration.

## ✔️ PR Todo


astegmaier:
@DeMoorJasper @padmaia @devongovett - I know you guys are pretty busy (thanks for all the hard work!), but if one of you gets a chance, I'd love to get some feedback on this. I think what we're running into here is a limitation of the current interface with `validator` plugins in general when there are cross-asset dependencies, and the right solution needs to fit into the long-term vision for the way threading/workers should behave. I'm happy to do the coding and testing, but any big-picture thoughts about the right direction would be helpful.

padmaia:
Thanks for putting this together @astegmaier! This is a really tough problem (you know what they say about cache invalidation). Option 3 seems like a decent medium term solution.

> How would a given Validator specify whether it needed to only be on a single thread (like typescript)

Maybe a separate function?

> What is the right way to update the WorkerFarm to support this?

The simplest thing, which wouldn't require a change to the WorkerFarm, would be to just run plugins that cannot be multi threaded in the main process. Another option would be a dedicated thread, kind of like what @DeMoorJasper did for watching in v1, but I'm not sure how much that would help. 

In the long term, it would be nice to do more things in parallel. Would it be possible to track what files are used in any given validation request? Then we could have only necessary invalidation connections for each request in the graph and add something to the RequestGraph/RequestTracker that keeps track of which events invalidated a node. Then we could send all files that should be invalidated with each request no matter which worker ends up handling it. Now that I'm thinking about it... Does option 3 handle the case where file `X` depends on file `Y`, and only file `Y` changes which may break the contract with `X`. `X` didn't change so it wouldn't be validated again. Would the validation of `Y` catch that?

astegmaier:
Thanks for the advice @padmaia! I've just started paternity leave, so my time available to work on this in the next month will be sporadic, but I'll try to make some progress.

> Does option 3 handle the case where file X depends on file Y, and only file Y changes which may break the contract with X. X didn't change so it wouldn't be validated again. Would the validation of Y catch that?

I added [a new unit test that simulates this case](https://github.com/parcel-bundler/parcel/pull/4237/commits/10a76925d6566cb28eec22d935a4feddbaf7738c), and unfortunately, it currently fails (🙁). I think the solution would involve changing the way we call the typescript language service (we currently call `getSemanticDiagnostics` for each file that changes, but not for files that don't). I can play around with the Typescript API to see if there is a better way.

>The simplest thing, which wouldn't require a change to the WorkerFarm, would be to just run plugins that cannot be multi threaded in the main process.

I'm trying to find a way to do this that accomplishes the goal with the minimal amount of disruption to the existing model, and I'd love your advice if you have any.

Here's what I see about the existing model:

1. When a set of files change, validation is kicked off for all of them within `AssetGraphBuilder`

   https://github.com/parcel-bundler/parcel/blob/da3cf2a5386d71d25eae5418ebef5d9bfb9e5d62/packages/core/core/src/AssetGraphBuilder.js#L216-L222

2. `runValidate` is a "handle" into the `WorkerFarm`, which will queue up a task that eventually gets run (in an individual worker) by the `runValidate` method of `worker.js`

   https://github.com/parcel-bundler/parcel/blob/da3cf2a5386d71d25eae5418ebef5d9bfb9e5d62/packages/core/core/src/AssetGraphBuilder.js#L99

   https://github.com/parcel-bundler/parcel/blob/da3cf2a5386d71d25eae5418ebef5d9bfb9e5d62/packages/core/core/src/worker.js#L46-L56

3. Only within `Validation`  (running in a single worker) is [the work done to load the configuration and figure out which validation plugins apply to a given file](https://github.com/parcel-bundler/parcel/blob/da3cf2a5386d71d25eae5418ebef5d9bfb9e5d62/packages/core/core/src/Validation.js#L53-L72) (which might vary depending on the file's extension, path, and `.parcelrc` files in the project).

So if our goal is to run some plugins in-proc and some plugins on on threads in the WorkerFarm, one way to accomplish this would be:
1. Add a parameter (e.g. `singleInstance`) to the constructor of `Validation` that lets us know if this particular validation is happening in-proc (or with a single-threaded guarantee).
2. Within `Validation.run()`, check this parameter, and call a different method (e.g. `singleThreadValidate` instead of `validate`, if it exists, and _don't_ call the `validate` method).
3. Call `new Validation({ singleInstance: true ...}).run()` (through a promise) within `AssetGraphBuilder.runValidate()`.

Is this roughly what you had in mind? I can work on a PR that tests out this idea in parallel, but if you can spot any obvious problems with it (or have any tips), let me know.

astegmaier:
@DeMoorJasper and @padmaia - I've overhauled this PR given the conversation above, and I think it's ready to merge, pending addressing any feedback from y'all. For the reasons I outlined above, this is unfortunately not a small change (to be fair, it deals with _both_ cache invalidation _and_ naming things 😄 ).

The main changes are:
1. Adds a new (optional) `Validator` contract. Validator plugins can implement (A) the current `getConfig`/`validate` functions _or_ (B) a new `validateAll` function. If they do (A), the plugin should support validating files one-by-one in parallel on separate threads (like `eslint`). If they do (B), parcel will guarantee that the plugin will run in a single thread, and it's appropriate for plugins (like `typescript`) that maintain their own project-wide cache and depend on the relationship between files.
2. Uses the new `validateAll` contract in `validator-typescript` to fix #4237.

Some notes on the implementation:
1. The same `Validation` class [can support running either `validate`- or `validateAll`-type validation plugins depending on whether it is instantiated with `dedicatedThread = true` or `false`](https://github.com/parcel-bundler/parcel/pull/4237/files#diff-bc32b3f6475317f3a66b1cbb60a6c25fR71-R120).
2. It [runs `validateAll`-type plugins on the main thread](https://github.com/parcel-bundler/parcel/pull/4237/files#diff-49d31bfc29cd74b44ca47235d9dbba16R223-R230). It seems like a good idea to eventually enhance this so they get their own dedicated thread, but I figured it would make sense to do this in a separate PR.
3. The diff produced by `git` seems larger than it really is - most of the code is re-used, but I refactored some of it into async helper methods for readability and DRY. (see [`buildAssetsAndValidators`](https://github.com/parcel-bundler/parcel/pull/4237/files#diff-bc32b3f6475317f3a66b1cbb60a6c25fR131-R170), [`handleResult`](https://github.com/parcel-bundler/parcel/pull/4237/files#diff-bc32b3f6475317f3a66b1cbb60a6c25fR172-R186), and [`getConfig`](https://github.com/parcel-bundler/parcel/pull/4237/files#diff-82caeb66e0bad40d35067e3f80f3338cR77-R97), [`tryCreateLanguageService`](https://github.com/parcel-bundler/parcel/pull/4237/files#diff-82caeb66e0bad40d35067e3f80f3338cR99-R127), and [`getValidateResultFromDiagnostics`]())
4. It does not address #4347, but it does add [a pending unit test that captures this issue](https://github.com/parcel-bundler/parcel/pull/4237/files#diff-a5f40351026ba62c971ab55c988f9a53R197-R237). 
5. I fixed a few small issues along the way:
   - [More expressive types](https://github.com/parcel-bundler/parcel/pull/4237/files#diff-82caeb66e0bad40d35067e3f80f3338cR20).
   - [Making sure the `ts-validation` tests don't fail when re-running them with `jest --watch`](https://github.com/parcel-bundler/parcel/pull/4237/files#diff-82caeb66e0bad40d35067e3f80f3338cR59-R62)
    - An issue where [another test was setting `process.browser` to `true`](https://github.com/parcel-bundler/parcel/pull/4237/files#diff-266472ddae511d80d5622d83e42d935bR1320-R1322), which caused `typescriptLanguageService.sys` to fail to instantiate on the main thread.

packages/core/core/src/AssetGraphBuilder.js
```diff
@@ -214,8 +216,18 @@ export default class AssetGraphBuilder extends EventEmitter {
   }
 
   async validate(): Promise<void> {
+    // Schedule validations on workers for all plugins that implement the one-asset-at-a-time "validate" method.
     let promises = this.assetRequests.map(request =>
-      this.runValidate({request, optionsRef: this.optionsRef}),
+      this.runValidate({requests: [request], optionsRef: this.optionsRef}),
+    );
+    // Schedule validations on the main thread for all validation plugins that implement "validateAll".
+    promises.push(
```
DeMoorJasper: Ideally we would run validateAll on a dedicated worker instead of the main thread. Although we probably need to benchmark if it's worth it. (as in it's pointless to multithread if main thread is idle)

packages/core/integration-tests/test/javascript.js
```diff
@@ -1317,6 +1317,9 @@ describe('javascript', function() {
     let output = await run(b);
     assert.ok(output.toString().indexOf('process.browser') === -1);
     assert.equal(output(), true);
+    // Running the bundled code has the side effect of setting process.browser = true, which can mess
+    // up the instantiation of typescript.sys within validator-typescript, so we want to reset it.
+    process.browser = undefined;
```
DeMoorJasper: awesome. thanks for this bugfix

packages/core/core/src/AssetGraphBuilder.js
astegmaier: I agree this is a good idea, and I'm happy to work on it. @padmaia mentioned [above](https://github.com/parcel-bundler/parcel/pull/4237#issuecomment-595572660) that you may have done something similar in v1. I'm sure I can figure it out the hard way, but do you have any pointers or examples I could look at to help?

DeMoorJasper: What I did in v1 is something quite different I just created a wrapper around chokidar which started up a worker.

What I wanted to do for this is assign a worker to it, every worker has an id so in theory we should be able to send along an id to a request to the workerfarm and we should be able to only execute that on that exact worker.

But it would require quite some refactoring so we definitely shouldn't be doing this as part of this PR and open a seperate PR for it instead.

astegmaier:
I'm a bit puzzled by the [failing tests](https://dev.azure.com/devongovett/devongovett/_build/results?buildId=8344&view=logs&j=9473ad5c-6f95-59ee-11bd-9c05c464119a&t=f0fb0729-43c2-589d-5bf6-0aba45e88633) in the latest CI benchmark - they only started occurring since I merged from v2 at the very end (which pulled commits from roughly today (3/21) and yesterday (3/20). I don't see any obvious connection with any of the changes in this PR.

The test failures seem to be intermittent/inconsistent, but when they occur and there is a clear test failure, it tends to be in tests that were written or modified by #4266. @mischnic - do any of these failures look familiar to you? Do you think this PR is safe to merge in spite of them?

DeMoorJasper:
@astegmaier I've been seeing it on v2 and in my PRs as well it's a bug in v2 unrelated to your PR

mischnic:
Thank you for investigating! I noticed that they were flakey as well, but was never able to reproduce them locally, I'll give Win+Node10 a try now.

mischnic:
> perhaps it is some kind of race condition

It is, because it was implicity depending on the order that packagers ran, but they run in parallel. They are not caused by your changes.

astegmaier:
I fixed the merge conflicts with #4371 - I think this is ready to merge @DeMoorJasper.

DeMoorJasper:
@astegmaier awesome, hopefully CI will finally pass so we can merge this :)

PR #4263, opened 2020-03-04
Add flow libdef for babel

mischnic:
# ↪️ Pull Request

For now, I've copied over the types that babel would generate with my PR (https://github.com/babel/babel/pull/11196), we would remove them when Babel includes them itself.

## ✔️ PR Todo

Fix the tests 🙈


packages/transformers/js/src/JSTransformer.js
```diff
@@ -58,7 +58,7 @@ export default new Transformer({
       version: '7.0.0',
       isDirty: false,
       program: parse(code, {
-        filename: this.name,
+        sourceFilename: this.name,
```
mischnic: The only actual bug in Parcel I came across in this endeavor

PR #4282, opened 2020-03-06
Hoist all "top-level" var declarations when wrapping a module

mischnic:
# ↪️ Pull Request

When wrapping a module, hoist all `var` declarations that are in the same function scope as the module itself (so don't traverse into FunctionDeclarations, FunctionExpressions, FunctionExpressions - is that enough?).

# Example

(from https://www.npmjs.com/package/util)

```js
const BigIntSupported = true;

if (BigIntSupported) {
	var bigIntValue = value => !!value; // <----------- 
}

function is(value) {
	return BigIntSupported && bigIntValue(value);
}
exports.is = is;
```


parcel-benchmark:
## Benchmark Results

#### Timings

| --- | --- | --- |
| Cold | 6.01s | -6.00ms |
| Cached | 3.15s | +101.00ms |

#### Cold Bundles

| dist/legacy/parcel.d5807e82.webp | 102.94kb | +0.00b | 187.00ms | -20.00ms 🚀 |
| dist/legacy/index.html | 709.00b | +0.00b | 748.00ms | -58.00ms 🚀 |
| dist/modern/index.html | 709.00b | +0.00b | 744.00ms | -58.00ms 🚀 |
| dist/legacy/kitchen-sink.8cc8c872.js | 418.00b | +0.00b | 728.00ms | -61.00ms 🚀 |

#### Cached Bundles

| dist/legacy/parcel.d5807e82.webp | 102.94kb | +0.00b | 5.00ms | +1.00ms ⚠️ |
| dist/modern/parcel.d5807e82.webp | 102.94kb | +0.00b | 3.00ms | +1.00ms ⚠️ |
| dist/legacy/styles.27b3555b.css | 29.00b | +0.00b | 6.00ms | +1.00ms ⚠️ |
| dist/modern/styles.27b3555b.css | 29.00b | +0.00b | 4.00ms | +1.00ms ⚠️ |

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 7.45s | +100.00ms |
| Cached | 3.06s | -103.00ms |

#### Cold Bundles

| dist/main/index.js | 66.09kb | -553.00b 🚀 | 1.58s | +58.00ms |

#### Cached Bundles

| dist/main/index.js | 66.09kb | -553.00b 🚀 | 6.00ms | +1.00ms ⚠️ |
| dist/module/index.js | 37.05kb | +0.00b | 2.00ms | -1.00ms 🚀 |

</p></details>

#### Timings

| --- | --- | --- |

#### Cold Bundles

#### Cached Bundles

</p></details>

**[Click here to view a detailed benchmark overview.](https://benchmark.parceljs.org/benchmark/259271365860786697)**

packages/shared/scope-hoisting/src/concat.js
```diff
@@ -270,46 +280,66 @@ function wrapModule(asset: Asset, statements) {
...
+    );
+    path.skip();
+  },
+};
+
+function wrapModule(asset: Asset, statements) {
+  let decls = [];
+  let fns = [];
+  let program = t.program(statements);
+  traverse(t.file(program), WRAP_MODULE_VISITOR, null, {decls, fns});
```
devongovett: What's the perf impact of this?

mischnic: I'll test it, but this is a `noScope` traverse that skips function and class bodies, so not too bad I think.

mischnic: Setup (similar to esbuild): 10 200k assets that each contain ~1000x `if(...){var foo1 = ...; let local = ...;}` and ~1000x `function bar1(){}`)

| | v2 | hoist-var |
|---|---|---|
| single `wrapModule` call | 2ms | 60ms
| all `wrapModule` "total time" | 18ms | 716ms
| build time | 14,7s | 15,8s

The question is now: how much of that is actually the overhead of the traverse and how much the fact the with v2, only half as many lines are processed.

mischnic: On the other hand, I don't see what we should do instead, the alternative would be to do a one-off reimplementation of babel-traverse. `babylon-traverse` doesn't really support removing nodes (because you don't have access to the parent and the position in the parent).

devongovett: That's a pretty significant difference, and I believe it's why I did a flat loop in the first place. But seems like it was insufficient, so it might be a cost we have to accept. We'll just have to find an equal or greater optimization elsewhere. :wink:

Did you profile to see what was expensive in there?

mischnic: > We'll just have to find an equal or greater optimization elsewhere.

Updating the npm ecosystem so wrapping isn't needed anymore...

> Did you profile to see what was expensive in there?

The visitor callbacks are red:
Screenshot from 2020-03-21 22-49-27

devongovett: It does seem to have a deeper visit depth than I expected. Are we missing a `path.skip()` somewhere?

mischnic: There isn't really anything to skip here:
```js
if (BigIntSupported) {
  var bigIntValue = () => true;
  let inner_let = false;
  if (inner_let) {
    throw new Error();
  }
}
// duplicated...

function testIsHoisted() {
  return BigIntSupported && bigIntValue();
}
// duplicated....
```

devongovett: How about all ExpressionStatements? We only need to process statements right?

mischnic: I've excluded all Expressions.

|  | small benchmark | bigger benchmark | [atlassian/parcel-stress-test](https://github.com/atlassian/parcel-stress-test)
|------|----|---|-----|
v2               | 5.7s | 18.2s | real: 1:30 - user: 3:05
custom traverse   | 5.0s | 20.1s | 1:22 - 3:00
hoist-var         | 5.7s | 20.8s | 1:24 - 3:00

"custom traverse" is a 80-line reimplmentaiton of babel-traverse that only does what we need here (`Path#replaceWith`, `Path#remove`, `Path#skip`).

mischnic: So in summary: this is an artificial benchmark and the `@babel/traverse` overhead is negligible in the real world.

mischnic: With the custom traverse, each of these traversals takes 28ms instead of 81ms, is that a significant difference?

PR #4326, opened 2020-03-13
refactor: Use `p-syncy` to wrap `deasync`

Richienb:
# ↪️ Pull Request

Use `p-syncy` to abstract the logic needed to use `deasync`.

## ✔️ PR Todo

- [x] Added/updated unit tests for this change (tests don't change)
- [x] Included links to related issues/PRs (related issues/PRs don't exist)


DeMoorJasper:
Not entirely sure if we should use this as it's no different from our current util which works fine and we no longer rely on deasync, besides for register which actually does not even work atm

PR #4345, opened 2020-03-19
One parcel config to rule them all

padmaia:
# ↪️ Pull Request
This PR changes the behavior of Parcel to only resolve a single parcel file once* (*see code comments for a note) instead of looking for nested config files for every file that is being transformed. The main motivation is performance, as this takes ~60s off of a 5m cold build.

Close T-127


packages/core/core/src/AssetGraphBuilder.js
```diff
@@ -158,6 +160,39 @@ export default class AssetGraphBuilder extends EventEmitter {
     assetGraph: AssetGraph,
     changedAssets: Map<string, Asset>,
   |}> {
+    let {config, configRef} = nullthrows(
+      await this.configRequestRunner.runRequest(null, {
```
padmaia: For now, the Parcel config file is being resolved twice (once in `Parcel.js` and once here). Once the `RequestGraph` supports sub-requests and is hoisted up the the `Parcel.js` level, then any request that requires the config will call that request directly, reusing the result from the graph as long as the config request is not invalid.

packages/core/core/src/Transformation.js
```diff
@@ -170,8 +179,13 @@ export default class Transformation {
     );
     let initialCacheEntry = await this.readFromCache(initialAssetCacheKey);
 
+    if (initialCacheEntry) {
+      console.log('FOUND CACHE ENTRY');
```
wbinnssmith: debug 😄

packages/core/integration-tests/test/babel.js
```diff
@@ -511,6 +512,7 @@ describe('babel', function() {
         );
 
       build();
+      console.log('FINISHED FIRST BUILD');
```
wbinnssmith: debug 😄

packages/core/core/src/requests/AssetRequestRunner.js
```diff
@@ -0,0 +1,155 @@
...
+  }
+
+  async run(request: AssetRequestDesc, api: RequestRunnerAPI) {
+    api.invalidateOnFileUpdate(
+      await this.options.inputFS.realpath(request.filePath),
+    );
+    let start = Date.now();
+    let {assets, configRequests} = await this.runTransform({
+      request: request,
+      optionsRef: this.optionsRef,
```
wbinnssmith: why do instance methods (this and `runValidate` in `AssetGraphBuilder`) need to be passed instance properties? do the properties change over time?

packages/core/core/src/requests/AssetRequestRunner.js
```diff
@@ -0,0 +1,155 @@
...
+      }
+    }
+
+    // TODO: this should no longer be needed once we have ConfigRequestRunner
+    let graph = this.tracker.graph;
+    let subrequestNodes = [];
+    // Add config requests
+    for (let {request, result} of configRequests) {
+      let id = generateRequestId('config_request', request);
+      let shouldSetupInvalidations =
```
wbinnssmith: nit: inline or move closer to its use below

packages/core/core/src/requests/AssetRequestRunner.js
```diff
@@ -0,0 +1,155 @@
...
+      subrequestNodes.push(subrequestNode);
+
+      // Add dep version requests
+      for (let [moduleSpecifier, version] of result.devDeps) {
+        let depVersionRequst = {
+          moduleSpecifier,
+          resolveFrom: result.resolvedPath, // TODO: resolveFrom should be nearest package boundary
+        };
+        let id = generateRequestId('dep_version_request', depVersionRequst);
+        let shouldSetupInvalidations =
```
wbinnssmith: nit: inline or move closer to its use below

packages/core/integration-tests/test/contentHashing.js
```diff
@@ -24,7 +24,6 @@ describe('content hashing', function() {
 
     let bundleHtml = () => bundle(path.join(__dirname, '/input/index.html'));
     await bundleHtml();
-
```
wbinnssmith: unrelated? this newline seems okay

packages/core/core/src/Transformation.js
padmaia: 🙈

packages/core/core/src/requests/AssetRequestRunner.js
padmaia: This actually has to be on this line because we are checking `!graph.hasNode(id)` and then right after that ensuring the node exists. We only want to set up invalidation connections if this is the first time this request has run or it was just run because it was invalid.

packages/core/core/src/requests/AssetRequestRunner.js
padmaia: https://github.com/parcel-bundler/parcel/pull/4345#discussion_r395880413

packages/core/core/src/requests/AssetRequestRunner.js
padmaia: `runTransform` and `runValidate` are both worker handles that don't have access to instance properties.

PR #4366, opened 2020-03-22
Custom faster ESM to CommonJS transformer

devongovett:
Currently, when not scope hoisting, we compile ES modules to CommonJS using `@babel/plugin-transform-modules-commonjs`. I noticed while benchmarking that this appeared to be quite slow - by far the slowest thing we do in Parcel. Looking through the source code, it appears to do multiple traversals over the entire AST. I'm sure there's a reason for that, but I don't think we need to do it for our use in Parcel. In addition, using `@babel/core` has some overhead over just using `@babel/traverse` directly.

This PR implements a new AST visitor that transforms ESM to CommonJS in a single pass over only the top-level statements in the program. Using scope information collected by babel traverse already, we can replace references without doing multiple full program traverses. On the esbuild benchmark, this improved overall build performance by ~25%!

In addition, builds may be slightly smaller because the helpers for various interop functionality are not duplicated in every file but included once in a separate asset.


parcel-benchmark:
## Benchmark Results
<details><summary>packages/benchmarks/three 🚨</summary><p>

#### Timings

| --- | --- | --- |

#### Cold Bundles

#### Cached Bundles

</p></details>

#### Timings

| --- | --- | --- |

#### Cold Bundles

#### Cached Bundles

</p></details>

<details><summary>packages/benchmarks/react-hn 🚨</summary><p>

#### Timings

| --- | --- | --- |

#### Cold Bundles

#### Cached Bundles

</p></details>

#### Timings

| --- | --- | --- |

#### Cold Bundles

#### Cached Bundles

</p></details>

**[Click here to view a detailed benchmark overview.](https://benchmark.parceljs.org/benchmark/260742428503310849)**

packages/transformers/js/src/helpers.js
```diff
@@ -0,0 +1,37 @@
+export function interopDefault(a) {
```
wbinnssmith: maybe this file could be called `esmodule-helpers.js` or something more specific?

mischnic:
You made casts to any and used `invariant(node.type ==` in many places.
Flow understands `if(isIdentifier(..)) { }` but not `if(t.isIdentifier(..)) { }`, so replace every occurence `t.is*` with `is*` and import them directly `import {isIdentifer} from '@babel/types'`.

And `NodePath<any>` should be `NodePath<Node>` (import that type from `@babel/types`).

devongovett:
> Flow understands if(isIdentifier(..)) { } but not if(t.isIdentifier(..)) { }

Oh, what? Why is that?

mischnic:
For no particular reason, it's just not implemented: https://github.com/facebook/flow/pull/7735

mischnic:
@devongovett Do you still want to merge this? Should I implement the flow changes I suggested?

devongovett:
Yeah I just haven’t had time yet, sorry!

mischnic:
In case we still want to do this:

1. The babel transformer replaces `this` in the module scope with `undefined` (if we don't do this in the transformer, the prelude wrapper could do `.call(module.isES6 ? undefined : module.exports)`)
2. Multiple reexports with (non)conflicting symbols should be handled gracefully: https://github.com/babel/babel/issues/11710

devongovett:
Finally had time to update this again. Made it even faster by completely removing `@babel/traverse` and using the traversal function I developed for the hoist/link perf PRs.

```
@babel/plugin-transform-modules-commonjs: 8.76s
custom transform using babel-traverse (existing PR): 6.92s (~21% faster)
custom transform without babel-traverse: 4.25s (~51% faster)
```

packages/transformers/js/src/visitors/modules.js
```diff
@@ -0,0 +1,376 @@
...
+  let helpersId;
+  let addHelpers = () => {
+    if (helpersId) {
+      return helpersId;
+    }
+
+    // Add a dependency so Parcel includes the helpers
+    let helperPath = path.join(__dirname, '..', 'esmodule-helpers.js');
+    let helperSpecifier = asset
+      ? relativePath(path.dirname(asset.filePath), helperPath)
```
mischnic, line 240: Could this instead be `@parcel/transformer-js/something/esmodule-helpers.js`? This relative path might contain a machine-specific part

packages/core/parcel/package.json
```diff
@@ -14,7 +14,7 @@
     "type": "git",
     "url": "https://github.com/parcel-bundler/parcel.git"
   },
-  "bin": "src/bin.js",
+  "bin": "lib/bin.js",
```
mischnic, line 17: 😉

packages/transformers/js/src/visitors/modules.js
```diff
@@ -0,0 +1,376 @@
...
+  imports: Array<
+    ImportDeclaration | ExportAllDeclaration | ExportNamedDeclaration,
+  >,
+  importNames: Map<
+    string,
+    {|name: string, default: string, namespace: string|},
+  >,
+  exports: Array<{|local: Expression, exported: Identifier|}>,
+  needsInteropFlag: boolean,
+  ...
```
mischnic, line 42: I don't see why this couldn't be an exact object type?

packages/transformers/js/src/visitors/modules.js
mischnic: This is probably causing this Windows test failure:
```
1) babel
     should support building with default babel config when running parcel globally:
   Error: Unknown pipeline ./D.
    at ResolverRunner.resolve (D:\a\1\s\packages\core\core\src\requests\/PathRequest.js:143:17)
    at Object.run (D:\a\1\s\packages\core\core\src\requests\/PathRequest.js:66:20)
    at RequestTracker.runRequest (D:\a\1\s\packages\core\core\src\/RequestTracker.js:535:20)
    at AssetGraphBuilder.runPathRequest (D:\a\1\s\packages\core\core\src\/AssetGraphBuilder.js:765:18)
    at PromiseQueue._runFn (D:\a\1\s\packages\core\utils\src\/PromiseQueue.js:84:7)
    at PromiseQueue._next (D:\a\1\s\packages\core\utils\src\/PromiseQueue.js:73:5)
```

mischnic:
We should also make sure that duplicate reexports works correctly (which [we fixed previously in Babel's transformer](https://github.com/babel/babel/issues/11710)): https://github.com/parcel-bundler/parcel/issues/4399#issuecomment-605636663

packages/transformers/js/src/visitors/modules.js
devongovett: hmm tried this, and it mostly works, except in a few tests that are libraries where "includeNodeModules" is implicitly false. Then the helpers aren't actually included in the bundle... I don't think we can get away with that because `@parcel/transformer-js` won't be a dependency of the package. Any ideas?

mischnic: I didn't think of that... and Yarn PnP would disallow it as well

devongovett: So go back to the relative path you think?

mischnic: Yeah, we don't have an alternative at the moment...

devongovett: Ok, actually ended up going back to the non-relative path because of one test that runs parcel "globally" which happens to be on a different drive letter on windows. To make this work, I had to add a `resolveFrom` option to dependencies so it would resolve relative to the plugin and not the source file, and also override the `env` of the dependency to always include node modules for that file. Currently `resolveFrom` is in addition to `sourceFile` (which wasn't publicly writable), and it falls back to `sourceFile` when not available. I'm not sure if these fields should be combined...

packages/utils/babylon-walk/src/scope.js
```diff
@@ -0,0 +1,219 @@
...
+          for (let id in ids) {
+            scope.addBinding(id, decl);
+            scope.addReference(ids[id]);
+          }
+        }
+      } else {
+        let ids = t.getBindingIdentifiers(node);
+        for (let id in ids) {
+          scope.addBinding(id, node);
+          scope.addReference(ids[id]);
```
mischnic, line 181: So references are not just references but also the identifier in the declaration?

mischnic:
This should print `1 4`, but after bundling it's `undefined 4` (namespace import includes default, only namespace reexport skips it)
```js
// other.mjs
import * as x from "./other.mjs";

export default 1;
export const a = 4;

console.log(x.default, x.a);
```

mischnic:
Does this support https://github.com/tc39/proposal-export-default-from ? We have the parser plugin enabled by default (`exportDefaultFrom` in babel-ast-utils)

devongovett:
No it doesn't. Should it given it's only a proposal?

packages/utils/babylon-walk/src/scope.js
devongovett: Hmm yeah I think this came from my other PR which was doing renaming. Could probably get rid of it

mischnic:
Should we remove the parser plugin as well then? At the moment it parses only if the babel transformer doesn't run and then you get an unhelpful assertion error from the babel cjs-esm plugin

https://github.com/parcel-bundler/parcel/issues/5452#issuecomment-743153918

devongovett:
Ah. Well maybe it should be the other way then? Disable the parser plugin by default, but do support it in the transformer. That way people can enable it by adding the syntax plugin to their babel config. If we didn't support transforming it, then they'd also have to use babel's modules transform but that might break some scope hoisting stuff...

devongovett:
Ok I actually removed all parser plugins from both babel transformer and js transformer. Both dynamic import and export namespace declarations are supported by default now. Export namespace is stage 4 and supported natively in chrome, so I've added support for that in the modules transformer. I added default export support as well, but it must be enabled by adding `@babel/plugin-syntax-export-default-from` to turn on the parser plugin.

packages/transformers/js/src/esmodule-helpers.js
```diff
@@ -28,6 +28,12 @@ exports.namespace = function(source, dest = {}) {
   return dest;
 };
 
+exports.namespace = function(source) {
+  let namespace = exports.exportAll(source, {});
+  namespace.default = source.default;
```
mischnic, line 33: But now `Object.keys` would list it?

mischnic (approved):
Maybe @wbinnssmith could give this a spin in case we forgot some obvious case?

wbinnssmith (changes requested):
I ran this on a codebase that used the following:

`export { default } from './foo';`

and

`export { default as foo } from './foo';`

and both are left behind as-is 😞

wbinnssmith (approved):
🎉

Issue #4389, opened 2020-03-26
parcelRequire is not defined (among other problems)


mischnic:
# 🐛 bug report

1. `parcelRequire` is used before it's defined
```js
(function () {
  var $a08e08a6fc4cf4272cd3bacd0a9f1$init = parcelRequire("012a08e08a6fc4cf4272cd3bacd0a9f1");
  var $parcel$modules = {};
  var $parcel$bundles = {};
  var globalObject = typeof globalThis !== 'undefined' ? globalThis : typeof self !== 'undefined' ? self : typeof window !== 'undefined' ? window : typeof global !== 'undefined' ? global : {};

  if (globalObject.parcelRequire == null) {
    globalObject.parcelRequire = function (name) {
      // Execute the bundle wrapper function if there is one registered.
      if (name in $parcel$bundles) {
        $parcel$bundles[name]();
        delete $parcel$bundles[name];
      }
// ...
```
2. Even when correcting problem (1), the module to import is not defined. Because the sibling bundle `<script>` tag in the html file was inserted after instead of before.

3. After fixing (2), notice that there are actually three bundles (two would make more sense here) and that it contains
```js
parcelRequire.registerBundle("012a08e08a6fc4cf4272cd3bacd0a9f1", function () {
  parcelRequire.register("012a08e08a6fc4cf4272cd3bacd0a9f1", $a08e08a6fc4cf4272cd3bacd0a9f1$exports);
});
```
so that exports object is missing. (This is the bundle that is dynamically imported).

The corresponding bundle graph is also very odd (an empty bundle?)

![BundleGraph](https://user-images.githubusercontent.com/4586894/77699303-75d55b00-6fb2-11ea-9d6e-19f0b6fa0ad0.png)

## 💻 Code Sample

```html
<script type="text/javascript" src="index.js"></script>
```

```js
// index.js
import lodash from "lodash";
console.log(lodash.add(1, 2));

import "./async";

// async.js
import("lodash").then(lodash => console.log(lodash.add(1, 2)));
```

## 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Parcel           | a1a2515ff8710c0ebc56fb6dc0aad72b6f6dec48

wbinnssmith:
> The corresponding bundle graph is also very odd (an empty bundle?)

This is because we unconditionally create new bundles when we encounter async dependencies, and then deduplicate assets when they're present in all ancestries in the graph.

[...]

PR #4375, opened 2020-03-24
Propagate symbols data in bundler

mischnic:
# ↪️ Pull Request

Would close #4370 and fix #4318

Instead of finding used imports per bundle, do this directly after inserting the runtimes. By directly updating `asset.meta.symbols` with this information, we can do cross-bundle treeshaking without major changes in the scope-hoisting code itself.

I've had to update the runtime mechanism to allow runtimes to insert themselves between an asset and a dependency (so act as a reexporting asset).

Potential downsides: `getUsedExports` is now single-threaded. But this is more of an API problem because https://github.com/parcel-bundler/parcel/blob/afd9efb9acc837eca8ac733d2d7cbd5cc27fbdff/packages/bundlers/default/src/DefaultBundler.js#L279-L281
is simply running per bundle and the resulting map could be merged later.

## 🚨 TODO

- [ ] shared bundles are still not shaken
- [ ] detect more dynamic import patterns (currently, only `import("./b.js").then(({ default: b }) => ...);` for testing)
- [ ] is a JSRuntime always correctly replacing a dependency?
- [ ] much else (CI)

For discussion:
- [ ] create new plugin (type) instead of `DefaultBundler#propagate`

PR #4295, opened 2020-03-09
Fix assets duplicated between sibling bundles, and parallel request limit behavior

devongovett:
When testing in site with lots of pages, I noticed two issues:

* Assets were duplicated between sibling bundles in the same bundle group
* When hitting the parallel request limit, we were skipping creating shared bundles if any bundle group had hit the limit instead of if all bundle groups had. This meant a single bundle group could prevent other bundle groups from being optimized.

I'm having trouble reproducing these issues in a smaller case, so opening the PR for feedback.

PR #4391, opened 2020-03-27
Sync and async dependencies on the same module in the same bundle

wbinnssmith:
Depends on #4407.

Related to #4389, but does not address matching sync and async imports originating in the module, at least not in the non-scope-hositing JSPackager [0].

This adds a new edge to the graph called `internal_async` to represent async dependencies that have been "internalized" — the bundler has determined that they don't need to point to an external bundle group. This allows the graph to reflect that there is no longer a relationship between the bundle and the bundle group created by the async dependency (there is no `'bundle'` edge between the bundle and the bundle group).

If all dependencies on a bundle group are internalized, the bundle group is removed, preventing empty and unused bundles from being created.

This also refactors JSRuntime and the BundleGraph API, removing `getExternalDependencies` in favor of `getDependenciesInBundle` and allowing the runtime access to all dependencies directly. `resolveExternalDependency` now can return an Asset, indicating that the asset can be safely required directly.

Notes:
- [x] ~This doesn't work with scope hoisting. Tests are failing because of this. Is there a way to reference an asset by id without adding a dependency in a way that is understood by scope hoisting? cc @mischnic~ This now works with scope hoisting.

Test Plan: Added tests for cases where the asset was available in the current bundle as well as in its ancestry. `yarn test`

[0] To address sync and async imports originating in the same module in the standard JSPackager, we will need to disambiguate synchronously requiring a module from synchronously requiring a runtime that asynchronously loads that module — currently they both use the same module specifier.


parcel-benchmark:
## Benchmark Results
<details><summary>packages/benchmarks/three 🚨</summary><p>

#### Timings

| --- | --- | --- |

#### Cold Bundles

#### Cached Bundles

</p></details>

#### Timings

| --- | --- | --- |

#### Cold Bundles

#### Cached Bundles

</p></details>

<details><summary>packages/benchmarks/react-hn 🚨</summary><p>

#### Timings

| --- | --- | --- |

#### Cold Bundles

#### Cached Bundles

</p></details>

#### Timings

| --- | --- | --- |

#### Cold Bundles

#### Cached Bundles

</p></details>

**[Click here to view a detailed benchmark overview.](https://benchmark.parceljs.org/benchmark/261138610344952329)**

packages/runtimes/js/src/JSRuntime.js
```diff
@@ -67,135 +67,80 @@ export default new Runtime({
...
-      let [firstBundle] = bundlesInGroup;
-      if (firstBundle.isInline) {
+      if (resolved.type === 'asset') {
+        // If this bundle already has the asset this dependency references,
+        // return a simple runtime of `Promise.resolve(parcelRequire("assetId"))`.
         assets.push({
-          filePath: path.join(__dirname, `/bundles/${firstBundle.id}.js`),
-          code: `module.exports = ${JSON.stringify(dependency.id)};`,
+          filePath: __filename,
+          code: `module.exports = Promise.resolve(parcelRequire(${JSON.stringify(
```
devongovett: I'm not sure it's always safe to use `parcelRequire` here - that relies on an implementation detail of the packager. For example, the commonjs and ESM output formats don't have this. We'll need to make a real dependency with `require` which will end up getting compiled correctly by the packager.

packages/bundlers/default/src/DefaultBundler.js
```diff
@@ -151,6 +169,31 @@ export default new Bundler({
         bundleGraph.addAssetGraphToBundle(asset, bundle);
       }
     }
+
+    for (let [dependency, {bundleGroup, resolution}] of asyncDependencies) {
```
devongovett: Should this be done after optimization? Will shared bundles affect this?

mischnic:
I'm probably missing something but why can't we let the async dependencies point at assets in the same bundle? So isAsync would mean "I am fine with getting this asynchronously, but Parcel can do what it thinks is better". In that case the JSRuntime would wrap the synchronous import (in case of scope hoisting just the exports object) with a Promise.

>  This doesn't work with scope hoisting. Tests are failing because of this. Is there a way to reference an asset by id without adding a dependency in a way that is understood by scope hoisting? 

A dependency signifies (or at least should signify) that the values in `dep.symbols` influence how the importing asset is executed, so no.
Would this change to how JSRuntimes are inserted work for you? This would be more correct anyway (since JSRuntimes are undoubtably just reexporting assets). (I implemented this already in https://github.com/parcel-bundler/parcel/pull/4375, you can copy that if you want)
Bildschirmfoto 2020-03-27 um 18 35 24

wbinnssmith:
> I'm probably missing something but why can't we let the async dependencies point at assets in the same bundle?

Dependencies can be present in multiple bundles, some of which may have the asset and some may not. There's a test for this 🙂

> So isAsync would mean "I am fine with getting this asynchronously, but Parcel can do what it thinks is better".

Yep, that's how I've been thinking about this as well.

packages/bundlers/default/src/DefaultBundler.js
wbinnssmith: Probably. I don't think it should affect any of the deduplication or heuristics we currently use.

packages/runtimes/js/src/JSRuntime.js
wbinnssmith: Yeah, this is what's breaking the scope hoisting tests.

How can we create a real dependency on the asset from here? `require()`? If so, what would we use as the module specifier?

mischnic:
> Dependencies can be present in multiple bundles, some of which may have the asset and some may not.

Why is this a problem? An async dependency would point to the asset directly if it's in the current bundle and otherwise to the bundle (like we do now).

wbinnssmith:
> An async dependency would point to the asset directly if it's in the current bundle and otherwise to the bundle (like we do now).

By "point to", do you mean an edge in the graph? The dependency is represented by a single node that can be reached by multiple bundles. How would the edge reflect a state for some bundles (that have the asset) but not others?

wbinnssmith:
> I was wondering about doing this the other way. Rather than keeping the asset in the same bundle, keep the separate bundle (prefer async over sync). Then hoist the async bundle to the same bundle group so it's loaded in parallel and available at the time of the sync require. 

How would this work with entry bundles, where it's not safe to add additional bundles to their group as it becomes unpredictable?

> Which approach do we think is better? What do other bundlers do?

Webpack produces one bundle in the first test case [0], where we currently produce two bundles each in different bundle groups, and this proposal would mean two bundles in a single group. This PR also only produces one bundle.

[0] https://github.com/parcel-bundler/parcel/pull/4391/files#diff-266472ddae511d80d5622d83e42d935bR2322

Webpack also produces two bundles in two separate groups for the second case [1], as does this PR. This proposal would mean two bundles in the initial group, and another in an additional group, yeah?

[1] https://github.com/parcel-bundler/parcel/pull/4391/files#diff-266472ddae511d80d5622d83e42d935bR2337

packages/bundlers/default/src/DefaultBundler.js
wbinnssmith: Another thing is we could move the "internalizing" of dependencies in bundles into core, but it would make bundle group creation/deletion asymmetric. It also feels better to put this in the control of the bundler plugin, even though it feels like we're leaking a detail requiring the bundler to do this.

packages/core/core/src/BundleGraph.js
```diff
@@ -253,37 +336,69 @@ export default class BundleGraph {
...
+        return;
+      }
+
+      if (descendant !== bundle && bundleHasReference(descendant)) {
+        isReferenced = true;
+        actions.stop();
+        return;
+      }
+
+      let similarSiblings = this.getSiblingBundles(descendant).filter(
```
devongovett, line 383: Does this fix the first issue in #4295?

packages/runtimes/js/src/JSRuntime.js
```diff
@@ -67,135 +67,86 @@ export default new Runtime({
...
-      if (firstBundle.isInline) {
+      if (resolved.type === 'asset') {
+        // If this bundle already has the asset this dependency references,
+        // return a simple runtime of `Promise.resolve(require("path/to/asset"))`.
         assets.push({
-          filePath: path.join(__dirname, `/bundles/${firstBundle.id}.js`),
-          code: `module.exports = ${JSON.stringify(dependency.id)};`,
+          filePath: path.join(options.projectRoot, 'JSRuntime.js'),
+          // Using Promise['resolve'] to prevent Parcel from inferring this is an async dependency.
+          // TODO: Find a better way of doing this.
```
devongovett, line 98: 😄

packages/core/core/src/BundleGraph.js
wbinnssmith: I think that issue is an issue in `isAssetReferencedByAncestorBundle` — the inverse of this function — where we don't detect at deduplication time that the asset _can_ be referenced, so we leave it in the bundle.

Issue #3351, opened 2019-08-11
Parcel 2: Elm transform


devongovett:
Create the `@parcel/transformer-elm` package in `packages/transforms/elm`. This package is a Parcel 2 transform plugin for Elm. It should be based on the ElmAsset from Parcel 1.

The `@parcel/transformer-elm` package is responsible for the following things:

* Ensuring the elm compiler is installed as a dependency
* Creating `elm.json` if it doesn't exist, or loading the existing config.
* Transforming Elm to JavaScript
* Collecting dependencies
* Injecting `elm-hot` runtime
* Compressing using the recommended terser settings? Possibly move this to an optimizer plugin if it makes sense to run it over the entire bundle rather than individual assets.

Related issues that could possibly also be fixed: #3188 #2818 #2508

Drapegnik:
Hi! Could I take it?

MattCheely:
I started looking into this a bit last week while on vacation, but ran into some build issues with the v2 which seem to have been fixed. I was going to look into it a bit more tonight and offer to pick the issue up, but I see @Drapegnik, you are also interested. Maybe we could collaborate? Either way I will probably plug away at it a bit more on my own, to get a better understanding of the internals of parcel 2 at least. I won't be put out if someone else implements it in the meantime though!

MattCheely:
I was able to get a basic transform working, but need to figure out how to properly track internal Elm dependencies, and work on compiler error formatting.

MattCheely:
After a decent amount of effort, I'm out of time to continue working on this.  I'm unable to get parcel to track changes to files beyond the main Elm entry point. It was [suggested on spectrum](https://spectrum.chat/parcel/contributing/parcel-v2-equivalent-of-includedinparent~e34c80d0-efe4-434a-9455-e05a7c6387e2) that I use `addConnectedFile` to set that up, but it doesn't seem to work, and I haven't been able to find anything in the core Parcel code where that would get set up that I can debug.  I was able to get a basic build working, and if anyone wants to pick up where I left off, [my elm-transform branch is here](https://github.com/MattCheely/parcel/tree/elm-transform).

harrysarson:
I had a look at your branch but the diff:

> Showing 742 changed files with 50,943 additions and 7,631 deletions. 

May I ask why there are so many changes and would it be possible to point me to the best place to start looking through them?

MattCheely:
> May I ask why there are so many changes and would it be possible to point me to the best place to start looking through them?

[...]

Issue #4264, opened 2020-03-04
Invalidate cache on env var dependency


ranisalt:
# 🙋 feature request

I have a configuration file that depends on env vars to select a couple values used throughout the system.

## 🤔 Expected Behavior

A file should be recompiled if it depends on environment variables that changed from the previous compilation.

## 😯 Current Behavior

If a file depends on environment variables that change, any file that depends on it will receive an outdated cached version and possibly the wrong values if they depend on that env var.

## 💁 Possible Solution

Maybe the cache should smartly tag which environment variables are used and store alongside the cached build, I think this is workable with a proxy to `process.env`

## 🔦 Context

I select the API endpoint to my requests depending on the environment, e.g. `process.env.ENDPOINT === 'staging' ? 'staging.example.com' : 'example.com`, and also a couple more values that depend on endpoint and build mode.

I can't use .env files since it is a monorepo and Parcel does not read them from workspaces (#4110) so I have been using cross-env to set them.

PR #4395, opened 2020-03-27
Transformer for Elm

nilshelmig:
# ↪️ Pull Request
This Pull Requests is going to implement #3351.
Feedback is welcomed and much appreciated :smiley: 

### Prerequirements
Create an simple project with an elm file included in a js file. Minimal example is given at parceljs.org/elm.html

### Transformer creates `elm.json`
When building with `parcel build/watch/serve index.html`
Then a `elm.json` is created

### Transformer optimizes code
When building with `parcel build [entry]`
Then the output code is optimized and unnecessary code is deleted

### Transformer don't optimizes code when explicit wanted 
When building with `parcel build --no-minify [entry]`
Then the output code is not optimized and unnecessary code is still there

### Transformer injects HMR-Runtime
When building with `parcel watch/serve [entry]`
And editing an `.elm` file
Then the change is applied to the output without needing to reload the website

### Transformer won't inject HMR-Runtime when explicit wanted
When building with `parcel watch/serve --no-hmr [entry]`
And editing an `.elm` file
Then the change is not applied to the output and the website won't change without reloading

### Transformer collects dependencies
I don't know how to test this!?

###  Transformer installs elm when not installed globally
When `elm` is not installed on the machine
And building with `parcel build/watch/serve [entry]`
Then `elm` should be installed as local dependency in the `package.json`
And build succeeds

### Transformer won't install elm when not installed globally and no autoinstall
When `elm` is not installed on the machine
And building with `parcel build/watch/serve --no-autoinstall [entry]`
Then build fails due to missing `elm` binary

### Transformer enables Elm debugger in debug mode
When building with `parcel watch/serve [entry]`
Then the elm time-travel-debugger is available

### Transformer disables Elm debugger in production mode
When building with `parcel build [entry]`
Then the elm time-travel-debugger is not available

### Transformer disables Elm debugger in debug mode when PARCEL_ELM_DISABLE_DEBUG is set (See #2818)
When building with `parcel watch/serve [entry]`
And environment variable `PARCEL_ELM_DISABLE_DEBUG` is set to `true`/`1`
Then the elm time-travel-debugger is not available

## ✔️ PR Todo

- [x] Test instructions for `Transformer collects dependencies`
- [x] Fix resolving of `elm` binary when installed via **autoinstall**
- [x] Fix installing `elm` fails on first run
- [x] Fix Flow type errors


nilshelmig:
I've managed to install elm via the `options.packageManager` but I got an error on the first run.
It seems to me that `options.packageManager.install` tries to load the installed module. `elm` can't be loaded as module because it's just a binary and has nothing to load from js. Anyone got an idea how to fix this?

nilshelmig:
I would like to challenge the autoinstall feature and ensuring `elm.json` file for this transformer.

Wouldn't every elm developer using parcel already has elm installed globally or as local dependency? 
Also I assume that the `elm.json` file is already created when someone creates an `.elm` file.

The only use case I know is the simple example of just installing parcel and create a `Main.elm` file. But is that a real life use case?

packages/transformers/elm/src/ElmTransformer.js
```diff
@@ -0,0 +1,125 @@
...
+import {minify} from 'terser';
+import ThrowableDiagnostic from '@parcel/diagnostic';
+
+export default new Transformer({
+  async getConfig({asset, options}) {
+    const installPackage = async name => {
+      if (!options.autoinstall) {
+        throw new ThrowableDiagnostic({
+          diagnostic: {
+            message: `Denpendency '${name}' is not installed and autoinstall is turned off. Either install dependency manually or enable autoinstall`,
```
mischnic: If you do `await options.packageManager.resolve('elm', asset.filePath, {autoinstall: options.autoinstall});` this will handled automatically,

mischnic:
- Could we use `options.minify` instead of `asset.env.minify`? `options.minify` is set by `--no-minify` and also controls if terser runs on js bundles.
- I think this doesn't watch the elm source files. These should be added with `asset.addIncludedFile("..path/to/included..")` so that elm is actually invoked again once a file has changed.

> Wouldn't every elm developer using parcel already has elm installed globally or as local dependency?

I would do it like this (maybe you have a elm.json and not added it as a npm dependency):
```js
if(!binaryInPath("elm"){
	// this will handle autoinstalling (or throw) automatically:
	await options.packageManager.resolve('elm', asset.filePath, {autoInstall: options.autoinstall});
}

// ...

childprocess.exec("elm", ....)
```

>Also I assume that the elm.json file is already created when someone creates an .elm file.
>
>The only use case I know is the simple example of just installing parcel and create a Main.elm file. But is that a real life use case?

I think we shouldn't automatically create config files, so we could throw an error like `No elm.json could be found, please run "elm init" to create one.`

nilshelmig:
@mischnic Thanks for your feedback :heart: 
> Could we use `options.minify` instead of `asset.env.minify`? `options.minify` is set by `--no-minify` and also controls if terser runs on js bundles.

I would love to, but the interface [`PluginOptions`](https://github.com/parcel-bundler/parcel/blob/9bb19fc6680817cf657e077097b45cbc2cfd869b/packages/core/types/index.js#L212) doesn't provide `.minify`. So I had to use `asset.env.minify`

> I think this doesn't watch the elm source files. These should be added with `asset.addIncludedFile("..path/to/included..")` so that elm is actually invoked again once a file has changed.

It's already done [here](https://github.com/parcel-bundler/parcel/pull/4395/files#diff-c741c351c2e6cfe757a793b982bd6db9R40) :)  
Watching files included in `Main.elm` works. I've tested it against https://github.com/rtfeldman/elm-spa-example

> I would do it like this (maybe you have a elm.json and not added it as a npm dependency):
> 
> ```js
> if(!binaryInPath("elm"){
> 	// this will handle autoinstalling (or throw) automatically:
> 	await options.packageManager.resolve('elm', asset.filePath, {autoInstall: options.autoinstall});
> }
> 
> // ...
> 
> childprocess.exec("elm", ....)
> ```

Autoinstalling elm as dependency currently brings a lot of hassle. After a lot of work I've managed to get it working, but only on the second run ob parcel. As mentioned [in my previous comment](https://github.com/parcel-bundler/parcel/pull/4395#issuecomment-605639034) resolving "elm" fails, because it is a npm package with only a binary file and no JavaScript.

If the transformer should be capable of autoinstalling elm, then `.install` of `NodePackageManager` should not resolve the installed package, as I assumed it. The transformer need a way to just install a dependency.  
If the transformer doens't need autoinstall then I would suggest a simple error message like
```
`elm` is not installed globally or as a local dependency. Please install elm from https://elm-lang.org or via the `elm` npm package
```

> > Also I assume that the elm.json file is already created when someone creates an .elm file.
> > The only use case I know is the simple example of just installing parcel and create a Main.elm file. But is that a real life use case?
> 
> I think we shouldn't automatically create config files, so we could throw an error like `No elm.json could be found, please run "elm init" to create one.`
Give that there is no use case that a user starts an elm project without running init, then you are right.  
And no need to check for `elm.json, because elm already gives a great user experience when compiling without `elm.json`:
```
@parcel/transformer-elm: Compilation failed
-- NO elm.json FILE ------------------------------------------------------------

It looks like you are starting a new Elm project. Very exciting! Try running:

    elm init

It will help you get set up. It is really simple!
```

mischnic:
> I would love to, but the interface PluginOptions doesn't provide .minify. So I had to use asset.env.minify

Sorry, I mixed up options.env and asset.env. You're right.

> because it is a npm package with only a binary file and no JavaScript.

Then do: `options.packageManager.resolve('elm/package.json')`

> NO elm.json FILE

I appreciate the effort, but to align this with the existing error message, it should be a bit more neutral. The `Diagnostic` interface also accepts a hint which seems appropriate here:
```js
import ThrowableDiagnostic from '@parcel/diagnostic';
  diagnostic: {
    message: "...no elm.json...",
    hints: ["...elm init..."],
	origin: "@parcel/transformer-elm",
	filePath: asset.filePath
  }
});
```

nilshelmig:
I have some problems with flow. How do I add a parameter in the type info `terserOptions`? It's missing the `rename` option

mischnic:
Here is the Terser libdef we are using: https://github.com/parcel-bundler/parcel/blob/v2/flow-libs/terser.js.flow

But I don't think there is a rename option? Not even in the official typescript definitons: https://github.com/terser/terser/blob/master/tools/terser.d.ts

nilshelmig:
Good catch! Then I can remove the option 🙂

nilshelmig:
> > because it is a npm package with only a binary file and no JavaScript.
> 
> Then do: `options.packageManager.resolve('elm/package.json')`

resolving `elm/package.json` don't install elm as an dependencie. So I don't know how to solve that parcel is auto installing elm :worried:

mischnic:
Please apply this change, it should work then:
```diff
diff --git packages/core/package-manager/src/installPackage.js packages/core/package-manager/src/installPackage.js
index 1fcfaab1e..596354300 100644
--- packages/core/package-manager/src/installPackage.js
+++ packages/core/package-manager/src/installPackage.js
@@ -65,7 +65,8 @@ async function installPeerDependencies(
   options,
 ) {
   let basedir = path.dirname(from);
-  const {resolved} = await resolve(fs, module.name, {
+  // This package might not have a `main` field, resolve `package.json` instead
+  const {resolved} = await resolve(fs, `${module.name}/package.json`, {
     basedir,
     range: module.range,
   });
```

nilshelmig:
> Please apply this change, it should work then:
> 
> ```diff
> diff --git packages/core/package-manager/src/installPackage.js packages/core/package-manager/src/installPackage.js
> index 1fcfaab1e..596354300 100644
> --- packages/core/package-manager/src/installPackage.js
> +++ packages/core/package-manager/src/installPackage.js
> @@ -65,7 +65,8 @@ async function installPeerDependencies(
>    options,
>  ) {
>    let basedir = path.dirname(from);
> -  const {resolved} = await resolve(fs, module.name, {
> +  // This package might not have a `main` field, resolve `package.json` instead
> +  const {resolved} = await resolve(fs, `${module.name}/package.json`, {
>      basedir,
>      range: module.range,
>    });
> ```

Same problem. Installs `elm` package but fails. When running the second time it works.
```
🚨 Build failed.
@parcel/transformer-elm: Cannot find module 'elm' from '/data/Dokumente/parcel/test'
Error: Cannot find module 'elm' from '/data/Dokumente/parcel/test'
    at /data/Dokumente/parcel/node_modules/resolve/lib/async.js:107:35
    at processDirs (/data/Dokumente/parcel/node_modules/resolve/lib/async.js:260:39)
    at callback (/data/Dokumente/parcel/node_modules/resolve/lib/async.js:267:32)
    at isDirectory (/data/Dokumente/parcel/packages/core/utils/src/resolve.js:113:9)
```

And when I'm using `options.packageManager.resolve('elm/package.json')` with the patch applied, then elm will not be installed
```
@parcel/workers: spawn /data/Dokumente/parcel/test/node_modules/elm/bin/elm ENOENT
Error: spawn /data/Dokumente/parcel/test/node_modules/elm/bin/elm ENOENT
    at Process.ChildProcess._handle.onexit (internal/child_process.js:267:19)
    at onErrorNT (internal/child_process.js:467:16)
    at processTicksAndRejections (internal/process/task_queues.js:84:21)
```

mischnic:
I have:
- create a new folder with index.html, index.js, Main.elm as per the example in the docs.
- checkout your branch, merge v2 into it, apply this change:
```diff
diff --git packages/core/package-manager/src/installPackage.js packages/core/package-manager/src/installPackage.js
index 1fcfaab1e..596354300 100644
--- packages/core/package-manager/src/installPackage.js
+++ packages/core/package-manager/src/installPackage.js
@@ -65,7 +65,8 @@ async function installPeerDependencies(
   options,
 ) {
   let basedir = path.dirname(from);
-  const {resolved} = await resolve(fs, module.name, {
+  // This package might not have a `main` field, resolve `package.json` instead
+  const {resolved} = await resolve(fs, `${module.name}/package.json`, {
     basedir,
     range: module.range,
   });
diff --git packages/transformers/elm/src/ElmTransformer.js packages/transformers/elm/src/ElmTransformer.js
index 7a0183dc2..31905f7e2 100644
--- packages/transformers/elm/src/ElmTransformer.js
+++ packages/transformers/elm/src/ElmTransformer.js
@@ -64,7 +64,7 @@ async function pathToElm(root, installPackage) {
   if (!commandExists.sync('elm')) {
     const elmBin = path.resolve(root, 'node_modules/elm/bin/elm');

-    if (!fs.existsSync(elmBin)) await installPackage('elm');
+    if (!fs.existsSync(elmBin)) await installPackage('elm/package.json');
     return elmBin;
   }
```

and it worked:
```
Building index.html...
Building index.js...
Building Main.elm...
Installing elm...
@parcel/package-manager: yarn: package.json: No license field
@parcel/package-manager: yarn: No license field
yarn: [1/4] Resolving packages
@parcel/package-manager: yarn: elm > request@2.88.2: request has been deprecated, see https://github.com/request/request/issues/3142
yarn: [2/4] Fetching packages
yarn: [3/4] Linking dependencies
yarn: [4/4] Building fresh packages
@parcel/package-manager: yarn: Saved lockfile.
@parcel/package-manager: yarn: No license field
@parcel/package-manager: yarn: Saved 47 new dependencies.
@parcel/package-manager: yarn: Direct dependencies
@parcel/package-manager: yarn: All dependencies
Bundling...
Packaging undefined...
Optimizing undefined...
Packaging undefined...
Optimizing undefined...
✨ Built in 18.20s

dist/elm.28e86ea4.js    105.5 KB     2.23s
├── Main.elm            88.19 KB    13.17s
└── index.js                94 B     721ms

dist/index.html             91 B     140ms
└── index.html             107 B     147ms
```

nilshelmig:
@mischnic
I've rebased the branch on the latest v2 branch. Tested it under linux and everything seems to work fine. I would appreciate if you take a look and tell me if something is missing. Would like to get this merged :)

packages/transformers/elm/src/ElmTransformer.js
```diff
@@ -0,0 +1,128 @@
...
+import {inject as injectElmHMR} from 'elm-hot';
+import {minify} from 'terser';
+import ThrowableDiagnostic from '@parcel/diagnostic';
+
+export default new Transformer({
+  async transform({asset, options}) {
+    const config = {
+      cwd: path.dirname(asset.filePath),
+      debug: !(
+        options.env.PARCEL_ELM_NO_DEBUG || options.mode === 'production'
```
mischnic, line 17: No, environment variables aren't factored into the cache key. So this would require you to delete the cache directory when changing that variable.

Can this somehow be read from elm.json?

packages/core/parcel/package.json
```diff
@@ -10,7 +10,7 @@
     "type": "git",
     "url": "https://github.com/parcel-bundler/parcel.git"
   },
-  "bin": "src/bin.js",
+  "bin": "lib/bin.js",
```
mischnic, line 13:
```suggestion
  "bin": "src/bin.js",
```

packages/transformers/elm/src/ElmTransformer.js
```diff
@@ -0,0 +1,128 @@
...
+async function pathToElm(root, installPackage) {
+  if (!commandExists.sync('elm')) {
+    const elmBin = path.resolve(root, 'node_modules/elm/bin/elm');
+
+    if (!fs.existsSync(elmBin)) await installPackage('elm');
+    return elmBin;
+  }
+
+  return undefined; // use globally installed elm
+}
```
mischnic, line 70: This would break with Yarn PnP, it should be something like this:
```js
async function ensureElmInstalled(asset, options) {
  if (!commandExists.sync('elm')) {
    await options.packageManager.resolve('elm/package.json', asset.filePath, {
      autoinstall: options.autoinstall,
    });
  }
}
```
and then you spawn `elm` while relying on npm/npx/Yarn to have injected the bin folder into the PATH.


I'm also still getting errors from `@parcel/package-manager` (`@parcel/transformer-elm: Cannot find module 'elm' from '.../elm/src'`), there is probably still some code where the `/package.json` part of `elm/package.json` is removed incorrectly.

packages/transformers/elm/src/ElmTransformer.js
nilshelmig: Didn't know that `options.env` is cached. Elm doens't allow third party options in `elm.json`.
Is `process.env` accessable in Transformers?

WalidBehlock:
Any update on the state of this PR?

packages/transformers/elm/src/ElmTransformer.js
mischnic: No, using `options.env` is correct.
The problem is that the resolver just doesn't rerun at all if only the environment variable changes. But this is really a bug : https://github.com/parcel-bundler/parcel/issues/4264. I guess it's fine for now..

nilshelmig:
@WalidBehlock I work a lot on other side projects at the moment. But I can spend some time today to rebase my stuff and maybe integrate all the feedback :)

sephii:
Thanks for taking the time to work on this! Do you have any news on the feedback integration? Can I help in any way? I'd love to see this merged!

packages/transformers/elm/src/ElmTransformer.js
domenkozar: @mischnic could autoinstall be optional for elm until someone implements it? It's a bit a shame this whole PR is stalled for that feature which the author doesn't care much about (I think that's understandable).

mischnic: Yes, let's do that

mischnic:
Alright, let's remove autoinstall for now:

1. so we don't need `installPackage` and `ensureElmIsInstalled` should merely check if `elm` is available as a binary and show an error otherwise (setting "hints" to a message how it can/should be installed)
2. It shouldn't automatically create a config file for elm if missing (https://github.com/parcel-bundler/parcel/pull/4395#issuecomment-605685860), but rather show an appropriate error (I think suggesting running `elm init?`)

nilshelmig:
@mischnic Alright thanks for that. I've rebased to the latest v2 branch and applied the your feedback.
Looking forward to finally get this merged :slightly_smiling_face:

packages/transformers/elm/package.json
```diff
@@ -0,0 +1,33 @@
...
+  "dependencies": {
+    "@parcel/diagnostic": "^2.0.0-beta.1",
+    "@parcel/plugin": "^2.0.0-beta.1",
+    "command-exists": "^1.2.8",
+    "terser": "^5.2.1"
+  },
+  "peerDependencies": {
+    "elm": "^0.19.1-3 || 0.19.0-no-deps",
+    "elm-hot": "^1.1.4",
+    "node-elm-compiler": "^5.0.4"
```
mischnic, line 31: These should be removed now, they're unnecessary with `options.packageManager.require`.

mischnic:
`commandExists.sync('elm')` doesn't work for me (at least on macOS), and `node-elm-compiler` doesn't find it either.

That's what `findElm` was for in my version here: https://github.com/parcel-bundler/parcel/pull/4395#issuecomment-727239638.

packages/transformers/elm/src/ElmTransformer.js
```diff
@@ -0,0 +1,125 @@
+import {Transformer} from '@parcel/plugin';
+import commandExists from 'command-exists';
+import path from 'path';
+import {minify} from 'terser';
+import ThrowableDiagnostic from '@parcel/diagnostic';
+
+export default new Transformer({
+  async transform({asset, options}) {
+    const elm = await setupElmCompiler(asset, options);
```
mischnic, line 9: Can you inline this call to `setupElmCompiler `, calling a helper function that also calls even more helper functions becomes hard to read...

packages/transformers/elm/src/ElmTransformer.js
```diff
@@ -0,0 +1,125 @@
...
+
+    const config = {
+      cwd: path.dirname(asset.filePath),
+      debug: !options.env.PARCEL_ELM_NO_DEBUG && options.mode !== 'production',
+      optimize: asset.env.minify,
+    };
+    asset.invalidateOnEnvChange('PARCEL_ELM_NO_DEBUG');
+    ResolveDependenciesFromAsset(elm, asset);
```
mischnic, line 17: Ideally run the for loop directly in `transform`, or at least rename it to be lower case. This is not a class

packages/transformers/elm/src/ElmTransformer.js
```diff
@@ -0,0 +1,125 @@
...
+  });
+}
+
+function ensureElmIsInstalled() {
+  if (!commandExists.sync('elm')) {
+    throw new ThrowableDiagnostic({
+      diagnostic: {
+        message: "Can't find 'elm' binary.",
+        hints: [
+          "You can add it as an dependency for your project by running 'npm add -D elm' or 'npm add -D elm'.",
```
mischnic, line 46: One of there should be Yarn instead of npm twice 😄

nilshelmig: True :sweat_smile:

packages/core/integration-tests/test/elm.js
```diff
@@ -1,59 +1,89 @@
...
-    });
+    await assertBundles(b, [
+      {
+        type: 'js',
+        assets: [
+          'HMRRuntime.js',
+          'Main.elm',
+          'index.js',
+          'react-refresh-runtime.development.js',
+          'ReactRefreshRuntime.js',
```
mischnic, line 46: Can you add these two files (containing `{}` and being empty, respectively):
```
	packages/core/integration-tests/test/integration/elm/package.json
	packages/core/integration-tests/test/integration/elm/yarn.lock
```

Then the React related HMR assets aren't bundled here (as it should be).

nilshelmig:
As I don't work with `flow` I'll just straight forward take your code! Thanks for that 😉

packages/core/integration-tests/test/elm.js
nilshelmig: Sorry I don't quite understand what you mean by adding those two files? Should they be added in the `assertBundles` list or in `assets` from the `js`?

mischnic: No, sorry 😄 . Create these files and commit them. This will make Parcel use `packages/core/integration-tests/test/integration/elm` as the root for this test fixture (instead of the repo root, where the package.json contains React leading to these assets).

mischnic:
Last two remaining items:
1. https://github.com/parcel-bundler/parcel/pull/4395#discussion_r528919936

2. I didn't think about this for my previous comment, but now you actually first get the error about the missing elm.json file (even if elm is not a dependency). So it would probably make sense to ensure that elm is available before suggesting running `elm init`:
```diff
diff --git packages/transformers/elm/src/ElmTransformer.js packages/transformers/elm/src/ElmTransformer.js
index 23c74153..cab74656 100644
--- packages/transformers/elm/src/ElmTransformer.js
+++ packages/transformers/elm/src/ElmTransformer.js
@@ -8,7 +8,11 @@ import nullthrows from 'nullthrows';
 import ThrowableDiagnostic from '@parcel/diagnostic';
 
 export default (new Transformer({
-  async loadConfig({config}) {
+  async loadConfig({config, options}) {
+    // Make sure to (tell the user to) first install elm
+    // and only then suggest running `elm init`.
+    await elmBinaryPath(config.searchPath, options);
+
     let elmConfig = await config.getConfig(['elm.json']);
     if (!elmConfig) {
       throw new ThrowableDiagnostic({
@@ -25,7 +29,7 @@ export default (new Transformer({
   },
 
   async transform({asset, options}) {
-    const elmBinary = await elmBinaryPath(asset, options);
+    const elmBinary = await elmBinaryPath(asset.filePath, options);
     const elm = await options.packageManager.require(
       'node-elm-compiler',
       asset.filePath,
@@ -77,11 +81,11 @@ async function elmBinaryPath(asset, options) {
   return elmBinary;
 }
 
-async function resolveLocalElmBinary(asset, options) {
+async function resolveLocalElmBinary(from, options) {
   try {
     let result = await options.packageManager.resolve(
       'elm/package.json',
-      asset.filePath,
+      from,
       {autoinstall: false},
     );
 ```

**Comment by nilshelmig** [comment]
All done! I've changed the resolving of `elm` binary that the path will be available in the configs so that the transformer don't need to resolve the binary twice :slightly_smiling_face:

**Comment by mischnic** [comment]
That's a valid point, but now the absolute path to the binary might be used as a cache key, which we generally try to avoid (this could also cause bundle hashes to be different across machines). What do you think about doing this instead:
```js
  async loadConfig({config, options}) {
    let elmConfig = await config.getConfig(['elm.json']);
    if (!elmConfig) {
      // Make sure to (tell the user to) first install elm
      // and only then suggest running `elm init`.
      await elmBinaryPath(config.searchPath, options);
      throw new ThrowableDiagnostic(...);
    } 
    config.setResult(elmConfig});
  } 
```

Now, `elmBinaryPath` is also only ever executed once.

nilshelmig:
> That's a valid point, but now the absolute path to the binary might be used as a cache key, which we generally try to avoid (this could also cause bundle hashes to be different across machines). What do you think about doing this instead:
> 
> ```js
>   async loadConfig({config, options}) {
>     let elmConfig = await config.getConfig(['elm.json']);
>     if (!elmConfig) {
>       // Make sure to (tell the user to) first install elm
>       // and only then suggest running `elm init`.
>       await elmBinaryPath(config.searchPath, options);
>       throw new ThrowableDiagnostic(...);
>     } 
>     config.setResult(elmConfig});
>   } 
> ```
> 
> Now, `elmBinaryPath` is also only ever executed once.
@mischnic 
I don't quite understand how the path could be used as an cache key? If it's from the configResult, then this would also apply to the content of `elm.json`. Is that the wanted behavior?

And how can the transform function access the elmBinaryPath? It's seems the result is not used

mischnic:
> to the content of elm.json. Is that the wanted behavior?

Yes

> I don't quite understand how the path could be used as an cache key?

We hash the config result object, if that changed, then the transformer is reexecuted. It might be possible that in the future, `asset.id` is derived from that hash. Another idea is having a global cache (used by multiple machines), this is where absolute paths don't work either.

> And how can the transform function access the elmBinaryPath? It's seems the result is not used

I meant "only every executed once per asset", it's still needed in these two locations:

```js
  async loadConfig({config, options}) {
    let elmConfig = await config.getConfig(['elm.json']);
    if (!elmConfig) {
      // Make sure to (tell the user to) first install elm
      // and only then suggest running `elm init`.
      await elmBinaryPath(config.searchPath, options);
      throw new ThrowableDiagnostic(...);
    } 
    config.setResult(elmConfig});
  } ,

  async transform({asset, options}) {
    const elmBinary = await elmBinaryPath(asset.filePath, options);
    const elm = await options.packageManager.require(
```

nilshelmig:
@mischnic I see the problem now. Now everything should be ready to merge :smiley:

packages/core/types/index.js
```diff
@@ -240,6 +240,7 @@ export type PackageJSON = {
   devDependencies?: PackageDependencies,
   peerDependencies?: PackageDependencies,
   sideEffects?: boolean | FilePath | Array<FilePath>,
+  bin: string | {|[string]: FilePath|},
```
mischnic, line 243:
```suggestion
  bin?: string | {|[string]: FilePath|},
```

packages/transformers/elm/src/ElmTransformer.js
```diff
@@ -0,0 +1,150 @@
...
+
+async function resolveLocalElmBinary(searchPath, options) {
+  try {
+    let result = await options.packageManager.resolve(
+      'elm/package.json',
+      searchPath,
+      {autoinstall: false},
+    );
+
+    let bin = nullthrows(result.pkg).bin;
```
mischnic, line 89:
```suggestion
    let bin = nullthrows(result.pkg?.bin);
```

mischnic:
The tests are failing on windows: 
Bildschirmfoto 2020-11-24 um 10 45 04

https://dev.azure.com/devongovett/devongovett/_build/results?buildId=13026&view=logs&j=5c638625-2435-5635-d5ca-9da4143f3051&t=6ab6a9ca-ea03-5f8b-897c-aaa501253c99

nilshelmig:
@mischnic `node-elm-compiler` had an old version of `cross-spawn` which didn't support worker threads. I've added the latest `cross-spawn` version as dependency and injected it in `node-elm-compiler`

nilshelmig:
@mischnic Yeah I'm out of ideas ☹️ 
https://dev.azure.com/devongovett/devongovett/_build/results?buildId=13028&view=logs&j=5c638625-2435-5635-d5ca-9da4143f3051&t=6ab6a9ca-ea03-5f8b-897c-aaa501253c99

mischnic:
This is caused by graceful-fs which overwrites `process.chdir` and doesn't copy over `process.chdir.disabled` (which is used by graceful-fs as you've already found out)...

https://github.com/isaacs/node-graceful-fs/blob/5a29f6c50ccdb412cb198b06ee248e65f365145b/polyfills.js#L17-L21

That property also undefined on other platforms, but the problematic code path is only used on Windows:
https://github.com/moxystudio/node-cross-spawn/blob/5d843849e1ed434b7030e0aa49281c2bf4ad2e71/lib/parse.js#L28-L33

You can try this out by putting this into ElmTransformer somewhere:
```
    console.log(process.chdir.toString(), process.chdir.disabled)
```
prints
```
function(d) {
  cwd = null
  chdir.call(process, d)
} undefined
```
The printed function is graceful-fs's wrapper.

I've opened a PR: https://github.com/isaacs/node-graceful-fs/pull/200
However, I don't see it getting merged any time soon. We'll probably have to find a workaround.

mischnic:
Let's do this workaround in the meantime so we can merge this:

```diff
diff --git a/packages/transformers/elm/src/ElmTransformer.js b/packages/transformers/elm/src/ElmTransformer.js
index 63f2b796..f088c42c 100644
--- a/packages/transformers/elm/src/ElmTransformer.js
+++ b/packages/transformers/elm/src/ElmTransformer.js
@@ -8,6 +8,14 @@ import {minify} from 'terser';
 import nullthrows from 'nullthrows';
 import ThrowableDiagnostic from '@parcel/diagnostic';

+let isWorker;
+try {
+  let worker_threads = require('worker_threads');
+  isWorker = worker_threads.threadId > 0;
+} catch (_) {
+  isWorker = false;
+}
+
 export default (new Transformer({
   async loadConfig({config, options}) {
     const elmConfig = await config.getConfig(['elm.json']);
@@ -49,6 +57,8 @@ export default (new Transformer({
       asset.addIncludedFile(filePath);
     }

+    process.chdir.disabled = isWorker;
+
     let code = await compileToString(elm, elmBinary, asset, compilerConfig);
     if (options.hot) {
       code = await injectHotModuleReloadRuntime(code, asset.filePath, options);

```

mischnic:
You'll need to run prettier on ElmTransformer.js

mischnic (approved):

nilshelmig:
@mischnic Seems that the pipeline ran into timeout
https://dev.azure.com/devongovett/devongovett/_build/results?buildId=13140&view=logs&j=6fe4638a-2d22-52c1-aa18-c5cc82156fc5

Should it be restarted?

PR #4397, opened 2020-03-28
Feature/v2 jsonld transformer

### LINKED ISSUE #3355: Parcel 2: JSONLD transformer
Create the `@parcel/transformer-jsonld` package in `packages/transforms/jsonld`. This package is a Parcel 2 transform plugin for JSONLD. It should be based on the `JSONLDAsset` from Parcel 1.

The `@parcel/transformer-jsonld` package is responsible for the following things:

* Parsing and collecting dependencies
* Rewriting urls to dependencies
* Compiling back to JSON

isaacadams:
# ↪️ Pull Request

This PR aims to fulfill requirements made in #3355 -- that is to port the jsonld asset v1 code into a v2 jsonld transformer

Closes #3355

## 💻 Examples

take a look at the [index.html][example] that is used for the jsonld integration test

## 🚨 Test instructions

Run the following: `yarn workspace @parcel/integration-tests test -g jsonld`
The integration test file for jsonld is defined [here][test]

## ✔️ PR Todo

## 🐛 Issues 

- [x] the jsonld is not outputting the json data back into the built [index.html][example] file in the proper format
    - the data injected into the index.html isn't a pure json object but is wrapped in JS code
    - now a `<div></div>` is being injected into the script tag where the jsonld object used to be
- [x] unable to get the `publicURL` config option from the available objects passed into the transform object
    - `--public-url https://place.holder/` is being passed to the `parcel build` command, however, logs do not show it appearing anywhere
    - update fom @DeMoorJasper: url replacement occurs further down the line after packaging and therefore no longer needs to happen inside the transformer
- [x] the `asset.addURLDependency()` is returning a number, which causes `urlJoin()` to throw an error
    - this logic is ported directly from v1 code, so it seems that the addURLDependency has changed what it returns for v2 changes??
    - update from @DeMoorJasper: the number being returned will eventually be replaced by the new asset path during the naming stage, therefore, replacing the old asset path with the number is sufficient
- [x] unit tests are failing because the jsonld transformer is not part of the [package.json](https://github.com/parcel-bundler/parcel/blob/v2/packages/configs/default/package.json) list
    - It seems I am meant to add it to the list in order to get that unit test to pass. 
    - However, I need to publish the transformer package first
    - Unfortunately and not surprisingly, I do not have permissions to publish the new @parcel/transformer-jsonld package
- [x] integration test is failing because it is using an old method to test the bundle
- [x] the [integration test][test] uses the function `assertBundles()` which is not seeing the expected .png and .jpeg dependencies, however, when using the local parcel cli to build the [index.html][example], those image assets get pulled through to the dist folder
     - 🤦‍♂ accidentally removed the code that parses the jsonld object -- put it back in and the images are getting pulled through again and the integration test is seeing them
- [x] unable to figure out how to assert the transformer output for an asset in the integration test

[example]: https://github.com/isaacadams/parcel/blob/feature/v2-jsonld-transformer/packages/core/integration-tests/test/integration/schema-jsonld/index.html
[test]: https://github.com/isaacadams/parcel/blob/feature/v2-jsonld-transformer/packages/core/integration-tests/test/schema-jsonld.js


packages/transformers/jsonld/src/JSONLDTransformer.js
```diff
@@ -0,0 +1,110 @@
...
+    return jsonArray;
+  }
+
+  transformString(value) {
+    let assetPath = this.asset.addURLDependency(value, {});
+    console.log(assetPath);
+
+    /* if (!isURL(assetPath)) {
+      // paths aren't allowed, values must be urls
+      assetPath = urlJoin(this.publicURL, assetPath);
```
isaacadams: - the `asset.addURLDependency()` is returning a number, which causes `urlJoin()` to throw an error, which is why it is currently commented out
- this logic is ported directly from v1 code
- am I using the `asset.addURLDependency()` function properly? Am I supposed to be passing options to it? Is there docs or good examples on how to use it?

packages/transformers/jsonld/src/JSONLDTransformer.js
```diff
@@ -0,0 +1,110 @@
...
+    parser.parse(jsonCode);
+    console.log("done");
+    console.log(jsonCode);
+    
+    // this will send the jsonld to the JSONTransformer
+    asset.type = 'json';
+    // setting it to jsonCode since the parser updates asset paths
+    asset.setCode(JSON.stringify(jsonCode));
+    return [asset];
+  },
```
isaacadams: - parcel v1 seems to inject the json object write back into the script tag on the html page
- The only difference is that any asset paths that was in the jsonld are now updated with the new asset paths
- However, my local parcel injects the json object wrapped in javascript code with `module.exports=` and syntax like that

packages/transformers/jsonld/src/JSONLDTransformer.js
```diff
@@ -0,0 +1,110 @@
...
+  'embedUrl',
+  'thumbnailUrl',
+  'video',
+  'contentUrl',
+  'url'
+];
+
+export default new Transformer({
+  async transform({ asset, config, logger, options, resolve }) {
+    // allowing any recieved jsonld to be in json5 format
```
isaacadams: from which of the available object (config or options?) would I be able to get the `publicURL`? Whenever I try to reference it like `option.publicURL` I get a compiler error from flow

DeMoorJasper: as far as I know publicUrl should no longer be directly used in a transformer as we do url replacement all the way at the end after packaging. We just insert references in transformers

packages/transformers/jsonld/src/JSONLDTransformer.js
DeMoorJasper: Yeah that makes sense... the json transformer will convert it into js.

We have to figure out a way to ignore this in some way, I think we were gonna query parameters for this like in https://github.com/parcel-bundler/parcel/pull/4209

But that PR also implements the query param support so maybe it should get merged asap so this can output json?

packages/transformers/jsonld/src/JSONLDTransformer.js
DeMoorJasper: I don't think you need to add anything here, it should replace the number in the naming stage.

packages/transformers/jsonld/src/JSONLDTransformer.js
mischnic: We could also put the JSON transformer on a named pipeline instead and insert that pipeline to JSON imports in JS (just as we do with url dependencies and `url:`)

DeMoorJasper: That's also a possibility and probably a better option in this case (as it will unblock this PR)

isaacadams: @DeMoorJasper & @mischnic let me know if there is anything I can do to help get this PR unblocked or if there is another PR I should be keeping my eyes on.

DeMoorJasper: You can use named pipelines for now. So nothing should be blocking this PR :)

mischnic:
>unit tests are failing because the jsonld transformer is not part of the package.json  list
> It seems I am meant to add it to the list in order to get that unit test to pass.

Yes, simply do that. You don't need to concern yourself with publishing.

We also have an existing integration test for jsonld from v1 which you should reenable/update: https://github.com/parcel-bundler/parcel/blob/v2/packages/core/integration-tests/test/schema-jsonld.js

packages/transformers/jsonld/src/JSONLDTransformer.js
isaacadams: Is there something I need to add to the code here in order to use the named pipelines? Or are you saying the code is fine how it is because the named pipelines will pick it up.

DeMoorJasper: change the `"*.{json,json5}": ["@parcel/transformer-json"],` into something like `"js:*.{json,json5}": ["@parcel/transformer-json"],`

Than `js:./something.json` will parse json into js and `url:./something.json` will just copy the json to the output directory and in your case not transform jsonld beyond your transformer

packages/core/integration-tests/test/schema-jsonld.js
```diff
@@ -1,26 +1,62 @@
...
+          "width": 180,
+          "height": 120
+        },
+      "image": ["images/image.jpeg", "images/image.jpeg"]
+    };
+
+    // I need to assert that the current (v2) output is the same as the v1 output
+    // aka, is there a test util that can be used to grab onto the transformed output?
+    
+    assert.ok(false);
```
isaacadams: need some help here figuring out how / where I can get the output of the jsonld transformer to assert its form

packages/transformers/jsonld/src/JSONLDTransformer.js
```diff
@@ -0,0 +1,94 @@
...
+    let rawCode = await asset.getCode();
+    // allowing any recieved jsonld to be in json5 format
+    let jsonCode = json5.parse(rawCode);
+    
+    let parser = new JSONLDParser(asset);
+    parser.parse(jsonCode);
+    
+    // should these be used to tell the transformer that the output is meant to be placed back into the index.html file?
+    //asset.isInline = true;
+    //asset.meta.inlineType = 'string';
```
isaacadams: so locally I made the change that @DeMoorJasper suggested [here](https://github.com/parcel-bundler/parcel/pull/4397#discussion_r399685120) . After that, the built index.html file has `<div></div>` inside the script tag where the jsonld object should be. I am unsure why its putting a div in there and not the transformed jsonld object (which really should be the original jsonld object with updated asset paths).

After doing some research in the code, I was thinking that these options might be close to what I need? I tried them out and the output remains the same. Not sure how to proceed to get it to write the jsonld object into the index.html file.

packages/core/integration-tests/test/schema-jsonld.js
isaacadams: just found `outputFS, distDir` being used from the test utils, so I will be using that

isaacadams: @DeMoorJasper or @mischnic is there any reason why 'chai' is not included for testing? Would I be able to add it in? No worries if not.

mischnic: No particular reason, but the builtin assert package provides everything that we need.

isaacadams: 🤦‍♂ you are right, i found the deep equals function I was looking for in mocha's assert -- don't know why I have always relied on chai's deep equals for that.

packages/transformers/jsonld/src/JSONLDTransformer.js
isaacadams: take a look at the [result][1] of the unit test I wrote, it shows that output being the div instead of the expected output of the json object

[1]: https://dev.azure.com/devongovett/devongovett/_build/results?buildId=8730&view=logs&j=3dbdc9c4-c12f-5bc5-3f55-0178d7f6c002&t=59bda16f-6692-523f-b778-442dd3cca32c&l=761

isaacadams: I think I have it figured out in my latest commit --- I changed `asset.type="jsonld"` to `asset.type="html"`

mischnic: Setting it to `html` is definitely not what we want. It should be `json`, but we'll have to figure out how to to always convert JSON assets into JS modules for that.

packages/transformers/jsonld/src/JSONLDTransformer.js
```diff
@@ -0,0 +1,89 @@
...
+    
+    let rawCode = await asset.getCode();
+    // allowing any recieved jsonld to be in json5 format
+    let jsonCode = json5.parse(rawCode);
+    
+    let parser = new JSONLDParser(asset);
+    parser.parse(jsonCode);
+
+    // json should be injected back into the html page
+    asset.type = 'html';
```
isaacadams: as @mischnic mentioned [here](https://github.com/parcel-bundler/parcel/pull/4397#discussion_r399711888) the `asset.type` should be set to json. However, when it is set to that, the json does not get outputted to the index.html file in the proper format, which is what it was doing in v1. How should I go about getting this fixed. Should this be handled in a different PR / Issue?

isaacadams:
the pipeline build is failing at the lint stage and the error it gives is `error Command failed with exit code 1.` (see it [here](https://dev.azure.com/devongovett/devongovett/_build/results?buildId=8736&view=logs&j=81a0a40d-e1c6-55f1-066f-a450d17357d3&t=08540eb4-2586-54e4-c72e-edd69d7b0b03&l=18))

I am unsure how to get that resolved.

mischnic:
Prettier says that the files packages/core/integration-tests/test/schema-jsonld.js and packages/transformers/jsonld/src/JSONLDTransformer.js aren't formatted correctly. You could invoke something like `yarn prettier --write packages/core/integration-tests/test/schema-jsonld.js packages/transformers/jsonld/src/JSONLDTransformer.js`.

packages/transformers/jsonld/src/JSONLDTransformer.js
isaacadams: @mischnic & @DeMoorJasper what should I do here? Should I change it to 'json' for this PR or keep it as is?

isaacadams: another thing that isn't clear to me -- is parcel 2 supposed to output the jsonld into its own json file for the bundle dist or should it output back into the bundled index.html file like parcel v1 does?

DeMoorJasper: use `asset.type = 'json';`?

mischnic: > use asset.type = 'json';?

Then the json transformer will output a js asset... 
And implementing the change to run that only for js assets is non-trivial.

> is parcel 2 supposed to output the jsonld into its own json file for the bundle dist or should it output back into the bundled index.html file like parcel v1 does?

If the jsonld was inlined in a html file, it will be inserted back again.
`<script src="..." type="application/ld+json"></script>` would create a separate bundle.

isaacadams: @DeMoorJasper & @mischnic but when I set it to `asset.type = "json"` it does NOT insert the json back into the script tag. Either it inserts a `<div></div>` inside the script tag or it inserts it into the script tag as if it were a js asset with `module.exports=...` syntax, which is again not what parcel v1 does.

isaacadams: > If the jsonld was inlined in a html file, it will be inserted back again.
> `<script src="..." type="application/ld+json"></script>` would create a separate bundle.
thats good to know and makes perfect sense

so again, in the unit test case, the jsonld IS inlined html and parcel v2 is NOT inserting the transformed jsonld object back into the script tag. See my previous comment.

isaacadams:
1. unit tests are [failing][1] and I am unsure why as they look unrelated to the changes I made
2. is there any direction for me based on the conversation [here][2]?

[1]: https://dev.azure.com/devongovett/devongovett/_build/results?buildId=8820&view=logs&j=6fe4638a-2d22-52c1-aa18-c5cc82156fc5&t=f25df677-184c-5b96-065c-7345903c14c5&l=750
[2]: https://github.com/parcel-bundler/parcel/pull/4397#discussion_r399820896

mischnic:
> unit tests are failing and I am unsure why as they look unrelated to the changes I made

They are indeed unrelated.

> is there any direction for me based on the conversation here?

Either:
- (if you have time/willing to dig deep into the Parcel core) try to implement my suggestion, though I'm not sure either if it would work out/debug why `type = jsonld` emits html
- I'll look into this myself in the next few days

mischnic:
- Please remove the `JSONLDParser` class by turning the methods into functions and calling them in the transformer.

- Removing the current JSON-to-JS transformer is a more complicated endeavour, so we will just leave the type at `jsonld` for now and fix a bug in the HTMLPackager to actually insert the content (and not `<div></div>`:
```diff
diff --git packages/packagers/html/src/HTMLPackager.js packages/packagers/html/src/HTMLPackager.js
index f980274b5..a4f955d14 100644
--- packages/packagers/html/src/HTMLPackager.js
+++ packages/packagers/html/src/HTMLPackager.js
@@ -1,10 +1,16 @@
 // @flow strict-local
 import type {Bundle, BundleGraph} from '@parcel/types';
 
 import assert from 'assert';
+import {Readable} from 'stream';
 import {Packager} from '@parcel/plugin';
 import posthtml from 'posthtml';
-import {replaceURLReferences, urlJoin} from '@parcel/utils';
+import {
+  bufferStream,
+  replaceInlineReferences,
+  replaceURLReferences,
+  urlJoin,
+} from '@parcel/utils';
 import nullthrows from 'nullthrows';
 
 // https://www.w3.org/TR/html5/dom.html#metadata-content-2
@@ -122,7 +128,10 @@ async function replaceInlineAssetContent(
 
     if (newContent != null) {
       let {contents, bundle} = newContent;
-      node.content = contents;
+      node.content = (contents instanceof Readable
+        ? await bufferStream(contents)
+        : contents
+      ).toString();
 
       if (bundle.env.outputFormat === 'esmodule') {
         node.attrs.type = 'module';
diff --git packages/transformers/jsonld/src/JSONLDTransformer.js packages/transformers/jsonld/src/JSONLDTransformer.js
index 63c8b3499..1208299bc 100644
--- packages/transformers/jsonld/src/JSONLDTransformer.js
+++ packages/transformers/jsonld/src/JSONLDTransformer.js
@@ -29,8 +29,6 @@ export default new Transformer({
     let parser = new JSONLDParser(asset);
     jsonCode = parser.parse(jsonCode);
 
-    // json should be injected back into the html page
-    asset.type = 'html';
     // setting it to jsonCode since the parser updates asset paths
     asset.setCode(JSON.stringify(jsonCode));
     return [asset];
```

isaacadams:
>     * Please remove the `JSONLDParser` class by turning the methods into functions and calling them in the transformer.

I can do that, but that will mean I will then need to pass around `asset` through each function, which feels a little sloppy to me. Whenever I find a url dependency to add, I can easily grab `this.asset` and call the `addUrlDependency` method right then and there without having to pass around `asset` through each function. This is what I wrapped it in a class -- makes it much easier to reference `this.asset` whenever or wherever I need to call `addUrlDepdendency`.

Is this still something you want me to do?

mischnic:
> Is this still something you want me to do?

You can leave it like it is if you want

isaacadams:
@mischnic I created a unit test that checks if the injected jsonld into the html output is what is expected and with the HTMLPackager fixes you suggested, it is now injecting the json into the html now, however, the image asset urls are not being transformed as expected

```html
<script type="application/ld+json">
      {
        "@context": "http://schema.org",
        "@type": "LocalBusiness",
        "description": "This is your business description.",
        "name": "Parcel's parcel",
        "telephone": "555-111-2345",
        "openingHours": "Mo,Tu,We,Th,Fr 09:00-17:00",
        "logo": {
          "@type": "ImageObject",
          "url": "d7bbaabe2edc086ba84f5ed6a9c205cd",
          "width": 180,
          "height": 120
        },
        "image": [
          "883ce35a1e55486f1a620e1699429309",
          "883ce35a1e55486f1a620e1699429309"
        ]
      }
    </script>
```

mischnic:
The HTMLPackager only replaces its own hashes with urls, but not inside inline bundles.
You need to create a new packager (`@parcel/packager-raw-url` ?) which is exactly that same as the existing `@parcel/packager-raw` except for:
```diff
 import assert from 'assert';
 import {Packager} from '@parcel/plugin';
+import {replaceURLReferences} from '@parcel/utils';

 export default new Packager({
-  package({bundle}) {
+  async package({bundle, bundleGraph}) {
     let assets = [];
     bundle.traverseAssets(asset => {
       assets.push(asset);
     });

     assert.equal(assets.length, 1, 'Raw bundles must only contain one asset');
-    return {contents: assets[0].getStream()};
+    return {
+      contents: replaceURLReferences({
+        bundle,
+        bundleGraph,
+        contents: await assets[0].getCode(),
+        relative: false,
+      }).contents,
+    };
   },
 });
```

- All our tests are essentially integration tests, so you can simply merge your two tests into one.
- Please don't hardcode bundle hashes into the test (`'/image.ba250946.jpeg'`), that hash might not be stable, for example use a regex `image\.[a-f0-9]+\.jpeg`.

isaacadams:
@mischnic how should I properly include the new "packager-url-raw" in index.json config? I assumed the following would work.

```diff
diff --git a/packages/configs/default/index.json b/packages/configs/default/index.json
index b3f19515..1226a9dc 100644
--- a/packages/configs/default/index.json
+++ b/packages/configs/default/index.json
@@ -51,7 +51,7 @@
     "*.css": "@parcel/packager-css",
     "*.js": "@parcel/packager-js",
     "*.ts": "@parcel/packager-ts",
-    "*": "@parcel/packager-raw"
+    "*": ["@parcel/packager-raw", "@parcel/packager-raw-url"]
   },
   "resolvers": ["@parcel/resolver-default"],
   "reporters": [
```

However, I am getting an error: 

```shell
@parcel/core: Invalid Parcel Config
..\node_modules\@parcel\config-default\index.json:108:8
  107 |     "*.ts": "@parcel/packager-ts",
> 108 |     "*": [
>     |          ^
> 109 |       "@parcel/packager-raw",
>     | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> 110 |       "@parcel/packager-raw-url"
>     | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> 111 |     ]
>     | ^^^^^ Expected type string
  112 |   },
  113 |   "resolvers": [
```

mischnic:
```diff
diff --git a/packages/configs/default/index.json b/packages/configs/default/index.json
index b3f19515..1226a9dc 100644
--- a/packages/configs/default/index.json
+++ b/packages/configs/default/index.json
@@ -51,7 +51,7 @@
     "*.css": "@parcel/packager-css",
     "*.js": "@parcel/packager-js",
     "*.ts": "@parcel/packager-ts",
+    "*.jsonld": "@parcel/packager-raw-url",
     "*": "@parcel/packager-raw"
   },
   "resolvers": ["@parcel/resolver-default"],
   "reporters": [
```
(You can only have one packager per type, because running two packagers on one bundle doesn't actually make sense.)

isaacadams:
new error that is complaining about the use of `getBundleGroupsReferencedByBundle()`

```console
Building index.html...
Building other.css...
Building logo.png...
Building image.jpeg...
Bundling...
Packaging logo.[hash].png...
Packaging image.[hash].jpeg...
Packaging index.html...
Packaging dev-tests.[hash].jsonld...
Packaging other.[hash].css...
× Build failed.
@parcel/packager-html: bundleGraph.getBundleGroupsReferencedByBundle is not a function or its return value is not iterable
TypeError: bundleGraph.getBundleGroupsReferencedByBundle is not a function or its return value is not iterable
    at replaceURLReferences (C:\source\isaacadams\parcel\packages\packagers\raw-url\node_modules\@parcel\utils\lib\replaceBundleReferences.js:72:20)
    at Object.package (C:\source\isaacadams\parcel\packages\packagers\raw-url\lib\RawUrlPackager.js:29:49)
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
    at async PackagerRunner.package (C:\source\isaacadams\parcel\packages\core\core\lib\PackagerRunner.js:201:14)
    at async PackagerRunner.getBundleResult (C:\source\isaacadams\parcel\packages\core\core\lib\PackagerRunner.js:170:20)
    at async getInlineBundleContents (C:\source\isaacadams\parcel\packages\core\core\lib\PackagerRunner.js:216:21)
    at async getAssetContent (C:\source\isaacadams\parcel\packages\packagers\html\lib\HTMLPackager.js:90:26)
    at async Object.replaceInlineAssetContent (C:\source\isaacadams\parcel\packages\packagers\html\lib\HTMLPackager.js:111:22)
```

mischnic:
> getBundleGroupsReferencedByBundle

I can't find that function being used/defined/called anywhere in the repo.

Did you run `yarn build` before testing? (You can also invoke `<repo>/packages/core/parcel/src/bin.js` directly without building)

mischnic:
You specified an old package version in package.json:
```diff
--- packages/packagers/raw-url/package.json
+++ packages/packagers/raw-url/package.json
@@ -17,6 +17,6 @@
   },
   "dependencies": {
     "@parcel/plugin": "^2.0.0-alpha.3.1",
-    "@parcel/utils": "2.0.0-alpha.3"
+    "@parcel/utils": "^2.0.0-alpha.3.1"
   }
 }
```

isaacadams:
@mischnic see azure pipeline [error] (https://dev.azure.com/devongovett/devongovett/_build/results?buildId=9444&view=logs&j=c9e7f81a-e301-5408-62aa-d2082bd3a1e5&t=ef7d0775-db49-594e-ab10-3d888f0b1c3a&l=783)

is `assert.match()` not valid?

isaacadams:
@mischnic @DeMoorJasper given all the benchmarks are good, this PR will be ready for official review.

packages/core/integration-tests/test/schema-jsonld.js
```diff
@@ -1,26 +1,88 @@
-import {bundle, assertBundleTree} from '@parcel/test-utils';
+import {bundle, assertBundles, distDir, outputFS} from '@parcel/test-utils';
+import assert from 'assert';
+import path from 'path';
 
-describe.skip('schema ld+json', function() {
+function getPathToFile(relativePathToFile) {
```
mischnic, line 5: These functions are used only once, so you can inline them.

packages/transformers/jsonld/README.md
```diff
@@ -0,0 +1,35 @@
+# jsonld transformer
```
mischnic, line 1: I think this file can be removed now?

packages/core/integration-tests/test/schema-jsonld.js
```diff
@@ -1,26 +1,88 @@
...
+      logo: {
+        '@type': 'ImageObject',
+        url: 'images/logo.png',
+        width: 180,
+        height: 120,
+      },
+      image: ['image\\.[a-f0-9]+/\\.jpeg', 'images/image.jpeg'],
+    }; */
+
+    let file = await getBundleFile();
```
mischnic, line 65: As I said already, there is no value in having two tests for this:
- the second test shouldn't read the output of the first test (as it does now). I'm actually surprised that this works as at all because the output filesystem is cleared before every test..
- if the first test fails, the second one will fail as well (if the dependencies aren't collected).

packages/transformers/jsonld/README.md
isaacadams: We could remove it, but is there harm in keeping it around? Could serve as a nice starter manual for developers working on new transformers? I was really confused when I started working on this and that README is where I put everything I learned. If anything, we could clean it up and move it somewhere else where developer documentation goes.

Up to you.

packages/core/integration-tests/test/schema-jsonld.js
isaacadams: that test's intention was to make sure that the content being inserted inbetween the script tags was actually a json object rather than the 'div' tag that was previously appearing there. Since then, it has changed and become redundant, as you pointed out. I will remove the test. 

However, if you think the original intention of the test is a worthy pursuit, (make sure a json object is inserted in between the script tag) I can write that. The true test for that would be to extract the content inbetween the script tag, like its currently doing, but then just test that the content is a valid json object instead of looking at the content of its properties / assets.

packages/core/integration-tests/test/schema-jsonld.js
isaacadams: addressed in 7f508f52750e83eba91e1f8647dbf6b94031789f

packages/core/integration-tests/test/schema-jsonld.js
mischnic: Sorry, I should have written that in the beginning: I meant that you should merge these two tests, so:
```js
it('...', async () => {
	let b = bundle();
	
	assertBundles(b, {
	...
	});

    let regex = new RegExp(
      /<\s*script \s*type="application\/ld\+json"\s*>.*<\/\s*script\s*>/gm,
    );
    let firstMatch = regex.exec(file).map(m => {
      // removes script html tags to extract the json object in between them
      return m.replace(/<\/?\s*script( \w+=".*"\s*)*\s*>/g, '');
    })[0];

    let actual = JSON.parse(firstMatch);
    match(actual.logo.url, /logo\.[a-f0-9]+\.png/);
    match(actual.image[0], /image\.[a-f0-9]+\.jpeg/);
    match(actual.image[1], /image\.[a-f0-9]+\.jpeg/);
});
```
(Not really important, but you can simplify that regex logic to `let firstMatch = file.match(/...../gm)[1]` with the regex `<\s*script \s*type="application\/ld\+json"\s*>(.*)<\/\s*script\s*>` (exactly what you wrote but using a capture group for the script tag contents))

packages/transformers/jsonld/README.md
mischnic: Proper (plugin) docs are in the works. And the part that applies to transformers in general is quite short

packages/core/integration-tests/test/schema-jsonld.js
isaacadams: see 3e6f25904390c5a1ebda60e69c135c5f6be47495

Issue #1746, opened 2018-07-17
Self-signed Certificate security warning


germanftorres:
# 🐛 bug report
Environment: Parcel 1.9.7 in Windows 10.

Using --https option generates a self-signed certificate issued to IP 127.0.0.1 and not to localhost. In my system, navigating to https://locahost:1234 causes a security warning even after installing the certificate under Trusted Root Certification Authorities. In contrast, If I navigate to https://127.0.0.1:1234 the certificate is shown as valid.

```
>parcel src\index.html --https
Generating SSL Certificate...
√  Built in 2.94s.
```
## 🤔 Expected Behavior

I would expect to navigate directly to https://locahost:1234 and see no security warning with self signed certifcate.

## 😯 Current Behavior
Security warning!!

## 🌍 Your Environment
Parcel 1.9.7
Windows 10
npm 5.6.0

DeMoorJasper:
This line should probably be updated to `hmr-hostname` and fallback to localhost
https://github.com/parcel-bundler/parcel/blob/75188057e01515f1ee399fa6cda30a5304d1cf17/src/utils/generateCertificate.js#L100

shobhitchittora:
@DeMoorJasper I'd like to pick this up as my first contribution.

DeMoorJasper:
@shobhitchittora feel free to work on it :)

dehuszar:
I'm seeing issues with this as well even when trying to access `https://127.0.0.1:1234`.  The odd thing is if I clear out all certs it works using localhost on the first cert generated, but any subsequent certificates (either after deleting the .cache folder or when creating a new parcel project on the same machine) it starts throwing errors around reused serials.

I'm wondering if [this bit](https://github.com/parcel-bundler/parcel/blob/75188057e01515f1ee399fa6cda30a5304d1cf17/src/utils/generateCertificate.js#L30) needs to be set to a random number or uuid or something.

dehuszar:
Yup.  Just ticked the hard-coded '01' value up to '02' and the issue went away.  I took a stab at requiring `uuid` as a dependency, but that caused an unhandled promise rejection error.  I've used this module in a ton of other projects, so I'm not sure what's causing things to explode here, but it seems that having someway to ensure a unique value for every cert generated by parcel will resolve the issue.

DeMoorJasper:
@dehuszar what about Date.now() ?

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

[...]

PR #4402, opened 2020-03-29
Add cli -ip option to generateCertificate() "altNames" alternative to 127.0.0.1

ivnnv:
Related Issue: ([1746](https://github.com/parcel-bundler/parcel/issues/1746))

**Problem**
Self-signed Certificate always use 127.0.0.1 as its IP and there is no way to change it.

## 💻 Examples
Self-signed certificates work well for local machine development, it also work well for testing websites on iOS simulator, as its smart enough to understand 127.0.0.1 is its host IP, and it will trust the exported & installed certificate.

But devices on the same network aren't able to trust the certifcate even after manually installing it and trying to access the hostname "hostmachine.local" from bonjour. They resolve the host and load the page, but wont trust the certificate.

This PR is intended to add a -ip argument to the cli so it will override the hardcoded 127.0.0.1 when generating a self-signed certificate.

## ✔️ PR Todo

### DIFF
--- a/packages/core/parcel/README.md
+++ b/packages/core/parcel/README.md
@@ -127,6 +127,7 @@ console.log("Hello World");
     - [`--target [name]`](#--target-name)
     - [`--open, -o [browser]`](#--open--o-browser)
     - [`--host <host>`](#--host-host)
+    - [`--ip <ip>`](#--ip-ip)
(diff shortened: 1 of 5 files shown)


devongovett:
Should we use the existing --host option for this instead of adding a new one?

DeMoorJasper:
closing due to inactivity

PR #4404, opened 2020-03-30
Fix typescript rechecking when dependencies change

### LINKED ISSUE #4347: @parcel/validator-typescript can report incorrect errors when dependencies change
# 🐛 bug report

In testing PR #4237 (which tries to fix issues that can occur with `@parcel/validator-typescript` when multiple files change at the same time), I discovered a pre-existing problem where `@parcel/validator-typescript` will not report correct errors in situations where a dependency (such as an interface definition) changes, but the files that depends on the changed file (e.g. one that consumes the interface) does not. 

Thanks to @padmaia for asking [the right question](https://github.com/parcel-bundler/parcel/pull/4237#issuecomment-595572660) to help uncover this.

***.parcelrc:***
```js
{
    "validators": {
    }
}
```
***package.json:***
```
{
  "version": "1.0.0",
  "main": "index.html",
  "private": true,
  "scripts": {
  },
  "devDependencies": {
    "@parcel/validator-typescript": "2.0.0-nightly.151", // In order to see this problem, you need to use a build created after #4187 is merged.
    "parcel": "2.0.0-nightly.149",
    "typescript": "3.7.5"
  }
}
```

***src/index.ts:***
```ts
import { returnMessage } from "./returnMessage";
const message: string = "My Message!";
export const output = returnMessage(message);
```

***src/returnMessage.ts***
```ts
export function returnMessage(message: string): string {
  return message;
}
```
## 🤔 Expected Behavior

When you start the above project with parcel (by running `parcel src/index.ts`), it should build with no warnings (because the initial types of both `index.ts` and `App.ts` are valid).

Now, change _only_ `returnMessage.ts` to:
```ts
export function returnMessage(message: number):number { // we've changed the type to 'number', which is not what index.ts is expecting
    console.log(message);
}
```
Parcel should re-validate the project and show this error:

```
🚨 Build failed.
@parcel/validator-typescript: Argument of type 'number' is not assignable to parameter of type 'string'.
/Users/Andrew/Projects/parcel2-typescript-validator-bug/src/index.ts
/Users/Andrew/Projects/parcel2-typescript-validator-bug/src/index.ts:3:13
  2 | const message: string = "My Message!";
> 3 | showMessage(message);
>   |             ^^^^^^^^ Argument of type 'string' is not assignable to parameter of type 'number'.
  4 | 
```

## 😯 Current Behavior

Parcel does not report any errors.

## 💁 Possible Solution

Currently, `@parcel/validator-typescript` will only call `getSemanticDiagnostics` for the files that have changed. Since `index.ts` did not change, `getSemanticDiagnostics` will not be called, and the error will go unreported.

I took a look at the way that `fork-ts-checker-webpack-plugin` addressed this issue, and it appears that they take the simple-but-slow approach of [re-analyzing _every_ file in the typescript program whenever _any_ file changes](https://github.com/TypeStrong/fork-ts-checker-webpack-plugin/blob/54a51c707d064f8a52c4f599ac592355094c2f6e/src/IncrementalChecker.ts#L230-L247).

This would work as a backup, but we might also be able to improve performance if we were able to use parcel's dependency graph to only re-analyze the files that depend on the files that changed.

It might also be possible to use Typescript itself to perform the same analysis - see [this (still open) issue where someone was asking for this feature](https://github.com/Microsoft/TypeScript/issues/5192).

## 🔦 Context

## 💻 Code Sample

See the `dependencies-change-issue` branch of [this repo](https://github.com/astegmaier/parcel2-typescript-validator-bug/tree/dependencies-change-issue) for a demonstration.

## 🌍 Your Environment

| Parcel           | 2.0.0-nightly.149
| Node             | 10.6.3
| Yarn         | 1.21.1

astegmaier:
# ↪️ Pull Request

This is a work-in-progress PR to facilitate discussion about how to fix #4347. The problem it is trying to solve is: when `@parcel/validator-typescript` re-checks a project after a change, if the change breaks (or fixes) a contract with another file that did _not_ change, the validator will not ask for diagnostics on that unchanged file, and the fix (or break) will not be reflected in the result.

@DeMoorJasper or @padmaia - I'd love to get your feedback on this - in particular the questions below about how to implement option number 3. One possibility would be to merge this PR as-is (which makes things correct, but decreases performance), and then work on on a faster solution in a separate PR.

## 💻 Examples

Start with a simple, two-file project with @parcel/validator-typescript enabled:

***.parcelrc:***
```js
{
    "validators": {
    }
}
```

***src/index.ts:***
```ts
import { returnMessage } from "./returnMessage";
const message: string = "My Message!";
export const output = returnMessage(message);
```

***src/returnMessage.ts***
```ts
export function returnMessage(message: string): string {
  return message;
}
```
Then, edit `returnMessage.ts` in a way that breaks the contract being used by `index.ts`:

```ts
export function returnMessage(message: number):number { // we've changed the type to 'number', which is not what index.ts is expecting
    console.log(message);
}
```
`@parcel/validator-typescript` should report an error for `index.ts`, but it currently does not - this PR aims to fix this.

## 💡Possible Solutions

I can think of 3 possible solutions here. This PR implements number 1, but I think number 3 is the most promising one in the long-term.

1. **Correct But Slow** ***(Implemented in this PR)*** - Whenever any file changes, call `getSemanticDiagnostics` for _all_ files in the project.
    - This is definitely slower than what we do now, but it has the lowest risk of being incorrect.
    - It's [what `fork-ts-checker-webpack-plugin` does](https://github.com/TypeStrong/fork-ts-checker-webpack-plugin/blob/master/src/IncrementalChecker.ts#L230-L247) when the [`useTypescriptIncrementalApi `](https://github.com/TypeStrong/fork-ts-checker-webpack-plugin/blame/master/README.md#L165-L167) is `false`.
2. **Use typescript incremental API** - Typescript supports a pattern where a tool can create a `Watch` program along with a `WatchCompilerHost` (see [this PR](https://github.com/Microsoft/TypeScript/pull/20234) and [documentation](https://github.com/microsoft/TypeScript/wiki/Using-the-Compiler-API#writing-an-incremental-program-watcher)). The idea is that the watch program will maintain a cache of dependencies, so when you call `getSemanticDiagnostics` on it, it can provide you with _all_ diagnostics in the program in an efficient manner without re-checking the whole project. It's [what `fork-ts-checker-webpack-plugin` uses](https://github.com/TypeStrong/fork-ts-checker-webpack-plugin/blob/master/src/CompilerHost.ts#L144-L230) when the [`useTypescriptIncrementalApi`](https://github.com/TypeStrong/fork-ts-checker-webpack-plugin/blame/master/README.md#L165-L167) is `true`. I created a proof-of-concept implementation for parcel [in this branch](https://github.com/parcel-bundler/parcel/compare/v2...astegmaier:fix-typescript-rechecking-dependencies-watcher).
    - In my tests, this proof-of-concept is still quite slow (slower than option 1 above) - I'm not 100% sure why. There is a _lot_ of hacks (see [here](https://github.com/microsoft/TypeScript/issues/29176), [here](https://github.com/microsoft/TypeScript/issues/32385) and [here](https://github.com/TypeStrong/fork-ts-checker-webpack-plugin/blob/master/src/CompilerHost.ts#L232-L261)) within the `fork-ts-checker-webpack-plugin` code, and I suspect if we went this route, we'd have to do a lot of onion-peeling.
    - The API does not naturally fit within the parcel validator model because it assumes that the typescript watcher program is directly monitoring the filesystem, rather than being handed a list of changed files directly. You have to hack around this by [implementing `watchFile`/`watchDirectory` methods in the `WatchCompilerHost`](https://github.com/parcel-bundler/parcel/compare/v2...astegmaier:fix-typescript-rechecking-dependencies-watcher#diff-8e07c4f793a820a013f0e57f1934c9e8R184-R221), storing the callbacks that typescript provides, and then invoking them within `validateAll`. Not pretty.
3. **Use parcel's dependency graph**. Parcel appears to already have a mechanism to keep track of dependencies. We could (somehow) re-use this to allow a `Validator` to get (unchanged) files that depend on the changed files, and make sure to call `getSemanticDiagnostics` on those as well.
    - I think this the most promising option to get both correctness and good performance.
    - I could use some tips on how to make this work (if you agree it's a good idea):
        - The basic problem is that even though the `Asset` class has [a `getDependencies` method](https://github.com/parcel-bundler/parcel/blob/790a2469e5bf9e231949256c1655562c61dda79d/packages/core/core/src/public/Asset.js#L149-L151), calling it within the context of a `validateAll` implementation always returns an empty array.
        - this is because of the way [assets are re-built from requests within the `Validation` class](https://github.com/parcel-bundler/parcel/blob/790a2469e5bf9e231949256c1655562c61dda79d/packages/core/core/src/Validation.js#L180-L213) - it does not populate the internal `dependencies` array or have `id` that relates back to the `AssetGraph`.
        - I'm still getting my head around all the different things that happen inside of `AssetGraphBuilder`, and I don't fully understand it. Can you offer any suggestions about how we might enable a `Validator` to look up dependencies of a changed asset within the context of `validateAll`?

## 🚨 Test instructions

I had [previously written a unit test to cover this scenario](https://github.com/parcel-bundler/parcel/blob/790a2469e5bf9e231949256c1655562c61dda79d/packages/core/integration-tests/test/ts-validation.js#L197-L237) as part of PR #4237 . When this PR is complete, we can turn this unit test on.

## ✔️ PR Todo


astegmaier:
@DeMoorJasper or @padmaia - when you have a moment, I was hoping to get some feedback about this.

In particular, I'm currently spinning my wheels a bit investigating option 3. I'd love to hear your thoughts about:
* Is it possible/a good idea to somehow use parcel's various asset or request graphs to determine what file(s) depend on a typescript file that has changed? (e.g. if only `returnMessage.ts` changes, use parcel to figure out that `index.ts` depends on `returnMessage.ts` file, and ask typescript for diagnostics about both files).
* If yes, can you provide any pointers to which graph might contain this information and how best to access it in the context of the [`Validation`](https://github.com/parcel-bundler/parcel/blob/v2/packages/core/core/src/Validation.js) or [`TypeScriptValidator`](https://github.com/parcel-bundler/parcel/blob/v2/packages/validators/typescript/src/TypeScriptValidator.js) classes?

devongovett:
Doesn't typescript already track all that information in its own graph? Could you just ask it for diagnostics for all files whenever anything changes? If you only invalidate the files that changed, TS itself should only recheck what's necessary right?

astegmaier:
> Doesn't typescript already track all that information in its own graph? Could you just ask it for diagnostics for all files whenever anything changes? If you only invalidate the files that changed, TS itself should only recheck what's necessary right?

I wish this were the case, but unfortunately it is not, at least for the language service API (which is what we currently use). See [this issue](https://github.com/Microsoft/TypeScript/issues/5192) and [this closed PR](https://github.com/microsoft/TypeScript/pull/3687) for details. The TS team de-prioritized adding this ability to the language service when they released the "incremental" API. (Switching from the language service to the "incremental" API was what I called "option 2" <- see above for details about some of the challenges I encountered testing this).

astegmaier:
> Doesn't typescript already track all that information in its own graph? Could you just ask it for diagnostics for all files whenever anything changes? If you only invalidate the files that changed, TS itself should only recheck what's necessary right?

I wanted to clarify that "Ask for diagnostics for all files when anything changes" is what this PR currently does (i.e. "option 1" above). It definitely works to solve #4347, the only problem is that it's slower. The typescript language service tracks which files are a part of the program, but it's not particularly smart about avoiding repeat work. I think it will avoid re-parsing, unless you've invalidated the file in the `LanguageServiceHost`, but it doesn't cache much beyond that - it will still re-check files that haven't changed and aren't dependent on files that did change. I did some basic `console.time` logging on [this unit test](https://github.com/parcel-bundler/parcel/blob/68b5050ae0e84c4a65ebdc2e57415c1ed0ceacd2/packages/core/integration-tests/test/ts-validation.js#L147-L195) to measure this, and here's the results:

***Baseline*** (current `master` branch, which calls `getSemanticDiagnostics()` only on changed files)

```
RUN 0: ran getSemanticDiagnostics() for ./packages/core/integration-tests/test/ts-validator-multi-change-errors/index.ts: 655.041ms
RUN 0: ran getSemanticDiagnostics() for ./packages/core/integration-tests/test/ts-validator-multi-change-errors/returnMessage.ts: 3.158ms
RUN 0 FINISHED! Total Time: 658.810ms
RUN 1: ran getSemanticDiagnostics() for ./packages/core/integration-tests/test/ts-validator-multi-change-errors/index.ts: 26.732ms
RUN 1: ran getSemanticDiagnostics() for ./packages/core/integration-tests/test/ts-validator-multi-change-errors/returnMessage.ts: 0.617ms
RUN 1 FINISHED! Total Time: 27.539ms
```

***This PR*** (which uses `languageService.getProgram().getSourceFiles()` to call `getSemanticDiagnostics()` on all files in the program whenever any file changes):

```
RUN 0: ran getSemanticDiagnostics() for ./node_modules/typescript/lib/lib.d.ts: 3.499ms
RUN 0: ran getSemanticDiagnostics() for ./node_modules/typescript/lib/lib.es5.d.ts: 153.522ms
RUN 0: ran getSemanticDiagnostics() for ./node_modules/typescript/lib/lib.dom.d.ts: 904.074ms
RUN 0: ran getSemanticDiagnostics() for ./node_modules/typescript/lib/lib.webworker.importscripts.d.ts: 0.367ms
RUN 0: ran getSemanticDiagnostics() for ./node_modules/typescript/lib/lib.scripthost.d.ts: 7.711ms
RUN 0: ran getSemanticDiagnostics() for ./packages/core/integration-tests/test/ts-validator-multi-change-errors/returnMessage.ts: 3.129ms
RUN 0: ran getSemanticDiagnostics() for ./packages/core/integration-tests/test/ts-validator-multi-change-errors/index.ts: 7.816ms
RUN 0 FINISHED! Total Time: 1739.063ms
RUN 1: ran getSemanticDiagnostics() for ./node_modules/typescript/lib/lib.d.ts: 0.709ms
RUN 1: ran getSemanticDiagnostics() for ./node_modules/typescript/lib/lib.es5.d.ts: 104.020ms
RUN 1: ran getSemanticDiagnostics() for ./node_modules/typescript/lib/lib.dom.d.ts: 755.711ms
RUN 1: ran getSemanticDiagnostics() for ./node_modules/typescript/lib/lib.webworker.importscripts.d.ts: 0.250ms
RUN 1: ran getSemanticDiagnostics() for ./node_modules/typescript/lib/lib.scripthost.d.ts: 2.916ms
RUN 1: ran getSemanticDiagnostics() for ./packages/core/integration-tests/test/ts-validator-multi-change-errors/returnMessage.ts: 0.770ms
RUN 1: ran getSemanticDiagnostics() for ./packages/core/integration-tests/test/ts-validator-multi-change-errors/index.ts: 3.062ms
RUN 1 FINISHED! Total Time: 884.168ms
```

You can see that it's significantly slower, both on the initial run, and the second time (when, presumably, TypeScript is avoiding the need to re-parse unchanged files). You could imagine this scaling up linearly for a large project with lots of files.

So I think it would be goodness to look for ways to optimize the perf here. However, one possibility is that we fix the bug now and optimize performance in a separate PR later - what do you think?

padmaia:
> However, one possibility is that we fix the bug now and optimize performance in a separate PR later - what do you think?

@astegmaier Yeah, I was going to suggest this.

I would love to brainstorm Option 3 more, but I've been busy with other priorities recently. Maybe we could arrange a chat in a week or two to dive into this?

astegmaier:
> Yeah, I was going to suggest this.

Sounds good - in that case, I think this is ready to merge from my perspective. (The test failure seems to be intermittent and not related to this PR - let me know if you disagree).

> I would love to brainstorm Option 3 more, but I've been busy with other priorities recently. Maybe we could arrange a chat in a week or two to dive into this?

A chat sounds great - I'll email you with more of my contact info and availability details.

Issue #3761, opened 2019-11-10
Add an error location when dependent files are not found, or external modules are not allowed


mhintz:
# 🙋 feature request
When an error is thrown during compilation, currently you just get an error message and a stack trace from inside parcel. Ideally there would also be some information about the location in the source where the error was generated.

## 🤔 Expected Behavior
For example, when including an external module when that isn't allowed, or when a needed dependency is missing, it would be good to have an indication of which line in the source generated this problem.

## 😯 Current Behavior
Currently the output is just something like:
```
External modules are not supported when building for browser
... long stack trace from parcel ...
```

DeMoorJasper:
Should be fairly straightforward with the new diagnostics in parcel 2

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

mischnic:
This particular error will become more useful/rich after https://github.com/parcel-bundler/parcel/pull/3695 is merged, because currently the component that combines all the sources has no idea where the code came from (they are reparsed from a string internally).

devongovett:
This error is likely because you have inadvertently put Parcel into library mode with a `main` or `module` field in your package.json. This will cause Parcel to exclude node_modules by default, which would trigger this error at packaging time.

We can definitely make this error message a lot better though.

mischnic:
`require.resolve("./something.js")` results in:
`🚨 @parcel/packager-js: Got unexpected undefined`

Issue #4364, opened 2020-03-22
Fails to reexport property at build time


garthenweb:
# 🐛 bug report

Using `@sentry/browser` fails at build time with `Error: node_modules/@sentry/utils/esm/index.js does not export 'timestampWithMs'`.

## 🎛 Configuration (.babelrc, package.json, cli command)

See https://github.com/garthenweb/parcel-bug-report-examples/tree/bug/sentry-does-not-export-timestampWithMs

## 🤔 Expected Behavior
 
Should create a proper build that is executable in the browser.

## 😯 Current Behavior

Fails with `Error: node_modules/@sentry/utils/esm/index.js does not export 'timestampWithMs'`. See https://github.com/garthenweb/parcel-bug-report-examples/tree/bug/sentry-does-not-export-timestampWithMs for more.

## 💁 Possible Solution

`timestampWithMs` is exported correctly in `node_modules\@sentry\utils\esm\misc.js` which is reexported in `node_modules/@sentry/utils/esm/index.js`.

`timestampWithMs` uses another export called `crossPlatformPerformance`, maybe this is the source of the issue?

Maybe related to https://github.com/parcel-bundler/parcel/issues/4216 but the error is different.

Works without scope hoisting.

## 💻 Code Sample

See https://github.com/garthenweb/parcel-bug-report-examples/tree/bug/sentry-does-not-export-timestampWithMs

## 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Parcel           | 2.0.0-nightly.163
| Node             | v12.8.1
| npm         | v6.14.3
| Operating System | Windows (WSL v1/ Ubuntu Bash)

mischnic:
I've identified the problem, will be fixed after the refactoring in https://github.com/parcel-bundler/parcel/pull/4411 landed.

(Note to my future self: tests and an ugly fix can be found in https://github.com/parcel-bundler/parcel/tree/resolveSymbol-wrapped-reexport)

xtrinch:
@mischnic Are these fixes ever going to make it into 1.x?

mischnic:
No, also because Parcel 2's scope hoisting implementation has diverged (improved) too much at this point.

> If not, could one avoid tree shaking this particular module only?

That isn't possible.

drewsynan:
@xtrinch fwiw (in the sort term) pinning`@sentry/browser` to version "5.9.1" resolved this issue for me

xtrinch:
[...]

PR #4411, opened 2020-03-31
Emit codeframes for invalid imports when scopehoisting

mischnic:
# ↪️ Pull Request

Closes https://github.com/parcel-bundler/parcel/issues/3761
Close T-214

Actually, the scope-hoisting sourcemaps weren't useful for most cases (imports & exports) because we replace all identifiers that reference the imported binding with a new identifier, so the source location of the actual import is lost. For this reason,  `(Asset|Dependency}#symbolsLocs` were added, otherwise we could only error on the referencing identifier, not on the import specifier.

This is also makes it possible to synchronously throw a codeframe, by deferring the file reading into the reporter.

## 💻 Examples

Bildschirmfoto 2020-03-30 um 19 36 05
Bildschirmfoto 2020-03-31 um 23 42 44

Bildschirmfoto 2020-04-01 um 00 11 51

## Discussion

- Should the codeframe/error messages contain paths relative to projectRoot or cwd?
- I hope these changes don't have a measurable impact on performance for successful builds.


packages/core/core/src/public/Asset.js
```diff
@@ -120,10 +121,14 @@ class BaseAsset {
     return this.#asset.value.sideEffects;
   }
 
-  get symbols(): Map<Symbol, Symbol> {
+  get symbols(): $ReadOnlyMap<Symbol, Symbol> {
```
devongovett, line 124: 👍

devongovett:
> Actually, the scope-hoisting sourcemaps weren't useful for most cases (imports & exports) because we replace all identifiers that reference the imported binding with a new identifier, so the source location of the actual import is lost.

Should we attach the original location to the replacement node so sourcemaps work too?

mischnic:
> Should we attach the original location to the replacement node so sourcemaps work too?

You mean the location of the import declaration? How?
```js
// b.js
export const b = 4;

// a.js
import {b} from "./b.js"
console.log(b);
```
output:
```js
const $b$export$b = 4;
// <-- nowhere to attach to import declaration position
console.log($b$export$b); // <-- `$b$export$b` is should obviously map to the `b` argument
```

devongovett:
Ah I see

mischnic:
Ready for review!

packages/core/types/index.js
```diff
@@ -311,7 +316,9 @@ export interface BaseAsset {
   +isSplittable: ?boolean;
   +isSource: boolean;
   +type: string;
-  +symbols: Map<Symbol, Symbol>;
+  // (symbol exported by this -> name of binding to export)
+  +symbols: $ReadOnlyMap<Symbol, Symbol>;
+  +symbolsLocs: $ReadOnlyMap<Symbol, SourceLocation>;
```
devongovett, line 321: What do you think of keeping a single map for this rather than two maps? E.g. Map<Symbol, {identifier: Symbol, loc: SourceLocation}>

Also, should we make these internal and only expose the accessor methods? Might be nice to be able to change the data structures later if needed.

packages/core/types/index.js
```diff
@@ -286,8 +287,12 @@ export interface Dependency {
   +target: ?Target;
   +sourceAssetId: ?string;
   +sourcePath: ?string;
-  +symbols: Map<Symbol, Symbol>;
+  // (imported symbol -> variable that it is used as)
+  +symbols: $ReadOnlyMap<Symbol, Symbol>;
+  +symbolsLocs: $ReadOnlyMap<Symbol, SourceLocation>;
```
devongovett, line 292: Same idea here: Map<Symbol, {identifier: Symbol, loc: SourceLocation>, and make internal?

devongovett:
Minor comments above. Nothing blocking.

packages/core/types/index.js
mischnic: > What do you think of keeping a single map for this rather than two maps? E.g. Map<Symbol, {identifier: Symbol, loc: SourceLocation}>

For the internal assets, yes.

> Also, should we make these internal and only expose the accessor methods? Might be nice to be able to change the data structures later if needed.

What about a new `Symbols` type with `setSymbol, getSymbol, hasExportSymbol, hasLocalSymbol, getSymbols`? Otherwise the methods are duplicated between Dependency and Asset.

This would also go well with the fix for https://github.com/parcel-bundler/parcel/issues/4364, where we will need to have a distinction between "asset has no symbol" and "symbols not statically analyzable" (currently this is done via `asset.meta.isCommonJS`, but that cannot be used in `BundleGraph#resolveSymbol`), so `symbols` could be null in that case.

devongovett: Sounds reasonable to me!

packages/core/core/src/public/Symbols.js
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+  hasLocalSymbol(local: Symbol): boolean {
+    for (let s of this.#symbols.values()) {
+      if (local === s.local) return true;
+    }
+    return false;
+  }
+
+  getAll(): $ReadOnlyMap<Symbol, {|local: Symbol, loc: ?SourceLocation|}> {
```
devongovett, line 42: I know the flow type is read only, but it's technically writable because JS. Looks like this is only used to iterate over the symbols. So maybe we implement the iterable interface for Symbols instead? e.g. `[Symbol.iterator]`

mischnic: Done (and it looks cleaner anyway)

PR #4285, opened 2020-03-07
V2 > CLI > watch command: start HMR server by default on port 1234

wis:
# ↪️ Pull Request
fixes #4284

## 💻 Examples

described in #4284

## 🚨 Test instructions
listed in #4284  Code Sample section
I'd add:
- open the compiled HTML file in the browser without a server, the url should look like so:
`file:///home/user/dev/parcel-webextension/dist/index.html`

## ✔️ PR Todo

- [ ] Added/updated unit tests for this change
- [x] Filled out test instructions (In case there aren't any unit tests)
- [x] Included links to related issues/PRs
- [ ] use options in hmr-runtime to start the client on the right port,
instead of the currently hard-code port number: 1234.

Issue #4284, opened 2020-03-07
V2: watch command should (optionally?) start HMR server


wis:
# 🐛 bug report
it looks like in Parcel 2, unlike Parcel 1 the `watch` command doesn't start a HMR server despite the mention of `--hmr-port` and `--hmr-host` in the output of `parcel watch --help`

## 🎛 Configuration (.babelrc, package.json, cli command)
`parcel watch index.html --public-dir './'`

## 🤔 Expected Behavior
Parcel V2's watch command should have the option to start a HMR server like the serve command and V1's watch command.

## 😯 Current Behavior
no option to start HMR server with the watch command.

also when using `--hmr-port` and `--hmr-host` with the watch command as described in its help page,
the parser doesn't accept them:
```
error: unknown option `--hmr-port'
error: unknown option `--hmr-host'
```

## 💁 Possible Solution
add the option or make it default behavior to start HMR server and no HTTP server with the watch command.

## 🔦 Context
as I [explained](https://github.com/parcel-bundler/parcel/issues/4186#issuecomment-596039229) in issue #4186, I'm developing a WebExtension with Parcel V2 and React and I'm trying to get the awesome default support for Fast Refresh of Parcel V2 to work with web pages not served from a server.

## 💻 Code Sample
**index.html**:
```html
<!DOCTYPE html>
<html lang="en">
    <head>
        <meta charset="utf-8" />
        <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no" />
        <base href=".">
        <title>📦 Parcel</title>
    </head>
    <body>
        <noscript>You nttneed to enable JavaScript to run this application 😰</noscript>
        <div id="app"></div>

        <script src="./src/index.tsx"></script>
    </body>
</html>
```
**src/index.tsx**:
```js
import * as React from "react";
import { render } from "react-dom";
import { Componentx } from "./componentx";

class App extends React.Component {
  render() {
    return (
      <>
        <Componentx></Componentx>
      </>
    );
  }
}

render(<App />, document.getElementById("app"));
```
**src/componentx**:
```js
import React from "react";
export function Componentx() {
  return (
    <>
      <div>hello good world! </div>
    </>
  );
}
```

## 🌍 Your Environment
| Software         | Version(s) |
| ---------------- | ---------- |
| Parcel           | 2.0.0-alpha.3.2
| Node             | v13.9.0
| npm/Yarn         | 1.22.0
[...]

PR #4412, opened 2020-04-01
Fix HMR for file based loading and UXP by starting by default on port 1234

krisnye:
This is based on Wis's previous HMR PR here: https://github.com/parcel-bundler/parcel/pull/4285

# ↪️ Pull Request

Makes the --port option common to both server and watch modes.
Changes "hot" option type to { port?: number } | boolean
Allows HMR to function even when server is not started
Fixes client HMR to use "localhost:1234" when not loaded via HTTP. (Note, we cannot pull the port from the URL when loading by file so we use the default port).
Update the HMR unit tests.

## 💻 Examples

described in #4284

## 🚨 Test instructions

Listed in #4284 Code Sample section

Also, open the compiled HTML file in the browser without a server, the url should look like so:
file:///home/user/dev/parcel-webextension/dist/index.html

## ✔️ PR Todo

### DIFF
--- a/packages/core/parcel/package.json
+++ b/packages/core/parcel/package.json
@@ -10,7 +10,7 @@
     "type": "git",
     "url": "https://github.com/parcel-bundler/parcel.git"
   },
-  "bin": "src/bin.js",
+  "bin": "lib/bin.js",
   "main": "lib/bin.js",
   "source": "src/bin.js",
   "engines": {
--- a/packages/core/parcel/package.json
+++ b/packages/core/parcel/package.json
@@ -10,7 +10,7 @@
     "type": "git",
     "url": "https://github.com/parcel-bundler/parcel.git"
   },
-  "bin": "lib/bin.js",
+  "bin": "src/bin.js",
   "main": "lib/bin.js",
   "source": "src/bin.js",
   "engines": {
--- a/packages/runtimes/hmr/src/loaders/hmr-runtime.js
+++ b/packages/runtimes/hmr/src/loaders/hmr-runtime.js
@@ -27,12 +27,12 @@ var checkedAssets, assetsToAccept, acceptedAssets;
 // eslint-disable-next-line no-redeclare
 var parent = module.bundle.parent;
 if ((!parent || !parent.isParcelRequire) && typeof WebSocket !== 'undefined') {
-  var hostname = location.origin.startsWith('http')
-    ? location.hostname
-    : 'localhost';
-  var port = location.port ? ':' + location.port : ':1234';
+  var hostname = __PARCEL_HMR_HOST || location.hostname;
+  var port = __PARCEL_HMR_PORT || location.port;
   var protocol = location.protocol === 'https:' ? 'wss' : 'ws';
-  var ws = new WebSocket(protocol + '://' + hostname + port + '/');
+  var ws = new WebSocket(
+    protocol + '://' + hostname + (port ? ':' + port : '') + '/',
+  );
   ws.onmessage = function(event) {
     checkedAssets = {};
     assetsToAccept = [];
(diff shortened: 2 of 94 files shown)


packages/runtimes/hmr/src/loaders/hmr-runtime.js
```diff
@@ -27,8 +27,10 @@ var checkedAssets, assetsToAccept, acceptedAssets;
 // eslint-disable-next-line no-redeclare
 var parent = module.bundle.parent;
 if ((!parent || !parent.isParcelRequire) && typeof WebSocket !== 'undefined') {
-  var hostname = location.hostname;
-  var port = location.port ? ':' + location.port : '';
+  var hostname = location.origin.startsWith('http')
+    ? location.hostname
+    : 'localhost';
+  var port = location.port ? ':' + location.port : ':1234';
```
devongovett, line 33: We'll need to get this value from the options rather than hard coding 1234. I'd recommend writing a variable like `__PARCEL_HMR_PORT` [here](https://github.com/parcel-bundler/parcel/blob/v2/packages/runtimes/hmr/src/HMRRuntime.js#L23) and accessing it in place of this string. You can get the port from `options.hot.port`.

krisnye: So options.hot.port is in scope for the client at this point?

mischnic: No:
https://github.com/parcel-bundler/parcel/blob/003305315a3f96db9a3d5e40488c8e9b28d7dfd7/packages/runtimes/hmr/src/loaders/hmr-runtime.js#L28-L32
https://github.com/parcel-bundler/parcel/blob/003305315a3f96db9a3d5e40488c8e9b28d7dfd7/packages/core/parcel/src/cli.js#L283-L284
But we shouldn't set `process.env.HMR_HOSTNAME ` in the cli but rather in `@parcel/core` or an existing plugin.

krisnye: I couldn't see how to get environment properties into the client but I found I could add the variables here: https://github.com/parcel-bundler/parcel/blob/e3cbe12600c1ea30c42a22c822bdfe9016761dab/packages/runtimes/hmr/src/HMRRuntime.js#L23

packages/core/parcel/package.json
```diff
@@ -10,7 +10,7 @@
     "type": "git",
     "url": "https://github.com/parcel-bundler/parcel.git"
   },
-  "bin": "src/bin.js",
+  "bin": "lib/bin.js",
```
mischnic, line 13: Commited by accident

yarn.lock
```diff
@@ -3862,6 +3862,11 @@ cloneable-readable@^1.0.0:
     process-nextick-args "^2.0.0"
     readable-stream "^2.3.5"
 
+clones@^1.2.0:
```
mischnic, line 3865: Why did the lockfile change even though you didn't change any package.json?

packages/core/parcel/package.json
krisnye: I never changed that, is some automated script changing that or something?

yarn.lock
krisnye: There was a linting error saying that shared/babel-ast-utils/ was missing a dependency. I added it but then saw that something was merged back into master probably solving same problem. I will try to revert yarn.lock.

packages/core/parcel/package.json
mischnic: Yes, `yarn build` would do this.

krisnye: Seems weird. That's why I like to copy my package.json to a dist/ folder which is .gitignored prior to doing any modifications for publishing.

mischnic: The problem is that the binary needs to be specified in package.json, if we just left it as `lib/bin.js`, the `parcel` binary wouldn't work in development.
The `lib` folders are ignored by git.

mischnic: The `parcel` binary needs to be specified in package.json, if we didn't do this, running `parcel` wouldn't work inside the monorepo.

krisnye:
I butchered this PR branch trying to get linting errors fixed. I'm going to close this one and open a new one here: https://github.com/parcel-bundler/parcel/pull/4417

@mischnic @devongovett

PR #4417, opened 2020-04-02
Knye/hmr

krisnye:
This is based on Wis's previous HMR PR here: https://github.com/parcel-bundler/parcel/pull/4285

I messed up the git history and state so bad on the previous PR that I opened this one to fix things. (I have limited experience with advanced git commands like rebasing, reverting etc)

# ↪️ Pull Request

Makes the --host --port option common to both server and watch modes.
Changes "hot" option type to { port?: number, host?: number }
Allows HMR to function even when server is not started
Fixes client HMR to use "localhost:1234" when not loaded via HTTP. (Note, we cannot pull the port from the URL when loading by file so we use the default port).
Update the HMR unit tests.

## 💻 Examples

described in #4284

## 🚨 Test instructions

Listed in #4284 Code Sample section

Also, open the compiled HTML file in the browser without a server, the url should look like so:
file:///home/user/dev/parcel-webextension/dist/index.html

## ✔️ PR Todo


krisnye:
@mischnic please review this PR if you can. I closed the old one.

PR #4421, opened 2020-04-02
Share bundles when their entries exist in others

wbinnssmith:
Depends on #4391.

When a bundle `a`'s entry asset is present in bundle `b`, share bundle `a` by placing it (and its siblings) in `b`'s bundle group. This takes place in optimization in a similar way as other shared bundle extraction. See tests for  examples. 

This required making scope hoisted bundles always register init functions. @mischnic may have a better idea or may remove this with the symbol propagation work.

Test Plan: `yarn test`

### DIFF
--- a/packages/bundlers/default/src/DefaultBundler.js
+++ b/packages/bundlers/default/src/DefaultBundler.js
@@ -160,14 +160,60 @@ export default new Bundler({
   },
 
   optimize({bundleGraph}) {
-    // Step 2: remove assets that are duplicated in a parent bundle
+    // Step 2: Remove asset graphs that begin with entries to other bundles.
+    bundleGraph.traverseBundles(bundle => {
+      if (bundle.isInline || !bundle.isSplittable) {
+        return;
+      }
+
+      let mainEntry = bundle.getMainEntry();
+      if (mainEntry == null) {
+        return;
+      }
+
+      let siblings = bundleGraph
+        .getSiblingBundles(bundle)
+        .filter(sibling => !sibling.isInline);
+      let candidates = bundleGraph.findBundlesWithAsset(mainEntry).filter(
+        containingBundle =>
+          containingBundle.id !== bundle.id &&
+          // Don't add to BundleGroups for entry bundles, as that would require
+          // another entry bundle depending on these conditions, making it difficult
+          // to predict and reference.
+          !containingBundle.isEntry &&
+          !containingBundle.isInline &&
+          containingBundle.isSplittable,
+      );
+
+      for (let candidate of candidates) {
+        let bundleGroups = bundleGraph.getBundleGroupsContainingBundle(
+          candidate,
+        );
+        if (
+          Array.from(bundleGroups).every(
+            group =>
+              bundleGraph.getBundlesInBundleGroup(group).length <
+              OPTIONS.maxParallelRequests,
+          )
+        ) {
+          bundleGraph.removeAssetGraphFromBundle(mainEntry, candidate);
+          for (let bundleGroup of bundleGroups) {
+            for (let bundleToAdd of [bundle, ...siblings]) {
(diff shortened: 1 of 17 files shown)


packages/core/integration-tests/test/html.js
```diff
@@ -1016,24 +1016,24 @@ describe('html', function() {
       {production: true, scopeHoist: true},
     );
 
-    // a.html should point to a CSS bundle containing a.css and b.css.
-    // It should not point to the bundle for b.css from b.html.
+    // a.html should point to a CSS bundle containing a.css as well as
+    // reuse the b.css bundle from b.html.
```
wbinnssmith, line 1020: This is happening intentionally now 😄

packages/shared/scope-hoisting/src/formats/global.js
```diff
@@ -97,7 +101,7 @@ export function generateExports(
     exported.add(exportsId);
 
     statements.push(
-      EXPORT_TEMPLATE({
+      EXPORT_FN_TEMPLATE({
```
devongovett, line 104: Not sure I understand this change. Could you explain?

What about other output formats, e.g. ESM/CommonJS?

mischnic: With esm and commonjs, we directly import from the bundle, but with the global one, we register the exports via `parcelRequire.register(id, exports)`.
If a bundle has the asset twice in it's ancestry, both as a child bundle and as a shared bundle, this happened:
```js
// parent bundle 1 (child, 
// would naively not need a init because it is initialized by being loaded at runtime)
parcelRequire.register("1234", $1234$exports)
```
```js
// parent bundle 2 (sibling)
parcelRequire.register("1234", $1234$init)
```
Now we got a collision and don't know if we need to do `parcelRequire("1234")()` or not.

(Thinking about it, it actually does make a difference which of these you get because some state could be set in the asset - but that is another issue).

I hope I got this right, @wbinnssmith.

devongovett: Ah ok, that makes sense.

devongovett: Could we add a comment to the code explaining this?

wbinnssmith: done 😄

packages/bundlers/default/src/DefaultBundler.js
```diff
@@ -160,14 +160,60 @@ export default new Bundler({
...
+          !containingBundle.isInline &&
+          containingBundle.isSplittable,
+      );
+
+      for (let candidate of candidates) {
+        let bundleGroups = bundleGraph.getBundleGroupsContainingBundle(
+          candidate,
+        );
+        if (
+          Array.from(bundleGroups).every(
```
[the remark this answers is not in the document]
wbinnssmith: Done 😄 I've extracted the logic into a function we can use in #4295 as well.

packages/bundlers/default/src/DefaultBundler.js
```diff
@@ -160,14 +160,54 @@ export default new Bundler({
...
+
+      for (let candidate of candidates) {
+        let bundleGroups = bundleGraph.getBundleGroupsContainingBundle(
+          candidate,
+        );
+        if (!exceedsParallelRequestLimit(bundleGraph, bundleGroups)) {
+          bundleGraph.removeAssetGraphFromBundle(mainEntry, candidate);
+          for (let bundleGroup of bundleGroups) {
+            for (let bundleToAdd of [bundle, ...siblings]) {
+              bundleGraph.addBundleToBundleGroup(bundleToAdd, bundleGroup);
```
wbinnssmith, line 196: My only concern using async/entry bundles as shared bundles is that these bundles have asset entries — code that will run immediately once they are evaluated — unlike the shared bundles created below. This doesn't seem to have been a problem so far. What do you think, @devongovett?

PR #4434, opened 2020-04-05
 @parcel/transformer-babel: remove core-js dep

### LINKED ISSUE #4433: @parcel/transformer-babel: remove core-js dep?
# 🐛 bug report

It seems that this dependency is redundant:
https://github.com/parcel-bundler/parcel/blob/c690f91bcb28081ba370eacdb91b9b1b9a1baa6a/packages/transformers/babel/package.json#L32

Searching the code base, I only found `core-js` in the tests.

## 🔦 Context

`core-js` is particularly harmful due to https://github.com/zloirock/core-js/issues/548, https://github.com/zloirock/core-js/issues/611, https://github.com/zloirock/core-js/issues/708.

Any chance it can be removed?

fregante:
# ↪️ Pull Request

`core-js` isn't a direct dependency of `@parcel/transformer-babel` so it should not be listed as one.

Closes #4433 

## ✔️ PR Todo

- [/] Added/updated unit tests for this change
- [/] Filled out test instructions (In case there aren't any unit tests)

### DIFF
--- a/packages/transformers/babel/package.json
+++ b/packages/transformers/babel/package.json
@@ -29,7 +29,6 @@
     "@parcel/plugin": "^2.0.0-alpha.3.1",
     "@parcel/utils": "^2.0.0-alpha.3.1",
     "browserslist": "^4.6.6",
-    "core-js": "^3.2.1",
     "nullthrows": "^1.1.1",
     "semver": "^5.7.0"
   },


fregante:
I think that windows test failed due to an unrelated memory issue. Maybe try re-running it.

mischnic:
As I've already commented in https://github.com/parcel-bundler/parcel/issues/4433, `@babel/preset-env` uses `core-js` if `useBuiltins` is set, which we do here for the default babel config:
https://github.com/parcel-bundler/parcel/blob/f01437e7f646f4aa74a5d3cf48ed7c4f21822352/packages/transformers/babel/src/env.js#L39-L40

According to babel, you should install `core-js` in that case: 
> WARNING: We noticed you're using the `useBuiltIns` option without declaring a core-js version. Currently, we assume version 2.x when no version is passed. Since this default version will likely change in future versions of Babel, we recommend explicitly setting the core-js version you are using via the `corejs` option.
>
>You should also be sure that the version you pass to the `corejs` option matches the version specified in your `package.json`'s `dependencies` section. If it doesn't, you need to run one of the following commands:
>
>  `npm install --save core-js@2` `npm install --save core-js@3`
>  `yarn add core-js@2` `yarn add core-js@3`

https://github.com/babel/babel/blob/aeebc0823415ed4770c633cedf570cf774ae89c6/packages/babel-preset-env/src/normalize-options.js#L211-L222

fregante:
> According to babel, you should install `core-js` in that case:

Indeed, so https://github.com/parcel-bundler/parcel/issues/4433 is invalid and should be closed?

PR #4447, opened 2020-04-07
Random fixes

padmaia:
# ↪️ Pull Request
Some one line fixes.


packages/core/parcel/src/cli.js
```diff
@@ -72,7 +72,7 @@ var hmrOptions = {
   '--no-hmr': 'disable hot module replacement',
   '-p, --port <port>': [
     'set the port to serve on. defaults to 1234',
-    parseInt,
+    value => parseInt(value, 10),
```
padmaia, line 75: Commander passes the default as the second arg to the given parse function.

Issue #2180, opened 2018-10-21
Conflicting asset ids leads to wrong file being required


pselden:
# 🐛 bug report

I ran into an issue only seen when building in production that when lodash's clone method was called it would lead to the following error:

```
_baseClone.js:142 Uncaught TypeError: e is not a constructor
    at Y (_baseClone.js:142)
    at n (clone.js:33)
    at r.get (memo.js:123)
    at e.get (Transaction.js:53)
    at On (mobx.module.js:3120)
    at e.computeValue (mobx.module.js:1602)
    at e.trackAndCompute (mobx.module.js:1591)
    at e.get (mobx.module.js:1550)
    at e.<anonymous> (mobx.module.js:3479)
    at e.get (mobx.module.js:1055)
```

I tracked it down to this line here: https://github.com/lodash/lodash/blob/master/.internal/baseClone.js#L200 and added some extra logging around it. What I found was that the "Stack" variable, which should be a reference to an imported lodash function, was actually the string '/authorize_change_trust.5c3cd8f3.png', which is the generated filename for one of the images that import in another file.

I tracked it down further and found that the requires map to the same ID: 

```
"./authorize_change_trust.png":"3+VT"
"./_Stack":"3+VT"
```
So somehow, only in the production build, these things are being mapped to the same id. Everything is fine in dev/watch mode.

## 🎛 Configuration (.babelrc, package.json, cli command)

.babelrc
```js
{
  "plugins": [
    [
      "direct-import",
      [
        "@material-ui/core",
        "@material-ui/icons",
        "react-router",
        "react-router-dom"
      ]
    ],
    "transform-decorators-legacy",
    "transform-export-extensions",
    "transform-class-properties",
    "transform-object-rest-spread",
    [
      "babel-plugin-transform-builtin-extend",
      {
        "globals": ["Error"]
      }
    ]
  ],
  "presets": [
    [
      "env",
      {
        "targets": {
          "browsers": ["last 2 versions", "> 1%"]
        },
        "useBuiltIns": "entry"
      }
    ],
    "react"
  ]
}

```

## 💻 Code Sample

This is happening here: https://github.com/stellarguard/stellarguard but it's happening in a dependency of this project, not directly in the call site.

I will try to come up with a smaller reproduction step but I'm not sure I'll be able to since the bug probably relies on a lot of little things falling into place to cause a clash.

## 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
[...]

PR #4449, opened 2020-04-08
Use a more concise public id to identify assets and bundles

### LINKED ISSUE #4446: Consider shorter module name aliases
# 💬 RFC

Parcel v2 uses 32 symbol hex string (i.e. `98a8b4be7438864cba762c83cd5f7e6d`) to identify a module while v1 used various length `[A-Za-z0-9]` string (i.e. `Sh3W`). This results in larger builds in v2, and gzip won't help here. For example, the same file went from 95 KB to 103 KB (+ 8%) for me simply when switching to v2. Not a big deal, but still may be worth considering going back to older scheme.

wbinnssmith:
Resolves #4446

This makes Assets and Bundles use a more concise "public id" to identify themselves globally. Currently, we write entire asset ids (128bit values originating from md5, encoded as 32-character hexadecimal strings) into bundles frequently. This should reduce the size of bundles. It also does the same for when bundles are referred to in the global runtime manifest in js files.

~**This reduces the keyspace of ids and increases the likelihood of collisions**~. This now addresses collisions by beginning with the first 5 characters of the base64 version of the id, and growing it as necessary to avoid collisions.

The public id is now implemented by encoding the ~first 2 bytes of the~ 128bit id as base64, and truncating it to the first 5 characters. If collisions are encountered, the id grows to accommodate unless it's identical, in which case an exception is thrown. I'm happy to discuss alternatives, and we should test in real-world applications. 

Assets are generated in parallel during transformation, where their ids are determined, and are not mutated after this. Therefore, it would be difficult to detect collisions and e.g. find a neighboring id to use, free of the collision. In comparison, webpack uses a 4 character id, but says that it may use longer ids to avoid collisions [0]. Parcel 1 also used 4 character ids but @devongovett and @demoorjasper mentioned that it ran into collisions. 

Test Plan: `yarn test`

[0] https://webpack.js.org/plugins/hashed-module-ids-plugin/


packages/core/core/src/public/utils.js
```diff
@@ -0,0 +1,17 @@
...
+    id.match(/^[0-9a-f]{32}$/),
+    `id ${id} must be a 32-character hexadecimal string`,
+  );
+
+  let buf = Buffer.alloc(8);
+  // Add the integer values to the buffer
+  // https://stackoverflow.com/questions/8044543/how-can-i-store-an-integer-in-a-nodejs-buffer/53550757#53550757
+  buf.writeUInt32BE(parseInt(id.slice(0, 8), 16), 0);
+  buf.writeUInt32BE(parseInt(id.slice(8, 16), 16), 4);
+  return buf.toString('base64').slice(0, 6);
```
wbinnssmith, line 16: We could also use the url-friendly base64url [0], and the namer could even use that in the bundle's filename as well. Or maybe base62 — at least in the bundle case — to avoid special characters all together.

[0] https://tools.ietf.org/html/rfc4648#page-7

DeMoorJasper: I don't think we should shrink the id at all, it will probably reintroduce the issues we had in parcel 1, https://github.com/parcel-bundler/parcel/issues/2180 and https://github.com/parcel-bundler/parcel/pull/2999

We could encode it with base62 or something to not lose any data as we don't have to remove any non-alphanumeric characters

mischnic:
Would fix https://github.com/parcel-bundler/parcel/issues/4446?

packages/core/core/src/public/Bundle.js
```diff
@@ -168,20 +184,38 @@ export class NamedBundle extends Bundle implements INamedBundle {
...
     bundle: InternalBundle,
     bundleGraph: BundleGraph,
     options: ParcelOptions,
   ) {
+    super(sentinel, bundle, bundleGraph, options);
+    this.#bundle = bundle; // Repeating for flow
+    this.#bundleGraph = bundleGraph; // Repeating for flow
+  }
+
+  static get(
```
wbinnssmith, line 197: This resolves a bug that caused `Bundle`s to be returned instead of `NamedBundle`s as the `super()` call found the memoized `Bundle` and made that `this`.

packages/core/core/src/public/Bundle.js
```diff
@@ -168,20 +184,38 @@ export class NamedBundle extends Bundle implements INamedBundle {
   #bundleGraph; // BundleGraph
 
   constructor(
+    sentinel: mixed,
```
wbinnssmith, line 187: This prevents anything but something in this file from invoking this constructor 😄

wbinnssmith:
One thing we should consider is that public asset ids, whose length will vary depending on the other assets present in the graph because of the potential for collisions, may impact the contents of what is stored in the package cache. If a cache entry for a bundle from a prior build is used, the public id for an asset may be different than how it is referenced in another bundle.

Maybe we should just continue with truly universal public ids like we have now, but encode them as base64 to go from 32 characters to 22. cc @padmaia @devongovett

packages/core/core/src/BundleGraph.js
```diff
@@ -41,6 +41,8 @@ type BundleGraphEdgeTypes =
   | 'internal_async';
 
 export default class BundleGraph {
+  _assetPublicIds: Set<string>;
+  _publicIdByAssetId: Map<string, string>;
```
wbinnssmith, line 45: These need to be store on the BundleGraph because we'll need them when we merge the runtimes graph into the main bundle graph. We want to re-use any generated public id for an asset (thus the asset id => public id map), but otherwise want to quickly detect if there's a collision (the set of public ids).

packages/core/core/src/utils.js
```diff
@@ -51,3 +52,35 @@ export function registerCoreWithSerializer() {
 
   coreRegistered = true;
 }
+
+export function getPublicId(
+  id: string,
+  alreadyExists: string => boolean,
```
DeMoorJasper, line 58: Wouldn't this introduce the possibility of generating different id's for each build?
Weren't ids supposed to remain the same across builds for caching?

packages/core/core/src/utils.js
```diff
@@ -51,3 +52,35 @@ export function registerCoreWithSerializer() {
...
+    // https://stackoverflow.com/questions/8044543/how-can-i-store-an-integer-in-a-nodejs-buffer/53550757#53550757
+    buf.writeUInt32BE(
+      parseInt(id.slice(byteOffset * 2, byteOffset * 2 + 8), 16),
+      byteOffset,
+    );
+  }
+
+  // 128-bit values are represented in the first 22 characters
+  // of the base64 string. The remaining two are always '=='.
+  let base64 = buf.toString('base64').slice(0, 22);
```
DeMoorJasper, line 77: Maybe good to remove special characters like we did in Parcel 1 instead (as we never really use the entire string anyway), there's also other characters in here besides `=`

wbinnssmith: All base64 characters should be fine in a string literal where they're currently used. Did you have another use in mind?

DeMoorJasper: It was just a nitpick in case it would ever be used as part of a variable name or something, like we did with scope-hoisting, but that still uses the long id's. You can leave it as is

wbinnssmith:
> One thing we should consider is that public asset ids, whose length will vary depending on the other assets present in the graph because of the potential for collisions, may impact the contents of what is stored in the package cache. If a cache entry for a bundle from a prior build is used, the public id for an asset may be different than how it is referenced in another bundle.
> 
> Maybe we should just continue with truly universal public ids like we have now, but encode them as base64 to go from 32 characters to 22. cc @padmaia @devongovett

Any thoughts here @padmaia @devongovett? I think this is likely the only blocker for landing this PR.

packages/core/core/src/public/Asset.js
```diff
@@ -196,6 +196,10 @@ export class Asset extends BaseAsset implements IAsset {
   get stats(): Stats {
     return this.#asset.value.stats;
   }
+
+  get publicId(): string {
```
devongovett, line 200: Should we expose `asset.id` at all anymore? Should `asset.id` be the public id?

packages/core/core/src/public/Bundle.js
```diff
@@ -48,29 +48,47 @@ export function bundleToInternalBundleGraph(bundle: IBundle): BundleGraph {
...
+    if (sentinel !== _private) {
+      throw new Error('Unexpected public usage');
     }
 
     this.#bundle = bundle;
     this.#bundleGraph = bundleGraph;
     this.#options = options;
+  }
+
+  static get(
```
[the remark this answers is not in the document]
wbinnssmith: This is explained here: https://github.com/parcel-bundler/parcel/pull/4449/files/e9481aafd066fd23a3a314505e5614a7b61f6f25#r405938493

packages/core/core/src/public/utils.js
wbinnssmith: I like the idea of using base62 here combined with using @devongovett's suggestion of only ever exposing the public id. base64 won't be safe to use in a variable name, but as long as there's a prefix (variable names can't begin with a digit), base62 would work in that case as well.

packages/core/core/src/applyRuntimes.js
```diff
@@ -128,7 +138,7 @@ export default async function applyRuntimes({
       }
     }, runtimeNode);
 
-    runtimesGraph.traverse((node, _, actions) => {
+    runtimesGraph._graph.traverse((node, _, actions) => {
```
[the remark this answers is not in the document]
wbinnssmith: `runtimesGraph` is now a BundleGraph (which has a Graph), not just a Graph.

I can't remember if BundleGraph ever had a traverse method itself, but it would make sense to have one.

packages/core/core/src/public/Asset.js
wbinnssmith: Unfortunately we need the entire asset graph to determine asset public ids, as they may need to be extended in length to mitigate collisions. This means that during transformation, these public ids won't be available, as the graph is not yet complete and we can't independently generate these shorter ids. Asset ids are used in at least hoisting in JS transformation.

I'd like to use public ids for the identifiers we create in scope hoisting, but unfortunately we need them in hoisting (transformation). We could insert placeholders and replace them during `concat` or `link` in packaging. What do you think, @devongovett @mischnic?

PR #4469, opened 2020-04-12
Correct bundle order inside a bundle group

mischnic:
# ↪️ Pull Request

- Script tags for sibling JS bundles were inserted in a random order, instead keep use insertion order (of `addBundleToBundleGroup`):
graph

- Insert `defineInteropFlag` calls for every referenced assets (previously they were missing for "top-level assets" (because they were only inserted when processing `$parcel$require(...)` calls (which where intra-bundle, not inter-bundle).

# Questions

Is it enough to simply rely on `addBundleToBundleGroup`? The DefaultBundler is quite overwhelming


mischnic:
(`run` in test-utils can now run execute all scripts tags in a html bundle in the correct order.)

PR #4476, opened 2020-04-14
feat(cli): Add --global option in commonOptions in cli

marosca:
# ↪️ Pull Request
With this pr I want to add --global option to cli like this other pr that was merge into Parcel1 code:
https://github.com/parcel-bundler/parcel/pull/453/files

The purpose is to reestablish the functionality of having access globally to the exports

### DIFF
--- a/packages/core/parcel/src/cli.js
+++ b/packages/core/parcel/src/cli.js
@@ -47,10 +47,11 @@ if (typeof process.env.PARCEL_BUILD_ENV === 'string') {
 
 program.version(version);
 
-// --no-cache, --cache-dir, --no-source-maps, --no-autoinstall, --global?, --public-url, --log-level
+// --no-cache, --cache-dir, --no-source-maps, --no-autoinstall, --public-url, --log-level
 // --no-content-hash, --experimental-scope-hoisting, --detailed-report
 
 const commonOptions = {
+  '--global <variable>': 'expose your module through a global variable',,
   '--no-cache': 'disable the filesystem cache',
   '--cache-dir <path>': 'set the cache directory. defaults to ".parcel-cache"',
   '--no-source-maps': 'disable sourcemaps',
--- a/packages/core/parcel/src/cli.js
+++ b/packages/core/parcel/src/cli.js
@@ -320,5 +320,6 @@ async function normalizeOptions(command): Promise<InitialParcelOptions> {
     env: {
       NODE_ENV: nodeEnv,
     },
+    global: command.global
   };
 }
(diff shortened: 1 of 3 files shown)


mischnic:
With your change, the option would do nothing - to actually make it work:

1. It needs to be added here as well:
https://github.com/parcel-bundler/parcel/blob/3a4e13fa424ef7c17dfc3d52ee1638a61dd58fdf/packages/core/types/index.js#L210
and
https://github.com/parcel-bundler/parcel/blob/3a4e13fa424ef7c17dfc3d52ee1638a61dd58fdf/packages/core/types/index.js#L214-L229
2. Needs to be added to the cli handling:
https://github.com/parcel-bundler/parcel/blob/3a4e13fa424ef7c17dfc3d52ee1638a61dd58fdf/packages/core/parcel/src/cli.js#L305-L321
3. Actually set in in the `JSPackager` (just like in the PR you linked):
https://github.com/parcel-bundler/parcel/blob/3a4e13fa424ef7c17dfc3d52ee1638a61dd58fdf/packages/packagers/js/src/JSPackager.js#L188
4. We need to figure something out for scope-hoisting, probably in this function: 
https://github.com/parcel-bundler/parcel/blob/3a4e13fa424ef7c17dfc3d52ee1638a61dd58fdf/packages/shared/scope-hoisting/src/formats/global.js#L69

marosca:
Thanks for your help
For this comment (We need to figure something out for scope-hoisting, probably in this function:) I dont know what I have to do exactly.

mischnic:
> I dont know what I have to do exactly.

I'll have to think about it a bit more, but `${global} = ${entry.meta.exportsIdentifier}` should be inserted if the global option is used and if `bundle` is the main bundle (not sure how to determine that) in `packages/shared/scope-hoisting/src/formats/global.js#generateExports`.

Will report back once I have found a solution.

marosca:
Really, thank you very much. In our project we need to have global access to exports and we do it with Parcel1. We are interested in using Parcel2 but we miss that functionality.

We greatly appreciate your help

devongovett:
I don't think this should necessarily be exposed as a CLI option, especially now that we have multiple targets. Perhaps it should be part of the target config in package.json? Should we make this part of a more general UMD output format?

jbaiter:
> I'll have to think about it a bit more, but ${global} = ${entry.meta.exportsIdentifier} should be inserted if the global option is used and if bundle is the main bundle (not sure how to determine that) in packages/shared/scope-hoisting/src/formats/global.js#generateExports.
>
> Will report back once I have found a solution.

Has there been progress on this? We're relying on this feature as well and would love to try out Parcel v2 :-)

mischnic:
Superseded by #7240

Issue #3839, opened 2019-11-23
parce@2.0.0-alpha.3.2 corrupts module styles when CSS files are edited


bvaughn:
I'm using PostCSS [as suggested here](https://parceljs.org/css.html#postcss) with the following `.postcssrc` configuration:
```
{
  "modules": true,
  "plugins": {
  }
}
```
It seems that any time I edit a `.css` file, Parcel loses all of the styles defined in that module and I need to wipe out the `.parcel-cache` directory entirely and restart Parcel to fix it.

mischnic:
Doesn't seem to happen with: 🤔 
```html
<body>hello</body>
<script type="text/javascript" src="./index.js"></script>
```
```js
import { Main } from "./index.css";

document.body.className = Main;
```
```css
.Main {
	color: blue;
	background-color: orange;
	font-size: 200;
}
```

bvaughn:
My apologies. I should have created a smaller repro case. I have done so now:
https://github.com/bvaughn/parcel-alpha-postcss-bug

In this smaller project, the bug dose not trigger _every time_ :frown: If I edit `index.css` and `App.css` a couple of times though, the bug _does_ trigger (empty imported `styles` object), requiring me to remove `.parcel-cache`. If I am able to come up with a more predictable repro case, I'll update this issue. (Sorry!)

May also be worth pointing out that every time I edit a `.css` file, I see the following error in the console:
```
Uncaught SyntaxError: Unexpected token '.'
    at new Function (<anonymous>)
    at hmrApply (HMRRuntime.js:180)
    at HMRRuntime.js:60
    at Array.forEach (<anonymous>)
    at WebSocket.ws.onmessage (HMRRuntime.js:61)
```

devongovett:
I think this is a bug in HMR. Seems to be sending CSS assets over the web socket, and then trying to evaluate them as JS. Oops. 😬

mischnic:
The second part yes, V1's `parcel-bundler/src/builtins/css-loader.js` isn't ported yet.

Still couldn't reproduce the caching issue

bvaughn:
Yeah, I think the two issues are separate :grin: Just mentioned the HMR one while I was here since I was also seeing it.

Still not sure how to make the first issue reproducible without just monkeying around a bit :frown: Sorry

KatSick:
I have the same issue when changing .module CSS file. I see an error in the console:

```
Uncaught SyntaxError: Unexpected token '.'
[...]

PR #4491, opened 2020-04-17
Fix currently broken CSS HMR by appending payload as a new style element

krisnye:
This fixes the "Unexpected token '.'" errors reported in #3839 . I'm not sure if it relates to the original bug or not.

# ↪️ Pull Request

Currently editing .css files results in HMR error in the browser:

```
Uncaught SyntaxError: Unexpected token '.'
    at new Function (<anonymous>)
    at hmrApply (HMRRuntime.js:180)
    at HMRRuntime.js:60
    at WebSocket.ws.onmessage (HMRRuntime.js:61)
```

This is because all updates are treated as Javascript. This checks if the value is type "css" and then appends it as a new stylesheet to the bottom of the document.body. This behavior is not perfect since it should ideally replace only the style section that it maps to, but it is much better than throwing an error and not working at all.

## 🚨 Test instructions

In any sample that uses normal plain .css imports, edit the css while serving the file with HMR.
Check the console, verify there should be no error, and the css change should be applied.

## ✔️ PR Todo


mischnic:
Ah, I get this as well now.

```html
<script type="text/javascript" src="./index.js"></script>
```

```js
import "./style.css"
module.hot.accept();
```

```css
body {
	background: blue;
}
```

Parcel 1 had constant hashes in development and on a HMR, we added a query to the script src to force a reload.
This way, many HMR updates could pile up. Not sure what the best solution is

Issue #2749, opened 2019-03-08
Cannot read property 'type' of undefined


WayneHiller:
# 🐛 bug report

I have a small VueJS project. I am running "parcel watch index.html --log-level 4 --out-dir wwwroot" from npm. On version 1.11.0 it would stop reacting to file changes after a few saves. I updated to version 1.12.0 and the same thing is happening but now I get an error (see below).

Note: I downgraded to parcel version 1.10.3 and everything works as expected when watching. 

## 🎛 Configuration (.babelrc, package.json, cli command)

```package.json
{
  "name": "abs-setup",
  "description": "ABS Setup",
  "author": "Wayne Hiller",
  "scripts": {
    "dev": "parcel watch index.html --log-level 4 --out-dir wwwroot",
    "build": "parcel build index.html --out-dir wwwroot",
    "publish": "npm run build && del-cli bin/Release/netcoreapp2.2/publish && dotnet publish -c Release /p:PublishProfile=Properties\\\\PublishProfiles\\\\FolderProfile.pubxml"
  },
  "devDependencies": {
    "@aspnet/signalr": "1.1.2",
    "@babel/core": "7.3.4",
    "@babel/preset-env": "7.3.4",
    "@vue/component-compiler-utils": "^2.6.0",
    "axios": "0.18.0",
    "bootstrap": "4.3.1",
    "bootstrap-vue": "2.0.0-rc.11",
    "less": "^3.9.0",
    "parcel-bundler": "1.12.0",
    "parcel-plugin-clean-easy": "1.0.2",
    "vue": "2.6.7",
    "vue-axios": "2.1.4",
    "vue-template-compiler": "^2.6.7",
    "del-cli": "1.1.0"
  },
  "dependencies": {
    "vue-hot-reload-api": "^2.3.3"
  },
  "parcelCleanPaths": [
    "wwwroot/*.*",
    "wwwroot/debug"
  ]
}
```

## 🤔 Expected Behavior

No error

## 😯 Current Behavior

Error being generated after 5 or 6 saves of the same .vue file.

`
[12:13:38 PM]: ×  Cannot read property 'type' of undefined
[12:13:38 PM]:     at Bundler.createBundleTree (E:\Development\AIN 2\ABS\Setup\ABS Setup\node_modules\parcel-bundler\src\Bundler.js:652:54)
[...]

PR #4501, opened 2020-04-20
Fixed crash "Cannot read property 'type' of undefined"

lunarzale:
# ↪️ Pull Request

This fixes an error where `bundle.findCommonAncestor()` will return `undefined`, which crashes when rebuilding a project due to a file change. I believe this fixes the bug #2749

## 🚨 Test instructions

To test: start with https://github.com/ourcade/phaser3-parcel-template, clone and follow instructions. Without the fix, `npm run start` will crash with above exception when a watched file is changed (for example, modifying HelloWorldScene.js). With this PR, the automatic rebuild will succeed.

## ✔️ PR Todo


devongovett:
We're no longer updating parcel 1.

PR #4506, opened 2020-04-20
added no-content-hashing option. fixed css-hmr, content hashing is disabled if HMR is active

krisnye:
Makes output bundles use stable names instead of varying based on content hash when HMR is active.
Changes CSS HMR to append timestamp to all css links in order to force reload.

Close T-278

# ↪️ Pull Request

Bundles use stable names if options.hot is true, otherwise normal content hashed names.
Client HMR now reloads all local style links upon receiving an update.

## 🚨 Test instructions

Edit a .css file while running parcel with HMR and verify that you see the change and the URL of all style links has changed. Edit again to verify that the CSS reloading still works after initial url change.

## ✔️ PR Todo


packages/core/core/src/PackagerRunner.js
```diff
@@ -566,12 +576,17 @@ function assignComplexNameHashes(hashRefToNameHash, bundles, bundleInfoMap) {
...
     ];
 
-    hashRefToNameHash.set(
-      bundle.hashReference,
-      md5FromString(
-        includedBundles.map(bundleId => bundleInfoMap[bundleId].hash).join(':'),
-      ).slice(-8),
-    );
+    // we want stable names with Hot Module Reloading
+    // we want names based on content during production for cacheability
```
mischnic, line 580: Should the names be stable only with HMR or always in development mode?

krisnye: We only need it with HMR and if I change it so that it's stable in any mode but 'production' then it breaks two unit tests that expect output file names to change with content changes. If I swap those unit tests over to mode: 'production' then it breaks 40+ tests for some reason that I haven't been able to determine.

devongovett: I think ideally this would be behind a flag for content hashing, like --no-content-hash from v1. That should be false by default in development mode.

krisnye: I added no-content-hashing.

content-hashing is automatically set to false if HMR is enabled.

padmaia:
Would it work to conditionally add a hash reference in the default namer based on the new contentHash option? (See [here](https://github.com/parcel-bundler/parcel/blob/v2/packages/namers/default/src/DefaultNamer.js#L94)) The packager makes some optimizations when there are no hash references present in a bundle. You'll need to use something else to avoid naming collisions. For now you could probably use the bundle id until [this PR](https://github.com/parcel-bundler/parcel/pull/4449) lands, in which case you would probably use the public id.

packages/runtimes/hmr/src/loaders/hmr-runtime.js
```diff
@@ -170,6 +170,39 @@ function getParents(bundle, id) {
...
+    var links = document.querySelectorAll('link[rel="stylesheet"]');
+    for (var i = 0; i < links.length; i++) {
+      var absolute = /^https?:\/\//i.test(links[i].getAttribute('href'));
+      if (!absolute) {
+        updateLink(links[i]);
+      }
+    }
+
+    cssTimeout = null;
+  }, 50);
```
wbinnssmith, line 203: what's the rationale for `50`? Can this be in a comment?

mischnic: This is just copied from Parcel 1: https://github.com/parcel-bundler/parcel/commit/2cc14c8da8a8b8b699e44f4f5fa87cc731e0ed68#diff-4471b25fd54ef2af9a791fe450751033
So we should ask @devongovett 😄

mischnic:
> The packager makes some optimizations when there are no hash references present in a bundle

Just to make sure: is that the case now that the paths returned by namers never contain the `HASH_REF_PREFIX`?

padmaia:
@mischnic yes

Issue #4518, opened 2020-04-23
Single command that can both "watch" and "serve" files


krisnye:
# 🙋 feature request

In a very simple project I can run "parcel watch src/index.html" and it will build the results into "./dist" folder and watch for changes (as well as run HMR server).

If I run "parcel serve src/index.html" it will NOT build that folder but it will serve the results over http.

I want a single command that will do both of the above.

I DO NOT want to have to run both of these as it's a pain and using & to run them in parallel leaves one as a background process.

## 🤔 Expected Behavior

Either 'parcel watch' should also serve files over http or else 'parcel serve' should also build to ./dist (if there are not multiple targets specified in the package.json)

## 😯 Current Behavior

@see above

## 💁 Possible Solution

I'll attach a PR for this later.

## 🔦 Context

We have a special native browser at Adobe which needs to use the output in the filesystem from 'parcel watch'.

We also need to test in a browser over http.

I want to be able to run a single command to serve up both with HMR.

mischnic:
There is a difference between "watch mode with HTTP server" and "serve mode with that outputs to `dist`" if you don't specify any targets.

Would `parcel serve src/index.html --dist-dir ./something` work for you?

krisnye:
Yes, that would work, and my PR actually makes it so that would work (which it wouldn’t before since serve would always force usage of .parcel-cache/dist).

I don't think it's ideal though.

Let me see if I understand current behavior.

If you have multiple defined targets in package.json then:

watch mode will build and incrementally rebuild all of them.
serve mode will throw an error if you don’t specify a target as it can't host them all on the same port.
if you specify a single target then it DOES build it to the specified dist directory.

If you do NOT have any targets defined then:

watch mode will build out to ./dist
serve mode will NOT build out to the ./dist directory

It seems for consistency that having no targets should be equivalent to having one default target which writes to ./dist in which case `parcel serve index.html` should write to ./dist.

devongovett:
Proposal:

* Write to `./dist` by default if there are no targets defined in package.json
* If targets are defined, then write into .parcel-cache/dist by default, unless a target to build is specified on the cli with `--target`.
* If `--dist-dir` is specified, then write into that

mischnic:
At some point, this was changed so that `parcel serve` also outputs to `./dist` by default instead of `.parcel-cache/dist`

PR #4519, opened 2020-04-23
Serve mode will now use --dist-dir if supplied else .parcel-cache/dist if package.targets is supplied else ./dist

krisnye:
# ↪️ Pull Request

When starting in 'watch' mode, this will now also start the server.
This also fixes some hard coded repetitions in referring to the distDir.

Related discussion: #4518 

## 💻 Examples

If you run "parcel watch src/index.html" in addition to building the /dist folder it will also serve the files on the default (or specified) port.

## 🚨 Test instructions

Run any sample parcel watch command and then open your browser to http://localhost:1234

## ✔️ PR Todo


packages/reporters/dev-server/src/Server.js
```diff
@@ -176,6 +185,21 @@ export default class Server extends EventEmitter {
     );
   }
 
+  getFilePathFromBundleGraph(name: string) {
```
devongovett, line 188: Possibly not needed now that we're using the distDir from options?

krisnye: I removed it.

packages/core/core/src/TargetResolver.js
```diff
@@ -176,7 +176,16 @@ export default class TargetResolver {
             // For serve, write the `dist` to inside the parcel cache, which is
             // temporary, likely in a .gitignore or similar, but still readily
             // available for introspection by the user if necessary.
-            distDir: path.resolve(this.options.cacheDir, DEFAULT_DIST_DIRNAME),
+            // unless there are no package targets in which case we write to the ./dist consistent with watch
+            // @see https://github.com/parcel-bundler/parcel/issues/4518
+            distDir:
+              this.options.distDir ??
+              path.resolve(
```
devongovett, line 183: This should never happen now, because you defaulted `this.options.distDir` in resolveOptions.js.

packages/core/parcel/package.json
```diff
@@ -10,7 +10,7 @@
     "type": "git",
     "url": "https://github.com/parcel-bundler/parcel.git"
   },
-  "bin": "src/bin.js",
+  "bin": "lib/bin.js",
```
mischnic, line 13: Revert this (artifact from `yarn build`)

packages/core/types/index.js
```diff
@@ -220,6 +220,7 @@ export interface PluginOptions {
   +autoinstall: boolean;
   +logLevel: LogLevel;
   +rootDir: FilePath;
+  +distDir: FilePath;
```
mischnic, line 223: I would say this property should rather be on the `ServerOptions` to not clearly separate "server root" from the "bundle output dist dir per target".

packages/core/core/src/TargetResolver.js
krisnye: That was left from before I defaulted options.distDir. I removed this code.

ParcelOptions.distDir is now a required value so downstream consumers can rely on it and not choose or hardcode their own directories.

packages/core/parcel/package.json
krisnye: Fixed. How do you guys avoid this happening all the time?

packages/core/types/index.js
krisnye: I know that the I need a distDir to server from here: https://github.com/parcel-bundler/parcel/blob/edab2589cfe689d009d428165e6188afe563699f/packages/reporters/dev-server/src/ServerReporter.js#L30

@devongovett  what do you think on this?

mischnic: @devongovett ?

devongovett: 🤷 I think it's ok... it does reflect the original distDir option that's passed to the API.

robinheinze:
This is awesome! Exactly the feature I'm looking for! Any chance this will be in a release soon?

mischnic:
Soon™️

PR #4531, opened 2020-04-25
delete dead branches in process visitor to reduce dead dependencies

### LINKED ISSUE #4524: Dead Code of Dependencies not removed
# 🐛 bug report

When using if-else cases to differentiate between e.g. browser and server but also for different build env, unused dependencies are not removed.

The code is not failing in any way, it is just bigger than it needs to be.

See https://github.com/garthenweb/parcel-bug-report-examples/tree/bug/dead-code-elemination-dependency-wrong-branch

## 🤔 Expected Behavior

Should remove all unused code.

## 😯 Current Behavior

Keeps the library code.

## 💁 Possible Solution

* insert Buffer library after dead code elimination?
* remove Buffer library after dead code elimination?
* find out that Buffer is not used and split it out?
* insert the Buffer library into the branch where it is used in?

## 💻 Code Sample

```javascript
const btoa = process.browser
  ? globalThis.btoa
  : (str) => Buffer.from(str.toString(), "binary").toString("base64");

console.log(btoa("test"));
```
https://github.com/garthenweb/parcel-bug-report-examples/blob/bug/dead-code-elemination-dependency-wrong-branch/src/index.js

The same happens for the following code

```javascript
import { act } from 'react-dom/test-utils'
const fireAsAct = (cb: () => void) => {
  if (process.env.NODE_ENV === "test") {
    act(() => cb());
  } else {
    cb();
  }
};

fireAsAct(() => console.log('test'));
```

For the latter example, I am actually not sure if it is a bug, as the `react-dom` does not have `sideeffects: false` set in package.json. But inserting the Buffer in the first example is unexpected.

Rewriting the `react-dom` example to the following works as expected
```javascript
const fireAsAct = (cb: () => void) => {
  if (process.env.NODE_ENV === "test") {
    const { act } = require("react-dom/test-utils");
    act(() => cb());
  } else {
    cb();
  }
};

fireAsAct(() => console.log('test'));
```

## 🌍 Your Environment

| Parcel           | 2.0.0-nightly.226
| Node             | v12.8.1
| npm         | v6.14.4
| Operating System | Windows (WSL v1/ Ubuntu Bash)

garthenweb:
# ↪️ Pull Request

This tries to fix the second part of #4524.

It looks like dependencies that are only used in a dead branch after the process visitor was executed are still included in the bundle.

By removing the dead branches directly this should be solved.

Closes #4524

@mischnic I am not 100% sure if this is the solution to the issue. First, the test is not passing. Not sure if I just took the wrong approach to test it.

But also, when running parcel with or without the removed dead branches on the example in the test, it results in the same code: It never contains code for dependency `a.js` (which is the wanted behavior anyhow).

But why is the test failing? Did I miss something? Do we need to have another approach? And is this an issue at all? Maybe you have some ideas on that :)

## 💻 Examples

``` javascript
import { act } from 'react-dom/test-utils'
const fireAsAct = (cb: () => void) => {
  if (process.env.NODE_ENV === "test") {
    act(() => cb());
  } else {
    cb();
  }
};

fireAsAct(() => console.log('test'));
```

## ✔️ PR Todo

- [ ] Find out why the tests are not passing for the dead code check
- [ ] Discuss if this is actually the solution to the original issue
- [ ] Maybe find some better solutions to remove the branches


packages/core/integration-tests/test/javascript.js
```diff
@@ -1333,6 +1333,36 @@ describe('javascript', function() {
...
+      path.join(__dirname, `/integration/dead-branch-elimination/deps.js`),
+      {
+        scopeHoist: true,
+      },
+    );
+
+    let output = await run(b);
+    assertBundles(b, [
+      {
+        assets: ['deps.js'],
```
garthenweb, line 1360: This is failing as it always contains `a.js`. Not sure how to fix that.

packages/transformers/js/src/visitors/process.js
```diff
@@ -30,6 +32,8 @@ export default {
       } else {
         morph(node, types.booleanLiteral(true));
       }
+      deleteClosestFalsyBranch(ancestors);
+      asset.setAST(ast);
```
garthenweb, line 36: Not 100% sure this should be added here as I am not 100% aware what it does. But it looked logically to not just have it for `process.env` but also for `process.browser`.

packages/transformers/js/src/visitors/utils.js
```diff
@@ -70,6 +72,40 @@ export function isInFalsyBranch(ancestors: Array<Node>) {
...
+      isClosestTestNode ||
+      isClosestUnaryExpression ||
+      isClosestBinaryExpression
+    ) {
+      const res = evaluateExpression(conditionNode.test);
+      if (res && res.confident) {
+        const index = ancestors.indexOf(conditionNode);
+        ancestors[index] = conditionNode.alternate;
+        if (res.value) {
+          Object.assign(conditionNode, conditionNode.consequent);
```
garthenweb, line 100: This was the best I came up with to mutate the node/ remove the dead branch. Are there better ways? Feels dirty to me.

mischnic:
> But why is the test failing? Did I miss something? Do we need to have another approach? And is this an issue at all? Maybe you have some ideas on that :)

Yeah, it's actually a bit more complicated. This basic idea is:

-

mischnic:
> But why is the test failing? Did I miss something? Do we need to have another approach? And is this an issue at all? Maybe you have some ideas on that :)

Yeah, it's actually a bit more complicated:

- remove dead branches completely so that the imported binding referenced at all (your current changes)
- In the `hoist.js` visitor, only add import declarations that are actually referenced to `dep.symbols` (the reason for the previous step is that this is now a very cheap operation).
The non-scope-hoisting `dependencies.js` visitor has no symbol handling at all, so doing this for `--no-scope-hoist` and watch mode would be a bit more work.
- the transformer doesn't know if a dependency has `sideEffects: false`, so in `@parcel/core` we can exclude side-effect-free dependencies that have no used symbols (similar to https://github.com/parcel-bundler/parcel/blob/f88eb6822d76002c9c900ab7ec50154f8490688f/packages/core/core/src/requests/DepPathRequestRunner.js#L63) - it might also work out of the box with the current `deferring` mechanism

packages/transformers/js/src/visitors/utils.js
mischnic: There isn't, but you should use the `morph` function defined above, maybe that makes you feel better 😄

devongovett:
I think this should be handled in the Rust rewrite. We use SWC's branch optimizer which should completely remove dead branches, and this is run prior to dependency collection and process/global insertion.

PR #4542, opened 2020-04-29
Fix issue so that typescript-validator can report results across multiple files.

### LINKED ISSUE #4488: @parcel/validator-typescript only reports errors from one file
# 🐛 bug report

`@parcel/validator-typescript` will only ever report errors from a single file. If multiple typescript files contain semantic errors, it will only report problems from one of them.

***.parcelrc:***
```js
{
    "validators": {
    }
}
```
***package.json:***
```
{
  "version": "1.0.0",
  "main": "index.html",
  "private": true,
  "scripts": {
  },
  "devDependencies": {
    "@parcel/validator-typescript": "2.0.0-nightly.225",
    "parcel": "2.0.0-nightly.221",
    "typescript": "3.7.5"
  }
}
```

***src/index.ts:***
```ts
import { message2 } from "./message2";
const message: number = "My Message!"; // This is a type error!
export const output = message2 + message;
```

***src/message2.ts***
```ts
export const message2: string = 123; // This is also a type error!
}
```
## 🤔 Expected Behavior

Parcel should show diagnostics for both type errors.

## 😯 Current Behavior

Parcel only shows diagnostics for one of the errors.

## 💁 Possible Solution

This code within `Validation` will throw a `ThrowableDiagnostic` error as soon as the first `ValidatorResult` that contains an error is returned from any validator plugin:

https://github.com/parcel-bundler/parcel/blob/c8d127623e419f23b6216d3fc68e1d45adeecbd0/packages/core/core/src/Validation.js#L164-L177

Although I refactored this code in #4237, I also was able to repro the issue with an earlier build (`2.0.0-nightly.115`), so I beleive this was a pre-existing problem.

One way to fix the error would be to wait until all validator plugins returned, and then throw a single error with all the diagnostics.

## 🔦 Context

## 💻 Code Sample

See the `mutli-file-errors` branch of [this repo](https://github.com/astegmaier/parcel2-typescript-validator-bug/tree/mutli-file-errors) for a demonstration.

## 🌍 Your Environment

| Parcel           | 2.0.0-nightly.221
| Node             | 10.6.3
| Yarn         | 1.21.1

astegmaier:
# ↪️ Pull Request

This fixes #4488 and allows `@parcel/validator-typescript` to report semantic diagnostic errors across multiple files. It should also address the problem for any future validators that use the `validateAll` API.

PR #4521 also contained the same fix, but I broke this into a separate PR because after talking to @padmaia about #4521, it looks like that one will take some more thought before it's ready. 

## 💻 Examples

Create a simple project with `@parcel/validator-typescript` enabled where there are two typescript files, each of which contains a type error:

***.parcelrc:***
```js
{
    "validators": {
    }
}
```

***src/index.ts:***
```ts
import { message2 } from "./message2";
const message: number = "My Message!"; // This is a type error!
export const output = message2 + message;
```

***src/message2.ts***
```ts
export const message2: string = 123; // This is also a type error!
}
```
Without this fix, only one error will be reported. With this fix, both errors are reported.

## 🚨 Test instructions

I modified the existing single-file unit test to include a second file (and make sure that diagnostics are emitted for both files). You can also test manually using the `mutli-file-errors` branch of [this repo](https://github.com/astegmaier/parcel2-typescript-validator-bug/tree/mutli-file-errors).

## ✔️ PR Todo


astegmaier:
@mischnic - thanks for taking a look at this! Was there someone else who you wanted to review before merging?

PR #4545, opened 2020-04-29
Import asset from reachable bundle in ESM output format

devongovett:
Previously, we would import from the first bundle containing the asset when it wasn't in the current bundle in scope hoisting. This was incorrect because it could have been from a different part of the site completely rather than a bundle that was already loaded in the current bundle's ancestry.

This PR finds a reachable bundle containing the asset to import from the current bundle. This includes sibling bundles as well as ancestors and siblings of ancestors. It follows a similar algorithm to the one I added in #4295, but returns the first bundle it finds containing the asset rather than checking if all of the ancestors contain it. This is verified by ensuring that any ESM import is already loaded by the parent HTML file.

### DIFF
--- a/packages/core/core/src/public/BundleGraph.js
+++ b/packages/core/core/src/public/BundleGraph.js
@@ -119,19 +119,15 @@ export default class BundleGraph implements IBundleGraph {
   }
 
   isAssetReachableFromBundle(asset: IAsset, bundle: IBundle): boolean {
-    let internalNode = this.#graph._graph.getNode(bundle.id);
-    invariant(internalNode != null && internalNode.type === 'bundle');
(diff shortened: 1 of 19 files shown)


packages/core/core/src/public/BundleGraph.js
```diff
@@ -127,6 +127,21 @@ export default class BundleGraph implements IBundleGraph {
     );
   }
 
+  findReachableBundleWithAsset(bundle: IBundle, asset: IAsset): ?IBundle {
+    let internalNode = this.#graph._graph.getNode(bundle.id);
```
wbinnssmith, line 131: `bundleToInternalBundle`?

Issue #4616, opened 2020-05-16
Require.cache causes errors


m-allanson:
# 🐛 bug report

Parcel does not transform `require.cache` when bundling, which cause an error when running the bundled file.

## 🎛 Configuration (.babelrc, package.json, cli command)

```json5
// in package.json
"esm": "dist/index.js",
"targets": {
"esm": {
  "includeNodeModules": true,
  "outputFormat": "esmodule",
  "isLibrary": true
}
},
```

The bundle command is `parcel build index.js --target esm --no-minify`

## 🤔 Expected Behavior

The bundled output should run without error.

## 😯 Current Behavior

Running the bundled file will cause the following error:

```
file:///<path>/parcel-repro/dist/index.js:4
for (const file in require.cache) {
                           ^

ReferenceError: require is not defined
    at file:///<path>/parcel-repro/dist/index.js:4:28
    at ModuleJob.run (internal/modules/esm/module_job.js:110:37)
    at async Loader.import (internal/modules/esm/loader.js:179:24)
```

## 💁 Possible Solution

Parcel 1 created a shim for `require.cache`. See this earlier issue and PR:

- https://github.com/parcel-bundler/parcel/issues/1959
- https://github.com/parcel-bundler/parcel/pull/1960

In the v2 branch there is [a skipped test for the shim](https://github.com/parcel-bundler/parcel/blob/9f71164b910bb7fef050d57d8ef9036f3aa17af6/packages/core/integration-tests/test/javascript.js#L2061). 

## 🔦 Context

I want to package `node_modules` dependencies into my bundle. Some of these dependencies make use of `require.cache`.

## 💻 Code Sample

Repro here: https://github.com/m-allanson/parcel-require-repro

The `dist` directory is committed too. 

Example code that will cause the error: 

```js
// in src file index.js
for (const file in require.cache) {
  if (/color/g.test(file)) {
    result[file] = true;
  }
}
```

## 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Parcel           | 2.0.0-alpha.3.2
| Node             | 12 or 14 (for ESM support)
| npm/Yarn         | npm
| Operating System | macOS

mischnic:
Needs to be added here:
https://github.com/parcel-bundler/parcel/blob/9f71164b910bb7fef050d57d8ef9036f3aa17af6/packages/shared/scope-hoisting/src/hoist.js#L266-L280

m-allanson:
[...]

PR #4621, opened 2020-05-18
Stub `require.cache` with an empty object

m-allanson:
# ↪️ Pull Request

Closes https://github.com/parcel-bundler/parcel/issues/4616

I wonder if more tests are needed here? e.g. verify that `require.cache` is _not_ stubbed when the output format is cjs. Guidance welcome :)
## 🚨 Test instructions

`yarn test:integration`

## ✔️ PR Todo

### DIFF
--- a/packages/shared/scope-hoisting/src/hoist.js
+++ b/packages/shared/scope-hoisting/src/hoist.js
@@ -276,7 +276,7 @@ const VISITOR: Visitor<MutableAsset> = {
     }
 
     if (t.matchesPattern(path.node, 'require.cache') && !asset.env.isNode()) {
-      path.replaceWith(t.identifier('{}'));
+      path.replaceWith(t.objectExpression([]));
     }
 
     if (t.matchesPattern(path.node, 'module.bundle')) {
--- a/packages/shared/scope-hoisting/src/hoist.js
+++ b/packages/shared/scope-hoisting/src/hoist.js
@@ -275,6 +275,10 @@ const VISITOR: Visitor<MutableAsset> = {
       path.replaceWith(t.identifier('null'));
     }
 
+    if (t.matchesPattern(path.node, 'require.cache') && !asset.env.isNode()) {
+      path.replaceWith(t.identifier('{}'));
+    }
+
     if (t.matchesPattern(path.node, 'module.bundle')) {
       path.replaceWith(t.identifier('parcelRequire'));
     }
--- a/packages/core/integration-tests/test/integration/formats/commonjs/require-cache.js
+++ b/packages/core/integration-tests/test/integration/formats/commonjs/require-cache.js
@@ -1,0 +1,1 @@
+module.exports = require.cache;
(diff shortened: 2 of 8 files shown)


packages/shared/scope-hoisting/src/hoist.js
```diff
@@ -275,6 +275,10 @@ const VISITOR: Visitor<MutableAsset> = {
       path.replaceWith(t.identifier('null'));
     }
 
+    if (t.matchesPattern(path.node, 'require.cache') && !asset.env.isNode()) {
+      path.replaceWith(t.identifier('{}'));
```
mischnic, line 279:
```suggestion
      path.replaceWith(t.objectExpression([]));
```

mischnic:
> I wonder if more tests are needed here? e.g. verify that require.cache is not stubbed when the output format is cjs. Guidance welcome :)

You're right. I didn't think of that. (these tests should go into `formats/{esmodule,commonjs}`)

mischnic:
> cjs 

There is however the problem that the modules bundled by Parcel won't be listed in `require.cache` but there's nothing we can do there

packages/shared/scope-hoisting/src/hoist.js
m-allanson: 👍 thanks

m-allanson:
I've added a few tests here, one of them is failing. 

```
  output formats
    commonjs
      1) should support commonjs output with require.cache
    esmodule
      ✓ should support esmodule output (stub require.cache) (91ms)
    global
      ✓ should support global module output (stub require.cache) (84ms)

  scope hoisting
    commonjs
      ✓ should stub require.cache (84ms)


  3 passing (2s)
  1 failing

  1) output formats
       commonjs
         should support commonjs output with require.cache:

      AssertionError [ERR_ASSERTION]: false == true
      + expected - actual

      -false
      +true

      at Context.<anonymous> (test/output-formats.js:52:7)
```

If I understand correctly `require.cache` should _not_ be stubbed when the output format is commonjs. The failing test shows that this this is not happening. `require.cache` is being stubbed for all output formats. Is there a way to check the output format before applying the stub? Or is there a better way to handle this?

mischnic:
> Is there a way to check the output format before applying the stub? 

That doesn't seem to be possible in hoist.js (because with multiple output formats, that file is only executed once).
So we need to move this to the MemberExpression visitor in `scope-hoisting/link.js`.
The same applies to the `module.require` replacement as well (the `isNode` check should also be for commonjs instead).

m-allanson:
> So we need to move this to the MemberExpression visitor in scope-hoisting/link.js. The same applies to the module.require replacement as well

👍 I'll dig into it and update the PR. Thanks for bearing with me :)

m-allanson:
I have spent as long as I can on this, for now. Unfortunately I haven't made any further progress. 

I think I've missed something obvious, but I couldn't see how to inspect `scope-hoisting/link.js` from the integration or unit tests. 

If I add `console.log()`s they don't seem to output, and any breakpoints in that file don't get hit. Are there some other tests, or some other way to examine what's happening in that file at runtime? 

I'm going to close this for now. If I make any further progress I will re-open it :)

mischnic:
I always do `console.log` and then running parcel like this `parcel build ... | cat` to make sure they are not overwritten

PR #4638, opened 2020-05-21
Fix #4429: dev server entry detection

### LINKED ISSUE #4429: 404 Page not found when serving arbitrary file
# 🐛 bug report

I'm trying to use Parcel 2 in no-config mode, for a very simple use case, serving a file named `test.html`. That fails with a 404 Page not found in the browser. There is no error shown anywhere.

node_modules/.bin/parcel serve test.html

## 💁 Possible Solution

Rename the file to `index.html`.

## 🔦 Context

I'm trying to serve different HTML files.

Must I create targets in `package.json`? I've read the README on that and it's not quite clear. For example, what does [dist/browser/index.js](https://github.com/parcel-bundler/parcel/blob/v2/packages/core/parcel/README.md#target-configuration) represent? The [Sources](https://github.com/parcel-bundler/parcel/blob/v2/packages/core/parcel/README.md#sources) section says at `1.` that specifying the file on the CLI should work.

## 🌍 Your Environment

| Node             | v13.11.0
| npm         | 6.14.3
| Operating System | Linux

moeriki:
# ↪️ Pull Request

Fixes #4429.

## 💻 Examples

Before the entry detection served the first `index.html` bundle it found.

This would yield a 404 when running any other html file.

```
parcel serve test.html
```

I updated the logic, as suggested, to serve any single HTML entry.

## 🚨 Test instructions

I didn't see any tests for the dev server. Let me know if or where I can do so.

## ✔️ PR Todo

### DIFF
--- a/packages/reporters/dev-server/src/Server.js
+++ b/packages/reporters/dev-server/src/Server.js
@@ -142,25 +142,17 @@ export default class Server extends EventEmitter {
 
   sendIndex(req: Request, res: Response) {
     if (this.bundleGraph) {
-      // If the main asset is an HTML file, serve it
-      let htmlBundle = this.bundleGraph.traverseBundles(
-        (bundle, context, {stop}) => {
-          if (bundle.type !== 'html' || !bundle.isEntry) return;
-
-          if (!context) {
-            context = bundle;
-          }
-
-          if (
-            context &&
+      // If there is exactly one entry HTML file, serve it
+      const htmlBundles = this.bundleGraph
+        .getBundles()
+        .filter(
+          bundle =>
+            bundle.type === 'html' &&
+            bundle.isEntry &&
             bundle.filePath &&
-            bundle.filePath.endsWith('index.html')
-          ) {
-            stop();
-            return bundle;
-          }
-        },
-      );
+            bundle.filePath.endsWith('.html'),
+        );
+      const htmlBundle = htmlBundles.length === 1 ? htmlBundles[0] : undefined;
 
       if (htmlBundle) {
         req.url = `/${path.relative(
--- a/packages/reporters/dev-server/src/templates/404.html
+++ b/packages/reporters/dev-server/src/templates/404.html
@@ -10,7 +10,7 @@
         margin: 0;
         font-family: sans-serif;
       }
-      
+
       .error-title {
         color: #282c33;
         font-size: 3rem;
@@ -25,10 +25,16 @@
         padding: 0 25px;
         margin: 0;
       }
+
+      .long-desc {
+        color: #282c33;
+        padding: 0 25px;
+      }
     </style>
   </head>
   <body>
     <h1 class="error-title">404</h1>
     <h2 class="short-desc">Page not found.</h2>
+    <p class="long-desc">You may have zero or multiple HTML entries.</p>
   </body>
 </html>


Moeriki:
I wasn't sure on how to use the `bundleGraph`. If there is a better way than to use `getBundles()` I will gladly update.

packages/reporters/dev-server/src/Server.js
```diff
@@ -142,25 +142,17 @@ export default class Server extends EventEmitter {
...
-            bundle.filePath.endsWith('index.html')
-          ) {
-            stop();
-            return bundle;
-          }
-        },
-      );
+            bundle.filePath.endsWith('.html'),
+        );
+      const htmlBundle = htmlBundles.length === 1 ? htmlBundles[0] : undefined;
```
DeMoorJasper, line 155: Would be nice to get a test for this.

packages/reporters/dev-server/src/Server.js
```diff
@@ -142,25 +142,17 @@ export default class Server extends EventEmitter {
...
+            bundle.isEntry &&
             bundle.filePath &&
-            bundle.filePath.endsWith('index.html')
-          ) {
-            stop();
-            return bundle;
-          }
-        },
-      );
+            bundle.filePath.endsWith('.html'),
```
mischnic, line 153:
```suggestion
            bundle.type === '.html',
```

Moeriki:
I will check all your comments and add a few tests.

But now I'm looking at existing tests I broke.

It seems all html bundles have `isEntry: true` although the tests in question run with a single entry filepath: `bundler('/integration/html/index.html')`. Would it be possible there is something else going wrong there?

mischnic:
> It seems all html bundles have isEntry: true although the tests in question run with a single entry filepath

isEntry mostly means: no hash.

For example:
https://parcel2-repl.now.sh/#JTdCJTIyY3VycmVudFByZXNldCUyMiUzQSUyMkphdmFzY3JpcHQlMjIlMkMlMjJvcHRpb25zJTIyJTNBJTdCJTIybWluaWZ5JTIyJTNBZmFsc2UlMkMlMjJzY29wZUhvaXN0JTIyJTNBdHJ1ZSUyQyUyMnNvdXJjZU1hcHMlMjIlM0FmYWxzZSUyQyUyMnB1YmxpY1VybCUyMiUzQSUyMiUyMiUyQyUyMnRhcmdldFR5cGUlMjIlM0ElMjJicm93c2VycyUyMiUyQyUyMnRhcmdldEVudiUyMiUzQW51bGwlMkMlMjJvdXRwdXRGb3JtYXQlMjIlM0FudWxsJTJDJTIycmVuZGVyR3JhcGhzJTIyJTNBdHJ1ZSUyQyUyMnZpZXdTb3VyY2VtYXBzJTIyJTNBZmFsc2UlN0QlMkMlMjJhc3NldHMlMjIlM0ElNUIlNUIlMjJhLmh0bWwlMjIlMkMlMjIlM0NoMSUzRUhlbGxvJTNDJTJGaDElM0UlNUNuJTNDYSUyMGhyZWYlM0QlNUMlMjJjLmh0bWwlNUMlMjIlM0VMaW5rJTNDJTJGYSUzRSUyMiUyQzElNUQlMkMlNUIlMjJiLmh0bWwlMjIlMkMlMjIlM0NoMSUzRWIlM0MlMkZoMSUzRSUyMiUyQzElNUQlMkMlNUIlMjJjLmh0bWwlMjIlMkMlMjJjJTIyJTVEJTVEJTdE

generates this bundle graph (a & b are "CLI entries", c is merely linked from a): 
parcel2-repl now sh_

Moeriki:
I see.

> if there are 2 or more html bundles: 404 (because of the ambiguity, maybe the error page should state this as well)

Per suggest I updated logic from "serve first index.html" to serve any single HTML file.

The problem is there are two tests failing with multiple HTML bundles where it serves a 404 now instead of the first `index.html` as before.

https://github.com/parcel-bundler/parcel/blob/v2/packages/core/integration-tests/test/server.js#L135
https://github.com/parcel-bundler/parcel/blob/v2/packages/core/integration-tests/test/server.js#L391

Both use the fixtures at `integration/html/` which contain both `index.html` and `other.html`. I hoped `isEntry` would filter out the other, but it doesn't. Not sure how to filter this correctly.

PR #4703, opened 2020-06-06
set postcss transformer as default for pcss files

smeijer:
# ↪️ Pull Request

I've updated the default config for `Parcel v2`, to transform `*.pcss` files with `postcss`.

## 💻 Examples

While moving [testing-playground](https://github.com/smeijer/testing-playground) to parcel v2, I've noticed that postcss files were no longer transformed by default.

To fix that issue, I added a `.parcelrc` file to my project root:

```json
{
  "transformers": {
    "*.pcss": ["@parcel/transformer-postcss"],
    "*.css": ["@parcel/transformer-postcss"],
  },
}
```

This small change results in supporting the `.pcss` extension again, and make the `.parcelrc` above unnecessary. Simplifying the migration from parcel v1 to parcel v2.


mischnic:
That doesn't quite work, this is needed:

```diff
diff --git packages/configs/default/index.json packages/configs/default/index.json
index 2f734f5b..fad73284 100644
--- packages/configs/default/index.json
+++ packages/configs/default/index.json
@@ -18,7 +18,7 @@
     "*.{styl,stylus}": ["@parcel/transformer-stylus"],
     "*.{sass,scss}": ["@parcel/transformer-sass"],
     "*.less": ["@parcel/transformer-less"],
-    "*.css": ["@parcel/transformer-postcss", "@parcel/transformer-css"],
+    "*.{css,pcss}": ["@parcel/transformer-postcss", "@parcel/transformer-css"],
     "*.sss": ["@parcel/transformer-sugarss"],
     "*.{htm,html}": [
       "@parcel/transformer-posthtml",
diff --git packages/transformers/postcss/src/PostCSSTransformer.js packages/transformers/postcss/src/PostCSSTransformer.js
index 0292b184..dc8fa398 100644
--- packages/transformers/postcss/src/PostCSSTransformer.js
+++ packages/transformers/postcss/src/PostCSSTransformer.js
@@ -48,6 +48,7 @@ export default new Transformer({
   },

   async transform({asset, config, options, resolve}) {
+    asset.type = 'css';
     if (!config) {
       return [asset];
     }
```

Otherwise, a `foo.pcss` bundle is generated because the type of the `pcss` assets is never set to `css`.

smeijer:
Thanks! That explains why I still needed a "proxy file". My `styles.pcss` was indeed being parsed, but the browsers didn't like the `.pcss` file. 

To workaround that issue, I simply imported all my `.pcss` files in an `app.css`, which was working.

I've pushed the change.

mrassili:
@smeijer Thanks for making those changes, do you have an idea why parcel can't resolve styles from the bundle `import style from "bundle-text:./style.css";`
it gives `Cannot resolve dependency 'bundle-text:./style.css'`

mischnic:
@mrassili This has nothing to do with this PR.
Are you actually using Parcel 2? Please open a new issue

mrassili:
yes sorry will do that :)

PR #4670, opened 2020-05-27
Make isAssetReachableFromBundle only look in preceding siblings

mischnic:
# ↪️ Pull Request

Close T-475

PR #4735, opened 2020-06-14
Fix issue with circular bundles

### LINKED ISSUE #4667: Importing entire @material-ui/core creates a bundle with a TypeError using 2.0.0-alpha.3.2
# 🐛 bug report

When importing this library with `* as mui` and then using `mui.Paper` builds a bundle with a TypeError
```typescript
import * as mui from '@material-ui/core'
ReactDOM.render(<mui.Paper>Hello</mui.Paper>, root)
```

Interestingly, it does _not_ happen if the import is changed to: 

```typescript
import {Paper} from '@material-ui/core'`
```

There is a repo to reproduce here:
https://github.com/danr/parcel-import-bug

## 🤔 Expected Behavior

The bundle should not raise an exception.

## 😯 Current Behavior

```
Uncaught TypeError: $f is not a function
    at Ny (withStyles.js:6)
    at Paper.js:109
    at main.tsx:10
```

## 🌍 Your Environment

| Node             | 14.2.0
| npm/Yarn         | yarn 1.22.4

It also happens with 2.0.0-nightly.275

devongovett:
Closes T-475. Closes #4670.

This fixes an issue where there is both a synchronous and asynchronous dependency on the same asset from the same bundle. Previously, step 2 in the default bundler would remove the asset from the bundle and add the async bundle to the same bundle group. The problem with this is that it can introduce a circular dependency, which won't work correctly when split into separate bundles.

This PR skips reusing separate bundles for async dependencies when a synchronous dependency is also present in a reachable parent bundle. This will be internalized into the parent bundle, so the async dependency is basically a promise wrapper around the asset in the same bundle. If the async bundle isn't used anywhere else, it will be deleted, and the asset will only exist in the bundle with the sync dependency.

In addition, there was an issue with this internalized asset where the dependency would resolve to the original asset rather than the wrapped one. This was due to the wrapper asset also having a dependency on the original (in order to wrap it), and that was getting included in the bundle when merging with the runtimes graph.

There's another issue that I wasn't able to solve which is that because of this wrapper asset we end up building the entire sub-graph of the wrapped asset again when building the runtimes graph. The asset graphs aren't shared, so we cannot really avoid this. Ideally, we'd have a way to add a dependency to an existing asset id without rebuilding it. To discuss separately.


mischnic:
Doesn't seem to be related to #4667?

devongovett:
Oops. Meant #4670.

PR #4736, opened 2020-06-14
Added fallback port to environment variable $PORT

notsofunctional:
# ↪️ Pull Request

Reading the documentation it shows something like this:
Imgur
But trying it on a project i'm working on and viewed your code, you never use the $PORT variable, so it always fallback in the port 1234.

## 💻 Examples

`PORT=3000 parcel serve [...rest]`
It is always 1234.

## ✔️ PR Todo


mischnic:
> Reading the documentation it shows something like this:

Technically, these are the Parcel 1 docs. But given that this is somewhat of a convention and Parcel 1 supported this as well, I don't see why not.

Issue #4695, opened 2020-06-05
V2: JSRuntime should reference child bundle dependencies with a require call when using commonjs


jfrconley:
# 🙋 feature request

JSRuntime should reference non-async child bundle dependencies with a require call when using commonjs, instead of falling back to the default url runtime.  This would allow for the bundler to split nodejs applications into multiple bundles.

## 🤔 Expected Behavior

When the bundler generates child bundles for a commonjs application, the JSRuntime should reference them using a require call.

## 😯 Current Behavior

No specific behaviour is in place for handling synchronous child bundles in commonjs so the runtime falls back to a url export.

## 💁 Possible Solution

Include something like this in the JSRuntime
```js
      if (bundle.env.outputFormat === 'commonjs') {
        assets.push({
          filePath: __filename,
          dependency,
          code: `module.exports = require("./" + ${getRelativePathExpr(bundle, mainBundle)})`
        });
        continue;
      }
```

## 🔦 Context

I am creating a custom bundler that allows me to split nodejs bundles at given files. This was fairly simple and generates correct graphs, but the runtime doesn't reference the child bundles correctly

mischnic:
Makes sense

> I am creating a custom bundler that allows me to split nodejs bundles at given files.

Could you share that so that we can add a test case as well when fixing this?

jfrconley:
Sure! https://github.com/jfrconley/parcel-bundler-splitable
All I did was add a package.json, "bundleSplits" (name up for debate) that is an array of files to split the bundle on. All together a less then 10 line change.

mischnic:
In the short term, https://github.com/jfrconley/parcel-runtime-splitable/commit/e8550941939c6d4fc36f2fa50d81da9a106b3a01 could definitely be upstreamed as a bugfix (with a stripped down version of your bundle in the test case).

> what the official configuration method for it would be

> project philosophy

We plan on having these options
https://github.com/parcel-bundler/parcel/blob/e95fbaeac705c170719f771580e502c1f7f5865a/packages/bundlers/default/src/DefaultBundler.js#L15-L19
configureable by the user (probably via some config file), but that's not decided yet.

PR #4744, opened 2020-06-16
Fix: #4695 Reference CommonJS child bundles with a require call

jfrconley:
# ↪️ Pull Request

This pull request adds support to JSRuntime for referencing CommonJS child bundles with a require call. No handling of this was present, so the current behavior was to use a URL reference.

Resolves: #4695

## 💻 Examples

There isn't an easy way to see this in action without a custom bundler. I created a simple one (https://github.com/jfrconley/parcel-bundler-splitable) that allows listing files you want to split the bundle on. The integration test relies on a custom bundler with a hardcoded path to split on. 

## 🚨 Test instructions

Testing is covered by added integration test: commonjs-bundle-require

## ✔️ PR Todo


jfrconley:
@mischnic Is there a timeline for when this can be merged? Keeping a custom runtime plugin up to date is causing some issues for us.

mischnic:
I'll take a look.
Could you slim down the bundler plugin you added in the test so that it only includes the bare minimum to make the test work? For example the `optimize` step likely doesn't do anything here

jfrconley:
I removed optimize and deduplicate. Additionally, I've resolved merge conflicts and pulled in upstream changes

jfrconley:
@mischnic Is there any other cleanup I need to do before this can be merged?

packages/core/integration-tests/test/integration/commonjs-bundle-require/package.json
```diff
@@ -0,0 +1,18 @@
...
+  "engines": {
+    "node": "8"
+  },
+  "targets": {
+    "main": {
+      "engines": {
+        "node": "8"
+      }
+    }
+  }
```
mischnic, line 17:
```suggestion
  }
```

packages/core/integration-tests/test/integration/commonjs-bundle-require/node_modules/parcel-bundler-splitable/package.json
```diff
@@ -0,0 +1,4 @@
+{
+  "name": "parcel-bundler-splitable",
+  "version": "1.2.3"
```
mischnic, line 3:
```suggestion
  "version": "1.2.3",
  "engines": {
    "parcel": "^2.0.0-alpha.1"
  }
```

mischnic:
I would have suggested support ESM as well while we're at it, but that doesn't work yet for other reasons...

Issue #2622, opened 2019-02-07
It doesn't resolve image file pointed in src attribute of <amp-img> elements


honey32:
# 🐛 bug report
The entry point is a html file.
The bundler doen't seem to detect the resource file which is specified in `<amp-img>` tag's `src` attribute.
The path in output file is the same as the input file.
```html
<amp-img src="/path/to/image.jpg">
```
## 🎛 Configuration (.babelrc, package.json, cli command)

`parcel index.html`

## 🤔 Expected Behavior

It should work as same as it does for `<img>`  tag.
```html
<amp-img src="/image.(hash goes here).jpg">
```

## 😯 Current Behavior

It doesn't.

## 💁 Possible Solution

Furthurmore, I think there should be **no limitation** of the tag name (or **black-list** instead of current black-list) on checking `src` attribute, because there are plenty types of amp elements using media file (See: [Components - Amp](https://www.ampproject.org/docs/reference/components)).

## 🔦 Context

Parcel is now not friendly with AMP.

## 💻 Code Sample
[https://github.com/honey32/snippet-of-parcel-bug](https://github.com/honey32/snippet-of-parcel-bug)

## 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Parcel           | 1.11.0
| Node             | 8.9.3
| npm/Yarn         | 5.5.1
| Operating System | Windows 10

BegoConstantino:
Any update to this? I have the same issue

honey32:
sorry, i'm not sure. I'm not dealing with a project related to this problem.

2020年6月12日(金) 16:02 BegoConstantino <notifications@github.com>:

BegoConstantino:
I fixed it!
Just need to add 'amp-img' in the array of src to watch. File:
\node_modules\parcel-bundler\src\assets\HTMLAsset.js

honey32:
thank you!
i appreciate your work.

2020年6月17日(水) 16:01 BegoConstantino <notifications@github.com>:

honey32:
Maybe the file below is the same as one you mentioned.
[parcel/packages/transformers/html/src/dependencies.js](https://github.com/parcel-bundler/parcel/blob/v2/packages/transformers/html/src/dependencies.js)
However, i'm not sure if 'parcel' should support AMP.

BegoConstantino:
It seems like the same file (in my installed parcel has more const defined at the begining)

[...]

PR #4745, opened 2020-06-17
Adding support for <amp-img src=”path/to/img.jpg”>

honey32:
I enabled finding dependency of src attribute in <amp-img> tag.
Otherwise, that tag won't work.

I've made an issue [here(#2622)](https://github.com/parcel-bundler/parcel/issues/2622).

# ↪️ Pull Request

If I write as below
```html
<amp-img src="/path/to/image.jpg">
```
the path to image didn't seem to be recognized as a dependency.
So, I added a portion of code to treat `src` attribute of `<amp-img>` tag in the same way as one of `` tag is treated.

## 💻 Examples

https://github.com/honey32/snippet-of-parcel-bug

## ✔️ PR Todo


honey32:
You mean I should do one of these ?
- add an example under packages/examples/
- modify html example in that directory to include `<amp-img>`

DeMoorJasper:
@honey32 no I mean an integration test

You can find the integration tests for html here: https://github.com/parcel-bundler/parcel/blob/v2/packages/core/integration-tests/test/html.js

honey32:
I'm sorry.
I'm not familiar with testing and it seems too difficult for now, so could you or anyone do that please?
If not, I'll manage to do this for a week.

DeMoorJasper:
@honey32 sure I'll approve the PR for now and might add a test later, would've just been better to prevent this from breaking in the future

DeMoorJasper:
@honey32 sure I'll approve the PR for now and might add a test later, would've just been better to prevent this from breaking in the future

honey32:
Thank you very much.

PR #4786, opened 2020-06-22
Add scripts and styles referenced from HTML to the same bundle group

### LINKED ISSUE #4302: Inlining shared, small modules can yield non-standard behavior
# 🐛 bug report

If you have 2 `<script>` tags that share a dependency, that dependency will get chunked out if it is above a certain size (is this size threshold documented somewhere?).

However, this can yield non-standard behavior compared to using normal ES modules. For example, exporting an object, all importers will get a reference to the _same_ object. So one function change the object, other importers should see those changes.

Please see [repro repo](https://github.com/surma-dump/parcel-shared-objects)

## 🤔 Expected Behavior

`obj.js` should land in its own chunk.

## 😯 Current Behavior

`obj.js` is inlined into both `component-1.js` and `component-2.js`.

## 💁 Possible Solution

See “Expected Behavior”.

## 🔦 Context

E.g: A module exporting a global config object that gets augmented by other modules.

## 💻 Code Sample

Please see [repro repo](https://github.com/surma-dump/parcel-shared-objects)
## 🌍 Your Environment

| Node             | v12.8.0
| npm/Yarn         | 6.12.1
| Operating System | MacOS 10.15.3 (19D76)

### LINKED ISSUE #4304: Generated script tags don’t have `defer`
# 🐛 bug report

I have 2 entry points, `index.html` and `profile.html`, each of loading their own specific piece of JavaScript, `index.js` and `profile.js`. Both these JS files depend on a common dependency `utils.js`. If `utils.js` is big enough to become its own chunk, the resulting HTML contains an additional `<script>` tag for the common dependencies. 

Please see [repro repo](https://github.com/surma-dump/parcel-js-entry-cascade)

## 🤔 Expected Behavior

The `<script>` tag for the common dependency should have `defer` if the original `<script>` tag has a `defer` attribute. Same for `async`.

## 😯 Current Behavior

The `<script>` tag for the common dependency does not have any attributes and is therefore renderblocking.

## 🔦 Context

Normal web development.

## 💻 Code Sample

Please see [repro repo](https://github.com/surma-dump/parcel-js-entry-cascade)

## 🌍 Your Environment

| Node             | v12.8.0
| npm/Yarn         | 6.12.1
| Operating System | MacOS 10.15.3 (19D76)

devongovett:
Fixes #4302.

This is a refactor to the way the bundler works for scripts and styles referenced from HTML, along with inline bundles. Previously, each script/style would create separate a bundle group. This meant that assets were unable to be deduplicated between multiple scripts/styles on the same page, leading to duplication and even incorrect runtime output (e.g. #4302). In addition, sibling bundles (e.g. CSS imported from JS) would not be deduplicated across multiple bundles.

This changes the behavior so that scripts and styles are placed in the same bundle group as the HTML itself. This is more correct since they JS/CSS is expected to load together with the HTML, and it allows Parcel to deduplicate things between multiple script/styles.

In the future, we could also decide to combine multiple `<link rel="stylesheet">` and `<script>` tags together into one. To discuss. For now, they are kept separate (treated as a code split point).

In addition, there are some API changes/improvements in here:

* `bundle.getMainEntry()` now returns the asset that originally created the bundle, rather than just the last entry. This is stored as a new property on the internal bundle object.
* Sibling bundles that are created when the asset type changes (e.g. importing CSS from JS) now have multiple entries if there are multiple assets added to them, rather than just the first one.
* Sibling bundles are now not shared with other bundle groups (bundler step 2) unless there is only one entry. This was leading to unused CSS being loaded by other pages for example.
* `BundleGraph#addEntryToBundle` is a new method similar to `addAssetGraphToBundle`, but which also adds the asset to the list of entry assets.
* Sibling bundles no longer have a main entry since there can be multiple entries. `getMainEntry` should only be used to determine the asset that should be exported to other module systems (e.g. CJS/ESM).
* Dependencies now have an `isIsolated` option, which allows them to create a new bundle without creating a new bundle group.
* Setting `isEntry` to `false` explicitly on a a dependency allows you to avoid marking the bundle as an entry even if its parent bundle is an entry. For example, in HTML this is used to properly content hash scripts/styles even though they are in the same bundle group as the HTML (entry). But in JS, this is not set, so if your entry point was a JS file, its sibling CSS would not be hashed (to allow you to load it with a predictable filename).
* `BundleGraph#getReferencedBundle` is a new method that resolves a dependency to a bundle. It's similar to `resolveExternalDependency` but that method resolves to a bundle group (I also renamed it to `resolveAsyncDependency` to better describe it).


packages/bundlers/default/src/DefaultBundler.js
```diff
@@ -15,7 +15,7 @@ import nullthrows from 'nullthrows';
 const OPTIONS = {
   minBundles: 1,
   minBundleSize: 30000,
-  maxParallelRequests: 5,
+  maxParallelRequests: 6,
```
devongovett, line 18: The HTML is now counted as one of the requests.

jakearchibald:
This is great!

What happens if HTML isn't involved? As in, if you do a build with two script entry points, which both import a third, will it duplicate the module between the two?

devongovett:
@jakearchibald currently, yes. We currently don't split code out of entries because you'd need to somehow know to load that third script. It would need a predictable name in order to do so, and even with that, if a new duplicated module occurred later, you'd need to know to add that script as well.

As for only keeping one copy of the module, as opposed to splitting it out into a separate bundle, each entry is in its own bundle group at the moment. It's hard to know whether your entries will be loaded on separate pages or all on the same page. We assume that each entry is a separate page at the moment, which seems like the safest assumption. Also, even if they were on the same page, we don't know which order you'll load them in, so it's hard to say which one should keep the duplicated module and which one should reference the other one. Perhaps there could be an option to enable entries to be placed in the same bundle group, but at the moment an HTML entry point is the best way to do this.

PR #4788, opened 2020-06-22
Optimize for HTTP/2 by default and support configuring bundler options

devongovett:
Depends on #4786.

This changes the default options for the bundler to optimize for HTTP/2. In particular, it raises the number of parallel requests allowed from 6 to 25, and lowers the minimum bundle size from 30kb to 20kb. This should allow more granular caching while keeping good performance for apps served with HTTP/2. This is based on the work the Chrome team did with the Next.js team on more granular chunking. See https://web.dev/granular-chunking-nextjs/.

Since HTTP/2 is very widely supported by all browsers and most CDNs/servers, we believe that this is a good default. However, if you are still using HTTP/1.1, or otherwise wish the change these thresholds, you can now configure the bundler. This is done through the package.json in the project root.

This will use the default options for HTTP/1.1:

```json
{
  "@parcel/bundler-default": {
    "http": 1
  }
}
```

If you'd like to override individual options, you can do so as well:

```json
{
  "@parcel/bundler-default": {
    "minBundles": 1,
    "minBundleSize": 0,
    "maxParallelRequests": 10
  }
}
```

I'm not 100% sure about this method of configuring vs others (e.g. `.bundlerrc` file, targets?). Let me know what you think!

### DIFF
--- a/packages/bundlers/default/package.json
+++ b/packages/bundlers/default/package.json
@@ -20,6 +20,7 @@
     "parcel": "^2.0.0-alpha.1.1"
   },
   "dependencies": {
+    "@parcel/diagnostic": "^2.0.0-beta.1",
     "@parcel/plugin": "2.0.0-beta.1",
     "@parcel/utils": "2.0.0-beta.1",
     "nullthrows": "^1.1.1"
--- a/packages/bundlers/default/src/DefaultBundler.js
+++ b/packages/bundlers/default/src/DefaultBundler.js
@@ -5,17 +5,29 @@ import type {
   Bundle,
   BundleGroup,
   MutableBundleGraph,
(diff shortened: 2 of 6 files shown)


wbinnssmith:
This is awesome 🤩 

I think the config needs to be captured in the BundleGraph's cache key somehow, yeah? cc @padmaia

devongovett:
@wbinnssmith yeah see todo comment. I think it's gonna depend on making the bundler a request.

padmaia:
It will need to be part of the cache key, which wouldn't require making it a request, but it would best be done by supporting the `loadConfig` method in bundler plugins. Making bundling a request and adding invalidation will make sure the config file causes a rebuild.

Issue #3823, opened 2019-11-21
Generated asset id is prone to collisions -- breaking imports


ArsenyYankovsky:
# 🐛 bug report

The asset id generation method used in https://github.com/parcel-bundler/parcel/blob/291f3e6815a1a857a6165fafb1691ceeb878b8f6/packages/core/parcel-bundler/src/Asset.js#L208 is prone to collisions. One example of relative paths that will generate same id is:
```
../../node_modules/lodash-es/clone.js
```
```
../../node_modules/core-js/library/modules/_an-object.js
```
both of these assets will generate asset id "ToFw". Having different assets with the same id could and does break some applications. 

## 🎛 Configuration (.babelrc, package.json, cli command)

The build should be a production build or have scope hoisting enabled. This affects only Parcel 1.

## 🤔 Expected Behavior

Asset ids should not have collisions. 

## 😯 Current Behavior

Asset ids collide leading to importing of wrong modules. I.e. importing the `_an-object` results in importing the `lodash-es/clone`.

## 💁 Possible Solution

Check if the hash already exists in the cache with the same relative name and use more characters from the full md5 if required.

mischnic:
> This also affects both v1 and v2 of the parcel.

Parcel 2 uses the full hash, do you have an example where it still fails in this case?

ArsenyYankovsky:
The link above is for the file in the `v2` branch. I assumed it is the branch used for the v2 development.

mischnic:
`v2` is branched off of `master` and Parcel 1 is still in `packages/core/parcel-bundler` so that we can easily merge in v2 without rewriting the whole git history.
Parcel 2 is basically everything in the `v2` branch except for `packages/core/parcel-bundler` 😄

ArsenyYankovsky:
Could you please point me to the updated Asset.js from the Parcel 2 then?

mischnic:
https://github.com/parcel-bundler/parcel/blob/291f3e6815a1a857a6165fafb1691ceeb878b8f6/packages/core/core/src/InternalAsset.js#L60-L65

ArsenyYankovsky:
Ok, thank you. I updated the issue description.

MatAlps:
Hi @mischnic , 

I have had the exact same issue. Now fixed by changing a folder name in my project to avoid collision.

Could this be fixed in parcel 1? 

Thx!

Issue #3747, opened 2019-11-06
Different modules have the same hash value


MrKou47:
# 🐛 bug report

## 🎛 Configuration (.babelrc, package.json, cli command)

https://github.com/MrKou47/parcel-playground

## 🤔 Expected Behavior

The file generated by `parcel build`, different module has same hashname.You can see [this file](https://github.com/MrKou47/parcel-playground/blob/master/build/src.7b44d61f.js),Line-147064, the './block-view' has hashname 'yOt4',and Line-59696,'./registry' has hashname 'yOt4' too.

## 😯 Current Behavior

If i just run `parcel`,everything would be ok.I guess it because use filepath instead of hash to require file.

## 💁 Possible Solution

Maybe we need a config like [webpack optimization.moduleids](https://webpack.js.org/configuration/optimization/#optimizationmoduleids) to solve this problem.

## 💻 Code Sample

https://github.com/MrKou47/parcel-playground

## 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Parcel           |1.12.4
| Node             |8.11.1
| npm/Yarn         |1.5.1
| Operating System |macos 10.14

mischnic:
Probably a hash collision. Parcel 2 uses longer hashes by default

MrKou47:
> Probably a hash collision. Parcel 2 uses longer hashes by default

Thanks for your reply.I tried parcel@2.0.0-alpha.2.1, `parcel index.html` works fine, but `parcel build index.html` got an error:
```bash
$ parcel build ./src/index.html
(node:34183) Warning: N-API is an experimental feature and could change at any time.
🚨 External modules are not supported when building for browser
    at Object.generateExternalImport (/Users/koubo/workspace/parcel-playground/node_modules/@parcel/scope-hoisting/lib/formats/global.js:37:9)
    at exit (/Users/koubo/workspace/parcel-playground/node_modules/@parcel/scope-hoisting/lib/link.js:540:23)
    at NodePath._call (/Users/koubo/workspace/parcel-playground/node_modules/@babel/traverse/lib/path/context.js:55:20)
    at NodePath.call (/Users/koubo/workspace/parcel-playground/node_modules/@babel/traverse/lib/path/context.js:42:17)
    at NodePath.visit (/Users/koubo/workspace/parcel-playground/node_modules/@babel/traverse/lib/path/context.js:99:8)
[...]

Issue #1655, opened 2018-07-03
babel transpile node_module dependencies


tmchng:
# ❔ Question

I need a way to babel transpile some dependencies in `node_modules`, and there doesn't seem to be a way to do so with parcel.

I've read these relevant issues: #948  #13 , and I'm aware of #1101  which solves the case for local repo symlink.

I'm using `file:` dependencies to reference local repos in my project and I need to run babel for those, but no existing solution covers this case.

What would the problems be if we simply specify the target `node_modules` that need to be transpiled with babel? I'm curious why this wasn't considered as an alternative solution to #1101

Having the repos flag themselves as "source" only solves part of the problem, and being able to manually specify packages from the root project seems a lot more intuitive and flexible. I agree that the community needs to collectively come up with a cleaner, more thought-out solution for managing code compatibility for released packages/modules, but until then I believe it is up to the package consumers to make sure that every code they import and release into production is transpiled and made compatible with their desired browser targets. Parcel is a tool for package consumers, therefore I feel it should support such use case.

## 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Parcel           |1.9.3
| Node             |8.9.1
| npm/Yarn         |5.5.1/1.3.2
| Operating System | MacOS HighSierra 10.13.3

icopp:
Right now a project I'm currently working on is using along the lines of:

```js
// .browserslistrc.packages
node 10.11

// package.json
{
  "scripts": {
    "postinstall": "npm-run-all -p \"postinstall:*\"",
    "postinstall:p-retry": "cpy --rename=.browserslistrc .browserslistrc.packages node_modules/p-retry",
    "postinstall:query-string": "cpy --rename=.browserslistrc .browserslistrc.packages node_modules/query-string"
  }
}
```

It's super hacky, but it more or less works.

ford04:
Babel 7 allows custom transpilations of dependencies in node_modules with project-wide `babel.config.js` and `overrides` property ([Link](https://babeljs.io/docs/en/options#overrides)).

```
module.exports = {
  ...
  overrides: [
    {
      test: ["./node_modules/myPackage/dist"],
      plugins: ["@babel/plugin-transform-modules-commonjs"]
    }
  ]
}
```

Can we somehow emulate this on Parcel 1.x ? Will be such scenarios supported in Parcel 2, where a package consumer can decide how he transpiles a package? 

[...]

PR #4800, opened 2020-06-25
fix(*): use the full hash to avoid collisions

ArsenyYankovsky:
# ↪️ Pull Request

Fixes [3823](https://github.com/parcel-bundler/parcel/issues/3823) and [3747](https://github.com/parcel-bundler/parcel/issues/3747). Switching to Parcel 2 is not an option for us since it's in the Beta and doesn't support multi-target builds. I think it's a pretty critical bug fix and should be fixed in the Parcel 1.

## 💻 Examples

One example of relative paths that will generate same id is:
`../../node_modules/lodash-es/clone.js`
`../../node_modules/core-js/library/modules/_an-object.js`

## ✔️ PR Todo


DeMoorJasper:
We don't really update parcel 1 anymore, is there any way we can help you transition to Parcel 2 instead?

ArsenyYankovsky:
I tried switching to Parcel 2 and almost immediately I found a blocking issue for us which is the absence of multi-target builds.  I see that it's in the [todo list](https://github.com/parcel-bundler/parcel/projects/5#card-24711928). I don't think it's good to drop support in the form of bug fixes for Parcel 1 until Parcel 2 is released and considered production-ready.

Thanks for a fast response!

mischnic:
What exactly do you mean by "multi-target builds"? Parcel 1 didn't support this either?
Builds with multiple entry points and/or multiple targets do work. It's just that assigning specific entry points to specific targets isn't possible

ArsenyYankovsky:
We are running the following command to build our apps: `parcel build ./react/apps/*.tsx --no-content-hash` which builds all the apps in that directory. 

Using parcel@nightly currently with the same command gives me a following error:
```
Error: Bundles must have unique filePaths
```

If I just build one of them it gives me files with names like gulpfile.js and gulpfile.css. I guess I need to configure my entries using programmatic API?

mischnic:
> Error: Bundles must have unique filePaths

This usually happens because you have somehow specified an output path for the output bundle but are using multiple entries (which obviously cannot all have the same output filename).

>  I guess I need to configure my entries using the programmatic API?

No, does your package.json contain have a `main`/`module`/`browser` field?

ArsenyYankovsky:
Thank you very much, I had `main` entry in my `package.json` which I removed and now can successfully build. 

However, we need to support IE11 and that means using es5 as target. My output files contain arrow functions and I'm not sure why.

My `.babelrc`:

```
{
  "presets": [
    [
      "@parcel/babel-preset-env",
      {
        "forceAllTransforms": true
      }
    ],
    "@babel/preset-typescript"
  ],
  "plugins": [
    [
      "@babel/plugin-proposal-decorators",
      {
        "legacy": true
      }
    ],
    [
      "@babel/plugin-proposal-class-properties",
      {
        "loose": true
      }
    ],
    [
      "@babel/plugin-transform-react-jsx"
    ]
  ]
}
```

My `tsconfig.json`:

```
{
  "compilerOptions": {
    "module": "commonjs",
    "target": "es5",
    "noImplicitAny": false,
    "outDir": "./dist",
    "lib": [
      "es5",
      "DOM"
    ],
    "sourceMap": true,
    "listEmittedFiles": true,
    "esModuleInterop": true,
    "jsx": "react"
  },
  "esModuleInterop": true,
  "include": [
    "react/**/*"
  ]
}
```

and my `package.json` has following section:

```
  "browserslist": [
    "last 1 version",
    "> 1%",
    "maintained node versions",
    "not dead",
    "IE 10"
  ]
```

I might've missed something but I can't honestly see why. Thank you again.

mischnic:
- tsconfig is ignored by `@babel/preset-typescript`.
- maybe these arrow functions come from node_modules? Your project's babelrc isn't used for npm dependencies.

ArsenyYankovsky:
Yes, they actually do come from the node_modules. All the arrow functions from my code are transformed correctly. Should I use overrides section in my .babelrc to process my dependencies or is there a nicer way? Thank you again.

ArsenyYankovsky:
I tried using `babel.config.js` and `babel.config.json` instead of my `.babelrc` to make it a project-wide configuration but that didn't seem to change anything.

mischnic:
babel.config.* is monorepo-wide, but still not for node_modules.

Ideally, the "includes" field in babelrc would be respected (https://github.com/parcel-bundler/parcel/issues/1655).

DeMoorJasper:
I don't actually think `node_modules` are being transpiled with babel in Parcel 1 either as far as I remember so it's not really different, but we definitely need to support babel's includes at some point

ArsenyYankovsky:
I think parcel 1 uses babel 6 which transpiles `node_modules` by default. I tried different things but what seems to have worked is using `@babel/register`. I will try using it to require all my modules tomorrow and will let you know. Thanks again for great support. You guys are awesome.

ArsenyYankovsky:
I tried using `@babel/register` to process my `node_modules` and it initially complained about reprocessing itself so I ended up ignoring some of the babel and parcel related modules like that:

```
require('@babel/register')({
  ignore:
    [
      './node_modules/@babel/*',
      './node_modules/@parcel/babel-preset-env/*',
      './node_modules/@parcel/transformer-json/*',
      './node_modules/@parcel/transformer-postcss/*',
      './node_modules/@parcel/transformer-css/*',
      './node_modules/@parcel/packager-js/*',
      './node_modules/@parcel/packager-css/*',
      './node_modules/@parcel/optimizer-cssnano/*',
      './node_modules/@parcel/optimizer-terser/*',
    ],
})
```

Unfortunately, there are still arrow functions in the resulting file. I'm running build command with the `--no-cache` flag so there's nothing cached.

ArsenyYankovsky:
I've worked around by removing a dependency that had arrow functions. I had other issues which I applied various workaround to fix. The last thing I encountered is the optimization being so slow it times out on CI (take more than 10 minutes to optimize css). It's been anything but the hassle to make it work. I still don't understand why bugfixes for parcel 1 can't be merge, especially since parcel 2 is in this state. Guess I'll switch to webpack.

DeMoorJasper:
@ArsenyYankovsky Parcel 2 is in a state that is almost as stable as Parcel 1, just lacks a bit of docs and plugin api stability. There's also some issues with cache stability but Parcel 1 also had those...

PR #4807, opened 2020-06-26
Added stub for require.cache

navaru:
# ↪️ Pull Request
 
Fixes: #4616 

Following the previous #4621 discussion on the same issue, I've made the changes.

## ✔️ PR Todo


DeMoorJasper:
Looks good to, @mischnic could you have a look to validate the changes to scope-hoisting are done properly?

packages/shared/scope-hoisting/src/link.js
```diff
@@ -603,6 +603,17 @@ export function link({
...
+      enter(path) {
+        const isCommonJS = bundle.env.outputFormat == 'commonjs';
+
+        if (t.matchesPattern(path.node, 'module.require') && !isCommonJS) {
+          path.replaceWith(t.identifier('null'));
+        }
+
+        if (t.matchesPattern(path.node, 'require.cache') && !isCommonJS) {
+          path.replaceWith(t.objectExpression([]));
+        }
```
mischnic, line 615: 1.
```js
{
	let module = {
		require() {
			return 1;
		},
	};
	const x = module.require("fs");

	console.log(module, x);
}
```
Is transformed to `...; var x = null("fs"); ...`, you need `&& !path.scope.hasBinding("module")` (and same for `require`).

Then this should work:
```js
{
	function require() {
		return 1;
	}
	require.cache = [1, 2, 3];
	let module = {
		require,
	};
	const x = module.require("fs");

	console.log(module, module.require("fs"), require.cache[0]);
}
```

2. The following would throw a cryptic error (`Property left of AssignmentExpression expected node to be of a type ["LVal"] but instead got "ObjectExpression"`) even with the `hasBinding` check.
```js
{
	require.cache = [1, 2, 3];
	let module = {
		require() {
			return 1;
		},
	};
	const x = module.require("fs");

	console.log(module, module.require("fs"), require.cache[0]);
}
```

(And these should probably be integration tests as well)

navaru:
Currently I'm overwhelmed by stuff at work, closing for now, will be back in a couple of weeks.

Issue #4796, opened 2020-06-24
Parcel 2 fails to bundle TypeScript code that re-exports types


evanw:
# 🐛 bug report

Now that Parcel 2 is out of alpha, I thought I'd try adding it to [esbuild's bundler performance benchmarks](https://github.com/evanw/esbuild/blob/32f756b16d2e62c8cfcf3115221322c364a02a5e/README.md#benchmarks). However, Parcel 2 currently fails to build both of the benchmarks. This issue is about the TypeScript benchmark.

## 🎛 Configuration (.babelrc, package.json, cli command)

Here is how to reproduce the issue:

```
git clone https://github.com/evanw/esbuild
cd esbuild
make bench-rome-parcel2
```

## 🤔 Expected Behavior

I expected Parcel 2 to be able to build my TypeScript benchmark. Parcel 1 can build the benchmark fine. The benchmark is a copy of the [Rome bundler](https://github.com/romejs/rome) source code.

## 😯 Current Behavior

My first attempt at building this benchmark with Parcel 2 failed immediately with a TypeScript parse error:

```
🚨 Build failed.
@parcel/transformer-babel: /Users/evan/dev/esbuild/parcel2/bench/rome/src/@romejs/core/client/Client.ts: `import =` is not supported by @babel/plugin-transform-typescript
Please consider using `import <moduleName> from '<moduleName>';` alongside Typescript's --allowSyntheticDefaultImports option.
  32 | import {PartialMasterQueryRequest} from '../common/bridges/MasterBridge';
  33 | import {loadUserConfig, UserConfig} from '../common/userConfig';
> 34 | import stream = require('stream');
     | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  35 |
  36 | import net = require('net');
  37 |
```

I did some digging and found #2023, which says that while Parcel 1 uses the TypeScript compiler to parse TypeScript files, Parcel 2 will use Babel instead. And it looks like Babel's support for TypeScript syntax is incomplete. I implemented the suggested workaround of using `@parcel/transformer-typescript-tsc` instead:

```json
{
  "extends": ["@parcel/config-default"],
  "transformers": {
    "*.ts": ["@parcel/transformer-typescript-tsc"]
  }
}
```

That let Parcel 2 successfully complete the parsing phase. However, it then failed at the bundling stage:

```
🚨 Build failed.
[...]

PR #4841, opened 2020-07-03
Assetgraph propagate symbols

### LINKED ISSUE #4565: Parcel2 v. Webpack - parcel bundle size is significantly larger
# 🐛 bug report

I've started trying to use Parcel2 in a real-life situation, and I'm noticed that the parcel bundle size (1.1MB) is significantly larger than webpack's (666 KiB) for the same app. (It's also significantly slower - see #4566).

Like all real-life situations, it's complicated, so I tried to create a simplified repro environment in the `fluentui-button-babel` branch of [this repo](https://github.com/astegmaier/parcel-webpack-comparer/tree/fluentui-button-babel). It's a simple react app with a single `Button` component from the [`@fluentui/react-northstar`](https://fluentsite.z22.web.core.windows.net/) library. This library is alpha-quality, and there's [known issues](https://github.com/microsoft/fluentui/issues/12953) about bundle size that the team is trying to address. However, webpack still does a significantly better job than parcel right now. The main driver seems to be that [this optimization](https://medium.com/@marijanajdova/fluentui-react-how-we-cut-more-than-30-of-components-bundle-size-by-creating-icons-package-474019d3123) that the library authors made is picked up by webpack but not by parcel.

***Webpack*** (full config [here](https://github.com/astegmaier/parcel-webpack-comparer/blob/fluentui-button-babel/webpack.config.js)):
- Using `babel-loader` with this config:
```
presets: [
   "@babel/preset-env",
   "@babel/preset-react",
   "@babel/preset-typescript",
]
```
- `production` flag set to `true`, which enables minification through [`TerserWebpackPlugin`](https://webpack.js.org/plugins/terser-webpack-plugin/) and scope-hoisting/tree-shaking through [`ModuleConcatinationPlugin`](https://webpack.js.org/plugins/module-concatenation-plugin/), among other optimizations.

***Parcel configuration***:

- No `.parcelrc` file, which means that the default babel transformer is being used.
- No custom babel configuration.
- Ran `parcel build src/index.html`

***src/index.html***:
```html
<!DOCTYPE html>
<html lang="en">
  <head>
    <meta charset="utf-8" />
    <title>Sample App</title>
  </head>
  <body>
    <noscript>You need to enable JavaScript to run this app.</noscript>
    <div id="root"></div>
    <script src="./index.tsx"></script>
  </body>
</html>
```

***src/index.tsx***:
```typescript
import React from "react";
import ReactDOM from "react-dom";
import App from "./components/App";

ReactDOM.render(<App />, document.getElementById("root"));
```

***src/components/App.ts***:
```typescript
import React from "react";
import { Provider, themes, Button } from "@fluentui/react-northstar";

const App = () => (
  <Provider theme={themes.teams}>
    <Button content="Hello from FluentUI" />
  </Provider>
);

export default App;
```

## 🤔 Expected Behavior

Parcel and webpack build size should be roughly the same (or maybe Parcel should be better 😄).

## 😯 Current Behavior

Here's the bundle size comparison:
| Scenario                                                                                      | Bundle Size |
| --------------------------------------------------------------------------------------------- | ----------- |
| Build with parcel (default babel config with scope-hoisting and minification)                 | 1.11 MB     |
| Build with webpack (`babel-loader` with scope-hoisting and minification)                      | 660 KiB     |
| Build with webpack (`ts-loader` with `module: es6` with scope-hoisting and minification)      | 660 KiB     |
| Build with webpack (`ts-loader` with `module: commonjs` with scope-hoisting and minification) | 1.21 MB    |

## 💁 Possible Solution

By using `source-map-explorer` and [a custom node script](https://github.com/astegmaier/parcel-webpack-comparer/blob/fluentui-button-babel/scripts/analyze.ts), I generated a visualization of the stuff that is in the parcel bundle, but not the webpack bundle:

image

You can see that the main driver of the increased size are icons from the `@fluentui/react-icons-northstar`. The fluentui team recently made [an optimization](https://medium.com/@marijanajdova/fluentui-react-how-we-cut-more-than-30-of-components-bundle-size-by-creating-icons-package-474019d3123) to allow webpack to tree-shake these icons, and it seems that it works with webpack, but not with parcel.

I'm not sure why this is happening - most likely something at the scope-hoisting phase that doesn't like the way this library was written. Even if there's something the library authors could do to improve this, it seems like a good goal to do our best to make sure that libraries that are optimized for webpack remain optimized with parcel.

## 🔦 Context

Trying to build a real-world app that uses [`@fluentui/react-northstar`](https://fluentsite.z22.web.core.windows.net/) and parcel2.

## 💻 Code Sample

See the `fluentui-button-babel` branch of [this repo](https://github.com/astegmaier/parcel-webpack-comparer/tree/fluentui-button-babel).

## 🌍 Your Environment

| Parcel           |  2.0.0-nightly.248
| Node             | 12.16.3
| Yarn         | 1.22.4
| Operating System | OSX 10.15.4

mischnic:
# ↪️ Pull Request

Benefits of symbol propagation (e.g. with calling resolveSymbol repeatedly in the bundler):

1. Tree shaking across (shared) bundles
2. Better/faster tree shaking with many namespace reexports (e.g. react component libraries)
3. Not implemented in this PR: Tree shaking dynamic imports
4. Not implemented in this PR: Somehow operate on symbols cross assets (e.g. skip TS type reexports)
5. Not implemented in this PR (not sure how slower it would be): actually do the symbol resolution in the graph. Could catch conflicting reexports (spec compliant)

Benefits of assetgraphbuilder symbol propagation (doing it after building the asset graph):
1. Exclude a subgraph of assets during bundling (e.g. to also exclude CSS imports of unused&side-effect free assets or not creating async bundles that are imported from dead assets)

This PR: Benefits of assetgraph(builder) symbol propagation (as opposed to doing it in the bundle graph):

1. Incremental ("faster")
2. Deferring is always correct (except with namespace reexports - chained named reexports do work)
3. Way way slower...

What this PR implements:

1. Propagate the actually used symbols down the asset graph (`dependencyNode.usedSymbolsDown`) by aligning imports with reexports as best as possible.
2. To handle namespace reexports correctly (where the imports have to be distributed to every reexport), there is another pass that walks the finished asset graph in post-order and propagates the used & available symbols back up. (This is actually very fast - 80ms for the ak-editor. Could also expanded later to solve https://github.com/parcel-bundler/parcel/issues/4796).
So for every dependency node, `usedSymbolsDown` is a superset of the actual `usedSymbolsUp` (which is exposed in the public BundleGraph )

Now comes the nasty part: When a dependency is added/removed (and the asset is resolves to is already in the graph), the dependency node (and part of the subgraph) has to be walked to add/remove symbols, this causes some nodes to be visited multiple times (`{ downVisitedCount: 27699, upVisitedCount: 16001 }`, `upVisitedCount` is the true number of nodes).
This is what causes a slowdown of 10 seconds when building the asset graph for ak-editor.

`usedSymbolsDown` is actually a map of `symbol -> Set<Incoming Dependencies that caused this symbol request>` to facilitate incremental updates

~At the moment, the build size is not deterministic 😬  (the builds seem to be working however) - I'm guessing some updates to the used symbols are lost~

Closes #4565

## 💻 Examples

ak-editor: 500k smaller but ~10s~ 18s slower (this is the worst case, because almost everywhere in `@atlaskit/*`, named reexports are used)

```
this:
 build . --no-cache --no-source-maps --no-minify --log-level warn  288.47s user 42.57s system 383% cpu 1:26.36 total
 16015360	dist

v2:
 build . --no-cache --no-source-maps --no-minify --log-level warn  212.17s user 27.54s system 350% cpu 1:08.31 totaldist/  16527360 dist


building only the asset graph:
this:
 PARCEL_WORKERS=0  build . --no-cache --no-source-maps --no-minify --profile    151.91s user 33.03s system 152% cpu 2:01.12 total
v2:
 PARCEL_WORKERS=0  build . --no-cache --no-source-maps --no-minify --profile    93.44s user 19.82s system 152% cpu 1:14.16 total
```

`@fluentui/react-northstart` as a react component library with many namespace reexports (#4565)
```
$ time parcel2 build src/index.html --no-cache
✨ Built in 17.37s

dist/index.html               245 B    4.51s
dist/index.173315dd.js    232.84 KB    7.41s
 build src/index.html --no-cache  52.73s user 6.22s system 314% cpu 18.718 total
$ time parcel2 build src/index.html --no-cache
✨ Built in 28.39s

dist/index.html               245 B     1.35s
dist/index.b261fe70.js    1012.9 KB    14.82s
 build src/index.html --no-cache  78.04s user 9.98s system 286% cpu 30.757 total
```

## ✔️ PR Todo

- [x] Build size is non-deterministic !?!?
- [ ] Is deleting a dependency, which causes a cascade (orphans), handled correctly?
- [ ] Faster? - visit fewer nodes multiple times
      - Maybe keep the previous defer heuristic and be less aggressive about revisiting nodes ?
      - Maybe batch dirty dependencies?
- [ ] Remove debug comments


packages/core/core/src/AssetGraph.js
```diff
@@ -187,100 +206,324 @@ export default class AssetGraph extends Graph<AssetGraphNode> {
...
+
+    if (changedIncomingDeps.length === 0) {
+      // Root in the runtimes Graph
+      assetUsedSymbols.get('*').add(null);
+      namespaceReexportedSymbols.get('*').add(null);
+    } else {
+      for (let incomingDep of changedIncomingDeps) {
+        // TODO isIsolated?
+        if (incomingDep.value.isEntry || incomingDep.value.isIsolated) {
+          isEntry = true;
```
mischnic, line 338: The toplevel asset is special because it has no incoming dependencies with symbols and so it would simply be deferred. Not sure if `isEntry || isIsolated` is correct here (JS entry files, JS script tags)

packages/core/core/src/AssetGraph.js
```diff
@@ -315,8 +558,13 @@ export default class AssetGraph extends Graph<AssetGraphNode> {
           dependentAssets.push(dependentAsset);
         }
       }
+
+      // TODO ???
+      let assetNode = this.getNode(asset.id) ?? nodeFromAsset(asset);
```
mischnic, line 563: Previously, `assetNode.usedSymbols` was being discarded/overwritten - is this the correct solution?

mischnic:
ak-editor:

- this (propagate during assetgraph building, meaning aggressive deferring):
286.06s user 42.96s system 382% cpu 1:26.07 total
16015360 bytes

- alternative (old deferring heuristic, propagate after building assetgraph, skip subgraphs in bundler):
211.91s user 28.20s system 366% cpu 1:05.57 total
16117760 bytes
(I have no idea how the bundler works, so it might be possible to make smaller)

- v2:
219.03s user 28.07s system 340% cpu 1:12.59 total
16527360 bytes

PR #4850, opened 2020-07-05
Use excludeNodeModules instead of includeNodeModules

### LINKED ISSUE #4846: excludeNodeModules instead of includeNodeModules
# 🙋 feature request

The ability to use `excludeNodeModules` rather than `includeNodeModules` in each target config. Many more people need externals than need to forcibly bundle a specific module, so this seems like something that would semantically make sense.

## 🤔 Expected Behavior

```
{
  "targets": {
    "library": {
      "excludeNodeModules": {
        "react": true,
        "jquery": "$"
      }
    }
  }
}
```
A string value should be treated as the name of a global, while a boolean should be whether to keep the `require` or not (though, in my opinion, a value of `true` should cause an error in a browser build).

The same array syntax as `includeNodeModules` should be allowed, but only for exclusion of modules.

## 😯 Current Behavior

With `includeNodeModules`, not only are globals impossible, the syntax to exclude multiple (but not all) modules is cumbersome in comparison to a single array.

## 💁 Possible Solution

The Parcel source must already have some way to exclude any given node module. To add support for globals, Parcel must dynamically generate a fake dependency, i.e. to make `jquery` map to `$`, Parcel could create an asset with solely `module.exports=$`.

## 🔦 Context

This would fix #3305, which I have a similar use case for. Effectively, I'd like to emulate the behavior of [`parcel-plugin-externals`](https://github.com/FlorianRappl/parcel-plugin-externals).

101arrowz:
# ↪️ Pull Request

WIP. Fixes #4846, #3305.

Should allow the use of globals in addition to externals. Also allows a semantically nicer API. Untested.

## 💻 Examples

TODO

## 🚨 Test instructions

TODO

## ✔️ PR Todo


packages/utils/node-resolver-core/src/NodeResolver.js
```diff
@@ -275,8 +275,13 @@ export default class NodeResolver {
       };
     }
 
-    if (!this.shouldIncludeNodeModule(env, filename)) {
+    let shouldInclude = this.shouldIncludeNodeModule(env, filename);
+    if (!shouldInclude) {
       return null;
+    } else if (typeof shouldInclude === 'string') {
```
101arrowz, line 281: Is this all that's needed to allow support for globals?

mischnic: Yes (so theoretically, you could create a resolver plugin that does this - without any changes to Parcel itself).

101arrowz: Alright. I thought adding globals support as a core feature would be useful, which is why I made this PR, but if it's out of scope for the project I could remove it and make a plugin. Which is preferable?

mischnic: We are not against more features, but renaming `includeNodeModules` is a rather breaking change.

101arrowz: I was hoping that breaking changes wouldn't be a major annoyance since Parcel v2 is still in beta and not that many people are using it in comparison to v1. If necessary I'll create a plugin, but I don't like the idea of fragmenting similar functionality into separate places, especially when globals should be configurable on a target-by-target basis.

devongovett:
I don't think we should switch `includeNodeModules` to `excludeNodeModules`. It's less flexible that way. For example, I use `includeNodeModules` to only include specific node modules when building a library, rather than needing to list every possible module to exclude. This wouldn't be possible with `excludeNodeModules`. Meanwhile, excluding modules is already possible by setting them to `false` with `includeNodeModules`.

101arrowz:
The main feature this PR tries to add is globals; `excludeNodeModules` is just semantics. I'll make another PR with only globals support and no `excludeNodeModules`. My only concern is semantic correctness; when you specify a string value in the `includeNodeModules` object, you would assume that the module would be included (since strings are truthy), but in this case the string means that the module is not bundled.

devongovett:
True. What do you think about using the `alias` field in package.json for this instead of `includeNodeModules`?

101arrowz:
See [my comment](https://github.com/parcel-bundler/parcel/issues/4846#issuecomment-654419954). I think it's certainly a good idea but there has to be a secondary method for some use cases to specify globals by target.

PR #4853, opened 2020-07-07
Add globals support

### LINKED ISSUE #3305: Support "external" modules in Parcel 2

101arrowz:
# ↪️ Pull Request

Adds support for globals using the `alias` in `package.json`.

Based off of #4846 and #4850.

Fixes #3305.

## 💻 Examples

In `package.json`:
```json
{
    "alias": {
        "react": { "global": "React" },
        "myAliasedPackageName": { "fileName": "./myPkg.js" },
        "finalPackage": "finalPackage.js"
    }
}
```
All of the above syntax forms work, for globals and normal aliases respectively. <strike>The pure string syntax works less reliably because it has to guess that it's not an alias based on whether or not the file exists.</strike> (Removed)

## ✔️ PR Todo

### DIFF
--- a/packages/core/integration-tests/test/globals.js
+++ b/packages/core/integration-tests/test/globals.js
@@ -1,0 +1,20 @@
+import assert from 'assert';
+import path from 'path';
+import {bundle, outputFS} from '@parcel/test-utils';
+
+describe('global alias', function() {
+  it('should support all alias syntax', async function() {
+    let b = await bundle(
+      path.join(__dirname, '/integration/global-alias/index.js'),
+    );
+
+    let index = await outputFS.readFile(
+      b.getBundles().find(b => b.name.startsWith('index')).filePath,
+      'utf8',
+    );
+
+    assert(/module\.exports\ =\ React/.test(index));
+    assert(/module\.exports\ =\ \$/.test(index));
+    assert(/module\.exports\ =\ \_/.test(index));
+  });
+});
--- a/packages/core/integration-tests/test/globals.js
+++ b/packages/core/integration-tests/test/globals.js
@@ -1,6 +1,6 @@
 import assert from 'assert';
 import path from 'path';
-import {bundle, outputFS} from '@parcel/test-utils';
+import {bundle, run} from '@parcel/test-utils';
 
 describe('global alias', function() {
   it('should support global alias syntax', async function() {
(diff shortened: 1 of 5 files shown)


mischnic:
Why do we need so many variants of specifying a global? Does some other tool have this?

I think just adding `{"global": "React"}` is less confusing (and not ambigious)
```json
{
    "alias": {
        "react": { "global": "React" },
        "lodash": "always-a-filename-alias.js"
    }
}
```

101arrowz:
I personally like multiple ways of specifying options. For sure, I think having it possible to do just
```json
{
     "jquery": "$"
}
```
would be important because developers would find that more intuitive and less clunky, especially since they're very unlikely to have a file named `$`. I could remove the `["global", "$"]` method (although right now it's just a byproduct of how I'm representing aliases and could just remain undocumented).

If you'd like to make it a rigid structure, I can do that for sure, but it wouldn't be as user-friendly.

mischnic:
> would be important because developers would find that more intuitive and less clunky,

macOS has a case-insensitive filesystem, so `"react": "React"` wouldn't even be stable across OSs
(if it would check `if(does package of that name exist){ usePackage()} else { useGlobal() }`)

101arrowz:
Fair point, but they could always just use `window.React` if absolutely necessary. I'll remove the string syntax if you insist, though.

101arrowz:
Any feedback on the implementation?

packages/core/integration-tests/test/globals.js
```diff
@@ -0,0 +1,18 @@
...
+    let b = await bundle(
+      path.join(__dirname, '/integration/global-alias/index.js'),
+    );
+
+    let index = await outputFS.readFile(
+      b.getBundles().find(b => b.name.startsWith('index')).filePath,
+      'utf8',
+    );
+
+    assert(/module\.exports\ =\ React/.test(index));
```
mischnic, line 16: Testing this using  `module.exports = React.createElement...` and
```js
let output = await run(b, {React: {createElement(tag){ return "ok"; }}});

assert.equals(output, "ok");
```
would be less brittle.

101arrowz: Fixed in 825a3eace291139a66f74e3df8d5a072488fbb97.

101arrowz:
@mischnic If there's nothing else that needs changes, I'd recommend merging before the next beta. I have a couple projects I'd be able to move from Webpack to Parcel 2 if Parcel 2 supported globals. By the way, the reason I recommended string syntax was for [similarity to Webpack](https://webpack.js.org/configuration/externals/).

mischnic:
The tests are failing

> I think just adding {"global": "React"} is less confusing (and not ambigious)

We still think this is the right path to take (only having the object and not an array). If someone reads the docs for the first time, the first question would probably be: "why are there two ways to achieve the same thing? There has to be *something*"

> recommended string syntax

Makes sense, but that does unfortunately not really work with `alias`

101arrowz:
I have already removed the string syntax (check the commit history). I have now fixed the failing tests (at least on my machine), thanks for pointing it out.

101arrowz:
I have already removed the string syntax (check the commit history). I have now fixed the failing tests (at least on my machine), thanks for pointing it out.

PR #4881, opened 2020-07-12
[T-6] Image transformer

### LINKED ISSUE #3477: [RFC] Import query params

### LINKED ISSUE #3736: Parcel 2: WebP Transformer

### LINKED ISSUE #3737: Parcel 2: Image Resizing Transformer

DeMoorJasper:
# ↪️ Pull Request

Allow resizing of images using query parameters and sharp

Closes #3737
Closes #3736
Closes #3477

Closes T-6

## 💻 Examples

```
import image from 'url:./image.jpg?height=200';
import image from 'url:./image.webp?height=200&width=300';
import image from 'url:./image.png?width=300';
```

## 🚨 Test instructions

Run kitchen sink example

## ✔️ PR Todo


parcel-benchmark:
## Benchmark Results

#### Timings

| --- | --- | --- |
| Cold | 6.66s | +834.00ms ⚠️ |
| Cached | 2.13s | -9.00ms |

#### Cold Bundles

| dist/legacy/parcel.d5807e82.webp | 102.94kb | -12.74kb 🚀 | 260.00ms | -3.00ms |
| dist/modern/parcel.d5807e82.webp | 102.94kb | -12.74kb 🚀 | 259.00ms | -4.00ms |
| dist/legacy/index.42516b4a.js | 1.10kb | +0.00b | 1.52s | +105.00ms ⚠️ |
| dist/modern/index.8a1ced31.js | 1.10kb | +0.00b | 1.52s | +105.00ms ⚠️ |
| dist/legacy/index.html | 701.00b | +0.00b | 596.00ms | -70.00ms 🚀 |
| dist/modern/index.html | 701.00b | +0.00b | 593.00ms | -71.00ms 🚀 |
| dist/modern/index.e480862d.css | 77.00b | +0.00b | 1.39s | +90.00ms ⚠️ |

#### Cached Bundles

| dist/legacy/parcel.d5807e82.webp | 102.94kb | -12.74kb 🚀 | 199.00ms | -68.00ms 🚀 |
| dist/modern/parcel.d5807e82.webp | 102.94kb | -12.74kb 🚀 | 198.00ms | -69.00ms 🚀 |

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 26.27s | +346.00ms |
| Cached | 2.31s | -19.00ms |

#### Cold Bundles

| dist/logo.24c8bf9e.png | 274.00b | -54.00b 🚀 | 124.00ms | -83.00ms 🚀 |

#### Cached Bundles

| dist/logo.24c8bf9e.png | 274.00b | -54.00b 🚀 | 243.00ms | -3.00ms |

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 2.77m | -2.01s |
| Cached | 4.82s | +119.00ms |

#### Cold Bundles

| dist/index.d6c6b729.js | 2.73mb | +3.00b ⚠️ | 1.57m | -3.69s |
| dist/pdfRenderer.cd742f5e.js | 1.11mb | +0.00b | 48.17s | -4.02s 🚀 |
| dist/Toolbar.33a83ae3.js | 92.34kb | +0.00b | 54.38s | -43.74s 🚀 |
| dist/DatePicker.cbe44a9b.js | 21.17kb | +0.00b | 10.35s | +526.00ms ⚠️ |
| dist/EmojiPickerComponent.898d8be6.js | 18.23kb | +0.00b | 10.35s | +518.00ms ⚠️ |
| dist/EmojiPickerComponent.624bf3f4.js | 3.58kb | +0.00b | 10.35s | +518.00ms ⚠️ |
| dist/dropzone.2f1a1e17.js | 3.23kb | +0.00b | 48.12s | +18.67s ⚠️ |
| dist/Modal.2835744c.js | 3.08kb | +0.00b | 4.22s | -496.00ms 🚀 |
| dist/card.cadccd96.js | 2.03kb | +0.00b | 27.32s | -2.04s 🚀 |
| dist/ResourcedEmojiComponent.d9a49606.js | 2.00kb | +0.00b | 3.93s | -756.00ms 🚀 |
| dist/date.d49e64f8.js | 1.73kb | +0.00b | 6.86s | +366.00ms ⚠️ |
| dist/images.80270b1f.js | 1.67kb | +0.00b | 6.98s | -46.60s 🚀 |
| dist/16.86a8c7d4.js | 1.66kb | +0.00b | 6.87s | +435.00ms ⚠️ |
| dist/feedback.8b7c7928.js | 1.64kb | +0.00b | 7.32s | +388.00ms ⚠️ |
| dist/list-number.4632b459.js | 1.46kb | +0.00b | 52.76s | +46.13s ⚠️ |
| dist/code.af9ca2b3.js | 1.38kb | +0.00b | 6.86s | +366.00ms ⚠️ |
| dist/16.c1ed07f8.js | 1.33kb | +0.00b | 6.87s | +435.00ms ⚠️ |
| dist/16.249c644a.js | 1.31kb | +0.00b | 6.85s | +364.00ms ⚠️ |
| dist/link.d12cc130.js | 1.30kb | +0.00b | 52.76s | +46.13s ⚠️ |
| dist/heading6.4c4017db.js | 1.30kb | +0.00b | 7.32s | +388.00ms ⚠️ |
| dist/heading3.93606b9c.js | 1.29kb | +0.00b | 7.32s | +388.00ms ⚠️ |
| dist/emoji.e603e70e.js | 1.23kb | +0.00b | 6.98s | -46.60s 🚀 |
| dist/16.06d5a67b.js | 1.18kb | +0.00b | 6.85s | +365.00ms ⚠️ |
| dist/heading5.dde15cec.js | 1.17kb | +0.00b | 7.32s | +388.00ms ⚠️ |
| dist/expand.6889aead.js | 1.16kb | +0.00b | 10.35s | +525.00ms ⚠️ |
| dist/media-card-analytics-error-boundary.6a25b970.js | 1.15kb | +0.00b | 48.10s | +18.66s ⚠️ |
| dist/16.ecf8a702.js | 1.13kb | +0.00b | 7.07s | +608.00ms ⚠️ |
| dist/16.da43aa07.js | 1.13kb | +0.00b | 6.87s | +435.00ms ⚠️ |
| dist/heading2.e219309a.js | 1.11kb | +0.00b | 52.76s | +45.91s ⚠️ |
| dist/16.dd07505e.js | 1.10kb | +0.00b | 3.93s | -756.00ms 🚀 |
| dist/component.40194190.js | 1.07kb | +0.00b | 3.93s | -756.00ms 🚀 |
| dist/mention.b1828447.js | 1.07kb | +0.00b | 7.04s | +418.00ms ⚠️ |
| dist/heading4.5316f6f2.js | 1.06kb | +0.00b | 7.32s | +388.00ms ⚠️ |
| dist/layout.ac6734f9.js | 1.05kb | +0.00b | 6.98s | +350.00ms ⚠️ |
| dist/divider.d2be55a1.js | 1.03kb | +0.00b | 6.98s | -46.60s 🚀 |
| dist/action.cb287eb0.js | 1.00kb | +0.00b | 6.86s | +366.00ms ⚠️ |
| dist/decision.91f3af14.js | 1010.00b | +0.00b | 6.98s | -46.60s 🚀 |
| dist/list.057cb664.js | 978.00b | +0.00b | 52.76s | +46.14s ⚠️ |
| dist/panel.9dbc048e.js | 897.00b | +0.00b | 7.19s | +442.00ms ⚠️ |
| dist/media-picker-analytics-error-boundary.553055f6.js | 837.00b | +0.00b | 50.14s | -2.60s 🚀 |
| dist/index.html | 119.00b | +0.00b | 808.00ms | +126.00ms ⚠️ |

#### Cached Bundles

| dist/pdfRenderer.cd742f5e.js | 1.11mb | +4.00b ⚠️ | 595.00ms | +28.00ms |
| dist/media-viewer.bc661bce.js | 3.92kb | +0.00b | 595.00ms | +34.00ms ⚠️ |

</p></details>

<details><summary>Three.js x4 🚨</summary><p>

#### Timings

| --- | --- | --- |

#### Cold Bundles

#### Cached Bundles

</p></details>

**[Click here to view a detailed benchmark overview.](https://benchmark.parceljs.org/benchmark/270847872156566029)**

packages/configs/default/index.json
```diff
@@ -27,6 +27,7 @@
     "*.pug": ["@parcel/transformer-pug"],
     "*.coffee": ["@parcel/transformer-coffeescript"],
     "*.mdx": ["@parcel/transformer-mdx"],
+    "*.{png,jpg,jpeg,webp}": ["@parcel/transformer-image", "..."],
```
devongovett, line 30: So, currently if you want to import an image from JS you need to use the url: pipeline. This would make it so you do not. And if you did, then you wouldn’t get resizing capabilities because it wouldn’t use this pipeline... Should url: be required? For non JS it isn’t though, so not sure how we’d implement that.

packages/utils/node-resolver-core/src/NodeResolver.js
```diff
@@ -49,6 +52,26 @@ type Module = {|
   code?: string,
 |};
 
+const parseFilePath = (
```
devongovett, line 55: Wonder which parts of specifier parsing should be up to the resolver and which should be up to core. I guess the pipeline is parsed by core. Should query params be parsed by core as well? No strong opinion just wondering.

packages/configs/default/index.json
DeMoorJasper: Not sure, we could add it if you prefer. Probably makes more sense to stay consistent

packages/utils/node-resolver-core/src/NodeResolver.js
DeMoorJasper: Probably?

packages/configs/default/index.json
```diff
@@ -32,6 +32,7 @@
     "script:*.vue": ["@parcel/transformer-vue"],
     "style:*.vue": ["@parcel/transformer-vue"],
     "custom:*.vue": ["@parcel/transformer-vue"],
+    "url:*.{png,jpg,jpeg,webp}": ["@parcel/transformer-image", "..."],
```
devongovett, line 35: hmm so now you couldn't use it from CSS/HTML without the `url:` prefix either... :/

DeMoorJasper: that's probably why I originally did it the other way...

DeMoorJasper: so should I change it back?

packages/transformers/image/src/ImageTransformer.js
```diff
@@ -0,0 +1,49 @@
...
+
+    let inputBuffer = await asset.getBuffer();
+    let width = asset.query.width ? parseInt(asset.query.width, 10) : null;
+    let height = asset.query.height ? parseInt(asset.query.height, 10) : null;
+    let quality = asset.query.quality
+      ? parseInt(asset.query.quality, 10)
+      : undefined;
+    let format = asset.query.as ? asset.query.as.toLowerCase().trim() : null;
+
+    let imagePipeline = sharp(inputBuffer);
```
devongovett, line 28: If none of the options are set, should we skip running sharp altogether? Maybe faster?

DeMoorJasper: But the output might be non minified

devongovett: Hmm it's not an optimizer... I guess we could check that option too. I was more thinking in dev it might be slow/unnecessary.

DeMoorJasper: Yeah... I can fix that in a follow-up PR if you want, sharp isn't always smaller anyway, I've had a couple cases in which it significantly enlarged output images...

PR #4888, opened 2020-07-13
Fix config merging

### LINKED ISSUE #4883: Extending multiple configs fails
# 🐛 bug report

When extending multiple config files, merging them doesn't work.

Possibly related: #4842 

`.parcelrc`:
```json
{
  "extends": ["@parcel/config-default", "parcel-config-precache-manifest", "parcel-config-pwa-manifest"]
}
```

`parcel-config-precache-manifest/.parcelrc`:
```json
{
  "runtimes": {
    "service-worker": ["...", "parcel-runtime-precache-manifest"]
  }
}
```
`parcel-config-pwa-manifest/.parcelrc`:
```json
{
  "transformers": {
    "*.{htm,html}": ["@parcel/transformer-html", "parcel-transformer-pwa-manifest"],
    "*.{xml,png,webp,tiff,jpg,webmanifest}": ["@parcel/transformer-raw"]
  },
  "namers": ["parcel-namer-pwa-manifest", "..."]
}
```

## 🤔 Expected Behavior

Both the modified transformers and namers from `parcel-config-pwa-manifest` and the modified runtimes from `parcel-config-precache-manifest` are applied.

## 😯 Current Behavior

The build silently fails.
If I add `extends: "@parcel/config-default"` to both the plugin configs, only the pwa-manifest one applies.
If I move `@parcel/config-default` to the end of the `extends` array, only the precache-manifest one applies.
In both cases, if I swap the pwa-manifest and precache-manifest configs, the configs that apply swap as well.

## 💁 Possible Solution

I tried tinkering with the source code but couldn't find any way to make both plugins apply without breaking the build.

## 🔦 Context

I'm trying to use two plugins at once, both of which provide a config package for easier use.

## 🌍 Your Environment

| Parcel           | 2.0.0-beta.1
| Node             | 12
| Operating System | Ubuntu (WSL)

101arrowz:
# ↪️ Pull Request

Modifies the behavior of config merging.

Previously, Parcel would iterate down the `extends` array and have each element be the base config to be merged with the original file. Now, each config in `extends` merges with the next (effectively aggregating together) before finally merging with the original config. This provides a much more natural-feeling merging strategy and allows for multiple incomplete config packages to properly merge with the default.

Fixes #4883

## 💻 Examples

See #4883 configuration section.

## ✔️ PR Todo


101arrowz:
Should I add unit tests for these changes? If so, which plugins could I use?

101arrowz:
Are there any packages in the core that have a partial config? Given that it's impossible to create local plugins right now, it could be difficult to test the partial-config-merging functionality.

DeMoorJasper:
@101arrowz true, just leave it as is for now as there's not really a partial config for any of the plugins

mischnic:
You can create packages in node_modules inside the test fixture and commit that folder: https://github.com/parcel-bundler/parcel/tree/v2/packages/core/integration-tests/test/integration/commonjs-bundle-require/node_modules/parcel-bundler-splitable

101arrowz:
Alright, I've added an integration test using @mischnic's suggestion at #4910.

PR #4892, opened 2020-07-15
Vue Transformer

### LINKED ISSUE #1183: 🐛 In VueAsset `supplemental` is minified independently from `js` which sometimes leads to weird shadowing.
What the issue's title says.

It can lead to very confusing bugs, for example your imports stop being what they should be, but only in production build, etc.

I think it's enough to change `src/assets/VueAsset.js`
```js
    let supplemental = `
      var ${optsVar} = exports.default || module.exports;
      if (typeof ${optsVar} === 'function') {
        ${optsVar} = ${optsVar}.options;
      }
    `;

    supplemental += this.compileTemplate(generated, scopeId, optsVar);
    supplemental += this.compileCSSModules(generated, optsVar);
    supplemental += this.compileHMR(generated, optsVar);
```
to
```js
    let supplemental = `
      (function(exports, module){
      var ${optsVar} = exports.default || module.exports;
      if (typeof ${optsVar} === 'function') {
        ${optsVar} = ${optsVar}.options;
      }
    `;

    supplemental += this.compileTemplate(generated, scopeId, optsVar);
    supplemental += this.compileCSSModules(generated, optsVar);
    supplemental += this.compileHMR(generated, optsVar);
    supplemental += "})(exports, module);"
```
Notice the added IIFE.

### LINKED ISSUE #1333: Vue SFC code splitting
# 🙋 feature request

I would like to be able to split a Vue Single File Component (SFC) into separate files:

- Template
- Script
- Style

This is briefly documented here: [What About Separation of Concerns?](https://vuejs.org/v2/guide/single-file-components.html#What-About-Separation-of-Concerns)

## 🤔 Expected Behavior

The script/style file should be built by Parcel.

## 😯 Current Behavior

I have tried the method documented on the link above, but the script/style file is not built by Parcel, and it is not included in the output.

## 💁 Possible Solution

I believe the solution is either obvious or someone will know better than I do.

When a `script` tag has a `src` attribute, then the script file (JavaScript, TypeScript, etc) should be built by Parcel. I am not sure if the embedded script should be ignored in this case.

When a `style` tag has a `src` attribute, then the style file (CSS, LESS, SASS/SCSS, etc) should be built by Parcel. I am not sure if the embedded styles should be ignored in this case.

## 🔦 Context

I started using Parcel, Vue.js and vue-router recently, and I organize my Vue components as Single File Components (SFC) with the template, script and styles in these `*.vue` files due to its simplicity. I do not know the best practices, but as some of my `*.vue` files are growing in length, I find it more inconvenient to navigate between the template/script/style sections.

I also see no reason to not support this feature because it is documented usage of Vue.js.

If there are no good alternatives, then I also see no reason to not support this feature, because separation is oftentimes a good thing.

Different people can work on the template, script and styles without also crossing the other "layers".

## 💻 Examples

Taken from [What About Separation of Concerns?](https://vuejs.org/v2/guide/single-file-components.html#What-About-Separation-of-Concerns):
```
<!-- my-component.vue -->
<template>
  <div>This will be pre-compiled</div>
</template>
<script src="./my-component.js"></script>
<style src="./my-component.css"></style>
```

`my-component.vue` contains mostly the template; `./my-component.js` contains the script; `./my-component.css` contains the styles.

### LINKED ISSUE #2721: Vue SFC: missing CSS sourcemaps
# 🐛 bug report

No CSS sourcemaps are genereted for Vue Single File Components.

## 💁 Possible Solution

The css concatenation here has to be handled correctly:
https://github.com/parcel-bundler/parcel/blob/f68f0df06e3af97fe3ff21061f182a272f6c069d/packages/core/parcel-bundler/src/assets/VueAsset.js#L129-L135
https://github.com/parcel-bundler/parcel/blob/f68f0df06e3af97fe3ff21061f182a272f6c069d/packages/core/parcel-bundler/src/assets/VueAsset.js#L220-L225

## 💻 Code Sample

```html
<!-- app.vue -->
<template lang="pug">
  .container Hello {{bundler}}
</template>

<script lang="ts">
import Vue from "vue";

export default Vue.extend({
  data() {
    return {
      bundler: "Parcel"
    };
  }
});
</script>

<style lang="scss" scoped>
.container {
  color: green;
}
</style>
```

```js
// index.js
import Vue from 'vue';
import App from './app.vue';

new Vue(App).$mount('#app')
```

```html
<!-- index.html -->
<div id="app"></div>
```
## 🌍 Your Environment

| Parcel           |master
| Node             |11
| npm/Yarn         |
| Operating System |macOS

### LINKED ISSUE #3364: Parcel 2: Vue transformer
Create the `@parcel/transformer-vue` package in `packages/transforms/vue`. This package is a Parcel 2 transform plugin for Vue. It should be based on the `VueAsset` from Parcel 1.

The `@parcel/transformer-vue` package is responsible for the following things:

* Parsing Vue SFCs with `@vue/component-compiler-utils`
* Extracting scripts, styles, and templates as child assets
* Compile templates
* Compile CSS modules
* Compile scoped styles
* Insert HMR runtime
* Combine the results into JS and CSS child assets

Additional features above what Parcel 1 offers to consider:

* Support for external script and style tags (#1333)
* Support for CSS source maps (#2721)
* Remove minification from the transformer - it will be handled by the optimizer, and sometimes causes issues like #1183

101arrowz:
# ↪️ Pull Request

Uses the Vue v3 `@vue/compiler-sfc` API.
Closes #3364
Closes #1333
Fixes #2721
Fixes #1183
Heavily inspired by [DarioDaF](https://github.com/parcel-bundler/parcel/issues/3364#issuecomment-598786210)'s transformer.

WORKING:
- Core functionality
- Any template/scripting/styling languages
- Scoped CSS
- Requiring other SFCs
- Autoinstalling missing template/scripting/styling packages
- CSS Modules (including custom names)
- Custom blocks

BUGS:
- SFC is parsed multiple times because Parcel considers different pipelines to be totally separate (`canReuseAST` is never called as a result)

UNTESTED:
- HMR (likely working but not sure how to test it vs. standard HMR)
- `vue-router`, `vuex`, etc. (though these should work just fine since they're merely JavaScript)

## 💻 Examples

See `vue` integration tests

## ✔️ PR Todo


101arrowz:
@mischnic I wrote a quick integration test `vue-full`, and it has started breaking in the rewrite with named pipelines because for some reason, the assets imported from the JS code generated by the transformer aren't added to the asset graph. Could you take a look?

101arrowz:
Any ideas on how to fix the main bug (the SFC is parsed once per pipeline, which means parsing each SFC multiple 
times)? @mischnic

101arrowz:
Opening for review because the transformer supports most operations. I still need to fix the tests, but I could definitely use some suggestions as to how to test this because unlike before, `run(bundleGraph)` won't work because Vue references browser globals.

packages/transformers/vue/src/VueTransformer.js
```diff
@@ -0,0 +1,358 @@
...
+import ThrowableDiagnostic from '@parcel/diagnostic';
+import type {Diagnostic} from '@parcel/diagnostic';
+import type {TransformerResult} from '@parcel/types';
+import SourceMap from '@parcel/source-map';
+import semver from 'semver';
+import {basename, extname} from 'path';
+
+const MODULE_BY_NAME_RE = /\.module\./;
+
+// TODO: flow
```
DeMoorJasper, line 14: I assume this TODO can be removed?

packages/transformers/vue/src/VueTransformer.js
```diff
@@ -0,0 +1,358 @@
...
+export default (new Transformer({
+  canReuseAST({ast}) {
+    return ast.type === 'vue' && semver.satisfies(ast.version, '3.0.0-beta.20');
+  },
+  async parse({asset, options}) {
+    // TODO: This parses the vue component multiple times. Fix?
```
DeMoorJasper, line 20: Not sure if we can work around this, maybe you could use the `postProcess` function for some of this?

packages/transformers/vue/src/VueTransformer.js
101arrowz: Yes, probably, but most variables are untyped; not sure if that's OK, but I'll remove it anyway.

DeMoorJasper: As long as there's no `$FlowFixMe` all over the place it's fine

packages/transformers/vue/src/VueTransformer.js
101arrowz: All this really does is parse the SFC using Vue's API, so moving it to `postProcess` can't work. I''ll leave this comment as a future optimization.

101arrowz:
I've fixed the tests, all of them pass (according to CI, not Windows Node 12, but that seems like a bug to me).

Right now, the transformer is almost at 1:1 feature parity with `vue-loader` (just a few things, like custom block support, are missing). However, for the VAST majority of use cases, the current feature set will do.

I'm also wondering if making the transformer configurable (say with a `.vuerc`) would be useful.

101arrowz:
Are there any specific features that should be added or is the current feature-set fine?

101arrowz:
I've added support for custom blocks but the syntax is quite different from Webpack's `vue-loader`.

In the Vue config (`.vuerc`, `vue.config.js`, etc.):
```json
{
    "customBlocks": {
        "docs": "./src/docs-preprocessor.js"
    }
}
```
In `src/docs-preprocessor.js`:
```js
export default function(component, blockContent, blockAttrs) {
    if (blockAttrs.brief)
        component.__briefDocs = blockContent;
    else
        component.__docs = blockContent;
}
```
In any Vue component (say `HomePage.vue`):
```vue
<template>
  <div>Home Page</div>
</template>

<docs>
This component represents the home page of the application.
</docs>

<docs brief>
Home Page
</docs>
```
In a parent component (say `App.vue`):
```vue
<template>
  <div>
    <child></child>
    docs: {{docs.standard}}
    in brief: {{docs.brief}}
  </div>
</template>

<script>
import Child from './HomePage';
export default {
  components: {
    child: Child
  },
  data() {
    return {
      docs: {
        standard: Child.__docs,
        brief: Child.__docsBrief
      }
    }
  }
}
</script>
```

101arrowz:
I've added support for custom blocks but the syntax is quite different from Webpack's `vue-loader`.

In the Vue config (`.vuerc`, `vue.config.js`, etc.):
```json
{
    "customBlocks": {
        "docs": "./src/docs-preprocessor.js"
    }
}
```
In `src/docs-preprocessor.js`:
```js
export default function(component, blockContent, blockAttrs) {
    if (blockAttrs.brief)
        component.__briefDocs = blockContent;
    else
        component.__docs = blockContent;
}
```
In any Vue component (say `HomePage.vue`):
```vue
<template>
  <div>Home Page</div>
</template>

<docs>
This component represents the home page of the application.
</docs>

<docs brief>
Home Page
</docs>
```
In a parent component (say `App.vue`):
```vue
<template>
  <div>
    <child></child>
    docs: {{docs.standard}}
    in brief: {{docs.brief}}
  </div>
</template>

<script>
import Child from './HomePage';
export default {
  components: {
    child: Child
  },
  data() {
    return {
      docs: {
        standard: Child.__docs,
        brief: Child.__docsBrief
      }
    }
  }
}
</script>
```

At this point, the transformer has full feature parity with `vue-loader` and is ready for merging. @DeMoorJasper

DeMoorJasper:
@101arrowz the CI is probably a new bug either discovered or caused by this PR.
Might look into it later.
The PR in general looks really good.

101arrowz:
Somehow, allowing `.vue` extensions to [resolve without an extension](https://github.com/parcel-bundler/parcel/pull/4892/commits/e4ebfc8f96f005d4a9d96c5e911b5b79be03caac#diff-55ff0e36af2108f79d8af002e7dc6841R29) is the bug that caused HMR to fail on macOS. I think you might want to take a look into the Parcel core because that seems like (very) unintended behavior.

PR #4912, opened 2020-07-18
Reload css from same origin

jongacnik:
# ↪️ Pull Request

This PR updates the check in `reloadCSS()` which determines whether a css resource is an absolute url or not, to include urls which originate from the same origin as the window. The motivation is to allow hot reloading functionality to continue to work on css resources which are of the same origin as your application. This can often be the case when using `parcel watch`.

## 💻 Examples

This PR will allow this to work with css reloading:

```html
<link rel="stylesheet" href="http://localhost:9001/assets/bundle.css" />
```

Whereas currently, you must include css assets using a relative url in order to get css reloading:

```html
<link rel="stylesheet" href="/assets/bundle.css" />
```

### DIFF
--- a/packages/runtimes/hmr/src/loaders/hmr-runtime.js
+++ b/packages/runtimes/hmr/src/loaders/hmr-runtime.js
@@ -197,8 +197,10 @@ function reloadCSS() {
   cssTimeout = setTimeout(function() {
     var links = document.querySelectorAll('link[rel="stylesheet"]');
     for (var i = 0; i < links.length; i++) {
-      var href = links[i].getAttribute('href')
-      var absolute = /^https?:\/\//i.test(href) && href.indexOf(window.location.href) !== 0
+      var href = links[i].getAttribute('href');
+      var absolute =
+        /^https?:\/\//i.test(href) &&
+        href.indexOf(window.location.origin) !== 0;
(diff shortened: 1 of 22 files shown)


DeMoorJasper:
Great work, could you run `yarn lint` and fix the issues that come up. So CI can pass?

packages/runtimes/hmr/src/loaders/hmr-runtime.js
```diff
@@ -197,7 +197,8 @@ function reloadCSS() {
   cssTimeout = setTimeout(function() {
     var links = document.querySelectorAll('link[rel="stylesheet"]');
     for (var i = 0; i < links.length; i++) {
-      var absolute = /^https?:\/\//i.test(links[i].getAttribute('href'));
+      var href = links[i].getAttribute('href')
+      var absolute = /^https?:\/\//i.test(href) && href.indexOf(window.location.href) !== 0
```
mischnic, line 201: Shouldn't this rather be `href.indexOf(window.location.origin) !== 0`?

jongacnik: Yes, it should, good catch. Updated and `yarn format`ed.

PR #4918, opened 2020-07-21
Autoinstall missing styling packages in Vue

101arrowz:
# ↪️ Pull Request

Complete #4892 by autoinstalling missing styling packages.

## ✔️ PR Todo


packages/transformers/vue/src/VueTransformer.js
```diff
@@ -337,6 +338,36 @@ ${
...
+                diagnostic: {
+                  message: `Unknown style language: "${style.lang}"`,
+                  origin: '@parcel/transformer-vue',
+                  filePath: asset.filePath,
+                },
+              });
+          }
+          if (options.autoinstall && toInstall) {
+            await options.packageManager.require(toInstall, asset.filePath, {
+              autoinstall: true,
```
mischnic, line 368: If you remove `options.autoinstall from the condition` and the value directly to the package manager, an error would be thrown.

I'm guessing `compileStyleAsync` fails as well when the package is missing?

101arrowz: Yes. If you'd like I could do what you are suggesting instead, since a Parcel diagnostic would be thrown with your suggestion as opposed to a standard error from `compileStyleAsync`.

Issue #3408, opened 2019-08-15
Add support for pnpm 


anhnhoktvn:
# 🙋 Feature request

I'd like to replace npm/yarn with pnpm.

## 🤔 Expected Behavior
Parcel can use pnpm to install packages.

## 😯 Current Behavior
Parcel can not use pnpm to install packages.

## 💁 Possible Solution
Have a config to use pnpm package managers.

## 💻 Examples
[https://github.com/pnpm/pnpm](https://github.com/pnpm/pnpm)

devongovett:
We're working on abstracting the package manager interface in v2, so it should be possible to pass a custom implementation to the Parcel API to use a custom package manager. We can look into supporting pnpm by default as well.

nthypes:
This should be closed since https://github.com/parcel-bundler/parcel/pull/3471 or only after official v2 release?

mischnic:
That PR only refactored the code, it didn't add a pnpm "backend".

PR #4920, opened 2020-07-22
Package manager pnpm

dishuostec:
# ↪️ Pull Request

Try to add [pnpm](https://github.com/pnpm/pnpm) suport

#3408


packages/core/package-manager/src/installPackage.js
```diff
@@ -128,13 +129,19 @@ async function determinePackageInstaller(
...
     'package-lock.json',
+    'pnpm-lock.yaml',
   ]);
   let hasYarn = await Yarn.exists();
 
   // If Yarn isn't available, or there is a package-lock.json file, use npm.
   let configName = configFile && path.basename(configFile);
-  if (!hasYarn || configName === 'package-lock.json') {
-    return new Npm();
+  if (!hasYarn || configName) {
```
101arrowz, line 138: If the Yarn binary doesn't exist and no lockfile is present, NPM should be used by default. However, if there is no lockfile, this new logic will cause Yarn to be used even when the binary doesn't exist.

dishuostec: > could you add some integration tests? You can take inspiration from how we did it for yarn 2



> If the Yarn binary doesn't exist and no lockfile is present, NPM should be used by default. However, if there is no lockfile, this new logic will cause Yarn to be used even when the binary doesn't exist.

Yes, I made a mistake.

dishuostec:
> could you add some integration tests? You can take inspiration from how we did it for yarn 2

I couldn't find which part is the test for yarn2, do you mean `@parcel/integration-test`?

DeMoorJasper:
@dishuostec yes that's the package with all our integration tests.

For inspiration I meant https://github.com/parcel-bundler/parcel/blob/v2/packages/core/integration-tests/test/pnp.js which is part of our test suite

mischnic:
(The PnP tests are only testing the resolver, not the autoinstall part)

dishuostec:
Pnpm use symlinked node_modules structure. If a package is dependent on by several package, then the package will be resolved as different position. It will be bundle multiple times.

I created new pr https://github.com/parcel-bundler/parcel/pull/4939.

aminya:
Really looking forward to this. Having to manage many packages, the time and storage that pnpm saves me is a big advantage. However, the fact that Parcel does not work with pnpm is a bummer!

PR #4925, opened 2020-07-22
Use fastest-levenshtein instead of js-levenshtein

ka-weihe:
`fastest-levenshtein` ([link](https://github.com/ka-weihe/fastest-levenshtein)) is a lot faster than `js-levenshtein`. Please consider using it instead :slightly_smiling_face: 

### Benchmarks
N = length of string
Numbers are in ops/sec
Higher is better.

| Test Target               | N=4   | N=8   | N=16  | N=32 | N=64  | N=128 | N=256 | N=512 | N=1024 |
|---------------------------|-------|-------|-------|------|-------|-------|-------|-------|--------|
| fastest-levenshtein       | 44423 | 23702 | 10764 | 4595 | 1049  | 291.5 | 86.64 | 22.24 | 5.473  |
| js-levenshtein            | 21261 | 10030 | 2939  | 824  | 223   | 57.62 | 14.77 | 3.717 | 0.934  |

### Relative Performance
x-axis = "times faster compared to `js-levenshtein`"
Benchmark


DeMoorJasper:
Not sure if it matters for or use-case and maybe it's best to use the package that is used most?
We only really use this whenever an error occurs on a very small amount of strings so speed is not really that important.

ka-weihe:
Well, fastest-levenshtein is now used under the hood in the most used package (fast-levenshtein).

DeMoorJasper:
@ka-weihe awesome thanks for the update

PR #4926, opened 2020-07-22
Use TSC by default when tsconfig is present

### LINKED ISSUE #4022: Parcel 2: TypeScript automatic zero configuration not type checking or reading tsconfig.json
I've been following #1378 for a long time, and I'm thrilled to see Parcel 2 taking on better first-class support for TypeScript. There's currently some problems with Parcel 2 out of the box, and I didn't see any threads tracking this, so I'm making an issue.

# 🐛 bug report

In 2.0.0-alpha.3.2, given a very basic TypeScript project, there are three issues: 

- the `typescript` package is not automatically installed
- settings in `tsconfig.json` are completely ignored
- typescript errors are silently ignored and compiled anyway

Here's all of the files:

- `package.json` created from two commands:
    - `yarn init -y`
    - `yarn install -D parcel@v2`
- `index.html`
	```html
    <!DOCTYPE html>
    <html lang="en">
      <head>
        <meta charset="utf-8" />
        <title>My First Parcel App</title>
        <script src="index.ts"></script>
      </head>
      <body>
        <h1>Hello, World!</h1>
      </body>
    </html>
	```
- `index.ts`
	```typescript
	console.log("Hello, von Nuemann!")
	const failMe: BadType = 5; // this should fail, but it compiles
	```
- optionally, I added a junk `tsconfig.json` to test, and it's obviously being ignored because parcel still compiles:
    ```json
	{
	  ALDIFJ SLDIFJ :LEIRJ LWEIJR;
	}}}}}
      "compilerOptions": {
    ```

## 🤔 Expected Behavior

As part of parcels "zero configuration", parcel automatically detects TypeScript and configures the project to correctly use TypeScript.

- TypeScript compilation or validation should fail by default when there are type errors
- `typescript` package should be installed automatically
- `tsconfig.json` should be read and used by whatever is compiling it, if it exists

## 😯 Current Behavior

- the `typescript` package is not automatically installed
- settings in `tsconfig.json` are completely ignored
- typescript errors are silently ignored and compiled anyway

## 💁 Possible Solution

I'm guessing one of these two potential solutions:

2. The current default `transformer-babel` is correctly configured to behave as expected with TypeScript, extending from #3083
1. `typescript` and `@parcel/transformer-typescript(-tsc?)` is automatically installed when Parcel 2 detects a `.ts` file.

Either one is fine, I think, as long as the TypeScript compilation actually behaves like TypeScript by default!

## 🔦 Context

- Seems to have started in this PR that defaults to `babel` for TypeScript: https://github.com/parcel-bundler/parcel/pull/3083
- Type checking seems to have been merged, but it's not happening: https://github.com/parcel-bundler/parcel/pull/3232

## 💻 Code Sample

See files above

## 🌍 Your Environment

| Parcel           | 2.0.0-alpha.3.2 |
| Node             | v10.15.3 |
| npm/Yarn         | 1.21.1 | 
| Operating System | Mac 10.14.6 |

101arrowz:
# ↪️ Pull Request

(NOT COMPLETE) Typechecks and compiles with `tsc` for compiling when `tsconfig.json` present, and babel otherwise. This allows for rapid builds out of the box (and for benchmarks) but safer, typechecking builds otherwise.

I think this is as close to a middle ground as is possible - Those who want zero-config and high speed can use babel, and those who want little config with high safety can have that too with `tsc`.

Note that the test cases were cut down quite a bit.

## ✔️ PR Todo

### DIFF
--- a/packages/configs/default/index.json
+++ b/packages/configs/default/index.json
@@ -4,12 +4,13 @@
     "types:*.{ts,tsx}": ["@parcel/transformer-typescript-types"],
     "bundle-text:*": ["@parcel/transformer-inline-string", "..."],
     "data-url:*": ["@parcel/transformer-inline-string", "..."],
-    "*.{js,mjs,jsm,jsx,es6,cjs,ts,tsx}": [
+    "*.{js,mjs,jsm,jsx,es6,cjs,ts-babel,tsx-babel}": [
       "@parcel/transformer-react-refresh-babel",
       "@parcel/transformer-babel",
       "@parcel/transformer-js",
       "@parcel/transformer-react-refresh-wrap"
     ],
+    "*.{ts,tsx}": ["@parcel/transformer-typescript-tsc"],
     "*.{json,json5}": ["@parcel/transformer-json"],
     "*.jsonld": ["@parcel/transformer-jsonld"],
     "*.toml": ["@parcel/transformer-toml"],
--- a/packages/configs/default/package.json
+++ b/packages/configs/default/package.json
@@ -59,6 +59,7 @@
     "@parcel/transformer-stylus": "2.0.0-beta.1",
     "@parcel/transformer-sugarss": "2.0.0-beta.1",
     "@parcel/transformer-toml": "2.0.0-beta.1",
+    "@parcel/transformer-typescript-tsc": "2.0.0-beta.1",
     "@parcel/transformer-typescript-types": "2.0.0-beta.1",
     "@parcel/transformer-vue": "2.0.0-beta.1",
     "@parcel/transformer-yaml": "2.0.0-beta.1"
--- a/packages/core/integration-tests/test/typescript.js
+++ b/packages/core/integration-tests/test/typescript.js
@@ -10,199 +10,162 @@ import {
 } from '@parcel/test-utils';
 import {readFileSync} from 'fs';
 
-const configPath = path.join(
-  __dirname,
-  '/integration/typescript-config/.parcelrc',
-);
-
-const tscConfig = {
-  ...JSON.parse(readFileSync(configPath)),
-  filePath: configPath,
-};
-
 describe('typescript', function() {
-  // This tests both the Babel transformer implementation of typescript (which
-  // powers typescript by default in Parcel) as well as through the Typescript
-  // tsc transformer. Use a null config to indicate the default config, and the
-  // tsc config to use the tsc transformer instead.
-  //
-  // If testing details specific to either implementation, create another suite.
-  for (let config of [
-    null /* default config -- testing babel typescript */,
-    tscConfig,
-  ]) {
-    it('should produce a ts bundle using ES6 imports', async function() {
-      let b = await bundle(
-        path.join(__dirname, '/integration/typescript/index.ts'),
-        {config},
-      );
+  it('should produce a ts bundle using ES6 imports', async function() {
+    let b = await bundle(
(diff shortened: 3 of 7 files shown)


mischnic:
The TSC transformer doesn't do typechecking.

101arrowz:
In that case, I'll try to convert the TSC transformer to a typechecking one, because having the issues emitted after the build completes seems suboptimal. Sorry for the confusion.

packages/configs/default/index.json
```diff
@@ -4,12 +4,13 @@
     "types:*.{ts,tsx}": ["@parcel/transformer-typescript-types"],
     "bundle-text:*": ["@parcel/transformer-inline-string", "..."],
     "data-url:*": ["@parcel/transformer-inline-string", "..."],
-    "*.{js,mjs,jsm,jsx,es6,cjs,ts,tsx}": [
+    "*.{js,mjs,jsm,jsx,es6,cjs,ts-babel,tsx-babel}": [
```
DeMoorJasper, line 7: Not a fan of this `ts-babel` extension, seems like a hack. These transformers should not depend on each other, it kind of defeats the purpose of having the plugins as seperate entities.

DeMoorJasper: Note that this would also just end up with using tsc by default on 99% of Typescript projects as every typescript project should have a tsconfig anyway... otherwise why are you using Typescript in the first place if you don't do typechecking...

101arrowz: Yeah I honestly don't want to do this either. But there's literally no other way to make babel process it instead of TSC in such situations.

Typechecking must be integrated into the bundler for a variety of reasons, and one of them is that Vue TypeScript can't be processed.

I'm considering just dropping this PR altogether but I really, really want TypeScript to be checked by default before the build finishes.

101arrowz: I know that most TypeScript projects would end up using `tsc` because of this but that's kind of the point; few TypeScript users actually want to use Babel, but Parcel can still be called "blazing-fast" if you decide to use Babel.

DeMoorJasper: Like @mischnic already mentioned adding `@parcel/transformer-typescript-tsc` won't do anything more than using Babel, typechecking is done by the validator plugin... and it also doesn't support vue (yet)

DeMoorJasper: Also Devon mentioned in the Typescript issue even if we wanted to enable type-checking by default the performance and reliability of our typechecking validator is not good enough to do so, mainly because tsc isn't really built for what we use it but we can work around most of that.

Fixing most issues with the typechecker diagnostics plugin seems more important than this, which is just a config change and removal of tests.

So if you really want this to be the default, help us ensure our typescript transformer and validator are top-notch so we don't have to worry about the various bugs we encountered in Parcel 1 and performance issues we might experience now.

The transformer mainly had too much freedom in Parcel 1 and probably still does in Parcel 2, we should ignore probably 90% of compiler options and only whitelist a few we want to allow, so it does not break any further compilation down the line.

Besides this we also want to support the paths field in jsconfig and tsconfig in our resolver at some point and want typechecking to be faster. We've even thought about a command like `parcel validate` to just run the validators without any compilation to run linting and typechecking a lot faster than the tools can do themselves by only running over changed files

101arrowz: I'll make more useful contributions instead of this PR given your feedback. Closing.

PR #4931, opened 2020-07-24
Don't remove subgraphs that are still needed for other dependencies in the bundle

wbinnssmith:
This has `removeAssetGraphFromBundle` stop if it encounters an asset where it and its subgraph are still needed by other dependencies in the bundle. Thanks to @devongovett for the idea of using a first pass for identifying subgraph dependency nodes.

This is described by this scenario drawn by @mischnic, where either asset 2 or asset 3 is removed, removing asset 1 along with it, which is necessary for the other:

### DIFF
--- a/packages/core/core/src/BundleGraph.js
+++ b/packages/core/core/src/BundleGraph.js
@@ -860,7 +860,8 @@ export default class BundleGraph {
     }
 
     return this._graph
-      .findAncestors(node, node => node.type === 'dependency')
+      .getNodesConnectedTo(node)
+      .filter(node => node.type === 'dependency')
       .map(node => {
         invariant(node.type === 'dependency');
         return node.value;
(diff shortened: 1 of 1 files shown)


packages/core/core/src/BundleGraph.js
```diff
@@ -860,7 +896,8 @@ export default class BundleGraph {
     }
 
     return this._graph
-      .findAncestors(node, node => node.type === 'dependency')
+      .getNodesConnectedTo(node)
```
mischnic, line 899: Doesn't this cause e.g. child bundle entries to have no incoming dependencies because the dependency itself points to the bundle(group)?

wbinnssmith: Nothing seems to be depending on that right now. Would you say that's expected behavior?

mischnic: getIncomingDependencies is used in very few places at the moment, but I would say it's expected behavior. Otherwise, there is no way to get the incoming deps for a child bundle entry via the public `BundleGraph`.

PR #4935, opened 2020-07-26
Fix Vue HMR

### LINKED ISSUE #5325: No transformers found for .vue with Vue 3 and Parcel 2
# 🐛 bug report

This seems like a rookie mistake, but I'm getting this error.
```
🚨 Build failed.
Error: No transformers found for "src/js/components/App.vue".
```

```js
  "scripts": {
    "dev": "parcel src/html/index.html",
  },
  "dependencies": {
    "parcel": "^2.0.0-beta.1",
    "postcss-modules": "^3.2.2",
    "tailwindcss": "^1.9.6",
    "vue": "^3.0.2"
  },
  "devDependencies": {
    "sass": "^1.28.0"
  }
```

## 😯 Current Behavior
It seems like the basic doesn't work for me right now, I think I have double checked everything but can't figure out what's wrong.

## 🔦 Context
Trying to get started with Parcel 2 and Vue 3.

## 💻 Code Sample

```html
<!DOCTYPE html>
<html lang="en">
<head>
  <meta charset="utf-8" />
  <title>My First Parcel App</title>
  <link rel="stylesheet" href="../css/index.pcss">
</head>
<body class="font-body">

<div id="app"></div>

<script src="../js/index.js"></script>
</body>
</html>
```

index.js
```js
import { createApp } from 'vue'
import App from './components/App.vue'

createApp(App).mount('#app')
```

App.vue
```vue
<template>
  <div class="container mx-auto py-4">

    <a href="#" class="btn">Test button</a>
  </div>
</template>
<script>

export default {
}
</script>

```

101arrowz:
# ↪️ Pull Request

Sorry for the constant PRs.

Vue HMR didn't work previously, it's now been partially fixed. State isn't preserved but unfortunately at the moment it doesn't seem to be possible to do so due to some limitations with Parcel's HMR API. I find the behavior of hot-reloading assets before their dependencies peculiar, but I worked around it with an (admittedly hacky) `setTimeout(fn, 0)`.

Closes #5325


101arrowz:
I realized that HMR isn't actually working properly. I wish Parcel's HMR API were the same as webpack's. Also, if assets are supposed to reload after their dependencies, then I found a bug. All I did is `console.log` in the `module.hot.accept` callbacks for all of my assets, and the dependents updated before the dependencies.

mischnic:
> I wish Parcel's HMR API were the same as webpack's, it would make feature parity much easier

What's missing? 

> All I did is console.log in the module.hot.accept callbacks *for all of my assets*, and the dependents updated before the dependencies

But if you accept an update in a module, it doesn't bubble up?

I'll try to dig up the Webpack HMR test suite, maybe we can run that. I'm not even sure myself how HMR should behave exactly

101arrowz:
In Webpack, you can [specify which dependencies you want to watch](https://webpack.js.org/api/hot-module-replacement/) for a specific `module.hot.accept` callback. That may seem trivial to work around, but for managing local state it can be extremely annoying.

mischnic:
Ah, I thought we already did that. Looking into it now

101arrowz:
Another thing that's making HMR difficult is that Parcel doesn't accurately detect which assets have changed. Whenever even a single line of, for example, the `style` block of a `.vue` file changes, the assets associated with all of its components (including, for example, the `script` block) are "updated" by HMR, which makes it impossible to distinguish between a change in logic and style during HMR.

101arrowz:
@mischnic Any updates?

mischnic:
I've started in this branch but ran out of time (there a TODO comment where I stopped).
https://github.com/parcel-bundler/parcel/compare/webpack-hmr-api
You can continue from there if you want

101arrowz:
I've worked on expanding that branch (finishing `.accept` and adding `.decline`), but the reported changed assets by the HMR server is still not accurate. Even if a source asset changes, the resulting generated assets may not all change, but `AssetGraphBuilder` assumes that all assets are modified. Because of this, changes in template vs. script are indistinguishable for `.vue`.

mischnic:
Aligning the API with webpack (which is the client-side in the runtime) and this Vue HMR problem (server-side) are two separate topics.

This is the first transformer that extensively uses multiple child asset (apart from PostCSS CSS modules)...

101arrowz:
I'm aware, but I'm just mentioning that the issue persists in context of this PR. Unfortunately, it looks like there's quite a long way to go before Vue HMR will be fully possible.

mischnic:
(I think the `changedAssets` behaviour shouldn't be too hard to fix)

101arrowz:
I'm going to open for review because Vue doesn't properly reload during development, and this, at least for now, fixes that (even if it's not real HMR). I'm not adding a test because this isn't actual HMR yet + I'm lazy.

packages/transformers/vue/src/VueTransformer.js
```diff
@@ -394,7 +390,7 @@ ${
           let styleAsset = {
             type: 'css',
             content: styleComp.code,
-            sideEffects: !style.module,
+            sideEffects: true,
```
101arrowz, line 393: Not sure if this is correct...

mischnic: 👍  I can't think of a case where a CSS asset should be ever have `sideEffects: false`.

KaKi87:
Hello @mischnic, please review and merge so that we can all benefit from this fix, thanks :)

mischnic:
Are there any Vue HMR tests? (https://github.com/parcel-bundler/parcel/pull/4935#issuecomment-663957170)

KaKi87:
Oh I didn't see that you were waiting for additional elements and I was only interested about resolving #5325, I'm sorry for the disturbance.

mischnic:
> you were waiting for additional elements

(No worries, even I forgot about that)

101arrowz:
@mischnic Is a test really necessary? Despite the title, this isn't actual HMR - state is not preserved between updates. Adding a test for it is pretty much pointless (unless you want to skip it by default and enable it once HMR is fully fixed).

mischnic (approved):
Alright

PR #4939, opened 2020-07-27
Resolve realpath

dishuostec:
# ↪️ Pull Request

Use realpath when resolving module.

## 💻 Examples

Support [symlinked `node_modules` structure used by pnpm](https://pnpm.js.org/en/symlinked-node-modules-structure).

## ✔️ PR Todo


mischnic:
Git doesn't create symlinks on Windows by default, so committing them doesn't really work

dishuostec:
> Git doesn't create symlinks on Windows by default, so committing them doesn't really work

I should create symlinks before test and then remove.
It seems pnpm use [`symlink-dir`](https://www.npmjs.com/package/symlink-dir) to create junction on windows. Should I import that package or use `fs.symlink`?

mischnic:
So far, we've either wrapped these `fs.symlink` calls with a try catch to skip that test if the symlink creation failed: https://github.com/parcel-bundler/parcel/blob/1bb0498339087cea54b1044d56279813293f00a2/packages/core/parcel-bundler/test/resolver.js#L37
or do the whole thing in the `MemoryFS` abstraction (which would be ideal because sometimes filesystem based tests are flakey): https://github.com/parcel-bundler/parcel/blob/da42ce98ce61d7c6dd6cd62719da69fc58bfb999/packages/core/integration-tests/test/ts-types.js#L203-L206

dishuostec:
I used `fs.symlink` to create junction for windows, it passed.

jasonswearingen:
@mischnic could we move forward with a review of this?  

Sorry for the bump, I don't know a better way to ping on this.

mischnic:
In case you haven't noticed: we've added resolver unit tests since you opened this PR, and many of them are now failing

packages/utils/node-resolver-core/src/NodeResolver.js
```diff
@@ -538,11 +538,12 @@ export default class NodeResolver {
...
         let stats = await this.fs.stat(moduleDir);
         if (stats.isDirectory()) {
+          const realModuleDir = this.options.inputFS.realpathSync(moduleDir);
           return {
             moduleName: moduleName,
             subPath: subPath,
-            moduleDir: moduleDir,
-            filePath: path.join(dir, 'node_modules', filename),
+            moduleDir: realModuleDir,
+            filePath: path.join(realModuleDir, subPath || ''),
```
aminya, line 546:
```suggestion
            moduleDir: realModuleDir ?? moduleDir,
            filePath: path.join(realModuleDir ?? moduleDir, subPath || ''),
```

aminya: @mischnic This might solve the issue. The fix can return to its traditional mode when `realModuleDir` is not found for some reason.

packages/utils/node-resolver-core/src/NodeResolver.js
```diff
@@ -538,11 +538,12 @@ export default class NodeResolver {
         let moduleDir = path.join(dir, 'node_modules', moduleName);
         let stats = await this.fs.stat(moduleDir);
         if (stats.isDirectory()) {
+          const realModuleDir = this.options.inputFS.realpathSync(moduleDir);
```
aminya, line 541:
```suggestion
		  let realModuleDir = moduleDir;
		  try {
		    realModuleDir = this.options.inputFS.realpathSync(moduleDir);
		  } catch (error) {
		  
		  }
```

aminya:
I ran the tests locally after applying my suggestions and all of them pass now:

image

Previously:
image

aminya:
@DeMoorJasper There is also https://github.com/parcel-bundler/parcel/pull/4939#discussion_r475326924 😁

aminya:
Now the unit tests pass, but the integration tests fail. The issue I was trying to solve was that `this.options` is `undefined` sometimes. If you know the root of the cause, it would be better to solve that too.

packages/utils/node-resolver-core/src/NodeResolver.js
```diff
@@ -538,11 +538,13 @@ export default class NodeResolver {
         let moduleDir = path.join(dir, 'node_modules', moduleName);
         let stats = await this.fs.stat(moduleDir);
         if (stats.isDirectory()) {
+		      const realModuleDir = this.options?.inputFS.realpathSync(moduleDir);
```
aminya, line 541:
```suggestion
          const realModuleDir = this.fs.realpathSync(moduleDir);
```

OK found it. At some point, the names of the variables changed. That's why `this.options` does not exist anymore.

aminya:
By applying the changes, the actual issue is revealed. These failing tests seem to conflict with this PR
image

aminya:
Any thoughts on [this](https://github.com/parcel-bundler/parcel/pull/4939#issuecomment-684463349)? As I said current tests conflict with what this PR does. To fix them, we should change the behavior of `source field` tests.

mischnic:
The original reason was that packages in a monorepo (which are symlinked in node_modules) should always be bundled via the `source` entry

aminya:
@mischnic 

So what will be the solution? Does this mean that we should add an option to detect monorepos?

 This PR adds pnpm support #4920, and I am really looking forward to it. I want to understand the relation of this PR and that one.

aminya:
Can we get this revived?

@mischnic I am still awaiting your response to my question in https://github.com/parcel-bundler/parcel/pull/4939#issuecomment-687660743

mischnic:
I'm investigating how we can support both the `source` field and pnpm...

dishuostec:
Sorry guys, I have accidentally deleted the branch. 

:stuck_out_tongue_winking_eye:

dishuostec:
> By applying the changes, the actual issue is revealed. These failing tests seem to conflict with this PR. I don't think there should be a difference between the symlinked and non-symlined starting from this PR, and these unit tests should be changed.
> 
> image

I created a new pr https://github.com/parcel-bundler/parcel/pull/5183, and fixed these tests

Issue #3365, opened 2019-08-12
Parcel 2: Rust transformer


devongovett:
Create the `@parcel/transformer-rust` package in `packages/transforms/rust`. This package is a Parcel 2 transform plugin for Rust to WASM. It should be based on the `RustAsset` from Parcel 1.

The `@parcel/transformer-rust` package is responsible for the following things:

* Installing the rust toolchain if needed
* Resolving Cargo.toml and running cargo if available.
* Otherwise, running rustc
* Collecting dependencies from the rust compiler and adding them as connected files so Parcel watches them
* Generating WASM

It is possible that the current approach in Parcel 1 is out of date. Additional research on the current Rust/WASM ecosystem is needed here. Some related issues:

* Importing JS from WASM - #647
* ESM integration - #1325
* HMR - #2688

There are also some plugins that have been working on some related stuff:

* https://github.com/mysterycommand/parcel-plugin-wasm-pack
* https://github.com/catsigma/parcel-plugin-wasm.rs
* https://github.com/rustwasm/rust-parcel-template

If someone from the community would like to own this feature in Parcel 2 core, that would be very helpful and appreciated! 😍 

cc. @xtuc @linclark @littledan @alexcrichton - what's the current state of WASM + Rust + ESM? What should we be aiming for in Parcel 2?

talentlessguy:
Would be cool to see Golang transformer as well. It has good support for WebAssembly. Also Go now has modules which makes development easier (and I guess making a transformer will be easier)

Related issues:

#2903

devongovett:
@talentlessguy let's keep this issue about Rust. #2903 seems to cover Go well.

alexcrichton:
Thanks for the cc @devongovett! The current state of Rust/WASM/ESM is that tooling like `wasm-bindgen` and `wasm-pack` have a "bundler" target which is intended to be consumed by Parcel/Webpack/etc. This "bundler" target assumes that the output wasm file is interpreted as an ES module and follows the wasm ESM integration spec. 

From our perspective having Parcel act as a polyfill for the [wasm ESM integration spec](https://github.com/WebAssembly/esm-integration) would cover all our use cases!

devongovett:
Thanks for the info @alexcrichton. I'll take a look through that spec.

kellytk:
FWIW https://github.com/parcel-bundler/parcel/issues/2973 is another issue related to Rust and `wasm-pack`.

justinas:
[...]

Issue #3269, opened 2019-07-26
Parcel support for import.meta


KaliedaRik:
# ❔ Question

Are there any plans for Parcel to support the import.meta Javascript object in the near future?
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import.meta

## 🔦 Context

I have a requirement to use Javascript's import.meta in my module code, which I then use to build a url string to download and start a web worker. My code is entirely ES6 modular. 

## 💻 Code Sample

const buildFilterWorker = function () {

	let path = import.meta.url.slice(0, -('factory/filter.js'.length))
	let filterUrl = `${path}worker/filter.js`;
	return new Worker(filterUrl);
};

Running Parcel leads to the following build error:

$ parcel index.html
Server running at http://localhost:1234
🚨  factory/filter.js:191:19: Support for the experimental syntax 'importMeta' isn't currently enabled (191:20):
  189 | const buildFilterWorker = function () {
  190 |
> 191 | 	let path = import.meta.url.slice(0, -('factory/filter.js'.length))
      | 	                 ^

## 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Parcel           | 1.12.3
| Node             | 12.6.0
| npm/Yarn         | npm 6.9.0
| Operating System | MacOSX

mischnic:
> Support for the experimental syntax 'importMeta' isn't currently enabled

This isn't an error from Parcel but from babel. It seems like you have an older version of `@babel/parser` installed.

But I don't think it would work after it parsed successfully.

Related: https://github.com/webpack/webpack/issues/6719

KaliedaRik:
Many thanks for the speedy response and update.

I've been searching through the documentation, and the issues, and it appears that Parcel will run all code through Babel - Issue 1207 says this is a feature, not a bug (https://github.com/parcel-bundler/parcel/issues/1207).

As I'm not looking for my code to run through Babel - the library I'm building won't be supporting IE/Edge - I'm happy to let this issue be closed as "will not fix" - unless this is something you want to include in Parcel at some point in the future.

mischnic:
> it appears that Parcel will run all code through Babel

Yes. 

> unless this is something you want to include in Parcel at some point in the future.

Yes. `import.meta` is a stage 3 proposal and likely to become part of the standard.
Though I'm not quite sure (yet) whether this will require some explicit handling by Parcel or if the Babel behaviour will change.

[...]

PR #4970, opened 2020-08-03
Adds support for Rust/wasm assets via wasm-pack

mysterycommand:
# ↪️ Pull Request

This PR is a first implementation of support for Rust/wasm assets via wasm-pack in Parcel 2. It's based on https://github.com/mysterycommand/parcel-plugin-wasm-pack for Parcel 1 (which I also wrote). It has been discussed in https://github.com/parcel-bundler/parcel/discussions/4422#discussioncomment-23292 and in #3365.

## 💻 Examples

There are examples in `packages/examples/wasm-single-crate` and `packages/examples/wasm-multiple-crates`. They both use an `index.html` as the entry for `parcel build/serve`. The html adds a `js` script file which does something like `import {run} from '../Cargo.toml';`.

I updated [`packages/configs/default/index.json`](https://github.com/parcel-bundler/parcel/compare/v2...mysterycommand:v2-wasm-pack#diff-89039bac63e920d9ab1ea1bdf8d794ba) so that imports of `Cargo.toml` flow through a new [`WasmPackTransformer`](https://github.com/parcel-bundler/parcel/compare/v2...mysterycommand:v2-wasm-pack#diff-6f6b71ef3fcd1ce4730efbffd67d228a) which runs `wasm-pack` and generates a loader and some glue code to wire the wasm module and it's imports object together.

Probably the most "controversial" part of this PR are some changes to [`JSPackager`](https://github.com/parcel-bundler/parcel/compare/v2...mysterycommand:v2-wasm-pack#diff-660e8fe5d1c2ec7c9f720b52951595e4) that are based on [the original plugin implementation's `WasmPackPackager`](https://github.com/mysterycommand/parcel-plugin-wasm-pack/blob/master/src/WasmPackPackager.js) which [registered itself](https://github.com/mysterycommand/parcel-plugin-wasm-pack/blob/master/index.js#L7) to handle JS files in the plugin implementation. I wasn't sure if (now that I'm contributing to core) there was a better way/place to make these changes, but this seems to work at least … as a starting point.

## 🚨 Test instructions

I am very open to ways of making the integration test better/more meaningful … it's really just a scaffold right now that asserts the names of the generated bundle's assets. Weirdly, it appears that [one of the bundles includes the `JSRuntime` asset twice](https://github.com/parcel-bundler/parcel/compare/v2...mysterycommand:v2-wasm-pack#diff-85db19d3082a98226bb57566ceaba21bR34-R35) … is that normal/expected? Seems weird to me.

That said, both the examples mentioned above work, and navigating into one of those folders and running `yarn build` or `yarn start` will generate a valid bundle that logs something to the console from the wasm module.

## ✔️ PR Todo

- [ ] Added/updated unit tests for this change <-- need help!


devongovett:
This looks great! Thanks for working on it. Will look more in depth soon. 😄

mischnic:
Which of the problems that you ran into and mentioned in the various other issues are still relevant? 

> Weirdly, it appears that one of the bundles includes the JSRuntime asset twice … is that normal/expected? Seems weird to me.



You mean according to `--detailed-report` (or something similar)? That is normal, i.e. every dynamic import is replaced with a small asset that always has `JSRuntime.js` as it's path: https://github.com/parcel-bundler/parcel/blob/82747eb5778248d8db35186ec424687f91db5942/packages/runtimes/js/src/JSRuntime.js#L292

mysterycommand:
> Which of the problems that you ran into and mentioned in the various other issues are still relevant?

The issues described here: https://github.com/parcel-bundler/parcel/issues/3365#issuecomment-613046348 (apart from item 1) are largely unaddressed because I find myself agreeing with @qwerty2501's conclusion here: https://github.com/parcel-bundler/parcel/issues/3365#issuecomment-617828021.

@Pauan's comments here: https://github.com/parcel-bundler/parcel/issues/3365#issuecomment-613298005 are mostly addressed with the exceptions of the first and last items there. In the case of the first, I think I prefer to use wasm-pack's verbose output in Parcel's `logger.progress` when running `wasm-pack build` over a "crate". In the case of the last, I think unless we can target ES Modules we need to do some work on wasm-pack's output to fetch and instantiate the wasm module, and to create the imports object … indeed removing this boilerplate appears to be one of the key motivations behind [WebAssembly/ESM integration spec](https://github.com/WebAssembly/esm-integration/tree/master/proposals/esm-integration#motivation)

> You mean according to `--detailed-report` (or something similar)? That is normal, i.e. every dynamic import is replaced with a small asset that always has `JSRuntime.js` as it's path:

Pauan:
@mysterycommand It is not necessary to have ESM integration, instead you can use `--target web`. This is what the [Rollup plugin](https://github.com/wasm-tool/rollup-plugin-rust#readme) does. In that case the boilerplate, fetching, and imports object are all handled by wasm-pack.

mysterycommand:
@Pauan super interesting, thanks! You're reminding me of a couple of things that may or may not be issues for using `--target web` in the `wasm-pack build` args. In a first draft of `parcel-plugin-wasm-pack` I actually split the `wasm-pack --target` (`bundler`, `web`, `nodejs`, or `no-modules`) based on the `parcel --target` (`browser`, `electron`, or `node`).

If we could split it up this we we wouldn't need the loaders (you're right … in fact, even in the case of the Rollup plugin I think using `wasm-pack --target nodejs` would let you avoid adding the `loadFile` function to the asset prelude since the `wasm-pack`-generated entry does that for you as well in that case, though synchronously).

However, I think there's a couple things I'm not so jazzed about with this approach:

1. In Parcel, I believe JS is passed through Babel and I think it still needs [a plugin](https://babeljs.io/docs/en/babel-plugin-syntax-import-meta) for [`import.meta`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import.meta) (which is used by the `init` function generated by `wasm-pack --target web` … I guess you don't have this issue in Rollup since it has a more "ESM first" approach?
    - if we want to do it this way, maybe we can just auto-install that plugin?
2. In the case of multiple crates being compiled to wasm and imported separately, the `wasm-pack`-generated module includes a (duplicate) loader for each module. This is probably not a huge deal, just a personal annoyance of mine.
3. The usage I'd like to achieve involves loading the wasm, and swapping them out in the bundle cache before running the entry ([the "controversial" stuff in `JSPackager`](https://github.com/parcel-bundler/parcel/pull/4970/files#diff-660e8fe5d1c2ec7c9f720b52951595e4)) … all this to make it so that someone working with Rust/wasm can do like (for example):
    ```
    import { greet } from './Cargo.toml';

    greet();
    ```
    … or in the case of multiple crates like:
    ```
    import { hello } from './crates/hello/Cargo.toml';
    import { world } from './crates/world/Cargo.toml';

    hello();
    world();
    ```

Super open to feedback on all of this … my use case/desired usage might not be in exactly in line with Parcel's? I'm really not sure what other users/authors might want or expect either.

mischnic:
> I think it still needs a plugin for import.meta

That's enabled by default since Babel 7.10. Though we probably need to handle it in Parcel itself: https://github.com/parcel-bundler/parcel/issues/3269

Pauan:
@mysterycommand I personally would not recommend using `--target nodejs`, because it is synchronous. Instead you can use `--target web` with NodeJS, you just need to pass a `Promise<Buffer>` to the `init` function:

```js
init(promisify(readFile)(wasmPath))
```

This works because the `init` function accepts a string *or* a Buffer/ArrayBuffer/Uint8Array (or a Promise which resolves to one of those).

This allows you to load the `.wasm` however you wish, while still having the initialization/imports object handled by wasm-bindgen.

> In the case of multiple crates being compiled to wasm and imported separately, the wasm-pack-generated module includes a (duplicate) loader for each module. This is probably not a huge deal, just a personal annoyance of mine.

That's unavoidable, since the imports object, JS heap, and linear memory must be different for each crate.

So the only things which could be deduplicated are a handful of small utility functions like `cachedTextDecoder`, `cachedTextEncoder`, and `load`. So the amount of duplicated code is small, just ~1 KB (which will become much smaller with compression).

Any sort of deduplication would have to be done in wasm-bindgen, I don't think this plugin is the right place to do it.

> The usage I'd like to achieve involves loading the wasm, and swapping them out in the bundle cache before running the entry (the "controversial" stuff in JSPackager) … all this to make it so that someone working with Rust/wasm can do like (for example)

I don't know how Parcel works internally, and I don't know what Parcel plugins are capable of doing, so I don't know if that's possible without ESM integration or top-level await.

Rollup does not work with top-level await (it simply relies on the browser's implementation of top-level await), so that's why the Rust Rollup plugin needed to use a less convenient API. Whereas Webpack *does* transform top-level await, so it has great support for `.wasm`.

aminya:
Does this PR add anything reusable for #5598?

PR #4981, opened 2020-08-06
More stable hashes (package-based instead of projectRoot-relative)

mischnic:
Do we actually want this (by default)?

- This system also applies to local monorepo packages
- Would deduplicate cases where Yarn wasn't able to do it (smaller bundles)
- Danger: some patches to node_modules could be discarded/applied to all consumers of that package
   - factor `fs.readFile(asset.filePath)` into the cache key?

### TODO

- [ ] Add test
- [x] factor input contents into cache key?
	- only with `isSource: false`
- [ ] cascading bundle invalidation
	- don't make this the actual id, only use for deduplicating asset group ids. Content-hashed asset ids would still use name+version

PR #5003, opened 2020-08-10
Asset: content hashes

mischnic:
# ↪️ Pull Request

1. Use base62 (uppercase,lowercase,numbers) for public asset ids
	- but not bundle ids because of case-insensitive filesystems...
2. Remove concise ids altogether
3. For the asset public ids, use content hash (if `--content-hash-assets`) of the individual assets or the id otherwise (including a `publicIdReference` system)

When you want to enable this content-hash mode:
- With multiple Parcel builds running in the same context (be it intentionally or unintentionally) as it decreases the chance of hash collisions

Downsides:
- This would cause much more unnecessary bundle hash changes where the above case doesn't provide any benefit

## ✔️ PR Todo

- [ ] "`--content-hash-assets`" ?
- [ ] add test


packages/core/core/src/CommittedAsset.js
```diff
@@ -33,9 +34,13 @@ export default class CommittedAsset {
 
   getContent(): Blob | Promise<Buffer | string> {
     if (this.content == null) {
-      if (this.value.contentKey != null) {
-        return this.options.cache.getStream(this.value.contentKey);
-      } else if (this.value.astKey != null) {
+      if (
+        this.value.hasContent ||
+        // TODO ???
+        this.options.cache.blobExists(nullthrows(this.value.contentKey))
```
mischnic, line 40: TODO

packages/core/integration-tests/test/image.js
```diff
@@ -6,7 +6,8 @@ import sharp from 'sharp';
 describe('image', function() {
   this.timeout(10000);
 
-  it('Should be able to resize images', async () => {
+  // FIXME
+  it.skip('Should be able to resize images', async () => {
```
mischnic, line 10: Fails with "input buffer has unsupported file format", although the buffer contains the exact same bytes....

devongovett:
Would a hash of the input file contents work or do we really need to use an output hash after all transformation? We already have an `asset.hash` property that could be used for that potentially. Seems like there's currently a lot of complexity to get the output hashes and replace with references.

Also, same concerns as I mentioned in https://github.com/parcel-bundler/parcel/pull/4981#issuecomment-674458327

mischnic:
> Would a hash of the input file contents work or do we really need to use an output hash after all transformation?

What's not handled with the input hash is: `process.env`, differing Babel (or other) configs

Issue #5035, opened 2020-08-18
Upgrade to Terser 5


mischnic:
We should upgrade to Terser 5 as Terser 4 will not receive any more updates. Relevant breaking changes:
- `minify()` is now an async function, which isn't a problem for us.

### Solution

1. Update semver range in https://github.com/parcel-bundler/parcel/blob/64ebc3757a7c36432d55ee4d791e96e30dab242f/packages/optimizers/terser/package.json#L28
2. Add await: https://github.com/parcel-bundler/parcel/blob/64ebc3757a7c36432d55ee4d791e96e30dab242f/packages/optimizers/terser/src/TerserOptimizer.js#L42

AbderrahimSoubaiElidrissi:
I will take this

AbderrahimSoubaiElidrissi:
Done https://github.com/parcel-bundler/parcel/pull/5038

tbroyer:
Is there any chance that this gets backported to v1?

Currently, with Safari 14 now being a thing, using `last 2 chrome versions, last 2 firefox versions, last 2 edge versions, last 2 safari versions, last 2 opera versions` as browserslist, nullish coaslescing and optional chaining are no longer transformed, and then `parcel build` fails when minifying because it uses Terser 3.

So, either I use Parcel 1.x and have to include `safari 12` back in my browserslist to make Babel transform those (and probably others, leading to a less optimal code), or I move to `parcel@nightly` to benefit from this change but possibly meeting some bugs, or I move to something else entirely (webpack? rollup?) but nothing seems to use HTML as the entrypoint like Parcel.
For now I chose to _deoptimize_ my code and keep Parcel 1.x, but Parcel 1.x is starting to decay and Parcel 2.x is still not ready so it would be great if there was a stable and maintained solution somewhere.
Having a Parcel 1.13 with an updated Terser would be great.

alexburner:
I'm still not able to use optional chaining in Parcel 2, with a minimal TypeScript setup. 

Terser throws an error when trying to build: 
<img width="483" alt="terser-error" src="https://user-images.githubusercontent.com/409114/109995511-35528700-7cc3-11eb-830c-0c531dd23fdb.png">

Inspecting `package-lock.json`, looks like Parcel is bringing in Terser 4.8.0:
```json
"@parcel/optimizer-terser": {
  "version": "2.0.0-beta.1",
[...]

PR #5038, opened 2020-08-18
chore: upgrade Terser Optimizer

AbderrahimSoubaiElidrissi:
PR: Upgrade teaser 4 to 5 https://github.com/parcel-bundler/parcel/issues/5035

### DIFF
--- a/packages/optimizers/terser/package.json
+++ b/packages/optimizers/terser/package.json
@@ -25,7 +25,7 @@
     "@parcel/source-map": "2.0.0-alpha.4.13",
     "@parcel/utils": "2.0.0-beta.1",
     "nullthrows": "^1.1.1",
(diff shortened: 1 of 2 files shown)


mischnic:
1. Apparently, there was another breaking change: now calling `minify` throws an error instead of returning it:

```js
let result = minify();

if(result.error){
    let {...} = result.error;
    ...
}
```
should now be
```js
let result;
try{
    result = await minify();
} catch(err){
    let {...} = err
    ...
}
```

That should make the integration tests pass on CI.

2. I was looking whether this change would install two version of Terser and the only other place where it was used is `packages/core/utils/src/serializeObject.js`. But that function is never called, so that file (and the `terser` entry in `packages/core/utils/package.json`) can be removed as well.

PR #5054, opened 2020-08-23
ESBuild Optimizer

DeMoorJasper:
# ↪️ Pull Request

This PR adds an optimizer that uses ESBuild as a minifier, getting sourcemaps for this is very hacky, besides that it does seem to work.

This is an experiment and shouldn't be used as the default optimizer any time soon unless esbuild becomes as complete and stable as terser

## Benchmarking

After some back and forth between the team and Evan we figured out how to optimise this and now the builds are looking very similar in size and way better performance wise for ESBuild.

I still need to redo this with sourcemaps, but I doubt Terser is gonna be faster at that as their sourcemaps implementation isn't very fast.

| Benchmark | Terser | ESBuild |
| ---| --- | --- |
| Three (--no-source-map) | 120secs - 4.5M | 75secs - 4.5M  |
| Three | - | - |


devongovett:
Hmm did you measure the time taken by esbuild itself vs terser rather than the overall build time? Kinda surprised to see this.

DeMoorJasper:
@devongovett yeah I've just profiled it and the main performance bottleneck that makes the difference close to non is that most time is spent in the Graph.

The ESBuild optimiser runs for about 1ms

devongovett:
Ah so we're not testing it on a big enough project for terser to be a large portion of the build?

DeMoorJasper:
@devongovett this was in the Threejs example we have, I expected that to be large enough as the total build takes about 70seconds but it's a fairly minimal difference, the benchmarks I mentioned are probably very inaccurate as I was doing other things as well during the tests... but it's nothing significant. But I doubt they're of by more than 5 seconds

devongovett:
cc. @evanw

aminya:
The final size which affects the runtime in the production is way more important than the build time which happens during development. From what I see in the benchmarks, Esbuild is performing very badly in this regard, and the final size is almost double the size that Terser generates.

evanw:
I'm not sure how to run this benchmark myself but I can try to help out. Here's what I tried:

* Clone https://github.com/parcel-bundler/parcel-benchmark-action
* Run `make benchmarks/three`
* Do `npm install parcel@next`
* Run `npx parcel build --no-cache benchmarks/three/entry.js`

This produced a file called `dist/entry.js` that's 2.3mb. This doesn't match the benchmark described above, which says the output should be 20mb. So I probably have the benchmark instructions wrong. My apologies.

Regardless, I wanted to see how esbuild did on the same input, even though it's not the same benchmark that is described here. I modified `node_modules/@parcel/optimizer-terser/lib/TerserOptimizer.js` to write out the `code` variable before it's passed to `terser.minify()`. This produced a file that's 4.0mb. Running `esbuild --minify` on this file also produces a 2.3mb minified file. So for this case esbuild's output is approximately the same size as terser's output (esbuild's is 2,318.2kb while terser's is 2,318.1kb).

It would be interesting for me to run tests on the benchmark described in the PR description. Perhaps I have the wrong benchmark? I picked https://github.com/parcel-bundler/parcel-benchmark-action since it seemed to be what the parcel-benchmark bot above uses.

DeMoorJasper:
@evanw thanks for having a look at this.

The 20M was the entire size of the dist folder, so including sourcemaps... Just redid the test with sourcemaps disabled and some other tweaks. The benchmark is in `packages/examples/three`.

The results are still similar, esbuild outputs larger files when doing a legacy build (targetting older browsers), not sure if this is actually an issue with esbuild or just a bug in our Terser optimizer where it doesn't respect the browser target. Maybe I configured esbuild incorrectly?

Also not sure how to let esbuild map to an old sourcemap, is this possible?

evanw:
Thanks for providing more context.

After inspecting the results, it looks like the size difference is because the modern build is done inside of an IIFE closure while the legacy build isn't. The name minification pass inside esbuild doesn't minify top-level names because that would change the meaning of the code. It looks like terser has this same behavior too, but the legacy build for terser uses the special `toplevel: true` flag to change terser's default behavior. Is there a reason why the legacy build doesn't wrap everything in an IIFE closure too just like the modern build? Then having to change the name minification behavior for top-level names wouldn't be necessary.

There is logic in esbuild for remapping using an input source map. I believe it's currently hooked up to the bundler but not the transformer. When bundling with source maps enabled, source files with a `//# sourceMappingURL=` comment will read that source map and use that to remap the final output all the way back to the original input files. I can figure out a way to expose this for the transform API if it's not already exposed.

DeMoorJasper:
Great, thanks for figuring this out. I personally don't know why we do this but I'm sure @mischnic knows everything about this.

Will look further into the API of transforms and figure out how to properly do the sourcemaps

mischnic:
Using `browser` as a target name, it's interpreted as the `main` field for browsers, so a CommonJS library build.

This wraps both bundles in an IIFE:
```diff
-  "browser": "dist/legacy/index.js",
+  "browserLegacy": "dist/legacy/index.js",
   "browserModern": "dist/modern/index.js",
   "targets": {
     "browserModern": {
@@ -18,7 +18,7 @@
         ]
       }
     },
-    "browser": {
+    "browserLegacy": {
       "engines": {
         "browsers": [
           "> 0.25%"
```

evanw:
I just remembered: if you want to pass a source map to `service.transform(code, options)` you can use an inline comment such as ``service.transform(`${code}\n//# sourceMappingURL=data:application/json;base64,${base64(json(map))}`, options)``. Does that work for you?

DeMoorJasper:
@evanw awesome, thanks for helping out with the sourcemaps. Just tried it out and I think there's a bug with the source paths. it returns paths like these: `"./three/dist/modern/./three/src/copy4/lights/RectAreaLight.js"`

kzc:
`three.js`, or most JS libraries for that matter, are not ideal test cases for minification because most code is retained. Comparable post-gzip size results can be obtained with just whitespace elimination and symbol mangling alone - with `compress` disabled in UglifyJS terminology.

A complete application like https://github.com/marmelab/react-admin/tree/master/examples/simple better exercises a bundler and dead code elimination. See the following thread for results for comparisons between rollup+terser and esbuild: https://github.com/evanw/esbuild/issues/81#issuecomment-633185889. I didn't try parcel.

devongovett:
@DeMoorJasper do we want to merge this? Looks like the sourcemap changes were merged already...

DeMoorJasper:
gonna close this for now as it's not reliable, produces invalid sourcemaps and is probably not worth maintaining at this point

packages/optimizers/esbuild/src/ESBuildOptimizer.js
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    let relativeBundlePath = path.relative(
+      options.projectRoot,
+      bundle.filePath,
+    );
+    let code = await blobToString(contents);
+    if (map && options.sourceMaps) {
+      let vlqMappings = map.toVLQ();
+      vlqMappings = {
```
devongovett, line 27: Why not use `map.stringify`?

packages/optimizers/esbuild/src/ESBuildOptimizer.js
```diff
@@ -0,0 +1,64 @@
...
+        file: normalizePath(relativeBundlePath + '.map'),
+        sourceRoot: options.projectRoot,
+      };
+      code += `\n//# sourceMappingURL=data:application/json;charset=utf-8;base64,${Buffer.from(
+        JSON.stringify(vlqMappings),
+      ).toString('base64')}`;
+    }
+
+    let {code: js, map: jsSourceMap} = await service.transform(code, {
+      sourcemap: options.sourceMaps,
```
devongovett, line 45: should we use `sourcemap: 'external'` to avoid esbuild adding the `sourceMappingURL` comment? https://esbuild.github.io/api/#sourcemap

packages/optimizers/esbuild/src/ESBuildOptimizer.js
DeMoorJasper: mainly performance, there's some overhead for using async and the additional logic in stringify. Probably fairly minimal, so we could use it the optimizers don't run that often anyway

packages/optimizers/esbuild/src/ESBuildOptimizer.js
DeMoorJasper: oh interesting, that must be new or maybe I missed it last time, docs were limited at the time

devongovett (approved):
LGTM. Do we want to somehow remove the IIFE wrapper (if any) and re-add it after minification so esbuild actually removes dead code? Or do that separately? See https://github.com/evanw/esbuild/issues/639

DeMoorJasper:
@devongovett seeing that the issue has been closed I assume we can do this in a follow-up by just updating esbuild and maybe passing an additional option?

devongovett:
Yeah we'd also have to strip our IIFE and allow ESBuild to add theirs instead.

PR #5061, opened 2020-08-24
Reload css form local HMR server

### LINKED ISSUE #5100: [Parcel-2] HMR does not update CSS files if entry point is not *.html
# 🐛 bug report

I use parcel to run a dev server and compile assets from `*.js/*.ts` entries, and then include those assets into html. It can be a file on the disc or an html generated by some other framework (php, ruby, etc.). When I make a change in CSS, the parcel correctly updates/hot-reloads js bundle, but not the CSS file. This leads to the page to be broken and requires to manually reload the whole page.

[Parcel-hmr-css](https://github.com/maksimsemenov/parcel-hmr-css)

## 🤔 Expected Behavior

Parcel should reload the CSS file as well

## 😯 Current Behavior

Because the CSS file has a different origin from the current page (the current page is either a static file, or is served from a different port on localhost), the service worker script treats this CSS file as a global and does not reload it.

## 💁 Possible Solution

I created a PR with the possible fix: #5061

## 🔦 Context

It breaks the dev experience

## 💻 Code Sample

[Parcel-hmr-css](https://github.com/maksimsemenov/parcel-hmr-css)

## 🌍 Your Environment

| Parcel           | 2.0.0-beta.1 |

maksimsemenov:
# ↪️ Pull Request

Closes #5100

The CSS file can be included in the HTML file or in the HTML template
of a web framework (PHP or Ruby, etc). In this case,
the location origin will not match with CSS href and CSS will
not be reloaded. To fix that we check if CSS is served from
the localhost and its port is the same as parcel HMR port

## 💻 Examples
This [repo](https://github.com/maksimsemenov/parcel-hmr-css) illustrates the broken use case: CSS file won't reload, which leads to broken styles on the page

## 🚨 Test instructions
In the [repo](https://github.com/maksimsemenov/parcel-hmr-css):
1. Install dependencies `npm i`
2. Run `npm run dev`
3. Open index.html in browser
4. Change any style in `App.css`

## ✔️ PR Todo


Jogai:
@DeMoorJasper Can someone review this? It fixes a problem for me..

DeMoorJasper (approved):

PR #2872, opened 2019-03-30
Strip `type="module"` from output <scripts> tags

twome:
# ↪️ Pull Request

This addresses the feature request in  #1401. Because Parcel assumes a non-strict non-module execution context for its output ES5, running the output from a `type="module"` script tag will hit a ReferenceError when `parcelRequire` tries to implicitly define itself as a global variable using undeclared assignment. With this, users won't need to know/remember to only use vanilla <script> tags in their source HTML. 

## 💻 Examples

See #2809's use-case.

## 🚨 Test instructions

[New test](https://github.com/parcel-bundler/parcel/pull/2872/files#diff-728d2fc29ee35fbd3dee59dfdc80d85dR798): `html -> should strip 'type' attribute from ES-module and plain script tags`

Same repro as as #2809's. Only errors on macOS 10.14.2 are `should produce a basic kotlin bundle` (no local Java). Only [error from CI](https://dev.azure.com/devongovett/devongovett/_build/results?buildId=1581&view=logs&jobId=520d356b-b8e2-5387-8fcd-807640ab874e&taskId=bc978590-48aa-5958-925c-56704fd94fd3&lineStart=244&lineEnd=245&colStart=1&colEnd=1) is:

```
2019-03-30T07:38:15.1147470Z parcel-bundler:   1 failing
2019-03-30T07:38:15.1158420Z parcel-bundler:   1) babel
2019-03-30T07:38:15.1159230Z parcel-bundler:        "before all" hook:
2019-03-30T07:38:15.1160400Z parcel-bundler:      SyntaxError: unknown: Identifier 'T' has already been declared (29:5)
```

Is this different from the CI results from the current official master? I can't find any way to see those CI results in the GitHub / Azure Pipelines UIs, and when I tried to reset all my changes from the old version of this PR to trigger the CI tests, GitHub automatically closed the PR 😅!

## ✔️ PR Todo

- [x] Added/updated unit tests for this change
- [x] Filled out test instructions (In case there aren't any unit tests)
- [x] Included links to related issues/PRs

PR #5077, opened 2020-08-27
Support <script type="module"> when serving with HMR

### LINKED ISSUE #4773: Specifying type="module" for script tag breaks HMR
# 🐛 bug report

When I add a type="module" attribute in a script tag in my index.html, the HMR fails when parcel wants to reload the page on saving changes to the index.html file.

In the browser the error is the following
```
HMRRuntime.js:56 Uncaught TypeError: Cannot read property 'parcelRequire' of undefined
    at HMRRuntime.js:56
    at WebSocket.ws.onmessage (HMRRuntime.js:53)
```

## 🎛 Configuration

No configuration, just a html file like the following

```html
<!DOCTYPE html>
<html>
	<head>
		<script src="./index.js" type="module"></script>
	</head>
	<body>
		<div>test</div>
	</body>
</html>
```

oliverdunk:
# ↪️ Pull Request

Hot Module Replacement was failing when a `<script type="module">` was present on the
page. Building for development now ignores the type attribute, allowing HMR to be
used. The attribute is preserved when using the parcel build command.

In development, scripts with the `nomodule` attribute are now removed entirely. This is under the assumption a `<script type="module">` is also present, which will run in all browsers since the type was removed.

Resolves #4773 

## Thought Process

I discussed a solution with @mischnic in https://github.com/parcel-bundler/parcel/issues/4773. The plan was to remove type="module" scripts, under the assumption a nomodule script would be present, and to remove the nomodule attribute so that script ran in all scenarios. I decided to do things the other way around, because:

- Tests like [this one](https://github.com/parcel-bundler/parcel/blob/v2/packages/core/integration-tests/test/integration/html-inline-js-shared/index.html#L7) use module scripts alone and expect them to remain. It'd be fiddly to change this expectation.
- I can imagine someone new to Parcel adding a `<script type="module">` without the `nomodule` counterpart. Keeping module scripts and stripping the type means the one script they've added won't disappear.

I also chose to add the `defer` attribute, as suggested [here](https://github.com/parcel-bundler/parcel/issues/1401#issuecomment-656171260). This makes sense because it makes our script more closely resemble the fact that this is [added by default](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script#attr-defer) on module scripts.

I came across https://github.com/parcel-bundler/parcel/pull/2872, but I believe that applied in production, making it a more significant change. This MR is limited to development builds, as HMR is now the only thing broken by module scripts - scripts with type="module" work in production just fine.

## 🚨 Test instructions

Run the new integration tests.

You can also copy the react-hmr project from `packages/examples`, add `type="module"` to the script, and check that HMR works when running `parcel serve`.

## ✔️ PR Todo


oliverdunk:
@mischnic I'd love some feedback here. Is it something you could look at, or is there someone else that might be better?

mischnic:
After some discussion, we think that removing `nomodule` scripts and removing `type="module"` should only be done if there are actually two corresponding module/nomodule tags. 😬 

This should be doable without traversing twice by keeping adding the script tags to a list and modifying/removing/`addDependency()` them after traversal is done.

> I also chose to add the defer attribute

👍

oliverdunk:
@mischnic Thanks for taking a look, that sounds like fair feedback. Happy to make the changes, I should be able to get to those towards the end of the week.

oliverdunk:
Hey @mischnic - I'm sorry for leaving this one a crazy amount of time. I should be close to being able to finish this off - but, just to confirm beforehand, does this MR still make sense and would you still approach it in the same way?

mischnic:
https://github.com/parcel-bundler/parcel/pull/5398 is going to change how `<script type=module>` is handled: with that PR, `<script type=module>` is turned into two script tags by Parcel, so the user doesn't specify two module/nomodule scripts. This would simply handling this in dev mode (simply remove `type=module`). We should probably wait until that PR is merged?

devongovett:
This shouldn't be needed anymore

PR #5123, opened 2020-09-09
Write message when ctrl-c is received

thebriando:
# ↪️ Pull Request
Writes a message to the console when `ctrl-c` is received to notify the user that Parcel is shutting down.

## 🚨 Test instructions

Run Parcel in a directory and exit with `ctrl-c` to see the message that is outputted.

## ✔️ PR Todo


padmaia (approved):
Looks good to me, the only thing I can think of is maybe adding a chalk color? @thebriando could you maybe provide a picture to show what it looks like?

thebriando:
@padmaia Does this look ok? 
image

packages/core/parcel/src/cli.js
```diff
@@ -223,6 +223,9 @@ async function run(entries: Array<string>, command: any) {
           // We don't use the SIGINT event for this because when run inside yarn, the parent
           // yarn process ends before Parcel and it appears that Parcel has ended while it may still
           // be cleaning up. Handling events from stdin prevents this impression.
+          INTERNAL_ORIGINAL_CONSOLE.log(
+            chalk.bold.yellowBright('Parcel is shutting down'),
```
wbinnssmith, line 227: nit: Maybe "Parcel is shutting down..."?

mischnic:
Should this message be displayed with a small delay (via setTimeout)? I haven't run into a situation yet where Parcel seemed to hang after pressing Ctlr-C (as in: my projects weren't big enough 😄 ).

thebriando:
@mischnic Hmm how long should the timeout be?

mischnic:
I would say long enough so that you don't ask yourself if Parcel crashed for big projects and short enough that you don't even see the message for small(er) projects. Probably below 1 second?

Issue #5175, opened 2020-09-23
Received URL without a pathname #logotype.


FRSgit:
# 🐛 bug report

Parcel fails when it stumbles across html fragment with svg sprite used like this:

```html
<svg xmlns:xlink="http://www.w3.org/1999/xlink">
   <use xlink:href="#logotype" href="#logotype"></use>
</svg>
```

## 🎛 Configuration (.babelrc, package.json, cli command)

Default configs, happens when running both `parcel` and `parcel build`.

```js
{
  "scripts": { "build": "parcel build src/index.html" }
}
```

## 🤔 Expected Behavior

The HTML should be bundled without an error.

## 😯 Current Behavior

Parcel throws an error:
![image](https://user-images.githubusercontent.com/10456649/94004561-2a530c80-fd9d-11ea-9544-55688a71a487.png)

## 💁 Possible Solution

Seems like there was already similar issue for CSS files (#3873), seems like there should be some kind of ignoring happening for urls that are beginning with a `#` sign - either it would be done automatically by parcel or by some `ignore-paths` configuration option.

## 🔦 Context

This bug basically makes it impossible to create SVG sprites with use of SVG `<use>` tag.

## 💻 Code Sample

Create new parcel project, paste the code from summary to the index.html and run `parcel` or `parcel build`.

## 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Parcel           | 2.0.0-beta.1
| Node             | 10.21.0
| Yarn         | 1.22.4
| Operating System | MacOS

mischnic:
We need to add a special case somewhere in here:

https://github.com/parcel-bundler/parcel/blob/v2/packages/transformers/html/src/dependencies.js#L121

And a test case here: https://github.com/parcel-bundler/parcel/blob/v2/packages/core/integration-tests/test/html.js

FRSgit:
Closing as this was fixed by #5188.

PR #5188, opened 2020-09-27
fix: svg use referencing local symbol via id

FRSgit:
add test case to cover case of local symbols

# ↪️ Pull Request

Parcel have failed when referenced local SVG symbols via `<use xlink:href="#sth"></use>` (more explanation in the bug report - #5175).

**Added only integration test, please give me a sign, if it's needed from me to add anything else to this PR** (ping @mischnic)

## 💻 Examples

Now the bundler correctly ignores paths starting with a hash (id references), for example:
```html
<svg xmlns:xlink="http://www.w3.org/1999/xlink">
   <use xlink:href="#logotype" href="#logotype"></use>
</svg>
```

## 🚨 Test instructions

In the examples.

## ✔️ PR Todo

### DIFF
--- a/packages/transformers/html/src/dependencies.js
+++ b/packages/transformers/html/src/dependencies.js
@@ -160,6 +160,11 @@ export default function collectDependencies(asset: MutableAsset, ast: AST) {
         continue;
       }
 
+      // Check for id references
+      if (attrs[attr][0] === '#') {
+        continue;
+      }
+
       let elements = ATTRS[attr];
       if (elements && elements.includes(node.tag)) {
         let depHandler = getAttrDepHandler(attr);
--- a/packages/core/integration-tests/test/html.js
+++ b/packages/core/integration-tests/test/html.js
@@ -703,6 +703,19 @@ describe('html', function() {
     ]);
   });
 
+  it('should ignore svgs referencing local symbols via <use xlink:href="#">', async function() {
+    let b = await bundle(
+      path.join(__dirname, '/integration/html-svg-local-symbol/index.html'),
+    );
+
+    assertBundles(b, [
+      {
+        name: 'index.html',
+        assets: ['index.html'],
+      },
+    ]);
+  });
+
   it('should bundle svg files using <image xlink:href=""> correctly', async function() {
     let b = await bundle(
       path.join(__dirname, '/integration/html-svg-image/index.html'),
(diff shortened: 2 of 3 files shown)


packages/transformers/html/src/dependencies.js
```diff
@@ -160,6 +160,11 @@ export default function collectDependencies(asset: MutableAsset, ast: AST) {
         continue;
       }
 
+      // Check for id references
+      if (attrs[attr][0] === '#') {
```
mischnic, line 164: I think this should also check `tag === 'use' && (attr === 'href' || attr === 'xlink:href')`?

FRSgit: Yea, I thought about that, but isn't always hash at the beginning of the `src`/`href`/any other "path-related" attribute means an id-based selector?

mischnic: For `<a href="#xyz">` yes, but `` should probably still be an error.

FRSgit: Yeees and no. `` isn't invalid (on the other hand, `` is treated as invalid by the browsers) and because of that it's even sometimes used as a "placeholder src value" when e.g. the `src` is meant to be later changed by js (https://stackoverflow.com/questions/19613841/img-src-means-in-html).

It's not a strong use-case, that's true, but I'm not sure if parcel should fail on a valid HTML? But also, you're the parcel member here, so if you say that I should check for `use` element & attr name here, that's what I will do

FRSgit: Out of curiosity [I've hacked a small codepen](codepen.io/kkkubas/pen/zYqXvVO) and tried out all of the elements & attributes which are being handled by this html transformer with id-based path: `#sth`.
It seems that none of the elements is failing (at least not because of having the id-based path set - I've commented out `script` tag under `ATTRS.src`, because the `script` element have tried to load the current page as an js script - which obviously have failed).

BTW. surprisingly the `object` element have loaded the empty HTML page when given `#sth` as the `data` - the same way as `iframe` did. I didn't know it would work like this 😄

FRSgit: Hey @mischnic what would be the decision here? Let's limit the condition to check for tag & attr name and call it a day?
Or do we leave current solution?

mischnic (approved):

FRSgit:
Oh, wow, wasn't expecting that to be merged just like that, thanks @mischnic! <3

PR #5192, opened 2020-09-28
GLSL Transformer

### LINKED ISSUE #3352: Parcel 2: GLSL transformer
Create the `@parcel/transformer-glsl` package in `packages/transforms/glsl`. This package is a Parcel 2 transform plugin for GLSL. It should be based on the `GLSLAsset` from Parcel 1.

The `@parcel/transformer-glsl` package is responsible for the following things:

* Collecting dependencies in GLSL files using `glslify-deps` and resolving with the Parcel resolver.
* Bundling dependencies with `glslify-bundle` and exposing as JavaScript
* Remove unneeded whitespace

We could also reconsider the way this works, and split this into actual Parcel plugins for each part:

* A transformer to collect deps for the immediate asset (not any children)
* A packager to combine those assets together
* Some way of inlining the result into JS if needed (perhaps an inline packager, similar to HTML)

This would have the benefit of allowing packaging GLSL to an output GLSL file, or inline into some other language besides JS as well. Perhaps the infra developed for #3301 will be useful here as well.

fedegratti:
# ↪️ Pull Request

Closes #3352

## 💻 Examples

```JS
import vertex_shader from './vertex_shader.vert';
import fragment_shader from './fragment_shader.frag';
```

## ✔️ PR Todo

### DIFF
--- a/packages/configs/default/index.json
+++ b/packages/configs/default/index.json
@@ -14,6 +14,7 @@
     "*.jsonld": ["@parcel/transformer-jsonld"],
     "*.toml": ["@parcel/transformer-toml"],
     "*.yaml": ["@parcel/transformer-yaml"],
+    "*.{glsl,vert,frag}": ["@parcel/transformer-glsl"],
     "*.{gql,graphql}": ["@parcel/transformer-graphql"],
     "*.{styl,stylus}": ["@parcel/transformer-stylus"],
     "*.{sass,scss}": ["@parcel/transformer-sass"],
--- a/packages/configs/default/package.json
+++ b/packages/configs/default/package.json
@@ -41,6 +41,7 @@
     "@parcel/transformer-babel": "2.0.0-beta.1",
     "@parcel/transformer-coffeescript": "2.0.0-beta.1",
     "@parcel/transformer-css": "2.0.0-beta.1",
+    "@parcel/transformer-glsl": "2.0.0-beta.1",
     "@parcel/transformer-graphql": "2.0.0-beta.1",
     "@parcel/transformer-html": "2.0.0-beta.1",
     "@parcel/transformer-inline-string": "2.0.0-beta.1",
--- a/packages/core/integration-tests/test/glsl.js
+++ b/packages/core/integration-tests/test/glsl.js
@@ -1,26 +1,11 @@
 import assert from 'assert';
 import path from 'path';
 import fs from 'fs';
-import {
-  bundle,
-  run,
-  assertBundleTree,
-  normaliseNewlines,
-} from '@parcel/test-utils';
+import {bundle, run, normaliseNewlines} from '@parcel/test-utils';
 
-describe.skip('glsl', function() {
+describe('glsl', function() {
   it('should support requiring GLSL files via glslify', async function() {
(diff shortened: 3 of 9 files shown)


mischnic:
We probably want to 
- call `asset.addIncludedFile(f)` on the files that were bundled 
- use our own resolver mechanism instead of `glsify-deps`'s default resolver:

https://github.com/parcel-bundler/parcel/blob/v2/packages/core/parcel-bundler/src/assets/GLSLAsset.js

fedegratti:
> We probably want to
> 
> * call `asset.addIncludedFile(f)` on the files that were bundled
> * use our own resolver mechanism instead of `glsify-deps`'s default resolver:
> 
> https://github.com/parcel-bundler/parcel/blob/v2/packages/core/parcel-bundler/src/assets/GLSLAsset.js

@mischnic Thank you for the quick reply! I tried to do these changes, but I didn't find a way to use your resolver on Parcel 2. Could you show me an example of how to use your own resolver on Parcel 2?
Thank you

mischnic:
V2 passes a `resolve` function to the `transform` function just like V1. The SASS/LESS transformers also use this.

Use that function in the callback for glslify-deps like here: https://github.com/parcel-bundler/parcel/blob/b2da8a221b9289b626cbef2742894101ca620677/packages/core/parcel-bundler/src/assets/GLSLAsset.js#L27

fedegratti:
@mischnic I did the proper changes to use the V2 resolver mechanism but I have some questions: 

-  `resolve` function (in the way I'm currently using it) doesn't resolve paths without a file extension. Is this the expected behavior? or am I missing something?

- I didn't completely get the meaning of `asset.addIncludedFile(f)`, so I put it where I thought it was right. Please check if it is in the right place.

mischnic:
> resolve function (in the way I'm currently using it) doesn't resolve paths without a file extension. Is this the expected behavior? or am I missing something?

This is determined by the list here: https://github.com/parcel-bundler/parcel/blob/b5f4d1c9d952e0513545f863a04dcff6640ea2e8/packages/resolvers/default/src/DefaultResolver.js#L31

But we should leave it as is. Extensionless specifiers have some performance and caching problems...

> I didn't completely get the meaning of asset.addIncludedFile(f), so I put it where I thought it was right. Please check if it is in the right place.

So `glslifyBundle` inlines theses imports into the parent file. But Parcel needs to know which files are bundled here in additon to `asset.filePath` so that the transformer reruns if a included file changed.

That said, I don't know whether what you or what V1 did
https://github.com/parcel-bundler/parcel/blob/b2da8a221b9289b626cbef2742894101ca620677/packages/core/parcel-bundler/src/assets/GLSLAsset.js#L44-L48
is more correct....

fedegratti:
Got it, thank you for the explanation. I decided to put `asset.addIncludedFile(f)` inside a function (similar to v1) so it is easier to follow. Please let me know if I have to do anything else.

devongovett (approved):

devongovett:
Thanks! 🤩

KonradLinkowski:
Hey, I tried using the feature, but typing `import fragment from '../shaders/fragment.glsl';` doesn't work.
Do I need to add another module or change something in parcel config?

PR #5218, opened 2020-10-06
Fix mistyped calls to MutableAsset::addURLDependency

rhendric:
# ↪️ Pull Request

I found two places where `MutableAsset::addURLDependency` was being called with a string as the second argument, where that function expects an object. This PR corrects those calls and adds some types to help prevent similar errors in the future.

## 💻 Examples

N/A

## 🚨 Test instructions

Type checking should cover it. (Revert just the calls to `addURLDependency` while leaving the surrounding type annotations, and the type checker should fail.)

## ✔️ PR Todo


packages/transformers/less/src/LessTransformer.js
```diff
@@ -15,6 +15,68 @@ type LessConfig = {
   ...
 };
 
+type LessFileInfo = {|
```
wbinnssmith, line 18: can these be moved to a type definition for less itself?

mischnic (approved):

PR #5219, opened 2020-10-07
Partially fix asset query handling

rhendric:
The test case for an image referenced from HTML is broken for now; there seems
to be a deeper conflict with assets in named pipelines and some aspect of the
HTML pipeline.

# ↪️ Pull Request

The handling of asset query strings, introduced in #4881 in support of `@parcel/transformer-image`, didn't work when the asset dependency was marked as being a URL, such as in CSS or HTML files. This PR fixes the query string handling issue, and adds test cases for transforming images from CSS and HTML. Unfortunately, there seems to be an additional issue with the HTML case; I think there's some kind of conflict between the `url:` named pipeline and some aspect of the HTML pipeline. I don't entirely know how all of this is supposed to go; should the pipeline name be stripped from the module identifier at some point in the process? If so, where? If not, how is the `replaceURLReferences` call in `HTMLPackager` supposed to deal with it?

I'd love it if someone more knowlegable could take a look at the HTML case added here and point me in the direction of a fix. I can do the implementation work if I know what ought to be happening.

## ✔️ PR Todo

### DIFF
--- a/packages/core/core/src/requests/PathRequest.js
+++ b/packages/core/core/src/requests/PathRequest.js
@@ -151,20 +151,25 @@ export class ResolverRunner {
       filePath = dependency.moduleSpecifier;
     }
 
+    let queryPart = null;
     if (dependency.isURL) {
       let parsed = URL.parse(filePath);
       if (typeof parsed.pathname !== 'string') {
         throw new Error(`Received URL without a pathname ${filePath}.`);
       }
       filePath = decodeURIComponent(parsed.pathname);
+      if (parsed.query != null) {
+        queryPart = decodeURIComponent(parsed.query);
+      }
+    } else {
+      let matchesQuerystring = filePath.match(QUERY_PARAMS_REGEX);
+      if (matchesQuerystring && matchesQuerystring[2] != null) {
+        filePath = matchesQuerystring[1];
+        queryPart = matchesQuerystring[2].substr(1);
+      }
     }
-
-    let matchesQuerystring = filePath.match(QUERY_PARAMS_REGEX);
-    if (matchesQuerystring && matchesQuerystring.length > 2) {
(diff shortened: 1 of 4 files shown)


packages/core/core/src/requests/PathRequest.js
```diff
@@ -151,20 +151,25 @@ export class ResolverRunner {
...
-
-    let matchesQuerystring = filePath.match(QUERY_PARAMS_REGEX);
-    if (matchesQuerystring && matchesQuerystring.length > 2) {
-      filePath = matchesQuerystring[1];
-      query = matchesQuerystring[2]
-        ? querystring.parse(matchesQuerystring[2].substr(1))
-        : {};
+    if (queryPart != null) {
+      query = querystring.parse(queryPart);
     }
```
DeMoorJasper, line 173: This should probably be `query = queryPart != null ? querystring.parse(queryPart) : {};`

rhendric: `query` is initialized to `{}` back on line 128; why is the ternary preferable?

DeMoorJasper:
@devongovett wrote the original named pipelines, so maybe he knows. I might have a look into this as well tonight

packages/core/core/src/requests/PathRequest.js
DeMoorJasper: It's not I just missed that line, my bad

DeMoorJasper:
Had a look at it and I think the logic behind `getURLReplacement` was not handling this properly, opened a PR for it with a ton of new unit tests. #5228

Not entirely sure this is the way to fix it, but I don't think we should modify the moduleSpecifier so seemed like the best way to me. Feel free to have a look at it, thanks for bringing this issue up and fixing the query params issues.

PR #5239, opened 2020-10-14
Use a unique parcelRequire global per project

### LINKED ISSUE #4953: Change global parcelRequire in v2
# 💬 RFC

Scenario:
- Two different bundles are on one page; one compiled with Parcel 1.x, one with with Parcel 2.x.
- The code compiled with 2.x loads multiple modules.
- We receive the following:
Screen Shot 2020-07-28 at 3 36 24 PM
- Presumably it is because there are API differences between v1 and v2 builds.

This could be solved in a few different ways (one or more may apply):
- change `parcelRequire` to `parcelRequire2` in v2
- allow users the ability to customize the name of `parcelRequire` (this was a feature of some bundlers in the past, and was useful in ensuring that when 3rd party code was included on pages it wouldn't introduce side effects via unknown globals)
- automagically change `parcelRequire` to include a hash (per parcel version? per some other criteria? this seems a bit difficult since the all produced bundles would also have their corresponding hashes changed?) 

## 🔦 Context

I'm trying to load a Parcel 1.x project with a Parcel 2.x project in the same page with each multiple modules that load async.

However, presumably this could come up in other cases as well:
- hash collision between the two builds because two assets at `node_modules/react/index.js` in two different projects could have a different package version but still the same asset id in `parcelRequire("<<id>>")`
- since this considered "internal API" (i think?) it could vary between minor versions, and cause issues similar to the one I'm seeing between 1.x and 2.x

## 💻 Examples

See scenarios; I'm sure @mischnic and @devongovett have more 👍

devongovett:
Fixes #4953.

This changes the global `parcelRequire` function to use a unique generated name per project. This should make it possible to have multiple parcel builds on the same page at once. It's generated by hashing the ~~projectRoot~~ package.json#name. Probably not perfect but better than nothing. 🤷 If anyone has a better idea on how to generate a unique id for a project that's stable over time let me know. 😄

### DIFF
--- a/packages/core/core/src/BundleGraph.js
+++ b/packages/core/core/src/BundleGraph.js
@@ -1155,7 +1155,7 @@ export default class BundleGraph {
     this.traverseBundles((childBundle, _, traversal) => {
       if (childBundle.isInline) {
         bundles.push(childBundle);
-      } else {
+      } else if (childBundle.id !== bundle.id) {
         traversal.skipChildren();
       }
     }, bundle);
--- a/packages/core/core/src/BundleGraph.js
+++ b/packages/core/core/src/BundleGraph.js
@@ -1131,10 +1131,8 @@ export default class BundleGraph {
     return hashHex;
   }
 
-  getHash(bundle: Bundle): string {
-    let hash = crypto.createHash('md5');
-    hash.update(bundle.id);
-
+  getInlineBundles(bundle: Bundle): Array<Bundle> {
+    let bundles = [];
     let seen = new Set();
     let addReferencedBundles = bundle => {
       if (seen.has(bundle.id)) {
@@ -1146,7 +1144,7 @@ export default class BundleGraph {
       let referencedBundles = this.getReferencedBundles(bundle);
       for (let referenced of referencedBundles) {
         if (referenced.isInline) {
-          hash.update(this.getContentHash(referenced));
+          bundles.push(referenced);
           addReferencedBundles(referenced);
         }
       }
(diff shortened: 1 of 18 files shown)


packages/shared/scope-hoisting/src/hoist.js
```diff
@@ -271,7 +271,10 @@ const VISITOR: Visitor<MutableAsset> = {
       path.replaceWith(t.identifier('null'));
     }
 
-    if (t.matchesPattern(path.node, 'module.bundle')) {
+    if (
+      t.matchesPattern(path.node, 'module.bundle.root') ||
```
mischnic, line 275: Is there some precedent (in Webpack?) for `module.bundle.root`?

devongovett: AFAIK module.bundle in general is a parcel thing... I chose root here because it represents the root bundle on the page - the current parcelRequire global. Open to other names.

packages/core/core/src/BundleGraph.js
```diff
@@ -1146,23 +1144,40 @@ export default class BundleGraph {
...
+
+    let inlineBundles = this.getInlineBundles(bundle);
+    for (let inlineBundle of inlineBundles) {
+      hash.update(this.getContentHash(inlineBundle));
+    }
+
+    this.traverseBundles((childBundle, _, traversal) => {
+      hash.update(childBundle.id);
+      traversal.skipChildren();
+    }, bundle);
```
mischnic, line 1179: Does this do the same as before? Doesn't this traverse only the `childBundle` itself?

wbinnssmith: I believe `bundleGraph.getChildBundles()` does exactly this

mischnic: Maybe?
```suggestion
    for(let childBundle of this.getChildBundles()) {
      hash.update(childBundle.id);
    }
```

mischnic (approved):

wbinnssmith (approved):

Issue #4298, opened 2020-03-10
Disable AutoInstall by default


erhathaway:
## 💬 RFC

From my perspective, it is a massive security flaw that packages are implicitly downloaded behind the scenes by default. 

Multiple times now, myself and others I know, have miss-typed dependencies in a file import statement and Parcel downloaded a package that matched the off-by-one spelling. 

I purpose that this behavior is turned off by default.

## 🔦 Context

Undesired packages are installed when mistakes are made on import statements. These packages are, in turn, able to execute post-install scripts and thus easily execute malicious behavior.

It is unrealistic for NPM or other registries to guard against off-by-one misspellings.

From a security first perspective, Parcel should not implicitly download packages - especially ones the Parcel team doesn't audit.

## 💻 Examples

The relevant PR would disable the AutoInstall behavior by default and users would need to explicitly enable it if they desired it.

mischnic:
Related: https://github.com/parcel-bundler/parcel/issues/2081

erhathaway:
@mischnic thanks for the related link!

Do you know who would be best to pull into this RFC conversation to get consensus on this issue?

samsch:
Another case where autoinstall was both a major inconvenience and potential security issue: I'm converting a codebase which used something like Babel `module-resolver`, with import paths like `import xyz from 'xyz/path/to/file';`, where `xyz` is a folder in `src/`. Running parcel against this installed several packages with src folder names (aka, `xyz`).

samsch:
This is still an outstanding issue

samsch:
As noted here: https://github.com/parcel-bundler/parcel/pull/5267#issuecomment-716040487, v2 doesn't auto-install deps.

This can probably be closed?

PR #5267, opened 2020-10-22
Change cli autoinstall default to false (not properly tested)

samsch:
# ↪️ Pull Request

#4298 and #2081.

Reverses the default for autoinstall in the cli to be disabled by default, and enabled for non-production builds with `--autoinstall`.

Changes are not properly tested, and were made naively. But maybe enough to get the ball rolling.

There are also some documentation changes which should go along with this.

## 💻 Examples

`parcel index.html` Now never autoinstalls

`parcel index.html --autoinstall` now autoinstalls

## 🚨 Test instructions

I was unfortunately... not able to get autoinstall to work at all? Neither following the example instructions in CONTRIBUTING.md nor just installing parcel@2.0.0-beta.1 in completely separate project lead to a successful autoinstall.

That would actually put the cli doing what I'd want already, but this PR (blindly) makes that an official change.

I unfortunately ran out of time to continue attempting to get some solidly tested working stuff. The existing tests seem to pass locally fwiw though.

## ✔️ PR Todo

### DIFF
--- a/packages/core/parcel/src/cli.js
+++ b/packages/core/parcel/src/cli.js
@@ -56,7 +56,7 @@ if (typeof process.env.PARCEL_BUILD_ENV === 'string') {
 
 program.version(version);
 
-// --no-cache, --cache-dir, --no-source-maps, --no-autoinstall, --global?, --public-url, --log-level
+// --no-cache, --cache-dir, --no-source-maps, --autoinstall, --global?, --public-url, --log-level
 // --no-content-hash, --experimental-scope-hoisting, --detailed-report
 
 const commonOptions = {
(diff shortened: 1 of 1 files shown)


devongovett:
This is affecting Parcel 2 which already doesn't support auto install for packages referenced in source code. The only thing Parcel 2 auto installs is dev dependencies like compilers (e.g. sass, typescript, etc.). I think this is reasonable and expected, and if you don't want even that to auto install you can use --no-autoinstall.

Going to close this.

samsch:
Ah, that explains why it auto installing wasn't working.

I made an assumption based on --no-autoinstall existing and the v1 functionality. Maybe a note about this should be included in the v1->v2 migrating doc, and I'm guessing the --no-autoinstall doc is probably planned to elaborate more on what it's actually doing.

Thanks for the update!

PR #3439, opened 2019-08-19
[WIP] Add WebextManifestTransformer

Drapegnik:
# ↪️ Pull Request

Depends on #3404, Resolves #3362, Resolves #1039

## ✔️ PR Todo

- [x] Added/updated unit tests for this change
- [x] Included links to related issues/PRs

PR #5225, opened 2020-10-09
API audit notes

devongovett:

Issue #4840, opened 2020-07-03
.parcelrc "transformer" key gives Uncaught SyntaxError: Cannot use import statement outside a module


codrin-iftimie:
# 🐛 bug report

## 🎛 Configuration (.babelrc, package.json, cli command)

```json
// .parcelrc
{
  "extends": ["@parcel/config-default"],
  "transformers": {}
}

// package.json

{
  "scripts": {
    "clean": "rm dist/bundle.js",
    "start": "parcel src/index.html",
    "build": "parcel build src/index.html",
    "test": "jest"
  },
  "dependencies": {
    "axios": "^0.19.2",
    "connected-react-router": "^6.8.0",
    "history": "^4.10.1",
    "immer": "^7.0.5",
    "normalizr": "^3.6.0",
    "querystring": "^0.2.0",
    "react": "^16.13.1",
    "react-dom": "^16.13.1",
    "react-redux": "^7.2.0",
    "react-router": "^5.2.0",
    "react-router-dom": "^5.2.0",
    "redux": "^4.0.5",
    "redux-devtools-extension": "^2.13.8",
    "redux-thunk": "^2.3.0",
    "reselect": "^4.0.0"
  },
  "alias": {
    "services": "./src/services",
    "state": "./src/state",
    "utils": "./src/utils",
    "components": "./src/components",
    "pages": "./src/pages",
    "assets": "./src/assets"
  },
  "devDependencies": {
    "@parcel/config-default": "^2.0.0-beta.1",
    "@parcel/transformer-babel": "^2.0.0-beta.1",
    "@parcel/transformer-js": "^2.0.0-beta.1",
    "@parcel/transformer-svg-react": "^2.0.0-beta.1",
    "@parcel/transformer-svgo": "^2.0.0-beta.1",
    "babel-jest": "^26.1.0",
    "eslint": "^7.3.1",
    "eslint-config-prettier": "^6.11.0",
    "eslint-plugin-prettier": "^3.1.4",
    "eslint-plugin-react": "^7.20.3",
    "express": "^4.17.1",
    "http-proxy-middleware": "^1.0.4",
    "jest": "^26.1.0",
    "parcel": "^2.0.0-beta.1",
    "prettier": "^2.0.5"
  }
}

// index.js

import React from "react";
import ReactDOM from "react-dom";

var mountNode = document.getElementById("app");
ReactDOM.render(<div>test</div>, mountNode);
```

## 🤔 Expected Behavior
It should not break the build

[...]

PR #5304, opened 2020-10-31
Web Extension Transformer

### LINKED ISSUE #1039: Support WebExtensions manifest.json

101arrowz:
# ↪️ Pull Request

Adds support for bundling web extensions from a single `manifest.json`. Supports all standard features, plus a few non-standard but popular ones.

Rewrite of #3439
Closes #3439 
Closes #1039

I'm not sure how necessary this is in Parcel's core, but I think it would give Parcel a massive advantage over bundlers like Webpack, for which massive amounts of config are required to get web extensions working even with community packages.

## 💻 Examples

`manifest.json`:
```json
{
  "manifest_version": 2,
  "name": "Parcel WebExt",
  "version": "1.2.3",
  "default_locale": "en_US",
  "icons": {
    "100": "src/assets/foo.png"
  },
  "background": {
    "scripts": ["./src/background.js"],
    "persistent": true
  },
  "browser_action": {
    "default_icon": {
      "100": "src/assets/foo.png"
    },
    "default_popup": "src/popup.html"
  },
  "content_scripts": [{
    "matches": ["https://v2.parceljs.org/*"],
    "js": ["src/content.js"],
    "css": ["src/content.css"]
  }],
  "dictionaries": {
    "en-US": "./dicts/tmp.dic"
  },
  "devtools_page": "src/devtools.html",
  "web_accessible_resources": [
    "src/assets/**/*.txt"
  ]
}
```

If a file named `manifest.json` is given as an entry point and looks like a web extension manifest, it is validated for cross-browser compatibility, and all URL references are replaced. Assets that should not change location, such as `_locales`, are added as entry assets to prevent renaming.

## 🚨 Test instructions

Integration tests provided with this PR

## ✔️ PR Todo

- [x] <strike>Filled out test instructions (In case there aren't any unit tests)</strike> N/A


packages/core/parcel/package.json
```diff
@@ -14,7 +14,7 @@
     "type": "git",
     "url": "https://github.com/parcel-bundler/parcel.git"
   },
-  "bin": "src/bin.js",
+  "bin": "lib/bin.js",
```
mischnic, line 17:
```suggestion
  "bin": "src/bin.js",
```

101arrowz: Not sure how this happened... I never touched this file. Thanks for checking it

mischnic: Probably something like this:
```
yarn build
git commit -a
```

fregante:
I have a couple of extensions set up with Parcel 1 + https://github.com/kevincharm/parcel-plugin-web-extension

How can I test this PR on [my repo](https://github.com/tanmayrajani/notifications-preview-github/blob/35accac48e28da78e0a5b86bc45c44c20f5f0d98/package.json#L6-L7)?

101arrowz:
To be completely honest, I'm not sure how you could integrate it into your existing repo. However, what you could try is cloning, then copy-pasting all of your files into `packages/core/integration-tests/test/integration/webext`. Then replace the code inside `packages/core/integration-tests/test/webext.js` with the following:
```js
import path from 'path';
import {ncp} from '@parcel/fs';
import {prettyDiagnostic} from '@parcel/utils'
import {bundle, assertBundles, inputFS, outputFS, distDir} from '@parcel/test-utils';

describe('webext', function() {
  it('should resolve a full webext bundle', async function() {
    try {
      let b = await bundle(
        path.join(__dirname, '/integration/webext/manifest.json'),
      );
      await ncp(outputFS, distDir, inputFS, './webext-out');
    } catch(e) {
      for (const diag of e.diagnostics.map(prettyDiagnostic)) {
        console.error(diag.message);
        console.error(diag.codeframe || diag.stack);
        for (const h of diag.hints) console.error(h);
      }
    }
  });
});
```
Run `yarn test:integration -g webext` in the root directory. You should see your web extension in `packages/core/integration-tests/webext-out`.

mischnic:
Two options:
1. Clone the repo/checkout the branch: `git clone https://github.com/101arrowz/parcel --branch transformer-webext`
2. run `yarn`
3. Go into your project and run `/path/to/parcel/packages/core/parcel/src/bin.js build src/index.html` (or whatever entry you need). This works as long as you don't have any Parcel 2 packages installed in your project.

Alternatively:
3. Copy your project into the `packages/examples` folder, run `yarn`. Use the `parcel` command in the example folder (via `yarn` or `npx`).

wachunei:
Hey @101arrowz this very cool, thanks a lot. I'm testing this with my extension and I'll report here what I find.

packages/transformers/webext/src/schema.js
```diff
@@ -0,0 +1,402 @@
...
+
+// Non-comprehensive, but should cover the vast majority of users
+// SVGs not supported in WebKit/Blink
+const IMG_EXT = new Set(['jpg', 'jpeg', 'png', 'svg', 'webp', 'gif']);
+
+const validateVersion = (ver: string): ?string => {
+  const parts = ver.split('.', 5);
+  if (parts.length == 5) return 'Extension versions to have at most three dots';
+  // greater than to allow for NaN to fail
+  if (parts.every(part => +part[0] > 0 && +part < 65536)) return;
```
wachunei, line 13: I found this bug already:

Screen Shot 2020-11-01 at 18 16 54

I think it should be 
```js
if (parts.every(part => +part >= 0 && +part < 65536)) return;
```

mischnic: (Apart from that, please use `Number(part)`)

packages/transformers/webext/src/schema.js
```diff
@@ -0,0 +1,402 @@
...
+    type: 'string',
+    __validate: validateImage,
+  },
+};
+
+const actionProps = {
+  // FF only
+  browser_style: {type: 'boolean'},
+  // You can also have a raw string, but not in Edge, apparently...
+  default_icon: icons,
```
wachunei, line 36: According to documentation, [`browser_action.icons` should be `string` or `object`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/browser_action):

I guess something like:

```ts
const icons: SchemaOneOf = {
  oneOf: [{
    type: 'object',
    properties: {},
    additionalProperties: {
      type: 'string',
      __validate: validateImage,
    },
  },
  {
    type: 'string',
    __validate: validateImage,
  }]
};
```

wachunei:
Next error I'm getting is resolving `.ttf` as assets, but I guess it is not related to this transformer. @mischnic do you know if this is correct? 

Screen Shot 2020-11-01 at 18 47 42

However, when removing every `.ttf` reference in my codebase the build succeeds, but they do not work:

<a href="https://user-images.githubusercontent.com/1024246/97816453-2f488d00-1c74-11eb-90b6-78e8d1f69c93.png">Screen Shot 2020-11-01 at 18 57 18</a>

above is the background script on top and the options page in the bottom of the screenshot

packages/transformers/webext/src/schema.js
101arrowz: I'm aware, this is intentional to support Edge. The goal was maximizing compat.

mischnic:
If you are not actually specifying these files in the manifest.json, please open a new issue describing how/where you are including them

packages/transformers/webext/src/schema.js
```diff
@@ -0,0 +1,402 @@
...
+  if (!IMG_EXT.has(ext))
+    return `Extensions do not support ${ext} images and icons`;
+};
+
+const icons: SchemaEntity = {
+  type: 'object',
+  properties: {},
+  additionalProperties: {
+    type: 'string',
+    __validate: validateImage,
```
mischnic, line 28: I don't think this should be that strict. Maybe you could even have a pipeline that converts between image formats. In that case, these "invalid" formats should be allowed here.

packages/transformers/webext/src/schema.js
wachunei: Edge (which is the same API as chrome) [supports both `string` or `object`](https://developer.chrome.com/extensions/browserAction) as well. To maximize compatibility we should support both cases too.

packages/transformers/webext/src/schema.js
101arrowz: I hadn't considered other transformers. Will fix.

packages/transformers/webext/src/schema.js
101arrowz: [This](https://docs.microsoft.com/en-us/microsoft-edge/extensions/api-support/supported-manifest-keys#browser_action-or-page_action-keys) disagrees.

mischnic: This if for the "legacy" Edge up to version 44: https://docs.microsoft.com/en-us/microsoft-edge/extensions/api-support/supported-manifest-keys#browser_action-or-page_action-keys

This applies to the modern Edge based on Chromium: https://developer.chrome.com/extensions/browserAction

wachunei: Good catch, very interesting because the extension I'm testing against uses a string:

https://github.com/wachunei/directUC/blob/9583cbc832f802977437cfbc04aa0441b17d2bf9/src/manifest/base.json#L22

And it works (passes Edge store validations), it is [published here.](https://microsoftedge.microsoft.com/addons/detail/directuc/hnlkkfgdllbagiaommmbnagdbnciddfl).

From the set of restrictions though, should we just use the less restrictive? Since this is not a tool for manifest validations, in case Chrome/Edge enforces the array syntax should be up to the dev to update it. What do you think?

mischnic: > this is not a tool for manifest validations, ... should it be up to the dev to update it

👍

101arrowz: That's true. In hindsight, the massive schema may not have been necessary per se since validation isn't the goal, but I suppose it (mostly) doesn't hurt.

wachunei: I think it is essential actually, super helpful for the user to catch errors early in development 👌

wachunei:
After updating the `.ttf` imports the build succeeds but it does not work in the browser.

Screen Shot 2020-11-01 at 19 36 12


--- 

I'm going to leave here instructions to test against the extension I'm using to test in case anyone wants to dig deeper.

- Extension code is here: https://github.com/wachunei/directUC/ 
- It currently relies on parcel@v1 + [parcel-plugin-web-extension](https://github.com/kevincharm/parcel-plugin-web-extension)
- to build: clone and run `yarn && yarn build:firefox`, the build output will be in `build/firefox` folder

To test using this PR branch
- If you ran `yarn build:firefox` a `src/manifest.json` should be there, otherwise run `yarn manifest:firefox`
- prepend `url:` to font imports in [`src/js/ui/components/GlobalStyles.jsx`](https://github.com/wachunei/directUC/blob/develop/src/js/ui/components/GlobalStyles.jsx?rgh-link-date=2020-11-01T22%3A20%3A10Z#L2-L5) 
  ```js
  import RedHatTextRegular from "url:../../../fonts/RedHatText-Regular.ttf";
  import RedHatTextMedium from "url:../../../fonts/RedHatText-Medium.ttf";
  import RedHatDisplayRegular from "url:../../../fonts/RedHatDisplay-Regular.ttf";
  import RedHatDisplayMedium from "url:../../../fonts/RedHatDisplay-Medium.ttf";
  ```
- using the method decribed in https://github.com/parcel-bundler/parcel/pull/5304#issuecomment-720065829 run from the extension root:
  ```sh
  TARGET=firefox node /path/to/parcel/packages/core/parcel/src/bin.js build src/manifest.json --no-content-hash --no-cache --dist-dir build/webextransformer
  ```
- build output should be in `build/webextransformer` folder

mischnic:
The error I get (`parcelRequired is not a function` with `--no-minify`) seems to be caused by Parcel trying to deduplicate code between `options_ui.page`, `browser_action.default_popup`  and `background.scripts`. Changing the options passed to  `addDependency` doesn't seem to change anything.

101arrowz:
Tried `isIsolated` when adding the deps, but it seems to change nothing.

wachunei:
> Tried `isIsolated` when adding the deps, but it seems to change nothing.
> 
> EDIT: Forcing `isEntry: true` for everything seems to fix it. @mischnic Is that an acceptable solution?

with _for everything_ you mean every `addURLDependency` inside `collectDependencies`?

101arrowz:
Yes, I mean just that. Theoretically I could exclude the images and such, since they obviously can't really have this kind of issue (given it's caused by scope hoisting), but every script will have to be declared an entry and therefore will not have their path rewritten

mischnic:
There should definitely be a dependency option to do this with the side effects of `isEntry` (e.g. no hash in path). Not sure if `isIsolated`'s behaviour is a bug or intended (or what it exactly signifies).

101arrowz:
I couldn't really find any other way to do it, so I just used made all JS assets entries. It makes logical sense, but I don't like having half the assets retain the same name and the other half not doing so. Since this is a web extension, it may be appropriate to force all assets to retain their names, given that the versioning provided by fingerprints is unnecessary. I'm considering just hardcoding `isEntry: true` for all assets.

101arrowz:
By the way, I discovered that while testing HMR (which, surprisingly, is mostly OOB already) that Content-Security-Policy can reject HMR, so I'm going to add support for auto-patching CSP in development mode.

wachunei:
I tested the latest version and it works 👌, this is great!

101arrowz:
@mischnic There are two things I'm having trouble with automating: preventing the server from running and adding support for source maps. While the server running doesn't make it impossible to use the packager, it's certainly a waste of resources. Source mapping only works inline in web extensions, so I need a way to force source maps to be inline.

Beyond those (non-critical) features, this PR is ready for merging.

101arrowz:
Any chance this can get reviewed? I'd really love to be able to use this officially.

packages/configs/default/index.json
```diff
@@ -10,6 +10,8 @@
       "@parcel/transformer-js",
       "@parcel/transformer-react-refresh-wrap"
     ],
+    "manifest.json": ["@parcel/transformer-webext", "@parcel/transformer-json"],
```
mischnic, line 13:
```suggestion
    "manifest.json": ["@parcel/transformer-webext"],
```

The json transformer isn't needed here

packages/configs/default/index.json
```diff
@@ -60,6 +62,7 @@
     "*.css": "@parcel/packager-css",
     "*.js": "@parcel/packager-js",
     "*.ts": "@parcel/packager-ts",
+    "*.json": "@parcel/packager-json",
     "*.jsonld": "@parcel/packager-raw-url",
```
mischnic, line 66:
```suggestion
    "*.{json,jsonld}": "@parcel/packager-raw-url",
```

I don't see what your json packager does differently compared to that one

packages/transformers/webext/src/schema.js
```diff
@@ -0,0 +1,391 @@
...
+const arrStr = {
+  type: 'array',
+  items: {
+    type: 'string',
+  },
+};
+
+// This has *some* Chrome bias, but let's be real here...
+// It's mainly intended to be highly cross-browser compatible
+export default {
```
mischnic, line 41: Can you add a type assertion like this: `export default ({...}: SchemaEntity);`

packages/transformers/webext/src/WebExtTransformer.js
```diff
@@ -0,0 +1,261 @@
...
+        filePath: asset.filePath,
+      },
+      '@parcel/transformer-webext',
+      'Invalid Web Extension manifest',
+    );
+    return {
+      type: 'json-source-map',
+      version: '0.6.1',
+      program: map,
+    };
```
mischnic, line 246: Using `parse` doesn't have any benefit here, you can just move that code into `transform`.

packages/transformers/webext/src/WebExtTransformer.js
```diff
@@ -0,0 +1,261 @@
...
+              codeHighlights: [
+                ((getJSONSourceLocation(
+                  ptrs['/default_locale'],
+                  err,
+                ): any): DiagnosticCodeHighlight),
+              ],
+            },
+          },
+        ],
+      });
```
mischnic, line 72: 1. The error doesn't say where the locale folder should be
2. Don't just cast a flow error away: add your current message to the highlight and change the other message to something more top-level:

```js
        diagnostic: [
          {
            message: 'Failed to process WebExtension manifest',
            filePath,
            codeFrame: {
              codeHighlights: [
                {
                  ...getJSONSourceLocation(ptrs['/default_locale'], err),
                  message: `Localization directory${
                    err == 'value' ? ' for ' + program.default_locale : ''
                  } does not exist: ${relative(
                    dirname(filePath),
                    join(locales, program.default_locale),
                  )}`,
                },
              ],
            },
          },
        ],
      });
```

Bildschirmfoto 2020-11-14 um 23 29 24

packages/transformers/webext/src/WebExtTransformer.js
```diff
@@ -0,0 +1,261 @@
...
+      const dictFile = program.dictionaries[dict];
+      if (extname(dictFile) != '.dic') {
+        throw new ThrowableDiagnostic({
+          diagnostic: [
+            {
+              message: 'Dictionaries must be .dic files',
+              origin: '@parcel/transformer-webext',
+              filePath,
+              codeFrame: {
+                codeHighlights: [((sourceLoc: any): DiagnosticCodeHighlight)],
```
mischnic, line 119: (as above regarding the message)

packages/transformers/webext/src/WebExtTransformer.js
```diff
@@ -0,0 +1,261 @@
...
+    for (let i = 0; i < loc.length - 1; ++i) {
+      obj = obj[loc[i]];
+      if (!obj) break;
+    }
+    if (!obj) continue;
+    const parent = obj,
+      lloc = loc[loc.length - 1];
+    obj = obj[lloc];
+    if (!obj) continue;
+    if (typeof obj == 'string')
```
mischnic, line 170: Please make this more readable:
- the short variable names
- that `DEP_LOCS` variable has a very strange format. Maybe use objects instead of arrays?

packages/transformers/webext/src/WebExtTransformer.js
```diff
@@ -0,0 +1,261 @@
...
+      parent[lloc] = asset.addURLDependency(
+        // TODO: not this, for sure
+        (extname(obj) == '.json' ? 'raw:' : '') + obj,
```
mischnic, line 173: Is `raw` needed for other fields apart from `web_accessible_resources`? Not really a fan of this but we'll have to work around this for now...

packages/transformers/json/src/JSONTransformer.js
```diff
@@ -5,15 +5,22 @@ import json5 from 'json5';
...
-    // https://v8.dev/blog/cost-of-javascript-2019#json.
-    // Apply `JSON.stringify` twice to make it a valid string literal.
-    asset.setCode(
-      `module.exports = JSON.parse(${JSON.stringify(
-        JSON.stringify(json5.parse(await asset.getCode())),
-      )});`,
-    );
+    // This indicates a previous transformer (e.g. WebExt) has applied special
+    // handling to this already
+    if (asset.meta.handled) {
```
mischnic, line 10: Why is this needed, just for pretty formatting? In that case we could make `raw:...` "pretty" be default

packages/packagers/json/src/JSONPackager.js
```diff
@@ -0,0 +1,24 @@
+// @flow strict-local
+
+import assert from 'assert';
+import {Packager} from '@parcel/plugin';
+import {replaceURLReferences} from '@parcel/utils';
+
+export default (new Packager({
```
mischnic, line 7: As I've written in the default config, this is probably not needed.

packages/core/integration-tests/test/webext.js
```diff
@@ -0,0 +1,39 @@
...
+      {assets: ['foo.png']},
+      {assets: ['popup.html']},
+      {assets: ['devtools.html']},
+      {assets: ['content.js']},
+      {assets: ['content.css']},
+      {assets: ['background.js']},
+    ]);
+  });
+  // TODO: Test error-checking
+  // TODO: Test certain types of plugins fully rather than just one with everything
```
mischnic, line 38: 1. Yes, tests for errors would indeed be great
2. I think most of the settings in the manifest don't influence each other, so I think this isn't really necessary...

packages/configs/default/index.json
101arrowz: It actually is useful in case someone has a file named `manifest.json` that isn't a web app manifest. Thoughts?

packages/configs/default/index.json
101arrowz: Removes absolute paths and uses relative ones only

packages/transformers/webext/src/WebExtTransformer.js
101arrowz: It allows us to use another JSON transformer if we have a file that happens to be named `manifest.json` (for example, a web app manifest).

packages/packagers/json/src/JSONPackager.js
101arrowz: [Rationale](https://github.com/parcel-bundler/parcel/pull/5304#discussion_r523477732)

packages/transformers/json/src/JSONTransformer.js
101arrowz: If we don't do this, we don't allow users to have files named `manifest.json` that aren't web extension manifests.

packages/transformers/webext/src/WebExtTransformer.js
101arrowz: This is needed for anything with JSON because the default JSON transformer transforms JSON to javascript. In terms of WebExt, this is needed for that and `storage.managed_schema`.

packages/transformers/webext/src/WebExtTransformer.js
101arrowz: How would you propose I make `DEP_LOCS` use objects?

packages/transformers/webext/src/WebExtTransformer.js
```diff
@@ -0,0 +1,261 @@
...
+function cspPatchHMR(policy: ?string) {
+  if (policy) {
+    const csp = parseCSP(policy);
+    policy = '';
+    if (!csp['script-src']) {
+      csp['script-src'] = ["'self'", "'unsafe-eval'", 'blob: filesystem:'];
+    }
+    if (!csp['object-src']) {
+      csp['object-src'] = ["'self' blob: filesystem:"];
+    }
```
101arrowz, line 208: I don't think this is necessary and am not sure why I added it.

packages/transformers/webext/src/WebExtTransformer.js
mischnic:
```js
const DEP_LOCS = [
  {path: ['icons']},
  {path: ['browser_action', 'default_icon']}
  {path: ['browser_action', 'default_popup'], isEntry: true},
```

mischnic:
1. To resolve the "someone has a file named manifest.json that isn't a web app manifest" situation and the workarounds that are needed to make this work, we should instead have a `@parcel/config-webextension` with something like this:
```js
{
  "transformers": {
    "manifest.json": ["@parcel/transformer-webextension"],
    "raw:*": ["@parcel/transformer-raw"]
  }
}
```

Then the current json transformer doesn't have to be modified. We plan to properly fix the "every JSON is transformed to JS" in the near future (then that `raw:` pipeline could be removed entirely).

2. Is `isEntry` fine at the moment for files specified in manifest.json? I think hashes don't provide any benefit here anyway?

101arrowz:
1. I agree, do you want me to make the JSON transformer switch in this PR or just wait for it to land?
2. Yes, `isEntry` is fine for all files in the Web Extension because they have custom versioning. Looking at the code again, I'm not quite sure why I didn't just do that.

mischnic:
> just wait for it to land

Not sure when that will be solved. I would say that we should use the config as I've posted above (and I think you can revert your changes to the JSON transformer) so this PR isn't blocked.

> do you want me to make the JSON transformer switch

Not sure what you mean there, does my previous answer clear this up?

101arrowz:
I was just wondering if you wanted me to switch to the new JSON transformer or wait for an actual core maintainer to do it, then switch, but you already answered that.

I've been quite busy over the past few days, I can probably work on this next week.

mischnic:
> create and switch to the new JSON transformer

(This will probably be solved in the parcel config level, not in a transformer: https://github.com/parcel-bundler/parcel/pull/5225/files#diff-3ca400b6b91bc49c60c3c75951d144849778233b303de0a706863daa69bc50d6)

fregante:
I finally got around to trying this PR on https://github.com/npmhub/npmhub by following the instructions in https://github.com/parcel-bundler/parcel/pull/5304#issuecomment-720065829

---

The `persistent` property isn't required by the browsers and parcel, but:

```js
🚨 Build failed.
@parcel/transformer-webext: Invalid Web Extension manifest
/source/manifest.json:22:17
  21 |   "minimum_chrome_version": "80.0",
> 22 |   "background": {
>    |                 ^
> 23 |     "scripts": [
>    | ^^^^^^^^^^^^^^^^
> 24 |       "background.js"
>    | ^^^^^^^^^^^^^^^^^^^^^
> 25 |     ]
>    | ^^^^^
> 26 |   },
>    | ^^^ Missing property persistent
  27 |   "permissions": [
  28 |     "https://registry.npmjs.org/",
```

My extension didn't use Babel, but I get this issue (I do use Svelte though)

```js
🚨 Build failed.
Error: Cannot find module '@babel/core' from /npmhub/source'
Error: Cannot find module '@babel/core' from /npmhub/source'
    at /parcel/node_modules/resolve/lib/async.js:107:35
    at processDirs (/parcel/node_modules/resolve/lib/async.js:260:39)
    at callback (/parcel/node_modules/resolve/lib/async.js:267:32)
    at isDirectory (/parcel/packages/core/utils/src/resolve.js:117:9)
```

---

Unfortunately my testing stops here because of https://github.com/parcel-bundler/parcel/issues/4840

fregante:
However, I updated my other extension and it seems to work without issues! https://github.com/parcel-bundler/parcel/pull/5304

101arrowz:
Thanks for the reminder @fregante, I plan to fully fix this up in 2-3 days but right now I can't unfortunately. I'll fix everything mischnic suggested and remove the persistent property.

101arrowz:
I've incorporated the changes @mischnic but there's a bug with the ParcelConfigRequest, and it seems that the new config doesn't work at all. Could you take a look at the latest commit?

mischnic:
You need to an empty file here: `test/integration/webext/yarn.lock` (.parcelrc are only used when in the project root).

101arrowz:
This is confusing, there wasn't an issue with named pipelines in `addURLDependency` before I merged with upstream.

mischnic:
So apparently the argument to `addURLDependency` should always be a proper URL, so you should do this instead:
```js
      asset.addURLDependency(path.posix.join('_locales', locale, 'messages.json'), {
        isEntry: true,
        pipeline: 'raw',
      });
```
(and for the other cases as well). We really need some better docs here...

(Also `path.join` uses backslashes on Windows, and URLs use forward slashes, so `path.posix.join` does what we want.)

Change the types so that flow passes:
```diff
diff --git packages/core/types/index.js packages/core/types/index.js
index ca56d39a..ad88137a 100644
--- packages/core/types/index.js
+++ packages/core/types/index.js
@@ -417,6 +417,7 @@ export type DependencyOptions = {|
   +loc?: SourceLocation,
   +env?: EnvironmentOpts,
   +meta?: Meta,
+  +pipeline?: string,
   +target?: Target,
   +symbols?: $ReadOnlyMap<
     Symbol,
```

This was already implemented but never listed in the public API for some reason...

Let me know if you run into any other problems

101arrowz:
Right, thanks for the clarification. I'll fix that soon. On a similar note, I wish Parcel were better documented in all honesty; I wasn't really able to use Parcel 2 before spending a few hours reading the codebase, and making plugins has been similarly painful. It's been more "setup" than Webpack, but I believe Parcel 2 has the potential to offer an actual replacement for Webpack if it's more accessible. Since it's already in the beta phase, I'd consider that a top priority.

101arrowz:
Facing one issue @mischnic, for some reason there's still a hash/fingerprint in the generated "URL" after specifying the pipeline as either `raw` or `url`. Not sure why.

mischnic:
And even worse, because isEntry is false, the path in `_locale/..../` isn't preserved...

Not sure yet why that's there,  the tests still pass when removing the `asset.isIsolated` condition....

Can you commit your changes anyway (even if the tests/CI fails), so that we're on the same page?

101arrowz:
I'm not sure why the `isIsolated` flag was added to raw assets, but I'm completely confused as to why `isIsolated` being true forces `isEntry` to be true. Shouldn't isolated assets always be entries, not never?

101arrowz:
@mischnic Any ideas? Is simply removing the `isIsolated` check acceptable, or should we make `isIsolated` force `isEntry` to true instead of false, or should we remove the addition of the attribute from the transformer?

mischnic:
Removing the `isIsolated` condition https://github.com/parcel-bundler/parcel/blob/064ba0a166ea63ccd2b623dc11324d4cd7ffb526/packages/bundlers/default/src/DefaultBundler.js#L105 is probably the way to go (and all tests still pass). I hope to know in the next couple day why the check is there in the first place (and if there's a case not covered by our tests).

101arrowz:
I've pushed a version that passes all tests and uses the `config-webext` you mentioned. Hopefully once `isIsolated` is cleared up, this can be merged.

packages/runtimes/hmr/src/loaders/hmr-runtime.js
```diff
@@ -40,7 +40,7 @@ var parent = module.bundle.parent;
 if ((!parent || !parent.isParcelRequire) && typeof WebSocket !== 'undefined') {
   var hostname = getHostname();
   var port = getPort();
-  var protocol = location.protocol === 'https:' ? 'wss' : 'ws';
+  var protocol = HMR_SECURE ? 'wss' : 'ws';
```
101arrowz, line 43: This is sketchy if a user is, for example, running through ngrok. I couldn't think of any other way to make content scripts HMR work. Perhaps the behavior can be rewritten to choose HTTPS if `(HMR_SECURE || (location.protocol == 'https:' && hostname != 'localhost'))`.

packages/configs/default/index.json
```diff
@@ -60,7 +60,7 @@
     "*.css": "@parcel/packager-css",
     "*.js": "@parcel/packager-js",
     "*.ts": "@parcel/packager-ts",
-    "*.jsonld": "@parcel/packager-raw-url",
+    "*.{json,jsonld}": "@parcel/packager-raw-url",
```
mischnic, line 63: Please move this to the webextension transformer instead:
```diff
--- packages/configs/webext/index.json
+++ packages/configs/webext/index.json
@@ -2,5 +2,8 @@
   "extends": "@parcel/config-default",
   "transformers": {
     "manifest.json": ["@parcel/transformer-webext"]
+  },
+  "packagers": {
+    "manifest.json": "@parcel/packager-raw-url"
   }
 }
```

packages/transformers/webext/src/WebExtTransformer.js
```diff
@@ -0,0 +1,244 @@
...
+      });
+    else {
+      for (const k of Object.keys(obj)) {
+        obj[k] = asset.addURLDependency(obj[k], {
+          isEntry: true,
+          loc: {
+            filePath,
+            ...getJSONSourceLocation(ptrs[location + '/' + k], 'value'),
+          },
+          ...(extname(obj[k]) == '.json' && {pipeline: 'url'}),
```
mischnic, line 187: do this instead: `pipeline: condition ? ... : undefined`

packages/configs/webext/package.json
```diff
@@ -0,0 +1,21 @@
...
+    "url": "https://opencollective.com/parcel"
+  },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/parcel-bundler/parcel.git"
+  },
+  "main": "index.json",
+  "dependencies": {
+    "@parcel/config-default": "2.0.0-beta.1",
+    "@parcel/transformer-webext": "2.0.0-beta.1"
```
mischnic, line 19: `@parcel/packager-raw-url` is missing here

packages/runtimes/hmr/src/loaders/hmr-runtime.js
```diff
@@ -40,7 +40,12 @@ var parent = module.bundle.parent;
 if ((!parent || !parent.isParcelRequire) && typeof WebSocket !== 'undefined') {
   var hostname = getHostname();
   var port = getPort();
-  var protocol = location.protocol === 'https:' ? 'wss' : 'ws';
+  var protocol =
+    HMR_SECURE ||
+    (location.protocol == 'https:' &&
+      !/localhost|127.0.0.1|0.0.0.0/.test(hostname))
+      ? 'wss'
+      : 'ws';
```
mischnic, line 48: Why is this change needed?

101arrowz: It was necessary to make content scripts HMR work. Effectively, on an HTTPS site, say `https://google.com` the runtime previously tried to access `wss://google.com:1234` by default. If you set the host to `localhost` from the command line, you'd still get `wss://localhost:1234`, which didn't work. Now, the system will only access `wss` if either `HMR_SECURE` was set from the server, or we're not on a local network. One would assume that on a local network, proxies that add HTTPS support are rare. Therefore, this change does not break any environments but adds support for accessing the correct URL with content scripts.

101arrowz:
@mischnic Could you take another look?

101arrowz:
I've been experimenting with this branch locally and found that after reloading a content script, the HMR changes aren't saved because the extension internally still reports the old version. To solve this, I need to attach some code to the content scripts and either create or add code to the background scripts, but I'm not sure how to do this. If I add a runtime, it fails when there is no background script. If I try to add the dependencies from the transformer (with `addURLDependency` and a local file), it fails because Parcel can't resolve the local file `parcel/transformers/webext/src/autoreload.js` from the project directory. Any ideas @mischnic?

mischnic:
> t fails when there is no background script.

You could always add a script if there is none (similar to https://github.com/parcel-bundler/parcel/pull/5414)? But how do you plan to restrict that runtime to only append to that background script and not all JS bundles?

> can't resolve the plugin file from the project directory

That should be solved with https://github.com/parcel-bundler/parcel/pull/4366#discussion_r540677355

101arrowz:
Well, I managed to do it by effectively injecting an import with `path.relative` from `asset.filePath`, not sure if it's acceptable but it's the only way I could find to do it.

101arrowz:
Take another look @mischnic?

packages/core/utils/.eslintrc.js
```diff
@@ -5,16 +5,16 @@ const RESTRICTED_CONFIG = [
...
 ];
 
 module.exports = {
   extends: '@parcel/eslint-config',
   rules: {
     'no-restricted-imports': RESTRICTED_CONFIG,
-    'no-restricted-modules': RESTRICTED_CONFIG
-  }
+    'no-restricted-modules': RESTRICTED_CONFIG,
+  },
```
fregante, line 19: This PR includes a lot of unrelated stylistic changes probably due to prettier. I think these should be reverted (this isn't the only file)

101arrowz: `yarn.lock` is necessary for the integration test to pass because Parcel looks for a lockfile when resolving the root directory of a project. [See here](https://github.com/parcel-bundler/parcel/pull/5304#issuecomment-731567435).

Also, the Prettier changes I agree shouldn't have happened but I'm not sure it's all that important that they're removed. I'll do it if necessary.

packages/core/parcel/package.json
```diff
@@ -14,7 +14,7 @@
...
   "main": "lib/bin.js",
   "source": "src/bin.js",
```
mischnic, line 19: This should be reverted

packages/core/utils/.eslintrc.js
fregante: Honestly I’m surprised that CI doesn’t complain. Is there no code linter?

mischnic: Our `lint`  npm script only checks `./packages/*/*/{src,bin,test}/**/*.{js,json,md}` with prettier. Some time ago we enabled trailing commas and apparently only the files matching the glob were converted.

packages/core/parcel/package.json
101arrowz: Whoops.

packages/core/utils/.eslintrc.js
101arrowz: So should these changes be undone or left as-is?

packages/core/parcel/package.json
```diff
@@ -37,4 +37,4 @@
   "devDependencies": {
     "@parcel/babel-register": "2.0.0-beta.1"
   }
-}
+}
```
mischnic, line 40: ...

packages/core/utils/.eslintrc.js
mischnic: You can leave them in.

packages/core/parcel/package.json
fregante: This is one of the unnecessary changes I was talking about. 

If either one of you is using Refined GitHub, you might be able to undo every file that has unrelated changes (however if the change is “add a trailing newline”, I think that it won’t be able to undo it)

101arrowz: This is my fault, I never add newlines at the end of files and my code editor (VSCode) refuses to as well. Sorry for all the trivial mistakes.

101arrowz:
Is there anything else that needs to be fixed before this can be merged?

101arrowz:
I hate to keep bothering the maintainers, but because this PR introduces so many changes, it's getting increasingly difficult to keep up with merge conflicts. @mischnic if you have time, I'd really appreciate if you could take another look soon.

packages/transformers/webext/src/WebExtTransformer.js
```diff
@@ -0,0 +1,273 @@
...
+          });
+        }
+      }
+      if (hot && sc.js && sc.js.length) {
+        needRuntimeBG = ptrs[`/content_scripts/${i}/js/${sc.js.length - 1}`];
+        sc.js.push(
+          asset.addURLDependency(
+            relative(dirname(asset.filePath), RUNTIME),
+            {},
+          ),
```
mischnic, line 111: Use `resolveFrom` instead of the relative path: https://github.com/parcel-bundler/parcel/blob/aff1426c88f465cdea1a977c4b6633417e61eb69/packages/transformers/js/src/visitors/modules.js#L287-L295

packages/transformers/webext/src/WebExtTransformer.js
```diff
@@ -0,0 +1,273 @@
...
+              ...getJSONSourceLocation(
+                ptrs[`/content_scripts/${i}/${k}/${j}`],
+                'value',
+              ),
+            },
+          });
+        }
+      }
+      if (hot && sc.js && sc.js.length) {
+        needRuntimeBG = ptrs[`/content_scripts/${i}/js/${sc.js.length - 1}`];
```
mischnic, line 106: You're using an entry in the pointers array as a boolean?

packages/transformers/webext/src/WebExtTransformer.js
```diff
@@ -0,0 +1,273 @@
...
+    const csp = parseCSP(policy);
+    policy = '';
+    if (!csp['script-src']) {
+      csp['script-src'] = ["'self' 'unsafe-eval' blob: filesystem:"];
+    }
+    if (!csp['script-src'].includes("'unsafe-eval'")) {
+      csp['script-src'].push("'unsafe-eval'");
+    }
+    for (const k in csp) {
+      policy += k + ' ' + csp[k].join(' ') + ';';
```
mischnic, line 229:
```suggestion
      policy += `${k} ${csp[k].join(' ')};`;
```

packages/transformers/webext/src/WebExtTransformer.js
```diff
@@ -0,0 +1,273 @@
...
+
+export default (new Transformer({
+  async transform({asset, options}) {
+    const code = await asset.getCode();
+    const parsed = jsm.parse(code);
+    const data: any = parsed.data;
+    if (BASE_KEYS.some(key => !data.hasOwnProperty(key))) {
+      // This is probably just another file that happens to be named manifest.json
+      return [asset];
+    }
```
mischnic, line 248: I think this can be removed now that we have a dedicated webextension config?

packages/transformers/webext/src/runtime/autoreload.js
```diff
@@ -0,0 +1,6 @@
+/* global window */
+window.addEventListener('beforeunload', function() {
```
mischnic, line 2: Would `module.hot.dispose` work here?

101arrowz: If we did that, the module would reload on every hot reload, which is not only expensive but also means the web extension itself won't have true HMR. I've tested this locally and this is the best solution I found.

mischnic:
1. Please rename the config and transformer packages to `-webextension` instead of `-webext`
2. eslint fails

No blocker, just a heads up: this here https://github.com/101arrowz/parcel/blob/transformer-webext/packages/transformers/webext/src/WebExtTransformer.js#L154-L169 doesn't cause a rebuild if a file is added, but we'll need to extend the API for that. For now you'll have to clear the cache if a file matching that glob is removed or added.

101arrowz:
Could you describe how exactly to clear cache whenever the glob updates? I would have assumed that if the transformer were running, the cache would already be expecting a refresh...

mischnic:
> Could you describe how exactly to clear cache whenever the glob updates? I would have assumed that if the transformer were running, the cache would already be expecting a refresh...

`rm -rf .parcel-cache` 😬 

The problem is that the transformer *doesn't* rerun if you add/remove files that are specified by the glob. Only if the file itself changes.

mischnic (approved):

mischnic:
Thank you and sorry for the long delay. Ideally this would have been split up into the modifications of the schema util and then the transformer itself. The various reformattings also make it hard to read...

101arrowz:
It's not a problem. I know I was being pretty pushy with this PR and that others have also had to wait a while to get PRs merged. Thank you for all your efforts - not just here, but generally on this project. I appreciate it a lot ❤️

fregante:
Wooot! Looking forward to upgrading my projects to parcel v2!

mgranderath:
I was just testing it out in a Typescript project and everything works great, however, it seems that while the `web_accessible_resources` do get compiled, the actual file endings in the output `manifest.json` are not changed to be `.js`. I think that is a bug but let me know if it is actually correct. If you want me to I can open an issue and build minimal reproducing example in case it's a bug.

101arrowz:
Yeah, I just realized that's an issue with the system. I can fix it soon. Out of curiosity, why are you using JS files in `web_accessible_resources`? Can't you just import normally in content scripts?

mgranderath:
We use it to inject scripts into the page "context" since content scripts cannot access all properties of the `window` object. We usually use it along the lines of 
```javascript
// Inject the page context script
const script = document.createElement("script");
script.src = browser.runtime.getURL("content/page/page-context.js");
if (document.documentElement) {
  document.documentElement.appendChild(script);
}
```

101arrowz:
That's reasonable, I'll create a new PR with the fix at some point.

mgranderath:
Amazing! Thanks for all of your work on this! 💯

fregante:
I'm not able to use this in the nightly of Parcel, how am I supposed to use it?

`npm i parcel@nightly` does not install the `@parcel/transformer-webextension` package. Even if I manually install `@parcel/transformer-webextension`, parcel only generates `manifest.js`

```
❯ npx parcel --version
2.0.0-nightly.540+bdc45e49

❯ npm ls @parcel/transformer-webextension
notifications-preview-github@ ./
└── @parcel/transformer-webextension@2.0.0-nightly.2164+bdc45e49

❯ npx parcel build extension/manifest.json 
✨ Built in 1.21s

dist/manifest.js                     841 B    161ms
├── extension/manifest.json          762 B    761ms
└── Code from unknown sourcefiles     79 B      0ms
```

101arrowz:
@fregante check the docs [here](https://v2.parceljs.org/recipes/web-extension/).

Also, sorry for the delays on the issues you were facing @mgranderath, I'll draft a PR ASAP.

ranisalt:
Will this play well along non-extension `manifest.json` files? I currently name my manifest files `*.webmanifest` to avoid them being packaged as JSON by parcel. I wonder if this will work for any manifest file or it will just break it in another way.

101arrowz:
Yes, you need to enable a config for this to apply. The web manifest transformer is [still in the works](https://github.com/parcel-bundler/parcel/pull/4313), but at this point there are so many merge conflicts that I doubt it'll ever get merged. We'll probably need to create a new PR.

fregante:
This is working great by the way! I opened a feature request about another API used to register scripts and styles: https://github.com/parcel-bundler/parcel/issues/5758

PR #5330, opened 2020-11-09
Added package def for sinon

AGawrys:
Agawrys/P2X-321-flow-types-for-sinon-pkg

# ↪️ Pull Request

This PR adds `sinon` to flow-types

See also PR made to [sinon flow type](https://github.com/flow-typed/flow-typed/pull/3934) for relevant changes

## 🚨 Test instructions

Can test imports by adding` // flow strict enable` to files that use `sinon`

Also running ` yarn flow check` and seeing there are 0 errors

### DIFF
--- a/packages/core/core/test/ParcelConfig.test.js
+++ b/packages/core/core/test/ParcelConfig.test.js
@@ -1,4 +1,6 @@
-// @flow
+// @flow strict-local
+
+// flowlint untyped-import:off
 import ParcelConfig from '../src/ParcelConfig';
 import assert from 'assert';
 import path from 'path';
@@ -189,7 +191,6 @@ describe('ParcelConfig', () => {
         inputFS,
         false,
       );
-
       let parcelVersion = require('../package.json').version;
       let pkgJSON = path.join(
         __dirname,
--- a/packages/core/core/test/ParcelConfig.test.js
+++ b/packages/core/core/test/ParcelConfig.test.js
@@ -190,6 +190,7 @@ describe('ParcelConfig', () => {
         inputFS,
         false,
       );
+      // $FlowFixMe
       let parcelVersion = require('../package.json').version;
       let pkgJSON = path.join(
         __dirname,
--- a/packages/core/core/test/ParcelConfig.test.js
+++ b/packages/core/core/test/ParcelConfig.test.js
@@ -1,6 +1,5 @@
 // @flow strict-local
 
-// flowlint untyped-import:off
 import ParcelConfig from '../src/ParcelConfig';
 import assert from 'assert';
 import path from 'path';
--- a/packages/core/integration-tests/test/hmr.js
+++ b/packages/core/integration-tests/test/hmr.js
@@ -1,4 +1,4 @@
-// @flow
+// @flow strict-local
 import assert from 'assert';
 import path from 'path';
 import {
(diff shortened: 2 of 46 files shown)


AGawrys:
@wbinnssmith Sinon definition updated 😀

packages/core/integration-tests/test/hmr.js
```diff
@@ -9,6 +9,7 @@ import {
   overlayFS,
   sleep,
 } from '@parcel/test-utils';
+// flowlint untyped-import:off
```
wbinnssmith, line 12:
```suggestion
// flowlint-next-line untyped-import:off
```

packages/core/core/test/ParcelConfig.test.js
```diff
@@ -1,4 +1,6 @@
-// @flow
+// @flow strict-local
+
+// flowlint untyped-import:off
```
wbinnssmith, line 3: why is this needed?

packages/core/integration-tests/test/hmr.js
```diff
@@ -23,7 +24,7 @@ async function closeSocket(ws: typeof WebSocket) {
   ws.close();
   await new Promise(resolve => (ws.onclose = resolve));
 }
-
+// flowlint-next-line unclear-type:off
```
wbinnssmith, line 27: This should be resolvable once we have typings for the ws package :)

packages/core/integration-tests/test/hmr.js
AGawrys: The reason I didn't include the `next-line `here is that I wanted the rule applied for the next 4 lines, which all are untyped imports. the flowlint would only apply to the rest of the file, and not before so it wouldn't affect the previous imports.

packages/core/core/test/ParcelConfig.test.js
AGawrys: ahh mistake

packages/core/integration-tests/test/hmr.js
AGawrys: Do you suggest removing this?

packages/core/core/test/ParcelConfig.test.js
wbinnssmith: what remains untyped in this file?

packages/core/integration-tests/test/hmr.js
wbinnssmith: not now — when we add typings for ws :)

packages/core/integration-tests/test/hmr.js
wbinnssmith: Ah, got it. Personally I would suppress each import individually, as this suppresses the violation for any imports added after this line in the future. But this is fine to me, especially if we can get typings for the remaining packages and remove it soon.

thoughts @padmaia @mischnic?

mischnic: > Personally I would suppress each import individually, as this suppresses the violation for any imports added after this line in the future.

👍

packages/core/package-manager/test/NodePackageManager.test.js
```diff
@@ -1,5 +1,6 @@
-// @flow
+// @flow strict-local
 
+// flowlint untyped-import:off
```
wbinnssmith, line 3: same here

packages/core/integration-tests/test/hmr.js
wbinnssmith: Another option might be to group all the untyped imports and enable the lint rule afterwards with `// flowlint untyped-import:error`.

wbinnssmith (approved):

PR #5340, opened 2020-11-11
Add flow types for browserslist

thebriando:
# ↪️ Pull Request
Adds a flow type lib def for `browserslist`.

## 🚨 Test instructions

Run `yarn flow` to ensure that there are no flow errors.

@padmaia @wbinnssmith

### DIFF
--- a/flow-libs/browserslist.js.flow
+++ b/flow-libs/browserslist.js.flow
@@ -1,0 +1,21 @@
+// @flow strict-local
+
+declare module 'browserslist' {
+  import type {ConfigResult} from '@parcel/types';
+
+  declare type BrowserslistConfig = {|[string]: string | Array<string>|};
+  // declare export type Brwos = {
+  //   parseConfig: parseConfig,
+  //   loadConfig: loadConfig,
+  //   ...
+  // }
+  // declare export default (
+  //   parseConfig,
+  //   loadConfig,
+  // ) => Array<string>;
+  declare export class BrowsersList extends Array<string> {
+    constructor(browsers: string | Array<string>): Array<string>;
+    static parseConfig(configResult: ConfigResult): Promise<?BrowserslistConfig>;
+    static loadConfig({|opts: any|}): any;
+  }
+}
--- a/flow-libs/browserslist.js.flow
+++ b/flow-libs/browserslist.js.flow
@@ -1,13 +1,21 @@
 // @flow strict-local
 
 declare module 'browserslist' {
-  import type {ConfigResult} from '@parcel/types';
-
   declare type BrowserslistConfig = {|[string]: string | Array<string>|};
+  declare type Options = {
+    path?: string,
+    env?: string,
+    stats?: string,
+    config?: string,
+    ignoreUnknownVersions?: boolean,
+    dangerousExtend?: boolean,
+    mobileToDesktop?: boolean,
+    ...
+  };
 
   declare module.exports: {|
-    (browsers: string | Array<string>): Array<string>,
-    parseConfig(configResult: ConfigResult): Promise<?BrowserslistConfig>,
-    loadConfig({[path: FilePath]: FilePath, ...}): string,
+    (query: string | Array<string>, opts?: Options): Array<string>,
+    parseConfig(configResult: string): BrowserslistConfig,
+    loadConfig(opts: {[path: string]: string, ...}): string,
   |};
 }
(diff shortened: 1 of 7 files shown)


flow-libs/browserslist.js.flow
```diff
@@ -0,0 +1,13 @@
+// @flow strict-local
+
+declare module 'browserslist' {
+  import type {ConfigResult} from '@parcel/types';
+
+  declare type BrowserslistConfig = {|[string]: string | Array<string>|};
+
+  declare module.exports: {|
+    (browsers: string | Array<string>): Array<string>,
+    parseConfig(configResult: ConfigResult): Promise<?BrowserslistConfig>,
```
wbinnssmith, line 10: looks like this takes a string: https://github.com/browserslist/browserslist/blob/88f7d5141f9be0ac3d712f8b52f4654fadbdeeff/node.js#L260

flow-libs/browserslist.js.flow
```diff
@@ -0,0 +1,13 @@
...
+    loadConfig({[path: FilePath]: FilePath, ...}): string,
```
wbinnssmith, line 11: not sure how `FilePath` works here. A string is probably better, or we could define `FilePath` inside this module as an alias to `string`.

wbinnssmith, line 11: the first argument, options, is defined in browserlist's js api docs: https://github.com/browserslist/browserslist#js-api

packages/core/core/src/public/Environment.js
```diff
@@ -86,6 +86,7 @@ export default class Environment implements IEnvironment {
     return this;
   }
 
+  // flowlint unsafe-getters-setters: off
```
wbinnssmith, line 89: this should probably be hoisted to the top of the file

packages/transformers/babel/src/getBabelTargets.js
```diff
@@ -66,6 +66,7 @@ export default async function getBabelTargets(
       browsers != null &&
       !Array.isArray(browsers)
     ) {
+      // flowlint sketchy-null-string: off
       let env = process.env.NODE_ENV || 'development';
```
wbinnssmith, line 70: @padmaia @devongovett looks like this is using the real environment's `NODE_ENV`. 

@thebriando `process.env.NODE_ENV ?? 'development` addresses this, though it changes the semantics, probably for the better. In the future, we can avoid disabling lint by making the falsy check explicit like `(foo !== '' && foo != null) ? foo : 'fallback'`

packages/transformers/babel/src/getBabelTargets.js
```diff
@@ -66,11 +66,11 @@ export default async function getBabelTargets(
       browsers != null &&
       !Array.isArray(browsers)
     ) {
-      let env = process.env.NODE_ENV || 'development';
+      let env = process.env.NODE_ENV ?? 'development';
       browsers = browsers[env] || browsers.defaults;
     }
 
-    if (browsers) {
+    if (browsers !== null && browsers !== undefined) {
```
wbinnssmith, line 73:
```suggestion
    if (browsers != null) {
```

flow-libs/browserslist.js.flow
```diff
@@ -0,0 +1,21 @@
+// @flow strict-local
+
+declare module 'browserslist' {
+  declare type BrowserslistConfig = {|[string]: string | Array<string>|};
+  declare type Options = {
```
wbinnssmith, line 5: can this be exact?

packages/core/core/src/public/Environment.js
```diff
@@ -1,4 +1,5 @@
-// @flow
+// @flow strict-local
+// flowlint unsafe-getters-setters: off
```
wbinnssmith, line 2: @devongovett @mischnic what do you think of allowing getters and setters globally? They do technically introduce potential unsoundness.

mischnic: 👍  We do use them extensively in the public interfaces. And I think as long as we keep them side-effect-free (which no linter can verify anyway), it should be fine regarding soundness.

flow-libs/browserslist.js.flow
```diff
@@ -0,0 +1,20 @@
...
+    config?: string,
+    ignoreUnknownVersions?: boolean,
+    dangerousExtend?: boolean,
+    mobileToDesktop?: boolean,
+  |};
+
+  declare module.exports: {|
+    (query?: string | Array<string>, opts?: Options): Array<string>,
+    parseConfig(configResult: string): BrowserslistConfig,
+    loadConfig(opts: {[path: string]: string, ...}): string,
```
wbinnssmith, line 18: can this be exact?

wbinnssmith (approved):

Issue #5344, opened 2020-11-12
Parcel v2: Setting composite to true breaks transformer-typescript-types


paeolo:
# 🐛 bug report

`@parcel/transformer-typescript-types` is broken when setting `composite: true` in the `compilerOptions` of `tsconfig.json`
- When building with the example provided below I have this error:
```console
@parcel/transformer-typescript-types: File '~/parcel-issue-6786/src/add.ts' is not listed within the file list of project ''. Projects must list all files or use an 'include' pattern.
/home/paul/parcel-issue-6786/src/index.ts:1:15
> 1 | export * from './add';
>   |               ^^^^^^^^ File '/home/paul/parcel-issue-6786/src/add.ts' is not listed within the file list of project ''. Projects must list all files or use an 'include' pattern.
  2 | export default 'Hello World!';
```
- Looking at the `index.d.ts` file there are no types emitted.
## 🎛 Configuration (.babelrc, package.json, cli command)

https://github.com/paeolo/parcel-issue-5344

## 🤔 Expected Behavior

- `transformer-typescript-types` should make the types as if `composite: true` were not set. [composite](https://www.typescriptlang.org/docs/handbook/project-references.html#composite) shouldn't have an impact on type generation.
- Adding (patching) `composite: false` to [this line](https://github.com/parcel-bundler/parcel/blob/3bfef2821b6fa8031e8ec7778132581296266faa/packages/transformers/typescript-types/src/TSTypesTransformer.js#L34) does the trick for me. Types are emitted correctly. But maybe the fix is somewhere else;)

mischnic:
> Adding (patching) composite: false to this line does the trick for me

AFAICT, that sounds good

paeolo:
@mischnic fine;) Do you want me to do a PR for this?

mischnic:
Sure

PR #5346, opened 2020-11-12
fix: typescript createProgram doesn't support incremental mode

paeolo:
# ↪️ Pull Request

This PR solves https://github.com/parcel-bundler/parcel/issues/5344

`transformer-typescript-types` uses the typescript API through the `ts.createProgram` function. But this function is not meant to be used with [composite](https://www.typescriptlang.org/docs/handbook/project-references.html#composite). Indeed, the typescript's team provide another API specifically for this case: [`createIncrementalProgram`](https://github.com/microsoft/TypeScript/pull/31432).

Incremental build is irrelevant for Parcel here, but on the other hand the user could have a `tsconfig.json` with composite enabled, for instance to enable some `vscode` features (ie. project references). So we should make sure that `composite` is set to false.

## 🚨 Test instructions

I provided an example repository to test the fix: https://github.com/paeolo/parcel-issue-5344. But the idea is juste to set `composite: true` in the `compilerOptions` and have a non-empty `index.d.ts` when you build.

## ✔️ PR Todo

### DIFF
--- a/packages/core/integration-tests/test/integration/ts-types/composite/add.ts
+++ b/packages/core/integration-tests/test/integration/ts-types/composite/add.ts
@@ -1,0 +1,1 @@
+export const add = (a: number, b: number) => a + b;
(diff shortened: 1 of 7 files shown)


DeMoorJasper:
Could you add a test case?

paeolo:
@DeMoorJasper I added a test based on the minimal repro I linked in the issue. I also made sure the test fails without the fix and pass with it, see the latest commit;)

DeMoorJasper (approved):
awesome, thanks

PR #5357, opened 2020-11-17
Fix undefined dev dependency versions

thebriando:
# ↪️ Pull Request

At some point a bug was introduced where we are not actually resolving versions of the Parcel plugins we use.

## 💻 Examples

This is the log for one of the configs after running `yarn demo` in the `simple` example project. The versions are undefined for each devDep.

```
console: Map {
  '@parcel/transformer-react-refresh-babel' => undefined,
  '@parcel/transformer-babel' => undefined,
  '@parcel/transformer-js' => undefined,
  '@parcel/transformer-react-refresh-wrap' => undefined
}
```

## 🚨 Test instructions

Checkout the branch and add `console.log(config.devDeps)` to line 87 in `ConfigLoader.js`.
 https://github.com/thebriando/parcel/blob/10f4b4213ce0eb9234096631f64ec0a725ce73a0/packages/core/core/src/ConfigLoader.js#L87

Run `yarn parcel` on a project and ensure that there are version numbers for each devDep.

```
console: Map {
  '@parcel/transformer-react-refresh-babel' => '2.0.0-beta.1',
  '@parcel/transformer-babel' => '2.0.0-beta.1',
  '@parcel/transformer-js' => '2.0.0-beta.1',
  '@parcel/transformer-react-refresh-wrap' => '2.0.0-beta.1'
}
```

## ✔️ PR Todo

@padmaia @wbinnssmith


packages/core/core/src/ParcelConfig.js
```diff
@@ -332,6 +422,47 @@ export default class ParcelConfig {
...
+    Array<{|
+      name: string,
+      resolved: FilePath,
+      resolveFrom: FilePath,
+      version: Semver,
+    |}>,
+  > {
+    return this.resolvePlugins(this._getResolverNodes());
+  }
+
```
padmaia, line 465: Could you add `resolveBundler`, `resolveRuntimes`, `resolvePackager`, and `resolveOptimizers` as well? 🙏

packages/core/core/src/Transformation.js
```diff
@@ -762,7 +762,6 @@ function getImpactfulConfigInfo(configs: ConfigMap) {
     for (let [moduleName, version] of devDeps) {
       devDepsObject[moduleName] = version;
     }
-
```
padmaia, line 765: (nitpick) unnecessary change

padmaia (changes requested):

packages/core/core/src/ParcelConfig.js
thebriando: For these, should they also be called in `ConfigLoader.js`? Are there `actionTypes` for bundler, runtimes, packager, optimizers? 

https://github.com/parcel-bundler/parcel/blob/676499e3565d531e86178d978f89375d86fddf28/packages/core/core/src/ConfigLoader.js#L61-L75

packages/core/core/src/ParcelConfig.js
```diff
@@ -116,23 +122,107 @@ export default class ParcelConfig {
     };
   }
 
-  loadPlugin<T>(
+  async resolvePlugin(
+    node: ParcelPluginNode,
+  ): Promise<{|resolved: FilePath, pkg?: ?PackageJSON|}> {
+    let resolved, pkg;
+    let autoinstall = this.autoinstall;
+    try {
```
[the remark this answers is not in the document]
thebriando: Should we also store `pkg` in the cache then? Since resolve returns `pkg`

Issue #5378, opened 2020-11-22
Possible Bug: Parcel is commenting out my JS


john01dav:
# 🐛 bug report

This is related to [this](https://github.com/parcel-bundler/parcel/discussions/5377), so a misunderstanding on my part is possible, but it appears to be a bug since there is an HTML file that is valid by itself but fails when compiled through Parcel. When I have the following HTML file as source:
```html
<html>
<head>
<!-- The same thing happens if the script is referenced via <script src="… -->
<script>
function myFunction(){
  alert("An alert.");
}
//window.myFunction = myFunction; //If this is absent, then an error takes place. If this is present, then no error takes place
</script>
</head>
<body>
<button onclick="myFunction()">Show an alert</button> <!-- Error: myfunction() doesn't exist outside the module -->
</body>
</html>
```
I get errors in the JS console and, more importantly, no alert on the screen, when I click on the button. The fact that the commented-out `window.myFunction` line is able to fix the issue suggests that the script is functioning as a module, despite not being marked as such in the source code and nothing in the documentation specifying how, exactly, that works. Furthermore, if it were a module, then imports of other modules in my code would work from that script tag, and they do not (see the linked discussion for more detail);

## 🎛 Configuration (.babelrc, package.json, cli command)
An MCVE is attached which shows all configuration. The parcel version is specified in package.json and pacakge-lock.json, so build it like this:
```
npm install
npm run build # this will, I think, use the exact version of parcel that I'm using
```

## 🤔 Expected Behavior
I expect an alert to appear

## 😯 Current Behavior
An error appears in the JS console and there is no alert.

## 💁 Possible Solution
The only workaround I know of is the `window.…` line in my example.

## 🔦 Context
Without being able to import modules or access code in modules from `onclick` handlers, I'm forced to put every shared function in the `window` object, which is obviously problematic for code organization purposes.

## 💻 Code Sample
See attached

## 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Parcel           |See package-lock.json and package.json, it is the latest version as of a few minutes beforet his issue report. I'm not quite sure how to read package-lock.json, so I regretfully can't be of more help on this point.
| Node             |15.2.1
| npm/Yarn         |npm 6.14.8
| Operating System | Arch Linux, updated earlier today

[Download the MCVE here](https://github.com/parcel-bundler/parcel/files/5579135/parcel.zip)

john01dav:
[...]

PR #5398, opened 2020-11-29
Distinguish between module and classic scripts

devongovett:
Whew, this turned out way bigger than I expected... 😂 

This changes the way script tags are treated in HTML to distinguish between classic scripts and ES modules, improves Parcel's web compatibility, and introduces automatic modern-by-default optimizations for production builds.

* `import`, `export`, `require`, and all Node polyfill features are now only allowed in `<script type="module">` contexts (both inline and linked). Classic `<script>` elements can still have dynamic `import()`, `Worker`, `SharedWorker`, and `navigator.serviceWorker.register` dependencies. This matches what browsers support more closely, and allows better interoperability with legacy scripts where e.g. `require` being processed by Parcel was unexpected.
* Classic `<script>` now works just like the browser - global variables are truly global. The bundle is not wrapped in a function, so top-level variables are accessible between scripts on the page, in HTML event handler attributes, etc. See e.g. #5378. By default, when there are no dependencies, the script is not processed at all by Parcel. However, when an async dependency is seen that may require us to inject a runtime, we do wrap the bundle in a function to avoid unintended globals, but the globals seen in the original source are hoisted for compatibility.
* In addition, `<script type="module">` is now automatically compiled to both a modern ES module version and also a legacy `<script nomodule>` version if the browser targets do not all support ES modules natively. If no browserslist or target config is specified, then only a modern ES module version is output. For inline scripts, only a single script is output in order to avoid duplicating content - a modern version if all targets support it, otherwise legacy.
* The same rules also apply in other scenarios, including web workers and service workers. These now require the `type: 'module'` option just like in the browser to use ES modules, otherwise they will behave just like as described above for classic scripts. Workers are compiled to ES modules natively if all browser targets support them, otherwise they are compiled to a classic script instead.

When an import or export is found in a classic script, we show an error message that indicates both where the dependency is, as well as where the environment was originally instantiated.

image

This also resulted in adding support for multiple contexts at once in an environment. For example, an asset may have both "browser" and "script" as its environment. This may have other uses later on. I also added a method to the environment object to allow easier checking if it supports a particular feature. There's basic data for things like esmodules, dynamic imports, etc. so far, but ideally we'd get this data from somewhere else that we don't need to maintain ourselves. Finally, environments also now support a `loc` property to track the source location where they were originally created, mostly for error messaging purposes.

One other change due to the multi-environment support is that I changed the config format to no longer configure runtimes per context, and instead just have a single pipeline. The reality was that all of the contexts used the same runtime plugin anyway, and if the runtimes are context specific, they can check internally themselves.

Close T-554, Close T-136

### DIFF
--- a/packages/core/core/src/BundleGraph.js
+++ b/packages/core/core/src/BundleGraph.js
@@ -23,7 +23,12 @@ import assert from 'assert';
 import invariant from 'assert';
 import crypto from 'crypto';
 import nullthrows from 'nullthrows';
-import {flatMap, objectSortedEntriesDeep, unique} from '@parcel/utils';
+import {
+  flatMap,
+  objectSortedEntriesDeep,
+  unique,
+  isSubset,
+} from '@parcel/utils';
 
 import {getBundleGroupId, getPublicId} from './utils';
 import Graph, {ALL_EDGE_TYPES, mapVisitor, type GraphOpts} from './Graph';
@@ -615,7 +620,7 @@ export default class BundleGraph {
       visitedBundles.add(descendant);
       if (
         descendant.type !== bundle.type ||
-        descendant.env.context !== bundle.env.context
+        !isSubset(descendant.env.context, bundle.env.context)
       ) {
         actions.skipChildren();
         return;
@@ -630,7 +635,7 @@ export default class BundleGraph {
       let similarSiblings = this.getSiblingBundles(descendant).filter(
         sibling =>
           sibling.type === bundle.type &&
-          sibling.env.context === bundle.env.context,
+          isSubset(sibling.env.context, bundle.env.context),
       );
       if (
         similarSiblings.some(
@@ -713,7 +718,7 @@ export default class BundleGraph {
             // Don't deduplicate when context changes
             if (
               node.type === 'bundle' &&
-              node.value.env.context !== bundle.env.context
+              !isSubset(node.value.env.context, bundle.env.context)
             ) {
               actions.skipChildren();
             }
@@ -766,7 +771,7 @@ export default class BundleGraph {
             // Stop when context changes
             if (
               node.type === 'bundle' &&
-              node.value.env.context !== bundle.env.context
+              !isSubset(node.value.env.context, bundle.env.context)
             ) {
               actions.skipChildren();
             }
--- a/packages/core/core/src/Environment.js
+++ b/packages/core/core/src/Environment.js
@@ -17,6 +17,7 @@ export function createEnvironment({
   isLibrary = false,
   scopeHoist = false,
   sourceMap,
+  loc,
 }: EnvironmentOpts = {}): Environment {
   if (context == null) {
     if (engines?.node) {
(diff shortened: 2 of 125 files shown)


packages/utils/posthtml-parser/index.js
```diff
@@ -0,0 +1,198 @@
+const {Parser} = require('htmlparser2');
+
```
devongovett, line 2: I will be sending a PR to posthtml for this

packages/shared/babel-ast-utils/src/index.js
```diff
@@ -36,6 +36,8 @@ export async function parse({
         strictMode: false,
         sourceType: 'module',
         plugins: ['exportDefaultFrom', 'exportNamespaceFrom', 'dynamicImport'],
+        // $FlowFixMe
+        startLine: asset.meta.loc?.start.line,
```
devongovett, line 40: This offsets location information for both error messages and source maps of inline JS assets inside HTML so it starts at the correct line number of the original file

packages/transformers/html/src/dependencies.js
```diff
@@ -154,6 +146,70 @@ export default function collectDependencies(asset: MutableAsset, ast: AST) {
...
+            context,
+            outputFormat: 'global',
+            loc,
+          },
+        });
+
+        seen.add(copy);
+      }
+
+      attrs.src = asset.addURLDependency(attrs.src, {
```
devongovett, line 197: A thought: both of these dependencies count against the parallel request limit, but only one will actually be downloaded. How should we represent that in the graph? 🤔

mischnic:
I think this should resolve https://github.com/parcel-bundler/parcel/issues/333?

packages/core/core/src/Environment.js
```diff
@@ -28,65 +29,71 @@ export function createEnvironment({
...
+        case 'electron-renderer':
+          includeNodeModules = false;
+          break;
+        case 'browser':
+        case 'web-worker':
+        case 'service-worker':
+        default:
+          includeNodeModules = true;
+          break;
+      }
```
mischnic, line 67: So it depends on the order of the contexts? Maybe actually throw an error if there are conflicting values?

packages/core/test-utils/src/utils.js
```diff
@@ -207,46 +208,83 @@ type RunOpts = {require?: boolean, ...};
...
-      throw new Error('Unknown target ' + target);
   }
 
+  if (!ctx) {
+    throw new Error(
+      'Unknown context ' + [...entryAsset.env.context].join(', '),
+    );
+  }
+
+  let prepareCode = (code, b) => {
```
mischnic, line 244: ~Should we do https://github.com/parcel-bundler/parcel/pull/5548 instead?~

`prepareCode` shouldn't be necessary anymore

packages/core/core/src/BundleGraph.js
```diff
@@ -615,7 +620,7 @@ export default class BundleGraph {
       visitedBundles.add(descendant);
       if (
         descendant.type !== bundle.type ||
-        descendant.env.context !== bundle.env.context
+        !isSubset(descendant.env.context, bundle.env.context)
```
mischnic, line 623: `isSubset()` checks `a subset of b OR b subset of a`. But doesn't the previous "context changes" translate to "child context contains something that the parent doesn't support"? So only `child not subset of parent`?

(so context gets more specific/refined vs is widened)

mischnic:
Are JS entry points assigned the `module` context?

devongovett:
> Are JS entry points assigned the module context?

At the moment they have neither script or module. Actually, `module` is pretty much unused, we only check for `script`.

mischnic:
In the longer term, this could enable us to do something like Webpack 5's strict module mode (e.g. because of `type: "modlue"` in package.json) which disables some deprecated features.

devongovett:
Closing in favor of #6247

Issue #943, opened 2018-03-04
Hot reload doesn't work for an empty HTML page without external CSS/JS imports


RIP21:
### 🐛 bug report

No files or configs

### 🤔 Expected Behavior

I have empty HTML without any external resources (such as local CSS files or JS). When I change it and save I should get an immediate update in the browser without refresh while in dev mode.

### 😯 Current Behavior

I change anything in the HTML file, it gets rebuilt but not refreshed automatically in the browser. I need to click refresh.

### 💁 Possible Solution

When I add any import. For example <link> to the CSS external local file and then refresh the page once, all changes to HTML start to immediately displayed on saving without a need to press refresh

### 💻 Code Sample

```html
<!DOCTYPE html>
<html lang="en">
<head>
  <meta charset="UTF-8">
  <title>Title</title>
</head>
<body>
<p style="color: red">Hello world</p>
</body>
</html>
```

### 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Parcel           | latest to date
| Node             | 8.9.3
| Yarn         | 1.5
| Operating System | Mac OS High Sierra

mischnic:
There is no javascript bundle to append the HMR runtime to.
https://github.com/parcel-bundler/parcel/blob/420ed63ed18c6a09e8b25754d0142b3b87ebcd71/src/packagers/JSPackager.js#L185-L191

We would have to inject a script tag into the html file.

RIP21:
Glad that my bug contribution is not false positive :) Waiting for the fix ;)

RIP21:
@vsui I think you could. It's quite annoying.

cacheflow:
I'll take this on if it's still open @RIP21 @vsui.

akselinurmio:
This sure needs to be fixed. I’m not familiar with the internals of Parcel, but I guess that to support auto reloading without JS, we’d have to:

1. create a browser module including HMR runtime, and then
2. inject it to the HTML.

[Browsersync](https://github.com/Browsersync/browser-sync) injects its client script after the opening `<body>` tag, would that be wise to do here too?

MegaJSDeveloper:
## 💁 Possible Solution

if page has no one scripts, you must automatically add reload helper script, to the beginning of the head element, like:

```html
<script id="__parcel__helper__">
  /* code */
  /* code */
  /* code */

  // clear html from helper script tag, just in case
  var helperScriptElement = document.getElementById('__parcel__helper__')
[...]

PR #5414, opened 2020-12-03
Enable HMR without script tag in source html

faergeek:
# ↪️ Pull Request

This is a brute-force solution for https://github.com/parcel-bundler/parcel/issues/943. As explained, HMR code is injected into js bundles, and if there's no js bundle, HMR doesn't work. So in this PR I add empty script to if HMR is enabled. Right now it doesn't matter if there is one already (not sure if I need to add checks for that).

Closes https://github.com/parcel-bundler/parcel/issues/943

## ✔️ PR Todo


mischnic:
1. We should track whether there is already a script tag in the HTML file (`collectDependencies` could track that).
2. It shouldn't be an inline script tag, but rather this so that it becomes a separate bundle:
```js
ast.program.push({tag: 'script', attrs: {src: asset.addDependency("hmr-script"))}, content: []});
// ...
return [asset, {type: "js", uniqueKey: "hmr-script", content: ""}, ...];
```

faergeek:
Thanks for feedback!

1. Shouldn't then both `collectDependencies` and `extractInlineAssets` track if we have a script already? Because HMR code is injected into inline scripts too. And what's a good way to track that in your opinion? Pass an object with a boolean prop, which is mutated inside of those functions?

mischnic:
I would return a boolean in `collectDependencies` and return `{hasScript: true/false, assets: ...}` in `extractInlineAssets`

faergeek:
Now `hasScripts` flags are returned from `collectDependencies` and `extractInlineAssets` to ensure we need to add script tag for HMR to work. I also pushed an example to demonstrate invariant violation. If you run `yarn start` for it, the script is injected and everything is fine. But whenever you uncomment script tag in html, this invariant in `AssetGraph` is violated for inserted (and now being removed) node:
 
https://github.com/parcel-bundler/parcel/blob/e44a16888b8177af6d2863139a0096a0d4c5c72a/packages/core/core/src/AssetGraph.js#L162

I'm not sure if that's because I did something wrong. I'm new to parcel codebase. Looking for help

mischnic:
> this invariant in AssetGraph is violated for inserted

Your code is correct. I've run into this even without your PR, but haven't figured it out yet.

faergeek:
So the example is not needed anymore? Should I remove it? Please let me know if I need to do something more

mischnic:
Yes, remove that example. Instead add an integration test similar to this one: https://github.com/parcel-bundler/parcel/blob/90a3f012ccada0e5f80d9aac0dda3e08f21053d1/packages/core/integration-tests/test/html.js#L79-L94

In that assert call, the JS bundle should contain the `hmr-runtime.js` asset.

faergeek:
I removed an example and added integration test. There's also a react refresh runtime being injected. Is it correct? Also I'm not sure why pr benchmarks failed.

mischnic:
Add these two files:
```
packages/core/integration-tests/test/integration/html-no-js/yarn.lock
packages/core/integration-tests/test/integration/html-no-js/package.json
```
(yarn.lock is empty, package.json contains `{}`)

faergeek:
I added `package.json` and `yarn.lock`, and fixed test.

faergeek:
Is there anything else to be done with this PR?

mischnic:
No, but this apparently exaggerates the assertion error from https://github.com/parcel-bundler/parcel/pull/5414#issuecomment-742764000. So I'd like to get that fixed first before merging your PR.

faergeek:
@mischnic is there any issue to track this? I noticed that this problem only happens when assets returned from transform get removed. So the error happens if you remove inline script, but when you remove script with `src`, it doesn't. The same thing with `<link rel="stylesheet" href="...">` vs `<style>...</style>`. And when removing inline styles the error happens as well. So now I'm not sure why this invariant is needed there.

Also I noticed that `<style>...</style>` tags and inline styles don't get hot-reloaded. I'm seeing an error `Uncaught SyntaxError: Unexpected token '<'` in browser devtools.

mischnic:
PR to fix the assertion: https://github.com/parcel-bundler/parcel/pull/5522

I've opened a new issue with some findings about HMR update for inline assets: https://github.com/parcel-bundler/parcel/issues/5523

packages/transformers/html/src/HTMLTransformer.js
```diff
@@ -23,12 +23,45 @@ export default (new Transformer({
...
+    // when there's no html/head/body in source html.
+    if (options.hot && !(hasScripts || hasInlineScripts)) {
+      ast.program.push({
+        tag: 'script',
+        attrs: {
+          src: asset.addURLDependency('hmr.js', {
+            isAsync: false,
+            isEntry: false,
+            isIsolated: true,
+          }),
```
mischnic, line 50:
```suggestion
          src: asset.addURLDependency('hmr.js'),
```

packages/core/integration-tests/test/html.js
```diff
@@ -93,6 +93,28 @@ describe('html', function() {
     ]);
   });
 
+  it('should insert empty script tag for HMR', async function() {
+    let b = await bundle(
+      path.join(__dirname, '/integration/html-no-js/index.html'),
+      {
+        hot: {
+          port: 42,
+        },
```
mischnic, line 102:
```suggestion
        hot: {},
```

mischnic (approved):

DeMoorJasper (approved):

devongovett (approved):

mischnic:
Thank you!

aarongeorge:
Hey everyone, thanks for the work on this.

There is an edge case that seems to be missing from this fix. If the file is a `.pug` (or anything other than `.html`) this HMR is still not injected.

I'm thinking we should be checking the entry file for a script tag, instead of the `.html` file for it (maybe we do both?)

Thoughts?

mischnic:
It should also work for Pug because the HTML that was generated from the Pug file get analyzed for these script tags.

faergeek:
Hey @aarongeorge 

I believe this is a better place to fix pug/jade https://github.com/parcel-bundler/parcel/issues/5523

faergeek:
@mischnic I'm not entirely sure, but I remember there were some checks for file extension. So I assume something may not work for pug now

Issue #5387, opened 2020-11-24
An array `package.json#source` should build these specified entries.


mischnic:
Running `parcel build my-package/` should build `a.html, b.html` if `my-package/package.json` contains:
```json
{
  "source": ["a.html", "b.html"]
}
```

So be equivalent to `cd my-package && parcel build a.html b.html`.

The docs already say this should work: https://v2.parceljs.org/configuration/package-json/#source

Relevant code:
https://github.com/parcel-bundler/parcel/blob/064ba0a166ea63ccd2b623dc11324d4cd7ffb526/packages/core/core/src/requests/EntryRequest.js#L98-L126

alphabet-codes:
@mischnic I'd love to pick this issue up.

alphabet-codes:
Thanks, @mischnic

alphabet-codes:
@mischnic how do you run build if there's a source file `{"source": "src/index.js"}` in your package.json for instance?

mischnic:
Pass the directory which contains the package.json as an argument:

> Running `parcel build my-package/` should build `a.html, b.html` if `my-package/package.json` contains:

alphabet-codes:
@mischnic I keep getting an error that says "no entries found" when I run the above command why?

mischnic:
Example: https://github.com/mischnic/parcel-issue-5387

alphabet-codes:
@mischnic, please help me review #5451. some checks are not passing

alphabet-codes:
@mischnic, please review this pull request, so that I can pick up a new issue and work on it.

mischnic:
You can also work on something else even before you PR is merged, if you want.

PR #5494, opened 2020-12-17
An array `package.json#source` should build specified entries

AceTheCreator:
# ↪️ Pull Request
Running parcel build my-package/ should build a.html, b.html if my-package/package.json contains: 
`{
  "source": ["a.html", "b.html"]
}`
#5387


mischnic:
Please add an integration test

packages/core/core/src/requests/EntryRequest.js
```diff
@@ -122,6 +122,40 @@ class EntryResolver {
           entries: [{filePath: source, packagePath: entry}],
           files: [{filePath: pkg.filePath}],
         };
+      } else if (pkg && Array.isArray(pkg.source)) {
+        let source;
+        const entries = [];
```
wbinnssmith, line 127: nit: `let entries = [];`, even though it's not reassigned this is a style thing in Parcel (though I personally also would prefer const)

wbinnssmith (commented):
Looks good! As @mischnic mentioned, this should have an integration test.

packages/core/core/src/requests/EntryRequest.js
alphabet-codes: Ok. Thanks for the review

alphabet-codes: @wbinnssmith I have submitted an updated version of this pull request, kindly review it thanks.

wbinnssmith: looks good! Mind writing an integration test that exercises this code? Should be good to go from there 😄

alphabet-codes: @wbinnssmith I'll do that later today.

mischnic:
@AceTheCreator Would be great if you could add some tests so that we can merge this!

AceTheCreator:
> @AceTheCreator Would be great if you could add some tests so that we can merge this!

I'll do that later today @mischnic

AceTheCreator:
> > @AceTheCreator Would be great if you could add some tests so that we can merge this!
> 
> I'll do that later today @mischnic



> @AceTheCreator Would be great if you could add some tests so that we can merge this!

AceTheCreator:
@mischnic I'm having an issue writing test for this. Any idea on how I can do it?

mischnic:
Add a new folder to `packages/core/integration-tests/test/integration/` containing multiple source files and a package.json with a `source` field. Similar to `packages/core/integration-tests/test/integration/html-pkg-source` but with an array instead.

And then run that test in `packages/core/integration-tests/test/html.js`.

PR #5511, opened 2020-12-21
Flow typings for htmlnano

AGawrys:
# ↪️ Pull Request

This PR is for flow typing for htmlnano. The file is sufficient for our usage, although there are many more configuration options. Was unsure since I couldn't find a d.ts file.  

## 💻 Source

Source: https://github.com/posthtml/htmlnano 

## 🚨 Test instructions

I ran `yarn flow check` with 0 errors found after removing `// flowlint-next-line untyped-import:off` .


flow-libs/htmlnano.js.flow
```diff
@@ -0,0 +1,11 @@
+// @flow
+
+declare module 'htmlnano' {
+  declare type Options = {|
+    minifyJs?: boolean,
+    minifyCss?: boolean,
+    userConfig?: mixed,
```
mischnic, line 7:
```suggestion
```

`userConfig` is definitely not an option

AGawrys: Ah my mistake, we were passing in a config of type any. I made it inexact now and removed it...

mischnic (approved):

PR #5516, opened 2020-12-22
Devserver port search fix

### LINKED ISSUE #5495: Devserver should not find a new port if a specific one is requested
# 🐛 bug report

Parcel's devserver finds another suitable port if one is not available. This makes sense in the default case when a specific port number is not requested.

To reproduce, run

`parcel serve src/index.html --port 3000`

and then run it again: `parcel serve src/index.html --port 3000`

## 🤔 Expected Behavior

We should consider exiting the process with a non-zero status code. This "fail fast" signal is useful if the Parcel devserver is being expected to be listening on a specific port, rather than going ahead with starting Parcel. This might occur if the devserver is behind a reverse proxy, running in a container with a specific published port, or if running the devserver is automated, such as if it's run by a supervisor or init process.

## 😯 Current Behavior

The devserver finds another suitable port and listens on it.

## 💁 Possible Solution

Exit the process with a non-zero code and do not bind on another port.

golota60:
# ↪️ Pull Request

Added a change after which devserver is not searching for a new port if a specific one was requested

Closes #5495

## 💻 Examples

If a user has explicitly wanted to serve on a specific port, then if that port is taken it won't search for a new one.

## ✔️ PR Todo

### DIFF
--- a/packages/core/parcel/src/cli.js
+++ b/packages/core/parcel/src/cli.js
@@ -349,8 +349,9 @@ async function normalizeOptions(command): Promise<InitialParcelOptions> {
         chalk.bold.yellowBright(`⚠️  Port ${command.port} could not be used.`),
       );
       // If port was provided using -p or --port flag then throw an Error
-      if (isCustomPort && nodeEnv === 'development')
+      if (isCustomPort && nodeEnv === 'development') {
         throw new Error(`port ${command.port} taken`);
+      }
     }
   }
 
--- a/packages/core/parcel/src/cli.js
+++ b/packages/core/parcel/src/cli.js
@@ -183,7 +183,7 @@ async function run(entries: Array<string>, command: any) {
   try {
     options = await normalizeOptions(command);
   } catch (e) {
-    return process.exit(-1);
+    return process.exit(1);
   }
   let fs = new NodeFS();
   let packageManager = new NodePackageManager(fs);
@@ -349,8 +349,8 @@ async function normalizeOptions(command): Promise<InitialParcelOptions> {
         chalk.bold.yellowBright(`⚠️  Port ${command.port} could not be used.`),
       );
       // If port was provided using -p or --port flag then throw an Error
-      if (isCustomPort && nodeEnv === 'development') {
-        throw new Error(`port ${command.port} taken`);
+      if (isCustomPort) {
+        throw new Error(`Port ${command.port} could not be used`);
       }
     }
   }
--- a/packages/core/parcel/src/cli.js
+++ b/packages/core/parcel/src/cli.js
@@ -349,7 +349,8 @@ async function normalizeOptions(command): Promise<InitialParcelOptions> {
         chalk.bold.yellowBright(`⚠️  Port ${command.port} could not be used.`),
       );
       // If port was provided using -p or --port flag then throw an Error
-      if (isCustomPort) throw new Error(`port ${command.port} taken`);
+      if (isCustomPort && nodeEnv === 'development')
+        throw new Error(`port ${command.port} taken`);
     }
   }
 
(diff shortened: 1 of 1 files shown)


packages/core/parcel/src/cli.js
```diff
@@ -179,7 +179,12 @@ async function run(entries: Array<string>, command: any) {
     return;
   }
   let Parcel = require('@parcel/core').default;
-  let options = await normalizeOptions(command);
+  let options;
+  try {
+    options = await normalizeOptions(command);
+  } catch (e) {
+    return process.exit(-1);
```
DeMoorJasper, line 186: Shouldn't this be `process.exit(1)` ?

packages/core/parcel/src/cli.js
```diff
@@ -340,6 +348,10 @@ async function normalizeOptions(command): Promise<InitialParcelOptions> {
       INTERNAL_ORIGINAL_CONSOLE.warn(
         chalk.bold.yellowBright(`⚠️  Port ${command.port} could not be used.`),
       );
+      // If port was provided using -p or --port flag then throw an Error
+      if (isCustomPort && nodeEnv === 'development') {
```
DeMoorJasper, line 352: Is the development check neccessary?

packages/core/parcel/src/cli.js
golota60: I changed it cause otherwise we would exit with status code 0 - if there's a better way to handle that, please let me know! 😄

packages/core/parcel/src/cli.js
golota60: I checked and I think it wasn't - thanks

packages/core/parcel/src/cli.js
DeMoorJasper: Yeah, I believe more than just that line can fail, opened a seperate PR for it, https://github.com/parcel-bundler/parcel/pull/5517

DeMoorJasper (approved):

mischnic:
If we do something like this, passing a port via the `PORT` env variable is also handled:

```diff
diff --git packages/core/parcel/src/cli.js packages/core/parcel/src/cli.js
index ca7cf45bb..812327837 100755
--- packages/core/parcel/src/cli.js
+++ packages/core/parcel/src/cli.js
@@ -90,7 +90,7 @@ var hmrOptions = {
   '-p, --port <port>': [
     'set the port to serve on. defaults to $PORT or 1234',
     value => parseInt(value, 10),
-    parseInt(process.env.PORT, 10) || 1234,
+    process.env.PORT != null ? parseInt(process.env.PORT, 10) : undefined,
   ],
   '--host <host>':
     'set the host to listen on, defaults to listening on all interfaces',
@@ -331,15 +331,24 @@ async function normalizeOptions(command): Promise<InitialParcelOptions> {
   }
 
   let serve = false;
-  let {port, host} = command;
+  let {port = 1234, host} = command;
+  let originalPort = port;
+  console.log(originalPort)
   if (command.name() === 'serve' || command.hmr) {
     port = await getPort({port, host});
 
-    if (command.port && port !== command.port) {
-      // Parcel logger is not set up at this point, so just use native INTERNAL_ORIGINAL_CONSOLE.
-      INTERNAL_ORIGINAL_CONSOLE.warn(
-        chalk.bold.yellowBright(`⚠️  Port ${command.port} could not be used.`),
-      );
+    if (originalPort && port !== originalPort) {
+      if (command.port != null) {
+        // If port was provided using -p or --port flag or via the PORT env variable then throw an Error
+        throw new Error(`Port ${originalPort} could not be used`);
+      } else {
+        // Parcel logger is not set up at this point, so just use native INTERNAL_ORIGINAL_CONSOLE.
+        INTERNAL_ORIGINAL_CONSOLE.warn(
+          chalk.bold.yellowBright(
+            `⚠️  Port ${originalPort} could not be used.`,
+          ),
+        );
+      }
     }
   }
 
```

It would also make sense to only log a warnings if no error is thrown.

PR #5562, opened 2021-01-02
Fix native package not found issue

DeMoorJasper:
# ↪️ Pull Request

This PR includes the native packages in lerna and yarn workspaces to prevent the package not found error.

This fails the CI/CD environment because there's a custom script that seems to do basically the same? I'm kind of confused why this is necessary? cc @devongovett

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,4 +1,4 @@
 {
-  "packages": ["packages/*/*"],
+  "packages": ["packages/*/*", "native-packages/*"],
   "version": "independent"
 }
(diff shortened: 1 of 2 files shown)


mischnic:
I think the idea was to use the latest published version from npm and have an opt-in for the locally built version.
But there was no nightly release after the merge yet...

devongovett:
Right. You can run `yarn link-native` to symlink the native packages into the monorepo. I didn't want to force all contributors to install a rust toolchain even if they weren't contributing to that package...

DeMoorJasper:
@devongovett Alright, makes total sense, thanks for the explanation. We'll just have to wait for a nightly before this actually works :)

PR #5565, opened 2021-01-02
Remove Parcel v1 from monorepo

DeMoorJasper:
# ↪️ Pull Request

This PR removes Parcel v1 from the `v2` branch monorepo, as it has caused confusion with some people wanting to patch bugs in the past and doesn't really add any value anymore in the v2 monorepo as we pretty much have feature parity with v1 at this point?


devongovett (approved):
Probably a good idea...

PR #5568, opened 2021-01-03
Add pnpm autoinstall support

aminya:
↪️ Pull Request

Adds pnpm support

Originally posted by @dishuostec at https://github.com/parcel-bundler/parcel/pull/4920

That PR was not merged every, so I thought it might be still needed.

 
# TODO

The changes in `packages/core/parcel-bundler/src/utils/installPackage.js` should be moved to `packages/core/package-manager/src/installPackage.js`

I don't have access to your repo, so you should do this.


ssalbdivad:
Would really love to see this merged ASAP so I can start using Parcel :bowtie:

aminya:
@ssalbdivad You can add this `.npmrc` file that hoists the folder making pnpm folder structure like npm. This works, however, Parcel uses a JavaScript implementation to follow symlinks (instead of `fs.realPath.native`) which may increase the build time. This should be fixed.
```
public-hoist-pattern[]=*
```
https://github.com/aminya/solid-simple-table/blob/master/.npmrc

mischnic:
pnpm should generally work. The only thing this adds is using pnpm for autoinstalling missing packages.

mischnic:
-> https://github.com/parcel-bundler/parcel/pull/5951

PR #5571, opened 2021-01-03
Implement audit changes

DeMoorJasper:
# ↪️ Pull Request

This pull requests implements the options part of the API audit notes.

This is currently very much a work in progress

Related #5225


packages/core/core/src/AssetGraphBuilder.js
```diff
@@ -818,7 +818,7 @@ export default class AssetGraphBuilder extends EventEmitter {
   }
 
   async readFromCache(): Promise<?Array<Event>> {
-    if (this.options.disableCache) {
+    if (this.options.shouldDisableCache) {
```
mischnic, line 821: I thought we had talked about this in the call but maybe this should rather be `shouldEnableCache`.

DeMoorJasper: All changes have been taken from the linked PR by Devon, which are the changes we agreed on I think.

mischnic: Yeah, but at the same time I remember Will saying that double negations should be avoided 🤷‍♂️

DeMoorJasper:
Closing in favor of #5572 will do these changes in smaller incremental PRs this PR will take ages to complete and be a nightmare to keep in sync with v2

packages/core/core/src/AssetGraphBuilder.js
DeMoorJasper: Yeah it was commented out so probably not final yet

PR #5575, opened 2021-01-03
Support manually dispatching nightly releases

devongovett:
Not sure if this syntax is correct... YAML is a mystery.

### DIFF
--- a/.github/workflows/nightly-release.yml
+++ b/.github/workflows/nightly-release.yml
@@ -2,6 +2,7 @@ name: release-nightly
 on:
   schedule:
     - cron: '0 0 * * *'
+  workflow_dispatch: {}
 jobs:
   build:
     strategy:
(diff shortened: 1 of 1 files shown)


.github/workflows/nightly-release.yml
```diff
@@ -2,6 +2,7 @@ name: release-nightly
 on:
   schedule:
     - cron: '0 0 * * *'
+  workflow_dispatch: {}
```
mischnic, line 5:
```suggestion
  workflow_dispatch:
```
If I understand https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#example-using-multiple-events-with-activity-types-or-configuration correctly 🤷‍♂️

DeMoorJasper: I think the `{}` is optional but recommended in yaml?

mischnic: I recommend not using yaml at all 😂

PR #5576, opened 2021-01-03
Reduce linux binary size for fs-search package

devongovett:
Strips debug symbols. Due to https://github.com/rust-lang/rust/issues/46034


mischnic (commented):
Do you want to do that on macOS as well?

mischnic (approved):

PR #5589, opened 2021-01-05
Replace asset.setEnvironment with transformer.getEnvironment

devongovett:
asset.setEnvironment was problematic because it replaced the environment but did not affect the id of the asset which had already been computed. Instead, this PR replaces it with an optional transformer method which does the same thing but is run before the asset is even loaded, and again for each pipeline (if the asset changes types or new assets are added). This allows the environment in the cache key to be correct as well so hopefully we'll see more cache hits as opposed to unnecessary re-transformations.

Closes T-885

### DIFF
--- a/packages/core/core/src/Transformation.js
+++ b/packages/core/core/src/Transformation.js
@@ -16,6 +16,7 @@ import type {
   ConfigRequestDesc,
   ParcelOptions,
   ReportFn,
+  Environment as InternalEnvironment,
 } from './types';
 
 import invariant from 'assert';
@@ -50,8 +51,10 @@ import {
 } from './assetUtils';
 import summarizeRequest from './summarizeRequest';
 import PluginOptions from './public/PluginOptions';
+import Environment from './public/Environment';
 import {PARCEL_VERSION} from './constants';
 import {optionsProxy} from './utils';
+import {createEnvironment, mergeEnvironments} from './Environment';
 
 type GenerateFunc = (input: UncommittedAsset) => Promise<GenerateOutput>;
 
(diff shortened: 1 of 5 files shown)


packages/core/core/src/Transformation.js
```diff
@@ -492,36 +511,52 @@ export default class Transformation {
...
+      filePath,
+      pipeline,
+      this.request.isURL,
+    );
+
+    // Allow transformers in the pipeline to override the requested environment.
+    // For example, some transformers may not be affected by certain options
+    // (e.g. library mode, scope hoisting, or output formats).
+    let env = transformers.reduce((env, {plugin}) => {
+      if (typeof plugin.getEnvironment === 'function') {
```
mischnic, line 526: Nit: so far, something like `if(plugin.getEnvironment)` has been used in this file for checking optional methods

packages/core/core/src/Transformation.js
```diff
@@ -492,36 +511,52 @@ export default class Transformation {
...
+    let transformers = await this.parcelConfig.getTransformers(
+      filePath,
+      pipeline,
+      this.request.isURL,
+    );
+
+    // Allow transformers in the pipeline to override the requested environment.
+    // For example, some transformers may not be affected by certain options
+    // (e.g. library mode, scope hoisting, or output formats).
+    let env = transformers.reduce((env, {plugin}) => {
```
mischnic, line 525: Does this mean that e.g. with Less asset, the LessTransformer would see `asset.scopeHoist === true` because it's only cleared in the following `*.css` pipeline (or something like that for 3 chained pipelines)? I guess there's nothing we can do about this though...

mischnic: 2. In my mind, it should rather be "if every transformer doesn't care about `scopeHoist`, then it can be dropped". Instead of `if any transformer.....`
3. In this case
```json
{
	"*.a": ["parcel-transformer-a-to-b", "parcel-transformer-unused"],
	"*.b": ["parcel-transformer-b"]
}
```
`parcel-transformer-unused` never runs because the asset type changes before it's reached. But `getEnvironment` is still applied, which might be unexpected as well.

PR #5481, opened 2020-12-15
Implement fs searching in rust

devongovett:
This adds a new package, `@parcel/fs-search`, which implements some very common filesystem searching methods natively in Rust, with a fallback to a JS implementation. These include finding a file by searching up a directory tree, finding a node_modules directory, and checking a number of extensions. Using the native implementation instead of JS is much faster because it avoids crossing the JS -> C++ barrier for every FS call and instead only crosses it once per search.

```
rust 46104.62 opts/sec (mean: 0.022ms, stddev: 0.016ms, 2500 samples)
js 4051.36 opts/sec (mean: 0.247ms, stddev: 0.185ms, 2500 samples)
```

This results in ~15% faster builds overall.

TODO

- [x] Figure out how to build and publish native modules as part of the monorepo
- [x] Fix the circular dependency which breaks the tests due to import not being lazy (NodeFS -> utils -> fs-search -> NodeFS)

PR #5596, opened 2021-01-06
Windows 32 bit support

aminya:
This runs the CI on Windows 32 bit. 

Fixes https://github.com/parcel-bundler/parcel/issues/5597

[fs-search](https://github.com/parcel-bundler/parcel/pull/5481) broke Parcel on Windows 32bit. This PR is adding Win 32bit to CI to prevent this kind of breakage.

## 🚨 Test instructions

The CI itself runs the tests.

## ✔️ PR Todo

### DIFF
--- a/azure-pipelines-template.yml
+++ b/azure-pipelines-template.yml
@@ -17,6 +17,7 @@ jobs:
       - task: NodeTool@0
         inputs:
           versionSpec: $(node_version)
+          force32bit: ${{ eq(parameters.name, 'Windows-32bit') }}
         displayName: 'Install Node.js'
 
       # Install Rust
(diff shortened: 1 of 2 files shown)


mischnic:
Bildschirmfoto 2021-01-06 um 20 19 19

devongovett:
I don't think we really need to run the tests against win32. If we were to support it, we would only need to do so in the publishing actions (like we do for arm for example).

Regardless, it may be easier to simply fall back to the JS implementation on unsupported platforms.

aminya:
> I don't think we really need to run the tests against win32. If we were to support it, we would only need to do so in the publishing actions (like we do for arm for example).
> 
> Regardless, it may be easier to simply fall back to the JS implementation on unsupported platforms.

I could not find a place that I can specify the arch for the Rust toolchain. In C++, you can easily do this using `--arch=ia32`.

https://github.com/parcel-bundler/source-map/blob/794a36d529f421ca897f2dd8ca6f2cdad928a85b/package.json#L18

PR #5598, opened 2021-01-06
AssemblyScript transformer

aminya:
↪️ Pull Request

WIP. Traying to port the rollup plugin
https://github.com/surma/rollup-plugin-assemblyscript/blob/master/index.js

## ✔️ PR Todo


packages/transformers/assemblyscript/src/ASTransformer.js
```diff
@@ -0,0 +1,132 @@
...
+
+      // wasm file
+      asset.type = 'wasm'
+      asset.setCode(await fsp.readFile(wasmFilePath))
+      if (options.sourceMapURLPattern) {
+        asset.setMap( await fsp.readFile(sourceMapFilePath) )
+      }
+
+      // wasm wrapper
+      let wasmWrapper = new BaseAsset() // ?? no documentation or examples
```
mischnic: Instead do
```js
return [asset, {type: 'js', map: ..., code: ....}];
```
at the end.

aminya: Thanks. I changed it.

packages/transformers/assemblyscript/src/ASTransformer.js
```diff
@@ -0,0 +1,144 @@
...
+      asset.setMap(await fsp.readFile(sourceMapFilePath));
+    }
+
+    // wasm wrapper for browser (TODO support Nodejs)
+    const wasmWrapper = {
+      type: 'js',
+      // TODO port Rollup improts
+      code: `
+          import {compileStreaming} from "${SPECIAL_IMPORT}";
+          const wasmUrl = import.meta.ROLLUP_FILE_URL_${asset.id}
```
aminya: Not sure how to port this

packages/transformers/assemblyscript/src/ASTransformer.js
```diff
@@ -0,0 +1,144 @@
...
+const compileStreamingCode = `
+  export async function compileStreaming(respP) {
+              if('compileStreaming' in WebAssembly) {
+                return WebAssembly.compileStreaming(respP);
+              }
+              return respP
+                .then(resp => resp.arrayBuffer())
+                .then(buffer => WebAssembly.compile(buffer));
+            }
+        `;
```
aminya: Not sure how I can use these in Parcel
https://github.com/surma/rollup-plugin-assemblyscript/blob/da03273c4da0979e84d432cd0827a69e2443ecff/index.js#L86-L102

packages/transformers/assemblyscript/src/ASTransformer.js
```diff
@@ -0,0 +1,144 @@
...
+
+    await new Promise((resolve, reject) => {
+      const errorCollector = streamCollector();
+      const params = [
+        asset.filePath,
+        '-b',
+        wasmFilePath,
+        ...(options.sourceMapURLPattern
+          ? [
+              `--sourceMap=${renderNamePattern(options.sourceMapURLPattern, {
```
aminya: Not sure how sourceMaps should be handled (renderNamePatter is a Rollup specific thing)

packages/transformers/assemblyscript/src/ASTransformer.js
```diff
@@ -0,0 +1,144 @@
...
+              }
+              return respP
+                .then(resp => resp.arrayBuffer())
+                .then(buffer => WebAssembly.compile(buffer));
+            }
+        `;
+
+export default (new Transformer({
+  async transform({asset, options}) {
+    const asCompiler = await getASCompilerPath(asset.filePath, options);
```
mischnic: Looks this like this was taken from the Elm transformer? This workaround was needed there because the `elm` package has no JS entry and so `require("elm")` always fails.

I think you can instead simply do 
```js
let asCompiler  = await options.packageManager.require('assemblyscript/cli/asc', asset.filePath, {
	shouldAutoInstall: options.shouldAutoInstall,
}),
```
here (or whatever subpath needs to be imported)

packages/transformers/assemblyscript/src/ASTransformer.js
mischnic: Conceptually, this should be the result of `asset.addDependency()` which gets replaced with the correct url.

packages/transformers/assemblyscript/src/ASTransformer.js
```diff
@@ -0,0 +1,144 @@
...
+          const msg = new TextDecoder().decode(stderr);
+          return reject(msg);
+        }
+        resolve();
+      });
+    });
+
+    // wasm file
+    asset.type = 'wasm';
+    asset.setCode(await fsp.readFile(wasmFilePath));
```
mischnic: Are there downsides to using this variant of the compiler instead? Would be cleaner than writing & reading from a tmpdir and would also work in the browser (from https://github.com/AssemblyScript/assemblyscript/tree/master/cli#api).

Bildschirmfoto 2021-01-09 um 16 20 52

packages/transformers/assemblyscript/src/ASTransformer.js
mischnic: This might work:
```js
let wasmCode = asc.compile(/*...*/);

asset.type = "js";
asset.setCode(`
import wasmURL from "my-wasm";
export const modulePromise = /*@__PURE__*/(() => compileStreaming(fetch(wasmUrl)))();
export const instancePromise = /*@__PURE__*/(() => modulePromise.then(module => WebAssembly.instantiate(module, {})))();
export default wasmUrl;
`)
asset.addURLDependency('my-wasm');

return [asset, {type: 'wasm', uniqueKey: 'my-wasm', code: wasmCode}];
```

packages/transformers/assemblyscript/src/ASTransformer.js
mischnic: To share code between transformer outputs, you can put this into a separate file inside the transformer package like here:

https://github.com/parcel-bundler/parcel/blob/e183e32c811f066310fff0610aa856ca73a2fcb6/packages/transformers/js/src/visitors/modules.js#L281-L296 which includes
https://github.com/parcel-bundler/parcel/blob/e183e32c811f066310fff0610aa856ca73a2fcb6/packages/transformers/js/src/esmodule-helpers.js

packages/transformers/assemblyscript/package.json
```diff
@@ -0,0 +1,28 @@
...
+    "url": "https://github.com/parcel-bundler/parcel.git"
+  },
+  "main": "lib/ASTransformer.js",
+  "source": "src/ASTransformer.js",
+  "engines": {
+    "node": ">= 12.0.0",
+    "parcel": "^2.0.0-alpha.1.1"
+  },
+  "peerDependencies": {
+    "assemblyscript": "*"
```
mischnic: ~This shouldn't be a peer dependency. The package will be imported relative to the users's project when using `packageManger.require`~ In the meantime, we have changed this approach

101arrowz:
I'm a huge fan of AS, and this will definitely make using it more painless. However, I have a few concerns with this PR so far:
- It's not necessary to use the CLI version over the string compiler, so I'd think it's better to use the string version
- I think we should handle exports the same as with Rust - make it seamless without having `modulePromise` and `instancePromise`.
- We could use a config, e.g. for [transformations](https://www.assemblyscript.org/transforms.html).
- We might want to support stuff like `process.env` and `process.browser` and all the other runtime globals. This can be done with a Parcel internal transformation for AS.

Overall though, this looks very promising!

mischnic:
> I think we should handle exports the same as with Rust - make it seamless without having modulePromise and instancePromise.

Is there a case where you need to pass an importObject with memory or imported functions to an AS wasm module?

101arrowz:
Apparently so, AssemblyScript/assemblyscript#158...

mischnic:
Also found it here: https://www.assemblyscript.org/exports-and-imports.html#imports

So I guess we have to provide some why to pass them in.

This is also what I dislike about the proposal for importing Wasm with ES imports. You only get the exports but there is no way to provide imports. So there's no standardized way....

101arrowz:
The data to be imported from WASM could theoretically be provided via a config file, but I don't like the idea myself. At the same time, having different ways to import Rust WASM and AS WASM is just confusing and I'd rather sacrifice the ability to import from WASM than have to use both. The JS-WASM bridge is so slow that most WASM users probably don't need JS imports. A common strategy is to split logic between WASM and JS to avoid two passes across the bridge (the call and returning the value).

aminya:
The upstream package is updated:
https://github.com/surma/rollup-plugin-assemblyscript/blob/main/index.js

packages/transformers/assemblyscript/src/ASTransformer.js
aminya: Fixed

aminya:
I tried to understand the Parcel API and mimic what the Rollup plugin does, but there are flaws in my implementation. Feel free to push to this branch.

mischnic:
I think Rollup and Parcel differ enough that it makes more sense to reimplement the features instead of a 1:1 port.

I've made a minimal version work (no automatic JS wrapper/runtime, the generated Wasm simply replaces the previous AS file) and added an example

We probably also need a config file for the assemblyscript compiler options

packages/transformers/assemblyscript/src/ASTransformer.js
```diff
@@ -0,0 +1,28 @@
...
+
+import * as asc from 'assemblyscript/cli/asc';
+
+export default (new Transformer({
+  async transform({asset}) {
+    await asc.ready;
+
+    let code = await asset.getCode();
+    const {binary} = asc.compileString(code, {
+      // optimize: true,
```
aminya, line 14: The user should be able to modify the options here

mischnic: Yes, that's what I meant with 

> We probably also need a config file for the assemblyscript compiler options

aminya:
As-bind is a killer feature and should be also implemented.

aminya:
Automatic wasm instantiation would be nice, but as seen in the Rollup plugin, it requires supporting a notion like 
```
import "asc:something.as"
```

PR #5615, opened 2021-01-08
Bump parcel cli dependencies

aminya:
# ↪️ Pull Request

This PR bumps the dependencies for the cli package.

These dependency bumps include various performance, bug, and security fixes.

For example:
- ora switched to new streams in https://github.com/sindresorhus/ora/pull/163
- chalk had massive performance improvements in https://github.com/chalk/chalk/pull/337

## 🚨 Test instructions

CI and the tests pass.

## ✔️ PR Todo


aminya:
Chalk has the types included in its package. Doesn't flow understand typescript types?

mischnic:
No, Flow doesn't understand Typescript definitons.
chalk used to also contain Flow definitions, but not anymore:
https://unpkg.com/browse/chalk@2.4.2/ vs https://unpkg.com/browse/chalk@4.1.0/

It the API hasn't changed, you could take https://unpkg.com/chalk@2.4.2/index.js.flow and add it to https://github.com/parcel-bundler/parcel/tree/v2/flow-libs (and wrapping the contents with `declare module` like in this file: https://github.com/parcel-bundler/parcel/blob/v2/flow-libs/ansi-html.js.flow)

aminya:
@mischnic There is [`flowgen`](https://www.npmjs.com/package/flowgen) to automatically generate the flow definitions from .d.ts files.

mischnic:
1. There's no `declare module`, these are all global types and not scoped to the package.
2. `type chalk$Chalk` uses non exact types `{+bgWhiteBright: chalk$Chalk, ...}` so accessing a non-existing color wouldn't case a type error.

aminya:
> There's no `declare module`, these are all global types and not scoped to the package.

I added this manually. This can be easily integrated into flowgen using an optional flag.

> And there's no `// @flow` comment at the top, so flow doesn't even read that file

There is a flag for this. `--add-flow-header`.

aminya:
@mischnic Could you check why it still does not detect the types?

mischnic:
Renaming the file to `chalk.js.flow` worked, but apparently the types are wrong because I get
```
Cannot call `chalk.gray` because property `gray` is missing in  object type 
```

aminya:
@mischnic I found a way to make flowgen output usable. You just have to remove the line which does `npm$namespace` and instead of that wrap the file in a declare module. 

Examples:
https://github.com/parcel-bundler/parcel/pull/5615/commits/f33b731b7170f4e4d9ad0b93072fee83d84b66d0
https://github.com/parcel-bundler/parcel/pull/5615/commits/e84638f33834414e30b85a28a37bd73bb3f69ef9

vidstige:
Who can review this? I'm getting security warnings just from installing latest from npm. :roll_eyes:

aminya:
> Who can review this? I'm getting security warnings just from installing latest from npm. 🙄

This is ready to go unless other PRs create conflicts.

mischnic (approved):

Issue #5072, opened 2020-08-25
Heavy memory usage


Banou26:
# 🐛 bug report/question

Trying to import `ffmpeg.js` results in a process taking up 5GBs+ of RAM in a few seconds.
The imported file is 10mb so i can understand it taking up a lot of ram but that much ?

This is something that has been recurring over my projects, parcel processes taking up a lot of ram.
Do you guys have any plans on reducing the memory footprint ?

DeMoorJasper:
We're definitely working on addressing this, it's currently one of our top priorities together with stability

download13:
I've been able to mitigate this problem by passing `PARCEL_WORKERS=1` to it. It seems like the memory usage has something to do with how many worker threads it's trying to run. Maybe they're duplicating work?

roman-petrov:
I have (it seems) exactly the same issue with `ffmpeg.js` and `Parcel 1.12.4`: it runs out of memory when I import `ffmpeg` and start `parcel serve`. Is there any existing workaround, maybe some Parcel CLI arguments?

Banou26:
@roman-petrov You should try using Parcel 2 with the workaround literally on top of your comment.
> I've been able to mitigate this problem by passing PARCEL_WORKERS=1 to it.

roman-petrov:
@Banou26 , thank you. I will try to upgrade to Parcel 2  and use `PARCEL_WORKERS=1`

aminya:
I think Parcel should set the default number of Workers to 1 until the issue is fixed. Using workers is **SLOWER** than just 1 thread! 

Running Parcel on a laptop with a modern processor with 16 threads (AMD Ryzen 7 4800H) and 16GB of RAM:

First setting  the number of workers to 1, the build takes 618 ms.
```
❯ $env:PARCEL_WORKERS=1
❯ ( Measure-Command -Expression { npm run build }  ).Milliseconds
618
```

Then unsetting the number of workers, the build now takes 710 ms and I hit the #4628 issue. So there is no benefit in using the workers.
```
❯ $env:PARCEL_WORKERS=""
❯ ( Measure-Command -Expression { npm run build }  ).Milliseconds
console: (node:19400) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 17 error listeners added to [Socket]. Use emitter.set
MaxListeners() to increase limit
(Use `node --trace-warnings ...` to show where the warning was created)
[...]

Issue #4628, opened 2020-05-19
MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 bundled listeners added to [Bundler]. Use emitter.setMaxListeners() to increase limit


vertangelx:
# 🐛 bug report

`MaxListenersExceededWarning` warning when running app built with Parcel.

> MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 bundled listeners added to [Bundler]. Use emitter.setMaxListeners() to increase limit

## 🎛 Configuration (.babelrc, package.json, cli command)

**package.json**
```js
{
  "scripts": {
    "react-start": "export NODE_OPTIONS=--trace-warnings; parcel -p 3000 index.html --out-dir build",
    "start": "concurrently \"cross-env BROWSER=none yarn react-start\" \"wait-on http://localhost:3000 && electron . \"",
    ...
}
```

## 🤔 Expected Behavior

No such warnings should be raised.

## 😯 Current Behavior

After running `yarn start`, which includes running the React app with `--trace-warnings`, the Electron+React app starts up without any problems. However, we see the `MaxListenersExceededWarning` in the terminal:

```
(node:62770) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 bundled listeners added to [Bundler]. Use emitter.setMaxListeners() to increase limit
[0]     at _addListener (events.js:390:17)
[0]     at Bundler.addListener (events.js:406:10)
[0]     at Bundler.once (events.js:437:8)
[0]     at Server.<anonymous> (/Users/vertangelx/Desktop/electron-app/node_modules/parcel-bundler/src/Server.js:48:15)
[0]     at Server.emit (events.js:315:20)
[0]     at parserOnIncoming (_http_server.js:779:12)
[0]     at HTTPParser.parserOnHeadersComplete (_http_common.js:117:17)
```

## 💁 Possible Solution

Might be a problem with the file watcher `chokidar` for hot reloading?

## 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Parcel           | 1.12.4
| Node             | 14.2.0
| npm/Yarn         | 1.22.4
| Operating System | Mac OSX

daviddarx:
Same here with vue app, seems to be new, didn't get the error before this week.

vazra:
I have the same issue . It can be reproduced at https://github.com/moctrash/parcel2-react-refresh 

[...]

PR #5617, opened 2021-01-09
Limit the number of threads (fix memory usage)

aminya:
# ↪️ Pull Request

See [this comment](https://github.com/parcel-bundler/parcel/issues/5072#issuecomment-757090390) for the analysis behind this PR.

Fixes https://github.com/parcel-bundler/parcel/issues/5072
Fixes https://github.com/parcel-bundler/parcel/issues/4628

## 💻 Examples
Parcel does not utilize all the threads it spawns so there is no point in using long-lived workers when they do not do anything. This results in huge memory usage and slow-down.

For example, here as you see, other than workers 0 and 1, the others almost do nothing. 
image

## ✔️ PR Todo

### DIFF
--- a/packages/core/workers/src/cpuCount.js
+++ b/packages/core/workers/src/cpuCount.js
@@ -41,7 +41,10 @@ export function detectRealCores(): number {
 }
 
 let cores;
-export default function getCores(bypassCache?: boolean = false): number {
+export default function getCores(
+  bypassCache?: boolean = false,
+  limit: number = 4,
+): number {
   // Do not re-run commands if we already have the count...
   if (cores && !bypassCache) {
     return cores;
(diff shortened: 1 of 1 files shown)


aminya:
@mischnic @devongovett As I mentioned in [this comment](https://github.com/parcel-bundler/parcel/issues/5072#issuecomment-757371251) using workers has no measurable benefits. However, if we want to keep that discussion open, I can move the limiting part to another PR, and in this PR I just fix the correct core number detection for Windows. This core number detection should fix the linked issues.

mischnic:
> I can move the limiting part to another PR, and in this PR I just fix the correct core number detection for Window

Please do that

PR #5636, opened 2021-01-13
Update PostHTML

### LINKED ISSUE #5600: Update PostHTML version
(Using `2.0.0-beta.1`)

`posthtml ^0.11` is currently required, which requires `htmlparser2 ^3.9.2` (through `posthtml-parser ^0.4.1`).

Versions of `htmlparser2` less than `4.0.0` are quite broken when some optional HTML5 tags are missing (fixed in https://github.com/fb55/htmlparser2/commit/a40ddc2cf4958b8c875d6cf3c00384455b062037).

Could the version requirement be expanded so that `posthtml-parser` allows for `0.6.0` (as https://github.com/posthtml/posthtml-parser/commit/58258b4c36520e8d59febe7c010b3c9df3f8bc68 updated its `htmlparser2` dependency)?

thewilkybarkid:
Closes #5600.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -9963,7 +9963,7 @@ posthtml-obfuscate@^0.1.5:
   resolved "https://registry.yarnpkg.com/posthtml-obfuscate/-/posthtml-obfuscate-0.1.5.tgz#d65f6642c96349002d33a2208845b2f7517d0fc8"
   integrity sha1-1l9mQsljSQAtM6IgiEWy91F9D8g=
 
-posthtml-parser@^0.4.1, posthtml-parser@^0.4.2:
+posthtml-parser@^0.4.2:
   version "0.4.2"
   resolved "https://registry.yarnpkg.com/posthtml-parser/-/posthtml-parser-0.4.2.tgz#a132bbdf0cd4bc199d34f322f5c1599385d7c6c1"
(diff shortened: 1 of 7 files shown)


yarn.lock
```diff
@@ -9994,15 +9994,7 @@ posthtml-render@^1.2.3, posthtml-render@^1.4.0:
...
-posthtml@^0.11.3:
-  version "0.11.6"
-  resolved "https://registry.yarnpkg.com/posthtml/-/posthtml-0.11.6.tgz#e349d51af7929d0683b9d8c3abd8166beecc90a8"
-  integrity sha512-C2hrAPzmRdpuL3iH0TDdQ6XCc9M7Dcc3zEW5BLerY65G4tWWszwv6nG/ksi6ul5i2mx22ubdljgktXCtNkydkw==
-  dependencies:
-    posthtml-parser "^0.4.1"
-    posthtml-render "^1.1.5"
-
-posthtml@^0.12.0, posthtml@^0.12.3:
+posthtml@^0.12.0:
```
thewilkybarkid, line 9997: Seems to be `htmlnano` not using a peer dependency, and so now requires an older version (was requiring a _newer_ version).

thewilkybarkid: Looking at the `htmlnano` code, it only uses its own copy when running outside of `PostHTML` so is redundant here.

DeMoorJasper (approved):

PR #5638, opened 2021-01-13
Fix sourcemap generation with comments

### LINKED ISSUE #5630: Sourcemap problems (bad sourcecode references) after running npm audit --fix
# 🐛 bug report
I'm not sure how to categorize this bug, but the result of it is that debugging source-code in developer tools (I only tried Chrome) becomes difficult or impossible, presumably due to incorrect sourcemap references.   For example, If I set a break point via `debugger` or in developer tools, it will break and enter the debugger but on the wrong line in the source code, and then when stepping through the code it jumps around to different lines and source files, seemingly at random.

It seems to have come from doing an `npm audit --fix`. I normally specify `next` for Parcel version in my package.json but npm complains that there are a lot of security vulnerabilities and suggests I perform an `npm audit --fix`. So I did that. I am not sure what changes were made except that the version of Parcel in my package.json was updated from `next` to `2.0.0-nightly.524`, and debugging in Chrome developer tools becomes difficult because the lines of code being pointed to are wrong, for example when attempting to set a break point or step through the source code. I presume this is a problem with the sourcemaps. I was able to get it working again by changing the Parcel version back to `next` in `package.json`, wiping my node_modules folder and re-`npm install`-ing, but then npm complains about 46 vulnerabilities again.

I dont have a `.babelrc`, only `package.json`. 
 
in `package.json`,
```js
{
  "devDependencies": {
    "@babel/core": "^7.12.10",
    "documentation": "^13.1.0",
    "eslint": "^7.17.0",
    "eslint-config-prettier": "^6.15.0",
    "parcel": "next",
    "parcel-plugin-nuke-dist": "^1.0.1",
    "prettier": "^2.2.1"
  },
}
```
Parcel `next` version comes out to `2.0.0-beta.1` for most Parcel related things in `package-lock.json`.

## 🤔 Expected Behavior
Opening developer tools using Parcel's provided sourcemaps should allow a developer to set breakpoints in the code and pause at the correct line of the source code.

## 😯 Current Behavior
Breakpoints seem to be observed but happen on the wrong lines and in the wrong files.

## 💁 Possible Solution
Ignore npm's complaints about vulnerabilities and do not perform the suggested `npm audit --fix`

👉  Yes, I did donate some money!  Thanks!!

devongovett:
Fixes #5630. astring has an optimization that only checks for newlines at the end of the string of code being printed rather than in the middle. This broke our custom comment generator which added the next indent at the end of the string rather than being printed separately causing astring not to count lines properly. Added a test for this as well.

### DIFF
--- a/packages/core/integration-tests/test/integration/sourcemap-comments/index.js
+++ b/packages/core/integration-tests/test/integration/sourcemap-comments/index.js
@@ -1,0 +1,9 @@
+console.log('foo');
+// comment
+console.log('bar');
+/* block comment line */
+console.log('baz');
+/* multi line
+  block comment 
+ */
+console.log('idhf');
--- a/packages/core/integration-tests/test/sourcemaps.js
+++ b/packages/core/integration-tests/test/sourcemaps.js
@@ -1197,4 +1197,62 @@ describe('sourcemaps', function() {
     await test(false);
     await test(true);
   });
+
+  it('should handle comments correctly in sourcemaps', async function() {
+    let sourceFilename = path.join(
+      __dirname,
+      '/integration/sourcemap-comments/index.js',
+    );
+    let b = await bundle(sourceFilename, {
+      scopeHoist: true,
+    });
+
+    let filename = b.getBundles()[0].filePath;
+    let raw = await outputFS.readFile(filename, 'utf8');
+    let mapUrlData = await loadSourceMapUrl(outputFS, filename, raw);
+    if (!mapUrlData) {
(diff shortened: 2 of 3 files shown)


5olitaire:
Please help unsubscribe me from this

devongovett:
@5olitaire did you try the unwatch button at the top right of this page?

Screen Shot 2021-01-13 at 10 14 38 AM

DeMoorJasper (approved):

PR #5642, opened 2021-01-13
Add async load to NodePackageManager

aminya:
# ↪️ Pull Request

Tries to fix the issue that was discussed in [this comment](https://github.com/parcel-bundler/parcel/issues/5072#issuecomment-757375571)

## ✔️ PR Todo


packages/core/package-manager/src/NodePackageManager.js
```diff
@@ -71,15 +71,55 @@ export class NodePackageManager implements PackageManager {
...
+    const cachedModule = Module._cache[filePath];
+    if (cachedModule !== undefined) {
+      return cachedModule.exports;
+    }
+
+    let m = new Module(filePath, Module._cache[from] || module.parent);
+    Module._cache[filePath] = m;
+
+    // Patch require within this module so it goes through our require
+    m.require = async id => {
```
mischnic, line 99: Does this actually work? Doesn't `require()` return a promise now instead of the exports?

aminya: Reverted this part. There must be a workaround. 

https://github.com/parcel-bundler/parcel/pull/5642/commits/ce0f80291bfc7704d85d12b732fc03a2fcd5893b

aminya: https://github.com/jaydenlin/node-async-require

packages/core/package-manager/src/NodePackageManager.js
```diff
@@ -71,15 +71,68 @@ export class NodePackageManager implements PackageManager {
     |},
   ): Promise<any> {
     let {resolved} = await this.resolve(name, from, opts);
-    return this.load(resolved, from);
+    return await this.load(resolved, from);
```
aminya, line 74: The linter complains about this `await`!

mischnic: That await isn't needed. If you await, the function implicitly returns a promise. That's basically the same as returning the promise returned by `load` directly.

mischnic:
Is this version faster on your machine?

packages/core/package-manager/src/NodePackageManager.js
aminya: Right. I am used to awaiting all the promises no matter if they are on return or not. But this works too.

aminya:
I don't know how to benchmark it because this is a mono repo. What's your procedure for testing a dev branch?

mischnic:
If your project doesn't use a custom parcelrc, you can simply do:
1. run `yarn build` in the monorepo root
2. remove any parcel deps from your project
3. in your project, run `/path/to/monorepo/parcel/packages/core/parcel/lib/bin.js build index.js`

Alternatively, since this is only one file, you could build the monorepo as above and copy `packages/core/package-manager/lib/NodePackageManager.js` to your project's `node_modules/@parcel/package-manager/lib/NodePackageManager.js`

PR #5671, opened 2021-01-16
Bundle CSSNano optimizer (make Parcel 2-5 s faster)

aminya:
# ↪️ Pull Request

This PR bundles CSSNano which results in a 2-5 seconds build time reduction in the projects that use CSS.

When CSSNano is used, 2-5 seconds is spent to load all the packages required for optimizing CSS. However, by bundling, this time which is spent for reading files from the disk is removed.

image

Related to this issue: 
https://github.com/parcel-bundler/parcel/issues/5072#issuecomment-761254092

NOTE: 
The original PR is [this PR](https://github.com/cssnano/cssnano/pull/985) which uses ES modules for bundling. This PR works on the CJS outputs. Based on my tests, both work as expected without issues. However, the bundle resulting from the ESM code might result in better tree-shaking.

## 💻 Examples

For example, running this PR on [solid-simple-table](https://github.com/aminya/solid-simple-table) results in 5 seconds improvement on the build time.

After:
image
[profile-20210115-194214.zip](https://github.com/parcel-bundler/parcel/files/5823683/profile-20210115-194214.zip)

Before:
image
[profile-20210115-142429.zip](https://github.com/parcel-bundler/parcel/files/5823685/profile-20210115-142429.zip)

## 🚨 Test instructions

Trigger the bundle build:
```
> $env:PARCEL_BUILD_ENV='production'
> yarn
```

Test wherever you want.

## ✔️ PR Todo

## Future Work
We can do the same thing for other optimizers and transformers. For example, @babel/preset-env also takes 1 second to load

### DIFF
--- a/packages/optimizers/cssnano/package.json
+++ b/packages/optimizers/cssnano/package.json
@@ -35,6 +35,7 @@
         "css-declaration-sorter": false,
         "browserslist": false,
         "util-deprecate": false,
+        "cosmiconfig": false,
         "@parcel/source-map": false,
         "@parcel/plugin": false
       },
--- a/packages/optimizers/cssnano/package.json
+++ b/packages/optimizers/cssnano/package.json
@@ -24,5 +24,19 @@
     "@parcel/source-map": "2.0.0-alpha.4.19",
     "cssnano": "^4.1.10",
     "postcss": "^8.0.5"
+  },
+  "scripts": {
+    "build": "cross-env NODE_ENV=production parcel build --target main src/CSSNanoOptimizer.js"
+  },
+  "targets": {
+    "main": {
+      "context": "node",
+      "includeNodeModules": {
+        "css-declaration-sorter": false,
+        "browserslist": false,
+        "util-deprecate": false
+      },
+      "isLibrary": true
+    }
   }
 }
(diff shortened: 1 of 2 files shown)


mischnic:
The downside is that cssnano cannot be upgraded independently (e.g. a cssnano bugfix would require a new release of the optimizer).

aminya:
> The downside is that cssnano cannot be upgraded independently (e.g. a cssnano bugfix would require a new release of the optimizer).

cssnano is already a dependency and it is imported statically. Yo

mischnic:
No,  the range `"cssnano": "^4.1.10"` means that your package manager could decide to use a newer (patch or minor) version.

aminya:
> No, the range `"cssnano": "^4.1.10"` means that your package manager could decide to use a newer (patch or minor) version.

You should decide between dynamicsim or faster build. This is similar to using JavaScript vs Rust. Rust ships a prebuilt binary but JavaScript can dynamically use any arbitrary code at different times and dates.

devongovett:
We're not going to distribute our own copy of cssnano. In fact, we should really change it to match most other plugins which require their dependencies from the user's project rather than having a dependency in the plugin itself.

I recommend that you raise an issue with the cssnano maintainers. However, I doubt that bundling would be a good solution there either given that presets and plugins are configurable and meant to be in separate packages in the first place.

PR #5681, opened 2021-01-18
Updating compileTemplate to use `id` instead of `compilerOptions.scopeId`

### LINKED ISSUE #5629: Vue warning: compileTemplate now requires the `id` option.
# 🐛 bug report

There's a warning when running the integration tests (and probably also when actually using the Vue transformer in a project):

```
[@vue/compiler-sfc] compileTemplate now requires the `id` option.`.
```

## Possible solution

Looking at 
https://github.com/vuejs/vue-next/blob/13fc8ff9dd7007e4ef7d2f0b6e5c973013ccc8b8/packages/compiler-sfc/src/compileTemplate.ts#L189-L212

I would say this should instead use `id` and not `compilerOptions.shortId`?
https://github.com/parcel-bundler/parcel/blob/6e81ebff219f13e2049638923b8819f0a63d40be/packages/transformers/vue/src/VueTransformer.js#L252-L254

But I don't actually know what `id` does...

cc @101arrowz

cemenson:
Closes #5629 

Vue have changed their API to require the `id` field instead of `compilerOptions.scopeId`. They also remove the `data-v-` prefix.

### DIFF
--- a/packages/transformers/vue/src/VueTransformer.js
+++ b/packages/transformers/vue/src/VueTransformer.js
@@ -70,7 +70,7 @@ export default (new Transformer({
     let baseId = md5FromObject({
       filePath: asset.filePath,
     }).slice(-6);
-    let scopeId = 'data-v-' + baseId;
+    let scopeId = baseId;
     let hmrId = baseId + '-hmr';
     let basePath = basename(asset.filePath);
     let {template, script, styles, customBlocks} = nullthrows(
@@ -249,9 +249,7 @@ async function processPipeline({
         source: content,
         inMap: template.src ? undefined : template.map,
         isFunctional,
-        compilerOptions: {
-          scopeId,
-        },
+        id: scopeId,
       });
       if (templateComp.errors.length) {
         throw new ThrowableDiagnostic({
--- a/packages/core/integration-tests/test/vue.js
+++ b/packages/core/integration-tests/test/vue.js
@@ -55,7 +55,7 @@ describe('vue', function() {
     );
     let output = (await run(b)).default;
     assert.equal(typeof output.render, 'function');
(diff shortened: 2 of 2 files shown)


packages/transformers/vue/src/VueTransformer.js
```diff
@@ -70,7 +70,7 @@ export default (new Transformer({
     let baseId = md5FromObject({
       filePath: asset.filePath,
     }).slice(-6);
-    let scopeId = 'data-v-' + baseId;
+    let scopeId = baseId;
```
101arrowz, line 73: You may consider removing this/renaming variables for clarity

101arrowz (approved):

packages/transformers/vue/src/VueTransformer.js
cemenson: I considered that but I thought that keeping `scopeId` provided more clarity than just `id`, and kept changes minimal.

cemenson:
@mischnic is the PR ok to merge? Not sure if I need to do anything about the failed checks - but honestly wouldn't know where to start.

101arrowz:
To see the cause of the failed checks, click the "Details" menu for one of the tests, then click the small link in gray "View more details on Azure Pipelines". From there, you can look at the failing part on the sidebar, which is Integration tests in this case. The error message is that scoped styles are failing. You should be able to test and fix the issue locally. You can try `npm/yarn test:integration -g vue` to test the Vue integration test specifically and debug the issue in the transformer and/or test.

In this case, maybe try logging what `output.__scopeId` actually is, maybe it doesn't match the regex for some reason, but it should be a 6-digit hexadecimal number AFAIK.

mischnic:
Superseded by #6027

PR #5625, opened 2021-01-12
Resolver invalidation

devongovett:
This adds support for tracking files that if created or modified would cause resolution to be invalidated. In addition to the existing support for watching glob patterns for file create events, it also adds support for tracking full file paths, as well as a faster mechanism for matching files of a certain pattern that are created above another file in the directory hierarchy. This is very common (e.g. a package.json, node_modules folder, or `.babelrc` created somewhere above a file), so it makes sense to optimize it. There is no way to represent this as a glob: while e.g. `**/package.json` works, it is inefficient because it matches a package.json anywhere in the project, not just relative to a certain file. In addition, globs are slow to match, so having 1000s of glob nodes would bog down file change event processing.

Instead, this is represented as a sort of trie structure within the graph. For a given pattern, e.g. `node_modules/foo`, we create `file_name` nodes for each parts with `dirname` edges between them. In the above case, that would be two nodes  and one edge. The start and end of the chain are connected to a `file` node via an `invalidated_by_create_above` edge, and this is in turn connected to the request it invalidates by an `invalidated_by_create` edge.

image

When matching, we will start from the first node in the chain, and continue following it to parent directories until there is an `invalidated_by_create_above ` edge pointing to a `file` node that also points to the start of the chain. This indicates a complete match, and any requests attached to the `file` node via an `invalidated_by_create` edge are invalidated. The above case matches when a file or directory called `foo` inside a directory called `node_modules` is created, above `/src/nested/index.js` in the filesystem hierarchy.

This structure is efficient both in terms of memory and CPU. The number of nodes is limited because common names are deduplicated (e.g. `node_modules` and `package.json`), and the `file` nodes are reused with the parent source file node that we already have. And matching can be performed in constant time since there's no need to test all possible globs.

The other case where we could potentially optimize is for extensionless module specifiers, e.g. `./foo` instead of `./foo.js`. In some cases, we need to check many different extensions, so this might result in a lot of extra file nodes in the request graph. We will need to measure the impact of this, but it will only happen in a couple cases: when the file type changes (e.g. a `.ts` file imports a `.js` file), and when a specifier turns out to resolve to a directory rather than a file (all extensions are checked first). We could potentially introduce another special node type to the graph for this case (e.g. filePath minus extension) and match against that instead if the number of nodes proves to be too great.

Close T-662

To do:

- [x] Determine how to handle resolver inside transformer 😬 
- [x] Move more existing glob invalidations to file above invalidations (e.g babel)
- [ ] Test performance and graph size implications
[...]

PR #5686, opened 2021-01-19
Fixes less files imported dependencies not watched

gmougeolle:
# ↪️ Pull Request

This PR adapts the LessTransformer to allow it to include imported files to the asset's included files, resulting in those files being watched properly.

## 💻 Examples

Take a `.less` file that imports other files (using less without CSS Modules, it's very common to have a single file acting as an entrypoint) :

```less
//index.less
import "other.less";
```

```less
//other.less
body {
    background: red;
}
```

If you run `yarn parcel index.less`, the first build works, but any further change to the `other.less` file will remain unnoticed by the bundler, requiring a change to the `index.less` file in order to take changes into account. Others have reported this as well in #5299.

This PR is inspired by the SASS Transformer which does this by default.

## 🚨 Test instructions

I'm not sure whether the watcher part of Parcel supports this kind of tests (I didn't find anything about this in the documentation). The SASS transformer does not include any kind of tests about this case.

## ✔️ PR Todo

- [] Added/updated unit tests for this change
- [] Filled out test instructions (In case there aren't any unit tests)


mischnic:
Does Less read sourcemap comments? In that case, that array could refer to non-existent (or irrelevant) files.

gmougeolle:
I did some quick testing and after trying to reference other less files in different ways (content strings, comments), it didn't end up in the sourcemap file list. However I couldn't find the code responsible for parsing this in the less compiler source so take that with a grain of salt.
Is there a utility in the Parcel core that could check that the file exists as an additional security mesure ? In the meantime, I've published a `parcel-transformer-more` package that simply add those lines for those who want to test it with their code.

mischnic:
Yeah, either `output.imports` or we could also track this ourselves in `resolvePathPlugin` (not sure if `output.imports` is the resolved or unresolved path).

devongovett:
Was this possibly fixed by #5625? See the changes [here](https://github.com/parcel-bundler/parcel/pull/5625/files#diff-8e49c8d44b06d801d9f372ea303a129551a0e8d86e90639084bfdc4a36c19caf)

mischnic:
Your examples appears to work now even without these changes

PR #5690, opened 2021-01-21
Fix web_accessible_resources for web extensions

101arrowz:
# ↪️ Pull Request
[From #5304](https://github.com/parcel-bundler/parcel/pull/5304#issuecomment-759576125):
> I was just testing it out in a Typescript project and everything works great, however, it seems that while the web_accessible_resources do get compiled, the actual file endings in the output manifest.json are not changed to be .js. I think that is a bug but let me know if it is actually correct. If you want me to I can open an issue and build minimal reproducing example in case it's a bug.

## ✔️ PR Todo

- [ ] Added/updated unit tests for this change N/A
- [ ] Filled out test instructions (In case there aren't any unit tests) N/A

/cc @mgranderath


mischnic (approved):

mgranderath:
Any way to get this merged soon? This is the last thing preventing us to fully move over to parcel 2. Again thank you for all of the work on this!

101arrowz:
Not sure why the "generate benchmarks" check is failing, never done that before... any ideas @mischnic or is it not a blocker?

DeMoorJasper:
@101arrowz the benchmark is a non-blocking check, the failing ci does block the merge though...

mischnic:
A test would also be good

mischnic:
As far as I can tell, your change resolves the globs and lists all files in `web_accessible_resources`. How does this relate to

> the actual file endings in the output manifest.json are not changed to be .js

101arrowz:
I was thinking that the packager would now be able to replace the URL references when the glob is expanded. For some reason it seems to no longer work, even though I used this myself earlier...

101arrowz:
Fixed the issue, TypeScript definitely works in `web_accessible_resources` now.

packages/core/integration-tests/test/webextension.js
```diff
@@ -31,8 +36,51 @@ describe('webextension', function() {
...
+      {assets: ['single.js', 'esmodule-helpers.js']},
+    ]);
+    const manifest = await outputFS.readFile(
+      b.getBundles().find(b => b.name == 'manifest.json').filePath,
+      'utf8',
+    );
+    assert(/injected\/index.js/.test(manifest));
+    assert(/injected\/nested\/other.js/.test(manifest));
+    assert(/injected\/index-jsx.js/.test(manifest));
+    assert(/injected\/single.js/.test(manifest));
```
mischnic, line 83: You could actually parse the manifest and read out the value of `web_accessible_resources`.

101arrowz: Fixed.

Issue #5013, opened 2020-08-12
Using --detailed-report <arg> is confusing because it can consume an entry parameter


mischnic:
# 🐛 bug report

Changing
`parcel build a.html b.html`
to
`parcel build --detailed-report a.html b.html`
actually runs
`parcel build b.html --detailed-report`
because `a.html` is used as an argument for the number of assets to print (although it's not even a valid number and therefore ignored)

PR #5701, opened 2021-01-22
Custom parsing / more descriptive logging for options that have choices

joeyslater:
# ↪️ Pull Request

- Display choices if none provided (ie --log-level : none, info, warn, etc)
- Indicate to user if option is invalid (--detailed-report must be an integer)

**Dev Notes:**
* Upgrade to Commander 7 to take advantage of choices which will automatically return to user if not in the enums provided.
* With the upgrade, provided a new way to add more complex options to a command.
* Set a flag on the Commander program to use options the same way. Some re-work to fully support the new Commander API style that wasn’t worth investment quite yet.
* While choices adds messaging to the error message when adding an invalid value, it does not add the messaging if no argument is present. Ergo, I had to override the optionMissingArgument function to print the original message as well as the option description. With choices, this will automatically add messaging to direct what the user could use.

## 💻 Examples

`yarn parcel build src/index.html  --log-level`

previously returns: 
```
error: option '--log-level <level>' argument missing
```

now returns: 
```
error: option `--log-level <level>' argument missing
set the log level (choices: "none", "error", "warn", "info", "verbose")
```

`yarn parcel build src/index.html  --log-level test`

previous: does not validate and continues on

now returns: 
```
error: option '--log-level <level>' argument 'test' is invalid. Allowed choices are none, error, warn, info, verbose.
```

`yarn parcel build src/index.html  --detailed-report test`

previous: does not validate and continues on with default value

now returns: 
```
error: option '--detailed-report [depth]' argument 'test' is invalid. Must be an integer.
```
This might help with #5013 as it should at least alert the user that the argument isn't valid. 

## 🚨 Test instructions

- Currently no unit tests for CLI
- Ran through each example with the listed scripts
- Added the targeted options in the kitchen sink, where the examples above stem.


mischnic:
> error: option '--detailed-report [depth]' argument 'test' is invalid. Must be an integer.


Actually, `[foo]` is an optional parameter. So `--detailed-report` should also work (and use the default of 10).

Also, we should rename `depth` to `count` because it shows the x largest assets in every bundle.

joeyslater:
> > error: option '--detailed-report [depth]' argument 'test' is invalid. Must be an integer.
> 
> Actually, `[foo]` is an optional parameter. So `--detailed-report` should also work (and use the default of 10).
> 
> Also, we should rename `depth` to `count` because it shows the x largest assets in every bundle.

`--detailed-report` still works without an argument. It prevents when the next available argument is invalid, you can see this behavior in #5013 . 

```
$ rm -rf dist && parcel build --detailed-report src/index.html
No entries found
```

I can however update the argument value to count.

mischnic:
Ok, if it's indeed accepting either nothing or a number then that's fine. I was just confused by your description:

> `yarn parcel build src/index.html --detailed-report`
> now returns:
> `error: option '--detailed-report [depth]' argument 'test' is invalid. Must be an integer.`

joeyslater:
> Ok, if it's indeed accepting either nothing or a number then that's fine. I was just confused by your description:
> 
> > `yarn parcel build src/index.html --detailed-report`
> > now returns:
> > `error: option '--detailed-report [depth]' argument 'test' is invalid. Must be an integer.`

Fair play, let me update the description!

wbinnssmith:
Are the flow types for commander 2 accurate for commander 7?

joeyslater:
> Are the flow types for commander 2 accurate for commander 7?

There are not currently flow types defined for Commander. From a cursory typescript type check, I don't see anything breaking. 

I can add flow types in for both 2 (Parcel 1) and 7 (Parcel 2) to verify. Is that something that the maintainers would prefer be a separate request or bundled here?

mischnic:
I tried adding the definitions for commander 2 from here: https://github.com/flow-typed/flow-typed/blob/master/definitions/npm/commander_v2.x.x/flow_v0.104.x-/commander_v2.x.x.js and there are a few new methods in v7 compared to v2.

Parcel 1 didn't use Flow.

I'd just add them to this PR.

mischnic:
You need to put `declare module "commander" {` around the flowgen output

packages/core/parcel/src/cli.js
```diff
@@ -101,10 +104,12 @@ var hmrOptions = {
 
 function applyOptions(cmd, options) {
   for (let opt in options) {
-    cmd.option(
-      opt,
-      ...(Array.isArray(options[opt]) ? options[opt] : [options[opt]]),
-    );
+    const option = options[opt];
+    if (option.flags) {
```
mischnic, line 108: Does this check the same as `option instanceof commander.Option` or are there actually `Option` objects with no `flags` and we want to discern them?

mischnic: Did you undo the fix for this on purpose?

joeyslater: Looks like a rebase fail on my part

mischnic (approved):

Issue #5775, opened 2021-02-04
parcel 2 --open param treats empty string as a browser


jeffpeck10x:
Consider this:
```
"scripts": {
    "dev": "parcel serve src/index.html --open $BROWSER"
  },
```

v.s. this:
```
"scripts": {
    "dev": "parcel serve src/index.html --open \"$BROWSER\""
  },
```

In the first example without quotes, if `$BROWSER` is unset, it will open the default browser as expected.

In the second example with quotes, if `$BROWSER` is unset, nothing happens.

## 🤔 Expected Behavior

`--open $BROWSER` should behave in the same way as `--open "$BROWSER"`

## 😯 Current Behavior

`--open $BROWSER` is not the same as `--open "$BROWSER"`

## 💁 Possible Solution

Change this line to check for empty string:
https://github.com/parcel-bundler/parcel/blob/017e8d795b3237c61c1cdf82c5e68052dda5bf28/packages/core/utils/src/openInBrowser.js#L52
```
-    typeof browser === 'string' ? {app: [getAppName(browser)]} : undefined;
+    typeof browser === 'string' && browser.length > 0 ? {app: [getAppName(browser)]} : undefined;
```

## 🔦 Context

I would like the option to specify a custom browser such that if it is not set, it will use the default browser. The path to the browser might contain spaces, such that the path needs to be quoted.

## 💻 Code Sample

N/A

## 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Parcel           | 2.0.0-nightly.566+dba96de0
| Node             | v12.20.1
| npm/Yarn         | yarn
| Operating System | MacOS 10.15.7 (19H15)

PR #5776, opened 2021-02-04
empty string for --open cli param is treated as undefined

jeffpeck10x:
# ↪️ Pull Request

Makes it so the `--open` option treats an empty string in the same way as `undefined` (i.e. will open the default browser)

So,  `--open $BROWSER` will behave in the same way as this `--open "$BROWSER"`

For this issue: https://github.com/parcel-bundler/parcel/issues/5775

## 💻 Examples

In package.json you should be able to do this:
```
  "scripts": {
    "dev": "parcel serve src/index.html --open \"$BROWSER\""
  },
```
and if `$BROWSER` is unset, it will open the default browser.

## 🚨 Test instructions

Confirm that this:
```
  "scripts": {
    "dev": "parcel serve src/index.html --open \"$BROWSER\""
  },
```

behaves the same way as this:
```
  "scripts": {
    "dev": "parcel serve src/index.html --open $BROWSER"
  },
```

when `$BROWSER` is an actual string and when it is empty.

### DIFF
--- a/packages/core/utils/src/openInBrowser.js
+++ b/packages/core/utils/src/openInBrowser.js
@@ -49,7 +49,7 @@ function getAppName(appName: string): string {
 export default async function openInBrowser(url: string, browser: string) {
   try {
     const options =
-      typeof browser === 'string' ? {app: [getAppName(browser)]} : undefined;
+      typeof browser === 'string' && browser.length > 0 ? {app: [getAppName(browser)]} : undefined;
 
     await open(url, options);
   } catch (err) {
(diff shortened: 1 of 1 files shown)


DeMoorJasper (approved):

packages/core/utils/src/openInBrowser.js
```diff
@@ -49,7 +49,7 @@ function getAppName(appName: string): string {
 export default async function openInBrowser(url: string, browser: string) {
   try {
     const options =
-      typeof browser === 'string' ? {app: [getAppName(browser)]} : undefined;
+      typeof browser === 'string' && browser.length > 0 ? {app: [getAppName(browser)]} : undefined;
```
mischnic, line 52:
```suggestion
      typeof browser === 'string' && browser.length > 0
        ? {app: [getAppName(browser)]}
        : undefined;
```
This should calm down prettier

PR #5824, opened 2021-02-11
Support for a custom source per target with same semantics as package#source

### LINKED ISSUE #3302: Parcel 2: multi entry/target builds
Split out of #2574.

For SSR, it would be useful to support multiple targets with different entries simultaneously.

Three options came up in the previous issue:

## Explicit entries per target

```js
const parcel = new Parcel({
  entries: {
    browserEntry_page1: '/path/to/browser/entry/of/page1.js',
    browserEntry_page2: '/path/to/browser/entry/of/page2.js',
    serverEntry: '/path/to/server.js',
  },

  targets: {
    browserEntry_page1: {
      "browsers": ["> 1%", "not dead"]
    },
    browserEntry_page2: {
      "browsers": ["> 1%", "not dead"]
    },
    serverEntry: {
      "node": ["^8.0.0"]
    },
  }
});
```

## An array of parcel options

```js
const parcel = new Parcel([
  {
    entries: ['page1.js', 'page2.js']
    targets: {
      browser: {
        browser: ['>1%', 'not dead']
      }
    }
  },
  {
    entries: ['server.js']
    targets: {
      node: {
        node: ['^8.0.0']
      }
    }
  }
});
```

## Multiple Parcel instances, sharing a worker farm

We were thinking of making the worker farm an option anyway, so this might work for free. By sharing a worker farm instance, multiple Parcel instances could run in parallel.

```js
const workerFarm = new WorkerFarm();

const browserParcel = new Parcel({
  workerFarm,
  entries: ['page1.js', 'page2.js']
  targets: {
    browser: {
      browser: ['>1%', 'not dead']
    }
  }
});

const serverParcel = new Parcel({
  workerFarm,
  entries: ['server.js']
  targets: {
    node: {
      node: ['^8.0.0']
    }
  }
});
```

Thoughts?

cc. @brillout @padmaia @wbinnssmith

krisnye:
# ↪️ Pull Request

This provides a support for multi-entry builds where a given target has it's own entry point specified.

This is accomplished by adding a `source` property as an option on each target specified in package.json targets.

Any target with a specific source specified will only be built with that source, ignoring all other entries.

Resolves #3302

PR to update website documentation is here: https://github.com/parcel-bundler/website/pull/814

## 💻 Examples

    "source": ["one.js", "two.js"],
    "targets": {
        "a": {
        },
        "b" {
          "source": "three.js"
        }
      }

In this example, the new `source` is specified for target *b*. As a result, source one and two are only built for target *a* and target *b* is only built with source three.

## 🚨 Test instructions

I added example 'target-source' to the examples. It compiles two packages simultaneously, both of which use a source override.

I also added new unit tests for the EntryRequest and TargetRequest.

## ✔️ PR Todo


packages/core/core/src/TargetDescriptor.schema.js
```diff
@@ -112,6 +115,9 @@ export const DESCRIPTOR_SCHEMA: SchemaEntity = {
     distEntry: {
       type: 'string',
     },
+    source: {
+      type: 'string',
+    },
```
mischnic, line 120: This shouldn't be needed because you already added it to `PACKAGE_DESCRIPTOR_SCHEMA.properties `

mischnic:
(In case you haven't seem them, there are a few Flow type errors.)

krisnye:
@mischnic Ya, I'm looking at the flow errors now.

packages/core/core/src/requests/EntryRequest.js
```diff
@@ -143,7 +165,9 @@ class EntryResolver {
     throw new Error(`Unknown entry ${entry}`);
   }
 
-  async readPackage(entry: FilePath) {
+  // Disabling flow because it's complaining that this function is typed as 'any'... but it really should be typed as 'any'
+  // $FlowFixMe
+  async readPackage(entry: FilePath): any {
```
mischnic, line 170: We have a `PackageJSON` type in `@parcel/types`. So maybe `{|...PackageJSON, filePath: FilePath|}` works

krisnye: I'm a Typescript pro but a noob at Flow. Can you pull and take a look at that one. I tried this and some similar types, but then it pushes type problems up to the readPackage call which are beyond my skill to resolve.

`PackageJSON & {| filePath: FilePath, source?: string |} | null`

mischnic: Do this:

```diff
diff --git packages/core/core/src/requests/EntryRequest.js packages/core/core/src/requests/EntryRequest.js
index 6692b71f..8a31bb5d 100644
--- packages/core/core/src/requests/EntryRequest.js
+++ packages/core/core/src/requests/EntryRequest.js
@@ -1,6 +1,6 @@
 // @flow strict-local
 
-import type {Async, FilePath, File} from '@parcel/types';
+import type {Async, FilePath, File, PackageJSON} from '@parcel/types';
 import type {StaticRunOpts} from '../RequestTracker';
 import type {Entry, ParcelOptions} from '../types';
 
@@ -97,12 +97,13 @@ export class EntryResolver {
     if (stat.isDirectory()) {
       let pkg = await this.readPackage(entry);
       if (pkg && typeof pkg.source === 'string') {
-        let source = path.join(path.dirname(pkg.filePath), pkg.source);
+        let pkgSource = pkg.source;
+        let source = path.join(path.dirname(pkg.filePath), pkgSource);
         try {
           stat = await this.options.inputFS.stat(source);
         } catch (err) {
           throw new Error(
-            `${pkg.source} in ${path.relative(
+            `${pkgSource} in ${path.relative(
               this.options.inputFS.cwd(),
               pkg.filePath,
             )}#source does not exist`,
@@ -111,7 +112,7 @@ export class EntryResolver {
 
         if (!stat.isFile()) {
           throw new Error(
-            `${pkg.source} in ${path.relative(
+            `${pkgSource} in ${path.relative(
               this.options.inputFS.cwd(),
               pkg.filePath,
             )}#source is not a file`,
@@ -124,7 +125,7 @@ export class EntryResolver {
         if (pkg.targets) {
           for (let targetName in pkg.targets) {
             let target = pkg.targets[targetName];
-            if (target.source) {
+            if (target.source != null) {
               let filePath = path.join(entry, target.source);
               entries.push({filePath, packagePath: entry, target: targetName});
             }
@@ -165,9 +166,9 @@ export class EntryResolver {
     throw new Error(`Unknown entry ${entry}`);
   }
 
-  // Disabling flow because it's complaining that this function is typed as 'any'... but it really should be typed as 'any'
-  // $FlowFixMe
-  async readPackage(entry: FilePath): any {
+  async readPackage(
+    entry: FilePath,
+  ): Promise<?{...PackageJSON, filePath: FilePath, ...}> {
     let content, pkg;
     let pkgFile = path.join(entry, 'package.json');
     try {
```

The problems you ran into are:
1. `path.dirname(pkg.filePath)` is a function call and invalidates the `typeof === "string"` refinement.
2. Async functions need to return `Promise<actual return value>`
1. (For objects in Flow, `{...PackageJSON, something else}` is preferred over `&`.)

krisnye: Thanks, that seems to work.

krisnye:
@mischnic, I am closing this PR and I have opened a new one I created from scratch so that I could more easily track and resolve the unit test failures.

https://github.com/parcel-bundler/parcel/pull/5861

Issue #5816, opened 2021-02-10
`...$export$getBaseURL` is not defined


mischnic:
# 🐛 bug report

An error is thrown with scope-hoisting and strict mode.

cc @surma

## 🤔 Expected Behavior

Works

## 😯 Current Behavior

```
Uncaught ReferenceError: $10da859f8926f997eca867931986be87$export$getBaseURL is not defined
```

```
function $abc$exec() {
  $abc$exports = {};
  $abc$var$bundleURL = null;
  $abc$export$getBundleURL = $abc$var$getBundleURLCached;
  $abc$exports.getBundleURL = $abc$export$getBundleURL;
  $abc$export$getBaseURL = $abc$var$getBaseURL; // <---
  $abc$export$getOrigin = $abc$var$getOrigin;
  $abc$exports.getOrigin = $abc$export$getOrigin;
}
```

`$abc$export$getBaseURL` is never declared (and indeed unused), so that assignment should just be omitted.

## 💁 Possible Solution

The problem is probably somewhere in here: https://github.com/parcel-bundler/parcel/blob/0d3abd6ffa8fc21bbefe6f48bd6edb9713512edc/packages/shared/scope-hoisting/src/link.js#L842-L887

## 💻 Code Sample

(`type="module"`is just so that it runs in strict mode, this isn't specific to ESM)
```html
<script type="module" src="index.js"></script>
```
```js
if (typeof process !== "undefined" && process.env.TARGET_DOMAIN) {
} else {
	let worker = new Worker("./worker.js");
}

// worker.js
console.log("worker");
```

Smaller reproduction:
```js
if (Date.now() > 0) {
	console.log(require("./bundle-url").getBundleURL());
}

// bundle-url.js
function getBundleURL() {
  return "bundle";
}

function getBaseURL(url) {
  return "base";
}

exports.getBundleURL = getBundleURL;
exports.getBaseURL = getBaseURL;
```

## 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Parcel           | `2.0.0-nightly.579`

PR #5837, opened 2021-02-14
Worklet support

ProLoser:
# ↪️ Pull Request

Support for worklets.
## 💻 Examples
import url from 'worklet:./my-worklet.js'

## ✔️ PR Todo


packages/transformers/clear/src/ClearTransformer.js
```diff
@@ -0,0 +1,10 @@
+// @flow strict-local
+
+import {Transformer} from '@parcel/plugin';
+
+export default (new Transformer({
+  transform({asset}) {
+    asset.symbols.clear();
```
devongovett, line 7: Can you give us some context about why this plugin is necessary?

ProLoser: @mischnic perhaps you could speak to this?

ProLoser: I'm not really sure how `this.#value.symbols` is used, the worklet or worker will likely not be able to share dependencies with the main javascript thread (for now) so I'm curious if this is an attempt to prevent the asset from having additional globals or other dependencies incorrectly bundled into the asset during packaging?

mischnic: This is purely to workaround the `foo.js does not export 'default'` error when doing `import x from "worker:./foo.js"`, has nothing to do with shared bundles.

An alternative is special-casing `asset.isIsolated` in the symbol propagation logic itself.

packages/configs/default/index.json
```diff
@@ -4,6 +4,16 @@
     "types:*.{ts,tsx}": ["@parcel/transformer-typescript-types"],
     "bundle-text:*": ["@parcel/transformer-inline-string", "..."],
     "data-url:*": ["@parcel/transformer-inline-string", "..."],
+    "worker:*": [
+      "@parcel/transformer-worker",
+      "...",
+      "@parcel/transformer-clear"
+    ],
+    "worklet:*": [
+      "@parcel/transformer-worklet",
```
ProLoser, line 13: Since the source code for these is identical, maybe it would be more efficient to make 1 transformer and merely have 2 pipelines here point to the same transformer configuration?

mischnic: They aren't identical. Worklets cannot use `importScripts`. That's the TODO part of this branch

ProLoser:
I wouldn't need this branch merged if I knew how to run parcel commands without having it needs with the dependencies, is there a flag for that?

mischnic:
1. (Turn your project into a monorepo)
2. Add the newly added transformers to your project as packages
3. Add them to the parcelrc

ProLoser:
> 1. Already a monorepo
> 2. I copied transformers/clear, transformers/worker, transformers/worklet, configs/default, core/core, core/types
> 3. Add them to the `.parcelrc`

Cleared node modules and reinstalled, I'm getting lots of packaging headaches. I've also tried making the parcel repo a submodule (on the `worker-url` branch) and adding `"parcel/packages/*/*"` to my `package.json#workspaces[]` but it's nothing but dependency headaches and version mismatches.

image

mischnic:
I haven't had any problems so far with adding a Parcel plugin to a monorepo, for example here: https://github.com/mischnic/parcel-resolver-root. `example` would be your project

ProLoser:
Okay I got it working and it was really easy in the end:

`cd myProject; git submodule add -b worker-url https://github.com/parcel-bundler/parcel.git parcel`

and then I updated my monorepo config:

```json
// myProject/package.json
{
	"workspaces": [
		  "packages/*", // mine
		  "parcel/packages/*/*" // yours
	],
	"dependencies": {
		"parcel": "2.0.0-beta.1"
	}
}
```

The problem I had was I was previously using nightly releases, but the packages from `worker-url` are set to `2.0.0-beta.1` and I really don't want to keep manually updating those with every nightly.

So instead, I decided it's easier to treat the entire parcel dependency tree as a submodule that I can checkout whatever version I want. To get the dependency to work correctly, I had to ensure that my root `package.json` was pointing to the same version as the one specified in the submodule, or else the submodule would not be considered a recent enough release to satisfy the dependency (even though the source code itself IS in fact newer).

I would recommend these instructions for simplicity and minimal dependency headache in the future ^

ProLoser:
I should mention that the way you guys are setting up your dependencies on `@babel/core` is causing multiple copies of babel to be resolved in different packages.

I would recommend peaking at your `yarn.lock` and running `yarn why @babel/core` to see how many versions are being resolved.

ProLoser:
I'm still seeing this when I do a build instead of server
image

ProLoser:
image

I've seen this typically when running built files. The build succeeds but running the code does not. Serving the files succeeds and running succeeds.

mischnic:
Sounds like https://github.com/parcel-bundler/parcel/issues/5816

ProLoser:
Yup that looks like what I'm experiencing and it looks like it was fixed after this branch was created

ProLoser:
I rebased this branch on top of the latest v2 locally and tried building again but it's still showing up for me :/ any tips what I should be looking at?

PR #5828, opened 2021-02-12
allow array in pkg#source

DeMoorJasper:
# ↪️ Pull Request

Running parcel build my-package/ should build a.html, b.html if my-package/package.json contains:

```json
{ "source": ["a.html", "b.html"] }
```

credits to @AceTheCreator for creating the original PR

Closes #5387
Closes #5494

TODO:
- [ ] Write a test

PR #5861, opened 2021-02-17
Support for a custom source per target with same semantics as package (second try) #source #5824

### LINKED ISSUE #3302: Parcel 2: multi entry/target builds

krisnye:
# ↪️ Pull Request

This provides a support for multi-entry builds where a given target has it's own entry point specified.

This is accomplished by adding a `source` property as an option on each target specified in package.json targets.

Any target with a specific source specified will only be built with that source, ignoring all other entries.

Resolves #3302

PR to update website documentation is here: https://github.com/parcel-bundler/website/pull/814

## 💻 Examples

    "source": ["one.js", "two.js"],
    "targets": {
        "a": {
        },
        "b" {
          "source": "three.js"
        }
      }

In this example, the new `source` is specified for target *b*. As a result, source one and two are only built for target *a* and target *b* is only built with source three.

## 🚨 Test instructions

I added example 'target-source' to the examples. It compiles two packages simultaneously, both of which use a source override.

I also added new unit tests for the EntryRequest and TargetRequest.

## ✔️ PR Todo


krisnye:
@mischnic, @devongovett , This look good or do we need anything else?

mischnic:
https://github.com/parcel-bundler/parcel/pull/5828 has caused a merge conflict. And ideally that change would applied to the per-target property as well:
```
"targets": {
    "b" {
      "source": ["three.js", "four.js"]
    }
  }
```

krisnye:
I resolved the merge conflict with #5828 and added support for arrays to the target.source as well to keep semantics the same. Also modified the integration test to verify target.source array works.

packages/core/core/src/requests/EntryRequest.js
```diff
@@ -102,42 +102,74 @@ class EntryResolver {
...
+          for (let targetName in pkg.targets) {
+            let target = pkg.targets[targetName];
+            if (target.source != null) {
+              targetsWithSources++;
+              let targetSources = Array.isArray(target.source)
+                ? target.source
+                : [target.source];
+              for (let source of targetSources) {
+                let filePath = path.join(entry, source);
+                entries.push({
```
devongovett, line 122: Should we do the same validation here as for the top level source field? Eg that the path exists and is a file

krisnye: I refactored the exist/file checks into a shared function and now apply them to both. Also added a new EntryRequest.test.js file and added unit tests for both of those checks on package#source and package#targets.*.source values.
Screen Shot 2021-02-23 at 1 43 23 PM

packages/core/types/index.js
```diff
@@ -133,6 +133,7 @@ export type PackageTargetDescriptor = {|
   +isLibrary?: boolean,
   +optimize?: boolean,
   +scopeHoist?: boolean,
+  +source?: string,
```
devongovett, line 136: Should be `string | Array<string>`?

devongovett (approved):
LGTM aside from one type issue. cc. @mischnic @DeMoorJasper

packages/core/types/index.js
krisnye: Seems it should be consistent with this:
https://github.com/parcel-bundler/parcel/blob/3b709981c5a15aa11a5c9998ad719b4d36d42636/packages/core/types/index.js#L237

Isn't that one wrong though. I'm not super familiar with flow types, but shouldn't that be `FilePath | Array<FilePath>`

devongovett: Yeah looks like they should both be updated. 😬

krisnye: I updated both. This caused a flow error in NodeResolver.js which I fixed but I want some more experienced eyes on.

packages/utils/node-resolver-core/src/NodeResolver.js
```diff
@@ -799,9 +799,12 @@ export default class NodeResolver {
...
     for (let pkgKey of pkgKeys) {
-      let alias = await this.getAlias(filename, pkg, pkg[pkgKey]);
-      if (alias != null) {
-        return alias;
+      let pkgKeyValue = pkg[pkgKey];
+      if (!Array.isArray(pkgKeyValue)) {
+        let alias = await this.getAlias(filename, pkg, pkgKeyValue);
+        if (alias != null) {
+          return alias;
+        }
```
krisnye, line 807: @devongovett @mischnic When I correctly changed the PackageTargetDescriptor.source to "FilePath | Array<FilePath" it caused an error here which I resolved with the explicit Array check.

I'm going to commit this and it didn't break any unit tests, but we should verify that this is the correct aliasing behavior when a source array is provided.

It looks like if source is a single FilePath then it's an alias, if source is an Array then none of the values are aliases.

mischnic: I didn't know that `source` could also behave like `alias`. @devongovett ?

You've already removed that alias variant in the types package, which previously was
```js
	source?: FilePath | {[FilePath]: FilePath, ...},
```
 So maybe we could just not process `source` here anymore? (And then the added condition isn't needed anymore)

packages/core/core/src/requests/TargetRequest.js
```diff
@@ -135,64 +158,80 @@ export class TargetResolver {
           return matchingTarget;
         });
       } else {
+        optionTargets;
```
mischnic, line 161: That can be removed

packages/core/core/src/requests/EntryRequest.js
```diff
@@ -102,42 +130,67 @@ class EntryResolver {
...
+          pkg != null &&
+          pkg.targets != null &&
+          Object.keys(pkg.targets).length === targetsWithSources;
+
+        if (!allTargetsHaveSource && pkg.source != null) {
+          let pkgSources = Array.isArray(pkg.source)
+            ? pkg.source
+            : [pkg.source];
+          for (let pkgSource of pkgSources) {
+            if (typeof pkgSource === 'string') {
```
mischnic, line 181: This condition is always true

packages/core/core/src/requests/TargetRequest.js
```diff
@@ -135,64 +158,80 @@ export class TargetResolver {
...
+            );
+            if (distDir == null) {
+              let optionTargetsString = JSON.stringify(
+                optionTargets,
+                null,
+                '\t',
+              );
+              throw new ThrowableDiagnostic({
+                diagnostic: {
+                  message: `Missing distDir for target "${name}"`,
```
mischnic, line 180:
```suggestion
                  message: md`Missing distDir for target "${name}"`,
```

packages/core/core/src/requests/TargetRequest.js
```diff
@@ -135,64 +158,80 @@ export class TargetResolver {
...
+                  descriptor.optimize !== false,
+                shouldScopeHoist:
+                  this.options.defaultTargetOptions.shouldScopeHoist &&
+                  descriptor.scopeHoist !== false,
+                sourceMap: normalizeSourceMap(
+                  this.options,
+                  descriptor.sourceMap,
+                ),
+              }),
+            }: Target);
```
mischnic, line 220: The cast on the right in `let target: Target = ({...}: Target)` isn't needed in any case

packages/core/core/src/requests/EntryRequest.js
```diff
@@ -63,7 +64,34 @@ async function run({input, api, options}: RunOpts): Promise<EntryResult> {
   return result;
 }
 
-class EntryResolver {
+async function asyncThrowIfNotFile(
```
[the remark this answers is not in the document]
krisnye: how about `rejectIfNotFile`? I just added `async` to make sure that anyone using it understood that it's an asynchronous function so you must await it's completion. If `reject` implies that then that seems fine. If you don't think we need any indication that a void function is async, then I can change to either `assertFile` or `checkFile` whichever you think is more standard.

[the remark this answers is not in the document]
krisnye: For now, I'll just change to `assertFile` since that implies a runtime if it's not.

packages/core/core/src/requests/TargetRequest.js
krisnye: it was needed (along with some other things) at one point in my flow type checker wrangling... but yes, it's not needed any longer. Removed.

packages/core/core/src/requests/EntryRequest.js
krisnye: removed.

packages/core/core/src/requests/TargetRequest.js
krisnye: ya, not sure how that ended up in there. removed.

packages/utils/node-resolver-core/src/NodeResolver.js
devongovett: Yeah that may have been from Parcel 1. I think the idea was to map compiled code back to source when you use a symlinked package.

mischnic: So, should we remove it or keep it (though it's untested at the moment)?

mischnic (approved):
(Apart from https://github.com/parcel-bundler/parcel/pull/5861#discussion_r582369564)

packages/utils/node-resolver-core/src/NodeResolver.js
devongovett: It is documented on the v1 site: https://parceljs.org/transforms.html#third-party-modules.

Whether we want to keep that (and other Parcel specific resolution features) for v2 is questionable. I'm fine with merging as is and deciding outside this PR.

devongovett (approved):

krisnye:
@devongovett @mischnic This website documentation update should also be merged now. https://github.com/parcel-bundler/website/pull/814

Issue #5808, opened 2021-02-09
V2: @parcel/config-webextension: generates unacceptable folder names __


avalanche1:
# 🐛 bug report
Running `parcel` for devving web-extn locally (`yarn start` from `package.json` below).
If extn `manifest.json` references a file outside `src` folder, say `..assets/icon.png` - Parcel will copy that file to `dist/__/assets/icon.png`.
Pls notice the intermediary `__` folder. This is prohibited for chrome extns. When I try to load the unpackaged extn into the browser via pointing to `dist/` - it will throw this err:
`Cannot load extension with file or directory name __. Filenames starting with "_" are reserved for use by the system`

This doesn't happen when running the `build` script, mind you.
## 💻 Code Sample

`manifest.json`:
```json
{
  ...
  "browser_action": {
    "default_icon": {
      "128": "assets/icon/icon.png"
    }
  },
  "icons": {
    "128": "assets/icon/icon.png"
  }
}
```
folder structure:
```
    .
    ├── assets           
      └── icon
        └── icon.png
    ├── dist
      └── __
        └── assets
          └── icon
            └── icon.png
      └── manifest.json
      └── ...
    ├── src
      └── manifest.json
      └── ...
```

## 🎛 Configuration (.babelrc, package.json, cli command)

`.parcelrc`:
```json
{
  "extends": "@parcel/config-webextension",
  "transformers": {
    "*.{ts,tsx}": ["@parcel/transformer-typescript-tsc"]
  }
}
```
` package.json`
```json
{
  "scripts": {
    "start": "parcel src/manifest.json --host localhost --target webext-dev",
    "build": "parcel build src/manifest.json --target webext-prod",
  },
  "targets": {
    "webext-dev": {
      "sourceMap": {
        "inline": true,
        "inlineSources": true
      }
    },
    "webext-prod": {}
  },

}
```
## 🤔 Expected Behavior

Do not generate folders with chrome-restricted names

## 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Parcel           |  `^2.0.0-nightly.574`
| Operating System | `osX 10.14`

avalanche1:
Hmm... after killing `.parcel-cache` and `dist` folders - the issue is gone. What was that about?🤔

avalanche1:
> Hmm... after killing `.parcel-cache` and `dist` folders - the issue is gone. What was that about?🤔

[...]

PR #5893, opened 2021-02-20
Avoid reserved characters in namer when transforming ../../

fregante:
# ↪️ Pull Request

Tentatively fixes https://github.com/parcel-bundler/parcel/issues/5808

Closes #6223

The issue is that some characters are considered reserved by some systems, for examples when names start with a `.` or `_`.

The namer here attempted to escape the `..` representation of a _parent directory_ by replacing the dots with underscores. `__/__/` looks close enough and is understandable, but again underscores have special meanings in some system.

I can't think of non-magic ASCII characters that aren't common letters:

- `--/--/` would be the next candidate but `--` is used for CLI flags so that could cause issues.
- `up/up/` or `parent/parent/` could work as well, but then they have a higher risk of conflict, clobbering files.

So… the closest solution is a combination of both: a short word + an uncommon character _not at the beginning_:

- `up_/up_/`

Thoughts? Alternative ideas?

## 🚨 Test instructions

I think this happens when your `index.html` uses a resource in the upper level, like `<script src="../script.js">`

## ✔️ PR Todo


mischnic:
One alternative would be stripping the `__/` parts and throwing an error if there are colliding names (to make the user rename the sourcefile(s) to remove that collision).

fregante:
Unfortunately I don't think I can help implementing that. I'll close the PR

packages/namers/default/src/DefaultNamer.js
```diff
@@ -114,7 +114,7 @@ function nameFromContent(
 
     return path
       .join(path.relative(entryRoot, path.dirname(entryFilePath)), name)
-      .replace(/\.\.(\/|\\)/g, '__$1');
+      .replace(/\.\.(\/|\\)/g, 'up_$1');
```
fregante, line 117: Let me know if you prefer a different name, like 
```suggestion
      .replace(/\.\.(\/|\\)/g, 'dist/$1');
```

mischnic (approved):

PR #5802, opened 2021-02-08
Parcel plugin dev dep cache invalidation and hot reloading

devongovett:
T-377 T-383

This tracks dev dependencies (e.g. parcel plugins) in the request graph so that we invalidate when they are updated. There are dev_dep_request nodes in the graph which are connected to all of the files that the plugins include via require (tracked by the NodePackageManager). They also store a hash of all of these files. When a file changes, it invalidates the dev_dep_request, which in turns invalidates the asset_request. When the asset request is re-run, all of the valid dev_dep_requests are sent to the worker with their hash. This way we avoid recomputing the hash on each run, except if the dev dep is invalid.

Screen Shot 2021-02-08 at 10 17 39 AM

The dev_dep_request nodes are shared between multiple asset_requests. This reduces the number of edges in the graph because only one edge is needed from asset_request to dev_dep_request rather than an edge to each individual file. It is possible, however, for multiple dev_dep_requests for the same plugin to exist, for example, due to conditional requires. A plugin may require different dependencies depending on the asset (e.g. parcel uses lazy requires that only load a module once it is used). The dependencies are recomputed for each request (though individual file hashes are cached during builds), and a new node will be created for each unique hash.

Aside from enabling cache invalidation for plugins (supporting manual editing of node_modules for debugging, patch-package, local plugins, etc.), this also enables HMR for parcel plugins and their dependencies. Since we already watch all dependencies and trigger a rebuild, we just need to clear the node require cache for that plugin and pretty much get HMR for free. This lets you work on a Parcel plugin or other dev dependency and see the results immediately without restarting the process. 😄 

https://user-images.githubusercontent.com/19409/107240356-d1a4a900-69f7-11eb-9dc9-2c059dca9e45.mov

PR #5483, opened 2020-12-15
Bump emphasize

mischnic:
Closes https://github.com/parcel-bundler/parcel/issues/5694

PR #5900, opened 2021-02-20
Make sure the absolute path isn't contained in the cache

mischnic:
Depends on https://github.com/parcel-bundler/parcel/pull/5802

Closes T-870

This should also always use `/` as path separators

Grepping the absolute path in .parcel-cache didn't yield any results after this change (for a simple test case).

- [x] I wasn't yet able to figure out why the the remaining few tests fail. Something with cache invalidation after file changes is wrong.
- [x] Move the new utils into `utils.js` in `@parcel/core`, no reason to put them into `@parcel/utils` if they're not used outside of core.
- [x] locs
- [ ] still use micromatch 4?
- [x] throw an error if namers return an absolute path
- [x] Windows CI fails?


parcel-benchmark:
## Benchmark Results

#### Timings

| --- | --- | --- |

#### Cold Bundles

#### Cached Bundles

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 8.99s | -202.00ms |
| Cached | 447.00ms | +20.00ms |

#### Cold Bundles

| dist/logo.1e014c76.png | 274.00b | +0.00b | 203.00ms | -63.00ms 🚀 |

#### Cached Bundles

| dist/logo.1e014c76.png | 274.00b | +0.00b | 225.00ms | -75.00ms 🚀 |

</p></details>

#### Timings

| --- | --- | --- |

#### Cold Bundles

#### Cached Bundles

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 6.74s | -151.00ms |
| Cached | 431.00ms | +39.00ms ⚠️ |

#### Cold Bundles

*No bundle changes detected.*

#### Cached Bundles

*No bundle changes detected.*

</p></details>

**[Click here to view a detailed benchmark overview.](https://benchmark.parceljs.org/benchmark/303424942251180609)**

packages/core/core/test/TargetRequest.test.js
```diff
@@ -179,6 +179,7 @@ describe('TargetResolver', () => {
             sourceMap: {},
           },
           loc: {
+            // TODO?
             filePath: path.join(COMMON_TARGETS_FIXTURE_PATH, 'package.json'),
             start: {
```
mischnic: I guess the locs should also be relative to the root

packages/core/core/src/types.js
```diff
@@ -46,6 +45,7 @@ export type ExtendableParcelConfigPipeline = $ReadOnlyArray<
 >;
 
 export type ProcessedParcelConfig = {|
+  // TODO ProjectPath?
   extends?: PackageName | FilePath | Array<PackageName | FilePath>,
```
mischnic: Haven't determined yet if this is ever an absolute path (and if it's stored in the cache at all)

packages/core/core/src/projectPath.js
```diff
@@ -0,0 +1,52 @@
...
+
+export opaque type ProjectPath = string;
+
+export function toProjectPath_(
+  projectRoot: FilePath,
+  p: FilePath,
+): ProjectPath {
+  // TODO we could save the `./` at the start of every path
+  // return p != null ? relativePath(projectRoot, p, false) : p;
+  return p != null ? relativePath(projectRoot, p) : p;
```
mischnic: ...

packages/core/core/src/RequestTracker.js
```diff
@@ -572,15 +592,15 @@ export class RequestGraph extends Graph<
...
+            this.invalidateFileNameNode(fileNameNode, _filePath, above);
           }
         }
 
         for (let id of this.globNodeIds) {
           let globNode = this.getNode(id);
           invariant(globNode && globNode.type === 'glob');
 
-          if (isGlobMatch(filePath, globNode.value)) {
+          if (isGlobMatch(filePath, fromProjectPathRelative(globNode.value))) {
```
devongovett, line 618: So `filePath` here is absolute, but the glob is relative? Won't that not match except if the glob starts with `**/`? I was thinking about how to do this for the glob resolver. Should the glob in the invalidation be relative to the project root? If so, then I think we'd need to make it absolute here.

mischnic: `filePath` is also relative here, further up there's

```js
let filePath = fromProjectPathRelative(_filePath);
```

mischnic: But I ran into https://github.com/parcel-bundler/parcel/pull/5483#issuecomment-746479825 again, so I guess I'll have to use the absolute paths anyway (to make micromatch happy).

As a workaround, I've downgraded to micromatch 3 for now 😬

packages/core/core/src/projectPath.js
```diff
@@ -0,0 +1,52 @@
...
+}
+
+export const toProjectPath: ((
+  projectRoot: FilePath,
+  p: FilePath,
+) => ProjectPath) &
+  ((projectRoot: FilePath, p: FilePath | void) => ProjectPath | void) &
+  // $FlowFixMe Not sure how to type properly
+  ((projectRoot: FilePath, p: ?FilePath) => ?ProjectPath) = toProjectPath_;
```
mischnic, line 30: 🤷‍♂️

packages/core/core/src/types.js
mischnic: This property was apparently unused, so I just removed it...

elliottkember:
Hi! Great work here. this PR would be a really big deal for CI / deployment environments where the build directory changes between builds. Several orders of magnitude difference. At the moment it seems the Parcel cache uses the absolute directory.

packages/core/types/index.js
```diff
@@ -445,10 +445,9 @@ export type DependencyOptions = {|
   +meta?: Meta,
   +pipeline?: string,
   +resolveFrom?: FilePath,
-  +target?: Target,
```
mischnic: This was apparently neither used in the plugins nor inside `addDependency`

devongovett:
One interesting problem is what to do about source maps. The source map object stores the project root at the moment, and source map buffers are stored in the cache. Perhaps we can make the project root a required parameter of the sourcemap constructor even if you are creating it from a buffer, and not include that as part of the serialized payload? @DeMoorJasper

devongovett:
Unfortunately, all of the path conversion seems to add around ~300ms to the esbuild benchmark. We might need to find a faster way of doing this... 😞

devongovett:
Did some optimizations and now the perf is pretty much the same as v2.

devongovett:
Also now storing paths that are outside the project root as absolute rather than relative since they are not portable anyway. This means that at least if the project root moves, the file references with still work (unless the referenced files also moved). Thoughts?

elliottkember:
> Also now storing paths that are outside the project root as absolute rather than relative since they are not portable anyway. This means that at least if the project root moves, the file references with still work (unless the referenced files also moved). Thoughts?

Sounds fine to me!

devongovett:
I think this is ready. I added a util that verifies that there are no references to the project root inside the cache folder after each cache test. After upgrading the source maps package to the latest release, all the tests pass.

packages/core/integration-tests/test/cache.js
```diff
@@ -69,13 +75,33 @@ async function testCache(update: UpdateFn | TestConfig, integration) {
     }
   }
 
+  let resolvedOptions = await resolveOptions(
+    getParcelOptions(getEntries(entries), getOptions(options)),
+  );
+
   let b = await runBundle(entries, options);
 
+  await assertNoFilePathInCache(
```
wbinnssmith, line 84: 🤩

wbinnssmith (approved):
It's a lot, but from what I can tell LGTM. cc @mischnic to review changes from @devongovett

PR #5916, opened 2021-02-23
Fix getBundleURL calls across bundles

### LINKED ISSUE #5906: Relative async imports resolve to incorrect paths in some scenarios
# 🐛 bug report

## 🎛 Configuration (.babelrc, package.json, cli command

I have a minimal-ish reproduction here: https://github.com/marcins/parcel-path-bug-repro. See the README for repro steps.

The build command is:
`parcel build --no-scope-hoist src/packages/entry/entry-a.js src/foo/entry-b.js`

Note the issue does _not_ occur when scope hoisting is enabled.

In terms of the setup - essentially I have two entry points in different paths: `src/packages/entry/entry-a.js` and `src/foo/entry-b.js`, these get bundled to non-root paths in `dist`: `dist/packages/entry/entry-a.js` and `dist/foo/entry-b.js` (if the entry points are in the _same_ path they end up in the dist dir root).

`entry-b` imports a deeply nested module `import deep from './one/two/three/deep.js';`, that deeply nested module has an async import `import("../../four/also-deep")`.  At runtime this import resolves to `http://127.0.0.1:8080/dist/packages/entry/../also-deep.79273ef1.js` which is the wrong path since `also-deep` is at the root of `dist`.

I suspect it's something to do with the `../../..` relative `import` that causes the problem.

## 🤔 Expected Behavior

Parcel should try to load `http://127.0.0.1:8080/dist/also-deep.79273ef1.js`

## 😯 Current Behavior

Parcel tries to load `http://127.0.0.1:8080/dist/packages/entry/../also-deep.79273ef1.js`.

## 💁 Possible Solution

I haven't had a chance to dig into the Parcel sources yet to understand what's causing this. I assume there's an incorrect assumption about some path resolution somewhere.

## 🔦 Context

The scenario is based on the real world structure of Jira's FE codebase, and I ran into this issue when trying to bundle Jira with Parcel 2.

## 💻 Code Sample

https://github.com/marcins/parcel-path-bug-repro

## 🌍 Your Environment

| Parcel           | `2.0.0-nightly.606`
| Node             | `v12.20.1`
| npm/Yarn         | `7.5.5`
| Operating System | `macOS 10.15.7`

mischnic:
# ↪️ Pull Request

Closes #5906

We don't really have a way to test this. `getBundleURL` uses `document.currentScript` in the test environment

Not sure how to test this


parcel-benchmark:
## Benchmark Results

#### Timings

| --- | --- | --- |

#### Cold Bundles

#### Cached Bundles

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 9.90s | -122.00ms |
| Cached | 464.00ms | -5.00ms |

#### Cold Bundles

| dist/index.js | 486.89kb | +45.00b ⚠️ | 5.61s | +88.00ms |

#### Cached Bundles

| dist/index.js | 486.89kb | +45.00b ⚠️ | 5.57s | +63.00ms |
| dist/logo.1e014c76.png | 274.00b | +0.00b | 212.00ms | -25.00ms 🚀 |

</p></details>

#### Timings

| --- | --- | --- |

#### Cold Bundles

#### Cached Bundles

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 7.19s | -42.00ms |
| Cached | 421.00ms | +1.00ms |

#### Cold Bundles

*No bundle changes detected.*

#### Cached Bundles

*No bundle changes detected.*

</p></details>

**[Click here to view a detailed benchmark overview.](https://benchmark.parceljs.org/benchmark/305067993395626561)**

packages/runtimes/js/src/bundle-url.js
```diff
@@ -1,3 +1,4 @@
+// eslint-disable-next-line no-unused-vars
 /* globals document:readonly */
```
mischnic: These two lines can be removed altogether

packages/core/integration-tests/test/output-formats.js
```diff
@@ -614,7 +614,7 @@ describe('output formats', function() {
       await assertESMExports(b, {a: 2, c: 5});
     });
 
-    it.skip('should support esmodule output (re-export namespace as)', async function() {
+    it.only('should support esmodule output (re-export namespace as)', async function() {
```
mischnic: This tests was skipped previously because it is a known limitation of scope-hoisting. (And I just left the test case in at the time so that we can reenable it once it's supported)

packages/core/integration-tests/test/output-formats.js
```diff
@@ -1153,8 +1155,8 @@ describe('output formats', function() {
       );
       assert(
         new RegExp(
-          'Promise.all\\(\\[.+?getBundleURL\\(\\) \\+ "' +
-            path.basename(asyncCssBundle.filePath) +
+          "Promise.all\\(\\[.+?getBundleURL\\('[a-zA-Z0-9]+'\\) \\+ \"" +
+            asyncCssBundle.name +
```
mischnic: This is still wrong

packages/core/test-utils/src/utils.js
```diff
@@ -588,14 +589,41 @@ function prepareBrowserContext(
...
-    currentScript: {
-      src: 'http://localhost/script.js',
-    },
+
+    currentScript: null,
   };
 
   var exports = {};
+
+  class customError {
```
wbinnssmith: nit: UpperCamelCase? and maybe a better name

Also, I wonder if this should be a function with a `.prototype` property that's the current Error.prototype. Otherwise assumptions about this being the "real" Error may fail.

wbinnssmith (approved):

mischnic:
Maybe `PatchedError.name = "Error"` fixes that failing test

packages/core/integration-tests/test/integration/differing-bundle-urls/foo/one/deep.js
```diff
@@ -0,0 +1,2 @@
+sideEffect("deep");
+console.log("deep");
```
mischnic: The test shouldn't contain log statements so that the mocha output doesn't become cluttered with them

Bildschirmfoto 2021-04-19 um 20 34 50

packages/runtimes/js/src/bundle-url.js
```diff
@@ -1,5 +1,3 @@
-/* globals document:readonly */
-
 var bundleURL = null;
 function getBundleURLCached() {
```
mischnic: The original bugfix got lost somewhere along the way: [`4237582` (#5916)](https://github.com/parcel-bundler/parcel/pull/5916/commits/4237582f322c48b2d55484844fcbbe8a13dd7ead#diff-39112519f42223a8e331d0dac27d01f8a70db6fe3d515ad8016856931a6ccbda)

packages/runtimes/js/src/bundle-url.js
```diff
@@ -1,28 +1,22 @@
-/* globals document:readonly */
-
-var bundleURL = null;
-function getBundleURLCached() {
-  if (!bundleURL) {
-    bundleURL = getBundleURL();
+var bundleURL = {};
+function getBundleURLCached(id) {
+  let value = bundleURL[id];
```
mischnic:
```suggestion
  var value = bundleURL[id];
```

packages/core/test-utils/src/utils.js
```diff
@@ -588,14 +589,51 @@ function prepareBrowserContext(
...
+  function PatchedError(message) {
+    const patchedError = new Error(message);
+    const stackStart = patchedError.stack.indexOf('at new Error');
+    const stackEnd = patchedError.stack.includes('at Script.runInContext')
+      ? patchedError.stack.indexOf('at Script.runInContext')
+      : patchedError.stack.indexOf('at runNextTicks');
+    const stack = patchedError.stack.slice(stackStart, stackEnd).split('\n');
+    stack.shift();
+    stack.pop();
+    stack.unshift(`    at getBundleURL (${bundle.name})`);
```
mischnic: So now every stack traces will contain this `getBundleURL` entry, even if the error was thrown somewhere else?

gorakong: perhaps we could change it to `    at (${bundle.name})`?

wbinnssmith: The problem isn't the function name, it's that we're effectively hardcoding where the function is defined, even if it's defined in another bundle.

devongovett:
Should we rebase this and get it into the rc?

packages/runtimes/js/src/bundle-url.js
```diff
@@ -1,28 +1,22 @@
...
-  }
-
   try {
     throw new Error();
   } catch (err) {
     var matches = ('' + err.stack).match(/(https?|file|ftp):\/\/[^)\n]+/g);
     if (matches) {
-      return getBaseURL(matches[0]);
+      // Use the last occurrence so that the URL of the calling bundle is returned, not of the bundling containing this
+      return getBaseURL(matches[matches.length - 1]);
```
devongovett: Couldn't one bundle call another bundle which then calls getBundleUrl and then the last item in the stack would be from the first bundle rather than the bundle that called getBundleUrl?

devongovett (approved):

Issue #5745, opened 2021-02-02
parcel does not handle the 'distDir' option correctly.


Tarocch1:
# 🐛 bug report

parcel does not handle the 'distDir' option correctly.

## 🎛 Configuration (.babelrc, package.json, cli command)

**package.json**

```json
{
  "name": "test",
  "version": "0.1.0",
  "private": true,
  "pageIndex": "src/pages/Index/index.html",
  "targets": {
    "pageIndex": {
      "distDir": "dist/index",
      "publicUrl": "."
    }
  },
  "devDependencies": {
    "parcel": "^2.0.0-beta.1",
    "react": "^17.0.1",
    "react-dom": "^17.0.1"
  }
}
```

**Build command**

```bash
$ npx parcel build src/pages/Index/index.html
```

## 🤔 Expected Behavior

Bundled files should be outputted in `dist/index` folder.

## 😯 Current Behavior

Bundled files are outputted in `src/pages/Index` with source code file. My source code file `index.html` has even been modified by parcel.

![image](https://user-images.githubusercontent.com/15124068/106546411-189d2500-6546-11eb-82cb-8d346d7206f5.png)

## 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Parcel           |2.0.0-beta.1|
| Node             |v14.15.4|
| npm/Yarn         |npm 6.14.10|
| Operating System | Windows 10|

Issue #5281, opened 2020-10-26
Parcel2 build command outputs bundles in my src folder instead of dist folder.


ajm113:
# 🐛 Parcel2 build command outputs bundles in my src folder instead of dist folder.

When I run the `parcel build ./src/index.html --dist-dir ./dist` the bundle will get outputted to my `src` directory instead of the `dist` directory. Running `parcel serve ./src/index.html` works as expected.

## 🎛 Configuration (.babelrc, package.json, cli command)

```js
{
  "name": "my app",
  "version": "1.0.0",
  "description": "This is a app",
  "main": "src/index.html",
  "scripts": {
    "start": "parcel serve ./src/index.html",
    "build": "parcel build ./src/index.html --dist-dir ./dist",
    "test": "echo \"Error: no test specified\" && exit 1"
  },
  "author": "Andrew",
  "license": "ISC",
  "alias": {
    "components": "./src/components",
    "containers": "./src/containers",
    "style": "./src/style",
    "assets": "./src/assets",
    "libs": "./src/libs",
    "workers": "./src/workers",
    "db": "./src/db"
  },
  "devDependencies": {
    "parcel": "next",
    "sass": "^1.26.11"
  },
  "dependencies": {
    "bulma": "^0.9.1",
    "lunr": "^2.3.9",
    "react": "^16.13.1",
    "react-dom": "^16.13.1",
    "react-router-dom": "^5.2.0"
  }
}
```

## 🤔 Expected Behavior

It should build to `./dist` even without the additional flag.

## 😯 Current Behavior

Throws all bundle files and replaces my `index.html` with the bundle version.

## 🔦 Context

I'm building a very basic app using React to try out Parcel v2. I never experienced this issue using Parcel v1. Was really hoping to migrate the team to v2. :cry: 

## 💻 Code Sample

N/A

## 🌍 Your Environment

Arch Linux 64x

| Software         | Version(s) |
| ---------------- | ---------- |
| Parcel           |  2.0.0-beta.1
| Node             | v12.19.0
| npm/Yarn         |  6.14.7/1.22.10
| Operating System | Arch Linux

mischnic:
Remove this line:
```
	"main": "src/index.html",
```

Duplicate of https://github.com/parcel-bundler/parcel/issues/4255 (because the src folder should never be overwritten because of a configuration error).

(Also, `--dist-dir ./dist` isn't needed, that's the default already).

ajm113:
[...]

Issue #5041, opened 2020-08-19
parcel 2 build command may dangerously overwrite source files in workspace folder


gogiii:
# 🐛 bug report

Parcel 2 drops build result into current (project) folder instead of ./dist by default (comparing to parcel1)!
it also **overwrites** html source file but creates scripts with hashed names.
project example of two files: test.js, test.html
>parcel build test.html
```
√ Built in 6.05s
test.html           167 B    930ms
test.8db25ec3.js     70 B    1.26s
```

Using "dist-dir" flag doesn't work - it's ignored!
>parcel build test.html --dist-dir dist

Same behaviour using both yarn or npm.

## 🎛 Configuration (.babelrc, package.json, cli command)

package.json
```js
{
  "name": "test",
  "version": "1.0.0",
  "main": "test.html",
  "dependencies": {
  },
  "devDependencies": {
    "parcel": "^2.0.0-beta.1"
  },
  "scripts": {
    "build": "parcel build test.html --dist-dir dist"
  }
}
```

## 🤔 Expected Behavior

Original files not changed by parcel.
Build result appear in ./dist folder.

## 😯 Current Behavior

test.html is **overwritten** by parcel's result in current (project) folder:

```<html><head><meta charset="utf8"><title>Test</title></head><body> <div id="root"></div> <script src="/test.8db25ec3.js" type="text/javascript"></script> </body></html>```

test.js name is hashed correctly as test.8db25ec3.js avoiding being overwritten.

## 💻 Code Sample

test.js
``
console.log("hello world");
``

test.html
```
<html>
    <head>
        <meta charset="utf8"/>
        <title>Test</title>
    </head>
    <body>
        <div id="root"></div>
        <script src="test.js" type="text/javascript"></script>
    </body>
</html>
```

## 🌍 Your Environment

| Software  | Version(s) |
| ------------ | ---------- |
| Parcel       | 2.0.0-beta.1
| Node        | v12.18.3
| Yarn         | yarn v1.22.4
| npm         | 6.14.6
| Operating System | Windows 7 64-bit

cameronelliott:
I almost lost hours of work because of this change when I switched from v1 to v2. This really got my attention.

devongovett:
Your main field in package.json is pointing to the same html file as your input. Parcel uses this field as the _output_, thus it overwrites it. Removing the main field should fix the issue. It's invalid anyway because only JS files can be referenced from that field.

mischnic:
[...]

PR #5941, opened 2021-03-01
Prevent users from accidentally overwriting their source code

### LINKED ISSUE #4255: Error if package#source == package#main or package#module
# 🐛 bug report

Parcel 2 shouldn't overwrite the source in these situations:

```json
{
	"module": "src/index.js",
	"source": "src/index.js"
}
```

`parcel build .`

---

```json
{
	"main": "src/index.js",
}
```

`parcel build src/index.js`

JessieFrance:
# ↪️ Pull Request

This PR is for [#4255](https://github.com/parcel-bundler/parcel/issues/4255), and aims to prevent users from accidentally overwriting their own source code.

Users accidentally overwrite their own code when they configure their package.json main, module, browser, or distDir fields to be the same as the source entry file. This happened earlier in duplicates [#5281](https://github.com/parcel-bundler/parcel/issues/5281), and [#5041](https://github.com/parcel-bundler/parcel/issues/5041), and possibly also in [#5745](https://github.com/parcel-bundler/parcel/issues/5745).

This code patch would catch these previous cases, and would throw a diagnostic error with hints for package.json configurations, with a link to relevant Parcel V2 documentation.

## 💻 Examples
- Similar to the examples @mischnic gave in the [#4255](https://github.com/parcel-bundler/parcel/issues/4255) bug report, typing ```parcel build .``` in the terminal will overwrite src/index.js with this library mode package.json configuration:
```
{
	"main": "src/index.js",
	"source": "src/index.js"
}
```
In the above, replacing "main" with either "module" or "browser" will also overwrite src/index.js.

- Likewise, ```yarn run build``` with the custom target package.json configuration below will also overwrite source code:
```
{
  "targets": {
    "app": {
      "distDir": "src/index.html"
    }
  },
  "scripts": {
    "build": "parcel build src/index.html"
  }
}
```

## 🚨 Test instructions

Please see the proposed integration tests and run with
```yarn test:integration -g 'overwrite an entry'```

## ✔️ PR Todo

- [x] Added/updated ~~unit~~ integration tests for this change

Thank you for reviewing this PR. It's been fun working on this bug, and I'd love to hear your feedback.

Fixes #4255

### CODE REVIEW

**Comment by mischnic** [comment]
That check should rather be performed in here: https://github.com/parcel-bundler/parcel/blob/806cc65e0e430b1120cdcf1676e2f58738326d30/packages/core/core/src/requests/TargetRequest.js#L87-L108

(It would still execute in the same order as your current version, but its cleaner to put it into the TargetRequest itself)

**Comment by JessieFrance** [comment]
Thank you @mischnic for the helpful feedback on this. I moved the check over to the TargetRequest where you suggested (and did a rebase in the process to get rid of the last fix).

**Comment by JessieFrance** [comment]
@mischnic -- Sorry to disturb you, but I noticed that my integration test scenarios fail on the Windows CI pipeline, succeed on Linux, and possibly succeed on Mac. I originally created these test cases on my local Linux machine. Since I do not have good access to Windows to debug this, I set up a trial Azure account and ran the Parcel Windows node v12/v14 pipelines on this with some logging for the four proposed integration test cases. 

If I did this correctly, it appears that the 'browser', 'custom-target', and 'main' package.json test scenarios are not triggering the desired Parcel diagnostic error on Windows at all (despite correctly doing so on Linux), and the 'module' case on Windows triggers a different Parcel error - "esmodule output is not supported without scope hoisting" - instead of the intended error message that I get on Linux. Do you have any suggestions? Many thanks in advance for your thoughts.

File: packages/core/core/src/requests/TargetRequest.js [hunk 1]
```diff
@@ -91,6 +91,24 @@ async function run({input, api, options}: RunOpts) {
   );
   let targets = await targetResolver.resolve(input.packagePath, input.target);
 
+  for (const currTarget of targets) {
+    if (
+      currTarget.name !== 'default' &&
+      currTarget.distEntry != null &&
+      currTarget.distDir + '/' + currTarget.distEntry === input.filePath
```
DeMoorJasper, line 98: not sure if this'll help in resolving the windows test case issue but you probably want to use path utils for this, something like `path.join(currTarget.distDir, currTarget.distEntry) === input.filePath` or even `path.join(currTarget.distDir, currTarget.distEntry) === path.normalize(input.filePath)`

JessieFrance: @DeMoorJasper you're right. That code looks a lot cleaner. I'll use one of those if I'm able to debug this. Thank you.

JessieFrance:
@DeMoorJasper your suggestion on using path utilities solved the previous issues my four test cases had on Windows. Thank you!

devongovett:
@JessieFrance hope you don't mind me pushing to your branch. I added a bit more detail to the diagnostics, e.g. a code frame.

image
image

devongovett (approved):

Issue #5943, opened 2021-03-02
Invalid Version: undefined


LeoMsky:
Everything is OK yesterday，but I've had this problem since I installed it again this morning
![image](https://user-images.githubusercontent.com/18640050/109591153-fd730480-7b47-11eb-9a89-4ef5c55fb0b3.png)

skjb:
FYI reverting to 1.12.3 seems to work

DarinTayTwoXAR:
Pinning @babel/preset-env to 7.13.8 is what worked for me.

The version check that is failing here was added in 7.13.9 here - https://github.com/babel/babel/commit/c155caf1e8f229e546b22980ecb53313b90334d0#diff-cddb13dc00454073d19373b40529b8047e25b9141ab627eb8be482601f16eabe

vk-sourcefuse:
Yes, reverting to 1.12.3 is working

bemnlam:
Even a minimal repo with Parcel `1.12.4` installed does not work:

```bash
npm init -y && npm i parcel && touch index.js && echo "<html><body>Minimal Parcel<script src=\"./index.js\"></script></body></html>" > index.html
```

Add a `dev` command in `package.json`:
```json
"scripts": {
  "dev": "parcel index.html"
},
```

And then run `npm run dev`. This is the error:

```bash
Server running at http://localhost:1234 
🚨  /Users/ben.lam/poc/testbuild/index.js: Invalid Version: undefined
    at new SemVer (/Users/ben.lam/poc/testbuild/node_modules/@babel/preset-env/node_modules/semver/semver.js:314:11)
    at compare (/Users/ben.lam/poc/testbuild/node_modules/@babel/preset-env/node_modules/semver/semver.js:647:10)
    at lt (/Users/ben.lam/poc/testbuild/node_modules/@babel/preset-env/node_modules/semver/semver.js:688:10)
    at /Users/ben.lam/poc/testbuild/node_modules/@babel/preset-env/lib/index.js:276:22
    at Object.default (/Users/ben.lam/poc/testbuild/node_modules/@babel/helper-plugin-utils/lib/index.js:22:12)
    at getEnvPlugins (/Users/ben.lam/poc/testbuild/node_modules/parcel/src/transforms/babel/env.js:62:34)
    at getEnvConfig (/Users/ben.lam/poc/testbuild/node_modules/parcel/src/transforms/babel/env.js:12:25)
[...]

PR #5946, opened 2021-03-02
Pass version to presetEnv

bemnlam:
# ↪️ Pull Request

Fixing #5943

[This change](https://github.com/babel/babel/pull/12934/files#diff-cddb13dc00454073d19373b40529b8047e25b9141ab627eb8be482601f16eabeR301) in babel breaks Parcel if version doesn't pass into `presetEnv()`. By setting version as 7.13.0 `presetEnv()` will continue other checking, just like the logic before.

### DIFF
--- a/packages/transformers/babel/src/env.js
+++ b/packages/transformers/babel/src/env.js
@@ -47,9 +47,8 @@ export default async function getEnvOptions(
 }
 
 function getNeededPlugins(targets: BabelTargets): Array<PresetEnvPlugin> {
-  const { version } = require("@babel/core")
   return presetEnv(
-    {version: version , assertVersion: () => true},
+    {version: require('@babel/core').version, assertVersion: () => true},
     {targets: targets},
   ).plugins.filter(p => p[0]);
 }
--- a/packages/transformers/babel/src/env.js
+++ b/packages/transformers/babel/src/env.js
@@ -47,8 +47,9 @@ export default async function getEnvOptions(
 }
 
 function getNeededPlugins(targets: BabelTargets): Array<PresetEnvPlugin> {
+  const { version } = require("@babel/core")
   return presetEnv(
-    {version: '7.13.0', assertVersion: () => true}, // version is required since https://github.com/babel/babel/pull/12934/
+    {version: version , assertVersion: () => true},
     {targets: targets},
   ).plugins.filter(p => p[0]);
 }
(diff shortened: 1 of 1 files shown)


devongovett:
I think this is a v1 only bug? This code path won't actually be hit in v2 because config.isSource is always true.

nicolo-ribaudo:
Q: If that code path won't be hit, the code can just be deleted?

packages/transformers/babel/src/env.js
```diff
@@ -48,7 +48,7 @@ export default async function getEnvOptions(
 
 function getNeededPlugins(targets: BabelTargets): Array<PresetEnvPlugin> {
   return presetEnv(
-    {assertVersion: () => true},
+    {version: '7.13.0', assertVersion: () => true}, // version is required since https://github.com/babel/babel/pull/12934/
```
nicolo-ribaudo, line 51: This should be 7.12.0 (https://github.com/parcel-bundler/parcel/blob/806cc65e0e430b1120cdcf1676e2f58738326d30/packages/transformers/babel/package.json#L43), or `require("@babel/core").version`.

Ideally, the safest approach would be `{ ...require("@babel/core"), assertVersion: () => true }`.

mischnic:
Yes, 

https://github.com/parcel-bundler/parcel/blob/806cc65e0e430b1120cdcf1676e2f58738326d30/packages/transformers/babel/src/config.js#L38-L41

means that this whole if statement could be removed because it's never reached if `!config.isSource`:

https://github.com/parcel-bundler/parcel/blob/806cc65e0e430b1120cdcf1676e2f58738326d30/packages/transformers/babel/src/env.js#L30-L41

bemnlam:
@nicolo-ribaudo @mischnic so looks like this is a v1-only issue and what can I do if I also want to apply this change to v1? (sorry I am a newbie and I want to help 🙏)

nicolo-ribaudo:
@bemnlam Parcel 1 is on the `master` branch, but I'd wait for the ok from a maintainer before opening a PR for Parcel 1.

agilgur5:
@devongovett @mischnic could y'all review this change, particularly for v1? #5943 is being felt quite massively across the community as I'm sure you're seeing. I was going to write this PR myself (even though I'm only indirectly impacted) before finding it was already done here.

Currently, the only solution right now is either a `resolution` or a not-so-easy migration to the v2 beta -- this fix is considerably simpler than either.

devongovett:
v1.12.5 has a fix. Please see https://github.com/parcel-bundler/parcel/issues/5943#issuecomment-799899064

PR #5959, opened 2021-03-03
ReactRefreshBabelTransform change to check peerDependencies in addition to dependencies.

krisnye:
# ↪️ Pull Request

Currently the ReactRefreshBabelTransform plugin is NOT applied if your react dependency is in your package.json#peerDependencies. This change fixes it so that it will be.

## 💻 Examples

This should cause react refresh to be used.

package.json
`
"peerDependencies": {
    "react": "16.13.1"
},
`

## 🚨 Test instructions

In any project, move your `react` dependency to peerDependencies, restart parcel and verify HMR still works with react refresh.

## ✔️ PR Todo

- [n/a] Added/updated unit tests for this change
- [n/a] Included links to related issues/PRs


mischnic (approved):

krisnye:
@mischnic any idea why the build is failing. Seeing an error in loadParcelPlugin.js which doesn't seem like it could be caused by this trivial change.

mischnic:
Don't worry about the test failures, they're indeed unrelated. We can also rerun CI without you having to commit something

krisnye:
OK, in that case I'm done.

PR #6142, opened 2021-04-15
Fix regular expression for file:/// URLs

### LINKED ISSUE #6141: Worker bundle URL failing to get file:/// origin
# 🐛 bug report

I'm trying to import a web worker inside an Electron app. This works fine in parcel most of the time, but then fails when I bundle the app. The code worked in Parcel v1, but I'm having all kinds of problems in Parcel v2.

I've tried all these things:

```
worker = new Worker("./ledWorker.js");
```
```
import ledWorker from './ledWorker';
worker = new Worker(ledWorker);
```
```
 const worker = new Worker(new URL('./ledWorker.js', import.meta.url), {
   type: 'module'
 });
```
```
const worker = new Worker(new URL(workerUrl));
```

But all I get is this:
```
Uncaught Error: Origin not found
    at Object.n.getOrigin (index.8ca53705.js:190)
    at e.exports (marketing.433cd28f.js:1)
    at Object.45QHR../get-worker-url (marketing.433cd28f.js:1)
    at _ (marketing.433cd28f.js:1)
    at c (marketing.433cd28f.js:1)
```

The issue is that the worker file is a local file, `file:///Users/elliott/.../worker.js`. `getOrigin` is returning undefined on the `file:///` URL instead of finding `file://`. 

https://github.com/parcel-bundler/parcel/blob/v2/packages/runtimes/js/src/get-worker-url.js#L6

## 🤔 Expected Behavior

I should be able to import a web worker with `new Worker('./myWorker')`

## 😯 Current Behavior

It fails to load the web worker and then raises an exception

image

## 💁 Possible Solution

I think getOrigin here:

https://github.com/parcel-bundler/parcel/blob/7717e2547f510c89ea57bbfec62147bfb912ba49/packages/runtimes/js/src/bundle-url.js#L40

needs to be updated to handle file URLs correctly. It's not currently matching file:/// URLs. Is this intended, or am I doing something wrong?

```
> url = "file///Users/elliott/file.js"
'file///Users/elliott/file.js'
> ('' + url).match(/(https?|file|ftp):\/\/[^/]+/);
null
```

I tried writing a better regular expression for this to find a host, but I am not great at regular expressions.

## 🌍 Your Environment

| Parcel           | v2
| Node             | v15.9
| npm/Yarn         | latest
| Operating System | OSX

elliottkember:
It seems as though file:/// URLs aren't currently matched correctly on their origins.

The `file:///` URL has an extra slash when finding a local file, which isn't accounted for in this regular expression. This PR adds an extra optional `/` to the end of the regular expression's matcher for origin.

Closes https://github.com/parcel-bundler/parcel/issues/6141

# ↪️ Pull Request

## ✔️ PR Todo

### DIFF
--- a/packages/runtimes/js/src/bundle-url.js
+++ b/packages/runtimes/js/src/bundle-url.js
@@ -38,6 +38,7 @@ function getBaseURL(url) {
 // TODO: Replace uses with `new URL(url).origin` when ie11 is no longer supported.
 function getOrigin(url) {
   let matches = ('' + url).match(/(https?|file|ftp):\/\/[^/]+/);
+  let matches = ('' + url).match(/(https?|file|ftp):\/\/(\/)?[^/]+/);
   if (!matches) {
     throw new Error('Origin not found');
   }
--- a/packages/runtimes/js/src/bundle-url.js
+++ b/packages/runtimes/js/src/bundle-url.js
@@ -37,7 +37,10 @@ function getBaseURL(url) {
 
 // TODO: Replace uses with `new URL(url).origin` when ie11 is no longer supported.
 function getOrigin(url) {
-  let matches = ('' + url).match(/(https?|file|ftp):\/\/(\/)?[^/]+/);
+  if (url.slice(0, 7) === 'file://') {
+    return 'file://';    
+  }
+  let matches = ('' + url).match(/(https?|file|ftp):\/\/[^/]+/);
   if (!matches) {
     throw new Error('Origin not found');
   }
--- a/packages/runtimes/js/src/bundle-url.js
+++ b/packages/runtimes/js/src/bundle-url.js
@@ -31,14 +31,13 @@ function getBundleURL() {
 
 function getBaseURL(url) {
   return (
-    ('' + url).replace(/^((?:https?|file|ftp):\/\/.+)\/[^/]+$/, '$1') + '/'
+    ('' + url).replace(/^((?:https?|file|ftp):\/\/.+)\/+$/, '$1') + '/'
   );
 }
 
 // TODO: Replace uses with `new URL(url).origin` when ie11 is no longer supported.
 function getOrigin(url) {
-  let matches = ('' + url).match(/(https?|file|ftp):\/\/[^/]+/);
-  let matches = ('' + url).match(/(https?|file|ftp):\/\/(\/)?[^/]+/);
+  let matches = ('' + url).match(/(https?|file|ftp):\/\/+/);
   if (!matches) {
     throw new Error('Origin not found');
   }
--- a/packages/runtimes/js/src/bundle-url.js
+++ b/packages/runtimes/js/src/bundle-url.js
@@ -37,7 +37,7 @@ function getBaseURL(url) {
 
 // TODO: Replace uses with `new URL(url).origin` when ie11 is no longer supported.
 function getOrigin(url) {
-  let matches = ('' + url).match((https?|file|ftp):\/\/(\/)?[^/]+);
+  let matches = ('' + url).match(/(https?|file|ftp):\/\/(\/)?[^/]+/);
   if (!matches) {
     throw new Error('Origin not found');
   }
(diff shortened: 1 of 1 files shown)


elliottkember:
On further inspection, the regular expression should also be finding the host for http(s):// URLs, and right now it's only finding the scheme.

elliottkember:
OK, so on further investigation I got this a bit wrong. The "origin" of a `file:///` url is `file://`, and doesn't include the first section of the URL.

packages/runtimes/js/src/bundle-url.js
```diff
@@ -37,6 +37,9 @@ function getBaseURL(url) {
 
 // TODO: Replace uses with `new URL(url).origin` when ie11 is no longer supported.
 function getOrigin(url) {
+  if (url.slice(0, 7) === 'file://') {
+    return 'file://';
```
mischnic: The linter is complaining about the spaces after the semicolon

mischnic:
We should also add an integration test similar to 
https://github.com/parcel-bundler/parcel/blob/1ec530f64c227b6a0cc9f814382bae21bc6944b8/packages/core/integration-tests/test/javascript.js#L542

packages/runtimes/js/src/bundle-url.js
```diff
@@ -37,6 +37,9 @@ function getBaseURL(url) {
 
 // TODO: Replace uses with `new URL(url).origin` when ie11 is no longer supported.
```
DeMoorJasper, line 38: Do we still need to support IE11 or can we polyfill `URL` dynamically if the user requests IE11 support? Seems it would simplify this code a lot. (potential polyfill https://yarnpkg.com/package/url-polyfill)

This is kinda unrelated to this PR getting merged, is mainly something for core team to consider

cc @mischnic

mischnic: Every time we included `let` or some newer API like `"...".includes` in a runtime/the prelude, someone opened an issue that it doesn't work on IE11, so I guess we should still support it.

And in this case, `getOrigin()` is 170 bytes, while that polyfill is 6.2kb (both minified but not gzip).

DeMoorJasper: Right, so there's still more need for IE11 than I thought. Thanks for clearing that up.

The polyfill was just a suggestion and should've probably been inserted only if there's IE11 in browserslist, so the bundle would probably be smaller on newer browsers as we don't do any kind of polyfill and always used built-in URL

packages/runtimes/js/src/bundle-url.js
elliottkember: Thanks. Sorry I didn't catch that. My auto-formatting was causing me headaches so I had it turned off.

elliottkember:
I've been testing this with `"postinstall": "cp bundle-url.js node_modules/@parcel/runtime-js/lib/bundle-url.js",` using the version of `bundle-url.js` that's in this PR. It seems to work very nicely and finally got me all working in Electron. I'm using this to load a couple of Web Workers.

DeMoorJasper:
Awesome, can you add an integration test so we can merge this?

elliottkember:
> Awesome, can you add an integration test so we can merge this?

I've been trying to do this, but I'm not sure how since this code is executed at runtime in the browser a URL with `self.location.origin`. 

I don't think I'm doing it right at all. I've tried copying a few other cases, but I keep hitting the error:

```
  1) javascript
       should support bundling workers with file:/// URLs:
     index.js:158
navigator.serviceWorker.register(require("service-worker.js"), {
^

ReferenceError: navigator is not defined
```

Maybe a unit test on `getOrigin` would work?

DeMoorJasper:
@elliottkember yes a unit test would be easier to write and probably sufficient for this

mischnic:
I would use a regular web worker, then you can do something like this

https://github.com/parcel-bundler/parcel/blob/3baf990a686d1c1dd0c5f4360296de890cc00c47/packages/core/integration-tests/test/javascript.js#L1195-L1212

PR #6194, opened 2021-04-27
Add bundle builds

aminya:
# ↪️ Pull Request

- This PR tries to add bundle builds for Parcel to speed up loading time. The plan is to provide a `parcel-bundle` binary that uses the bundled builds instead of the source files. This is the second take on #5671, but without modifying the default behavior of the packages (with `parcel` binary).

- It's always a very good test to use a tool on itself. Bundling Parcel using Parcel can reveal many bugs, and also result in improving the Parcel code itself.

## ✔️ PR Todo

### DIFF
--- a/packages/core/parcel/package.json
+++ b/packages/core/parcel/package.json
@@ -55,7 +55,8 @@
         "@parcel/package-manager": false,
         "@parcel/fs-search": false
       },
-      "isLibrary": true
+      "isLibrary": true,
+      "outputFormat": "commonjs"
     }
   }
 }
--- a/packages/core/parcel/package.json
+++ b/packages/core/parcel/package.json
@@ -59,5 +59,8 @@
       "isLibrary": true,
       "outputFormat": "commonjs"
     }
+  },
+  "alias": {
+    "@parcel/core": "@parcel/core/lib/index.bundle.js"
   }
 }
--- a/packages/core/parcel/package.json
+++ b/packages/core/parcel/package.json
@@ -53,8 +53,7 @@
         "@parcel/core": false,
         "@parcel/fs": false,
         "@parcel/package-manager": false,
-        "@parcel/fs-search": false,
-        "highlight.js": false
+        "@parcel/fs-search": false
       },
       "isLibrary": true
     }
--- a/packages/core/parcel/package.json
+++ b/packages/core/parcel/package.json
@@ -14,8 +14,12 @@
     "type": "git",
     "url": "https://github.com/parcel-bundler/parcel.git"
   },
-  "bin": "src/bin.js",
+  "bin": {
+    "parcel": "lib/bin.js",
+    "parcel-bundle": "lib/bin.bundle.js"
+  },
   "main": "lib/bin.js",
+  "bundle": "lib/bin.bundle.js",
   "source": "src/bin.js",
   "engines": {
     "node": ">= 12.0.0"
@@ -38,5 +42,22 @@
   },
   "devDependencies": {
     "@parcel/babel-register": "2.0.0-beta.2"
+  },
+  "scripts": {
+    "bundle": "cross-env NODE_ENV=production parcel build --target bundle lib/bin.js --no-scope-hoist"
+  },
+  "targets": {
+    "bundle": {
+      "context": "node",
+      "includeNodeModules": {
+        "@parcel/core": false,
+        "@parcel/fs": false,
+        "@parcel/package-manager": false,
+        "@parcel/fs-search": false,
+        "highlight.js": false,
+        "node-forge": false
+      },
+      "isLibrary": true
+    }
   }
 }
--- a/packages/core/parcel/package.json
+++ b/packages/core/parcel/package.json
@@ -44,7 +44,7 @@
     "@parcel/babel-register": "2.0.0-beta.2"
   },
   "scripts": {
-    "bundle": "cross-env NODE_ENV=production parcel build --target bundle lib/bin.js --no-scope-hoist"
+    "bundle": "cross-env NODE_ENV=production parcel build --target bundle src/bin.js --no-scope-hoist"
   },
   "targets": {
     "bundle": {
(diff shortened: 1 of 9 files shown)


packages/core/parcel/package.json
```diff
@@ -38,5 +42,20 @@
...
     "@parcel/babel-register": "2.0.0-beta.2"
+  },
+  "scripts": {
+    "bundle": "cross-env NODE_ENV=production parcel build --target bundle lib/bin.js --no-scope-hoist"
+  },
+  "targets": {
+    "bundle": {
+      "context": "node",
+      "includeNodeModules": {
+        "@parcel/core": false,
```
aminya, line 53: Failed because of `require.resolve('./worker')`

packages/core/parcel/package.json
```diff
@@ -38,5 +42,20 @@
...
+        "@parcel/fs": false,
+        "@parcel/package-manager": false,
```
aminya, line 55: Creates a cryptic error when the bundle is used.
```js
Error: Name already registered with serializer
    at registerSerializableClass (C:\parcel\packages\core\core\lib\serializer.js:43:11)
    at Object.<anonymous> (C:\parcel\packages\core\fs\lib\NodeFS.js:229:39)
    at Module.n._compile (C:\parcel\packages\core\parcel\lib\bin.bundle.js:64:1076211)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1137:10)
    at Module.load (node:internal/modules/cjs/loader:988:32)
    at Function.Module._load (node:internal/modules/cjs/loader:828:14)
    at Module.require (node:internal/modules/cjs/loader:1012:19)
    at o (C:\parcel\packages\core\parcel\lib\bin.bundle.js:64:1075905)
    at Object.<anonymous> (C:\parcel\packages\core\fs\lib\index.js:21:15)
    at Module.n._compile (C:\parcel\packages\core\parcel\lib\bin.bundle.js:64:1076211)
 ERROR  Command failed with exit code 1.
```

packages/core/parcel/package.json
```diff
@@ -38,5 +42,20 @@
...
+        "@parcel/fs-search": false
```
aminya: The prebuild relies on the location of the file.

packages/core/parcel/package.json
```diff
@@ -38,5 +42,20 @@
   },
   "devDependencies": {
     "@parcel/babel-register": "2.0.0-beta.2"
+  },
+  "scripts": {
+    "bundle": "cross-env NODE_ENV=production parcel build --target bundle lib/bin.js --no-scope-hoist"
```
aminya: Scope hoisting results in this error when the bundle is used:
```js
TypeError: Class extends value #<Object> is not a constructor or null
    at $e0ff572d13646e06f40e076cefd6686f$exec (C:\parcel\packages\core\parcel\lib\bin.bundle.js:25:1424680)
    at $e0ff572d13646e06f40e076cefd6686f$init (C:\parcel\packages\core\parcel\lib\bin.bundle.js:25:1443053)
    at $ee4cb4f5c2d240a4c58df653433665f4$var$_commander (C:\parcel\packages\core\parcel\lib\bin.bundle.js:25:1447963)
    at Object.<anonymous> (C:\parcel\packages\core\parcel\lib\bin.bundle.js:25:1454346)
    at Module._compile (node:internal/modules/cjs/loader:1108:14)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1137:10)
    at Module.load (node:internal/modules/cjs/loader:988:32)
    at Function.Module._load (node:internal/modules/cjs/loader:828:14)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:76:12)
    at node:internal/main/run_main_module:17:47
```

packages/core/parcel/package.json
```diff
@@ -54,8 +54,7 @@
         "@parcel/fs": false,
         "@parcel/package-manager": false,
         "@parcel/fs-search": false,
-        "highlight.js": false,
-        "node-forge": false
```
aminya: `node-forge` makes the bundle very large. Although not all of its exports are used.

packages/core/parcel/package.json
```diff
@@ -54,8 +54,7 @@
         "@parcel/fs": false,
         "@parcel/package-manager": false,
         "@parcel/fs-search": false,
-        "highlight.js": false,
```
aminya: `highlight-js` makes the bundle very large. Although many of its required languages are not used.

aminya: Made a pull request to fix this issue
#6196

packages/core/parcel/package.json
aminya: Made a pull request to fix this issue: #6198

mischnic:
Is this significantly faster for you? Because all of the plugins as specified in the parcelrc are still unbundled

aminya:
> Is this significantly faster for you? Because all of the plugins as specified in the parcelrc are still unbundled

This isn't ready yet. It's just the start. Also, the improvements shouldn't be measured only based on build time. As I mentioned in the OP, it's always a very good test to use a tool on itself. Bundling Parcel using Parcel can reveal many bugs, and also result in improving the Parcel code itself.

So far I have extracted these two pull requests from this pull request which are helpful even for the default Parcel binary.
#6196
#6198

mischnic:
> As I mentioned in the OP, it's always a very good test to use a tool on itself. Bundling Parcel using Parcel can reveal many bugs, and also result in improving the Parcel code itself.

This isn't the first effort to do that: https://parcel2-repl.vercel.app

aminya:
Thanks. In addition to the benefits I mentioned before, these changes (like the fixes I have done in the other two PRs) will reduce the payload size significantly.

I am anticipating that the main build-time performance benefits start to show themselves when we bundle the plugins such as cssnano. I already showed this in #5671. Just bundling CSSNano reduced the build time by 2-5 seconds. This is a "constant overhead" that anyone who uses CSS will pay! By the new `parcel-bundle` binary, these overheads are removed (for those who optionally use it).

I am going to take a look at the source code of the REPL to see which parts I can borrow for `parcel-bundle`.

PR #6198, opened 2021-04-27
Lazy load node-forge only if needed (Make Parcel ~300KB smaller)

aminya:
# ↪️ Pull Request

This changes the usage of `generateCertificate` so it is only loaded when a HTTPS server is going to be created. Node-forge which is used by `generateCertificate` is ~300KB or more which is now lazy loaded.

## 🚨 Test instructions

The current server/hmr integration tests cover this.

## ✔️ PR Todo

### DIFF
--- a/packages/core/utils/src/generateCertificate.js
+++ b/packages/core/utils/src/generateCertificate.js
@@ -1,6 +1,7 @@
 // @flow
 import type {FileSystem} from '@parcel/fs';
-import forge from 'node-forge';
+import {default as forgePki} from 'node-forge/lib/pki';
+import {default as forgeMd} from 'node-forge/lib/md';
 import path from 'path';
 import logger from '@parcel/logger';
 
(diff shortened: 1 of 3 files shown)


mischnic:
This should make **no** difference because we use the lazy mode of Babel's esm-to-cjs transform https://github.com/parcel-bundler/parcel/blob/cb9b82e3693c9c64dbb3aba1b420308d294d2407/packages/dev/babel-preset/index.js#L22

aminya:
If we can find a way to remove node-forge altogether, it would make the code much smaller.

aminya:
Not sure why the tests fail.

aminya:
OK. I fixed the tests. The dynamic import is surprisingly wrapping the exports of the module in one another object. Not sure if this is caused by Flow or Babel. So I had to write it like this:
```js
const generateCertificate = (await import('./generateCertificate.js'))
      .default.default;
```

aminya:
It seems that Flow doesn't see this extra `default`. I am guessing this is added by Babel.

aminya:
> This should make **no** difference because we use the lazy mode of Babel's esm-to-cjs transform

Using Node's built-in async `import` is better than using a Babel shim which wraps sync `require`. Also, in #6194, `node-forge` is added to the bundle. Parcel only does splitting whenever dynamic imports are used.

PR #6230, opened 2021-05-03
Rewrite JS transformer and scope hoisting in Rust

devongovett:
This is a rewrite of the current JS transformer, including scope hoisting, in Rust. It utilizes [SWC](https://swc.rs) for parsing and codegen. All of the features of the existing JS transformer have been ported, including dependency detection, process.env replacement, fs inlining, node global insertion, and scope hoisting. In addition, SWC replaces Babel by default for transpilation when a `browserslist` is set, as well as for compiling JSX and TypeScript, and React Fast Refresh.

**Overall, build times are up to 10x faster!**

On the esbuild benchmark, without terser:

* **Current v2:** 34s
* **This PR:** 3.2s

With terser:

* **Current v2:** 56s
* **This PR:** 18s

Still plenty of optimization potential to go, but this is a great foundation to build on!

It's also really exciting that something like this is even possible. Parcel's language-agnostic core made this entirely possible in plugins. In fact, the old Babel-based plugins and scope hoisting implementation still work entirely without modification. In the future, we will be able to swap out additional plugins to improve performance or output quality as well.

## Scope hoisting rewrite

Aside from performance, this PR was motivated by a need to rethink how Parcel's scope hoisting/tree shaking implementation works. We ran into some bugs especially having to do with preserving execution ordering (e.g. #5659) and duplicate execution  that necessitated us to rethink how we handle wrapped assets via a module registry. The result is a hybrid between scope hoisting where we can, but falling back wrapping modules in a CJS-style function where needed. The packager is also redesigned to work entirely on strings rather than ASTs, which improves performance massively. You can read more about how this works in the [design document](https://github.com/parcel-bundler/parcel/blob/swc-scope-hoisting/packages/transformers/js/hoist.md) for the new scope hoisting implementation.

* Fixes #5659
* Fixes #5658
* Fixes #5657
* Fixes #5662
* Fixes #5606
* Fixes #5591
* Fixes #5892
* Fixes #5883
* Fixes #5782
* Fixes #3989
* Fixes #6051
* Fixes #6056
* Fixes #6071
* Fixes #6076
* Fixes #6167

Possibly fixes (needs verification) #5064 #5914

Non scope hoisting issues: Fixes #5560

## Migration

[...]

PR #5871, opened 2021-02-17
[T-417] Babel ast location remapping

DeMoorJasper:
# ↪️ Pull Request

This PR remaps all Babel ast nodes to the original sourcemap location if there's a sourcemap available.

This is an alternative solution to #5831 

### Performance

AK-Editor (best of 3): Before: 42 secs, After: 40 secs

### TODO

- [x] Remap location data in ast nodes
- [x] Remove old remapping logic
- [x] Ensure source files can be read and embedded into the sourcemap

fixes #5620

## ✔️ PR Todo

- [x] Added/updated unit tests for this change
- [x] Filled out test instructions (In case there aren't any unit tests)
- [x] Included links to related issues/PRs

PR #6238, opened 2021-05-06
Babel ast location, diagnostic, and source location remapping

wbinnssmith:
#5871 applied to #6230.

This also adds a fallback to read and use the original asset's code as sourceContent when an existing sourcemap is not found.

Fixes T-1042

Test Plan: 
* See reproduction in T-1042
* ~This should be implemented as an integration test.~ Added an integration test


packages/core/core/src/Transformation.js
```diff
@@ -158,6 +162,14 @@ export default class Transformation {
           },
         ]);
       }
+
+      if (existing == null) {
+        // If no existing sourcemap was found, initialize sourcesContent with
+        // the asset's filepath and its original contents.
+        asset.sourcesContent = {
+          [asset.value.filePath]: await asset.getCode(),
```
mischnic: This has to definitely be a relative path of some sorts

packages/core/core/src/applyRuntimes.js
```diff
@@ -76,8 +79,20 @@ export default async function applyRuntimes({
...
+
+            let sourcemap = SourceMap.generateEmptyMap({
+              projectRoot: pluginOptions.projectRoot,
+              sourceName,
+              sourceContent: code,
+            });
+
             let assetGroup = {
               code,
+              mapBuffer: sourcemap.toBuffer(),
```
wbinnssmith: This is still going to be an issue as it (along with the code) gets serialized into the bundlegraph. cc @DeMoorJasper @devongovett

packages/core/core/src/Transformation.js
DeMoorJasper: The other sourcescontents are relative to projectroot as enforced by the sourcemap library

DeMoorJasper (approved):

packages/core/core/src/Transformation.js
```diff
@@ -161,6 +162,16 @@ export default class Transformation {
...
         ]);
       }
+
+      if (existing == null) {
+        // If no existing sourcemap was found, initialize sourcesContent with
+        // the asset's filepath and its original contents.
+        asset.sourcesContent = {
+          [path
+            .relative(this.options.projectRoot, asset.value.filePath)
+            .replace(new RegExp(path.sep, 'g'), '/')]: await asset.getCode(),
```
mischnic: Should this be https://github.com/parcel-bundler/parcel/blob/8b240d03dec1bec95bb232070de7cb60890171b1/packages/transformers/babel/src/babel7.js#L30 
instead?

parcel-benchmark:
## Benchmark Results

#### Timings

| --- | --- | --- |

#### Cold Bundles

#### Cached Bundles

</p></details>

#### Timings

| --- | --- | --- |

#### Cold Bundles

#### Cached Bundles

</p></details>

#### Timings

| --- | --- | --- |

#### Cold Bundles

#### Cached Bundles

</p></details>

#### Timings

| --- | --- | --- |

#### Cold Bundles

#### Cached Bundles

</p></details>

**[Click here to view a detailed benchmark overview.](https://benchmark.parceljs.org/benchmark/298774518238806538)**

packages/packagers/js/src/ScopeHoistingPackager.js
```diff
@@ -173,6 +173,26 @@ export class ScopeHoistingPackager {
 
     res += this.outputFormat.buildBundlePostlude();
 
+    if (sourceMap) {
+      let promises = [];
+      // Traverse the bundle to get all source contents
+      this.bundle.traverseAssets(asset => {
```
devongovett: This doesn't seem quite right. Not all of the assets in the bundle will necessarily be included in the final output. Some might be skipped e.g. due to no used symbols. Also, shouldn't the sourcemap library handle this when merging the source maps from the original asset?

packages/core/core/src/Transformation.js
```diff
@@ -158,6 +163,17 @@ export default class Transformation {
           },
         ]);
       }
+
+      if (existing == null) {
+        // If no existing sourcemap was found, initialize sourcesContent with
+        // the asset's filepath and its original contents.
+        asset.sourcesContent = {
```
devongovett: Why is this stored separately on the asset rather than in the source map? If there's no existing source map, should we just initialize an empty one containing only the original code?

packages/packagers/js/src/ScopeHoistingPackager.js
DeMoorJasper: I think you're right the new scope hoisting no longer needs this as now you're using the sourcemap library already and not relying on the AST for all of this.

Previously we were relying on the Babel AST for this and the AST does not keep a reference to the source content so we had to extract that from the maps and manually add this. We can probably refactor this a bit further to make it a bit more lightweight now that the scope-hoisting implementation already supports the source-maps properly.

packages/core/core/src/applyRuntimes.js
devongovett: Yes. Do we actually need the sourcemap here? It's empty and only contains the same code as is already passed through the request. If there's no existing map, won't it already be set as the source content during transformation? I guess we should have an integration test for runtimes + sourcemaps.

packages/shared/babel-ast-utils/src/index.js
```diff
@@ -24,6 +25,42 @@ import {generator, expressionsPrecedence} from './generator';
 
 export {babelErrorEnhancer};
 
+export function remapAstLocations(ast: BabelNodeFile, map: SourceMap) {
+  // remap ast to original mappings
+  // This improves sourcemap accuracy and fixes sourcemaps when scope-hoisting
```
devongovett, line 30: I guess this is only used if you are using babel and there's an existing sourcemap now?

devongovett:
Changed a few things in #6286. LMK if I missed something.

packages/shared/babel-ast-utils/src/index.js
DeMoorJasper: Yes

packages/core/core/src/Transformation.js
DeMoorJasper: Because of this line: https://github.com/parcel-bundler/parcel/blob/v2/packages/core/core/src/UncommittedAsset.js#L102

packages/core/core/src/applyRuntimes.js
devongovett: Another issue is that with empty maps all of the columns mappings are zero when the empty map is extended. If we overwrote the `filePath` here though it breaks a lot of tests that expect `JSRuntime.js` as the name. Maybe the test util should just ignore runtime assets?

PR #6247, opened 2021-05-09
Implement support for classic scripts and automatic module/nomodule

### LINKED ISSUE #333: Some dependencies that rely on window don't work

### LINKED ISSUE #3011: Parcel 2: Differential Bundling

devongovett:
This implements #5398 on top of the new SWC-based JS transformer. It takes a slightly different approach: rather than adding support for multiple contexts to the environment, it adds a `sourceType` field, which can either be `script` for classic scripts, or `module` for ES modules.

The other difference from the original implementation is the way scripts hoist their top-level variables into the global scope. In the previous implementation, this was done by attempting to find all top-level vars and assign them to the global object. In this implementation, the code for the entry asset of script bundles itself is moved outside the bundle wrapper. Any `require` calls that the entry asset might do (e.g. for runtimes) are replaced with a `parcelRequire` call.

Otherwise, everything is the same as the previous PR. See there for details.

~~The posthtml parser fork for adding location information will be removed once https://github.com/posthtml/posthtml-parser/pull/63 is merged and released.~~ DONE

Close T-136. Fixes #333. Closes #3011.


packages/transformers/js/src/JSTransformer.js
```diff
@@ -246,17 +293,20 @@ export default (new Transformer({
...
+        asset.env.shouldScopeHoist && asset.env.sourceType !== 'script',
+      source_type: asset.env.sourceType === 'script' ? 'Script' : 'Module',
+      supports_module_workers: supportsModuleWorkers,
     });
 
     let convertLoc = loc => ({
       filePath: relativePath,
       start: {
-        line: loc.start_line,
+        line: loc.start_line + Number(asset.meta.startLine ?? 1) - 1,
```
devongovett: This is kind of a hack. Should we do something in core for this?

packages/transformers/js/src/lib.rs
```diff
@@ -258,13 +261,16 @@ fn transform(ctx: CallContext) -> Result<JsUnknown> {
...
-                },
+                Optional::new(
+                  EnvReplacer {
+                    replace_env: config.replace_env,
+                    env: config.env,
+                    is_browser: config.is_browser,
+                    decls: &decls,
+                    used_env: &mut result.used_env
+                  },
+                  config.source_type != SourceType::Script
```
devongovett: Should process.env be supported in scripts?

packages/core/types/index.js
```diff
@@ -148,6 +148,8 @@ export type TargetDescriptor = {|
   +distEntry?: FilePath,
 |};
 
+export type SourceType = 'script' | 'module';
```
mischnic, line 151: ~It doesn't change anything effectively, but I think it would makes if this were~
```suggestion
export type SourceType = null | 'script' | 'module';
```
~instead. Because there really are 3 types and not 2?~ Looks like there are only 2 and `Environment#sourceType` should not be nullable?

packages/transformers/html/src/dependencies.js
```diff
@@ -157,6 +149,62 @@ export default function collectDependencies(
...
+        attrs.nomodule = '';
+        attrs.src = asset.addURLDependency(attrs.src, {
+          // Keep in the same bundle group as the HTML.
+          priority: 'parallel',
+          env: {
+            sourceType,
+            outputFormat: 'global',
+            loc,
+          },
+        });
```
mischnic, line 188: Do these statements do anything? `attrs` is a shallow clone and it was already merged into `copy`

packages/core/core/src/Environment.js
```diff
@@ -13,10 +13,12 @@ export function createEnvironment({
   engines,
   includeNodeModules,
   outputFormat,
+  sourceType,
```
wbinnssmith: (assuming `null` is a desirable type for `SourceType`)

```suggestion
  sourceType = 'module',
```

packages/core/core/src/Environment.js
```diff
@@ -84,10 +86,12 @@ export function createEnvironment({
     engines,
     includeNodeModules,
     outputFormat,
+    sourceType: sourceType ?? 'module',
```
wbinnssmith: (along with above, assuming `null` is a desirable type for `SourceType`)

```suggestion
    sourceType,
```

packages/core/core/src/public/Environment.js
```diff
@@ -110,6 +137,10 @@ export default class Environment implements IEnvironment {
     return this.#environment.outputFormat;
   }
 
+  get sourceType(): ?SourceType {
```
wbinnssmith: is this really a maybe type? whenever `createEnvironment()` is used it always has a SourceType

packages/core/types/index.js
wbinnssmith: Yeah, this should address one of my concerns above. However, in `createEnvironment` it's currently not possible to explicitly set it to null.

packages/core/core/test/Environment.test.js
```diff
@@ -6,7 +6,7 @@ import {createEnvironment} from '../src/Environment';
 describe('Environment', () => {
   it('assigns a default environment with nothing passed', () => {
     assert.deepEqual(createEnvironment(), {
-      id: '1051b306b4fe6e64',
+      id: '2163092a3dd83a5f',
```
wbinnssmith, line 9: 😞

packages/transformers/js/src/lib.rs
mischnic: I guess it would be nice? But we could also add that later if it comes up. Was there a reason you disabled it? 
The only thing we can't support if course is the `process` Node global replacement.

packages/core/core/test/Environment.test.js
wbinnssmith: Would be neat if we used/made something along the lines of Jest's stringMatching matcher: https://jestjs.io/docs/expect#expectstringmatchingstring--regexp

packages/core/integration-tests/test/sourcemaps.js
```diff
@@ -428,7 +428,7 @@ describe('sourcemaps', function() {
       source: inputs[0],
       generated: raw,
       str: 'const local',
-      generatedStr: 'const o',
+      generatedStr: 'const t',
```
wbinnssmith, line 431: 😞

packages/packagers/js/src/CJSOutputFormat.js
```diff
@@ -148,7 +148,7 @@ export class CJSOutputFormat implements OutputFormat {
     return [res, lines];
   }
 
-  buildBundlePostlude(): string {
-    return '';
+  buildBundlePostlude(): [string, number] {
```
wbinnssmith, line 151: nit: Maybe a type alias or a comment explaining what these values in the return type mean?

packages/transformers/js/core/src/global_replacer.rs
```diff
@@ -70,6 +70,7 @@ impl<'a> Fold for GlobalReplacer<'a> {
               attributes: None,
               is_optional: false,
               is_helper: false,
+              source_type: Some(SourceType::Module),
```
mischnic, line 73: Should this be `None` as in `fs.rs`?

packages/transformers/js/src/lib.rs
devongovett: Mainly I was just trying to make scripts as close to real browser JS as possible, and not enable any CommonJS/Node features.

packages/transformers/js/core/src/global_replacer.rs
devongovett: file dependencies don't need a source type, but in this case it's referring to a require that should be treated as a module.

packages/transformers/html/src/dependencies.js
devongovett: `attrs` is referenced by `copy`, so this is modifying the attrs of the copy.

wbinnssmith (approved):
LGTM

rhendric:
Found my way here from the new "Browser scripts cannot have imports or exports" error. Why does this restriction exist, and can it be suppressed if we know that Parcel is going to bundle away all of the import statements?

devongovett:
You just need to add `type="module"` to your script tag and then it should work fine. If your browser targets defined in your browserslist don't all support modules natively, Parcel will auto generate a nomodule version as well.

rhendric:
What if I only want to produce one bundle for all browsers? If everything's being bundled together, what does it matter if the bundle is being loaded as a module or not? Seems unusually opinionated for Parcel to force there to be two essentially identical versions of the bundle; what are we getting for this?

devongovett:
The main reason is to ship less code to most browsers. [92.64%](https://caniuse.com/es6-module) of global web traffic is via a browser that supports ES modules natively. This also means that we can ship a lot more modern JS features like classes etc. without transpilation which results in reduced bundle sizes. Basically, unless you have to support IE 11 still, you will get one bundle anyway, and if you do, then most of your users will need to download less code, and only IE 11 users will need to download the larger legacy bundle.

rhendric:
Hmm, okay. I still don't see why Parcel couldn't just support script tags without the module type by doing exactly what it does to generate a nomodule bundle, or what transpilation of other JS features has to do with modules and imports/exports (which are completely eliminated in bundles regardless of what features the browser supports), but I can try targeting modules and seeing if that causes any problems for me.

devongovett:
Yeah it's because scripts behave differently in other ways as well. Whereas modules are self contained, scripts run in the global context meaning any variables created become global. This might be seen as a bad idea but there are lots of legacy libraries like jquery that rely on this assumption. The goal of Parcel is to work exactly like if you ran your code natively in a browser but also make your code portable to older browsers as well if needed. Treating scripts as modules in Parcel v1 was a mistake that we're trying to correct in v2.

rhendric:
Ah, so this is a way to support
```html
<script src="jquery.js"></script>
<script type="module" src="app.js"></script>
```
such that on old browsers, jquery.js is loaded as-is but app.js is wrapped in an IIFE or something to keep it from adding to the global scope?

devongovett:
Yeah exactly. It solves two separate problems that we've had for a long time - legacy script support, and automatic modern/legacy builds for reduced bundle sizes. It'll be a transition for everyone to get used to for sure. If you think there's an improvement we can make to the error message, we'd love to hear suggestions. Maybe we can add a link to the docs with some more info or something.

ivoba:
This is awesome!
@devongovett is this already in 2.0.0-nightly.746 ?

I have classic imports but they still seem to be wrapped as modules.

```
    <script src="/node_modules/jquery/dist/jquery.min.js"></script>
    <script src="/js/jquery.cycle.lite.js"></script>
```
which results in:

>$(...).cycle is not a function

mischnic:
> is this already in 2.0.0-nightly.746 ?

Yes

> $(...).cycle is not a function

Can you open an issue for that with a full code sample (not sure if there's more needed that you already posted)?

PR #3933, opened 2019-12-20
RFC: Scope hoisting: support bundling browserify-built bundles

wbinnssmith:
Closes https://github.com/parcel-bundler/parcel/issues/4634

This adds support for bundling bundles that have already been built using browserify. 

The browserify prelude tests for an existing global `require` using `typeof require === 'function'`, ~~which we previously statically rewrote to `'function' === 'function'`,~~ leading it to attempt to invoke a require that does not exist. Should we find a way of conditionally rewriting these, maybe only if `require` is only used as a function in a statically analyzable way?

Test Plan: Added a test using a browserify bundle as a fixture.

PR #6260, opened 2021-05-10
Allow bundling browserify bundles by replacing free requires with undefined

### LINKED ISSUE #4634: Uncaught ReferenceError: require is not defined
# 🐛 bug report

Since some nightly versions (not 100% sure when it started) I get an error for optimized JS files `Uncaught ReferenceError: require is not defined`.

It looks like it happens when a node module is requested which contains a commonjs module.

See https://github.com/garthenweb/parcel-bug-report-examples/tree/bug/require-is-not-a-function

## 🤔 Expected Behavior

Should not throw at runtime.

## 😯 Current Behavior

Throws.

## 💁 Possible Solution

I guess the required calls need to be replaced in the scope hoisting process.

## 🔦 Context

I saw this yesterday the first time in nightly version 266 with the `object-hash` library but had it today in another project using `react-hot-loader`.

## 💻 Code Sample

https://github.com/garthenweb/parcel-bug-report-examples/tree/bug/require-is-not-a-function

## 🌍 Your Environment

| Parcel           | 2.0.0-nightly.273
| Node             | v12.8.1
| npm | v6.14.5
| Operating System | Windows / WSL 1 Ubuntu

mischnic:
Closes #3933, Closes #4634, Closes T-1002

~This replaces `typeof require == "function" && require` (and commutative equivalents, or with `===`) with `undefined`, so that browserify bundles don't throw. `require` is never actually called, but still assigned to a local variable.~

~Previously, it was replaced with `"function" == "function" && require`.~

This replaces free `require`s (so everything except `typeof require` and `require("some...string")` with `undefined`.

### DIFF
--- a/packages/transformers/js/src/dependency_collector.rs
+++ b/packages/transformers/js/src/dependency_collector.rs
@@ -300,12 +300,10 @@ impl<'a> Fold for DependencyCollector<'a> {
           attributes = Some(attrs);
         }
       }
-    }
-
-    // importScripts() accepts multiple arguments. Add dependencies for each
-    // and replace with require calls for each of the specifiers (which will
-    // return the resolved URL at runtime).
-    if kind == DependencyKind::ImportScripts {
+    } else if kind == DependencyKind::ImportScripts {
+      // importScripts() accepts multiple arguments. Add dependencies for each
+      // and replace with require calls for each of the specifiers (which will
+      // return the resolved URL at runtime).
       let mut node = node.clone();
       node.args = node
         .args
@@ -375,10 +373,31 @@ impl<'a> Fold for DependencyCollector<'a> {
 
       // Track the returned require call to be replaced with a promise chain.
       self.require_node = Some(call.clone());
-      return call;
+      call
+    } else if kind == DependencyKind::Require {
+      // Don't continue traversing so that the `require` isn't replaced with undefined
+      node
+    } else {
+      node.fold_children_with(self)
+    }
+  }
+
+  fn fold_unary_expr(&mut self, node: ast::UnaryExpr) -> ast::UnaryExpr {
+    // Don't traverse `typeof require` further to `require` with undefined
+    if let ast::UnaryExpr {
+      op: ast::UnaryOp::TypeOf,
+      arg,
+      ..
+    } = &node
+    {
+      if let ast::Expr::Ident(ast::Ident { sym, .. }) = &**arg {
+        if sym == &js_word!("require") && !self.decls.contains(&(sym.clone(), node.span.ctxt())) {
+          return node;
+        }
+      }
     }
 
-    return node.fold_children_with(self);
+    node.fold_children_with(self)
   }
 
   fn fold_new_expr(&mut self, node: ast::NewExpr) -> ast::NewExpr {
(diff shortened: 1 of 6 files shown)


wbinnssmith (approved):

parcel-benchmark:
## Benchmark Results

#### Timings

| --- | --- | --- |

#### Cold Bundles

#### Cached Bundles

</p></details>

#### Timings

| --- | --- | --- |

#### Cold Bundles

#### Cached Bundles

</p></details>

#### Timings

| --- | --- | --- |

#### Cold Bundles

#### Cached Bundles

</p></details>

#### Timings

| --- | --- | --- |

#### Cold Bundles

#### Cached Bundles

</p></details>

**[Click here to view a detailed benchmark overview.](https://benchmark.parceljs.org/benchmark/298498718266556936)**

devongovett:
Would it be simpler to only look for `require` as a free identifier? Right now we don't do anything with that. If we replaced it with `undefined` then that might also fix this. I can't think of a case where a free require would work anyway.

mischnic:
It now replaces free `require`s instead. That wasn't really possible before the swc transformer because there was no `__parcel__require__` for the runtimes.

packages/transformers/js/src/dependency_collector.rs
```diff
@@ -373,10 +371,31 @@ impl<'a> Fold for DependencyCollector<'a> {
...
+    } else if kind == DependencyKind::Require {
+      // Don't continue traversing so that the `require` isn't replaced with undefined
+      node
+    } else {
+      node.fold_children_with(self)
     }
+  }
 
-    return node.fold_children_with(self);
+  fn fold_unary_expr(&mut self, node: ast::UnaryExpr) -> ast::UnaryExpr {
```
devongovett, line 385: Shouldn't `fold_expr` already take care of this? It should run before `fold_unary_expr` I think.

mischnic: `UnaryExpr` is an enum variant of `Expr`, so `Expr` is visited before the `UnaryExpr`.

Do you mean I should move the logic from `fold_unary_expr` to `fold_expr`?

devongovett: I mean, we already handle `UnaryExpr` inside the `fold_expr` visitor, and return without visiting children, so I'm not sure why this additional visitor is needed.

mischnic: > I mean, we already handle UnaryExpr inside the fold_expr visitor

I think you are looking at hoist.rs? All of my current changes are in dependency_collector.rs now

devongovett: Oh right 🙈

devongovett (approved):

packages/transformers/js/src/dependency_collector.rs
```diff
@@ -375,10 +373,31 @@ impl<'a> Fold for DependencyCollector<'a> {
...
+    {
+      if let ast::Expr::Ident(ast::Ident { sym, .. }) = &**arg {
+        if sym == &js_word!("require") && !self.decls.contains(&(sym.clone(), node.span.ctxt())) {
+          return node;
+        }
+      }
+    }
+
+    node.fold_children_with(self)
   }
```
kdy1, line 402: Seems like `fold_member_expr` is missing.
All passes touching identifier should implement it and fold `prop` only if `computed` is true.

kdy1: Seems like a cause of https://github.com/parcel-bundler/parcel/issues/6323

kdy1: I was wrong. Sorry for the mistake.

PR #6117, opened 2021-04-12
Not mutate original PostCSS config file.

hydrognomik:
# ↪️ Pull Request

Related to #6094.

## ✔️ PR Todo

- [ ] Added/updated unit tests for this change
- [ ] Filled out test instructions (In case there aren't any unit tests)
- [x] Included links to related issues/PRs

PR #6266, opened 2021-05-12
Prevents function from being called twice

n370:
The PostCSS transformer is calling this function twice and that is causing build and serve to fail intermittently (Once a successful pass is cached it always work for the same command). I know this is not the right solution but it might give y'all a hint on what is going on. 

During build or serve this function ends up being called twice and at the second pass the config object shows up modified, missing the original plugins property value.

I tried to mess up with this a little bit but didn't insisted too much as I don't have my local environment configured with the proper tooling to debug child processes.

I noticed that commenting out https://github.com/parcel-bundler/parcel/blob/a3aa86815e098c90fda72a08d29646ff059ab74d/packages/transformers/postcss/src/loadConfig.js#L42 doesn't fix the issue but breaks the code path somewhere else.

Sorry for the messy bug report. Let me know if I can help in any way.

# ↪️ Pull Request

## ✔️ PR Todo

### DIFF
--- a/packages/transformers/postcss/src/loadConfig.js
+++ b/packages/transformers/postcss/src/loadConfig.js
@@ -80,6 +80,8 @@ async function configHydrator(
   });
 }
 
+let count = 0;
+
 export async function load({
   config,
   options,
@@ -89,6 +91,7 @@ export async function load({
   options: PluginOptions,
   logger: PluginLogger,
 |}): Promise<void> {
+  if (count > 0) return;
   let configFile: any = await config.getConfig(
     ['.postcssrc', '.postcssrc.json', '.postcssrc.js', 'postcss.config.js'],
     {packageKey: 'postcss'},


devongovett:
Can you give us a bit more detail about what the problem was? Is there a way we can reproduce the issue?

n370:
> Can you give us a bit more detail about what the problem was? Is there a way we can reproduce the issue?

Hi @devongovett Thanks for your help. Yes I can. 

I have a GH action that is reproducing the issue at https://github.com/n370/rolltip/runs/2563636326?check_suite_focus=true

And here is the codebase:

https://github.com/n370/rolltip/tree/f06fc9233015e5026c3f067319c62b8f1e441c6d/demo

GH environment is ubuntu linux running

node v14.17.0
npm v7.12.1

I first noticed the issue in my mac notebook

node 12.22.1
npm 7.11.1

mischnic:
I think this is already fixed by https://github.com/parcel-bundler/parcel/pull/6117.

Try it with a nightly, the PR was merged after the beta.2

n370:
> I think this is already fixed by #6117.

@mischnic It is. Thanks for jumping in.

In my investigation I tried cloning at https://github.com/parcel-bundler/parcel/blob/a3aa86815e098c90fda72a08d29646ff059ab74d/packages/transformers/postcss/src/loadConfig.js#L41

But it didn't worked. I knew this was some issue with mutable references but couldn't quite spot where the change was needed.

PR #6286, opened 2021-05-15
Simplify source content handling

devongovett:
Goes into #6238. Opened a separate PR for feedback. Basically removed `asset.getSourcesContent` and set the source content in core.

Also I think we don't need to extract the source content from existing source maps because `map.extends` should handle copying over the correct source content from the original map. We only need to set the source content if there was no original map. Could be wrong though. Maybe we should add an integration test for this.


parcel-benchmark:
## Benchmark Results

#### Timings

| --- | --- | --- |

#### Cold Bundles

#### Cached Bundles

</p></details>

#### Timings

| --- | --- | --- |

#### Cold Bundles

#### Cached Bundles

</p></details>

#### Timings

| --- | --- | --- |

#### Cold Bundles

#### Cached Bundles

</p></details>

#### Timings

| --- | --- | --- |

#### Cold Bundles

#### Cached Bundles

</p></details>

**[Click here to view a detailed benchmark overview.](https://benchmark.parceljs.org/benchmark/298693636730651144)**

packages/transformers/babel/src/BabelTransformer.js
```diff
@@ -64,16 +64,6 @@ export default (new Transformer({
...
-    if (sourcesContent) {
-      for (let sourcesContentFileName of Object.keys(sourcesContent)) {
-        map.setSourceContent(
-          sourcesContentFileName,
-          sourcesContent[sourcesContentFileName],
-        );
-      }
-    }
-
     if (originalSourceMap) {
```
DeMoorJasper, line 67: `map.extends` should be removed in this transformer... this is probably a merge conflict that got resolved incorrectly. The AST should contain the mappings and the source contents should get set as Babels AST has no way to contain these

packages/shared/babel-ast-utils/src/index.js
```diff
@@ -156,16 +156,6 @@ export async function generate({
     options,
   });
 
-  let sourcesContent = await asset.getSourcesContent();
```
DeMoorJasper, line 159: This is still necessary as Babel does not embed this in it's AST

packages/core/core/src/UncommittedAsset.js
```diff
@@ -62,7 +60,7 @@ export default class UncommittedAsset {
   options: ParcelOptions;
   content: ?(Blob | Promise<Buffer>);
   mapBuffer: ?Buffer;
-  sourcesContent: ?SourcesContentDictionary;
```
DeMoorJasper, line 65: Why did this change to just one source content? Compiled modules can have multiple sources?

DeMoorJasper (changes requested):

packages/transformers/babel/src/BabelTransformer.js
devongovett: Oh I thought extends would take care of setting all the source content?

packages/core/core/src/UncommittedAsset.js
devongovett: Extends should take care of setting source content when there is a previous source map right?

packages/transformers/babel/src/BabelTransformer.js
DeMoorJasper: It does but it also remaps mappings which isn't necessary in this case as the babel ast should already contain the original mappings. So we just need to set source contents

packages/core/core/src/UncommittedAsset.js
DeMoorJasper: In case of babel it's part of the ast so no extends should be called as it would also remap all mappings as mentioned in other reply

packages/transformers/babel/src/BabelTransformer.js
devongovett: Ah. Ok well maybe we can still avoid an extra method in core by copying over the source contents from the original map here rather than extracting and storing it in core even when unnecessary.

devongovett: Do we have an integration test that covers this btw?

packages/core/core/src/UncommittedAsset.js
devongovett: Ok. So now `sourceContent` is only used when there is no previous source map. In that case, it is set by Transformation.js to the original source content, and used by `setMap` below. If there is an original source map, the transformer will either call `extends` on it, or in the case of babel, apply the source map to the AST and copy the original source contents over in `generate`. Does that make sense?

DeMoorJasper: Kind of, the only issue here is that the map will be erased whenever `commit` is called in `UncommittedAsset.js` so those source contents get lost pretty quickly if we ever re-use the AST... which is maybe no longer obvious after the JS Transformer rewrite but still a bug...

DeMoorJasper: This was how I originally fixed this issue, before I saw this line: https://github.com/parcel-bundler/parcel/blob/v2/packages/core/core/src/UncommittedAsset.js#L102

packages/transformers/babel/src/BabelTransformer.js
DeMoorJasper: Yes, Will added a test for it, thought I also had one in the original PR but apparently not

packages/core/core/src/UncommittedAsset.js
devongovett: Oh yeah I guess it would only be an issue if the packager uses an AST but that's not the case anymore. Kinda wondering if we should just remove support for that.

DeMoorJasper: Sounds like a good idea to remove support for it. Would probably cause issues at some point otherwise

DeMoorJasper (approved):

PR #6294, opened 2021-05-16
fix parcel CLI passed config

Magi-KS:
- parcel CLI config option was NOT used before, updated cli.js so that
the passed config is actually passed to parcel.

`parcel build --config .myparcelrc` does not work because config was not passed when initializing parcel.

notable line in file(assisting review)
` packages/core/parcel/src/cli.js:218`

PR that added the CLI option:
https://github.com/parcel-bundler/parcel/pull/4927


DeMoorJasper (approved):

Magi-KS:
would like this to be merged if there's no objections.

PR #6298, opened 2021-05-17
fix github action benchmark rustup not found

Magi-KS:
GitHub Action benchmark is failing, this PR is attempt at fixing it.

turns out GitHub Action executes shell script in non-interactive mode, so .bashrc is not loaded by default.
https://github.community/t/self-hosted-not-using-bashrc/18358

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -6,9 +6,11 @@ jobs:
   prBenchmarks:
     name: PR Benchmarks
     runs-on: ubuntu-latest
+    defaults:
+      run:
+        shell: bash -leo pipefail {0}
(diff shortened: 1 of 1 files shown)


DeMoorJasper (approved):

DeMoorJasper (changes requested):
approved this a bit too quickly, doesn't fix the benchmark

Magi-KS:
changes here are not needed.
fixes over at https://github.com/parcel-bundler/parcel-benchmark-action/pull/37

DeMoorJasper:
Thanks for looking into this and fixing the benchmark

PR #6372, opened 2021-05-29
build-native.js tweaked xcodeDir path to toolchain

### LINKED ISSUE #6316: `yarn build-native` throws error
# 🐛 bug report

On running `yarn build-native` in the parcel monorepo, I get an error stating

`error: failed to run custom build command for `ring v0.16.20``

## 🤔 Expected Behavior

This should work as expected and stated in https://github.com/shawsourav/parcel/blob/v2/CONTRIBUTING.md

## 😯 Current Behavior

Build fails

## 💁 Possible Solution

Not sure

## 🔦 Context

I was going through the Contributing page and this command `yarn build-native` didn't work.

## 🌍 Your Environment

| Parcel           | v2
| Node             | 12.20.0
| npm/Yarn         | 6.14.0
| Operating System | Mac OS Catalina v10.15.7

LironHazan:
Following PR fixes #6316
Doing so will resolve the path to clang and clang++ which fixed the "ring" dependency compilation error 

```
  process.env.CC = `${xcodeDir}/usr/bin/clang`;
  process.env.CXX = `${xcodeDir}/usr/bin/clang++`;
```

Afterwards build-native will be ended successfully 

```
 Compiling ring v0.16.20
   Compiling jemalloc-sys v0.3.2
........
```


mischnic:
>  process.env.CC = `${xcodeDir}/usr/bin/clang`;

That file does not exist on my machine, and I get a build error.

Can you try to execute `xcrun -f clang`? That correctly returns `/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang` for me, maybe we can use the result from that for `process.env.CC`

LironHazan:
@mischnic yes you're right that would be much better,
my output btw (as suspected) 
`/Library/Developer/CommandLineTools/usr/bin/clang`

scripts/build-native.js
```diff
@@ -12,7 +12,7 @@ async function build() {
     setupMacBuild();
   }
 
-  let packages = glob.sync('packages/*/*')
+  let packages = glob.sync('packages/*/*');
```
LironHazan, line 15: @mischnic now I see that my IDE's global linter definition kicked in :) I can revert that if you want

scripts/build-native.js
```diff
@@ -43,9 +43,13 @@ async function build() {
...
-  process.env.CC = `${xcodeDir}/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang`;
-  process.env.CXX = `${xcodeDir}/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang++`;
+  let xcodeClangPath = execSync('xcrun -f clang', {
+    encoding: 'utf8',
+  }).trim();
+  let sysRoot = execSync('xcrun --sdk macosx --show-sdk-path', {
+    encoding: 'utf8',
+  }).trim();
+  process.env.CC = xcodeClangPath;
+  process.env.CXX = `${xcodeClangPath}++`;
```
LironHazan: @mischnic do you find this line a bit flaky? it works

mischnic: I think we should also call `xcrun -f clang++` for that. This kind of assumption is exactly why it was broken before 😄

scripts/build-native.js
mischnic: It's fine

scripts/build-native.js
LironHazan: @mischnic lol the cost of "being DRY" :) done

LironHazan: @mischnic the first 2 workflows failed, I think re-triggering them will resolve it (there's no sense in failing for ts type validation issues as the change is on a js file :) + I ran the tests locally just to be 100% sure there's nothing I'm missing.

mischnic: 👍  Thanks

You can ignore the failures, CI has been flakey for some time now.

mischnic (approved):
Still works correctly for me.

@wbinnssmith and/or @gorakong were also having the same issue you had with the CLT

PR #6373, opened 2021-05-29
Switch to xxhash3 for improved hash performance

devongovett:
T-961

Parcel relies heavily on hashing for many things, including cache keys, content hashes, etc. We've used md5 for this so far since it is a built in hash function in Node.js. However, it shows up pretty high in profiles of overall performance. This PR replaces md5 with [xxhash](http://cyan4973.github.io/xxHash/), which is one of the fastest hash functions out there.

Specifically, this is an napi wrapper around [this Rust implementation](https://github.com/DoumanAsh/xxhash-rust) of xxhash. There were some existing npm modules available for this, but most were either unmaintained or only accepted Buffers and not strings. This offered some performance increase, but the overhead of converting between strings and buffers was too high.

This benchmark shows the difference between md5, xxhash-addon (an existing npm module that only accepts buffers), and xxhash-rust.

```
md5 209,999.15 opts/sec (mean: 0.005ms, stddev: 0.029ms, 500 samples)
xxhash-addon 415,511.89 opts/sec (mean: 0.002ms, stddev: 0.003ms, 500 samples)
xxhash-rust 1,517,985.09 opts/sec (mean: 658.768ns, stddev: 996.69ns, 500 samples)
```

Overall this improves performance on the esbuild benchmark by ~5%.


flow-typed/npm/hash_v2.x.x.js
```diff
@@ -0,0 +1,11 @@
+// @flow
```
devongovett: Is there a way to put these declarations inside the actual package without creating a wrapper around the native module?

mischnic: Move this file to `packages/utils/hash/index.js.flow` and remove the `declare module` wrapper

devongovett:
One thing to consider: xxhash3 is a 64-bit hash, whereas md5 is a 128-bit hash. There is also a 128-bit version of xxhash that we could consider, though it is slightly slower. I'm not sure if this matters in practice. A 64-bit hash should probably give us a low enough probability of collision for our uses I think.

packages/core/core/src/requests/AssetRequest.js
```diff
@@ -47,16 +48,16 @@ const type = 'asset_request';
...
-    query: input.query ? objectSortedEntries(input.query) : null,
-  });
+  return hashString(
+    type +
+      input.filePath +
+      input.env.id +
+      String(input.isSource) +
+      String(input.sideEffects) +
+      (input.code ?? '') +
+      (input.pipeline ?? '') +
```
mischnic: `code = "", pipeline = "foo"` and `code="foo", "pipeline = ""` would produce the same hash (same goes for `query`)

One solution would be to join them with `:` or something like that

Or a utility `hashFromStrings(...args) { return hashFromString(args.join(":")); }` ?

packages/core/core/src/assetUtils.js
```diff
@@ -73,16 +68,18 @@ type AssetOptions = {|
...
-    pipeline: options.pipeline ?? '',
-    queryString,
-  });
+  return hashString(
+    idBase +
+      options.type +
+      options.env.id +
+      uniqueKey +
+      (options.pipeline ?? '') +
+      queryString,
```
mischnic: Same problem with collisions

packages/core/core/src/requests/AssetRequest.js
devongovett: Good point.

packages/core/core/src/UncommittedAsset.js
```diff
@@ -129,12 +129,10 @@ export default class UncommittedAsset {
...
-      [
-        this.value.hash,
-        pipelineKey,
-        await getInvalidationHash(this.getInvalidations(), this.options),
-      ].join(':'),
+    this.value.outputHash = hashString(
+      (this.value.hash ?? '') +
+        pipelineKey +
+        (await getInvalidationHash(this.getInvalidations(), this.options)),
     );
```
mischnic, line 136: ...

parcel-benchmark:
## Benchmark Results

#### Timings

| --- | --- | --- |
| Cold | 1.58s | -90.00ms 🚀 |
| Cached | 392.00ms | -1.00ms |

#### Cold Bundles

*No bundle changes detected.*

#### Cached Bundles

| dist/modern/parcel.d5807e82.webp | 102.94kb | +0.00b | 74.00ms | +7.00ms ⚠️ |
| dist/legacy/index.html | 701.00b | +0.00b | 76.00ms | +4.00ms ⚠️ |

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 10.35s | -266.00ms |
| Cached | 675.00ms | +18.00ms |

#### Cold Bundles

| dist/index.js | 491.15kb | -98.00b 🚀 | 5.75s | -1.00ms |

#### Cached Bundles

| dist/index.js | 491.15kb | -98.00b 🚀 | 97.00ms | +10.00ms ⚠️ |
| dist/NotFound.c3254f12.js | 417.00b | +0.00b | 96.00ms | +9.00ms ⚠️ |

</p></details>

#### Timings

| --- | --- | --- |

#### Cold Bundles

#### Cached Bundles

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 7.24s | -261.00ms |
| Cached | 569.00ms | -30.00ms 🚀 |

#### Cold Bundles

*No bundle changes detected.*

#### Cached Bundles

| dist/Three.js | 579.69kb | +0.00b | 71.00ms | -22.00ms 🚀 |

</p></details>

**[Click here to view a detailed benchmark overview.](https://benchmark.parceljs.org/benchmark/300420267431166477)**

packages/core/core/src/UncommittedAsset.js
devongovett: This one should be ok? There will always be a pipelineKey and invalidation hash.

packages/core/core/src/Transformation.js
```diff
@@ -333,29 +329,26 @@ export default class Transformation {
...
     for (let transformer of pipeline.transformers) {
       let key = `${transformer.name}:${transformer.resolveFrom}`;
-      hash.update(
+      hashes +=
         this.request.devDeps.get(key) ??
-          this.devDepRequests.get(key)?.hash ??
-          '',
-      );
+        this.devDepRequests.get(key)?.hash ??
+        '';
```
mischnic: In theory, the same thing could happen here?

packages/core/core/src/Dependency.js
```diff
@@ -33,13 +33,13 @@ type DependencyOpts = {|
...
-      env: opts.env.id,
-      target: opts.target,
-      pipeline: opts.pipeline,
-    });
+    hashString(
+      (opts.sourceAssetId ?? '') +
+        opts.moduleSpecifier +
+        opts.env.id +
+        (opts.target ? JSON.stringify(opts.target) : '') +
+        (opts.pipeline ?? ''),
```
mischnic, line 41: target and/or pipeline can be unset here

devongovett: but target will be a JSON object and it's very unlikely that pipeline would look like that

packages/core/core/src/Transformation.js
devongovett: There will always be at least on dev dep - the plugin itself. I think the `?? ''` is just there to appease flow.

mischnic: Should it be `nullthrows` instead of `??` then?

devongovett: Actually seems I was wrong. It will be null on initial builds.

DeMoorJasper (approved):

mischnic (approved):

mischnic:
Can you leave the benchmark script in there? [`d003549` (#6373)](https://github.com/parcel-bundler/parcel/pull/6373/commits/d0035492bf8fb9da61d08a4f6b721fba852d3321#diff-89d72a5991f686f52f33d685f11aa52bea711090ec139b133adfc08cf43e3841)

Might be interesting for the future or the wasm version

wbinnssmith (approved):

PR #6404, opened 2021-06-04
added cio headers

### LINKED ISSUE #6220: "SharedArrayBuffer will require cross-origin isolation" with the dev server
# 🐛 bug report

Using `SharedArrayBuffer` will require/already requires setting some headers on the server:

https://developer.chrome.com/blog/enabling-shared-array-buffer/

## 💁  Solution

We already set the "normal" CORS headers:
https://github.com/parcel-bundler/parcel/blob/a67a5f02a7a4f6fdf9e849847c1a7d86994bcd64/packages/reporters/dev-server/src/Server.js#L33-L41

The dev server should set the headers described on the linked Chrome page, probably these:
```
Cross-Origin-Embedder-Policy: require-corp
Cross-Origin-Opener-Policy: same-origin
Cross-Origin-Resource-Policy: same-origin
```

The question is whether we can just always set them or if this can cause problems

## 🌍 Your Environment

| Parcel           | v2

LironHazan:
Following PR is corollated to: "SharedArrayBuffer will require cross-origin isolation" with the dev server #6220
1. Fetched from upstream + yarn install
2. Added the cross-origin isolation headers
3. Ran integration tests (https://github.com/parcel-bundler/parcel/blob/v2/packages/core/integration-tests/test/server.js)
first time before the fix for making sure all really pass and second time after the fix - **all is green** 

Screen Shot 2021-06-04 at 8 50 28

@mischnic LMK if there's another thing needed to be checked 

Closes #6220


LironHazan:
actually after I pulled updates from v2 and ran yarn install + yarn test I see many failures, I'll checkout to v2 for running again but it seems that the tests are broken

mischnic:
The tests are passing on CI, you probably didn't recompile the Rust modules

packages/reporters/dev-server/src/Server.js
```diff
@@ -40,6 +40,9 @@ function setHeaders(res: Response) {
     'Access-Control-Allow-Headers',
     'Origin, X-Requested-With, Content-Type, Accept, Content-Type',
   );
+  res.setHeader('Cross-origin-Opener-Policy', 'same-origin');
```
mischnic:
```suggestion
  res.setHeader('Cross-Origin-Opener-Policy', 'same-origin');
```

LironHazan: :)

DeMoorJasper (approved):

packages/reporters/dev-server/src/Server.js
```diff
@@ -40,6 +40,9 @@ function setHeaders(res: Response) {
     'Access-Control-Allow-Headers',
     'Origin, X-Requested-With, Content-Type, Accept, Content-Type',
   );
+  res.setHeader('Cross-Origin-Opener-Policy', 'same-origin');
+  res.setHeader('Cross-Origin-Resource-Policy', 'same-origin');
+  res.setHeader('Cross-Origin-Embedded-Policy', 'require-corp');
```
mischnic:
```suggestion
  res.setHeader('Cross-Origin-Opener-Policy', 'same-origin');
  res.setHeader('Cross-Origin-Embedder-Policy', 'require-corp');
```

1. It's called `Cross-Origin-Embedder-Policy `: https://developer.chrome.com/blog/enabling-shared-array-buffer/#cross-origin-isolation
2. As far as I can tell, `Cross-Origin-Resource-Policy` isn't really needed here

LironHazan: Fixed the typo,  good catch again,  I'm sorry for causing many iterations from your side for a task that should've been a quick win :|

mischnic (approved):

bminer:
Can override this COEP header by [doing this](https://github.com/parcel-bundler/parcel/issues/6220#issuecomment-885837128)

PR #6472, opened 2021-06-16
Fix TypeScript and other sourcemaps by always creating an initial sourcemap

### LINKED ISSUE #6191: Adding .babelrc breaks browserslist in serve
# 🐛 bug report

I have a project with no `.babelrc` file that builds and runs fine, both in `parcel serve` and `parcel build`.  When I add a `.babelrc` file, `parcel build` behaves as before but `parcel serve` produces a bundle that needs regeneratorRuntime and has downlevel transpiled generator functions, even though I did not ask for it.

browserslist in `package.json`:
```json
  "browserslist": {
    "production": [
      "last 3 chrome version",
      "last 2 firefox version",
      "last 1 safari version",
      "firefox esr"
    ],
    "development": [
      "last 1 chrome version",
      "last 1 firefox version",
      "last 1 safari version"
    ]
  }
```

`.babelrc` (only in the the broken build; the working build doesn't have this file at all):
```json
{
	"presets": [
		"@parcel/babel-preset-env",
		"@babel/preset-react"
	],
	"plugins": [
		[
			"@babel/plugin-transform-typescript",
			{ "isTSX": true }
		]
	]
}
```

Cli commands:
```json
    "start": "parcel serve ./public/index.html --dist-dir .parcel-dev --public-url / --port 3000",
    "build": "tsc && rimraf ./build && parcel build ./public/index.html --public-url . --dist-dir build --reporter @parcel/reporter-bundle-analyzer --no-scope-hoist",
```

## 🤔 Expected Behavior

Both versions should produce identical output.

## 😯 Current Behavior

In the working version:
`npm start` works
`npm run build` produces a working static build

In the broken version:
`npm start` produces a build that fails at runtime missing regeneratorRuntime
`npm run build` produces a working static build

## 🔦 Context

I want to make some changes to .babelrc, but to do that first I need to have a .babelrc, so my first step is to create a .babelrc that makes the application build and behave just like before, but I am unable to do that.

## 💻 Code Sample

This is full code to the entire project.  If you'd like to test it out, `npm install && npm start` should be sufficient; then see what is being served up at http://localhost:3000
The working version: https://github.com/nattthebear/ff12characterplanner/tree/babelrc-without
The broken version: https://github.com/nattthebear/ff12characterplanner/tree/babelrc-with

## 🌍 Your Environment

| Parcel           | 2.0.0-nightly.643
| Node             | 14.16.0
| npm/Yarn         | npm 6.14.11
| Operating System | Windows 10 Home

### LINKED ISSUE #6457: Sourcemaps only showing compiled code after upgrade to v2
# 🐛 bug report

After upgrading to Parcel v2, sourcemaps are showing compiled code.

babel.config.json

```js
{
  "presets": [
    "@babel/preset-env",
    "@babel/preset-react",
    "@babel/preset-typescript"
  ]
}

```

package.json
```json
"scripts": {
    "build": "parcel build public/index.html --no-scope-hoist",
    "dev": "parcel public/index.html",
```

## 🤔 Expected Behavior

Expecting to see sourcemaps, like when using parcel v1

## 😯 Current Behavior

Only seeing compiled code sourcemaps in `__parcel_source_root`
The end of the index.js file shows `//# sourceMappingURL=index.ed731f93.js.map`

## 🌍 Your Environment

| Parcel           | 2.0.0-beta.3 and 2.0.0-beta.3.1
| Node             | v15.14.0 and v16
| npm/Yarn         |  7.13.0
| Operating System | Mac OS Big Sur

gorakong:
Fixes #6191. Fixes #6457.

# ↪️ Pull Request

Unconditionally create an initial sourcemap for assets. This should fix issues with sourcemaps showing transpiled code instead of source contents in typescript (and other languages not included in the set) projects.


parcel-benchmark:
## Benchmark Results

#### Timings

| --- | --- | --- |

#### Cold Bundles

#### Cached Bundles

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 9.67s | +90.00ms |
| Cached | 435.00ms | -28.00ms 🚀 |

#### Cold Bundles

*No bundle changes detected.*

#### Cached Bundles

*No bundle changes detected.*

</p></details>

#### Timings

| --- | --- | --- |

#### Cold Bundles

#### Cached Bundles

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 6.83s | +77.00ms |
| Cached | 373.00ms | -17.00ms |

#### Cold Bundles

*No bundle changes detected.*

#### Cached Bundles

*No bundle changes detected.*

</p></details>

**[Click here to view a detailed benchmark overview.](https://benchmark.parceljs.org/benchmark/302675243581833741)**

yarn.lock
```diff
@@ -16,6 +16,13 @@
   dependencies:
     "@babel/highlight" "^7.12.13"
 
+"@babel/code-frame@^7.14.5":
```
mischnic, line 19: You didn't change any package.json, was this intentional?

devongovett:
Do we want to do it this way or move the assignment to asset.sourceContent out of this if statement? This would also be a problem for other languages like coffeescript. I think the original purpose of the extensions list was for loading source map comments. What do you think, @DeMoorJasper?

DeMoorJasper:
I think the other solution is way better a ts file will never have a sourcemap comment so there's no point in reading it? Probably just wastes processing time

joeyslater:
Is the test failing because we don't have the preset? 
> Cannot find module '@babel/preset-typescript' Require stack: - 
https://dev.azure.com/devongovett/devongovett/_build/results?buildId=16995&view=ms.vss-test-web.build-test-results-tab&runId=1393488&resultId=100976&paneView=debug

https://github.com/parcel-bundler/parcel/blob/daece49d003ba804bbdaa3a7ed3d6aaf446f166d/packages/dev/babel-preset/package.json

I notice it has the flow, react, etc? I also noticed the typescript wasn't listed in the yarn.lock

mischnic:
They should be added here:

https://github.com/parcel-bundler/parcel/blob/3492ad0449b7a7143e165d8b0c11d78f6672266a/packages/core/integration-tests/package.json#L15-L20

it's more of a coincidence that the flow and react presets area already listed in some other monorepo packages

DeMoorJasper (approved):

packages/core/core/src/Transformation.js
```diff
@@ -140,13 +140,13 @@ export default class Transformation {
...
-        // If no existing sourcemap was found, initialize asset.sourceContent
-        // with the original contents. This will be used when the transformer
-        // calls setMap to ensure the source content is in the sourcemap.
-        asset.sourceContent = await asset.getCode();
-      }
+    if (existing == null) {
+      // If no existing sourcemap was found, initialize asset.sourceContent
+      // with the original contents. This will be used when the transformer
+      // calls setMap to ensure the source content is in the sourcemap.
+      asset.sourceContent = await asset.getCode();
```
devongovett, line 155: Will this end up creating a string representation of binary content? Maybe we can do this on demand when calling setMap?

wbinnssmith (approved):

wbinnssmith:
Approved, pending https://github.com/parcel-bundler/parcel/pull/6472#discussion_r654006905

packages/core/core/src/Transformation.js
wbinnssmith: This would be an issue as a common pattern is `let map = asset.getMap(); myMap.extend(map); asset.setMap(map);`, but by the time we lazily initialize a map inside setMap, getMap would already be called and it returns empty then.

What about doing this lazily in getMap instead?

devongovett: Well we only want to do it for the first transformer in a pipeline

wbinnssmith: Isn't that a separate problem from getMap/setMap?

gorakong: Not sure if I'm missing something here, but wouldn't `asset.getCode()` return transpiled code by the time we get to `set/getMap`?

devongovett (approved):

packages/core/core/src/Transformation.js
```diff
@@ -146,7 +146,7 @@ export default class Transformation {
       // If no existing sourcemap was found, initialize asset.sourceContent
       // with the original contents. This will be used when the transformer
       // calls setMap to ensure the source content is in the sourcemap.
-      asset.sourceContent = await asset.getCode();
+      asset.sourceContent = await asset.getBuffer();
```
wbinnssmith: Does this mean that assets with (probably large) stream content will be read into memory at once and stored on the asset cache entry?

wbinnssmith:
~~Just pushed a change that changes asset's `getStream` api to always return a new copy of a stream and `setStream` to accept a function that returns a stream.~~ This is now #6516

However, a couple issues remain, and can probably be addressed outside this PR:
* We're still inlining content even though the user may not has asked for it. We should be checking `options.sourceMap.inlineSources` and not reading these in for most cases (aside from runtimes)
* Large JS sources read as streams due to their size are buffered entirely into memory to read them as sourceContents. While this could probably be addressed by allowing the sourcemaps library to read and write streams (cc @DeMoorJasper), the JSTransformer still needs to work with the entire buffer (and probably will continue to unless we develop a streaming JS transformer in the future), see: https://github.com/parcel-bundler/parcel/pull/6516#issuecomment-868896437

cc @devongovett @gorakong

packages/core/integration-tests/test/sourcemaps.js
```diff
@@ -1340,4 +1341,99 @@ describe('sourcemaps', function() {
...
+    checkSourceMapping({
+      map: sourceMap,
+      source: sourceContent,
+      generated: raw,
+      str: 'document.getElementById(',
+      sourcePath,
+    });
+  });
+
+  it('retains sourcesContent from the original sources from a large text file', async () => {
```
wbinnssmith, line 1390: This passed before and I'm not sure adds much value, but it verifies the fact that even though streams are not read to create sourceContent, this works since the sourcemaps code always buffers streams first, see: https://github.com/parcel-bundler/parcel/pull/6516#issuecomment-868896437

DeMoorJasper (approved):

PR #5297, opened 2020-10-29
Add TypeScript types

101arrowz:
# ↪️ Pull Request

Adds TypeScript typings to (as far as I can tell) Parcel's entire public API. I haven't checked the validity of these since a few weeks ago; any corrections are welcome. Of note is that this addition documents much of `@parcel/core` with rich comments. These additions will help TypeScript users (and JS users in most IDEs) to have much better autocomplete and more clarity with types. In addition, the types massively facilitate the creation of a Parcel plugin.

FIxes #3912.

## 💻 Examples

```ts
import { Transformer } from '@parcel/plugin'; // well typed
// Generic type allows for well-defined AST and config types
export default new Transformer({
  // the method and its parameters have full typing
  parse({ asset }) {
    return whatever;
  },
  transform({ asset, config }) {
    // No more typing out long names and hoping to catch typos
    // Since most IDEs don't support Flow, TS types help write correct code faster
    asset.addIncludedFile(...);
  }
})
```
## 🚨 Test instructions

N/A for the most part (maybe add a linter in the future?)

## ✔️ PR Todo

- [x] <strike>Added/updated unit tests for this change</strike> N/A
- [x] Filled out test instructions (In case there aren't any unit tests)
- [x] Included links to related issues/PRs

PR #6484, opened 2021-06-20
Add TypeScript definitions for public packages

### LINKED ISSUE #3912: [Parcel 2] Typescript Definitions
# 🙋 feature request

Type definitions for this package, whether they'd be included in the package or in @types/parcel.

## 🤔 Expected Behavior

Import types from `@types/parcel` :
```
import Parcel from 'parcel'
```

## 😯 Current Behavior
Import has squiggly lines of death explaining that there are no types.

`@types/parcel` exists:
```
import Parcel from 'parcel'
```

`@types/parcel-bundler` exists:
```
import Parcel from 'parcel-bundler'
```

## 💁 Possible Solution

Generate types using [`flow-to-typescript` package](https://github.com/bcherny/flow-to-typescript). Unsure on any restrictions in the project.

## 🔦 Context

I want to test out this package, but will be using typescript strictly. I'm using the bundler API to prerender static HTML pages for a react application.

devongovett:
Closes #5297. Fixes #3912.

This adds TypeScript definition files for public packages including `@parcel/core`, `@parcel/types`, and `@parcel/plugin` along with all dependencies. This will make it easier for people to develop plugins and use the Parcel API in TypeScript (or even just for autocomplete in VSCode).

I tried to use [flow-to-ts](https://github.com/Khan/flow-to-ts) on these packages, and then run `tsc` to generate `.d.ts` files, but ran into some issues because the TS type system and Flow type system work significantly differently and I was getting lots of errors when TS checked the converted code.

Instead, I opted for a hybrid approach. I used flow-to-ts to convert the type-only files (e.g. `@parcel/types`), but wrote some hand written types for some packages containing code. This way, the bulk of changes will be handled automatically and kept in sync with the Flow types, but we don't need to worry about passing two type systems on all source code.


packages/core/diagnostic/src/diagnostic.js
```diff
@@ -1,5 +1,4 @@
 // @flow strict-local
-import type {FilePath} from '@parcel/types';
```
devongovett, line 2: Avoiding circular dependency... Perhaps we should consider whether the diagnostic types should move into the core types package and only the implementation should stay in this package?

packages/core/fs/src/MemoryFS.js
```diff
@@ -195,11 +195,11 @@ export class MemoryFS implements FileSystem {
   }
 
   // eslint-disable-next-line require-await
-  async readFile(filePath: FilePath, encoding?: buffer$Encoding): Promise<any> {
+  async readFile(filePath: FilePath, encoding?: Encoding): Promise<any> {
```
devongovett, line 198: We can't rely on these global flow types

packages/core/plugin/src/PluginAPI.d.ts
```diff
@@ -0,0 +1,47 @@
...
+  Bundler as BundlerOpts,
+  Namer as NamerOpts,
+  Runtime as RuntimeOpts,
+  Packager as PackagerOpts,
+  Optimizer as OptimizerOpts,
+  Reporter as ReporterOpts,
+  Validator as ValidatorOpts,
+} from '@parcel/types';
+  
+export declare class Transformer<T> {
```
devongovett, line 13: Did this file manually because flow allows the type parameter on the constructor rather than the class but TS doesn't. Wanted to avoid the type param on the class in Flow otherwise you have to explicitly type the config on every plugin rather than it being inferred. This is because flow doesn't infer the types of exports in type first mode and forces you to type it explicitly (which is really annoying).

packages/core/types/index.js
```diff
@@ -875,6 +878,36 @@ export type TransformerResult = {|
 
 export type Async<T> = T | Promise<T>;
 
+export interface PluginLogger {
```
devongovett, line 905: Moved this type here to avoid another circular dependency. Plus, better to expose an interface to plugins rather than an implementation.

parcel-benchmark:
## Benchmark Results

#### Timings

| --- | --- | --- |

#### Cold Bundles

#### Cached Bundles

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 10.56s | +112.00ms |
| Cached | 458.00ms | -15.00ms |

#### Cold Bundles

*No bundle changes detected.*

#### Cached Bundles

*No bundle changes detected.*

</p></details>

#### Timings

| --- | --- | --- |

#### Cold Bundles

#### Cached Bundles

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 7.85s | -86.00ms |
| Cached | 397.00ms | -30.00ms 🚀 |

#### Cold Bundles

*No bundle changes detected.*

#### Cached Bundles

| dist/Three.js | 579.69kb | +0.00b | 5.77s | -404.00ms 🚀 |

</p></details>

**[Click here to view a detailed benchmark overview.](https://benchmark.parceljs.org/benchmark/302439871186731533)**

zxbodya:
Hi,
I am building a tool for converting flow code into TypeScript - [flowts](https://github.com/zxbodya/flowts).

And, I happened to also experiment with it on parcel codebase, on which there is pretty good progress - rebased it on current latest parcel version:

https://github.com/zxbodya/parcel/tree/ts

If one day, you decide to migrate parcel to TypeScript - that might be a good starting point 😄 

There is still some type errors to be fixed, and also to review fixes I did manually after converting it to typescript, but generally it already should be in a pretty good shape, and can at least - can be used for generating type definitions.

Drafted scripts to bundle `d.ts` files for publishing: 
- rollup config to bundle `index.d.ts` files from `tsc` output in converted version - https://github.com/zxbodya/parcel/tree/dts
- generated files - https://github.com/zxbodya/parcel/tree/generated-dts

It should be usable, only thing - probably some type dependencies to be added in package.json files before merging it.

wbinnssmith (approved):

PR #6501, opened 2021-06-22
Support automatic JSX runtime and configuration via tsconfig.json

### LINKED ISSUE #614: 🐛 jsconfig.json (TS) is not supported and allowJS compiler option ignored in tsconfig.json
**Choose one:**  🐛 bug report 

I usually use TypeScript to transpile regular JS files instead of babel. For intance, TypeScript allows to define a `jsconfig.json` file to do so. However, this behavior is not supported and apparently parcel tries to transpile with babel.

On the other side, the current behavior allows for TS files to be transpiled if `tsconfig.json` file is present. However, the `tsconfig.json` file also allows for `.js` files to be transpiled if setting the `{ allowJS: true }` compiler option but parcel still doesn't transpile `.js` files using this configuration.

Using `tsconfig.json`:
```js
{
  "compilerOptions": {
    "allowJs": true, // <-- Should allow .js as entry file
    "checkJs": true,
    "target": "es5",
    "module": "commonjs",
    "removeComments": false,
    "lib": [
      "es2015"
    ],
    "moduleResolution": "node",
    "baseUrl": "src/"
  },
  "include": [ "src/*" ],
  "exclude": [ "node_modules" ]
}
```
Using `jsconfig.json`:
```js
{
  "compilerOptions": {
    "checkJs": true,
    "target": "es5",
    "module": "commonjs",
    "removeComments": false,
    "lib": [
      "es2015"
    ],
    "moduleResolution": "node",
    "baseUrl": "src/"
  },
  "include": [ "src/*" ],
  "exclude": [ "node_modules" ]
}
```

### 🤔 Expected Behavior

It should respect compiler options in `tsconfig.json` if `{ allowJS: true }` and allow transpiling `.js` files using the TS Compiler instead of babel.

If `jsconfig.json` file is found, it should transpile `.js` files using the TypeScript Compiler instead of trying to use babel.

### 😯 Current Behavior

Parcel doesn't respect the compiler options in `tsconfig.json` for transpiling JS files. If using `jsconfig.json` the configuration is completely ignored.

### 💁 Possible Solution

Not to make it confusing to use, a possible solution may be for parcel to compile `.js` files using the TS Compiler when `jsconfig.json` is found instead of babel.

One reason why we do so is because we can gradually port parts of the application from JS to TS. However, the entry point still remains a `.js` file. For instance:
```
entry.js // imports example.ts
example.ts
```

### 🔦 Context

We don't always use babel. Most of the times, we use TS to transpile regular JS files to es5 and to be able to use TS modules from JS without having to transpile those first.

### 💻 Code Sample

`entry.js`:
```js
import fn from './tsFile'

console.log(
	'This is just a JS file',
	fn(1, 'string'),
)
```
`tsFile.ts` (Note: it's a TS file)
```ts
export default (a: number, b: string) => `${a}${b}`
```

### 🌍 Your Environment

| Parcel           | 1.4.1
| Node             | v9.3.0
| npm/Yarn         | NPM v5.5.1
| Operating System | MacOS

### LINKED ISSUE #3287: Support custom JSX pragmas in TSX

### LINKED ISSUE #4622: Parcel 2 should require less manual configuration of Babel to build Typescript (with decorators)
# 🙋 feature request

I request that Parcel 2 require less manual configuration of Babel to work with Typescript. 

In Parcel 1, Typescript *mostly* worked for me out of the box the way that I wanted it to, in a "zero configuration" way.  It was a long time before I even realized that Parcel was using Babel under the hood to transpile, rather that `tsc`.

In Parcel 2, I immediately had a bunch of difficulties with my Typescript code which tool a long time to resolve, and it took me a lot further in to the weeds of configuring Babel and installing Babel plugins than I would have liked.  

Specifically, my code uses "[decorators](https://www.typescriptlang.org/docs/handbook/decorators.html)", which did not work out-of-the-box with Parcel 2.  

The error message about this was reasonably clear about the problem but not terribly clear about the solution.

```
@parcel/transformer-babel: Support for the experimental syntax 'decorators-legacy' isn't currently enabled (6:5):
```

From this, and doing some googling, I gathered that I needed to have a `.babelrc` file which contained the following in the `plugins` section:

```
      ["@babel/plugin-proposal-decorators", { "legacy": true }],
      ["@babel/plugin-proposal-class-properties", { "loose" : true }]
```

but unfortunately when I did this I started getting syntax errors:

```
Uncaught SyntaxError: Unexpected token '<'
```

So it took a bunch more googling to figure out that I would need to add `@babel/plugin-transform-react-jsx` to the plugin list.    Since JSX worked correctly out-of-the-box,  it took me a while to realize that I had inadventently broken JSX by creating the .babelrc.  By creating the .babelrc to solve the decorators problem, I lost the settings which allowed it to process JSX.

I gather from #3911 that this behaviour has changed between Parcel 1 and 2, where in parcel 1 it merged your .babelrc and the Parcel default one where in Parcel 2, the project's .babelrc will override the default one.  While I can see how the old behaviour might have created some issues with incorrect merging, the new behaviour is also troubling because I lost good defaults like the JSX processing, and I had to spend extra time researching how to re-add that functionality. 

## 💁 Possible Solution

Here are a few possibilties on how the situation could be improved:

* Parcel's default babelrc could include the support for "experimental decorators" like it presumably did in Parcel 1 (my decorator code works fine in Parcel 1 without manual babelrc configuration).  Why was this removed?
* If there's a good reason not to include decorator support by default, and if Parcel is going to attempt to be "zero-configuration" by configuring babel for the user out of the box, then Parcel could be smarter about whether to include decorator support by checking the `tsconfig.json` to see if the `experimentalDecorators` setting is `true`, and then configure babel accordingly.  
* Parcel 2 could go back to the old behaviour of Parcel 1, merging the .babelrc project customizations with parcel's defaults.  (I'm not sure what the issues with that approach were, but is the new approach really the lesser of two evils?)
* There could be an "eject" feature which writes out parcel's default configuration to the .babelrc file as a starting point, if the user wants to customize it.   (Even after I read #3911 and knew that creating the .babelrc would override the defaults, I was having a hard time figuring out what the defaults were.  I could not find the default .babelrc settings by doing a search of this repo)
* Consider using `tsc` as the default transpiler for Typescript, to avoid issues due to Babel ignoring `tsconfig.json` and/or transpiling differently than `tsc` would.   (debated extensively in #4022).   I'm supportive of both sides of this debate.  I see the performance benefit of using Babel by default, but also see that Babel is not looking at all like a "zero configuration" option for my Typescript code the way that it is set up in Parcel 2.
* Typescript page (https://parcel2-docs.now.sh/recipes/typescript/) could be more detailed about babel configuration issues.  A lot of Typescript users aren't really familiar with using Babel to transpile, since they use `tsc` normally.   
   * Add a link to the default .babelrc so that it is clear what the defaults are in case you need to override them like I did
   * Add more clarity over what features of tsconfig map to what features of Babel
   * Add an explanation of the `.parcelrc` snippet that is in there which has `"@parcel/transformer-typescript-tsc"` added to the `transformers` section.  It is not clear what that is for, but I mistakenly thought it would solve some of my issues, when in fact it did not seem to be required.

### LINKED ISSUE #6389: New JSX transform doesn't work when using react v17+
# 🐛 bug report

react v17+ removes the need for `import React from "react";` statements but looks like the zero-config of Parcel doesn't work with it.

.babelrc: none

```
// package.json
...
    "react": "^17.0.1",
    "react-dom": "^17.0.1",
...
```

CLI command: `"parcel serve ./app/index.html --port 3000"`

## 🤔 Expected Behavior

Below code should work fine when using react v17+

```tsx

render(
    <h1>Hello World</h1>,
  document.getElementById("root")
);
```

## 😯 Current Behavior

Nothing is rendered on the screen.
Browser console error says "React is not defined"

## 💁 Possible Solution

This config works:
https://stackoverflow.com/a/67759434/7435656

## 🔦 Context

Migrating existing codebase from create-react-app to parcel. Followed the [Generic Webapp guide from the docs](https://v2.parceljs.org/getting-started/webapp/) but it doesn't look like those steps work for projects using the new JSX transform.

## 💻 Code Sample

Adding `import React from "react";` makes the application work but that shouldn't be required when using react v17+

## 🌍 Your Environment

| Parcel           | ^2.0.0-beta.3.1 |
| Node             | v14.17.0 |
| npm/Yarn         | 6.14.13 |
| Operating System | Ubuntu 20 LTS |

devongovett:
Fixes #6389. Fixes #4622. Fixes #3287. Fixes #614. Closes T-1038.

This adds support for the new automatic JSX runtime added in React 17. This is automatically enabled when the version of React the app depends on is >= 17.0.0.

Additionally, this adds support for reading several options from `tsconfig.json`.

* `jsx` - Enable JSX even when an auto-detected library isn't installed. Can also be used to enable the automatic JSX runtime if set to `react-jsx`.
* `jsxImportSource` - The import source for the automatic JSX runtime (e.g. `react` or `preact`)
* `jsxFactory` - The pragma for the classic JSX transform.
* `jsxFragmentFactory` - The fragment factory for the classic JSX transform.
* `experimentalDecorators` enables support for legacy decorators, as used by TypeScript.


parcel-benchmark:
## Benchmark Results

#### Timings

| --- | --- | --- |

#### Cold Bundles

#### Cached Bundles

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 10.22s | +12.00ms |
| Cached | 468.00ms | +4.00ms |

#### Cold Bundles

*No bundle changes detected.*

#### Cached Bundles

| dist/logo.1e014c76.png | 274.00b | +0.00b | 287.00ms | -19.00ms 🚀 |

</p></details>

#### Timings

| --- | --- | --- |

#### Cold Bundles

#### Cached Bundles

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 7.15s | +68.00ms |
| Cached | 402.00ms | -5.00ms |

#### Cold Bundles

*No bundle changes detected.*

#### Cached Bundles

*No bundle changes detected.*

</p></details>

**[Click here to view a detailed benchmark overview.](https://benchmark.parceljs.org/benchmark/302429439000576521)**

mischnic:
Should it also look at jsconfig.json? Might be weird to have to add a tsconfig file to change React settings even if you're not using Typescript

devongovett:
I wondered the same. I couldn't find definitive information about it though. Does jsconfig.json support compilerOptions even though it isn't being compiled by TSC presumably?

mischnic:
This makes it sound like it is:

> jsconfig.json is a descendant of tsconfig.json, which is a configuration file for TypeScript. jsconfig.json is tsconfig.json with "allowJs" attribute set to true.
> (https://code.visualstudio.com/docs/languages/jsconfig)

And there is an example with `compilerOptions.target` on that page

devongovett:
Hmm ok. Also should tsconfig only apply to .ts and .tsx files or also to .js files in the same project? And same for the inverse? Or should they just be treated as interchangeable?

mischnic:
I would say prefer tsconfig over jsconfig in `getConfigFrom`, but I don't know how tsc/VS Code behave.

mischnic:
> This is automatically enabled when the version of React the app depends on is >= 17.0.0.

Preact >=10.5.0 support the automatic JSX runtime: https://github.com/preactjs/preact/releases/tag/10.5.0

devongovett:
Found this "bug" about jsconfig.json not being picked up when a tsconfig.json is present, and it claims it's by design so we'll follow it. https://github.com/microsoft/TypeScript/issues/15869

packages/transformers/js/src/JSTransformer.js
```diff
@@ -515,7 +598,10 @@ export default (new Transformer({
           priority: dep.kind === 'DynamicImport' ? 'lazy' : 'sync',
           isOptional: dep.is_optional,
           meta,
-          resolveFrom: dep.is_helper ? __filename : undefined,
+          resolveFrom:
+            dep.is_helper && !dep.specifier.endsWith('/jsx-runtime')
```
mischnic, line 602: What is the jsx-runtime part for?

mischnic (approved):

wbinnssmith (approved):

Rechdan:
We can't don't import React when on v17?

PR #6504, opened 2021-06-22
Don't attempt to resolve URLs starting with '#'

### LINKED ISSUE #6471: Bundler fails when processing CSS clip-path URL
# 🐛 bug report

I'm in the process of migrating our team's build from Parcel v1 to v2. It seems that the build fails when it gets to a Stylus file containing a clip-path or fill url: `Received URL without a pathname #clip-path`.
Screen Shot 2021-06-16 at 4 27 55 PM

I'm not sure if I should be using some kind of transformer to get Parcel to ignore these CSS rules.

.babelrc
```
{
  "presets": [
    "jest"
  ],
  "plugins": [
    "transform-react-handled-props",
    "@babel/plugin-proposal-function-sent",
    "@babel/plugin-proposal-throw-expressions"
  ]
}
```

```js
{
  "extends": "@parcel/config-default"
}
```

## 🤔 Expected Behavior

Parcel should ignore clip-path or fill URLs starting with `#` in Stylus files and bundle them correctly.

## 😯 Current Behavior

An error is thrown and the build fails.

```
🚨 Build failed.
@parcel/core: Received URL without a pathname #clip-path.
CardDetails/Graphics/styles.styl:38:19
  37 |   .cls-2
> 38 |     clip-path url('#clip-path')
>    |                   ^^^^^^^^^^^^
  39 |
  40 |   .cls-3
```

## 💁 Possible Solution

Seems somewhat similar to the issue raised in #5175

## 🔦 Context

I'm trying to migrate our build from v1 to v2 to keep our code base up-to-date as development has stopped on v1.

## 💻 Code Sample

[Minimum test case repo](https://github.com/hbilles/parcel2-minimum-case)

## 🌍 Your Environment

| Parcel           | v2.0.0-beta.3.1
| Node             | v14.17.0
| npm/Yarn         | npm v6.14.13
| Operating System | macOS 11.3.1

hbilles:
# ↪️ Pull Request

This resolves the issue I came across in #6471 where Stylus files containing a rule like `clip-path: url('#clip-path')` caused the build to fail and throw an error (`Received URL without a pathname #clip-path`).

At first I thought the issue was coming from the `@parcel/transformer-stylus` package, but after digging into it I discovered the error is being thrown _before_ the stylus transformer is hit.

After looking around I came across this [issue](https://github.com/parcel-bundler/parcel/issues/5175) with `#` in paths of SVG files which was solved in #5188. I chose to update `PathRequest.js` to check if a path starts with `#` and return `null` if so.

I can't think of any cases where we would want Parcel to attempt resolving a path starting with `#`, but I could be wrong.

Let me know if there's a different path I should take or if there's any issues with my proposed fix.

Thanks!

## 💻 Examples

This is an example of a Stylus rule that causes Parcel to fail and is fixed by this PR:
```
.svg-background .cls-1
  clip-path: url('#clip-path')
```

## ✔️ PR Todo

### DIFF
--- a/packages/core/core/src/requests/PathRequest.js
+++ b/packages/core/core/src/requests/PathRequest.js
@@ -168,12 +168,15 @@ export class ResolverRunner {
         }
       }
     } else {
-      if (
-        dep.specifierType === 'url' &&
-        dependency.specifier.startsWith('//')
-      ) {
-        // A protocol-relative URL, e.g `url('//example.com/foo.png')`
-        return null;
+      if (dep.specifierType === 'url') {
+        if (dependency.specifier.startsWith('//')) {
+          // A protocol-relative URL, e.g `url('//example.com/foo.png')`
+          return null;
+        }
+        if (dependency.specifier.startsWith('#')) {
+          // An ID-only URL, e.g. `url(#clip-path)` for CSS rules
+          return null;
+        }
       }
       filePath = dependency.specifier;
     }
--- a/packages/core/integration-tests/test/integration/stylus-id-url/index.js
+++ b/packages/core/integration-tests/test/integration/stylus-id-url/index.js
@@ -1,0 +1,5 @@
+require('./index.styl');
+
+module.exports = function () {
+  return 2;
+};
--- a/packages/core/integration-tests/test/integration/stylus-id-url/index.styl
+++ b/packages/core/integration-tests/test/integration/stylus-id-url/index.styl
@@ -1,0 +1,2 @@
+.svg-background .clip-1
+  clip-path: url("#clip-path")
--- a/packages/core/integration-tests/test/stylus.js
+++ b/packages/core/integration-tests/test/stylus.js
@@ -96,6 +96,31 @@ describe('stylus', function() {
     );
   });
 
+  it('should ignore paths starting with "#" when resolving with stylus url()', async function() {
+    let b = await bundle(
+      path.join(__dirname, '/integration/stylus-id-url/index.js'),
+    );
+
+    assertBundles(b, [
+      {
+        name: 'index.js',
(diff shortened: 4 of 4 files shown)


packages/core/core/src/requests/PathRequest.js
```diff
@@ -168,12 +168,15 @@ export class ResolverRunner {
         }
       }
     } else {
-      if (
-        dep.specifierType === 'url' &&
-        dependency.specifier.startsWith('//')
-      ) {
-        // A protocol-relative URL, e.g `url('//example.com/foo.png')`
-        return null;
+      if (dep.specifierType === 'url') {
```
devongovett, line 171: Ideally we would let plugins that wanted to handle urls do so. Maybe this should be in the default resolver plugin instead?

hbilles: I agree it would make sense to have a check like this in something like the default resolver plugin. However I can't see another way of bypassing the `if (typeof parsed.pathname !== 'string')` block around line 184/187 below which throws an error when it attempts to parse a "path" like `#clip-path`.

I tried handling the case in `packages/resolvers/default/src/DefaultResolver.js` but nothing seems to work, I suspect because the error is being thrown before any resolver plugins get hit. Part of the reason I chose to modify the code in `PathRequest` is because it saw it handling a similar case for paths starting with `//`.

I'm not that familiar with the structure of the Parcel codebase, so if you have any suggestions for a better way to handle the case of a path starting with `#` I'm happy to try it.

devongovett: Ok, thanks. Given that this is already the case, this is fine for now.

devongovett (approved):

PR #6516, opened 2021-06-26
Always return new streams from asset.getStream()

wbinnssmith:
Previously, there was only one stateful stream associated with an asset with stream content, and when called, `asset.getStream()` would return a reference to the stream and unset it on the asset. `asset.setStream()` now expects a stream-returning function.

This makes it possible to call `asset.getStream()` multiple times, each returning a new stream. For transformers that set streams, *this means that transformations may happen multiple times*, which seems safer than unsetting it.

Test Plan: `yarn test`

### DIFF
--- a/packages/core/core/src/UncommittedAsset.js
+++ b/packages/core/core/src/UncommittedAsset.js
@@ -18,6 +18,7 @@ import type {
 } from './types';
 
 import invariant from 'assert';
+import fs from 'fs';
 import {Readable} from 'stream';
 import SourceMap from '@parcel/source-map';
 import {
@@ -44,7 +45,7 @@ import {BundleBehaviorNames} from './types';
 type UncommittedAssetOptions = {|
   value: Asset,
   options: ParcelOptions,
-  content?: ?Blob,
+  content?: ?(string | Buffer),
   mapBuffer?: ?Buffer,
   ast?: ?AST,
   isASTDirty?: ?boolean,
@@ -56,7 +57,7 @@ type UncommittedAssetOptions = {|
 export default class UncommittedAsset {
   value: Asset;
   options: ParcelOptions;
-  content: ?(Blob | Promise<Buffer>);
+  content: ?(string | Buffer | Promise<Buffer> | (() => Readable));
   mapBuffer: ?Buffer;
   sourceContent: ?string;
   map: ?SourceMap;
@@ -177,7 +178,7 @@ export default class UncommittedAsset {
     if (typeof content === 'string' || content instanceof Buffer) {
       return content.toString();
     } else if (content != null) {
-      this.content = bufferStream(content);
+      this.content = bufferStream(content());
       return (await this.content).toString();
     }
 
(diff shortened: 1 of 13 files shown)


packages/core/core/src/UncommittedAsset.js
```diff
@@ -223,7 +224,7 @@ export default class UncommittedAsset {
     this.clearAST();
   }
 
-  setStream(stream: Readable) {
+  setStream(stream: () => Readable) {
```
wbinnssmith, line 227: Maybe this needs a new name?

wbinnssmith:
One thing I noticed re: streams of JS is that we always buffer the entire stream into memory to detect existing sourcemaps [0]  (this goes for CSS and others that can have sourcemaps) and to transform JS at all [1]. Since there isn't a well-known streaming JS transformer (that I know of), this is probably fine, but worth noting that it de-opts one of the key points of streaming. We should continue to benefit from it for things like copying large binary files, or other transformers like the ImageTransformer that transform with streams.

[0] https://github.com/parcel-bundler/parcel/blob/88f2c7c2242e8530d62502e28ea0c08296227e58/packages/core/core/src/UncommittedAsset.js#L236
[1] https://github.com/parcel-bundler/parcel/blob/88f2c7c2242e8530d62502e28ea0c08296227e58/packages/transformers/js/src/JSTransformer.js#L196

parcel-benchmark:
## Benchmark Results

#### Timings

| --- | --- | --- |

#### Cold Bundles

#### Cached Bundles

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 10.16s | -143.00ms |
| Cached | 489.00ms | +26.00ms ⚠️ |

#### Cold Bundles

*No bundle changes detected.*

#### Cached Bundles

*No bundle changes detected.*

</p></details>

#### Timings

| --- | --- | --- |

#### Cold Bundles

#### Cached Bundles

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 7.16s | +80.00ms |
| Cached | 407.00ms | -8.00ms |

#### Cold Bundles

*No bundle changes detected.*

#### Cached Bundles

*No bundle changes detected.*

</p></details>

**[Click here to view a detailed benchmark overview.](https://benchmark.parceljs.org/benchmark/302419072758317581)**

packages/core/core/src/UncommittedAsset.js
```diff
@@ -213,6 +210,10 @@ export default class UncommittedAsset {
     return blobToStream(this.content ?? Buffer.alloc(0));
   }
 
+  makeStream(): Readable {
```
devongovett, line 213: Remove?

packages/core/core/src/UncommittedAsset.js
devongovett: I do wonder about this API. What if we instead still accepted a stream, but immediately wrote it into the cache. And when calling `getStream` we created a new stream to read it from the cache. Seems easy for plugins to mess up and always return the same stream from the function they pass or something.

packages/transformers/image/src/ImageTransformer.js
```diff
@@ -48,7 +48,7 @@ export default (new Transformer({
         });
       }
 
-      asset.setStream(imagePipeline);
+      asset.setStream(() => imagePipeline);
```
devongovett, line 51: Won't this return the same stream multiple times?

mischnic:
Re-reading the stream from cache instead of rerunning the transformer also sounds like a better solution to me.

devongovett:
I attempted to implement my suggestion, but tbh I'm not sure it's worth it since the LMDB cache implementation doesn't support streams anyway and has to buffer the whole thing into memory...

packages/core/core/src/UncommittedAsset.js
findmet: I request you don't

PR #6531, opened 2021-06-30
JS runtime improvements

### LINKED ISSUE #6261: Workers don't work in library mode
# 🐛 bug report

I recently updated my project to use a worker:
```
    const worker = new Worker(
      new URL("./workers/my-worker.ts", import.meta.url)
    );
```

When I import the package from another package in the monorepo with `parcel serve`, things work as expected. 

However, when I `parcel build` the package, this happens:
```
$ parcel build ./src/index.ts --no-source-maps --no-optimize --no-cache
🚨 Build failed.
@parcel/packager-js: External modules are not supported when building for browser
```

I am not targeting `browser` in my package.json.

It is not clear if this is an actual bug, or a misunderstanding on my part, but the error message is not clear anyway.

But, if I remove the line about the worker, it builds.

babel.config.json:
```json
{
  "plugins": [
    "@parcel/babel-plugin-transform-runtime",
    ["inline-svg", { "disableSVGO": true }]
  ],
  "presets": [
    "@parcel/babel-preset-env",
    "@babel/preset-react",
    "@babel/preset-typescript"
  ]
}
```

package.json (just the relevant parts)
```
{
  ...
  "main": "dist/main/index.js",
  "module": "dist/module/index.js",
  "source": "src/index.ts",
  "types": "index.d.ts",
  "scripts": {
    "build": "parcel build ./src/index.ts --no-source-maps --no-optimize --no-cache"
  },
  "browserslist": "> 0.2%",
  "targets": {
    "types": false,
    "main": {
      "includeNodeModules": [
        ... list of internal packages ...
      ]
    },
    "module": {
      "includeNodeModules": [
        ... list of internal packages ...
      ]
    }
  },
  "webpackEntry": "src/index.ts"
  ...
}
```

## 🤔 Expected Behavior

`parcel build` should successfully build, just like it does when there is no web worker. Or at least, the error message should be clearer.

## 😯 Current Behavior

```
🚨 Build failed.
@parcel/packager-js: External modules are not supported when building for browser
```

## 💁 Possible Solution

Not sure

## 🔦 Context

This is blocking me from building a project that uses a worker.

## 💻 Code Sample

```
    const worker = new Worker(
      new URL("./workers/my-worker.ts", import.meta.url)
    );
```

## 🌍 Your Environment

| Parcel           | 2.0.0-beta.2
| Node             | 12.18.3
| npm/Yarn         | 1.22.10
| Operating System | MacOS 10.15.7

devongovett:
Some improvements to the JS runtime to reduce bundle size and improve caching.

* The ESM output format now uses the runtime manifest in the entry bundle just like the global output format. This should result fewer cascading cache invalidations.
* We use `new URL(relativePath, import.meta.url)` in ESM output when targeting browsers supporting `import.meta.url` natively. This avoids needing to include the runtime that attempts to determine the script url by parsing an error stack.
* The relative path utility function is removed. Rather than the manifest storing the full path from the target dist dir, it only stores the base name of the bundle since the part of the path that's most likely to change is just the hash. The relative path part (if any) is computed at build time and inlined into the bundle to avoid the need for this large-ish function.
* Don't include worker url helper and bundle url helper in library builds. Fixes #6261.

To do:

- [x] tests
- [x] questions below


packages/runtimes/js/src/JSRuntime.js
```diff
@@ -570,13 +587,35 @@ function getRelativePathExpr(
...
-  return JSON.stringify(relativeBundlePath(from, to, {leadingDotSlash: false}));
+  return JSON.stringify(relativePath);
+}
+
+function getAbsoluteUrlExpr(relativePathExpr: string, bundle: Bundle) {
+  if (
+    bundle.env.outputFormat === 'esmodule' &&
+    (bundle.env.isLibrary || bundle.env.supports('import-meta-url'))
+  ) {
+    return `new __parcel__URL__(${relativePathExpr}, import.meta.url).toString()`;
```
devongovett, line 601: For library builds, this might not fully fix the problem because the `new URL` call ends up in a separate module with the worker doing something like `new Worker(parcelRequire('...'))`. That's not gonna be statically analyzable by another bundler. Not really sure how to fix this though. Runtimes don't modify the compiled code of the original module...

packages/runtimes/js/src/JSRuntime.js
```diff
@@ -570,13 +587,35 @@ function getRelativePathExpr(
...
+  } else if (bundle.env.outputFormat === 'commonjs') {
+    return `require('path').join(__dirname, ${relativePathExpr})`;
```
devongovett: For commonjs libraries, what should the output be? `import.meta` is technically only supported in ESM, but this pattern won't be picked up by other bundlers. Should we still output the `new URL` pattern for all library output, even CJS?

parcel-benchmark:
## Benchmark Results

#### Timings

| --- | --- | --- |

#### Cold Bundles

#### Cached Bundles

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 8.36s | -101.00ms |
| Cached | 412.00ms | -17.00ms |

#### Cold Bundles

| dist/index.js | 487.47kb | -591.00b 🚀 | 4.80s | -25.00ms |
| dist/logo.1e014c76.png | 274.00b | +0.00b | 191.00ms | -13.00ms 🚀 |

#### Cached Bundles

| dist/index.js | 487.47kb | -591.00b 🚀 | 4.77s | -23.00ms |
| dist/logo.1e014c76.png | 274.00b | +0.00b | 220.00ms | +27.00ms ⚠️ |

</p></details>

#### Timings

| --- | --- | --- |

#### Cold Bundles

#### Cached Bundles

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 6.10s | +140.00ms |
| Cached | 349.00ms | -6.00ms |

#### Cold Bundles

*No bundle changes detected.*

#### Cached Bundles

*No bundle changes detected.*

</p></details>

**[Click here to view a detailed benchmark overview.](https://benchmark.parceljs.org/benchmark/303490315905401408)**

packages/runtimes/js/src/JSRuntime.js
```diff
@@ -181,17 +181,9 @@ export default (new Runtime({
         }),
       );
 
-      if (bundle.env.outputFormat === 'commonjs' && mainBundle.type === 'js') {
```
devongovett, line 184: This came from #4744 but I had to revert it because it breaks URL dependencies to a JS file in CJS output. I think the real solution to what the PR was trying to accomplish would be to make the dependency async.

devongovett:
For libraries, `new Worker(new URL('worker.js', import.meta.url))` now becomes `new Worker(new URL('HASH', import.meta.url))` during transformation, and the HASH is replaced with the final URL during packaging similar to URL dependencies in other languages like CSS. This way the output is statically analyzable, unlike our previous output of `new Worker(require('some-worker-runtime'))`. That output is still used for non-libraries though, because we want to be able to use the bundle manifest.

wbinnssmith (approved):

Issue #5590, opened 2021-01-05
Extract CSS from the bundle


aminya:
# 🙋 feature request

 I have a component library that imports CSS into the components. I do this so Parcel transforms and bundles CSS for me. But I don't want them to be required into the JS bundle which breaks the workflow of anyone who does not use Parcel or similar transformers.

https://github.com/aminya/solid-simple-table/blob/20c4c543fe4604acd3877426055abc474416d67a/src/SimpleTable.tsx#L2

## 🤔 Expected Behavior

An option to remove `require('something.css')` after the bundle is created.

## 😯 Current Behavior

There is no way to remove `require('something.css')`.

## 💁 Possible Solution
I couldn't find a bundler agnostic CSS extractor, but probably there is one out there.
https://www.npmjs.com/search?q=extract%20css

## 🔦 Context
Related to https://github.com/parcel-bundler/parcel/discussions/5246

## 💻 Examples

```tsx
import "./MyStyle.less"
console.log('Hey')
```
I want the bundle to only include
```tsx
console.log('Hey')
```

mischnic:
One workaround would be using alias in package.json to exclude the JS import: `alias: { "./src/MyStyle.less": false }`

aminya:
@mischnic Using alias causes the `less` file to not be transformed at all.

aminya:
I found a babel plugin that does this! 🎉 

However, Babel transformation happens "before" parcel performs the CSS bundling. If you can integrate this babel plugin and somehow run it in the end after everything is done, that exactly solves the issue.

https://www.npmjs.com/package/@iwatakeshi/babel-plugin-remove-import-styles

For now, I wrote a custom script and used babel's cli:

package.json
```
    "style": "babel dist --out-dir dist --config-file ./babel.style.config.json",
```
babel.style.config.json
```json
{
  "plugins": [
    [
      "@iwatakeshi/babel-plugin-remove-import-styles",
      {
        "extensions": [".css", ".less", ".sass"]
      }
    ]
  ]
}
```

aminya:
Can we get this babel plugin integrated into Parcel? It should run after all the CSS transformations.

mischnic:
You'd need to create another transformer that runs after all default JS transformers

PR #6532, opened 2021-06-30
Changed some JS to be IE11 compatible.

### LINKED ISSUE #6502: Bundle contains non IE11 compatible code
# 🐛 bug report

Bundle contains lots of non IE11 compatible code, even with `"browserslist": ["IE 11"]`.

Empty project, just a dummy index.html + index.js.

package.json
`"browserslist": ["IE 11"]`

CLI
`npx parcel serve .\index.html`

## 🤔 Expected Behavior

Bundle should run in IE11.

## 😯 Current Behavior

Syntax error in IE11. Bundle contains multiple non IE11-code. This is some injected boilerplate code and does not originate from my simple index.js.

Some examples from the bundle:

IE doesn't like the trailing comma:

```
      modules[name][0].call(
        module.exports,
        localRequire,
        module,
        module.exports,
        this,
      );
```

Arrow function:
```
let assets = data.assets.filter((asset)=>asset.envHash === HMR_ENV_HASH
```

IE doesn't support `for of`:

```
for (let ansiDiagnostic of data.diagnostics.ansi){
```

## 💁 Possible Solution

Bundle should be free of non IE11. If this is injected boilerplate, that code should be compatible with IE11. Either by changing the boilerplate template or by running it through some transpilation.

## 🔦 Context

N/A

## 💻 Code Sample

I think my examples above are enough to find the issue. If not, I'm happy to push a simple repro.

## 🌍 Your Environment

| Parcel           | @next
| Node             | 16
| npm/Yarn         |7.13
| Operating System |Win10

hallsbyra:
# ↪️ Pull Request

Partial fix for #6502

Had to revert changes done by `yarn format` to `dev-prelude.js` since it would re-introduce trailing commas that break IE11 compatibility. Maybe these files should be excluded from `yarn format` somehow?

## ✔️ PR Todo

- [ ] Added/updated unit tests for this change - N/A
- [ ] Filled out test instructions (In case there aren't any unit tests) - N/A


mischnic:
> Had to revert changes done by yarn format to dev-prelude.js since it would re-introduce trailing commas that break IE11 compatibility. Maybe these files should be excluded from yarn format somehow?

That should already be handled by this:
https://github.com/parcel-bundler/parcel/blob/d1e0a1a0781d65fdfb84354e52ed85313717bc4b/.prettierrc#L9-L11

mischnic:
When building this with `parcel build a.js`:
```js
// a.js
console.log(require("./b.js"));

// b.js
module.exports.x = 123;
```

the output contains arrow functions:
```js
(function () {
...
parcelRequire.register("fPYqQ", function(module, exports) {

$parcel$export(module.exports, "x", () => $b87a7ef68e90c5ee$export$10b1f2ceae7ab64e, (v) => $b87a7ef68e90c5ee$export$10b1f2ceae7ab64e = v);
var $b87a7ef68e90c5ee$export$10b1f2ceae7ab64e;
$b87a7ef68e90c5ee$export$10b1f2ceae7ab64e = 123;

});


console.log((parcelRequire("fPYqQ")));
```
This is what I mentioned in the second part of https://github.com/parcel-bundler/parcel/issues/6502#issuecomment-871333088

hallsbyra:
Ok, thanks. Took care of that second part of https://github.com/parcel-bundler/parcel/issues/6502#issuecomment-871333088 in latest commit.

devongovett:
Maybe we should do this based on the environment so we don't have to ship larger output to modern browsers?

aminya:
> Maybe we should do this based on the environment so we don't have to ship larger output to modern browsers?

Yes, please! I do not want to support IE11. It's better to break the websites for IE11 users, so they start to think about upgrading.

hallsbyra:
> Maybe we should do this based on the environment so we don't have to ship larger output to modern browsers?

Yes, that'd be good. This PR is now ready to merge. I suggest we merge this as it is to get the bug fixed. Then we open a separate issue/PR with an improvement to the current behavior. Keeping PRs small is good.

danieltroger:
Sounds good @hallsbyra. I vote for merging.

danieltroger:
@mischnic what is blocking this from being merged?

aminya:
Its regressions for those who don't want to support IE11:
https://github.com/parcel-bundler/parcel/pull/6532#issuecomment-871534264

danieltroger:
@aminya that's not a reason to not merge this. It's like saying you don't want firefighters to put out the burning house of your neighbour because it would wake you up.

You get like 50 bytes extra, while my efforts to support older browsers stop being in vain. It doesn't break anything for you, while it fixes something for me. And I agree that we should use arrow functions in modern browsers. I think terser converts them anyways if you use the right options.

aminya:
You can run this in the meantime.
```
    "ie11": "babel dist --out-dir dist --config-file ./babel.ie11.config.json",
```

babel.ie11.config.json
```
{
  "targets": {
    "ie": "11"
  }
}
```

danieltroger:
Thanks @aminya. Running that on parcel's output will also transpiled all core-js polyfills, is there any indication showing it will work? I'll probably try it later on but I give it a 40% success rate because I read in some of @zloirock's issues that that messes with some core-js Symbol.iterator implementation

mischnic:
> You get like 50 bytes extra, while my efforts to support older browsers stop being in vain. It doesn't break anything for you, while it fixes something for me. And I agree that we should use arrow functions in modern browsers.

If you have a large bundle with 3000 of these export declarations, this change will make it 14*3000=42kb larger (even when minified).

> I think terser converts them anyways if you use the right options.

Terser cannot do this safely (= by default), it requires `unsafe_arrows: true`

hallsbyra:
> Yes, please! I do not want to support IE11. It's better to break the websites for IE11 users, so they start to think about upgrading.

Dropping support for IE11 is indeed an option. But that decision is out of scope of this PR. I suggest that is discussed somewhere else, and keep this PR as the bug fix it is intended to be.

mischnic:
Doesn't IE 11 not really support `let`? Can you replace them with `var` in the various files you already touched?

hallsbyra:
> Doesn't IE 11 not really support `let`? Can you replace them with `var` in the various files you already touched?

`let` works in IE except in `for`-loops. (https://caniuse.com/?search=let) I reviewed the code and could not find any such cases.

danieltroger:
@hallsbyra it might work in IE11 but it doesn't work on older androids. You'd do me a favor if you'd changed it in the screenshot that I sent [here](https://github.com/parcel-bundler/parcel/issues/6502#issuecomment-882620767).

aminya:
Instead of transforming all of these manually, there could be a babel transformer that runs on the final output when you select something ancient like IE11. 

A babel transformer that runs on the final output is useful in general (see [this](https://github.com/parcel-bundler/parcel/issues/5590#issuecomment-865657074) as an example)

hallsbyra:
@danieltroger , I replaced the `let`s.

@aminya , a transformer would be great, but I'm not versed enough in the codebase to do it myself.

danieltroger:
@hallsbyra TYSM for replacing the `let`s. Can't wait for this to get merged.

@aminya FYI I tried babel but couldn't find a browserslist query for this userAgent: `Mozilla/5.0 (Linux; Android 5.0.2; Android SDK built for x86_64 Build/LSY66K) AppleWebKit/537.36 (KHTML, like Gecko) Version/4.0 Chrome/37.0.0.0 Mobile Safari/537.36`.
I tried `android 5` (which is accurate) - it still had `let`, `android 3` which still had `let`, `ChromeAndroid 37` which babel didn't support and a couple more.

I manually replaced all `let ` with `var ` in the file and everything ran flawlessly (!). Apparently my bundle doesn't contain any wrapped assets where the arrow function bug comes to show.

devongovett:
Hey, hope you don't mind me pushing to your branch. I made 3 changes:

1. Rather than removing use of arrow functions completely for all targets, we now detect whether they are supported in the target environment and only fall back to normal functions if necessary. This should result in smaller bundles for a majority of builds.
2. Rather than manually fixing issues in runtime helpers, transpile them with babel before publishing. In this case the transpilation required is very minor (mainly let -> var and fixing trailing commas) so it doesn't have much of an effect on build size. Eventually we might do this within the parcel build itself, but right now we don't transpile node_modules.
3. Added linting for IE 11 compatibility in files that don't get transpiled - right now just the dev prelude. This should prevent future regressions.

danieltroger:
Yay, that's amazing!

devongovett (approved):

Issue #2901, opened 2019-04-07
SVG as an Asset Type for image references


schwjm:
# 🙋 feature request

Support parsing .svg files to scan and update for references like inline SVG in HTML is already done.

## 🤔 Expected Behavior

An `<image>` tag inside a .svg file should be able to reference another .svg file via `href` or `xlink:href`. This is supported in Firefox 66, Chrome 73, IE 11. These references should be supported in SVGs included in the bundle.

```svg
<image href="./examplebase.a2a5d7d4.svg" x="15" y="20" width="70" height="70" />
```

## 😯 Current Behavior

SVG files are not treated as a built-in asset type--only HTML inline SVG is massaged, so references in .svg files do not know that the .svg files in the bundle are now under a different filename with the hash.

```svg
<image href="./examplebase.svg" x="15" y="20" width="70" height="70" />
```

## 💁 Possible Solution

.svg files could be parsed and have their contents examined. It didn't seem like it was as easy as just running .svg as an `HTMLAsset` unfortunately, a separate parser or configuration may be required. Image references could perhaps even be inlined instead of linked.

## 🔦 Context

I am constructing a set of SVG icons, and separate from that, a tool that wants to use these icons in a composition or with modifications. For example, the same icon could be reused in different contexts. The SVG references are all in my JavaScript because the HTML is dynamically generated, and I would prefer to separate the concern of image data from JavaScript code (i.e. avoid code to compose new SVG images if I can just reference a file.)

### examplebase.svg
```svg
<svg xmlns="http://www.w3.org/2000/svg"
    width="100" height="100" viewBox="0 0 100 100"
>
<defs>
    <linearGradient id="bg"
        x1="0%" y1="0%" x2="100%" y2="0%"
    >
        <stop offset="0%" stop-color="#737eaf" />
        <stop offset="100%" stop-color="#3f457b" />
    </linearGradient>
</defs>
<rect x="5" y="5" width="90" height="90"
    rx="2"
    fill="url(#bg)"
/>
</svg>
```
![examplebase](https://user-images.githubusercontent.com/20408541/55686392-3d161f00-5926-11e9-9463-7a85366b9cb2.png)

### example.svg
```svg
<svg xmlns="http://www.w3.org/2000/svg"
    width="100" height="100" viewBox="0 0 100 100"
>
<image href="./examplebase.svg" x="15" y="20" width="70" height="70" />
<text x="50" y="30" width="100" text-anchor="middle"
[...]

PR #6535, opened 2021-07-01
Support xml-stylesheet processing instructions

thewilkybarkid:
# ↪️ Pull Request

Adds `@parcel/transformer-xml`, starting with support for the `xml-stylesheet` processing instruction.

Relates to #2901.

## 💻 Examples

A HTML file containing:

```html
<object data="image.svg"></object>
```

`image.svg`:

```svg
<?xml-stylesheet href="style.css"?>
<svg xmlns="http://www.w3.org/2000/svg">
  <rect width="100" height="100"/>
</svg>
```

`style.css`:

```css
:root {
  fill: red;
}
```

The square should appear red rather than black.

## 🚨 Test instructions

(I've added an integration test, but unable to get it working locally.)

## ✔️ PR Todo


thewilkybarkid:
There's more that can be added (`xlink:href` is the obvious one), but some SVG specifics too in a complementary SVG transformer (more details in #2901).

yarn.lock
```diff
@@ -8032,6 +8150,39 @@ jsdom@^15.2.1:
...
+    symbol-tree "^3.2.4"
+    tough-cookie "^4.0.0"
+    w3c-hr-time "^1.0.2"
+    w3c-xmlserializer "^2.0.0"
+    webidl-conversions "^6.1.0"
+    whatwg-encoding "^1.0.5"
+    whatwg-mimetype "^2.3.0"
+    whatwg-url "^8.5.0"
+    ws "^7.4.5"
+    xml-name-validator "^3.0.0"
```
thewilkybarkid, line 8184: This is the only local change, not sure where the other changes have come from.

thewilkybarkid:
How do I get the integration test to pick up the new transformer? Obviously, it's not published (yet), but I assumed these would use the local versions?

mischnic:
You need to rename "packages/transformers/xml/src/index.js" to "XMLTransformer.js"

mischnic:
I'm not entirely sure whether a single transformer for atom, rss, svg and xml is the right approach.

And another problem with your changes is that `import x from "./icon.svg";` is allowed, while it should require `url:./icon.svg` like other images (as it is now).

thewilkybarkid:
I imagined having a SVG transformer to use in combination with this, so this handles generic XML features and the other the SVG format. I included Atom/RSS since they're common extensions using XML, but they could have their own transformers too if needed.

Ok to change the format (though I'll be changing it back locally).

thewilkybarkid:
Thinking about, XLink isn't so generic: `<svg:a>` should behave differently to `<svg:image>` (the former should have a stable path).

thewilkybarkid:
Given https://github.com/parcel-bundler/parcel/pull/6565 should I remove the `url:` prefix?

packages/transformers/xml/src/XMLTransformer.js
```diff
@@ -0,0 +1,63 @@
...
+export default (new Transformer({
+  canReuseAST({ast}) {
+    return ast.type === 'jsdom' && semver.satisfies(ast.version, '^16.6.0');
+  },
+
+  async parse({asset}) {
+    return {
+      type: 'jsdom',
+      version: '16.6.0',
+      program: new JSDOM(await asset.getBuffer(), {
```
devongovett, line 17: JSDOM seems a bit heavyweight for this use case. Should we use a simpler XML parser for this instead?

packages/core/integration-tests/test/integration/xml-stylesheet-processing-instruction/styled.xml
```diff
@@ -0,0 +1,10 @@
+<?xml-stylesheet href="style1.css"?>
+<?xml-stylesheet href="style2.css?>
+<?xml-smilesheet href="style2.css"?>
```
devongovett, line 3: Is that intentional?

packages/configs/default/index.json
```diff
@@ -40,6 +40,7 @@
     "style:*.vue": ["@parcel/transformer-vue"],
     "custom:*.vue": ["@parcel/transformer-vue"],
     "*.{png,jpg,jpeg,webp}": ["@parcel/transformer-image"],
+    "*.{atom,rss,svg,xml}": ["@parcel/transformer-xml"],
```
devongovett, line 43: Seems like we should probably have different transformers for each of these. Atom, RSS, and SVG will all have different tags and attributes that we'll need to handle to create dependencies for sub resources. Maybe keep this transformer specific to `.xml` files?

devongovett:
Curious what your use case is for this? How common is it to have a `.xml` file with a stylesheet attached? I've never seen that before.

thewilkybarkid:
It's how Graphviz attaches stylesheets (refs #6544).

thewilkybarkid:
Probably better to just make this an SVG transformer (as the name's now free). It might be that an SVG using a PI containing (S)CSS (as opposed to others like XSLT) should become a `<style>@import(...)</style>` so it can be inlined and then optimised. (External files aren't loaded when using an ``.)

If there are future uses for processing XML-based formats then it could be revisited, but the expected behaviour might be different anyway.

thewilkybarkid:
Closing in favour of #6595.

PR #6536, opened 2021-07-02
Disable workers with string literals and improve diagnostics

devongovett:
Closes T-1087

This disables workers of all types from being created with a string literal. It's now required to use `new Worker(new URL('filename.js', import.meta.url), {type: 'module'})`. This is because strings passed to the worker constructor are supposed to be resolved relative to the page URL, not to the current file. Using the `URL` constructor solves this issue so that Parcel is more web compatible. This is also how other bundlers like webpack now work.

This also makes some formatting improvements to diagnostics, especially when multiple diagnostics are displayed at once. The code frame and hints are now indented underneath each error message so it's clear what it is associated with.

image

In addition, it's now possible to have multiple code frames pointing to different files within the same diagnostic. This way, errors can show context for the error in another file without creating a separate diagnostic which looks like a different error.

image

There's also some other small diagnostic improvements in this PR which you can see in the code.


packages/core/diagnostic/src/diagnostic.js
```diff
@@ -63,7 +65,7 @@ export type Diagnostic = {|
   language?: string,
 
   /** A code frame points to a certain location(s) in the file this diagnostic is linked to (optional) */
-  codeFrame?: ?DiagnosticCodeFrame,
+  codeFrame?: ?DiagnosticCodeFrame | ?Array<DiagnosticCodeFrame>,
```
devongovett: We could possibly make this always an array if we wanted. I just didn't want to have to change all the tests. Any opinions?

parcel-benchmark:
## Benchmark Results

#### Timings

| --- | --- | --- |

#### Cold Bundles

#### Cached Bundles

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 11.40s | +95.00ms |
| Cached | 505.00ms | -17.00ms |

#### Cold Bundles

*No bundle changes detected.*

#### Cached Bundles

| dist/logo.1e014c76.png | 274.00b | +0.00b | 359.00ms | +30.00ms ⚠️ |

</p></details>

#### Timings

| --- | --- | --- |

#### Cold Bundles

#### Cached Bundles

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 7.93s | -38.00ms |
| Cached | 396.00ms | -14.00ms |

#### Cold Bundles

*No bundle changes detected.*

#### Cached Bundles

*No bundle changes detected.*

</p></details>

**[Click here to view a detailed benchmark overview.](https://benchmark.parceljs.org/benchmark/303324846783201857)**

packages/core/diagnostic/src/diagnostic.js
mischnic: I think `BuildFailureEvent` originally also contained either a diagnostic or an array of them, and it was rather cumbersome to use because you always had to somehow normalize them into an array anyway yourself.

devongovett: Okay, I've refactored it to rename `codeFrame` to `codeFrames` and make it always an array. Also moved `language` and `filePath` into code frames rather than on the top-level diagnostic object.

devongovett:
Turns out `importScripts()` also has the same problem: the specified urls are supposed to resolve relative to the entry worker script, which we cannot know at build time. I've added error diagnostics for this as well. Unfortunately, it means there is no way to import into a worker at all unless you use `{type: 'module'}` since the `new URL` trick won't work due to `import.meta.url` not being supported in scripts.

image

image

DeMoorJasper (approved):
Looks good to me, not totally sure if we should remove filepath if there's no codeframe?

devongovett:
Yeah I wondered the same. It just seemed weird to me to have filePath at the top-level if each code frame can refer to a different file. I think it's still possible to have a code frame with only a file path though, and no code highlights. Not really sure when that would happen though.

PR #6539, opened 2021-07-02
Add nomodule for sibling script tags

### LINKED ISSUE #4304: Generated script tags don’t have `defer`

mischnic:
Fixes #4304.

Previously, this would happen when a sibling (shared) bundle is created with module/nomodule (and a browserslist that includes non-ESM browsers):

```html
<script type="module" src="/a.614f575b.js"></script>  <!-- esmodule sibling -->
<script src="/a.e0e78886.js"></script>  			  <!-- global   sibling, missing nomodule -->
<script type="module" src="/a.d4305cfd.js"></script>  <!-- esmodule entry -->
<script src="/a.e8114d18.js" nomodule=""></script>    <!-- global   entry -->
```

So the sibling to the unused nomodule entry would be loaded unnecssarily.


parcel-benchmark:
## Benchmark Results

#### Timings

| --- | --- | --- |

#### Cold Bundles

#### Cached Bundles

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 10.35s | +380.00ms |
| Cached | 461.00ms | -31.00ms 🚀 |

#### Cold Bundles

| dist/index.js | 486.89kb | +3.00b ⚠️ | 5.68s | -14.00ms |
| dist/logo.1e014c76.png | 274.00b | +0.00b | 213.00ms | -24.00ms 🚀 |

#### Cached Bundles

| dist/index.js | 486.89kb | +3.00b ⚠️ | 5.72s | +130.00ms |

</p></details>

#### Timings

| --- | --- | --- |

#### Cold Bundles

#### Cached Bundles

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 7.20s | -176.00ms |
| Cached | 425.00ms | -8.00ms |

#### Cold Bundles

*No bundle changes detected.*

#### Cached Bundles

| dist/Three.js | 579.69kb | +0.00b | 5.91s | +368.00ms ⚠️ |

</p></details>

**[Click here to view a detailed benchmark overview.](https://benchmark.parceljs.org/benchmark/304409550200504898)**

packages/packagers/html/src/HTMLPackager.js
```diff
@@ -163,21 +163,20 @@ function insertBundleReferences(siblingBundles, tree) {
...
     } else if (bundle.type === 'js') {
       bundles.push({
         tag: 'script',
         attrs: {
           type: bundle.env.outputFormat === 'esmodule' ? 'module' : undefined,
-          src: urlJoin(
-            nullthrows(bundle.target).publicUrl,
-            nullthrows(bundle.name),
-          ),
+          nomodule:
```
devongovett: Shouldn't this only be true if there is a corresponding module? (Not sure how to determine that)

mischnic: Maybe determine the main bundle in the bundle group that `bundle` belongs to, check if the main bundle's script tag has `nomodule`, and add `nomodule` to `bundle` in that case?

mischnic:
I noticed that https://github.com/parcel-bundler/parcel/issues/4304 is actually the same issue (just with `defer`/`async` instead of `nomodule`). So we need a proper solution to propagate these HTML attributes anyway

packages/packagers/html/src/HTMLPackager.js
devongovett: I think the module and nomodule scripts all get placed in the same bundle group with the HTML...

devongovett: We may need to add an `asset.meta.nomodule` flag, and check the entry asset for that. Though not sure what about shared bundles... I guess we wouldn't want to make a shared bundle containing both `nomodule` and non-nomodule (i.e. still global output format but not from a nomodule source), but idk if this is even possible.

devongovett:
We should probably add `defer` for `nomodule` scripts as well so they have the same semantics as the original `type="module"` script.

devongovett:
So:

- `<script async>` should have `isBundleSplittable: false` added to it so it cannot be split into a shared bundle. This is required given that async scripts can execute in any order.
- Normal `<script>` should also have `isBundleSplittable: false` so that any injected runtimes they contain cannot be shared with `<script nomodule>`.
- This should make it safe to add `nomodule` to all non-esm shared bundles injected in the packager I think (?).
- We should also make all `nomodule` scripts, created by both the transformer and packager, have `defer` so their semantics match the `type="module"` version which have `defer` by default.

devongovett (approved):

PR #6587, opened 2021-07-12
Fix bug where stylusrc imports are not being hot reloaded

### LINKED ISSUE #5989: Changes in stylusrc imported files not being picked up
# 🐛 bug report

I have a stylus file that I use to declare variables / functions. I include that file in the `imports` section of my stylusrc, so it's included automatically at the top of each styl file. I noticed that changes to this file don't trigger reloads on my server. Also I noticed that the contents of these files are not considered when looking up cached compilations.

Vanilla parcel configuration, but here is my .stylusrc.js

```js
const path = require('path')

module.exports = {
  "imports": ["global.styl"],
  "paths": [path.join(__dirname, "src")]
}
```

CLI command `parcel index.html --open`

## 🤔 Expected Behavior

Changes in stylus global includes should trigger reloads
Changes in stylus global includes should factor into which cached compilation gets served.

## 😯 Current Behavior

Changes to stylus files included this way don't reload my dev server.
Also currently, the cache isn't operating as expected. I can, say, add `html { color: red; }` in my config-included file. I can see this change if I then make a change to a stylus file that appears in the normal dependency graph (say, one pulled in by my js / html). I can just add a comment or something. If I then delete that comment, the page reloads as if a recompilation has occurred, but it will have reverted my red text.

Will include an example repo below.

## 💁 Possible Solution

I notice in the stylus transformer that we're doing some work with the paths. Maybe we can also iterate through the imports configuration and add them to the watchlist / dependency tree? (I'm not super familiar with the source, sorry if that's not helpful)

## 🔦 Context

Just trying to use the dev server. Having to keep restarting the server to get my updates.

## 💻 Code Sample

Here's a minimum recreation
https://github.com/conlanpatrek/parcel-stylus-bug

## 🌍 Your Environment

| Parcel           | 2.0.0-nightly.623
| Node             | v14.15.4
| Yarn         | 1.22.10
| Operating System | macOS 10.14.6 (I know)

conlanpatrek:
# ↪️ Pull Request

Resolves #5989

This PR adds global imports (declared in the stylusrc) to the Stylus Transformer's dependency list.

The comments in the issue suggested we just run the imports through `asset. addIncludedFile()` (now `asset.invalidateOnFileChange()`), but that wasn't so simple. The global imports are resolved in stylus as though they are any `@import` statement at the top of each file, so in some cases, we'll need to resolve the entries to a file path the "stylus" way.

This resolves both the caching problem _and_ the refresh problem mentioned in the issue (stale caches were possible because these global files were not part of the cache key hash I guess).

## 🚨 Test instructions

I added imports into the integration test suite (into the stylus-deps test) to add a little extra safeguards, but it does not directly test that the dev server hot reloads when changes happen.

Here is the minimum recreation of the bug, along with steps to recreate the issue yourself.
https://github.com/conlanpatrek/parcel-stylus-bug

tl:dr;
1. Create a stylusrc with an imports option that imports a file
2. Start the dev server
2. Make a change to that imported file
3. Assert the dev server automatically updates the page

## ✔️ PR Todo

### DIFF
--- a/packages/core/integration-tests/test/integration/stylus-deps/.stylusrc.js
+++ b/packages/core/integration-tests/test/integration/stylus-deps/.stylusrc.js
@@ -1,3 +1,4 @@
 module.exports = {
-  paths: [__dirname + '/deps']
+  paths: [__dirname + '/deps', 'global'],
+  imports: ['b']
 };
--- a/packages/core/integration-tests/test/integration/stylus-deps/global/b.styl
+++ b/packages/core/integration-tests/test/integration/stylus-deps/global/b.styl
@@ -1,0 +1,2 @@
+.b
+  color blue
--- a/packages/core/integration-tests/test/stylus.js
+++ b/packages/core/integration-tests/test/stylus.js
@@ -56,6 +56,7 @@ describe('stylus', function() {
     let css = await outputFS.readFile(path.join(distDir, 'index.css'), 'utf8');
     assert(css.includes('.index'));
     assert(css.includes('.a'));
+    assert(css.includes('.b'));
     assert(css.includes('-webkit-box'));
     assert(css.includes('.foo'));
   });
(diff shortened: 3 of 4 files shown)


devongovett (approved):
Great work, thanks!

PR #6595, opened 2021-07-14
Add SVG transformer

thewilkybarkid:
# ↪️ Pull Request

Adds `@parcel/transformer-svg`, starting with support for the `xml-stylesheet` processing instructions and (`xlink:`)`href` on `<a>` and `<use>`.

Relates to #2901 and replaces #6535.

Close T-1093

## 💻 Examples

A HTML file containing:

```html
<object data="image.svg"></object>
```

`image.svg`:

```svg
<?xml-stylesheet href="style.css"?>
<svg xmlns="http://www.w3.org/2000/svg">
  <rect width="100" height="100"/>
</svg>
```

`style.css`:

```css
:root {
  fill: red;
}
```

The square should appear red rather than black.

## 🚨 Test instructions

Includes integration tests.

## ✔️ PR Todo


thewilkybarkid:
I think this is ready and resolves some of the comments mentioned in #6535.

There's clearly a lot more features to add; handling `<style>` elements is probably the priority, but this will require a packager.

packages/transformers/svg/src/SVGTransformer.js
```diff
@@ -0,0 +1,47 @@
+// @flow
+
+import {Transformer} from '@parcel/plugin';
+import nullthrows from 'nullthrows';
+import semver from 'semver';
+import collectDependencies from './dependencies';
+import parse from 'posthtml-parser';
+import render from 'posthtml-render';
```
devongovett, line 8: We should probably use an actual XML parser for SVG, not an HTML parser. I think there are some subtle differences.

thewilkybarkid: Despite the name it is a HTML _and_ XML tool (as is htmlparser2).

Hadn't thought of using it initially (hence JSDOM), but it does the job here and has the benefit of already being used.

packages/transformers/svg/src/SVGTransformer.js
```diff
@@ -0,0 +1,47 @@
...
+      version: '0.4.1',
+      program: parse(await asset.getCode(), {
+        directives: [
+          {
+            name: /^\?/,
+            start: '<',
+            end: '>',
+          },
+        ],
+        lowerCaseAttributeNames: true,
```
devongovett: I think this should be false? SVG attributes are case sensitive. They are only case insensitive when embedded in an HTML file, but that should already be handled by the HTML transformer.

DeMoorJasper (approved):

packages/configs/default/index.json
```diff
@@ -40,6 +40,7 @@
     "style:*.vue": ["@parcel/transformer-vue"],
     "custom:*.vue": ["@parcel/transformer-vue"],
     "*.{png,jpg,jpeg,webp}": ["@parcel/transformer-image"],
+    "*.svg": ["@parcel/transformer-svg"],
```
mischnic, line 43: This means that `import img from "./logo.svg";` doesn't cause a build error anymore (currently it needs to be `from "url:./logo.svg"`). But at the moment, there's nothing we can do about that

devongovett (approved):

Issue #4585, opened 2020-05-08
How do you minify and process your images with Parcel?


sergiojunene:
Is there any way to process our images or media files with this excellent bundler? I'm using right now `imagemin` and its plugin but I don't know if I'm doing it right. I've this warning on my terminal. 

![screenshot 2020-05-08 a las 21 47 15](https://user-images.githubusercontent.com/22520811/81443436-ae518280-9175-11ea-8110-4275bb0074bf.jpg)

mischnic:
You mean `parcel-plugin-imagemin`? Are you sure that warning is caused by that plugin?
@DeMoorJasper is the author

DeMoorJasper:
I doubt this issue is caused by the parcel-plugin-imagemin plugin.

This error occurs when you try to use `readFileSync` with anything that cannot be statically analyzed, meaning using anything that isn't constant like a variable in the filepath.

sergiojunene:
Yes, is exactly that @DeMoorJasper but this warning is caused by imagemin on async await anonymous function. But I'd like how can I work with images using Parcel?

sergiojunene:
I didn't used `parcel-plugin-imagemin` but just only `imagemin` with its plugins

DeMoorJasper:
@sergiojunene I guess you could share some of your code so we can reproduce this or give some guidance. `how can I work with images using Parcel` is pretty much out of the box, just require an image and it'll return a url, using the imagemin plugin it would minify it...

sergiojunene:
Yeah, I know, but how do I configure on `package.json` my scripts, the entrypoint on parcel must be a file that not an image. I don't have a code yet, so I want to know how it's done.

DeMoorJasper:
I really don't see how this is related to parcel in any way. You're asking us how you should setup your image pipeline?

I really don't understand what you're asking here. If you want to run a node script to optimize images ahead of time just run it using node... If you want parcel to do it, use a plugin.

sergiojunene:
Yes, which is the code for images pipeline

mischnic:
> Yes, which is the code for images pipeline

The question is: what do you want to achieve? 
- Get the URL to an image from JS? `import foo from "./logo.png"; console.log(foo);`
- Include a image in HTML? `<img src="logo.png">`

The warning you posted in the beginning is just that, a warning. Usually it means that you didn't actually mean to bundle that module because it won't work as expected in the browser.

sergiojunene:
[...]

Issue #3739, opened 2019-11-03
Parcel 2: Compression and Image Optimization


devongovett:
This is an epic level RFC to summarize a project to support compression and image optimization in Parcel 2. The details of each specific plugin will be contained in their own tickets, linked below.

## Overview

Parcel is a build tool for your entire application. Unlike other bundlers, it includes native support for all of your code and assets, not just your JavaScript. Because of this, we can perform a lot more application wide optimizations that other tools cannot. One of these areas we can support is automatic compression and image optimization across your whole app, regardless of where your images are referenced from (e.g. HTML, CSS, JS, etc.). We are well positioned to provide best practice optimizations when building for production completely automatically.

## Text asset compression

Text assets like HTML, CSS, JS, and SVG all benefit from compression with formats like gzip and Brotli. We should support plugins to let users enable this for their applications. Sometimes this is automatically handled by servers, but other times, such as when uploading to a file storage service like S3, it needs to be done in advance as part of the build. For this reason, we won't include these compression plugins in the default config, but it's fairly easy for users to enable them via `.parcelrc`.

```json
{
  "extends": "@parcel/config-default",
  "optimizers": {
    "*.{js,css,html,svg}": ["...", "@parcel/optimizer-gzip"]
  }
}
```

The `"..."` syntax ensures that any other optimizers defined in `@parcel/config-default` for JS, CSS, etc. (e.g. Uglify) are run prior to gzipping.

Specific issues:

* [@parcel/optimizer-gzip](https://github.com/parcel-bundler/parcel/issues/3731)
* [@parcel/optimizer-brotli](https://github.com/parcel-bundler/parcel/issues/3732)

## Image compression and resizing

Images can be some of the largest assets on a website, so it is imperative that they are resized to their displayed size and optimized prior to serving them to users. Parcel is in a great position to do this automatically as we already process all of the source code that would reference images, e.g. HTML, CSS, JS, etc. We should support plugins to optimize various image formats with best of breed image optimization tools, along with plugins to resize images to display sizes automatically.

Issues for image optimizers:

* [@parcel/optimizer-mozjpeg](https://github.com/parcel-bundler/parcel/issues/3733)
* [@parcel/optimizer-pngquant](https://github.com/parcel-bundler/parcel/issues/3734)
* [@parcel/optimizer-svgo](https://github.com/parcel-bundler/parcel/issues/3735)

Note: these are just a few to start with. Others could be added later.

[...]

Issue #627, opened 2018-01-23
Image optimization


windware:
🙋 **Feature request**

Is it possible to add image optimizers?
I assume [webpack's image loader](https://github.com/tcoopman/image-webpack-loader) (or [another](https://www.npmjs.com/package/imagemin-webpack-plugin)) does it.

### 🤔 Expected Behavior

When importing images, they will be piped through image optimizers.

### 😯 Current Behavior

Not optimized.

### 🔦 Context

Possible option can be,

--optimize-image [quality] (If quality is not specified, does it losslessly)

### 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Parcel           | 1.4.1
| Node             | 8.9.4
| npm         | 5.6.0
| Operating System | Ubuntu 16.04.3

DeMoorJasper:
Should this be a plugin or a core feature? I've not seen that many people request it so far. I'm intrested in writing it though

windware:
If it goes in the core, maybe it shouldn't be enabled by default as it takes a while to optimize a large amount of images, giving wrong impressions for first time users but it could be handy if it's built in and users opt-in for it.

brandon93s:
I think this fits best as a plugin. Seems very useful, but not everyone will want their images altered by their bundler.

DeMoorJasper:
Initial version of the plugin is ready here: https://github.com/DeMoorJasper/parcel-plugin-imagemin

windware:
I just ran the example and it showed an error at the jpeg file,

🚨  Error in file: parcel-plugin-imagemin/example/pexels-image.jpeg

write EPIPE
write EPIPE
    at _errnoException (util.js:1022:11)
    at WriteWrap.afterWrite [as oncomplete] (net.js:867:14)

Also, I ran it on my own files and it stopped with this, (I placed the imagemin.config.js at the project root where package.json is and also placed it at the directory where the entry point file is.)

🚨  asset.getConfig is not a functio
    at ImagePackager.addAsset (node_modules/parcel-plugin-imagemin/lib/ImagePackager.js:22:30)
    at Bundle._addDeps (node_modules/parcel-bundler/src/Bundle.js:106:20)
    at Bundle._package (node_modules/parcel-bundler/src/Bundle.js:89:18)
    at <anonymous>

DeMoorJasper:
@windware i'll close down this issue, feel free to open any bug issues on https://github.com/DeMoorJasper/parcel-plugin-imagemin

Issue #6444, opened 2021-06-10
transformer-image automatically rotates 3:4 ratio images


Shaokun-X:
# 🐛 bug report

I have a bunch of images that I use Parcel/transformer-image to resize, I found that all the images of 3:4 ratio were automatically rotated 90 degrees thus became of 4:3 ratio. And I didn't find any way to configure this behavior.

## 🎛 Configuration (.babelrc, package.json, cli command)

`package.json`:
```js
  "scripts": {
    "build": "parcel build index.html",
    "dev": "parcel index.html --open",
    "start": "npm run build && npm run dev"
  },
  "devDependencies": {
    "@parcel/packager-raw-url": "^2.0.0-beta.3.1",
    "@parcel/transformer-image": "^2.0.0-beta.3.1",
    "@parcel/transformer-sass": "^2.0.0-beta.3.1",
    "@parcel/transformer-webmanifest": "^2.0.0-beta.3.1",
    "parcel": "^2.0.0-beta.3.1",
    "posthtml-include": "^1.7.1",
    "sharp": "^0.28.3"
  }
```

Dev server:
```
$ npm run start
```

## 🤔 Expected Behavior

Converting images without rotation.

## 😯 Current Behavior

Converting all and only 3:4 images with 90 degrees rotation.

## 💻 Code Sample

I used HTML:
```html
<picture>
  <source srcset="url:./img/m1.jpg?as=webp&width=1280" type="image/webp">
  <img src="url:./img/m1.jpg?width=1280">
  <source srcset="url:/img/m1.jpg">
</picture>
```
Here `m1.jpg` is of 3024 × 4032 (3:4), the converted image became 1280 × 960 (4:3).

## 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Parcel           | 2.0.0-beta.3.1 |
| Node             | 14.15.4 |
| npm/Yarn         | 6.14.10 |
| Operating System | Windows 10 AMDx64

Shaokun-X:
I haven't tested 4:3 images

DeMoorJasper:
We need to do a rotate operation on the images to ensure the correct rotation https://sharp.pixelplumbing.com/api-operation#rotate 

If anyone wants to implement this feel free to do so, otherwise I'll look into this tomorrow

LironHazan:
Will adding rotate() to ImageTransformer be enough? :)

```
      if (width || height) {
        imagePipeline.rotate().resize(width, height);
      }
```

PR #6624, opened 2021-07-21
Add image optimizer

thewilkybarkid:
# ↪️ Pull Request

Adds an image optimizer using ~[Squoosh](https://squoosh.app/) or~ Sharp.

Has reasonably safe defaults, and supports a config file.

Refs https://github.com/parcel-bundler/parcel/discussions/5544, https://github.com/parcel-bundler/parcel/issues/4585, https://github.com/parcel-bundler/parcel/issues/627, https://github.com/parcel-bundler/parcel/issues/3739 and others.

## ✔️ PR Todo


packages/core/integration-tests/test/image.js
```diff
@@ -88,4 +88,20 @@ describe('image', function() {
...
+      },
+    });
+
+    const imagePath = b.getBundles().find(b => b.type === 'jpg').filePath;
+
+    const buffer = await outputFS.readFile(imagePath);
+    const image = await sharp(buffer).metadata();
+
+    assert.strictEqual(image.width, 1920);
+    assert.strictEqual(image.size, 282274);
```
thewilkybarkid: Works on my machine. 😅

packages/optimizers/squoosh/src/SquooshOptimizer.js
```diff
@@ -0,0 +1,43 @@
...
+
+export default (new Optimizer({
+  async optimize({bundle, contents}) {
+    if (!bundle.env.shouldOptimize) {
+      return {contents};
+    }
+
+    if (!imagePool) {
+      imagePool = new ImagePool();
+    }
```
thewilkybarkid: There doesn't seem to be a way to set-up/tear-down the optimizer.

devongovett:
How does this compare with the existing image transformer, which uses sharp?

thewilkybarkid:
Not actually checked, but it doesn't (and probably shouldn't?) set the mozjpeg option for example so would definitely be bigger.

thewilkybarkid:
Tried out Squoosh based on https://github.com/parcel-bundler/parcel/discussions/5544#discussioncomment-252279, but after a bit of a search have found https://github.com/vercel/next.js/issues/27055 which is concerning. Maybe a Sharp optimiser would be better.

DeMoorJasper:
Dont think sharp can really be used as an image optimizer, as far as I've tried tools like imagemin yield way better results size wise. Sharp is great for transforming images though which libraries like imagemin and squoosh only do the bare minimum of. We can definitely experiment with sharp and see how far we can take it.

thewilkybarkid:
https://github.com/vercel/next.js/discussions/17141 has some useful comparisons (not Squoosh though).

thewilkybarkid:
At least for that one image, using the same quality value as Squoosh's default in Sharp produces the same result. (Actually 10 bytes smaller.)

packages/optimizers/sharp/src/SharpOptimizer.js
```diff
@@ -0,0 +1,39 @@
+// @flow
+
+import {Optimizer} from '@parcel/plugin';
+import {blobToBuffer} from '@parcel/utils';
+import path from 'path';
+import sharp from 'sharp';
+
+const OPTIONS = {
+  jpeg: {
+    quality: 75,
```
thewilkybarkid: To match https://github.com/GoogleChromeLabs/squoosh/blob/2bf0b904cd3ebf831c9b47e6f59ab508ae968f81/src/features/encoders/mozJPEG/shared/meta.ts#L23 (rather than proposing it as a new default).

DeMoorJasper:
If sharp produces similar results size wise, maybe we should change some of the defaults in the image transformer and not run sharp a second time in the optimize stage, not sure how much difference this makes but I imagine it's gonna be a bit slower

thewilkybarkid:
That assumes there was a transformation...

thewilkybarkid:
Going to try other formats: if I understand correctly Sharp will be bigger as it doesn't use an optimiser.

The other option is imagemin, but I've only just seen that it's now marked as unmaintained so isn't suitable.

thewilkybarkid:
Couldn't get Squoosh to reduce the image anywhere near as much as Sharp with the `palette` option.

thewilkybarkid:
Given Squoosh is still reasonably experimental, maybe the way forward is to just use Sharp. The double pass could be removed if it's possible to know whether the transformer has already processed the image?

thewilkybarkid:
Or just run every image through the transformer?

DeMoorJasper:
@thewilkybarkid it's probably fine to keep it separate, also just realised we can't access shouldOptimize in the transformer so we should run the transformer for each image if we'd move this which will probably be worse than just running the optimiser.

Pretty nice how good sharp is with optimisation, didn't know that, thanks for researching all these optimisation options

packages/optimizers/sharp/src/SharpOptimizer.js
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    const ext = path.extname(bundle.name).substr(1);
+    const format = FORMATS.get(ext);
+
+    if (!format) {
+      throw new Error(`Sharp does not support ${ext} images.`);
+    }
+
+    const optimized = await sharp(await blobToBuffer(contents))
```
DeMoorJasper: Not sure if we should add `.rotate()` to this as well to normalise the rotation metadata of images. We used to have an issue in the transformer where images would be rotated if they have a certain format, might also be caused by resizing not sure.

devongovett:
I'm a little conflicted about whether image optimization should be opt in or not. If it's always lossless in terms of quality I guess that's ok, but if it's lossy then I think it would be potentially unexpected. Do we need some way to configure that?

thewilkybarkid:
I was thinking of reading a Sharp config file, which has the output settings for each file type. As for defaults, could be quite conservative (i.e. Sharp's defaults).

packages/optimizers/sharp/src/SharpOptimizer.js
thewilkybarkid: Just seen #6444; looks like it shouldn't happen, but maybe worth having a test with the EXIF Orientation tag (more for the transformer though).

thewilkybarkid: Will expand on https://github.com/parcel-bundler/parcel/pull/6647.

thewilkybarkid: Turns out it _did_ happen, so I've solved it in https://github.com/parcel-bundler/parcel/pull/6624/commits/28fa471800b77b7e1f1e539e6c3bd1ae9a93b466.

packages/optimizers/sharp/src/SharpOptimizer.js
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    const options = {...DEFAULT_OPTIONS[format], ...config?.[format]};
+
+    const optimized = await sharp(await blobToBuffer(contents))
+      [format](options)
+      .rotate()
+      .toBuffer();
+
+    return {contents: optimized};
```
DeMoorJasper: Should we add a check to ensure the size is smaller and return the original if that's smaller?

thewilkybarkid: Hmm, technically you could just turn on making images progressive which would make them a bit bigger.

DeMoorJasper: Why would you want to make them bigger? I meant just checking if the output buffer is smaller than the input buffer and return the input buffer if that is smaller as sometimes input images are already optimised a lot

yarn.lock
```diff
@@ -2137,6 +2182,36 @@
     detect-libc "^1.0.3"
     globby "^11.0.3"
 
+"@parcel/types@2.0.0-beta.3.1":
+  version "2.0.0-beta.3.1"
+  resolved "https://registry.yarnpkg.com/@parcel/types/-/types-2.0.0-beta.3.1.tgz#0bc5255a4454c198d58403b84a66f059106e1610"
+  integrity sha512-8Cmh5/43LZ7wQeqawFMenZ/CCh+YAbyRuwBBApMCDLU3+RSpajUYyTpqa1HZVV8FggQb8E0xvPkmUbj/UHz/gw==
+
+"@parcel/utils@2.0.0-beta.3.1":
+  version "2.0.0-beta.3.1"
```
thewilkybarkid: Not sure why Yarn is showing these as I bumped the versions to `rc.0` in in https://github.com/parcel-bundler/parcel/pull/6624/commits/8dfd3d7754dc0ef4a81c982d0b89f44f24074b02?

thewilkybarkid:
```console
$ yarn list --pattern @parcel/utils
yarn list v1.22.10
├─ @parcel/transformer-js@2.0.0-rc.0
│  └─ self-published@2.0.0-beta.3.1
│     └─ @parcel/utils@2.0.0-beta.3.1
└─ @parcel/utils@2.0.0-rc.0
✨  Done in 1.00s.
```

thewilkybarkid: Unrelated changes; happens on my `v2` branch locally too.

DeMoorJasper (approved):

mischnic:
> I'm a little conflicted about whether image optimization should be opt in or not. If it's always lossless in terms of quality I guess that's ok, but if it's lossy then I think it would be potentially unexpected. Do we need some way to configure that?

So right now, it processes images by default with Sharp's default settings. Are those lossless (and guaranteed to not make the image bigger)?

DeMoorJasper:
@mischnic it should be pretty much lossless and there's a check to fallback to the original image if the original is smaller (which is also what imagemin does)

~Just had a look at the sharp docs and by default `webp`, `avif` and `heif` are set to lossless: true, should we change those to false in our default configs?~ apparently I messed up or they changed it

Also did a small test to test the performance of sharp w/ and without compressions and compressing is about 3-4 times slower than without compression so my original idea of moving this to the transformer is a bad one

devongovett:
I don't think it is lossless. Even decoding and re-encoding a JPEG is a lossy operation (JPEG encoding is inherently lossy). And the current defaults in this PR use indexed PNG which would definitely be lossy (only 256 colors). There _are_ lossless image compression libraries out there (e.g. jpegtran and optipng), but sharp doesn't appear to use them.

> Just had a look at the sharp docs and by default webp, avif and heif are set to lossless: true

I don't think this is true, the defaults listed [here](https://sharp.pixelplumbing.com/api-output) are false.

Again, I'm not sure this should be on by default if it's not fully lossless. Especially if there's no way to opt out for only certain images.

Perhaps we should leave this to the image transformer so users can opt-into image optimization via query params for each image. Using the `as` or `quality` query param is basically equivalent to what is in this optimizer already. We could probably enable `mozjpeg` as well, and potentially expose more options as well.

We could also have a global configuration to enable the image transformer for all images potentially. This would allow people to opt into image optimization for all images without manually adding query params to every image.

To sum up, I think it's too risky to enable lossy image optimization by default without people knowing and no way to opt-out per image. Plus, the image transformer already does most of this, and this would reduce performance because images would get processed twice.

devongovett:
I used a couple Rust libraries to build an image optimizer that does lossless compression in https://github.com/parcel-bundler/parcel/pull/6762. Perhaps we can also offer a config file to enable the image transformer on all images to opt into lossy compression without manually adding query params. What do you think?

thewilkybarkid:
Yeah, that would work. I understand the hesitancy over making it opt-out rather than opt-in, but some config to turn it on globally would be enough.

devongovett:
Ok. I'm going to close this then since #6672 is merged. If you want to send a PR for a config file for the image transformer, feel free.

PR #6646, opened 2021-07-27
Add prerender transformer

aminya:
# ↪️ Pull Request

This adds a [Prerender](https://www.npmjs.com/package/@prerenderer/prerenderer) transformer.
Ported from [the deprecated Parcel 1 plugin](https://github.com/ABuffSeagull/parcel-plugin-prerender).

## ✔️ PR Todo

### DIFF
--- a/packages/transformers/prerender/src/PrerenderTransformer.js
+++ b/packages/transformers/prerender/src/PrerenderTransformer.js
@@ -1,0 +1,72 @@
+import {promises as fs} from 'fs';
+import path from 'path';
+import Prerenderer from '@prerenderer/prerenderer';
+import Puppeteer from '@prerenderer/renderer-puppeteer';
+import {Transformer} from '@parcel/plugin';
+
+export default (new Transformer({
+  async loadConfig({config, options}) {
+    let routes = ['/']; // the default route
+    let rendererConfig = {};
+
+    let userConfig = await config.getConfigFrom(
+      path.join(options.projectRoot, 'index'),
+      [
+        '.prerender.json',
+        '.prerender.config.json',
+        '.prerenderrc.json',
+        '.prerenderrc',
+        '.prerender.config.js',
+      ],
+    );
+
+    if (userConfig) {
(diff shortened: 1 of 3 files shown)


packages/transformers/prerender/src/PrerenderTransformer.js
```diff
@@ -0,0 +1,72 @@
...
+      if (isJavascript) {
+        config.invalidateOnStartup();
+      }
+    }
+
+    return {routes, rendererConfig};
+  },
+  async transform({options, config}) {
+    const {routes, rendererConfig} = config;
+    const outDir = options.outDir || path.resolve('./dist'); // TODO removed in parcel 2
```
aminya, line 47: I couldn't find a replacement for this in the new API

packages/transformers/prerender/src/PrerenderTransformer.js
```diff
@@ -0,0 +1,72 @@
+import {promises as fs} from 'fs';
+import path from 'path';
+import Prerenderer from '@prerenderer/prerenderer';
+import Puppeteer from '@prerenderer/renderer-puppeteer';
+import {Transformer} from '@parcel/plugin';
+
+export default (new Transformer({
```
aminya, line 7: This transformer should run when everything is bundled. Not sure how to force that using the current API. The output of this transformer should then be fed to htmlnano, posthtml, etc.

mischnic: > This transformer should run when everything is bundled 

It has to be an optimizer (or packager) for that, this would make it run before the existing htmlnano optimizer:

```json
{
	"optimizers": { "*.html": ["your-thing", "..."] }
}
```

aminya: The optimizer doesn't support returning multiple assets. How can we use that?

mischnic:
So there were some technical problems, and this probably also makes more sense as a third-party plugin anyway.

PR #6445, opened 2021-06-11
added rotate() fn to the sharp img pipeline (#6444)

LironHazan:
suggestion to resolve --> https://github.com/parcel-bundler/parcel/issues/6444

PR #6647, opened 2021-07-27
Retain image EXIF data in the transformer

thewilkybarkid:
# ↪️ Pull Request

Following https://github.com/parcel-bundler/parcel/pull/6624#discussion_r677170694 I created a test for #6445.

Turns out that the transformer strips EXIF data, which probably isn't meant (should be in https://github.com/parcel-bundler/parcel/pull/6624).

So, this keeps EXIF data, as well as testing the consumption of the image orientation.

The test image comes from https://www.galloway.me.uk/2012/01/uiimageorientation-exif-orientation-sample-images/

## 💻 Examples

See https://github.com/parcel-bundler/parcel/issues/6444.

## ✔️ PR Todo


DeMoorJasper (approved):

devongovett (approved):

devongovett:
Seems ok, though not exactly sure what the use case is. The image optimizer should strip this for prod builds.

mischnic:
If you have an image that is rotated via EXIF (not in the actual image data), then applying the image transformer would previously output an image that has a different orientation than the original.

DeMoorJasper:
> If you have an image that is rotated via EXIF (not in the actual image data), then applying the image transformer would previously output an image that has a different orientation than the original.

@mischnic that should already be handled by the `image.rotate()` in the image transformer

thewilkybarkid:
> Seems ok, though not exactly sure what the use case is.

I was just adding a missing test case, wasn't expecting to see the transformer strip EXIF data (so only images that are resized etc lose it).

> The image optimizer should strip this for prod builds.

Yep, I've added another test case for this. (Also stopped the optimizer from running on non-prod builds too.)

devongovett:
Sorry I missed this was getting merged. Should we check that metadata doesn't get added in production builds? The optimizer should strip it out, but only for JPEG and PNGs.

KevinDillaerts:
Hello all, I seem to have a similar issue still running Parcel 2.
Below are the versions I'm running:

`"devDependencies": {
    "@parcel/packager-raw-url": "^2.0.1",
    "@parcel/transformer-sass": "^2.0.1",
    "@parcel/transformer-webmanifest": "^2.0.1",
    "parcel": "^2.0.1",
    "posthtml-include": "^1.7.1",
    "sharp": "^0.29.3"
  },`

I have some images in portrait mode that work fine on localhost, however after build when running serve these are rotated.
In the out folder I can see these files have been rotated, so I assume this is related to the issue with the transformer as well.
This is the localhost result:
localhost

And this is from 'serve':
serve

Here are the images in question, the first one is the original file I have in my src folder, the second what ends up in the out folder:
riks_cafe_004(src folder)
riks_cafe_004(out folder)


Other portrait mode images I'm using work fine, so this seems to be related to these aspect ratios.
I did resize these earlier running an Automator script I use where the longest side gets scaled to 1000px.
This works fine for the other images on the site, as these were all running from the same script.

PR #6664, opened 2021-07-31
Handle mutations of process.env

### LINKED ISSUE #5493: Broken output for `delete` on an process.env property
# 🐛 bug report

Don't replace `delete process.env.SOMETHING` with `delete undefined;`

## 🤔 Expected Behavior

Remove the whole statement or replace with an empty statement.

## 😯 Current Behavior

`delete undefined;` is invalid in strict mode. This is even verified by terser.

## 💁 Possible Solution

`process.env.*` is inlined here: https://github.com/parcel-bundler/parcel/blob/a4cf3e958ea1136b4829eb93615fba5328213471/packages/transformers/js/src/visitors/process.js#L26

This special case is described by `t.isUnaryExpression(path.parent, {operator: 'delete', argument: node})`.

devongovett:
Fixes #5493. Fixes T-1036.

Mutating process.env properties can't be supported if we still want to statically replace them with their values at build time. In source code this is flagged as an error. In node_modules it is a warning in case the code isn't actually used at runtime. In that case, the behavior won't be as expected but at least it will compile.

In order to do this, the diagnostics in Rust now have a severity property, which can be `Error`, `Warning`, or `SourceError`. `SourceError` means error if `asset.isSource` and warning otherwise.

Also fixed an error message for script tags which had a copy/paste mistake I noticed.

### DIFF
--- a/packages/core/integration-tests/test/integration/env-computed/index.js
+++ b/packages/core/integration-tests/test/integration/env-computed/index.js
@@ -1,3 +1,2 @@
 let name = "ABC";
-process.env[name] = "abc";
 module.exports = process.env[name];
--- a/packages/core/integration-tests/test/html.js
+++ b/packages/core/integration-tests/test/html.js
@@ -1242,9 +1242,7 @@ describe('html', function() {
               ],
             },
           ],
-          hints: [
-            'Add type="module" as a second argument to the <script> tag.',
-          ],
+          hints: ['Add the type="module" attribute to the <script> tag.'],
         },
       ]);
 
@@ -1523,9 +1521,7 @@ describe('html', function() {
               ],
             },
           ],
-          hints: [
-            'Add type="module" as a second argument to the <script> tag.',
-          ],
+          hints: ['Add the type="module" attribute to the <script> tag.'],
         },
       ]);
 
--- a/packages/core/integration-tests/test/integration/env-mutate/index.js
+++ b/packages/core/integration-tests/test/integration/env-mutate/index.js
@@ -1,0 +1,4 @@
+process.env.SOMETHING = "foo";
+process.env.SOMETHING += "foo";
+delete process.env.SOMETHING;
+process.env.SOMETHING++;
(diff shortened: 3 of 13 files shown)


packages/core/integration-tests/test/integration/env-computed/index.js
```diff
@@ -1,3 +1,2 @@
 let name = "ABC";
-process.env[name] = "abc";
```
devongovett, line 2: Do we still want to allow mutations with computed properties?? Seems weird to allow that but not static properties.

mischnic: It will always be possible to mutate them anyway by assigning `process.env` to a variable and then changing that.

parcel-benchmark:
## Benchmark Results

#### Timings

| --- | --- | --- |

#### Cold Bundles

#### Cached Bundles

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 9.75s | -9.00ms |
| Cached | 490.00ms | +1.00ms |

#### Cold Bundles

*No bundle changes detected.*

#### Cached Bundles

*No bundle changes detected.*

</p></details>

#### Timings

| --- | --- | --- |

#### Cold Bundles

#### Cached Bundles

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 7.23s | -49.00ms |
| Cached | 409.00ms | -27.00ms 🚀 |

#### Cold Bundles

*No bundle changes detected.*

#### Cached Bundles

*No bundle changes detected.*

</p></details>

**[Click here to view a detailed benchmark overview.](https://benchmark.parceljs.org/benchmark/305669281499054658)**

packages/core/integration-tests/test/integration/env-computed/index.js
devongovett: Sure, but the process.env object won't have any properties besides the ones added at runtime, so it's not great to rely on that anyway. IMO better to be consistent when it's not assigned to a variable unless there's a really good reason to allow this.

mischnic (approved):

DeMoorJasper (approved):

PR #6544, opened 2021-07-04
Transform Graphviz files

thewilkybarkid:
# ↪️ Pull Request

Transforms Graphviz files (`.gv` or the alternative `.dot`) into SVGs.

## 💻 Examples

With a HTML file containing:

```html
<object data="graph.gv"></object>
```

and a Graphviz file containing:

```gv
digraph graphname {
    a -> b -> c;
    b -> d;
}
```

which is turned into an SVG files like:

```svg
<?xml version="1.0" encoding="UTF-8" standalone="no"?>
<!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN"
 "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd">
<!-- Generated by graphviz version 2.47.3 (20210621.0720)
 -->
<!-- Title: graphname Pages: 1 -->
<svg width="134pt" height="188pt"
 viewBox="0.00 0.00 134.00 188.00" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
<g id="graph0" class="graph" transform="scale(1 1) rotate(0) translate(4 184)">
<title>graphname</title>
<polygon fill="white" stroke="transparent" points="-4,4 -4,-184 130,-184 130,4 -4,4"/>
<!-- a -->
<g id="node1" class="node">
<title>a</title>
<ellipse fill="none" stroke="black" cx="63" cy="-162" rx="27" ry="18"/>
<text text-anchor="middle" x="63" y="-157.8" font-family="Times,serif" font-size="14.00">a</text>
</g>
<!-- b -->
<g id="node2" class="node">
<title>b</title>
<ellipse fill="none" stroke="black" cx="63" cy="-90" rx="27" ry="18"/>
<text text-anchor="middle" x="63" y="-85.8" font-family="Times,serif" font-size="14.00">b</text>
</g>
<!-- a&#45;&gt;b -->
<g id="edge1" class="edge">
<title>a&#45;&gt;b</title>
<path fill="none" stroke="black" d="M63,-143.7C63,-135.98 63,-126.71 63,-118.11"/>
<polygon fill="black" stroke="black" points="66.5,-118.1 63,-108.1 59.5,-118.1 66.5,-118.1"/>
</g>
<!-- c -->
<g id="node3" class="node">
<title>c</title>
<ellipse fill="none" stroke="black" cx="27" cy="-18" rx="27" ry="18"/>
[...]

PR #6686, opened 2021-08-05
Export core class in TypeScript

thewilkybarkid:
Working on a small test suite for turning https://github.com/parcel-bundler/parcel/pull/6544 into an independent library, and found the core class isn't currently exported in the TypeScript definitions.

### DIFF
--- a/packages/core/core/index.d.ts
+++ b/packages/core/core/index.d.ts
@@ -2,7 +2,7 @@ import type {InitialParcelOptions, BuildEvent, BuildSuccessEvent, AsyncSubscript
 import type {FarmOptions} from '@parcel/workers';
 import type WorkerFarm from '@parcel/workers';
 
-declare class Parcel {
+export default class Parcel {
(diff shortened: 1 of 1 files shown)


packages/core/core/index.d.ts
```diff
@@ -2,7 +2,7 @@ import type {InitialParcelOptions, BuildEvent, BuildSuccessEvent, AsyncSubscript
 import type {FarmOptions} from '@parcel/workers';
 import type WorkerFarm from '@parcel/workers';
 
-declare class Parcel {
+export default class Parcel {
```
devongovett, line 5: Is the `declare` needed there?

thewilkybarkid: No; declarations need either an `export` or `declare`. I think you can include both, but it's redundant.

DeMoorJasper (approved):

DeMoorJasper (approved):

PR #6696, opened 2021-08-09
Resolver improvements

### LINKED ISSUE #670: Support Web Workers coming from node_modules

### LINKED ISSUE #3492: Parcel 2: Specific resolvers per file type
The way Parcel currently resolves all files is the same across all file types, whether that be JS, CSS, HTML, or anything else, but sometimes it makes sense for there to be differences. For Parcel 2, we are considering making resolvers configurable per file type. Something like this:

```json
"resolvers": {
  "*.{js,mjs,jsm,jsx,es6,ts,tsx,coffee}": ["@parcel/resolver-path", "@parcel/resolver-alias", "@parcel/resolver-node"],
  "*.{css,styl,stylus,sss,sass,scss,less}": ["@parcel/resolver-url", "@parcel/resolver-path", "@parcel/resolver-alias", "@parcel/resolver-node"],
  "*.{htm,html}": ["@parcel/resolver-url", "@parcel/resolver-path", "@parcel/resolver-alias"]
}
```

These would run as pipelines, similar to other plugin types (e.g. transformers, optimizers, etc.).  Each resolver would run, and the result if any would be passed to the next resolver and so on. Allowing resolvers to be configured this way would allow different languages to resolve slightly differently, and for users to adjust smaller aspects of the resolution more easily than rewriting the entire default resolver plugin.

The proposed resolver plugins above are the following:

* `@parcel/resolver-url` - resolves URL specifiers to file paths, e.g. from CSS and HTML.
* `@parcel/resolver-path` - resolves file path specifiers to absolute, e.g. tilde, relative, etc.
* `@parcel/resolver-alias` - resolves aliases defined in package.json
* `@parcel/resolver-node` - resolves node_modules.

The actual resolution algorithms for each language are described in the following sections.

## JavaScript

The JS resolution algorithm is a modified version of the [Node resolution algorithm](https://nodejs.org/api/modules.html#modules_all_together). It also supports Parcel's tilde and absolute paths, aliases, and a few other main fields like `source`, `browser` and `module`. This is essentially the same as how Parcel 1 resolved modules.

One change from Parcel 1 is to begin treating ES module import specifiers are URLs rather than file paths. This means resolving with URL semantics rather than platform specific path semantics, and allowing things like query strings and percent escaping. See also #3477.

Some examples:

```javascript
import 'foo'; // node_modules/foo
import 'foo/bar'; // node_modules/foo/bar.js
import './foo.js'; // ./foo.js
import './foo'; // ./foo.js
import '~/foo'; // <nearest dir with package.json>/foo.js
import '/foo'; // <projectRoot>/foo.js
```

Extensionless specifiers are allowed, but only for the following extensions: `'.js', '.jsx', '.mjs', '.json', '.ts', '.tsx',  '.coffee'`. This is different from Parcel 1, which allowed importing any extension supported by Parcel without specifying it. This was slow since Parcel needed to check for every possible extension as it searched for files. Also in Parcel 2 there is no definitive list of extensions that are supported, as we use globs to match files instead, and I don't think it was really ever expected to be able to import non-JS without specifying the extension anyway. It is mostly for compatibility with the Node CommonJS ecosystem.

There are a few exceptions, where resolution is treated differently. These are for JS features that produce dependencies other than ES modules and CommonJS, which are generally treated as URLs. Examples include workers, service workers, etc.

```javascript
new Worker('./foo') // ./foo.js
new Worker('foo') // ./foo.js || node_modules/foo
```

The main difference with these specifiers is that they are always treated as URLs, and bare specifiers are resolved as relative by default but fall back to node_modules. See #670. I'm still debating whether extensionless specifiers should be supported here. We could support them only for relative paths and not bare specifiers, or as above at all times. Would like feedback here.

## CSS

The resolution algorithm for CSS is similar to the JS one, but has some additional restrictions. It still allows relative, tilde, and absolute path specifiers, but all specifiers are treated as URLs. This means that in addition to allowing query params, percent encoding, etc., relative URLs are allowed without prefixing with `./`. Unfortunately, this conflicts with `node_modules` resolution, but that's less expected than in JS. For CSS, we will first check whether a relative path exists for a bare specifier, and if not, check if a `node_modules` path exists.

Some examples:

```css
@import "foo"; /* node_modules/foo */
@import "foo/bar.css"; /* node_modules/foo/bar.css */
@import "foo.css"; /* ./foo.css || node_modules/foo.css */
@import "./foo"; /* ERROR - extensionless imports not allowed */
@import "./foo.css"; /* ./foo.css */
@import "~/foo.css"; /* <nearest dir with package.json>/foo.js */
@import "/foo.css"; /* <projectRoot>/foo.js */
```

Extensionless imports are not allowed (unlike JS, unlike Parcel 1), and directory index imports are also not allowed (e.g. no `foo/index.css`). This more closely matches the CSS spec, which treats imports are URLs.

When resolving from `node_modules`, the `package.json#style` field is considered as an entry point. Other entry fields are mostly JS specific, so are not supported for CSS.

## HTML

The resolution algorithm for HTML is much simpler than for JS and CSS. `node_modules` resolution is not generally needed or expected in HTML, so it is not supported by default. That just leaves relative, tilde, and absolute paths, and URLs.

Some examples:

```html
<a href="foo.html"></a> <!-- ./foo.html -->
<script src="foo.js"></script> <!-- ./foo.js -->
<script src="./foo.js"></script> <!-- ./foo.js -->
<script src="~/foo.js"></script> <!-- <nearest dir with package.json>/foo.js -->
<script src="/foo.js"></script> <!-- <projectRoot>/foo.js -->
<script src="./foo"></script> <!-- ERROR - extensionless -->
<script src="foo"></script> <!-- ERROR - node_modules resolution is not allowed -->
```

Extensionless imports and directory index imports are not supported. All specifiers are treated as URLs, and are resolved with URL semantics. Bare specifiers are treated as relative URLs, just like normal HTML.

## Other languages

Other languages generally use their own import semantics and are handled by the individual transformer plugins. A few have Parcel specific extensions:

* SASS, LESS, Stylus - these languages are extended to allow node_modules resolution, along with absolute and tilde paths (the CSS resolution path described above). If the Parcel resolver doesn't find anything, then the transformer falls back to the default resolver for those languages.

### LINKED ISSUE #5484: CSS @imports are not the first rule in the css output file
# 🐛 bug report

The `@import` rule is supposed to be the first rule in a CSS file. When two css files are combined, the `@import` rule is kept in the import order and not moved to the start of the file.

From the official specs (present since CSS2):
> Any `@import` rules must precede all other valid at-rules and style rules in a style sheet (ignoring `@charset`), or else the `@import` rule is invalid.

https://drafts.csswg.org/css-cascade-4/#at-ruledef-import

```js
import 'style.css'
import 'another.css'
```

```css
/* style.css */
body {
    ...
}
```
```css
/* another.css */
@import url('https://fonts.googleapis.com/icon?family=Material+Icons');

div {
    ...
}
```

## 🤔 Expected Behavior

```css
/* output css file */

@import url('https://fonts.googleapis.com/icon?family=Material+Icons');

body {
    ...
}

div {
    ...
}
```

## 😯 Current Behavior

```css
/* output css file */

body {
    ...
}

@import url('https://fonts.googleapis.com/icon?family=Material+Icons');

div {
    ...
}
```

## 💁 Possible Solution

The `@import` rules should be detected and, either moved to the start of the file or splitted into its own file.

## 🔦 Context

Each component of my app contains its own css file, by changing the order the components are loaded, the `@import` rule might not be the first rule in the final css file, meaning it won't work at all.

I used to build the same codebase using Webpack which worked well with weirdly-ordered css imports, this is the only issue I've seen after moving over to Parcel.

## 🌍 Your Environment

| Parcel           | 2.0.0-beta.1 (I've also seen the issue in v1) |
| Node             | 13.5.0 |
| npm/Yarn         | Yarn 1.21.1 |
| Operating System | Windows 10 x64 |

devongovett:
While trying to document the resolver, I came across some inconsistencies and opportunities for improvement.

* File extensions are always required for URL dependencies. For ESM and CJS specifiers, only the following extensions may be omitted: `['ts', 'tsx', 'js', 'jsx', 'json']`.
* Added some additional reserved pipeline names, which conflict with common URL schemes. Some of these may be handled by the default resolver in the future, or by custom resolver plugins.
* Don't support resolving directories from URL dependencies (either package.json or index files)
* Allow resolver plugins to handle more aspects of URL parsing. Previously, core would return early for URL dependencies with schemes that didn't match a pipeline, or throw for ESM. Core also removed query parameters before passing the specifier to resolver plugins. Now it does none of this, and lets plugins decide what to do. Plugins can return query params that they parsed and didn't handle. This also fixes parsing ESM specifiers as URLs, so that things like percent encoding are supported properly.
* CommonJS no longer supports query parameters. This matches node's behavior, where the `?` character is allowed in filenames (just not on Windows...).
* Adds support for an `npm:` scheme, which allows URL dependencies to resolve to `node_modules` packages rather than as a relative path. Useful for e.g. CSS and web workers. Fixes #670. Closes #3492.
* Fixes the CSS transformer to pass external URLs through resolvers rather than leaving them inline. Hoists all `@import` declarations to the top of the packaged bundle. Fixes #5484.
* Renames the `filePath` option of resolver plugins to `specifier` since it may include things like query params and URL schemes. Note that this _is_ slightly different from `dependency.specifier` in cases where a known named pipeline is used. In this case, the `specifier` passed to resolvers already has the pipeline stripped off.


parcel-benchmark:
## Benchmark Results

#### Timings

| --- | --- | --- |
| Cold | 1.58s | +12.00ms |
| Cached | 263.00ms | +9.00ms |

#### Cold Bundles

| dist/legacy/parcel.7cdb0fad.webp | 102.94kb | +0.00b | 461.00ms | -38.00ms 🚀 |
| dist/legacy/parcel.7cdb0fad.webp | 102.94kb | +0.00b | 462.00ms | -38.00ms 🚀 |
| dist/modern/parcel.7cdb0fad.webp | 102.94kb | +0.00b | 462.00ms | -37.00ms 🚀 |

#### Cached Bundles

| dist/legacy/parcel.7cdb0fad.webp | 102.94kb | +0.00b | 471.00ms | -27.00ms 🚀 |
| dist/legacy/parcel.7cdb0fad.webp | 102.94kb | +0.00b | 471.00ms | -27.00ms 🚀 |
| dist/modern/parcel.7cdb0fad.webp | 102.94kb | +0.00b | 472.00ms | -27.00ms 🚀 |

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 8.06s | -105.00ms |
| Cached | 414.00ms | -2.00ms |

#### Cold Bundles

*No bundle changes detected.*

#### Cached Bundles

*No bundle changes detected.*

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 57.06s | +1.05s |
| Cached | 1.51s | -4.00ms |

#### Cold Bundles

| dist/index.04eaaaa3.js | 1.77mb | +4.00b ⚠️ | 16.53s | +322.00ms |
| dist/index.d2f714b9.js | 694.59kb | +15.00b ⚠️ | 43.83s | +1.10s |
| dist/Modal.6d102ab0.js | 45.30kb | +15.00b ⚠️ | 43.34s | +1.06s |

#### Cached Bundles

| dist/index.d4ffc1c1.js | 1.77mb | +6.00b ⚠️ | 16.05s | -123.00ms |
| dist/EmojiPickerComponent.513df8b1.js | 146.66kb | +0.00b | 17.53s | -8.80s 🚀 |
| dist/esm.d037b5a8.js | 33.13kb | +0.00b | 17.53s | -8.80s 🚀 |
| dist/DatePicker.69b3c4ae.js | 22.96kb | +0.00b | 17.53s | -8.80s 🚀 |
| dist/js.0f0bb621.js | 17.24kb | +0.00b | 31.95s | -10.14s 🚀 |
| dist/png-chunks-extract.94b5b9fc.js | 3.56kb | +0.00b | 31.95s | -10.14s 🚀 |
| dist/feedback.e18b45f0.js | 1.75kb | +0.00b | 17.53s | -8.80s 🚀 |
| dist/workerHasher.567cfc57.js | 1.62kb | +0.00b | 17.53s | -8.80s 🚀 |
| dist/heading6.b3b946d1.js | 1.49kb | +0.00b | 17.53s | -8.80s 🚀 |
| dist/heading5.e617db66.js | 1.37kb | +0.00b | 17.53s | -8.80s 🚀 |
| dist/expand.1f17de7f.js | 1.27kb | +0.00b | 17.53s | -8.80s 🚀 |

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 6.03s | +30.00ms |
| Cached | 342.00ms | -12.00ms |

#### Cold Bundles

*No bundle changes detected.*

#### Cached Bundles

*No bundle changes detected.*

</p></details>

**[Click here to view a detailed benchmark overview.](https://benchmark.parceljs.org/benchmark/307033392207102528)**

packages/resolvers/default/src/DefaultResolver.js
```diff
@@ -8,7 +8,7 @@ import NodeResolver from '@parcel/node-resolver-core';
 const WEBPACK_IMPORT_REGEX = /\S+-loader\S*!\S+/g;
 
 export default (new Resolver({
-  resolve({dependency, options, filePath}) {
+  resolve({dependency, options, specifier}) {
```
wbinnssmith, line 11: 👍 this name makes so much more sense

wbinnssmith (approved):

packages/core/types/index.js
```diff
@@ -1452,6 +1452,7 @@ export type ResolveResult = {|
   /** An absolute path to the file. */
   +filePath?: FilePath,
   +pipeline?: ?string,
+  +query?: QueryParameters,
```
mischnic, line 1457: Can you document this? (That these are the query params the resolver didn't handle and should be forwarded to transformers)

packages/core/core/src/requests/PathRequest.js
```diff
@@ -176,82 +167,29 @@ export class ResolverRunner {
...
-        };
-      }
-      filePath = decodeURIComponent(parsed.pathname);
-      if (parsed.query != null) {
-        queryPart = parsed.query;
+      [, pipeline, specifier] = match;
+      if (!validPipelines.has(pipeline)) {
+        // This may be a url protocol or scheme rather than a pipeline, such as
+        // `url('http://example.com/foo.png')`. Pass it to resolvers to handle.
+        // return {assetGroup: null};
```
mischnic:
```suggestion
```

Issue #6269, opened 2021-05-12
`parcel build` with web worker produces code that cannot execute


jeffpeck10x:
# 🐛 bug report

`parcel build` with web worker produces code that cannot execute

```
  const worker = new Worker(
      new URL("./workers/my-worker.ts", import.meta.url),
      { type: "module" }
    );
```

```
$ parcel build ./src/index.ts --no-source-maps --no-optimize --no-cache
```

The result in `dist/module/index.js`:
```
$37fff9a7ec0a7105da2f7eb5f3df596b$exports = $53630ef4c3b85e8cc60065ddf3d18aba$init()("../main/my-worker.ed75ea8e.js");
```

which fails when imported in another application with:
```
Error: Origin not found
```

Why? Because `getBundleURL()` is returning a `/`:
https://github.com/parcel-bundler/parcel/blob/v2/packages/runtimes/js/src/bundle-url.js#L12

And then `getOrigin()` fails to regex match the origin:
https://github.com/parcel-bundler/parcel/blob/v2/packages/runtimes/js/src/bundle-url.js#L39

I am wondering if I am missing something, but I am not sure what to even try next.

Honestly, I would really appreciate if the answer is: "Parcel does not yet support web workers" just so I don't have to spend time trying to make this work, and can focus on a different solution.

## 🎛 Configuration (.babelrc, package.json, cli command)

.babel.config.json
```json
{
  "plugins": [
    "@parcel/babel-plugin-transform-runtime",
    ["inline-svg", { "disableSVGO": true }]
  ],
  "presets": [
    "@parcel/babel-preset-env",
    "@babel/preset-react",
    "@babel/preset-typescript"
  ]
}

```

## 🤔 Expected Behavior

`parcel build` will produce code that will not crash when encountering a web worker.

## 😯 Current Behavior

```
Error: Origin not found
```

## 🔦 Context

Cannot use web workers with parcel2 at the moment.

## 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Parcel           | 2.0.0-beta.2
| Node             | 12.18.3
| npm/Yarn         | 1.22.10
| Operating System | MacOS 10.15.7

jeffpeck10x:
I have found a workaround, prepending "data-url:" to the path:
```
    const worker = new Worker(
      new URL("data-url:./workers/my-worker.ts", import.meta.url),
      { type: "module" }
    );
```

[...]

Issue #4731, opened 2020-06-12
Compatibility issues with node workers / `comlink`


lgarron:
# 🐛 bug report

Using the `comlink` library with a `node` target in Parcel 2 fails to produce working code. :-(

## 🎛 Configuration (.babelrc, package.json, cli command)

`package.json`:
```js
{
  "main": "dist/index.js",
  "targets": {
    "main": {
      "context": "node"
    },
    ...
  }
```

CLI command:

    // --no-minify helps with debugging, but does not cause/affect the issues in questions.
    npx parcel build --no-minify src/main.ts
    node dist/index.js

## 🤔 Expected Behavior

The resulting code runs in `node` with a worker, and prints the correct output (4).

## 😯 Current Behavior

There are two main issues.

Without `"includeNodeModules": true` in `package.json`:

    > cd issue1
    > npm install
    > npx parcel build src/main.ts

    🚨 Build failed.
    @parcel/packager-js: External modules are not supported when building for browser
    /Users/lgarron/parcel-bug-worker-node-external-modules/issue1/src/worker.ts:1:1
    > 1 | import * as Comlink from "comlink";
    >   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      2 | import nodeEndpoint from "comlink/dist/esm/node-adapter";
      3 | import { parentPort } from "worker_threads";

With `"includeNodeModules": true` in `package.json`:

    > cd issue2
    > npm install
    > npx parcel build src/main.ts
    > node dist/index.js
    
    (node:17465) UnhandledPromiseRejectionWarning: Error: Origin not found
        at Object.L [as getOrigin] (/Users/lgarron/parcel-bug-worker-node-external-modules/issue2/dist/index.js:1:3467)
        at k (/Users/lgarron/parcel-bug-worker-node-external-modules/issue2/dist/index.js:1:3708)
        at P (/Users/lgarron/parcel-bug-worker-node-external-modules/issue2/dist/index.js:1:3873)
        at /Users/lgarron/parcel-bug-worker-node-external-modules/issue2/dist/index.js:1:10995
        at s (/Users/lgarron/parcel-bug-worker-node-external-modules/issue2/dist/index.js:1:4954)
        at Generator._invoke (/Users/lgarron/parcel-bug-worker-node-external-modules/issue2/dist/index.js:1:4707)
[...]

PR #6742, opened 2021-08-19
Add default origin for workers

egasimus:
# ↪️ Pull Request

* Changes the `getOrigin` helper in `@parcel/runtime-js` to return `self.location.origin` if the passed URL starts with `/`.
* Possible fix for #4731, #6141, #6269.

## 💻 Examples

```json
// .parcelrc
{ "extends": "@parcel/config-default",
  "transformers": {
    "*.ts": ["@parcel/transformer-typescript-tsc"]},
  "validators": {
    "*.ts": ["@parcel/validator-typescript"]} }
```

```html
// index.html
<html>
<head></head>
<body><script type="module" src="index.ts"></script></body>
</html>
```

```js
// index.js
const worker = new Worker(
  new URL('./worker.ts', import.meta.url),
  { type: 'module' }))
```

```ts
// worker.ts
import something from './somewhere'
console.log(something())
```

```ts
// somewhere.ts
export default () => 'it works!'
```

## 🚨 Test instructions

```sh
parcel index.html
```

## ✔️ PR Todo

### DIFF
--- a/packages/runtimes/js/src/helpers/bundle-url.js
+++ b/packages/runtimes/js/src/helpers/bundle-url.js
@@ -32,7 +32,11 @@ function getBaseURL(url) {
 
 // TODO: Replace uses with `new URL(url).origin` when ie11 is no longer supported.
 function getOrigin(url) {
-  let matches = ('' + url).match(/(https?|file|ftp):\/\/[^/]+/);
+  url = '' + url;
+  if (url[0] === '/') {
+    return self.location.origin
+  }
+  let matches = url.match(/(https?|file|ftp):\/\/[^/]+/);
   if (!matches) {
     throw new Error('Origin not found');
   }


packages/runtimes/js/src/helpers/bundle-url.js
```diff
@@ -32,7 +32,11 @@ function getBaseURL(url) {
 
 // TODO: Replace uses with `new URL(url).origin` when ie11 is no longer supported.
 function getOrigin(url) {
-  let matches = ('' + url).match(/(https?|file|ftp):\/\/[^/]+/);
+  url = '' + url;
```
jeffpeck10x, line 35: Instead of mutating the `url` arg, maybe try `if (typeof url === 'string' && url[0] === '/')` and keep the `('' + url).match(...` below.

egasimus: My assumption was that `(''+url)` is an implicit conversion to string? I'm unfamiliar with this particular idiom - I'd much rather do `if (typeof url !== 'string') { throw new Error('you accidentally a url') }` and not append it to an empty string at all (memory copy optimization?)

jeffpeck10x: I believe that is what it is doing. `''+undefined === "undefined"`, so if `url` is undefined or null, etc., you are not going to get a match anyway. But, I suppose it ensures that `.match` never fails.

PR #6743, opened 2021-08-19
Process CSS in SVGs

thewilkybarkid:
# ↪️ Pull Request

SVGs can include CSS, this will see it processed.

Follows #6595 (currently unmerged, so those changes appear here too) and refs https://github.com/parcel-bundler/parcel/issues/2901#issuecomment-483482271.

Leaving in draft for the moment as there's definitely more tests to add (and probably changes to make).

## 💻 Examples

A SVG with CSS will have it processed, so:

```svg
<svg viewBox="0 0 240 80" xmlns="http://www.w3.org/2000/svg">
  <style>
    @import './test.css';
  </style>
  <text>Should be red</text>
</svg>
```

becomes

```svg
<svg viewBox="0 0 240 80" xmlns="http://www.w3.org/2000/svg">
  <style>
    :root {
      fill: red
    }
  </style>
  <text>Should be red</text>
</svg>
```

## 🚨 Test instructions

Integration test included.

## ✔️ PR Todo


packages/core/integration-tests/test/svg.js
```diff
@@ -35,17 +58,56 @@ describe('svg', function() {
...
+    );
+
+    assert(!svg.includes('style1.css'));
+    assert(!svg.includes('style3.css'));
+    assert(svg.includes('style2.css'));
+    assert(svg.includes('<?xml-not-a-stylesheet'));
+    assert(svg.includes('<style>'));
+    assert(!svg.includes('@import'));
+    assert(svg.includes(':root {\n  fill: red;\n  font-family: serif;\n}'));
+    assert(svg.includes(':root {\n  font-family: monospace;\n}\n'));
```
thewilkybarkid: Probably better to (safely) compare the output as a whole.

thewilkybarkid:
~The test failures seem to be that the packager (i.e. PostHTML) doesn't produce self-closing tags, which I assume the optimizer would handle anyway? Relates to https://github.com/parcel-bundler/parcel/pull/6743#discussion_r691951410, in that it's probably better to compare XML documents rather than strings.~

thewilkybarkid:
~Found some conflicts with SVGO (namely plugins that use `computeStyle`, e.g. `removeHiddenElems`), hitting `Empty sub-selector`.~

~(Had to turn off `removeHiddenElems`, `minifyStyles`, `convertPathData` and `mergePaths` to get it to build.)~

thewilkybarkid:
~SVGO's `inlineStyles` seems to be problematic too, some selectors not working when used.~ Seems to be an SVGO problem (https://github.com/svg/svgo/issues/1359).

packages/transformers/svg/src/dependencies.js
```diff
@@ -0,0 +1,79 @@
...
+  let isDirty = false;
+  const stylesheets = [];
+  PostHTML().walk.call(ast.program, node => {
+    if (typeof node === 'string' && node.startsWith('<?xml-stylesheet')) {
+      const [, href] = node.match(/(?<=(?:^|\s)href\s*=\s*")(.+?)(?=")/i) ?? [];
+
+      if (!href) {
+        return node;
+      }
+      stylesheets.push(href);
```
devongovett: This is interesting. In HTML, we keep separate `<link rel="stylesheet">` elements separate even after bundling. What was the reasoning for merging them here?

thewilkybarkid: External resources in an SVG won't load when using ` ` (but will with `<object>`), but was thinking about this and it's the wrong place to handle it: if anywhere the packager could recognise if the SVG is used in an `` and inline resources.

devongovett (approved):

devongovett:
Thanks!

PR #6757, opened 2021-08-24
Make sure (non-React) SVGs are in separate bundles

thewilkybarkid:
# ↪️ Pull Request

`@parcel/packager-raw-url` expects them to be in separate bundles, so currently throws an error when using them from JS. (I don't use `@parcel/transformer-svg-react`, but that seems to need the opposite behaviour.)

## 💻 Examples

See test.

## 🚨 Test instructions

Test included.

## ✔️ PR Todo


packages/transformers/svg-react/src/SvgReactTransformer.js
```diff
@@ -34,6 +34,7 @@ export default (new Transformer({
     );
 
     asset.type = 'jsx';
+    asset.bundleBehavior = 'inline';
```
devongovett: I think this should be `null` instead. `inline` means it will create a separate bundle, and then inline it, but we really just want it to be in the same bundle as the parent javascript.

```suggestion
    asset.bundleBehavior = null;
```

devongovett (approved):

Issue #3738, opened 2019-11-03
Parcel 2: HTML Image Transformer


devongovett:
Create the `@parcel/transformer-html-images` (name TBD) package. This should include a Transformer plugin that transforms `<img>` tags in HTML to `<picture>` elements with multiple sources. It should work with the [@parcel/transformer-webp](https://github.com/parcel-bundler/parcel/issues/3736) and [@parcel/transformer-image-resize](https://github.com/parcel-bundler/parcel/issues/3737) plugins to create multiple dependencies that are transformed by those plugins and re-inserted later as URLs. This handles generating multiple versions of an image with different sizes and formats from a source HTML file and image completely automatically.

HTML like this:

```html
<img src="./image.jpg" width="200">
```

should be transformed into HTML like this, with dependencies created for each referenced asset:

```html
<picture>
  <source type="image/webp" srcset="webp:./image.jpg?width=200, webp:./image.jpg?width=400 2x">
  <source type="image/jpeg" srcset="./image.jpg?width=200, ./image.jpg?width=400 2x">
  <img src="./image.jpg?width=200" width="200">
</picture>
```

This automatically handles creating both WebP and JPG versions of the image, at multiple resolutions.

In addition, the `srcset` attribute should be supported to generate multiple versions with specific sizes from the same input image:

```html
<img srcset="image.jpg 480w, image.jpg 1080w">
```

becomes:

```html
<picture>
  <source type="image/webp" srcset="webp:image.jpg?width=480 480w, webp:image.jpg?width=1080 1080w">
  <source type="image/jpeg" srcset="image.jpg?width=480 480w image.jpg?width=1080 1080w">
  <img src="./image.jpg?width=1080w">
</picture>
```

Note that the protocol and query param syntax here is internal to Parcel and would not be shipped to users. After the images are transformed, the final URL would be placed into the HTML file as expected.

## Open Questions

* What should the default resolutions include? 1x, 2x, 3x? Should this happen by default at all or should we require users specify `srcset` in their code themselves? Seems like it should be fairly common to want this...
* What should the default formats include? WebP and the original format? WebP and JPEG always?
* What should happen if there is no `width` or `height` set on an `<img>`? Just convert to WebP without resizing?

mischnic:
`width` and `height` will be used to calculate the aspect ratio which can reduce layout jank, so we should consider adding these attributes (automatically? - does that have layout sideeffects?):
https://twitter.com/addyosmani/status/1169813271009886208

[...]

PR #6783, opened 2021-08-27
Set image width/height attributes in HTML

thewilkybarkid:
# ↪️ Pull Request

This sets the `width` and `height` attributes on HTML `` elements to help layout in browsers (refs https://github.com/parcel-bundler/parcel/issues/3738#issuecomment-549504886).

I had been using [posthtml-img-autosize](https://github.com/posthtml/posthtml-img-autosize) (with the `processEmptySize` option), but that's incompatible with things like the image transformer (since it runs before transformations).

## 💻 Examples

See the tests.

## 🚨 Test instructions

Tests included.

## ✔️ PR Todo


thewilkybarkid:
The potential layout problem is if someone has something like `img { width: 100% }` without `height: auto` (since the image will stretch). This is combination 3 on https://github.com/WICG/intrinsicsize-attribute/issues/16#issuecomment-500917583.

Or are using the width/height to control the size (refs https://github.com/parcel-bundler/parcel/pull/6783#discussion_r697663129 and https://github.com/parcel-bundler/parcel/issues/3738#issue-516907103).

Feels like the benefits far outweigh this edge case though.

packages/core/integration-tests/test/integration/html-image-size/index.html
```diff
@@ -0,0 +1,12 @@
+<!doctype html>
+<html>
+<body>
+<img id="image1" src="100x100.png">
+<img id="image2" src="100x100.png?width=150">
+<img id="image3" src="100x100.png?height=150">
+<img id="image4" src="100x100.png?width=200&height=150">
+<img id="image5" src="200x200.png" width="100">
+<img id="image6" src="200x200.png" height="100">
```
thewilkybarkid, line 9: As it stands, these are both fixed to be the real size of the image.

flow-libs/image-size.js.flow
```diff
@@ -0,0 +1,5 @@
+// @flow
+
+declare module 'image-size' {
+  declare module.exports: (input: Buffer) => { width: number, height: number, ...};
```
mischnic: Apparently, `width` and `height` can also be undefined: https://github.com/image-size/image-size/blob/a38b56fe7303898ce6811adb213e518a57593d10/lib/types/interface.ts#L1-L10

thewilkybarkid: I don't think those types match the implementation (I believe there should be union types of the different combinations), but better to match it.

thewilkybarkid:
~I'm planning on experimenting with injecting a little CSS to avoid misshapen images (which is hopefully tree-shaken away in most cases).~ (Too much for now.)

packages/packagers/html/src/HTMLPackager.js
```diff
@@ -112,6 +114,51 @@ async function getAssetContent(
...
+
+    const image = bundleGraph.getResolvedAsset(dependency);
+
+    if (!image) {
+      continue;
+    }
+
+    if (
+      typeof image.meta.width !== 'number' &&
+      typeof image.meta.height !== 'number'
```
devongovett: Are these supposed to be set somewhere (e.g. the image transformer)?

packages/packagers/html/src/HTMLPackager.js
```diff
@@ -112,6 +114,51 @@ async function getAssetContent(
...
+    ) {
+      const size = imageSize(await image.getBuffer());
+      image.meta.width = size.width;
+      image.meta.height = size.height;
```
devongovett: I'm not sure we should mutate assets in packagers. I'm actually surprised this works... 😬

packages/packagers/html/src/HTMLPackager.js
```diff
@@ -112,6 +114,51 @@ async function getAssetContent(
...
+    }
+
+    node.attrs.width = image.meta.width;
+    node.attrs.height = image.meta.height;
```
devongovett, line 142: I think we should only do this if a `width` or `height` attribute is not already set on the element?

mischnic:
We never want to make Parcel "break" your code after bundling, so if it worked correctly before bundling, it should still behave the same afterwards. That is unfortunately not always the case here:

https://codesandbox.io/s/serene-ives-legx7?file=/index.html

So I'm torn between automatically adding them to avoid layout shift and not randomly breaking the projects of unknowing users (especially because this is hard to debug if you don't know what you're looking for. In this case, you need to set the other "free" dimension to `auto` instead of `initial` to retain the aspect ratio).

packages/packagers/html/src/HTMLPackager.js
thewilkybarkid: Ideally, yes.

packages/packagers/html/src/HTMLPackager.js
thewilkybarkid: > I think we should only do this if a `width` or `height` attribute is not already set on the element?

Probably. I left this in for the moment (https://github.com/parcel-bundler/parcel/pull/6783#discussion_r697663129) for feedback (primarily about the CSS issue), but in the longer term it would be better for these to be "fixed" earlier in the pipeline (refs https://github.com/parcel-bundler/parcel/issues/3738#issue-516907103).

packages/packagers/html/src/HTMLPackager.js
thewilkybarkid: It could be that the image transformer sets the sizes (even if it's not actually transforming the image), and this sets them if present.

thewilkybarkid:
> So I'm torn between automatically adding them to avoid layout shift and not randomly breaking the projects of unknowing users (especially because this is hard to debug if you don't know what you're looking for. In this case, you need to set the other "free" dimension to `auto` instead of `initial` to retain the aspect ratio).

I do wonder about this too; the packager doesn't feel the right place to do it but I'm not sure where is (the HTML transformer's too early?).

I did wonder about injecting some CSS (https://github.com/parcel-bundler/parcel/pull/6783#issuecomment-913039779), but that also could be thorny (and I'm not sure how and _where_ it should happen: inline and before any CSS in all HTML files that contain images?).

thewilkybarkid:
If it helps, this could be split in two: the first part that adds the width/height as metadata on image assets, then a follow-up that uses it to set the width/height in HTML. Since the latter has potential CSS problems, I could release this as a separate plugin. (The former could also be a separate plugin, but doesn't seem unreasonable to make that information generally available for image assets?)

> the packager doesn't feel the right place to do it but I'm not sure where is (the HTML transformer's too early?).

Looking at the plugin system again, would it be an optimizer? (It's not 'optimization' as such, but needs to be a late-running transformation.)

thewilkybarkid:
> Looking at the plugin system again, would it be an optimizer? (It's not 'optimization' as such, but needs to be a late-running transformation.)

I had a go at writing an optimizer and managed to get it to work, it just seems a bit tricky to find the asset from the `src` (which is now the final path, e.g. `/some-image.ab12345.jpg`).

First attempt looks like:

```js
tree.match({tag: 'img', attrs: {src: true}}, node => {
  let image;

  for (const dependency of dependencies) {
    const asset = bundleGraph.getResolvedAsset(dependency, bundle)

    if (!asset) {
      continue;
    }

    const bundles = bundleGraph.getBundlesWithAsset(asset);

    if (node.attrs['src'] !== bundles[0].name) {
      continue;
    }

    image = asset;
    break;
  }

  // use imageSize on image and update attributes on node

  return node;
});
```

which works, but is rather ugly. Is there a simple way that I've missed?

Does either of these approaches make sense @devongovett @mischnic? I've happy to update this PR or close it and release something standalone.

packages/core/integration-tests/test/integration/html-image-size/index.html
thewilkybarkid: Would be changed by https://github.com/parcel-bundler/parcel/pull/7390

PR #5728, opened 2021-01-28
Fix CSS tree shaking with 'build --no-scope-hoist'

mischnic:
The used symbols data is only valid when scope-hoisting ran. However, for CSS assets `env.scopeHoist` is reset to deduplicate more assets. 

Previously `parcel build --no-scope-hoist` would still shake all CSS declarations because `mode` was indeed production.

PR #6796, opened 2021-08-28
Do not tree shake CSS classes when scope hoisting is disabled

### LINKED ISSUE #6780: CSS modules do not build correctly without scope hoisting
# 🐛 bug report

Running `parcel build --no-scope-hoist` causes used styles to be removed.

I built an example repo here that demonstrates the bug: https://github.com/asmockler/parcel-css-module-bug.

There are 2 build commands, one with scope hoisting on and the other with scope hoisting off. The build to different dist dirs to make it easier to compare the output.

## 🤔 Expected Behavior

All referenced classes from CSS modules should be included whether or not scope hoisting is turned on.

## 😯 Current Behavior

With `--no-scope-hoist`, all or almost all classes are removed during build. In a larger project where I first noticed the bug, I often wind up with just a single class in the output css. 

I will also note that it seems like the build step takes a long time on `Optimizing ModuleName.[hash].css` when it exhibits this behavior, which makes me think the styles are included and then tree-shaken, but I'm not sure. 

## 💁 Possible Solution

Really not sure -- running with `--no-optimize` doesn't seem to fix it, either, though.

## 🔦 Context

I think that with `--no-scope-hoist`, it would be ok to disable all tree-shaking from CSS modules.

## 💻 Code Sample

https://github.com/asmockler/parcel-css-module-bug

## 🌍 Your Environment

| Parcel           | ^2.0.0-rc.0 |
| Node             | Observed on `14.16` and `16.0` |
| npm/Yarn         | Yarn `1.22.10` |
| Operating System | Mac OS Big Sur |

asmockler:
# ↪️ Pull Request

Attempt to resolve https://github.com/parcel-bundler/parcel/issues/6780

Disables CSS tree-shaking when `--no-scope-hoist` is present.

## 💻 Examples

I have an example of the bug in this repo: https://github.com/asmockler/parcel-css-module-bug

## 🚨 Test instructions

Before this change, most or all CSS from CSS modules would be tree-shaken when `--no-scope-hoist` was passed; now, all CSS should be included.

## ✔️ PR Todo

### DIFF
--- a/packages/packagers/css/src/CSSPackager.js
+++ b/packages/packagers/css/src/CSSPackager.js
@@ -152,10 +152,41 @@ async function processCSSModule(
 ): Promise<[Asset, string, ?Buffer]> {
   let ast: Root = postcss.fromJSON(nullthrows((await asset.getAST())?.program));
 
-  let usedSymbols = bundleGraph.getUsedSymbols(asset);
   let localSymbols = new Set(
     [...asset.symbols].map(([, {local}]) => `.${local}`),
   );
+
+  if (asset.env.shouldScopeHoist) {
+    treeShakeUnusedSymbols(logger, ast, bundleGraph, asset, localSymbols);
+  }
+
+  let {content, map} = await postcss().process(ast, {
+    from: undefined,
+    to: options.projectRoot + '/index',
+    map: {
+      annotation: false,
+      inline: false,
+    },
+    // Pass postcss's own stringifier to it to silence its warning
+    // as we don't want to perform any transformations -- only generate
+    stringifier: postcss.stringify,
+  });
+
+  let sourceMap;
+  if (bundle.env.sourceMap && map != null) {
+    sourceMap = new SourceMap(options.projectRoot);
+    sourceMap.addVLQMap(map.toJSON());
+  }
+
+  if (media.length) {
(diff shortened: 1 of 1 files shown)


devongovett:
cc. @mischnic 

I think another question is why we are storing an AST at all when scope hoisting is disabled...

mischnic:
I would be surprised if this actually works. Because we reset `env.scopeHoist` in the CSS transformer to allow for more deduplication, this means that the CSS packager doesn't actually know if scope hoisting is enabled or not.

`asset.env.shouldScopeHoist` is always false in the packager.

asmockler:
You're right...I thought I had fixed it, but turns out I had just broken the normal behavior 😅  what if `scopeHoist` was passed through in the CSS transformer as well?

i.e. add `shouldScopeHoist: asset.env.shouldScopeHoist,` to https://github.com/parcel-bundler/parcel/blob/v2/packages/transformers/css/src/CSSTransformer.js#L56-L63

I would be happy to add that, but I am not sure if it is going to have other consequences.

mischnic:
My plan for this is to make `getUsedSymbols` return `Array<Symbol> | null`, so that you can differentiate the cases "nothing is used" and "symbol propagation didn't run"/"unkown".

devongovett:
Should be fixed by #5728.

PR #6798, opened 2021-08-29
Add support for injecting manifest into service workers

devongovett:
This adds a `@parcel/runtime-service-worker` plugin, which can be used to inject a manifest of bundles along with a version hash into a service worker. There is also a "browser" target for this plugin that you can import in a service worker to access the manifest/version.

For example (code based on https://bundlers.tooling.report/non-js-resources/service-worker/):

```js
import {manifest, version} from '@parcel/runtime-service-worker';

async function install() {
  const cache = await caches.open(version);
  await cache.addAll(manifest);
}
addEventListener('install', e => e.waitUntil(install()));

async function activate() {
  const keys = await caches.keys();
  await Promise.all(keys.map(key => key !== version && caches.delete(key)));
}
addEventListener('activate', e => e.waitUntil(activate()));
```


parcel-benchmark:
## Benchmark Results

#### Timings

| --- | --- | --- |
| Cold | 1.57s | +6.00ms |
| Cached | 277.00ms | +17.00ms ⚠️ |

#### Cold Bundles

*No bundle changes detected.*

#### Cached Bundles

*No bundle changes detected.*

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 8.47s | -67.00ms |
| Cached | 418.00ms | +6.00ms |

#### Cold Bundles

| dist/logo.c5bb83f1.png | 246.00b | +0.00b | 4.54s | +534.00ms ⚠️ |

#### Cached Bundles

*No bundle changes detected.*

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 56.91s | -164.00ms |
| Cached | 1.60s | +86.00ms ⚠️ |

#### Cold Bundles

| dist/index.334769ba.js | 1.77mb | -8.00b 🚀 | 16.41s | +4.00ms |
| dist/index.2017f979.js | 694.59kb | +19.00b ⚠️ | 43.89s | +47.00ms |

#### Cached Bundles

| dist/index.ed2b7d05.js | 1.77mb | -6.00b 🚀 | 16.91s | +400.00ms |
| dist/index.fefdc2ee.js | 694.58kb | +4.00b ⚠️ | 45.28s | +1.50s |
| dist/Modal.b23a7ff6.js | 45.30kb | +15.00b ⚠️ | 44.77s | +1.44s |

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 6.15s | +51.00ms |
| Cached | 346.00ms | +1.00ms |

#### Cold Bundles

*No bundle changes detected.*

#### Cached Bundles

*No bundle changes detected.*

</p></details>

**[Click here to view a detailed benchmark overview.](https://benchmark.parceljs.org/benchmark/308418670481113665)**

packages/runtimes/service-worker/package.json
```diff
@@ -0,0 +1,32 @@
...
+  },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/parcel-bundler/parcel.git"
+  },
+  "main": "./lib/ServiceWorkerRuntime.js",
+  "source": {
+    "./lib/ServiceWorkerRuntime.js": "./src/ServiceWorkerRuntime.js"
+  },
+  "browser": "browser.js",
```
mischnic: This will definitely break when building the REPL.

Could we instead override the dependency like the JSRuntime does?

devongovett: The thing is, runtimes don't override a dependency, they add an _additional_ resolution - the original is still there. For replacing async dependencies this works because we prefer resolutions in the same bundle, but for sync dependencies you'll get the original asset.

devongovett: I guess one option would be to create another package, eg @parcel/service-worker... but I was trying avoid that

devongovett: I guess that wouldn't be too bad. You'd have to install the package manually into your project anyway (e.g. in Yarn 2) even if the runtime is in the default config 🤷

packages/core/core/src/requests/EntryRequest.js
```diff
@@ -248,7 +248,9 @@ export class EntryResolver {
         if (!allTargetsHaveSource && pkg.source != null) {
           let pkgSources = Array.isArray(pkg.source)
             ? pkg.source
-            : [pkg.source];
+            : typeof pkg.source === 'string'
+            ? [pkg.source]
+            : [];
```
mischnic: Are there tests somewhere for this change?

packages/runtimes/service-worker/src/ServiceWorkerRuntime.js
```diff
@@ -0,0 +1,50 @@
...
+    let manifest = [];
+    bundleGraph.traverseBundles(b => {
+      if (b.bundleBehavior === 'inline' || b.id === bundle.id) {
+        return;
+      }
+
+      manifest.push(urlJoin(b.target.publicUrl, b.name));
+    });
+
+    let code = `import {_register} from '@parcel/service-worker';
```
mischnic, line 35: So this imports `@parcel/service-worker` relative to the runtime? It should rather be resolved from the user's package. Yarn 2 would still break

packages/core/core/src/requests/EntryRequest.js
devongovett: Well, it's not really a change, this was done to appease Flow...

mischnic (approved):

Issue #6306, opened 2021-05-18
Class constructor cannot be invoked without 'new'


driescroons:
# 🐛 bug report

For my WebXR game, I'm migrating to parcel v2. However, I've ran into an issue with the Text component from the [Drei](https://www.npmjs.com/package/@react-three/drei). It uses [troika-three-text](https://www.npmjs.com/package/troika-three-text) internally.

I've setup a repro repo to showcase the issue. You can find that here: https://github.com/driescroons/parcelv2troika3dtext

When I serve my project, I get the following error:
![image](https://user-images.githubusercontent.com/27679518/118667207-165c8f00-b7f4-11eb-9621-6cd583bc6fdf.png)

```
Uncaught TypeError: Class constructor InstancedBufferGeometry cannot be invoked without 'new'
    at new GlyphsGeometry
```

Here is the location this error is thrown:
![image](https://i.imgur.com/lMxavLV.png)

However, the weird thing is that when I build the project and use live-server to host the example, I don't get any issue:
![image](https://user-images.githubusercontent.com/27679518/118667390-47d55a80-b7f4-11eb-9957-de3755577e27.png)

It seems like whenever I'm serving, parcel is using the umd distributable, but when building, using the esm?
![image](https://user-images.githubusercontent.com/27679518/118679480-34c78800-b7fe-11eb-9879-66fbc5f992c7.png)

umd file from troika-3d-text:
![image](https://i.imgur.com/aImvrcq.png)

esm file from troika-3d-text:
![image](https://i.imgur.com/khjs5Jg.png)

I've tried adding the browserslist flag to my package.json, but to no avail.

## 🎛 Configuration (.babelrc, package.json, cli command)

package.json
```
{
  "name": "parcelv2troika3dtext",
  "scripts": {
    "start": "parcel serve ./src/index.html",
    "build": "parcel build ./src/index.html"
  },
  "devDependencies": {
    "parcel": "^2.0.0-beta.2"
  },
  "dependencies": {
    "@react-three/drei": "^5.0.0",
    "@react-three/fiber": "^6.1.4",
    "react": "^17.0.2",
    "react-dom": "^17.0.2",
    "three": "^0.128.0"
  }
}
```

## 🤔 Expected Behavior

The serve command should work similar to the build command.

## 😯 Current Behavior

For some reason, the serve command, I think, uses the umd distributable file from troika-3d-text, while the build command uses the esm file. 

## 💻 Code Sample

[...]

Issue #6176, opened 2021-04-23
Parcel 2 gives wrong error message when node_modules contains types


danieltroger:
# 🐛 bug report

I'm compiling this: https://github.com/getsentry/sentry-javascript/tree/master/packages/browser/src to es2021 because sentry only ships es5 or es3 builds.

To not have to change every dependency in every file and not have to read and understand their build documentation I just symlinked all directories into node_modules.
Now my file structure looks like this:
```
package.json
node_modules <-- containing parcel
src
  lazy_sentry
      node_modules
```

But parcel's don't-transpile-node-modules somehow doesn't only recognize the node_modules containing parcel but also refuses to run the lazy_sentry node_modules through babel.

## 🎛 Configuration (.babelrc, package.json, cli command)

Standard parcel configuration with `@babel/preset-typescript` in the outer directory. `node_modules` but no package.json in the inner one.

## 🤔 Expected Behavior

Either that parcel just transpiles the code in the second node_modules or that it shows something like this:
> You tried to use something that needs to be transpiled inside of a folder called `node_modules`
> Please see this issue: https://github.com/parcel-bundler/parcel/issues/1655
> and run `sed -i 's/!filePath.includes(NODE_MODULES)/true/' node_modules/\@parcel/core/lib/summarizeRequest.js` to enable transpilation

## 😯 Current Behavior

```
🚨 Build failed.
@parcel/transformer-js: This experimental syntax requires enabling one of the following parser plugin(s): 'flow, typescript' (12:7)
SyntaxError: This experimental syntax requires enabling one of the following parser plugin(s): 'flow, typescript' (12:7)
    at Parser._raise ([project_dir]node_modules/@babel/parser/lib/index.js:776:17)
    at Parser.raiseWithData ([project_dir]node_modules/@babel/parser/lib/index.js:769:17)
    at Parser.expectOnePlugin ([project_dir]node_modules/@babel/parser/lib/index.js:9750:18)
    at Parser.isExportDefaultSpecifier ([project_dir]node_modules/@babel/parser/lib/index.js:13485:16)
    at Parser.maybeParseExportDefaultSpecifier ([project_dir]node_modules/@babel/parser/lib/index.js:13386:14)
    at Parser.parseExport ([project_dir]node_modules/@babel/parser/lib/index.js:13339:29)
    at Parser.parseStatementContent ([project_dir]node_modules/@babel/parser/lib/index.js:12365:27)
[...]

PR #6821, opened 2021-08-30
Readme.md: Make summary about parcel more accurate

danieltroger:
# ↪️ Pull Request

`node_modules` can't be transpiled, see https://github.com/parcel-bundler/parcel/issues/1655 and https://github.com/parcel-bundler/parcel/issues/6306#issuecomment-846641326

Yet the readme claimed they could. I propose to remove that claim.

## 💻 Examples

Me right now that needs solid-js to not make my scripts throw syntax errors in older browsers because they contain arrow functions because parcel refuses to transpile them

## 🚨 Test instructions

Clone solid-js, import something from it, add something like this to package.json

```
"targets": {
    "ie11": {
      "engines": {
        "browsers": "ie 11"
      }
    }
  }
```
and build -> you have arrow functions

## ✔️ PR Todo

- [X ] Filled out test instructions (In case there aren't any unit tests)
- [ X] Included links to related issues/PRs

### DIFF
--- a/packages/core/parcel/README.md
+++ b/packages/core/parcel/README.md
@@ -14,7 +14,7 @@
 
 - 🚀 **Blazing fast** bundle times - multicore compilation, and a filesystem cache for fast rebuilds even after a restart.
 - 📦 Out of the box support for JS, CSS, HTML, file assets, and more - **no plugins to install**.
-- 🐠 **Automatically transforms modules** using Babel, PostCSS, and PostHTML when needed - even `node_modules`.
+- 🐠 **Automatically transforms modules** using Babel, PostCSS, and PostHTML when needed.
 - ✂️ Zero configuration **code splitting** using dynamic `import()` statements.
 - 🔥 Built in support for **hot module replacement**
 - 🚨 Friendly error logging experience - syntax highlighted code frames help pinpoint the problem.


danieltroger:
In case anyone wonders thanks to @mischnic I have the ultimate workaround.
Basically do this: https://www.ryanelainska.com/how-to-patch-an-npm-package-with-yarn-patch

With a patch like this

```
diff --git a/lib/summarizeRequest.js b/lib/summarizeRequest.js
index d13efe8fcaf441d5488b60c0e666deff568389d9..6736674448e373f7e9dea9df7e3e640cb6002143 100644
--- a/lib/summarizeRequest.js
+++ b/lib/summarizeRequest.js
@@ -56,6 +56,14 @@ async function summarizeRequest(fs, req) {
 }
 
 function isFilePathSource(fs, filePath) {
+  if (process.env.FORCE_TRANSPILE) {
+    for (const thing of process.env.FORCE_TRANSPILE.split(",")) {
+      if (filePath.includes(thing)) {
+        console.log("FORCE TRANSPILING", filePath);
+        return true;
+      }
+    }
+  }
   return !filePath.includes(NODE_MODULES);
 }

```

(use [resolutions](https://next.yarnpkg.com/configuration/manifest#resolutions) in yarn berry and target `@parcel/core`

And then you can force transpile things via an environment variable

devongovett:
FWIW, Solid doesn't support IE11 anyway so not sure what good transpiling it will do https://github.com/solidjs/solid#browser-support

danieltroger:
@devongovett yeah, that's my next problem :( but I can deal with that from my javascript code and disable the solid-parts or just let it fail silently like it does WHEN the code runs. But if it has arrow functions it doesn't run at all so even my non-solid code can't do anything about it nor provide its functionality

danieltroger:
@devongovett also it, transpiled and with polyfills, works in for example firefox 48 which isn't officially supported. So why not support it when all I need for that is a little transpilation?

devongovett:
I think we are probably going to change this we just haven't figured out exactly how it should work yet. I think only some syntax should be allowed, eg no non-standard stuff like JSX or TypeScript and no future JS proposals.

danieltroger:
> no non-standard stuff like JSX or TypeScript 

Hahaha did building sentry with typescript as node_modules in #6176 feel so wrong?

For my part I'd obviously appreciate if everything was supported because I need to find another way to build sentry otherwise but I totally understand it's strictly speaking "abuse" of what parcel as tool is intended to do.

>  think we are probably going to change this we just haven't figured out exactly how it should work yet.

But yeah, I totally agree. You need to solve it somehow, rather sooner than later, but it's a very difficult problem while maintaining a "zero-config" approach. Like, I don't even have any good ideas to solve it lol. But I'm sure you'll come up with something ingenious 👍 

For now it's not supported tho, alas this PR.

And also I really needed to get my stuff done today and not in a month which is why I had to figure out the patching.

danieltroger:
@devongovett FYI I got solid working in IE11 :dances furiously:

mischnic:
The readme was rewritten in the meantime and that statement was removed

danieltroger:
btw I think you're still saying "including node modules" on opencollective https://opencollective.com/parcel

> 🐠 Automatically transforms modules using Babel, PostCSS, and PostHTML when needed - even node_modules.

Issue #1354, opened 2018-05-12
parcel dev server will serve index.html when `--public-url` is set to  `./` (dev server will serve index.html to anything unrecognized)


doomleika:
# 🐛 bug report

As title. I have used `--public-url ./` as a temporary workaround within VSCode's sourcemap behavior [documented here](https://github.com/Microsoft/vscode-chrome-debug/issues/653#issuecomment-386656211). This will made asset linked within compiled html from `<script src="/js.8e7f6a41.js">` to `<script src="js.8e7f6a41.js">`. Without the slash the dev server will consider this unknown path and serve index.html. By using other dev server solution like `http-server` to serve `dist` will not exhibit such problem.

I believe this behavior is more confusing than helpful to most users.

## 🤔 Expected Behavior

Have dev server act more similarly to regular http server than automatically try to 'correct' unknown path 

## 😯 Current Behavior

Dev server will serve any path unregonized with index.html, and many magic seems to be confusing to me.

DeMoorJasper:
This is to support Single Page Applications, as these might have internal front-end routing, that can't work without redirecting everything to the index.html file

doomleika:
Thanks for your reply. but is serve index.html with request like `js.8e7f6a41.js` as `index.html` seems to be rather confusing. Is there anyway to workaround that. Without resorting external package like `http-server` of course.

doomleika:
This solution fits my use perfectly, but I am not really a frontend expert. Not sure if this will break somebody else's setup.

domenkozar:
> This is to support Single Page Applications, as these might have internal front-end routing, that can't work without redirecting everything to the index.html file

maybe that could be non-default behaviour enabled by `--spa-routing`?

patrickhulce:
I'm running into this issue as well even on just the index page. I guess my question is why is this considered an unknown path? When on `/`/`/index.html`/etc, `./bundle.0af421.js` resolves to the exact same URL as `/bundle.0af421.js` so shouldn't it just serve the correct file like normal?

alex-benoit:
running into this as well

I use parcel to build a simple html css js page (using it for minification, autoprefixer, scss, etc.)
I used to run it like this `parcel index.html`

Just noticed the multiple page support so decided to add a `privacy.html` and now run parcel as `parcel index.html privacy.html` and build as such `parcel build index.html privacy.html --public-url . --out-dir docs/` (for github pages support)

[...]

PR #6917, opened 2021-09-12
add ability to perform internal rewrite on serve

Magi-KS:
# ↪️ Pull Request

might help with:
Issue https://github.com/parcel-bundler/parcel/issues/1354
have conflicting interest with:
PR https://github.com/parcel-bundler/parcel/pull/3114

I am using parcel as a static site generator `parcel serve './src/contents/**/*.*'`
contents are then converted into `*.html` in one of the html page there will be a link to another html file, but I wanted the path to not contain `.html`.
`article/js` :heavy_check_mark: 
`article/js.html` :x: 
with an [internal rewrite](https://www.npmjs.com/package/http-rewrite-middleware) this is achievable.

## important-ish
had to remove `req.originalUrl` for this thing to work, check the conflicting interest [PR](https://github.com/parcel-bundler/parcel/pull/3114).
if `req.originalUrl` is necessary, this PR is still useful because it is possible to perform a HTTP redirect from `article/js` to `article/js.html`

## difference with proxy
the difference between this rewrite and the [proxy module](https://v2.parceljs.org/features/development/#api-proxy) is that the proxy module makes a http request to the target with/without path rewrite.
while the rewrite module changes the request `path` before passing it along to parcel's hander

## 💻 Examples
when i visit `/article/js` it should internally fetch `/article/js.html` 

## 🚨 Test instructions
create folder with 
`cake.html`
```
contains cake
```
`cookie.html`
```
contains cookie
```
`.rewriterc`
```
[
{"from": "^([^.]+)$", "to": "$1.html"}
]
```
the rule basically check for paths without `.` and appends `.html`
run `/path/to/parcel/packages/core/parcel/src/bin.js serve --log-level verbose '*'`
visit `/cake` should show contains cake
visit `/cookie` should show contains cookie
## ✔️ PR Todo

- [ ] debate `req.originalUrl || req.url`
- [ ] act on debate decision.

### DIFF
--- a/packages/reporters/dev-server/package.json
+++ b/packages/reporters/dev-server/package.json
@@ -26,7 +26,6 @@
     "connect": "^3.7.0",
     "ejs": "^3.1.6",
     "http-proxy-middleware": "^1.0.0",
-    "http-rewrite-middleware": "^0.1.6",
     "nullthrows": "^1.1.1",
     "serve-handler": "^6.0.0",
     "ws": "^7.0.0"
--- a/packages/reporters/dev-server/package.json
+++ b/packages/reporters/dev-server/package.json
@@ -26,6 +26,7 @@
     "connect": "^3.7.0",
     "ejs": "^3.1.6",
     "http-proxy-middleware": "^1.0.0",
+    "http-rewrite-middleware": "^0.1.6",
     "nullthrows": "^1.1.1",
     "serve-handler": "^6.0.0",
     "ws": "^7.0.0"
--- a/packages/reporters/dev-server/src/Server.js
+++ b/packages/reporters/dev-server/src/Server.js
@@ -28,6 +28,7 @@ import ejs from 'ejs';
 import connect from 'connect';
 import serveHandler from 'serve-handler';
 import {createProxyMiddleware} from 'http-proxy-middleware';
+import rewriteModule from 'http-rewrite-middleware';
 import {URL} from 'url';
 
 function setHeaders(res: Response) {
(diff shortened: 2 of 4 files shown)


mischnic:
Is this not already possible with https://v2.parceljs.org/features/development#api-proxy?

Magi-KS:
I've tried this
`.proxyrc`
```
{
  "/": {
    "target": "http://localhost:1234",
    "pathRewrite": {
      "^([^.]+)$": "$1.html"
    }
  }
}
```
and got this
`console: [HPM] Error occurred while proxying request localhost:1234/prototype.html to http://localhost:1234/ [EADDRNOTAVAIL] (https://nodejs.org/api/errors.html#errors_common_system_errors)`

seems to have problem when it's calling it self.

Magi-KS:
how can I help to move this forward?

mischnic:
This `.proxyrc.js` would work if we removed the `req.originalUrl` usages:

```js
module.exports = function (app) {
  app.use((req, res, next) => {
    if (!req.url.includes(".")) {
      req.url += ".html";
    }
    next();
  });
};
```

I would rather not add another dev server config file in addition to proxyrc.

`.originalUrl` was added here: https://github.com/parcel-bundler/parcel/pull/3114. Maybe we should just revert that, being able to rewrite the URL seems useful. @DeMoorJasper

Magi-KS:
oh nice, didn't realize:
```
a .proxyrc.js file allows you to attach any connect-compatible middleware.
```
so we just need to determine to remove the `.originalUrl` or not, then i just need to have `.proxyjs` file setup and it will work as intended.

Magi-KS:
I've reverted all my changes and made a commit to revert what was done in https://github.com/parcel-bundler/parcel/pull/3114.
if this is desired, feel free to merge.
if not feel free to close the PR as well if `originalUrl` is to be kept.

I am able to achieve what I wanted with what @mischnic suggested using `.proxyrc.js` and replacing the `originalUrl` instead of `url`.

Magi-KS:
I'll close this since it is not getting any traction.

PR #6922, opened 2021-09-14
Store graph edges in SharedArrayBuffer

lettertwo:
# Store graph edges in SharedArrayBuffer

This changes the low level storage mechanism for graph adjacency lists from multiple [JS Maps] to a custom hashmap (😅 ) implemented on top of [SharedArrayBuffer]. 

---
Background: Why would we want to implement our own hashmap?

## Parallelism

Parcel has the ability to run many operations in parallel, but Node does not have any way to share data between threads except by serializing data and sending it to *each worker*, where it is then deserialized *by each worker* before work commences. Even for modestly sized Parcel projects, this overhead is significant. 

JavaScript _does_ have support for sharing memory between threads via [SharedArrayBuffer], but unfortunately, there is currently nothing higher level (like a `SharedMap` or something) in JS or Node that can take advantage of this in a significant way.

Thus, having a custom data structure built on top of `SharedArrayBuffer` means that Parcel can build a very large graph and share some significant portion of that graph with workers *for free*.

## Caching

Parcel caches build state frequently, and that involves serializing data to be written to cache. an `ArrayBuffer` of data serializes more quickly than a `Map`. 

## Other Solutions Considered

We evaluated other options, such as [Protocol Buffers] and [FlatBuffers] (either of which may still be viable for storing node data in the future), but ruled them out for this work for the following reasons:
- Any transport format still requires serialization and deserialization, which is inherently less efficient than sharing memory
-  These solutions generally incur a build step to generate bindings for reading and writing data to the underlying transport format
- The adjacency list only stores integers (node ids, edge types) which makes it good candidate for storage  in an `ArrayBuffer`

---

We've added a small set of unit tests to exercise the basic assumptions about our implementation, but we are mostly relying on the existing suites of tests to verify this work.

We have also done extensive benchmarking to measure the impact of these changes in a real-world, significantly sized app.

---

> **EDIT:** Refactoring based on feedback seems to have improved the overall performance enough to eliminate most of the regressions!

Benchmarks: Testing the impact of these changes in a real app

---

These benchmarks were taken on a Parcel app with a bundle graph containing **~92,421 nodes** and **~255,538 edges** and a request graph containing **~163,633 nodes** and **~1,351,038 edges**.

Some definitions:

<dl>
<dt>Cold Build</dt>
<dd>A parcel production build with no cache.</dd>
<dt>Warm Build</dt>
<dd>A parcel production build from cache.</dd>
<dt>Cold Start</dt>
<dd>A parcel dev start with no cache.</dd>
<dt>Warm Start</dt>
<dd>A parcel dev start from cache.</dd>
<dt>Total {Serialization,Deserialization} Time</dt>
<dd>How much time is spent serializing or deserializing the graphs (including node data)</dd>
<dt>Edge {Serialization,Deserialization} Time</dt>
<dd>How much time is spent serializing or deserializing the adjacency list (just the edge data)</dd>
</dl>

| Cold Build                 |         v2 |    this PR |      change | change (%) |
| -------------------------- | ---------: | ---------: | ----------: | ---------: |
| Total Memory               | 3397.57 mb | 1755.89 mb | -1641.68 mb |   48% less |
| Total Serialization Time   |    4941 ms |    3309 ms |    -1632 ms | 33% faster |
| Edge Serialization Time    |    2372 ms |    1456 ms |     -916 ms | 39% faster |
| Total Deserialization Time |    5738 ms |    4123 ms |    -1615 ms | 28% faster |
| Edge Deserialization Time  |    2994 ms |     518 ms |    -2476 ms | 83% faster |
| init                       |       1 ms |       1 ms |       +0 ms | 16% slower |
| build                      |  163023 ms |  135714 ms |   -27309 ms | 17% faster |
| shutdown                   |    2991 ms |    4517 ms |    +1525 ms | 51% slower |
| ✨ Built in                |   146.39 s |   135.69 s |    -10.70 s |  7% faster |

| Warm Build                 |         v2 |   this PR |     change | change (%) |
| -------------------------- | ---------: | --------: | ---------: | ---------: |
| Total Memory               | 1304.12 mb | 946.77 mb | -357.35 mb |   27% less |
| Total Serialization Time   |    6469 ms |   5379 ms |   -1090 ms | 17% faster |
| Edge Serialization Time    |    3109 ms |   1438 ms |   -1671 ms | 54% faster |
| Total Deserialization Time |    6223 ms |   4563 ms |   -1660 ms | 27% faster |
| Edge Deserialization Time  |    3028 ms |    495 ms |   -2533 ms | 84% faster |
| init                       |    1994 ms |    744 ms |   -1250 ms | 63% faster |
| build                      |    4007 ms |   3304 ms |    -703 ms | 18% faster |
| shutdown                   |    2678 ms |   3901 ms |   +1223 ms | 46% slower |
| ✨ Built in                |     4.95 s |    3.79 s |    -1.16 s | 23% faster |

| Cold Start                 |         v2 |    this PR |     change | change (%) |
| -------------------------- | ---------: | ---------: | ---------: | ---------: |
| Total Memory               | 2337.65 mb | 1384.07 mb | -953.58 mb |   41% less |
| Total Serialization Time   |    4110 ms |    3049 ms |   -1061 ms | 26% faster |
| Edge Serialization Time    |    2358 ms |    1329 ms |   -1029 ms | 44% faster |
| Total Deserialization Time |    6441 ms |    3311 ms |   -3130 ms | 49% faster |
| Edge Deserialization Time  |    3141 ms |     356 ms |   -2785 ms | 89% faster |
| init                       |       1 ms |       2 ms |      +1 ms | 41% slower |
| build                      |   79695 ms |   69485 ms |  -10211 ms | 13% faster |
| save                       |     628 ms |     415 ms |    -213 ms | 34% faster |
| shutdown                   |    2565 ms |    3490 ms |    +925 ms | 36% slower |
| ✨ Built in                |    76.64 s |    69.47 s |    -7.17 s | 19% faster |

| Warm Start                 |         v2 |   this PR |     change | change (%) |
| -------------------------- | ---------: | --------: | ---------: | ---------: |
| Total Memory               | 1195.51 mb | 807.47 mb | -388.04 mb |   32% less |
| Total Serialization Time   |    6756 ms |   4198 ms |   -2558 ms | 38% faster |
| Edge Serialization Time    |    2326 ms |   1498 ms |    -828 ms | 36% faster |
| Total Deserialization Time |    6254 ms |   4073 ms |   -2181 ms | 35% faster |
| Edge Deserialization Time  |    2897 ms |    501 ms |   -2396 ms | 83% faster |
| init                       |    1934 ms |    907 ms |   -1028 ms | 53% faster |
| build                      |    3382 ms |   2418 ms |    -964 ms | 29% faster |
| save                       |     999 ms |    464 ms |    -535 ms | 54% faster |
| shutdown                   |    2564 ms |   3067 ms |    +503 ms | 20% slower |
| ✨ Built in                |     2.57 s |    2.42 s |    -0.15 s |  6% faster |

---

## Summary

- Total memory usage is down significantly across all tested conditions
- Serialization and deserialization are both significantly faster in all tested conditions
- ~~With warm cache,~~ builds are on par **or faster** in all tested conditions
- Shutdown times seem to be slower, which we did not expect, and requires further investigation.

~~While we have seen some regression with cold build times, and not much improvement with warm times,~~ We have seen no regression (and even some improvement!) in build times, and we have also seen significant improvement in memory usage and serialization and deserialization times in all build scenarios.  However, shutdown times are slower than expected.

The main reasons we think these are acceptable trade-offs are:

- We believe we have not significantly regressed the most common DX scenarios
- These results reflect one part of the graph improvements we are working toward. Our hope and expectation is that we will ultimately improve all of these metrics in future by backing the graph nodes in shared buffers like we have done for the edges here.

For this work, we did not address data stored in the graph nodes, focusing solely on the edges. Since there are generally many more edges than nodes in most of Parcel's graphs, this change represents a measurable improvement, even though there is still more to do to improve the efficiency of the graph structure. 

[JS Maps]: https://github.com/parcel-bundler/parcel/blob/fb9818a/packages/core/graph/src/Graph.js#L573-L623
[SharedArrayBuffer]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/SharedArrayBuffer
[Protocol Buffers]: https://developers.google.com/protocol-buffers
[FlatBuffers]: https://google.github.io/flatbuffers/


packages/core/graph/src/AdjacencyList.js
```diff
@@ -0,0 +1,1195 @@
+// @flow
+import assert from 'assert';
+import {fromNodeId, toNodeId} from './types';
+import type {NullEdgeType, AllEdgeTypes} from './Graph';
+import type {NodeId} from './types';
+
+/**
```
lettertwo, line 7: The implementation described here was inspired by [This dissection of the v8 Map internals](https://itnext.io/v8-deep-dives-understanding-map-internals-45eb94a183df)

packages/core/graph/src/AdjacencyList.js
```diff
@@ -0,0 +1,1195 @@
...
+  nodeCapacity?: number,
+|};
+
+opaque type EdgeHash = number;
+
+/** The index of the edge in the edges array. */
+opaque type EdgeIndex = number;
+
+// From https://gist.github.com/badboy/6267743#32-bit-mix-functions
+function hash32shift(key: number): number {
```
lettertwo, line 256: This crazy bit mixing function improves the uniformity of our hash table a fair amount without adding any noticeable time to hashing. We found this in the [v8 Map hash utils](https://github.com/nodejs/node/blob/238104c531219db05e3421521c305404ce0c0cce/deps/v8/src/utils/utils.h#L213)

packages/core/graph/src/AdjacencyList.js
```diff
@@ -0,0 +1,1195 @@
...
+  }
+
+  /**
+   * Returns a serializable object of the nodes and edges in the graph.
+   */
+  serialize(): SerializedAdjacencyList<TEdgeType> {
+    return {
+      nodes: this.#nodes,
+      edges: this.#edges,
+    };
```
lettertwo, line 399: The improvements to serialization and deserialization times come here; `nodes` and `edges` are both `SharedArrayBuffer` instances.

packages/core/graph/src/AdjacencyList.js
```diff
@@ -0,0 +1,1195 @@
...
+      | NullEdgeType
+      | Array<TEdgeType | NullEdgeType> = 1,
+  ): $ReadOnlySet<NodeId> {
+    return new Set(this.getNodesConnectedFrom(from, type));
+  }
+
+  /**
+   * Get the list of nodes connected from this node.
+   */
+  getNodesConnectedFrom(
```
thebriando: These functions can probably be renamed to `getNodeIdsConnected..` to match the ones in `Graph.js`.

parcel-benchmark:
## Benchmark Results

#### Timings

| --- | --- | --- |
| Cold | 1.95s | +39.00ms |
| Cached | 235.00ms | -26.00ms 🚀 |

#### Cold Bundles

*No bundle changes detected.*

#### Cached Bundles

| dist/legacy/index.96016b08.js | 1.59kb | +0.00b | 784.00ms | -41.00ms 🚀 |
| dist/legacy/index.8aaa89c9.js | 1.20kb | +0.00b | 785.00ms | -40.00ms 🚀 |
| dist/modern/index.6be20f01.js | 1.13kb | +0.00b | 784.00ms | -41.00ms 🚀 |

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 9.37s | -20.00ms |
| Cached | 402.00ms | -41.00ms 🚀 |

#### Cold Bundles

*No bundle changes detected.*

#### Cached Bundles

*No bundle changes detected.*

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 1.02m | -83.00ms |
| Cached | 1.27s | -280.00ms 🚀 |

#### Cold Bundles

| dist/EmojiPickerComponent.07ae4435.js | 147.12kb | +0.00b | 18.94s | -9.59s 🚀 |
| dist/esm.f1a592e5.js | 33.15kb | +0.00b | 18.94s | -9.59s 🚀 |
| dist/DatePicker.f2994e45.js | 22.98kb | +0.00b | 18.94s | -9.59s 🚀 |
| dist/js.660b75db.js | 17.25kb | +0.00b | 34.16s | -11.87s 🚀 |
| dist/png-chunks-extract.5c9c0fb9.js | 3.58kb | +0.00b | 34.16s | -11.87s 🚀 |
| dist/feedback.c7e69064.js | 1.77kb | +0.00b | 18.93s | -9.59s 🚀 |
| dist/workerHasher.3639b1f9.js | 1.63kb | +0.00b | 18.94s | -9.59s 🚀 |
| dist/heading6.a693146e.js | 1.51kb | +0.00b | 18.93s | -9.59s 🚀 |
| dist/heading5.8bfafe21.js | 1.38kb | +0.00b | 18.93s | -9.59s 🚀 |
| dist/expand.a44d10fe.js | 1.29kb | +0.00b | 18.93s | -9.59s 🚀 |

#### Cached Bundles

*No bundle changes detected.*

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 6.52s | +219.00ms |
| Cached | 318.00ms | -99.00ms 🚀 |

#### Cold Bundles

*No bundle changes detected.*

#### Cached Bundles

*No bundle changes detected.*

</p></details>

**[Click here to view a detailed benchmark overview.](https://benchmark.parceljs.org/benchmark/315376143717892674)**

packages/core/graph/src/AdjacencyList.js
```diff
@@ -0,0 +1,1195 @@
...
+   */
+  hash(from: NodeId, to: NodeId, type: TEdgeType | NullEdgeType): EdgeHash {
+    // A crude multiplicative hash, in 3 steps:
+    // 1. Serialize the args into an integer that reflects the argument order,
+    // shifting the magnitude of each argument by the sum
+    // of the significant digits of the following arguments,
+    // .e.g., `hash(10, 24, 4) => 10244`.
+    // $FlowFixMe[unsafe-addition]
+    // $FlowFixMe[incompatible-type]
+    let hash = '' + from + to + type - 0;
```
devongovett: This seems like it could be a perf issue. Converting numbers to strings and back is expensive and this is a very hot function.

Also, your hash32shift function only works on 32 bit integers, and this could easily create a number larger than that. Your node ids are up to 32 bits a piece already and you are concatenating them here.

Did you try some kind of hash combining function? No idea how good this on is but it might be faster?

```js
let hash = 17;
hash = hash * 37 + hash32shift(from);
hash = hash * 37 + hash32shift(to);
hash = hash * 37 + hash32shift(type);
```

packages/core/graph/src/AdjacencyList.js
```diff
@@ -0,0 +1,1195 @@
...
+ * This cache is used to speed up iterations that are grouped by type,
+ * such as `getEdgesConnectedFrom` and `getEdgesConnectedTo`,
+ * as well as `getOutboundEdgesByType` and `getInboundEdgesByType`.
+ */
+class TypeMap<TEdgeType: number = 1> {
+  #data: AdjacencyList<TEdgeType>;
+  /** A map of node ids from => through types => to node ids. */
+  #from: Map<NodeId, Map<TEdgeType | NullEdgeType, Set<NodeId>>> = new Map();
+  /** A map of node ids to => through types => from node ids. */
+  #to: Map<NodeId, Map<TEdgeType | NullEdgeType, Set<NodeId>>> = new Map();
```
devongovett: How much do these extra maps affect performance? We're basically paying the cost of 3 hash maps instead of just one, so basically the cost of our old implementation PLUS the cost of your new implementation (minus serializing these maps). Explains why it is slightly slower in some cases.

Would it make sense to adjust the main adjacency list data structure instead to be more oriented around iterating by type? It's pretty rare that we access all of the edges connected to a node across all edge types. Most of the time, we are asking for a single type. One idea could be to change the NEXT_IN, NEXT_OUT, etc. pointers to be the next edge of the same type rather than having a single linked list of every edge connected to a node. The nodes would gain a type in addition to FIRST_IN/FIRST_OUT etc., so basically there'd be a "node" per id + type. This way you'd basically have a separate linked list per edge type and could simply walk that to iterate the connected edges. If you wanted all edge types, you could instead pass a list of those (which we know upfront).

packages/core/graph/src/AdjacencyList.js
```diff
@@ -0,0 +1,1195 @@
...
+    if (fromNodeId(from) < 0 || fromNodeId(from) >= this.#nodes[COUNT]) {
+      throw new Error(`Unknown node ${String(from)}`);
+    }
+    if (fromNodeId(to) < 0 || fromNodeId(to) >= this.#nodes[COUNT]) {
+      throw new Error(`Unknown node ${String(to)}`);
+    }
+    if (type <= 0) throw new Error(`Unsupported edge type ${0}`);
+
+    // The edge is already in the graph; do nothing.
+    if (this.hasEdge(from, to, type)) return false;
```
devongovett: I think this results in hashing twice when the edge is not already in the graph (the common case). You could inline this check below after hashing maybe?

packages/core/graph/src/AdjacencyList.js
```diff
@@ -0,0 +1,1195 @@
...
+    let count = 0;
+    let len = this.#nodes.length;
+    for (let i = NODES_HEADER_SIZE; i < len; i += NODE_SIZE) {
+      if (count++ >= max) break;
+      yield nodeAt(i);
+    }
+  }
+
+  /** Iterate over outgoing edge hashes from the given `nodeId` the `AdjacencyList`. */
+  *iterateOutgoingEdges(nodeId: NodeId): Iterator<EdgeIndex> {
```
devongovett: How much do generators affect performance vs say using an array?

packages/core/graph/src/AdjacencyList.js
```diff
@@ -0,0 +1,1195 @@
...
+ * }
+ *
+ * ┌────────────────────────────────────┐
+ * │          EDGES_HEADER_SIZE         │
+ * ├────────────┬───────────┬───────────┤
+ * │  CAPACITY  │   COUNT   │  DELETES  │
+ * └────────────┴───────────┴───────────┘
+ *
+ * Each edge is represented with 8 4-byte chunks:
+ * The first 4 bytes are the edge type.
```
devongovett: We probably don't need to allocate that much space for the type? 1 byte (max 256 types) would be more than enough. Though I guess that makes things a little more complex and you'd have to deal with alignment...

packages/core/graph/src/AdjacencyList.js
lettertwo: Very interesting! We thought string conversion seemed like a not great idea, but it was never indicated as a bottleneck in our profiling, so we left it. However, I've done some microbenching with your suggested impl, and it is indeed faster, but does not appear to degrade uniformity at all, which seems like a no-brainer:

'hash' is string conversion
'hash2' is prime multiplication

163,633 nodes, 1,351,038 edges:

|       | ops/sec | max collisions | avg collisions |
| :---- | ------: | -------------: | -------------: |
| hash  | 505,108 |              8 |           0.46 |
| hash2 | 884,441 |              9 |           0.46 |

92,421 nodes, 255,538 edges:

|       |   ops/sec | max collisions | avg collisions |
| :---- | --------: | -------------: | -------------: |
| hash  |   924,997 |              5 |           0.34 |
| hash2 | 1,384,921 |              7 |           0.34 |

packages/core/graph/src/AdjacencyList.js
```diff
@@ -0,0 +1,1205 @@
...
+const MIN_EDGE_CAPACITY = 2;
+/** The largest possible edge capacity. */
+const MAX_EDGE_CAPACITY = Math.floor(
+  // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/Invalid_array_length#what_went_wrong
+  (2 ** 31 - 1 - EDGES_HEADER_SIZE) / EDGE_SIZE / BUCKET_SIZE,
+);
+/** The size after which to grow the edge capacity by the minimum factor. */
+const PEAK_EDGE_CAPACITY = 2 ** 18;
+
+export const ALL_EDGE_TYPES: AllEdgeTypes = '@@all_edge_types';
```
thebriando: nit: You can get rid of this declaration and import it from `Graph`

packages/core/graph/src/AdjacencyList.js
```diff
@@ -0,0 +1,1210 @@
...
+  inspect(): any {
+    return {
+      nodes: this.#nodes.inspect(),
+      edges: this.#edges.inspect(),
+    };
+  }
+}
+
+/**
+ * `SharedTypeMap` is a hashmap of items,
```
lettertwo, line 498: The implementation described here was inspired by [This dissection of the v8 Map internals](https://itnext.io/v8-deep-dives-understanding-map-internals-45eb94a183df)

packages/core/graph/src/AdjacencyList.js
```diff
@@ -0,0 +1,1210 @@
...
+    hash = hash * 37 + hash32shift((to: any));
+    hash = hash * 37 + hash32shift((type: any));
+    // Finally, we map the hash to a value modulo the edge capacity.
+    hash %= this.capacity;
+    return hash;
+  }
+}
+
+// From https://gist.github.com/badboy/6267743#32-bit-mix-functions
+function hash32shift(key: number): number {
```
lettertwo, line 1167: This crazy bit mixing function improves the uniformity of our hash table a fair amount without adding any noticeable time to hashing. We found this in the [v8 Map hash utils](https://github.com/nodejs/node/blob/238104c531219db05e3421521c305404ce0c0cce/deps/v8/src/utils/utils.h#L213)

packages/core/graph/src/AdjacencyList.js
```diff
@@ -0,0 +1,1210 @@
...
+    if (prevIn !== null) this.#edges.linkIn(prevIn, edge);
+
+    // Link this edge to the node's list of outgoing edges.
+    let prevOut = this.#nodes.linkOut(fromNode, edge);
+    if (prevOut !== null) this.#edges.linkOut(prevOut, edge);
+
+    return true;
+  }
+
+  *getAllEdges(): Iterator<{|
```
lettertwo, line 311: This is the now the only scenario where we generate values for iteration. All other traversals are now using arrays.

packages/core/graph/src/Graph.js
```diff
@@ -292,19 +221,6 @@ export default class Graph<TNode, TEdgeType: number = 1> {
...
-    type: TEdgeType | NullEdgeType = 1,
-  ): void {
-    this._assertHasNodeId(fromNodeId);
-    for (let parent of this.inboundEdges.getEdges(fromNodeId, type)) {
-      this.addEdge(parent, toNodeId, type);
-      this.removeEdge(parent, fromNodeId, type);
-    }
-    this.removeNode(fromNodeId);
-  }
-
```
lettertwo, line 307: This method appears to be unused.

packages/core/graph/src/AdjacencyList.js
```diff
@@ -0,0 +1,1210 @@
...
+    this.#edges = copy.#edges;
+  }
+
+  /**
+   * Adds a node to the graph.
+   *
+   * Returns the id of the added node.
+   */
+  addNode(): NodeId {
+    let id = this.#nodes.getId();
```
lettertwo, line 229: Note that we don't _actually_ add an entry to the node map until it is connected  to another node id in `addEdge`.

packages/core/graph/src/AdjacencyList.js
```diff
@@ -0,0 +1,1210 @@
...
+      // If we're in danger of overflowing the `nodes` array, resize it.
+      if (this.#nodes.load >= LOAD_FACTOR) {
+        this.resizeNodes(increaseNodeCapacity(this.#nodes.capacity));
+        // We need to update our indices since the `nodes` array has changed.
+        toNode = this.#nodes.addressOf(to, type);
+        fromNode = this.#nodes.addressOf(from, type);
+      }
+    }
+    if (toNode === null) toNode = this.#nodes.add(to, type);
+    if (fromNode === null) fromNode = this.#nodes.add(from, type);
```
lettertwo, line 295: If this edge is the first _of this type_ for either of the nodes, we add a new entry to the node map.

packages/core/graph/src/AdjacencyList.js
```diff
@@ -0,0 +1,1210 @@
...
+
+  inspect(): any {
+    return {
+      nodes: this.#nodes.inspect(),
+      edges: this.#edges.inspect(),
+    };
+  }
+}
+
+/**
```
lettertwo, line 497: The implementation described here was inspired by [This dissection of the v8 Map internals](https://itnext.io/v8-deep-dives-understanding-map-internals-45eb94a183df)

packages/core/graph/src/AdjacencyList.js
```diff
@@ -0,0 +1,1210 @@
...
+  static ITEM_SIZE: number = 2;
+  /** The offset at which a link to the next item in the same bucket is stored. */
+  static #NEXT: 0 = 0;
+  /** The offset at which an item's type is stored. */
+  static #TYPE: 1 = 1;
+
+  /** The number of items to accommodate per hash bucket. */
+  static BUCKET_SIZE: number = 2;
+
+  data: Uint32Array;
```
lettertwo, line 578: There would be significant memory saved if we used `DataView` instead, which would allow us to store the `type` field in 1 byte (instead of 4), but it also would complicate the implementation significantly.

The savings would come from the following:
- each node adds an entry for every unique edge type it is connected to, so in the worst case (every node had at least one edge of every type), that's `n * t * 3` bytes saved, where `n` is the number of nodes and `t` is the number of unique edge types in the graph.
- each edge keeps its type in its entry, which would be `e * 3` bytes saved, where `e` is the number of edges in the graph.

Additionally, typed arrays are _technically_ not portable, since they use the originating system's [endianess](https://en.wikipedia.org/wiki/Endianness)

Some of the reasons we haven't implemented this yet:
- `DataView` is likely not as well optimized as type arrays
- memory allocation is more complex, since we cannot rely on the uniform item size of a typed array
- reads and writes must now be aware of endianness
- In practice, most systems are LE, so the non-portable nature of typed array is probably a non-issue
- memory savings from this implementation are already significant compared to v8 Maps

packages/core/graph/src/AdjacencyList.js
lettertwo: We refactored the data structure just as you suggested; These maps are gone now! 🎉

packages/core/graph/src/AdjacencyList.js
lettertwo: This is a really good point, and something I expanded on a bit in [this comment](https://github.com/parcel-bundler/parcel/pull/6922#discussion_r718652525).

thebriando (commented):
https://github.com/parcel-bundler/parcel/blob/8aa9c25df02bcfa21d74c59ed4c4bd31a6caaafa/packages/core/core/src/BundleGraph.js#L1210-L1213

I wonder if creating a new function to do this would have any benefit since we can traverse backwards with the doubly linked list now.

devongovett (approved):

devongovett:
🥳 🎉

PR #6948, opened 2021-09-16
Use automatic react runtime for v18.0.0

### LINKED ISSUE #7595: JSX automatic transform doesn't get used for 18+ prereleases
# 🐛 bug report

If you try using `react@next` or `react@experimental`, Parcel uses the old JSX transform, so it forces React to be in scope. But it should use the new one. Not sure how CRA is detecting it but it's worth aligning on the detection method IMO.

alexeyraspopov:
# ↪️ Pull Request

When using `react@next` and `react-dom@next`, NPM installs `v18.0.0-alpha-XXXXXXX`. Parcel uses `>= 17.0.0` to determine whether to use automatic runtime or React.createElement. The check fails for v18.x probably because of semver: `>=17.0.0` does not go beyond the major release.

## 🚨 Test instructions

Using a fresh nightly build `parcel@^2.0.0-nightly.838` with `react@^18.0.0-alpha-a8cabb564-20210915` (and similar `react-dom`, or simply `react@next`), writing any JSX without `import React from 'react'` throws an error `React is not defined` in browser.

## ✔️ PR Todo

### DIFF
--- a/packages/transformers/js/src/JSTransformer.js
+++ b/packages/transformers/js/src/JSTransformer.js
@@ -22,7 +22,7 @@ const JSX_PRAGMA = {
   react: {
     pragma: 'React.createElement',
     pragmaFrag: 'React.Fragment',
-    automatic: '>= 17.0.0',
+    automatic: '>= 17.0.0 || >= 18.0.0',
   },
   preact: {
     pragma: 'h',


mtgto:
> The check fails for v18.x probably because of semver: `>=17.0.0` does not go beyond the major release.

It seems to be a bit wrong. The problem is `semver.satisfies` returns false when target version is [prerelease](https://github.com/npm/node-semver#prerelease-tags).

I test with RunKit.
https://npm.runkit.com/semver

```js
console.log(semver.satisfies('18.0.0', '>= 17.0.0')); // true
console.log(semver.satisfies('18.0.0-alpha-a8cabb564-20210915', '>= 17.0.0')); // false
```

BTW, I suggest to use [coercion of semver](https://github.com/npm/node-semver#coercion) to compare versions including prelease libraries.

```js
console.log(semver.valid(semver.coerce('18.0.0-alpha-a8cabb564-20210915'))); // "18.0.0"
console.log(semver.satisfies(semver.valid(semver.coerce('18.0.0-alpha-a8cabb564-20210915')), '>= 17.0.0')); // true
```

alexeyraspopov:
@mtgto, I do agree `coerce()` makes sense. I guess the issue will be resolved once v18 is released, so there will be no need to update the version range, but every following alpha/beta/rc/nightly will keep breaking the condition

mischnic:
This looks like a cleaner solution:
```js
require("semver").satisfies("18.0.0-rc.0-next-13036bfbc-20220121", ">= 17.0.0", {includePrerelease: true})
===
true
```

@alexeyraspopov Could you fix that?

gaearon:
The approach doesn’t seem ideal because it won’t work for `0.0.0` versions tagged as `@experimental`. I’d suggest checking for the presence of `/jsx-runtime` entry point in the package.

mischnic:
We don't really want to do a path resolution for every single asset to determine this. I think it's safe to assume that using an experimental React version means that it's a recent one which supports the automatic runtime (and if not, these very few projects could still use the manualy jsconfig setting). So the check should be something like
`>=17 (including prereleases) OR 0.0.0 (including prereleases)`

gaearon:
Fair.

gaearon:
>for every single asset 

(I don't understand this part but your approach sounds good to me too.)

devongovett:
Opened a new PR to implement the above suggestion: #7642

PR #7035, opened 2021-10-07
Unmark defer for dependency that become used ('does not export')

AGawrys:
# ↪️ Pull Request

This PR addresses the nondeterministic "does not export" bug caused by a deferred dependency not unmarked deferred upon usage.

## 🚨 Test plan
 
- [x] Added integration test derived from @mischnic small reproduction 
- [x] Verify all passing integration tests
- [ ] Tested in large application

## ✔️ PR Todo


mischnic (approved):
The bug here was that when a dependency was resolved to an already existing (and deferred) asset group, `markParentsWithHasDeferred` didn't get called. This caused the new dependency to not be marked `hasDeferred=true`, which meant that it was never revisited in the asset graph traversal, so it was actually never checked if the asset group should be undeferred.

The fix is to call `markParentsWithHasDeferred` if the (potentially new) dependency is not marked with `hasDeferred=true`. Previously this "previously deferred" was checked on the asset group, which might have been added by another dependency.

Graph of a reproduction, the order causes another dependency to be added to a deferred asset group:

AssetGraph_Main_failed Kopie

devongovett (approved):

Issue #6061, opened 2021-03-27
HMR fails to connect when linked externally.


timdorr:
# 🐛 bug report

Since #6005 landed, I am unable to use HMR with Parcel acting as an external server. That is, I have another server that renders our HTML and includes the Parcel build output via a `<script>` tag. It uses the port of the HTML rendering server when trying to connect to the HMR WebSocket.

## 🎛 Configuration (.babelrc, package.json, cli command)

CLI command:

```sh
yarn parcel serve client/index.tsx --https --cert ./tls.crt --key ./tls.key
```

```html
<script defer src="https://localhost:1234/index.js"></script>
```

## 🤔 Expected Behavior

It should use the port number explicitly in the generated output for the HMR runtime.

## 😯 Current Behavior

My HTML rendering server runs on port 10000, so I get this message in my browser:

```
WebSocket connection to 'wss://localhost:10000/' failed: 
```

## 💁 Possible Solution

Side note: I noticed that `--hmr-port` does set the port in the runtime, but doesn't actually open a server on that port for updates. That might be a possible solution.

Maybe add a `--no-hmr-port` option to resolve the issue #6005 was trying to address, but revert the change for the default case. 

## 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Parcel           | 2.0.0-nightly.631 |
| Node             | 14.15.1 |
| npm/Yarn         | yarn@1.22.10 |
| Operating System | macOS Big Sur |

mischnic:
So all combinations I can think of are:
1. `parcel serve` and opening `http://localhost:1234/index.html`
2. `parcel watch` and bringing your own HTTP server
3. 1 but proxied (e.g. ngrok) to a different domain/port (`https://abc.ngrok.io`), ideally `http://localhost:1234` would still work fine
4. 1 but serving HTML youself  (Parcel serves JS)

101arrowz:
I just found that #6005 also breaks HMR for web extensions with `serve` because `location.port` doesn't actually work in an extension context. I guess a working solution is to specify a different `--hmr-port` but it's not optimal. 

EDIT: After looking again everything is fine with `watch` but sometimes you like to debug the extension pages from a web context instead of using the extension directly.

drmercer:
What I'm seeing on a fresh install (version 2.3.2) after following [the official Web Extension recipe](https://parceljs.org/recipes/web-extension/):

[...]

PR #7050, opened 2021-10-11
Add support for Web Extension manifest V3

### LINKED ISSUE #5865: V2: web-ext: generated manifest.json doesn't reference imported css
# 🐛 bug report

When I import a .css file into a .ts file, which is mentioned in `manifest.json` as a content-script - the aforementioned .css is not referenced in the `content_scripts` section of the generated `manifest.json`.

## 💻 Code Sample

```
`cs.ts`:
```ts
import './foo.css';
```

## 🤔 Expected Behavior

`manifest.json`:
```json
{
  "content_scripts": [
    {
      "matches": ["https://goo.gl/"],
      "js": ["/entry/content-script/cs.js", "/autoreload.6f291371.js"],
      "css": ["entry/content-script/cs.css"]
    }
  ],
}
```

## 😯 Current Behavior

`manifest.json`:
```json
{
  "content_scripts": [
    {
      "matches": ["https://goo.gl/"],
      "js": ["/entry/content-script/cs.js", "/autoreload.6f291371.js"]
    }
  ],
}
```
Plz note, that `entry/content-script/cs.css` **is** being correctly bundled and **is** present on disk.

## 💁 Possible Solution

Fix the algo that writes to `manifest.json` to include `"content_scripts": [     { "css": }  ]`property.

## 🔦 Context

Devving a chrome extn.

`.parcelrc`:
```json
{
  "transformers": {
    "*.{ts,tsx}": ["@parcel/transformer-typescript-tsc"]
  }
}

## 🌍 Your Environment

<!--- Include as many relevant details about the environment you experienced the bug in -->

| Parcel           | `2.0.0-nightly.579`

<!-- Love parcel? Please consider supporting our collective:
👉  https://opencollective.com/parcel/donate -->

_BTW, I I have a strong opinion that in `Issue template` the `Code Sample` section should go 2nd from beginning; and ` 🎛 Configuration` section - 2nd from the end._

### LINKED ISSUE #6079: Support manifest v3 for web extensions
# 🙋 feature request
Chrome (and Edge) support a newer version of the extension manifest, [v3](https://developer.chrome.com/docs/extensions/mv3/intro/mv3-overview/) but the webextension config plugin only supports version 2. The two are very similar, but some of the fields that name resources are different.

## 🤔 Expected Behavior

The `@parcel/config-webextension` package should support extensions that are defined with v3 of the manifests (i.e. point `parcel` at the `manifest.json` and let it package things up).

## 😯 Current Behavior

Currently when following the [Web Extension recipe](https://v2.parceljs.org/recipes/web-extension/) and using a v3 manifest, `parcel build` throws an error:
```
@parcel/transformer-webextension: Invalid Web Extension manifest
/path/to/my/extension/manifest.json:5:23
  4 |   "version": "1.0",
> 5 |   "manifest_version": 3,
>   |                       ^ Possible values: "2"
  6 |   "action": {
  7 |     "default_popup": "popup.html"
  ```

## 💁 Possible Solution

From parcel's point of view, the manifests are fairly similar.
- The [schema check](https://github.com/parcel-bundler/parcel/blob/a3aa86815e098c90fda72a08d29646ff059ab74d/packages/transformers/webextension/src/schema.js#L46) needs to allow version 3
- The new [`background.service_worker` property](https://developer.chrome.com/docs/extensions/mv3/getstarted/#background)
- The CSP setting has [changed from being a simple string](https://developer.chrome.com/docs/extensions/mv3/intro/mv3-migration/#fcontent-security-policy) so the [HMR logic](https://github.com/parcel-bundler/parcel/blob/a3aa86815e098c90fda72a08d29646ff059ab74d/packages/transformers/webextension/src/WebExtensionTransformer.js#L241-L261) will need to be modified to rewrite multiple policies instead of just one.
- The actions, which can refer to resources, have [moved under a new toplevel key](https://developer.chrome.com/docs/extensions/mv3/intro/mv3-migration/#action-api-unification)
- The schema for `web_accessible_resources` [changed slightly](https://developer.chrome.com/docs/extensions/mv3/intro/mv3-migration/#web-accessible-resources)
- There are additional [restrictions on how code is executed](https://developer.chrome.com/docs/extensions/mv3/intro/mv3-migration/#code-execution) and I do not know how this affects the bundles and HMR code for Parcel.

## 🔦 Context

Chrome is pushing this as the new way to write extensions and has removed much of the v2 documentation in favor of [pushing v3 by default](https://developer.chrome.com/docs/extensions/). A big architecture change for extensions seems to be the move to [using service workers + message passing instead of a background page](https://developer.chrome.com/docs/extensions/mv3/migrating_to_service_workers/). I am starting to write a new extension (not having done so before), so I am trying to follow the documentation on the Chrome site and set this project up for having the right long-term architecture, so following the v3 manifest feels like the best course, rather than creating a v2 manifest.

## 💻 Examples

There don't seem to be many official examples yet - even the official repo has a [limited selection](https://github.com/GoogleChrome/chrome-extensions-samples/blob/main/examples/hello-world/manifest.json).

### LINKED ISSUE #6494: [Web Extension Transformer] doesn't support `background.page`
# 🙋 feature request

`@parcel/transformer-webextension` doesn't pack the background page specified in the source `manijest.json` file.

## 🤔 Expected Behavior

The packed extension should have the `background.html` page with all assets (scripts, css, images, etc.)

## 😯 Current Behavior

Adding a background page to the `manifest.json`, like it's shown below, is ignored by the transformer:

```
{
...
  "background": {
    "page": "background.html",
    "persistent": true
  },
...
}
```

### LINKED ISSUE #7385: Web extension build not working with _locales, trying to process ".DS_Store"
Parcel version: 2.0.1
@parcel/config-webextension: 2.0.1
Node: v16.13.0
OS: macOS Monterey

Run the following to build web extension:
`parcel build src/manifest.json --target webext-prod`

Error:

> > parcel build src/manifest.json --target webext-prod
> 
> 🚨 Build failed.
> 
> @parcel/core: Failed to resolve '_locales/.DS_Store/messages.json' from './src/manifest.json'
> 
> @parcel/resolver-default: Cannot load file './_locales/.DS_Store/messages.json' in './src'.
> 💡 Did you mean '_./_locales/de/messages.json_'?
> 💡 Did you mean '_./_locales/fr/messages.json_'?
> 

The workaround is to use "--no-cache".
`parcel build src/manifest.json --target webext-prod --no-cache`

I also had to delete "_locales/.DS_Store", otherwise a similar error happens.

### LINKED ISSUE #7568: image url: pipeline fails for web-extension config production build
# 🐛 bug report

When using the url: pipeline for images in my react code, parcel non-production works fine but production (`parcel build`) fails with bundle resolution errors. The images are *not* used in a content script – only in a React app that is displayed in a devtools panel or popup window.

.parcelrc
```
{
  "extends": "@parcel/config-webextension"
}
```

package.json
```json
  "targets": {
    "webext-dev": {
      "source": [
        "src/manifest.json",
        "src/app/index.html"
      ],
      "distDir": "dist",
      "sourceMap": {
        "inline": true,
        "inlineSources": true
      }
    },
    "webext-prod": {
      "source": [
        "src/manifest.json",
        "src/app/index.html"
      ],
      "distDir": "dist",
      "sourceMap": false
    }
  },
  "scripts": {
    "start": "rimraf dist && parcel watch --no-hmr --target webext-dev",
    "start:hmr": "rimraf dist && parcel watch --host localhost --hmr-port 1235 --target webext-dev",
    "build": "rimraf dist && parcel build --target webext-prod --detailed-report",
    "test": "jest"
  },
```

## 🤔 Expected Behavior

No bundle resolution errors for my url: pipeline images.

## 😯 Current Behavior

When running my extension which spawns a popup window with my react app, an error for each url: image shows as
image

When looking at the minified file I can see that the three images I'm having trouble with are all required before they are later registered:
image
image

## 💁 Possible Solution

I assume this is due to the optimization processes that occur during a production build, but I'm unable to understand what is causing the issue.

## 💻 Code Sample

How I'm using the url: pipeline in code:
```jsx
import searchIcon from 'url:../../images/search.svg';
```

How I'm running my build
```
yarn build
```
which corresponds to the package.json entry shown above.

All three images causing issues are used in this way.

## 🌍 Your Environment

| Parcel           |2.2.1
| Node             |14.18.0
| npm/Yarn         |1.22.15
| Operating System |Windows 10

### LINKED ISSUE #7808: Webextension managed storage
# 🐛 bug report

I tried to setup a managed storage (https://developer.chrome.com/docs/extensions/reference/storage/)

```json
// manifest.json
"storage": {
    "managed_schema": "schema.json"
  },
```

but my `schema.json` is compiled to a `schema.js` and in the manifest is changed to

```json
"storage": {
    "managed_schema": "/schema.js"
  },
  ```
  
  two issue there, the first is that managed_schema should'nt have the leading slash

`Extension failed to load:  storage.managed_schema must be a relative path without .. `

and the second is that the json file should only be copy as is because managed_schema does not support js file

`Extension failed to load: Line: 1, column: 1, Unexpected token.`

## 🤔 Expected Behavior

`/schema.js` -> `schema.json`

## 🌍 Your Environment

"@parcel/config-webextension": "2.3.2",
"@parcel/config-default": "2.3.2",
"parcel": "2.3.2"

Is this realy an issue or I missed something ?

thanks.

### LINKED ISSUE #7835: transformer-webextension: Cannot add Vue generated .css file to web_accessible_resources
# 🐛 Bug report

I'm trying to inject a Vue app into a web page from a web extension content script, but I can't load the Vue generated CSS file from its Asset URL into the page because Parcel doesn't let me put the generated file's name into the `web_accessible_resources` property of the manifest.

### `.babelrc`
```js
{
    "extends": "@parcel/config-webextension"
}
```

## 🤔 Expected Behavior

Parcel should keep the `content.css` entry in `web_accessible_resources`.

## 😯 Current Behavior

`content.css` is removed from `web_accessible_resources` in the manifest during the build.

## 💁 Possible Solution

I don't really understand what the transformer does but I guess this TODO might have something to do with the fact the generated file is not found: https://github.com/parcel-bundler/parcel/blob/e294eafd9a49c056fb4b223c5ace5c0653428ede/packages/transformers/webextension/src/WebExtensionTransformer.js#L176

## 🔦 Context

I'm building a web extension that injects stuff into web pages using its content script.

One of the thing I want to inject is a Vue app, I managed to make the Vue app mount to the injected DOM element but one thing is missing: the components style.

I tracked down that a `content.css` file is generated by Vue but since I'm using a content script, Parcel understandably doesn't know how to inject it anywhere (normally it would inject it as a `<link rel="stylesheet">` element in the generated HTML file).

My solution was to try injecting the CSS file into the page myself using its Asset URL however when I try to add `content.css` to the 
`web_accessible_resources` in the manifest, it gets removed by Parcel.

I also tried to add a placeholder `content.css` file in my source files but then Parcel complains about the non-unique bundle name.

### `manifest.json`
```json
{
    "manifest_version": 2,
    "name": "Example",
    "description": "Example extension",
    "version": "0.0.1",
    "content_scripts": [
        {
            "matches": ["*://*/*"],
            "js": ["content.js"]
        }
    ],
    "web_accessible_resources": ["content.css"]
}
```

### `content.js`
```js
import browser from "webextension-polyfill";
import { createApp } from "vue";

import App from "./App.vue";

function injectStyle(url) {
    const linkElem = document.createElement("link");
    linkElem.rel = "stylesheet";
    linkElem.href = url;
    document.head.appendChild(linkElem);
}

function injectVue() {
    const appElem = document.createElement("div");
    appElem.id = "app";
    document.body.appendChild(appElem);

    createApp(App).mount("#app");
}

injectStyle(browser.runtime.getURL("content.css"));
injectVue();
```

### `App.vue`
```vue
<template>
    <p>Hello! :)</p>
</template>

<style scoped>
p {
    color: red;
}
</style>
```
## 🌍 Your Environment

| Parcel           | 2.3.2
| Node             | v14.19.0
| npm/Yarn         | yarn v1.22.17
| Operating System | Linux

101arrowz:
# ↪️ Pull Request

Adds support for Manifest V3 to `@parcel/transformer-webextension`.
Closes #6079
Fixes #6494
Fixes #7568 
Fixes #7385 
Fixes #7835 (mostly)
Fixes #7808 
Fixes #5865 

HMR is not working due to MV3 restrictions on CSP. Refreshing the page does update the extension though.

## ✔️ PR Todo


101arrowz:
While trying to add HMR, I came across [this issue](https://github.com/parcel-bundler/parcel/issues/6061#issuecomment-948226782), which is probably not a blocker because most people will just `parcel watch` but it is something to note.

I'm moving this out of draft stage because I really don't know how to get HMR working but this MV3 transformer does seem to work properly beyond that, so until we can work something out it's a useful addition.

packages/transformers/webextension/src/WebExtensionTransformer.js
```diff
@@ -265,32 +319,38 @@ function cspPatchHMR(policy: ?string) {
...
+        (function() {
+          ${AUTORELOAD_BG}
+        })();
+        ${code}
+      `);
+      return [asset];
+    } else {
+      const parsed = jsm.parse(code);
+      const data: any = parsed.data;
+      validateSchema.diagnostic(
```
101arrowz: I'm planning to rewrite this to check first which manifest version was provided because right now it just says "`manifest_version` must be `3`" when you provide an invalid MV2 extension.

101arrowz:
Finished adding tests, this PR should now be ready for merging.

packages/transformers/webextension/src/WebExtensionTransformer.js
```diff
@@ -267,30 +321,44 @@ function cspPatchHMR(policy: ?string) {
...
-    await collectDependencies(
-      asset,
-      data,
-      parsed.pointers,
-      Boolean(options.hmrOptions),
-    );
-    if (options.hmrOptions) {
-      // To enable HMR, we must override the CSP to allow 'unsafe-eval'
-      data.content_security_policy = cspPatchHMR(data.content_security_policy);
+    if (asset.pipeline == 'mv3-bg-sw') {
```
mischnic: The "proper" solution for this would be a runtime. `@parcel/runtime-hmr` achieves exactly this.

101arrowz: Should this be moved into the HMR runtime? If so how would we differentiate between a background script and a normal script in the runtime (to figure out what to inject)?

mischnic: Maybe create another specific runtime plugin for this? And make it add the extra entry only if the context is service-worker (or something like that, you need to prevent it from being added to every JS bundle).

101arrowz: Makes sense. This would also solve the problem of getting HMR working with a background page instead of background scripts in MV2. We'd need a way to inject some JS into an HTML file, would that also be possible from the same runtime?

mischnic: You can add additional JS entry assets to JS bundles, not to HTML.
But we already have some logic to ensure that in development, every HTML asset has at least one script tag so that HMR works:
https://github.com/parcel-bundler/parcel/blob/c8a64c9a533072c12681ca0805d1e4d7481ee53d/packages/transformers/html/src/HTMLTransformer.js#L47-L48

So it should be possible, you just have to figure out which bundle in the bundlegraph you want to add the asset to.

101arrowz: Switching to a runtime might take a bit, and I won't have much time to work on it for the next month so it might be best to make that change in a separate PR and just have the HMR issues for `background.page` and the suboptimal architecture be known limitations for now.

findmet:
Unsafe unfortunately not the right one for the job

LexSwed:
Thanks for the efforts! Can this work be revived?

101arrowz:
I don't really want to work on this PR anymore since it does work, it just has some design issues that can be fixed in the future (I'd be willing to work on a new PR to fix them next week). @mischnic could this be merged?

sarahkadar:
any update on this? it'd be great to have this

Q0:
Very much waiting for these changes. New extensions in webstore can only be load manifest V3.
Thanks for working on this PR!

irgendwr:
Thanks for your work on this @101arrowz!

This patch seems to work great on simple examples, however I ran into some issues on more complex ones.
I hope my issues are reproducible with the following descriptions. It might be possible that I somehow messed something up by building it myself and applying the patch via [patch-package](https://github.com/ds300/patch-package).

1. Specifying `web_accessible_resources` in a manifest v3 causes a build error.

```json
    "resources": ["images/*"],
    "matches": ["<all_urls>"]
}],
```

The following error refers to [this line here](https://github.com/parcel-bundler/parcel/blob/b71d9866bbf2e2434f665e2026669e31c2898950/packages/transformers/webextension/src/WebExtensionTransformer.js#L191).
```
@parcel/transformer-webextension: The "path" argument must be of type string. Received an instance of Array

  TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received an instance of Array
  at new NodeError (node:internal/errors:372:5)
  at validateString (node:internal/validators:120:11)
  at Object.join (node:path:1172:7)
  at collectDependencies (@parcel/transformer-webextension/lib/WebExtensionTransformer.js:215:67)
  at Object.transform (@parcel/transformer-webextension/lib/WebExtensionTransformer.js:370:13)
```

2. ~~It seems like it's not possible to use `import` in the background service worker.~~ (whoops, my bad; see [response below](https://github.com/parcel-bundler/parcel/pull/7050#issuecomment-1071433977))

```json
"background": {
    "service_worker": "background.ts"
}
```

```
@parcel/transformer-js: Service workers cannot have imports or exports without the `type: "module"` option.

  background.ts:1:1
  > 1 | import browser from 'webextension-polyfill';
  >   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    2 | import optionsStorage from './options-storage.js';
    3 | 

  💡 Add {type: 'module'} as a second argument to the navigator.serviceWorker.register() call.
  📝 Learn more: https://parceljs.org/languages/javascript/#classic-scripts
```

101arrowz:
> Specifying `web_accessible_resources` in a manifest v3 causes a build error.

This shouldn't happen, I'll fix that.

> It seems like it's not possible to use import in the background service worker.

It is but you need to specify that it's a module type service worker. This behavior is [consistent with browsers](https://stackoverflow.com/a/66408379/10930232).
```json
"background": {
    "service_worker": "background.ts",
    "type": "module"
}
```

irgendwr:
Thanks for the quick response :)

> you need to specify that it's a module type service worker. This behavior is [consistent with browsers](https://stackoverflow.com/a/66408379/10930232).

Whoops, my bad, I totally missed that. Now it works :sweat_smile:

101arrowz:
Seems there are more `web_accessible_resources` issues than I thought (ref #7835). I'll have to see if this is something that can be resolved easily.

fregante:
> Seems there are more `web_accessible_resources` issues than I thought

It'd be great to fix them but I really wish this PR could be merged since Parcel support for extensions is DOA for new extensions at the moment.

I hope some Parcel maintainer merges this soon. _Any_ support is better than no support.

irgendwr:
If anyone wants to give this patch a try, I've written down my steps on how to use it in a project [here](https://gist.github.com/irgendwr/8bdeefc8d715333a69e09fe83548a717). (sharing this here since it took me a while to figure out ^^)

Be aware that, as discussed above, there are still known issues that need to be resolved.

101arrowz:
I just fixed CSS inclusion in content scripts, runtime use for HMR support, `web_accessible_resource` bugginess, and some other issues but when I merged suddenly prettier reformatted the entire codebase :/ @DeMoorJasper @devongovett any way to fix this and/or get this merged now that the migration is complete?

packages/transformers/webextension/src/WebExtensionTransformer.js
101arrowz: I've finished the switch to a runtime system now, this logic is removed. Hopefully this should work better.

101arrowz:
In related news, the newly merged codebase for Parcel crashes with no error message while building itself after 3 seconds so I can't work on this PR anymore. I'm guessing it's because I'm using WSL, and there's some issue with a native package. So that's nice.

EDIT: Luckily managed to get it working again by building one package at a time (I think too many file handles were being created or something). Should be good to merge for sure at this point.

irgendwr:
> when I merged suddenly prettier reformatted the entire codebase :/

I strongly recommend fixing this, otherwise there is no way that this will be merged since it's nearly impossible to review the changes. Try the steps below and feel free to let me know if you need help.

```sh
# Warning: This will remove ALL local changes, commit or stash anything you want to keep.
# Reset to the commit before the messy merge
git reset --hard 8145bb7~1
# Replace 'upstream' with the upstream remote (pointing to this repo here), if it is named differently
git fetch upstream
git merge upstream/v2
# Then, resolve the merge conflict:
# Edit the file 'packages/configs/webextension/package.json', accept the *incoming* change
# Stage the change
git add packages/configs/webextension/package.json
# Edit the file 'packages/transformers/webextension/src/runtime/autoreload-bg.js', accept the *current* change
# Stage the change
git add packages/transformers/webextension/src/runtime/autoreload-bg.js
# Commit the merge with the resolved conflicts
git commit -m "Merge 'upstream/v2'"
# Cherry-pick the commits you did after the merge
git cherry-pick '69e355e^..49f08de'
# Finally, force push (because the history changed)
git push -f
```

I've cleaned this up in [a branch of my fork](https://github.com/irgendwr/parcel/commits/101arrowz/web-extension-mv3), you can view the clean changes [here](https://github.com/parcel-bundler/parcel/compare/v2...irgendwr:101arrowz/web-extension-mv3?expand=1).

101arrowz:
@irgendwr Thanks for the suggestions! I'm not very good with Git :) I managed to get it working just by reformatting with the latest version of Prettier instead of fiddling with the source tree. Only like 2 style changes now (which I think are correct anyway). Let me know if this new version works on your extension, thanks for your help!

irgendwr:
@101arrowz I've encountered another issue while testing:

`content_scripts` are not resolved correctly. The paths should be relative to the manifest and not the CWD (current working directory).

Example:

`src/manifest.json`
```
"content_scripts": [
	{
		"matches": ["<all_urls>"],
		"js": ["inject.js"],
		"run_at": "document_start"
	}
],
```

`parcel watch src/manifest.json --dist-dir distribution --no-cache --no-hmr`

Error: `no such file or directory, open '/path/to/repo/inject.js'`

Expected: It should open `src/inject.js`
Actual: It opens `inject.js` and fails, because that's not the correct path.

101arrowz:
Doesn't that happen because it thinks inject.js is a module since there's no leading `./`? As in it's a Parcel resolution thing, not this transformer.

This probably won't work either:

```js
// index.js
import code from 'inject.js';
```

fregante:
From what I understand, node_module imports are only followed in JS whereas other types need to specifically use `"npm:package-name"`. So in this case `"file.js"` should still be supported and relative to the current file.

- https://parceljs.org/features/dependency-resolution/#url-schemes
- complete example in a CSS file: https://parceljs.org/languages/css/#%40import
- example without `./` in an HTML file: https://parceljs.org/languages/html/#scripts

irgendwr:
> Doesn't that happen because it thinks inject.js is a module since there's no leading `./`?

Without the patch this works as expected (and this also worked the last time I tried this patch @ b71d986). The documentation also shows that it should work without leading `./`: https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/content_scripts

101arrowz:
After debugging a bit I found the issue. Apparently the function `asset.invalidateOnFileChange()` takes a path relative to the project root, while every other function takes a path relative to the asset's path. I wish this behavior were documented somewhere! Anyway it's fixed now.

irgendwr:
works fine now \o/

101arrowz:
@mischnic sorry to keep bothering you but could you take a look at this PR now? I think it's pretty close to completely functional, and I've written updated docs at https://github.com/parcel-bundler/website/pull/1014.

dougwithseismic:
Anything needed to push this over the edge? Happy to help.

101arrowz:
@dougwithseismic Not really beyond bothering the maintainers, this PR should be complete.

devongovett:
Thanks @101arrowz, and sorry for the delay. I will review tomorrow.

packages/core/integration-tests/test/webextension.js
```diff
@@ -78,11 +74,24 @@ describe('webextension', function () {
...
+        name: 'manifest.json',
+        assets: ['manifest.json'],
+      },
+      {
+        assets: ['background.js'],
+      },
+      {assets: ['popup.html']},
+      {assets: ['popup.css']},
+      {assets: ['popup.js', 'esmodule-helpers.js', 'bundle-url.js']},
+      {assets: ['content-script.js']},
```
devongovett, line 110: Should there be some assertions about the content of the manifest? I.e. covering the stuff done by the packager/runtime, including injecting referenced bundle urls etc.

packages/packagers/webextension/src/WebExtensionPackager.js
```diff
@@ -0,0 +1,67 @@
...
+              .map(relPath),
+          ),
+        ),
+      ];
+
+      war.push({
+        matches: contentScript.matches,
+        extension_ids: [],
+        resources: referencedBundles
+          .flatMap(b => bundleGraph.getChildBundles(b))
```
devongovett: Should this be recursive? Right now, this would only include referenced bundles and their immediate children. For example, what if a JS file referenced a CSS file which referenced an image?

packages/transformers/webextension/src/WebExtensionTransformer.js
```diff
@@ -75,10 +77,12 @@ async function collectDependencies(
       });
     }
     for (const locale of await fs.readdir(locales)) {
-      asset.addURLDependency(`_locales/${locale}/messages.json`, {
-        needsStableName: true,
-        pipeline: 'raw',
-      });
+      if (await fs.exists(path.join(locales, locale))) {
```
devongovett: Is this check needed? You just read the directory, it's unlikely the file got deleted in the meantime.

packages/transformers/webextension/src/WebExtensionTransformer.js
```diff
@@ -88,8 +92,10 @@ async function collectDependencies(
       for (const k of ['css', 'js']) {
         const assets = sc[k] || [];
         for (let j = 0; j < assets.length; ++j) {
+          if (k == 'js') {
+            asset.invalidateOnFileChange(path.join(assetDir, assets[j]));
```
devongovett: Why do we need to invalidate the manifest when the JS file changes (but not CSS?)? We also add a dependency below, so that should already capture the JS change...

packages/transformers/webextension/src/WebExtensionTransformer.js
```diff
@@ -279,18 +361,24 @@ export default (new Transformer({
       '@parcel/transformer-webextension',
       'Invalid Web Extension manifest',
     );
+    asset.setEnvironment({
```
devongovett: What's the reason for this? Generally we should avoid it as it can cause issues. If it is needed, you might also need to move it to the top of the transformer so that it is set before adding any dependencies. Setting the environment overrides the asset's id, which is also used in those dependencies, so  setting it later means the dependencies might not point to the right asset id anymore.

packages/transformers/webextension/src/WebExtensionTransformer.js
```diff
@@ -207,38 +221,97 @@ async function collectDependencies(
...
+      if (needRuntimeBG && !program.background?.page) {
+        if (!program.background) {
+          program.background = {};
+        }
+        if (!program.background.scripts) {
+          program.background.scripts = [];
+        }
+        if (program.background.scripts.length == 0) {
+          program.background.scripts.push(
+            asset.addURLDependency('./runtime/default-bg.js', {
```
devongovett, line 275: I assume this empty script is needed for some reason?

packages/runtimes/webextension/src/WebExtensionRuntime.js
```diff
@@ -0,0 +1,50 @@
...
+    );
+    let firstInsertableBundle;
+    bundleGraph.traverseBundles((b, _, actions) => {
+      if (b.type == 'js') {
+        firstInsertableBundle = b;
+        actions.stop();
+      }
+    }, insertBundle);
+
+    // Add autoreload
```
devongovett, line 41: I guess this moved from the transformer into a runtime? I'm curious about why that was. Would you mind explaining a bit?

devongovett (commented):
A few code comments would have been nice, but looks great overall! A few questions. Thanks again!

packages/core/integration-tests/test/webextension.js
101arrowz: I wasn't sure how to deal with content hashing, is there a way to handle that?

packages/packagers/webextension/src/WebExtensionPackager.js
101arrowz: Yes AFAIK any resources accessed from the extension, direct or not, must be specified in `web_accessible_resources`. How do we recursively get children? `traverseBundles`?

packages/transformers/webextension/src/WebExtensionTransformer.js
101arrowz: Now that you mention it that does seem weird but I think it fixed an issue where files like `.DS_Store` were accessed? I'll check again.

packages/transformers/webextension/src/WebExtensionTransformer.js
101arrowz: Actually this is specifically to handle the case that this JS file is updated to reference a CSS bundle, because without this line, that doesn't trigger the `WebExtensionPackager` again so the CSS doesn't get added to the manifest.

packages/transformers/webextension/src/WebExtensionTransformer.js
101arrowz: This is before `collectDependencies`, but I can move it to the top for clarity maybe. I know this is a bad practice but for web extensions it is literally impossible to use it in any environment but a browser, and this makes it way easier for people to use the web extension config without modification. The `inline` and `inlineSources` source mapping options especially are critical for source mapping to work by default.

packages/transformers/webextension/src/WebExtensionTransformer.js
101arrowz: Yes, we need it to inject the auto-reloader for content-script page refresh. That's `autoreload-bg.js` in the runtime.

packages/runtimes/webextension/src/WebExtensionRuntime.js
101arrowz: Sure. We needed to move to a runtime because there was no way to inject the autoreload into specifically the background pages or the service worker without adding another specialized transformer for those files, or hijacking the JS transformer. Of course, the original strategy of using the transformer for this worked fine for `background.scripts` users, but that is MV2.

packages/transformers/webextension/src/WebExtensionTransformer.js
101arrowz: Oh right I was going to check if `messages.json` existed but I accidentally forgot to write out the full path. Fixing now.

packages/transformers/webextension/src/WebExtensionTransformer.js
101arrowz: Actually this wouldn't work for nested dependencies...any solutions?

packages/core/integration-tests/test/webextension.js
devongovett: Any kind of assertions are good. Like asserting the correct number of referenced files, some regex on the name, etc.

packages/packagers/webextension/src/WebExtensionPackager.js
devongovett: Yep, that should work.

packages/transformers/webextension/src/WebExtensionTransformer.js
devongovett: Ah hmm. That's more of a problem with caching in packagers. At the moment if you access other bundles from within a packager it won't be invalidated. Let me think about that.

packages/core/integration-tests/test/webextension.js
101arrowz: I added a few more tests that should hopefully cover most common use cases. Full testing of e.g. the HMR functionality would probably be a bit too cumbersome.

packages/transformers/webextension/src/WebExtensionTransformer.js
101arrowz: How is CSS importing support in JS when the JS is referenced from HTML then? We should be able to replicate the strategy here.

devongovett: The HTML packager only needs to list directly referenced bundles, and their siblings. `getReferencedBundles` handles this. It's possible that it works because the JS and CSS get placed in the same bundle group (i.e. loaded together) as the HTML file. That's due to `priority: 'parallel'` on the dependency. You could potentially also try that here.

However, that still wouldn't work for nested dependencies, i.e. async `import()` in the JS. I guess those would also need to be listed in `web_accessible_resources`?

101arrowz: Yes, I'm looking into solutions for that right now. I found that `priority: 'parallel'` works well for the content scripts, but as you expected, CSS imports from lazy-loaded children in content scripts are not propagated to the manifest. We do need those bundles in `web_accessible_resources` as well.

There's also another issue: since CSS is parsed outside the extension environment for content scripts, [relative paths don't work](https://stackoverflow.com/a/8864212/10930232). <strike>We need to prepend `chrome-extension://__MSG_@@extension_id__/` to any bundle referenced from a content-script CSS bundle. Do we have a way to do that?</strike> Actually for cross-browser compatibility, the only way to make this work is to inline all the bundles in the CSS. Not sure how to make that work but it could be a known issue until we find a way to fix it.

devongovett: Yeah, if you're comfortable, we could probably just merge this as is for now, and address the caching issue separately. Maybe having something is better than being perfect? Should be a somewhat rare edge case when adding/removing a bundle.

101arrowz: Yeah that's probably for the best, that edge case seems a bit difficult to tackle right now, and the MV3 support from this PR is, as far as I can tell, already way cleaner than what's available in any other bundler.

101arrowz:
I've updated the docs again and fixed every issue beyond the edge case for CSS imports within an `import()`ed module - that can be left a known issue for now (it's even specified in the updated docs). Hopefully this is good enough to release in the next version.

devongovett (approved):
Thanks again!

marcomow:
great job @101arrowz ! Thank you so much! 

Is there a way we can "Buy you a coffe/beer"?

And of course the same goes for the reviewer @devongovett  :)

sergey-shpak:
@101arrowz thank you very much! :rocket: 
@devongovett, @mischnic we are waiting for the release of this feature)

Q0:
@101arrowz thank you very much for your work! 🚀
And thanks to everyone who participated in this PR.

We look forward to the release and start updating extensions 🎉

devongovett:
In a few hours, you'll be able to test it out in tonight's nightly release. Please do, and report any issues you find. It'll be officially released in the next minor, which should be in a few days if all goes well.

fregante:
<p align="center">The time has come</p>

a8cef12e365f6474e44e8af606a14eb1

101arrowz:
> In a few hours, you'll be able to test it out in tonight's nightly release

Awesome, thanks you so much for reviewing!

> Is there a way we can "Buy you a coffee/beer"?

Feel free to donate to the [Open Collective](https://opencollective.com/parcel) for Parcel, the maintainers deserve the support more than I do :)

101arrowz:
Also for anyone testing please refer to the docs at https://github.com/parcel-bundler/website/pull/1014 (live at [this link](https://website-f530twpc7-devongovett.vercel.app/recipes/web-extension/)). @devongovett The docs should probably be implemented before the next minor release.

dougwithseismic:
> Also for anyone testing please refer to the docs at [parcel-bundler/website#1014](https://github.com/parcel-bundler/website/pull/1014) (live at [this link](https://website-f530twpc7-devongovett.vercel.app/recipes/web-extension/)). @devongovett The docs should probably be implemented before the next minor release.

Any tips for a novice on how to grab the nightly?

101arrowz:
`npm install parcel@nightly` or `yarn add parcel@nightly` should work.

devongovett:
(Nightly releases are at midnight UTC, so you'll need to wait until then.)

mischnic:
(And also make sure to install the nightly versions of both `parcel` and `@parcel/config-webextension`)

Stvad:
Hmm, seems nightly release action is failing: https://github.com/parcel-bundler/parcel/runs/5993495011?check_suite_focus=true last successfuly published nightly was 4 days ago 🤔

fregante:
It seems that the manifest validator is too strict and doesn't say why. It immediately fails when the `browser_specific_settings` key is found:
```
@parcel/transformer-webextension: Invalid Web
Extension manifest
```

Also it automatically adds a broken/unnecessary `web_accessible_resources` key:

<details>
<summary>Input manifest</summary>

```json
{
	"name": "Awesome Extension",
	"version": "0.0.0",
	"description": "An awesome new browser extension",
	"homepage_url": "https://github.com/fregante/browser-extension-template",
	"manifest_version": 3,
	"minimum_chrome_version": "100",
	"icons": {
		"128": "icon.png"
	},
	"permissions": [
		"storage"
	],
	"host_permissions": [
		"https://github.com/*"
	],
	"content_scripts": [
		{
			"matches": [ "https://github.com/fregante/browser-extension-template/*" ],
			"js": [ "content.js" ],
			"css": [ "content.css" ],
			"run_at": "document_end"
		}
	],
	"options_ui": {
		"browser_style": true,
		"page": "options.html"
	},
	"background": {
		"service_worker": "background.js",
		"type": "module"
	}
}
```

</details>

<details>
<summary>Output manifest</summary>

```json
{
	"name": "Awesome Extension",
	"version": "0.0.0",
	"description": "An awesome new browser extension",
	"homepage_url": "https://github.com/fregante/browser-extension-template",
	"manifest_version": 3,
	"minimum_chrome_version": "100",
	"icons": { "128": "icon.da91f5eb.png" },
	"permissions": ["storage"],
	"host_permissions": [
		"https://github.com/*"
	],
	"content_scripts": [
		{
			"matches": ["https://github.com/fregante/browser-extension-template/*"],
			"js": ["manifest.e36317c5.js"],
			"css": ["manifest.d108b4c6.css"],
			"run_at": "document_end"
		}
	],
	"options_ui": { "browser_style": true, "page": "options.29b7abd3.html" },
	"background": {
		"service_worker": "background.2784684d.js",
		"type": "module"
	},
	"web_accessible_resources": [
		{
			"matches": ["https://github.com/fregante/browser-extension-template/*"],
			"extension_ids": [],
			"resources": []
		}
	]
}
```

</details>

And chrome throws with:

```
Failed to load extension from: ./distribution. Invalid value for 'web_accessible_resources[0]'. Invalid match pattern.
```


Live test at https://github.com/fregante/browser-extension-template/pull/75

mischnic:
Please open new issues for bugs.

I just tried a small example and it does say why it failed for browser_specific_settings:
```
$ yarn parcel build manifest.json
yarn run v1.22.18
warning package.json: No license field
$ node_modules/.bin/parcel build manifest.json
🚨 Build failed.

@parcel/transformer-webextension: Invalid Web Extension manifest

  manifest.json:9:3
     8 |   },
  >  9 |   "browser_specific_settings": {
  >    |   ^^^^^^^^^^^^^^^^^^^^^^^^^^^ Possible values: "default\_locale", "icons", "author", "chrome\_settings\_overrides", "chrome\_url\_overrides", "commands", "content\_scripts", "devtools\_page", "dictionaries", "externally\_connectable", "file\_browser\_handlers", "file\_system\_provider\_capabilities", "homepage\_url", "incognito", "minimum\_chrome\_version", "offline\_enabled", "omnibox", "optional\_permissions", "options\_ui", "permissions", "protocol\_handlers", "requirements", "short\_name", "sidebar\_action", "storage", "theme", "tts\_engine", "user\_scripts", "version\_name", "action", "content\_security\_policy", "host\_permissions", "web\_accessible\_resources"
    10 |     "gecko": {
    11 |       "id": "addon@example.com",

```

fregante:
> I just tried a small example and it does say why it failed for browser_specific_settings:

It didn't for me:

Screen Shot 12

Until I did a clean install of everything:


163381100-0089e648-e9d6-4fa4-b255-66754ac5ea99

ricksmit3000:
I've installed the nighly's but when I build I get the following error:
```
yarn run v1.22.17
$ parcel build manifest.json --target webext-prod

🚨 Build failed.

@parcel/transformer-js: Service workers cannot have imports or exports without the `type: "module"` option.

  /Users/rick/dev/react-parcel-web-extension/src/scripts/background.ts:1:1
  > 1 | import browser, { Tabs } from "webextension-polyfill";
  >   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

  💡 Add {type: 'module'} as a second argument to the navigator.serviceWorker.register() call.
  📝 Learn more: https://parceljs.org/languages/javascript/#classic-scripts

```
This used to work and I couldn't find anything in the new docs that I needed to change

devongovett:
Did you try the suggestion in the error message?

irgendwr:
@rickiesmooth See the [comment](https://github.com/parcel-bundler/parcel/pull/7050#issuecomment-1071433977) from @101arrowz:

> you need to specify that it's a module type service worker. This behavior is [consistent with browsers](https://stackoverflow.com/a/66408379/10930232).
> 
> ```json
> "background": {
>     "service_worker": "background.ts",
>     "type": "module"
> }
> ```

ricksmit3000:
@devongovett in my app code I couldn't find the `navigator.serviceWorker.register` call, but as @irgendwr mentioned I should just add it to the manifest! sorry about that

sergey-shpak:
running production on top of parcel nightly release - thug dev life :shrug:

devongovett:
This was released in v2.5.0 earlier today fyi

sergey-shpak:
@101arrowz  @devongovett found few issues: https://github.com/parcel-bundler/parcel/issues/7997

PR #7061, opened 2021-10-13
Dynamic imports priority fix closes #6980

### LINKED ISSUE #6980: Dynamic imports loaded as low-priority scripts when compiled
Native dynamic imports (e.g. `import('./foo')`) are typically loaded by browsers with high priority. However, when targeting browsers without native or polyfillable support, Parcel inserts `<script async>` tags into the head, which are loaded with low priority. See https://addyosmani.com/blog/script-priorities/.

With that said, we should probably still include the `async` attribuite on these scripts as their execution order isn't important. One workaround to boost priority of async scripts is to also insert a `<link rel="preload" as="script">` with an `href` with the same value as the script's `src`.

We should consider adding these link tags as well when loading js targeted at these environments.

Related: #6981

highvoltag3:
# ↪️ Pull Request

Fixes #6980 by leveraging the `<link rel=preload>+<script async>` "hack" which makes both {loading,scheduling} priority high in engines that do not support native esmodules when Parcel inserts `<script async>` tags into the head which are loaded with low priority.

## 💻 Examples

See: https://addyosmani.com/blog/script-priorities/

## ✔️ PR Todo


packages/core/integration-tests/test/javascript.js
```diff
@@ -868,6 +868,29 @@ describe('javascript', function() {
...
+
+    let output = await run(b);
+    let headChildren = await output.default;
+
+    assert(headChildren[1].tag === 'link');
+    assert(headChildren[1].rel === 'preload');
+    assert(headChildren[1].as === 'script');
+
+    assert(headChildren[2].tag === 'script');
+    assert(headChildren[2].src.match(/async\..*\.js/));
```
wbinnssmith: Would be helpful to assert that the src here matches the href value in the link tag above

wbinnssmith:
Looks like some existing tests are now failing, likely because they assert the number of things in `<head>`. Should be straightforward to adjust the test to those new values assuming everything else remains correct.

packages/core/integration-tests/test/javascript.js
highvoltag3: Just saw this, sure let me add that really quick

highvoltag3:
> Looks like some existing tests are now failing, likely because they assert the number of things in `<head>`. Should be straightforward to adjust the test to those new values assuming everything else remains correct.

Odd... seems to be passing for me even after merging the latest V2 branch.

wbinnssmith (approved):

PR #7056, opened 2021-10-12
fix(vuetransfomer): fix unknown asset language for vue scripts/templa…

RuyiLi:
…tes/styles

# ↪️ Pull Request

Address an issue where attempting to load a script/template/style from another file would fail due to VueTransformer not being able to detect the language from the file name.

## 💻 Examples

#6546

## ✔️ PR Todo

- [ ] Added/updated unit tests for this change
- [ ] Filled out test instructions (In case there aren't any unit tests)
- [ ] Included links to related issues/PRs

PR #7069, opened 2021-10-14
Fix `src` tag for Vue SFCs

101arrowz:
# ↪️ Pull Request

Fixes the `src` tag in Vue SFCs, which previously didn't work because the `type` was detected as `.js` instead of `js`.

## 💻 Examples

`App.vue`:
```js
<script src="./other.tsx" />
```

`other.tsx`:
```js
import { defineComponent, h } from 'vue'

export default defineComponent({
  render() {
    return <h1>Hi there!</h1>
  }
})
```

Just noticed this while testing out the v2.0.0 stable release (Congrats! 🎉) I'd prefer to make a PR for tests later, though a regression test could be nice.

### DIFF
--- a/packages/transformers/vue/src/VueTransformer.js
+++ b/packages/transformers/vue/src/VueTransformer.js
@@ -207,7 +207,7 @@ async function processPipeline({
             await resolve(asset.filePath, template.src),
           )
         ).toString();
-        template.lang = extname(template.src);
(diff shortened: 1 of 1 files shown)


DeMoorJasper:
Could you add a test for this?

101arrowz:
I think this can be moved back to #7056, which already did the same thing.

PR #7172, opened 2021-10-23
chore(README): cite swc for core compiler

sno2:
Edit: Just saw this again 4 years later. I was in the wrong here. Sorry about this.

The README is misleading and I've corrected a couple of people on discord that had thought that Parcel had wrote the whole Rust code for the JavaScript compiler themselves.  This fix should clear things up for people.


devongovett:
This is intentional. Yes, we use SWC's parser and code generator, and optionally, for transpilation, but what this is talking about is the core scope hoisting compiler, dependency collector, and other  bundler specific transforms implemented by Parcel. Listing SWC here makes it sound like SWC did all the work and Parcel is just using it (people already assume this), but that's not true. It's actually more custom than not, and a lot of work went into it beyond "leveraging SWC". It would be like saying webpack "leveraged acorn" which is technically true but kinda not the point.

sno2:
> This is intentional. Yes, we use SWC's parser and code generator, and optionally, for transpilation, but what this is talking about is the core scope hoisting compiler, dependency collector, and other bundler specific transforms implemented by Parcel.

I disagree.  You specifically state "Parcel's JavaScript compiler is written in Rust"  within this section.  The "JavaScript compiler" utilizes swc for parsing, ast manipulation, and transpilation.  The language-agnostic plugin system was built by Parcel.

> Listing SWC here makes it sound like SWC did all the work and Parcel is just using it (people already assume this), but that's not true. It's actually more custom than not, and a lot of work went into it beyond "leveraging SWC".

If you don't like the wording, then how would "Parcel is built upon swc, a JavaScript compiler built in Rust, for native performance" sound?  Therefore, everyone did some work.

> SWC did all the work and Parcel is just using it (people already assume this)

I don't know what you're talking about as there are many people who are surprised to learn about swc when I tell them about it, yet they have used and went through Parcel 2's info.

> It would be like saying webpack "leveraged acorn" which is technically true but kinda not the point.

I disagree and I think that swc should get credit for a _huge_ factor of why Parcel is now faster.  Furthermore, if you check out [webpack's readme](https://www.npmjs.com/package/webpack), then you will notice that they include a link to Babel and how it handles the loading of JS code and transpilation just underneath the introduction.  Parcel, however, does not even mention swc on their README.

My final point is that swc is way smaller in comparison to babel and is completely done by volunteers.  They rely on donations to survive and I believe that it is a social responsibility to include a link to something that your 20x performance improvement in Parcel 2 is completely dependent on.  If frameworks do not link to SWC when they use it, then it reduces exposure that it deserves to flourish as an open-source project and community.

devongovett:
Have you actually looked at the code? There's a lot more custom Rust than you are giving credit for. We previously used Babel's parser and didn't cite that on the readme either. I don't see how this is different. On the website we do mention SWC and describe exactly what parcel implements as well.

PR #7173, opened 2021-10-24
perf(rust): remove useless macro calls

sno2:
Every single one of the macro calls slows down compilation as it needs to run rust functions and pre-process your code.  This should help bring those compilation speeds down.


devongovett:
Do you have any compilation time numbers that show a significant improvement? Somehow I doubt that this has much of an effect.

PR #7184, opened 2021-10-25
Update package.json to include the repository

v-rr:
Hi there!
This change adds the repository property to your package.json file(s). Having this available provides a number of benefits to security tooling. For example, it allows for greater trust by checking for signed commits, contributors to a release and validating history with the project. It also allows for comparison between the source code and the published artifact in order to detect attacks on authors during the publication process.
We validate that we're making a PR against the correct repository by comparing the metadata for the published artifact on [npmjs.com](www.npmjs.com) against the metadata in the package.json file in the repository.
This change is provided by a team at Microsoft -- we're happy to answer any questions you may have. (Members of this team include [@s-tuli](https://github.com/s-tuli), [@iarna](https://github.com/iarna), [@rancyr](https://github.com/v-rr), [@Jaydon Peng](https://github.com/v-jiepeng), [@Zhongpeng Zhou](https://github.com/v-zhzhou) and [@Jingying Gu](https://github.com/v-gjy)). If you would prefer that we not make these sorts of PRs to projects you maintain, please just say. If you'd like to learn more about what we're doing here, we've prepared a document talking about both this project and some of our other activities around supply chain security here: [microsoft/Secure-Supply-Chain](https://github.com/microsoft/Secure-Supply-Chain)
This PR provides repository metadata for the following packages:
* @parcel/babel-preset
* @parcel/babylon-walk
* @parcel/create-react-app
* @parcel/fs-write-stream-atomic

### DIFF
--- a/packages/dev/babel-preset/package.json
+++ b/packages/dev/babel-preset/package.json
@@ -3,6 +3,11 @@
   "version": "2.0.0",
   "private": true,
   "license": "MIT",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/parcel-bundler/parcel.git",
+    "directory": "packages/dev/babel-preset"
+  },
   "dependencies": {
     "@babel/plugin-proposal-class-properties": "^7.1.0",
     "@babel/plugin-proposal-nullish-coalescing-operator": "^7.4.4",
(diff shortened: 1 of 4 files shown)


devongovett:
Seems fine to me. Why did you only add it to these packages? Should we add it to all of them?

iarna:
I believe that the other packages already have repository keys, it was just these that we found missing. Did we miss some ourselves?

devongovett:
Ah, I just assumed they weren't there. Looks like they are. Would be nice to add the "directory" key though. Can do that separately. Thanks!

PR #7198, opened 2021-10-27
Add large blob storage to Cache

lettertwo:
## New methods on `Cache`
- `hasLargeBlob(key: string): Promise<boolean>`
- `getLargeBlob(key: string): Promise<Buffer>`
- `setLargeBlob(key: string, contents: Buffer | string): Promise<void>`

In `FSCache`, these are basically aliases for `has()`, `getBlob()` and `setBlob()`, but they use a modified key to differentiate large blobs from other cached values.

In `LMDBCache`, they interact with the configured filesystem rather than the LMDB store.

## Caching RequestGraph as a large blob

In testing #6922 with a very large Parcel project, rebuilds would hard crash while attempting to deserialize the cached `RequestGraph`:
```
Assertion failed: (length <= imp::kMaxLength && "too large buffer"), function NewBuffer, file ../node_modules/nan/nan.h, line 890.
```

It seems this is because lmdb-store decodes binary data from the database into a Node Buffer which has a hard-coded [upper limit on the size of the buffer](https://github.com/nodejs/nan/blob/8502c2417b9c1d0ea86c1e0aea6e975d26bf682b/nan.h#L892) for [various reasons](https://bugs.chromium.org/p/v8/issues/detail?id=3505).

To work around this limit, the `RequestGraph` is now cached using these new large blob methods.

## Caching other graphs as large blobs

Other graphs (`AssetGraph`, `BundleGraph`) are also cached as large blobs by updating the `RequestTracker` `getRequestResult()` and `writeToCache()` methods to use large blob caching, under the assumption that only graph requests are made with cache keys (which seems to be the mechanism by which `RequestTracker` decides to cache a result).

## Streams

Large assets are already threaded through Parcel as streams, and now the `LMDBCache` also streams those values to and from the underlying FS. (Previously, streams were buffered and then written to the LMDB store).

## Fallback to FS caching for large values

~~In addition to the new methods that allow explicitly caching large blobs, LMDBCache has also been updated to automatically fallback to FS cache when a value will exceed the max Node buffer size.~~ Auto fallback has been removed in favor of marking streams as large blobs and using the cache `{get, set}Stream` methods, with the blob methods becoming the default (previously, the stream methods were always used, even if the asset was not streamed).

## Questions
- [x] What other large entities should be treated this way? AssetGraph and BundleGraph?
> All the graphs are now treated as large blobs
- [x] What about large assets like video files?
> large assets are now streamed in and out of the cache


parcel-benchmark:
## Benchmark Results

#### Timings

| --- | --- | --- |
| Cold | 1.79s | -8.00ms |
| Cached | 283.00ms | +19.00ms ⚠️ |

#### Cold Bundles

| dist/legacy/parcel.7cdb0fad.webp | 102.94kb | +0.00b | 552.00ms | +30.00ms ⚠️ |
| dist/modern/parcel.7cdb0fad.webp | 102.94kb | +0.00b | 553.00ms | +32.00ms ⚠️ |

#### Cached Bundles

| dist/legacy/parcel.7cdb0fad.webp | 102.94kb | +0.00b | 521.00ms | -34.00ms 🚀 |

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 9.08s | -75.00ms |
| Cached | 420.00ms | +21.00ms |

#### Cold Bundles

*No bundle changes detected.*

#### Cached Bundles

*No bundle changes detected.*

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 1.00m | -208.00ms |
| Cached | 1.41s | -7.00ms |

#### Cold Bundles

*No bundle changes detected.*

#### Cached Bundles

*No bundle changes detected.*

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 6.36s | +56.00ms |
| Cached | 368.00ms | +52.00ms ⚠️ |

#### Cold Bundles

*No bundle changes detected.*

#### Cached Bundles

*No bundle changes detected.*

</p></details>

**[Click here to view a detailed benchmark overview.](https://benchmark.parceljs.org/benchmark/315986605614563906)**

packages/core/cache/src/LMDBCache.js
```diff
@@ -35,8 +39,8 @@ export class LMDBCache implements Cache {
     };
```
mischnic: You have to add `fs` here as well

lettertwo: 🙈 Nice catch, thanks!

thebriando:
Is there any perf difference between caching the graph to LMDB vs writing it to FS?

lettertwo:
> Is there any perf difference between caching the graph to LMDB vs writing it to FS?

Nothing noticeable at the macro level (build times are comparable), but maybe I can capture some startup/shutdown metrics.

lettertwo:
Ran some profiling builds to see how this change might effect performance:

|             | Cold Build | Warm Build | Cold Start | Warm Start |
| ----------- | ---------- | ---------- | ---------- | ---------- |
| serialize   | -165 ms    | +438 ms    | +461 ms    | -261 ms    |
| deserialize | +242 ms    | +382 ms    | +118 ms    | -98 ms     |
| init        | +1 ms      | -71 ms     | +1 ms      | -97 ms     |
| build       | -924 ms    | -2 ms      | +1298 ms   | +78 ms     |
| save        | -          | -          | -6 ms      | +11 ms     |
| shutdown    | +115 ms    | -162 ms    | -24 ms     | +365 ms    |
| ✨ Built in | -0.75 s    | -0.30 s    | +0.95 s    | +0.07 s    |

The above shows the difference between the same build with and without caching the request graph as a large blob. They seem to be pretty comparable overall, with the differences between them looking similar to the differences between any two builds regardless of caching mechanism.

For context: This build typically runs in about ~135s in prod and ~70s in dev.

thebriando (approved):

mischnic:
`LMDBCache` will always write to the "real" FS(= `NodeFS`) anyway, so do we actually need to make the FS configurable? When would you want to use LMDB on the real FS but write the large blobs to a MemoryFS? 

Also currently the change to the `LMDBCache` constructor would be a breaking change

mischnic:
> What about large assets like video files?

For that, a stream-based API is better anyway to reduce memory usage (this is why transformers, packagers and optimizers have a stream-based API in the first place). That would at the least require `(s)etLargeStream` methods in the cache. Not sure how straight forward it is to detected in core which method to use for large enough assets and bundles.

lettertwo:
> `LMDBCache` will always write to the "real" FS(= `NodeFS`) anyway, so do we actually need to make the FS configurable? When would you want to use LMDB on the real FS but write the large blobs to a MemoryFS?
> 
> Also currently the change to the `LMDBCache` constructor would be a breaking change

I see what you mean about not needing to configure the FS for `LMDBCache` (it is only used when the FS is `NodeFS` anyway). I updated the PR to not change the `LMDBCache` constructor API.

lettertwo:
> > What about large assets like video files?
> 
> For that, a stream-based API is better anyway to reduce memory usage (this is why transformers, packagers and optimizers have a stream-based API in the first place). That would at the least require `(s)etLargeStream` methods in the cache. Not sure how straight forward it is to detected in core which method to use for large enough assets and bundles.

👍 For the purposes of this PR (solving the Node buffer size limit), would it make sense to just assert that any buffer being written to LMDBCache is under that size?

lettertwo:
> 👍 For the purposes of this PR (solving the Node buffer size limit), would it make sense to just assert that any buffer being written to LMDBCache is under that size?

I've opted instead to auto fallback to using FS cache when any blob being written to `LMDBCache` might exceed the Node buffer size.

This is kind of a punt on solving for large media, but currently, large media is already buffered completely into memory to be cached by `LMDBCache`, so this will at least let things continue to work if something very large is streamed into cache, though it doesn't deliver the memory savings that a proper stream API would allow.

packages/core/cache/src/LMDBCache.js
```diff
@@ -40,45 +44,77 @@ export class LMDBCache implements Cache {
...
+      if (this.store.get(key) != null) await this.store.remove(key);
+      await this.setLargeBlob(key, contents);
+    } else {
+      await this.store.put(key, contents);
+    }
+  }
+
+  async getBuffer(key: string): Promise<?Buffer> {
+    let buffer = this.store.get(key);
+    if (buffer == null && (await this.hasLargeBlob(key))) {
```
devongovett: I'm worried that this will result in a lot of extra `stat` calls to the filesystem since it will be checked for every key that is missing from the cache (everything on initial builds). Can we keep checking for large blobs explicitly opt-in? I thought we only needed it for certain known large entries (e.g. graphs)?

lettertwo:
> I've opted instead to auto fallback to using FS cache when any blob being written to LMDBCache might exceed the Node buffer size.

I reverted this fallback behavior. Now, LMDBCache uses FS read/write streams instead of converting between stream and buffer, and I added some `isLargeBlob` signals for cases where streams are being written to cache, which is then used to read them from cache as streams.

So, large assets should now be streamed to/from cache rather than being buffered in memory.

packages/core/cache/src/LMDBCache.js
lettertwo: I reverted this in favor marking large assets for streaming to and from cache ([comment](https://github.com/parcel-bundler/parcel/pull/7198#issuecomment-972389465))

packages/core/cache/src/LMDBCache.js
```diff
@@ -53,16 +57,28 @@ export class LMDBCache implements Cache {
...
+    await this.setBlob(key, serialize(value));
   }
 
   getStream(key: string): Readable {
-    return blobToStream(this.store.get(key));
+    let filename = path.join(this.dir, key);
+    if (this.fs.existsSync(filename)) {
+      return this.fs.createReadStream(filename);
+    } else {
+      // If the file doesn't exists, return an empty stream.
```
devongovett: Hmm, should it error?

devongovett (approved):
Looks good. Going to perf test it later

packages/core/cache/src/LMDBCache.js
lettertwo: Yeah, I suppose it should, as that seems to be what FSCache does.

packages/core/cache/src/LMDBCache.js
```diff
@@ -73,12 +81,41 @@ export class LMDBCache implements Cache {
   }
 
   async setBlob(key: string, contents: Buffer | string): Promise<void> {
+    invariant(
+      !isLargeBlob(contents),
+      'Cannot store large blobs in the cache. You may want to use `setLargeBlob` instead.',
+    );
```
lettertwo: Maybe we won't benefit from this check any more, since our most likely sources of large blobs are either being stored explicitly (graphs), or are being streamed (assets).

devongovett (approved):

devongovett:
Tested and saw no perf regression anymore.

PR #7236, opened 2021-11-01
Reject browser js loader promise with Error object

### LINKED ISSUE #6052: JSRuntime: browser js loader rejects with an error Event, not an Error
# 🐛 bug report

If loading js were to fail in a browser context with the global output format, the promise would be rejected with an error _Event_, not an Error object. This is not intuitive, and is inconsistent with real `import()`s in browsers and across how we implement this in other environments.

## 🤔 Expected Behavior

Should reject with an Error object instead. Real `import()`s can reject with a TypeError such as `TypeError: Failed to fetch dynamically imported module: http://localhost:5000/foo.js`.

## 😯 Current Behavior

If fetching the module fails, the promise returned by `import()` rejects with a DOM Error event.

https://github.com/parcel-bundler/parcel/blob/dea272f0708fdabfe11f844f189df6421e7a3b65/packages/runtimes/js/src/loaders/browser/js-loader.js#L20

## 💁 Possible Solution

Create our own error in the browser js loader that is shaped like what real `import()`s in browsers reject with in this case.

mrkldshv:
# ↪️ Pull Request

If loading of JS in the browser fails, reject promise of JS loader with Error object with message instead of DOM Error event.
Fixes #6052.

## 🚨 Test instructions

Fake JS loader fail to get rejected promise and checked printed out error.

## ✔️ PR Todo


packages/runtimes/js/src/helpers/browser/js-loader.js
```diff
@@ -18,10 +18,13 @@ module.exports = cacheLoader(function loadJSBundle(bundle) {
     script.type = 'text/javascript';
     script.charset = 'utf-8';
     script.src = bundle;
-    script.onerror = function (e) {
+    script.onerror = function(e) {
+      var error = new TypeError(
```
devongovett, line 22: This should probably just be `new Error`. I don't think TypeError makes sense here

packages/runtimes/js/src/helpers/browser/js-loader.js
```diff
@@ -18,10 +18,13 @@ module.exports = cacheLoader(function loadJSBundle(bundle) {
...
+        `Failed to fetch dynamically imported module: ${bundle}.`,
+      );
```
devongovett, line 24: Should we somehow attach the original error so people can determine why it failed?

devongovett (changes requested):

packages/runtimes/js/src/helpers/browser/js-loader.js
mischnic: Chrome and Firefox do use `TypeError`:
```
> import("http://google.com")
Uncaught (in promise) TypeError: Failed to fetch dynamically imported module: http://google.com/
```

I don't think `TypeError` is specced somewhere though because the behaviour is implementation defined

https://tc39.es/ecma262/#sec-import-call-runtime-semantics-evaluation
https://tc39.es/ecma262/#sec-hostimportmoduledynamically

https://github.com/GoogleChromeLabs/dynamic-import-polyfill/blob/5ae321654a96c5e99b23b7cdabe10bc6159caaaa/initialize.mjs#L59

mrkldshv: Do you think we can keep `TypeError` to be inline with browsers behaviour though?

packages/runtimes/js/src/helpers/browser/js-loader.js
mrkldshv: We can do something like: 
```js
var error = new TypeError(
    `Failed to fetch dynamically imported module: ${bundle}. Error: ${e.message}`,
);
```
What do you think?

mrkldshv: Should be fixed now.

mischnic:
Ci is failing because you need to reformat the file with Prettier.

mrkldshv:
@mischnic Formatting should be fixed now

devongovett (approved):

mrkldshv:
Thank you for review!

PR #7248, opened 2021-11-04
fix: add @parcel/diagnostic as dependency of @parcel/transformer-typescript-types

susickypavel:
# ↪️ Pull Request

Fixes missing `@parcel/diagnostic` dependency in `@parcel/transformer-typescript-types`. Otherwise, it produces unexpected behavior with modern package managers (such as yarn v2+).  

Specifically, in yarn berry, it can be solved using `packageExtensions` like so:

```yml
packageExtensions:
  "@parcel/transformer-typescript-types@*":
    dependencies:
      "@parcel/diagnostic": "^2.0.0"
```

However, it's better when packages themselves list their dependencies for more fluent DX and stability.


mischnic (approved):
Thanks

PR #7251, opened 2021-11-04
fix: getRootDir behaviour for single directory causing wrong project root

### LINKED ISSUE #7145: Wrong Project Root Directory in watch mode
# 🐛 bug report
Configuring targets in package.json and running `parcel watch` leads to wrong project root directory - as result the file watcher pointed to the system root (or any parent root with package.json) and watching all the possible files (which makes initial and future timeouts and limits to system file watchers).

```json
"targets": {
  "default": {
    "source": "./src/app/index.js"
   }
},
"scripts": {
  "watch": "parcel watch"
}
```

## 🤔 Expected Behavior
Correct project root dir, and fast `parcel watch`

## 😯 Current Behavior
Huge(8-10 seconds) timeouts when starting `parcel watch` and wrong project root dir

## 💁 Possible Solution
I had to debug parcel, and the problem is quite simple - when running `parcel watch` it uses current working directory as an entry point, and since current directory path doesn't contain trailing slash it is ignored when trying to find project root directory (it walks only through the parents looking for package-lock.json), then wrong project root directory is used to listen to system file changes.

*tried to run `parcel watch $PWD/` but still no luck, because entry point is resolved(by path) and trailing slash removed, which leads to the same bug

## 🌍 Your Environment
Node@16.9, Ubuntu 20.4, Parcel@2.0.0-rc.0

akaltar:
Linked issue: #7145 

# ↪️ Pull Request

The getProjectRoot command was returning a wrong root when passed in a folder value. This caused the wrong project root to be determined when running in watch mode. In some cases this caused the entire filesystem to be watched.

The reason this is a draft is that this change breaks about ~100 integration tests, I'm happy to look into fixing those as well, but I'd like to first get feedback and confirm that my change aligns with the intended behaviour of the `getProjectRoot` function.

I tested this on a repo that was not working due to this bug, and now the .proxyrc.js file was loaded correctly and the site could be loaded. In fact some extra config files (.babelrc) also got loaded thanks to this.

Even if this is the intended change it's probably a huge breaking change, so please advise on how should I proceed.

Thank you for the amazing work on this project, adding a new test and making this fix was a breeze.

## 💻 Examples

Loading .proxyrc.js from project didn't work and explicitly setting the path for it wasn't possible.

## ✔️ PR Todo

### DIFF
--- a/packages/core/utils/src/getRootDir.js
+++ b/packages/core/utils/src/getRootDir.js
@@ -4,6 +4,8 @@ import type {FilePath} from '@parcel/types';
 import {isGlob} from './glob';
 import path from 'path';
 
+// Returns the common root of the given paths.
+// If there is no common root, returns the current working directory.
 export default function getRootDir(files: Array<FilePath>): FilePath {
   let cur = null;
 
@@ -27,10 +29,16 @@ export default function getRootDir(files: Array<FilePath>): FilePath {
       }
 
       cur.dir = i > 1 ? curParts.slice(0, i).join(path.sep) : cur.root;
+      cur.name = '';
+      cur.base = '';
     }
   }
 
-  return cur ? cur.dir : process.cwd();
+  if (!cur) {
+    return process.cwd();
+  }
+
+  return path.join(cur.dir, cur.name);
 }
 
 // Transforms a path like `packages/*/src/index.js` to the root of the glob, `packages/`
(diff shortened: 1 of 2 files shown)


mischnic:
Superseded by https://github.com/parcel-bundler/parcel/pull/7537

Issue #7063, opened 2021-10-13
pnpm workspace not work


ckken:
![image](https://user-images.githubusercontent.com/3890513/137075426-5274a101-40f8-4d64-b017-74b5f122b57b.png)

ckken:
Mark

ch99q:
I've experienced the same thing, but I have been investigating a bit into this.

I found two things resulting in this.
1. It won't identify the folder you have your project in cause it won't have a lock file.
https://github.com/parcel-bundler/parcel/blob/d2a9b6bdd93855b468fa9911dad27a49e54e9557/packages/core/core/src/resolveOptions.js#L52-L59
2. If we add package.json or "node_modules" we return to issue 2; it won't necessarily find the root folder cause of the entries. So to make it work now, we have to do `parcel build src/index.ts`; this prevents us from using `"source": "src/index.ts"` in the `package.json`.

I've gone so deep into the code; I can't figure where it comes from; To me, it seems like it doesn't handle the `"source"` in the beginning but first later in the process. @devongovett can probably give some good insight into this.

artmsv:
Just ran into this issue. A bad workaround is to put an empty lock file into workspace which works with parcel.
Can we consider to fix this or review PR from @ch99q?

virketrang:
I installed the needed dependencies in the root of my workspace and it started working.

PR #7265, opened 2021-11-07
PNPM Workspaces, entry resolving for no input and accepting node_modules.

ch99q:
# ↪️ Pull Request
This pull request resolves issue #7063 and allows for Parcel to install dependencies in workspaces. It changes following things.
1. Adding `node_modules` to the list of acceptable names to identify a folder as a root folder.
2. Validating if the current directory could be the root directory if the current directory is the same as the entry.
e.g., `parcel build .`

## 💻 Examples
The test instructions tell an excellent example; the only difference is it's now supported to resolve the root dir as the folder where package.json lives in. (In case your current dir is the same as the entry in Parcel, e.g., running `parcel build .` would result in this).

## 🚨 Test instructions
1. Create a simple repository with pnpm as package manager, then add a project as workspace in the folder, e.g., `packages/playground.`
2. Add Parcel to the playground, and add the`"types"` field to trigger the auto module install function.
3. Parcel won't be able to find the playground as a project due to its missing lock files or .git

The result is the bundler is finding the root project and installing the dependencies in that folder, or it is simply failing to install in some cases.

## ✔️ PR Todo

### DIFF
--- a/packages/core/core/src/resolveOptions.js
+++ b/packages/core/core/src/resolveOptions.js
@@ -49,12 +49,18 @@ export default async function resolveOptions(
     entries = [path.resolve(inputCwd, initialOptions.entries)];
   }
 
-  let entryRoot = getRootDir(entries);
+  let entryRoot;
+  if(entries[0] === inputCwd) {
+    entryRoot = getRootDir([`${inputCwd}/package.json`])
+  } else {
+    entryRoot = getRootDir(entries);
+  }
+
   let projectRootFile =
     (await resolveConfig(
       inputFS,
       path.join(entryRoot, 'index'),
-      [...LOCK_FILE_NAMES, '.git', '.hg'],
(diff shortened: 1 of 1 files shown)


ch99q:
Please confirm it's not messing something else up; it's a pretty significant change to resolving root directories.

mischnic:
The project root is supposed to be the monorepo root. I don't think pnpm does anything different compared to Yarn?

ch99q:
> The project root is supposed to be the monorepo root. I don't think pnpm does anything different compared to Yarn?

But suppose I wish to use parcel in a repo inside a monorepo and try to trigger the parcel package manager to install dependencies. In that case, it won't be able to find the repo to install the dependencies. It will always look up to the place where the lock file is. And that's not always correct to assume that's the place; it's also a good thing to consider the folder containing the `node_modules` folder is the repo root.

Issue #6676, opened 2021-08-03
Parcel 2 - Bundle codegen contains non-ie11 compatible code - but only in watch mode


jamesrichards:
# 🐛 Bug

I have a project with multiple bundles using Parcel 1. We bundle on a component level so there's quite a few as you can imagine. I've tried upgrading to Parcel 2 this week but I can't seem to get IE11 working using Parcel 2's **watch mode** - but the **build** option is working fine.

## 🔦 Context

We use an IE11 polyfill bundle that we conditionally serve only in IE11. We don't use targets or anything fancy to serve this, its literally just added into the HTML (example below).

If I run a prod build - everything appears to work fine. But if I try to run the local dev watch task, IE throws a syntax error for all of the bundle files because there appears to still be non-compatible IE11 code present in the bundle.

## 💻 Examples

**package.json**

```
{
  "scripts": {
    "dev": "parcel watch ParcelBundles/*.js --public-url . --no-cache",
    "build": "parcel build ParcelBundles/*.js --public-url ."
  },
  "browserslist": [
    "> 1%",
    "last 2 versions",
    "ie 11",
    "not ie <= 10"
  ],
  "devDependencies": {
    "@parcel/transformer-sass": "^2.0.0-beta.3.1",
    "parcel": "^2.0.0-beta.3.1"
  },
  "dependencies": {
    "@parcel/babel-plugin-transform-runtime": "^2.0.0-nightly.1823",
    "@parcel/babel-preset-env": "^2.0.0-alpha.3",
    "whatwg-fetch": "^3.6.2"
  }
}
```

**babel.config.json**

```
{
    "presets": [
      [
        "@parcel/babel-preset-env",
        {
          "useBuiltIns": "entry",
          "corejs": 3
        }
      ]
    ],
    "plugins": ["@parcel/babel-plugin-transform-runtime"]
}
```

**ie11-polyfill.js**

```
import "core-js/stable";
import "regenerator-runtime/runtime";
import "whatwg-fetch";
```

This is what we get for every bundle file in IE11 when running watch mode:

![image](https://user-images.githubusercontent.com/5166247/128012074-10910fcf-803b-4b64-bafc-b01e1f8afb1b.png)

As I said before, running a prod build works fine.

Thanks in advance for taking a look, Parcel has been fantastic for us and we hope to continue with it in future.

jamesrichards:
[...]

PR #7016, opened 2021-10-05
Ensure named exports are prioritized over wildcard re-exports

devongovett:
Fixes #6344. Fixes #6970.

The ES module spec says that named exports should be prioritized over wildcard re-exports. This updates our helpers to skip keys that are already defined when performing a wildcard re-export. This is safe because we hoist named exports to the top of the module, so by the time the helper is called, all named exports are already defined.

Previously, we either errored when trying to read the previous value (non scope-hoisting), errored when trying to re-define the property (scope hoisting with multiple re-exports), or overwrote the value with the later one (scope hoisting). Now the value is consistent across all cases.

Note: duplicate wildcard re-exports is still not fully spec compliant, but we can address this separately. According to the spec, ambiguous re-exports should be completely omitted from the namespace object, but should error when performing a named import. We currently will return the first value instead. Unfortunately, it seems most tools are inconsistent in this regard.

PR #7293, opened 2021-11-12
Fix generated code order when using TypeScript, ES6 and decorators

thewilkybarkid:
# ↪️ Pull Request

This fixes the problem I mentioned in https://github.com/parcel-bundler/parcel/issues/6676#issuecomment-949504559. I'm not sure I understand the actual problem, but it's something to do with the hoisting difference between `var` and `let`, and decorators. (I didn't manage to recreate it without decorators.)

Moving the call to `parcelHelpers.export()` to the end (i.e. after the `let`) seems to resolve the problem.

## 💻 Examples

I've managed to reproduce the problem in https://github.com/thewilkybarkid/parcel-access-before-initialization, where you can see the compilation result and the failure. (It's basically the same as the test case in the PR.)

## ✔️ PR Todo


thewilkybarkid:
Only ran a subset of tests locally (as they take so long); will investigate those.

basher:
Thanks @thewilkybarkid for doing this.

Hopefully it will fix [#6676](https://github.com/parcel-bundler/parcel/issues/6676) so I can finally upgrade to Parcel 2 and still support certain legacy browsers!

basher:
Thanks @thewilkybarkid for doing this.

Hopefully is will finally fix [#6676](https://github.com/parcel-bundler/parcel/issues/6676).

mischnic:
Not sure if this entirely safe to do, circular imports might be problematic and also:

https://github.com/parcel-bundler/parcel/pull/7016:

> This updates our helpers to skip keys that are already defined when performing a wildcard re-export. This is safe because we hoist named exports to the top of the module, so by the time the helper is called, all named exports are already defined.

thewilkybarkid:
Are you able to write a test that breaks?

thewilkybarkid:
@mischnic As far as I'm aware the export order is still fine (it's just at the end of the module). The test I added contains circular imports too, but maybe you had a different idea.

I've just tried this out on a real project and it works fine, but still has the same problem with scope hoisting enabled (as the order is changed back)...

mischnic:
That just came to mind because we had problems with this in the past as well. I haven't had time yet to try and test such a situation

packages/core/integration-tests/test/typescript-tsc.js
```diff
@@ -56,4 +56,19 @@ describe('typescript tsc', function () {
...
+        config: path.join(
+          __dirname,
+          '/integration/typescript-es6-decorators/.parcelrc',
+        ),
+      },
+    );
+
+    const output = await run(b);
+    assert.strictEqual(output, 'ClassOne');
+  });
```
devongovett, line 73: Should we also run this test case with scope hoisting enabled? We should verify that the code behaves the same way in dev and prod.

devongovett:
Here's an example:

a.mjs
```js
import { c } from "./b.mjs";
console.log(c);
```

b.mjs
```js
export const foo = "foo";
export { c } from "./c.mjs";
```

c.mjs
```js
import { foo } from "./b.mjs";
export const c = "c:" + foo;
```

When run, this should throw "Cannot access 'foo' before initialization" (as it does when running the modules natively in Node or browsers). However, after applying this patch, it logs "c:undefined".

thewilkybarkid:
Thanks @devongovett, I've opened #7350 with that test case. I'll move this back to draft and will keep looking.

thewilkybarkid:
I've just been able to find https://github.com/angular/angular/commit/401ef71ae5b01be95d124184a0b6936fc453a5d4 and https://github.com/microsoft/TypeScript/issues/27519 which seem to describe this problem: it's a TypeScript limitation. I'm not sure how it was working in Parcel `2.0.0-beta1`, but I'll look more thoroughly at the generated output there.

Not sure what to do with this problem now though.

(The project where I've had this problem is inherited, and I've never used decorators before as they are known to be experimental. This gives me the excuse to remove them.)

devongovett:
I think the reason it may have "worked" in a previous beta is because we weren't spec compliant yet. Are you sure this circular dependency works with normal TSC or Babel?

thewilkybarkid:
> Are you sure this circular dependency works with normal TSC or Babel?

Seems to work with TSC: I've expanded https://github.com/thewilkybarkid/parcel-access-before-initialization to show a failure with Parcel but success with TSC.

PR #7295, opened 2021-11-13
Fix RangeError in `not export` error with other file type

### LINKED ISSUE #7278: RangeError: Invalid count value
# 🐛 bug report

Compiling the file below should cause a build-time "does not export" error because a missing CSS module class is imported.

But the diagnostic is either wrong (try deleting some lines in the repro) or cannot be printed at all.

## 😯 Current Behavior

Either
```
RangeError: Invalid count value
    at String.repeat (<anonymous>)
    at codeFrame (node_modules/@parcel/codeframe/lib/codeframe.js:214:44)
    at prettyDiagnostic (node_modules/@parcel/utils/lib/prettyDiagnostic.js:104:55)
    at async writeDiagnostic (node_modules/@parcel/reporter-cli/lib/CLIReporter.js:207:9)
```

or a codeframe that points to the wrong position in the source.

## 🔦 Context

https://github.com/parcel-bundler/parcel/issues/7272

## 💻 Code Sample

(After reformatting the file, error is different!)

An empty `index.module.css` file and

```js
import * as styles from "./index.module.css";

export const Foo = () => {
  const apply = async () => {
    await Promise.all([
      ...a.map((p) =>
        foo({
          access: p.access,
          poolUuid: p.poolUuid,
          userUuid: user.uuid,
        })
      ),
      ...(b.image
        ? [
            uploadFile(
              "/a/b/",
              await foo(
                b.a,
                b.b
              )
            ),
          ]
        : []),
    ]);
  };

  return (
    <div>
      <div className={styles.someClass}>
        <div>
          {isDirty && (
            <>
              <button
                className=""
                type="button"
                onClick={close}
              >
                Cancel
              </button>
              <button
                disabled={isSubmitting}
              >
                Update
              </button>
            </>
          )}
        </div>
      </div>
    </div>
  );
};
```

## 🌍 Your Environment

| Parcel           | 2.0.1

Shinyaigeek:
# ↪️ Pull Request

Fixes: #7278

Hi team ! 👋 
I'm interested in contributing to parcel, and I have been doing some research on #7278, and I noticed that this RangeError is caused by language mismatch in highlighting. For example, in the situation of this issue, the JavaScript highlight that is the source of the css import is being done as a css one, and the resulting change in the number of lines is causing a RangeError.

I fixed this Range error by setting the language of codeFrame in `A does not export B` error to the file type of the file from which A imports (JavaScript in the above example). However, I am not yet familiar with the inner code of parcel, and there may be some context that I missed. If you find something I missed, please feel free to point it out to me.

## 💻 Examples

```javascript
import React from "react";
import * as styles from "./hoge.module.css";

const App = function() {
  return (
    <div className={styles.notExisting} />
  )
}
```

```css
// empty css
```

## 🚨 Test instructions

I was wondering how to test it, and I wrote it in `scope-hoisting` integration test, but it may not be the right test to write here. If you have a better idea, I'd appreciate it if you could point it out.

For now, I wrote a test to see if the language of the codeFrame is javascript instead of css in the `hoge.module.css does not export notExisting` error in a case like an example above

## ✔️ PR Todo


packages/core/core/src/requests/AssetGraphRequest.js
```diff
@@ -554,7 +554,9 @@ export class AssetGraphBuilder {
                           this.options.projectRoot,
                           loc?.filePath,
                         ) ?? undefined,
-                      language: assetNode.value.type,
+                      language:
+                        incomingDep.value.sourceAssetType ??
+                        assetNode.value.type,
```
mischnic: `assetNode.value.type` is never correct, so this should rather be
```suggestion
                      language: incomingDep.value.sourceAssetType ?? undefined,
```

mischnic (commented):
Great find! Could you also move the test to `packages/core/integration-tests/test/postcss.js` and call it something like "should throw an error when importing a missing class"?

Here's an example for a test using scope hoisting in that file:
https://github.com/parcel-bundler/parcel/blob/f65182989334638e5fca7d4ce637bdf471b9b874/packages/core/integration-tests/test/postcss.js#L156-L195

packages/core/core/src/requests/AssetGraphRequest.js
Shinyaigeek: OK! Thank you for helping 👍

Shinyaigeek:
Thank you for reviewing and helping! I moved the test into `packages/core/integration-tests/test/postcss.js`, and fixed the test label to "should throw an error when importing a missing class”? in f5ca1bc 👍

mischnic (approved):

Issue #7306, opened 2021-11-15
Trying to augment a module causes @parcel/transformer-typescript-types to output a mostly blank .d.ts file


astegmaier:
# 🐛 bug report

If you try to use the [module augmentation](https://www.typescriptlang.org/docs/handbook/declaration-merging.html#module-augmentation) feature of typescript in the context of a library project that is using `@parcel/transformer-typescript-types` to generate a `.d.ts` file, parcel will output an incorrect `.d.ts` file with many elements missing.

## 🎛 Configuration (.babelrc, package.json, cli command)

Create a `library-original` package with the following export:
```ts
export class Person {
    constructor(public name: string) {}
}
```

Build a `library-augmenter` package with parcel:

***package.json***
```json
{
  "name": "library-augmenter",
  "types": "dist/index.d.ts",
  "source": "src/index.ts",
  "main": "dist/index.js",
  "scripts": {
    "build": "parcel build",
  },
  "dependencies": {
    "library-original": "1.0.0"
  },
  "devDependencies": {
    "@parcel/packager-ts": "^2.0.1",
    "@parcel/transformer-typescript-types": "^2.0.1",
    "parcel": "^2.0.1",
  }
}
```
***src/index.ts***
```js
import { Person } from "library-original";
Person.prototype.greet = function() { return `Hello ${this.name}!` }

// This will be present in the .d.ts file generated by parcel:
export const anotherThing: string = "hello";

// This will get dropped:
declare module "library-original" {
  interface Person {
    greet(): string;
  }
}

// This will also get dropped, if the "declare module" statement above is present:
export const somethingElse: string = "goodbye";
```

## 🤔 Expected Behavior

Parcel should ouput a complete `.d.ts` file, similar to `tsc`:
```ts
export declare const anotherThing: string;
declare module "library-original" {
    interface Person {
        greet(): string;
    }
}
export declare const somethingElse: string;
```

## 😯 Current Behavior

Parcel outputs a `.d.ts` file that is missing both the module augmentation code and any other exports that are declared after it:
```ts
export const anotherThing: string;
```

## 💁 Possible Solution

It looks like we're doing some tree shaking of the `.d.ts` file in these lines, and we probably didn't account for this case.

https://github.com/parcel-bundler/parcel/blob/145bfb8243c516765bcf5c59e381d0bc45fc5f9e/packages/transformers/typescript-types/src/TSTypesTransformer.js#L58-L69

## 🔦 Context

[...]

PR #7315, opened 2021-11-16
Fix module augmentation in @parcel/transformers-typescript-types

astegmaier:
# ↪️ Pull Request

**Edit: This PR is ready for a review - it works to solve the issues below, and includes a unit test to cover this case.**

I'd appreciate any feedback - I'm new to this part of parcel, and I want to make sure I don't break anything.

There are two issues that this PR tries to solve, both identified in #7306:

1. The current implementation of `@parcel/transformer-typescript-types` (intentionally) flattens _all_ module declarations, including declarations that are intended to [augment an external modules](https://www.typescriptlang.org/docs/handbook/declaration-merging.html#module-augmentation). Then, it will tree-shake things that it finds in the module augmentation, which (most of the time) will result in them being removed. **This PR solves the issue by avoiding flattening "nested" module declarations - e.g. module declarations that appear within other module declarations.**
2. There is a bug where if you have a module declaration in the middle of your source file, then exports declared after that declaration get (mistakenly) removed, because we think they are unused e.g.:
   ```ts
   import { Person } from "original";
   Person.prototype.greet = function() { return `Hello ${this.name}!` }
   // Exports declared before the augmentation declaration are present.
   export const anotherThing: string = "hello";

   declare module "original" {
     interface Person {
       greet(): string;
     }
   }
   // Everything after this point will be missing from the .d.ts file.
   export const somethingElse: string = "goodbye"; 
   ```
   **This PR solves the issue by using a stack to keep track of what level of module declaration you are in, and making sure to associate each node with the correct module even as you visit deep nodes, and then jump back out to the original parent node.**

## 🚨 Test instructions

See the `"should work with module augmentation"` test in `packages/core/integration-tests/test/ts-types.js`.

## ✔️ PR Todo


devongovett (approved):
Thank you!!

PR #7318, opened 2021-11-17
Feature: check empty string reference for other resource in HTML(and SVG)

### LINKED ISSUE #7076: Build fails due to iframe with empty 'src' property
# 🐛 bug report

An 'iframe' tag with an empty 'src' property causes the build to fail with an unhelpful error message.

package.json:

```js
{
  "name": "test",
  "source": "index.html",
  "scripts": {
    "start": "parcel"
  },
  "devDependencies": {
    "parcel": "2.0.0"
  }
}
```

## 🤔 Expected Behavior

The build succeeds, or in case it has to fail — a helpful error message is displayed.

## 😯 Current Behavior

```
$ yarn start
yarn run v1.22.15
$ parcel
🚨 Build failed.

@parcel/bundler-default: The expression evaluated to a falsy value:

(0, _assert().default)(context != null)


  AssertionError [ERR_ASSERTION]: The expression evaluated to a falsy value:

  (0, _assert().default)(context != null)

  at enter
  (<PROJECT_DIR>/node_modules/@parcel/bundler-default/lib/DefaultBundler.js:174:31)
  at <PROJECT_DIR>/node_modules/@parcel/graph/lib/Graph.js:520:16
  at <PROJECT_DIR>/node_modules/@parcel/graph/lib/Graph.js:520:16
  at walk (<PROJECT_DIR>/node_modules/@parcel/graph/lib/Graph.js:385:26)
  at walk (<PROJECT_DIR>/node_modules/@parcel/graph/lib/Graph.js:407:22)
  at walk (<PROJECT_DIR>/node_modules/@parcel/graph/lib/Graph.js:407:22)
  at walk (<PROJECT_DIR>/node_modules/@parcel/graph/lib/Graph.js:407:22)
  at ContentGraph.dfs (<PROJECT_DIR>/node_modules/@parcel/graph/lib/Graph.js:433:12)
  at ContentGraph.traverse
  (<PROJECT_DIR>/node_modules/@parcel/graph/lib/Graph.js:336:17)
  at ContentGraph.filteredTraverse
  (<PROJECT_DIR>/node_modules/@parcel/graph/lib/Graph.js:344:17)
```

## 💻 Code Sample

index.html:

```html
<!DOCTYPE html>
<html lang="en">
  <head>
    <meta charset="utf-8" />
  </head>
  <body>
    <iframe src=""></iframe>
  </body>
</html>
```

## 🌍 Your Environment

| Parcel           | 2.0.0
| Node             | 16.11.1
| npm/Yarn         | yarn 1.22.15

Shinyaigeek:
# ↪️ Pull Request

Fixes: #7076 

Hi team :wave:

I investigated about #7076, and I found this is caused by an empty string `uniqueKey` in `Asset`. `uniqueKey`’s default value is an empty string “”, and this and asset with empty string specifier breaks dependency graph at https://github.com/parcel-bundler/parcel/blob/v2/packages/core/core/src/AssetGraph.js#L517-L519. This is the cause of the strange error.

Of course, If I set a string other than an empty string to uniqueKey, the dependency graph will be built as expected, but this empty string reference will be resolved as `/` path and causes another strange behavior.

After consideration, I fixed this issue via checking whether or not a reference value in HTML (and also SVG) is empty string in `Transforming` as `SassTransformer` checks whether the imported file exists.

If there is some context I missed or you have some opinion about this approach, I will be happy if you point it out or feel free to close this PR.

## 💻 Examples

```html
<html><body><img src="" /></body><script src="" type="module"></script></html>
```

スクリーンショット 2021-11-18 1 36 11

## 🚨 Test instructions

I made a test case in `html` in integration tests to check `parcel` correctly throw an error or not with html like an above example.

## ✔️ PR Todo


DeMoorJasper (approved):

packages/transformers/html/src/dependencies.js
```diff
@@ -156,7 +156,7 @@ export default function collectDependencies(
       let href = attrs.href;
       if (attrs.rel === 'manifest') {
         // A hack to allow manifest.json rather than manifest.webmanifest.
-        // If a custom pipeline is used, it is responsible for running @parcel/transformer-webmanifest.
+        // If a custom pipestart is used, it is responsible for running @parcel/transformer-webmanifest.
```
devongovett:
```suggestion
        // If a custom pipeline is used, it is responsible for running @parcel/transformer-webmanifest.
```

devongovett:
Should we error on this, or should we just ignore empty strings? We would currently ignore if the attribute is not there entirely. Maybe someone has a reason for doing this?

packages/transformers/html/src/dependencies.js
Shinyaigeek: :+1: thanks!

mischnic:
Browsers don't ignore empty src attributes either:
https://humanwhocodes.com/blog/2009/11/30/empty-image-src-can-destroy-your-site/
https://stackoverflow.com/a/5946696

Shinyaigeek:
According to whatwg, a reference to the other resource such as `href` in <link /> or `src` in  should be non empty string valid URL, so HTMLTransformer (and also SVG) throws an error when handling empty string href. However, I did not know this context( > We would currently ignore if the attribute is not there entirely. ).

I think it is also reasonable to ignore empty string href, just as Parcel ignores the attribute which is not there entirely. On the other hand, I also think that giving an error on an invalid href will lead more kindful Developer Experience, what do you think about this?

FYI: https://html.spec.whatwg.org/multipage/semantics.html#attr-link-href
https://html.spec.whatwg.org/multipage/embedded-content.html#attr-img-src

devongovett (approved):
Makes sense! Thanks.

folknor:
This is presenting me warnings like these ones:
```
@parcel/transformer-html: value should not be empty string

  /.../index.html:968:11
    967 |                   <div class="form-check mb-3">
  > 968 |                     <input class="form-check-input" type="checkbox" value="" id="test1"
  >     |                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  > 969 |                       checked>
  >     | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    970 |                     <label class="form-check-label" for="test1">
    971 |                       ...

@parcel/transformer-html: checked should not be empty string

  /.../index.html:968:11
    967 |                   <div class="form-check mb-3">
  > 968 |                     <input class="form-check-input" type="checkbox" value="" id="test1"
  >     |                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  > 969 |                       checked>
  >     | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    970 |                     <label class="form-check-label" for="test1">
    971 |                       ...

```

But according to https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Input/checkbox both these values can be omitted:
> If the value attribute was omitted, the default value for the checkbox is on, so the submitted data in that case would be subscribe=on.

and

> To make a checkbox checked by default, you give it the checked attribute. See the below example:

(example uses an empty `checked` attribute)

PR #7369, opened 2021-11-26
Fix require statements with plain template literals

### LINKED ISSUE #5492: Recognize `require` call with template literals (but no placeholders)
# 🐛 bug report

```
require(`lodash`)
```
should behave just like `require("lodash")` but are currently ignored.

<br>

```
require(`./${thing}`)
```
should still be ignored.

## 💁 Possible Solution

This case should be normalized here:

https://github.com/parcel-bundler/parcel/blob/a4cf3e958ea1136b4829eb93615fba5328213471/packages/transformers/js/src/visitors/dependencies.js#L56-L61

and also normalized to `"lodash"`.

Add a test similar to this one: https://github.com/parcel-bundler/parcel/blob/46af2476ecb78921af21432f978d863fbffafd9d/packages/core/integration-tests/test/javascript.js#L24-L35

### LINKED ISSUE #7368: Require statements with simple template literal does not get detected
# 🐛 bug report

Require statements with template literal does not get detected by Parcel. With regular quotes it does work.

```js
const x = require(`my-package`);
```

```js
{
}
```

## 🤔 Expected Behavior

It should detect the dependency and add it to the bundle.

## 😯 Current Behavior

Require statement stays untouched and thus cannot be found when running the code.

## 💁 Possible Solution

Converts template literals without any arguments inside of it to regular string

## 🔦 Context

Trying to bundle a library that has node_modules which contains require statements with template literals.

## 💻 Code Sample

```
/*
 * This module is a side-effect filled module to load in the proper logger.
 */

export const startLogger = () => {
  // if (!process.env.GATSBY_LOGGER) {
  //   if (
  //     semver.satisfies(process.version, `>=8`) &&
  //     !isCI() &&
  //     typeof jest === `undefined`
  //   ) {
  //     process.env.GATSBY_LOGGER = `ink`
  //   } else {
  //     process.env.GATSBY_LOGGER = `yurnalist`
  //   }
  // }
  // // if child process - use ipc logger
  // if (process.send && !process.env.GATSBY_WORKER_POOL_WORKER) {
  //   // FIXME: disable IPC logger when inside worker. IPC messages crash jest-worker.
  //   // This is just workaround to not crash process when reporter is used in worker context.
  //   // process.env.FORCE_COLOR = `0`
  //   // TODO move to dynamic imports
  //   require(`./loggers/ipc`).initializeIPCLogger()
  // }
  if (process.env.GATSBY_LOGGER.includes(`json`)) {
    // TODO move to dynamic imports
    require(`./logger/test.js`).initializeJSONLogger();
  } else if (process.env.GATSBY_LOGGER.includes(`yurnalist`)) {
    // TODO move to dynamic imports
    require(`./logger/test.js`).initializeYurnalistLogger();
  } else {
    // TODO move to dynamic imports
    require(`./logger/test.js`).initializeINKLogger();
  }
};
```

## 🌍 Your Environment

| Parcel           |  2.0.1
| Node             | 14.17.5
| npm/Yarn         | 1.22.5

wardpeet:
# ↪️ Pull Request

Converts plain template literals into a string so dependency collector does find the dependency.

Fixes #7368
Fixes #5492

## 💻 Examples

```
export const startLogger = () => {
  require("./test");
  if (process.env.GATSBY_LOGGER.includes(`json`)) {
    // TODO move to dynamic imports
    require(`./logger/test.js`).initializeJSONLogger();
  } else if (process.env.GATSBY_LOGGER.includes(`yurnalist`)) {
    // TODO move to dynamic imports
    require(`./logger/test.js`).initializeYurnalistLogger();
  } else {
    // TODO move to dynamic imports
    require(`./logger/test.js`).initializeINKLogger();
  }
}
```

## ✔️ PR Todo

### DIFF
--- a/packages/transformers/js/core/src/dependency_collector.rs
+++ b/packages/transformers/js/core/src/dependency_collector.rs
@@ -539,6 +539,19 @@ impl<'a> Fold for DependencyCollector<'a> {
     }
 
     let node = if let Some(arg) = node.args.get(0) {
+      let mut arg = arg.clone();
+
+      // convert require(`./name`) to require("./name")
+      if let ast::Expr::Tpl(_tpl) = &*arg.expr {
+        if _tpl.quasis.len() == 1 && _tpl.exprs.is_empty() {
+          let tpl_str = &_tpl.quasis[0].raw;
+          arg.expr = Box::new(ast::Expr::Lit(ast::Lit::Str(ast::Str {
+            value: tpl_str.clone().value,
+            ..tpl_str.clone()
+          })));
+        }
+      }
+
       if kind == DependencyKind::ServiceWorker || kind == DependencyKind::Worklet {
         let (source_type, opts) = if kind == DependencyKind::ServiceWorker {
           match_worker_type(node.args.get(1))
--- a/packages/core/integration-tests/test/integration/commonjs-template-literal-interpolation/index.js
+++ b/packages/core/integration-tests/test/integration/commonjs-template-literal-interpolation/index.js
@@ -1,0 +1,7 @@
+const fn = 'add';
+
+module.exports = function (a, b) {
+  const add = require(`lodash/${fn}`);
+
+  return add(a, b);
+};
--- a/packages/core/integration-tests/test/integration/commonjs-template-literal-interpolation/package.json
+++ b/packages/core/integration-tests/test/integration/commonjs-template-literal-interpolation/package.json
@@ -1,0 +1,13 @@
+{
+  "name": "commonjs-require",
+  "private": true,
+  "main": "dist/index.js",
+  "targets": {
+    "main": {
+      "context": "node"
+    }
+  },
+  "dependencies": {
+    "lodash": "*"
+  }
+}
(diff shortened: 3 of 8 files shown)


packages/transformers/js/core/src/dependency_collector.rs
```diff
@@ -539,6 +539,19 @@ impl<'a> Fold for DependencyCollector<'a> {
...
+      if let ast::Expr::Tpl(_tpl) = &*arg.expr {
+        if _tpl.quasis.len() == 1 && _tpl.exprs.is_empty() {
+          let tpl_str = &_tpl.quasis[0].raw;
+          arg.expr = Box::new(ast::Expr::Lit(ast::Lit::Str(ast::Str {
+            value: tpl_str.clone().value,
+            ..tpl_str.clone()
+          })));
+        }
+      }
+
```
wardpeet, line 554: I'm not really familiar with Rust so if this code is not optimized or can be rewritten to something better, I'll take it :)

mischnic:
Can you add 2 new tests: one for the test you already changed (and revert that existing one), and one test that does something like this (which should still be ignored)
```js
let x = 2;
require(`foo${x}`)
```

wardpeet:
Updated, found another piece that needed to be updated.

devongovett (approved):
Thanks! We could probably make a util function for this but I can do that separately.

devongovett:
Here's that util: #7376

Issue #7312, opened 2021-11-16
Unable to create UMD bundle


bhovhannes:
# 🐛 bug report

I cannot get Parcel to create UMD bundle.

I've found https://github.com/parcel-bundler/parcel/issues/766 and https://github.com/parcel-bundler/parcel/pull/453, which makes me think that producing UMD bundle is supported, but it doesn't look like the UMD prelude code is in place in Parcel v2 and I didn't found any mentions of UMD or AMD formats in documentation either.

I was not able to join discord and ask if this is producing UMD bundles is even supported, so opening an issue. At the end it will be awesome to add documentation around this topic to stop any confusion (I'll be happy to do a PR).

## 🎛 Configuration (.babelrc, package.json, cli command)

No other config, I only have a `package.json`:
```json
{
  "name": "parcel-umd",
  "private": true,
  "scripts": {
    "build": "parcel build"
  },
  "targets": {
    "app": {
      "source": "src/app.js",
      "optimize": true,
      "context": "browser",
      "includeNodeModules": true,
      "distDir": "dist/"
    }
  },
  "devDependencies": {
    "parcel": "2.0.1"
  },
  "dependencies": {
    "regenerator-runtime": "^0.13.9"
  }
}

```

## 🤔 Expected Behavior

Generated bundle should contain UMD prelude.

## 😯 Current Behavior

Generated bundle does not contain UMD prelude and cannot be loaded using RequireJS.

## 💁 Possible Solution

I don't know, but I am ready to help :) 

## 🔦 Context

We're using webpack to combine polyfills imported from `es-shims` and `regenerator-runtime` packages into a single JS file, which is minified and loadable using RequireJS.

The command was something like:
```
webpack ./src/app.js -o ./dist/app.js --mode production --output-library-target umd --optimize-max-chunks 1
```

I wanted to do the same using Parcel and I cannot get Parcel to create UMD bundle. I've put a link to the repo below.

## 💻 Code Sample

Repo: https://github.com/bhovhannes/parcel-umd-build-issue

Do `npm install` and then `npm run build` to invoke parcel. Output will be in `dist/app.js`. It does not contain UMD prelude (searching for `define.amd` inside `dist/app.js` returns no matches).

## 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Parcel           | 2.0.1 |
| Node             |  14.17.3 |
| npm/Yarn         | npm 6.14.13 |
[...]

PR #7381, opened 2021-11-29
feat(core): add AMD output format support

bhovhannes:
# ↪️ Pull Request

This PR adds a new `outputFormat` - `amd`, which. will create bundle in [AMD format](https://requirejs.org/docs/whyamd.html#amd).

This is related to https://github.com/parcel-bundler/parcel/issues/7312, and most probably code in this MR will be slightly refactored later to reuse as much as it is possible for UMD support.  

But in any case this is a separate feature, and deserves a separate MR.

## 💻 Examples

Essentially, I want this to work:
```
{
  "source": "src/index.js",
  "main": "dist/index.js",
  "targets": {
    "main": {
      "context": "browser",
      "outputFormat": "amd",
      "includeNodeModules": {
        "lodash": false,
        "react": false,
        "react-dom": false
      }
    }
  }
}
```

## ✔️ PR Todo

### DIFF
--- a/packages/packagers/js/src/AMDOutputFormat.js
+++ b/packages/packagers/js/src/AMDOutputFormat.js
@@ -1,0 +1,94 @@
+// @flow
+import type {
+  ScopeHoistingPackager,
+  OutputFormat,
+} from './ScopeHoistingPackager';
+
+export class AMDOutputFormat implements OutputFormat {
+  packager: ScopeHoistingPackager;
+
+  constructor(packager: ScopeHoistingPackager) {
+    this.packager = packager;
+  }
+
+  getMainEntryAssetId(): ?string {
+    const mainEntryAsset = this.packager.bundle.getMainEntry();
+
+    // Sometimes there is no main asset, e.g. in the case of shared bundles.
+    // In that case we don't need AMD wrapper as parcelRequire will be used for
+    // loading shared bundle.
+    if (!mainEntryAsset) {
+      return;
+    }
+
+    const assetId = mainEntryAsset.id;
+    // We don't need AMD wrapper for assets wrapped in parcelRequire.
+    // For example, dynamically loaded chunk does not need to be an AMD.
+    if (this.packager.wrappedAssets.has(assetId)) {
+      return;
+    }
+    return assetId;
+  }
+
+  buildBundlePrelude(): [string, number] {
+    const assetId = this.getMainEntryAssetId();
+    if (!assetId) {
+      return ['', 0];
+    }
+
+    const envSupportsArrowFunctions = this.packager.bundle.env.supports(
+      'arrow-functions',
+      true,
+    );
+    let prelude = '';
+    let preludeLines = 0;
+
+    let depArgs = [`$${assetId}$exports`];
+    for (const [, specifiers] of this.packager.externals) {
+      let arg = specifiers.get('*');
+      if (!arg) {
+        arg = `$${assetId}$arg${depArgs.length}`;
+      }
+
+      if (envSupportsArrowFunctions) {
+        let destructuredItems = [];
+        for (const [originalName, mangledName] of specifiers) {
+          if (originalName !== '*') {
+            destructuredItems.push(`${originalName}: ${mangledName}`);
+          }
+        }
+        if (destructuredItems.length > 0) {
+          prelude += `const {${destructuredItems.join(', ')}} = ${arg};\n`;
+          preludeLines++;
+        }
+      } else {
+        for (const [originalName, mangledName] of specifiers) {
+          if (originalName !== '*') {
+            prelude += `var ${mangledName} = ${arg}['${originalName}'];\n`;
+            preludeLines++;
+          }
+        }
+      }
+
+      depArgs.push(arg);
+    }
+    depArgs = depArgs.join(', ');
+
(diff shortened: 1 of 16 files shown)


bhovhannes:
In this MR dynamic imports are not supported. I am still working on handling them properly.

bhovhannes:
Being new to Parcel codebase I have a couple questions:

1. Which type of tests are preferred? I've identified about 10 possible scenarios I want to cover with tests, and I've noticed [`packages/core/integration-tests/test/output-formats.js`](https://github.com/parcel-bundler/parcel/blob/v2/packages/core/integration-tests/test/output-formats.js), but I am not sure if it is desirable to increase number of integration tests that much (because they are slow). I didn't find any unit tests for [`ScopeHoistingPackager.js`](https://github.com/parcel-bundler/parcel/blob/v2/packages/packagers/js/src/ScopeHoistingPackager.js), which confuses me.

2. Should we use modern ES6 syntax in generated code to save a few bytes for supported environments or we can write es5-compatible code in all cases to make Parcel code shorter? Asking because I saw [`this.packager.bundle.env.supports('arrow-functions', true)` condition in `GlobalOutputFormat.js`](https://github.com/parcel-bundler/parcel/blob/v2/packages/packagers/js/src/GlobalOutputFormat.js#L15-L17) but without that conditions the code of `AMDOutputFormat.js` will be significantly shorter. Not sure if we care about adding extra bytes to each bundle here.

3. I have named function arguments in AMD prelude and worry about possible collisions between them and variables in existing source code. While I understand that there is no way to avoid collisions holistically, is there any established pattern used across codebase which will help minimize such collisions?

devongovett:
Thanks for working on this!

1. Yes, we generally prefer integration tests. This lets us more easily change the implementation without breaking the tests (we reused a lot of tests from v1 in v2). The `output-formats.js` file is the right place.
2. I don't mind complexity in Parcel if it makes for smaller output. But I'll leave this up to you for the initial implementation. We can always optimize later if you want.
3. All top-level variable names within the generated code will already have been renamed to something like `$assetId$var$xyz`, so it should generally be safe to use variable names that don't fall under that pattern. The pattern you've used looks ok, though you could perhaps use `this.packager.bundle.publicId` as the prefix rather than the main asset id. Sometimes there is no main asset, e.g. in the case of shared bundles.

packages/packagers/js/src/AMDOutputFormat.js
```diff
@@ -0,0 +1,67 @@
...
+      (envSupportsArrowFunctions
+        ? `define(${deps}, (${depArgs}) => {\n`
+        : `define(${deps}, function (${depArgs}) {\n`) + prelude;
+    preludeLines++;
+
+    return [prelude, preludeLines];
+  }
+
+  buildBundlePostlude(): [string, number] {
+    return ['});', 0];
```
devongovett, line 92: Should the function return something? e.g. so the module could expose its main asset. I guess only for library entries?

bhovhannes: I am using this form of AMD, thus no return statement:
```
define(['exports', 'lodash'], (exports, _) => {
   exports.weirdSum = (a, b) => _.uppercase(a + b)
})
```

devongovett: Ah got it. 👍

bhovhannes:
Thank you for review @devongovett, and for your work on Parcel! You did a huge amount of work there.

> 1. Yes, we generally prefer integration tests. This lets us more easily change the implementation without breaking the tests (we reused a lot of tests from v1 in v2). The `output-formats.js` file is the right place.

ok, I'll add test there.

> 2. I don't mind complexity in Parcel if it makes for smaller output. But I'll leave this up to you for the initial implementation. We can always optimize later if you want.

ok, I'll add tests covering that as well.

> 3. All top-level variable names within the generated code will already have been renamed to something like `$assetId$var$xyz`, so it should generally be safe to use variable names that don't fall under that pattern. The pattern you've used looks ok, though you could perhaps use `this.packager.bundle.publicId` as the prefix rather than the main asset id. Sometimes there is no main asset, e.g. in the case of shared bundles.

That is very helpful, I'll change to use `this.packager.bundle.publicId`. Thanks!

bhovhannes:
Hi @devongovett , @mischnic , Happy New Year!

I've added tests to AMD output format implementation, so this MR is ready for final review.

I was not able to find a way to create a sample repo where Parcel produces shared bundle for the 2 JS entries. Producing shared chunks works for HTML entries (like is written [here](https://parceljs.org/features/code-splitting/#shared-bundles)), but I didn't find a way to generate shared chunks for JS entries. Thus, I was not able to test that case, although I think I covered that case by checking for empty main asset it and short-circuiting if so.  
Currently, only the main asset code will be wrapped in AMD wrapper, I exclude dynamically loaded chunks as well because they are loaded by `parcelRequire`.

I think I also need to mention `AMD` format in https://parceljs.org/features/targets/#outputformat. ~~What is the repo behind Parcel docs website so I can send an MR there as well?~~ found it, never mind.

bhovhannes:
Friendly ping 🙂

oe:
any news?

jamenai:
Any news here?

PR #7390, opened 2021-12-01
Resize HTML images based on the width and height attributes

thewilkybarkid:
# ↪️ Pull Request

This is one of the steps towards https://github.com/parcel-bundler/parcel/issues/3738: it resizes images in HTML based on the `width` and/or `height` attribute.

## 💻 Examples

```html
<img src="200x100.png" width="100" height="50">
```

is processed as if it were

```html
<img src="200x100.png?width=100&height=50" width="100" height="50">
```

## ✔️ PR Todo


packages/transformers/html/src/dependencies.js
```diff
@@ -123,6 +124,20 @@ export default function collectDependencies(
...
+      const src = url.parse(attrs.src, true);
+      if (!src.protocol && !src.query.width && !src.query.height) {
+        if (attrs.width) {
+          src.query.width = attrs.width;
+        }
+        if (attrs.height) {
+          src.query.height = attrs.height;
+        }
+
+        attrs.src = url.format({pathname: src.pathname, query: src.query});
```
thewilkybarkid, line 137: Was surprised to find that `url.format(src)` doesn't use the updated `query`. I know that the Node URLs are deprecated in favour of the WHATWG form, but that doesn't support relative URLs (which is the only case we want to process here).

So... this works, but there's probably a better way of doing this.

PR #7406, opened 2021-12-04
Add error message for disabled autoinstall

### LINKED ISSUE #6165: Different error message for disabled autoinstall
# 🐛 bug report

At the moment, the error message for `parcel build --no-autoinstall` and
- a missing parcel plugin specified in the config that usually would get autoinstalled (e.g. the SASS transformer) or
- `packageManager.resolve` and `packageManager.require`

is just `Cannot find module foo, did you mean bar`.

In this case, it should say something along the lines of `Autoinstall is disabled, please install foo yourself`.

# Relevant code

https://github.com/parcel-bundler/parcel/blob/3baf990a686d1c1dd0c5f4360296de890cc00c47/packages/core/core/src/loadParcelPlugin.js#L82-L126

astralbody:
# ↪️ Pull Request

Closes https://github.com/parcel-bundler/parcel/issues/6165

## ✔️ PR Todo


packages/core/integration-tests/test/javascript.js
```diff
@@ -5308,8 +5309,10 @@ describe('javascript', function () {
         name: 'BuildError',
         diagnostics: [
           {
-            message: `Failed to resolve '@swc/helpers' from '${normalizePath(
-              require.resolve('@parcel/transformer-js/src/JSTransformer.js'),
+            message: `Failed to resolve '@swc/helpers' from '${escapeMarkdown(
```
astralbody: I added `escapeMarkdown` because my working directory contains an underscore. The test is failed if the path is not escaped.

packages/core/package-manager/src/installPackage.js
```diff
@@ -182,9 +185,16 @@ async function determinePackageInstaller(
...
-let queue = new PromiseQueue({maxConcurrent: 1});
+const promiseQueueOptions = {maxConcurrent: 1};
+let queue = new PromiseQueue(promiseQueueOptions);
 let modulesInstalling: Set<string> = new Set();
 
+// This function is needed to clear _error field between different unit test files
+// PromiseQueue only saves a first error and returns it each time even though another error is thrown
+export function clearPromiseQueue() {
+  queue = new PromiseQueue(promiseQueueOptions);
+}
```
astralbody, line 196: I didn't find a solution that is better than adding this function. You could reproduce my problem with the next steps:
1. add `.only` to `should error with a codeframe if a plugin is not resolved`
2. remove the next line:
https://github.com/parcel-bundler/parcel/blob/aa2bf861506c4df5b3c90d17f111bfbb72f6a721/packages/core/core/test/ParcelConfig.test.js#L274
3. add `.only` to `NodePackageManager`
4. run `yarn test:unit`

It's worth mentioning:
https://github.com/parcel-bundler/parcel/blob/aa2bf861506c4df5b3c90d17f111bfbb72f6a721/packages/core/utils/src/PromiseQueue.js#L87-L91

packages/core/package-manager/src/installPackage.js
```diff
@@ -182,9 +185,16 @@ async function determinePackageInstaller(
...
 
-let queue = new PromiseQueue({maxConcurrent: 1});
+const promiseQueueOptions = {maxConcurrent: 1};
+let queue = new PromiseQueue(promiseQueueOptions);
 let modulesInstalling: Set<string> = new Set();
 
+// This function is needed to clear _error field between different unit test files
+// PromiseQueue only saves a first error and returns it each time even though another error is thrown
+export function clearPromiseQueue() {
+  queue = new PromiseQueue(promiseQueueOptions);
```
astralbody, line 195: I didn't find a solution that is better than adding this function. You could reproduce my problem with the next steps:
1. add `.only` to `should error with a codeframe if a plugin is not resolved`
2. remove the next line:
https://github.com/parcel-bundler/parcel/blob/aa2bf861506c4df5b3c90d17f111bfbb72f6a721/packages/core/core/test/ParcelConfig.test.js#L274
3. add `.only` to `NodePackageManager`
4. run `yarn test:unit`

It's worth mentioning:
https://github.com/parcel-bundler/parcel/blob/aa2bf861506c4df5b3c90d17f111bfbb72f6a721/packages/core/utils/src/PromiseQueue.js#L87-L91

Pi-Bouf (approved):

packages/core/package-manager/src/index.js
```diff
@@ -5,3 +5,4 @@ export * from './Pnpm';
 export * from './Yarn';
 export * from './MockPackageInstaller';
 export * from './NodePackageManager';
+export * from './installPackage';
```
devongovett: Maybe we can avoid exporting this from the package root and just import from the subpath directly in the test? This shouldn't really be public...

Pi-Bouf (approved):

packages/core/package-manager/src/index.js
astralbody: Done.

PR #7418, opened 2021-12-08
Update: allow empty string toml config

### LINKED ISSUE #7371: getConfig returns null for empty config file
# 🐛 bug report

I found it very unexpected that `config.getConfig[From]` returns null (usually meaning that no config file could be found) if a config file was found.

Especially because there are config formats where an empty file is valid (yaml, toml).

## 🤔 Expected Behavior

Just parse the file, an empty json config file would throw a parse error (which would be fine/correct).

## 😯 Current Behavior

Null is returned

## 💁 Possible Solution

Remove this line https://github.com/parcel-bundler/parcel/blob/b9426b83f86809143ff122bde3ee217fb64a8d78/packages/core/utils/src/config.js#L85

## 🌍 Your Environment

| Parcel           | b9426b83f86809143ff122bde3ee217fb64a8d78

Shinyaigeek:
# ↪️ Pull Request

Fixes: #7371

Hi! team 👋 
Currently, `loadConfig` module early returns null when configuration file is an empty string. However, some files such as `.toml` and `.yaml` allow an empty string file, so `loadConfig` should not early return null when the config file is an empty string.

I fixed this behavior to parse also empty string files (, and a file parser throws an error when a file parser such as JSON does not allow empty string).

## 💻 Examples

with empty json

スクリーンショット 2021-12-09 1 23 29

## 🚨 Test instructions

- parse a valid JSON config file
- `loadConfig` should throw an error with the invalid empty string JSON
- `loadConfig` can load a valid empty string toml

## ✔️ PR Todo


mischnic (approved):
Thanks!

devongovett (approved):

Issue #7400, opened 2021-12-03
Invalid TypeScript declaration files (.d.ts) emitted when mono-repo modules are being referred.


SayBGM:
# 🐛 bug report
It seems Parcel fails to emit valid TypeScript declaration files (.d.ts) when the input source is importing any of its mono-repo related modules.

https://github.com/Mint-SayBGM/parcel-typescript-interface-error 

## 🎛 Configuration (.babelrc, package.json, cli command)
package.json
```json
{
  "name": "root",
  "private": true,
  "devDependencies": {
    "@babel/core": "^7.16.0",
    "@babel/preset-env": "^7.16.0",
    "@parcel/packager-ts": "^2.0.1",
    "@parcel/transformer-typescript-types": "^2.0.1",
    "@parcel/config-default": "^2.0.1",
    "@parcel/transformer-typescript-tsc": "^2.0.1",
    "@swc/helpers": "^0.2.13",
    "@typescript-eslint/eslint-plugin": "^5.1.0",
    "@typescript-eslint/parser": "^5.1.0",
    "eslint": "^7.11.0",
    "eslint-plugin-react": "^7.26.1",
    "lerna": "^4.0.0",
    "parcel": "^2.0.1",
    "typescript": "^4.4.4"
  },
  "scripts": {
    "prepare": "lerna run prepare",
    "dev": "lerna run dev --stream",
    "build": "yarn & lerna run build --stream",
    "bootstrap": "lerna bootstrap --use-workspaces"
  },
  "workspaces": [
    "packages/*"
  ]
}

```
packages/sdk/package.json
```json
{
  "name": "@modules/sdk",
  "version": "0.0.1",
  "license": "ISC",
  "baseUrl": "./src",
  "types": "dist/index.d.ts",
  "main": "dist/index.js",
  "targets": {
    "main": {
      "optimize": true
    }
  },
  "scripts": {
    "prepublishOnly": "yarn build",
    "dev": "yarn build",
    "build": "yarn parcel build src/index.ts",
    "start": "yarn parcel src/index.ts"
  }
}
```
packages/types/package.json
```json
{
  "name": "@modules/types",
  "version": "1.0.0",
  "private": true,
  "description": "> TODO: description",
  "homepage": "",
  "license": "ISC",
  "types": "dist/index.d.ts",
  "main": "dist/index.js",
  "scripts": {
    "prepare": "yarn build",
    "build": "yarn parcel build index.ts"
  }
}
```

[...]

Issue #5978, opened 2021-03-10
typescript not using semver, should peerDependencies be > instead of ^?


fuxingloh:
# 🐛 bug report

With auto install, parcel attempts to download the typescript transformer this fails due to peer deps resolution. I use typescript `4.2.2` for my development, while peerDeps for the transformer package is set to `^3.0.0`. This implies peer deps from `3.0.0` to `3.9.9`, however typescript is not using semver and `4.0.0` is not a major release non a breaking change, so it should be `>=3.0.0` instead?

Typescript is not using semver: https://github.com/microsoft/TypeScript/issues/14116

https://github.com/parcel-bundler/parcel/blob/0a3b4372a84c90bfd22654c7257a4c7ec133f6b6/packages/transformers/typescript-types/package.json#L28-L33

## 🎛 Configuration (.babelrc, package.json, cli command)

```js
{
  "devDependencies": {
    "@defichain/testcontainers": "0.0.0",
    "parcel": "2.0.0-nightly.620",
    "typescript": "4.2.2"
  }
}
```

## 🤔 Expected Behavior

Auto install should install without throwing peerDependencies error as typescript is not using semver.

## 😯 Current Behavior

Auto install fail if there is a conflicting peerDependencies. (Meanwhile, I force install with `npm  i --force` and use `--no-autoinstall`)

## 💁 Possible Solution

Maybe use `>=3.0.0` instead for all typescript peerDeps?

## 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Parcel           | 2.0.0-nightly.620
| Node             | v14.15.0
| npm/Yarn         | 7.5.6
| Operating System | MacOS

PR #7426, opened 2021-12-10
Fix createImportSpecifier with typescript 4.5+

### LINKED ISSUE #7325: [bug] Syntax error in TypeScript type defination files
### `package.json`

```json
{
    "source": "src/index.ts",
    "main": "dist/index.js",
    "module": "dist/index.esm.js",
    "types": "dist/index.d.ts",
    "dependencies": {
        "@swc/helpers": "^0.2.14",
        "web-utility": "^2.9.6"
    },
    "peerDependencies": {
        "katex": "^0.15.1",
        "react": ">=16",
        "react-bootstrap": "^1.6.4"
    },
    "devDependencies": {
        "@parcel/packager-ts": "^2.0.1",
        "@parcel/transformer-typescript-types": "^2.0.1",
        "@types/jest": "^27.0.2",
        "@types/katex": "^0.11.1",
        "@types/react": "^16.14.21",
        "husky": "^7.0.4",
        "jest": "^27.3.1",
        "katex": "^0.15.1",
        "lint-staged": "^12.0.3",
        "open-cli": "^7.0.1",
        "parcel": "^2.0.1",
        "prettier": "^2.4.1",
        "react": "^16.14.0",
        "react-bootstrap": "^1.6.4",
        "ts-jest": "^27.0.7",
        "typedoc": "^0.22.9",
        "typedoc-plugin-mdn-links": "^1.0.4",
        "typescript": "^4.5.2"
    },
    "browserslist": "> 0.5%, last 2 versions, not dead",
    "targets": {
        "main": {
            "optimize": false
        }
    }
}
```

### `tsconfig.json`

```json
{
    "compilerOptions": {
        "target": "ES5",
        "esModuleInterop": true,
        "jsx": "react",
        "lib": ["ES2021", "DOM"]
    },
    "include": ["src/**/*"]
}
```

## 🤔 Expected Behavior

Generate right **Type Defination** files for libraries, as what Parcel 2.0.0 did.

## 😯 Current Behavior

`dist/index.d.ts`

```typescript
import { PickData as  } from "web-utility";
import { PropsWithoutRef as , RefCallback as , PropsWithChildren as  } from "react";
import { type _PickData1 as  } from "web-utility/source/data";
```

## 🌍 Your Environment

| Parcel           | 2.0.1 |
| TypeScript | 4.5.2 |
| Node             | 14.17.6 |
| Yarn         | 1.22.15 |
| Operating System | Windows 10 Pro 21H1 |

astegmaier:
# ↪️ Pull Request

This fixes #7325 and possibly also #7400. The issue, as @benpigchu pointed out [here](https://github.com/parcel-bundler/parcel/issues/7325#issuecomment-988697578), was that Typescript 4.5 added support for [type modifiers on specific import names](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-5.html#type-modifiers-on-import-names). For example, it is now valid to write:
```ts
import { someFunc, type BaseType } from "./some-module.js";
```
Unfortunately, the way that the compiler API was changed to support this is not backward compatible :-(

Here's what the pre-4.5 API looked like:
```ts
createImportSpecifier(propertyName: Identifier | undefined, name: Identifier): ImportSpecifier;
```
And here's the new API ([see PR](https://github.com/microsoft/TypeScript/pull/45998)):
```ts
createImportSpecifier(isTypeOnly: boolean, propertyName: Identifier | undefined, name: Identifier): ImportSpecifier;
```
Since there's no way that I know of to check a function's arity at runtime, I had to resort to a version check.

There were several other backward-incompatible changes in that PR (to `updateImportSpecifier`, `createExportSpecifier` and `updateExportSpecifier`), and I verified that we don't use these APIs in parcel.

## 💻 Examples

Without this change, when you have typescript 4.5+, `@parcel/transformer-typescript-types` will generate output like:
```ts
import { something as } from "./foo"
```
instead of
```ts
import { something } from "./foo
```
This change ensures that we do the right thing.

## 🚨 Test instructions

Several of the existing unit tests will break if you upgrade to typescript 4.5+ and don't apply this change, so it didn't seem necessary to add new ones.

I manually verified that with this change, and typescript 4.2.4 (the old version in the repo), all the unit tests still pass. You can test it yourself by modifying `yarn.lock` to revert back to `4.2.4`, running `yarn install`, then `yarn test:integration`.

## ✔️ PR Todo

### DIFF
--- a/packages/transformers/typescript-types/src/utils.js
+++ b/packages/transformers/typescript-types/src/utils.js
@@ -1,5 +1,6 @@
 // @flow
 import typeof TypeScriptModule from 'typescript'; // eslint-disable-line import/no-extraneous-dependencies
+import type {Identifier, ImportSpecifier} from 'typescript';
 
 export function getExportedName(ts: TypeScriptModule, node: any): ?string {
   if (!node.modifiers) {
@@ -25,4 +26,23 @@ export function isDeclaration(ts: TypeScriptModule, node: any): boolean {
     ts.isEnumDeclaration(node) ||
     ts.isTypeAliasDeclaration(node)
   );
+}
+
+export function createImportSpecifier(
+  ts: TypeScriptModule,
+  propertyName: Identifier | void,
+  name: Identifier,
+  isTypeOnly: boolean = false,
+): ImportSpecifier {
+  const [majorVersion, minorVersion] = ts.versionMajorMinor
+    .split('.')
+    .map(num => parseInt(num, 10));
+  // The signature of createImportSpecifier had a breaking change in Typescript 4.5.
+  // see: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-5.html#type-modifiers-on-import-names
+  if (majorVersion > 4 || (majorVersion === 4 && minorVersion >= 5)) {
+    // $FlowFixMe
+    return ts.createImportSpecifier(isTypeOnly, propertyName, name);
+  } else {
+    return ts.createImportSpecifier(propertyName, name);
+  }
 }
--- a/packages/transformers/typescript-types/src/shake.js
+++ b/packages/transformers/typescript-types/src/shake.js
@@ -4,7 +4,7 @@ import type {TSModuleGraph} from './TSModuleGraph';
 
 import ts from 'typescript';
 import nullthrows from 'nullthrows';
-import {getExportedName, isDeclaration} from './utils';
+import {getExportedName, isDeclaration, createImportSpecifier} from './utils';
 
 export function shake(
   moduleGraph: TSModuleGraph,
@@ -238,7 +238,8 @@ function generateImports(moduleGraph: TSModuleGraph) {
         );
       } else {
         namedSpecifiers.push(
-          ts.createImportSpecifier(
+          createImportSpecifier(
+            ts,
             name === imported ? undefined : ts.createIdentifier(imported),
             ts.createIdentifier(name),
           ),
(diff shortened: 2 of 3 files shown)


packages/transformers/typescript-types/src/utils.js
```diff
@@ -26,3 +27,22 @@ export function isDeclaration(ts: TypeScriptModule, node: any): boolean {
...
+  ts: TypeScriptModule,
+  propertyName: Identifier | void,
+  name: Identifier,
+  isTypeOnly: boolean = false,
+): ImportSpecifier {
+  const [majorVersion, minorVersion] = ts.versionMajorMinor
+    .split('.')
+    .map(num => parseInt(num, 10));
+  // The signature of createImportSpecifier had a breaking change in Typescript 4.5.
+  // see: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-5.html#type-modifiers-on-import-names
```
devongovett, line 41: Wowwwwww... TS doesn't follow semver?? Maybe we should pin our dependency on it?

devongovett (approved):

packages/transformers/typescript-types/src/utils.js
mischnic: The other side of that coin: https://github.com/parcel-bundler/parcel/issues/5978

Jack-Works:
> there's no way that I know of to check a function's arity at runtime

Yes you can

image

mischnic:
Though that would break if Typescript implemented `createImportSpecifier` using something like `createImportSpecifier(...args)` in the future.

PR #7435, opened 2021-12-11
Fix: transpiling React.createElement into Preact h behavior

### LINKED ISSUE #7433: Automatic JSX runtime doesn't work when React is aliased to Preact
# 🐛 bug report

```json
{
    "scripts": {
        "start": "parcel index.html",
        "build": "parcel build index.html"
    },
    "devDependencies": {
        "parcel": "^2.0.1"
    },
    "dependencies": {
        "preact": "^10.6.4"
    },
    "alias": {
        "react": "preact/compat",
        "react-dom/test-utils": "preact/test-utils",
        "react-dom": "preact/compat",
        "react/jsx-runtime": "preact/jsx-runtime"
    }
}
```

## 🤔 Expected Behavior

The automatic JSX runtime should be used (= no import of React is necessary) even if Preact is aliased to React.

## 😯 Current Behavior

Without an alias, you don't need to import `h` of `React` when just using Preact/React.

With alias, you need to import `React`.

## 💁 Possible Solution

Related code:

https://github.com/parcel-bundler/parcel/blob/b27515ec0eb557971dfe83ad3692fd208a9ee397/packages/transformers/js/src/JSTransformer.js#L147

## 💻 Code Sample

```jsx
import React from "react"; // this shouldn't be needed

function App() {
	return "hello";
}

render(<App />, document.getElementById("root"));
```

## 🌍 Your Environment

| Parcel           | 2.0.1

Shinyaigeek:
# ↪️ Pull Request

Hi team :wave:

Fixes: #7433

`jsxImportSource` will be determined with `reactLib` variable, but `reactLib` variable is always “react” string literal if `alias.react is setted`. I think `jsx-runtime` (and automatic transformation) is not related.

I fixed this

## 💻 Examples

**package.json**
```json
{
  "alias": {
    "react": "preact/compat"
  }
}
```

## 🚨 Test instructions

I made a test case to check jsx runtime transformation with preact with `alias.react` package.json.

## ✔️ PR Todo


packages/core/integration-tests/test/transpilation.js
```diff
@@ -120,7 +120,7 @@ describe('transpilation', function () {
     await bundle(path.join(__dirname, '/integration/jsx-react-alias/index.js'));
 
     let file = await outputFS.readFile(path.join(distDir, 'index.js'), 'utf8');
-    assert(file.includes('React.createElement("div"'));
```
Shinyaigeek: This test case seems wrong. In this test case, `React.createElement` should be transpiled into `h`.

packages/transformers/js/src/JSTransformer.js
```diff
@@ -126,6 +126,37 @@ const SCRIPT_ERRORS = {
   },
 };
 
+function convertAliasReactIntoPragma(
```
Shinyaigeek: In order to convert `preact/compat` into `preact` pragma, I made a function to do it, but I feel this approach is a little muddy. If you have another idea to implement this, I will be happy if you comment on it.

packages/core/integration-tests/test/transpilation.js
devongovett: Are we sure this is right? It should map to `preact/compat` in this case, which should have `createElement` (and may be different than pure Preact for compatibility reasons).

Shinyaigeek: Oh I see ! I’m sorry I misunderstood :bow:
As you say, `preact/compat` exports `createElement` and does not export `h` and `React` is aliased to `preact/compat`, so `React.createElement` should be transpiled into `React.createElement`, so this test case is not wrong.

Thank you for pointing out!

Shinyaigeek:
I think this approach is wrong. I thought #7433  behavior is caused by this line(https://github.com/parcel-bundler/parcel/blob/v2/packages/transformers/js/src/JSTransformer.js#L160) in which reactLib is fixed to `react` when aliased . However, As this comment (https://github.com/parcel-bundler/parcel/pull/7435#discussion_r768001211), `reactLib` should be fixed to `react` when aliased.

I think the cause is not this line, but these line(https://github.com/parcel-bundler/parcel/blob/v2/packages/transformers/js/src/JSTransformer.js#L201). In these line, `JSTransformer` decides whether `parcel` (and backend `swc`) should automatically transform `createElement` to new JSX Runtime API, and decides which `JSXImportSource` should be used. 
The cause for this issue is that `reactLib` is fixed to `react` so this processing runs in the case only `react`. I feel I should fix these line to insert the specific processing when `pkg.alias` is configured.


It's 3am in my timezone, so once I back this PR  to draft and sleep, and I’ll work tomorrow :bow:.

Shinyaigeek:
I fixed 👍

packages/core/integration-tests/test/transpilation.js
Shinyaigeek: 09b9844 👍

mischnic:
Does the current version still have the problem that `reactLib` (and therefore `jsxImportSource`) is set to "preact" even though it should be "react"?

I think something like this should work. The other react-like libraries don't support this automatic runtime anyway

```diff
--- packages/transformers/js/src/JSTransformer.js
+++ packages/transformers/js/src/JSTransformer.js
@@ -199,11 +199,15 @@ export default (new Transformer({
         jsxImportSource = compilerOptions?.jsxImportSource;
         automaticJSXRuntime = true;
       } else if (reactLib) {
-        let automaticVersion = JSX_PRAGMA[reactLib]?.automatic;
+        let effectiveReactLib =
+          (pkg?.alias && pkg.alias['react']) === 'preact/compat'
+            ? 'preact'
+            : reactLib;
+        let automaticVersion = JSX_PRAGMA[effectiveReactLib]?.automatic;
         let reactLibVersion =
-          pkg?.dependencies?.[reactLib] ||
-          pkg?.devDependencies?.[reactLib] ||
-          pkg?.peerDependencies?.[reactLib];
+          pkg?.dependencies?.[effectiveReactLib] ||
+          pkg?.devDependencies?.[effectiveReactLib] ||
+          pkg?.peerDependencies?.[effectiveReactLib];
         let minReactLibVersion =
           reactLibVersion != null && reactLibVersion !== '*'
             ? semver.minVersion(reactLibVersion)?.toString()
```

Shinyaigeek:
As you say, it's not `JSTransformation`'s responsibility to resolve the alias, I overlooked it. There are many libraries that support `jsxRuntime`, but only "preact" supports `{reactLib}/jsx-runtime` alias, and I feel that even if more library support that alias and `parcel` supports it in the future, It's enough to abstract it to at that time. I fixed it in e05804a :+1:

Thank you for helping!

packages/core/integration-tests/test/transpilation.js
```diff
@@ -183,6 +183,18 @@ describe('transpilation', function () {
...
+  it('should support the automatic JSX runtime with preact with alias', async function () {
+    let b = await bundle(
+      path.join(
+        __dirname,
+        '/integration/jsx-automatic-preact-with-alias/index.js',
+      ),
+    );
+
+    let file = await outputFS.readFile(b.getBundles()[0].filePath, 'utf8');
+    assert(file.includes('_jsxRuntime.jsx("div"'));
```
mischnic: I think we should also add this to make sure it's actually importing from React (using a Regex because `react/jsx-runtime` is a substring of `preact/jsx-runtime`).

```suggestion
    let file = await outputFS.readFile(b.getBundles()[0].filePath, 'utf8');
    assert(/\Wreact\/jsx-runtime\W/.test(file));
    assert(file.includes('_jsxRuntime.jsx("div"'));
```

Shinyaigeek: Sure. I fixed in 4efec78 :+1: Thanks!

mischnic (approved):

devongovett (approved):

PR #7506, opened 2021-12-30
fix: add undeclared dependencies

merceyz:
# ↪️ Pull Request

`@parcel/graph` depends on `@parcel/utils` and `@parcel/cache` depends on `@parcel/fs` but neither declare it so this PR updates them to declare their dependencies.


mischnic (approved):
Thanks!

PR #5674, opened 2021-01-16
Remove postProcess

DeMoorJasper:
# ↪️ Pull Request

This PR removes the postprocess functionality as discussed in the api audit

Related to #5225

PR #7508, opened 2021-12-31
Add transformer for svelte (#7503).

### LINKED ISSUE #7503: Parcel 2: Svelte transformer
There is already a third party transformer package. It is seemingly unmaintained but mostly functional. I have previously fixed some of the bugs in the published package (orlov-vo/parcel-transformer-svelte#24) and would be willing to maintain it here.

It seems rude to just outright fork the original especially considering they already have npmjs.com name reserved; however, the original author has given permission to port the transformer over to the mainline parcel repo (orlov-vo/parcel-transformer-svelte#28).

The package would live at `packages/transformers` and would be published as `@parcel/transformer-svelte`. There would be commensurate documentation as well on the official website.

Would a PR for this be accepted? Also, should a PR for this be accepted, I have a number of "tests" which are really just example projects which should compile properly. Where should those be placed in this repo?

gcoakes:
This is the implementation of the plugin discussed in #7503. There are still some loose ends I would like some help on, but it is tested for the non-preprocessor scenario.

Closes #7503

# ↪️ Pull Request

Adds a transformer plugin for [svelte](https://svelte.dev) files. It optionally preprocesses the files using svelte's builtin mechanism for typescript, scss, etc. It then compiles the svelte file into a js and css asset. Preprocessing only occurs if `svelte.config.js` specifies a `preprocess` key whose value is truthy or if the `svelte-preprocess` package is importable.

## 🚨 Test instructions

An integration test has been added for the simple case of a svelte file. I'm a little at a loss for how to automate testing the typescript scenario since it would require additional packages (i.e.: `svelte-preprocess`) which I don't know how to add to the integration test.

Ideally, a repository like this [example](https://github.com/gcoakes/example-parcel-svelte-ts) would compile perfectly fine. It includes `svelte-preprocess` in its package.json, so it should convert the typescript script block from within svelte to a javascript asset.

## ✔️ PR Todo

### DIFF
--- a/packages/transformers/svelte/src/SvelteTransformer.js
+++ b/packages/transformers/svelte/src/SvelteTransformer.js
@@ -43,6 +43,9 @@ export default (new Transformer({
     let source = await asset.getCode();
     const filename = relativeUrl(options.projectRoot, asset.filePath);
 
+    // If the preprocessor config is never defined in the svelte config, attempt
+    // to import `svelte-preprocess`. If that is importable, use that to
+    // preprocess the file. Otherwise, do not run any preprocessors.
     if (preprocessConf === undefined) {
       let preprocessor = null;
       logger.verbose({
@@ -50,7 +53,7 @@ export default (new Transformer({
           'No preprocess specified; attempting to use `svelte-preprocess`.',
       });
       try {
-        preprocessor = await import('svelte-preprocess');
+        preprocessor = require('svelte-preprocess');
       } catch (e) {
         logger.verbose({
           message:
@@ -58,10 +61,11 @@ export default (new Transformer({
         });
       }
       if (preprocessor) {
-        preprocessConf = preprocessor.default();
+        preprocessConf = preprocessor();
       }
     }
 
+    // Only preprocess if there is a config for it.
     if (preprocessConf) {
       logger.verbose({message: 'Preprocessing svelte file.'});
       const processed = await catchDiag(
--- a/packages/transformers/svelte/src/SvelteTransformer.js
+++ b/packages/transformers/svelte/src/SvelteTransformer.js
@@ -2,8 +2,16 @@
 import {Transformer} from '@parcel/plugin';
 import SourceMap from '@parcel/source-map';
 import {relativeUrl} from '@parcel/utils';
-import {compile, preprocess} from 'svelte/compiler';
+import {compile, preprocess, walk} from 'svelte/compiler';
 import ThrowableDiagnostic from '@parcel/diagnostic';
+import {createMakeHot} from 'svelte-hmr';
+
+const makeHot = createMakeHot({
+  walk,
+  meta: 'module',
+  absoluteImports: false,
+  versionNonAbsoluteImports: false,
+});
 
 export default (new Transformer({
   async loadConfig({config, options, logger}) {
@@ -12,7 +20,13 @@ export default (new Transformer({
       '.svelterc.json',
       'svelte.config.js',
     ]);
-    if (!svelteConfig) return {};
+    if (!svelteConfig)
+      return {
+        compilerOptions: {
+          css: false,
+          dev: options.mode !== 'production',
+        },
+      };
     if (svelteConfig.filePath.endsWith('.js')) {
       // TODO: Is there a better way of handling this warning? Probably just
       // mention it in the documentation and silently invalidate.
@@ -36,7 +50,7 @@ export default (new Transformer({
 
   async transform({
     asset,
-    config: {preprocess: preprocessConf, compilerOptions},
+    config: {preprocess: preprocessConf, compilerOptions, hmrOptions},
     options,
     logger,
   }) {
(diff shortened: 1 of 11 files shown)


packages/transformers/svelte/src/SvelteTransformer.js
```diff
@@ -0,0 +1,147 @@
+import {Transformer} from '@parcel/plugin';
```
mischnic, line 2: This enable type checking with Flow for this file
```suggestion
// @flow
import {Transformer} from '@parcel/plugin';
```

packages/transformers/svelte/src/SvelteTransformer.js
```diff
@@ -0,0 +1,147 @@
...
+    const filename = relativeUrl(options.projectRoot, asset.filePath);
+
+    if (!config.preprocess) {
+      let preprocessor = null;
+      try {
+        // This seems sufficiently batteries included. Only apply
+        // svelte-preprocess if it is importable. Otherwise, assume the lack of
+        // a preprocess entry is intentional.
+        // TODO: Is it possible to utilize parcel's pipelines in lieu of
+        // svelte-preprocess?
```
mischnic: The Vue transformer uses named pipelines to first extract the individual assets from the input file and then process the parts again.

gcoakes: So, if I'm understanding the Vue transformer correctly, it transforms the vue code once to produce some "glue" code which imports the disparate named pipelines. Those named pipelines get processes by the Vue transformer separately and each get rendered into the glue code. My use case would be to transform the embed code as a callback within the preprocess function.

I don't think I want to take up that change in this PR. It's probably better to keep it simple. It seems pretty likely to go wrong with svelte, if I were to follow the same pattern as Vue by generating glue code. It would also be unexpected from a UX perspective.

devongovett: It might be weird if TypeScript/SASS/PostCSS behave differently in a Svelte file than outside...

We used to have a feature that allowed returning multiple assets for processing, and then accessing them in a `postProcess` method. It was originally designed for cases like this, but we removed it because no one was using it. #5674

packages/transformers/svelte/src/SvelteTransformer.js
```diff
@@ -0,0 +1,147 @@
...
+        // TODO: Call this out in the documenation.
+        logger.verbose({message: 'Attempting to use svelte-preprocess.'});
+        preprocessor = await import('svelte-preprocess');
```
devongovett: Note: using ESM currently means Parcel cannot invalidate the cache properly when this dependency changes.

packages/transformers/svelte/src/SvelteTransformer.js
gcoakes: I think it's actually the opposite. As it stands right now, preprocessing is optional and only occurs when the user has explicitly set the `preprocess` key in the `svelte.config.js` or `svelte-preprocess` is importable. In the first case, the user is fully in control of what and how the embed code gets processed. In the second case, `svelte-preprocess` [tells](https://github.com/sveltejs/svelte-preprocess/blob/main/docs/getting-started.md) the user to install the tsc compiler. Perhaps a good solution would be to document the inconsistency and prompt the user to install the tsc transformer if using svelte?

As to the `postProcess` function, I would love to use this if the project is willing to maintain the API and if it works for this use case. I don't think I can commit to that maintenance though as it would likely require deeper knowledge of the inner workings of parcel. I have concerns whether it would actually work for this situation though. Would I be able to fully run an embed pipeline to completion within the transform function? `svelte/compiler:preprocess` takes as input callbacks to preprocess the embed languages of a svelte file. The output of those callbacks need to be vanilla js, css, or markup. I imagine optimization would also impact that. I don't think I would be able to take the multi-pass approach that Vue took.

The most conservative solution would be for me to remove preprocessing altogether so there's no backwards incompatible changes in the defaults at any point. Though I think the current defaults requiring the user to opt into `svelte-preprocess` is sufficient.

packages/transformers/svelte/src/SvelteTransformer.js
gcoakes: Can I just `require('svelte-preprocess')` instead?

gcoakes: That was probably an unnecessary question... I fixed it in the gcoakes/parcel@d896bfd.

gcoakes: I'm still uncertain how to write an integration test for the negative case of this import. It seems like the test harness isn't designed for optional dependencies. I think it's fairly safe to leave untested though as long as the catch block remains minimal.

Shinyaigeek:
@gcoakes  Hi! 👋  Is there some reason to close this PR? I want to take over this great work if you like. If there is some reason to abort this work and you could tell me, I would appreciate it.

gcoakes:
> Is there some reason to close this PR?

My original intention was just to port over the svelte transformer which I already understood pretty well. However, in order to fit into Parcel's UX, a proper implementation appears to require utilizing parcel transformers for the embedded languages. That was a significant deviation from the original implementation. I wasn't particularly keen on making those changes.

PR #7537, opened 2022-01-07
Fix: load .env value also if entry target is specified with `source` value in package.json

### LINKED ISSUE #7515: Environment variables are not loaded from .env files when using "source" from package.json as parcel entry
# 🐛 bug report
When using `parcel` or `parcel build` and setting the `source` property in the package.json as the entry file, the `.env` files seem to be ignored as the environment variable values are undefined. However, if the entry is specified in the CLI command, the environment variables resolve correctly.

.env
```
TEST_THING=hello
```

package.json
```json
{
 ...
 "source": "src/index.tsx"
 ...
 "scripts": {
   "start": "parcel",
   "build": "parcel build"
 }
 ...
}
```

## 🤔 Expected Behavior

When running the above commands, accessing environment variables should resolve in the value from .env values. IE `process.env.TEST_THING` should return the value of `"hello"`.

## 😯 Current Behavior

When running the above commands, accessing any environment variables resolves with `undefined`.

However, if the the scripts are changed to specify the entry file in the CLI like so:
```json
{
...
"scripts": {
  "start": "parcel src/index.tsx",
  "build": "parcel build src/index.tsx"
}
...
}
```

The environment variables are resolved correctly.

## 🔦 Context

This is not that major of an issue as there is a workaround, but it is unexpected behavior that not everyone may be aware of

## 💻 Code Sample

See above.

## 🌍 Your Environment

| Parcel           | 2.0.1
| Node             | 16.13.1
| npm/Yarn         | yarn 1.22.10
| Operating System | macOS 12.0.1

Shinyaigeek:
# ↪️ Pull Request

Fixes: #7515 

Hi team :wave:

While I was investigating this issue, I noticed the cause of this issue.

When entry file is specified with `source` value in package.json ( and not specified with cli `build` target value), entries will be determined as `["."]` in https://github.com/parcel-bundler/parcel/blob/v2/packages/core/parcel/src/cli.js#L219-L221, and resolved into project root directory in https://github.com/parcel-bundler/parcel/blob/v2/packages/core/parcel/src/cli.js#L223.
These entry files are passed into `getRootDir` to detect the entry root directory. `.env` is loaded from this entry root directory. However, when the entry file is specified with `source` value in package.json ( and not specified with cli `build` target value), these entry files are resolved into the project root directory, so `getRootDir` returns the parent of the project root directory, so `.env` file cannot be loaded due to this wrong path.

I fixed this

## 💻 Examples

**package.json**
```json
{
 ...
 "source": "src/index.tsx"
 ...
 "scripts": {
   "start": "parcel",
   "build": "parcel build"
 }
 ...
}
```

## 🚨 Test instructions

I added the integration test to check this behavior

## ✔️ PR Todo


packages/core/integration-tests/test/javascript.js
```diff
@@ -2955,6 +2955,26 @@ describe('javascript', function () {
...
+        entries: [
+          path.join(__dirname, '/integration/env-file-with-package-source'),
+        ],
+      },
+    );
+
+    let output = await run(b);
+    assert.equal(output, 'bartest');
+  });
+
```
Shinyaigeek: This test case depends on the implementation of `cli.js`, so it feels muddy. If you know the better way to test this(such as invoking parcel-bundle via CLI instead of `parcel/test-utils/bundle`), please let me know

packages/core/integration-tests/test/javascript.js
```diff
@@ -2955,6 +2955,26 @@ describe('javascript', function () {
...
+      ),
+      {
+        // when the entry file is specified with a source value in package.json and not specified with cli build target value,
+        // the entry file is resolved to the entry root directory in packages/core/core/src/resolveOptions.js as InitialParcelOptions.
+        // So, entry root directory is specified as InitialParcelOptions' entries value.
+        entries: [
+          path.join(__dirname, '/integration/env-file-with-package-source'),
+        ],
+      },
+    );
```
mischnic: You don't need that entries setting if you just pass the directory as an entry directly instead of the index file.
```suggestion
    let b = await bundle(
      path.join(__dirname, '/integration/env-file-with-package-source'),
    );
```

packages/core/core/src/resolveOptions.js
```diff
@@ -49,7 +49,15 @@ export default async function resolveOptions(
     entries = [path.resolve(inputCwd, initialOptions.entries)];
   }
 
-  let entryRoot = getRootDir(entries);
+  let entryRoot = getRootDir(
+    entries.map(entry =>
+      // if the entry refers entryRootDir, getRootDir returns the parent directory of entryRootDir.
+      // In this case, each entry should refers some file, but without cli build target entry refers entryRootDir directory.
+      // So, we add /. to entry to make entry refers some (virtual) file in such case.
+      path.extname(entry) !== '' ? entry : entry + '/.',
```
mischnic: This would break if you have extensionless entries (which should work fine right now if there's a matching glob in parcelrc).
So maybe something like this instead?
```js
        (await inputFS.stat(entry)).isFile() ? entry : path.join(entry, 'index'),
```
(and use `path.join(entry, 'index')` instead of `./`. This pattern is used in various places, for example also when determining the projectRoot a few lines further down)

packages/core/integration-tests/test/javascript.js
Shinyaigeek: I see, I did not know that I can set a directory as entries when I pass a directory as an entry directly. Thank you for helping!

packages/core/core/src/resolveOptions.js
Shinyaigeek: Thank you! As you said, it is better to use `stat` than to judge with extension. I fixed in [4b2d4a6](https://github.com/parcel-bundler/parcel/pull/7537/commits/4b2d4a6934967dfe511e64fed412e97c073600d8) 👍

packages/core/core/src/resolveOptions.js
```diff
@@ -50,7 +50,19 @@ export default async function resolveOptions(
     entries = [path.resolve(inputCwd, initialOptions.entries)];
   }
 
-  let entryRoot = getRootDir(entries);
+  let entryRoot = getRootDir(
+    await Promise.all(
+      entries.map(async entry =>
+        // if the entry refers entryRootDir, getRootDir returns the parent directory of entryRootDir.
+        // In this case, each entry should refers some file, but without cli build target entry refers entryRootDir directory.
+        // So, we resolve with index as entry to make entry refers some (virtual) file in such case.
```
mischnic:
```suggestion
        // getRootDir treats the input as files, so getRootDir(["/home/user/myproject"]) returns "/home/user".
        // Instead we need to make the entries refer to some file inside the specified folders.
```
I was having trouble understanding your comment 😄

mischnic (approved):

packages/core/core/src/resolveOptions.js
Shinyaigeek: Looks good ☺️ The comment you changed is easier to understand! Thank you!

mischnic:
**See my next comment**
~Actually, these `entries` can be entries so we definitely need something like~
```js
      entries.map(async entry => {
        // getRootDir treats the input as files, so getRootDir(["/home/user/myproject"]) returns "/home/user".
        // Instead we need to make the entries refer to some file inside the specified folders.
        let isFolder = false;
        try {
          isFolder = (await inputFS.stat(entry)).isDirectory()
        } catch(e) {
          // Ignore error if entry is a glob or doesn't exist, assume it's a file
        }

        return isFolder
          ? path.join(entry, 'index')
          : entry;
      })
```

And for reference, some cases where something `extname`-based doesn't work correctly:
```js
path.extname("home/my-project.new/") === '.new'
path.extname("project/*") === ''
```

devongovett:
I'm not sure assuming a glob is a file is right either though. If we want to be 100% correct, we'd need to expand the globs first. But that is expensive, and already done in the entry request. This is just duplicating all that again. Ideally we'd re-order this so the root dir gets resolved after entries are resolved. Not sure that's possible though. For finding the project root, a best guess at a starting point to search upward for `yarn.lock` etc. is probably fine.

devongovett:
I'm not sure assuming a glob is a file is right either though. If we want to be 100% correct, we'd need to expand the globs first. But that is expensive, and already done in the entry request. This is just duplicating all that again. Ideally we'd re-order this so the root dir gets resolved after entries are resolved. Not sure that's possible though. For finding the project root, a best guess at a starting point to search upward for `yarn.lock` etc. is probably fine.

mischnic:
We only need to do one stat call because this problematic behaviour of getRootDir only happens if there is exactly one entrypoint and that entrypoint is the project root folder (and not a glob).

```patch
diff --git packages/core/core/src/resolveOptions.js packages/core/core/src/resolveOptions.js
index 42c529792..c26d79b88 100644
--- packages/core/core/src/resolveOptions.js
+++ packages/core/core/src/resolveOptions.js
@@ -14,7 +14,7 @@ import {hashString} from '@parcel/hash';
 import {NodeFS} from '@parcel/fs';
 import {LMDBCache, FSCache} from '@parcel/cache';
 import {NodePackageManager} from '@parcel/package-manager';
-import {getRootDir, relativePath, resolveConfig} from '@parcel/utils';
+import {isGlob, getRootDir, relativePath, resolveConfig} from '@parcel/utils';
 import loadDotEnv from './loadDotEnv';
 import {toProjectPath} from './projectPath';
 import {getResolveFrom} from './requests/ParcelConfigRequest';
@@ -50,6 +50,19 @@ export default async function resolveOptions(
     entries = [path.resolve(inputCwd, initialOptions.entries)];
   }

+  if (entries.length === 1 && !isGlob(entries[0])) {
+    let [entry] = entries;
+    let isDirectory = false;
+    try {
+      isDirectory = (await inputFS.stat(entry)).isDirectory();
+    } catch (e) {
+      // ignore failing stat call
+    }
+    if (isDirectory) {
+      entries[0] = path.join(entry, 'index');
+    }
+  }
+
   let entryRoot = getRootDir(entries);
   let projectRootFile =
     (await resolveConfig(
```
This prevents unnecessarily many stat calls and works correctly with globs.
Cases for which this approach works correctly:
```
parcel '/Users/niklas/my-project/'
parcel '/Users/niklas/my-project/*'
parcel '/Users/niklas/my-project/packages'
parcel '/Users/niklas/my-project/package.json'
parcel '/Users/niklas/my-project/packages/*'
parcel '/Users/niklas/my-project/packages/**/*.md'
```

Shinyaigeek:
yes, as @mischnic said (and be described in this PR's description), this issue can happen only if the one entry point(and this entry point can not be glob). I think the proposed way is good because this prevents unnecessary stat calls.

mischnic:
@Shinyaigeek Could you change your implementation to use that approach instead?

Shinyaigeek:
@mischnic [7c9d5a8](https://github.com/parcel-bundler/parcel/pull/7537/commits/7c9d5a856d7f1a58db35dd51e20e0311e1f72109) DONE 👍  

While I changed the code, I noticed that adding the side-effects to `entries` will cause a reference error after `getRootDir` process because arranged entries refer to a non-existing file. 

This process is only for `getRootDir`, so I thought about adding this feature to `getRootDir`, but this process is needed only when solving entries given by package.json or CLI, so this feature should not be included in `getRootDir`, parcel’s utility module. so I added a process to change the arguments passed into `getRootDir` (and not change `entries` itself).

Shinyaigeek:
~~By the way, if such a `/index` needs to be added to entries, the value returned by `getRootDir` should be the one before the `/index` is added ( ["/home/myproject"] -> ["/home/myproject/index"] -> "home/myproject"), then we can just return `entries[0]` instead of invoking `getRootDir`.~~

~~This is more intuitive (and I can determine a more suitable variable name), so I changed it that way in [c90d470](https://github.com/parcel-bundler/parcel/pull/7537/commits/c90d4707a1e9de5d77af6d7bc319bd6450f19acf).~~

I noticed this does not work fine in the case `["/home/myproject/packages"]` so I revert this

packages/core/core/src/resolveOptions.js
```diff
@@ -50,7 +50,22 @@ export default async function resolveOptions(
     entries = [path.resolve(inputCwd, initialOptions.entries)];
   }
 
-  let entryRoot = getRootDir(entries);
+  let shouldMakeEntryReferFolder = false;
+  // if entries have the only one entry and the one is not glob and refers to directory, we need to make the entry refer to some file.
```
mischnic:
```suggestion
```

There's already a comment further down and also this one essentially rehashes the condition of the if.

mischnic (approved):
Thanks! I forgot to check whether `entries` was also used in other places 👍

PR #7566, opened 2022-01-17
Update pnpm to allow installing dependencies in a monorepo

PetterRuud:
# ↪️ Pull Request
As with Yarn 2, PNPM needs the -W parameter to install packages in root in a monorepo.


devongovett (approved):
Thanks!

PR #7564, opened 2022-01-17
CSS fixes and dependency upgrades

devongovett:
* Fixes #7556. This was caused by a change I made to make the old CSS transformer in line with `@parcel/css`, but it resulted in a bug. Now we preserve the quotes, and I added them to Parcel CSS as well to match. That way we don't need to deal with escaping. That will be handled by the minifier if needed.
* Fixes https://github.com/parcel-bundler/parcel-css/issues/56. Back in #4838 we removed the default browser targets for perf reasons with Babel at the time. But now that we are on SWC it shouldn't make much difference if we default to the latest browser versions. That way features that aren't supported anywhere yet like CSS nesting will get compiled.

Also updates a bunch of dependencies to fix security warnings and reduce the number of duplicates we have:

* Fixes #7543. This bumps htmlnano to 2.0.0, which removes uncss. That results in many very old dependencies being removed, such as jsdom and old postcss versions.
* Removes css-module-loader-core which is unmaintained, and switches to the copy inside postcss-modules. Also upgrades postcss-modules. This removes all old versions of PostCSS from the lock file.
* Bumps the Vue compiler to the latest as well, to get rid of old PostCSS dep.
* Bumps SVGR to v6
* Normalizes version of JSON5 to v2
* http-proxy-middleware v2
* And several other random bumps noted in the security audit

Hopefully these solve the security audit issues for now, and starts to reduce the number of dependencies we have.

PR #6958, opened 2021-09-20
Update Micromatch

NovaAtWarren:
# ↪️ Pull Request

Updates micromatch to remove security vulnerability, as described in #6942.
EDIT: probably closes #6942 

## ✔️ PR Todo

- [x] Included links to related issues/PRs

PR #6155, opened 2021-04-19
Add dependency validation script

lettertwo:
# ↪️ Pull Request

This PR adds a script (`yarn lint:dependencies`) that checks for the following:

- That packages have compatible semver expressions (see [Version mismatch](#version-mismatch) example below)
  
- That dependencies of packages have compatible semver expressions (see [Dependency mismatch](#dependency-mismatch) example below)

The intent of these checks is to avoid unintentionally installing duplicate incompatible versions of packages or dependencies (both in the repository and in Parcel projects).

### Version mismatch:
  
```shell
PackageMismatchError: @parcel/config-default has a parcel dependency "@parcel/optimizer-data-url@2.0.0-beta.1"
that is incompatible with the current version "@parcel/optimizer-data-url@2.0.0-beta.2".
```

This error is meant to catch the case where a Parcel package lists another Parcel package dependency with a semver expression _that doesn't match_ the current version.

### Dependency mismatch:
  
```shell
DependencyMismatchError: @parcel/utils has a dependency "json5@^1.0.1"
that is incompatible with a dependency "json5@^2.1.0" from @parcel/transformer-json
```

This error is meant to catch the case where a package lists a dependency with a semver expression that _does not overlap_ with one or more other packages' semver expressions for the same dependency.

### Ignoring errors via excludes

Packages can be ignored by adding their names to the [PACKAGE_EXCLUDES](https://github.com/parcel-bundler/parcel/compare/lettertwo/dependency-validation?expand=1#diff-24b399476d1ee97b815fb69ca60db72741f4df85c56800d6cc801845db651b80R8) set.

Dependencies can be ignored by adding their names to the [DEPENDENCY_EXCLUDES](https://github.com/parcel-bundler/parcel/compare/lettertwo/dependency-validation?expand=1#diff-24b399476d1ee97b815fb69ca60db72741f4df85c56800d6cc801845db651b80R11) set.

## 🚨 Test instructions

Script can be run via `yarn lint:dependencies`

## ✔️ PR Todo

- [x] Resolve current errors
   <details>
     <summary>Found 39 errors.</summary>

    ```shell
    PackageMismatchError: @parcel/config-default has a parcel dependency "@parcel/optimizer-data-url@2.0.0-beta.1" that is incompatible with the current version "@parcel/optimizer-data-url@2.0.0-beta.2".

[...]

PR #7576, opened 2022-01-19
Reduce the number of npm dependencies needed by parcel

devongovett:
Previously, installing the `parcel` package downloaded 493 dependencies from npm. This was made better from the over 600 deps we used to have in #7564, but is still pretty ridiculous. The aim of this PR is to reduce this further. This is accomplished in a few ways:

* Pre-bundling some dependencies with Parcel itself rather than loading them from npm. This is done mainly for small utility dependencies, not for things that applications would normally need to control the versions of. Specifically, for the following packages: `@parcel/fs`, `@parcel/package-manager`, `@parcel/utils`, `@parcel/codeframe`, `@parcel/reporter-cli`, and `@parcel/reporter-dev-server`. This reduces dependencies to 334. It also has the benefit that any random security warnings produced by npm will only affect the Parcel team and not downstream users. Most of the security warnings we receive are not actually exploitable in Parcel anyway, so this really is just about reducing annoyance for everyone. Both Next.js and Vite have taken a similar approach.
* Auto installing node builtin polyfills on demand (e.g. `buffer`, `stream`, etc.). The `@parcel/node-libs-browser` package pulled in over 100 dependencies, which were pretty rarely used these days. Now you'll see a warning when one is encountered, and it'll be auto installed into your project. In combination with the above, this reduces dependencies to 226.
* Removing built-in Babel and PostCSS dependencies, and installing them into projects on demand, only when actually used. This includes builtin flow and legacy css modules support. In combination with the above, this reduces dependencies to 183.
* Removing CSS nano once we switch entirely to `@parcel/css` will reduce this further to 150.

There are other potential optimizations we can do in the future, but this already helps a lot, reducing dependencies by ~70%. It should result in faster install times, and less dependencies to audit for everyone using Parcel.


packages/core/core/src/ParcelConfig.js
```diff
@@ -29,7 +29,7 @@ import ThrowableDiagnostic, {
 } from '@parcel/diagnostic';
 import json5 from 'json5';
 
-import {makeRe} from 'micromatch';
+import {globToRegex} from '@parcel/utils';
```
devongovett, line 32: Centralized all uses of micromatch into the `@parcel/utils` package so we only get one copy

packages/core/parcel/src/bin.js
```diff
@@ -2,7 +2,10 @@
 
 'use strict';
 
-if (process.env.PARCEL_BUILD_ENV !== 'production') {
+if (
+  process.env.PARCEL_BUILD_ENV !== 'production' ||
+  process.env.PARCEL_SELF_BUILD
```
devongovett, line 7: This new environment variable is needed so we can replace PARCEL_BUILD_ENV while parcel is building itself...

packages/dev/babel-preset/index.js
```diff
@@ -1,36 +1,63 @@
...
-      {
-        lazy: true,
-      },
-    ],
-  ],
-  env: {
-    production: {
+module.exports = api => {
+  let name = api.caller(caller => caller && caller.name);
+  if (name === 'parcel') {
```
devongovett, line 3: A different Babel config is used when Parcel is building itself since we don't want to compile modules etc.

packages/transformers/postcss/src/PostCSSTransformer.js
```diff
@@ -79,10 +77,22 @@ export default (new Transformer({
     let cssModules: ?{|[string]: string|} = null;
     if (config.hydrated.modules) {
       asset.meta.cssModulesCompiled = true;
+
+      // TODO: should this be resolved from the project root?
```
devongovett, line 81: actually a question for all dev deps. What if there were a CSS module in node_modules? we would never want to install dev deps there...

packages/utils/node-resolver-core/src/NodeResolver.js
```diff
@@ -602,7 +653,16 @@ export default class NodeResolver {
       if (isExplicitNode) {
         filename = filename.substr(5);
       }
-      return {filePath: builtins[filename] || empty};
+
+      // By default, exclude node builtins from libraries unless explicitly opted in.
```
devongovett, line 657: Is this a good idea? Alternatively we could have some separate config for it, but it seems like when building libraries you'd usually want to exclude these...

packages/core/core/src/ParcelConfig.js
lettertwo: I recall that there was some breaking change in micromatch that thwarted a previous attempt to consolidate it (see [this comment](https://github.com/parcel-bundler/parcel/pull/6155#issuecomment-822626449) from @mischnic on a PR I forgot about 😨 ). Not sure if it's still relevant?

packages/utils/node-resolver-core/src/NodeResolver.js
Shinyaigeek: Since it is ideal to resolve node builtins polyfill with the bundler of the library user, I feel we do not want to include node builtins if a bundling target is a library working in the browser. On the other hand, I think some library author wants to host their library on `esm.sh` etc. and use it as pure ESModules from the browser. In that case, the bundle of the library should include node builtins.


So, I think it is good to make it the default behavior, but I think also that an emergency hatch should be prepared as a config or flag for such use cases. (it would be simple to judge that with whether or not the output is `esmodule`, but in reality, it shouldn't be so difficult ….)

I'm sorry if I overlooked the context and said something strange

packages/utils/node-resolver-core/src/NodeResolver.js
```diff
@@ -273,6 +288,44 @@ export default class NodeResolver {
...
+                : [],
+            },
+          ],
+          documentationURL:
+            'https://parceljs.org/features/node-emulation/#polyfilling-%26-excluding-builtin-node-modules',
+        });
+
+        await packageManager.resolve(builtin, this.projectRoot + '/index', {
+          saveDev: true,
+          shouldAutoInstall: true,
```
mischnic, line 317: Looks like this would override `--no-autoinstall` which should not be the case?

devongovett: The package manager is not passed in as an option from the DefaultResolver if shouldAutoInstall is false

Edit: should not be. I thought I did this but apparently not.

packages/core/core/src/ParcelConfig.js
mischnic: At least one occurence was fixed in https://github.com/parcel-bundler/parcel/pull/6958

packages/utils/node-resolver-core/test/resolver.js
```diff
@@ -297,8 +297,8 @@ describe('resolver', function () {
             aboveFilePath: path.join(rootDir, 'foo.js'),
           },
           {
-            fileName: 'package.json',
-            aboveFilePath: require.resolve('browserify-zlib'),
+            fileName: 'node_modules/browserify-zlib',
```
mischnic, line 300: I'm surprised that `invalidateOnFileCreate` can be used to watch for folder creations...

xvello:
Thanks for this @devongovett, do you have plans to cut a release including these changes soon?

devongovett:
@xvello i hope so. We just want to do some testing to ensure nothing is broken. If you want, you could try out the nightly release and report back if it works for you?

xvello:
@devongovett I tried upgrading my [simple project](https://github.com/xvello/letsblockit/blob/main/src/assets/package.json) to the nightlies, and the generated output is the same, so good enough for me I guess? ;)

> npm install --dev parcel@2.0.0-nightly.991
> npm install -dev @parcel/transformer-sass@2.0.0-nightly.993+90050e4e

I reached out to know whether I should wait for an upcoming release, or commit these updates.

PR #7600, opened 2022-01-23
simplify macros

danieleades:
simplify some test macros slightly. kinda makes you wonder whether they need to exist at all...


danieleades:
@devongovett is this PR not of any interest? Personally I would prefer to remove these macros entirely. Not sure they're adding a lot of value, just needless indirection

PR #7672, opened 2022-02-08
fix(transformers): errors.map is not a function

sweetliquid:
# ↪️ Pull Request

## ✔️ PR Todo


Shinyaigeek:
https://github.com/parcel-bundler/parcel/blob/v2/packages/transformers/html/src/dependencies.js#L117

The errors `collectDependencies` throw seems to be only array, so errors.map should be a valid function. Is there some case that parcel throws “errors.map is not a function” error in HTMLTransformer? ( if there is, please share reproduction)

sweetliquid:
The error is thrown by other functions called by collectDependencies like `QUERIES[i].select` in browserslist package
```js
throw new BrowserslistError(
  'Unknown version ' + version + ' of ' + name
)
```

sweetliquid:
I think errors those are not directly thrown by `collectDependencies` can be ignored(re-thown).

sweetliquid:
If I create the `.browserslistrc` file with any wrong content like: `xxx`, it thrown `emit error:  TypeError: errors.map is not a function` when I run parcel build

Shinyaigeek:
Oh I see, I cannot reproduce it but I feel this can happen and this change is reasonable. Thank you !

devongovett (approved):

PR #7688, opened 2022-02-10
chore: Update CONTRIBUTING

LekoArts:
# ↪️ Pull Request

I've been following the guide and was hit by not seeing local changes. I used the `ts-example` as a template which unfortunately didn't have the `--no-cache` flag. So I'm update the guide here + the example to save future people from that struggle.


wbinnssmith:
Can you describe the steps you took to not have changes reflected? That seems like the real issue in need of a solution that's not running without cache.

LekoArts:
I was editing Rust code in the JS transformer and building with napi-rs my changes. Then when I tried out a build in the `ts-example` I didn't see my debug logging I've added. @mischnic then told me that I should use the `--no-cache` flag that is already the default on the other examples. I just had bad luck and chose the one without that :D

mischnic:
I have also had the impression in the past that native Node modules aren't properly being invalidated on changes, but in this specific case it would never work because the project root is `packages/examples` and the transformer is outside of that. So `--no-cache` is needed anyway?

mischnic (approved):

PR #7694, opened 2022-02-10
Upgrade to version 2 of detect-libc

lovell:
# ↪️ Pull Request

Hi, the latest `detect-libc` has been modernised to use the non-blocking Node.js Report API where available (Node.js >= 12), falling back to a more efficient single child process only when needed.

It's a major version bump for `detect-libc` as it drops support for versions of Node.js < 8 but `parcel` already requires >= 12 so this change is OK to include in a patch release.

https://www.npmjs.com/package/detect-libc

## 💻 Examples

Should anyone be interested in the background to this, please see lovell/detect-libc#14

## 🚨 Test instructions

Hopefully existing unit tests in this repo cover its use.

`detect-libc` has an extensive integration test suite, e.g. https://github.com/lovell/detect-libc/actions/runs/1717414863

## ✔️ PR Todo

### DIFF
--- a/packages/optimizers/image/src/ImageOptimizer.js
+++ b/packages/optimizers/image/src/ImageOptimizer.js
@@ -48,8 +48,8 @@ async function loadOnMainThreadIfNeeded() {
     process.platform === 'linux' &&
     WorkerFarm.isWorker()
   ) {
-    let {family, version} = require('detect-libc');
-    if (family === 'glibc' && parseFloat(version) <= 2.17) {
+    let {GLIBC, family, version} = require('detect-libc');
+    if ((await family()) === GLIBC && parseFloat(await version()) <= 2.17) {
       let api = WorkerFarm.getWorkerApi();
       await api.callMaster({
         location: __dirname + '/loadNative.js',
--- a/packages/optimizers/image/native.js
+++ b/packages/optimizers/image/native.js
@@ -1,7 +1,7 @@
 let parts = [process.platform, process.arch];
 if (process.platform === 'linux') {
-  const {MUSL, family} = require('detect-libc');
-  if (family === MUSL) {
+  const {MUSL, familySync} = require('detect-libc');
+  if (familySync() === MUSL) {
     parts.push('musl');
   } else if (process.arch === 'arm') {
     parts.push('gnueabihf');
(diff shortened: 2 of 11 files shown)


packages/optimizers/image/src/ImageOptimizer.js
```diff
@@ -48,8 +48,8 @@ async function loadOnMainThreadIfNeeded() {
     process.platform === 'linux' &&
     WorkerFarm.isWorker()
   ) {
-    let {family, version} = require('detect-libc');
-    if (family === 'glibc' && parseFloat(version) <= 2.17) {
+    let {GLIBC, family, version} = require('detect-libc');
+    if ((await family()) === GLIBC && parseFloat(await version()) <= 2.17) {
```
wbinnssmith, line 52: nit: consider

```js
let [_family, _version] = await Promise.all([family(), version()]);
// ...
```

wbinnssmith:
> Hopefully existing unit tests in this repo cover its use.

We don't run on old Centos to test the segfault issue with the ImageOptimizer, so I don't think that codepath is taken. Probably worth a manual test to confirm it functions as before.

@devongovett once we get CI stable, it may be worth considering running the integration tests on a system with old glibc too.

packages/optimizers/image/src/ImageOptimizer.js
lovell: Thanks for the review. For those on Node.js 12+, which I think will be all parcel users, the call to `family()` is very fast but the call to `version()` will still involve spawning a child process, a small I/O overhead. Keeping these separate allows the call to the former to possibly shortcut the need to call the latter. That's bit of a micro-optimisation though, so I'm happy to switch to your suggestion for more readable code if you'd prefer.

wbinnssmith: 👍 makes sense

lovell:
I've rebased to remove the conflict and I think the one open conversation can now be resolved. Does the "Waiting" status mean you're waiting for me to do something? Please do let me know if there's anything else that's required, thank you.

lovell:
Closing as no longer required since https://github.com/parcel-bundler/parcel/pull/8838

PR #7718, opened 2022-02-15
Fix: escape double quote of url value in CSS `url()`

### LINKED ISSUE #7710: Broken url strings in CSS
# 🐛 bug report

Since @parcel/transformer-css-~~2.2.0~~ 2.2.1 single quotes in CSS `url()` seem to be replaced with double quotes. This breaks url strings that contain double quotes.

No relevant config.

## 🤔 Expected Behavior

I’m expecting the following line to stay as is in the output CSS:

```css
background-image: url('data:image/svg+xml;utf8,<svg xmlns="http://www.w3.org/2000/svg" version="1.1" width="32" height="24" viewBox="0 0 32 24"><polygon points="0,0 32,0 16,24" style="fill: rgb%2819, 80, 91%29"></polygon></svg>');`
```

## 😯 Current Behavior

From ~~2.2.0~~ 2.2.1 on `'` is changed to `"`, resulting in invalid CSS:

```css
background-image: url("data:image/svg+xml;utf8,<svg xmlns="http://www.w3.org/2000/svg" version="1.1" width="32" height="24" viewBox="0 0 32 24"><polygon points="0,0 32,0 16,24" style="fill: rgb%2819, 80, 91%29"></polygon></svg>");`
```

## 💁 Possible Solution

Is there some config that would allow me turn this off again? I've experimented with transformer settings but no luck.

For now, pinning @parcel/transformer-css to 2.1.1 is a workaround:

```json
  "resolutions": {
    "@parcel/transformer-css": "2.1.1"
  },
```

## 🔦 Context

There are no build errors but the resulting CSS is invalid and, in the above example, wouldn’t show the svg.

I cannot change the affected url string because it comes from a 3rd party dependency ([foundation-sites](https://github.com/foundation/foundation-sites/blob/master/scss/util/_mixins.scss#L151)).

## 💻 Code Sample

See example above.

## 🌍 Your Environment

| Parcel           | 2.3.1 |
| Node             | 14.17.4 |
| Yarn         | 1.22.17 |
| Operating System | Ubuntu 21.10 |

Shinyaigeek:
Fixes: #7710 

# ↪️ Pull Request

Hi team! 👋 

In [css transforming](https://github.com/parcel-bundler/parcel/blob/v2/packages/transformers/css/src/CSSTransformer.js#L187), `CSSTransformer` encloses URL value in CSS `url()` with `"` double quote, so I fixed this simply by escaping `”` double quote in url value in `url()` , but honestly I'm worried that this way is too straightforward and I overlooked some context.

If there is something I missed, please be free to point out or close this PR 🙇 

## 💻 Examples

```html
<!DOCTYPE html>
<html>
  <head>
    <meta charset="utf-8" />
    <title>CSS Double</title>
  </head>
  <body>
    <h1>CSS Double</h1>
    <div class="background"></div>
    <style>
      .background {
        background-image: url('data:image/svg+xml;utf8,<svg xmlns="http://www.w3.org/2000/svg" version="1.1" width="32" height="24" viewBox="0 0 32 24"><polygon points="0,0 32,0 16,24" style="fill: rgb%2819, 80, 91%29"></polygon></svg>');
        width: 100px;
        height: 100px;
      }
    </style>
  </body>
</html>

```

## 🚨 Test instructions

simply, I checked with the case base64 encoded url value with `"`

## ✔️ PR Todo


wbinnssmith (approved):

packages/transformers/css-experimental/src/CSSTransformer.js
```diff
@@ -83,7 +83,8 @@ export default (new Transformer({
             symbols: new Map([['*', {local: '*', isWeak: true, loc}]]),
           });
         } else if (dep.type === 'url') {
-          asset.addURLDependency(dep.url, {
+          // if dependency imported with url() includes ", escape " double quotes because this url value will be enclosed with double quote "
+          asset.addURLDependency(dep.url.replace(/"/g, '\\"'), {
```
devongovett: hmm I think this might cause an issue for resolvers because they won't expect pre-escaped values. Perhaps this replacement should be done in the packager. The util [here](https://github.com/parcel-bundler/parcel/blob/9775b9fab1b29d26c655de2aa84d3edc486ca578/packages/core/utils/src/replaceBundleReferences.js#L31) is what does the replacement of placeholders to resolved urls. It might need an option to escape the url.

Shinyaigeek: Oh, indeed. As you say, escaping the background image’s URL in Transformer will cause the not found error in resolver. I will fix this :+1: Thank you for your advice!

I think escaping the quote in URL is needed in anyway because
- if URL is not enclosed with quotes, it is not a valid property that the quote exists in the URL
- if URL is enclosed with quotes, the quote should be escaped

https://github.com/parcel-bundler/parcel/pull/7564 as you can see in this PR, I think we need to enclose the URL in quotes and then escape the quotes.

Shinyaigeek: I was thinking about this issue, and I notice that parcel's resolver cannot resolve the filepath of the image where the quotes are escaped though such a URL is the correct URL in the first place. So I think escape should be done with CSSTransformer and the resolver should be modified.

Shinyaigeek: I thought about making the resolver able to handle escaped quotes, but I was concerned about performance degradation due to increased computation, so I made a change to remove escaped quotes from URLs with the CSSTransformer and re-escape them with the CSSPackager. [ea56294](https://github.com/parcel-bundler/parcel/pull/7718/commits/ea56294c0c1f7ab09e416276061e3dd993708c29)

packages/transformers/css/src/CSSTransformer.js
```diff
@@ -83,12 +83,16 @@ export default (new Transformer({
             symbols: new Map([['*', {local: '*', isWeak: true, loc}]]),
           });
         } else if (dep.type === 'url') {
-          asset.addURLDependency(dep.url, {
-            loc,
-            meta: {
-              placeholder: dep.placeholder,
+          asset.addURLDependency(
+            // when URL is escaped, parcel resolver cannot resolve URL, so we need to unescape it in transformer
+            dep.url.replace(/\\"/g, '"').replace(/\\'/g, "'"),
```
devongovett: I don't think this is necessary. The url will already be unescaped before it is passed to us.

devongovett:
Thanks for getting this started @Shinyaigeek! I refactored it a little so that we handle the escaping differently depending on the asset type. For example, in HTML, quotes become `&quot;`. Also supported escaping `\` itself in CSS. Hopefully the changes make sense.

packages/core/utils/src/replaceBundleReferences.js
```diff
@@ -33,13 +33,15 @@ export function replaceURLReferences({
...
   map,
+  getReplacement,
   relative = true,
 }: {|
   bundle: NamedBundle,
   bundleGraph: BundleGraph<NamedBundle>,
   contents: string,
   relative?: boolean,
   map?: ?SourceMap,
+  getReplacement: string => string,
```
mischnic: This is technically a breaking change, though the question is how many plugins are out there using this function.

Shinyaigeek:
Thank you for your help @devongovett  !! I saw your commit, and feels reasonable. Sure, escaping only CSS text was short-circuited and should have been able to escape other files in the parcel pipeline! I learned a lot from that 👍

Issue #7579, opened 2022-01-20
incorrect `projectRoot` when using parcel in monorepo


oe:
# 🐛 bug report

I'm using lerna with yarn workspace just like you guys do.
When I run parcel within a sub package, it can't read the `.env` `.proxyrc.js` files inside the package, instead it reads them from the root  of repo.

After digging into the code, turns out that `projectRoot` was set incorrectly, and there is no way to specify its value. Here is the code in @parcel/core 

https://github.com/parcel-bundler/parcel/blob/v2/packages/core/core/src/resolveOptions.js#L52-L61

## 🎛 Configuration

with all  default Configurations

## 🤔 Expected Behavior

parcel can reads configurations  inside the sub package

## 😯 Current Behavior

parcel  reads configurations  from the root of repo

## 🔦 Context
possible solutions:
1. change parcel behavior in monorepo
2. add an option `--project-root`( `projectRoot` in API) to allow user to specify `projectRoot`

## 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Parcel           | 2.2.1
| Node             | v16.9.1
| npm/Yarn         | yarn 1.22.11 / npm 7.21.1
| Operating System | macOS 12.1

steve-taylor:
The documentation on monorepos is quite thin. It suggests that the only supported way to use Parcel in a monorepo is by running commands against the root. It would be nice to have some clarity around whether running Parcel inside individual packages is supported at all.

chasewalden:
Agreed on adding or expanding on the monorepo documentation. It very briefly mentions that parcel ***can*** be run from the root by passing a glob, but not that it ***must***. 

If parcel is intended to be run from the root only, it should be mentioned somewhere. I hope this isn’t the case, however, as this effectively forces users to opt-out of other tools like Lerna or Turborepo, or even PNPM’s built in multi/recursive feature. 

For context: I’m working on converting one of my projects to a monorepo. Currently frontend and backend are in separate repos. The frontend is built with parcel, but backend is just tsc and turborepo. The apis are tRPC, so the frontend is devDependant on the backend. Both have some pipeline steps for build, lint, unit test, etc. There is also a third repo with some playwright e2e tests. 

At the moment, I can’t get pnpm, turborepo and parcel to play nicely this way. Maybe it’s actually a pnpm issue and yarn would work better, but turbo only supports yarn v1.

jpapini:
[...]

PR #7727, opened 2022-02-17
feat(@parcel/transformer-js): Rewrite `__dirname`/`__filename` to be relative to `asset.filePath`

### LINKED ISSUE #7611: Rewrite `__dirname`/`__filename` to be relative from `asset.filePath`
# 🙋 feature request

This is a more detailed report of https://github.com/parcel-bundler/parcel/issues/5090 and one that I put up after following some private conversions with the Parcel Core Team to discuss the specifics in the open.

This feature (request) addressed one bulletpoint here: https://github.com/parcel-bundler/parcel/issues/2493

> Rewrite `__dirname` and `__filename`. Currently we don't do this for Node targets, but we probably should so that they refer to the source file rather than the dist.

## 🤔 Expected Behavior

`file.ts`:

```ts
console.log(`${__dirname}/foo/bar`)
```

Gets compiled with Parcel to JS into the directory `.cache/compiled`.

`.cache/compiled/file.js`:

```js
console.log(`${__dirname + "/../.."}/foo/bar`)
```

## 😯 Current Behavior

`file.ts`:

```ts
console.log(`${__dirname}/foo/bar`)
```

`.cache/compiled/file.js`:

```js
console.log(`${__dirname}/foo/bar`)
// Which will result in `.cache/compiled/foo/bar`
```

## 💁 Possible Solution

First step would be to set a default for `node` target. That could/would be:
Treat it as relative from `options.projectRoot` (what is shown in the expected behavior example).

So for example: If `outputDir` is `.cache/compiled` and current `cwd()`/projectRoot is `/` the relative path to it would be `../../`

In a next step options similar to https://webpack.js.org/configuration/node/#node__dirname could be added.

## 🔦 Context

This might also impact https://github.com/parcel-bundler/parcel/issues/6925 maybe?

## Example Use Case

`/usr/project/file.ts`:

```ts
console.log(`${__dirname}/foo/bar`)
```

`/usr/project/.cache/compiled/file.js`:

```js
console.log(`../../foo/bar`)
```

`/usr/project` is then the projectRoot because the rest of the scripts runs from this root directory and then the `file.js` is accessed.

LekoArts:
# ↪️ Pull Request

This adds a way of accomplishing https://github.com/parcel-bundler/parcel/issues/7611 - this functionality will be enabled by default in node environment.

Fixes https://github.com/parcel-bundler/parcel/issues/7611

## 💻 Examples

See linked issue, but TL;DR is:

Some code in e.g. `<root>/index.ts` has a `${__dirname}/data` string in it.
When it'll be compiled to e.g. `<root>/dist` the path should be changed to:

```js
const path = require("path")
const dirname = path.resolve(__dirname, "..")
return `${dirname}/data`
```

## 🚨 Test instructions

I've added a test to the integration tests, it has to have the node env (so e.g. `engines: node` key).

## ✔️ PR Todo


packages/core/core/src/resolveOptions.js
```diff
@@ -50,13 +50,24 @@ export default async function resolveOptions(
...
   let projectRootFile =
     (await resolveConfig(
       inputFS,
       path.join(entryRoot, 'index'),
       [...LOCK_FILE_NAMES, '.git', '.hg'],
       path.parse(entryRoot).root,
     )) || path.join(inputCwd, 'index'); // ? Should this just be rootDir
+  */
+
+  let projectRootFile =
```
LekoArts: This was necessary in local testing because of https://github.com/parcel-bundler/parcel/issues/7579

packages/transformers/js/src/JSTransformer.js
```diff
@@ -392,6 +399,7 @@ export default (new Transformer({
       replace_env: !asset.env.isNode(),
       inline_fs: Boolean(config?.inlineFS) && !asset.env.isNode(),
       insert_node_globals: !asset.env.isNode(),
+      relative_context: Boolean(config?.relativeContext),
```
LekoArts: `relative_context` was just an arbitrary name chosen by me to get things going.

I'll need input here on the proper name + shape of the API how it should look

packages/transformers/js/src/JSTransformer.js
```diff
@@ -248,6 +252,8 @@ export default (new Transformer({
     let rootPkg = result?.contents;
 
     let inlineEnvironment = config.isSource;
+    const relativeContext =
```
LekoArts: I couldn't get this to work with `rootPkg` (like a couple of lines below)

```
relativeContext = rootPkg['@parcel/transformer-js']?.relativeContext ?? relativeContext;
```

This didn't work :/

mischnic (commented):
Looks good overall!

IMO `relativeContext` doesn't really convey that this is about `__dirname` and `__filename` 🤷‍♂️ 

---

There's a similar existing test here: https://github.com/parcel-bundler/parcel/blob/47379bf8fabeb2cfe03ade8802d942388b153e5b/packages/core/integration-tests/test/javascript.js#L2646

So duplicate that and add a package.json with that new config option, then you can delete the example you've committed.

---

Regarding the `path` import: this is similar to how an import for `buffer` is added for the `Buffer` global, that is done here: https://github.com/parcel-bundler/parcel/blob/47379bf8fabeb2cfe03ade8802d942388b153e5b/packages/transformers/js/core/src/global_replacer.rs#L82

LekoArts:
Ok, so I pushed some more changes to the Rust code and I have two problems now:

1. Actually the generated relative path is incorrect
2. I'm not seeing my changes in the tests

Let me explain.

---

When we have a source file at `<root>/index.ts` and we compile to `<root>/dist/index.js` we would expect that the relative path to the `projectRoot` is "..".
But it isn't.

Here is the relevant line: https://github.com/LekoArts/parcel/blob/eca555e83e8ba3ac7ad0eded8dd9c0723a21713e/packages/transformers/js/core/src/global_replacer.rs#L144

`dirname` here is the path of the `<root>/index.ts` file, which is identical with `projectRoot`. So the relative path created will be "" and not "..".

How can I get the `distDir` here to construct a correct relative path?

---

I'm not seeing my Rust code changes in the tests. I _think_ I've added the test and the new option correctly, but when I print out the `js` (currently commented out) I'm not seeing the changes I did when I still had the example project in `packages/examples`.

LekoArts:
Let me know @mischnic if my above questions/comments make sense and you can help me? Appreciate it a lot! Thanks

mischnic:
> How can I get the distDir here to construct a correct relative path?

The transformer doesn't know the final path of the bundle that the asset will be put into. This is only known after bundling/packaging (https://parceljs.org/plugin-system/overview/), so you have to do the `path.relative(root, bundle.filePath)` computation and insertion in a packager (so `@parcel/packager-js`).

---

I don't know why the changes wouldn't show up in the tests (apart from forgetting to recompile the Rust code), so I guess it has to be that your change doesn't modify the behaviour somehow.

LekoArts:
Ok, thanks for that information. I'll look into the tests, I probably made an error somewhere.

I'm trying to wrap my head around how to go about this then. Do I need to keep the changes I made to the transformer so far and add some "identifier" so that later the packager can modify/replace things, or do I completely remove the Rust logic and make all my changes in the packager? From my understand I'd need both, so that in the transformer I actually replace the call and then the packager can change the import path?

mischnic:
Exactly. The logic to determine whether `__dirname` wasn't declared as a local variable definitely needs to happen in the transformer.

Then in the packager, one possibility would be replacing that placeholder literally or (though somewhat sketchy) do the same as  `$parcel$global` and just assume that this variable is unused by the user and emit it in the transformer, and then prepend a variable declaration in the packager. Though in this case you probably need one variable per asset

https://github.com/parcel-bundler/parcel/blob/13649c365d36b40e548e34e2e278bdb4cd1de48a/packages/transformers/js/core/src/hoist.rs#L835

LekoArts:
Ok, I was sick 🤧  for a bit, I'll start working on this again now

LekoArts:
Ok, so I got it working in my example (currently still committed in this PR) in how I expect it:

Bildschirmfoto 2022-03-18 um 12 54 01

But I'll remove the example and add tests.

The API and naming etc. is pretty gross and I'm also still using same placeholder everywhere, not per asset.

So it works, but still lot to do.

@mischnic

packages/packagers/js/src/index.js
```diff
@@ -75,6 +78,12 @@ export default (new Packager({
       }));
     }
 
+    if (config.relativeContext) {
+      const relPath = path.relative(bundle.target.distDir, options.projectRoot);
+      contents = contents.replace('$parcel$dirnameReplace', relPath);
+      contents = contents.replace('$parcel$filenameReplace', relPath);
```
devongovett: Probably need to use a regex to replace all instances. I think `replace` of a string only replaces the first one.

LekoArts: @devongovett I've added a second call to dirname/filename to my test file and the resulting output is:

```js
var __dirname1 = require("path").join(__dirname, "../test/integration/env-node-replacements");
var __filename = require("path").join(__dirname, "../test/integration/env-node-replacements", "index.js");
const fs = require('fs');
const path = require('path');
module.exports = function() {
    const data = fs.readFileSync(path.join(__dirname1, 'data', 'test.txt'), 'utf8');
    const firstDirnameTest = `${__dirname1}/data`;
    const secondDirnameTest = `${__dirname1}/other-data`;
    const firstFilenameTest = __filename;
    const secondFilenameTest = `${__filename}?query-string=test`;
    return {
        data,
        firstDirnameTest,
        secondDirnameTest,
        firstFilenameTest,
        secondFilenameTest
    };
};
```

So it replaces the calls with only one instance of $parcel$dirnameReplace/$parcel$filenameReplace. In which instances would I have multiple of those?

LekoArts:
@mischnic Can you approve CI runs on this PR?

> First-time contributors need a maintainer to approve running workflows.

Want to see if CI passes :)

packages/core/integration-tests/test/javascript.js
```diff
@@ -2656,6 +2656,47 @@ describe('javascript', function () {
...
+    assert(
+      output.firstFilenameTest.includes(
+        'integration-tests/test/integration/env-node-replacements/index.js',
+      ),
+    );
+    assert(
+      output.secondFilenameTest.includes(
+        'integration-tests/test/integration/env-node-replacements/index.js?query-string=test',
+      ),
+    );
```
mischnic: I assume that these fail on Windows because there'll be backslashes

LekoArts: Good callout. I'll copy something from our Gatsby repo to deal with this 😅

packages/core/integration-tests/test/fs.js
```diff
@@ -201,9 +201,8 @@ describe('fs', function () {
       assert(contents.includes("require('fs')"));
       assert(contents.includes('readFileSync'));
 
-      await outputFS.writeFile(path.join(distDir, 'test.txt'), 'hey');
       let output = await run(b);
-      assert.equal(output, 'hey');
+      assert.equal(output, 'hello');
```
LekoArts: Previously this was creating a new file in the distDir because the compiled code still had `__dirname` in it which resolved to the distDir. Now it points to the original directory so creating a new file is not necessary (the original file has "hello" in it)

LekoArts:
Mhh, seems like CI doesn't like me 😆  Not running 💤 
Ah: https://www.githubstatus.com/incidents/tyc8wpsgr2r8

packages/core/integration-tests/test/fs.js
mischnic: But now the test doesn't actually assert anymore that inlining doesn't happen...

packages/core/integration-tests/test/javascript.js
```diff
@@ -2678,6 +2681,59 @@ describe('javascript', function () {
...
+    let f = await run(b);
+    let output = f();
+    assert.equal(output.data, 'hello');
+    assert(
+      output.firstDirnameTest.includes(
+        xplatPath(
+          'integration-tests/test/integration/env-node-replacements/data',
+        ),
+      ),
+    );
```
mischnic: I think something like
```js
assert.equals(output.firstDirnameTest, path.join(__dirname, '/integration/env-node-replacements/data'));
```
would be better here. That also covers using the appropriate path separators for the current platform

packages/core/integration-tests/test/fs.js
LekoArts: Yeah, you're right. I'll think of something to have the old behavior. Creating the file inside the projectRoot directory should work?

packages/core/integration-tests/test/javascript.js
LekoArts: @mischnic done

packages/core/integration-tests/test/fs.js
LekoArts: I've did now what I proposed above

LekoArts:
@mischnic Can you trigger the CI again? 🙈 Hopefully next PR it should just work

mischnic:
(Done, for some reason it wanted me to approve you again, but not in the regular UI but only in the checks tab after selecting that workflow)

LekoArts:
Mhh, not sure why the benchmarks CI fails.

The windows CI fails for two things:

```
2) javascript
       should throw a codeframe for a missing file in worker constructor with URL and import.meta.url:
     AssertionError [ERR_ASSERTION]: Expected values to be strictly deep-equal:
+ actual - expected ... Lines skipped

  Comparison {
    diagnostics: [
...
      {
        hints: [
+         "Did you mean '__./index.js__'?",
+         "Did you mean '__./dynamic.js__'?"
-         "Did you mean '__./dynamic.js__'?",
-         "Did you mean '__./index.js__'?"
        ],
...
    ],
    name: 'BuildError'
  }
```

I don't think my change should affect that? Is that a race condition?

And second:

```
3) javascript
       should replace __dirname and __filename with path.join calls relative to projectRoot:
     Error: ENOENT: no such file or directory, open 'D:\a\parcel\parcel\packages\core\integration-tests\dist\..	estintegrationenv-node-replacements\data\test.txt'
      at Object.openSync (fs.js:497:3)
      at Object.readFileSync (fs.js:393:35)
      at NodeFS.readFileSync (D:\a\parcel\parcel\packages\core\fs\src\/NodeFS.js:81:17)
      at OverlayFS.readFileSync (D:\a\parcel\parcel\packages\core\fs\src\/OverlayFS.js:31:14)
      at Object.readFileSync (D:\a\parcel\parcel\packages\core\test-utils\src\/utils.js:889:30)
      at f (index.js:151:21)
      at Context.<anonymous> (test\/javascript.js:2705:18)
```

So yeah, still some work needed to construct the paths correctly for windows. I can test it next week with my windows machine for faster feedback cycle.

mischnic:
> Mhh, not sure why the benchmarks CI fails.

You can ignore them. They don't block merging

> I don't think my change should affect that? Is that a race condition?

Correct, it's unrelated. Yes, it is.

LekoArts:
I've changed it to `path.resolve` now to hopefully fix the pathing issue on windows.

Arghh, @mischnic the CI always wants approval on any new commit 😭 
Maybe if you merge https://github.com/parcel-bundler/parcel/pull/7688 after that I'll have the "power" 😆

LekoArts:
@mischnic I'm really sorry to bug you like this but seems like the windows test failed on something unrelated, can you please restart? Ideally I want to see today if I fixed all tests so that we can merge this week. Thanks!

LekoArts:
Mhh, it seems to fail on the HMR tests. Is that something I changed?

mischnic:
No, we have two flaky tests: the one where the order of filepaths is different, and this native crash (which is particularly hard to debug because there's no error message).

LekoArts:
Ah, those are the worst :/ Let's see if it all passes now. Okay, they do not, I'll have to test more

LekoArts:
Set up my Windows PC now with all build tools to try this locally:

image

I can fix this tomorrow then. Man, OS paths are hard :D

mischnic:
It would be better to use some of the already existing path normalization functions we have in https://github.com/parcel-bundler/parcel/blob/v2/packages/core/utils/src/path.js as opposed to writing your own in the integration test and the packager.

LekoArts:
Ahh, thanks for the pointer. I figured you'd have such a thing 👍  I'll let the tests run and then use the existing helpers

LekoArts:
Ah, depending on when https://github.com/parcel-bundler/parcel/pull/7870 gets merged either my or your PR needs a little update then.

I've used the path utils now, let's see if all works (locally all tests pass on windows & mac)

LekoArts:
I'm at a loss here, no idea why it always randomly fails on different things

mischnic:
Don't worry too much about failing CI if it works locally. It's our responsibility that they're not reliable, not yours.

So while passing CI means that your changes don't break the tests, a failing flakey CI doesn't necessarily mean that there's a bug in your changes 😉 .

LekoArts:
Ok, seems like the tests passed now 😅 
Just to make sure: This comment wasn't overlooked, right? https://github.com/parcel-bundler/parcel/pull/7727#discussion_r833111579

packages/packagers/js/src/index.js
mischnic: An example:
```js
// index.js
const fs = require("fs"); const path = require("path");
require("./src/other");
const data = fs.readFileSync(path.join(__dirname, "test.txt"), "utf8");
console.log(data);

// src/other.js
const fs = require("fs"); const path = require("path");
const data = fs.readFileSync(path.join(__dirname, "..", "test2.txt"), "utf8");
console.log(data);
```
`__dirname` has to be different in both cases

LekoArts: Okay, makes sense 👍  Then I'll adjust the test to incorporate that. How again should I make `$parcel$dirnameReplace` unique per asset? I already forgot 🙈

mischnic: Do the replacement per asset, so on the return value of `asset.getCode()`. That would be in `DevPackager#package` and on  the`code` param of `ScopeHoistingPackager#buildAsset`

(One potential performance optimization here would be marking assets using dirname via `asset.meta` and only doing this replacement conditionally.)

packages/packagers/js/src/DevPackager.js
```diff
@@ -116,6 +121,20 @@ export class DevPackager {
...
         wrapped += ']';
 
+        // For node environment @parcel/transformer-js replaces __dirname and __filename calls with
+        // path.join containing placeholders. The packager now replaces those placeholders with
+        // relative paths so that in the end the __dirname/__filename calls resolve to the correct files
+        if (this.bundle.env.isNode()) {
+          const relPath = normalizeSeparators(
+            path.relative(
+              this.bundle.target.distDir,
+              path.dirname(asset.filePath),
```
LekoArts, line 131: @mischnic Ok, after a bit of a treasure hunt (which was fun!) I managed to put the correct code paths into DevPackager and ScopeHoistingPackager.

How can I do the `asset.meta` optimization you talked about?

Oh, also, I changed it from `projectRoot` to `asset.filePath` as otherwise it wouldn't actually resolve to the correct file

LekoArts: I think I understand it. In the node_replacer rust code, when this gets run I should set some boolean. Then in the transformer I can access that boolean and set the `meta`, e.g. like https://github.com/parcel-bundler/parcel/blob/v2/packages/transformers/js/src/JSTransformer.js#L789

LekoArts: @mischnic Let me know if I did [`d249b88` (#7727)](https://github.com/parcel-bundler/parcel/pull/7727/commits/d249b880d304dfe4eef50440a7a6cba556c70e41) correctly :)

LekoArts:
The Integration Tests failure seems unrelated, also other PRs are failing on this

devongovett:
Code looks good to me! I'll test it out tomorrow. 😊

mischnic:
I think there's still a problem:
if I do
```js
const path = require("path")
console.log(__dirname, __filename)
```
both in `/Users/niklas/Desktop/y/index.js`  and `/Users/niklas/Desktop/y/sub/index.js` (and import the sub one in the main one), I get this output

```sh
$ node dist/index.js # (parcel output)
/Users/niklas/Desktop/y/sub /Users/niklas/Desktop/y/sub/sub/index.js
/Users/niklas/Desktop/y /Users/niklas/Desktop/y/index.js

$ node index.js # (the source)
/Users/niklas/Desktop/y/sub /Users/niklas/Desktop/y/sub/index.js
/Users/niklas/Desktop/y /Users/niklas/Desktop/y/index.js
```

and I think the double `sub/sub` on the second line shouldn't be there.
The bundle contains:
```js
var $0400fe44487afb84$var$__filename = $65Jaf$path.resolve(__dirname, "../sub", "sub/index.js");
```

LekoArts:
Thanks! Will look into it and add a test

LekoArts:
Good catch @mischnic 👍 

The issue was that I've used [pathdiff](https://docs.rs/pathdiff/latest/pathdiff/index.html)::[diff_paths](https://docs.rs/pathdiff/latest/pathdiff/fn.diff_paths.html) for `__filename` where I didn't need to use it. It just worked in the past but it seems with `/folder-name/index.js` it doesn't give back `index.js` but `/folder-name/index.js`.

packages/transformers/js/core/src/lib.rs
```diff
@@ -396,6 +400,34 @@ pub fn transform(config: Config) -> Result<TransformResult, std::io::Error> {
...
+                    filename: Path::new(&config.filename),
+                    decls: &mut decls,
+                    global_mark,
+                    scope_hoist: config.scope_hoist,
+                    has_node_replacements: &mut has_node_replacements,
+                  },
+                  config.node_replacer
+                ),
+                // Inject SWC helpers if needed.
+                helpers::inject_helpers(),
```
mischnic: 1. We don't need to run `inject_helpers` another time
2. ~I'd just move the NodeReplacer directly between `GlobalReplacer` and `preset_env` directly above this~ I see, you did this because of the borrow checker

LekoArts: Bildschirmfoto 2022-04-07 um 11 51 49

Bildschirmfoto 2022-04-07 um 11 52 00

I tried that at the beginning but I couldn't figure out how to solve this borrow issue :/
My Rust knowledge isn't sufficient enough for this I'm afraid

mischnic: There's no way around this borrow issue. You have to do this separately, so just: 
```rust
            let mut has_node_replacements = false;
            let module = module.fold_with(&mut Optional::new(
              NodeReplacer {
                source_map: &source_map,
                items: &mut global_deps,
                globals: HashMap::new(),
                project_root: Path::new(&config.project_root),
                filename: Path::new(&config.filename),
                decls: &mut decls,
                global_mark,
                scope_hoist: config.scope_hoist,
                has_node_replacements: &mut has_node_replacements,
              },
              config.node_replacer,
            ));
```

LekoArts: Done 👍

packages/packagers/js/src/ScopeHoistingPackager.js
```diff
@@ -538,6 +544,14 @@ ${code}
       this.needsPrelude = true;
     }
 
+    if (this.bundle.env.isNode() && asset.meta.has_node_replacements) {
+      const relPath = normalizeSeparators(
+        path.relative(this.bundle.target.distDir, path.dirname(asset.filePath)),
+      );
+      code = code.replace('$parcel$dirnameReplace', relPath);
+      code = code.replace('$parcel$filenameReplace', relPath);
+    }
```
mischnic: This should happen further up in this function (e.g. near `if (code.includes('$parcel$global'))`) because there are recursive calls inside `buildAsset` where it inlines dependencies. Right now, `code` might contain the code for multiple assets and then `asset.filePath` is wrong.

mischnic (approved):
Looks good to me!

LekoArts:
Awesome 🎉  I'll let you re-run the CI, otherwise with my commit trigger I'd invalidate your review

packages/packagers/js/src/DevPackager.js
```diff
@@ -116,6 +121,20 @@ export class DevPackager {
...
+          asset.meta.has_node_replacements === true
+        ) {
+          const relPath = normalizeSeparators(
+            path.relative(
+              this.bundle.target.distDir,
+              path.dirname(asset.filePath),
+            ),
+          );
+          wrapped = wrapped.replace('$parcel$dirnameReplace', relPath);
+          wrapped = wrapped.replace('$parcel$filenameReplace', relPath);
```
devongovett, line 135: Is there a reason we have two different variables that get replaced with the same path? Were these meant to be different, or do we only need one of them?

packages/transformers/js/core/src/node_replacer.rs
```diff
@@ -0,0 +1,218 @@
...
+            attributes: None,
+            is_optional: false,
+            is_helper: false,
+            source_type: Some(SourceType::Module),
+            placeholder: None,
+          });
+
+          *self.has_node_replacements = true;
+        }
+        "__dirname" => {
```
devongovett, line 126: Some of this could probably be combined with the above but ok for now

devongovett (approved):
Looks good overall! Thanks again!

LekoArts:
I will optimize this (i.e. address the two new review comments) in a follow-up PR -- I'd like to get this into a nightly release so that we can try it out at Gatsby :)

LekoArts:
I think you'll need to re-run a bunch of times for the flaky tests 😅

LekoArts:
With https://github.com/parcel-bundler/parcel/pull/7931 merged I guess I need to change some `ast::Lit::Str` occurrences. I'll do this in a bit

mischnic:
Sorry, I forgot that merging it would break this PR 😬

mischnic:
Looks like it compiles, the only breaking change from the swc update was related to `:ast:Number`. And these changes from `ast::Str{}` to `.into()` are just using a shorter syntax and are equivalent. So we can clean that up later.

LekoArts:
Awesome. I'll do that in the follow-up PR. Thanks for merging and help on the PR <3

LekoArts:
The nightly release that would include this failed here: https://github.com/parcel-bundler/parcel/actions/runs/2151865409

sharifzadesina:
How to disable this PR? we most of the time need the actual `__filename` and `__dirname` values.

PR #7735, opened 2022-02-18
Fix autoinstalling node builtins

### LINKED ISSUE #7697: Build error with version 2.3.1: @parcel/core: Failed to resolve 'process' from ...
# 🐛 bug report

I just upgraded to version 2.3.1 and now the build process fails.
Don't know exactly if this is a parcel or a sentry issue?

```js
{
    "presets": [
        [
            "@babel/preset-env",
            {
                "modules": false,
                "useBuiltIns": false,
                "targets": {
                    "node": "current"
                }
            }
        ],
        "@babel/preset-react"
    ],
    "plugins": [],
    "env": {
        "test": {
            "plugins": ["@babel/plugin-transform-modules-commonjs"]
        }
    }
}

```

## 🤔 Expected Behavior

Build should pass.

## 😯 Current Behavior

The build fails with the following exception:
```
🚨 Build failed.

@parcel/core: Failed to resolve 'process' from './node_modules/@sentry/utils/esm/node.js'

  /Users/*****/Sites/*****/node_modules/@sentry/utils/esm/node.js:18:43
    17 | /**
  > 18 |  * Requires a module which is protected against bundler minification.
  >    |                                           ^^^^^^^^
    19 |  *
    20 |  * @param request The module path to resolve
```

## 💁 Possible Solution

idk.

## 🌍 Your Environment

| Parcel           | 2.3.1
| Node             | 16.11.1
| npm/Yarn         | yarn: 1.22.17
| Operating System | mac osx

devongovett:
Fixes #7697

We need to pass the original builtin specifier (e.g. `process/browser.js` or `buffer/`) to resolve rather than just the package name (e.g. `process` or `buffer`) because the package name will resolve in node to a builtin rather than the browser version.

### DIFF
--- a/packages/utils/node-resolver-core/src/NodeResolver.js
+++ b/packages/utils/node-resolver-core/src/NodeResolver.js
@@ -313,7 +313,7 @@ export default class NodeResolver {
             'https://parceljs.org/features/node-emulation/#polyfilling-%26-excluding-builtin-node-modules',
         });
 
-        await packageManager.resolve(packageName, this.projectRoot + '/index', {
+        await packageManager.resolve(builtin, this.projectRoot + '/index', {
           saveDev: true,
           shouldAutoInstall: true,
         });


packages/utils/node-resolver-core/src/NodeResolver.js
```diff
@@ -313,7 +313,7 @@ export default class NodeResolver {
             'https://parceljs.org/features/node-emulation/#polyfilling-%26-excluding-builtin-node-modules',
         });
 
-        await packageManager.resolve(packageName, this.projectRoot + '/index', {
+        await packageManager.resolve(builtin, this.projectRoot + '/index', {
           saveDev: true,
```
jtojnar, line 317: By the way, should not this be a non-dev dependency?

eightyfive:
When will this be included in a `patch` release ?

devongovett:
Yes, it is in v2.3.2.

packages/utils/node-resolver-core/src/NodeResolver.js
devongovett: Maybe? Two reasons I made it a dev dep:

1. It was a dev dependency before when it was included in Parcel itself.
2. The "process" dependency might actually be used not in your project, but by a dependency in node_modules. We cannot add the dep to that, so we add it to the project root instead. So it's kinda an indirect dependency, needed by Parcel not by your project.

I could be convinced otherwise if you feel strongly.

jtojnar: I was just curious, not sure if there is any difference either way.

PR #7737, opened 2022-02-18
Use relative path for bundle labels in bundle analysis

ittaibaratz:
There is no value in providing the full path as the bundle label, and it can get quite long in CI/CD

Changing to relativePath will make the label shorter and easier to read

# ↪️ Pull Request

## ✔️ PR Todo


packages/reporters/bundle-analyzer/src/BundleAnalyzerReporter.js
```diff
@@ -142,7 +142,7 @@ async function getBundleNode(bundle: PackagedBundle, options: PluginOptions) {
   }
 
   return {
-    label: bundle.filePath,
+    label: relativePath,
```
mischnic: Is this what you had in mind?
```suggestion
    label: path.relative(options.projectRoot, bundle.filePath),
```

ittaibaratz: @mischnic yep thanks!

ittaibaratz:
@mischnic I would also like to enable bundle analysis on the modules only: (Right now it has both module and no-module, so everything is replicated)

image

Any quick pointers for how I would make it happen? Thanks!

i.e. where can I place the config for that behavior assuming that not everyone wants it.

cc @devongovett

mischnic:
I guess the hard part for that is determining which bundle is the nomodule half of a nomodule/module pair.
And a config that only lists module bundles in the report could incorrectly filter out non-module worker bundles...

mischnic (approved):
The only problem I can see with this change is projects including files from outside the project path, then you'd get paths like `../../foo`. But this situation isn't really supported anyway because these files aren not watched anyway

PR #7741, opened 2022-02-21
Fix transformer-svg-react not finding .svgrrc's

### LINKED ISSUE #8270: svgo.config.json doesn't work, but svgo.config.js does
When using `@parcel/transformer-svg-react`, I noticed that it was stripping the viewBox attribute.
I tried enabling it, but this didn't work:

```js
// svgo.config.json
{
  "plugins": [
    {
      "name": "preset-default",
      "params": {
        "overrides": {
          "removeViewBox": false
        }
      }
    }
  ]
}
```

After trying different things and removing .parcel-cache, changing config to a .js file worked:

```js
// svgo.config.js
module.exports = {
  plugins: [
    {
      name: 'preset-default',
      params: {
        overrides: {
          removeViewBox: false,
        },
      },
    },
  ],
};
```

Parcel packages used:
```
    "@parcel/config-webextension": "^2.6.0",
    "@parcel/transformer-inline-string": "^2.6.0",
    "@parcel/transformer-svg-react": "^2.6.0",
    "parcel": "^2.6.0",
```

coolreader18:
# ↪️ Pull Request

`componentName` was being passed as `filePath` instead of `componentName` and as a side effect svgr couldn't properly find a `.svgrrc` file, if there was one.

Closes https://github.com/parcel-bundler/parcel/issues/8270


coolreader18:
Though actually, wait, it should probably use parcel's own loadConfig stuff for caching and whatnot. One sec.

mischnic:
Can you add a test case for using a svgr config file?

coolreader18:
Alright, added. I'm realizing though that ideally the `svgrrc` shouldn't be necessary in the first place just to get it to work with preact (though still necessary for other things); would it be best to copy the JSX pragma detection code from JSTransformer, depend on it through package.json, or maybe some other method?

coolreader18:
(And maybe that should be done in a different PR? idk)

coolreader18:
Is the failure just a fluke?

mischnic:
According to the docs, `runtimeConfig: false` also disables the SVGO config search. So I think that will have to be loaded in `loadConfig` as well and then passed to the transform call. (A test for that would also be great).

The `assert(file.includes('h("svg"'));` assertion is failing on CI, that's not a fluke if it's on every platform.

> would it be best to copy the JSX pragma detection code from JSTransformer, depend on it through package.json, or maybe some other method?

Yeah, I think putting that code into some shared package would be the best way. But that's actually a separate problem for another PR.

packages/transformers/svg-react/src/SvgReactTransformer.js
```diff
@@ -3,41 +3,61 @@
...
+    ]);
+    let svgoResult = await config.getConfig([
+      '.svgorc',
+      '.svgorc.json',
+      '.svgorc.js',
+      '.svgorc.cjs',
+      'svgo.config.json',
+      'svgo.config.js',
+      'svgo.config.cjs',
+    ]);
```
devongovett, line 25: Looks like the SVG optimizer only supports `svgo.config.*`. Maybe we should match that here?

packages/transformers/svg-react/src/SvgReactTransformer.js
```diff
@@ -3,41 +3,61 @@
...
+    if (svgrResult) {
+      let isJavascript = path.extname(svgrResult.filePath) === '.js';
```
devongovett: These should also handle `.cjs`

devongovett (approved):

devongovett:
Test appears to be failing...

mischnic:
Forgot to rename the `.svgorc` file in the test 🙈

PR #7766, opened 2022-02-26
Human readable file size in bundle analyzer report

solstice23:
# ↪️ Pull Request

Make file sizes readable in the bundle analyzer report.

## 💻 Examples

image
image

### DIFF
--- a/packages/reporters/bundle-analyzer/client/index.js
+++ b/packages/reporters/bundle-analyzer/client/index.js
@@ -101,11 +101,11 @@ function translate3d(x, y, z) {
   return `translate3d(${x}px, ${y}px, ${z}px)`;
 }
 
-function formatSize(x){
-  const units = ['B', 'KB', 'MB', 'GB', 'TB', 'PB', 'EB', 'ZB', 'YB'];
+const UNITS = ['B', 'KB', 'MB', 'GB', 'TB', 'PB', 'EB', 'ZB', 'YB'];
+function formatSize(x) {
   let l = 0, n = parseInt(x, 10) || 0;
   while(n >= 1000 && ++l){
       n /= 1000;
   }
-  return(`${n.toFixed(l > 0 ? 2 : 0)} ${units[l]}`);
+  return(`${n.toFixed(l > 0 ? 2 : 0)} ${UNITS[l]}`);
 }
--- a/packages/reporters/bundle-analyzer/client/index.js
+++ b/packages/reporters/bundle-analyzer/client/index.js
@@ -102,10 +102,10 @@ function translate3d(x, y, z) {
 }
 
 function formatSize(x){
-  const units = ['bytes', 'KB', 'MB', 'GB', 'TB', 'PB', 'EB', 'ZB', 'YB'];
+  const units = ['B', 'KB', 'MB', 'GB', 'TB', 'PB', 'EB', 'ZB', 'YB'];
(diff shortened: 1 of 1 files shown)


packages/reporters/bundle-analyzer/client/index.js
```diff
@@ -100,3 +100,12 @@ function debounce(fn, delay) {
...
 }
+
+function formatSize(x){
+  const units = ['bytes', 'KB', 'MB', 'GB', 'TB', 'PB', 'EB', 'ZB', 'YB'];
+  let l = 0, n = parseInt(x, 10) || 0;
+  while(n >= 1024 && ++l){
+      n /= 1024;
+  }
+  return(`${n.toFixed(n < 10 && l > 0 ? 1 : 0)} ${units[l]}` + (l ? ` (${x} bytes)` : ''));
+}
```
kidonng: Parcel is using [`filesize`](https://github.com/parcel-bundler/parcel/blob/18b038d57043965afb31077db1f46dadb11a6a78/packages/reporters/cli/package.json#L39), therefore it would be better to have the same behavior:

```suggestion
function formatSize(x){
  const units = ['B', 'KB', 'MB', 'GB', 'TB', 'PB', 'EB', 'ZB', 'YB'];
  let l = 0, n = parseInt(x, 10) || 0;
  while(n >= 1000 && ++l){
      n /= 1000;
  }
  return(`${n.toFixed(n < 10 && l > 0 ? 1 : 0)} ${units[l]}`);
}
```

mischnic:
I think there should be some decimal places, like the Webpack bundle analyzer has: https://www.npmjs.com/package/webpack-bundle-analyzer

And in that regard, why not use `filesize` directly instead of handcoding that?
```
> require("filesize")(100000)
'97.66 KB'
```

kidonng:
> And in that regard, why not use `filesize` directly instead of handcoding that?

This script runs in the browser, so probably not worth the hassle to vendor an extra library.

solstice23:
The report is a single HTML file, so it might be better to calculate it without extra library.
Now it is rounded to 2 decimal places, same as the behavior of the `filesize`.

packages/reporters/bundle-analyzer/client/index.js
```diff
@@ -100,3 +100,12 @@ function debounce(fn, delay) {
...
 }
+
+function formatSize(x){
+  const units = ['B', 'KB', 'MB', 'GB', 'TB', 'PB', 'EB', 'ZB', 'YB'];
+  let l = 0, n = parseInt(x, 10) || 0;
+  while(n >= 1000 && ++l){
+      n /= 1000;
+  }
+  return(`${n.toFixed(l > 0 ? 2 : 0)} ${units[l]}`);
+}
```
mischnic:
```suggestion
const UNITS = ['B', 'KB', 'MB', 'GB', 'TB', 'PB', 'EB', 'ZB', 'YB'];
function formatSize(x) {
  let l = 0, n = parseInt(x, 10) || 0;
  while(n >= 1000 && ++l){
      n /= 1000;
  }
  return(`${n.toFixed(l > 0 ? 2 : 0)} ${UNITS[l]}`);
}
```

mischnic (approved):
Thanks! (I missed that this parts runs in the browser 👍 )

devongovett (approved):

PR #7775, opened 2022-02-28
Improve emoji support detection

### LINKED ISSUE #7767: Emoji support check is biased
# 🙋 feature request

Current code to check whether emoji is supported:

https://github.com/parcel-bundler/parcel/blob/18b038d57043965afb31077db1f46dadb11a6a78/packages/reporters/cli/src/emoji.js#L3-L4

This code is naive:

- It assumes **all** Windows version does not support emoji, which is not true. [According to Emojipedia](https://emojipedia.org/windows-emoji-support/), Windows supports emoji on Windows 7 or above. I doubt any Parcel user would not develop on a supported Windows version.
- It assumes **all** terminal's `$TERM` is `xterm-256color`. This is not true because obviously not all terminals are xterm. Some popular terminals use their own values (e.g. [kitty](https://sw.kovidgoyal.net/kitty/) uses `xterm-kitty`).

## 🤔 Expected Behavior

Parcel should correctly enable emoji on supported platforms.

## 😯 Current Behavior

Parcel does not correctly enable emoji on supported platforms.

## 💁 Possible Solution

- Remove Windows check all together, or at least add a major version check
- It's tricky to check actual emoji support, but we can safely assume that if a terminal supports 256 color, it probably supports emoji as well. The check can be done via e.g. [supports-color](https://github.com/chalk/supports-color).

kidonng:
# ↪️ Pull Request

Resolve #7767

Unfortunately [`is-unicode-supported`](https://github.com/sindresorhus/is-unicode-supported) has to be vendored since it's ESM-only.

## 🚨 Test instructions

Assuming you are not on Windows (ironically, but this is easier to test):

```sh
# 1. Inspect $TERM
$ echo $TERM
xterm-256color
# 2. Create a dummy
$ touch index.html
# 3. Build the dummy, you should be able to see an emoji
$ yarn parcel build index.html
✨ Built in 511ms
# 4. Pretend to be Linux console and build the dummy again, should be able to see the fallback symbols:
$ TERM=linux yarn parcel build index.html
√ Built in 511ms
```

## ✔️ PR Todo

### DIFF
--- a/packages/reporters/cli/src/emoji.js
+++ b/packages/reporters/cli/src/emoji.js
@@ -1,7 +1,23 @@
 // @flow strict-local
 
-const supportsEmoji =
-  process.platform !== 'win32' || process.env.TERM === 'xterm-256color';
+// From https://github.com/sindresorhus/is-unicode-supported/blob/8f123916d5c25a87c4f966dcc248b7ca5df2b4ca/index.js
+// This package is ESM-only so it has to be vendored
+function isUnicodeSupported() {
+  if (process.platform !== 'win32') {
+    return process.env.TERM !== 'linux'; // Linux console (kernel)
+  }
+
+  return (
+    Boolean(process.env.CI) ||
+    Boolean(process.env.WT_SESSION) || // Windows Terminal
+    process.env.ConEmuTask === '{cmd::Cmder}' || // ConEmu and cmder
+    process.env.TERM_PROGRAM === 'vscode' ||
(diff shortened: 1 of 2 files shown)


AGawrys:
Feel free to add me as a reviewer once you're satisfied with the PR TODO you listed :)

kidonng:
@AGawrys I don't think I have the permission to do that, but I have tested it locally and provided the test instructions.

devongovett (approved):

Issue #7166, opened 2021-10-23
Store warnings across builds


mischnic:
# 🙋 feature request

In transformers/packagers/resolvers/... you can throw a `ThrowableDiagnostic`. Because it is in a fatal error, rerunning Parcel will reevaluate the plugin.

But for warnings, you have to use `logger.warn` which is just a wrapper around `console.warn` to format the error. This means that warnings are only printed when the corresponding plugin runs, not when a cached result is used in a subsequent build.

We need a way to actually store warnings together with a successful plugin invocation (transformation, packaging, ...) without swallowing warnings.

## 💁 Possible Solution

- return diagnostics together with the return value
	- Not very elegant. We'd need this in every method (`Transformer.(loadConfig|transform|...}`, `Packager.(loadConfig|package)`, ...)
- store calls to `logger.*`
	- but where should the warnings be stored in a transformer that returns multiple assets? 

## 🔦 Context

Examples for such diagnostics:

- validators suffer from this problem (though this doesn't influence the API problem, only how we'd store the information internally)
- the verbose-mode tree shaking recommendations
- warnings about unnecessary Babel plugins

PR #7817, opened 2022-03-10
Bundle validator for preventing duplicate assets

AGawrys:
Co-Author @wbinnssmith 

### ↪️ Main Changes

Created a new validator, `NoDuplicateAssetsValidator`, which validates all bundles against some asset that should not be duplicated. Currently this is a hard coded value `noDup.js` for testing purposes.

Validators now return a`ValidationResult` which is then stored and reused on subsequent builds, thus, fixing the cache issue where diagnostics disappear after a build where nothing changed. I believe this is similar to [this issue](https://github.com/parcel-bundler/parcel/issues/7166). 

Whether or not validators are run is _no longer based solely_ on if there are changed assets.

### 🚨 Questions

- Should we move any cache-related validation tests to the cache tests? 

## ✔️ PR Todo

- [x] Added tests for this change
- [ ] Determine what assets should we *actually* enforce no duplication for 
- [ ] Or add way to customize what assets shouldn't be duplicated
- [ ] Remove hard coded `NoDup.js` and update test and cleanup

### DIFF
--- a/packages/core/core/src/ParcelConfig.js
+++ b/packages/core/core/src/ParcelConfig.js
@@ -64,7 +64,7 @@ export default class ParcelConfig {
   namers: PureParcelConfigPipeline;
   runtimes: PureParcelConfigPipeline;
   packagers: GlobMap<ParcelPluginNode>;
-  validators: GlobMap<ExtendableParcelConfigPipeline>;
+  validators: $ReadOnlyArray<ParcelPluginNode>;
   optimizers: GlobMap<ExtendableParcelConfigPipeline>;
   compressors: GlobMap<ExtendableParcelConfigPipeline>;
   reporters: PureParcelConfigPipeline;
@@ -83,7 +83,7 @@ export default class ParcelConfig {
     this.optimizers = config.optimizers || {};
     this.compressors = config.compressors || {};
     this.reporters = config.reporters || [];
-    this.validators = config.validators || {};
+    this.validators = config.validators || [];
     this.pluginCache = new Map();
     this.regexCache = new Map();
   }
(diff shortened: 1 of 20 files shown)


packages/core/core/src/ParcelConfig.js
```diff
@@ -64,7 +64,7 @@ export default class ParcelConfig {
   namers: PureParcelConfigPipeline;
   runtimes: PureParcelConfigPipeline;
   packagers: GlobMap<ParcelPluginNode>;
-  validators: GlobMap<ExtendableParcelConfigPipeline>;
+  validators: $ReadOnlyArray<ParcelPluginNode>;
```
wbinnssmith, line 67: nit:

```suggestion
  validators: PureParcelConfigPipeline;
```

Even though this isn't strictly a pipeline (plugins don't use the result from previous plugins), we seem to be using this for things like reporters so 🤷 

Alternatively, we could create another type alias that points to this that describes it more of a plugin "list".

mischnic: (Runtimes also don't use previous results, but they are applied in order so that might be another defining property of a pipline)

Speaking of that, doesn't removing the glob filters break the existing validators because they now run on every asset regardless of type?

wbinnssmith: Existing examples of validators have their own glob rules (in eslintrc and tsconfig) to determine if they should run on particular file globs. Worth covering with a test to confirm the behavior when this happens though.

AGawrys:
Closing as this hasn't been touched in a while & I don't plan to

PR #7834, opened 2022-03-16
Preserve query string when reloading CSS assets

dmitrijs:
# ↪️ Pull Request

Preserve query string when HMR reloads CSS assets.
This prevents situation when a dynamically loaded styles are not reloaded correctly.

## 💻 Examples

Styles can be loaded dynamically using code such as:
```html
<link rel="stylesheet" href="https://example.com/wp-admin/load-styles.php?c=1&amp;dir=ltr&amp;load%5Bchunk_0%5D=dashicons,admin-bar,common,forms,admin-menu,dashboard,list-tables,edit,revisions,media,themes,about,nav-menus,wp-pointer,widgets&amp;load%5Bchunk_1%5D=,site-icon,l10n,buttons,wp-auth-check&amp;ver=5.9.2" type="text/css" media="all">
```
However, triggering hmr changes the tag to:
```html
<link rel="stylesheet" href="https://example.com/wp-admin/load-styles.php?1647423873644" type="text/css" media="all">
```
New `href` attribute is missing the query string, which contained information that is required to generate correct styles.

## 🚨 Test instructions

1. Add a `<link>` attribute to `<head>` of `index.html`. `href` attribute should include a query string;
2. Run `npx parcel watch index.html`;
3. Trigger hot reload by changing source code.
Expected result: original query string is preserved, `__parcel_timestamp__=<timestamp>` is appended to the `href` attribute.
Actual result: query string is replaced with `<timestamp>`

## ✔️ PR Todo

- [X] Added/updated unit tests for this change *(no unit tests in hmr module)*
- [X] Included links to related issues/PRs *(did not find any issue or pull request)*

### DIFF
--- a/packages/runtimes/hmr/src/loaders/hmr-runtime.js
+++ b/packages/runtimes/hmr/src/loaders/hmr-runtime.js
@@ -241,10 +241,13 @@ function updateLink(link) {
       link.parentNode.removeChild(link);
     }
   };
+  var newLinkUrl = link.getAttribute('href').replace(/[?&]__parcel__timestamp__=\d+/, '');
+  newLinkUrl += (newLinkUrl.indexOf('?') < 0 ? '?' : '&') + '__parcel__timestamp__=' + Date.now()
+
   newLink.setAttribute(
     'href',
     // $FlowFixMe
-    link.getAttribute('href').split('?')[0] + '?' + Date.now(),
+    newLinkUrl,
   );
   // $FlowFixMe
   link.parentNode.insertBefore(newLink, link.nextSibling);


packages/runtimes/hmr/src/loaders/hmr-runtime.js
```diff
@@ -241,10 +241,13 @@ function updateLink(link) {
       link.parentNode.removeChild(link);
     }
   };
+  var newLinkUrl = link.getAttribute('href').replace(/[?&]__parcel__timestamp__=\d+/, '');
+  newLinkUrl += (newLinkUrl.indexOf('?') < 0 ? '?' : '&') + '__parcel__timestamp__=' + Date.now()
```
devongovett, line 245: Maybe instead of reloading all `<link>` elements across all domains, we could only reload the css from the current domain (e.g. parcel dev server). Might avoid unnecessary network requests. Would that also solve your issue?

devongovett, line 245: Maybe instead of reloading all `<link>` elements across all domains, we could only reload the css from the current domain (e.g. parcel dev server). Might avoid unnecessary network requests. Would that also solve your issue?

tcmulder: In the case of WordPress, the dynamically created https://example.com/wp-admin/load-styles.php?... stylesheet is from the current domain, the same domain as the site itself (https://example.com), so that won't work in that instance. I'm running into this issue also, although it works with Google Fonts dynamic URLs like https://fonts.googleapis.com/css2?family=Barlow+Condensed that are on a different domain.

PR #5975, opened 2021-03-08
Resolver: correctly determine sideEffects for subdir imports `pkg/subdir/`

mischnic:
For imports of subdirectories like `pkg/subdir`, `sideEffects` wasn't read in from `pkg/package.json`.

Prevents situations like

Main

(It's unclear whether this situation has caused actual bugs)

(And this also makes clear why resolvers should behave somewhat deterministically 😄 )

PR #7909, opened 2022-04-06
NodeResolver: Reload the closest package.json to an asset if it's a package entry

### LINKED ISSUE #7908: Resolved package.json does not reflect resolved file
Consider the following tree:

```
foo/
  package.json {sideEffects: false}
  real-bar.js
  bar/
    package.json {main: '../real-bar.js', has no defined sideEffects property}
```

Intuitively, it seems that real-bar.js's sideEffects value should be `false`, not `undefined`, as it's a child of the foo directory, which has a package.json defining `"sideEffects": false`. 

However, when `import 'foo/bar'` is currently resolved, it resolves correctly to `foo/real-bar.js`, but the associated package.json for the asset is `foo/bar/package.json`, not `foo/package.json`. This makes `foo/real-bar.js`'s sideEffects value `undefined` when it's required as `import 'foo/bar'`, but it's `false` when it's required as `import 'foo/real-bar.js'` (or `import './real-bar.js'` or similar).

We should consider conditionally re-resolving an asset's package.json when the initially resolved package.json points to a module that would otherwise resolve a different package.json.

cc @mischnic @devongovett @thebriando

wbinnssmith:
Closes #7908.

~~This heuristic is enough to solve the example in #7908, but it wouldn't hold in the case that the package.json defines a main as `"./in/some/nested-package/index.js"` where `"./in/some/nested-package/package.json"` also exists.~~ Updated to handle this case.

Test Plan: Added an integration test.


parcel-benchmark:
## Benchmark Results

#### Timings

| --- | --- | --- |

#### Cold Bundles

#### Cached Bundles

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 10.59s | +127.00ms |
| Cached | 516.00ms | +17.00ms |

#### Cold Bundles

| dist/logo.c5bb83f1.png | 246.00b | +0.00b | 5.48s | +707.00ms ⚠️ |

#### Cached Bundles

*No bundle changes detected.*

</p></details>

#### Timings

| --- | --- | --- |

#### Cold Bundles

#### Cached Bundles

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 7.62s | +32.00ms |
| Cached | 296.00ms | -10.00ms |

#### Cold Bundles

*No bundle changes detected.*

#### Cached Bundles

*No bundle changes detected.*

</p></details>

**[Click here to view a detailed benchmark overview.](https://benchmark.parceljs.org/benchmark/330223491933536850)**

packages/core/integration-tests/test/scope-hoisting.js
```diff
@@ -1964,6 +1964,31 @@ describe('scope hoisting', function () {
...
+      let bar;
+      b.traverse((node, _, actions) => {
+        if (
+          node.type === 'asset' &&
+          node.value.filePath.endsWith('real-bar.js')
+        ) {
+          bar = node.value;
+          actions.stop();
+        }
+      });
```
mischnic: We have a test utility for this: https://github.com/parcel-bundler/parcel/blob/a8a4f5114944512aad60df9baffa741f31705c8d/packages/core/test-utils/src/utils.js#L143-L146

mischnic:
For future reference, this bug is just another version of https://github.com/parcel-bundler/parcel/pull/5975

mischnic:
Also, instead or additionally adding a resolver unit test might be good

mischnic:
This update should definitely be done inside `loadAsFile`, but I think that additional parameter is kind of odd...
An alternative would be using `pkg.pkgDir !== path.basename(found) ? await this.readPackage() : pkg`.

wbinnssmith:
> An alternative would be using pkg.pkgDir !== path.basename(found) ? await this.readPackage() : pkg.

This doesn't really capture whether the file was loaded as a package entry, which is afaik the only thing that could invalidate the pkg we had loaded prior. It would also cause unnecessary reloads for files in packages with additional directories that were not themselves packages, yeah?

mischnic:
> file was loaded as a package entry,  which is afaik the only thing that could invalidate the pkg we had loaded prior

Right

> It would also cause unnecessary reloads for files in packages with additional directories that were not themselves packages

Every folder can contain a package.json, but there is only one case where it wouldn't match. So there shouldn't be unnecessary reloads.
And that `pkg.pkgDir !== path.basename(found)` condition should cause less loads in cases where the package entry field doesn't refer to a parent or sub folder?

(And if we do want to keep that flag, maybe rename `asPackageEntry` to `reloadPkg`?)

mischnic:
So some of these new invalidations are just duplicates, some make sense (e.g. for `should resolve node builtin modules`) but some are somehow just not needed, like this one? 
`node_modules/foo/package.json` already exists and it doesn't need to watch above that

```diff
    it('should resolve a node_modules index.js', async function () {
        let resolved = await resolver.resolve({
        env: BROWSER_ENV,
        filename: 'foo',
        specifierType: 'esm',
        parent: path.join(rootDir, 'foo.js'),
      });
      assert.deepEqual(resolved, {
        filePath: path.join(rootDir, 'node_modules', 'foo', 'index.js'),
        sideEffects: undefined,
        query: undefined,
        invalidateOnFileCreate: [
          {
            fileName: 'package.json',
            aboveFilePath: path.join(rootDir, 'index'),
          },
          {
            fileName: 'package.json',
            aboveFilePath: path.join(rootDir, 'foo.js'),
          },
          {
            fileName: 'node_modules/foo',
            aboveFilePath: path.join(rootDir, 'foo.js'),
          },
+         {
+           fileName: 'package.json',
+           aboveFilePath: path.join(rootDir, 'node_modules/foo/index.js'),
+         },
        ],
        invalidateOnFileChange: [
          path.join(rootDir, 'package.json'),
          path.join(rootDir, 'node_modules', 'foo', 'package.json'),
        ],
```

wbinnssmith:
This regressed the build time of a large app by about 3.4% (203s to 196s for a complete build), though it is arguably a correctness and compatibility issue with other bundlers.

packages/utils/node-resolver-core/test/resolver.js
```diff
@@ -850,6 +893,16 @@ describe('resolver', function () {
...
+            {
+              fileName: 'package.json',
+              aboveFilePath: path.join(
+                rootDir,
+                'node_modules',
+                'side-effects-false',
+                'src',
+                'index.js',
+              ),
+            },
```
mischnic, line 905: @devongovett Does `aboveFilePath` also stop the watching in parent folder after encountering `node_modules` (like `loadConfig` does)? Or is this fine?

devongovett:
> This regressed the build time of a large app by about 3.4%

Hmm that's unfortunate.

devongovett (approved):

PR #3562, opened 2019-09-25
Graph integrity: Check for node membership

wbinnssmith:
This adds checks for node membership before most graph methods run, including adding edges, removing nodes, replaceNodesConnectedTo, getting connected nodes, traversal, etc.

AssetGraph needed to be changed to add nodes before adding edges, as it added edges to temporarily non-existent nodes.

Test Plan: Added unit tests

PR #7927, opened 2022-04-08
Correctly remove orphaned non-tree subgraphs

mischnic:
# ↪️ Pull Request

`removeNode(1)`  first removes 0-1, then 1-2 (because 2 is an orphan), then 2-3 (because 2 is an orphan after 0-1 was removed). The the right branch is visited and 2-3 is removed again which fails

undefined

I think it makes more sense to first remove the outgoing edges and then the incoming edges. This way, the subgraph is removed bottom up and a nodes becomes only orphaned once the last parent of that node is removed. So there will always be exactly one anchestor node that is responsible for removing it and no double deletion as before.

## 💻 Examples

The asset graph looked like this, changing some target option in package.json would remove the subgraph and run into a `Edge from x to y not found`  error.

### DIFF
--- a/packages/core/graph/test/Graph.test.js
+++ b/packages/core/graph/test/Graph.test.js
@@ -318,7 +318,7 @@ describe('Graph', () => {
     assert.deepEqual(visited, [nodeA, nodeB, nodeD]);
   });
 
-  it('foo', () => {
+  it('correctly removes non-tree subgraphs', () => {
     let graph = new Graph();
     let nodeRoot = graph.addNode('root');
     let node1 = graph.addNode('1');
--- a/packages/core/graph/test/Graph.test.js
+++ b/packages/core/graph/test/Graph.test.js
@@ -317,4 +317,24 @@ describe('Graph', () => {
 
     assert.deepEqual(visited, [nodeA, nodeB, nodeD]);
   });
+
+  it('foo', () => {
+    let graph = new Graph();
+    let nodeRoot = graph.addNode('root');
+    let node1 = graph.addNode('1');
+    let node2 = graph.addNode('2');
+    let node3 = graph.addNode('3');
+
+    graph.addEdge(nodeRoot, node1);
+    graph.addEdge(node1, node2);
+    graph.addEdge(node1, node3);
+    graph.addEdge(node2, node3);
+
+    graph.setRootNodeId(nodeRoot);
+
+    graph.removeNode(node1);
+
+    assert.strictEqual(graph.nodes.size, 1);
+    assert.deepStrictEqual(Array.from(graph.getAllEdges()), []);
+  });
 });
--- a/packages/core/graph/test/Graph.test.js
+++ b/packages/core/graph/test/Graph.test.js
@@ -20,13 +20,6 @@ describe('Graph', () => {
     assert.equal(graph.nodes.get(id), node);
   });
 
-  it("errors when removeNode is called with a node that doesn't belong", () => {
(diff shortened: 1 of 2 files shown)


packages/core/graph/test/Graph.test.js
```diff
@@ -317,4 +317,24 @@ describe('Graph', () => {
 
     assert.deepEqual(visited, [nodeA, nodeB, nodeD]);
   });
+
+  it('foo', () => {
```
wbinnssmith: placeholder? 😁

mischnic:
```suggestion
  it('correctly removes non-tree subgraphs', () => {
```

wbinnssmith (approved):

parcel-benchmark:
## Benchmark Results

#### Timings

| --- | --- | --- |

#### Cold Bundles

#### Cached Bundles

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 9.78s | +431.00ms |
| Cached | 475.00ms | -11.00ms |

#### Cold Bundles

*No bundle changes detected.*

#### Cached Bundles

*No bundle changes detected.*

</p></details>

#### Timings

| --- | --- | --- |

#### Cold Bundles

#### Cached Bundles

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 6.55s | -84.00ms |
| Cached | 300.00ms | +20.00ms ⚠️ |

#### Cold Bundles

*No bundle changes detected.*

#### Cached Bundles

| dist/Three.js | 579.69kb | +0.00b | 5.40s | +728.00ms ⚠️ |

</p></details>

**[Click here to view a detailed benchmark overview.](https://benchmark.parceljs.org/benchmark/329120273400857170)**

mischnic:
Now this tests fails where b is removed and d is deleted twice
undefined

I wonder if we should just make the assertions  in the methods less strict.

mischnic:
@lettertwo @thebriando What do you think about removing some of these assertions in `removeNode` and `removeEdge`? I think finding an ordering/algorithm that doesn't have "double deletes" will be more effort (certainly to write it, and possibly also computationally) that just ignoring these additional deletions.

lettertwo:
> What do you think about removing some of these assertions in `removeNode` and `removeEdge`?

I was wondering this, too. I think we already do silent no-ops for the inverse operations (`addNode` and `addEdge`), so symmetrically, it would make sense to me. I don't currently know of a reason why would need to assert that a node or edge exists before removing it. I imagine any case that hinges on the success or failure of removal could be satisfied by returning a success boolean instead.

thebriando:
> What do you think about removing some of these assertions in removeNode and removeEdge?

Yep that makes sense to me as well.

lettertwo:
I can't find a place where the result of `removeEdge` is handled. The only places I see where `removeNode` results are handled are in [these](https://github.com/parcel-bundler/parcel/blob/84847ed76aca867ce6f5d11f3d783672dc8088fe/packages/core/graph/test/Graph.test.js#L25-L27) [two](https://github.com/parcel-bundler/parcel/blob/84847ed76aca867ce6f5d11f3d783672dc8088fe/packages/core/graph/test/Graph.test.js#L264-L270) tests

lettertwo:
I did find this old PR: https://github.com/parcel-bundler/parcel/pull/3562, but I'm not sure if the context around it is still relevant. /cc @wbinnssmith

thebriando (approved):

lettertwo (approved):

PR #7938, opened 2022-04-12
Fix web extension schema

101arrowz:
# ↪️ Pull Request

Minor fix for the web extension schema that makes it throw when the required fields aren't present.

## 💻 Examples

`manifest.json`:

```json
{
  "gobble": "degook",
  "this": "should throw an error"
}
```

The above used to pass successfully through the transformer and now errors.

### DIFF
--- a/packages/transformers/webextension/src/schema.js
+++ b/packages/transformers/webextension/src/schema.js
@@ -445,6 +445,7 @@ export const MV3Schema = ({
       },
     },
   },
+  required: ['manifest_version', 'name', 'version'],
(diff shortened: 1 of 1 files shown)


101arrowz:
The nightly publish seems not to have worked for the past few days, so I haven't been able to test the manifest V3 support fully yet. P.S. why is the CI always broken?

mischnic (approved):

mischnic:
> why is the CI always broken?

We would have fixed it by now if we knew 😬

PR #7954, opened 2022-04-14
Feature: replace environmental variable with in binary expression

### LINKED ISSUE #7904: `import` from `process` left behind in browser-targeted output
# 🐛 bug report

I'm trying to build a "universal" ES6 library - meaning, it should be able to load in modern Node, modern browsers, and probably Deno too.

The addition of `browserslist` appears to resolve all `process.env` references statically - however, one reference looks like it was left behind, and the resulting module leads with:

    import {platform as $gh3jN$platform, env as $gh3jN$env} from "process";

Shouldn't there be an error or warning for this, when the bundler knows it's building for a browser context, and the resulting build still has a Node specific `import` left in it?

It appears the lingering `process` import comes from the `ansi-colors`, which [contains this](https://github.com/doowb/ansi-colors/blob/a4794363369d7b4d1872d248fc43a12761640d8e/index.js#L13-L15):

```js
  if ('FORCE_COLOR' in process.env) {
    colors.enabled = process.env.FORCE_COLOR !== '0';
  }
```

Perhaps the `in` operator is not supported by static analysis?

Or maybe the static analysis is getting confused by the presence of two nested tests?

My attempted `package.json` (and the project) is located [here](https://github.com/mindplay-dk/inertion/blob/web-compat/package.json).

## 🤔 Expected Behavior

Probably at least a warning or error, for either one or both of these:

1. Unable to remove all `process` references from a browser target.
2. Unable to handle a `process.env` reference via static analysis for any reason.

Ideally, static analysis should probably be enhanced to support this though? It looks like both `rollup` and `webpack` are capable of handling this case, so it would probably be best to have parity with that, as it's rather influential on the ecosystem already.

## 😯 Current Behavior

(see above)

## 💁 Possible ~~Solution~~ Workaround

I could submit a PR to `ansi-colors` and try to make it more compatible with Parcel's static analysis. (assuming it's possible to somehow rewrite that `process.env` check in a way that works the same?)

## 🔦 Context

(see project referenced above.)

## 💻 Code Sample

(see references above.)

## 🌍 Your Environment

| Parcel           | `parcel@2.4.1`, `@parcel/packager-ts@2.4.1`, `@parcel/transformer-typescript-types@2.4.1`
| Node             | `v16.14.2`
| npm/Yarn         | n/a
| Operating System | Ubuntu 16.04.7 LTS, (Release:        16.04, Codename:       xenial)

Shinyaigeek:
# ↪️ Pull Request

Fixes: #7904 

I implemented the feature to replace `in` binary expression ,whose right branch is `process.env` and left branch is string literal value, with boolean literal value.

## 💻 Examples

```javascript
"SOME_VAR" in process.env
```

into

```javascript
true // or false
```

## 🚨 Test instructions

prepare the test case to cover this.

## ✔️ PR Todo

### DIFF
--- a/packages/transformers/js/core/src/env_replacer.rs
+++ b/packages/transformers/js/core/src/env_replacer.rs
@@ -43,30 +43,24 @@ impl<'a> Fold for EnvReplacer<'a> {
 
     if let Expr::Bin(ref binary) = node {
       if let Expr::Member(ref member) = &*binary.right {
-        if let Expr::Lit(ref lit) = &*binary.left {
-          if let Lit::Str(ref left) = lit {
-            let obj = &member.obj;
-            match &**obj {
-              Expr::Ident(ref ident) => {
-                if &ident.sym == "process" {
-                  if let MemberProp::Ident(ref prop) = member.prop {
-                    if &prop.sym == "env" {
-                      if self.env.contains_key(&left.value) {
-                        return Expr::Lit(Lit::Bool(Bool {
-                          value: true,
-                          span: DUMMY_SP,
-                        }));
-                      } else {
-                        return Expr::Lit(Lit::Bool(Bool {
-                          value: false,
-                          span: DUMMY_SP,
-                        }));
-                      }
-                    }
+        if let Expr::Lit(Lit::Str(ref left)) = &*binary.left {
+          if let Expr::Ident(ref ident) = &*member.obj {
+            if &ident.sym == "process" {
+              if let MemberProp::Ident(ref prop) = member.prop {
+                if &prop.sym == "env" {
+                  if self.env.contains_key(&left.value) {
+                    return Expr::Lit(Lit::Bool(Bool {
+                      value: true,
+                      span: DUMMY_SP,
+                    }));
+                  } else {
+                    return Expr::Lit(Lit::Bool(Bool {
+                      value: false,
+                      span: DUMMY_SP,
+                    }));
                   }
                 }
               }
-              _ => {}
             }
           }
         }
--- a/packages/transformers/js/core/src/env_replacer.rs
+++ b/packages/transformers/js/core/src/env_replacer.rs
@@ -41,6 +41,38 @@ impl<'a> Fold for EnvReplacer<'a> {
       }
     }
 
+    if let Expr::Bin(ref binary) = node {
+      if let Expr::Member(ref member) = &*binary.right {
+        if let Expr::Lit(ref lit) = &*binary.left {
+          if let Lit::Str(ref left) = lit {
+            let obj = &member.obj;
+            match &**obj {
+              Expr::Ident(ref ident) => {
+                if &ident.sym == "process" {
+                  if let MemberProp::Ident(ref prop) = member.prop {
+                    if &prop.sym == "env" {
+                      if self.env.contains_key(&left.value) {
+                        return Expr::Lit(Lit::Bool(Bool {
+                          value: true,
+                          span: DUMMY_SP,
+                        }));
+                      } else {
+                        return Expr::Lit(Lit::Bool(Bool {
+                          value: false,
+                          span: DUMMY_SP,
+                        }));
+                      }
+                    }
+                  }
+                }
+              }
+              _ => {}
+            }
+          }
+        }
+      }
+    }
+
     if let Expr::Member(ref member) = node {
       if self.is_browser && match_member_expr(member, vec!["process", "browser"], self.decls) {
         return Expr::Lit(Lit::Bool(Bool {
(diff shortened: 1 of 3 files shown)


packages/transformers/js/core/src/env_replacer.rs
```diff
@@ -41,6 +41,38 @@ impl<'a> Fold for EnvReplacer<'a> {
       }
     }
 
+    if let Expr::Bin(ref binary) = node {
+      if let Expr::Member(ref member) = &*binary.right {
+        if let Expr::Lit(ref lit) = &*binary.left {
+          if let Lit::Str(ref left) = lit {
```
mischnic: I'm getting some clippy warnings that 
1. the match can be replaced with an `if let` because there's only one branch.
2. you can save one `if let` by doing `if let Expr::Lit(Lit::Str(ref left))`.

Shinyaigeek: reasonable!! Thank you! 👍  [c45bcb7](https://github.com/parcel-bundler/parcel/pull/7954/commits/c45bcb791b3c8afa4d39df1068d9a7e87fe93d33)

packages/transformers/js/core/src/env_replacer.rs
```diff
@@ -41,6 +41,32 @@ impl<'a> Fold for EnvReplacer<'a> {
       }
     }
 
+    if let Expr::Bin(ref binary) = node {
+      if let Expr::Member(ref member) = &*binary.right {
+        if let Expr::Lit(Lit::Str(ref left)) = &*binary.left {
+          if let Expr::Ident(ref ident) = &*member.obj {
+            if &ident.sym == "process" {
```
mischnic: This also matches if there is a local variable called `process`.
So you need to do something along the lines of `if !decls.contains(ident)`. You find other places places where this is done by searching for `decls.contains`.

Shinyaigeek: I see. sorry, I did not notice this. Indeed, this process is done in `typeof_replacer`. [c8ba4ba](https://github.com/parcel-bundler/parcel/pull/7954/commits/c8ba4baf8e61237d2aba6cccfdfba1de382da455) 👍

packages/transformers/js/core/src/env_replacer.rs
```diff
@@ -41,6 +42,32 @@ impl<'a> Fold for EnvReplacer<'a> {
...
+                    return Expr::Lit(Lit::Bool(Bool {
+                      value: true,
+                      span: DUMMY_SP,
+                    }));
+                  } else {
+                    return Expr::Lit(Lit::Bool(Bool {
+                      value: false,
+                      span: DUMMY_SP,
+                    }));
+                  }
```
mischnic: You can remove the if by doing
```rust
					return Expr::Lit(Lit::Bool(Bool {
                      value: self.env.contains_key(&left.value),
                      span: DUMMY_SP,
                    }));
```

Shinyaigeek: Certainly! Thanks! 👍  9b66c82

packages/transformers/js/core/src/env_replacer.rs
```diff
@@ -41,6 +42,25 @@ impl<'a> Fold for EnvReplacer<'a> {
       }
     }
 
+    if let Expr::Bin(ref binary) = node {
+      if let Expr::Member(ref member) = &*binary.right {
+        if let Expr::Lit(Lit::Str(ref left)) = &*binary.left {
+          if let Expr::Ident(ref ident) = &*member.obj {
+            if !self.decls.contains(&id!(ident)) && &ident.sym == "process" {
+              if let MemberProp::Ident(ref prop) = member.prop {
+                if &prop.sym == "env" {
```
devongovett: Maybe we should use the `match_member_expr` utility to match `process.env` as below?

packages/transformers/js/core/src/env_replacer.rs
```diff
@@ -41,6 +42,25 @@ impl<'a> Fold for EnvReplacer<'a> {
       }
     }
 
+    if let Expr::Bin(ref binary) = node {
```
devongovett: We should probably check that the binary expression is actually an `in`. There are other binary expression operators that shouldn't be replaced.

devongovett (changes requested):

devongovett (approved):
Thanks! I did a little simplification to the code, hope you don't mind. 😄

Shinyaigeek:
I misunderstood that `match_member_expr` cannot be used in this usecase... Thanks!

PR #7969, opened 2022-04-18
web-ext | Add #schema key for json schema autocompletion

### LINKED ISSUE #7970: web-ext | Allow #schema property in manifest.json
# 🙋 feature request

When working with json file, having a `#schema` property can be very helpful to provide intellisense and autocompletion

## 🤔 Expected Behavior

When working with `manifest.json` file with web-ext, adding a `#schema` property pointing to `https://json.schemastore.org/chrome-manifest` would allow developer to have intellisense and autocompletion in editor supporting it.

## 😯 Current Behavior

web-ext schema bailed out saying the manifest.json is invalid when this property is defined

## 🔦 Context

Working with the `manifest.json` file (or any file for that matter) without any intelisense suck 😢

louisgv:
# ↪️ Pull Request

closes #7970

By allowing a `#schema` key in the manifest file, a developer can enjoy intelligence/autocompletion working with it.

## 💻 Examples

The schema file location for chrome extension manifest is as follow:

```json
  "$schema": "https://json.schemastore.org/chrome-manifest",
```

## 🚨 Test instructions

With nightly build:
- Setup a v3 manifest.json file
- Add the schema url like above, and test out the intellisense on vscode or any editor supporting json schema

## ✔️ PR Todo

### DIFF
--- a/packages/transformers/webextension/src/schema.js
+++ b/packages/transformers/webextension/src/schema.js
@@ -76,6 +76,7 @@ const warBase = {
 };
 
 const commonProps = {
+  "$schema": string,
   name: string,
   version: {
     type: 'string',


devongovett (approved):
Seems ok to me. @101arrowz?

101arrowz:
Won't cause an error in web extensions AFAIK but I feel like this is something that should be addressed outside of modifying the actual JSON. The same argument around `$schema` could be made for `.parcelrc` for instance, but I don't think it's supported. Nonetheless it's probably helpful for people.

101arrowz:
I just Googled this issue and [modifying user settings in VSCode](https://code.visualstudio.com/docs/languages/json#_mapping-in-the-user-settings) is a far better way of supporting this. I would advise against merging this PR, and instead directing people to modify their settings to match `manifest.json` with this schema (maybe we could add a note in the docs). Similar features exist [in other IDEs](https://www.jetbrains.com/help/idea/json.html#ws_json_using_schemas).

Maybe there could be a VSCode plugin for Parcel to do this kind of thing automatically. That would be pretty cool.

louisgv:
> I just Googled this issue and [modifying user settings in VSCode](https://code.visualstudio.com/docs/languages/json#_mapping-in-the-user-settings) is a far better way of supporting this. 

I politely disagree. The reason is that there are actually more than one `manifest.json` schema:

1. PWA manifest: https://web.dev/add-manifest/ 
2. AWS manifest: https://docs.aws.amazon.com/redshift/latest/dg/load-from-host-steps-create-manifest.html (many people just name it manifest.json xd....)

If a user configures `manifest.json` to point globally to just the web extension manifest schema, then when they need to work with other schemas, they would have to manually override the config. There is a way for vscode to specify a local config per project for the file association, but in a team where there are multiple editors, that setup is cumbersome. 

Furthermore, when the user have both a manifest for PWA and the manifest for web extension under the same monorepo (this is the case for new tab extension for example, where they are both a PWA app, and an extension) - they would have to config their setting to point to the specific path. I think the overall dev experience to config this would be quite annoying, esp when their team use different OS (windows like `\` path separator, vs Unix's `/`)...

> Maybe there could be a VSCode plugin for Parcel to do this kind of thing automatically. That would be pretty cool.

VSCode itself is actually aware of a couple json schema for `package.json` and `tsconfig`. I do think that'd be cool, but I believe that's a bit out of scope for this tiny change :p...

Also, I'd think that the manifest produced after the transform step would simply strip out this property anyway? (Or do I need to tweak the transform code to delete this key?)

101arrowz:
You bring up good points, I'll send a patch for this PR to remove $schema in a bit. There's a few other minor issues to address as well.

101arrowz:
It ended up being impossible to submit the patch so I made a new PR based off your branch. I think I would need to be a collaborator/member of Parcel to contribute directly.

louisgv:
> It ended up being impossible to submit the patch so I made a new PR based off your branch. I think I would need to be a collaborator/member of Parcel to contribute directly.

Awesome, I will just close this PR 👍

PR #7971, opened 2022-04-18
Fix `web_accessible_resources` and HMR

### LINKED ISSUE #7953: Web Extension Transformer: a broken `web_accessible_resources` field is added
- Follows https://github.com/parcel-bundler/parcel/pull/7050

# 🐛 bug report

When building a web extension, it seems that the hosts specified in the content scripts are copied to the `web_accessible_resources` field, even if unnecessary, breaking the manifest altogether in some cases.

No custom configuration. It happens on both build and watch.

## 🤔 Expected Behavior

No `web_accessible_resources` unless required:

```json
{
	"name": "Awesome Extension",
	"version": "0.0.0",
	"manifest_version": 3,
	"content_scripts": [
		{
			"matches": ["https://github.com/fregante/browser-extension-template/*"],
			"css": ["manifest.b7514168.css"]
		}
	]
}
```

or at least use a supported origin glob:

```json
{
	"name": "Awesome Extension",
	"version": "0.0.0",
	"manifest_version": 3,
	"content_scripts": [
		{
			"matches": ["https://github.com/fregante/browser-extension-template/*"],
			"css": ["manifest.b7514168.css"]
		}
	],
	"web_accessible_resources": [
		{
			"matches": ["https://github.com/*"],
			"extension_ids": [],
			"resources": []
		}
	]
}
```
## 😯 Current Behavior

```json
{
	"name": "Awesome Extension",
	"version": "0.0.0",
	"manifest_version": 3,
	"content_scripts": [
		{
			"matches": ["https://github.com/fregante/browser-extension-template/*"],
			"css": ["manifest.b7514168.css"]
		}
	],
	"web_accessible_resources": [
		{
			"matches": ["https://github.com/fregante/browser-extension-template/*"],
			"extension_ids": [],
			"resources": []
		}
	]
}
```

## 💻 Code Sample

Input manifest:

```json
{
	"name": "Awesome Extension",
	"version": "0.0.0",
	"manifest_version": 3,
	"content_scripts": [
		{
			"matches": [ "https://github.com/fregante/browser-extension-template/*" ],
			"css": [ "content.css" ]
		}
	]
}
```

## 🌍 Your Environment

| Parcel           | 2.0.0-nightly.1049
| parcel/config-webextension | 2.4.2-nightly.2674 |
| Node             | v16.4.2
| npm/Yarn         | 7.18.1
| Operating System | macOS 11

101arrowz:
# ↪️ Pull Request

Actually fixes #7953

This actually, really fixes `web_accessible_resources` generation for web extensions and HMR by using a runtime as a cache invalidator for the packager based on the state of the bundle graph. This is probably bad practice compared to just doing the generation from the runtime, but since JSON assets are automatically transformed to JS it's not possible to actually consume the result of the transformer in the packager without eval.

### DIFF
--- a/packages/packagers/webextension/src/WebExtensionPackager.js
+++ b/packages/packagers/webextension/src/WebExtensionPackager.js
@@ -42,22 +42,26 @@ export default (new Packager({
         ),
       ];
 
-      war.push({
-        matches: contentScript.matches,
-        extension_ids: [],
-        resources: jsBundles
-          .flatMap(b => {
-            const children = [];
-            const siblings = bundleGraph.getReferencedBundles(b);
-            bundleGraph.traverseBundles(child => {
-              if (b !== child && !siblings.includes(child)) {
-                children.push(child);
-              }
-            }, b);
-            return children;
-          })
-          .map(relPath),
-      });
+      const resources = jsBundles
+        .flatMap(b => {
+          const children = [];
+          const siblings = bundleGraph.getReferencedBundles(b);
+          bundleGraph.traverseBundles(child => {
+            if (b !== child && !siblings.includes(child)) {
+              children.push(child);
+            }
+          }, b);
+          return children;
+        })
+        .map(relPath);
+
+      if (resources.length > 0) {
+        // In the future, maybe use "matches" as well
+        war.push({
+          extension_ids: [],
+          resources,
+        });
+      }
     }
     manifest.web_accessible_resources = (
       manifest.web_accessible_resources || []
--- a/packages/packagers/webextension/src/WebExtensionPackager.js
+++ b/packages/packagers/webextension/src/WebExtensionPackager.js
@@ -11,14 +11,13 @@ export default (new Packager({
     bundle.traverseAssets(asset => {
       assets.push(asset);
     });
+    const manifestAssets = assets.filter(a => a.meta.webextEntry === true);
 
-    assert.equal(
-      assets.length,
-      1,
-      'Web extension manifest bundles must only contain one asset',
+    assert(
+      assets.length == 2 && manifestAssets.length == 1,
+      'Web extension bundles must contain exactly one manifest asset and one runtime asset',
     );
-    const asset = assets[0];
-    assert(asset.meta.webextEntry === true);
+    const asset = manifestAssets[0];
 
     const relPath = b =>
       relativeBundlePath(bundle, b, {leadingDotSlash: false});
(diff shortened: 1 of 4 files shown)


devongovett:
Looks like a couple tests are failing.

packages/packagers/webextension/src/WebExtensionPackager.js
```diff
@@ -56,9 +58,16 @@ export default (new Packager({
...
-          extension_ids: [],
+          matches: contentScript.matches.map(match => {
+            if (/^(((http|ws)s?)|ftp|\*):\/\//.test(match)) {
+              let pathIndex = match.indexOf('/', match.indexOf('://') + 3);
+              // Avoids creating additional errors in invalid match URLs
+              if (pathIndex == -1) pathIndex = match.length;
+              return match.slice(0, pathIndex) + '/*';
+            }
+            return match;
+          }),
```
devongovett, line 70: So if I'm reading this right, `"https://github.com/fregante/browser-extension-template/*"` would be transformed to `"https://github.com/*"`? What's the reason for that? Should we add a test?

We could also consider using a URL parser for this:

```js
  return new URL(match).origin + '/*'
```

packages/runtimes/webextension/src/WebExtensionRuntime.js
```diff
@@ -12,39 +12,50 @@ const AUTORELOAD_BG = fs.readFileSync(
...
+      const descendants = [];
+      bundleGraph.traverseBundles(b => {
+        descendants.push(b.id);
+      }, bundle);
       return {
         filePath: __filename,
-        code: AUTORELOAD_BG,
+        code: JSON.stringify(descendants),
         isEntry: true,
       };
```
devongovett, line 31: Nice hack. This is ok for now. We'll think about a better way to do this in core.

packages/packagers/webextension/src/WebExtensionPackager.js
101arrowz: In web extensions `web_accessible_resources` will error if the path is not a wildcard. We can't use the `new URL` constructor because there's a bunch of special syntax here like `*://something*.com/*` that doesn't get parsed properly by `new URL` but is covered by this logic.

Now in theory we could literally replace all this with `["<all_urls>"]` and it would work the same, but that makes fingerprinting by bad actors much easier.

101arrowz:
Not sure why the tests are failing now, it seems to be unrelated to the changes in this PR but clearly it's not. Investigating now...

EDIT: Actually the issue was me not removing the broken `priority: 'parallel'` logic to try to make the packager reload. Using a runtime seems to be the only way to update on bundle graph change rather than bundle change.

devongovett (approved):

fregante:
<details>Is this truncating `web_accessible_resources[*].resources` in every case? I have `resources: ['*.js']` in it and it's outputting `resources: []`

Same goes for for mv2: `web_accessible_resources: ['*.js']` is becoming `web_accessible_resources: []`</details>

Edit: false alarm, the array was being trimmed because there were no `*.js` files, I needed to use `../*.js`

PR #8000, opened 2022-04-23
Fix web extension issues

### LINKED ISSUE #7997: WebExtension manifestV3 doesn't work as expected
# 🐛 bug report

Using parcel to build webextension with manifest v3 causing few issues:
1. web_accessible_resources are not storing initial resource name if it is already used as an action source
2. all the sources bundled into root dir without any dir paths
3. manifest schema doesn't allow any additional props ([like `key` for example](https://developer.chrome.com/docs/extensions/mv2/manifest/key/))

default configuration

no babel config

## 🤔 Expected Behavior

1. Storing initial web_accessible_resources names to make them accessible
2. Preserving paths to sources in resulting bundle
3. Adding key prop, or allowing additional props in manifest schema

## 💻 WebExtension Manifest Sample
```json
{
  "name": "Test",
  "version": "0.0.0",
  "description": "Test manifest",
  "manifest_version": 3,
  "action": {
    "default_popup": "action.html"
  },
    "resources": ["action.html"],
    "matches": ["<all_urls>"]
  }]
}
```
results in 
```json
{
  "name":"Test",
  "version":"0.0.0",
  "description":"Test manifest",
  "manifest_version":3,
  "action":{
    "default_popup":"action.e0bbec54.html"
  },
  "web_accessible_resources":[{
    "resources":["action.e0bbec54.html"],
    "matches":["<all_urls>"]
  }]}
```
reusing compiled resource without preserving it's initial name is wrong for web_accessible_resources

## 🌍 Your Environment

| Parcel           | 2.5.0.
| Node             | 16
| npm/Yarn         |
| Operating System |

101arrowz:
# ↪️ Pull Request

Fixes #7997 

There were some issues with the web extension schema that this PR fixes. However, the main change is making the dependency merging logic more robust, which fixes an issue where two dependencies for an asset added as `needsStableName: true` and `needsStableName: false` but otherwise identical would sometimes result in a bundle with `needsStableName: false`. This was because the merging logic would simply update the first dependency's options with the newer dependency's options, when in reality the merged dependency should have a stable name if either source dependency needed it.

I also applied this logic to all the other fields that weren't factored into the dependency ID. For instance, if the dependency is optional in one dependency but not the other, the merged dependency should not be optional. The only field I wasn't sure about was `bundleBehavior`; I'm just guessing that the precedence should be `'isolated'` > `'inline'` > `null`. IMO `bundleBehavior` should actually be factored into the dependency ID but I wasn't sure so I didn't change that.

## 💻 Examples

`manifest.json`:
```json
{
{
    "manifest_version": 3,
    "version": "1",
    "name": "test",
    "action": {
      "default_popup": "popup.html"
    },
        "matches": ["<all_urls>"],
        "resources": ["popup.html"]
    }]
}
```

Previously this manifest resulted in the `popup.html` file receiving a content hash, which is undesired behavior because `web_accessible_resources` aren't really that accessible if their names aren't constant. (If `popup.html` were only present in `web_accessible_resources` and not in `action`, however, it would work)

This inconsistency is now fixed so if either dependency needs a stable name, the final merged dependency does too.


101arrowz:
Looks like the content hashing tests are failing due to the changes in `WriteBundlesRequest.js`. Not sure how to work around that...

EDIT: Found and fixed the issue, should be working now.

parcel-benchmark:
## Benchmark Results

#### Timings

| --- | --- | --- |

#### Cold Bundles

#### Cached Bundles

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 9.27s | -42.00ms |
| Cached | 439.00ms | -10.00ms |

#### Cold Bundles

*No bundle changes detected.*

#### Cached Bundles

*No bundle changes detected.*

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 1.67m | +1.51s |
| Cached | 2.58s | -34.00ms |

#### Cold Bundles

| dist/index.5c7ce5ea.js | 3.33mb | +144.00b ⚠️ | 32.37s | +678.00ms |

#### Cached Bundles

| dist/index.4ac13050.js | 3.33mb | +144.00b ⚠️ | 31.88s | -155.00ms |

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 6.75s | +79.00ms |
| Cached | 286.00ms | +3.00ms |

#### Cold Bundles

*No bundle changes detected.*

#### Cached Bundles

*No bundle changes detected.*

</p></details>

**[Click here to view a detailed benchmark overview.](https://benchmark.parceljs.org/benchmark/332415415059743315)**

packages/core/core/src/requests/WriteBundleRequest.js
```diff
@@ -66,11 +66,12 @@ export type WriteBundleRequest = {|
 export default function createWriteBundleRequest(
   input: WriteBundleRequestInput,
 ): WriteBundleRequest {
+  let name = nullthrows(input.bundle.name);
   let nameHash = nullthrows(
     input.hashRefToNameHash.get(input.bundle.hashReference),
   );
   return {
-    id: `${input.bundle.id}:${input.info.hash}:${nameHash}`,
+    id: `${input.bundle.id}:${input.info.hash}:${nameHash}:${name}`,
```
devongovett, line 74: What was the reason for this change?  When does a bundle change name where its hash doesn't change?

packages/core/core/src/Dependency.js
```diff
@@ -85,12 +85,29 @@ export function createDependency(
...
     }
   }
+  if (needsStableName) a.needsStableName = true;
+  if (isEntry) a.isEntry = true;
+  if (!isOptional) a.isOptional = false;
+  if (
+    bundleBehavior == BundleBehavior['isolated'] ||
+    a.bundleBehavior == null
+  ) {
+    a.bundleBehavior = bundleBehavior;
```
devongovett: @wbinnssmith this has implications for the bundling issue we were discussing. If we have two dependencies on the same asset, one isolated and one not, would we want two different bundles or should they be merged?

packages/core/core/src/Dependency.js
```diff
@@ -85,12 +85,29 @@ export function createDependency(
...
   if (a.symbols && symbols) {
     for (let [k, v] of symbols) {
       a.symbols.set(k, v);
     }
   }
+  if (needsStableName) a.needsStableName = true;
+  if (isEntry) a.isEntry = true;
+  if (!isOptional) a.isOptional = false;
+  if (
+    bundleBehavior == BundleBehavior['isolated'] ||
```
devongovett: Probably shouldn't be specific to isolated, inline would have the same issue. But I'm thinking this should probably produce two different dependencies (i.e. be included in the dependency id), and we should let the bundler handle it just like if the dep were in two different files.

packages/core/core/src/requests/WriteBundleRequest.js
101arrowz: This happens for bundles that have their `needsStableName` field changed.

packages/core/core/src/Dependency.js
101arrowz: Actually this semi-implicitly handles the inline case. If the first dependency's bundle behavior is `null` (i.e. unspecified), we always use the second one (so if the second one is inline, that case is covered). If the second dependency requests an isolated bundle explicitly, we override the first dependency even if it was inline (i.e. isolated takes priority over inline). Lastly, if the first dependency was already inline and the second one was inline or unspecified, we leave it as is.

101arrowz: Overall though I agree we should factor this into the dependency ID. I really doubt inline and isolated dependencies should be merged (this may have implications for #8052 though).

packages/core/core/src/Dependency.js
wbinnssmith: I think there are arguments for each: one would produce fewer overall bundles and there would be one canonical bundle for a given entry asset, and the other would produce multiple bundles, but each with its own optimal set of assets (non-isolated bundles can rely on existing loaded assets).

wbinnssmith: I think we could understand what is "best" with real-world tests.

packages/core/core/src/Dependency.js
devongovett: Discussed this in our core team meeting. I think we should update this to produce two dependencies (i.e. include bundleBehavior in the dependency id). The bundler is being updated in #8052 to handle this case better, but I think that should be unnecessary for your case.

101arrowz: Fixed.

devongovett (approved):

PR #8026, opened 2022-04-29
Feature(transformer): re-implement postProcess API

Shinyaigeek:
# ↪️ Pull Request

Part of #7503 

`postProcess` API was removed in #5674 because this API was unused, but this API is necessary for svelte preprocessing, so I start by re-implement this `postProcess` API in order to accomplish no-config svelte support.

## 🚨 Test instructions

Transformer’s API, such as transformer’s generate method, does not seems to be tested directly and a testing method does not seem to be established. In addition to these reasons, I thought Transformer’s API is tested indirectly by much of integration-test, so I did not write any unit-test about this API, but I think it would be better to write the unit-test about `postProcess`, so I will write a test if you want.

## ✔️ PR Todo


packages/core/types/index.js
```diff
@@ -186,7 +186,7 @@ export type EnvironmentOptions = {|
  */
 export type VersionMap = {
   [string]: string,
-  ...,
+  ...
```
Shinyaigeek, line 189: seems to be a change by the formatter, which runs pre-commit

packages/core/types/index.js
```diff
@@ -1045,6 +1041,14 @@ export type Transformer<ConfigType> = {|
     options: PluginOptions,
     logger: PluginLogger,
   |}): Async<Array<TransformerResult | MutableAsset>>,
+  /** Do some processing after the transformation  */
+  postProcess?: ({|
+    assets: Array<MutableAsset>,
+    config: ConfigType,
```
Shinyaigeek, line 1050: other than `config` type on this line and some comments and some invalidation logic are completely same to https://github.com/parcel-bundler/parcel/pull/5674/files

devongovett:
Once we have something that uses this, we should add some caching tests for this. One possible issue is if a transformer plugin returns a different list of assets from postProcess, the invalidations/dependencies in the original assets will be lost resulting in the cache/watcher not being invalidated on changes. Not sure if we should/can do this automatically in core, or whether we need to leave that to plugins. We'll see as this progresses further.

packages/core/types/index.js
```diff
@@ -1045,6 +1041,14 @@ export type Transformer<ConfigType> = {|
     options: PluginOptions,
     logger: PluginLogger,
   |}): Async<Array<TransformerResult | MutableAsset>>,
+  /** Do some processing after the transformation  */
```
devongovett: Can we mark this as `@experimental` since it still has some work left to be done on it? That should show up in the generated docs (though we'll have to double check).

mischnic:
```suggestion
  /**
   * Do some processing after the transformation
   * @experimental
   */
```

will look like this after https://github.com/parcel-bundler/website/pull/1026:

Bildschirmfoto 2022-05-08 um 18 21 02

Shinyaigeek: @mischnic Thanks! I applied your commit suggestion! 👍

PR #8031, opened 2022-04-30
Allow to define the "compilerOptions" of the VueTransformer

### LINKED ISSUE #8030: Vue.js transformer is missing the "isCustomElement" option
# 🙋 Feature request

When using custom elements within Vue.js templates (i.e elements that aren't Vue.js components nor native HTML elements), Vue emits a warning. The compiler (https://github.com/vuejs/core/tree/main/packages/compiler-sfc) has an option to work around that, allowing to specify which elements are custom and thus should be ignored.

## 💁 Possible Solution

It should be feasible, somewhere in the Parcel configuration, to define the `isCustomElement` option of the SFC compiler.

## 🔦 Context

Using custom elements isn't an edge case. E.g : web components, [MJML](https://mjml.io) components, etc...

abellion:
# ↪️ Pull Request

Fixes https://github.com/parcel-bundler/parcel/issues/8030


101arrowz:
Many of the compiler options can't be made in pure JSON, which means you need a JS config file. This includes `isCustomElement`. As a result caching will suffer substantially. Nonetheless this is probably a worthwhile change.

We might want to look into ways to declare invalidators within config files (i.e. an exported `invalidateConfig` function that prevents busting the cache when nothing relevant changes).

abellion:
Yep, true. However, as the Vue config can also come from JS files (`.vuerc.js`, `vue.config.js`), I think people won't try to configure what's not configurable though JSON (e.g. the `isCustomElement` setting).

101arrowz (approved):

KaKi87:
Hello,

> the Vue config can also come from JS files (`.vuerc.js`, `vue.config.js`), I think people won't try to configure what's not configurable though JSON (e.g. the `isCustomElement` setting).

How to make Parcel handle either of those files using the programmatic API (`await new Parcel(...).watch()` for dev & `await new Parcel(...).run()` for prod) ?

Thanks

PR #5316, opened 2020-11-04
HMR Sourcemaps

mischnic:
# ↪️ Pull Request

Closes #643

PR #8034, opened 2022-05-01
Add React error overlay and HMR source maps

### LINKED ISSUE #643: Bug: Source maps don't work with hot reload
In the console I get an output like `index.jsx:6 "test1"` and that's okay.
But when I save the file, it turns like `VM335:18 "test2"`.
I have to refresh the page to see this one `index.jsx:6 "test2"`.

So source maps don't work unless I refresh the page.

| Parcel           | 1.5.0
| Node             | 9.2.0
| npm/Yarn         | 5.6.0/1.3.2
| Operating System | MacOS 10.13.2

### LINKED ISSUE #7604: React Error Overlay
# 🙋 feature request

An error overlay for runtime errors like CRA has.

Not sure if this should also be enabled (or if it even works at all) when not using React.

## 💁 Possible Solution

I think we can reuse the `react-error-overlay` package from CRA

https://github.com/facebook/create-react-app/blob/a422bf227cf5294a34d68696664e9568a152fd8f/packages/react-dev-utils/webpackHotDevClient.js#L45

devongovett:
Closes #7604. Closes #643. Closes #5316.

image

This adds an overlay that appears when there is a runtime error in a React app, using the [react-error-overlay](https://github.com/facebook/create-react-app/tree/a422bf227cf5294a34d68696664e9568a152fd8f/packages/react-error-overlay) package from Create React App. It is implemented as part of `@parcel/runtime-react-refresh` since it is React-specific, and that seemed like the easiest place to put it. It will appear when a runtime error occurs, and is automatically dismissed when an HMR update comes in.

This required support for source maps in HMR, which has been a longstanding issue due to our use of `new Function`, which has poor support for source maps. Now, we use `eval` instead, which has better support in Chrome and Firefox. However, Safari still doesn't support `sourceMappingURL` or `sourceURL` in `eval` error stack traces, so we detect this case and fall back to loading the asset from a new dev server endpoint that serves asset contents by id.

As a side benefit, HMR should now fall back to loading URLs for changed assets via HTTP if `eval` is not allowed, e.g. with a strict CSP. This may help web extension v3 support, though I have not tested this. (cc. @101arrowz)

The error overlay also supports clicking the code frame to open it directly in your editor, which I also added to the standard parcel build error overlay and the error 500 page. This is implemented through a second new dev server endpoint which auto-detects which editor you're using, and opens it to the line and column of the error message.

Finally, the error 500 page has been improved to automatically reload when an HMR update comes in so you don't need to manually do that.

### DIFF
--- a/packages/core/utils/src/prettyDiagnostic.js
+++ b/packages/core/utils/src/prettyDiagnostic.js
@@ -10,10 +10,18 @@ import nullthrows from 'nullthrows';
 // $FlowFixMe
 import terminalLink from 'terminal-link';
 
+export type FormattedCodeFrame = {|
+  location: string,
+  code: string,
+|};
+
 export type AnsiDiagnosticResult = {|
   message: string,
   stack: string,
+  /** A formatted string containing all code frames, including their file locations. */
   codeframe: string,
+  /** A list of code frames with highlighted code and file locations separately. */
+  frames: Array<FormattedCodeFrame>,
   hints: Array<string>,
   documentation: string,
 |};
@@ -39,6 +47,7 @@ export default async function prettyDiagnostic(
       (skipFormatting ? message : mdAnsi(message)),
     stack: '',
     codeframe: '',
+    frames: [],
     hints: [],
     documentation: '',
   };
@@ -69,16 +78,20 @@ export default async function prettyDiagnostic(
         });
       }
 
-      result.codeframe +=
+      let location =
         typeof filePath !== 'string'
           ? ''
-          : chalk.gray.underline(
-              `${filePath}:${highlights[0].start.line}:${highlights[0].start.column}\n`,
-            );
+          : `${filePath}:${highlights[0].start.line}:${highlights[0].start.column}`;
+      result.codeframe += location ? chalk.gray.underline(location) + '\n' : '';
       result.codeframe += formattedCodeFrame;
       if (codeFrame !== codeFrames[codeFrames.length - 1]) {
         result.codeframe += '\n\n';
       }
+
+      result.frames.push({
+        location,
+        code: formattedCodeFrame,
+      });
     }
   }
 
--- a/packages/reporters/dev-server/src/HMRServer.js
+++ b/packages/reporters/dev-server/src/HMRServer.js
@@ -35,7 +35,7 @@ export type HMRMessage =
       type: 'error',
       diagnostics: {|
         ansi: Array<AnsiDiagnosticResult>,
-        html: Array<AnsiDiagnosticResult>,
+        html: Array<$Rest<AnsiDiagnosticResult, {|codeframe: string|}>>,
       |},
     |};
 
@@ -90,7 +90,10 @@ export default class HMRServer {
           return {
             message: ansiHtml(d.message),
             stack: ansiHtml(d.stack),
-            codeframe: ansiHtml(d.codeframe),
+            frames: d.frames.map(f => ({
+              location: f.location,
+              code: ansiHtml(f.code),
+            })),
             hints: d.hints.map(hint => ansiHtml(hint)),
             documentation: diagnostics[i].documentationURL ?? '',
           };
--- a/packages/reporters/dev-server/src/HMRServer.js
+++ b/packages/reporters/dev-server/src/HMRServer.js
@@ -1,6 +1,13 @@
 // @flow
 
-import type {BuildSuccessEvent, Dependency, PluginOptions} from '@parcel/types';
+import type {
+  BuildSuccessEvent,
+  Dependency,
+  PluginOptions,
+  BundleGraph,
+  PackagedBundle,
+  Asset,
+} from '@parcel/types';
 import type {Diagnostic} from '@parcel/diagnostic';
 import type {AnsiDiagnosticResult} from '@parcel/utils';
 import type {ServerError, HMRServerOptions} from './types.js.flow';
@@ -8,9 +15,11 @@ import type {ServerError, HMRServerOptions} from './types.js.flow';
 import WebSocket from 'ws';
 import invariant from 'assert';
 import {ansiHtml, prettyDiagnostic, PromiseQueue} from '@parcel/utils';
+import {HMR_ENDPOINT} from './Server';
 
 export type HMRAsset = {|
   id: string,
+  url: string,
   type: string,
   output: string,
   envHash: string,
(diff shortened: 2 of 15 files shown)


packages/core/utils/src/prettyDiagnostic.js
```diff
@@ -10,10 +10,18 @@ import nullthrows from 'nullthrows';
...
+  code: string,
+|};
+
 export type AnsiDiagnosticResult = {|
   message: string,
   stack: string,
+  /** A formatted string containing all code frames, including their file locations. */
   codeframe: string,
+  /** A list of code frames with highlighted code and file locations separately. */
+  frames: Array<FormattedCodeFrame>,
```
devongovett, line 24: A bit unfortunate, but couldn't think of a better way...

packages/reporters/dev-server/src/HMRServer.js
```diff
@@ -185,3 +201,34 @@ function getSpecifier(dep: Dependency): string {
...
+  if (asset.type === 'js') {
+    let publicId = bundleGraph.getAssetPublicId(asset);
+    output = `parcelHotUpdate['${publicId}'] = function (require, module, exports) {${output}}`;
+  }
+
+  let sourcemap = await asset.getMap();
+  if (sourcemap) {
+    let sourcemapStringified = await sourcemap.stringify({
+      format: 'inline',
+      sourceRoot: '/__parcel_source_root/',
```
devongovett, line 219: Somewhat related, but this source root seems to be displayed by react-error-overlay rather than either the original file path, or nothing. Not sure if there is a way to improve that?

101arrowz:
Yes, this should help manifest V3 hot reload. As long as the HMR server is on some kind of localhost domain, dynamic script loading should work fine via a similar CSP patching technique to MV2. I'll look into it more after this is merged.

packages/runtimes/hmr/src/loaders/hmr-runtime.js
```diff
@@ -92,8 +92,18 @@ if ((!parent || !parent.isParcelRequire) && typeof WebSocket !== 'undefined') {
...
+  var supportsSourceURL = false;
+  try {
+    (0, eval)('throw new Error("test"); //# sourceURL=test.js');
+  } catch (err) {
+    supportsSourceURL = err.stack.includes('test.js');
+  }
+
   // $FlowFixMe
-  ws.onmessage = function (event /*: {data: string, ...} */) {
+  ws.onmessage = async function (event /*: {data: string, ...} */) {
```
101arrowz, line 106: Does using `async/await` have any consequences for developing in old browsers? AFAICT most of our other runtimes are ES5 syntactically.

packages/runtimes/hmr/src/loaders/hmr-runtime.js
```diff
@@ -120,9 +130,15 @@ if ((!parent || !parent.isParcelRequire) && typeof WebSocket !== 'undefined') {
...
 
-        assets.forEach(function (asset) {
-          hmrApply(module.bundle.root, asset);
-        });
+        // Dispatch custom event so other runtimes (e.g React Refresh) are aware.
+        if (
+          typeof window !== 'undefined' &&
+          typeof CustomEvent !== 'undefined'
+        ) {
+          window.dispatchEvent(new CustomEvent('parcelhmraccept'));
```
101arrowz, line 138: Minor thing but you might instead use `parcel-hmr-accept` in styling convention with DOM events. Also since you're already using global state you could also a well-known symbol to `global` containing a `registerAcceptor` function or something to avoid polluting user code.

packages/runtimes/hmr/src/loaders/hmr-runtime.js
devongovett: It should get transpiled I believe.

parcel-benchmark:
## Benchmark Results

#### Timings

| --- | --- | --- |

#### Cold Bundles

#### Cached Bundles

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 8.79s | +58.00ms |
| Cached | 445.00ms | +14.00ms |

#### Cold Bundles

*No bundle changes detected.*

#### Cached Bundles

| dist/logo.c5bb83f1.png | 246.00b | +0.00b | 3.84s | -581.00ms 🚀 |

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 1.50m | -666.00ms |
| Cached | 2.41s | -92.00ms |

#### Cold Bundles

| dist/index.0605f25b.js | 3.32mb | +144.00b ⚠️ | 29.70s | -349.00ms |
| dist/16.069344b7.js | 905.00b | +0.00b | 33.73s | -3.84s 🚀 |
| dist/simpleHasher.46d6f2e5.js | 742.00b | +0.00b | 37.29s | -31.24s 🚀 |
| dist/index.html | 240.00b | +0.00b | 29.71s | -3.31s 🚀 |

#### Cached Bundles

| dist/esm.6190c9ac.js | 61.54kb | +0.00b | 1.13m | +30.73s ⚠️ |
| dist/workerHasher.e50d242f.js | 1.72kb | +0.00b | 1.13m | +30.73s ⚠️ |
| dist/16.1969624f.js | 1.08kb | +0.00b | 36.72s | +3.07s ⚠️ |
| dist/16.069344b7.js | 905.00b | +0.00b | 36.72s | +3.07s ⚠️ |
| dist/simpleHasher.46d6f2e5.js | 742.00b | +0.00b | 1.13m | +30.73s ⚠️ |
| dist/index.html | 240.00b | +0.00b | 1.13m | +38.42s ⚠️ |

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 6.04s | -26.00ms |
| Cached | 297.00ms | +25.00ms ⚠️ |

#### Cold Bundles

*No bundle changes detected.*

#### Cached Bundles

*No bundle changes detected.*

</p></details>

**[Click here to view a detailed benchmark overview.](https://benchmark.parceljs.org/benchmark/330481156792255059)**

packages/runtimes/hmr/src/loaders/hmr-runtime.js
devongovett: hmm I think most DOM events actually don't use dashes? And the CustomEvent example on MDN also doesn't... https://developer.mozilla.org/en-US/docs/Web/API/CustomEvent/CustomEvent. Anyway, doesn't really matter. This isn't meant to be public API.

packages/core/utils/src/prettyDiagnostic.js
101arrowz: You could make `codeframe` a getter that processes `frames` in the default manner so there's less duplication. Beyond disgusting regex I don't think there's any other way to support rewriting paths than saving the `frames` array.

101arrowz:
The logic here for the dev server is a bit problematic for web extensions since it depends on `parcel serve` rather than just `parcel watch` in order to serve the new chunks. To fix this we'd need to refactor a lot of serving logic. Additionally, some of the hardcoded `/__parcel_source_root` won't work because it will try to access the web extension context, we need to change it into absolute paths including hostname if `serveOptions.host != null`.

I'll try to do this at some point.

PR #8045, opened 2022-05-03
Add support for Vue 3 <script setup>

### LINKED ISSUE #6949: Support Vue 3's `<script setup>` transform
# 🐛 bug report

Parcel's Vue transformer currently does not support the new [`<script setup>` transform](https://v3.vuejs.org/api/sfc-script-setup.html).

```vue
<script setup>
	import { ref } from 'vue';

	let count = ref(0);

	function increment () {
		count.value++;
	}
</script>

<template>
	<div class='red'>It's red!</div>
	<button @click='increment'>{{ count }}</button>
</template>

<style scoped>
	.red {
		color: red;
	}
</style>
```

`package.json`
```json
{
	"type": "module",
	"name": "project",
	"scripts": {
		"serve": "parcel serve ./src/index.html",
		"build": "parcel build ./src/index.html"
	},
	"dependencies": {
		"vue": "^3.2.11"
	},
	"devDependencies": {
		"@parcel/transformer-vue": "2.0.0-nightly.2464",
		"parcel": "2.0.0-nightly.840"
	}
}
```

## 🤔 Expected Behavior

`count` should be initialized, and the button should show `0`

## 😯 Current Behavior

The `<script setup>` is ignored in the compiled code, leaving the button empty.

## 💻 Code Sample

https://github.com/intrnl/parcel-issue-vue-3-setup

## 🌍 Your Environment

| Parcel           | 2.0.0-nightly.840
| Node             | 16.9.1
| npm/Yarn         | pnpm 6.15.1
| Operating System | Arch Linux 5.14.2-197-tkg-cacule

Jengamon:
# ↪️ Pull Request

This PR is meant to close #6949, and add support for Vue 3's `<script setup>` tags.

## 💻 Examples

A component defined like:
```vue
<template lang="pug">
div Hello {{ name }}
</template>

<script setup lang="ts">
import { ref } from 'vue';

const name = ref("Vue");
</script>
```
should compile and say "Hello Vue", rather than "Hello ".

## 🚨 Test instructions

TBH I would want help with adding test cases. Basically, I'd want people to go ham on Vue 3 `<script setup>` SFCs and see if anything breaks (or even combine `<script>` and `<script setup>` as is allowed by Vue).

## ✔️ PR Todo

- [ ] Added more comprehensive test suite

### DIFF
--- a/packages/transformers/vue/src/VueTransformer.js
+++ b/packages/transformers/vue/src/VueTransformer.js
@@ -246,6 +246,7 @@ async function processPipeline({
         compilerOptions: {
           bindingMetadata: script ? script.bindings : undefined,
         },
+        isProd: options.mode === 'production',
         id,
       });
       if (templateComp.errors.length) {
--- a/packages/transformers/vue/src/VueTransformer.js
+++ b/packages/transformers/vue/src/VueTransformer.js
@@ -64,6 +64,9 @@ export default (new Transformer({
     }
 
     const descriptor = parsed.descriptor;
+    let id = hashObject({
+      filePath: asset.filePath,
+    }).slice(-6);
 
     return {
       type: 'vue',
@@ -73,7 +76,7 @@ export default (new Transformer({
         script:
           !!descriptor.script || !!descriptor.scriptSetup
             ? compiler.compileScript(descriptor, {
-                id: asset.id,
+                id,
                 isProd: options.mode === 'production',
               })
             : null,
--- a/packages/transformers/vue/src/VueTransformer.js
+++ b/packages/transformers/vue/src/VueTransformer.js
@@ -46,7 +46,7 @@ export default (new Transformer({
     };
   },
   canReuseAST({ast}) {
-    return ast.type === 'vue' && semver.satisfies(ast.version, '^3.0.0');
+    return ast.type === 'vue' && semver.satisfies(ast.version, '^3.2.0');
   },
   async parse({asset, options}) {
     // TODO: This parses the vue component multiple times. Fix?
@@ -65,11 +65,11 @@ export default (new Transformer({
 
     return {
       type: 'vue',
-      version: '3.0.0',
+      version: '3.2.0',
       program: {
         ...parsed.descriptor,
         script: compiler.compileScript(parsed.descriptor, {
-          id,
+          id: asset.id,
           isProd: options.mode === 'production',
         }),
       },
--- a/packages/transformers/vue/src/VueTransformer.js
+++ b/packages/transformers/vue/src/VueTransformer.js
@@ -74,7 +74,7 @@ export default (new Transformer({
       program: {
         ...descriptor,
         script:
-          !!descriptor.script || !!descriptor.scriptSetup
+          descriptor.script != null || descriptor.scriptSetup != null
             ? compiler.compileScript(descriptor, {
                 id,
                 isProd: options.mode === 'production',
--- a/packages/transformers/vue/src/VueTransformer.js
+++ b/packages/transformers/vue/src/VueTransformer.js
@@ -366,7 +366,8 @@ ${
             modules: style.module,
             preprocessLang: style.lang || 'css',
             scoped: style.scoped,
-            map: style.src ? undefined : style.map,
+            inMap: style.src ? undefined : style.map,
+            isProd: options.mode === 'production',
             id,
           });
           if (styleComp.errors.length) {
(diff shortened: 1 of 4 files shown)


Jengamon:
An example test of this would require the 'vue' package, but I'm not sure how to do this, if someone with more experience with writing integration tests could give me pointers.

mischnic:
There are various existing Vue tests that you can adapt: https://github.com/parcel-bundler/parcel/blob/v2/packages/core/integration-tests/test/vue.js

Jengamon:
> There are various existing Vue tests that you can adapt: https://github.com/parcel-bundler/parcel/blob/v2/packages/core/integration-tests/test/vue.js

I've looked through them, but none of the tests import 'vue' in the component.
I tried it naively (just using the component as-is), gives me `TypeError: Cannot destructure property 'expose' of 'undefined' as it is undefined.` when calling setup.

Jengamon:
Ok, added a basic test that tests for the existence of setup and render for a component that only uses <script setup>.

I'd also want to get the actual *contents* of the bundle tested (I hit some weird bugs.)

It would be cool if there were tests for the kinds of things you could do with <script setup> (combine it with <script>, or use it with v-bind for css) and inspecting the actual output of setup would be cool (I know where the `expose` error is coming from now.), but this should work for a first pass.

mischnic:
> I tried it naively (just using the component as-is), gives me TypeError: Cannot destructure property 'expose' of 'undefined' as it is undefined. when calling setup.

So it works fine after changing the order (as you've commited now)? Is that how Vue is supposed to work?

Jengamon:
Oh, no, it had nothing to do with the order of the blocks in the file. It was because I was *calling* setup with no args, but it actually should take 2, and "expose" was supposed to be a member in the second arg.

Relevant docs:
https://vuejs.org/api/composition-api-setup.html#accessing-props
https://vuejs.org/api/composition-api-setup.html#setup-context

I just hadn't read these going into this.

packages/transformers/vue/src/VueTransformer.js
```diff
@@ -63,10 +63,24 @@ export default (new Transformer({
...
+    }).slice(-6);
+
     return {
       type: 'vue',
       version: '3.0.0',
-      program: parsed.descriptor,
+      program: {
+        ...descriptor,
+        script:
+          !!descriptor.script || !!descriptor.scriptSetup
```
mischnic: Could this be rewritten as `... != null || ... != null`? I'd prefer that over `!!`

Jengamon: fixed~

packages/transformers/vue/src/VueTransformer.js
```diff
@@ -64,10 +64,24 @@ export default (new Transformer({
       });
     }
 
+    const descriptor = parsed.descriptor;
+    let id = hashObject({
+      filePath: asset.filePath,
+    }).slice(-6);
```
101arrowz: This seems a bit sketchy. Why does this ID only depend upon the asset's location?

EDIT: actually this is for consistency with the `compileStyle` ID. Not sure why we do that honestly, I'll look into it more.

101arrowz: [`vue-loader`](https://github.com/vuejs/vue-loader/blob/e9314347d75a1b0e54f971272d23a669fc3e6965/src/index.ts#L117-L121) and [Vite](https://github.com/vitejs/vite/blob/99740949b77509c442a27fb0e017fcbe376f3391/packages/plugin-vue/src/utils/descriptorCache.ts#L30) both use the source code for the Vue component as well in production. We should probably do the same. You can add this ID to the returned descriptor:

```js
let id = hashObject({
  filePath: asset.filePath,
  source: options.mode === 'production' ? code : null
});

return {
  type: 'vue',
  version: '3.0.0',
  program: {
    ...,
    id
  },
}
```

Later, when we generate the same ID at the top of `transform`, you can replace it with:

```js
let {template, script, styles, customBlocks, id} = nullthrows(
  await asset.getAST(),
).program;
```

Jengamon: Ok, changed and pushed!

101arrowz (approved):

ngustavo:
I tried using this today and it says properties are not defined on 2.5.0

Jengamon:
This is not in 2.5.0 afaik. Nightly, maybe, but this was merged after the release of 2.5.0.

PR #8076, opened 2022-05-09
Add multi module compilation for elm

### LINKED ISSUE #2508: Elm: Bulk compile assets to avoid duplication
# 🙋 feature request

Bulk compile Elm assets to avoid shared code duplication.

## 😯 Current Behavior

Given that Elm modules `A` and `B` have their own main functions, and both depend of common libraries such as `elm/browser`, `elm/json`, and then imported and initialized as follows:
```javascript
import A from './A.elm'
import B from './B.elm'

A.init(...)
B.init(...)
```
That way, each module gets compiled individually alongside their own dependencies, causing dependencies to be bundled twice, thus producing a much larger output in comparison to:
`elm make ./A.elm ./B.elm output=bundle.js`

## 💁 Possible Solution

One solution would be using glob imports:
```javascript
import Bundle from './*.elm'

Bundle.A.init(...)
Bundle.B.init(...)
```
Although being pretty cool, this might conflict with the existing glob import behavior.

A less ideal solution, would be defining an external file with bundle definitions:
```javascript
// elm-bundles.json
{
    "app": [
        "./A.elm",
        "./B.elm"
    ]
}
```
So `ElmAsset.js` could check if a given asset is present in a bundle, and then build the bundle instead of the individual asset.

A similar solution has been implemented in `elm-webpack-loader`.

## 🔦 Context

In my current setup, I have around a dozen Elm element apps being imported into the same application. When importing directly through parcel, the output is more than three times as large than `elm make A.elm B.elm ...` outputs.

### LINKED ISSUE #8263: Elm code with Debug.log fails with Javascript error
# 🐛 bug report
I have forgotten to delete the `Debug.log` code on elm and the build failed with javascript error.

Inside my elm code I had 
```
 case x of 
   ...

  OnChange value ->
      let
          _ =
               Debug.log "Test" value
      in
      .....

```

package.json

```
"devDependencies": {
        "@parcel/packager-raw-url": "^2.3.2",
        "@parcel/transformer-elm": "^2.3.2",
        "@parcel/transformer-webmanifest": "^2.3.2",
        "@vercel/node": "^1.15.0",
        "vercel": "^24.0.0"
    },
```

## 🤔 Expected Behavior

Show error that the code is running with `Debug` and its not allowed?

## 😯 Current Behavior

```
@parcel/transformer-elm: Cannot read property 'flatMap' of undefined

  TypeError: Cannot read property 'flatMap' of
  undefined
      at Object.transform
  (/Users/{path/to/project}/node_modules/@parcel/transformer-elm/lib/ElmTransformer.js:170:48)
      at async Transformation.runTransformer
  (/Users/{path/to/project}//node_modules/@parcel/core/lib/Transformation.js:653:5)
      at async Transformation.runPipeline
  (/Users/{path/to/project}//node_modules/@parcel/core/lib/Transformation.js:402:36)
      at async Transformation.runPipelines
  (/Users/{path/to/project}//node_modules/@parcel/core/lib/Transformation.js:265:40)
      at async Transformation.run
  (/Users/{path/to/project}//node_modules/@parcel/core/lib/Transformation.js:183:21)
      at async Child.handleRequest
  (/Users/{path/to/project}//node_modules/@parcel/workers/lib/child.js:217:9)

```

## 💁 Possible Solution
Prevent building if there are any `Debug`

## 🌍 Your Environment

| Parcel           |  v2
| Node             | v14.119.1
| npm/Yarn         | Yarn v1.22.18
| Operating System | MacOs Monterey

ChristophP:
Fixes https://github.com/parcel-bundler/parcel/issues/2508
Fixes https://github.com/parcel-bundler/parcel/issues/8263

# ↪️ Pull Request

Add multi module compilation for Elm](https://github.com/parcel-bundler/parcel/commit/58174b5e136948717e70772816adde586a50e6eb)

The Elm compiler can compile multiple entrypoints into one bundle, which
helps keep the asset size small since they all share the same runtime.

The approach used here has been inspired by the elm plugin for vite: https://github.com/hmsk/vite-plugin-elm#combine-multiple-main-files-experimental-from-270-beta1

## 💻 Examples

```js
import { Elm } from
'./Main.elm?with=./AnotherModule.elm,./YetAnotherModule.elm'

Elm.Main.init();
Elm.AnotherModule.init();
Elm.YetAnotherModule.init();
```

## 🚨 Test instructions

Use a js file like this:
```js
// src/index.js
import { Elm } from "../../../Main.elm?with=./ModuleB.elm";

document.body.innerHTML = '<div id="elm-1"></div><div id="elm-2"></div>'

Elm.Main.init({ node: document.getElementById('elm-1') });
Elm.ModuleB.init({ node: document.getElementById('elm-2') });
```
```elm
module Main exposing (main)

import Browser
import Html exposing (Html, button, div, text)
import Html.Events exposing (onClick)


type Msg
    = Increment
    | Decrement


update msg model =
    case msg of
        Increment ->
           ({ model | count = model.count + 1 }, Cmd.none)

        Decrement ->
           ({ model | count = model.count - 1 }, Cmd.none)


view model =
    div []
        [ button [ onClick Increment ] [ text "+1" ]
        , div [] [ text <| String.fromInt model.count ]
        , button [ onClick Decrement ] [ text "-1" ]
        ]


main =
    Browser.sandbox
        { init = \_ -> ({ count = 0 }, Cmd.none)
        , view = view
        , update = update
        , subscriptions = always Sub.none
        }
```
and another Elm module

```elm
module ModuleB exposing (main)

import Browser
import Html exposing (Html, button, div, text)
import Html.Events exposing (onClick)


type Msg
    = Increment
    | Decrement


update msg model =
    case msg of
        Increment ->
           ({ model | count = model.count + 1 }, Cmd.none)

        Decrement ->
           ({ model | count = model.count - 1 }, Cmd.none)


view model =
    div []
        [ button [ onClick Increment ] [ text "+1" ]
        , div [] [ text <| String.fromInt model.count ]
        , button [ onClick Decrement ] [ text "-1" ]
        ]


main =
    Browser.sandbox
        { init = \_ -> ({ count = 0 }, Cmd.none)
        , view = view
        , update = update
        , subscriptions = always Sub.none
        }
```


devongovett:
I'm not very familiar with Elm, but what's the difference between doing this vs making a new entry file and importing things there instead? The syntax here of importing files with a query string seems kinda strange to me.

ChristophP:
@devongovett Let me give you a bit more background on this. The difference is basically creating smaller bundles.

## Multiple compile steps vs a single compile step

```js
import { Elm } from "src/Main.elm";
import { Elm as Elm2 } from "src/Main2.elm";

Elm.Main.init();
Elm2.Main2.init();
```
and
```js
import { Elm } from "src/Main.elm?with=src/Main2.elm";

Elm.Main.init();
Elm.Main2.init();
```
is that
- the former creates **two** JS assets in two compilation steps, effectively duplicating the Elm runtime + any Elm code that may be shared between the entrypoints and their dependencies.
- the latter creates **one** JS asset in a single compilation step, combining multiple Elm entrypoints into the same output file. This way, the Elm runtime + any shared code gets shared in the compiled output between all entrypoints.

When using the elm compiler on the command line those cases be equivalent to this
```
elm make src/Main.elm --output=elm.js
elm make src/Main2.elm --output=elm2.js
# vs
elm make src/Main.elm src/Main2.elm --output=elm.js
```

### How common is this?
I wouldn't say this is super common, but if anyone wants to do this they couldn't use Elm with parcel right now.

## Query string syntax

I agree that it seems a bit odd. I was wondering how to do this and found that this is how it's done with Vite. https://github.com/hmsk/vite-plugin-elm#combine-multiple-main-files-experimental-from-270-beta1
If there's a better way of specifying that I'd be happy to change it.

ChristophP:
For reference, webpack handles it like this.
https://github.com/elm-community/elm-webpack-loader#files-default---path-to-required-file
It provides some extra config, but one of the things I love about parcel is that you specify most things where you import it.

devongovett:
Sure. I guess what I was asking is if you could make a new Elm file that imports all of your entry points and re-exports them somehow, and then import that from JS instead?

teppix:
> if you could make a new Elm file that imports all of your entry points and re-exports them somehow

Elm-applications are imported by their module names, and for a module to be considered an application, it has to contain a function named `main`. That means that you can't really define more than one application in the same module.

ChristophP:
Yes, what @teppix said. The Elm compiler doesn't allow more than one `main` per entry point.

devongovett:
So you can't do something like this?

```elm
module Main exposing (ModuleA, ModuleB)

import ModuleA
import ModuleB
```

and then from JS:

```js
import { Elm } from './Main.elm';

Elm.ModuleA.init();
Elm.ModuleB.init();
```

I have no idea if this is valid syntax, but that's what I meant. If possible, that seems nicer than doing it in the query string.

ChristophP:
Unfortunately not. In order to have an .init() function on the JS side you need to define a main function on the Elm side. Currently Elm only allows one main function per entry point module and it has to be in the topmost module (not imported by any other module).

devongovett:
Got it. Maybe that would be a good issue to open in Elm then. In the meantime, I guess this is fine. Would you mind adding a test? We'll need to add this to the documentation as well.

ChristophP:
@devongovett 

### Tests
sure, I'll add a test for this.

### Docs
I can also prepare some documentation if you'd like. Would that be another PR in this repo over here https://github.com/parcel-bundler/website/blob/v2/src/languages/elm.md ?

### Syntax / API design

I've been thinking a bit more on the not-so-perfect syntactical way of using this.
```js
import { Elm } from './Main.elm?with=./AnotherModule.elm,./YetAnotherModule.elm'
```
When there are more than one extra modules I separate them with a `,` withing the value of the query param `with`. This is kind of noisy and hard to read.

I thought maybe this could be improved a little by specifying the `with` param once for each module, like this.
```js
import { Elm } from './Main.elm?with=./AnotherModule.elm&with=./YetAnotherModule.elm'
```
and then get those values with `asset.query.getAll('with')`.
The upside is it seem a bit more readable and more natural with how URL params are used normally, the downside is that instead of one extra character per module (`,`) its 6 (`&with=`).

Do you have any thoughts on this? Improvement or just as bad?

devongovett:
Better I think!

ChristophP:
I made the changes regarding multiple `with` params as well as adding a test. I was still planning on improving error handling when one of the extra sources isn't there but haven't gotten around to it yet. Will hopefully do it this week.

ChristophP:
I have a question that I just thought about:

What would happen if someone imported this twice?
Assume Elm code is imported with parcel in two files like this.
```js
// in file1.js
import { Elm } from './Main.elm?with=./AnotherModule.elm&with=./YetAnotherModule.elm';
// in file2
import { Elm } from './Main.elm?with=./AnotherModule.elm&with=./YetAnotherModule.elm';
```
I would assume the asset is compiled once and imported in both files, right?
Would the same be true if someone did the same but changed the order of the query params?
```js
// in file1.js
import { Elm } from './Main.elm?with=./AnotherModule.elm&with=./YetAnotherModule.elm';
// in file2
import { Elm } from './Main.elm?with=./YetAnotherModule.elm&with=./AnotherModule.elm';
```
How does parcel treat the specifiers where the only difference is the order of the query params, i.e. when the name is the same, when the queries are the same, when the exact string is the same? Would parcel run the transformer twice and include the compiled JS twice in the resulting bundle?

mischnic:
Here's where the query comes from in the resolver:
https://github.com/parcel-bundler/parcel/blob/f2d0a3a27d6e493b23ddc2edbc8a4c0053ff34ab/packages/utils/node-resolver-core/src/NodeResolver.js#L603

and it's then turned into a string to store it in the cache.
https://github.com/parcel-bundler/parcel/blob/f2d0a3a27d6e493b23ddc2edbc8a4c0053ff34ab/packages/core/core/src/requests/PathRequest.js#L242

It looks like this won't normalize the order. So there will be two assets.

And theoretically you could also write a transformer where this order matters, so there's not much we can do here.

ChristophP:
@mischnic thanks for the clarification.

I see how that could be difficult to change, but also surprising at the same time. (i.e. when importing `./image.jpeg?width=100&height=200` vs `./image.jpeg?height=200&width=100` would be different assets).

I wonder if the current solution with the query string is the way to go then, because I could see that leading to unintentionally increasing the bundle size significantly without warning or specifying the path slightly diffently (i.e. no leading `./`).

So I'd like to propose a different solution for mutli-module compilation for Elm: In the package.json, one could add something like this:
```json
  "@parcel/transformer-elm": {
    // this means: whenever Main.elm is imported, also compile AnotherModule.elm and YetAnotherModule.elm with it
    "extraSources": {
      "./src/Main.elm": ["./src/AnotherModule.elm", "./src/YetAnotherModule.elm"]
    }
  }
```
This is more of a static config but it has a few benefits IMO
1. **No pitfalls:** It avoids the unintentional asset duplication problem
2. **Cleaner imports:** When importing the compiled Elm code multiple times you only have to `import { Elm } from "./src/Main.elm";` instead of  `import { Elm } from './Main.elm?with=./AnotherModule.elm&with=./YetAnotherModule.elm';` every time. Users have a single place to look at which Elm assets contain which modules.
3. **It scales better:** Having 10+ modules in an array in the package.json works, while having 10 extra files in the query string becomes unreadable. 

Any thoughts on this new approach? If you're in favor of this different approach, would you prefer me closing this PR and opening a new one or add commits to this one?

ChristophP:
I went ahead and implemented the approach described above and it works. nicely https://github.com/parcel-bundler/parcel/pull/8076/commits/e7162240b60f72b280e96cea4ca4bb336252d718
Please let me know if you're okay with the approach, then I can add tests and documentation.

mischnic:
The biggest difference I see between query param and the central config: is there a usecase for importing `A.elm?with=B.elm` and in some other file importing `A.elm?with=C.elm` (which wouldn't be possible with the config the `with` parameter is essentially fixed for all imports)?

Apart from the other advantages you've listed, if the only concern is deduplicating two different but equivalent `with` lists, then this would work:

```js
transform({asset}) {
    if(isQueryNormalized(asset.query)) {
        return whatTheElmTransformerDoesAtTheMoment(...);
    } else {
        let specifier = "./" + path.filename(asset.filePath) + "?" +
                        // Sort, normalize leading "./", etc.
                        normalizeQuery(asset.query).toString();
        asset.setCode(`export * from "${specifier}";`);
        asset.type = "js";
        return [asset];
    }
}
```

ChristophP:
Hi @mischnic ,

> is there a usecase for importing `A.elm?with=B.elm` and in some other file importing `A.elm?with=C.elm` (which wouldn't be possible with the config the with parameter is essentially fixed for all imports)?

I don't see one to be honest. The whole point of this multi-module compilation is saving on bundle size. If one needs `A.elm` and `B.elm` somewhere and `A.elm` and `C.elm` somewhere else, it would be better to have the three of them in one bundle together, because otherwise `A.elm` would be duplicated, which kinda defeats the purpose of doing mulit-module compilation in the first place.

> Apart from the other advantages you've listed, if the only concern is deduplicating two different but equivalent with lists, then this would work:

This is very interesting. Nice trick to sort the query string and turn it into a JS asset with a normalized query. Maybe that would also be good for the image transformer.

Despite of there being a solution to the duplication problem, I think advantages of the config approach outweigh the ones of the query string approach (which is mostly not having to have a config). Especially the fact that the query approach gets really tedious with long file paths and/or many extra sources. So I'm proposing to go with the config solution.

ChristophP:
Added tests and fixed flow errors.

teppix:
> > is there a usecase for importing `A.elm?with=B.elm` and in some other file importing `A.elm?with=C.elm` (which wouldn't be possible with the config the with parameter is essentially fixed for all imports)?
> 
> I don't see one to be honest. The whole point of this multi-module compilation is saving on bundle size. If one needs `A.elm` and `B.elm` somewhere and `A.elm` and `C.elm` somewhere else, it would be better to have the three of them in one bundle together, because otherwise `A.elm` would be duplicated, which kinda defeats the purpose of doing mulit-module compilation in the first place.

I would have to disagree. I think it's important to be able to control which elm-application ends up in which bundle. For instance, we have some smaller elm-applications shared among different pages, but we also have a rather large SPA which we definitely don't want to load by default on the landing page. In this case, having some smaller apps duplicated across bundles would be better than having the large SPA included in a single monolithic bundle.

As long as it's possible to get around this restriction e.g. by using different build targets, I suppose it would be OK though.

ChristophP:
@teppix Interesting, sounds like this kind of flexibilty might be useful for your case.

> As long as it's possible to get around this restriction e.g. by using different build targets, I suppose it would be OK though.

Can you elaborate what you mean by using different build targets?

Also, how do you feel about the query string solution, when there are 4-5+ modules you need to compile in terms of readability and so?

teppix:
> @teppix Interesting, sounds like this kind of flexibilty might be useful for your case.
> 
> > As long as it's possible to get around this restriction e.g. by using different build targets, I suppose it would be OK though.
> 
> Can you elaborate what you mean by using different build targets?

By targets i'm referring to the [targets](https://parceljs.org/features/targets/) feature of parcel.
I suppose this will only work if you somehow can specify different bundle configurations for elm for each target.
I don't know enough of parcels internals to have an informed opinion on this however.

> Also, how do you feel about the query string solution, when there are 4-5+ modules you need to compile in terms of readability and so?

Given the lack of better alternatives, i suppose it works, but if i'm being honest; not really a fan of that approach either :)

teppix:
Looking at the package.json-suggestion again, i suppose it's flexible enough, since you could create dummy-modules which acts as entrypoints to each build-configuration.

Another aspect to think about is source-navigation.
I personally really appreciate if my imports explicitly points to actual modules, making it possible to jump directly to the included source file. I don't think this works in any of the suggested approaches.

One (probably not desirable) solution might be to invent a new kind of manifest-file, which you can include in the js source. It might look like this:

`my-elm-manifest.elm.json`
```json
{
  "sources": ["./elm/App1.elm", "./elm/App2.elm"]
}
```

and include it via
```js
import Elm from './my-elm-manifest.elm.json';
```

I don't really think we want to invent too much new stuff here, but at least this might solve the source-navigation issue i mentioned.

Just some food for thought...

ChristophP:
Ok, here's my idea of how to accommodate the case for full control over what Elm modules are included in a build.

```js
import { Elm } from "./Main.elm?with=widgets";
```
where `widgets` would be defined in the `package.json` like this:
```json
  "@parcel/transformer-elm": {
    "extraSources": {
      "widgets": ["./src/WidgetA.elm", "./src/WidgetB.elm"],
      "spa": ["./src/AnotherModule.elm", "./src/YetAnotherModule.elm"],
    }
  }
```
In this case `"./Main.elm?with=widgets"` would result in `Main.elm` being compiled together with `WidgetA.elm` and `WidgetB.elm`. If you don't want the `Main.elm` in there you could import `WidgetA?with=widgets` for example and `WidgetA` will only appear once in the resulting bundle.
If you specify `with=doesntExist` you'll get a warning that no extra sources could be found for `doesntExist`

@teppix What do you say?

PS: Regarding the source navigation, I get that this would be nice, but this is already hard enough to get to work for all use cases, so I'd like to not increase the complexity even further with something like intermediate json files.

teppix:
Disclaimer: I'm certainly not in a position to make any decisions, so you are free to overlook any suggestions. These are just my personal opinions 🙂 

For what it's worth, I think any of the suggested alternatives would work as a semi-temporary solution. I guess that depends a bit on what parcel's policy is for breaking changes, since this is after all a core parcel-plugin.

Long term I think this would probably have to be revised anyway.

I think your latest suggestion is pretty nice, e.g.
```json
"@parcel/transformer-elm": {
    "extraSources": {
      "widgets": ["./src/WidgetA.elm", "./src/WidgetB.elm"],
      "spa": ["./src/AnotherModule.elm", "./src/YetAnotherModule.elm"],
    }
  }
```

but instead of importing by filename, you would simply import by the build-names you defined, and skipping the module path entirely
```js
import Elm from "elm:widgets"
```

(In this case `"extraSources"` is probably not the right name for the option, but rather `"builds"` or something similar)

My reasoning for this is that:
1. it looks cleaner (I can't tell if this syntax would be illegal from a javascript/typescript standpoint though.)
2. in reality you are not really just importing one module, but all of the modules that are compiled together, so it feels a bit wrong to specify a single one of the modules.
3. specifying `./Module.elm?with=my-build` would be a bit ambiguous. What happens if `Module.elm` is not included in the source-list for `my-build`?

mischnic:
> For what it's worth, I think any of the suggested alternatives would work as a semi-temporary solution. I guess that depends a bit on what parcel's policy is for breaking changes, since this is after all a core parcel-plugin.
>
> Long term I think this would probably have to be revised anyway.

Right now it's a new feature. So we wouldn't want to add a temporary solution if we intend to break it in the future. However:

> import Elm from "elm:widgets"

From a user's perspective, this is certainly nice. But this is rather hard to do with Parcel plugins, because `elm` looks like a pipeline (such as `url:./img.png`) which would also require a resolver plugin. (Which could totally pull this off with something like `if(pipeline === "elm") return lookupNameInConfig(specifier);`).

That makes me wonder if we should add `./A.elm?with=B.elm&with=C.elm` in the official Elm transformer. Then there could be a third party resolver plugin which takes the config and essentially rewrites the `elm:widgets` import into the equivalent `..?with=...` import.

> specifying `./Module.elm?with=my-build` would be a bit ambiguous. What happens if Module.elm is not included in the source-list for my-build?

I would read this as: take Module.elm plus everything specified under "my-build".

But if the usecase in reality is just `import Elm from "elm:widgets"`, then this split between main and other files is strange anyway.

ChristophP:
@teppix Thanks for weighing in.

I have also thought about what you mentioned here
> 2. in reality you are not really just importing one module, but all of the modules that are compiled together, so it feels a bit wrong to specify a single one of the modules.

It's true. The thing is that in parcel (AFAIK) has the concept of assets import that reference other assets. The assets have to be actual files, so you can't do something like this `import "thing-that-doesn't-exist.elm"` and parcel transformers run on individual assets.

Something like `import Elm from "elm:widgets"` would be the cleanest in terms of "you're actually importing a bundle", but problematic because there is no file called `elm:widgets`. There may be a way to hack into the resolver that resolves the specifier (`elm:widgets`) but I would still be hard to get around the restriction that a specifier needs to point to a single file that exists on the file system.

Given this restriction, I don't know of a better way to solve this.

To your other point
> 3. specifying ./Module.elm?with=my-build would be a bit ambiguous. What happens if Module.elm is not included in the source-list for my-build?

I was going for this behavior for `import { Elm } from "./Main.elm?with=widgets&with=spa";`.
`Main.elm` will always be part of the bundle.
The final list of sources will be `Main.elm` + `all modules in widgets` + `all modules in spa` (filtered for duplicates).
That way
- if you want to have `Main.elm` along with `widgets` it's easy to do (`Main.elm?with=widgets`).
- If you just want widgets you'll have to import one of the Modules that's in `widgets` (`WidgetA.elm?with=widgets`).

I know this solution isn't perfect, but
- It would be flexible enough for all use cases
- It works with what parcel gives us
- It allows you to structure your Elm modules into groups that you want to have bundled
- Keeps your query string short and readable

UPDATE: I didn't see @mischnic 's comment until after I posted. You're saying the official ElmTransformer could just support `?with=` and then you could use a third party resolver that rewrites grouped modules from a config to a query string to make it cleaner for people where the query string would be super long, right?

mischnic:
> You're saying the official ElmTransformer could just support ?with= and then you could use a third party resolver that rewrites grouped modules from a config to a query string to make it cleaner for people where the query string would be super long, right?

Yes

> but I would still be hard to get around the restriction that a specifier needs to point to a single file that exists on the file system.

(There is no such restriction. It just needs to map (pipeline + specifier) to either a file on disk or return some dynamically generated code)

mischnic:
Here's what such a resolver would look like. That appears to work fine with your previous version using `?with` instead of the current `extraSources` thing: https://gist.github.com/mischnic/32ebf448f2c14dd24756a11499f02db2

ChristophP:
Ok here's what I'll do.
- I'll revert my changes back to the query string solution
- then I'll add logic to normalize that query string to avoid asset duplication on equivalent but different queries
- sometime after this PR is merged I will publish a resolver handing the @mischnic described (with your permission, since most of  the code is provided by you :-) )

mischnic:
> (with your permission, since most of the code is provided by you :-) )

The main reason why I didn't create a repo was that I didn't want to have to maintain it. So please go ahead 😂

ChristophP:
> The main reason why I didn't create a repo was that I didn't want to have to maintain it. So please go ahead joy

Will do. Thx :smile:

ChristophP:
Ok, I reverted the logic back to the query param solution and implemented a query deduplication logic.
@mischnic could you give me some pointers on two things, please?
1. How do I best test that the query deduplication works, i.e. that the runtime and shared code are not duplicated if imported twice with a different but equivalent query string?
2. When importing like `from './A.elm?with=./B.elm&with=./C.elm'`, I would like to make the paths for `B.elm` and `C.elm` relative to the JS file that imports them (just like for `A.elm`). So far I could only get the paths to be relative to `A.elm`, which might be confusing. Is there any way I can reference the path of the importing JS file from within the `ElmTransformer` to compute the bases for those paths?

mischnic:
> How do I best test that the query deduplication works, i.e. that the runtime and shared code are not duplicated if imported twice with a different but equivalent query string?

If they're duplicated, then there will be two assets with the filepath `Main.elm`:
https://github.com/parcel-bundler/parcel/blob/3b3b13557dba9d1a403c86550915c6013dd78dfe/packages/core/integration-tests/test/javascript.js#L397-L402

> Is there any way I can reference the path of the importing JS file from within the ElmTransformer to compute the bases for those paths?

A transformer doesn't know who imported the asset. What you're describing would conceptually be a resolver's responsibility.

ChristophP:
@mischnic 
1. Regarding the duplication. I wrote this test https://github.com/parcel-bundler/parcel/pull/8076/commits/9353f45a3c6c23b38356d8a73abf91f750bc1d85 but keep getting `Main.elm` twice in the asset array, which surprises me because I implemented the deduplication logic according to your comment here https://github.com/parcel-bundler/parcel/pull/8076#issuecomment-1140472980 (at least I think I do). Unless I'm doing something wrong I think that parcel still regards this as two assets.
2. Regarding knowing where the asset was imported: If there's no way of knowing (which makes sense since it can be imported more than once), I guess the only choice is to have the paths in the `with` param be relative to the `Main.elm` instead of the importing JS file. If that's document well, I guess this won't be a huge deal.

ChristophP:
So I ended up removing the query string deduplication logic for two reasons
- I couldn't get it to work as I wanted it in the test. The `Main.elm` was always shown multiple times in the integration test.
- The deduplication would have prevented unintentional asset duplication when things are imported more than once with different order of query params. If people use this feature sparingly or with the special resolver like this https://github.com/parcel-bundler/parcel/pull/8076#issuecomment-1142684231, unintentional asset duplication won't be a huge deal anyway.

Other than adding documentation for this feature and publishing the resolver package I think the work here is done. Does this look like something we could merge?

mischnic:
> If they're duplicated, then there will be two assets with the filepath Main.elm:

> The Main.elm was always shown multiple times in the integration test.

(Sorry, I might have confused you there. There will be multiple assets called `Main.elm`, but only one of them will be the actual Elm bundle, and all other ones will just contain `export * from "...";`)

ChristophP:
@mischnic 

> (Sorry, I might have confused you there. There will be multiple assets called Main.elm, but only one of them will be the actual Elm bundle, and all other ones will just contain export * from "...";)

Ah ok, yeah I got confused a bit. Would there be a good way to test that the other ones are just `export * from ...` then?

If so I could do that and bring back the deduplication. But overall, I guess the deduplication logic is not that important when the recommended way to use this feature is the resolver you described.

mischnic:
I can't think of a better way than `asset.getCode().length < ???`.

But yeah, I don't think that the deduplication is essential

ChristophP:
Ok, if you're fine not having the deduplication logic then I guess this PR is ready.

Would you like me to create another PR for the documentation?

devongovett (approved):
Thanks! Yes updating the documentation would be really helpful. 😄

ChristophP:
Added some docs here https://github.com/parcel-bundler/website/pull/1038

Will still finish up the extra third party resolver package @mischnic was proposing. here https://github.com/parcel-bundler/parcel/pull/8076#issuecomment-1142684231

ChristophP:
Thanks @mischnic will look at some of the suggestions

ChristophP:
I also bumped terser to the latest version to fix some issues around production builds that hang forever
see https://github.com/terser/terser/issues/1212 and https://github.com/passiomatic/elm-designer/issues/58

mischnic:
You'll need to run `yarn` to update the lockfile after the Terser bump. And please also bump the version in https://github.com/parcel-bundler/parcel/blob/v2/packages/optimizers/terser/package.json

devongovett:
Handled terser bump separately

Issue #5911, opened 2021-02-23
@parcel/transformer-typescript-tsc does not work


ranisalt:
# 🐛 bug report

## 🎛 Configuration (.babelrc, package.json, cli command)
package.json:
```json
{
	"source": "src/index.ts",
	"module": "dist/index.js",
	"types": "dist/index.d.ts",
}
```

`src/index.ts`
```ts
export * from './src/components';
```

CLI: `parcel build index.ts`

## 🤔 Expected Behavior

It should build.

## 😯 Current Behavior

```
@parcel/transformer-typescript-types: node.exportClause.elements is not iterable
TypeError: node.exportClause.elements is not iterable
    at visit (/home/rsa/skyfall/frontend/node_modules/@parcel/transformer-typescript-types/lib/collect.js:66:47)
    at visitNodes (/home/rsa/skyfall/frontend/node_modules/typescript/lib/typescript.js:80382:48)
    at Object.visitEachChild (/home/rsa/skyfall/frontend/node_modules/typescript/lib/typescript.js:80739:56)
    at visit (/home/rsa/skyfall/frontend/node_modules/@parcel/transformer-typescript-types/lib/collect.js:112:15)
    at visitNode (/home/rsa/skyfall/frontend/node_modules/typescript/lib/typescript.js:80329:23)
    at Object.visitEachChild (/home/rsa/skyfall/frontend/node_modules/typescript/lib/typescript.js:80737:222)
    at visit (/home/rsa/skyfall/frontend/node_modules/@parcel/transformer-typescript-types/lib/collect.js:112:15)
    at visitNodes (/home/rsa/skyfall/frontend/node_modules/typescript/lib/typescript.js:80382:48)
    at visitLexicalEnvironment (/home/rsa/skyfall/frontend/node_modules/typescript/lib/typescript.js:80422:22)
    at Object.visitEachChild (/home/rsa/skyfall/frontend/node_modules/typescript/lib/typescript.js:80809:55)
```

## 🔦 Context

I just want to build this with types.

## 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Parcel           | parcel@2.0.0-nightly.579
| Node             | v15.9.0
| npm/Yarn         | 1.22.10
| Operating System | Arch Linux

mischnic:
I've tried this:
```
{
	"source": "index.ts",
	"types": "dist/index.d.ts",
	"scripts": {
		"build": "parcel build . --no-cache"
[...]

PR #8088, opened 2022-05-12
Support named module wildcard re-export when generating typescript types

rmja:
# ↪️ Pull Request

This is a proposal for resolving https://github.com/parcel-bundler/parcel/issues/5911.

It allows to specify the following in an index.ts file:

```ts
export * as SomeExportName from "./some-other-file";
```

It does not support 1) the case:
```ts
import * as SomeExportName from "./some-other-file";
export { SomeExportName };
```
and, it does not supported 2) nested exports.

If someone can update the test cases for the two unsupported cases so that I can see the expected result, then let me know.

## ✔️ PR Todo

### DIFF
--- a/packages/core/integration-tests/test/integration/ts-types/exporting/index.ts
+++ b/packages/core/integration-tests/test/integration/ts-types/exporting/index.ts
@@ -1,3 +1,8 @@
 export {createMessage} from './message';
 export * from './other';
 export {test as toUpperCase, default} from './test';
+export * as Module1 from './module1';
+
+// TODO: Make this work
+// import * as Module2 from './module2';
+// export { Module2 };
--- a/packages/core/integration-tests/test/integration/ts-types/exporting/expected.d.ts
+++ b/packages/core/integration-tests/test/integration/ts-types/exporting/expected.d.ts
@@ -13,5 +13,8 @@ export var a: number, b: number;
 export function toUpperCase(foo: string): string;
 declare const _default: "test";
 export default _default;
+export declare module Module1 {
+    export const hello = "world";
+}
 
 //# sourceMappingURL=types.d.ts.map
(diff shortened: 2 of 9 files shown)


rmja:
Can someone shed some light on why checks on some of the platforms fail?

mischnic:
You can ignore those. The tests are flakey, as long as one of the integration test jobs passes, you're fine.

packages/core/integration-tests/test/integration/ts-types/exporting/index.ts
```diff
@@ -1,3 +1,8 @@
 export {createMessage} from './message';
 export * from './other';
 export {test as toUpperCase, default} from './test';
+export * as Module1 from './module1';
```
devongovett, line 4: Adding `export {hello} from './module1'` here results in the types being duplicated, both outside and inside the `module` block. For strings it's not too bad, but for more complex types might be weird. Maybe we should still hoist the declarations out of the module as before, but create a module like this instead to export them?

```ts
export const hello = "world";
export declare module Module1 {
  export {hello};
}
```

rmja: That is a good idea. That could probably also include support for the not supported 1) above.
The problem is then that we very likely gets naming conflicts between the hoisted definitions and the exports. Should we generate some random part in the name to avoid these conflicts?

devongovett: I think the hoisting we already do should handle renaming in the top level scope. So you might need something like `export {hello1 as hello}` within the nested module in that case.

rmja: @devongovett could I ask you to roughly commit to this PR the (failing) tests that describes the intended behavior, then I will try and do the implementation.

ivanbanov:
hey, any timeline for getting this out of the door?

claudeduguay:
Is this really still pending? 
The pattern:  
`export * as X from "..."`
is still broken after a fix has been ready to merge for 2+ years? 
I need this fix please...

PR #8097, opened 2022-05-16
Support using packages in imports resolved by the glob resolver

### LINKED ISSUE #7945: Should @parcel/resolver-glob support importing from packages?
# 💬 RFC

The glob resolver (`@parcel/resolver-glob`) only supports a file path as a specifier, and will not do node style resolution if a package is specified.

If the import references a path within a package, should the glob resolver "resolve" the package to a path first? Or should it at least warn or error if the file list the glob resolves to is empty?

## 🔦 Context

I have a component I'm trying to bundle that currently uses [Webpack's dynamic expressions](https://webpack.js.org/api/module-methods/#dynamic-expressions-in-import), something like: ``import(`@scope/pkg/dist/i18n/${locale}`)``.

As part of migrating to Parcel, I tried replacing this with `import('@scope/pkg/dist/i18n/*.js')[locale]` however this did not work correctly, as it is assumed the specifier is a path - https://github.com/parcel-bundler/parcel/blob/8daaed8a8c263635ef9e6c678fd92fef8c042c74/packages/resolvers/glob/src/GlobResolver.js#L55

In the example above, this doesn't work because the resolver tries to find files in `/path/to/my/repo/src/path/@scope/pkg/dist/i18n/*.js` (assuming a source path of `/path/to/my/repo/src/path/`)

Changing the import to a relative path obviously makes it work (i.e. `import('../../../node_modules/@scope/pkg/dist/i18n/*.js')`) but is potentially fragile if things get moved around, and similar to the case above would fail to find any files silently.

Should the glob resolver resolve any packages to paths first? I could see the argument for either the current behaviour or resolving, so it's not cut and dry in my mind - hence the RFC.

Another potential option, if package resolution is determined to not be appropriate here, would be to at least warn (or error?) if the list of files is empty? Because I didn't realise what the problem was until I debugged the build and stepped through the glob resolver. i.e. after the line below, warn if `files.length === 0`: https://github.com/parcel-bundler/parcel/blob/8daaed8a8c263635ef9e6c678fd92fef8c042c74/packages/resolvers/glob/src/GlobResolver.js#L57

marcins:
# ↪️ Pull Request

Implements the ability for the glob resolver to resolve globs in packages, fixes #7945.

## 💻 Examples

With this change the glob resolver will work with a glob like `import('@scope/pkg/i18n/*.js')`, by first resolving `@scope/pkg` to a path, and then appending `/i18n/*.js`.

## 🚨 Test instructions

An integration test has been added for both a regular `require` as well as an async `import`. Please provide feedback if the tests added are not comprehensive enough.

## ✔️ PR Todo

- [x] Added/updated ~unit~ integration tests for this change 

### DIFF
--- a/packages/core/integration-tests/test/glob.js
+++ b/packages/core/integration-tests/test/glob.js
@@ -203,4 +203,39 @@ describe('glob', function () {
       ],
     });
   });
+
+  it('should require a glob of files from a package', async function () {
+    let b = await bundle(
+      path.join(__dirname, '/integration/glob-package/index.js'),
+    );
+    await assertBundles(b, [
+      {
+        name: 'index.js',
+        assets: ['*.js', '*.js', 'a.js', 'b.js', 'x.js', 'y.js', 'index.js'],
+      },
+    ]);
+  });
+
+  it('should require a glob of files from a package async', async function () {
+    let b = await bundle(
+      path.join(__dirname, '/integration/glob-package-async/index.js'),
+    );
+    await assertBundles(b, [
+      {
+        name: 'index.js',
+        assets: [
+          '*.js',
+          '*.js',
+          'bundle-url.js',
+          'cacheLoader.js',
+          'index.js',
+          'js-loader.js',
+        ],
+      },
+      {type: 'js', assets: ['a.js']},
+      {type: 'js', assets: ['b.js']},
+      {type: 'js', assets: ['x.js']},
+      {type: 'js', assets: ['y.js']},
+    ]);
+  });
 });
--- a/packages/core/integration-tests/test/glob.js
+++ b/packages/core/integration-tests/test/glob.js
@@ -214,6 +214,10 @@ describe('glob', function () {
         assets: ['*.js', '*.js', 'a.js', 'b.js', 'x.js', 'y.js', 'index.js'],
       },
     ]);
+
+    let output = await run(b);
+    assert.equal(typeof output, 'function');
+    assert.equal(await output(), 10);
   });
 
   it('should require a glob of files from a package async', async function () {
@@ -237,5 +241,9 @@ describe('glob', function () {
       {type: 'js', assets: ['x.js']},
       {type: 'js', assets: ['y.js']},
     ]);
+
+    let output = await run(b);
+    assert.equal(typeof output, 'function');
+    assert.equal(await output(), 10);
   });
 });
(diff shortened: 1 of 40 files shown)


packages/core/integration-tests/test/integration/glob-package-async/index.js
```diff
@@ -0,0 +1,7 @@
+const scoped = import('@scope/pkg/foo/*.js');
+const unscoped = import('pkg/bar/*.js');
+
+module.exports = async function () {
+    await promise.all([scoped, unscoped]);
+    return scoped.a + scoped.b + unscoped.x + unscoped.y;
```
devongovett: I think it should be `Promise.all`? And each key on the returned object from `import()` is a function that also resolves to a promise, so this should be `await scoped.a() + await scoped.b() ...`

packages/core/integration-tests/test/glob.js
```diff
@@ -203,4 +203,39 @@ describe('glob', function () {
...
+  it('should require a glob of files from a package', async function () {
+    let b = await bundle(
+      path.join(__dirname, '/integration/glob-package/index.js'),
+    );
+    await assertBundles(b, [
+      {
+        name: 'index.js',
+        assets: ['*.js', '*.js', 'a.js', 'b.js', 'x.js', 'y.js', 'index.js'],
+      },
+    ]);
```
devongovett, line 216: Probably should `run` the resulting bundle here as in the other tests to ensure it actually works

packages/resolvers/glob/src/GlobResolver.js
```diff
@@ -33,26 +55,65 @@ export default (new Resolver({
...
-              ]
-            : undefined,
-        },
+      throw errorToThrowableDiagnostic(error, dependency);
+    }
+
+    // if the specifier does not start with /, ~, or . then it's not a path but package-ish - we resolve
+    // the package first, and then append the rest of the path
+    if (!/^[/~.]/.test(specifier)) {
+      specifier = path.normalize(specifier);
```
devongovett: I _think_ globs are supposed to always use `/` as a separator, even on Windows. See https://github.com/micromatch/micromatch#backslashes. So this might not be needed.

devongovett (commented):
Looks good thanks!

devongovett (approved):
Thanks, this looks great!

PR #8108, opened 2022-05-18
Control hmr protocol via --hmr-secure cli flag

remotezygote:
# ↪️ Pull Request

This allows for setting of both the hmr host and the hmr protocol via the command line. This is needed for some environments.

## ✔️ PR Todo


devongovett:
What's the use case for needing a different protocol for the dev server and hmr? You can already use the `--https` option for this.

remotezygote:
The use case is when SSL is terminated in a gateway or other proxy in front. For example, I use Traefik as an application gateway and terminate SSL there, so to the outside world, it's https/wss, but inside it should not actually terminate https and should run pure http/ws.

packages/core/parcel/src/cli.js
```diff
@@ -117,6 +117,7 @@ var hmrOptions = {
   '--key <path>': 'path to private key to use with HTTPS',
   '--hmr-port <port>': ['hot module replacement port', process.env.HMR_PORT],
   '--hmr-host <host>': ['hot module replacement host', process.env.HMR_HOST],
+  '--hmr-secure <https>': ['hot module replacement use https?', process.env.HMR_SECURE],
```
remotezygote:
```suggestion
  '--hmr-secure <https>': ['use wss protocol for hot module replacement', process.env.HMR_SECURE],
```

remotezygote:
Most likely this is a better way to do it: https://github.com/parcel-bundler/parcel/pull/8137

101arrowz:
Superseded by #8137, and we found that even there the target use case is already satisfied by default Parcel.

Issue #7585, opened 2022-01-20
`parcel build` hangs on a random file


rjdellecese:
# 🐛 bug report

`parcel build` hangs, on a seemingly random file. My problem sounds identical to the one described in https://github.com/parcel-bundler/parcel/issues/6760. Passing the `--no-optimize` flag causes the build to succeed. I've also, through testing different configurations in my `.parcelrc`, discovered that removing the `@parcel/transformer-js` transformer allows the build to succeed, too (although the resulting build is unusable). `parcel serve` works fine.

My project comprises Elm and TypeScript (no `.js`, and literally one `.css` file with two lines, both of which are `@import`s).

## 🎛 Configuration

`package.json` excerpt

```json
{
  ...
  "engines": {
    "node": "^17.3.0"
  },
  "source": "src/index.html",
  "browserslist": "> 0.5%, last 2 versions, not dead",
  "pwaManifest": {
    ...
  },
  "devDependencies": {
    ...
    "@parcel/packager-raw-url": "^2.2.1",
    "@parcel/packager-xml": "^2.2.1",
    "@parcel/service-worker": "^2.2.1",
    "@parcel/transformer-elm": "^2.2.1",
    "elm": "latest-0.19.1",
    "parcel": "^2.2.1",
    "parcel-config-pwa-manifest": "^0.1.0",
    "typescript": "^4.5.4"
    ...
  },
  ...
}
```

`.parcelrc`

```json
{
  "extends": ["@parcel/config-default", "parcel-config-pwa-manifest"],
}
```

## 🤔 Expected Behavior

`parcel build` succeeds without incident.

## 😯 Current Behavior

`parcel build` hangs indefinitely on a random file.

https://user-images.githubusercontent.com/13547385/150435837-9f204470-961b-4a1f-9217-0c9ed1d5622c.mov

`parcel build --no-optimize` succeeds. `parcel serve` works fine.

## 🔦 Context

I just transitioned this project from Webpack 5. Everything has worked swimmingly, except for this!

## 💻 Code Sample

Unfortunately, the project in which I'm experiencing this issue is private, and I can't share the source code. I'm also not sure how to narrow the issue down any further than I have, but am more than willing to try if given suggestions on how!

## 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Parcel           | 2.2.1
| Node             | 17.3.0
| npm/Yarn         | 8.3.0
| Operating System | macOS Monterey 12.1

[...]

PR #8134, opened 2022-05-24
Update `transformer-elm` to use `swc` instead of `terser`

balazs-lengyel:
# ↪️ Pull Request

The elm transformer is broken for at least some bigger elm applications (see for example https://github.com/parcel-bundler/parcel/issues/7585). The reason is that terser doesn't really like the JS output by the elm compiler (see for example https://discourse.elm-lang.org/t/ot-webpack-terser-build-issue/7606)

In the discussion for https://github.com/parcel-bundler/parcel/issues/7585 @rjdellecese wrote about using `swc` instead of `terser` to solve the issue without turning of optimization, but there was no code shared. So I tried to recreate the work from the description.

The only issue have with this PR is that my practical JS knowledge is limited and I couldn't get the plugin to work locally for testing. It should be a trivial change though. Any feedback/pointer on how to get a local version of `transformer-elm` working on a nix-based system are appreciated!

## 💻 Examples

No publicly shareable codebase, but the issue seems to happen whenever terser is used to optimize elm applications above a certain size/complexity.

## ✔️ PR Todo

### DIFF
--- a/packages/transformers/elm/package.json
+++ b/packages/transformers/elm/package.json
@@ -22,12 +22,12 @@
   "dependencies": {
     "@parcel/diagnostic": "2.6.2",
     "@parcel/plugin": "2.6.2",
+    "@swc/core": "^1.2.218",
     "command-exists": "^1.2.8",
     "cross-spawn": "^7.0.3",
     "elm-hot": "^1.1.5",
     "node-elm-compiler": "^5.0.5",
-    "nullthrows": "^1.1.1",
-    "terser": "^5.14.2"
+    "nullthrows": "^1.1.1"
   },
   "peerDependencies": {
     "elm": "^0.19.1-5"
--- a/packages/transformers/elm/src/ElmTransformer.js
+++ b/packages/transformers/elm/src/ElmTransformer.js
@@ -3,7 +3,8 @@
 import {Transformer} from '@parcel/plugin';
 import spawn from 'cross-spawn';
 import path from 'path';
-import {minify} from 'terser';
+// $FlowFixMe
+import {minify} from '@swc/core';
 import ThrowableDiagnostic, {md} from '@parcel/diagnostic';
 // $FlowFixMe
 import elm from 'node-elm-compiler';
--- a/yarn.lock
+++ b/yarn.lock
@@ -2020,18 +2020,6 @@
   resolved "https://registry.yarnpkg.com/@napi-rs/cli/-/cli-2.6.2.tgz#9b6e3e8649c93d138cd8df8072cea8ddcf2ac6e6"
   integrity sha512-EmH+DQDEBUIoqMim0cc+X96ImtcIZLFjgW5WWORpzYnA9Ug7zNPO7jCLMhIQRd/p5AdWaXrT4SVXc/aip09rKQ==
 
-"@napi-rs/triples@^1.0.3":
-  version "1.0.3"
-  resolved "https://registry.yarnpkg.com/@napi-rs/triples/-/triples-1.0.3.tgz#76d6d0c3f4d16013c61e45dfca5ff1e6c31ae53c"
-  integrity sha512-jDJTpta+P4p1NZTFVLHJ/TLFVYVcOqv6l8xwOeBKNPMgY/zDYH/YH7SJbvrr/h1RcS9GzbPcLKGzpuK9cV56UA==
-
-"@node-rs/helper@^1.0.0":
-  version "1.2.1"
-  resolved "https://registry.yarnpkg.com/@node-rs/helper/-/helper-1.2.1.tgz#e079b05f21ff4329d82c4e1f71c0290e4ecdc70c"
-  integrity sha512-R5wEmm8nbuQU0YGGmYVjEc0OHtYsuXdpRG+Ut/3wZ9XAvQWyThN08bTh2cBJgoZxHQUPtvRfeQuxcAgLuiBISg==
-  dependencies:
-    "@napi-rs/triples" "^1.0.3"
-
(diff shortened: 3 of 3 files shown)


mischnic:
There's a paragraph in https://github.com/parcel-bundler/parcel/blob/v2/CONTRIBUTING.md#getting-started-with-bug-fixing:

> If you want to test out your change outside the monorepo, you can run `/path/to/monorepo/packages/core/parcel/src/bin.js build src/index.html` (provided that you don't have any `@parcel/*` plugins installed in this project).

mischnic:
Looks like swc expects the options to have a different format.

```
  1) elm
       should remove debugger in production:
     Failed to deserialize swc::config::JsMinifyOptions from json string (`{"minify":true,"jsc":{"minify":{"compress":{"keep\_fargs":false,"passes":2,"pure\_funcs":["F2","F3","F4","F5","F6","F7","F8","F9","A2","A3","A4","A5","A6","A7","A8","A9"],"pure\_getters":true,"unsafe":true,"unsafe\_comps":true},"mangle":{"reserved":["F2","F3","F4","F5","F6","F7","F8","F9","A2","A3","A4","A5","A6","A7","A8","A9"]}}}}`)

Caused by:
    unknown field `minify`, expected one of `compress`, `mangle`, `format`, `ecma`, `keepClassnames`, `keepFnames`, `module`, `safari10`, `toplevel`, `sourceMap`, `outputPath`, `inlineSourcesContent` at line 1 column 9
  Error: Failed to deserialize swc::config::JsMinifyOptions from json string (`{"minify":true,"jsc":{"minify":{"compress":{"keep_fargs":false,"passes":2,"pure_funcs":["F2","F3","F4","F5","F6","F7","F8","F9","A2","A3","A4","A5","A6","A7","A8","A9"],"pure_getters":true,"unsafe":true,"unsafe_comps":true},"mangle":{"reserved":["F2","F3","F4","F5","F6","F7","F8","F9","A2","A3","A4","A5","A6","A7","A8","A9"]}}}}`)
```

rjdellecese:
Hi @balazs-lengyel, thanks for following up on this!

Here are the options I'm providing to the `swc` `minify` function in my forked plugin:

```js
  {
    compress: {
      keep_fargs: false,
      passes: 2,
      pure_funcs: elmPureFuncs,
      pure_getters: true,
      unsafe_arrows: true,
      unsafe_comps: true,
      unsafe_Function: true,
      unsafe_math: true,
      unsafe_symbols: true,
      unsafe_methods: true,
      unsafe_proto: true,
      unsafe_regexp: true,
      unsafe_undefined: true,
    },
    mangle: false,
  }
```

It diverges more from the Terser options currently used by the Parcel Elm transformer because I based it off of the Terser configuration that I was previously using with Webpack.

balazs-lengyel:
Thanks @mischnic! It seems that I have gone full circle until I understood the error messages and what I needed to do. Now I've rebased and force pushed

Thanks @rjdellecese for sharing, did you need any of the additional settings for elm specific reasons or was it just for your project specific needs?

rjdellecese:
> Thanks @rjdellecese for sharing, did you need any of the additional settings for elm specific reasons or was it just for your project specific needs?

My project began its life using Webpack via https://github.com/halfzebra/create-elm-app, so it was pretty much just a translation of what had been configured there—that is, I have no particular reason to believe that just using the previously-specified options in Parcel's existing Elm transformer wouldn't work just fine (unless I'm forgetting something).

balazs-lengyel:
@mischnic: Updated - can you restart CI?

balazs-lengyel:
I don't see how to fix further tests (failed HMR/css) - can you give me a hint on how to approach that?

mischnic:
You can ignore them, these tests are flakey

balazs-lengyel:
> You can ignore them, these tests are flakey

If there is nothing else to do, I'd say this is done!

mischnic:
So what exactly is the motivation here? That for some inputs, Terser is very slow/doesn't terminate at all in a reasonable time?
Also, the reason Parcel itself still uses Terser and not swc for minification is that swc's output is sometimes bigger (and occasionally also broken)

balazs-lengyel:
> So what exactly is the motivation here? That for some inputs, Terser is very slow/doesn't terminate at all in a reasonable time? Also, the reason Parcel itself still uses Terser and not swc for minification is that swc's output is sometimes bigger (and occasionally also broken)

Yes, terser doesn't terminate for elm programs after they reach (some unknown) size/complexity or use a pattern where the elm compiler puts too much stress on terser.

While testing on a private project and waiting for this bug to occur, I found that the swc minified version was in the same ballpark as the terser one (usually much less then 1kB difference on the up to ~50kB results)

This has hit:
- me while migrating from webpack in two codebases at work
- me once in the private project a few weeks later
- [@rjdellecese as well](https://github.com/parcel-bundler/parcel/issues/7585) 
- at least one [elm/webpack user](https://discourse.elm-lang.org/t/ot-webpack-terser-build-issue/7606) as well. Probably some more who haven't reported this.

This PR fixed the build issue I had, with the only caveat that i couldn't test the full application as I needed to comment out some parts of the js/css while doing the `/path/to/monorepo/packages/core/parcel/src/bin.js build`-style build.

balazs-lengyel:
@mischnic just updated to rebase onto v2 with https://github.com/parcel-bundler/parcel/pull/8076 merged - can we merge this?

balazs-lengyel:
never mind! I just realized that this seems to be fixed in terser!

rjdellecese:
In https://github.com/terser/terser/issues/1212, it seems?

Thank you @ChristophP for reporting the issue!

PR #8137, opened 2022-05-24
Allow setting of full hmr connect uri from cli

### LINKED ISSUE #5268: --hmr-host is not available
# 🐛 bug report

I'm trying to configure the host that hot module replacement responds, since I am developing in a remote machine but accessing from local browser.

Zero configuration

## 🤔 Expected Behavior

Running `parcel serve ... --host 0.0.0.0 --hot-host 12.34.56.78` should make the hot module replacement connect to the socket in `wss://12.34.56.78:1234` instead of `wss://127.0.0.1:1234` or `wss://0.0.0.0:1234`

## 😯 Current Behavior

It errors with
```
error: unknown option '--hot-host'
```

Even if prepending `--hot` it does not work:
```
error: unknown option '--hot-host'
```

However, the README states those flags exist, and I see that [`HMRRuntime.js`](https://github.com/parcel-bundler/parcel/blob/v2/packages/runtimes/hmr/src/HMRRuntime.js) tries to use these. I assume that `--hot-port` will also not work.

## 🌍 Your Environment

| Parcel           | v2.0.0-nightly.429
| Node             | v14.12.0
| npm/Yarn         | v1.22.5
| Operating System | Ubuntu 20.04.1 LTS

remotezygote:
# ↪️ Pull Request

This allows setting of the entire web socket connect URI for environments where there is not a direct mapping from the serving of web sockets and how you address them (such as when behind a reverse proxy or gateway).

Fixes #5268 more fully.

## 💻 Examples

You have a gateway that forwards ws.domain:443 (and SSL is terminated here) to your internal service at parcel.internal:6464 (no SSL). From the outside world, the websocket connect url should be `wss://ws.domain`, but the service itself should listen with no ssl on port 6464. The current options for `host` and `port` do not allow for this, as they control both sides of the equation, which should not match.

## 🚨 Test instructions

Set `--hmr-uri` to a uri that maps to your parcel hmr service, and see that the frontend uses the specified url to connect, while the service listens according to the other parameters.

## ✔️ PR Todo


remotezygote:
I can't get `yarn lint` to work... and the ci output is hidden or something, so not sure there, sorry!

101arrowz:
I don't quite understand why this is necessary except in very niche cases. You can already use HMR over an HTTP proxy if your HMR port and serve port are the same. That way is also easier, you just need to reverse proxy a single port.

remotezygote:
This is all about how the _frontend_ code it embeds in your bundle during development addresses the websocket connection. Currently it can only connect the same way it listens, so if it needs to connect via a different port than it listens on, you're out of luck, or if you want it to use a different hostname, you're out of luck (until my recent PR that was merged). I don't think this is niche at all - it may not be common in your world, but it is common in the world, and I don't see an issue with merging this - it hurts nothing and helps at least me and this person: https://github.com/parcel-bundler/parcel/issues/5268#issuecomment-1039145193

Also the work is done. Whether you need it or not, why not merge it if I need it?

101arrowz:
I cannot merge PRs, only the core team can really do that. I said this was niche because you already have full control over the HMR host and HMR port.

> if it needs to connect via a different port than it listens on, you're out of luck

This is my main confusion, why do you need the HMR port and serving port to be different? Keep in mind that if you're using a reverse proxy, the frontend takes that into account by using the same port that the current page is on so that the HMR requests still go through the reverse proxy.

> if you want it to use a different hostname, you're out of luck

Does `--hmr-host` not work for you? We have logic [specifically to support this use case](https://github.com/parcel-bundler/parcel/blob/v2/packages/runtimes/hmr/src/HMRRuntime.js#L28-L40).

remotezygote:
Sorry, forgot to mention protocol. In my case I also have SSL terminated in front of the service, so while the service should not boot up with SSL, it should address it with `wss://` even though the service has https/ssl disabled. This is all in the PR description above. When you need a different host, port, and protocol, as the linked issue also says, the better solution is to offer this support of setting the full uri.

101arrowz:
I've done exactly what you're discussing before and I had no issues with the default configuration. In [this block](https://github.com/parcel-bundler/parcel/blob/v2/packages/runtimes/hmr/src/loaders/hmr-runtime.js#L86-L91), we check if you're on an HTTPS page and connect with `wss:` if you are. I feel like your provided example should run without issues if you set up your gateway and specify `--port 6464` for Parcel without specifying either `--hmr-port` or `--hmr-host`.

The only case I can think of where the current logic breaks is having no SSL on the webpage but having SSL on the HMR, which just seems...weird?

> This is all in the PR description above.

I know, I read it :)

remotezygote:
Wow! Holy crap that works. Thank you! It makes sense now after looking at the code yet again, but this might be something worth mentioning somewhere in docs (here maybe: https://parceljs.org/features/development/#hot-reloading). Maybe it's just me, but coming from other bundlers (webpack, fuse.js), I hadn't even tried that since I knew I was in a strange environment. My straight-forward conversion from another bundler kinda bit me a little. It's maybe worth mentioning that it will share the serve port for HMR if a port is not specified - that's the key to this, not having a separate port, even though it feels like you need one (and in other worlds you definitely do need one).

Honestly, I still see a case for merging this, but you have me going now, so I'm fine with whatever.

MLNW:
I might have a similar issue that I currently cannot find a way for hot reloading to work:

On a development machine we have an nginx running on port 8080 that serves the built HMI via the `dist` folder.
Developers connect to `http://localhost:8080` to access the HMI which first redirects to a keycloak for authentication and then redirects back to `http://localhost:8080`. Parcel is running on `http://localhost:1234` and rebuilds for any changes but to see these changes in the HMI one needs to reload the page manually.

Without any configuration for the HMR I see the warning in the console: `[parcel] 🚨 Connection to the HMR server was lost`. Around the same time an error occurs stating that `Firefox is unable to connect to the server via ws://localhost:8080`. Now when I configure `--hmr-port 1234` the same error message appears.

Is there a way to get HMR working with this setup?

101arrowz:
@MLNW I'm not sure I understand your setup properly, but you probably need to enable WebSocket forwarding within Nginx.

devongovett:
Let's wait until something comes up again. I think adding this option kinda conflicts with a bunch of other ones - e.g. what happens if you set both `--hmr-uri` and `--hmr-port` - which one wins? Perhaps adding the one missing piece which is `--hmr-secure` would be better, but as described above, it seems like it might not really be needed.

PR #8146, opened 2022-05-26
fix: eslint validator (upgraded to v8)

### LINKED ISSUE #8063: @parcel/validator-eslint won't work with ESLint v8+
# 🐛 bug report

The `@parcel/validator-eslint` plugin works fine with ESLint version 7 and earlier, but it breaks with version 8.

**index.js**
```js
// This unused variable should trigger an ESLint error.
const unusedVar = "Hello!";
```

**.parcelrc**
```json
{
  "validators": {
    "*.{js,mjs,jsm,jsx,es6,cjs,ts,tsx}": [
      "@parcel/validator-eslint"
    ]
  }
}
```

**.eslintrc.json**
```
{
    "env": {
        "browser": true
    },
    "extends": [
        "eslint:recommended"
    ],
    "parserOptions": {
        "ecmaVersion": 2020,
        "sourceType": "module"
    }
}
```

## 🤔 Expected Behavior

`@parcel/validator-eslint` should work with eslint v8, the same as it does with v7, by throwing this error:
```
🚨 Build failed.

@parcel/validator-eslint: ESLint found 1 errors and 0 warnings.

  C:\Users\ansteg\Projects\parcel-eslint-example\src\index.js:2:7
    1 | // This unused variable should trigger an ESLint error.
  > 2 | const unusedVar = "Hello!";
  >   |       ^^^^^^^^^^ 'unusedVar' is assigned a value but never used.
    3 |
```

## 😯 Current Behavior

With eslint v8, you see this error instead:

```
🚨 Build failed.

@parcel/validator-eslint: Must use import to load ES Module:
C:\Users\ansteg\Projects\parcel-eslint-example\node_modules\@eslint\eslintrc\universal.js
require() of ES modules is not supported.
require() of
C:\Users\ansteg\Projects\parcel-eslint-example\node_modules\@eslint\eslintrc\universal.js
from
C:\Users\ansteg\Projects\parcel-eslint-example\node_modules\eslint\lib\linter\linter.js
is an ES module file as it is a .js file whose nearest parent package.json
contains "type": "module" which defines all .js files in that package scope as ES
modules.
Instead rename universal.js to end in .cjs, change the requiring code to use
import(), or remove "type": "module" from
C:\Users\ansteg\Projects\parcel-eslint-example\node_modules\@eslint\eslintrc\package.json.


  Error [ERR_REQUIRE_ESM]: Must use import to load ES Module:
  C:\Users\ansteg\Projects\parcel-eslint-example\node_modules\@eslint\eslintrc\universal.js
  require() of ES modules is not supported.
  require() of
  C:\Users\ansteg\Projects\parcel-eslint-example\node_modules\@eslint\eslintrc\universal.js
  from
  C:\Users\ansteg\Projects\parcel-eslint-example\node_modules\eslint\lib\linter\linter.js
  is an ES module file as it is a .js file whose nearest parent package.json
  contains "type": "module" which defines all .js files in that package scope as
  ES modules.
  Instead rename universal.js to end in .cjs, change the requiring code to use
  import(), or remove "type": "module" from
  C:\Users\ansteg\Projects\parcel-eslint-example\node_modules\@eslint\eslintrc\package.json.

  at new NodeError (internal/errors.js:322:7)
  at Object.Module._extensions..js (internal/modules/cjs/loader.js:1102:13)
  at Module.load (internal/modules/cjs/loader.js:950:32)
  at NodePackageManager.load
  (C:\Users\ansteg\Projects\parcel-eslint-example\node_modules\@parcel\package-manager\lib\index.js:3376:15)
  at NodePackageManager.requireSync
  (C:\Users\ansteg\Projects\parcel-eslint-example\node_modules\@parcel\package-manager\lib\index.js:3354:21)
  at Module.m.require
  (C:\Users\ansteg\Projects\parcel-eslint-example\node_modules\@parcel\package-manager\lib\index.js:3367:25)
  at require (internal/modules/cjs/helpers.js:93:18)
  at Object.<anonymous>
  (C:\Users\ansteg\Projects\parcel-eslint-example\node_modules\eslint\lib\linter\linter.js:27:9)
  at Module._compile (internal/modules/cjs/loader.js:1085:14)
  at Object.Module._extensions..js (internal/modules/cjs/loader.js:1114:10)
```

In both versions (v8 and v7) using eslint from the command line works fine.

## 💁 Possible Solution

In v7, the `@eslint/eslintrc` `package.json` looked like this:
```
{
  "name": "@eslint/eslintrc",
  "version": "0.4.3",
  "main": "lib/index.js",
   ...
}
```
...and `lib/index.js` looks like a standard `commonjs` file with `module.exports =` and `require` statements.

In v8, that same package.json looks like this:
```
{
  "name": "@eslint/eslintrc",
  "version": "1.2.2",
  "type": "module",
  "main": "./dist/eslintrc.cjs",
  "exports": {
    ".": {
      "import": "./lib/index.js",
      "require": "./dist/eslintrc.cjs"
    },
    "./package.json": "./package.json",
    "./universal": {
      "import": "./lib/index-universal.js",
      "require": "./dist/eslintrc-universal.cjs"
    }
  },
  ...
}
```

And `universal.js` looks like this:
```js
// Jest (and probably some other runtimes with custom implementations of
// `require`) doesn't support `exports` in `package.json`, so this file is here
// to help them load this module. Note that it is also `.js` and not `.cjs` for
// the same reason - `cjs` files requires to be loaded with an extension, but
// since Jest doesn't respect `module` outside of ESM mode it still works in
// this case (and the `require` in _this_ file does specify the extension).

// eslint-disable-next-line no-undef
module.exports = require("./dist/eslintrc-universal.cjs");
```

So it looks like there was a change on the eslint side to use the new [exports](https://nodejs.org/api/packages.html#exports) package.json field. I'm not familiar with the details of how this is expected to work v. how parcel does it, but this is probably where you'd start looking for a root cause.

## 🔦 Context

I was trying to help answer [these](https://stackoverflow.com/questions/72103413/how-to-get-react-warnings-in-parcel-react-extension) [stackoverflow questions](https://stackoverflow.com/questions/71375095/how-do-i-configure-parcel-to-exit-build-with-an-error-if-eslint-does-not-validat)

## 💻 Code Sample

You can see an example of the broken behavior in [the v8 branch of this github repo](https://github.com/astegmaier/parcel-eslint-example/tree/v8). The [master branch](https://github.com/astegmaier/parcel-eslint-example/tree/master) shows v7 working fine.

## 🌍 Your Environment

| Parcel           | 2.5.0
| Node             | 14.18.2
| Yarn         | 1.22.17
| Operating System | Windows 11 Version 10.0.22000 Build 22000

VladimirMikulic:
# ↪️ Pull Request

ESLint validator fixed and also upgraded to the latest version 8.

Resolves: #8063

### DIFF
--- a/packages/validators/eslint/package.json
+++ b/packages/validators/eslint/package.json
@@ -25,9 +25,9 @@
     "chalk": "^4.1.0"
   },
   "devDependencies": {
-    "eslint": "^6.0.0"
+    "eslint": "^8.16.0"
   },
   "peerDependencies": {
-    "eslint": "^6.0.0"
+    "eslint": ">=6.0.0"
   }
 }
--- a/packages/validators/eslint/src/EslintValidator.js
+++ b/packages/validators/eslint/src/EslintValidator.js
@@ -4,32 +4,37 @@ import {type DiagnosticCodeFrame, escapeMarkdown} from '@parcel/diagnostic';
 import eslint from 'eslint';
 import invariant from 'assert';
 
-let cliEngine = null;
+let eslintInstance = null;
 
 export default (new Validator({
   async validate({asset}) {
-    if (!cliEngine) {
-      cliEngine = new eslint.CLIEngine({});
+    const isPostV8 = typeof eslint.ESLint !== 'undefined';
+
+    if (!eslintInstance) {
+      eslintInstance = isPostV8 ? new eslint.ESLint() : new eslint.CLIEngine();
     }
-    let code = await asset.getCode();
 
-    invariant(cliEngine != null);
-    let report = cliEngine.executeOnText(code, asset.filePath);
+    invariant(eslint != null);
 
-    let validatorResult = {
+    const code = await asset.getCode();
+    const report = isPostV8
+      ? eslintInstance.lintText(code, {filePath: asset.filePath})
+      : eslintInstance.executeOnText(code, asset.filePath);
+
+    const validatorResult = {
       warnings: [],
       errors: [],
     };
 
     if (report.results.length > 0) {
-      for (let result of report.results) {
+      for (const result of report.results) {
         if (!result.errorCount && !result.warningCount) continue;
 
-        let codeframe: DiagnosticCodeFrame = {
+        const codeframe: DiagnosticCodeFrame = {
           filePath: asset.filePath,
           code: result.source,
           codeHighlights: result.messages.map(message => {
-            let start = {
+            const start = {
               line: message.line,
               column: message.column,
             };
@@ -48,7 +53,7 @@ export default (new Validator({
           }),
         };
 
-        let diagnostic = {
+        const diagnostic = {
           origin: '@parcel/validator-eslint',
           message: `ESLint found **${result.errorCount}** __errors__ and **${result.warningCount}** __warnings__.`,
           codeFrames: [codeframe],


VladimirMikulic:
@mischnic mind taking a look? Thanks!

mischnic:
This would be a breaking change though for projects not using eslint 8 (yet)

VladimirMikulic:
Thanks for the fast reply @mischnic. That's a good point.
When could I expect the change to land? Parcel v3 or?

VladimirMikulic:
@mischnic ping! Thanks!

kyle1320:
I would like this as well. Could we support multiple versions of eslint (e.g. `>=6.0.0 <9.0.0`) and do feature checks to support them all?

VladimirMikulic:
I'd be happy with that approach as well. @mischnic any thoughts?

devongovett:
Either that or release 3.0.0 of only this plugin. It's not in the default Parcel config, so users install it anyway. If we can have a single version that works with either version of eslint I think that's better though because then people don't need to worry about which version of the parcel plugin corresponds to which version of eslint.

VladimirMikulic:
Sorry for wait time here - now it's ready for review. (cc: @mischnic) Thanks!

VladimirMikulic:
I didn't build the whole project the first time around, but now after testing, I get the error as in #8063.
It's definitely not eslint's error though. eslint works in a random JS file like:
```js
import { ESLint } from 'eslint';
// const { ESLint } = require("eslint") - also works
const eslint = new ESLint();
eslint.lintText('const a = 2;').then(console.log);
```


However, it fails when the file is required by Parcel as a plugin. (neither import works)
I'm not sure if Parcel transpiles modules which could lead to this error?

VladimirMikulic:
With this being said, I'd be up for fixing this, but I'm just not sure where to look?

mattsputnikdigital:
Im getting errors with node modules that are made with v8

🚨 Build failed.

@parcel/validator-eslint: Failed to load config "@gilbarbara/eslint-config" to extend from.
Referenced from: /node_modules/react-inlinesvg/package.json

PR #8147, opened 2022-05-26
fix: transforming node:builtin when targeting commonjs

### LINKED ISSUE #8049: Transforming node:builtin to built-in when targeting commonjs
# 🙋 feature request

Currently, Parceljs silently uses `node:builtin` inside commonjs although this doesn't work. There should be a transformer that converts this to normal nodejs imports when targeting commonjs.

```js
import { execFileSync } from "node:fs"
```

## 🤔 Expected Behavior

```js
const { execFileSync } = require("fs")
```

## 😯 Current Behavior

```js
const { execFileSync } = require("node:fs")
```

aminya:
# ↪️ Pull Request

Fixes #8049 

> Currently, Parceljs silently uses `node:builtin` inside commonjs although this doesn't work. There should be a transformer that converts this to normal nodejs imports when targeting commonjs.
> 
> ```js
> import { execFileSync } from "node:fs"
> ```
> 
> ## 🤔 Expected Behavior
> 
> ```js
> const { execFileSync } = require("fs")
> ```
> 
> ## 😯 Previous Behavior
> 
> ```js
> const { execFileSync } = require("node:fs")
> ```

## ✔️ PR Todo


devongovett:
Can you please add a test?

aminya:
I did, but I realized that this doesn't fix the issue fully. I added an alias property. The architecture of the code does not allow to do alias transformation inside the resolving code. It also seems that AssertGroup doesn't accept something like an alias.

aminya:
One cruel workaround is to create simple built-in Node packages that export the module, and then use them for resolving `node:` when the output is commonjs:

For example,

`node:fs` can be resolved to this package:

@parcel/node-builtin-fs
```js
exports = require("fs")
```

aminya:
@devongovett do you want to go with the workaround route, or do you want to change the API of the resolver?

PR #8149, opened 2022-05-26
fix: force the use of posix path separators for ts module graphs

GeeWizWow:
# ↪️ Pull Request

When resolving the name of the main module for use in TS module graphs, ensure we force the usage of posix paths as this is what `typescript` uses and returns.

This fixes an issue when running on Windows, where an assert would be thrown as a result of two identical paths using different path separators being compared for equality.

For some reason I'm unaware of, this only occurs if a `tsconfig.json` file with a `paths` entry is present in the root folder.

## 💻 Examples

The main module name for ts module graphs is resolved here:
https://github.com/parcel-bundler/parcel/blob/f2d0a3a27d6e493b23ddc2edbc8a4c0053ff34ab/packages/transformers/typescript-types/src/TSTypesTransformer.js#L54-L57

On windows, this returns a path with dos path separators `\`. 

Subsequently in `collect.js` when we use `typescript` to visit a node,  it returns the node's name using posix separators `/`, when we add that node to the module graph:

https://github.com/parcel-bundler/parcel/blob/f2d0a3a27d6e493b23ddc2edbc8a4c0053ff34ab/packages/transformers/typescript-types/src/collect.js#L26

It then attempts to compare it to the main module name resolved in `TSTypesTransformer` earlier:

https://github.com/parcel-bundler/parcel/blob/f2d0a3a27d6e493b23ddc2edbc8a4c0053ff34ab/packages/transformers/typescript-types/src/TSModuleGraph.js#L19-L24

Which fails due to the path separator mismatch.

Fast forward a bit, and `getAllExports` is called, and the assert is thrown due to `nullthrows(this.mainModule),` still being undefined/ null.

## 🚨 Test instructions

* Ensure you are running on windows.
* Follow the instructions for building a library (source, main and types entries in package.json)
* Ensure your source file is a typescript file that exports from another module, for example:

```
export * from './library.ts';
```

* Have `tsconfig.json` file in the same directory as your package.json. This file can contain anything, as long as it contains an arbitrary entry for `compilerOptions.paths`
* Run `parcel build`

## ✔️ PR Todo

### DIFF
--- a/packages/core/integration-tests/test/integration/ts-types/windows-paths/expected.d.ts
+++ b/packages/core/integration-tests/test/integration/ts-types/windows-paths/expected.d.ts
@@ -1,0 +1,18 @@
+type Params = {
+    hello: number;
+};
+interface Hello {
+    yo: string;
+}
+export class Test {
+    test(hello: Hello): string;
+}
+export function test(params: Params): number;
+export function foo(): number;
+export var x: number;
+export var hi: number;
+export declare module mod {
+    function bar(): number;
+}
+
+//# sourceMappingURL=types.d.ts.map
--- a/packages/core/integration-tests/test/integration/ts-types/windows-paths/index.ts
+++ b/packages/core/integration-tests/test/integration/ts-types/windows-paths/index.ts
@@ -1,0 +1,1 @@
+export * from './lib/index';
--- a/packages/core/integration-tests/test/integration/ts-types/windows-paths/index.ts
+++ b/packages/core/integration-tests/test/integration/ts-types/windows-paths/index.ts
@@ -1,1 +1,1 @@
-export * from './lib/index';
+export * from './src/index';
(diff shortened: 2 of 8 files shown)


mischnic:
There are some existing utils we have for this operation:
https://github.com/parcel-bundler/parcel/blob/cc9e352a22ab931d305dea4715c44103a7152e24/packages/core/package-manager/src/NodeResolverBase.js#L16

Could you add a test case to this file? We do run a CI job on Windows so that would cover this bug https://github.com/parcel-bundler/parcel/blob/8823770717bf451da2667b4d0d24b8915698856d/packages/core/integration-tests/test/ts-types.js#L12

GeeWizWow:
Thanks @mischnic I'll pick up those tasks tomorrow. 

Can you shed any light into why this code path is only triggered in the presence of a `tsconfig.json` file containing a `paths` entry? Keen to understand the issue further...

GeeWizWow:
Updated with the above comments:
- `TSTypesTransformer` now uses `normalizeSeparators` from `@parcel/utils`
- Added an integration test in `packages\core\integration-tests\test\ts-types.js`

GeeWizWow:
Test failure with: 
```
 1) typescript types
       should handle a tsconfig file with paths on windows:
     Failed to resolve './lib/index' from './index.ts'
```

Will investigate.

Edit: ✅

devongovett (approved):

PR #8158, opened 2022-05-28
update react-error-overlay to avoid unnecessary process polyfill

josephshen:
update react-error-overlay to avoid unnecessary process polyfill when checking process.env and incomplete treeshaking

# ↪️ Pull Request

## ✔️ PR Todo

### DIFF
--- a/packages/runtimes/react-refresh/package.json
+++ b/packages/runtimes/react-refresh/package.json
@@ -22,7 +22,7 @@
   "dependencies": {
     "@parcel/plugin": "2.6.0",
     "@parcel/utils": "2.6.0",
-    "react-error-overlay": "6.0.9",
+    "react-error-overlay": "6.0.11",
     "react-refresh": "^0.9.0"
   }
 }


devongovett:
I had pinned it because the newer version has a bug: https://github.com/facebook/create-react-app/issues/11771

josephshen:
@parcel/resolver-default: Auto installing polyfill for Node builtin module 
"process"...
  /home/freeman/Projects/Front/node_modules/@parcel/runtime-react-refresh/node_modules/react-error-overlay/lib/index.js:1:4450
  > 1 | s[o]:process.cwd();if("string"!=typeof a)throw new TypeError("Arguments 
  >   |      ^^^^^^^ used here

this code seems trigger the  unnecessary process polyfill install

> I had pinned it because the newer version has a bug: [facebook/create-react-app#11771](https://github.com/facebook/create-react-app/issues/11771)

josephshen:
> I had pinned it because the newer version has a bug: [facebook/create-react-app#11771](https://github.com/facebook/create-react-app/issues/11771)

yep, just trigger this bug, so it seem right now we should just keep 6.0.9 :)

Issue #8116, opened 2022-05-20
Parcel transpiles dynamic import wrong, when wrapped into `Promise.resolve()`


SmartArray:
# 🐛 bug report

We are trying to import a module from a mono repo. That monorepo is setup correctly but it imports another library with `import()` wrapped into a `Promise.resolve()`

However, when parcel transpiles the code, we got a runtime error with `res is not defined`.

I successfully created a minimalistic example that helps us to reproduce the issue. It doesn't seem to be an issue with `babel`. Though, I hope you guys can help me to identify the source of the issue.

## 🎛 Configuration (.babelrc, package.json, cli command)
This issue can be reproduced with parcel's default babel config, so I removed it completely.
The package json looks like this:

```json
{
  "name": "xxx",
  "version": "1.0.0",
  "main": "index.js",
  "license": "MIT",
  "source": "index.html",
  "dependencies": {
    "prando": "^6.0.1"
  },
  "scripts": {
    "start": "parcel index.html",
    "babel": "babel index.js --out-file babel/script-compiled.js  --presets=@babel/preset-env"
  },
  "devDependencies": {
    "@babel/cli": "^7.17.10",
    "@babel/core": "^7.18.0",
    "@parcel/babel-preset": "^2.0.0-alpha.2",
    "parcel": "^2.5.0"
  }
}
```

My `index.html`:
```html
<!doctype html>
<html lang="en">
  <head>
    <meta charset="utf-8"/>
    <title>My First Parcel App</title>
    <script type="module" src="index.js"></script>
  </head>
  <body>
    <h1>Hello, World!</h1>
  </body>
</html>
```

And last but not least, my index.js:

```javascript
class Test {
  constructor() {
    this.someModule = null;
    this.seed = 123;
  }

  async loadInvalid() {
    const this$4 = this;

    if (!this.someModule) {
      // An import wrapped into Promise.resolve() causes issues. The callbacks 
      // will be ordered incorrectly.
      return Promise.resolve(import('prando').then(function (m) {
        var _m$default;

        return (_m$default = m == null ? void 0 : m["default"]) != null ? _m$default : m;
      })).then(function (SomeModule) {
        this$4.someModule = new SomeModule(this$4.seed);
      });
    } 
  }

  async loadValid() {
    const this$4 = this;

    if (!this.someModule) {
      return import('prando').then(function (m) {
        var _m$default;

        return (_m$default = m == null ? void 0 : m["default"]) != null ? _m$default : m;
[...]

PR #8167, opened 2022-05-29
Fix transformation of import/requires wrapped into Promise.resolve().

SmartArray:
# ↪️ Pull Request
Previously we noticed that parcel incorrectly transforms `Promise.resolve(import('foo')).then(...)` as described in https://github.com/parcel-bundler/parcel/issues/8116.

The transformer that rewrites the nodes of the code `Promise.resolve().then(() => return require('foo'))` erroneously restructures the above code too.
However, there is no need to transform this require/import statement, when wrapped into `Promise.resolve()` because it is used to promisify the import in the first place.

This PR adds a unit test to illustrate the issue, and a condition that prevents the incorrect processing of the statement. 

## 💻 Examples
```javascript
// main.js
module.exports = Promise.resolve(require('./async')).then(x => x + 1335);
```

```javascript
// async.js
module.exports = 2;
```

### Expected behaviour
```javascript
const n = await import('./main');
n === 1337
```

### Current behaviour
The current code causes the runtime to crash because it injects a variable called `res` incorrectly, and thus destroys the code.

## 🚨 Test instructions

```
yarn workspace @parcel/integration-tests test -g Promise.resolve
```

Or use the minimal reproduction code [here](https://parcel2-repl-mischnic.vercel.app/#JTdCJTIyZmlsZXMlMjIlM0ElNUIlNUIlMjIlMkZzcmMlMkZpbmRleC5qcyUyMiUyQyU3QiUyMnZhbHVlJTIyJTNBJTIyUHJvbWlzZS5yZXNvbHZlKGltcG9ydCgnLiUyRm90aGVyJykpLnRoZW4oeCUyMCUzRCUzRSUyMGNvbnNvbGUubG9nKHgpKSU1Q24lMjIlMkMlMjJpc0VudHJ5JTIyJTNBdHJ1ZSU3RCU1RCUyQyU1QiUyMiUyRnNyYyUyRm90aGVyLmpzJTIyJTJDJTdCJTIydmFsdWUlMjIlM0ElMjJleHBvcnQlMjBjb25zdCUyMHglMjAlM0QlMjAyJTNCJTIyJTdEJTVEJTVEJTJDJTIyb3B0aW9ucyUyMiUzQSU3QiUyMmVudHJpZXMlMjIlM0ElNUIlNUQlMkMlMjJtaW5pZnklMjIlM0FmYWxzZSUyQyUyMnNjb3BlSG9pc3QlMjIlM0FmYWxzZSUyQyUyMnNvdXJjZU1hcHMlMjIlM0FmYWxzZSUyQyUyMnB1YmxpY1VybCUyMiUzQSUyMiUyRl9fcmVwbF9kaXN0JTIyJTJDJTIydGFyZ2V0VHlwZSUyMiUzQSUyMmJyb3dzZXJzJTIyJTJDJTIydGFyZ2V0RW52JTIyJTNBJTIyQ2hyb21lJTIwOTUlMjIlMkMlMjJvdXRwdXRGb3JtYXQlMjIlM0ElMjJlc21vZHVsZSUyMiUyQyUyMmhtciUyMiUzQWZhbHNlJTJDJTIybW9kZSUyMiUzQSUyMmRldmVsb3BtZW50JTIyJTJDJTIycmVuZGVyR3JhcGhzJTIyJTNBZmFsc2UlMkMlMjJ2aWV3U291cmNlbWFwcyUyMiUzQWZhbHNlJTJDJTIyZGVwZW5kZW5jaWVzJTIyJTNBJTVCJTVEJTJDJTIybnVtV29ya2VycyUyMiUzQTAlN0QlMkMlMjJ1c2VUYWJzJTIyJTNBZmFsc2UlMkMlMjJicm93c2VyQ29sbGFwc2VkJTIyJTNBJTVCJTVEJTJDJTIydmlld3MlMjIlM0ElNUIlMjIlMkZzcmMlMkZpbmRleC5qcyUyMiUyQyUyMiUyRnNyYyUyRm90aGVyLmpzJTIyJTJDJTIyT3B0aW9ucyUyMiU1RCUyQyUyMmN1cnJlbnRWaWV3JTIyJTNBMiU3RA==)

## ✔️ PR Todo

### DIFF
--- a/packages/transformers/js/core/src/dependency_collector.rs
+++ b/packages/transformers/js/core/src/dependency_collector.rs
@@ -458,10 +458,17 @@ impl<'a> Fold for DependencyCollector<'a> {
               // Promise.resolve().then(() => require('foo'))
               // Promise.resolve().then(() => { return require('foo') })
               // Promise.resolve().then(function () { return require('foo') })
+              //   but not
+              // Promise.resolve(require('foo'))
               if let Call(call) = &*member.obj {
                 if let Callee::Expr(e) = &call.callee {
                   if let Member(m) = &**e {
-                    if match_member_expr(m, vec!["Promise", "resolve"], self.decls) {
+                    if match_member_expr(m, vec!["Promise", "resolve"], self.decls) &&
+                      // Make sure the arglist is empty.
+                      // I.e. do not proceed with the below unless Promise.resolve has an empty arglist
+                      // because build_promise_chain() will not work in this case.                   
+                      !call.args.get(0).is_some()
+                    {
                       if let MemberProp::Ident(id) = &member.prop {
                         if id.sym.to_string().as_str() == "then" {
                           if let Some(arg) = node.args.get(0) {
--- a/packages/transformers/js/core/src/dependency_collector.rs
+++ b/packages/transformers/js/core/src/dependency_collector.rs
@@ -467,7 +467,7 @@ impl<'a> Fold for DependencyCollector<'a> {
                       // Make sure the arglist is empty.
                       // I.e. do not proceed with the below unless Promise.resolve has an empty arglist
                       // because build_promise_chain() will not work in this case.                   
-                      !call.args.get(0).is_some()
+                      call.args.is_empty()
                     {
                       if let MemberProp::Ident(id) = &member.prop {
                         if id.sym.to_string().as_str() == "then" {
--- a/packages/core/integration-tests/test/integration/require-async/resolve-chain.js
+++ b/packages/core/integration-tests/test/integration/require-async/resolve-chain.js
@@ -1,0 +1,1 @@
+module.exports = Promise.resolve(require('./async')).then(x => x + 1335);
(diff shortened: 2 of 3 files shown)


packages/transformers/js/core/src/dependency_collector.rs
```diff
@@ -458,10 +458,17 @@ impl<'a> Fold for DependencyCollector<'a> {
...
+              // Promise.resolve(require('foo'))
               if let Call(call) = &*member.obj {
                 if let Callee::Expr(e) = &call.callee {
                   if let Member(m) = &**e {
-                    if match_member_expr(m, vec!["Promise", "resolve"], self.decls) {
+                    if match_member_expr(m, vec!["Promise", "resolve"], self.decls) &&
+                      // Make sure the arglist is empty.
+                      // I.e. do not proceed with the below unless Promise.resolve has an empty arglist
+                      // because build_promise_chain() will not work in this case.                   
+                      !call.args.get(0).is_some()
```
mischnic: Is there a reason you didn't do it like this? I had to think for a moment when I first read `!(...).get(0).is_some()`
```suggestion
                      call.args.is_empty()
```

SmartArray: I thought that it would catch ‘undefined’/‘null’ but that was probably stupid.

Will fix that when I get home from my vacation

devongovett (approved):

PR #8169, opened 2022-05-30
Fix glob resolver package issue with deep entry points

marcins:
# ↪️ Pull Request

This is a follow-up from #8097 - when trying to implement the change in my repository I realised there was a step missed, if the package entry point resolves to a file not in the root, then the paths for the glob import will need to be relative to the entry point, where they should be able to be relative to the package root.

This change adds an additional step to perform a `findPackage` on the resulting `filePath` in order to resolve to the root of the package.

## 💻 Examples

Before if you had a package `@scope/pkg` with an entry point of `./dist/index.js` then the glob resolver would resolve the root to `@scope/pkg/dist/`, so if you used a path like `@scope/pkg/src/foo*.js` it would actually resolve to `@scope/pkg/dist/src/foo*.js` which is incorrect.  With this change the path resolves as expected.

## 🚨 Test instructions

Failing integration test with a deep entry point was added, change was made and confirmed all integration tests still pass.

## ✔️ PR Todo

- [x] Added/updated ~~unit~~ integration tests for this change

### DIFF
--- a/packages/resolvers/glob/src/GlobResolver.js
+++ b/packages/resolvers/glob/src/GlobResolver.js
@@ -58,6 +58,9 @@ export default (new Resolver({
       throw errorToThrowableDiagnostic(error, dependency);
     }
 
+    let invalidateOnFileCreate = [];
+    let invalidateOnFileChange = new Set();
+
     // if the specifier does not start with /, ~, or . then it's not a path but package-ish - we resolve
     // the package first, and then append the rest of the path
     if (!/^[/~.]/.test(specifier)) {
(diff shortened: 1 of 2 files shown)


packages/resolvers/glob/src/GlobResolver.js
```diff
@@ -102,14 +102,30 @@ export default (new Resolver({
...
           dependency,
         );
       } else if (result.diagnostics) {
         throw new ThrowableDiagnostic({diagnostic: result.diagnostics});
       }
+      // We keep Flow happy by copying the filePath at a point we know it's defined
+      const filePath = result.filePath;
 
-      specifier = path.resolve(path.dirname(result.filePath), rest);
+      const ctx = {
```
marcins: I wasn't sure what to do about this - `findPackage` requires a context, so I provided one with the invalidation info from the step above, however when we return the resolution at the end of this method around line 178, there's only `invalidateOnFileCreate: [{glob: normalized}],` included. Should we collect all these additional invalidations and add them there as well?

packages/resolvers/glob/src/GlobResolver.js
```diff
@@ -102,14 +102,30 @@ export default (new Resolver({
...
+        invalidateOnFileCreate: result.invalidateOnFileCreate || [],
+        invalidateOnFileChange: new Set(result.invalidateOnFileChange),
+        specifierType: dependency.specifierType,
+        loc: dependency.loc,
+      };
+
+      const resolvedPackage = await resolver.findPackage(filePath, ctx);
```
devongovett: I'm not sure this is quite right because there could be another package.json inside a sub-directory of the module. We can use the `resolver.resolveModule` method for this I think.

devongovett (approved):

Issue #8071, opened 2022-05-09
Parcel WebExtension fails to bundle modules under some conditions (build + mv2, or mv3 &  --no-scope-hoist)


Stvad:
# 🐛 bug report

Under some conditions/configurations Parcel fails to package dependecies into bundles and fails in runtime.  
Here is a set of conditions and results I've discovered so far (see configs below for price command definitions)

- Things work for both manifest types in `watch` mode (https://gist.github.com/Stvad/e441997e880a47026772cd6872501d2a#file-background-mv2-watch-js)
- Manifest v3 without `--no-scope-hoist` produces working build (https://gist.github.com/Stvad/e441997e880a47026772cd6872501d2a#file-background-mv3-scope-hoist-js)
- Manifest v3 with `--no-scope-hoist` seems to produce a working backrgound script, but not content script
  - broken content script: https://gist.github.com/Stvad/e441997e880a47026772cd6872501d2a#file-content-mv3-no-
scope-hoist-js
  - background script: https://gist.github.com/Stvad/e441997e880a47026772cd6872501d2a#file-background-mv3-no-scope-hoist-js
- Manifest v2 is broken regardless of Scope Hoist (https://gist.github.com/Stvad/e441997e880a47026772cd6872501d2a#file-background-mv2-js )
- (probably unrelated) if I were to remove the `--no-content-hash` from the build commands - I start hitting #4145 

### `.parcelrc`
```js
{
  "extends": "@parcel/config-webextension",
  "validators": {
	"*.{ts,tsx}": [
	  "@parcel/validator-typescript"
	]
  }
}

```

### `package.json` parts
```json
"watch": "parcel watch source/mv2/manifest.json --dist-dir build/dev --no-cache",
"build:chrome": "parcel build source/mv3/manifest.json --no-content-hash --no-source-maps --dist-dir build/chrome --no-cache --detailed-report 0",
"build:firefox": "parcel build source/mv2/manifest.json --no-content-hash --no-source-maps --dist-dir build/firefox --no-cache --detailed-report 0",

```

## 🤔 Expected Behavior

When I run either of the build commands above - all the dependencies are bundled

## 😯 Current Behavior

Some (all?) dependencies are missing from the bundle and the extension fails at load time with the error like:
```
Uncaught Error: Cannot find module 'lbnQL'
    at parcelRequire (background.c6210b5c.js:15:15)
    at background.c6210b5c.js:19:57
    at background.c6210b5c.js:33:3
```

[...]

PR #8175, opened 2022-05-31
Issue8141/support safari extensions

### LINKED ISSUE #8141: Support safari webextensions in runtime-js
# 🙋 feature request

Runtime-js currently has specific support for chrome and firefox extensions. It has logic to generate a url relative to the bundle root.  For Safari this logic is missing, and parcel generates non-functional URLs. (could be considered a bug, depending on how you look at it)

## 🤔 Expected Behavior

When accessing assets in a safari extension, URLs should be formed correctly by getBundleUrl/getBaseUrl/getOrigin.

## 😯 Current Behavior

URLs generated by getBundleUrl/getBaseUrl/getOrigin do not get transformed and are not generally resolvable in a safari extension.

## 💁 Possible Solution

Adapt the regex in bundle-url.js to also support safari extensions.
I wrote a fix here: https://github.com/parcel-bundler/parcel/commit/6b8231771555acc8d12da49d29ed9b97d4f0c165

Kjoep:
# ↪️ Pull Request

Adapt parcel js runtime to also support safari extensions.
Fixes https://github.com/parcel-bundler/parcel/issues/8141

## 💻 Examples

Any webextension build with parcel that uses getBundleUrl/getBaseUrl/getOrigin should now function.

## 🚨 Test instructions

This can only be tested by building a webextension and running it on safari.  Note that to do this, you need to package the webextension as an OSX application using xcode.

In your extension, add an asset (let's say test.png).  Then add code as follows to your content script:

```
    import icon32 from '~/assets/test.png';
    const icon = document.createElement('img');
    icon.src = browser.runtime.getUrl;
    document.querySelector('body').appendChild(icon);
```

this should add the test image to the document body when your extension is applied (i.e. the import resolves to a `safari-web-extension://` url at runtime).


packages/configs/webextension/package.json
```diff
@@ -23,5 +23,10 @@
   },
   "peerDependencies": {
     "@parcel/core": "^2.5.0"
+  },
+  "@parcel/bundler-default": {
+    "minBundles": 10000000,
+    "minBundleSize": 3000,
+    "maxParallelRequests": 20
```
101arrowz: We should address this in a separate PR, for now this should probably be removed.

packages/runtimes/js/src/helpers/bundle-url.js
```diff
@@ -14,7 +14,7 @@ function getBundleURL() {
     throw new Error();
   } catch (err) {
     var matches = ('' + err.stack).match(
-      /(https?|file|ftp|(chrome|moz)-extension):\/\/[^)\n]+/g,
+      /(https?|file|ftp|(chrome|moz|safari-web)-extension):\/\/[^)\n]+/g,
```
101arrowz, line 17: To verify, the scheme for Safari extensions is `safari-web-extension`, not `safari-webextension`, correct?

101arrowz (commented):
Looks good for the most part, just a few questions.

packages/configs/webextension/package.json
Kjoep: Indeed.  That's a workaround for https://github.com/parcel-bundler/parcel/issues/8071 that ended up in this one.  Will remove.

packages/runtimes/js/src/helpers/bundle-url.js
Kjoep: Yes.  (we've already done a release based on this patched version, so it's tested in the wild)

101arrowz (approved):

mischnic (approved):

PR #8180, opened 2022-06-01
Experimental bundler integration

AGawrys:
# ↪️ Pull Request

This PR handles 
- Bundling Multiple Targets (i.e. not sharing between targets)
- Internalizing async dependencies
- merging bundles of the same type into one bundle (bundlegroups allowing)
- Adding edge types to the **BundleRoot** Graph to differentiate between parallel & async relationships

## ✔️ PR Todo

- [x] Update description and annotate changes on PR (or contributing doc)
- [ ] Test on large app - missing module in large app -> Fix to come
- [x] Test on app (react spectrum)
- [x] add or alter test cases


parcel-benchmark:
## Benchmark Results

#### Timings

| --- | --- | --- |
| Cold | 1.47s | -101.00ms 🚀 |
| Cached | 358.00ms | +24.00ms ⚠️ |

#### Cold Bundles

| dist/legacy/index.html | 826.00b | +0.00b | 538.00ms | -30.00ms 🚀 |
| dist/modern/index.html | 749.00b | +0.00b | 537.00ms | -31.00ms 🚀 |

#### Cached Bundles

| dist/legacy/index.2c76ad23.js | 1.66kb | +0.00b | 431.00ms | -32.00ms 🚀 |
| dist/legacy/index.8aaa89c9.js | 1.20kb | +0.00b | 431.00ms | -32.00ms 🚀 |
| dist/modern/index.6be20f01.js | 1.13kb | +0.00b | 431.00ms | -31.00ms 🚀 |

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 9.39s | +15.00ms |
| Cached | 461.00ms | -2.00ms |

#### Cold Bundles

*No bundle changes detected.*

#### Cached Bundles

*No bundle changes detected.*

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 1.63m | -3.37s |
| Cached | 2.65s | -53.00ms |

#### Cold Bundles

| dist/index.html | 240.00b | +0.00b | 1.23m | +32.94s ⚠️ |

#### Cached Bundles

*No bundle changes detected.*

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 6.91s | -40.00ms |
| Cached | 281.00ms | +4.00ms |

#### Cold Bundles

*No bundle changes detected.*

#### Cached Bundles

*No bundle changes detected.*

</p></details>

**[Click here to view a detailed benchmark overview.](https://benchmark.parceljs.org/benchmark/335827656141439572)**

contributorDocs/Bundler.md
```diff
@@ -0,0 +1,102 @@
...
+
+**AncestorAssets**: A mapping of BundleRoots to ALL assets reachable from it (This means all the assets reachable from it's ancestors (sync, async, parallel), PLUS any sync assets available to itself via bundlegroup or self, PLUS any parallel siblings that are guarenteed to load before it)
+
+# Other Definitions, as they relate to the bundler
+
+**Isolated** : An isolated Dependency, or Bundle must contain all assets it needs to load.
+
+**URL Specifier Type** : Cannot be merged or internalized.
+
+## Test Changes (Remove this section before PR)
```
AGawrys: Before Final PR 😅

packages/bundlers/experimental/src/ExperimentalBundler.js
```diff
@@ -528,32 +586,69 @@ function createIdealGraph(
...
       if (stack[stack.length - 1]?.[0] === node.value) {
         stack.pop();
       }
     },
   });
 
-  // Step 3: Determine reachability for every asset from each bundleRoot.
-  // This is later used to determine which bundles to place each asset in.
+  // Step Merge Type Change Bundles: Clean up type change bundles within the exact same bundlegroups
+  for (let [nodeIdA, a] of bundleGraph.nodes) {
```
devongovett, line 595: Would it be faster to only loop through `typeChangeIds` here rather than every node in the graph? Same with inner loop. I see you are skipping non-type change ids anyway? Should be able to get node values with `getNode`.

packages/bundlers/experimental/src/ExperimentalBundler.js
```diff
@@ -528,32 +586,69 @@ function createIdealGraph(
...
+      }
+    }
+  }
+
+  /**
+   *  Step Determine Reachability: Determine reachability for every asset from each bundleRoot.
+   * This is later used to determine which bundles to place each asset in. We build up two
+   * structures, one traversal each. ReachableRoots to store sync relationships,
+   * and bundleRootGraph to store the minimal availability through `parallel` and `async` relationships.
+   * The two graphs, are used to build up ancestorAssets, a strcuture which holds all availability by
```
devongovett:
```suggestion
   * The two graphs, are used to build up ancestorAssets, a structure which holds all availability by
```

packages/bundlers/experimental/src/ExperimentalBundler.js
```diff
@@ -696,88 +870,282 @@ function createIdealGraph(
...
     // bundle does not contain the asset in its ancestry
-    reachable = reachable.filter(b => !asyncAncestorAssets.get(b)?.has(asset));
+    reachable = reachable.filter(b => !ancestorAssets.get(b)?.has(asset));
+
+    // Finally, filter out bundleRoots (bundles) from this assets
+    // reachable if they are subgraphs, and reuse that subgraph bundle
+    // by drawing an edge. Essentially, if two bundles within an assets
+    // reachable array, have an ancestor-subgraph relationship, draw that edge.
+    // This allows for us to reuse a bundle instead of making a shared bundle if
+    // a bundle represents the exact set of assets a set fo bundles would share
```
devongovett:
```suggestion
    // a bundle represents the exact set of assets a set of bundles would share
```

packages/core/integration-tests/test/cache.js
```diff
@@ -4374,85 +4374,217 @@ describe('cache', function () {
...
             },
             async update(b) {
-              let html = await overlayFS.readFile(
-                b.bundleGraph.getBundles().find(b => b.name === 'b.html')
-                  ?.filePath,
-                'utf8',
-              );
-              assert.equal(html.match(/<script/g)?.length, 7);
-
+              assertBundles(b.bundleGraph, [
```
devongovett, line 4377: Why did these cache tests change?

packages/bundlers/experimental/src/ExperimentalBundler.js
```diff
@@ -429,72 +483,80 @@ function createIdealGraph(
...
+            )[0];
+            let referencingBundle = nullthrows(
+              bundleGraph.getNode(referencingBundleId),
+            );
+            invariant(referencingBundle !== 'root');
+            let bundle;
+            bundleId = bundles.get(childAsset.id);
+
+            /**
+             * If this is an entry bundlegroup, we only allow on bundle per type in those groups
```
devongovett:
```suggestion
             * If this is an entry bundlegroup, we only allow one bundle per type in those groups
```

packages/bundlers/experimental/src/ExperimentalBundler.js
```diff
@@ -696,88 +870,282 @@ function createIdealGraph(
...
+    let reachableNonEntries = [];
+
+    // Filter out entries, since they can't have shared bundles.
+    // Neither can non-splittable, isolated, or needing of stable name bundles.
+    // Reserve those filtered out bundles since we add the asset back into them.
+    for (let a of reachable) {
+      if (
+        entries.has(a) ||
+        !a.isBundleSplittable ||
+        getBundleFromBundleRoot(a).needsStableName ||
```
devongovett, line 882: Are we sure this is necessary and not already covered by entries? I'm not sure if we want to conflate naming with bundling behavior?

devongovett (commented):
Looks good overall. Couple questions.

Would be great to have a post with diagrams, a slide deck, or something a bit more visual to go through all the steps. The code comments are helpful, but I struggled without visual aids and examples of scenarios we covered. Probably a good idea to do before we forget, but certainly not a blocker for this PR.

packages/bundlers/experimental/src/ExperimentalBundler.js
AGawrys: Only reason we didn't do that is that, if two `typeChangeIds` may be merged, then one will be effectively deleted (from all structures and the graph) so I believe we encountered problems trying to process a node that didn't exist anymore. 

Alternatively, we could defer merging till after the loops, but I think then we'd run into issues where bundle `a` and `b` can be merged, and `b` and bundle `c` could be merged but not `a`, `b`, and `c`. Then we'd have to think of a way to resolve that. (perhaps a reason for duplication ? I'm not sure)

mischnic:
The flow comments in 15085a262e40f05f36553fecd817ceae18a40d38 should rather be solved with `nullthrows(...)` instead. And then the one remaining `$FlowFixMe[incompatible-call]` before `ALL_EDGE_TYPES` is also not necessary because there's no Flow error.

packages/bundlers/experimental/src/ExperimentalBundler.js
AGawrys: I think we took this logic directly from the default bundler (see this [link to the code](https://github.com/parcel-bundler/parcel/blob/v2/packages/bundlers/default/src/DefaultBundler.js#:~:text=let%20entries%20%3D%20b,b.needsStableName%20%26%26)) 

Yeah I'm kind of confused about this myself, I know non-entry bundles can be non-splittable and needStableName so that's why that is there, but I'm not sure if **only** bundlingBehavior can cover that. `NeedsStableName` is determined **by** bundlingBehavior but not equal.

 As for `isBundleSplittable` , I believe that comes from the Asset itself, so I'm not sure how thats determined exactly but I think it's set before bundling occurs

packages/core/integration-tests/test/cache.js
AGawrys: We changed the cache tests because they didn't seem to test the config properly, now the test for `adding a config` and `removing` determine if minBundleSize was upheld, for bundles with a large asset.
 
The previous test was testing script tags within a bundle, so it didn't seem like it was **just** testing  bundler config.

packages/bundlers/experimental/src/ExperimentalBundler.js
devongovett: Oh hmm I don't remember why we never changed this. Ok, well we can come back to it.

devongovett (approved):
lgtm!

Issue #8188, opened 2022-06-02
@parcel/transformers-webextension does not detect declarative_net_request. rule_resources[].path json files dependencies in manifest.json


carterworks:
# 🐛 bug report

The `declarativeNetRequest` extension API requires that the developer define the Rulesets in the `manifest.json`. [Each ruleset has a `path` property](https://developer.chrome.com/docs/extensions/reference/declarativeNetRequest/#type-Ruleset), which is "the path of the JSON ruleset relative to the extension directory." In other words, it's the path to a JSON file that needs to be copy over to the output directory when building the extension. Currently, this does not happen.

## 🎛 Configuration (.babelrc, package.json, cli command)

`package.json`

```json
{
  // …
  "scripts": {
    "build": "parcel build --config @parcel/config-webextension manifest.json"
  },
  "devDependencies": {
    "@parcel/config-webextension": "2.6.0",
    "@parcel/core": "2.6.0"
  }
}
```

`manifest.json`

```json
{
  "name": "parcel example",
  "description": "Parcel WebExt example, with declarativeNetRequest",
  "version": "1.0",
  "manifest_version": 3,
  "permissions": ["declarativeNetRequest"],
  "declarative_net_request": {
    "rule_resources": [
      {
        "id": "ruleset_1",
        "enabled": true,
        "path": "rules.json"
      }
    ]
  }
}
```

`rules.json`

```json
[]
```

## 🤔 Expected Behavior

The `rules.json` file should be copied to the `dist/` folder when building the extension.

## 😯 Current Behavior

`rules.json` is not copied to the `dist/` folder when building. This causes the extension to be invalid, when loaded unpacked into the browser.

## 💁 Possible Solution

Fix `packages/transformers/webextension/src/WebExtensionTransformer.js` to pick up those json dependencies.

Maybe something like this in `collectDependencies()`

```js
  if (program.declarative_net_request?.rule_resources) {
    const rrs = program.declarative_net_request?.rule_resources ?? [];
    for (const { path } of rrs) {
      asset.addURLDependency(path, {
        needsStableName: true,
        pipeline: "raw"
      });
    }
  }
```

It should have more error checking, but this works. I'd make the PR myself, but I can't get my local parcel build environment to work correctly.

## 🔦 Context

I'm trying to create a Chrome web extension that uses the `declarativeNetRequest` API.

## 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
[...]

PR #8189, opened 2022-06-03
Include declarativeNetRequest JSON file rules as dependencies when bundling web extensions

carterworks:
# ↪️ Pull Request

Fixes issue #8188.

> The `declarativeNetRequest` extension API requires that the developer define the Rulesets in the `manifest.json`. [Each ruleset has a `path` property](https://developer.chrome.com/docs/extensions/reference/declarativeNetRequest/#type-Ruleset), which is "the path of the JSON ruleset relative to the extension directory." In other words, it's the path to a JSON file that needs to be copy over to the output directory when building the extension. Currently, this does not happen.

I have included the `declarative_net_request` property in the `manifest.json` schema validator. I have also written a portion in `@parcel/transformer-webextension` that adds JSON files specified in the `path` property as dependencies of the asset, complete with error message if the specified JSON file does not exist.

This is my first contribution to Parcel, so please let me know what I need to change to align with correct API usage, design, and coding standards.

## 💻 Examples

See my [example repo](https://github.com/carterworks/parcel-dnr-dependency-example). Clone it, run `npm install`, and then `npm run build` to build the extension. If you check `dist/`, there is no `rules.json` file in it. If you load `dist/` as an unpacked extension into Chrome, it will throw an error about how `rules.json` is missing.

## 🚨 Test instructions

Install my branch into my example repository, then build the extension again. You will see that `rules.json` has been copied into the `dist/` folder.

## ✔️ PR Todo


carterworks:
@101arrowz I believe you are the Parcel maintainer that has worked on web extensions most recently. Will you look at this?

packages/transformers/webextension/src/schema.js
```diff
@@ -179,6 +179,25 @@ const commonProps = {
...
+          properties: {
+            id: string,
+            enabled: boolean,
+            path: string,
+          },
+          additionalProperties: false,
+          required: ['id', 'enabled', 'path'],
+        },
+      },
+      additionalProperties: false,
```
101arrowz: As far as I can tell the docs for this say that `rule_resources` is a required key, maybe specify that here?

packages/transformers/webextension/src/schema.js
```diff
@@ -179,6 +179,25 @@ const commonProps = {
...
+        type: 'array',
+        items: {
+          type: 'object',
+          properties: {
+            id: string,
+            enabled: boolean,
+            path: string,
+          },
+          additionalProperties: false,
+          required: ['id', 'enabled', 'path'],
```
101arrowz, line 195: Are all these fields really required? Specifically, is the `enabled` field not true by default? I haven't used `declarativeNetRequest` before so forgive my lack of knowledge.

packages/transformers/webextension/src/WebExtensionTransformer.js
```diff
@@ -216,6 +216,46 @@ async function collectDependencies(
...
+        );
+        throw new ThrowableDiagnostic({
+          diagnostic: [
+            {
+              message: 'Invalid Web Extension manifest',
+              origin: '@parcel/transformer-webextension',
+              codeFrames: [
+                {
+                  filePath,
+                  language: 'json',
```
101arrowz: `language` isn't necessary here, and for consistency with the other diagnostics thrown I would just remove it.

packages/transformers/webextension/src/WebExtensionTransformer.js
```diff
@@ -216,6 +216,46 @@ async function collectDependencies(
...
+                  codeHighlights: [
+                    {
+                      ...pathSourceLoc,
+                      message: `Path does not exist`,
```
101arrowz: We could make this include the ruleset ID, e.g. do this for `codeHighlights[0]`:

```js
{
  ...getJSONSourceLocation(
    ptrs[`/declarative_net_request/rule_resources/${i}/path`],
    'value',
  ),
  message: `Ruleset file for static ruleset ${resources.id} does not exist`
}
```

This is just to accomodate the fact that in some environments codeframes aren't fully shown, so the original message would be too vague.

101arrowz (commented):
Looks pretty good, just a few suggestions.

packages/transformers/webextension/src/WebExtensionTransformer.js
```diff
@@ -216,6 +216,46 @@ async function collectDependencies(
     }
     program.web_accessible_resources = war;
   }
+  if (program.declarative_net_request?.rule_resources) {
```
101arrowz: If I'm right that `rule_resources` is a required field you just need to check for `program.declarative_net_request` here.

101arrowz:
Also make sure to fix flow issues (probably just by typecasting).

packages/transformers/webextension/src/schema.js
carterworks: [The documentation for Rulesets](https://developer.chrome.com/docs/extensions/reference/declarativeNetRequest/#type-Ruleset) doesn't specify any of the properties as "optional", so I'm inclined to believe that they are required.

packages/transformers/webextension/src/WebExtensionTransformer.js
carterworks: Done. I also added the invalid filepath to the error message.

carterworks:
@101arrowz I've made those changes, including fixing the flow typing issues.

Also, I don't think the test failures (at least on the previous run) are due to anything I've change.
```
  1) hmr
       hmr runtime
         should handle CSS Modules update correctly:
      AssertionError [ERR_ASSERTION]: Expected "actual" to be strictly unequal to:
'http://127.0.0.1:49512/index.43be9ba8.css'
      + expected - actual
      at Context.<anonymous> (test/hmr.js:976:16)
  2) transpilation
       tests needing the real filesystem
         should compile node_modules when symlinked with a source field in package.json:
      AssertionError [ERR_ASSERTION]: The expression evaluated to a falsy value:
  path.join(__dirname, '/integration/babel-core-js/index.js')
      + expected - actual
      -false
      +true
      at Context.<anonymous> (test/transpilation.js:457:7)
```

101arrowz (approved):

mischnic:
Would be nice to include this in the integration test: https://github.com/parcel-bundler/parcel/tree/v2/packages/core/integration-tests/test/integration/webextension



> Also, I don't think the test failures (at least on the previous run) are due to anything I've change.

Yes, the tests are flakey... don't worry about them.

carterworks:
> Would be nice to include this in the integration test: https://github.com/parcel-bundler/parcel/tree/v2/packages/core/integration-tests/test/integration/webextension

@mischnic I can look into that today or tomorrow.

carterworks:
I've modified `packages/core/integration-tests/test/webextension.js`, but how can I run the test with the local version of `@parcel/transformer-webextension`? When I run it with `yarn test:integration --grep "webext*"`, it fails with the message "Invalid WebExtension Manifest".

mischnic:
That should be working (and there's no need to run a build step, Babel register does this on the fly).
So I'd say commit your test how you think it should work and then we can figure out if there's something wrong

(You need to modify `packages/core/integration-tests/test/webextension.js` and `packages/core/integration-tests/test/integration/webextension-mv3/manifest.json`)

carterworks:
A passing integration test has been added. I messed up the git history of this branch by rebasing off of `main`, but the substances is there. Just be sure to do a squash and merge 😅

packages/transformers/webextension/src/WebExtensionTransformer.js
```diff
@@ -216,6 +216,26 @@ async function collectDependencies(
...
+        filePath,
+        ...getJSONSourceLocation(
+          ptrs[`/declarative_net_request/rule_resources/${i}/path`],
+          'value',
+        ),
+      };
+
+      asset.addURLDependency(resources.path, {
+        needsStableName: true,
+        pipeline: 'isolated',
```
101arrowz: I think you probably want `pipeline: 'raw'` here; there is no such thing as `'isolated'`. JSON files get transformed to JS by default but you can prevent that with the `'raw'` pipeline (this only works in web extension builds).

carterworks: Whoops, I confused `pipeline` with `bundleBehavior`.

packages/transformers/webextension/src/WebExtensionTransformer.js
```diff
@@ -216,6 +216,26 @@ async function collectDependencies(
...
+      const loc = {
+        filePath,
+        ...getJSONSourceLocation(
+          ptrs[`/declarative_net_request/rule_resources/${i}/path`],
+          'value',
+        ),
+      };
+
+      asset.addURLDependency(resources.path, {
+        needsStableName: true,
```
101arrowz: Sorry for not adding this earlier, but why does this need a stable name? We typically only reserve that for files that actually cannot move locations according to the web extension documentation, i.e. `_locales/en/messages.json`.

carterworks: Unless the transformer can change the path in the manifest to the destination of the transformed json files, it needs to exist at the same path. The rules cannot be inlined.

101arrowz: The replacement is done in the packager automatically. This is the reason you see hashes in the filenames for other built files in the manifest like background scripts - their paths are rewritten in the transformer and fixed in the packager. This technique is useful for caching purposes and avoiding name collisions, so I recommend removing `needsStableName`.

carterworks: If I remove `needsStableName: true` from the transformer and add the following code to the integration test

```js
console.log({rule_resources: manifest.declarative_net_request.rule_resources});
for (const {path: resourcePath} of manifest.declarative_net_request?.rule_resources ?? []) {
  console.log(`Does "${path.join(distDir, resourcePath)}" exist? ${await outputFS.exists(path.join(distDir, resourcePath))}`);
  assert(await outputFS.exists(path.join(distDir, resourcePath)));
}
```

it fails with this output

```
{
  rule_resources: [
    {
      id: 'ruleset_1',
      enabled: true,
      path: './rulesets/ruleset_1.json'
    }
  ]
}
Does "/workspaces/parcel/packages/core/integration-tests/dist/rulesets/ruleset_1.json" exist? false
    1) should resolve a full webextension bundle
    ✓ should resolve the web_accessible_resources globs (2427ms)
    ✓ should support web extension manifest v3 (2294ms)


  2 passing (10s)
  1 failing

  1) webextension
       should resolve a full webextension bundle:

      AssertionError [ERR_ASSERTION]: false == true
      + expected - actual

      -false
      +true
      
      at Context.<anonymous> (test/webextension.js:59:7)
```

Is there is a way to modify the path in the manifest? Is that what `bundleBehavior: "isolated"` does? But I couldn't find out how to do that.

101arrowz: Ah right, it's because you aren't using the return value from `addURLDependency`. Just set `resources.path = asset.addURLDependency(...)` and keep `needsStableName` removed

carterworks: Done. Sorry for the confusion.

101arrowz:
Think this should be good to go. Thanks for working on this!

devongovett (approved):

PR #8194, opened 2022-06-07
Fix atomic createWriteStream

### LINKED ISSUE #7813: Incorrect js file extension with --no-scope-hoist
image

This is what I get by using the following build command:

`parcel build --dist-dir build --no-scope-hoist`

Notice the wrong js file name. In the compiled html I get:

`<script src="/index.ce910175.js" defer>`

So the js file name would go as expected at some point, but it won't get the correct file name when the build gets completed.

It's actually a TypeScript project with index.html + index.ts.

If I remove `--no-scope-hoist` the js file gets the correct extension. But  I don't want hoist, I need to keep the original identifier names intact.

parcel@2.3.2

mischnic:
Closes https://github.com/parcel-bundler/parcel/issues/7813

I think this is also the cause of at least some CI flakiness (for tests that use the NodeFS as the output FS).

## Problem

(copied from the linked issue)

> The problem is that when writing the bundles
> 
> https://github.com/parcel-bundler/parcel/blob/80c728ceffdc0829eca568a6ec66c07e07a0b87b/packages/core/core/src/requests/WriteBundleRequest.js#L253-L256
> 
> 
> this rename here
> https://github.com/parcel-bundler/parcel/blob/80c728ceffdc0829eca568a6ec66c07e07a0b87b/packages/utils/fs-write-stream-atomic/index.js#L103
> 
> 
> isn't awaited but happens sometime afterwards asynchronously. (The whole `handleClose` callback is called late, it's not just the missing `fs.rename` callback)
> But it doesn't happen at all if Parcel stops the process forcefully:
> 
> https://github.com/parcel-bundler/parcel/blob/80c728ceffdc0829eca568a6ec66c07e07a0b87b/packages/core/parcel/src/cli.js#L259

## Probable cause

> Node 13.14.0 is working correctly, the next version 14.0.0 has this behaviour. Something regarding the timing of the `close` event on the `createWriteStream()` stream has changed.
> 
> Potentially [nodejs/node@e13a37e49d](https://github.com/nodejs/node/commit/e13a37e49d)

## Solution

~So with my limited understanding of Node streams, I don't know if implementing an atomic write stream is even possible (anymore). So the easy (but brute force) solution is to let the callee do call the `move` at the right time (which worked out for all usages we have). E.g.~

It looks like the problem was that `fs-write-stream-atomic` added a wrapper `Writeable` "proxy" over the actual fs write stream. Overriding the `fs.close` used by Node's `createWriteStream` to additionally move the files afterwards works


parcel-benchmark:
## Benchmark Results

#### Timings

| --- | --- | --- |
| Cold | 1.37s | +1.37s ⚠️ |
| Cached | 310.00ms | +311.00ms ⚠️ |

#### Cold Bundles

#### Cached Bundles

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 8.54s | -67.00ms |
| Cached | 416.00ms | +19.00ms |

#### Cold Bundles

*No bundle changes detected.*

#### Cached Bundles

*No bundle changes detected.*

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 1.40m | -899.00ms |
| Cached | 2.24s | -35.00ms |

#### Cold Bundles

| dist/16.069344b7.js | 905.00b | +0.00b | 30.95s | -3.60s 🚀 |
| dist/simpleHasher.46d6f2e5.js | 742.00b | +0.00b | 34.29s | -28.64s 🚀 |
| dist/index.html | 240.00b | +0.00b | 27.27s | -2.83s 🚀 |

#### Cached Bundles

| dist/esm.c7dc1640.js | 61.96kb | +0.00b | 1.05m | +28.36s ⚠️ |
| dist/workerHasher.e50d242f.js | 1.72kb | +0.00b | 1.05m | +28.36s ⚠️ |
| dist/16.1969624f.js | 1.08kb | +0.00b | 34.29s | +2.94s ⚠️ |
| dist/16.069344b7.js | 905.00b | +0.00b | 34.29s | +2.94s ⚠️ |
| dist/simpleHasher.46d6f2e5.js | 742.00b | +0.00b | 1.05m | +28.36s ⚠️ |
| dist/index.html | 240.00b | +0.00b | 1.05m | +35.55s ⚠️ |

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 6.15s | -88.00ms |
| Cached | 256.00ms | -2.00ms |

#### Cold Bundles

*No bundle changes detected.*

#### Cached Bundles

*No bundle changes detected.*

</p></details>

**[Click here to view a detailed benchmark overview.](https://benchmark.parceljs.org/benchmark/334340555413652052)**

101arrowz:
To avoid introducing a `move` function, why not intercept the `'finish'` event (maybe via EventEmitter internals) and do what is currently in `move` before re-emitting `'finish'`?

mischnic:
This intercepting of the finish event was exactly what the previous broken implementation did.

I've now instead overridden the `fs.close` function used by `fs.createWriteStream` to move the file after closing the fd (and doing async operations is fine in this callback). So no `await move()` anymore.

packages/core/fs/src/NodeFS.js
```diff
@@ -58,7 +58,64 @@ export class NodeFS implements FileSystem {
...
+              throw e;
+            }
+          }
+        }
+      }
+    };
+
+    let writeStream = fs.createWriteStream(tmpFilePath, {
+      ...options,
+      fs: {
```
devongovett, line 96: Where does this `fs` option come from? I don't see it documented in Node or graceful-fs.

101arrowz: It's documented in Node.js since Node v12. See the `fs` option in the docs for [`fs.createWriteStream`'s options](https://nodejs.org/api/fs.html#fscreatewritestreampath-options).

devongovett (approved):

Issue #8038, opened 2022-05-02
CSS @import wrong order


sineway:
Let's say you have 2 files:

```css
/* red.css */
body { color: red }
```

```css
/* green.css */
body { color: green }
```
Then you import them:
```css
@import "red.css";
@import "green.css";
```

Then comment first import and save file:
```css
/* @import "red.css"; */
@import "green.css";
```
Then uncomment first import and save file again.
Order becomes wrong:
```css
/* green.css */
body { color: green }
/* red.css */
body { color: red }
```

| Software         | Version(s) |
| ---------------- | ---------- |
| Parcel           | 2.5.0
| Node             | 17.9.0
| npm         | 8.5.5
| Operating System | Linux

PR #8212, opened 2022-06-15
[Symbol Propagation] Non-deterministic bundle hashes

gorakong:
# ↪️ Pull Request
In collaboration with @lettertwo 
This PR addresses non-determinism in bundle hashes in the absence of file content changes.
In our test case, 
- `foo` imports `bag`
- `bar` imports both `baz` and `bag`

We end up with either one of two output bundle contents (and hence hashes) depending on whether `foo` or `bar` is transformed first (example below):

## 💻 Examples
When transformation of `foo.js` is delayed, `baz` is ordered before `bag` since we see `baz` being used first (in `bar.js`):
```
$parcel$export($aca58ea34dfb6a71$exports, "baz", () => $20d363e97eb3a92a$export$eab138534834282b);	
$parcel$export($aca58ea34dfb6a71$exports, "bag", () => $20d363e97eb3a92a$export$35e48f1ff4f2bc86);	
```
The order of these exports are flipped when we delay `bar.js`'s transformation.

## 💡Proposed Fix
Keep track of dependencies that have changes to their used symbols and sort their used symbols after symbol propagation.

## 🚨 Test instructions

The test case implements a custom `readFile` that proxies `overlayFS` to delay transformation. This was done (as opposed to deferring) to avoid changing source content of assets being transformed when introducing these delays.


gorakong:
@mischnic would like to know what you think 🙂

parcel-benchmark:
## Benchmark Results

#### Timings

| --- | --- | --- |
| Cold | 1.43s | +7.00ms |
| Cached | 306.00ms | -5.00ms |

#### Cold Bundles

*No bundle changes detected.*

#### Cached Bundles

*No bundle changes detected.*

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 8.63s | -105.00ms |
| Cached | 424.00ms | -7.00ms |

#### Cold Bundles

*No bundle changes detected.*

#### Cached Bundles

*No bundle changes detected.*

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 1.44m | -1.12s |
| Cached | 2.37s | -30.00ms |

#### Cold Bundles

| dist/16.069344b7.js | 905.00b | +0.00b | 35.22s | +2.98s ⚠️ |
| dist/simpleHasher.46d6f2e5.js | 742.00b | +0.00b | 1.07m | +28.37s ⚠️ |
| dist/index.html | 240.00b | +0.00b | 30.70s | +2.26s ⚠️ |

#### Cached Bundles

*No bundle changes detected.*

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 6.31s | +135.00ms |
| Cached | 272.00ms | +15.00ms ⚠️ |

#### Cold Bundles

*No bundle changes detected.*

#### Cached Bundles

*No bundle changes detected.*

</p></details>

**[Click here to view a detailed benchmark overview.](https://benchmark.parceljs.org/benchmark/338268198793118292)**

lettertwo:
The dependency graph for test case currently looks like this:

```mermaid
graph TB;
  index[index.js]
  library[library/index.js]
  foo[library/foo.js]
  bar[library/bar.js]
  utils[utils/index.js]
  bag[utils/bag.js]
  empty[utils/empty.js]

  index-->|"import {foo, bar} from 'library'"|library
  library-->|"export {foo} from './foo'"|foo
  library-->|"export {bar} from './bar'"|bar
  foo-->|"import {bag} from '../utils'"|utils
  bar-->|"import {baz, bag} from '../utils'"|utils
  utils-->|"export * from './bag'"|bag
  utils-->|"export * from './empty'"|empty

```

We aren't sure if this is the _minimal_ form or not. Maybe the diamond shape isn't actually necessary?

mischnic:
I do think the diamond (plus the "open" diamond at the end) is needed. But that linear chain before the diamond (html, `index.js`) should be unnecessary in theory. The order of outgoing dependencies is fixed for uncached builds, but the order of incoming dependencies depends on the order of transformation.

(And you could also run into this situation if you do a build with only the "right" diamond branch, and then do another cached build with the "left" diamond branch added. The result would also have a different order of incoming dependencies compared to adding them in the reverse order).

So this is just an inherent problem of `getIncomingDependencies` not having a consistent order across builds. This function is also iterated over in
- the experimental bundler to call `internalizeAsyncDependency` (which might be fine if the order is irrelevant here)
- the CSS packager to find the `@media(...) {}` types that a CSS  asset should be wrapped with. This could definitely cause differing bundle outputs
- (all other usages just do `find` , `some` or `every`

The only solution I can think of is sorting the dependencies somehow. There is simply no existing/inherent order for incoming dependencies.

packages/core/core/src/requests/AssetGraphRequest.js
```diff
@@ -593,6 +599,12 @@ export class AssetGraphBuilder {
           }
         }
       }
+      // Sort usedSymbolsUp so they are a consistent order across builds.
+      // This ensures a consistent ordering of these symbols when packaging.
+      // See https://github.com/parcel-bundler/parcel/pull/8212
+      for (let dep of changedDeps) {
+        dep.usedSymbolsUp = new Set([...dep.usedSymbolsUp].sort());
+      }
```
mischnic: Does this intentionally happen inside the `propagateSymbolsUp(...)` callback? This would be called for every node.

lettertwo: ooh, yeah, this should be moved up a level!

packages/core/core/src/requests/AssetGraphRequest.js
```diff
@@ -396,6 +400,7 @@ export class AssetGraphBuilder {
           depUsedSymbolsDown.clear();
         }
         if (!equalSet(depUsedSymbolsDownOld, depUsedSymbolsDown)) {
+          changedDeps.add(dep);
```
lettertwo: Not sure that we actually need to track changes to `usedSymbolsDown`, because outgoing edge orders are probably deterministic, and we are also only sorting the `usedSymbolsUp`.

packages/core/core/src/requests/AssetGraphRequest.js
gorakong: Moved this out of the callback and verified that there were no significant regressions in build times on a large app 🙂

packages/core/core/src/requests/AssetGraphRequest.js
mischnic: Right, I think this case isn't needed (because of the second reason)

> because outgoing edge orders are probably deterministic

(No, they aren't: https://github.com/parcel-bundler/parcel/issues/8038. But AFAICT this could be fixed by using `asset.dependencies` as opposed to `getNodesConnectedFrom(asset)`)

gorakong: In that case, I'll remove that line

mischnic:
It has to be that new workerfarm which is causing CI to fail. Using the existing one should work better:
```diff
diff --git packages/core/integration-tests/test/scope-hoisting.js packages/core/integration-tests/test/scope-hoisting.js
index 3375ac891..7509826fb 100644
--- packages/core/integration-tests/test/scope-hoisting.js
+++ packages/core/integration-tests/test/scope-hoisting.js
@@ -2,7 +2,6 @@ import assert from 'assert';
 import path from 'path';
 import nullthrows from 'nullthrows';
 import {normalizePath} from '@parcel/utils';
-import {createWorkerFarm} from '@parcel/core';
 import {md} from '@parcel/diagnostic';
 import {
   assertBundles,
@@ -18,6 +17,7 @@ import {
   overlayFS,
   run,
   runBundle,
+  workerFarm,
 } from '@parcel/test-utils';

 const bundle = (name, opts = {}) => {
@@ -5535,10 +5535,6 @@ describe('scope hoisting', function () {
   });

   it('produce the same bundle hash regardless of transformation order', async function () {
-    let workerFarm = createWorkerFarm({
-      maxConcurrentWorkers: 0,
-    });
-
     let testDir = path.join(
       __dirname,
       'integration/scope-hoisting/es6/non-deterministic-bundle-hashes',
```
But that makes the test pass sometimes when commenting out the fix...

mischnic:
Another thing I noticed: `await workerFarm.end();` is missing, e.g. like here https://github.com/parcel-bundler/parcel/blob/a9193ce0b19e93415f264b3295ceee093ddeb8c7/packages/core/integration-tests/test/cache.js#L3932-L3938

So having two workerfarms still running and receiving events from each other probably triggers the assertion

gorakong:
@mischnic doesn't the workerFarm in `test-utils` use the same `createWorkerFarm` (from `@parcel/core`) under the hood? The test case loses its non-determinism without `maxConcurrentWorkers` set to 0 as it affects the proxied readFile, so I added a `workerFarmZeroConcurrent` export in `test-utils` and used that instead. Test is still failing in CI though..

packages/core/integration-tests/test/scope-hoisting.js
```diff
@@ -5532,4 +5533,82 @@ describe('scope hoisting', function () {
...
+      workerFarm: workerFarmZeroConcurrent,
+    });
+
+    let bundleHashDelayBar = b2
+      .getBundles()
+      .find(b => b.filePath.endsWith('.js') && b.filePath.includes('index'))
+      .filePath.split('.')[1];
+
+    assert.strictEqual(bundleHashDelayFoo, bundleHashDelayBar);
+    await workerFarmZeroConcurrent.end();
```
mischnic: Having a global `workerFarmZeroConcurrent` and then ending it in one test is weird, so it would be better to keep it as a local variable here. And also there should be a `try {..} finally { end(); }` around it in case the test fails.

---
The only other explanation I have for the CI error is that the `inputFS` memoryfs instance (and therefore also `overlayFS`) use a different worker farm than the one used in the `bundle()`:
https://github.com/parcel-bundler/parcel/blob/a9193ce0b19e93415f264b3295ceee093ddeb8c7/packages/core/test-utils/src/utils.js#L39-L40
So maybe try creating a new temporary `outputFS` and `overlayFS` but using the nonconcurrent farm.

gorakong: thanks! is this what you mean? 
https://github.com/parcel-bundler/parcel/pull/8212/files#diff-ee074b5a5bbe809758dc3be2de9a23752857e0022abe2ccb350a3cbe28ad4018R5585-R5590

mischnic (approved):

devongovett (approved):

Issue #6830, opened 2021-08-30
in v2  "type": "module" breaks pug support.


cekvenich2:
if you set package json to type: module, the Pug plugin can no longer read pug.config.js to set pretty: true and breaks pug support.

PR #8253, opened 2022-06-25
allow cjs config files on type module projects

danieldevlewis:
# ↪️ Pull Request

Allow plugins with `.js` config files to also work with a `.cjs` extension.

If you have a project with `"type": "module"` then currently it is impossible to have a javascript config file for a plugin.  Either you get a error telling you `require` isn't allowed using modules, or if you use modules you get an error telling you modules can't be used with `require`.  If you use a `cjs` file you get a JSON parse error.

This PR allows `loadConfig` to load a `cjs` file as `js` file instead of trying to load it as a `json` file.

It also updates all plugins that accept a js config file to also accept a cjs config file.

Also see previously reported #6830

## 💻 Examples

For example if you use the pug transformer, you can now have a config file called:

`pug.config.cjs`.

Previously it would only allow `pug.config.js`

## 🚨 Test instructions

Create a project with `"type": "module"` in the package.json.

Attempt to use a javascript config file with, for example, a pug template.

```
// .pugrc.js
module.exports = {
  filters: {
    'custom-filter': function (text, options) {
      return 'FILTERED: ' + text;
    }
  }
}
```

```
// index.pug
:custom-filter()
    Hello!
```

If the `.pugrc.js` is renamed `.pugrc.cjs` this will now work without an error.

## ✔️ PR Todo


packages/core/utils/src/config.js
```diff
@@ -71,7 +71,7 @@ export async function loadConfig(
 
     try {
       let extname = path.extname(configFile).slice(1);
-      if (extname === 'js') {
+      if (['js', 'cjs'].includes(extname)) {
```
devongovett: Could be slightly faster, avoids array allocation.

```suggestion
      if (extname === 'js' || extname === 'cjs') {
```

devongovett (commented):
Makes sense to me

packages/core/utils/src/config.js
danieldevlewis: Oh yes, good point

devongovett (approved):

optimistiks:
https://github.com/parcel-bundler/parcel/blob/v2/packages/transformers/postcss/src/loadConfig.js#L98

I believe it should check as well for `.cjs` otherwise it treats everything that is not `.js` as JSON which is incorrect if you have `.cjs` config file for PostCSS.

optimistiks:
https://github.com/parcel-bundler/parcel/blob/v2/packages/transformers/postcss/src/loadConfig.js#L98

It seems it should check for `.cjs` here as well

Issue #6111, opened 2021-04-11
Respect NODE_PATH environment variable


TLATER:
# 🐛 bug report

When trying to build with no `node_modules` directory, but all my dependencies in a directory listed in `NODE_PATH`, parcel will fail to run because it cannot resolve its plugins.

I believe this will also extend to bundling modules from the path, since the module resolution is custom and I can't find any references to `NODE_PATH` in the relevant source files.

## 🎛 Configuration (.babelrc, package.json, cli command)

This can be reproduced with no configuration at all! See the code sample section.

## 🤔 Expected Behavior

Parcel should resolve modules correctly according to `NODE_PATH`, and in the test case output an empty file (I think, I'll admit I'm too lazy to check what parcel outputs for an empty html file right now).

## 😯 Current Behavior

Parcel complains that even its built-in plugins are missing.

The exact error is:

```
@parcel/core: Cannot find extended parcel config
/home/tlater/Documents/Projects/tlaternet-templates/.parcelrc:2:15
  1 | {
> 2 |   "extends": ["@parcel/config-default"],
>   |               ^^^^^^^^^^^^^^^^^^^^^^^^ Cannot find module "@parcel/config-default"
  3 |   "transformers": {
  4 |     "*.mp3": [ "@parcel/transformer-raw" ]
```

## 💁 Possible Solution

Fix the module resolution to respect `NODE_PATH`.

## 🔦 Context

I'm trying to deploy a website on a NixOS host. I'm using [node2nix](https://github.com/svanderburg/node2nix) to achieve this, which uses the `NODE_PATH` variable to allow the package manager to take control of the `node_modules`.

## 💻 Code Sample

Easiest reproduction scenario:

1. `mkdir test && pushd test`
2. `npm install parcel`
    - Make sure you don't have a directory with a `node_modules` or `package*.json` in it as any parent directory up to the root.
3. `export NODE_PATH="$(pwd)/node_modules"`
4. `popd`
5. `touch empty.html`
6. `"$NODE_PATH/.bin/parcel" empty.html`

## 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Parcel           | 2.0.0-beta.2 |
| Node             | v12.21.0 |
| npm/Yarn         | npm 6.14.11 |
| Operating System | NixOS 20.09.20210401.da7f4c4 (Nightingale) |

mischnic:
`NODE_PATH` would have to be taken into account in these files:
[...]

PR #8273, opened 2022-07-02
Respect NODE_PATH environment variable

josephshen:
Respect NODE_PATH environment variable

## 💻 Examples
1. `find .`
 .
./src
./src/App.tsx
./src/index.html
./3rd
./3rd/web
./3rd/web/vender

2. `pushd ./3rd/web/vender`
   install modules `npm install parcel react react-dom`

3. `export NODE_PATH="$(pwd)/3rd/web/vender/node_modules"`

    run parcel like this ` node 3rd/web/vender/node_modules/parcel/lib/cli.js src/index.html`
    - before this patch, parcel will complain "Failed to resolve XXXX"
    - after this patch, parcel will work fine with respect NODE_PATH setting for module resolving.

fix this https://github.com/parcel-bundler/parcel/issues/6111#issue-855203386

### DIFF
--- a/packages/core/package-manager/src/NodeResolverBase.js
+++ b/packages/core/package-manager/src/NodeResolverBase.js
@@ -139,6 +139,19 @@ export class NodeResolverBase<T> {
     let dir = path.dirname(sourceFile);
     let moduleDir = this.fs.findNodeModule(moduleName, dir);
 
+    if (!moduleDir && process.env.NODE_PATH != null) {
+       for (const dir of process.env.NODE_PATH.split(path.delimiter)) {
+          try {
+            let nodePathDir = path.join(dir, moduleName);
+            let stats = this.fs.statSync(nodePathDir);
+            if (stats.isDirectory()) {
+              moduleDir = nodePathDir;
+              break;
+            };
+          } catch (err) {}
+       }
+    }
+
     ctx.invalidateOnFileCreate.push({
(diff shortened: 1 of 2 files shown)


devongovett:
Not sure this is a good idea. One problem is that an absolute path in `NODE_PATH` will make the cache non-portable, since it'll be different for each person that runs the build.

mischnic:
Another data point: https://github.com/webpack/webpack/issues/4479#issuecomment-286701441

> NODE_PATH should not be used. Even the node.js documentation discourage usage: [nodejs.org/api/modules.html#modules_loading_from_the_global_folders](https://nodejs.org/api/modules.html#modules_loading_from_the_global_folders)
> 
> This is omitted by inten[tion].

toastal:
> NODE_PATH is still supported, but is less necessary now

This isn’t doesn’t sound like it’s discouraging rather than pointing out a many use cases won’t need it.

A use case that _could_ benefit from it is Nix/Guix derivations where you can’t do network requests for `npm install` so often other tools can provide the packages for you on the `$NODE_PATH`. In these environments, you wouldn’t be using Parcel cache anyhow.

I just tried to use `npmlock2nix` to get my modules for `$NODE_PATH` then I couldn’t run `parcel --no-cache …` because it was claiming my plugins weren’t installed since I’m to using npm’s mechanism for resolution.

PR #8303, opened 2022-07-12
Fix missing module in large app from Experimental Bundler

AGawrys:
This PR addresses a missing module where only shared bundles were considered when attempting to remove a shared bundle from a group.

When determining what source bundles need the shared-bundle-to-remove's assets pushed back into it. Instead of consulting the total BundleGroup bundles, we consulted only shared bundles. This structure would then always be empty, and so shared bundles were removed from groups but not placed into source bundles at all. 

## Major Changes
- We've removed `sharedToSourceBundles` (A structure that held shared bundle ids to source bundle ids)
- We now handle the case where if a shared bundle is removed from a Bundle Group, and only used elsewhere once, it should just be fully removed
- We now consider reused bundles in parallel request limit removal
- Refactor reused bundle code (we attempt to reuse bundles before creating a shared bundle if a bundle is the **exact** subgraph of some other set of bundles) 

### Reused Bundles

Reused bundles are bundles that are shared (i.e. they have source bundles) but also have a mainEntryAsset. These bundles have a referencing bundle (i.e. something that imports it) but are also exact subgraphs of other bundles. So, they can be shared, but they should  also never be removed by parallel request limits since they have a referencing bundles (something that dynamically imported it, causing a new bundle to be created)  (See future todo for plans to remove these as an optimization)

### Diagrams
In this bundleGraph,

`Bar.js` requires `a.js` and `b.js` (and requires `Foo.js` through `a.js`)
`Foo.js` requires `a.js` and `b.js`
`a.js` requires `foo.js` 

Thus, `foo.js` is a subgraph of `bar.js` and creating a shared bundle would be wrong since it would be an exact copy of the `foo` bundle and leave an empty shell bundle behind. 

IdealBundleGraph_Final

## ✔️ PR Todo

- [x] Test on large app

## Future TODO 

- [ ] Remove `sourceBundle` property from bundles in favor of reading graph
- [ ] Remove a reused bundle if it has been removed by parallel request limit (currently this is not done because reused bundle have some referencing or parent bundle as it's main reason for being a bundle. However, that bundle may be small and better suited to simply be merged back in)
- [x] [FIXED] Potential error: What if a reused bundle has a shared bundle with some other bundle. Then, if that bundle group hits the parallel request limit, and is merged back in, only the reused bundles' assets would be merged in and the shared bundle is now **not connected**

## Shared Bundle under Reused Bundle Case Diagram

GRAPH


parcel-benchmark:
## Benchmark Results

#### Timings

| --- | --- | --- |
| Cold | 1.50s | +6.00ms |
| Cached | 344.00ms | +7.00ms |

#### Cold Bundles

*No bundle changes detected.*

#### Cached Bundles

*No bundle changes detected.*

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 9.14s | +126.00ms |
| Cached | 444.00ms | -11.00ms |

#### Cold Bundles

*No bundle changes detected.*

#### Cached Bundles

*No bundle changes detected.*

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 1.61m | -4.00ms |
| Cached | 2.50s | -45.00ms |

#### Cold Bundles

| dist/index.html | 240.00b | +0.00b | 1.21m | +33.32s ⚠️ |

#### Cached Bundles

| dist/esm.c7dc1640.js | 61.96kb | +0.00b | 1.20m | +31.79s ⚠️ |
| dist/workerHasher.e50d242f.js | 1.72kb | +0.00b | 1.20m | +31.79s ⚠️ |
| dist/16.1969624f.js | 1.08kb | +0.00b | 38.95s | +2.40s ⚠️ |
| dist/index.html | 240.00b | +0.00b | 1.20m | +36.87s ⚠️ |

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 6.67s | +82.00ms |
| Cached | 272.00ms | -1.00ms |

#### Cold Bundles

*No bundle changes detected.*

#### Cached Bundles

*No bundle changes detected.*

</p></details>

**[Click here to view a detailed benchmark overview.](https://benchmark.parceljs.org/benchmark/338364050929877586)**

packages/bundlers/experimental/src/ExperimentalBundler.js
```diff
@@ -895,39 +892,57 @@ function createIdealGraph(
...
+        canReuse.add(candidateSourceBundleRoot);
+        bundleGraph.addEdge(candidateSourceBundleId, reuseableBundleId);
+
+        let reusableBundle = bundleGraph.getNode(reuseableBundleId);
+        invariant(reusableBundle !== 'root' && reusableBundle != null);
+        reusableBundle.sourceBundles.add(candidateSourceBundleId);
+      } else {
+        // Asset is not a bundleRoot, but if its parent bundle can be
+        // reused as a subgraph of another bundle in its reachable, reuse it
+        for (let otherReuseCandidate of reachable) {
```
devongovett, line 920: So the algorithm here went from O(n) to O(n^2)? Did we see any real-world perf regressions?

packages/bundlers/experimental/src/ExperimentalBundler.js
```diff
@@ -895,39 +892,57 @@ function createIdealGraph(
...
+          if (candidateSourceBundleRoot === otherReuseCandidate) continue;
+          let reusableCandidateReachable = getReachableBundleRoots(
+            otherReuseCandidate,
+            reachableRoots,
+          ).filter(b => !ancestorAssets.get(b)?.has(otherReuseCandidate));
+          if (
+            reusableCandidateReachable.indexOf(candidateSourceBundleRoot) > -1
```
devongovett: You could use array.includes here

packages/bundlers/experimental/src/ExperimentalBundler.js
```diff
@@ -895,39 +892,57 @@ function createIdealGraph(
...
+      let reuseableBundleId = bundles.get(asset.id);
+      if (reuseableBundleId != null) {
+        canReuse.add(candidateSourceBundleRoot);
+        bundleGraph.addEdge(candidateSourceBundleId, reuseableBundleId);
+
+        let reusableBundle = bundleGraph.getNode(reuseableBundleId);
+        invariant(reusableBundle !== 'root' && reusableBundle != null);
+        reusableBundle.sourceBundles.add(candidateSourceBundleId);
+      } else {
+        // Asset is not a bundleRoot, but if its parent bundle can be
```
devongovett: Only the direct parent? Is there any case where we'd have to do this recursively?

devongovett (approved):
A few questions, but lets get this in so we can test

packages/bundlers/experimental/src/ExperimentalBundler.js
devongovett: nevermind, this was there before

packages/bundlers/experimental/src/ExperimentalBundler.js
AGawrys: So actually I think the comment isn't the best. Essentially what we're trying to do here is: 

Of the reachable assets for an asset, is there some common subset (under a bundleRoot) that may be reused. 
So I guess it's not so much as a parent, as it is any bundleRoot that synchronously imports the asset in question. (which to me is kind of a parent? Just not the **only** parent)

So I guess the better way to phrase would be "if **a bundleRoot of a bundle that this asset is in,** can be reused by some other bundleRoot in the asset's reachable, reuse it"

AGawrys: And it's important to note, this code path you're talking about was only added because we can't guarantee that we process a bundleRoot before any non-bundleRoot assets that are in it, when placing assets into bundles. 
So like if we process a non-bundleRoot asset, we need to make sure we don't prematurely place it in, where it could've been reused by some other bundleRoot

AGawrys: Diagrams for this section might be important 😄 ..working on it

PR #8333, opened 2022-07-23
Fix naming collisions between named exports and wildcard exports

### LINKED ISSUE #8331: @parcel/transformer-typescript-types generates incorrect .d.ts file when there is a naming conflict with a wildcard export
# 🐛 bug report

When bundling a library with parcel and relying on `@parcel/transformer-typescript-types` to generate a bundled `.d.ts` file, you'll get incorrect output if there is a naming conflict between a top-level export and the contents of a module that's exported with a wildcard export (e.g. `export * from "foo`).

Interestingly, the bug will appear or not depending on the _order_ of the top-level exports (see below)

***package.json***
```json
{
    "name": "test-parcel-library",
    "main": "dist/index.js",
    "types": "dist/types.d.ts",
    "scripts": {
        "build": "parcel build src/index.ts"
    }
    ...
}
```
***src/index.ts***
```typescript
export const nameConflict = { messageFromIndex: "this instance of nameConflict is from index.ts" };
export * from "./name-conflict"; // Note: this comes _after_ the top-level export above.
```

***src/name-conflict.ts***
```typescript
export const nameConflict = { messageFromNameConflict: "this instance of nameConflict is from name-conflict.ts" };
```

## 😯 Current Behavior

When you run the `parcel build`, parcel generates the following `index.d.ts` file:

***types.d.ts***
```typescript
export const nameConflict: {
    messageFromNameConflict: string;
};
declare const _nameConflict1: {
    messageFromIndex: string;
};
```
However, when you consume the library in a different project, the `nameConflict` export at runtime will be the top-level export from `index.ts`:

```typescript
import { nameConflict } from "test-parcel-library";
// @ts-expect-error: the d.ts file generated by parcel is incorrect
console.log(nameConflict.messageFromIndex); // prints "this instance of nameConflict is from index.ts"
```

## 🤔 Expected Behavior

The `index.d.ts` file generated by parcel should be:

```typescript
declare const _nameConflict1: {
    messageFromNameConflict: string;
};
export const nameConflict: {
    messageFromIndex: string;
};
```
Interestingly, if you switch the order of the exports in `index.ts`, you'll get this correct output:

***index.ts***
```typescript
export * from "./name-conflict"; // Note: this comes _before_ the top-level export below, which fixes the problem.
export const nameConflict = { messageFromIndex: "this instance of nameConflict is from index.ts" };
```

If you build the same project with `tsc` directly, the `.d.ts` files generated will correctly describe the package, regardless of the order of the exports (although  it looks slightly different from what you'd expect from parcel because `tsc` can't bundle the `.d.ts` files for commonjs projects).

## 💁 Possible Solution

In `TSModuleGraph.propegate` we first collect all the top-level export names in an array with a call to `getAllExports`, and iterate through them:

https://github.com/parcel-bundler/parcel/blob/62f64cae33a7bc1d2f5117472f9fd8bc2a931ab0/packages/transformers/typescript-types/src/TSModuleGraph.js#L202-L208

In the example above (where the bug reproduces), this array will have these (simplified) contents, in this order:
```
[ 
   { name: "nameConflict", module: <TSModule for 'index.ts"> }
   { name: "nameConflict", module: <TSModule for 'nameConflict.ts"> }
]
```
The call to `exportedNames.set(e.name, e.module)` will thus happen twice with the same name, over-writing the previous value. At the end of the loop, `exportedNames` will be:
```
{ "nameConflict":  <TSModule for 'nameConflict.ts"> }
```
Further down, when we assign unique names, we do this check:

https://github.com/parcel-bundler/parcel/blob/62f64cae33a7bc1d2f5117472f9fd8bc2a931ab0/packages/transformers/typescript-types/src/TSModuleGraph.js#L213-L217

`exportedNames.get(name) === m` will be `true` when we're passing through `nameConflict.ts`, so we skip renaming it. When we pass through `index.ts`, it will be false, so we'll proceed to assign a disambiguator to it.

If you flip the order of the exports in `index.ts` that will change the contents of `exportedNames`, and you'll get the opposite (correct) behavior - i.e. the export from `nameConflict.ts` gets the disambiguator, and the top-level export from `index.ts` does not.

One solution would be to check when we are about to over-write entries to `exportedNames` and somehow find a way to deterministically prefer top-level exports over implicit exports that happen through a wildcard.

## 🔦 Context

I was working on a PR add support for namespace exports to `@parcel/transformer-typescript-types` (fixing #5911), and I ran across this issue in the process.

## 💻 Code Sample

You can see a reproduction of the problem in [this github repo](https://github.com/astegmaier/parcel-naming-conflict-bug)

## 🌍 Your Environment

| Parcel           |  2.6.2
| Node             | 16.15.1
| Yarn         | 1.22.19
| Operating System | Windows 11 Version 10.0.22000 Build 22000

astegmaier:
# ↪️ Pull Request

This fixes #8331.

## 💻 Examples

Previously, when a name contained in a module exported through a wildcard export conflicted with another named export, we would sometimes generate incorrect `.d.ts` bundles that included the wildcard export name instead of the named export. For example:

***src/index.ts***
```typescript
export const nameConflict = { messageFromIndex: "this instance of nameConflict is from index.ts" };
export * from "./other"; // Note: this comes _after_ the top-level export above.
```

***src/other.ts***
```typescript
export const nameConflict = { messageFromOther: "this instance of nameConflict is from othert.ts" };
```
...would produce...

***types.d.ts***
```typescript
export const nameConflict: {
    messageFromOther: string;
};
declare const _nameConflict1: {
    messageFromIndex: string;
};
```
...instead of...

```typescript
export const nameConflict: {
    messageFromIndex: string;
};
```
With this change, we do the right thing, no matter the order of the exports.

I tested both parcel and tsc runtime behavior and in such a situation, the higher-level named export always "wins" - i.e. the `nameConflict` in `other.ts` is completely inaccessible, so it can be shaken.

I also verified that we will still do the right thing in cases where the lower-level name export is referenced by some other export. For example, if you add this file to the above program

***src/name-conflict.ts***
```typescript
import { nameConflict } from "./other1";
export const consumer: typeof nameConflict = { messageFromOther1: "This variable uses the type of the nameConflict1 variable defined in other1.ts.." };
```
...and export it at the top level...

***src/index.ts***
```typescript
export const nameConflict = { messageFromIndex: "this instance of nameConflict is from index.ts" };
export * from "./other";
export { consumer } from "./consumer"; // <--this was added
```
...then your `types.d.ts` bundle will be...

```typescript
declare const _nameConflict1: {
    messageFromOther1: string;
};
export const consumer: typeof _nameConflict1;
export const nameConflict: {
    messageFromIndex: string;
};
```
...which is correct.

Both of these cases are reflected in the unit test I added.

One other case that I considered, but where we already do the right thing, is where two wildcard exports point to modules that have a name conflict _within_ them. For example, if your program is...

***index.ts***
```typescript
export * from "./other1"
export * from "./other2"
```
***other1.ts***
```typescript
export const nameConflict = { messageFromOther1: "this instance of nameConflict is from other1.ts" };
```
***other2.ts***
```typescript
export const nameConflict = { messageFromOther2: "this instance of nameConflict is from other2.ts" };
```
... we will bubble up the following warning from `tsc` at compile-time:
```
    1 | export * from "./other1.ts";
  > 2 | export * from "./other2.ts";
  >   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Module "./other1" has already exported a member named 'nameConflict'. Consider explicitly re-exporting to resolve the ambiguity.
    3 |
```

## 🚨 Test instructions

The unit test should be sufficient, but in case you want to play around with other use-cases end-to-end (and test how the resulting `d.ts` files would be consumed by other packages), you can use [this test repo](https://github.com/astegmaier/parcel-naming-conflict-bug) combined with `yarn link`.

## ✔️ PR Todo


devongovett:
This is an interesting case where Parcel's behavior doesn't match TSC's I think. When compiling JS, the last one wins. But looks like TSC prioritizes named exports over export *? [Example](https://parcel2-repl-mischnic.vercel.app/#JTdCJTIyZmlsZXMlMjIlM0ElNUIlNUIlMjIlMkZzcmMlMkZpbmRleC5qcyUyMiUyQyU3QiUyMnZhbHVlJTIyJTNBJTIyaW1wb3J0JTIwJTdCeCU3RCUyMGZyb20lMjAnLiUyRmZpbGUxJyUzQiU1Q25jb25zb2xlLmxvZyh4KSUyMiUyQyUyMmlzRW50cnklMjIlM0F0cnVlJTdEJTVEJTJDJTVCJTIyJTJGc3JjJTJGb3RoZXIuanMlMjIlMkMlN0IlMjJ2YWx1ZSUyMiUzQSUyMmNsYXNzJTIwVGhpbmclMjAlN0IlNUNuJTIwJTIwcnVuKCklMjAlN0IlNUNuJTIwJTIwJTIwJTIwY29uc29sZS5sb2coJTVDJTIyVGVzdCU1QyUyMiklM0IlNUNuJTIwJTIwJTdEJTIwJTVDbiU3RCU1Q24lNUNuY29uc3QlMjB4JTIwJTNEJTIwMTIzJTNCJTVDbmV4cG9ydCUyMCU3QlRoaW5nJTJDJTIweCU3RCUzQiUyMiU3RCU1RCUyQyU1QiUyMiUyRnNyYyUyRmZpbGUxLmpzJTIyJTJDJTdCJTIydmFsdWUlMjIlM0ElMjJleHBvcnQlMjAqJTIwZnJvbSUyMCcuJTJGb3RoZXInJTNCJTVDbmV4cG9ydCUyMGNvbnN0JTIweCUyMCUzRCUyMDQ1NiUzQiUyMiU3RCU1RCU1RCUyQyUyMm9wdGlvbnMlMjIlM0ElN0IlMjJlbnRyaWVzJTIyJTNBJTVCJTVEJTJDJTIybWluaWZ5JTIyJTNBZmFsc2UlMkMlMjJzY29wZUhvaXN0JTIyJTNBdHJ1ZSUyQyUyMnNvdXJjZU1hcHMlMjIlM0FmYWxzZSUyQyUyMnB1YmxpY1VybCUyMiUzQSUyMiUyRl9fcmVwbF9kaXN0JTIyJTJDJTIydGFyZ2V0VHlwZSUyMiUzQSUyMmJyb3dzZXJzJTIyJTJDJTIydGFyZ2V0RW52JTIyJTNBbnVsbCUyQyUyMm91dHB1dEZvcm1hdCUyMiUzQW51bGwlMkMlMjJobXIlMjIlM0FmYWxzZSUyQyUyMm1vZGUlMjIlM0ElMjJwcm9kdWN0aW9uJTIyJTJDJTIycmVuZGVyR3JhcGhzJTIyJTNBZmFsc2UlMkMlMjJ2aWV3U291cmNlbWFwcyUyMiUzQWZhbHNlJTJDJTIyZGVwZW5kZW5jaWVzJTIyJTNBJTVCJTVEJTJDJTIybnVtV29ya2VycyUyMiUzQTAlN0QlMkMlMjJ1c2VUYWJzJTIyJTNBZmFsc2UlMkMlMjJicm93c2VyQ29sbGFwc2VkJTIyJTNBJTVCJTVEJTJDJTIydmlld3MlMjIlM0ElNUIlMjIlMkZzcmMlMkZpbmRleC5qcyUyMiUyQyUyMiUyRnNyYyUyRm90aGVyLmpzJTIyJTJDJTIyJTJGc3JjJTJGZmlsZTEuanMlMjIlNUQlMkMlMjJjdXJyZW50VmlldyUyMiUzQTIlN0Q=)

Might be bad if the compiled types don't match the code?

cc. @mischnic

astegmaier:
> Might be bad if the compiled types don't match the code?

Totally agree. But I think in this case, the current type definitions from parcel do not match the runtime behavior, and this PR changes it to align (see below).

> When compiling JS, the last one wins.

This is not what I'm seeing in my tests. The runtime behavior of both Parcel and Tsc seems to be "named exports always win over wildcard exports". Details are below - let me know if you think I might be overlooking something.

I have a test monorepo with two libraries that have the same source code, but one is built with tsc, and the other with parcel. (you can [run it yourself here](https://github.com/astegmaier/parcel-naming-conflict-bug)):

***index.ts***
```ts
// Case 1: the top-level export is _after_ the wildcard export.

export * from "./other1";
export const nameConflict1 = { messageFromIndex: "this instance of nameConflict1 is from index.ts" };

// Case 2: the top-level export is _before_ the wildcard export <-- this will give you incorrect d.ts in parcel

export const nameConflict2 = { messageFromIndex: "this instance of nameConflict2 is from index.ts" };
export * from "./other2";
```

***other1.ts***
```ts
export const nameConflict1 = { messageFromOther2: "this instance of nameConflict1 is from other1.ts" };
```

***other2.ts***
```ts
export const nameConflict2 = { messageFromOther2: "this instance of nameConflict2 is from other2.ts" };
```

I consume and test the libraries like this (as written, this test passes).

```ts
import * as Parcel from "test-library-parcel";
import * as Tsc from "test-library-tsc";

describe("Wildcard export names conflicting with named exports", () => {
    it("Tsc should prefer the named export when it is _after_ the wildcard export.", () => {
        expect(Tsc.nameConflict1.messageFromIndex).toBe("this instance of nameConflict1 is from index.ts");
    });
    it("Parcel should prefer the named export when it is _after_ the wildcard export.", () => {
        expect(Parcel.nameConflict1.messageFromIndex).toBe("this instance of nameConflict1 is from index.ts");
    });
    it("Tsc should prefer the named export when it is _before_ the wildcard export.", () => {
        expect(Tsc.nameConflict2.messageFromIndex).toBe("this instance of nameConflict2 is from index.ts");
    })

    // This is the bug:
    it("Parcel should prefer the named export when it is _before_ the wildcard export, but the type definitions don't align with this behavior.", () => {
        // @ts-expect-error: Property 'messageFromIndex' does not exist on type '{ messageFromOther2: string; }
        expect(Parcel.nameConflict2.messageFromIndex).toBe("this instance of nameConflict2 is from index.ts");
        // The d.ts file generated by parcel says "messageFromOther2" is a "string", but it is undefined at runtime.
        expect(Parcel.nameConflict2.messageFromOther2).toBe(undefined);
    });
})
```

When I debug the runtime bundle generated by parcel, I see the calls to `$parcel$exportWildcard()` appear at the bottom of the bundle -- after the named exports are defined. When those functions executed, they no-op because they discover that the destination object already has a property called `nameConflictX`:

```js
function $parcel$exportWildcard(dest, source) {
  Object.keys(source).forEach(function(key) {
    if (key === 'default' || key === '__esModule' || dest.hasOwnProperty(key)) {
     // We return early here in this case because dest.hasOwnProperty("nameConflictX") is true.
      return;
    }
    // Rest of the function
  });
```

I was assuming this is correct because it matches `tsc`. (Although perhaps both are wrong? Not sure where in the spec defines this behavior...)

astegmaier:
@devongovett - I noticed that I overlooked [the link to the example in the REPL that you posted](https://parcel2-repl-mischnic.vercel.app/#JTdCJTIyZmlsZXMlMjIlM0ElNUIlNUIlMjIlMkZzcmMlMkZpbmRleC5qcyUyMiUyQyU3QiUyMnZhbHVlJTIyJTNBJTIyaW1wb3J0JTIwJTdCeCU3RCUyMGZyb20lMjAnLiUyRmZpbGUxJyUzQiU1Q25jb25zb2xlLmxvZyh4KSUyMiUyQyUyMmlzRW50cnklMjIlM0F0cnVlJTdEJTVEJTJDJTVCJTIyJTJGc3JjJTJGb3RoZXIuanMlMjIlMkMlN0IlMjJ2YWx1ZSUyMiUzQSUyMmNsYXNzJTIwVGhpbmclMjAlN0IlNUNuJTIwJTIwcnVuKCklMjAlN0IlNUNuJTIwJTIwJTIwJTIwY29uc29sZS5sb2coJTVDJTIyVGVzdCU1QyUyMiklM0IlNUNuJTIwJTIwJTdEJTIwJTVDbiU3RCU1Q24lNUNuY29uc3QlMjB4JTIwJTNEJTIwMTIzJTNCJTVDbmV4cG9ydCUyMCU3QlRoaW5nJTJDJTIweCU3RCUzQiUyMiU3RCU1RCUyQyU1QiUyMiUyRnNyYyUyRmZpbGUxLmpzJTIyJTJDJTdCJTIydmFsdWUlMjIlM0ElMjJleHBvcnQlMjAqJTIwZnJvbSUyMCcuJTJGb3RoZXInJTNCJTVDbmV4cG9ydCUyMGNvbnN0JTIweCUyMCUzRCUyMDQ1NiUzQiUyMiU3RCU1RCU1RCUyQyUyMm9wdGlvbnMlMjIlM0ElN0IlMjJlbnRyaWVzJTIyJTNBJTVCJTVEJTJDJTIybWluaWZ5JTIyJTNBZmFsc2UlMkMlMjJzY29wZUhvaXN0JTIyJTNBdHJ1ZSUyQyUyMnNvdXJjZU1hcHMlMjIlM0FmYWxzZSUyQyUyMnB1YmxpY1VybCUyMiUzQSUyMiUyRl9fcmVwbF9kaXN0JTIyJTJDJTIydGFyZ2V0VHlwZSUyMiUzQSUyMmJyb3dzZXJzJTIyJTJDJTIydGFyZ2V0RW52JTIyJTNBbnVsbCUyQyUyMm91dHB1dEZvcm1hdCUyMiUzQW51bGwlMkMlMjJobXIlMjIlM0FmYWxzZSUyQyUyMm1vZGUlMjIlM0ElMjJwcm9kdWN0aW9uJTIyJTJDJTIycmVuZGVyR3JhcGhzJTIyJTNBZmFsc2UlMkMlMjJ2aWV3U291cmNlbWFwcyUyMiUzQWZhbHNlJTJDJTIyZGVwZW5kZW5jaWVzJTIyJTNBJTVCJTVEJTJDJTIybnVtV29ya2VycyUyMiUzQTAlN0QlMkMlMjJ1c2VUYWJzJTIyJTNBZmFsc2UlMkMlMjJicm93c2VyQ29sbGFwc2VkJTIyJTNBJTVCJTVEJTJDJTIydmlld3MlMjIlM0ElNUIlMjIlMkZzcmMlMkZpbmRleC5qcyUyMiUyQyUyMiUyRnNyYyUyRm90aGVyLmpzJTIyJTJDJTIyJTJGc3JjJTJGZmlsZTEuanMlMjIlNUQlMkMlMjJjdXJyZW50VmlldyUyMiUzQTIlN0Q=) . 

Unfortunately, when I click on it, it seems broken - here's what I see:

broken_repl

So I put the same code in a [test repo](https://github.com/astegmaier/parcel-devon-example) and ran it, and here's what I see:

results

The value `456` comes from the named export in `file1.js`, not the wildcard-exported file (`other.js`)

Changing the order of the exports in `file1.js` has no effect - it will still print `456`. The rule seems to be "named exports win over wildcard exports", rather than "last export wins". This is consistent with the test results I ran above.

mischnic:
Huh, I'll have to try it with Edge as some point, works on Chrome macOS:
Bildschirmfoto 2022-08-05 um 17 12 13

I haven't had time to look at this in detail yet, but:
> Although perhaps both are wrong? Not sure where in the spec defines this behavior...

A good rule of thumb is: assume v8 implements the spec correctly, which also shows that named exports have a higher precedence than export stars

astegmaier:
> A good rule of thumb is: assume v8 implements the spec correctly, which also shows that named exports have a higher precedence than export stars

Makes sense. So it seems that at least the goals of this PR - to make sure we respect that rule in the generated `.d.ts` are correct. Is there anything I can do to improve the implementation?

ivanbanov:
@devongovett is it expected to have this PR merged any time soon? :)

PR #8387, opened 2022-08-10
Verify version when resolving Node builtin polyfills

mischnic:
This will call `packageManager.resolve` for every resolution of a Node builtin with the correct range that should be used to catch cases where another instance of some Node builtin polyfill was pulled in by some other dependency, be it in the same package, or in a monorepo situation

If it's the wrong version, it will either autoinstall, or error (using the existing autoinstall logic):
```
@parcel/core: Failed to resolve 'util' from './a/index.js'
  /Users/nmischkulnig/Desktop/monorepo-builtin-versioning/a/index.js:1:20
  > 1 | import * as x from "util";
  >   |                    ^^^^^^
    2 |
    3 | console.log(x);


@parcel/package-manager: Could not find module "util" satisfying ^0.12.3.

  /Users/nmischkulnig/Desktop/monorepo-builtin-versioning/a/package.json:6:3
    5 |   "dependencies": {
  > 6 |     "util": "^0.4.9"
  >   |     ^^^^^^ Found this conflicting local requirement.
    7 |   }
    8 | }

```

Questions:
- [x] ~Previously, the polyfill was resolved from the actual source file but autoinstalled into the project root. We either need to install it into the local package (what I've done now), or resolve it relative from the monorepo root. (So that resolution base = autoinstall target)~ Now, all builtin polyfills are resolved from the project root
- [ ] Is it a concern that `packageManager.resolve` is slow? I suspect there won't be that many Node builtins anyway

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,6 +54,7 @@
     "mocha-junit-reporter": "^2.0.0",
     "mocha-multi-reporters": "^1.5.1",
     "prettier": "2.4.1",
+    "punycode": "^1.4.1",
     "rimraf": "^3.0.2",
     "semver": "^5.7.1",
     "sinon": "^7.3.1"
--- a/packages/utils/node-resolver-core/src/builtins.js
+++ b/packages/utils/node-resolver-core/src/builtins.js
@@ -4,35 +4,36 @@ import {builtinModules} from 'module';
 
 export const empty: string = require.resolve('./_empty.js');
 
-// $FlowFixMe
-let builtins: {[string]: any, ...} = Object.create(null);
+let builtins: {[string]: {|name: string, range: ?string|}, ...} =
+  // $FlowFixMe
+  Object.create(null);
 // use definite (current) list of Node builtins
 for (let key of builtinModules) {
-  builtins[key] = empty;
+  builtins[key] = {name: empty, range: null};
 }
 
-builtins.assert = 'assert/';
-builtins.buffer = 'buffer/';
-builtins.console = 'console-browserify';
-builtins.constants = 'constants-browserify';
-builtins.crypto = 'crypto-browserify';
-builtins.domain = 'domain-browser';
-builtins.events = 'events/';
-builtins.http = 'stream-http';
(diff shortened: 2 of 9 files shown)


packages/utils/node-resolver-core/src/builtins.js
```diff
@@ -4,35 +4,39 @@ import {builtinModules} from 'module';
...
+builtins.querystring = {name: 'querystring-es3/', range: '^0.2.1'};
+builtins.stream = {name: 'stream-browserify', range: '^3.0.0'};
+builtins.string_decoder = {name: 'string_decoder/', range: '^1.3.0'};
+builtins.sys = {name: 'util/util.js', range: '^0.12.3'};
+builtins.timers = {name: 'timers-browserify', range: '^2.0.11'};
+builtins.tty = {name: 'tty-browserify', range: '^0.0.1'};
+builtins.url = {name: 'url/', range: '^0.11.0'};
+builtins.util = {name: 'util/util.js', range: '^0.12.3'};
+builtins.vm = {name: 'vm-browserify', range: '^1.1.2'};
+builtins.zlib = {name: 'browserify-zlib', range: '^0.2.0'};
```
devongovett: Should we load these dependencies from the `devDependencies` in package.json so they are always in sync?

package.json
```diff
@@ -54,6 +54,7 @@
     "mocha-junit-reporter": "^2.0.0",
     "mocha-multi-reporters": "^1.5.1",
     "prettier": "2.4.1",
+    "punycode": "^1.4.1",
```
mischnic, line 57: The `url` polyfill package declares punycode as a dependency

https://github.com/defunctzombie/node-url/blob/5480ec001102457ac00c3c7878facdec39b536c8/package.json#L6-L9

but then does `require("punycode")` which uses the node polyfill (as opposed to `require("punycode/")`). I needed to declare this version to get the integration tests to pass....

devongovett (approved):

PR #8389, opened 2022-08-11
New context type for non-browser, non-node environments

mischnic:
```
{
  "targets": {
    "default": {
      "context": "edge"
    }
  }
}
```

For something like https://edge-runtime.vercel.app/ or https://developers.cloudflare.com/workers/learning/how-workers-works/

This enables building with the usual polyfills for the browser (Node builtins, replacing globals like `Buffer`, inlining readFileSync) while also not using the `browser` field for resolution (which is a problem for e.g. `react-dom/server`) and also by default having `includeNodeModules: true`

- [ ] Better ideas for naming? Alternatively `bare`?
- [ ] Look at places where env is used, e.g. JSTransformer doesn't apply any target for non-electron,non-node,non-browser contexts
- [ ] Any prior discussions regarding adding V8/JSC itself to browserslist? (Because there's no "browser" name that maps 1:1 and it's neither `engines.browser` nor `engines.node`

### DIFF
--- a/packages/core/core/src/Environment.js
+++ b/packages/core/core/src/Environment.js
@@ -75,6 +75,7 @@ export function createEnvironment({
       case 'browser':
       case 'web-worker':
       case 'service-worker':
+      case 'edge':
       default:
         includeNodeModules = true;
         break;
@@ -87,6 +88,9 @@ export function createEnvironment({
       case 'electron-main':
       case 'electron-renderer':
         outputFormat = 'commonjs';
+        break;
+      case 'edge':
+        outputFormat = 'esmodule';
         break;
       default:
         outputFormat = 'global';
--- a/packages/packagers/js/src/ScopeHoistingPackager.js
+++ b/packages/packagers/js/src/ScopeHoistingPackager.js
@@ -53,6 +53,7 @@ const GLOBALS_BY_CONTEXT = {
     ...Object.keys(globals.node),
     ...Object.keys(globals.browser),
   ]),
+  edge: new Set([...BUILTINS, ...Object.keys(globals.worker)]),
 };
 
 const OUTPUT_FORMATS = {
--- a/packages/core/core/src/TargetDescriptor.schema.js
+++ b/packages/core/core/src/TargetDescriptor.schema.js
@@ -36,6 +36,7 @@ export const PACKAGE_DESCRIPTOR_SCHEMA: SchemaObject = {
         'electron-main',
         'electron-renderer',
         'service-worker',
+        'edge',
       ],
     },
     includeNodeModules: {
--- a/packages/core/integration-tests/test/integration/edge/index.js
+++ b/packages/core/integration-tests/test/integration/edge/index.js
@@ -1,8 +1,8 @@
 import library from "./library";
 import process from "process";
 
-module.exports = {
+output({
   global: Buffer.from("abc").toString("hex"),
   builtin: process.cwd(),
   browserResolution: library
-}
+});
--- a/packages/core/integration-tests/test/integration/edge/index.js
+++ b/packages/core/integration-tests/test/integration/edge/index.js
@@ -1,0 +1,8 @@
+import library from "./library";
+import process from "process";
+
+module.exports = {
+  global: Buffer.from("abc").toString("hex"),
+  builtin: process.cwd(),
+  browserResolution: library
+}
--- a/packages/core/integration-tests/test/integration/edge/library/browser.js
+++ b/packages/core/integration-tests/test/integration/edge/library/browser.js
@@ -1,0 +1,1 @@
+export default "browser";
--- a/packages/core/integration-tests/test/integration/edge/library/main.js
+++ b/packages/core/integration-tests/test/integration/edge/library/main.js
@@ -1,0 +1,1 @@
+export default "main";
--- a/packages/core/integration-tests/test/integration/edge/library/package.json
+++ b/packages/core/integration-tests/test/integration/edge/library/package.json
@@ -1,0 +1,4 @@
+{
+  "main": "main.js",
+  "browser": "browser.js"
+}
--- a/packages/core/integration-tests/test/integration/edge/package.json
+++ b/packages/core/integration-tests/test/integration/edge/package.json
@@ -1,0 +1,7 @@
+{
+  "targets": {
+    "default": {
+      "context": "edge"
+    }
+  }
+}
--- a/packages/core/integration-tests/test/javascript.js
+++ b/packages/core/integration-tests/test/javascript.js
@@ -795,6 +795,18 @@ describe('javascript', function () {
     ]);
   });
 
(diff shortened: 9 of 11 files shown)


packages/core/core/src/Environment.js
```diff
@@ -88,6 +89,9 @@ export function createEnvironment({
       case 'electron-renderer':
         outputFormat = 'commonjs';
         break;
+      case 'edge':
+        outputFormat = 'esmodule';
+        break;
```
mischnic, line 94: At least Cloudflare Workers and Vercel Edge Runtime only support ESM and no CommonJS

parcel-benchmark:
## Benchmark Results

#### Timings

| --- | --- | --- |
| Cold | 1.77s | -0.00ms |
| Cached | 397.00ms | -15.00ms |

#### Cold Bundles

| dist/legacy/index.2c76ad23.js | 1.66kb | +0.00b | 490.00ms | -34.00ms 🚀 |
| dist/legacy/index.8aaa89c9.js | 1.20kb | +0.00b | 490.00ms | -34.00ms 🚀 |
| dist/modern/index.6be20f01.js | 1.13kb | +0.00b | 489.00ms | -35.00ms 🚀 |

#### Cached Bundles

*No bundle changes detected.*

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 11.18s | +126.00ms |
| Cached | 547.00ms | +5.00ms |

#### Cold Bundles

*No bundle changes detected.*

#### Cached Bundles

*No bundle changes detected.*

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 1.96m | -594.00ms |
| Cached | 3.06s | +3.00ms |

#### Cold Bundles

*No bundle changes detected.*

#### Cached Bundles

*No bundle changes detected.*

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 8.13s | -53.00ms |
| Cached | 321.00ms | -20.00ms 🚀 |

#### Cold Bundles

*No bundle changes detected.*

#### Cached Bundles

*No bundle changes detected.*

</p></details>

**[Click here to view a detailed benchmark overview.](https://benchmark.parceljs.org/benchmark/339784230456263252)**

packages/core/core/src/Environment.js
kiddkai: Hey, is that possible to have an option to generate a single commonjs bundle as well?

packages/packagers/js/src/ScopeHoistingPackager.js
```diff
@@ -53,6 +53,7 @@ const GLOBALS_BY_CONTEXT = {
     ...Object.keys(globals.node),
     ...Object.keys(globals.browser),
   ]),
+  edge: new Set([...BUILTINS, ...Object.keys(globals.worker)]),
```
kiddkai, line 56: I think using `globals.workers` may not be correct because it has: https://github.com/sindresorhus/globals/blob/main/globals.json#L1129-L1135
And `self: true` that edge workers may not have implemented.

packages/core/core/src/Environment.js
mischnic: This is just the default, you can always do
```
{
  "targets": {
    "default": {
      "context": "edge",
      "outputFormat": "commonjs"
    }
  }
}
```
(and as I've said, as far as I can tell most of these runtimes only support ESM, which is why I made that the default.).

By "single" you mean making dynamic imports behave like a synchronous require instead? I think that's out of scope for this PR (and also there have been people in the past who wanted this feature for Node as well)

packages/packagers/js/src/ScopeHoistingPackager.js
mischnic: The only effect this has is that Parcel doesn't not create some `var onclose = ...` variable (if it decides to try that for some reason), so that if there's a global of that name, it will rename it to `var _onclose = ...` etc...

https://github.com/parcel-bundler/parcel/blob/9e2d5d0d60d08d65b5ae6cd765c907a8753bbf39/packages/packagers/js/src/ScopeHoistingPackager.js#L351-L355

In that sense, I thought that workers is close enough here (and a conservative choice).

devongovett: Do we need to account for other globals in https://edge-runtime.vercel.app/features/available-apis, https://proposal-common-min-api.deno.dev/#index, or other runtimes somehow?

kiddkai: https://developers.cloudflare.com/workers/runtime-apis/web-standards/#:~:text=Cloudflare%20Workers%20uses%20the%20V8,with%20no%20need%20for%20transpilers.

@devongovett I think the tricky thing to support them in every custom runtime(wrappers on top of barebone js engine) is they include different subsets of web APIs into their runtime based on the usage. 

For more generic runtimes like `deno`, they may provide a larger amount of default globals.

But for some specialised runtimes with very thin layers on top of v8(https://github.com/denoland/rusty_v8) or isolated-vm(https://github.com/laverdet/isolated-vm). The build target of these kinds of runtimes will have nothing. All the compatibility layers will be hand-written by the runtime creators.

I think the purpose of this `edge` target is for environments that only contain: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects assuming the runtime is a pure js engine(v8/jscore/...etc).

mischnic: > Do we need to account for other globals in https://edge-runtime.vercel.app/features/available-apis, https://proposal-common-min-api.deno.dev/#index, or other runtimes somehow?

AFAICT, these are all subsets of the browser globals, which is why I chose the worker globals.


> I think the tricky thing to support them in every custom runtime(wrappers on top of barebone js engine) is they include different subsets of web APIs into their runtime based on the usage.

As I've said, the worst thing that can happen here is a variable being called `var _ReadableStreamBYOBReader` as opposed to `var ReadableStreamBYOBReader` even if `ReadableStreamBYOBReader` is not available in the runtime. This is **not** about polyfilling the `URLSearchParams` global for runtimes not supporting it.

And as far I can tell, all callers of `getTopLevelName` do `this.getTopLevelName("$"+foo+"$"+bar);` anyway, so I'm not even sure why we have these globals in the first place... @devongovett ?

devongovett: hmm I think maybe vestiges of an older implementation where we had exports that weren't prefixed.

mischnic:
> building with the usual polyfills for the browser (Node builtins, replacing globals like Buffer, inlining readFileSync) while also not using the browser field for resolution

Well, there's a problem with these goals:
If you want to polyfill node builtins, then `browser` resolution has to be used for at least these packages in any case: https://unpkg.com/browse/process@0.11.10/package.json
The question is whether it would be enough to still use the `browser` field for node polyfills, or if that would still leave some packages broken...

PR #8393, opened 2022-08-11
Codesplitting across reexports using symbol data

### LINKED ISSUE #7392: Parcel does not split side effect free re-exports across bundles
[Reproduction](https://parcel2-repl-mischnic.vercel.app/#JTdCJTIyZmlsZXMlMjIlM0ElNUIlNUIlMjIlMkZzcmMlMkZhc3luYy5qcyUyMiUyQyU3QiUyMnZhbHVlJTIyJTNBJTIyaW1wb3J0JTIwJTdCYmFyJTdEJTIwZnJvbSUyMCcuLiUyRmxpYiclM0IlNUNuY29uc29sZS5sb2coYmFyKSUzQiUyMiU3RCU1RCUyQyU1QiUyMiUyRnNyYyUyRmluZGV4LmpzJTIyJTJDJTdCJTIydmFsdWUlMjIlM0ElMjJpbXBvcnQlMjAlN0Jmb28lN0QlMjBmcm9tJTIwJy4uJTJGbGliJyUzQiU1Q24lNUNuY29uc29sZS5sb2coZm9vKSUzQiU1Q25pbXBvcnQoJy4lMkZhc3luYycpJTNCJTVDbiUyMiUyQyUyMmlzRW50cnklMjIlM0F0cnVlJTdEJTVEJTJDJTVCJTIyJTJGbGliJTJGcGFja2FnZS5qc29uJTIyJTJDJTdCJTIydmFsdWUlMjIlM0ElMjIlN0IlNUNuJTIwJTIwJTVDJTIybmFtZSU1QyUyMiUzQSUyMCU1QyUyMmxpYiU1QyUyMiUyQyU1Q24lMjAlMjAlNUMlMjJzaWRlRWZmZWN0cyU1QyUyMiUzQSUyMGZhbHNlJTVDbiU3RCUyMiUyQyUyMmlzRW50cnklMjIlM0FmYWxzZSU3RCU1RCUyQyU1QiUyMiUyRmxpYiUyRmluZGV4LmpzJTIyJTJDJTdCJTIydmFsdWUlMjIlM0ElMjJleHBvcnQlMjAlN0Jmb28lN0QlMjBmcm9tJTIwJy4lMkZmb28nJTNCJTVDbmV4cG9ydCUyMCU3QmJhciU3RCUyMGZyb20lMjAnLiUyRmJhciclM0IlMjIlN0QlNUQlMkMlNUIlMjIlMkZsaWIlMkZmb28uanMlMjIlMkMlN0IlMjJ2YWx1ZSUyMiUzQSUyMmV4cG9ydCUyMGNvbnN0JTIwZm9vJTIwJTNEJTIwMiUzQiUyMiU3RCU1RCUyQyU1QiUyMiUyRmxpYiUyRmJhci5qcyUyMiUyQyU3QiUyMnZhbHVlJTIyJTNBJTIyZXhwb3J0JTIwY29uc3QlMjBiYXIlMjAlM0QlMjAzJTNCJTIyJTdEJTVEJTVEJTJDJTIyb3B0aW9ucyUyMiUzQSU3QiUyMmVudHJpZXMlMjIlM0ElNUIlNUQlMkMlMjJtaW5pZnklMjIlM0FmYWxzZSUyQyUyMnNjb3BlSG9pc3QlMjIlM0F0cnVlJTJDJTIyc291cmNlTWFwcyUyMiUzQWZhbHNlJTJDJTIycHVibGljVXJsJTIyJTNBJTIyJTJGX19yZXBsX2Rpc3QlMjIlMkMlMjJ0YXJnZXRUeXBlJTIyJTNBJTIyYnJvd3NlcnMlMjIlMkMlMjJ0YXJnZXRFbnYlMjIlM0ElMjJsYXN0JTIwMSUyMENocm9tZSUyMHZlcnNpb24lMjIlMkMlMjJvdXRwdXRGb3JtYXQlMjIlM0ElMjJlc21vZHVsZSUyMiUyQyUyMmhtciUyMiUzQWZhbHNlJTJDJTIybW9kZSUyMiUzQSUyMnByb2R1Y3Rpb24lMjIlMkMlMjJyZW5kZXJHcmFwaHMlMjIlM0FmYWxzZSUyQyUyMnZpZXdTb3VyY2VtYXBzJTIyJTNBZmFsc2UlMkMlMjJkZXBlbmRlbmNpZXMlMjIlM0ElNUIlNUQlMkMlMjJudW1Xb3JrZXJzJTIyJTNBMCU3RCUyQyUyMnVzZVRhYnMlMjIlM0FmYWxzZSUyQyUyMmJyb3dzZXJDb2xsYXBzZWQlMjIlM0ElNUIlNUQlMkMlMjJ2aWV3cyUyMiUzQSU1QiUyMiUyRmxpYiUyRnBhY2thZ2UuanNvbiUyMiUyQyUyMiUyRnNyYyUyRmFzeW5jLmpzJTIyJTJDJTIyJTJGbGliJTJGYmFyLmpzJTIyJTJDJTIyJTJGbGliJTJGaW5kZXguanMlMjIlMkMlMjIlMkZsaWIlMkZmb28uanMlMjIlMkMlMjIlMkZzcmMlMkZpbmRleC5qcyUyMiUyQyUyMk9wdGlvbnMlMjIlNUQlMkMlMjJjdXJyZW50VmlldyUyMiUzQTYlN0Q=)

Many large libraries have many exports in an index file which are re-exports of other modules, for example a library of React components. If an application utilizes code splitting, and imports different components from different bundles, Parcel currently includes all used components in the entry bundle rather than only the components necessary. This is because the index file is marked as used, and with it comes all of its dependencies. We could be smarter about this and only include dependencies that have used symbols _in the current bundle_.

mischnic:
Closes https://github.com/parcel-bundler/parcel/issues/7392

Symbol propagation already does the traversals anyway, so now it just always tracks for each symbol in a dependency, what that symbol should resolve to.

Then, once it encounters a non-reexport dependency, it checks if all symbols in that dependency resolve to the same asset, and rewrites the edge from the dependency to the asset to point to the original asset that has the symbol (or at least as far down as possible, in case some deopt or non-static situation was encountered).

Since it's possible to rename symbols via reexports, I've had to add an additional map to the dependency node, e.g. a dependency to a side-effect-free asset that does `export {foo as bar} from "y";` would then be rewritten to point directly to `y` instead, and the `symbolTarget` map then has `foo -> bar` which is applied transparently in `bundleGraph.getUsedSymbols` and `bundleGraph.getSymbols` (which I think is a better solution compared to mutating the actual DependencyValue symbols entry).

- [x] The major outstanding known issue: if we rewrite the dependency edge proper, we don't have the original edge anymore on the next build (if it's cached). So I think we have to have another edge type, which would be used instead of the untyped edge when converting the asset graph into the bundle graph.
- [ ] This currently only works if a dependency imports symbols that all eventually resolve to the same asset, otherwise it behaves just as before. So I think splitting up the imports to be per-symbol should happen in the JS transformer, otherwise we'd have a single dependency (id) that actually resolves to different assets. We might have to change how the dependency id is derived for that
- [ ] add tests for new behaviour
- [ ] real world testing

A pleasant surprise:
Doing `import("lodash-es").then(({add}) => add(1,2))` will be redirected to behave like `import("lodash-es/add").then(...)`

### DIFF
--- a/packages/core/core/src/AssetGraph.js
+++ b/packages/core/core/src/AssetGraph.js
@@ -32,8 +32,9 @@ import {type ProjectPath, fromProjectPathRelative} from './projectPath';
 
 export const assetGraphEdgeTypes = {
   null: 1,
-  //TODO
-  original: 2,
+  // In addition to the null edge, a dependency can be connected to the asset containing the symbols
+  // that the dependency requested (after reexports were skipped).
+  redirected: 2,
 };
 
 export type AssetGraphEdgeType = $Values<typeof assetGraphEdgeTypes>;
--- a/packages/core/core/src/AssetGraph.js
+++ b/packages/core/core/src/AssetGraph.js
@@ -68,7 +68,7 @@ export function nodeFromDep(dep: Dependency): DependencyNode {
     usedSymbolsDownDirty: true,
     usedSymbolsUpDirtyDown: true,
     usedSymbolsUpDirtyUp: true,
-    symbolTarget: new Map(),
+    symbolTarget: null,
   };
 }
 
--- a/packages/core/core/src/AssetGraph.js
+++ b/packages/core/core/src/AssetGraph.js
@@ -30,6 +30,14 @@ import {ContentGraph} from '@parcel/graph';
 import {createDependency} from './Dependency';
 import {type ProjectPath, fromProjectPathRelative} from './projectPath';
 
+export const assetGraphEdgeTypes = {
+  null: 1,
+  //TODO
+  original: 2,
+};
+
+export type AssetGraphEdgeType = $Values<typeof assetGraphEdgeTypes>;
+
 type InitOpts = {|
   entries?: Array<ProjectPath>,
   targets?: Array<Target>,
@@ -43,7 +51,7 @@ type AssetGraphOpts = {|
 |};
 
 type SerializedAssetGraph = {|
-  ...SerializedContentGraph<AssetGraphNode>,
+  ...SerializedContentGraph<AssetGraphNode, AssetGraphEdgeType>,
   hash?: ?string,
   symbolPropagationRan: boolean,
 |};
@@ -110,7 +118,10 @@ export function nodeFromEntryFile(entry: Entry): EntryFileNode {
   };
 }
 
-export default class AssetGraph extends ContentGraph<AssetGraphNode> {
+export default class AssetGraph extends ContentGraph<
+  AssetGraphNode,
+  AssetGraphEdgeType,
+> {
   onNodeRemoved: ?(nodeId: NodeId) => mixed;
   hash: ?string;
   envCache: Map<string, Environment>;
--- a/packages/core/core/src/AssetGraph.js
+++ b/packages/core/core/src/AssetGraph.js
@@ -60,7 +60,7 @@ export function nodeFromDep(dep: Dependency): DependencyNode {
     usedSymbolsDownDirty: true,
     usedSymbolsUpDirtyDown: true,
     usedSymbolsUpDirtyUp: true,
-    symbolTarget: null,
+    symbolTarget: new Map(),
   };
 }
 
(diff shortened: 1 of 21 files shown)


mischnic:
Superseded by https://github.com/parcel-bundler/parcel/pull/8432

PR #8411, opened 2022-08-23
Generate typescript for SVGs when using svgr and typescript option

### LINKED ISSUE #8376: Standard SVG transform pipeline doesn't understand Typescript
# 🐛 bug report

The standard SVG transform pipeline using `@parcel/transformer-svg-react` doesn't understand Typescript emitted by SVGR and assumes it is pure JS.

.parcelrc
```js
{
    "resolvers": ["...", "@parcel/resolver-glob"],
    "transformers": {
        "*.svg": ["...", "@parcel/transformer-svg-react"]
    }
}

```
.svgrrc.yaml
```yaml
# When generating icons from raw svgs, we need to expose the color value as a prop to make it modifiable in React
replaceAttrValues: { "#011e41": "{props.color}", "#adb9c3": "{props.color}" }
icon: true
dimensions: true
typescript: true
```

## 🤔 Expected Behavior

When `typescript: true` is set in the svgrrc.yaml, the pipeline should process the output of `@parcel/transformer-svg-react` as TypeScript and output typed definitions of svgs

## 😯 Current Behavior

The pipeline throws an error on encountering a `:`, presumably the type definition of the SVG.

```sh
🚨 Build failed.

@parcel/transformer-js: Expected ',', got ':'
```

## 💁 Possible Solution

Forcing Parcel to use a Typescript transform in the pipeline resolves the issue and generates proper types for the SVGs

```js
{
    "resolvers": ["...", "@parcel/resolver-glob"],
    "transformers": {
        "*.svg": ["...", "@parcel/transformer-svg-react", "@parcel/transformer-typescript-tsc"]
    }
}
```

## 🔦 Context

Generate an Icon library from SVG files that works with a Typescript development environment.

## 🌍 Your Environment

| Parcel           | 2.7.0
| Node             | 16.13.1
| npm/Yarn         | 3.2.2
| Operating System | macOS 12.4

samjt:
# ↪️ Pull Request

Generate typescript using SVGR's `typescript:true` option in `svgrrc.json`
SVGR can generate typescript as well as plain javascript react components, but the option throws an error in Parcel because the input in the transformer is hard coded to `jsx`.

Fixes #8376
Related: #7741

## 🚨 Test instructions

Included integration tests svg-react-typescript use a `.svgrrc.json` file and a `package.json` to transform to tsx and generate types based on the interface defined in `react.ts`

run `yarn test:integration --grep 'SVG'` to get the specific tests run

## ✔️ PR Todo


packages/core/integration-tests/test/svg-react-typescript.js
```diff
@@ -0,0 +1,32 @@
+import assert from 'assert';
+import {bundle, outputFS} from '@parcel/test-utils';
+import path from 'path';
+
+describe('svg-react-typescript', () => {
```
mischnic: Please add this test to this existing file https://github.com/parcel-bundler/parcel/blob/58823e43b8717cf499e92667da380a9b942a4afe/packages/core/integration-tests/test/svg-react.js#L5 instead (and as one test with all assertions, similar to the existing one)

packages/transformers/svg-react/src/SvgReactTransformer.js
```diff
@@ -17,7 +17,13 @@ function getComponentName(filePath) {
 }
 
 export default (new Transformer({
-  async transform({asset}) {
+  async loadConfig({config}) {
+    let {contents} =
+      (await config.getConfig(['.svgrrc.json', '.svgrrc'])) || {};
+
+    return contents;
```
mischnic: Nit: this might be easier to read:

```suggestion
    let conf = await config.getConfig(['.svgrrc.json', '.svgrrc'])
    return conf?.contents;
```

packages/core/integration-tests/test/svg-react-typescript.js
samjt: @mischnic all done, thanks for the feedback

samjt:
@mischnic changes ready, thanks for the review

packages/core/integration-tests/test/svg-react.js
```diff
@@ -20,3 +20,26 @@ describe('svg-react', function () {
     assert(file.includes('_react.createElement("svg"'));
   });
 });
+
+describe('svg-react-typescript', function () {
```
mischnic: We usually have one top-level describe block per file
```suggestion

```

mischnic:
The lint check is complaining that `packages/transformers/svg-react/src/SvgReactTransformer.js` should be reformatted with prettier.

packages/core/integration-tests/test/svg-react.js
```diff
@@ -19,4 +19,24 @@ describe('svg-react', function () {
...
+    let b = await bundle(
+      path.join(__dirname, '/integration/svg-react-typescript/react.ts'),
+      {
+        defaultConfig: path.join(
+          __dirname,
+          'integration/custom-configs/.parcelrc-svg-react',
+        ),
+        defaultTargetOptions: {
+          typescript: true,
+        },
```
mischnic: This option doesn't exist
```suggestion
```

samjt: For some reason I thought those options were being passed to the plug-in, but that doesn’t make sense I realise

samjt:
Sorry, a lot of back and forth on chores…

samjt:
@mischnic I think this is all ready to merge

mischnic (approved):
Thank you!

PR #8424, opened 2022-08-28
Fix SW packaging in web extensions

olistic:
# ↪️ Pull Request

This PR fixes the packaging of Service Workers using ES Modules in the `webextension` config. 

The JS transform does not support ES Modules for service workers:

https://github.com/parcel-bundler/parcel/blob/3329469f50de9326c5b02ef0ab1c0ce41393279c/packages/transformers/js/src/JSTransformer.js#L575-L577

Instead, it sets `outputFormat` to `global` ([docs](https://parceljs.org/features/targets/#outputformat)). For that reason, the `type=module` is removed when packaging the extension.

Related issue: https://github.com/parcel-bundler/parcel/issues/1747.

## 🚨 Test instructions

Here's a repo in which builds fail with `2.7.0` but pass with this PR:

https://github.com/olistic/parcel-webextension-repro

## ✔️ PR Todo

### DIFF
--- a/packages/packagers/webextension/src/WebExtensionPackager.js
+++ b/packages/packagers/webextension/src/WebExtensionPackager.js
@@ -25,7 +25,7 @@ export default (new Packager({
     const manifest = JSON.parse(await asset.getCode());
 
     if (manifest.background?.type === 'module') {
-      // service workers are build with output format 'global'
+      // service workers are built with output format 'global'
       // see: https://github.com/parcel-bundler/parcel/blob/3329469f50de9326c5b02ef0ab1c0ce41393279c/packages/transformers/js/src/JSTransformer.js#L577
       delete manifest.background.type;
     }
--- a/packages/packagers/webextension/src/WebExtensionPackager.js
+++ b/packages/packagers/webextension/src/WebExtensionPackager.js
@@ -23,6 +23,13 @@ export default (new Packager({
       relativeBundlePath(bundle, b, {leadingDotSlash: false});
 
     const manifest = JSON.parse(await asset.getCode());
+
+    if (manifest.background?.type === 'module') {
+      // service workers are build with output format 'global'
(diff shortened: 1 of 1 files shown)


louisgv:
@olistic how would we implement real module support when it comes to importing code into background sw? Or is that not necessary since parcel would be the "bundler" anyhow :-?... Just thinking outloud, since this PR is more of a hack to get over real-import and bundle them all into the output file (instead of using importScript/proper module import (?))

packages/packagers/webextension/src/WebExtensionPackager.js
```diff
@@ -23,6 +23,13 @@ export default (new Packager({
       relativeBundlePath(bundle, b, {leadingDotSlash: false});
 
     const manifest = JSON.parse(await asset.getCode());
+
+    if (manifest.background?.type === 'module') {
+      // service workers are build with output format 'global'
```
olistic:
```suggestion
      // service workers are built with output format 'global'
```

olistic:
> @olistic how would we implement real module support when it comes to importing code into background sw? Or is that not necessary since parcel would be the "bundler" anyhow :-?... Just thinking outloud, since this PR is more of a hack to get over real-import and bundle them all into the output file (instead of using importScript/proper module import (?))

Right now it's not necessary since Parcel is bundling the SW code anyhow, and since it leaves the `type=module` in the manifest, it's failing as what Parcel outputs isn't actually a module.

louisgv:
> it's failing as what Parcel outputs isn't actually a module.

Yeah, so the way to actually fix that is to fix the js transformer to output service worker module... tbh I'm not sure how useful it will be, since bgsw are loaded in the background anyhow. 

BTW, thanks for the patch - I incorporated it into my fork of this packager. (used in plasmo: https://github.com/PlasmoHQ/plasmo/)

devongovett (approved):

olistic:
@devongovett Thanks for merging. Any idea when this will be released?

devongovett:
Working on testing and preparing a release soon (next few weeks hopefully). In the meantime, you could try out the nightly releases.

PR #8432, opened 2022-08-31
Code splitting across reexports using symbol data by splitting dependencies

### LINKED ISSUE #7392: Parcel does not split side effect free re-exports across bundles

mischnic:
Closes https://github.com/parcel-bundler/parcel/issues/7392

Symbol propagation already does the traversals anyway, so now it just always tracks for each symbol in a dependency, what that symbol should resolve to.

So for each used symbol, it now says where that symbol should resolve to. This is not exectly the same as `getSymbolResolution` though, as it stops at non-sideeffect-free assets (while `getSymbolResolution` does resolve through them).

This information is then used in `BundleGraph.fromAssetGraph` to split up dependencies to be per symbol (and point them at the resolved asset according to the previous analysis). The old dependency is kept in case some transformer used its placeholder. This changes the execution order because the old dependency is replaced with `dep.usedSymbolsUp.map(... => new Dependency(...))` so the arbitrary (sorted alphabetically) order of the symbols is used for this.

~Since it's possible to rename symbols via reexports, I've had to add an additional map to the dependency node, e.g. a dependency to a side-effect-free asset that does `export {foo as bar} from "y";` would then be rewritten to point directly to `y` instead, and the `symbolTarget` map then has `foo -> bar` which is applied transparently in `bundleGraph.getUsedSymbols` and `bundleGraph.getSymbols` (which I think is a better solution compared to mutating the actual DependencyValue symbols entry).~

- [x] Maybe get rid of `dependency.symbolTarget`

### DIFF
--- a/packages/runtimes/service-worker/src/ServiceWorkerRuntime.js
+++ b/packages/runtimes/service-worker/src/ServiceWorkerRuntime.js
@@ -11,7 +11,8 @@ export default (new Runtime({
     let asset = bundle.traverse((node, _, actions) => {
       if (
         node.type === 'dependency' &&
-        node.value.specifier === '@parcel/service-worker'
+        node.value.specifier === '@parcel/service-worker' &&
+        !bundleGraph.isDependencySkipped(node.value)
       ) {
         actions.stop();
         return bundleGraph.getResolvedAsset(node.value, bundle);
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -68,9 +68,9 @@ checksum = "8da52d66c7071e2e3fa2a1e5c6d088fec47b593032b254f5e980de8ea54454d6"
 
 [[package]]
 name = "ast_node"
-version = "0.8.1"
+version = "0.8.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "87549fcb780f81054407f313a1693d102396c223f5c49ccc5d90b46a6cbef34a"
+checksum = "1a36288803cd1605bc4f0e3189970a0db8e602bb01a39f8133889f35ece7ddde"
 dependencies = [
  "darling",
  "pmutil",
@@ -1299,9 +1299,9 @@ checksum = "925383efa346730478fb4838dbe9137d2a47675ad789c546d150a6e1dd4ab31c"
 
 [[package]]
 name = "preset_env_base"
-version = "0.2.4"
+version = "0.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5f68dc2366d2258e280ad44221403aa0af50868b3e6dc1cb9fb14a302cc01948"
+checksum = "ec3fc6fae03023ee1badb75ccbefc36379c419754b30e3848b77a4327f4ec1e2"
 dependencies = [
  "ahash",
  "anyhow",
@@ -1690,9 +1690,9 @@ checksum = "73473c0e59e6d5812c5dfe2a064a6444949f089e20eec9a2e5506596494e4623"
 
 [[package]]
 name = "swc_atoms"
-version = "0.2.13"
+version = "0.4.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0d99c0ac33707dd1162a3665d6ca1a28b2f6594e9c37c4703e417fc5e1ce532e"
+checksum = "454bf2d73485f6c4af9c91e70ce4fd4f17e9294d37b9f2037a3c4c2fe54b598d"
 dependencies = [
  "once_cell",
  "rustc-hash",
@@ -1703,9 +1703,9 @@ dependencies = [
 
 [[package]]
 name = "swc_cached"
-version = "0.1.1"
+version = "0.3.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "84fed4a980e12c737171a7b17c5e0a2f4272899266fa0632ea4e31264ebdfdb5"
+checksum = "395389d54bea607246bb4a400b9b3df2e55adfe8fcce7965a5b99fed7816cf4d"
 dependencies = [
  "ahash",
  "anyhow",
@@ -1718,9 +1718,9 @@ dependencies = [
 
 [[package]]
 name = "swc_common"
-version = "0.23.0"
+version = "0.27.13"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "68e76a324fa0d7240e790c78914f39fdecfa9d87ef4efed591124b58607a4a4a"
+checksum = "cba38a2f1291fcf3f78f357802b8cec72ecf5e95808e9d937783e60cd3570b93"
 dependencies = [
  "ahash",
  "ast_node",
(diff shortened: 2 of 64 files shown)


packages/runtimes/service-worker/src/ServiceWorkerRuntime.js
```diff
@@ -11,7 +11,8 @@ export default (new Runtime({
     let asset = bundle.traverse((node, _, actions) => {
       if (
         node.type === 'dependency' &&
-        node.value.specifier === '@parcel/service-worker'
+        node.value.specifier === '@parcel/service-worker' &&
+        !bundleGraph.isDependencySkipped(node.value)
```
mischnic, line 15: This was needed because there are now two such dependencies, one of the being excluded.

devongovett (commented):
Code looks good to me. Would like to do some testing on some large apps to make sure we don't break anything.

parcel-benchmark:
## Benchmark Results

#### Timings

| --- | --- | --- |
| Cold | 1.40s | -27.00ms |
| Cached | 342.00ms | -1.00ms |

#### Cold Bundles

*No bundle changes detected.*

#### Cached Bundles

*No bundle changes detected.*

</p></details>

#### Timings

| --- | --- | --- |

#### Cold Bundles

#### Cached Bundles

</p></details>

#### Timings

| --- | --- | --- |

#### Cold Bundles

#### Cached Bundles

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 6.80s | +275.00ms |
| Cached | 314.00ms | +21.00ms ⚠️ |

#### Cold Bundles

*No bundle changes detected.*

#### Cached Bundles

*No bundle changes detected.*

</p></details>

**[Click here to view a detailed benchmark overview.](https://benchmark.parceljs.org/benchmark/342116996563862100)**

devongovett (approved):
Let's get this into nightly releases to test.

marcins:
> Code looks good to me. Would like to do some testing on some large apps to make sure we don't break anything.

FYI - I have this change patched in to our very large app for a few days now without build issues, and no runtime issues we're yet aware of (dog-fooding is currently limited for the Parcel builds). It's provided an improvement in runtime performance (hard to quantify as we had a workaround in place rewriting imports with a transform, but it's provided a further 10% improvement over the workaround - I think we're roughly at 30-40% improvement over the baseline we had).

PR #8452, opened 2022-09-07
Add  parcel-query function to locate a symbol by local name

lettertwo:
Based on the output for `findAsset`.

Some examples of how it works:

An imported symbol:
```
> .findAssetWithSymbol $8f94277571a64835$export$e08c7d021b829b7a
bRRst node_modules/stylis/dist/stylis.mjs
imported as prefixer from node_modules/stylis/src/Enum.js
```

A declared symbol:
```
> .findAssetWithSymbol $3780a8199c4f575f$var$mapping
4LrcW node_modules/@parcel/runtime-js/lib/helpers/bundle-manifest.js
possibly defined as mapping
```

### DIFF
--- a/packages/dev/query/src/cli.js
+++ b/packages/dev/query/src/cli.js
@@ -155,16 +155,20 @@ function findAssetWithSymbol(local: string) {
   );
 
   let asset;
-  outer: for (let node of assetGraph.nodes.values()) {
-    if (node.type === 'asset') {
-      // Search against the id used by the JSTransformer and ScopeHoistingPackager,
-      // not the final asset id, as it may have changed with further transformation.
-      if (node.value.meta.id === assetId) {
-        asset = node;
-        break;
-      } else if (node.value.symbols) {
-        // If the asset couldn't be found by searching for the id,
-        // search for the local name in asset used symbols.
+  // Search against the id used by the JSTransformer and ScopeHoistingPackager,
+  // not the final asset id, as it may have changed with further transformation.
+  for (let node of assetGraph.nodes.values()) {
+    if (node.type === 'asset' && node.value.meta.id === assetId) {
+      asset = node;
+      break;
+    }
+  }
+
+  // If the asset couldn't be found by searching for the id,
+  // search for the local name in asset used symbols.
+  if (asset == null) {
+    outer: for (let node of assetGraph.nodes.values()) {
+      if (node.type === 'asset' && node.value.symbols) {
         for (let symbol of node.value.symbols.values()) {
           if (symbol.local === local) {
             asset = node;
--- a/packages/dev/query/src/cli.js
+++ b/packages/dev/query/src/cli.js
@@ -121,22 +121,29 @@ function getAsset(v: string) {
   }
 }
 
-function findAsset(v: string) {
+function _findAssetNode(v: string) {
   let assetRegex = new RegExp(v);
   for (let node of assetGraph.nodes.values()) {
     if (
       node.type === 'asset' &&
       assetRegex.test(fromProjectPathRelative(node.value.filePath))
     ) {
-      try {
-        console.log(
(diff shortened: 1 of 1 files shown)


packages/dev/query/src/cli.js
```diff
@@ -130,23 +131,105 @@ function getAsset(v: string) {
...
+  }
+}
+
+function findAssetWithSymbol(local: string) {
+  let [, assetId, binding, ref] = nullthrows(
+    local.match(/^\$([^$]+)\$([^$]+)\$(.*)$/),
+    `symbol ${local} could not be resolved`,
+  );
+
+  let asset = bundleGraph._graph.getNodeByContentKey(assetId);
```
mischnic: If you have plugins running after the JS transformer, then `assetId` will be the asset id at the point when the JS transformer ran, and not the final one in the graph. So ideally this would be something like `assetGraph.nodes.find(n => n.type === "asset" && n.value.meta.id === assetId)`

parcel-benchmark:
## Benchmark Results

#### Timings

| --- | --- | --- |
| Cold | 1.51s | -25.00ms |
| Cached | 342.00ms | -42.00ms 🚀 |

#### Cold Bundles

*No bundle changes detected.*

#### Cached Bundles

| dist/legacy/index.html | 826.00b | +0.00b | 516.00ms | -39.00ms 🚀 |
| dist/modern/index.html | 749.00b | +0.00b | 515.00ms | -40.00ms 🚀 |

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 9.80s | -103.00ms |
| Cached | 456.00ms | -5.00ms |

#### Cold Bundles

*No bundle changes detected.*

#### Cached Bundles

*No bundle changes detected.*

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 1.77m | +207.00ms |
| Cached | 2.30s | -16.00ms |

#### Cold Bundles

*No bundle changes detected.*

#### Cached Bundles

*No bundle changes detected.*

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 7.09s | -82.00ms |
| Cached | 277.00ms | +5.00ms |

#### Cold Bundles

*No bundle changes detected.*

#### Cached Bundles

*No bundle changes detected.*

</p></details>

**[Click here to view a detailed benchmark overview.](https://benchmark.parceljs.org/benchmark/350961030289949267)**

packages/dev/query/src/cli.js
lettertwo: Ah interesting! I thought I had found a shortcut 😆

packages/dev/query/src/cli.js
```diff
@@ -130,23 +131,113 @@ function getAsset(v: string) {
...
+  for (let node of assetGraph.nodes.values()) {
+    if (node.type === 'asset' && node.value.meta.id === assetId) {
+      asset = node;
+      break;
+    }
+  }
+
+  // If the asset couldn't be found by searching for the id,
+  // search for the local name in asset used symbols.
+  if (asset == null) {
```
mischnic, line 169: Did you actually run into such a case? I can't think of a situation where the first search would fail

lettertwo: I think you're right that this _shouldn't_ be possible, but if there is an undefined symbol at runtime (due to a bug in Parcel), it likely won't be discoverable via the asset graph, but _might_ still be discoverable in used symbols. (I have one such case that I am investigating and can share).

lettertwo: Here's the case i mentioned: https://github.com/parcel-bundler/parcel/pull/8468

AGawrys (approved):
I think this is super useful in debugging issues that manifest as missing symbol (even if its not a Parcel bug, but a user's mishap)

mischnic (approved):

PR #8491, opened 2022-09-20
Package on main thread for small rebuilds

thebriando:
# ↪️ Pull Request
This change prevents `WriteBundlesRequest` from serializing across all workers with small rebuilds which _should_ improve save/warm build times when small changes that only impact a single bundle are made to the project.

Tested saving a file with a small change in a large project

Before:
Save time | ✨ Built in 154.08s
-- | --
Packaging phase | 13.614s

After:
Save time | ✨ Built in 134.72s 🚀 12.56% faster
-- | --
Packaging phase | 1.948s 🚀 85.70% faster

## ✔️ PR Todo


packages/core/core/src/requests/WriteBundlesRequest.js
```diff
@@ -83,6 +79,20 @@ async function run({input, api, farm, options}: RunInput) {
     return true;
   });
 
+  let ref;
+  let dispose;
+
+  // Create shared reference in WorkerFarm if we need to change multiple bundles and if we
+  // can't skip the subrequests for all the bundles
+  if (
```
thebriando, line 88: Not sure if I got this logic quite right.. are there any other cases to look for here?

parcel-benchmark:
## Benchmark Results

#### Timings

| --- | --- | --- |
| Cold | 1.39s | +7.00ms |
| Cached | 319.00ms | -63.00ms 🚀 |

#### Cold Bundles

*No bundle changes detected.*

#### Cached Bundles

*No bundle changes detected.*

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 8.77s | -18.00ms |
| Cached | 428.00ms | -1.00ms |

#### Cold Bundles

*No bundle changes detected.*

#### Cached Bundles

*No bundle changes detected.*

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 1.47m | +411.00ms |
| Cached | 2.50s | -70.00ms |

#### Cold Bundles

*No bundle changes detected.*

#### Cached Bundles

*No bundle changes detected.*

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 6.31s | -92.00ms |
| Cached | 266.00ms | -17.00ms 🚀 |

#### Cold Bundles

*No bundle changes detected.*

#### Cached Bundles

*No bundle changes detected.*

</p></details>

**[Click here to view a detailed benchmark overview.](https://benchmark.parceljs.org/benchmark/346986744074207827)**

packages/core/types/index.js
```diff
@@ -1754,6 +1755,7 @@ export type PackagingProgressEvent = {|
   +type: 'buildProgress',
   +phase: 'packaging',
   +bundle: NamedBundle,
+  +bundleGraph?: InternalBundleGraph,
```
mischnic: 1. Do we really want to add the bundle graph to this event? (This is the public plugin API)
2. If yes, then it certainly needs to be the public bundle graph type

thebriando: Hmm I think the main reason why I added this was because the ReporterRunner gets the BundleGraph from the WorkerFarm, which wouldn't exist in this case since we don't serialize the BundleGraph, so I sent the BundleGraph through the packaging/optimizing events.

https://github.com/parcel-bundler/parcel/blob/cf5b3ffdaadb3de90104a6d24261e1aad9c496fd/packages/core/core/src/ReporterRunner.js#L67-L73

The invariant here also looks like it uses the private BundleGraph type?

devongovett: Maybe we should make an internal ReporterEvent type that has this instead?

packages/core/types/index.js
```diff
@@ -1764,6 +1768,11 @@ export type BundlingProgressEvent = {|
   +phase: 'bundling',
 |};
 
+export type BundleGraphEvent = {|
+  +type: 'bundleGraph',
+  +bundleGraph: BundleGraph<NamedBundle>,
+|};
```
mischnic, line 1775: What Devon meant in https://github.com/parcel-bundler/parcel/pull/8491#discussion_r1004751445 is that there could an internal event between PackagerRunner and ReporterRunner (which the reporter runner then just uses for setting the bundle graph but doesn't send it out to the reporter plugins).

But I think it's strange to have to split it up into two events in the first place. 
Why don't you use the same idea of passing either the bundlegraph ref or the bundlegraph object itself in the `packaging` event? The `PackagerRunnner` is already running on the main thread, so calling `report(something)` wouldn't perform any serialization anyway.

packages/core/core/src/requests/WriteBundlesRequest.js
mischnic: I'm not sure either, but I'm confused that the comment says "can't skip all the bundles = at least 1 bundle can't be skipped" but then the if condition checks "there are at least 2 bundles which can't be skipped".

devongovett:
I've refactored the WorkerFarm in a way that should make this a bit simpler to implement: #8589.

PR #8501, opened 2022-09-23
Fix #8500 by not adding offset to line numbers

### LINKED ISSUE #8500: VueTransformer uses nonsensical values for codeFrames.codeHighlights line numbers
I mentioned in my PR (https://github.com/parcel-bundler/parcel/pull/8499) that the code previews don't work and hypothesised I was doing something wrong.

Looking at the line numbers it was generating, they're all nonsensical:

```
line: err.loc.start.line + err.loc.start.offset,
```

I don't get what this is supposed to do, but it generates line numbers like `1483` in a `155` line file!

Looking at the Vue SFC compiler's code

https://github.com/vuejs/core/blob/fbd697a4b6c238147d661026836314862fd09221/packages/compiler-sfc/src/compileTemplate.ts#L307-L312

It's already adjusting the line numbers by an offset, and indeed if I change the code to _not_ add this offset, the code previews start to work fine:

```js
if (err.loc) {
  diagnostic.codeFrames = [{
    filePath,
    codeHighlights: [{
      start: {
        line: err.loc.start.line,
        column: err.loc.start.column
      },
      end: {
        line: err.loc.end.line,
        column: err.loc.end.column
      }
    }]
  }];
}
```

lol768:
Closes #8500 

Code snippet previews work after applying this patch, and the line numbers make sense again:

image

### DIFF
--- a/packages/transformers/vue/src/VueTransformer.js
+++ b/packages/transformers/vue/src/VueTransformer.js
@@ -192,11 +192,11 @@ function createDiagnostic(err, filePath) {
         codeHighlights: [
           {
             start: {
-              line: err.loc.start.line + err.loc.start.offset,
+              line: err.loc.start.line,
               column: err.loc.start.column,
             },
             end: {
-              line: err.loc.end.line + err.loc.end.offset,
+              line: err.loc.end.line,
               column: err.loc.end.column,
             },
           },


mischnic:
Can you add a test case to https://github.com/parcel-bundler/parcel/blob/v2/packages/core/integration-tests/test/vue.js for this example?

So duplicate https://github.com/parcel-bundler/parcel/tree/v2/packages/core/integration-tests/test/integration/vue-basic and make it fail with an error like in your screenshot, then add an assertion like this one https://github.com/parcel-bundler/parcel/blob/e3a39a1f3642d28f8fd25919cda698ad34502cee/packages/core/integration-tests/test/javascript.js#L1719-L1757

PR #8511, opened 2022-09-26
Add ESM support to glob resolver

samjt:
# ↪️ Pull Request

Checks if `dependency.specifierType` is esm and then writes imports and exports in and ES6 style instead of commonjs style requires from glob resolved javascript files.

If it is marked as ESM then generate an import using the format

```js
import ${key} from ${JSON.stringify(matches)};
```

instead of

```js
const ${key} = require(${JSON.stringify(matches)});
```

exports will be written as 

```js
export { _temp0 as "a" }
```

instead of

```js
module.exports = {"a": _temp0}
```

## 💻 Examples

When in a project with module set:
```js
import * as React from 'react';
import * as Icons from './svg/*.svg';

export default function () {
  const {SVG_Logo} = Icons;

  console.log('>> SVG_Logo should be a React functional component, SVG_Logo.default should be undefined', SVG_Logo, SVG_Logo.default);

  return (
    <>
      <SVG_Logo />
      <div>hello!</div>
    </>
  );
}
```

## 🚨 Test instructions

`yarn test:integration --g 'glob'`

## ✔️ PR Todo


samjt:
One point to note:

async imports are treated as esm based on the dependency specifiertype returned to the resolver.

packages/resolvers/glob/src/GlobResolver.js
```diff
@@ -205,19 +215,27 @@ function set(obj, path, value) {
...
     if (isAsync) {
-      return {
-        imports: '',
-        value: `() => import(${JSON.stringify(matches)})`,
-        count,
-      };
+      imports = `const ${key} = () => import(${JSON.stringify(matches)});`;
+    } else {
+      if (specifierType === 'esm') {
+        imports = `import ${key} from ${JSON.stringify(matches)};`;
```
devongovett: Should this be `import * as ${key} ...`? As it is, this will only return the default export, whereas the CommonJS and dynamic imports will return the full module namespace.

samjt: @devongovett  So the problem is that importing with `import * as key` seems to trigger the JavaScript packager to treat the import as requiring wrapping in an "esm in commonjs" compatibility layer (as I understand it)

The end result is you get this (pseudocode) as the export from the glob import:

```js 
{ 
  esmodule:true, 
  default: function(){ 
    return { 
      esmodule:true, 
      default: function()...
    }
  }
}
```

So each import is treated as requiring a wrapper inside the index which also gets a wrapper. This doesn't make sense to me because the outermost export _might_ need a compatibility wrapper if it were being built in a commonjs environment but the rest inside it shouldn't as they're being imported into an esm file.
Have I understood this correctly?

samjt: I’m also not sure how to change this behaviour, I’m assuming it’s in the JavaScript packager, but not sure where and what impact that might have

devongovett:
One question: where did this requirement come from? Did you have an issue that was caused by using CommonJS? I think the tree shaking implementation should handle both the same way, so just curious where this PR is coming from.

samjt:
@devongovett 
It came from attempting to build an SVG library that exported as pure esm. 
When using the glob resolver to do this I would get back an objects rather than functions:

```js
import * as Icons from  "./svg/*.svg"
const { AnIcon } = Icons;
console.log(AnIcon) // > { esmodule: true, default: function }
<AnIcon /> // Does not render
<AnIcon.default /> //Does render

```
I believed this was because the glob resolver was incorrectly exporting them as commonjs compatible, and by making the changes above, I was able to import directly and use the icons without having to reference default.

devongovett:
But with the change to `import *` rather than only the default, wouldn't this now have the same issue?

PR #8453, opened 2022-09-07
Attach internalized orphans and consider types in PRL

AGawrys:
# 🐰 Pull Request

This PR address experimental bundler issues with CompiledCss extraction enabled. 

**Problem 1** was that upon internalization of async bundles, we allowed for orphans to be left behind (of type change bundles) and thus were **deleted from the graph**. 

**Solution:** This PR adds the function `attachOrphans` which makes sure that any children of an async bundle in the bundleGraph (these will only be non async code split point bundles) are then connected to whatever parent the async bundle was internalized into.

I've also moved the merge bundles of type change step to be after internalization, so that we don't need to remerge any type change bundles after internalization, which could add a lot of extra work.

**Problem 2** arose after Issue 1 was fixed. We had been processing type change bundles within the parallel request limit code. This would've been okay, since we only merge back / alter bundles with source bundles. However, when a reused bundle (an async bundle with source bundles) gets merged back, we know it may have children, so we propagate the reused bundle's source bundles to its children and attach it to the new parent. This is because we assumed the children of an async bundle at that point would only be shared bundles. 
But that is not the case.

**Solution**:  Instead, upon merging back a reused bundle, its child of a different type must be either connected to the new parent OR its assets merged into the new parent's already existing bundle of that type. 

## 💻 Examples

The Grey Graphs below are the "bare bones" bundleGraph we use in the experimental bundler, which display only bundles loaded in parallel and their children. So all async bundles are all on the same "level" (i.e. connected to root) and their children are parallel or type change bundles, which are connected to what pulls them in. 

We maintain sync relationships and async referencing relationships in supplementary graphs which are not shown but the drawings display these relationships.

**INTERNALIZATION ORPHAN BUG**

First, here is a visual representation of the dependencies each file has. 

So, given that setup, we know `foo.js` should be **internalized** into `index.js`. But `foo.js` has two "children". One is displayed in the graph below as it is a type change and the other is `bazz.js` which is an async bundle but is referenced by or "pulled in" by foo.  

bundleGraphIdeal_POST_STEP1

We need to make sure we patch up these would-be orphans. So, connect them to the parent. In the `bazz.js` case we change around our supplementary graphs instead (i.e. remove `foo` async connections and patch them up)

bundleGraphIdeal_POST_INTERNALIZATION

Type bundles must be merged per bundleGroups so this is why I moved **STEP MERGE BUNDLES** to occur later, so that we don't need to re-merge. 
bundleGraphIdeal_POST_MERGE

[...]

PR #8550, opened 2022-10-17
Support multiple edge types in `Graph.hasEdge`

mattcompiles:
# ↪️ Pull Request

This PR adds some desired functionality discovered when creating #8453. @AGawrys 

Currently, `Graph.hasEdge` only accepts a single edge type to validate against. This PR adds support for:
- An array of possible edge types
- ~The `ALL_EDGE_TYPES` constant, allowing edge checks to be type agnostic.~

~As the graph isn't currently aware of all possible edge types without doing lookups, we (@alshdavid and I) have added a new `edgeTypes` property to the `AdjacencyList` which is appended inside `addEdge`. To keep things simple, we never remove edge types. This shouldn't cause any issues, but does mean it's possible to search for edge types that no longer exist in the graph and edge types are never cleaned up once they have existed.~

~It's possible we'll decide that supporting `ALL_EDGE_TYPES` isn't worth tracking the extra piece of state. If that's the case it should be easy to remove that code and keep just the passed array of edge types param.~ 

After a discussion with @AGawrys we've decided simple array support is enough and `ALL_EDGE_TYPES` isn't worth the complexity. 

## 💻 Examples

```js
graph.addEdge(a, b, 2);

graph.hasEdge(a, b, [1, 2]);
```

## ✔️ PR Todo

### DIFF
--- a/packages/bundlers/experimental/src/ExperimentalBundler.js
+++ b/packages/bundlers/experimental/src/ExperimentalBundler.js
@@ -329,9 +329,7 @@ function createIdealGraph(
   let bundleRootGraph: ContentGraph<
     BundleRoot | 'root',
     $Values<typeof bundleRootEdgeTypes>,
-  > = new ContentGraph({
-    edgeTypes: dependencyPriorityEdges,
-  });
+  > = new ContentGraph();
 
   let bundleGroupBundleIds: Set<NodeId> = new Set();
 
--- a/packages/bundlers/experimental/src/ExperimentalBundler.js
+++ b/packages/bundlers/experimental/src/ExperimentalBundler.js
@@ -329,7 +329,9 @@ function createIdealGraph(
   let bundleRootGraph: ContentGraph<
     BundleRoot | 'root',
     $Values<typeof bundleRootEdgeTypes>,
-  > = new ContentGraph();
+  > = new ContentGraph({
+    edgeTypes: dependencyPriorityEdges,
+  });
 
   let bundleGroupBundleIds: Set<NodeId> = new Set();
 
--- a/packages/core/graph/src/AdjacencyList.js
+++ b/packages/core/graph/src/AdjacencyList.js
@@ -25,6 +25,7 @@ export type SerializedAdjacencyList<TEdgeType> = {|
 export type AdjacencyListOptions<TEdgeType> = {|
   edgeCapacity?: number,
   nodeCapacity?: number,
+  edgeTypes: Uint8Array,
 |};
 
 /** The upper bound above which capacity should be increased. */
@@ -41,7 +42,7 @@ const SHRINK_FACTOR = 0.5;
 export default class AdjacencyList<TEdgeType: number = 1> {
   #nodes /*: NodeTypeMap<TEdgeType | NullEdgeType> */;
   #edges /*: EdgeTypeMap<TEdgeType | NullEdgeType> */;
-  #edgeTypes /* Uint8Array */;
+  #edgeTypes /*: Uint8Array */;
 
   constructor(
     opts?:
@@ -61,6 +62,7 @@ export default class AdjacencyList<TEdgeType: number = 1> {
       let {
         nodeCapacity = NodeTypeMap.MIN_CAPACITY,
         edgeCapacity = EdgeTypeMap.MIN_CAPACITY,
+        edgeTypes,
       } = opts ?? {};
       assert(
         nodeCapacity <= NodeTypeMap.MAX_CAPACITY,
@@ -72,7 +74,7 @@ export default class AdjacencyList<TEdgeType: number = 1> {
       );
       this.#nodes = new NodeTypeMap(nodeCapacity);
       this.#edges = new EdgeTypeMap(edgeCapacity);
-      this.#edgeTypes = new Uint8Array(0);
+      this.#edgeTypes = edgeTypes;
     }
   }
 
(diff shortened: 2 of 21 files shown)


parcel-benchmark:
## Benchmark Results

#### Timings

| --- | --- | --- |
| Cold | 1.90s | -87.00ms |
| Cached | 402.00ms | -21.00ms 🚀 |

#### Cold Bundles

| dist/legacy/parcel.7cdb0fad.webp | 102.94kb | +0.00b | 341.00ms | +28.00ms ⚠️ |
| dist/legacy/parcel.7cdb0fad.webp | 102.94kb | +0.00b | 342.00ms | +27.00ms ⚠️ |
| dist/modern/parcel.7cdb0fad.webp | 102.94kb | +0.00b | 343.00ms | +29.00ms ⚠️ |
| dist/legacy/index.b8ae99ba.css | 94.00b | +0.00b | 350.00ms | +29.00ms ⚠️ |
| dist/modern/index.31cedca9.css | 94.00b | +0.00b | 349.00ms | +28.00ms ⚠️ |

#### Cached Bundles

*No bundle changes detected.*

</p></details>

#### Timings

| --- | --- | --- |

#### Cold Bundles

#### Cached Bundles

</p></details>

#### Timings

| --- | --- | --- |

#### Cold Bundles

#### Cached Bundles

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 8.22s | -24.00ms |
| Cached | 340.00ms | -0.00ms |

#### Cold Bundles

*No bundle changes detected.*

#### Cached Bundles

*No bundle changes detected.*

</p></details>

**[Click here to view a detailed benchmark overview.](https://benchmark.parceljs.org/benchmark/345912305151640147)**

packages/core/graph/src/AdjacencyList.js
```diff
@@ -254,6 +260,10 @@ export default class AdjacencyList<TEdgeType: number = 1> {
     // The edge is already in the graph; do nothing.
     if (edge !== null) return false;
 
+    if (!this.#edgeTypes.includes(type)) {
```
thebriando: So is the point of `this.#edgeTypes` to support `ALL_EDGE_TYPES` in `hasEdge`? I think another possible solution here would be to override `BundleGraph` and `RequestGraph` to pass in their edge types for `hasEdge` which would prevent us from having to serialize another array.

mattcompiles: Yeah we considered that and starting doing it that way originally 😅 . 

We just weren't sure there was a point of truth for all edge types for all the graphs? I do think the costs of serializing for this Array would be insignificant though.

mattcompiles: [Guess this is what you mean for the](https://github.com/parcel-bundler/parcel/blob/a171677e90829b4733397160ad177890cb6f2cee/packages/core/core/src/BundleGraph.js#L42 ) `BundleGraph` ?

I'll have a look into doing it this way. I agree it's more elegant.

thebriando: > We just weren't sure there was a point of truth for all edge types for all the graphs? I do think the costs of serializing for this Array would be insignificant though.

Yeah good point since the list of edge types for any given graph is pretty small though I do think that determining what all the edge types are for a graph inside AdjacencyList is a little weird. Makes more sense to me to pass in each graph's possible edge types into AdjacencyList 🤷 but I think either solution is fine

> [Guess this is what you mean for the](https://github.com/parcel-bundler/parcel/blob/a171677e90829b4733397160ad177890cb6f2cee/packages/core/core/src/BundleGraph.js#L42) BundleGraph ?

Yep! These are the [ones for RequestGraph](https://github.com/parcel-bundler/parcel/blob/a171677e90829b4733397160ad177890cb6f2cee/packages/core/core/src/RequestTracker.js#L54-L61), and `AssetGraph` doesn't use edge types

packages/core/graph/src/Graph.js
```diff
@@ -34,17 +39,24 @@ export default class Graph<TNode, TEdgeType: number = 1> {
...
-    this.adjacencyList = adjacencyList
-      ? AdjacencyList.deserialize(adjacencyList)
-      : new AdjacencyList<TEdgeType>();
+
+    if (adjacencyList) {
+      this.adjacencyList = AdjacencyList.deserialize(adjacencyList);
+    } else {
+      // default edgeTypes to array containing just the null edge type
+      let edgeTypes = new Uint8Array(
+        opts?.edgeTypes ? objectValues(opts.edgeTypes) : [1],
```
mattcompiles: For now I have defaulted `edgeTypes` to just the null edge type. However, I think this is probably a bad idea as it could easily lead to mistakes if more graphs are created. We should probably make it a required property. This will lead to a larger change though as it will mean passing it to all constructors, including tests. 

Thoughts?

packages/bundlers/experimental/src/ExperimentalBundler.js
```diff
@@ -329,7 +329,9 @@ function createIdealGraph(
   let bundleRootGraph: ContentGraph<
     BundleRoot | 'root',
     $Values<typeof bundleRootEdgeTypes>,
-  > = new ContentGraph();
+  > = new ContentGraph({
+    edgeTypes: dependencyPriorityEdges,
```
mattcompiles: @AGawrys This new `edgeTypes` property needs to list all possible edge types for a graph. Is `dependencyPriorityEdges` an exhaustive list of edge types for the experimental bundler?

mattcompiles (commented):
@thebriando I made the changes we spoke about. Had a bit of trouble with the types, I think a lot of the serialize/deserialze types are not quite right which I ended up changing in some cases. 

Let me know what you think.

thebriando (approved):

AGawrys (approved):

PR #8566, opened 2022-10-22
Do not merge isolated bundles in experimental bundler

devongovett:
This make handling of `bundleBehavior` more consistent, and fixes bundle merging logic with isolated bundles. In addition, bundles of different contexts should not be merged. This fixes the react spectrum site

### DIFF
--- a/packages/bundlers/experimental/src/ExperimentalBundler.js
+++ b/packages/bundlers/experimental/src/ExperimentalBundler.js
@@ -540,7 +540,6 @@ function createIdealGraph(
                 typeChangeIds.add(bundleId);
               }
             } else {
-              // Otherwise, merge.
               bundle = bundleGraph.getNode(bundleId);
               invariant(bundle != null && bundle !== 'root');
 
--- a/packages/bundlers/experimental/src/ExperimentalBundler.js
+++ b/packages/bundlers/experimental/src/ExperimentalBundler.js
@@ -603,8 +603,8 @@ function createIdealGraph(
         b !== 'root' &&
         a !== b &&
         typeChangeIds.has(nodeIdB) &&
-        a.bundleBehavior !== 'inline' &&
-        b.bundleBehavior !== 'inline' &&
+        a.bundleBehavior == null &&
+        b.bundleBehavior == null &&
         a.type === b.type
       ) {
         let bundleBbundleGroups = getBundleGroupsForBundle(nodeIdB);
--- a/packages/bundlers/experimental/src/ExperimentalBundler.js
+++ b/packages/bundlers/experimental/src/ExperimentalBundler.js
@@ -511,9 +511,8 @@ function createIdealGraph(
             let bundleGroupRootAsset = nullthrows(bundleGroup.mainEntryAsset);
             if (
               entries.has(bundleGroupRootAsset) &&
-              bundleGroupRootAsset.type === childAsset.type &&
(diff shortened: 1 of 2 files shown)


parcel-benchmark:
## Benchmark Results

#### Timings

| --- | --- | --- |
| Cold | 1.43s | -18.00ms |
| Cached | 328.00ms | -15.00ms |

#### Cold Bundles

*No bundle changes detected.*

#### Cached Bundles

*No bundle changes detected.*

</p></details>

#### Timings

| --- | --- | --- |

#### Cold Bundles

#### Cached Bundles

</p></details>

#### Timings

| --- | --- | --- |

#### Cold Bundles

#### Cached Bundles

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 6.66s | +7.00ms |
| Cached | 287.00ms | -27.00ms 🚀 |

#### Cold Bundles

*No bundle changes detected.*

#### Cached Bundles

*No bundle changes detected.*

</p></details>

**[Click here to view a detailed benchmark overview.](https://benchmark.parceljs.org/benchmark/346464779750081107)**

packages/bundlers/experimental/src/ExperimentalBundler.js
```diff
@@ -670,8 +672,9 @@ function createIdealGraph(
             );
             if (
               bundle !== 'root' &&
-              bundle.bundleBehavior !== 'inline' &&
-              !bundle.env.isIsolated()
+              bundle.bundleBehavior == null &&
```
devongovett, line 674: This line is what caused the test change, where a bundle was previously internalized and now isn't. I think in the general case, however, that isolated bundles should not be reachable.

packages/bundlers/experimental/src/ExperimentalBundler.js
```diff
@@ -534,8 +534,16 @@ function createIdealGraph(
                     : referencingBundle.needsStableName,
               });
               bundleId = bundleGraph.addNode(bundle);
+
               // Store Type-Change bundles for later since we need to know ALL bundlegroups they are part of to reduce/combine them
-              typeChangeIds.add(bundleId);
+              if (parentAsset.type !== childAsset.type) {
+                typeChangeIds.add(bundleId);
```
AGawrys, line 540: Speaking of fixes, I notice in the default bundler we maintain `bundleByType`, which not only holds bundles created via type change, but also all others (lazy/ parallel etc). It looks like that is how we determine where we can merge a type change bundle. 
In experimental, we only attempt to merge type change with type change.... does this mean we're not merging as many bundles as we could in experimental ?

packages/bundlers/experimental/src/ExperimentalBundler.js
```diff
@@ -534,8 +534,16 @@ function createIdealGraph(
...
+              }
+            } else {
+              // Otherwise, merge.
```
AGawrys: This is my mistake, but this comment doesn't make sense here anymore since we don't merge until later

packages/bundlers/experimental/src/ExperimentalBundler.js
devongovett: I think we should not merge parallel/lazy bundles, because the transformer that created those dependencies might not expect it. For example, if you have two `<script>` tags in an HTML page, those have parallel priority, but merging them would result in both scripts pointing to the same URL and (possibly?) executing twice.

AGawrys (approved):

PR #6514, opened 2021-06-25
Update without bundling for a 'simple' non-dependency related change

joeyslater:
# ↪️ Pull Request

[UPDATE from @AGawrys ] This PR now makes the default behavior to skip bundles on "simple" cases. To opt-out, users will need to specify `--no-incremental`  or need to set `NO_INCREMENTAL=1` for tests.

This is automatically disabled for production builds, config changes, and dependency related bundling scenarios. 

This the most basic example of incrementally bundling, adding something like:

- add a comment
- update a string or number value
- add a console log
- updating a namer
- multiple files adding a comment (like a patch or stopping Parcel, making changes, then starting again)

The behavior now looks into whether the update should force a full re-bundle or re-use the existing bundle graph.

## When to do a full re-bundle

* if a dependency is added or removed
* if a symbol is no longer used
* if bundler config has updated
* if a new bundler has been introduced

## 🚨 Test instructions

- [x] Make sure all integration tests pass with the flag on and off.
- [x] Add new test integrations to start the incremental bundling suite for when it should bundle and when it shouldn't

## ✔️ PR Todo

- [x] Added/updated unit tests for this change
- [x] Filled out test instructions (In case there aren't any unit tests)
- [x] Included links to related issues/PRs
- [x] Make opt out @AGawrys + test

PR #8583, opened 2022-10-29
Improve performance of incremental bundling

devongovett:
This is a refactor of #6514, which improves performance even more.

* Runtimes and namers are now no longer applied when incremental bundling, unless one of their sub requests has been invalidated (e.g. config/dev dependency change).
* The asset graph request is moved to be a sub request of the bundle graph request. This way we can avoid making the bundle graph request id based on the asset graph hash, because it'll be naturally invalidated when the asset graph is. That avoids a traversal over the whole graph. We may be able to do this with the write bundles request as well, but I ran into a couple problems for now.
* The cache key for the bundle graph is simplified so it doesn't change on every build. This should also save disk space. The tradeoff is that if an invalidation occurred but nothing actually changed, we wouldn't hit the cache. Also if you undid a change. I think these cases are more rare than changing code, so we should optimize for that. In the tests, nothing hit the cache for these cases ever.

### DIFF
--- a/packages/core/core/src/requests/BundleGraphRequest.js
+++ b/packages/core/core/src/requests/BundleGraphRequest.js
@@ -145,8 +145,8 @@ export default function createBundleGraphRequest(
         res.changedAssets.set(id, asset);
       }
 
-      // $FlowFixMe Added in Flow 0.121.0 upgrade in #4381 (Windows only)
       dumpGraphToGraphViz(
+        // $FlowFixMe Added in Flow 0.121.0 upgrade in #4381 (Windows only)
         res.bundleGraph._graph,
         'BundleGraph',
         bundleGraphEdgeTypes,
--- a/packages/core/core/src/requests/BundleGraphRequest.js
+++ b/packages/core/core/src/requests/BundleGraphRequest.js
@@ -6,12 +6,14 @@ import type ParcelConfig, {LoadedPlugin} from '../ParcelConfig';
 import type {StaticRunOpts, RunAPI} from '../RequestTracker';
 import type {
   Asset,
+  AssetGroup,
   Bundle as InternalBundle,
   Config,
   DevDepRequest,
   ParcelOptions,
 } from '../types';
 import type {ConfigAndCachePath} from './ParcelConfigRequest';
+import type {AbortSignal} from 'abortcontroller-polyfill/dist/cjs-ponyfill';
 
 import invariant from 'assert';
 import assert from 'assert';
@@ -29,8 +31,8 @@ import {unique} from '@parcel/utils';
 import {hashString} from '@parcel/hash';
 import PluginOptions from '../public/PluginOptions';
 import applyRuntimes from '../applyRuntimes';
-import {PARCEL_VERSION} from '../constants';
-import {optionsProxy} from '../utils';
+import {PARCEL_VERSION, OPTION_CHANGE} from '../constants';
+import {assertSignalNotAborted, optionsProxy} from '../utils';
 import createParcelConfigRequest, {
   getCachedParcelConfig,
 } from './ParcelConfigRequest';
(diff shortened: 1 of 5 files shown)


packages/core/core/src/requests/BundleGraphRequest.js
```diff
@@ -202,52 +253,16 @@ class BundlerRunner {
...
-          {
-            bundleGraph: graph,
-            changedAssets: new Map(),
-          },
-          cacheKey,
-        );
-        return graph;
-      }
-    }
-
```
devongovett, line 234: These cases were never hit at all in the tests, and should be very rare. I think removing the traversal to compute the hashes in the common case outweighs the potential benefits for zero change builds.

parcel-benchmark:
## Benchmark Results

#### Timings

| --- | --- | --- |
| Cold | 1.71s | -40.00ms |
| Cached | 307.00ms | -62.00ms 🚀 |

#### Cold Bundles

*No bundle changes detected.*

#### Cached Bundles

*No bundle changes detected.*

</p></details>

#### Timings

| --- | --- | --- |

#### Cold Bundles

#### Cached Bundles

</p></details>

#### Timings

| --- | --- | --- |

#### Cold Bundles

#### Cached Bundles

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 8.06s | +299.00ms |
| Cached | 345.00ms | -18.00ms 🚀 |

#### Cold Bundles

*No bundle changes detected.*

#### Cached Bundles

| dist/Three.js | 580.02kb | +0.00b | 6.21s | +712.00ms ⚠️ |

</p></details>

**[Click here to view a detailed benchmark overview.](https://benchmark.parceljs.org/benchmark/346904562640093778)**

mischnic:
> The cache key for the request graph is simplified so it doesn't change on every build. This should also save disk space. The tradeoff is that if an invalidation occurred but nothing actually changed, we wouldn't hit the cache. Also if you undid a change. I think these cases are more rare than changing code, so we should optimize for that. In the tests, nothing hit the cache for these cases ever.

Did you commit that? I don't see where that's done

packages/core/core/src/requests/BundleGraphRequest.js
```diff
@@ -356,66 +380,19 @@ class BundlerRunner {
...
-    let {cacheKey: updatedCacheKey} = await this.getHashes(graph);
     this.api.storeResult(
       {
-        bundlerHash,
         bundleGraph: internalBundleGraph,
         changedAssets: new Map(),
+        assetRequests: [],
       },
-      updatedCacheKey,
+      this.cacheKey,
```
mischnic, line 389: Why do we need a cache key here? (And tangentially: what does `storeResult` even do compared to just the return value of the run function?)

devongovett:
Yeah, the getCacheKey function was deleted and replaced with this.cacheKey

packages/core/core/src/requests/BundleGraphRequest.js
devongovett: If you don't call storeResult, nothing is stored in the graph. I forget why... Maybe because in this case the cached result is different from the uncached one (eg changedAssets).

Without a cache key, the result is serialized directly in the request graph, but if a cache key is provided then it is stored as a separate cache entry. We wouldn't want to store the entire bundle graph inside the request graph in this case.

mischnic:
But neither RequestGraph nor RequestTracker contain changes? Did you mean

> The cache key for the ~request~ bundle graph is simplified

devongovett:
Oh, oops, yes that is exactly what I meant. 🤦

mischnic (approved):

packages/core/core/src/requests/BundleGraphRequest.js
```diff
@@ -90,24 +89,70 @@ export default function createBundleGraphRequest(
...
+          force: options.shouldBuildLazily && requestedAssetIds.size > 0,
+        },
+      );
+
+      assertSignalNotAborted(signal);
+
+      // If any subrequests are invalid (e.g. dev dep requests or config requests),
+      // bail on incremental bundling. We also need to invalidate for option changes,
+      // which are hoisted to direct invalidations on the bundle graph request.
+      let subRequestsInvalid =
```
AGawrys, line 115: Does this mean that changes like adding a runtime would stop incremental bundling from happening? Thats the only issue I encountered when trying to not apply runtimes for incremental.

devongovett: Yeah, if any config changes or plugin changes occurred, then bundling will not be incremental. I think this should be pretty rare so probably acceptable?

AGawrys (approved):

PR #8586, opened 2022-10-30
Resolve `publicUrl` and `distDir` from environment variables (.env)

villermen:
(I couldn't find any open or closed PRs & issues related to this functionality.)

# ↪️ Pull Request

It's very difficult to provide the desired `publicUrl` or `distDir` arguments to parcel when they are sourced from environment variables. It's near-impossible (to my knowledge) to do so in a cross-platform compatible way from either the CLI or package.json. The environment variables are injected by Parcel and can be used in the build pipeline just fine, but Parcel itself will not be able to pick them up. So while your code might know the correct public URL based on the environment, you will still have to specify it manually to the build command.

This PR proposes a solution similar to how you can specify `PORT` as a fallback environment variable.

## 💻 Examples

```bash
// .env
PARCEL_PUBLIC_URL=/path
PARCEL_DIST_DIR=public
```

Running `parcel build` or `parcel serve` will default to a public URL of "/path" when not specified otherwise (via CLI or targets).

## 🚨 Test instructions

I was unable to test this behavior in practice yet. That will require me to dive a bit deeper into how this repository works,, but I'll be happy to do so once I know whether this is a welcome change or not.

## ✔️ PR Todo

- [ ] Test and add automatic testing.

### DIFF
--- a/packages/core/core/src/resolveOptions.js
+++ b/packages/core/core/src/resolveOptions.js
@@ -97,12 +97,6 @@ export default async function resolveOptions(
     initialOptions?.defaultTargetOptions?.shouldOptimize ??
     mode === 'production';
 
-  let publicUrl = initialOptions?.defaultTargetOptions?.publicUrl ?? '/';
-  let distDir =
-    initialOptions?.defaultTargetOptions?.distDir != null
-      ? path.resolve(inputCwd, initialOptions?.defaultTargetOptions?.distDir)
-      : undefined;
-
   let shouldBuildLazily = initialOptions.shouldBuildLazily ?? false;
   let shouldContentHash =
     initialOptions.shouldContentHash ?? initialOptions.mode === 'production';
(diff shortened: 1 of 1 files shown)


devongovett:
You can't do something like this already?

```shell
parcel build --public-url $PUBLIC_URL
```

villermen:
@devongovett Where would I source `PUBLIC_URL` from? It's stored in the `.env` file, but using that in a single command that is cross-platform compatible has proven to be very difficult.

1. `source .env && parcel build --public-url $PUBLIC_URL` would not work because `$PUBLIC_URL` is interpreted before it is executed. It will not work outside of Linux systems either.
2. `source .env && bash - c 'parcel build --public-url \$PUBLIC_URL'` would probably work but would definitely not work outside of systems without bash support.
3. `PUBLIC_URL=/foo npm run build` would work but you would have to remember to set the envvar each time you build for a non-root domain.

I thought being able to leverage the already existing .env for this would be the better solution as it wouldn't introduce new concepts or possible room for user error while building. There might be a better solution out there that I'm just not seeing though.

Example for wanting this behavior is an application that configures `<base href={{ process.env.BASE_URL }}` because it can take an arbitrary amount of additional slugs. The Parcel references should still point to the location of the main index.html though.

alexplummersky:
Any updates on this? I really need the public url to be accessible from an env var also.

ppalmeida:
wow. Me as well. I've tried everything. I've tried many different solutions but none worked.

PR #8604, opened 2022-11-03
Adding missing dependencies to package.json files to clean up phantom dependencies

alshdavid:
#### Background 

Yarn workspaces flattens all dependencies declared in the workspace packages into the top level `node_modeules`. This, combined with Node's module resolution strategy, allows packages in the workspace to have access to a dependency declared in another package's package.json without declaring that dependency in their own package.json.

These are known as phantom dependencies and can cause friction when it comes the versioning of dependencies, unexpected breakages for consumers due to missing packages (that are available in the workspace but not when pulled from npm), can sometimes hide circular dependencies and cause various other annoyances.

#### CHANGELOG

- Identified the dependencies of each project and included those dependencies in its package.json.
- Identified packages that explicitly need hoisting and included them in the `.npmcrc`
  - We should probably correct the packages that require hoisting

#### Notes

I am not 100% sure what should be a `devDependency`, `peerDependency` and `dependency` as I lack the context on what is a library and what is a consumer. Please check my working in this regard


.npmrc
```diff
@@ -0,0 +1,12 @@
...
+public-hoist-pattern[]=@babel/plugin-proposal-nullish-coalescing-operator
+public-hoist-pattern[]=@babel/plugin-proposal-optional-chaining
+public-hoist-pattern[]=@babel/plugin-transform-modules-commonjs
+public-hoist-pattern[]=@babel/preset-env
+public-hoist-pattern[]=@babel/preset-flow
+public-hoist-pattern[]=@babel/preset-react
+public-hoist-pattern[]=babel-plugin-minify-dead-code-elimination
+public-hoist-pattern[]=babel-plugin-transform-inline-environment-variables
+public-hoist-pattern[]=read-pkg-up
+public-hoist-pattern[]=resolve
```
alshdavid, line 12: These are dependencies of the /dev/bable-preset which are required by the cli for transformations.

They need to be hoisted to the top level and made globally available due to the parcelBableRegister setting it's cwd to the project root

https://github.com/parcel-bundler/parcel/blob/alsh/rationalizing-phantom-dependencies/packages/dev/babel-register/index.js#L5

package.json
```diff
@@ -37,18 +37,20 @@
   },
   "devDependencies": {
     "@babel/core": "^7.12.0",
+    "@parcel/babel-register": "2.7.0",
+    "@parcel/babel-preset": "2.7.0",
```
alshdavid, line 41: These are also dependencies of /dev/bable-preset however due to them being workspaces packages, they cannot be explicitly hoisted in the `.npmrc` and instead referenced here.

As an alternative to the `.npmrc`, I can put those packages in the top level `package.json`

mischnic:
We also have this eslint rule which I would have expected to catch these undeclared dependencies (I removed nullthrows from the package.json):
```
parcel/packages/core/diagnostic/src/diagnostic.js
  4:24  error  'nullthrows' should be listed in the project's dependencies. Run 'npm i -S nullthrows' to add it  import/no-extraneous-dependencies
```

packages/core/test-utils/package.json
```diff
@@ -17,13 +17,21 @@
...
     "@parcel/package-manager": "2.7.0",
+    "@parcel/types": "2.7.0",
     "@parcel/utils": "2.7.0",
     "chalk": "^4.1.0",
     "ncp": "^2.0.0",
     "nullthrows": "^1.1.1",
     "posthtml": "^0.16.5",
     "posthtml-parser": "^0.10.1",
-    "resolve": "^1.12.0",
+    "resolve": "^1.22.1",
```
alshdavid: Updated this to the lasted minor version as there were issues with `linux-arm64`

packages/dev/eslint-config-browser/package.json
```diff
@@ -7,6 +7,6 @@
   },
   "peerDependencies": {
     "@babel/core": "^7.0.0",
-    "eslint": ">= 7.0.0"
+    "eslint": "^7.32.0"
```
alshdavid: Had to ensure `eslint` was the same version in all packages that use it due to it being a global workspace package.

packages/optimizers/image/package.json
```diff
@@ -39,6 +39,7 @@
     "detect-libc": "^1.0.3"
   },
   "devDependencies": {
+    "@types/node": "^12.12.6",
```
alshdavid: This should probably be added to packages that specifically need node types rather than relying on the global package on the top level

packages/transformers/elm/package.json
```diff
@@ -29,6 +29,9 @@
     "nullthrows": "^1.1.1",
     "terser": "^5.14.2"
   },
+  "devDependencies": {
+    "@parcel/types": "2.7.0"
+  },
   "peerDependencies": {
     "elm": "^0.19.1-5"
```
alshdavid: Worth noting that `elm` does not work on `linux-arm64` meaning Parcel cannot be built/developed on docker for MacOS, ARM Linux, Asahi, etc

https://github.com/elm/compiler/pull/2234

EDIT: Notified the elm team, they are addressing this

packages/utils/parcel-lsp/package.json
```diff
@@ -28,10 +28,10 @@
     "parcel": "^2.7.0"
   },
   "dependencies": {
-    "@parcel/diagnostic": "2.5.0",
-    "@parcel/plugin": "2.5.0",
-    "@parcel/types": "2.5.0",
-    "@parcel/utils": "2.5.0",
```
alshdavid: Were these intentionally set to use old versions of parcel?

If we are interested in syncing all packages in the repo against their declared workspace version, [pnpm](https://pnpm.io/workspaces) has phenomenal workspace support that avoids phantom dependencies and allows us to describe package dependencies with:
```json
"dependencies": {
    "@parcel/types": "workspace/*"
}
```
which automatically transforms the version to the version declared in the workspace package on package publishing.
```json
"dependencies": {
    "@parcel/types": "2.7.0"
}
```
If a dependency with a different version is required, it installs it locally to the package without creating conflicts with other packages

scripts/clean.bash
```diff
@@ -0,0 +1,174 @@
+#!/bin/bash
```
alshdavid, line 1: I normally just nuke all `node_modules` with `rm -rf **/node_modules` but we have a few we keep around under examples.

I introduced this script to make it easier to reinstall dependencies and clean the project

```bash
$ ./scripts/clean.bash --node_modules --rust --native --misc
$ ./scripts/clean.bash --all
```

pnpm-workspace.yaml
```diff
@@ -0,0 +1,2 @@
+packages:
+  - "packages/*/*"
```
alshdavid, line 2: Adds optional support for pnpm workspaces, this is how I found all the phantom dependencies.

I can remove this (or add it to a `./scripts/setup-pnpm.bash` script for those who want to opt-in) if it's too contentious though

parcel-benchmark:
## Benchmark Results

#### Timings

| --- | --- | --- |
| Cold | 1.42s | -7.00ms |
| Cached | 350.00ms | +5.00ms |

#### Cold Bundles

*No bundle changes detected.*

#### Cached Bundles

*No bundle changes detected.*

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 9.07s | -36.00ms |
| Cached | 431.00ms | +7.00ms |

#### Cold Bundles

*No bundle changes detected.*

#### Cached Bundles

*No bundle changes detected.*

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 1.59m | -1.17s |
| Cached | 2.13s | -38.00ms |

#### Cold Bundles

*No bundle changes detected.*

#### Cached Bundles

*No bundle changes detected.*

</p></details>

#### Timings

| --- | --- | --- |
| Cold | 6.50s | -79.00ms |
| Cached | 255.00ms | +5.00ms |

#### Cold Bundles

*No bundle changes detected.*

#### Cached Bundles

*No bundle changes detected.*

</p></details>

**[Click here to view a detailed benchmark overview.](https://benchmark.parceljs.org/benchmark/347806431374738003)**

package.json
```diff
@@ -57,7 +59,18 @@
...
+    "@babel/plugin-proposal-nullish-coalescing-operator": "^7.4.4",
+    "@babel/plugin-proposal-optional-chaining": "^7.2.0",
+    "@babel/plugin-transform-modules-commonjs": "^7.0.0",
+    "@babel/preset-env": "^7.1.0",
+    "@babel/preset-flow": "^7.0.0",
+    "@babel/preset-react": "^7.0.0",
+    "babel-plugin-minify-dead-code-elimination": "^0.5.0",
+    "babel-plugin-transform-inline-environment-variables": "^0.4.3",
+    "read-pkg-up": "^4.0.",
+    "resolve": "^1.22.1"
```
alshdavid: These are dependencies of the CLI when running in development mode. They are required by the cli for transformations. 

This lists them explicitly as dependencies that need to made available globally.

They are needed to be made available globally due to the parcelBableRegister setting it's cwd to the project root: https://github.com/parcel-bundler/parcel/blob/alsh/rationalizing-phantom-dependencies/packages/dev/babel-register/index.js#L5

Technically Yarn makes all dependencies global, but I just want to make sure they are explicitly listed as such

package.json
```diff
@@ -37,18 +37,20 @@
...
     "cross-env": "^7.0.0",
-    "eslint": "^7.20.0",
+    "eslint": "^7.32.0",
     "flow-bin": "0.184.0",
     "glob": "^7.1.6",
     "gulp": "^4.0.2",
     "gulp-babel": "^8.0.0",
     "husky": "^6.0.0",
     "lerna": "^3.22.1",
+    "@octokit/core": "3.6.0",
```
alshdavid: This is a peerDependency of lerna, adding this to suppress a warning

packages/runtimes/webextension/src/autoreload-bg.js
```diff
@@ -1,4 +1,4 @@
-/* global chrome, browser, addEventListener, fetch, Response, HMR_HOST, HMR_PORT */
+/* global chrome, browser, addEventListener, Response, HMR_HOST, HMR_PORT */
```
alshdavid: Linting error
```
1:46  error  'fetch' is already defined as a built-in global variable  no-redeclare
```

packages/core/utils/src/shared-buffer.js
```diff
@@ -1,5 +1,4 @@
 // @flow
-/* global MessageChannel:readonly */
```
alshdavid: Linting error
```
1:46  error  'MessageChannel' is already defined as a built-in global variable  no-redeclare
```

package.json
```diff
@@ -37,18 +37,21 @@
...
   "devDependencies": {
     "@babel/core": "^7.12.0",
+    "@parcel/eslint-config": "2.7.0",
+    "@parcel/babel-register": "2.7.0",
+    "@parcel/babel-preset": "2.7.0",
     "@khanacademy/flow-to-ts": "^0.5.2",
     "@napi-rs/cli": "^2.6.2",
-    "@parcel/babel-register": "2.7.0",
-    "@types/node": "^15.12.4",
+    "@types/node": "^12.20.55",
```
alshdavid: Making sure our node types are aligned with our engine support

package.json
```diff
@@ -57,7 +60,18 @@
...
+    "@babel/preset-env": "^7.1.0",
+    "@babel/preset-flow": "^7.0.0",
+    "@babel/preset-react": "^7.0.0",
+    "babel-plugin-minify-dead-code-elimination": "^0.5.0",
+    "babel-plugin-transform-inline-environment-variables": "^0.4.3",
+    "read-pkg-up": "^4.0.0",
+    "resolve": "^1.22.1"
   },
   "engines": {
     "node": ">= 12.0.0"
```
alshdavid, line 66: Should we consider moving to 18 given it's the current LTS?

alshdavid:
> @mischnic:  We also have this eslint rule which I would have expected to catch these undeclared dependencies (I removed nullthrows from the package.json)

In a previous project I worked on we also used an [eslint rule](https://www.npmjs.com/package/eslint-plugin-package-json-dependencies) to enforce packages explicitly declaring their dependencies and it worked pretty well. Also helped with blacklisting imports containing `/src/` and such.

We ended up moving to pnpm as, in workspaces, it structures `node_modules` in a way that ensures it's impossible to use a dependency without declaring it (each package has its own node_modules with symlinks to the actual packages - results in 0kb node_modules folders, instant&offline installs and super safe workspaces)

alshdavid:
Should I do a patch version bump on the packages that have been updated?

mischnic:
> Should I do a patch version bump on the packages that have been updated?

No, we do the version bumping as part of the release process.

> In a previous project I worked on we also used an [eslint rule](https://www.npmjs.com/package/eslint-plugin-package-json-dependencies) to enforce packages explicitly declaring their dependencies and it worked pretty well. Also helped with blacklisting imports containing /src/ and such.

It'd also say that it worked pretty well for Parcel. But apparently it's not working for intra-workspace packges (which the current eslint rule apparently silently allows)

> We ended up moving to pnpm

I guess that's a separate discussion. Detecting these undeclared dependencies statically in a linter seems better to me than relying on the import call to fail at runtime/in the tests.

> Should we consider moving to 18 given it's the current LTS?

For the unpublished (dev) packages in the monorepo it doesn't matter much but Parcel does support Node >= 12 for now, so changing that would be a breaking change.

> These are [Babel] dependencies of the CLI when running in development mode. They are required by the cli for transformations.

I had thought that Babel resolves plugins declared in presets relative to that preset, and they are declared over there: https://github.com/parcel-bundler/parcel/tree/v2/packages/dev/babel-preset
Otherwise you'd have this problem with every Babel preset from npm as well?

> Were these intentionally set to use old versions of parcel?

(`parcel-lsp`) No, but for some reason this step didn't bump the version on the last release
https://github.com/parcel-bundler/parcel/blob/932b6bd702b0e6af573713840cbb47cd3afd6e17/package.json#L33

> `"@parcel/types": "workspace/*"`

Is there some mechanism to replace that with the actual version on publish? Or is that only meant for private packages?

package.json
mischnic: For the unpublished (dev) packages in the monorepo it doesn't matter much but Parcel does support Node >= 12 for now, so changing that would be a breaking change.

package.json
mischnic: I had thought that Babel resolves plugins declared in presets relative to that preset, and they are declared over there: https://github.com/parcel-bundler/parcel/tree/v2/packages/dev/babel-preset
Otherwise you'd have this problem with every Babel preset from npm as well?

packages/utils/parcel-lsp/package.json
mischnic: > Were these intentionally set to use old versions of parcel?

No, but for some reason this step didn't bump the version on the last release
https://github.com/parcel-bundler/parcel/blob/932b6bd702b0e6af573713840cbb47cd3afd6e17/package.json#L33

> `"@parcel/types": "workspace/*"`

Is there some mechanism to replace that with the actual version on publish? Or is that only meant for private packages?

mischnic:
> Should I do a patch version bump on the packages that have been updated?

No, we do the version bumping as part of the release process.

> In a previous project I worked on we also used an [eslint rule](https://www.npmjs.com/package/eslint-plugin-package-json-dependencies) to enforce packages explicitly declaring their dependencies and it worked pretty well. Also helped with blacklisting imports containing /src/ and such.

It'd also say that it worked pretty well for Parcel so far. But apparently it's not working for intra-workspace packges (which the current eslint rule apparently silently allows)

> We ended up moving to pnpm

I guess that's a separate discussion. Detecting these undeclared dependencies statically in a linter seems better to me than relying on the import call to fail at runtime/in the tests.

packages/utils/parcel-lsp/package.json
alshdavid: > Is there some mechanism to replace that with the actual version on publish? Or is that only meant for private packages?

It's a mechanism to replace the version with the actual version when running `pnpm pack` or `pnpm publish` - it works for public or private packages as it's just string replacement when publishing.

Reduces the maintenance burden involved in tracking the versions of packages in the workspace and avoids us needing to commit absolute versions in the source repo which has caused me so much grief in the past 🤣

To be clear, this is a [feature of pnpm workspaces](https://pnpm.io/workspaces#publishing-workspace-packages) and I don't believe it's available in lerna or yarn - so it's currently unavailable to us, but is that something we'd be open to discussing?

package.json
alshdavid: Is this the engine supported for contributors developing parcel or for consumers of Parcel?

mischnic: This is the one for contributors.

package.json
alshdavid: You're right, I spent some time trying to understand where this was coming from and it turns out we use `@parcel/babel-preset` from the root level in the `babel.config.json`. So when I used strict workspace package isolation, babel would complain that it couldn't find our preset.

Turns out I just needed to add `@parcel/babel-preset` to the workspace `package.json`.

Removed the unused extras

packages/optimizers/image/src/ImageOptimizer.js
```diff
@@ -1,6 +1,5 @@
 // @flow
 import path from 'path';
-import process from 'process';
```
alshdavid: Flow error `Cannot resolve module 'process'.`

package.json
```diff
@@ -37,6 +37,9 @@
   },
   "devDependencies": {
     "@babel/core": "^7.12.0",
+    "@octokit/core": "3.6.0",
+    "@parcel/eslint-config": "2.7.0",
+    "@parcel/babel-preset": "2.7.0",
```
alshdavid: `@octokit/core` is an unmet peer dependency of lerna it was complaining about
`@parcel/eslint-config` Is used by `./.eslintrc.json`
`@parcel/babel-preset` Is used by `./babel.config.json`

packages/dev/babel-preset/index.js
```diff
@@ -55,7 +56,7 @@ module.exports = api => {
               ],
             },
           ],
-          'babel-plugin-minify-dead-code-elimination',
+          require('babel-plugin-minify-dead-code-elimination'),
```
alshdavid: These were causing resolution issues when strict/isolated workspace packages were enabled. They now resolve relative to this package and will always use the dependencies of this package first

package.json
alshdavid: That's cool, thanks

Issue #7934, opened 2022-04-10
MDX Plugin out of date


calebegg:
# 🐛 bug report

@parcel/transformer-mdx is out of date (doesn't support MDX 2.1.1)

devongovett:
See #7922

legobeat:
This was "closed as completed" but I assume this was incorrect? 

@alshdavid

TechQuery:
You can use my plugin for MDX 3: https://github.com/EasyWebApp/Parcel-transformer-MDX

TechQuery:
I'm improving Parcel MDX 3 plugin with MDX official team: mdx-js/mdx#2523

PR #8625, opened 2022-11-12
Add new transformer for MDX v2

Ayc0:
# ↪️ Pull Request

Other variant for https://github.com/parcel-bundler/parcel/pull/7922 to solve https://github.com/parcel-bundler/parcel/issues/7934.
The main differences are:
- we can have some custom config for MDX,
- pushed a new package `@parcel/transformer-mdx-2` instead of modifying the existing.

For the MDX config, you can add a `.mdxrc.js` or `.mdxrc.cjs` file at the root of your repo to be able to customise MDX's behavior (see https://mdxjs.com/packages/mdx/#optionsremarkplugins).

Note: I tried to build parcel locally but `yarn build-native` gets stuck. So i couldn't test this PR locally

Note 2: to make it work, we also need `@parcel/transformer-js` as MDX v2 generates ESM code:

```json
{
  "transformers": {
    "*.mdx": [ "parcel-transformer-mdx-2", "@parcel/transformer-js" ]
  }
}
```

## 💻 Examples

Previous MDX examples should still be applicable.

I also added a example here with some MDX config too: https://github.com/Ayc0/test-parcel/blob/7d05ca9/src/hello.mdx

## 🚨 Test instructions

Previous MDX tests should still be applicable.
But I also added a new one (not really sure if this is used).

I also created a repo here: https://github.com/Ayc0/test-parcel/blob/7d05ca9/src/hello.mdx

Notice the MDX 2 specific syntaxes like `{a}` or the MD in HTML, and that the strikethrough and the table work (provided by GFM):

Input | Config | Output
-|-|-
image|image| image

## ✔️ PR Todo

### DIFF
--- a/packages/transformers/mdx-2/src/MDXTransformer.js
+++ b/packages/transformers/mdx-2/src/MDXTransformer.js
@@ -1,0 +1,28 @@
+// @flow
+import {Transformer} from '@parcel/plugin';
+
+export default (new Transformer({
+  async loadConfig({config}) {
+    let configFile = await config.getConfig(['.mdxrc.js', '.mdxrc.cjs'], {
+      packageKey: 'mdx',
+    });
+
+    if (configFile) {
+      config.invalidateOnStartup();
+
+      return configFile.contents;
+    }
+  },
+  async transform({asset, config}) {
+    let code = await asset.getCode();
+    // @mdx-js/mdx is a ESM package, so we need to use `import()` to make it work in CJS envs
+    const {compile} = await import('@mdx-js/mdx');
+
+    let compiled = await compile(code, config);
+
+    asset.type = 'js';
+    asset.setCode(compiled.value);
+
+    return [asset];
+  },
+}): Transformer);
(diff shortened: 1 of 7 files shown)


Ayc0:
On rebuild, parcel crashes with this error:

image

This might be a bug within Parcel: the `import` resolves the first time:

image

But after a recompilation, it crashes.

Edit: it only happens in parcel 2.8.0, in parcel 2.7.0 it works fine:

https://user-images.githubusercontent.com/22725671/201500024-dbadf894-97d4-4d43-9fd4-8c594b4d5a07.mov

devongovett:
That is caused by v8-compile-cache I believe...

mischnic:
https://github.com/zertosh/v8-compile-cache/issues/30

packages/transformers/mdx-2/src/MDXTransformer.js
```diff
@@ -0,0 +1,41 @@
+const {Transformer} = require('@parcel/plugin');
+
+module.exports = new Transformer({
+  async loadConfig({config}) {
+    // config.getConfig only works with CJS
+    let configFile = await config.getConfig(['.mdxrc.js', '.mdxrc.cjs'], {
```
Ayc0, line 6: To follow https://github.com/mdx-js/mdx/blob/v1/packages/parcel-plugin-mdx/src/MDXAsset.js#L13, I could add `mdx.config.js` and `mdx.config.cjs`.

Or just use those 2 instead of `.mdxrc.js` (as usually the rc files are JSON files, not JS files)

PR #8641, opened 2022-11-17
Fix resolution of absolute paths inside project root

pastelsky:
# ↪️ Pull Request

When `NodeResolver` is trying to resolve an absolute path inside the parcel's project directory, it incorrectly resolves the path.

## 💻 Examples

For example, resolving `/projectDir/someFolder/index.js` when the project dir is `/projectDir`,
 `NodeResolver` will resolve the file to — `/projectDir/projectDir/someFolder/index.js` which is incorrect.

Instead, it should treat absolute paths inside the project directory differently, and perhaps resolve them without duplicating the project directory.

## 🚨 Test instructions

See test case added for an example.

## ✔️ PR Todo

### DIFF
--- a/packages/utils/node-resolver-core/src/NodeResolver.js
+++ b/packages/utils/node-resolver-core/src/NodeResolver.js
@@ -583,7 +583,9 @@ export default class NodeResolver {
 
         // Absolute path. Resolve relative to project root.
         dir = this.projectRoot;
-        filename = '.' + filename;
+        filename = filename.startsWith(dir)
+          ? path.relative(dir, filename)
+          : '.' + filename;
         break;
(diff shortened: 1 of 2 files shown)


lettertwo:
👋 Hi @pastelsky, could share a bit more about your use case here? I'm curious to know if you could just change your specifier to be absolute-ish without the root directory name, e.g., `/someFolder/index.js` instead of `/projectDir/someFolder/index.js`.

mischnic:
(So to complement @lettertwo's comment: the current behaviour is correct and working as expected: https://parceljs.org/features/dependency-resolution/#absolute-specifiers)

PR #8648, opened 2022-11-21
Fix dynamic import of modules with inlined imports

comp500:
# ↪️ Pull Request
When dynamically importing a module that itself has an inlined import (e.g. with bundle-text) a module loader is added for the inlined bundle, which references a file that is not generated (since it is inlined). This invalid reference causes the import to fail at runtime if the fetch fails before the correct import resolves; it appears to also cause a TypeError ("Cannot read properties of undefined") in tests.

This fix simply excludes external bundles in getLoaderRuntime that have a bundleBehavior of inline.

## 💻 Examples
From the unit test code:

`asyncTransitive.js`:
```js
export default import("./inline.js");
```
`inline.js`:
```js
import otherHTML from "bundle-text:./inline.html"
export default otherHTML;
```
`inline.html`:
```html
<p>test</p>
<script>console.log('hi')</script>
```

`asyncTransitive.js` is used as an entry point; generates 2 invalid references:

```js
// ...
module.exports = Promise.all([
    require("./helpers/browser/js-loader")(require("./helpers/bundle-url").getBundleURL("gfgQn") + "inline.4243f809.js" + "?" + Date.now()).catch((err)=>{ /* ... */ }),
    require("./helpers/browser/html-loader")(require("./helpers/bundle-url").getBundleURL("gfgQn") + "inline.0dc63d1b.html" + "?" + Date.now()).catch((err)=>{ /* ... */ }}),
    require("./helpers/browser/js-loader")(require("./helpers/bundle-url").getBundleURL("gfgQn") + "inline.03c983ec.js" + "?" + Date.now()).catch((err)=>{ /* ... */ })
]).then(()=>module.bundle.root("5L8Sl"));
// ...
```
(`inline.4243f809.js` and `inline.0dc63d1b.html` don't exist)

## 🚨 Test instructions
N/A (unit test added)

## ✔️ PR Todo


oe:
Confirmed issue! Nice mr!

This mr can fix error like this(caused by invalid references)
CleanShot 2022-12-09 at 22 49 24

but still not working when inlined module is a web worker, here is a demo <https://stackblitz.com/edit/github-fshd7k-xtd4gl?file=src/app.js> can reproduce this issue:
1.  download project (stackblitz not working online due to lack of supporting of native node modules)
CleanShot 2022-12-09 at 23 09 21

3. install and run project `yarn && yarn dev`
4. open browser and open devtools, check logs in console

comp500:
I can't seem to reproduce that issue with my change, I just get the compiled bundle of worker.js in the log: 
image

The worker itself doesn't run in development because Parcel doesn't know that the module is intended to be a web worker (as you're not using the intended syntax) so it emits the HMR runtime code, which isn't compatible with web workers. Running in production with the files produced from `parcel build` works fine.

PR #8651, opened 2022-11-21
Add support for babel transform in monorepos

jhsware:
If you have a monorepos with the following setup:
```
root/
  packages/
    app/
      package.json
      App.tsx
    component/
      package.json
      Component.tsx (requires babel transpile)
```

- `app` depends on `component`
- `app` and `component` need to be transpiled with Babeljs

**You are currently limited** to running Parceljs from a common parent folder with a shared Babeljs config.

However there are several cases where this might be undesirable. Especially if you have multiple application projects sharing the same development packages that are under continuous development.

This limitation doesn't exist when using Parceljs without requiring Babel transpile of development packages.

## Background
I ran into this problem when using Parceljs in an [Infernojs-project](https://www.infernojs.org/). When using Reactjs I could use this project structure. However, since Infernojs requires [babel-plugin-inferno](https://github.com/infernojs/babel-plugin-inferno) to create optimised transpiled code, Parceljs fails and transpiles the JSX-code of development packages to React.createElement calls.

## The Problem
What happens is that Parceljs won't pass projectRoot to babel when the asset being transpiled is outside projectRoot. This causes babel to fail to pick up a babel config file and thus skipping whatever transpilation is required by that package.

## The Solution
This PR checks if the asset to be transpiled is actually inside the project root. If not it checks for a package root (closest ancestor folder containing package.json) and passes this as current working directory to Babeljs. This allows babel to find the package specific babel config file and allows the asset to be transpiled properly.

## Tests
I have implemented tests for .js, .jsx, .ts and .tsx based on the babel-custom integration test.

### DIFF
--- a/packages/core/integration-tests/test/babel.js
+++ b/packages/core/integration-tests/test/babel.js
@@ -526,6 +526,69 @@ describe('babel', function () {
     assert.strictEqual(output.default, 123);
   });
 
+
+  it('should compile with different babel plugins in monorepos using js', async function () {
+    let b = await bundle(
+      path.join(__dirname, '/integration/babel-monorepos/packages/app/index.js'),
+    );
+
+    let file = await outputFS.readFile(b.getBundles()[0].filePath, 'utf8');
+    assert(!file.includes('class Test'));
+    assert(!file.includes('REPLACE_ME'));
+    assert(!file.includes('#private'));
+    assert(file.includes('---app---'));
+    assert(file.includes('---component---'));
+
+    let output = await run(b);
+    assert.strictEqual(typeof output, 'object');
+    assert.strictEqual(output.default, '---app---:---component---');
+  });
+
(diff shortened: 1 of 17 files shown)


mischnic:
Babel already has a mechanism for monorepos and package-specific overrides: https://babeljs.io/docs/en/config-files#monorepos.

An example from the Parcel test cases: https://github.com/parcel-bundler/parcel/tree/v2/packages/core/integration-tests/test/integration/babel-config-monorepo

Have you tried that already?

jhsware:
@mischnic I totally missed that even though I saw the test case. I will check it out, thanks for the links!

jhsware:
@mischnic From what I can tell it is working both with development server and build. Thank you for pushing me in the right direction! I tried a similar setup earlier but failed. I may have had caching issues or just bad luck... for what it is worth, it was interesting poking around in the project :)

PR #8654, opened 2022-11-23
Fix * re-exports from empty files

mattcompiles:
# ↪️ Pull Request

In some cases where a `*` export points to an empty file e.g. `export * from './file-with-no-contents'` Parcel incorrectly assign symbols to the file. This seems to be due to a [default `*` symbol assigned to these types of files](https://github.com/parcel-bundler/parcel/blob/711ae183f7b61be36128e9da0e424d874d836015/packages/transformers/js/src/JSTransformer.js#L821). While it's strange to export from an empty file, it can often occur in packages that don't remove their type exports from the runtime code. 

My fix is to pass a new `is_empty_file` value back from the native Js transformer which is set to true when a file has no exports, imports or declarations. 

## 🚨 Test instructions

I have replicated this issue but it doesn't seem to fail in the integration test suite.

### DIFF
--- a/packages/transformers/js/core/src/lib.rs
+++ b/packages/transformers/js/core/src/lib.rs
@@ -108,6 +108,7 @@ pub struct TransformResult {
   needs_esm_helpers: bool,
   used_env: HashSet<swc_atoms::JsWord>,
   has_node_replacements: bool,
+  is_empty_file: bool,
 }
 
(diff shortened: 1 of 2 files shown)


mattcompiles:
Closing this for now due to the issue being more complex than it first seemed.

PR #7918, opened 2022-04-07
Fix: join srcset in source with ", "

Shinyaigeek:
# ↪️ Pull Request

Fixes: #7884 

I found this is parcel’s issue. htmlnano depends on [srcset](https://github.com/sindresorhus/srcset), npm module to parse srcset attribute and after https://github.com/sindresorhus/srcset/pull/11 PR was merged `srcset` **module** cannot parse srcset **attribute** without whitespace after “,” because `hoge.png,bar.png` is a valid URL.

However, parsel joins multiple image URLs in srcset **attribute** with “,” so this issue happens. parcel should join multiple image URLs in srcset **attribute** with “, “(with space), so I fixed so.

## 💻 Examples

```html
<source srcset="hoge.png?as=webp&width=400, hoge.png?as=webp&width=800 2x">
```

into transformed into above by HTMLTransformer

```html
<source srcset="hoge.HASH.png,hoge.HASH.png 2x">
```

but `srcset` module cannot parse this srcset **attribute** as a two image URLs.

## 🚨 Test instructions

I fixed html srcset integration testcase to check output HTML's srcset is valid.

## ✔️ PR Todo

- [x] Added/updated unit tests for this change
- [x] Filled out test instructions (In case there aren't any unit tests)
- [x] Included links to related issues/PRs

PR #8671, opened 2022-12-03
fix srcset parsing

christianhaller:
# ↪️ Pull Request

When I was using some URLs with a comma separated query param to a srcset,
my parcel tried to resolve the wrong url.
It was splitting the srcset string on the wrong comma.

`@parcel/resolver-default: Cannot load file './9.739075709034898&amp;markers=size:tiny|52.345519011208346' in 
'./src/content`

I found a package which is parsing the srcset correctly.
I added a test for the affected method.


mischnic:
We don't really use unit tests, so please create an integration test instead (see https://github.com/parcel-bundler/parcel/pull/7918).

mischnic:
- Please remove packages/transformers/html/test/dependencies.test.js (and unexport `collectSrcSetDependencies`)
- We might as well leverage the `stringify` function of that package now, instead of the manual `w`/`x` appending, e.g. like this:
```js
import * as srcset from 'srcset';

function collectSrcSetDependencies(asset, v, opts) {
  let parsed = srcset.parse(v).map(({url, ...v}) => ({
    url: asset.addURLDependency(url, opts),
    ...v,
  }));
  return srcset.stringify(parsed);
}
```
- Please add this file at `flow-libs/srcset.js.flow`:
```js
// @flow
declare module 'srcset' {
  declare type SrcSetDefinition = {|
    url: string,
    width?: number,
    density?: number,
  |};

  declare type Options = {|
    strict?: boolean,
  |};

  declare export function parse(
    srcset: string,
    options?: Options,
  ): SrcSetDefinition[];
  declare export function stringify(
    srcSetDefinitions: SrcSetDefinition[],
    options?: Options,
  ): string;
}

```

christianhaller:
Thank you for the review Niklas!
Good point with the stringify method. This looks much better :-)

mischnic:
Looks like you have some npm registry override configured for Yarn. There shouldn't be that many changed lines in the lockfile

christianhaller:
> Looks like you have some npm registry override configured for Yarn. There shouldn't be that many changed lines in the lockfile

It took me hours to fix this :dizzy_face:

mischnic (approved):
Thanks!

PR #8688, opened 2022-12-12
add listUnusedSymbols

AGawrys:
# ↪️ Pull Request

Basic function for listing unused symbols on all assets of a project.

## ✔️ PR Todo

- [ ] Should we have this on the BundleGraph? Or Both ? 
- [ ] What would unused symbols for dependencies of the project look like
- [ ] Should we have some notion of "these symbols are unused across the **whole** **project**" ?


packages/dev/query/src/cli.js
```diff
@@ -366,6 +367,22 @@ function traverseBundle(v: string) {
...
+function listUnusedSymbols() {
+  for (let n of assetGraph.nodes.values()) {
+    if (n.type === 'asset') {
+      if (n.usedSymbols.size === 0) continue;
+      invariant(n.value.symbols != null);
+      let allSymbols = new Set([...n.value.symbols?.keys()]);
+      let unusedSymbols = setDifference(n.usedSymbols, allSymbols);
+      if (unusedSymbols.size === 0) continue;
+      console.log(
+        fromProjectPathRelative(n.value.filePath) + ' Does not use symbols: ',
```
mischnic: This is what you're computing here. (Not sure if this is from prototyping or if there was something unclear about the meaning of `unusedSymbols`)
```suggestion
        fromProjectPathRelative(n.value.filePath) + ' contains these unused exports: ',
```

mischnic:
> Should we have this on the BundleGraph? Or Both ?

The difference between the two is
- `if(assetGraph.isDependencySkipped(dependency))) { skipChildren(); return; }` and
- and whatever my recent dependency rewriting PR does


If you want to do source-level "linting" anyway, then the asset graph makes sense. Otherwise, you might miss some side-effect-free assets that are completely unused and therefore removed from the graph due to one of the two step above.

> What would unused symbols for dependencies of the project look like

As in node_modules? There's really not much the user can do about this though?
Like `unused symbols in node_modules/lodash/index.js: sum, groupBy, ...`
And if everything works as it should, these will be removed by Terser anyway.

> Should we have some notion of "these symbols are unused across the whole project" ?

(That's already what's stored in `(dep|asset).usedSymbols`. Unless I'm misunderstanding something)

Issue #8588, opened 2022-10-31
High level profiling / tracing


marcins:
# 💬 RFC

## 🔦 Context

Parcel currently provides a way to generate sampling CPU profiles using `--profile`, or using Ctrl-E during the build. These CPU sample based profiles can be used to identify bottlenecks and understand where Parcel is spending it's time at the code level, and are useful for these low-level purposes.

This proposal is to introduce the concept of a high-level trace, where Parcel can report on time spent at a higher level - for example, to be able to identify which transform is the most expensive in your build, or identify if there is a particular source file that causes additional overhead.

This is particularly useful for us internally when determining where to focus our efforts on build time improvements, in order to identify where time is being spent and where we should focus our optimisation efforts.

## Prior art

When I initially spiked this, @mischnic mentioned this had been previously raised by him in the Parcel project management tool (issue T-1131).

This RFC builds upon work originally created by @thebriando in our internal Parcel fork - Brian's implementation only ran on the main thread, and measured time spent in transform/bundling/packaging at a high level, however when I tried to adapt this approach to a more broad high level tracing (e.g. per transform) I ran into the following issues: a) passing a `Tracer` around through the various layers would involve a lot of changes in a lot of places, b) it would not be able to cross thread/serialisation boundaries.

## Spike approach

There is a very rough spike branch here: https://github.com/marcins/parcel/tree/mszczepanski/spike-system-tracing which at this stage is not comprehensive, nor particularly clean, but was created in order to help socialise the potential approach.

In this branch I have added additional "trace" functionality to the logger, mostly because I wanted to follow a similar approach - have a singleton `SystemTracer` that could be used to access the measurement functionality from anywhere. This `SystemTracer` lets you measure any code within Parcel with a pattern like:

```
 let measurement = SystemTracer.createMeasurement(transformer.name, {
  categories: ["transform"],
  args: {
    name: fromProjectPathRelative(initialAsset.value.filePath),
  },
});

// .. transformation happens here

measurement.end();
```

These trace events are caught by a TraceReporter which uses `chrome-trace-event` to produce an appropriate output.

This will produce a Chrome Trace event similar to:
```
{"ts":2726640,"pid":34523,"tid":3,"ph":"X","name":"@parcel/transformer-babel","cat":"transform","args":{"name":"node_modules/react/jsx-runtime.js"},"dur":288},
```

[...]

PR #6566, opened 2021-07-07
Remove 'Name already registered with serializer' error

devongovett:
Fixes #4064.

This issue seems to come up a lot because npm/yarn don't always deduplicate packages properly. It should be safe to not error because the version is part of the key being registered. The only case where it wouldn't work is if we relied on `instanceof`, but I didn't find any places where that's the case. So let's try just removing this error and see where it takes us. 🤷

PR #8695, opened 2022-12-16
Application profiling

marcins:
# ↪️ Pull Request

This is an implementation of the Tracing RFC - more detail on the background and motivations is in #8588. The main differences from the RFC is that the feature is called "application profiling" (and the existing CPU profiling is referred to as "Sampling Profiling").

Instrumentation has been added for:
* high-level phases - building, bundling, packaging
* transform (per-plugin)
* Babel transform (per-babel plugin visitor)
* bundling (bundle / optimize)
* packaging (package / optimize)
* other plugin types (resolving, naming, etc..)

## 🚨 Test instructions

An app profile can be generated by starting Parcel with `--profile-application`.

## ✔️ PR Todo

- [x] Inject the profiler for plugins
- [x] Create a low-overhead instrumentation method

See also documentation PR: https://github.com/parcel-bundler/website/pull/1075


packages/transformers/babel/src/babel7.js
```diff
@@ -8,6 +8,8 @@ import {relativeUrl} from '@parcel/utils';
 import {remapAstLocations} from './remapAstLocations';
 
 import packageJson from '../package.json';
+import {applicationProfiler} from '@parcel/profiler';
```
mischnic: I think it be better to have this be provided by core (so "dependency injection" similar to the `logger` right now).

This would break if Yarn doesn't hoist the package and then the calls `applicationProfiler.onTrace()` and `applicationProfiler.createMeasurement` would actually go to different instances.

We had this problem in the past: https://github.com/parcel-bundler/parcel/pull/6566

packages/core/core/src/PackagerRunner.js
```diff
@@ -394,7 +395,15 @@ export default class PackagerRunner {
 
     let packager = await this.config.getPackager(bundle.name);
     let {name, resolveFrom, plugin} = packager;
+    let measurement;
     try {
+      measurement = applicationProfiler.createMeasurement(name, {
```
lettertwo: Should this be gated by a check of `applicationProfiler.enabled`? Alternatively, could `applicationProfiler` expose category-specific methods that take positional args (instead of requiring object allocations) that are no-op by default?

packages/core/core/src/PackagerRunner.js
```diff
@@ -495,7 +505,14 @@ export default class PackagerRunner {
     };
 
     for (let optimizer of optimizers) {
+      let measurement;
       try {
+        measurement = applicationProfiler.createMeasurement(optimizer.name, {
```
lettertwo: Similarly, should this be gated or otherwise made to avoid the object allocations?

packages/core/core/src/Transformation.js
```diff
@@ -435,6 +436,16 @@ export default class Transformation {
     let resultingAssets = [];
     let finalAssets = [];
     for (let transformer of pipeline.transformers) {
+      let measurement = applicationProfiler.createMeasurement(
```
lettertwo: Here also? I am thinking that in the default case (profiling is not enabled), calls to `createMeasurement` should be a no-op, including skipping the allocation of new argument bags. I'm aware that this feels a bit like premature optimization; reasonable folks can disagree.

packages/core/core/src/requests/PathRequest.js
```diff
@@ -194,7 +195,14 @@ export class ResolverRunner {
     let invalidateOnFileChange = [];
     let invalidateOnEnvChange = [];
     for (let resolver of resolvers) {
+      let measurement;
       try {
+        measurement = applicationProfiler.createMeasurement(resolver.name, {
```
lettertwo: Flagging this one as part of the same series of comments. (not sure if i'm gonna catch em all :pokeball:)

packages/core/profiler/src/ApplicationProfiler.js
```diff
@@ -0,0 +1,82 @@
...
+    } finally {
+      measurement.end();
+    }
+  }
+
+  createMeasurement(
+    name: string,
+    data?: ApplicationProfilerMeasurementData = {categories: ['Core']},
+  ): ApplicationProfilerMeasurement {
+    const start = performance.now();
```
lettertwo: Should this method just no-op if `this.enabled === false`? Alternatively, call sites could be gated by that same check, which isn't a bad idea given that it could avoid allocating the data objects if profiling is disabled.

Another idea (mentioned in my first comment) could be to add category-specific methods that take positional arguments, which could allow call sites to avoid creating the complex `data` object. Those methods could just no-op by default, or build the args and call `createMeasurement` otherwise.

lettertwo (changes requested):

packages/transformers/babel/src/babel7.js
marcins: Yep, that's fair enough - you mean just where it's used in plugins as opposed to core? 

FWIW the initial implementation that Brian did internally passed around a "tracer" through core as well - this worked okay for the top level metrics, but once you wanted to instrument anything deeper it was a real pain to thread it through all the various layers.

packages/core/core/src/PackagerRunner.js
marcins: Yeah, good idea - given most of the calls follow the same pattern with just name, category and the filename changing can create a function for that use case with 3 string args that exits early if disabled (or as you say, change all of these to `applicationProfiler.enabled && applicationProfiler.createMeasurement(..)` - that would make `measurement` falsy which means later the `measurement && measurement.end()` would work.

packages/core/core/src/Transformation.js
marcins: I haven't done any significant perf measurement, but given most "prod" builds would be running with the profiler disabled avoiding unnecessary allocations / GC overhead seems like a reasonable idea.

packages/core/core/src/requests/PathRequest.js
marcins: I'll just search the code for all the `createMeasurement()` calls and address this.

packages/core/core/src/requests/ParcelBuildRequest.js
```diff
@@ -89,12 +94,14 @@ async function run({input, api, options}: RunInput) {
     ),
   });
 
+  let packaingMeasurement = applicationProfiler.createMeasurement('packaging');
```
mattcompiles:
```suggestion
  let packagingMeasurement = applicationProfiler.createMeasurement('packaging');
```

packages/core/core/src/requests/ParcelBuildRequest.js
```diff
@@ -89,12 +94,14 @@ async function run({input, api, options}: RunInput) {
...
   let writeBundlesRequest = createWriteBundlesRequest({
     bundleGraph,
     optionsRef,
   });
 
   let bundleInfo = await api.runRequest(writeBundlesRequest);
+  packaingMeasurement.end();
```
mattcompiles:
```suggestion
  packagingMeasurement.end();
```

packages/transformers/babel/src/babel7.js
mischnic: It's only a problem if some package X has state (such as that profiler) and X is declared as a dependency in multiple packages. 
So just importing it inside core would be fine (as there's only one dependency and no way Yarn creates two instances), though if it's only used inside of core anyway then it could maybe also just be separate file in core (as opposed to a separate package).

mischnic: > you mean [using dependency injection] just where it's used in plugins as opposed to core?

Yes, in the public API

marcins:
I've finally gotten to addressing the PR feedback from late last year. The main changes have been:
* rebasing on latest v2 branch
* “injecting” a `PluginApplicationProfiler` similar to `PluginLogger` for plugins to ensure the same instance is used
instrumenting the remaining plugin types in core

What’s still missing:
* adding some test coverage to the new functionality

packages/core/types/index.js
```diff
@@ -1931,3 +1976,31 @@ export interface IDisposable {
 export type AsyncSubscription = {|
   unsubscribe(): Promise<mixed>,
 |};
+
+export interface PluginApplicationProfiler {
+  /** Returns whether the application profiler is enabled. Use this to avoid possibly expensive calculations
+   * of arguments to `createMeasurement` - for example if you need to determine the entry of a bundle to pass it
+   * in as the <code>argumentName</code>, you would only do this if the application profiler is enabled.
+   */
+  get enabled(): boolean;
```
mischnic: Nit: for all the other public types, we've been using a readonly property (so `+enabled: boolean;`). For example `bundle.needsStableName` is in reality a getter in `core/src/public/Bundle.js`

packages/core/types/index.js
```diff
@@ -1931,3 +1976,31 @@ export interface IDisposable {
...
+   * function being executed (for example, the name of a Babel transform). The category will default to the name of your plugin,
+   * however it should be set to reflect the type of operation (for example, for a hypothetical operation
+   * to find CSS in an asset within a Compiled plugin you might set this to <code>find_css<code>).
+   *
+   * If this is an operation that executes multiple times on different things - whether that's assets, bundles, or
+   * otherwise - specify the name of the context object in <code>argumentName</code>.
+   *
+   * <code>otherArgs</code> can be used for specifying any other key/value pairs that should be written to the profile.
+   *
+   * For example: <code>profiler.createMeasurement('compiled', 'find_css', fromProjectPathRelative(asset.filePath), { meta: 'data' })</code>
```
mischnic: `fromProjectPathRelative` is an internal API. So either `asset.filePath` or `path.relative(options.projectRoot, asset.filePath)`

packages/reporters/application-profiler/src/ApplicationProfilerReporter.js
```diff
@@ -0,0 +1,85 @@
+// @flow
+import invariant from 'assert';
+import nullthrows from 'nullthrows';
+import {Reporter} from '@parcel/plugin';
+import {Tracer} from 'chrome-trace-event';
+
+// We need to maintain some state here to ensure we write to the same output, there should only be one
```
mischnic: Sidenote: This is exactly why `options.instanceId` was added. But in reality, this probably never happens anyway

packages/transformers/babel/src/babel7.js
```diff
@@ -66,6 +79,27 @@ export default async function babel7(
...
 
+  if (applicationProfiler.enabled) {
+    config.wrapPluginVisitorMethod = (
+      key: string,
+      nodeType: string,
+      fn: Function,
+    ) => {
+      return function () {
+        let measurement;
+        if (applicationProfiler.enabled) {
```
mischnic: Can the profiler be disabled mid-build or is the inner check unnecessary?

packages/core/parcel/src/cli.js
```diff
@@ -464,6 +465,13 @@ async function normalizeOptions(
     })),
   ];
 
+  if (command.profileApplication) {
+    additionalReporters.unshift({
+      packageName: '@parcel/reporter-application-profiler',
+      resolveFrom: __filename,
```
mischnic, line 469: This means that `@parcel/reporter-application-profiler` needs to be a dependency of the `parcel` CLI.

packages/core/core/src/requests/PathRequest.js
```diff
@@ -375,17 +387,7 @@ export class ResolverRunner {
...
-        // the potential for lazy require() that aren't executed until the request runs.
-        let devDepRequest = await createDevDependency(
-          {
-            specifier: resolver.name,
-            resolveFrom: resolver.resolveFrom,
-          },
-          this.previousDevDeps,
-          this.options,
-        );
-        this.runDevDepRequest(devDepRequest);
```
mischnic: Looks like an error during rebasing

packages/core/parcel/package.json
```diff
@@ -14,7 +14,7 @@
     "type": "git",
     "url": "https://github.com/parcel-bundler/parcel.git"
   },
-  "bin": "src/bin.js",
+  "bin": "lib/bin.js",
```
mischnic:
```suggestion
  "bin": "src/bin.js",
```

marcins: Oops, yes, good catch - I accidentally committed this!

packages/transformers/babel/src/babel7.js
marcins: Ah yep, I can remove this inner check.

marcins:
One thing to add - I did consider at some point whether instead of adding `PluginApplicationProfiler` everywhere I should just add the `createMeasurement` method to the existing `PublicLogger` as a profiling measurement can be argued to be a kind of log. In the end I talked myself out of it to avoid complexity with interdependencies between the logger & profiler, but happy to hear any other feels on this.

packages/transformers/babel/src/babel7.js
```diff
@@ -66,6 +79,25 @@ export default async function babel7(
...
 
+  if (applicationProfiler.enabled) {
+    config.wrapPluginVisitorMethod = (
+      key: string,
+      nodeType: string,
+      fn: Function,
+    ) => {
+      return function () {
+        let measurement;
+        measurement = applicationProfiler.createMeasurement(
```
mischnic:
```suggestion
        let measurement = applicationProfiler.createMeasurement(
```

lettertwo (approved):
😍

lettertwo (approved):

thebriando (approved):

devongovett:
This will be a really useful feature! My only question is around naming. I'm not sure what "application" refers to here. It's not really profiling the user's application so much as it is profiling Parcel. It seems more like a "trace" to me. Maybe we could have `--profile` and `--trace` as flags, and `trace` or `tracer` as the object passed to plugins? What do you think?

lettertwo:
Renamed ApplicationProfiler -> Tracer

Also renamed:
- @parcel/reporter-application-profiler -> @parcel/reporter-tracer
- PluginApplicationProfiler -> PluginTracer
- cli option `--profile-application` -> `--trace`
- PluginOptions.applicationProfiler -> PluginOptions.tracer

lettertwo:
@devongovett I think the rename is good, but with one small point of confusion: The `SamplingProfiler` has an accompanying `Trace` that is now very similar in name to `Tracer` (formerly called `ApplicationProfiler`.  Should we rename `Trace` to something that ties it to `SamplingProfiler`?

devongovett:
That's not exposed externally to parcel though right? If not then doesn't seem like too big a deal.

packages/core/workers/index.d.ts
```diff
@@ -11,6 +11,7 @@ export type FarmOptions = {
   workerPath?: FilePath,
   backend: BackendType,
   shouldPatchConsole?: boolean,
+  shouldProfileApplication?: boolean,
```
devongovett:
```suggestion
  shouldTrace?: boolean,
```

packages/transformers/babel/src/babel7.js
```diff
@@ -74,6 +76,28 @@ export default async function babel7(
     },
   };
 
+  if (tracer.enabled) {
+    config.wrapPluginVisitorMethod = (
```
devongovett, line 80: Does this wrap visiting each individual AST node, or just once for each plugin? How expensive is that?

devongovett (approved):

packages/transformers/babel/src/babel7.js
marcins: > Does this wrap visiting each individual AST node, or just once for each plugin? How expensive is that?

I just ran two full production Jira builds back-to-back, one with tracing & one without (both with `--no-cache`). The timings were 18m 29s without tracing, and 19m 1s with tracing (on an M1 Mac). So minimal overhead for tracing in general. 

The trace file produced is approximately 1.4GB - can't load it directly in the Perfetto UI (the WASM version), but works fine via the `trace_processor` CLI tool.

There are approximately 2.9m Babel transform entries - so yes, my understanding is that it wraps every visitor call for every AST type, but as you can see the overhead of this is relatively minimal.

devongovett (approved):

Issue #8673, opened 2022-12-05
Worker threads crash when unexpected execArgv are passed


maxpolun:
I'm running parcel as a bundler under [moonrepo](https://moonrepo.dev/), and one thing it does is set the process's title via the node arg `--title [task name]`. This should be fine for a regular node process, but parcel dies by default since it will try and use worker threads, and pass the node process args (in `process.execArgv`) to the worker thread's `execArgv`. However node worker threads will crash if given an unexpected `execArgv`, so the whole process crashes with an error like

```
Error [ERR_WORKER_INVALID_EXEC_ARGV]: Initiated Worker with invalid execArgv flags: --title
```

I'm currently working around this issue by setting `PARCEL_WORKER_BACKEND=process`, but it took me quite a while to figure this out. 

I'd say there are 3 options I can think of to improve this:
1. filter out any options known to throw (process level args, v8 tuning args)
2. just pass an empty list explicitly (so there's no execArgv in the child)
3. Catch the error on creating the worker and have a better error message (including mentioning `PARCEL_WORKER_BACKEND`)

I lean towards 2, just because it's simpler, will not break even with new options, and I don't think there's many execargv options that work with workers anyways, but I leave it to the project to decide. 

There is an issue in the node repo to change this behavior (which would be ideal), but it's been open for like a year: https://github.com/nodejs/node/issues/41103 

If one of these approaches sounds good, I can make a PR, they all seem pretty straightforward.

mischnic:
Option 1 is what we've been doing so far, so I'd say we just continue with that for now.
There has been a similar PR in the past: https://github.com/parcel-bundler/parcel/pull/5017

PR #8719, opened 2022-12-27
filter out title execArgv to workers

maxpolun:
# ↪️ Pull Request

As I reported in #8673 parcel will crash when a `--title [process name]` is passed to node and worker threads are being used. This PR filters that arg out so that parcel will not crash. 

## 💻 Examples

`--title [process name]` sets the name of the process (so it makes sense why threads can't take this arg). It can be set by some tools. I noticed it while using [moon](https://moonrepo.dev/).

## 🚨 Test instructions

There don't seem to be unit tests for the other filtered args, so I didn't add any here, but it's easy to test manually, just run `node --title test /path/to/parcel index.html`

## ✔️ PR Todo

### DIFF
--- a/packages/core/workers/src/Worker.js
+++ b/packages/core/workers/src/Worker.js
@@ -53,11 +53,11 @@ export default class Worker extends EventEmitter {
 
     for (let i = 0; i < filteredArgs.length; i++) {
       let arg = filteredArgs[i];
-      let hasArgWithParam =
+      let isArgWithParam =
         ((arg === '-r' || arg === '--require') &&
           filteredArgs[i + 1] === '@parcel/register') ||
         arg === '--title';
-      if (hasArgWithParam) {
+      if (isArgWithParam) {
         filteredArgs.splice(i, 2);
         i--;
       }
--- a/packages/core/workers/src/Worker.js
+++ b/packages/core/workers/src/Worker.js
@@ -53,10 +53,11 @@ export default class Worker extends EventEmitter {
 
     for (let i = 0; i < filteredArgs.length; i++) {
       let arg = filteredArgs[i];
-      if (
-        (arg === '-r' || arg === '--require') &&
-        filteredArgs[i + 1] === '@parcel/register'
-      ) {
+      let hasArgWithParam =
+        ((arg === '-r' || arg === '--require') &&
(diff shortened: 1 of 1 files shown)


packages/core/workers/src/Worker.js
```diff
@@ -53,10 +53,11 @@ export default class Worker extends EventEmitter {
...
     for (let i = 0; i < filteredArgs.length; i++) {
       let arg = filteredArgs[i];
-      if (
-        (arg === '-r' || arg === '--require') &&
-        filteredArgs[i + 1] === '@parcel/register'
-      ) {
+      let hasArgWithParam =
+        ((arg === '-r' || arg === '--require') &&
+          filteredArgs[i + 1] === '@parcel/register') ||
+        arg === '--title';
```
maxpolun, line 59: I can make this into a regex (like the single args filtered above) if you'd prefer. It's kind of pointless when there's only one, but I could do it for consistency.

packages/core/workers/src/Worker.js
```diff
@@ -53,10 +53,11 @@ export default class Worker extends EventEmitter {
 
     for (let i = 0; i < filteredArgs.length; i++) {
       let arg = filteredArgs[i];
-      if (
-        (arg === '-r' || arg === '--require') &&
-        filteredArgs[i + 1] === '@parcel/register'
-      ) {
+      let hasArgWithParam =
```
maxpolun: no `const` only `let` I guess, just matching the styles around here.

packages/core/workers/src/Worker.js
```diff
@@ -53,10 +53,11 @@ export default class Worker extends EventEmitter {
...
+        ((arg === '-r' || arg === '--require') &&
+          filteredArgs[i + 1] === '@parcel/register') ||
+        arg === '--title';
+      if (hasArgWithParam) {
```
mischnic:
```suggestion
      let isArgWithParam =
        ((arg === '-r' || arg === '--require') &&
          filteredArgs[i + 1] === '@parcel/register') ||
        arg === '--title';
      if (isArgWithParam) {
```

mischnic (approved):

Dezzj (approved):
Approved

PR #8802, opened 2023-01-31
feat: add @parcel/optimizer-svgo3 plugin

### LINKED ISSUE #8701: deprecation warning when installing: Modern JS already guarantees Array#sort() is a stable sort, so this library is deprecated.
# 🐛 bug report

```
$ npm i parcel
⸨#########⠂⠂⠂⠂⠂⠂⠂⠂⠂⸩ ⠙ idealTree:posthtml: sill placeDep node_modules/@parcel/transformer-html semver@5.7npm WARN deprecated stable@0.1.8: Modern JS already guarantees Array#sort() is a stable sort, so this library is deprecated. See the compatibility table on MDN: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort#browser_compatibility
```

## 🤔 Expected Behavior

Installing parcel should not show any deprecation warnings.

## 🌍 Your Environment

| Parcel           | v2.8.2
| Node             | v16.13.1
| npm         | 8.1.2
| Operating System | Linux debian 5.10.0-19-amd64 #1 SMP Debian 5.10.149-2 (2022-10-21) x86_64 GNU/Linux

jack0pan:
# ↪️ Pull Request

Fix #8701, but svgo 3.0 has [breaking changes](https://github.com/svg/svgo/releases/tag/v3.0.0):

- Node.js 14+ is required for version
- stable package is replaced with native stable sort (required node 12+)
- and some plugins' name was changed

In this major version of parcel:
1. introduce two new plugins for svgo 3.
  - `@parcel/optimizer-htmlnano3`
  - `@parcel/optimizer-svgo3`

2. use them with config:

```
{
  "optimizers": {
    "*.{html,xhtml}": ["@parcel/optimizer-htmlnano3"],
    "*.svg": ["@parcel/optimizer-svgo3"]
  }
}
```

In the next major of parcel:
1. remove `optimizers/htmlnano-3` and `optimizers/svgo-3`.
2. change `optimizers/htmlnano` and `optimizers/svgo` to work with svgo 3.

## ✔️ PR Todo


devongovett:
Are there any breaking changes in this release?

jack0pan:
> Are there any breaking changes in this release?

Add to the comment.

jack0pan:
Fix the params of svgo optimizer:

- [cleanupIds](https://github.com/svg/svgo/blob/main/plugins/cleanupIds.js)

mischnic:
> Node.js 14+ is required for version

Parcel (officially) supports Node 12, but since that's EOL we should probably move on anyway.

devongovett:
I'm more concerned that the plugin names seem to have changed. That would break people's config files.

jack0pan:
> > Node.js 14+ is required for version
> 
> Parcel (officially) supports Node 12, but since that's EOL we should probably move on anyway.



> I'm more concerned that the plugin names seem to have changed. That would break people's config files.

All right, wait the next major version?

mischnic:
We have a similar situation here as well: https://github.com/parcel-bundler/parcel/pull/7922

jack0pan:
OK. Let me try.

jack0pan:
@devongovett @mischnic Please review this pr.

devongovett:
Hey sorry @jack0pan. We haven't yet figured out what we want to do here. Introducing a whole bunch of new plugins that users must manually upgrade isn't ideal. Ideally new projects would automatically get the new version while existing projects would stick with the version they have until they decide to upgrade. I'm not sure if there is a way to do that. One idea was proposed in https://github.com/parcel-bundler/parcel/pull/7922#issuecomment-1100140323

> Bump a major version on @parcel/transformer-mdx. Change the dependency range in @parcel/config-default to e.g. ^2.4.1 || ^3.0.0. This way, existing projects will continue to work with the old version and would need to manually update the version installed in their projects, but new projects would automatically get the latest when auto installed. I think this is more optimal, but I have not tested whether it will actually work at the moment or whether we need to make changes to support it.

I still have not tested what package managers do in this case. Would be good to figure it out though because this is going to keep happening.

srbaeza-ai:
- [ ] https://github.com/jhines2k7/parcel...

Issue #8470, opened 2022-09-13
Support `process.env.npm_package_version` Inlining


graphemecluster:
…because Node already supports it.

Currently it seems that there are no way to include the version field in `package.json` without prepending `cross-env NPM_PACKAGE_VERSION=$npm_package_version` to the parcel commands. That's verbose, and requires `cross-env` to be installed.

Of course you might say, I could just import the JSON file directly by `import { version } from "../package.json"`. But that requires manually turning `resolveJsonModule` on. Moreover, this will including the whole JSON file, exposing other useless information in the bundle. I hate importing things outside `src` either.

Some people might want `npm_package_name` and `npm_package_description` too.

panstav:
According to the [docs](https://parceljs.org/features/node-emulation/#disabling-these-features), you can add this to your `package.json`:
```
"@parcel/transformer-js": {
    "inlineEnvironment": ["npm_package_version"]
}
```

tristan-f-r:
I do think that in the spirit of [node emulation](https://parceljs.org/features/node-emulation/), all `node_package_` variables should be inlined. ([NPM Docs](https://docs.npmjs.com/cli/v9/using-npm/scripts#packagejson-vars))

tomchen:
> According to the [docs](https://parceljs.org/features/node-emulation/#disabling-these-features), you can add this to your `package.json`:
> 
> ```
> "@parcel/transformer-js": {
>     "inlineEnvironment": ["npm_package_version"]
> }
> ```

Not only this, but user has to use `"context": "browser"` in `targets` like I just did in the commit to peerjs that's mentioned above - default context "node" can't inline `process.env.npm_package_version` 

But why with context "node" it doesn't inline `process.env.npm_package_version`? if `@parcel/transformer-js.inlineEnvironment` sets it explicitly, it should inline it.

PR #8806, opened 2023-02-02
feat: npm_package_* env variables

tristan-f-r:
# ↪️ Pull Request

This adds support for `process.env.npm_package_*` variables (for #8470)

## 💻 Examples

```js
console.log("Some cool site v" + process.env.npm_package_version)
```
## ✔️ PR Todo

### DIFF
--- a/packages/core/core/src/loadDotEnv.js
+++ b/packages/core/core/src/loadDotEnv.js
@@ -52,5 +52,27 @@ export default async function loadEnv(
     }),
   );
 
+  const packageFile = await resolveConfig(
+    fs,
+    filePath,
+    ['package.json'],
+    projectRoot,
+  );
+
+  // load npm_package_* variables from package.json (for node emulation)
+  if (packageFile != null) {
+    const packageJSON = await fs.readFile(packageFile, 'utf8').then(JSON.parse);
+
+    let packageEnv = packageJSON
+      .map((key, value) => {
+        if (typeof value === 'string') {
+          return 'npm_package_' + key.replace(/-/g, '_');
+        }
+      })
+      .filter(Boolean);
+
+    envs.push(packageEnv);
+  }
+
   return Object.assign({}, ...envs);
 }
--- a/packages/core/core/src/loadDotEnv.js
+++ b/packages/core/core/src/loadDotEnv.js
@@ -25,6 +25,8 @@ export default async function loadEnv(
     `.env.${NODE_ENV}.local`,
   ].filter(Boolean);
 
+  type Env = {[string]: string};
+
   let envs = await Promise.all(
     dotenvFiles.map(async dotenvFile => {
       const envPath = await resolveConfig(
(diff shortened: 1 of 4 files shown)


graphemecluster:
Thanks! Let's add a nested object key test case.

tristan-f-r:
> Thanks! Let's add a nested object key test case.

Thanks for the catch! Didn't realize that nested objects are also grabbed:

> Objects are flattened following this format, so if you had {"scripts":{"install":"foo.js"}} in your package.json, then you'd see this in the script ([source](https://docs.npmjs.com/cli/v9/using-npm/scripts#current-lifecycle-event))

graphemecluster:
It seems like currently NPM only copies part of the values: https://github.com/npm/run-script/blob/main/lib/package-envs.js, plus  `npm_package_json`, which is the path of `package.json`.
And I don't know if `npm_config_*` and `npm_lifecycle_*` should also be included.

tristan-f-r:
Thanks for the extra info -- I want to limit this PR solely for JSON information for the time being to make it easier to review, unless requested by one it the maintainers.

devongovett:
Doesn't npm already set these environment variables? Why do we need to do it ourselves?

tristan-f-r:
That's a very good point, but it seems the behavior for parcel doesn't actually load said environment variables.

Reproduction:

`package.json`

```json
{
	"version": "1.0.0",
	"dependencies": {
		"parcel": "^2.8.3"
	},
	"scripts": {
		"start": "parcel index.html",
		"start:node": "node index.js"
	}
}
```

`index.html`

```html
<script type="module">
    console.log(process.env.npm_package_version);
</script>
```

`index.js`

```js
console.log(process.env.npm_package_version);
```

However, thanks for bringing this up - I'm not sure if this is fine with parcel's internal cache system, but would it be a better idea to instead pass these missing environment variables from node directly.

~~Oddly enough, it seems that environment variables aren't passed (e.g. try `TEST=1 npm run start` and try to `console.log(process.env.TEST)`)~~

It seems that `loadDotEnv.js` [doesn't *actually* inject `process.env`](https://github.com/parcel-bundler/parcel/blob/1fd82417122fccd5ea465b2c237f736fc3d2d53d/packages/core/core/src/loadDotEnv.js#L16) and rather just looks for NODE_ENV (which I assume is for a good reason).

~~If it's **not** for a good reason and this is the issue (if someone else can triage that would be awesome)~~, then I don't mind remaking this PR to inject these environment variables (npm injects their own environment variables as @graphemecluster mentioned, so it would fix that as well.)

**EDIT:** Almost forgot about `inlineEnvironment`. Maybe instead a PR can be introduced that automatically adds `npm_package_*` environment variables as expected defaults?

devongovett:
loadDotEnv should only be loading `.env` files, which it merges with `process.env` here: https://github.com/parcel-bundler/parcel/blob/1fd82417122fccd5ea465b2c237f736fc3d2d53d/packages/core/core/src/resolveOptions.js#L113-L122 So it should already work. You might start by checking what `process.env` contains at that point.

graphemecluster:
On a side note, remember to handle the case of Array values like what the script I linked does.

tristan-f-r:
I've made a reproducable repository here: https://github.com/LeoDog896/parcel-implicit-env-test

The node version gets all 3 environment variables, while parcel only gets two.

(Node output - `ENV` comes from the `.env` file in the repo):

```
Shouldn't be 'ENV' (node):
undefined

Should be 'WOO':
WOO

Should be parcel-implicit-env-test:
parcel-implicit-env-test
```

(Parcel output): https://leodog896.github.io/parcel-implicit-env-test/

```
Should be 'ENV': ENV

Should be 'WOO': WOO

Should be parcel-implicit-env-test: undefined
```

This does confirm that parcel reads from env, ~~but for some odd reason, the node env injection isn't happening at that stage.~~ If I find what is *technically* happening in parcel / NPM causing the environment variables in node not to be read on parcel's side, I'll remake this PR to expose these NPM environment variables for inlining. 

EDIT: Added a python example - NPM is injecting the environment variables in the script side, but *parcel* is the one not reading them. I'll dig deeper.

devongovett:
Looks like it's due to this line: https://github.com/parcel-bundler/parcel/blob/1fd82417122fccd5ea465b2c237f736fc3d2d53d/packages/transformers/js/src/JSTransformer.js#L361

I don't fully remember why it was added, but my guess is performance since there are a ton of environment variables which would need to be serialized and sent from JS to Rust when most of the time they are unused. We could revisit this and test the performance. I think you could also probably configure Parcel to include them with a glob as described [in the docs](https://parceljs.org/features/node-emulation/#disabling-these-features).

tristan-f-r:
Awesome! I'll close this PR for now and see if the performance impact of including `npm_` is too large. If it is, I'll comment back here. Thanks for the help.

Issue #5784, opened 2021-02-05
`package.json#source` field is used for npm packages with pnpm


mischnic:
# 🐛 bug report

When using pnpm, the resolver tries to use the `"source"` field of a package from npm.

## 😯 Current Behavior

```
🚨 Build failed.
@parcel/core: Failed to resolve 'compute-scroll-into-view' from './index.js'
/index.js:1:35
> 1 | import computeScrollIntoView from "compute-scroll-into-view";
>   |                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^
  2 |
  3 | computeScrollIntoView();
@parcel/resolver-default: Could not load './src/index.ts' from module 'compute-scroll-into-view' found in package.json#source
/node_modules/compute-scroll-into-view/package.json:88:13
  87 |   },
> 88 |   "source": "src/index.ts",
>    |             ^^^^^^^^^^^^^^ './src/index.ts' does not exist, did you mean './dist/index.js'?'
  89 |   "umd:main": "umd/compute-scroll-into-view.min.js"
  90 | }
```

## 🤔 Expected Behavior

Ignore `package.json#source` for npm packages

## 🔦 Context

https://github.com/parcel-bundler/parcel/issues/5604#issuecomment-773799994

## 💻 Code Sample

```js
import computeScrollIntoView from "compute-scroll-into-view";

computeScrollIntoView();
```

```json
{
  "dependencies": {
    "@babel/core": "^7.12.13",
    "compute-scroll-into-view": "^1.0.16"
  }
}
```

Use `pnpm install`!

## 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Parcel           |  31f431d93da97d3edc3f489847053ef601412666

ngfk:
@mischnic Did you find a workaround for this issue? I'm also trying to load `compute-scroll-into-view` in a pnpm + parcel2 project.

mischnic:
I don't think there is a workaround, unfortunately.

This is the problematic code
https://github.com/parcel-bundler/parcel/blob/f28eaf60ac018d49d2176c563c86d09f84feaa8e/packages/utils/node-resolver-core/src/NodeResolver.js#L647-L654

Until pnpm, symlinks meant that you're using a monorepo and you're not importing an npm package but another monorepo package. The condition should probably be adjusted to make pnpm work.

ngfk:
Maybe something like this? This is very specific to pnpm though.

```js
[...]

PR #8807, opened 2023-02-02
New resolver implementation in Rust

### LINKED ISSUE #202: [TypeScript] `baseUrl` inside tsconfig.json not supporting.
🐛 bug report

`tsconfig.json`:
```js
{
    "compilerOptions": {
        "outDir": "./dist/",
        "baseUrl": "./src",
        "sourceMap": true,
        "module": "commonjs",
        "target": "es5", 
        "removeComments": true
    }
}
```

### 🤔 Expected Behavior 
structure: 
```js
- project
   - src
      - component
         Custom.ts
      Main.ts
   index.html
   tsconfig.json
```
component/Custom.ts:
```typescript
export default function hello() {
    console.log('Hello!');
};
```
Main.ts:
```typescript
import hello from 'component/Custom';

hello();
```
console should show 'Hello!'

### 😯 Current Behavior 
Running from shell:
```shell
$ parcel index.html
/Users/shunia/Documents/parcel-test/src/Main.ts:3:23: Cannot resolve dependency 'components/Custom'
  1 | import hello from "components/Custom";
  2 |
> 3 | hello();
```

### 💁 Possible Solution 
  The compiler should read `tsconfig.json` file right, and understands that `baseUrl` has made the `from` part of `import ... from ...` no need to be the actual relative path from the current file.

  In my case, because of `baseUrl` set to `./src`, the original `import hello from '../componet/Custom' can be replaced as `import hello from 'component/Custom'`. 

  This might because of the way typescript compiles the `import` part just as what is written in the codes:
```typescript
import hello from "components/Custom";
```
is just compiled to:
```js
var Custom_1 = require("components/Custom");
```
  And then, parcel-require can not recognize the path and throws error.

### 🔦 Context
  Our actual project are full of relative paths based on `baseUrl`, which we omitted the `../` parts when using relative import. With `webpack` and `awesome-typescript-loader` it works as expected. So when we are trying to get our hands on parcel, this problem really blocks us from going further.

### 🌍 Your Environment

| Parcel           |  1.1.0
| Node             |  v8.9.3
| npm/Yarn         |  Yarn 1.3.2
| Operating System |  macOS Sierra 10.12.6

Sorry for my poor english.

### LINKED ISSUE #4155: Use `package.json#exports` map
# 🙋 feature request

Use `exports` maps in the resolver

https://nodejs.org/api/esm.html#esm_package_exports
https://nodejs.org/api/esm.html#esm_conditional_exports
https://webpack.js.org/guides/package-exports/

## 💻 Examples

https://github.com/preactjs/preact/blob/master/package.json#L11
https://unpkg.com/browse/nanoid@3.1.18/package.json

## Relevant code

The resolver: https://github.com/parcel-bundler/parcel/blob/v2/packages/resolvers/default/src/DefaultResolver.js
which calls: https://github.com/parcel-bundler/parcel/blob/v2/packages/utils/node-resolver-core/src/NodeResolver.js

### LINKED ISSUE #4936: Implement TypeScript resolver plugin
TypeScript supports some [additional features](https://www.typescriptlang.org/docs/handbook/module-resolution.html) in its resolver on top of the Node resolution algorithm. These are configured in `tsconfig.json`.

* `baseUrl` - a root directory from which to resolve non-relative paths, in addition to `node_modules`. See also: #202.
* `paths` - additional directories, relative to `baseUrl`, that should also be searched.
* `rootDirs` - virtual directories that relative paths can be resolved from

There's also the legacy module resolution mode that TS supports. Not sure if that's something we should also support, but I guess someone might still use it. 🤷 

## Implementation

These resolution features should be supported in Parcel for imports declared in `.ts` or `.tsx` files. This can be done via a new resolver plugin, included in the default config: `@parcel/resolver-typescript`. This resolver should look at the `sourcePath` property of the dependency, and if it has a `.ts` or `.tsx` extension, it should resolve using the TS compiler API's `resolveModuleName` method. It should use a custom `ModuleResolutionHost` that ensures the Parcel file system is used instead of the Node one. This can be created using the `@parcel/ts-utils` package, which already includes an `FSHost` package implementing much of the required interface. 

If a module is not resolved, or the source file is not a `.ts` or `.tsx` file, then the resolver should return `null`. This will then fall back to the default Parcel resolver, which will handle Parcel specific features like aliases.

## Open questions

* We'll need to determine whether we can properly track all of the files that the TS resolver reads so we can invalidate the Parcel cache properly. This is a work in progress for the default resolver as well.
* It's unclear how the TS resolver should interact with Parcel features like aliases and named pipelines (e.g. `url:./foo.jpg` to import an image as a url). Theoretically we could fall back to the default resolver if TS doesn't find anything, but what if you alias `react` to `preact` but `react` *is* found by the TS resolver?

### LINKED ISSUE #8839: Import specifiers starting with a hash # resolve to the current asset
# 🐛 bug report

Import specifiers starting with a hash will resolve to the current asset.

## 🤔 Expected Behavior

Either error, or use the `imports` field.

## 😯 Current Behavior

The file reimports itself. 

## 🔦 Context

https://github.com/parcel-bundler/parcel/discussions/8836

## 💻 Code Sample

```
import { method } from "#/foo.js";
console.log(method());
```

with 
```
	"alias": {
		"#/*": "./$1"
	},
```

## 🌍 Your Environment

| Parcel           | 2.8.3

devongovett:
This is a new implementation of Parcel's resolver. It includes a few new features to improve compatibility with other tools:

* package.json exports and imports (Fixes #4155, Fixes https://github.com/parcel-bundler/parcel/issues/8839)
* tsconfig.json baseUrl, paths, and moduleSuffixes (Fixes #202, fixes #4936)

It also replaces 3 separate node resolver implementations with a single one. Previously, we had separate implementations for the default resolver plugin vs in the NodePackageManager. Now they all use a single library, written in Rust, which provides lots of feature flags to enable turning on or off different resolver features. This allows us to match Node's behavior when needed, and optionally turn on additional features to match TypeScript, webpack, previous Parcel versions, etc. It will also enable future features like ESM plugins, and prepare us for the future when more of Parcel core is written in Rust.

Since it is written in Rust, most file system operations can happen natively if you are using a `NodeFS` (and Yarn PnP is not enabled). Otherwise, we pass in JS callbacks that are called to read files. The API in the `@parcel/node-resolver-core` package is now a wrapper around the Rust implementation providing the same API as before (as long as you weren't using internal methods). Yarn PnP, diagnostics, and some other things like checking dependency versions and resolving builtin polyfills are still done in JS since they are rare and not performance critical.

For package.json exports, we try to match the [conditions that webpack](https://webpack.js.org/guides/package-exports/#conditions) applies based on Parcel's environment, though not all of them are supported. Currently, there is no way to specify custom conditions - not sure how we would want to do that.

There are a few questions about how some of these resolution features should work together. For example, should aliases apply to package.json exports or tsconfig paths? Would they go before or after? etc. See some comments in the code for more.

To do

- [x] Clean up code
- [x] Performance testing
- [x] Run a lot more test to ensure compatibility with other tools


packages/resolvers/glob/src/GlobResolver.js
```diff
@@ -94,22 +94,14 @@ export default (new Resolver({
...
-        loc: dependency.loc,
-        range: dependency.range,
-      };
-
       let result;
       try {
-        result = await resolver.resolveModule({
-          filename: pkg,
+        result = await resolver.resolve({
+          filename: pkg + '/package.json',
```
devongovett, line 93: Not sure how we should handle packages with an exports field here. This won't work in that case, and anyway we wouldn't want the glob to match based on files, it would need to be based on what is exposed by the exports field. Not sure if that would work in all cases.

packages/core/package-manager/src/NodePackageManager.js
```diff
@@ -26,8 +27,12 @@ import {getModuleParts} from '@parcel/utils';
 import {getConflictingLocalDependencies} from './utils';
 import {installPackage} from './installPackage';
 import pkg from '../package.json';
-import {NodeResolver} from './NodeResolver';
-import {NodeResolverSync} from './NodeResolverSync';
+import {Resolver} from '@parcel/node-resolver-core/index';
```
mischnic:
```suggestion
import {Resolver} from '@parcel/node-resolver-core';
```

packages/utils/node-resolver-core/package.json
```diff
@@ -14,12 +14,22 @@
     "url": "https://github.com/parcel-bundler/parcel.git"
   },
   "main": "lib/NodeResolver.js",
```
mischnic: Shouldn't make a different, but for consistency...
```suggestion
  "main": "lib/Wrapper.js",
```

mischnic:
Should we keep some tests that use `new URL` without the js extension? You said that you kept the behaviour the same?

Where does `zipfs` in the paths come from?

packages/core/package-manager/src/NodePackageManager.js
devongovett: Yeah I need to expose the underlying Rust implementation rather than the wrapper from that package for this

devongovett:
> Should we keep some tests that use new URL without the js extension?

Yeah I was gonna revert those test changes. I think they should pass as is now.

> Where does zipfs in the paths come from?

I wanted to test that we actually use the JS file system for PnP, so I added this to ensure that the paths returned didn't exist. That way the native Rust FS can't read them and we have to go through JS.

packages/utils/node-resolver-rs/src/lib.rs
```diff
@@ -0,0 +1,2392 @@
...
+    if let Some(ext) = from.extension() {
+      if ext == "ts" || ext == "tsx" || ext == "mts" || ext == "cts" {
+        flags |= RequestFlags::IN_TS_FILE;
+      } else if ext == "js" || ext == "jsx" || ext == "mjs" || ext == "cjs" {
+        flags |= RequestFlags::IN_JS_FILE;
+      }
+    }
+
+    if from.components().any(|c| c.as_os_str() == "node_modules") {
+      flags |= RequestFlags::IN_NODE_MODULES;
```
devongovett, line 283: This might not apply with yarn pnp?

packages/utils/node-resolver-rs/src/cache.rs
```diff
@@ -0,0 +1,169 @@
...
+      let mut pkg = PackageJson::parse(path.clone(), data).map_err(|e| JsonError::new(path, e))?;
+
+      // If the package has a `source` field, make sure
+      // - the package is behind symlinks
+      // - and the realpath to the packages does not includes `node_modules`.
+      // Since such package is likely a pre-compiled module
+      // installed with package managers, rather than including a source code.
+      if !matches!(pkg.source, SourceField::None) {
+        let realpath = fs.canonicalize(&pkg.path)?;
+        if realpath == pkg.path
```
devongovett, line 128: Not sure if maybe we should change this logic for pnpm... #5784

packages/core/core/src/RequestTracker.js
```diff
@@ -675,7 +675,7 @@ export class RequestGraph extends ContentGraph<
           requestGraphEdgeTypes.invalidated_by_create_above,
         ) &&
         isDirectoryInside(
-          path.dirname(fromProjectPathRelative(matchNode.value.filePath)),
+          fromProjectPathRelative(matchNode.value.filePath),
```
devongovett, line 678: This change reduces the number of invalidations we need because `aboveFilePath` refers to a directory rather than a file.

devongovett:
A follow up I'd like to do to improve performance is to create an instance of `NodeResolver` once per build rather than once per request. This way the cache can actually be reused. This will require a new method be introduced for `Resolver` plugins that runs on build start in order for the instance to be created. Not sure what to call that. Is it the same as `loadConfig`? If we added that, when would it run?

packages/utils/node-resolver-rs/src/lib.rs
mischnic: A package named `sub` might have this filepath with PnP: `/Users/niklas/.yarn/berry/cache/sub-file-8e8024d312-8.zip/node_modules/sub/index.js`. So there will be a node_modules in there (if that's what you were concerned about)

packages/resolvers/glob/src/GlobResolver.js
mischnic: The case of `import("@org/some-package/*")` is probably also rather uncommon? So I think it's fine to keep it using files for now.

packages/utils/node-resolver-rs/src/cache.rs
mischnic: I think this should work because the realpath will indeed contain node_modules: `/Users/niklas/Desktop/test/node_modules/.pnpm/file+sub/node_modules/sub/index.js`? Or which case are you thinking of that doesn't work with this logic?

devongovett: Ah ok I wasn't sure if node_modules was still in the path

packages/utils/node-resolver-rs/src/specifier.rs
```diff
@@ -0,0 +1,303 @@
...
+  if let Some(pos) = input.chars().position(|c| c == '?' || c == '#') {
+    (&input[0..pos], &input[pos..])
+  } else {
+    (input, "")
+  }
+}
+
+// https://url.spec.whatwg.org/#query-state
+fn parse_query<'a>(input: &'a str) -> (Option<&'a str>, &'a str) {
+  if !input.is_empty() && input.as_bytes()[0] == b'?' {
```
mischnic, line 209: ~`as_bytes()[0] == b'?'` is probably incorrect with multibyte characters?~ It can never be a starting byte in a multibyte utf8 character...

packages/resolvers/default/src/DefaultResolver.js
```diff
@@ -18,13 +18,6 @@ export default (new Resolver({
     const resolver = new NodeResolver({
       fs: options.inputFS,
       projectRoot: options.projectRoot,
-      // Extensions are always required in URL dependencies.
-      extensions:
-        dependency.specifierType === 'commonjs' ||
-        dependency.specifierType === 'esm'
-          ? ['ts', 'tsx', 'mjs', 'js', 'jsx', 'cjs', 'json']
-          : [],
-      mainFields: ['source', 'browser', 'module', 'main'],
```
mischnic, line 27: This could be a breaking change for some plugins. Should we try to keep supporting this? Or was the plan to do a major bump on node-resolver-core anyway?

packages/utils/node-resolver-core/src/Wrapper.js
```diff
@@ -0,0 +1,764 @@
...
+            return {
+              diagnostics: [diagnostic],
+              invalidateOnFileCreate: res.invalidateOnFileCreate,
+              invalidateOnFileChange: res.invalidateOnFileChange,
+            };
+          }
+        }
+
+        // TODO: invalidations?
+        return {isExcluded: true};
```
mischnic: Changing aliases/exports/... could probably influence whether it's external or not?

packages/utils/node-resolver-rs/src/lib.rs
```diff
@@ -0,0 +1,2479 @@
...
+    /// Parcel-style tilde specifiers resolved relative to nearest module root.
+    const TILDE_SPECIFIERS = 1 << 1;
+    /// The `npm:` scheme.
+    const NPM_SCHEME = 1 << 2;
+    /// The "alias" field in package.json.
+    const ALIASES = 1 << 3;
+    /// The settings in tsconfig.json.
+    const TSCONFIG = 1 << 4;
+    /// The "exports" field in package.json.
+    const EXPORTS = 1 << 5;
```
mischnic: That's for both exports and imports, right?

packages/utils/node-resolver-rs/src/lib.rs
```diff
@@ -0,0 +1,2479 @@
...
+    // Next, check tsconfig.json for the paths and baseUrl options.
+    if let Some(res) = self.resolve_tsconfig_paths()? {
+      return Ok(res);
+    }
+
+    self.resolve_node_module(module, subpath)
+  }
+
+  fn resolve_node_module(&self, module: &str, subpath: &str) -> Result<Resolution, ResolverError> {
+    // TODO: check if module == self
```
mischnic: Still relevant?

mischnic:
Should it also read from jsconfig.json? https://github.com/parcel-bundler/parcel/blob/9d1933b3d89a3f57f3ac1c70f4fb755738d78604/packages/transformers/js/src/JSTransformer.js#L184-L187

packages/resolvers/default/src/DefaultResolver.js
devongovett: Yeah it's a bit hard to support custom main fields with the way package.jsons are parsed now (rust discards unknown fields). Extensions could be done but idk if there are any use cases other than replicating the default logic from here. I think the safest thing would be to bump a major anyway because none of the methods other than `resolve` exist anymore, and someone might have been using them.

packages/utils/node-resolver-rs/src/lib.rs
devongovett: The [node docs](https://nodejs.org/api/packages.html#self-referencing-a-package-using-its-name) have a section on this, and it's even got a special [PACKAGE_SELF_RESOLVE](https://nodejs.org/api/esm.html#resolver-algorithm-specification) function in the spec. But I don't really understand why it's called out specifically, because self references work even without this special handling just by searching the node_modules folders as with any other package. I added some tests for it but 🤷

devongovett:
> Should it also read from jsconfig.json?

Yeah maybe but it would add yet another file that needs to be checked for in every directory. Also other tools like esbuild don't check for it either. I want to start with supporting tsconfig.json and see if we get feedback about needing jsconfig.json too.

fregante:
TypeScript *and* `exports`? 😱 🎉

jboler:
Will this new resolver support import path aliases starting with `@` ? e.g.

```ts
  // package.json
  "alias": {
    "@/**": "/src/$1"
  },

  // component.tsx
  import { thing } from '@/components/...'

```

fregante:
Doesn't it already? I think you just need `"@/**/*": "./src/$1/$2"`

jboler:
@fregante You're absolutely right - thanks for that!

nastynaz:
Does this actually work? I've been stuck for 4 hours trying to get it working. I'm on parcel 2.9.0. Fresh repo with the following tsconfig:

```json
{
  "compilerOptions": {
    "target": "es2022",
    "paths": {
      "shared/*": ["./src/shared/*"]
    },
    "include": ["src/**/*"]
  }
}
```

After running `parcel build` and then running the code with node, I get a "cannot find package 'shared'" error. The code doesn't exist but the imports are still trying to import a nonexistant 'shared' folder.

I've also tried adding various aliases to my package.json to no avail. If anyone could help me I'd greatly appreciate it!

fregante:
I think it’s opt-in at the moment. You need to customize it in your parcelrc file:

https://parceljs.org/blog/v2-9-0/#new-resolver

nastynaz:
@fregante Unfortunately that doesn't work.

@devongovett  Since Parcel uses swc under the hood, is it possible to just change the `.swcrc` file/config it uses?

Because of this issue I went back to using swc. It's pretty easy to set up there, you just copy your `tsconfig.json` `paths` statement directly into the `.swcrc`:
```json
{
  "jsc": {
    "parser": {
      "syntax": "typescript",
      "decorators": true
    },
    "target": "esnext",
    "paths": {
      "opensea/*": ["./src/shared/opensea/*"]
    }
  },
  "module": {
    "type": "commonjs"
  },
  "sourceMaps": true
}
```

PR #8826, opened 2023-02-10
Small changes helpful when developing Parcel

richardkmichael:
- `await` all GraphViz calls
- prefix GraphViz PNGs with `parcel-`, to make finding them easier

### DIFF
--- a/packages/core/core/src/Parcel.js
+++ b/packages/core/core/src/Parcel.js
@@ -274,7 +274,7 @@ export default class Parcel {
 
       this.#requestedAssetIds.clear();
 
-      dumpGraphToGraphViz(
+      await dumpGraphToGraphViz(
         // $FlowFixMe
         this.#requestTracker.graph,
         'RequestGraph',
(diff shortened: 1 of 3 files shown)


richardkmichael:
Unsure if you want these kinds of tiny changes.  I found them helpful, but I can maintain a local patch.

packages/core/utils/src/progress-message.js
```diff
@@ -6,7 +6,7 @@ import path from 'path';
 export function getProgressMessage(event: BuildProgressEvent): ?string {
   switch (event.phase) {
     case 'transforming':
-      return `Building ${path.basename(event.filePath)}...`;
+      return `Transforming ${path.basename(event.filePath)}...`;
```
mischnic: @devongovett This was always a bit of misleading, maybe a remnant from Parcel 1? Or should we leave it as is?

mischnic: Let's leave it as is for now so that we can merge the rest. 

```suggestion
      return `Building ${path.basename(event.filePath)}...`;
```

mischnic (approved):
Thanks (BTW the commits are going to be squashed anyway, so keep a clean commit history here doesn't really make a difference to us)

richardkmichael:
> Thanks (BTW the commits are going to be squashed anyway, so keep a clean commit history here doesn't really make a difference to us)

Ok thanks for mentioning.  Here, it was just faster to drop and push than to actually edit and commit, etc. :)

PR #8859, opened 2023-02-25
Doc: update wrong help option

benjamin-thomas:
$ parcel --version
2.8.3

$ parcel serve 

$ parcel serve --help | grep -A1 "\-\-host"
  --host <host>              set the host to listen on, defaults to listening
                             on all interfaces

# ↪️ Pull Request

## ✔️ PR Todo

### DIFF
--- a/packages/core/parcel/src/cli.js
+++ b/packages/core/parcel/src/cli.js
@@ -111,7 +111,7 @@ var hmrOptions = {
     process.env.PORT,
   ],
   '--host <host>':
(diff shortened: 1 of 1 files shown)


mischnic:
For better or for worse, by default it appears to listen on all interfaces (I can access it from my phone as well), even if just `localhost:1234` is printed.

benjamin-thomas:
Oh wow indeed. I think this is called a… gotcha ;)

Feel free to close then (unless you want to keep open for ref?)

PR #8860, opened 2023-02-25
Switch to SWC minifier by default

devongovett:
This replaces Terser with the SWC minifier by default. In testing large real-world apps, we see a significant performance improvement of 15-20% with slightly smaller output as well. Since SWC supports the same configuration options as terser, it is a drop in replacement. Next.js recently enabled it by default as well, so it seems stable at this point.

Aside from switching the default config over, this also fixes a few tests that now have slightly different output, and converts errors to Parcel diagnostics.


lettertwo (approved):

danieltroger:
FYI I had to switch back since my .terserrc is not supported:

```
{
  "mangle": {
    "eval": true,
    "properties": {
      "regex": "^(?!__core-js_shared__).*_$"
    }
  },
  "compress": {
    "unsafe": true,
    "arguments": true,
    "ecma": "2020",
    "passes": 4,
    "unsafe_arrows": true
  },
  "format": {
    "ecma": 2020
  }
}
```

mischnic:
@danieltroger Please open a new issue with the actual error message about which property isn't supported. Ideally in the swc repo
swc already made some changes to support all terserrcs in the past: https://github.com/parcel-bundler/parcel/issues/9174, https://github.com/parcel-bundler/parcel/issues/9041

danieltroger:
-> https://github.com/swc-project/swc/issues/7754

jboler:
When building my app with parcel 2.10.2 vs 2.8, my bundle is ~15% larger. I inspected the minified bundle and found that no comments were being removed at all.

I added this to a `.terserrc` and it removed most of the comments except some copyright comments:

```json
{
  "format": {
    "comments": "some"
  }
}
```

Whereas parcel 2.8 would leave only the copyright/license comments by default. Is the intention to default to leaving comments in the minified bundle in parcel 2.10?

devongovett:
Looks like there was a regression in SWC minifier recently. But appears to be fixed in the latest nightly version of `@swc/core` (in https://github.com/swc-project/swc/pull/8257).

PR #8862, opened 2023-02-27
Use BitSet for bundler intersections

mattcompiles:
# ↪️ Pull Request

In really large projects, the bundler can spend huge amounts of time performing expensive set intersections. This PR replaces Set's with a custom `BitSet` class implemented with `BigInt` when intersections are needed to be performed. 

The BitSet requires a list of all possible items before individual sets can be made. In the case of the bundler, this is the list of all known assets. 

Unfortunately, we don't really see any performance wins on most projects but a large internal project saw a 44 second improvement to dev bundle time.   

**Bonus change**
I upgraded our eslint config to es2020 to add support for BigInts

## 🚨 Test instructions

Tests added, plus existing bundler tests should be sufficient

## ✔️ PR Todo


packages/core/utils/src/BitSet.js
```diff
@@ -0,0 +1,121 @@
...
+
+    while (tmpValue > BIGINT_ZERO) {
+      // Get last set bit
+      i = tmpValue.toString(2).length - 1;
+
+      values.push(this._items[i]);
+
+      // Unset last set bit
+      tmpValue &= ~(BIGINT_ONE << numberToBigInt(i));
+    }
```
devongovett, line 117: You could do this in reverse order without the (probably expensive) string conversion on each iteration. Not sure if order is important here or not.

```js
for (let i = 0, mask = 1n; b != 0n; i++, mask <<= 1n) {
  if (b & mask) values.unshift(this._items[i]);
}
```

packages/core/utils/src/BitSet.js
```diff
@@ -0,0 +1,121 @@
...
+
```
devongovett, line 123: You could do this in reverse order without the (probably expensive) string conversion on each iteration.

```js
let v = this._value;
for (let i = 0; v != 0n; i++, v >>= 1n) {
  if (v & 1n) values.unshift(this._items[i]);
}
```

mattcompiles: I originally went with this approach, but the trade-off is you end up iterating from `0..largestIndex` times. The current approach only iterates once per item in the set. This trade-off works in favour of large lists of potential values but small Set sizes. The string length hack is only needed because `Math.log2` isn't implemented for BigInts. 

We could potentially make a hybrid approach where if `this._value.toString(2).length < SOME_THRESHOLD` we use standard iteration?

devongovett: Did you measure which is faster? My guess was that bit shifting that many times would be faster that converting to a string since it doesn't require any memory allocation.

mattcompiles: Yeah I did. It's tricky as the length approach is faster, but only when the set sizes are really low. However when testing internal projects this trade-off seems to pay off. My project has around 68000 assets but averages about 50 internalised assets. 

This is modelling that in my benchmarks
```bash
values (string.length)
3.67ms avg | 739.17ms total | 200 runs

values (loop)
36.22ms avg | 7243.16ms total | 200 runs
```

devongovett: Ok, nice. Might be good to add a comment to the code explaining that!

mattcompiles: Will do 👍

mattcompiles: Done

devongovett (approved):

Issue #8785, opened 2023-01-23
Background scripts for Firefox result in build error when using Manifest V3


circuitwarden:
# 🐛 bug report

Firefox requires background scripts to be included in the manifest.json using 'background.scripts' instead of 'background.service_worker':

```
"background": {
  "scripts": ["background-script.js"]
}
```

Including the above 'scripts' object in the manifest file results in 'Invalid Web Extension manifest' build failure.

## 🎛 Configuration (.babelrc, package.json, cli command)

Error encountered when running command: 'npm run build'

```
{
	"mangle": false,
	"output": {
		"beautify": true,
		"indent_level": 2
	}
}
```

```
// package.json
{
	"private": true,
	"scripts": {
		"build": "parcel build source/manifest.json --no-content-hash --no-source-maps --dist-dir distribution --no-cache --detailed-report 0",
		"lint": "run-p lint:*",
		"lint-fix": "run-p 'lint:* -- --fix'",
		"lint:css": "stylelint source/**/*.css",
		"lint:js": "xo",
		"test": "run-p lint:* build",
		"watch": "parcel watch source/manifest.json --dist-dir distribution --no-cache --no-hmr"
	},
	"browserslist": [
		"last 1 Chrome version",
		"last 1 Firefox version"
	],
	"xo": {
		"envs": [
			"browser",
			"webextensions"
		]
	},
	"stylelint": {
		"extends": "stylelint-config-xo"
	},
	"dependencies": {
		"webext-base-css": "^1.4.1",
		"webext-options-sync": "^3.1.0"
	},
	"devDependencies": {
		"@parcel/config-webextension": "^2.6.2",
		"npm-run-all": "^4.1.5",
		"parcel": "^2.6.2",
		"process": "^0.11.10",
		"stylelint": "^14.9.1",
		"stylelint-config-xo": "^0.21.0",
		"xo": "^0.50.0"
	},
	"webExt": {
		"sourceDir": "distribution",
		"run": {
			"startUrl": [
				"https://github.com/fregante/browser-extension-template"
			]
		}
	},
	"@parcel/bundler-default-bug": "https://github.com/parcel-bundler/parcel/issues/8071",
	"@parcel/bundler-default": {
		"minBundles": 10000000
	}
}
```

## 🤔 Expected Behavior

Parcel does not require Chrome specific manifest v3 parameters, including 'background.service_worker'

## 😯 Current Behavior
[...]

PR #8906, opened 2023-03-23
[webextension] Add support for `background.scripts` and `background.page`

kuroppe1819:
# ↪️ Pull Request

- Open https://github.com/parcel-bundler/parcel/issues/8785

## 🚨 Test instructions

Add this to your Firefox Add-ons manifest:

```
"background": {
  "scripts": ["background.js"]
}
```

```
"background": {
  "page": index.html
}
```

## ✔️ PR Todo


packages/transformers/webextension/src/WebExtensionTransformer.js
```diff
@@ -266,91 +266,81 @@ async function collectDependencies(
...
-        }
-        asset.meta.webextBGInsert = program.background.service_worker;
+        env: {
+          context: 'service-worker',
+          sourceType: program.background.type == 'module' ? 'module' : 'script',
+        },
+      },
+    );
+  }
+  if (hot) {
```
101arrowz: There are now two separate blocks of HMR handling code. This is going to cause issues with HMR: Chrome requires different handling of v2 and v3 to do HMR properly. I'm not sure which version is needed for Firefox's version of Manifest V3. Could you try to investigate?

101arrowz: One way to fix the double-handling of `hot` would be to keep the `isMV2` check but copy the `background.scripts` code into the `else` block as well. But you'd still need to investigate if this HMR code actually works in Firefox.

101arrowz (changes requested):

packages/transformers/webextension/src/WebExtensionTransformer.js
```diff
@@ -306,50 +307,93 @@ async function collectDependencies(
...
-            ...getJSONSourceLocation(
-              ptrs['/background/service_worker'],
-              'value',
-            ),
-          },
-          env: {
-            context: 'service-worker',
-            sourceType:
-              program.background.type == 'module' ? 'module' : 'script',
+    if (program.browser_specific_settings?.gecko) {
```
kuroppe1819, line 310: 📝 Using `browser_specific_settings.gecko` to switch between Firefox and Chrome. Firefox add-on need an ID in their manifest.json when submitted to AMO.

> All Manifest V3 extensions need an add-on ID in their manifest.json when submitted to AMO.
https://extensionworkshop.com/documentation/develop/extensions-and-the-add-on-id/

But, it's not necessary in debugging. I want to know some useful flags.

packages/transformers/webextension/src/WebExtensionTransformer.js
kuroppe1819: Thanks for review! I have pushed the commit with the fix. https://github.com/parcel-bundler/parcel/pull/8906/commits/1fbbe8bd6f39b94ca2b4ebe2be2964231a800621

I investigated whether HMR works in Firefox and having trouble getting HMR to work in Firefox.

In Manifest V3, [the script-src directives may only have 'self', 'none' and 'wasm-unsafe-eval'. ](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/content_security_policy#manifest_v3_syntax)



Overwrite at the following value in the `dist/manifest.json` file after the build, It will work HMR when saving the files specified in the `background.scripts`.

```
  "content_security_policy": {
    "extension_pages": "script-src 'self';"
  },
```

But, occur the following errors when saving the files specified in the content scripts

Screen Shot 2023-05-10 at 2 39 55

It's possible that the reason is the stricter CSP constraints in Manifest V3.

> In Manifest V3, all CSP sources that refer to external or non-static content are forbidden. The only permitted values are 'none', 'self', and 'wasm-unsafe-eval'. In Manifest V2, a source for a script directive is considered secure if it meets these criteria:
https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/content_security_policy

101arrowz: That makes sense: Chrome and Firefox both treat MV3 the same with regards to the Content Security Policy. Instead of splitting the HMR logic into two blocks for Chrome and Firefox when `isMV2` is false, you'll need to use the Chrome version in both cases, but modify it such that if it's running in Firefox, it also checks for `background.scripts`.

sailerinteractive:
I learned that this should be fixed in Parcel >= 2.10.0 However, when im trying to build a Firefox V3 extension with Parcel 2.10.2 im still getting the "Missing property service_worker" error as before with older versions of parcel. Do i somehow need to specify for which browser i would like to bulld in order to activate this fix?

mischnic:
@sailerinteractive 

> Do i somehow need to specify for which browser i would like to bulld in order to activate this fix?

No

The current logic says that (among others) these two versions are possible

```json5
{
	"background": {
		"service_worker": "xyz"
	}
	...
}
```

```json5
{
	"background": {
		"scripts": ["xyz"]
	}
	...
}
```

What does your manifest look like?

The corresponding JSON schema sections

https://github.com/parcel-bundler/parcel/blob/c72c336f0f6e2b8bc41fcbac979f7d800b7ad0e2/packages/transformers/webextension/src/schema.js#L462-L478

https://github.com/parcel-bundler/parcel/blob/c72c336f0f6e2b8bc41fcbac979f7d800b7ad0e2/packages/transformers/webextension/src/schema.js#L79-L87

sailerinteractive:
Here is a shortened version of the manifest i'm using:

```json
{
    "manifest_version": 3,
    "name": "My Extension",
    "short_name": "MyE",
    "description": "",
    "homepage_url": "https://example.com",
    "version": "10.2",
    "background": {
      "scripts": ["javascript/sw.js"],
      "persistent": false,
      "type": "module"
    },
    "content_scripts": [
    	...
    ],
    "action": {
    	...
    },
    "icons": {
      	...
    },
    "permissions": [
    	...
    ],
    "web_accessible_resources": [ 
    	...
    ],
    "host_permissions": [
    	...
    ]
}
```

mischnic:
The schema didn't allow `scripts`  together with `type`: https://github.com/parcel-bundler/parcel/pull/9385
You could remove `type` as a workaround

sailerinteractive:
Thanks! I can confirm the extension builds when removing the type property. I'm not sure yet what implications that might have when running the extension in Firefox.

PR #8941, opened 2023-04-13
fix: check for source map flag instead of always emitting source map

abusarah-tech:
# ↪️ Pull Request

This pull request should fix an issue I had where I needed to build Sass files with `@parcel/transformer-sass` and disable the source map generation. 

## 💻 Examples

When you create a `.sassrc` file to build Sass files with `@parcel/transformer-sass` and add the following content
```
{
   "sourceMap": false
}
``` 
The flag will be ignored as the sourceMap is being always emitted for some reason.

### DIFF
--- a/packages/transformers/sass/src/SassTransformer.js
+++ b/packages/transformers/sass/src/SassTransformer.js
@@ -38,11 +38,13 @@ export default (new Transformer({
     }
 
     // Always emit sourcemap
-    configResult.sourceMap = true;
-    // sources are created relative to the directory of outFile
-    configResult.outFile = path.join(options.projectRoot, 'style.css.map');
-    configResult.omitSourceMapUrl = true;
-    configResult.sourceMapContents = false;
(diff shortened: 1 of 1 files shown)


mischnic:
Have you tried `parcel build --no-source-maps` instead? That should also achieve this

abusarah-tech:
@mischnic That worked but it also ignore all source maps including for transpiled JS files. How can I just ignore the source maps of the CSS files?

devongovett:
Why do you need to do that?

abusarah-tech:
@devongovett I want to be able to control certain parameters through `.sassrc` but this one property is being ignored from the transformer

devongovett:
I mean why do you want JS source maps but not CSS source maps?

abusarah-tech:
I want to keep the JS source maps for debugging purposes but ignore the generation of CSS source maps. Toggling it for both seems to be limiting to configure Sass transformer. Especially that the official Sass configuration file offers a specific property.

PR #8965, opened 2023-04-27
Fixup DiagnosticCodeHighlight and SourceLocation columns

mischnic:
It was always (supposed to be) the case that:
- `DiagnosticCodeHighlight` is 1-based and the end location is inclusive
- `SourceLocation` is 1-based and the end is exclusive 

But due to two cancelling off-by-one errors, many `SourceLocation` had an incorrect end position.

This caused problems when bumping Babel (causing more accurate sourcemaps) which caused some wrong locations to be remapped.


devongovett (approved):

devongovett:
Seems like a test is failing

Issue #7271, opened 2021-11-09
Parcel wipes out Package.json when installing dependencies 


wesbos:
Just lost a bunch of work I did on my package.json.

I am running parcel 2.0.1, which was released yesterday to fix an issue with dependencies not being installed.

I have targets in my package.json like this:

```json
    "css": {
      "source": [
        "./public/stylesheets/style-ARG.styl",
        "./public/stylesheets/account.styl"
      ],
      "distDir": "./public/stylesheets/build"
    },
```

When I run `parcel`, it says `Installing @parcel/transformer-stylus...`

Then it both wipes out everything in my `node_modules` directory, and wipes out everything in my package.json and replaces it with:

```json
{
  "devDependencies": {
    "@parcel/transformer-stylus": "^2.0.1"
  }
}
```

Probably something to do with the new way you are saving deps? 

Pre-installing `@parcel/transformer-stylus` does not fix it either, same thing. 

I assume this is also happening with sass and other transformers

devongovett:
Parcel doesn't edit package.json, it just runs a package manager like yarn or npm, which may edit the package.json to add the dependency. Not sure what would be causing it to be removed completely. Could you provide more details? Are you using Yarn or NPM? What versions? Are you in a monorepo? What directory are you running Parcel from? Is it the project root with the package.json/lockfile, a sub directory, or outside?

This sounds similar to #2978, which I thought was fixed, but maybe there's another case.

devongovett:
Btw you can run with the `--no-autoinstall` flag to prevent automatic installation from being run, which might be a workaround while we figure this out.

wesbos:
Thanks Devon, I'm running it from the root directory where the package.json is, npm 7.15.1, node 16.3, no monorepo.

running it with --no-autoinstall and @parcel/transformer-stylus installed, gives me this error:

```
  > 24 |     "*.{styl,stylus}": ["@parcel/transformer-stylus"],
  >    |                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Cannot find module "@parcel/transformer-stylus
```

devongovett:
Hmm strange. So I made a example project with these files:

package.json:
```json
{
  "name": "test",
  "scripts": {
    "build": "parcel build test.styl"
  },
  "dependencies": {
    "parcel": "^2.0.1"
  }
}
```

test.styl:
```stylus
.foo
  background: red
```

[...]

PR #8969, opened 2023-04-28
Fix resolveOptions to use project root instead of disk root (#7271)

### LINKED ISSUE #8846: Autoinstall gets confused by symlinks and things break badly
Hi everybody, I'm having an issue which I think might in part be a bug in Parcel, so I thought I'd ask and potentially report here. I have a somewhat involved setup, but I think I can figure out a reasonable approximation. Consider the following folder structure:

```
shared/
  uploads/
    image.jpg
releases/
  1/
    site/
      index.html
      uploads -> absolute symlink to shared/uploads
    node_modules/
    package.json
current -> absolute symlink to releases/1
```

The `index.html` file contains this: ``.

When I run `parcel build "site/**/*.html"` within `current` (which really means within `releases/1`), weird things happen. Some are quite understandable, some less so:

 - Parcel resolves `uploads/image.jpg` to the real absolute path to `shared/uploads/image.jpg`
 - Then it tries to resolve `sharp` from the image's absolute path, which obviously fails, because `node_modules` and `package.json` are on a different subtree of the project structure
 - Then it tries to autoinstall `sharp` and gets stuck on that

I understand everything that happens up to the moment Parcel tries autoinstalling `sharp` - it's logical why module resolution would need to resolve things relative to the _real_ path of the source file instead of the symlink that led to it, and while it does throw quite a wrench in my setup, I'll figure out a way to work around it somehow (although I'm quite open to suggestions!).

But the really unsettling thing is what happens _during_ the attempt to install `sharp`. Not only does the build get stuck, but _something_ happens somewhere within there that causes my `node_modules` to get (at least partially) deleted - when I kill the Parcel build, I don't even _have_ a `parcel` directory within `node_modules` anymore (and quite a few others are missing as well).

I've found [this issue](https://github.com/parcel-bundler/parcel/issues/7271), which seems related, but it appears the consensus there is a rogue `package.json` somewhere up the directory tree - and I've checked for this, and found no such file on my system (nor `package-lock.json`, nor `yarn.lock`, nor `node_modules`, for that matter; neither in the tree leading up to the project directory, nor in `$HOME`). The Node installation on my system is owned by `root` and not writable for anyone else, so the `npm install` which Parcel is running behind the scenes can't be attempting to install the package globally (it would fail instantly if it did).

I don't know if there's anything to be done about this within Parcel, really - the only idea I had was that Parcel could attempt to resolve build tools like `sharp` from the current working directory, if resolving them from the file which needs them fails - and it could be argued that they're really a dependency of the _build tool_, ie. Parcel or one of Parcel's transformers, not the files they're supposed to work on - but I understand if that's not desirable. I thought I'd at least write this down in case somebody runs into the same issue. Again, if anyone has any tips on how to solve situations like these, I'm all ears!

Parcel: 2.8.3, NPM: 9.5.0, Node: 18.14.1, OS: Ubuntu Server 22.04, not that I think any of that matters in this situation.

EDIT: To be clear, the aspect I think might _possibly_ be a Parcel bug is the fact that autoinstalling a dependency deletes files it shouldn't delete - I know it's very likely not Parcel itself doing that, but seeing as Parcel _is_ triggering the process which is actually responsible, I think it is worth considering adding a check in Parcel for some of the common situations which might cause this to happen.

Kasea:
Relates to (and closes?) #7271 (and similar)
Closes #8846 

Changes resolveOptions function to use `path.parse().dir` as opposed to `path.parse().root` see [docs for more information](https://nodejs.org/api/path.html#pathparsepath).

With the old code, it would resolve to `C:/`, `D:/`, ... etc instead of the root we're interested in. This caused it to climb the tree as opposed to descend it, which causes numerous problems.

In my case it would break my pnpm monorepo setup when trying to run/build with parcel. It would select my monorepo root (the first case of a `lock` file from `D:/`). I noticed I could solve it by putting _any_ `yarn.lock` or `package-lock.json` inside of my app folder and it would select the correct root and build and start properly. This is due to the array `LOCK_FILE_NAMES` having `pnpm-lock.yaml` last.

## 💻 Examples

├─apps
│  └─examples
│      ├─package-lock.json
├─pnpm-lock.yaml
└─pnpm-workspace.yaml


Kasea:
@mischnic think it would be possible to have this released for next minor release? I'm fairly sure it should be a "patch" release, but some people might've done weird hacks to get around this issue, so don't want to go and potentially break those.

Kasea:
Sratch my previous comment, I no longer have any interest in Parcel; it's broken beyond repair.

I will leave the PR up in case someone wants to try and salvage this project

devongovett:
Sorry to hear that, what other issues did you have?

For this PR, I'm not sure this is right because the entry root (i.e. common directory between the entries), may not be the project root. For example, if the entries are `/path/to/project/website/pages/a.js` and `/path/to/project/website/pages/b.js`, then the common directory is `/path/to/project/src/website/pages`, and `path.parse('/path/to/project/src/website/pages').dir` is `'/path/to/project/src/website'`. But the project root should be `'/path/to/project'`. That's why we search upward from where the common path is up until we find a lock file or SCM directory, which indicates the project root.

Kasea:
won't get into it here and going backwards makes more sense than going upwards (you won't detect random files in the tree). You're more likely to have a random package-lock going up the tree than going down as the project itself will be checked first, which also gives you the option to debug faster and actually change how things get handled without changing things outside of the project.

PR #8981, opened 2023-05-03
Fix missing asset errors in runtime manifest bundles

mattcompiles:
# ↪️ Pull Request

This PR fixes an issue where runtime manifest bundles could cause bundles to be missing from the register code, causing a runtime error (`Error: Could not resolve bundle with id abKQ5`).

This issue would occur when a runtime manifest bundle was split out from the bundle group entry but one (or more) of it's parents runtimes had already been processed, causing the new bundle to be missing from the register code. 

Previously we were processing the bundles in reverse order to handle this issue, but that only minimised the issue and didn't eliminate it completely. The fix here is to process the bundles in reverse topological order which means all child bundles will be processed before their parents. 

## 🚨 Test instructions

As this is just a slight modification of the existing code I haven't added more tests scenarios.

## ✔️ PR Todo

### DIFF
--- a/packages/core/core/src/applyRuntimes.js
+++ b/packages/core/core/src/applyRuntimes.js
@@ -66,10 +66,15 @@ export default async function applyRuntimes<TResult>({
   let runtimes = await config.getRuntimes();
   let connections: Array<RuntimeConnection> = [];
 
-  // As manifest bundles may be added during runtimes we process them in reverse order.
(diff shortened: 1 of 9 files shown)


mischnic:
Is there some easy simple case you can add? Runtimes will (hopefully) be overhauled at some point in the future and this might break again

mattcompiles:
@mischnic Added a test case mate 👍

mischnic (approved):

Issue #8864, opened 2023-03-01
Bundles pointing to assets in the wrong dist when building with multiple targets


joeyslater:
# 🐛 bug report

Heya folks! I was testing the nightlies to take advantage of https://github.com/parcel-bundler/parcel/pull/8854 when I noticed the bundles were pointing to assets the other entry's dist. The node package being install will import a couple js / css files.

Put together a small repro here: https://github.com/joeyslater/parcel-entries-bug

Even though the repo is usingly the nightly, I also tested with 2.8.3 patching the removed edge fix.

## 🎛 Configuration (.babelrc, package.json, cli command)

Nothing too complex, just two targets pointing to two index sources that should build into the same named dist. 

```js
{
  "targets": {
    "main1": {
      "distDir": "./dist/main1",
      "source": "./src/ui/main1/index.html",
      "publicUrl": "./"
    },
    "main2": {
      "distDir": "./dist/main2",
      "source": "./src/ui/main2/index.html",
      "publicUrl": "./"
    }
  }}
```

## 🤔 Expected Behavior

I am guessing that it should be 
in `./main1/index.<hash>.js` I am seeing `import("./../main1/"+a("27Lyk").resolve`
in `./main2/index.<hash>.js` I am seeing `import("./../main2/"+a("27Lyk").resolve`

## 😯 Current Behavior

in `./main2/index.<hash>.js` I am seeing `import("./../main1/"+a("27Lyk").resolve`
in `./main1/index.<hash>.js` I am seeing `import("./../main2/"+a("27Lyk").resolve`

## 💻 Code Sample

https://github.com/joeyslater/parcel-entries-bug

## 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Parcel           | nightly |
| Node             | 18.13.0 |
| npm/Yarn         |  1.22.19 (Yarn 3.3.1 as well)| 
| Operating System | Apple M1 Pro @ 13.1 (22C65) |

PR #8991, opened 2023-05-06
Fix multiple entries pointing to wrong bundle in dist

AGawrys:
# ↪️ Pull Request

This aims to fix issue #8864 , Where specifying two dists, with an external shared bundle caused the wrong relative path to be placed in the loader code. I don't think this is an issue with the bundler algorithm though, since it still created separate shared bundles for each dist and the edges were fine. (see graph below) 

~Anyway, so it looks like the wrong path occurs when getting `externalBundles` in the runtime code. It grabs both shared bundles (for each target) and then picked the first match by asset, which didn't take into account targets. Not sure if this fix is best, so do let me know.~

Update: I've added the target name when keying for bundlegroups, which actually addresses the problem of having two identical bundles (same entry asset) under the same bundlegroup.
## 💻 Examples
BEFORE:
parcel-after_bundle
AFTER:
parcel-after_fix

## 🚨 Test instructions

Test case and @joeyslater 's reproduction 

## ✔️ PR Todo


packages/runtimes/js/src/JSRuntime.js
```diff
@@ -347,6 +347,10 @@ function getLoaderRuntime({
   }
 
   let externalBundles = bundleGraph.getBundlesInBundleGroup(bundleGroup);
+  externalBundles = externalBundles.filter(
+    externalBundle => externalBundle.target == bundle.target,
+  );
```
devongovett: One question is how bundles from multiple different targets are getting into the same bundle groups in the first place?

AGawrys: I believe I fixed it properly now 😅 @devongovett , added an updated graph pic as well where theres two separate bundlegroups for each target :) 
Only issue here could be if targets don't have a name but I thought they would always have one? Is that true?

mattcompiles (approved):
Nice one 🍻 🚀

PR #9000, opened 2023-05-10
Bumped version of "eslint-plugin-import" to support workspaces and type imports

alshdavid:
# ↪️ Pull Request

I noticed when installing Parcel via `pnpm`, I would get various warnings telling me there are missing peer dependencies. The exact missing dependencies seem to vary depending on the project.

image

In an effort to silence these warnings, I began investigating how we protect against imports that are not defined in the `package.json` of a project package.

Currently, we use a version of the `eslint-plugin-import` plugin that does not support workspaces and does not support type imports.

This PR bumps the version of the `eslint-plugin-import`, but leaves the configuration for workspaces and type imports disabled.

_(It's a bit tricky determining what is a `dependency`, `peerDependency` and/or `devDependency` so I figured that's something for a following PR)._

Currently we have 339 normal imports missing from their relevant `package.json`s and 60 type imports that are missing.

## ✔️ PR Todo

### DIFF
--- a/packages/dev/eslint-config/index.js
+++ b/packages/dev/eslint-config/index.js
@@ -45,7 +45,17 @@ module.exports = {
     '@parcel/no-self-package-imports': 'error',
     'import/first': 'error',
     'import/newline-after-import': 'error',
-    'import/no-extraneous-dependencies': 'error',
+    'import/no-extraneous-dependencies': [
+      'error',
+      {
+        // "includeTypes": true,
+        // "packageDir": require('../../../package.json')
+        //   .workspaces
+        //   .map(workspace => require('glob').sync(workspace, {
+        //     cwd: require('path').join(__dirname, '..', '..', '..')
+        //   })),
+      },
+    ],
     'import/no-self-import': 'error',
     'no-prototype-builtins': 'off',
     'no-console': 'error',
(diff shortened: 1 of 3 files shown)


packages/dev/eslint-config/index.js
```diff
@@ -45,7 +45,17 @@ module.exports = {
     '@parcel/no-self-package-imports': 'error',
     'import/first': 'error',
     'import/newline-after-import': 'error',
-    'import/no-extraneous-dependencies': 'error',
+    'import/no-extraneous-dependencies': [
+      'error',
+      {
+        // "includeTypes": true,
```
alshdavid, line 51: This enables detection of `import type {} from 'foo'` imports

packages/dev/eslint-config/index.js
```diff
@@ -45,7 +45,17 @@ module.exports = {
...
+        // "packageDir": require('../../../package.json')
+        //   .workspaces
+        //   .map(workspace => require('glob').sync(workspace, {
+        //     cwd: require('path').join(__dirname, '..', '..', '..')
+        //   })),
```
alshdavid, line 56: `packageDir` takes an array of packages to check. This is used to manually specify the locations of workspace packages. I do this by parsing the workspaces property on the workspace `package.json` and resolving the packages with `glob`

devongovett (approved):

PR #9014, opened 2023-05-16
Fix `inlineEnvironment` perf

mattcompiles:
# ↪️ Pull Request

Currently, when the `inlineEnvironment` config in the JS transformer is used with an array value, it does a glob match against every environment variable in the process. This leads to a large drop in performance as it is performed for every JS file in the build. 

`micromatch`, which is used for our glob matching, actually supports many to many matching which is much more performant for this type of case. This PR exposes a new glob API (`globMatch`) which is now used in the JS transformer. 

Just change took a local build completion time from `~140s` to  `~70s`. 

While this is good, I do wonder if there's a way to optimize this further as this work should only be done once per build? It is currently done once per file.

## 🚨 Test instructions

No new tests as functionality hasn't changed. 

## ✔️ PR Todo

### DIFF
--- a/packages/core/utils/src/glob.js
+++ b/packages/core/utils/src/glob.js
@@ -5,7 +5,7 @@ import type {FileSystem} from '@parcel/fs';
 
 import _isGlob from 'is-glob';
 import fastGlob, {type FastGlobOptions} from 'fast-glob';
-import {isMatch, makeRe, type Options} from 'micromatch';
+import micromatch, {isMatch, makeRe, type Options} from 'micromatch';
 import {normalizeSeparators} from './path';
 
 export function isGlob(p: FilePath): any {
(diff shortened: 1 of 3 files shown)


mischnic:
Some alternative ideas (the last two assuming that the proportion of assets using env variables is tiny):
- As you've said, something like `loadConfig` that runs once per build for projectRoot-level settings.
- Don't pass everything to Rust but instead use a JS callback to get used variables on demand
- Pass all env vars (as in the `else` case) and `config?.inlineEnvironment` to Rust, and make the glob matching there (also on demand)

devongovett (approved):

mattcompiles:
As discussed, I'm merging this as is and will follow up with global config loading for transformers.

PR #9016, opened 2023-05-16
Avoid leaking local types by the generated declarations

Andarist:
# ↪️ Pull Request

When working on a PR [here](https://github.com/bvaughn/react-resizable-panels/pull/148), I noticed that some type **stopped** being exported. This seemed quite odd because they were never exported in the first place.

We can verify this both by checking `src/index.ts` [here](https://github.com/bvaughn/react-resizable-panels/blob/5aafa03f30778b896e5bc9b61ad5fca2ad542251/packages/react-resizable-panels/src/index.ts) and by checking the generated `.d.ts` [here](https://unpkg.com/browse/react-resizable-panels@0.0.45/dist/react-resizable-panels.d.ts). 

## 🚨 Test instructions

`git clone git@github.com:bvaughn/react-resizable-panels.git && corepack pnpm i && pnpm run prerelease && pnpm run typescript`

With this script, I'd expect the typechecking to trip over [this import](https://github.com/bvaughn/react-resizable-panels/blob/5aafa03f30778b896e5bc9b61ad5fca2ad542251/packages/react-resizable-panels-website/src/utils/UrlData.ts#L20) (and some others as well). If we apply my patch then we will witness the expected behavior.

## ✔️ PR Todo

Note: I could use some pointers as to where I should put some tests for this and how they should look like (unit, integration, etc).

### DIFF
--- a/packages/transformers/typescript-types/src/TSTypesTransformer.js
+++ b/packages/transformers/typescript-types/src/TSTypesTransformer.js
@@ -165,6 +165,7 @@ export default (new Transformer({
 
     let code = nullthrows(host.outputCode);
     code = code.substring(0, code.lastIndexOf('//# sourceMappingURL'));
+    code += `\nexport {};\n`;
 
     let map = JSON.parse(nullthrows(host.outputMap));
     map.sources = map.sources.map(source =>


packages/transformers/typescript-types/src/TSTypesTransformer.js
```diff
@@ -165,6 +165,7 @@ export default (new Transformer({
 
     let code = nullthrows(host.outputCode);
     code = code.substring(0, code.lastIndexOf('//# sourceMappingURL'));
+    code += `\nexport {};\n`;
```
Andarist, line 168: You are using `outFile` [here](https://github.com/Andarist/parcel/blob/64099c070a8a76e50b0845fbd3f58784ffa06338/packages/transformers/typescript-types/src/TSTypesTransformer.js#L37) and that alters TS behavior when it comes to the generated code as it tries to concatenate files. On top of that, the emitted declarations assume a script mode (or something akin to it), and thus `export {};` is not appended to them, and thus all local declarations leak.

I understand that this might be seen as a breaking change so it's up to you to decide if this is a good change. I believe though that the change itself is good. As a library author, I want to control what I export and what I don't, I don't want a tool to leak things that were meant to be private.

If you decide to pick up this change then it's likely that [`shake.js`](https://github.com/parcel-bundler/parcel/blob/64099c070a8a76e50b0845fbd3f58784ffa06338/packages/transformers/typescript-types/src/shake.js) should be changed as well. After all, non-exported types could be pruned along with their otherwise unused dependencies.

Andarist: Actually, I tested this more and no changes to `shake.js` would be required. Those leaked types were leaked because they were still used by other retained things. So shaking works as expected.

devongovett: Could you explain more what you mean by leak? If they weren't exported in the type definitions, why would TypeScript allow importing them? And why would an empty export statement fix this? I have seen such a thing when there are no other exports in the file used to "tell" tools that something is ESM, but in this case, there _are_ other exports already so I would assume that it isn't necessary?

Andarist: > Could you explain more what you mean by leak? 

You can see in this [TS playground](https://www.typescriptlang.org/play?moduleResolution=3&ts=5.0.4#code/JYWwDg9gTgLgBAbzgBQIYDsCmAbA4lCAVzAHl0AZVATyPgF84AzAkOAIik1QGMYBaTgGdgAL1QAjbJj5gMOQWwDccAPQq4MKmEyCAXHAAMAOmMAWAKwAoS5u1wAgnAC8KOXgLEylGoRiLLanBBAHoA-JZAA) that the `PanelGroupOnLayout` is usable/importable and cross-reference it with the generated `.d.ts` to see that it has no `export` modifier nor it's exported through any other statement~: https://unpkg.com/browse/react-resizable-panels@0.0.45/dist/react-resizable-panels.d.ts#L7

> If they weren't exported in the type definitions, why would TypeScript allow importing them?

That's a good question. It's related to compatibility with declarations written in times when ESM was being born and stuff (that's my guess). It seems that `export` alone in **declaration file** doesn't truly make those declaration files treated as modules. I think that in the past the ESM-like syntax might have been reused for declaration files without its full semantics.

I've made some tests and it seems that you either need `export {}` (or `export { name }`) or `export { name } from './file.js'` or `export * from './file.js'` to make it work "as expected". Since you "flat bundle" declaration files through the `outFile`-based logic... you don't really have references to any other files in the output. So it creates this weird "leak".

I'm pretty sure that Andrew Branch from the TypeScript team could shed more light on this. But at the same time... it's probably not worth pinging him since this is how TypeScript behaves, we don't really have to understand it deeply - we just need to take this for granted. I totally agree though that this is weird and surprising.

Andarist: @devongovett friendly 🏓

PR #8631, opened 2022-11-14
fix for eslint rule protecting against importing packages that aren't in the package.json

alshdavid:
# ↪️ Pull Request

This PR corrects the misbehaving `eslint-plugin-import/no-extraneous-dependencies` rule, adding protection against using imports that are not defined in a package's package.json 

_Note:_ 
This currently excludes type imports like `import type { Foo } from 'foo'` due to the current version of [eslint-plugin-import](https://www.runpkg.com/?eslint-plugin-import@2.26.0/lib/rules/no-extraneous-dependencies.js#239) not including [the change which adds support for type imports](https://github.com/import-js/eslint-plugin-import/blob/main/src/rules/no-extraneous-dependencies.js#L272).

I am not sure when they will merge this, the last update was almost a year ago.

## 💻 Examples

```typescript
import('foo')

// Will fail to lint unless the closest "package.json" has "foo"
// in either the dependencies or devDependencies
```

## 🚨 Test instructions

As this is a linting rule, to test it you can try importing a package that isn't in that package.json to see if it fails

## ✔️ PR Todo

- [ ] Added/updated unit tests for this change
- [ ] Filled out test instructions (In case there aren't any unit tests)
- [ ] Included links to related issues/PRs

PR #9017, opened 2023-05-16
Bump some dependencies with "breaking" changes in patch versions

mischnic:
I copied some changes from #8631 that aren't related to undeclared dependencies (and caused the tests to fail once some deps were bumped in the lockfile)

Both of these happened in a patch version (that is covered by our dependency ranges):
- `fastest-levenshtein` added `__esModule: true`, so `import levenshtein from "..."` didn't work anymore.
- `xmldom` now correctly updates `element.parentNode` after removing the element:
```js
element.parentNode.removeChild(element);
element.parentNode.appendChild(...); // parentNode is null now
```
And for empty documents, `parsed.documentElement` is null

### DIFF
--- a/packages/transformers/xml/src/atom.js
+++ b/packages/transformers/xml/src/atom.js
@@ -34,11 +34,10 @@ export function content(
       break;
     case 'xhtml': {
       let fragment = element.ownerDocument.createDocumentFragment();
-      let child = element.firstChild;
-      while (child) {
+      let child;
+      while ((child = element.firstChild)) {
         element.removeChild(child);
         fragment.appendChild(child.cloneNode(true));
-        child = child.nextSibling;
       }
       contents = new XMLSerializer().serializeToString(fragment);
       break;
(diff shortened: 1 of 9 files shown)


packages/transformers/xml/src/atom.js
```diff
@@ -34,11 +34,10 @@ export function content(
       break;
     case 'xhtml': {
       let fragment = element.ownerDocument.createDocumentFragment();
-      let child = element.firstChild;
-      while (child) {
+      let child;
+      while ((child = element.firstChild)) {
         element.removeChild(child);
         fragment.appendChild(child.cloneNode(true));
-        child = child.nextSibling;
```
devongovett, line 41: so it only goes down the tree through firstChild instead of through all of the siblings of the element now?

mischnic: Previously, it removed all direct children of `element` (the first child and then its siblings).

Now, it repeatedly removes `element.firstChild` until that is null (= no more children, we are done).

devongovett (approved):

Issue #8519, opened 2022-10-05
Running out of heap memory caused by square brackets


JoshuaAmaju:
# 🐛 bug report

Running `parcel build` never completes and fails with the log show below. And increasing the heap memory (`--max-old-space-size`) didn't fix it.

## 🎛 Configuration (.babelrc, package.json, cli command)

```js
{
  "your": { "config": "here" }
}
```

## 🤔 Expected Behavior

The build should run fine with no build or memory issues

## 😯 Current Behavior

NodeJS runtime runs out of heap memory

```
<--- Last few GCs --->

[23194:0x7f814eb00000]   732134 ms: Scavenge 1947.5 (2001.5) -> 1940.6 (2005.0) MB, 8.8 / 0.0 ms  (average mu = 0.371, current mu = 0.368) task 
[23194:0x7f814eb00000]   732188 ms: Scavenge 1951.6 (2005.2) -> 1944.4 (2008.7) MB, 4.6 / 0.0 ms  (average mu = 0.371, current mu = 0.368) task 
[23194:0x7f814eb00000]   732234 ms: Scavenge 1954.5 (2008.7) -> 1947.7 (2012.2) MB, 4.8 / 0.0 ms  (average mu = 0.371, current mu = 0.368) task 


<--- JS stacktrace --->

FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory
 1: 0x10b6c4005 node::Abort() [/usr/local/bin/node]
 2: 0x10b6c4188 node::OnFatalError(char const*, char const*) [/usr/local/bin/node]
 3: 0x10b82e0f7 v8::Utils::ReportOOMFailure(v8::internal::Isolate*, char const*, bool) [/usr/local/bin/node]
 4: 0x10b82e093 v8::internal::V8::FatalProcessOutOfMemory(v8::internal::Isolate*, char const*, bool) [/usr/local/bin/node]
 5: 0x10b9cf345 v8::internal::Heap::FatalProcessOutOfMemory(char const*) [/usr/local/bin/node]
 6: 0x10b9d337b v8::internal::Heap::RecomputeLimits(v8::internal::GarbageCollector) [/usr/local/bin/node]
 7: 0x10b9cfc4c v8::internal::Heap::PerformGarbageCollection(v8::internal::GarbageCollector, v8::GCCallbackFlags) [/usr/local/bin/node]
[...]

PR #9020, opened 2023-05-16
Fix EntryResolver.resolveEntry infinite loop when entries look like globs

hacknlove:
# ↪️ Pull Request

https://github.com/parcel-bundler/parcel/issues/8519

## 💻 Examples

if you try to build any file whose path looks like a glob, like `/blog/[id].html` it will cause a infinite loop

## ✔️ PR Todo

~~I wasn't able to run test locally, I am having this error:~~

> ```
> /parcel/packages/core/core/src/AssetGraph.js: Cannot find module '@parcel/hash'
> ```

~~Even with v2 branch.~~

~~I couldn't find any solution. Maybe I need some guidance.~~

Also with the linter, It was erroring, but it was not showing any specific reason.

### DIFF
--- a/packages/core/core/src/requests/EntryRequest.js
+++ b/packages/core/core/src/requests/EntryRequest.js
@@ -165,23 +165,26 @@ export class EntryResolver {
   }
 
   async resolveEntry(entry: FilePath): Promise<EntryResult> {
-    if (await this.isTrulyGlob(entry)) {
-      let files = await glob(entry, this.options.inputFS, {
-        absolute: true,
-        onlyFiles: false,
-      });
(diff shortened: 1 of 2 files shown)


mischnic:
> Cannot find module '@parcel/hash'

You need to have Rust installed and then run `yarn build-native` in the monorepo root.

hacknlove:
I like better this last commit. I think it's simpler and faster, because it requires less stat checks operations.

But the previous one will do the job too.

joshamaju:
Any chance on reviewing and merging this?

mischnic (approved):
Thanks!

PR #9021, opened 2023-05-17
Load global config

mattcompiles:
# ↪️ Pull Request

This PR implements a new transformer hook called `loadGlobalConfig` that only runs once per build (rather than once per file). This is useful when you have a transform with project level configuration and only want to calculate it once. 

I have also updated the JS and CSS transformers to make use of the hook. The JS transformer is most impacted with the environment variable inline maps being calculated only once per build now. 

I had a few questions around implementation that I will leave as line specific comment below. 

## 🚨 Test instructions

I haven't added any tests yet as I wasn't sure how to approach testing this. However the existing tests for the JS transform should cover the JS specific functionality. I have also manually validated that the hook is being called when expected during re-builds. 

## ✔️ PR Todo

### DIFF
--- a/packages/core/core/src/Transformation.js
+++ b/packages/core/core/src/Transformation.js
@@ -77,6 +77,7 @@ import {
 import {invalidateOnFileCreateToInternal} from './utils';
 import invariant from 'assert';
 import {tracer, PluginTracer} from '@parcel/profiler';
+import {createBuildCache} from './buildCache';
 
 type GenerateFunc = (input: UncommittedAsset) => Promise<GenerateOutput>;
 
@@ -100,9 +101,14 @@ export type TransformationResult = {|
   devDepRequests: Array<DevDepRequest>,
 |};
 
+// Global configs are not file-specific, so we only need to
+// load them once per build.
+const globalConfigs = createBuildCache();
+
 export default class Transformation {
   request: TransformationRequest;
   configs: Map<string, Config>;
+  globalConfigs: Set<Config>;
   devDepRequests: Map<string, DevDepRequest>;
   pluginDevDeps: Array<InternalDevDepOptions>;
   options: ParcelOptions;
@@ -115,6 +121,7 @@ export default class Transformation {
 
   constructor({request, options, config, workerApi}: TransformationOpts) {
     this.configs = new Map();
+    this.globalConfigs = new Set();
     this.parcelConfig = config;
     this.options = options;
     this.request = request;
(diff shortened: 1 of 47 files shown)


packages/core/core/src/Transformation.js
```diff
@@ -714,22 +724,52 @@ export default class Transformation {
...
+        this.options,
+      );
+
+      globalConfigs.set(transformer.name, globalConfig);
+    }
+
+    if (globalConfig) {
+      // I believe this should be done per file?
+      for (let devDep of globalConfig.devDeps) {
+        await this.addDevDependency(devDep);
```
mattcompiles, line 753: I assumed I should be adding dev deps for all files using the global config even though it's already loaded. Just confirming this is correct?

packages/transformers/js/src/JSTransformer.js
```diff
@@ -152,6 +152,90 @@ type TSConfig = {
...
+        Object.keys(options.env),
+        inlineEnvironment,
+      )) {
+        publicEnv[match] = String(options.env[match]);
+      }
+    } else if (inlineEnvironment === true) {
+      // inlineEnvironment: true
+      // All files can inject all ENV vars
+      assignAllEnv(publicEnv);
+    } else if (inlineEnvironment === false) {
```
mattcompiles, line 225: While I think the existing `inlineEnvironment` behaviour is a bit confusing, this is attempting to replicate it exactly. 

Specifically, I'm not a huge fan of default behaviour being different to all the possible config options. We can discuss the actual behaviour separately though.

packages/core/core/src/Transformation.js
```diff
@@ -714,22 +724,52 @@ export default class Transformation {
...
     isSource: boolean,
   ): Promise<?Config> {
+    // Only load global config for a transformer once per build.
+    let globalConfig = globalConfigs.get(transformer.name);
+    if (globalConfig == null && transformer.plugin.loadGlobalConfig != null) {
+      globalConfig = createConfig({
+        plugin: transformer.name,
+        searchPath: toProjectPathUnsafe('index'),
+        // Project root should be considered source?
+        isSource: true,
```
mattcompiles, line 737: Wasn't sure about this one?

packages/core/core/src/Transformation.js
```diff
@@ -205,6 +212,7 @@ export default class Transformation {
     let configRequests = getConfigRequests([
       ...this.configs.values(),
       ...this.resolverRunner.configs.values(),
+      ...this.globalConfigs.values(),
```
mattcompiles, line 215: Just confirming this is correct?

PR #9024, opened 2023-05-18
Hoist exports to allow circular dependencies

AGawrys:
# 🤒 Pull Request

This PR aims to fix an issue where circular dependencies with a exported function is not defined at runtime. Both anonymous and named `export default function` should be hoisted, where as `export default class` **should not** as it is not supported by node.

## 😶‍🌫️  Examples

Without the fix, the test cases added below would result in a `function is not defined` or `b1() is not defined` type error.

## ✔️ PR Todo

- [x] Verified this behavior is supported by Node ESM
- [x] Verified that this behavior with `export default class` is **not** supported by Node ESM


mischnic:
Please remove the comments and the `debugger` call

mischnic (approved):
(I also renamed the tests to be more inline with the existing `test/integration/js-export-default-*` fixture folders)

PR #9045, opened 2023-05-26
Upgrade to svgo 3

### LINKED ISSUE #8948: Unmet peer dependency svgo 3 in optimizer-htmlnano
It seems htmlnano 2.0.4 used by optimizer-htmlnano wants svgo ^3.0.2 as a peer dep. optimizer-htmlnano loads svgo 2.8.0 instead. Please update accordingly.

mischnic:
Closes https://github.com/parcel-bundler/parcel/issues/8948

The breaking changes: https://github.com/svg/svgo/releases/tag/v3.0.0

- The plugin `cleanupIDs` was renamed to `cleanupIds`
- Some custom/third-party plugins might break as the API changed
- You need to use `overrides` instead of `extendDefaultPlugins` to change preset default settings

I'm still not entirely sure if this is safe to do...

### DIFF
--- a/packages/core/integration-tests/test/html.js
+++ b/packages/core/integration-tests/test/html.js
@@ -2910,7 +2910,11 @@ describe('html', function () {
 
     let output = await outputFS.readFile(b.getBundles()[0].filePath, 'utf8');
     assert(output.includes('<x-custom stddeviation="0.5"'));
-    assert(output.includes('<svg role="img" viewBox='));
+    assert(
+      output.includes(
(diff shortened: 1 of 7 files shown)


harryzcy:
@mischnic Is there any updates on this?

mischnic:
The question if still if we want to instead create a new plugin for svgo 3: https://github.com/parcel-bundler/parcel/pull/8802#issuecomment-1437856018

devongovett:
made another attempt at this in #9969, where I tried to detect if the config used any deprecated options and install the right version into the project...

PR #9079, opened 2023-06-07
Treat re-exports of '*' from empty files as ESM

lettertwo:
# ↪️ Pull Request

## Background

Many packages written in e.g., TypeScript are built for publishing to NPM by stripping types from the source files. This can lead to situations where ‘types-only’ files are stripped of all of their type definitions, leaving completely empty files behind. If there are imports from these ‘types-only’ files that are not also stripped (e.g., they re-export `*` from the module without the `type` keyword), then the resulting package ships with a dependency on a completely empty file.

## Current Behavior

This is a conundrum for Parcel, as it’s not clear what to do about the empty file when resolving. The current behavior is to treat these files as ‘dynamic CJS’ (to account for scenarios where `module.exports` might be modified outside of the module itself).

## The Problem

However, this is suboptimal for what is probably the most common case (as described in the first paragraph); ideally Parcel would treat these empty files as ESM with empty exports, and thus avoid the deopt to dynamic resolution altogether.

## Suggested Change

The compromise solution here is to treat these empty files as ESM **only if the asset is known to have no side effects** (e.g., the package defines `sideEffects: false`), and keep the 'dynamic CJS' behavior as the fallback.

## ✔️ PR Todo

### DIFF
--- a/packages/core/integration-tests/test/integration/scope-hoisting/es6/re-export-all-empty-no-side-effects/index.js
+++ b/packages/core/integration-tests/test/integration/scope-hoisting/es6/re-export-all-empty-no-side-effects/index.js
@@ -1,0 +1,3 @@
+import {bar} from 'lib';
+
+output = `foo ${bar}`;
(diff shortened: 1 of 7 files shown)


devongovett (approved):
I think this should be ok

mischnic (approved):

PR #9085, opened 2023-06-10
Enable automatic type directives for TypeScript Validator

### LINKED ISSUE #9072: TypeScript validator does not automatically find types
# 🐛 bug report

When I run `parcel watch`, incremental builds will attempt to validate everything in node_modules.

Run:
1. `npm run watch`
2. In another terminal `vim .parcelrc`
3. `ZZ` (save and exit)

See repo link below.

## 🤔 Expected Behavior

It should not validate `node_modules`.

## 😯 Current Behavior

```
jason@goodness> npm run watch                                                                                                           ~/src/pulumix/packages/pulumix-aws

> @bottech/pulumix-aws@0.1.0 watch
> parcel watch

🚨 Build failed.

@parcel/validator-typescript: Cannot find module 'http' or its corresponding type declarations.

  /Users/jason/src/pulumix/node_modules/aws-sdk/lib/config-base.d.ts:1:34
  > 1 | import {Agent as httpAgent} from 'http';
  >   |                                  ^^^^^^^ Cannot find module 'http' or its corresponding type declarations.
    2 | import {Agent as httpsAgent} from 'https';
    3 | import {AWSError} from './error';

@parcel/validator-typescript: Cannot find module 'https' or its corresponding type declarations.
```

## 🔦 Context

I'm using node workspaces but it still happens when I run `parcel watch` in a single workspace. The node_modules are still shared though so perhaps that is relevant.

## 💻 Code Sample

https://github.com/BotTech/pulumix/commit/b7ba4d737a11208a6b9d179725be28f926d7ca9e

## 🌍 Your Environment

| Parcel           | 2.9.1 |
| Node             | 18.16.0 |
| npm/Yarn         | 9.5.1 |
| Operating System | macOS 13.4 (Apple M1 Max) |

steinybot:
# ↪️ Pull Request

There is a difference between how the TypeScript Validator and tsc works. tsc automatically finds type directives without them needing to be specified in the `compilerOptions.types` setting of `tsconfig.json`

This is an important feature as TypeScript will check all imports and so your build can fail unless you manually specify all types that your dependencies use which is a maintenance nightmare.

This is done in [getAutomaticTypeDirectiveNames](https://github.com/microsoft/TypeScript/blob/v5.1.3/src/compiler/moduleNameResolver.ts#L768) and it requires the host to define both `directoryExists` and `getDirectories`.

The reason that the TypeScript Validator does not automatically find the type directives is because the `CompilerHost` does not define `getDirectories`.

Fixes #9072.

## 💻 Examples

`src/index.ts`:
```
// This should pass without explicitly listing node in compilerOptions.types.
// The types should be found automatically like they are for tsc.
export default function test(buffer: Buffer) {
  return;
}
```

`.parcelrc`:
```
{
  "validators": {
  }
}
```

`package.json`:
```
{
  "name": "@parcel/ts-typecheck-example2",
  "version": "2.9.1",
  "license": "MIT",
  "private": true,
  "module": "dist/index.js",
  "types": "dist/index.d.ts",
  "scripts": {
    "demo": "parcel build src/index.ts --no-cache",
    "check": "tsc --noEmit"
  },
  "devDependencies": {
    "@babel/core": "^7.0.0",
    "@parcel/babel-register": "2.9.1",
    "@parcel/core": "2.9.1",
    "@parcel/validator-typescript": "2.9.1",
    "@types/node": "18.16.0"
  }
}
```

`tsconfig.json`:
```
{
  "include": ["src/**/*"]
}
```

## 🚨 Test instructions

`yarn run check` and `yarn run demo` should both pass.

## ✔️ PR Todo


packages/examples/ts-typecheck-example2/.parcelrc
steinybot, line 1: I figure these probably shouldn't stay here but it is the easiest way to test this. How can I write a proper test for this?

Issue #8816, opened 2023-02-06
`parcel [serve] --lazy` fails to load an async bundle under certain conditions


rayinaway:
# 🐛 bug report

`parcel [serve]` fails to load an async bundle under the following conditions:

- `--lazy` flag is set.
- Along with the async bundle, the parent module statically imports some other module.
- Parcel cache has not been created yet.

These conditions seem exhaustive, a minimal repro below reflects that. After the cache is created, the next HMR invocation resolves the issue.

## 🎛 Configuration

Apart from the `--lazy` flag, out-of-the-box Parcel.

## 🤔 Expected Behavior

When the async bundle is requested, Parcel's development server responds with it.

## 😯 Current Behavior

When the async bundle is requested, the server responds with _`index.html`_ as a fallback instead, and a `SyntaxError` from failed parsing is thrown.

```
dynamic.51313f62.js:1 Uncaught SyntaxError: Unexpected token '<'
index.63aff760.js:61 Uncaught (in promise) Error: Cannot find module 'kmWTC'
    at newRequire (index.63aff760.js:61:19)
    at Function.newRequire (index.63aff760.js:45:18)
    at runtime-647fb96160fc59b6.js:1:248
```

## 💁 Possible Solution

Might be related to how Parcel initializes the cache.

## 🔦 Context

Faced the issue when tried to dynamically import app-specific devtools under certain conditions only.

## 💻 Code Sample

https://github.com/rayinaway/lazy-serve-parcel-bug-repro

## 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Parcel           | 2.8.3
| Node             | 16.13.2
| npm         | 8.1.2
| Operating System | Windows 10

marcins:
This should be fixed by #9093.

marcins:
This should be fixed by https://github.com/parcel-bundler/parcel/pull/9093.

PR #9093, opened 2023-06-14
Fix lazy compilation

marcins:
# ↪️ Pull Request

This fixes the reproduction provided in #8816 (Thanks @rayinaway), as well as a separate reproduction I created at https://github.com/marcins/lazy-testbed.

There were a couple of issues I found when trying to create a basic lazy compilation example:
* The decision as to whether a bundle is a placeholder isn't revisited after the bundle is created. A change has been made to set the placeholder value to false when a bundle is requested.
* Whether a bundle is a placeholder has been added to it's hash so that the bundle graph hash is invalidated when a bundle is no longer a placeholder.
* When a previously deferred dependency was no longer deferred, the asset graph was updated correctly, but this didn't re-trigger bundling with incremental bundling. A change has been made to skip incremental bundling when a dependency is "un-deferred".

A new integration test based on my testbed reproduction has been added that I first verified was failing on current `v2` branch, and verified that it now passes with these changes. In addition I tested the reproduction from #8816 and verified that Parcel with these fixes now loads those bundles correctly.

Note that due to my lack of deep knowledge in the areas touched the testing is possibly not exhaustive, I possibly haven't tested some scenarios that may still fail with these changes. I did manually test adding / removing dependencies and confirming that things re-compiled correctly.

Given the scope of the changes, they shouldn't affect anything outside of `--lazy` mode, so at least they're unlikely to break other functionality.

## 🚨 Test instructions

See repro repos above - these fail with current v2, but succeed when using Parcel linked with these changes.

## ✔️ PR Todo

### DIFF
--- a/packages/core/core/src/AssetGraph.js
+++ b/packages/core/core/src/AssetGraph.js
@@ -353,10 +353,12 @@ export default class AssetGraph extends ContentGraph<AssetGraphNode> {
         nodeId !== traversedNodeId
       ) {
         if (!ctx?.hasDeferred) {
+          this.safeToIncrementallyBundle = false;
           delete traversedNode.hasDeferred;
         }
         actions.skipChildren();
       } else if (traversedNode.type === 'dependency') {
+        this.safeToIncrementallyBundle = false;
         traversedNode.hasDeferred = false;
       } else if (nodeId !== traversedNodeId) {
         actions.skipChildren();
--- a/packages/core/core/src/Parcel.js
+++ b/packages/core/core/src/Parcel.js
@@ -321,6 +321,10 @@ export default class Parcel {
             };
           }
 
+          // If this bundle was previously a placeholder, it now no longer is as it has
+          // been requested.
+          bundleNode.value.isPlaceholder = false;
+
           for (let assetId of bundleNode.value.entryAssetIds) {
             this.#requestedAssetIds.add(assetId);
           }
--- a/packages/core/core/src/Parcel.js
+++ b/packages/core/core/src/Parcel.js
@@ -321,10 +321,6 @@ export default class Parcel {
             };
           }
 
-          // If this bundle was previously a placeholder, it now no longer is as it has
-          // been requested.
-          bundleNode.value.isPlaceholder = false;
-
           for (let assetId of bundleNode.value.entryAssetIds) {
             this.#requestedAssetIds.add(assetId);
           }
--- a/packages/core/integration-tests/test/lazy-compile.js
+++ b/packages/core/integration-tests/test/lazy-compile.js
@@ -7,6 +7,7 @@ import {
   getNextBuild,
   assertBundles,
   removeDistDirectory,
+  run,
 } from '@parcel/test-utils';
 
 const findBundle = (bundleGraph, nameRegex) => {
@@ -83,5 +84,40 @@ describe('lazy compile', function () {
 
     // Ensure the files match the bundle graph - lazy-2 should've been produced as it was requested
     assert(await distDirIncludes(['index.js', /^lazy-1\./, /^lazy-2\./]));
+  });
+
+  it('should lazy compile properly when same module is used sync/async', async () => {
+    const b = await bundler(
+      path.join(__dirname, '/integration/lazy-compile/index-sync-async.js'),
+      {
+        shouldBuildLazily: true,
+        mode: 'development',
+        shouldContentHash: false,
+      },
+    );
+
+    await removeDistDirectory();
+
+    const subscription = await b.watch();
+    let result = await getNextBuild(b);
+    result = await result.requestBundle(
+      findBundle(result.bundleGraph, /^index-sync-async\./),
+    );
+    result = await result.requestBundle(
+      findBundle(result.bundleGraph, /^uses-static-component\./),
+    );
+    result = await result.requestBundle(
+      findBundle(result.bundleGraph, /^uses-static-component-async\./),
+    );
+    result = await result.requestBundle(
+      findBundle(result.bundleGraph, /^static-component\./),
+    );
+
+    let output = await run(result.bundleGraph);
+    assert.deepEqual(await output.default(), [
+      'static component',
+      'static component',
+    ]);
+    subscription.unsubscribe();
   });
 });
--- a/packages/core/integration-tests/test/lazy-compile.js
+++ b/packages/core/integration-tests/test/lazy-compile.js
@@ -1,0 +1,87 @@
+import assert from 'assert';
+import path from 'path';
+import {
+  bundler,
+  outputFS,
+  distDir,
+  getNextBuild,
+  assertBundles,
+  removeDistDirectory,
+} from '@parcel/test-utils';
+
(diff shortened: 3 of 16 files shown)


packages/core/core/src/AssetGraph.js
```diff
@@ -353,10 +353,12 @@ export default class AssetGraph extends ContentGraph<AssetGraphNode> {
         nodeId !== traversedNodeId
       ) {
         if (!ctx?.hasDeferred) {
+          this.safeToIncrementallyBundle = false;
```
marcins, line 356: Do any of the other cases here need this added?

The symptoms I saw was that while the asset graph was correctly updated after a previously lazy bundle was requested, the bundle graph was not - so I've focused on cases where dependency nodes have changed from deferred to not deferred, rather than asset nodes.

packages/core/core/src/Parcel.js
```diff
@@ -321,6 +321,10 @@ export default class Parcel {
             };
           }
 
+          // If this bundle was previously a placeholder, it now no longer is as it has
+          // been requested.
+          bundleNode.value.isPlaceholder = false;
```
marcins: Not sure if this is the right place or way to do this - is it "safe" to mutate a `bundleNode` here?

packages/core/integration-tests/test/lazy-compile.js
```diff
@@ -0,0 +1,123 @@
...
+      },
+    ]);
+
+    subscription.unsubscribe();
+
+    // Ensure the files match the bundle graph - lazy-2 should've been produced as it was requested
+    assert(await distDirIncludes(['index.js', /^lazy-1\./, /^lazy-2\./]));
+  });
+
+  it('should lazy compile properly when same module is used sync/async', async () => {
```
marcins, line 89: This test still fails on this branch, it has been pulled in as an intentional broken test (hence the draft PR).

 The "fix" is to remove the conditions for `shouldBuildLazily` in `JSRuntime` (e.g.: https://github.com/parcel-bundler/parcel/blob/7ff54c7c1212b05ed0393e0a92e96c6002d1dc51/packages/runtimes/js/src/JSRuntime.js#L360-L369)

With the code as it is, the whole bundle group doesn't get loaded - so when the `uses-static-component-async` bundle loads, it doesn't trigger a load of the `static-component` bundle, so it can't resolve the module.

I know that  the"fix" (seen here: https://github.com/marcins/parcel/commit/a36a37d8add87b90d63fa941b759f4c3baf9dadb) means that there is a potential issue when a bundle group grows after lazy compilation - this means Parcel can get into a state where it requires a browser refresh in order to recover. However, with the current state, a refresh won't heal as the load order is just incorrect.

marcins: As discussed in the Core meeting, the "fix" for this has now been cherry-picked onto this branch. I've updated the comments to leave behind some trace that there was lazy build specific code previously there.

packages/core/core/src/Parcel.js
devongovett: hmm kinda weird but probably ok since we only store the bundle graph in the cache when parcel exits.

devongovett (approved):

packages/core/core/src/Parcel.js
marcins: Actually, I just had another look and I have a feeling this might no longer be required with the `this.safeToIncrementallyBundle = false;` fixes, as they'll cause the bundle graph to be recreated and the placeholder flag will be appropriately set.

My tests seem to pass without this line, I'll just validate it's the `safeToIncrementallyBundle` change that's fixing the issue.

marcins: Cool, I've removed this change as the other change (which I made later) in `AssetGraph` fixes the issue this was fixing.

jondlm:
Hey @marcins I tried out this patch but I'm still seeing some more issues with lazy mode. Here's another [minimal repro](https://github.com/jondlm/parcel-bug-lazy-shared-dep) that's exhibiting similar issues to what this PR was meant to fix. Maybe it'll be helpful to shake out more bugs! I can also file this as a separate issue if that's helpful.

I'm still trying to get comfortable with parcel internals to propose fixes but I'm not quite there yet.

Thank you for all your hard work on `--lazy`, I'm also quite interested in seeing it work properly on a large codebase we're (Stripe) piloting Parcel out on.

devongovett:
@jondlm did you build Parcel locally? This hasn't even been released in a nightly yet.

marcins:
> Hey @marcins I tried out this patch but I'm still seeing some more issues with lazy mode. Here's another [minimal repro](https://github.com/jondlm/parcel-bug-lazy-shared-dep) that's exhibiting similar issues to what this PR was meant to fix. Maybe it'll be helpful to shake out more bugs! I can also file this as a separate issue if that's helpful.

Thanks for that repro @jondlm - I haven't tried it, but looking at the description in the README that a refresh fixes it, I have a feeling you might've run into the issue which the conditions I removed in `JSRuntime` were trying to address.. with those conditions present, lazy compilation wouldn't work at all for our large application (even after refreshing), but without them it does work however I did intermittently see a "module not found" error that a refresh would fix. Presumably that's the issue the logic there was trying to avoid. 

As I couldn't find a reliable way to minimally repro this, we decided to merge the fixes for now as they definitely make things better than before, albeit requiring an occasional refresh to unstick things. Having a repro that fails with these fixes in place will definitely be handy in understanding the issue and working on a proper fix.

But just to confirm as Devon asked - you still see this issue by running a locally built Parcel with this patch applied? If that's the case, then yes please - do file a separate issue.

jondlm:
@devongovett I pulled commit 824ddbe70, compiled the JS runtime, copied it into my project's node_modules, and then ran Parcel via [bin.js](https://github.com/parcel-bundler/parcel/blob/v2/packages/core/parcel/src/bin.js). I _think_ that should pull in all the relevant changes from this PR.

Thanks for the context @marcins. I'll file a separate issue with the case I found 👍

marcins:
> I pulled commit [824ddbe](https://github.com/parcel-bundler/parcel/commit/824ddbe70ac5c513062a114eb6dc0f163e4d234c), compiled the JS runtime, copied it into my project's node_modules, and then ran Parcel via [bin.js](https://github.com/parcel-bundler/parcel/blob/v2/packages/core/parcel/src/bin.js). I _think_ that should pull in all the relevant changes from this PR.

You'll also need the changes in `AssetGraph` and `BundleGraph` to ensure bundles are rebuilt when they should be - though the symptoms of your issue (fixed by a refresh) definitely point to the JSRuntime part.

In any case, the latest `nightly` should now contain these fixes.

natan-stripe:
👋 I created a fork of your original lazy-testbed repo with the steps to reproduce the additional issue that @jondlm and I are bumping into which he mentioned above. We've spent some time trying to figure out how to fix it but haven't really been able to understand the expected behavior and usage of deferred modules properly.

It seems like it's _very_ close to working but there's just like one connecting piece that should be tying lazy async bundles to the relevant shared bundles that is maybe missing.

PR #9117, opened 2023-07-01
Query glibc version only once to speed up JSTransformer on Linux

pastelsky:
# ↪️ Pull Request

While profiling Parcel as it ran on a large app at Atlassian, I noticed that calls to `process.report.getReport()` stood out in the trace —

Screenshot 2023-07-02 at 1 28 48 AM

This function is in fact called once per transformation, to get the glibc version on linux and on an average takes 10ms per call. However, the version of glibc on a system isn't likely to change during a parcel run. Moving the call out so it is queried only once led to up to a ~35% improvement in overall build times on Linux.

This was tested on a 15 Core `c6i` AWS box, on Node `16.15.0`, with both the SWC and Babel transformers enabled. The % improvement might vary based on individual setups, however, it is still likely to be significant.

### DIFF
--- a/packages/transformers/js/src/JSTransformer.js
+++ b/packages/transformers/js/src/JSTransformer.js
@@ -940,23 +940,23 @@ export default (new Transformer({
 // See https://github.com/rust-lang/rust/issues/91979.
 let isLoadedOnMainThread = false;
 
-// $FlowFixMe
-let {glibcVersionRuntime} = process.report.getReport().header;
-
 async function loadOnMainThreadIfNeeded() {
   if (
     !isLoadedOnMainThread &&
     process.platform === 'linux' &&
     WorkerFarm.isWorker()
   ) {
+    // $FlowFixMe
+    let {glibcVersionRuntime} = process.report.getReport().header;
+
     if (glibcVersionRuntime && parseFloat(glibcVersionRuntime) <= 2.17) {
       let api = WorkerFarm.getWorkerApi();
       await api.callMaster({
         location: __dirname + '/loadNative.js',
         args: [],
       });
-
-      isLoadedOnMainThread = true;
     }
+
+    isLoadedOnMainThread = true;
   }
 }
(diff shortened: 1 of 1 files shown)


packages/transformers/js/src/JSTransformer.js
```diff
@@ -939,14 +939,16 @@ export default (new Transformer({
 // also load the native module on the main thread, so that it is not unloaded until process exit.
 // See https://github.com/rust-lang/rust/issues/91979.
 let isLoadedOnMainThread = false;
+
+// $FlowFixMe
+let {glibcVersionRuntime} = process.report.getReport().header;
+
 async function loadOnMainThreadIfNeeded() {
   if (
     !isLoadedOnMainThread &&
```
devongovett, line 945: hmm it's surprising to me that this is running more than once per thread given this check here. any idea why that's happening?

pastelsky: `isLoadedOnMainThread` is set to `true` only if the legacy glibc version condition is met — 

https://github.com/parcel-bundler/parcel/blob/db3bcae10497fa6a712fd9a135f93f26c5745454/packages/transformers/js/src/JSTransformer.js#L950-L957

Else, its called once per transform —

https://github.com/parcel-bundler/parcel/blob/db3bcae10497fa6a712fd9a135f93f26c5745454/packages/transformers/js/src/JSTransformer.js#L308-L314

I think moving the check outside might be better in that case?

```ts
    if (glibcVersionRuntime && parseFloat(glibcVersionRuntime) <= 2.17) {
      let api = WorkerFarm.getWorkerApi();
      await api.callMaster({
        location: __dirname + '/loadNative.js',
        args: [],
      });
    }

   isLoadedOnMainThread = true;
```

devongovett: Ah right

devongovett (approved):

pastelsky:
Moved the `isLoadedOnMainThread = true` out instead, which should be a neater way to do this.

pastelsky:
@devongovett do you mind approving the CI workflow whenever you can? Curious to see the effect this has on the parcel benchmark suite.

pastelsky:
Looks like the benchmark pipeline has been broken, and performance hasn't been published for a while according to —

> https://github.com/parcel-bundler/parcel-benchmark-action/pull/93#issuecomment-1541485353

So nevermind, feel free to merge this if all looks good.

PR #9131, opened 2023-07-13
Migrate to swc_core

marcins:
# ↪️ Pull Request

I wanted to pull in the fix for https://github.com/swc-project/swc/pull/7582, but it didn't appear to be published in the `swc_ecmascript` minor version we were using. Bumping the minor version to latest caused some other compilation issues, so I decided to yakshave.. 

Using `swc_ecmascript` etc directly is deprecated and the recommendation is to migrate to `swc_core` instead as a meta-pacakge that's semver stable.

This change makes that migration - it's mostly a case of updating a lot of imports, but also as a side-effect of bumping a few versions of SWC to pick up bugfixes there were some deprecated fields where code changes were needed.

One major change with the latest SWC is that it currently requires Rust nightly to build. I've updated the `rust-toolchain` to reflect this, but I'm not sure if there are any broader implications.


.github/workflows/ci.yml
```diff
@@ -15,6 +15,10 @@ jobs:
         with:
           cache: yarn
       - uses: actions-rs/toolchain@v1
+        with:
```
marcins, line 18: I tried using the `rust-toolchain` file support in this action for handling this, but couldn't get it working (in particular, this action doesn't support the [modern TOML format](https://rust-lang.github.io/rustup/overrides.html#the-toolchain-file) - it hasn't been updated for years).

Cargo.toml
```diff
@@ -1,4 +1,5 @@
 [workspace]
+resolver = "2"
```
mischnic, line 2: What was the reason for adding this?

marcins: Without this, you get this warning when building the JS Transformer: 
```
warning: some crates are on edition 2021 which defaults to `resolver = "2"`, but virtual workspaces default to `resolver = "1"`
note: to keep the current resolver, specify `workspace.resolver = "1"` in the workspace root's manifest
note: to use the edition 2021 resolver, specify `workspace.resolver = "2"` in the workspace root's manifest
```

From reading [the docs](https://doc.rust-lang.org/edition-guide/rust-2021/default-cargo-resolver.html), if everything is building with the 2021 resolver then it's all good.

mischnic (approved):

alshdavid (approved):

PR #9135, opened 2023-07-18
Bump oxipng from 6.0.0 -> 8.0.0

marcins:
# ↪️ Pull Request

This fixes an issue building for `aarch64-unknown-linux-musl` (e.g.: https://github.com/parcel-bundler/parcel/actions/runs/5433723752/jobs/9881567018) - where a C dependency of `oxipng`  (`cloudflare-zlib-sys`) was failing to compile. The latest version does not use this dependency.

## 🚨 Test instructions

If the integration tests still work, this should be okay.. the code change was only to remove a default.

I validated that `yarn build-release --target aarch64-unknown-linux-musl` in the image optimizer now works in the same Docker container as CI uses after applying this change, where previously it was failing.

## ✔️ PR Todo

- [n/a ] Added/updated unit tests for this change

### DIFF
--- a/packages/optimizers/image/src/lib.rs
+++ b/packages/optimizers/image/src/lib.rs
@@ -2,7 +2,7 @@ use mozjpeg_sys::*;
 use napi::bindgen_prelude::*;
 use napi::{Env, Error, JsBuffer, Result};
 use napi_derive::napi;
-use oxipng::{optimize_from_memory, Deflaters, Headers, Options};
+use oxipng::{optimize_from_memory, Headers, Options};
 use std::mem;
 use std::ptr;
 use std::slice;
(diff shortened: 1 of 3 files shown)


packages/optimizers/image/src/lib.rs
```diff
@@ -14,7 +14,6 @@ pub fn optimize(kind: String, buf: Buffer, env: Env) -> Result<JsBuffer> {
   match kind.as_ref() {
     "png" => {
       let options = Options {
-        deflate: Deflaters::Libdeflater,
```
marcins, line 17: This is no longer required as `Libdeflater` is the default: https://github.com/shssoichiro/oxipng/blob/b4e98a4dd16b370a99b31e0125dfa5854fc60dda/src/lib.rs#L313

mattcompiles (approved):

thebriando (approved):

PR #9137, opened 2023-07-18
[WIP] Remove package cycles

marcins:
# ↪️ Pull Request

Currently Parcel has several package cycles, all because of circular dependencies on `@parcel/types`.

This is problematic because currently if you have, say, a simple plugin that just depends on `@parcel/plugin` and `@parcel/types` you bring in a lot of transitive dependencies - like for example types depends on workers which depends on core as small example. Some are a bit worse because they bring in Rust based binaries.

This Draft PR removes dependencies from `@parcel/types` to any package that has a dependency on `@parcel/types`. This list was:

* @parcel/fs
* @parcel/cache
* @parcel/pacakge-manager
* @parcel/workers

Moving the types for FS, Cache, and Package Manager was relatively straightforward - the messier one was `@parcel/workers` - this is because currently the class for `Worker` and `WorkerFarm` are used as a type, and also there are a lot of required types to be moved and exported from `@parcel/types` to satisfy all of the dependencies.

This is a draft PR to foster some discussion on how to proceed, here are some open questions:

- [ ] Since `@parcel/types` is the public API, how can we mark things exported as "Not API"? For example, `WorkerFarm` is a property of the options object, which is why it was imported, but it's effectively an opaque type.. but also, because it's required by options, does that transitively actually make it API, and this cycle removal is actually just surfacing the underlying issue?
- [ ] Is there a better way to break the cycles without moving everything to `@parcel/types` for these packages?
- [ ] Changing `Worker`, `WorkerFarm`, and `Handle` to `interface`s instead of using the `class` as a type means we lose the `static` methods from the type signature. I don't think there's any way around that?

## 🚨 Test instructions

`yarn flow` has no errors.

### DIFF
--- a/packages/core/cache/src/FSCache.js
+++ b/packages/core/cache/src/FSCache.js
@@ -1,9 +1,8 @@
 // @flow strict-local
 
 import type {Readable, Writable} from 'stream';
-import type {FilePath} from '@parcel/types';
-import type {FileSystem} from '@parcel/fs';
-import type {Cache} from './types';
+import type {FilePath, FileSystem} from '@parcel/types';
+import type {Cache} from '@parcel/types';
 
 import stream from 'stream';
 import path from 'path';
--- a/packages/core/cache/src/IDBCache.browser.js
+++ b/packages/core/cache/src/IDBCache.browser.js
@@ -1,5 +1,5 @@
 // @flow strict-local
-import type {Cache} from './types';
+import type {Cache} from '@parcel/types';
 
 import {Readable} from 'stream';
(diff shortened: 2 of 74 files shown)


marcins:
On the "public" API question - it looks like re-exports currently aren't handled by the API generation docs, a patch would be needed as the current AST breaks the `generate-api-docs` tool in `parcel-website`.

This is a nice side-effect of putting those types extracted from other packages into separate modules, since they need to be re-exported to be accessed by the package that uses them, they don't end up in the Public API docs.

We could also use JSDoc to mark these types as `@private` for the IDEs that can surface that information.

PR #9172, opened 2023-08-01
Fix new ESM loader runtime by moving to abs URLs

mattcompiles:
# ↪️ Pull Request

This PR fixes the ESM loader runtime by making the bundle-manifest return absolute URLs. It also makes the following changes:
- Faster manifest registration by using a Map and entries list rather than an object
- Removes the ability to custom name runtime asset bundles
  - This is because it could break the absolute URL resolution

## ✔️ PR Todo

### DIFF
--- a/packages/core/core/src/applyRuntimes.js
+++ b/packages/core/core/src/applyRuntimes.js
@@ -50,8 +50,11 @@ function nameRuntimeBundle(
   let {hashReference} = bundle;
 
   let name = nullthrows(siblingBundle.name)
-    .replace(`.${bundle.type}`, `.runtime.${hashReference}.${bundle.type}`)
-    .replace(siblingBundle.hashReference, 'sib-hash');
+    // Remove the existing hash from standard file patterns
+    // e.g. 'main.[hash].js' -> 'main.js' or 'main~[hash].js' -> 'main.js'
+    .replace(new RegExp(`[\\.~\\-_]?${siblingBundle.hashReference}`), '')
+    // Ensure the file ends with 'runtime.[hash].js'
+    .replace(`.${bundle.type}`, `.runtime.${hashReference}.${bundle.type}`);
 
   bundle.name = name;
   bundle.displayName = name.replace(hashReference, '[hash]');
--- a/packages/core/core/src/applyRuntimes.js
+++ b/packages/core/core/src/applyRuntimes.js
@@ -6,7 +6,6 @@ import type {SharedReference} from '@parcel/workers';
 import type {
   Asset,
   AssetGroup,
-  Bundle,
   Bundle as InternalBundle,
   Config,
   DevDepRequest,
@@ -41,6 +40,23 @@ type RuntimeConnection = {|
   isEntry: ?boolean,
 |};
 
+function nameRuntimeBundle(
+  bundle: InternalBundle,
+  siblingBundle: InternalBundle,
+) {
+  // We don't run custom namers on runtime bundles as the runtime assumes that they are
+  // located at the same nesting level as they're owning bundle. Custom naming could
+  // be added in future as long as the custom name is validated.
+  let {hashReference} = bundle;
+
+  let name = siblingBundle.name
+    .replace(`.${bundle.type}`, `.runtime.${hashReference}.${bundle.type}`)
+    .replace(siblingBundle.hashReference, 'sib-hash');
+
+  bundle.name = name;
+  bundle.displayName = name.replace(hashReference, '[hash]');
+}
+
 export default async function applyRuntimes<TResult>({
   bundleGraph,
   config,
@@ -51,7 +67,6 @@ export default async function applyRuntimes<TResult>({
   previousDevDeps,
   devDepRequests,
   configs,
-  nameRuntimeBundle,
 }: {|
   bundleGraph: InternalBundleGraph,
   config: ParcelConfig,
@@ -62,7 +77,6 @@ export default async function applyRuntimes<TResult>({
   previousDevDeps: Map<string, string>,
   devDepRequests: Map<string, DevDepRequest>,
   configs: Map<string, Config>,
-  nameRuntimeBundle: (bundle: Bundle) => Promise<void>,
 |}): Promise<Map<string, Asset>> {
   let runtimes = await config.getRuntimes();
   let connections: Array<RuntimeConnection> = [];
@@ -152,7 +166,7 @@ export default async function applyRuntimes<TResult>({
               }
               bundleGraph.createBundleReference(bundle, connectionBundle);
 
-              await nameRuntimeBundle(connectionBundle);
+              nameRuntimeBundle(connectionBundle, bundle);
             }
 
             connections.push({
--- a/packages/core/core/src/applyRuntimes.js
+++ b/packages/core/core/src/applyRuntimes.js
@@ -45,7 +45,7 @@ function nameRuntimeBundle(
   siblingBundle: InternalBundle,
 ) {
   // We don't run custom namers on runtime bundles as the runtime assumes that they are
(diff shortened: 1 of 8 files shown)


packages/core/core/src/applyRuntimes.js
```diff
@@ -41,6 +40,23 @@ type RuntimeConnection = {|
   isEntry: ?boolean,
 |};
 
+function nameRuntimeBundle(
+  bundle: InternalBundle,
+  siblingBundle: InternalBundle,
+) {
+  // We don't run custom namers on runtime bundles as the runtime assumes that they are
+  // located at the same nesting level as they're owning bundle. Custom naming could
```
marcins:
```suggestion
  // located at the same nesting level as their owning bundle. Custom naming could
```

packages/runtimes/js/src/JSRuntime.js
```diff
@@ -632,39 +624,42 @@ function getRegisterCode(
...
-  return (
-    "require('./helpers/bundle-manifest').register(JSON.parse(" +
-    JSON.stringify(JSON.stringify(idToName)) +
-    '));'
-  );
-}
+  let baseUrl =
+    entryBundle.env.outputFormat === 'esmodule' &&
+    entryBundle.env.supports('import-meta-url')
+      ? 'new __parcel__URL__("").toString()' // <-- this isn't ideal. We should use `import.meta.url` directly but it get's replaced currently
```
marcins:
```suggestion
      ? 'new __parcel__URL__("").toString()' // <-- this isn't ideal. We should use `import.meta.url` directly but it gets replaced currently
```

packages/runtimes/js/src/helpers/bundle-manifest.js
```diff
@@ -1,18 +1,17 @@
-var mapping = {};
+var mapping = new Map();
 
-function register(pairs) {
-  var keys = Object.keys(pairs);
-  for (var i = 0; i < keys.length; i++) {
-    mapping[keys[i]] = pairs[keys[i]];
+function register(baseUrl, pairs) {
+  for (var i = 0; i < pairs.length; i++) {
+    mapping.set(pairs[i][0], {baseUrl, path: pairs[i][1]});
```
marcins: What's the benefit of changing this to a map?

packages/core/core/src/applyRuntimes.js
```diff
@@ -41,6 +40,23 @@ type RuntimeConnection = {|
...
+  // be added in future as long as the custom name is validated.
+  let {hashReference} = bundle;
+
+  let name = siblingBundle.name
+    .replace(`.${bundle.type}`, `.runtime.${hashReference}.${bundle.type}`)
+    .replace(siblingBundle.hashReference, 'sib-hash');
```
marcins: Does `sib-hash` get replaced somewhere?

marcins (approved):

packages/runtimes/js/src/helpers/bundle-manifest.js
marcins: Never mind, I read the PR description :P
> Faster manifest registration by using a Map and entries list rather than an object

packages/runtimes/js/src/helpers/bundle-manifest.js
```diff
@@ -1,18 +1,17 @@
-var mapping = {};
+var mapping = new Map();
```
devongovett, line 1: Are we ok to drop browsers that don't support `Map`? I think Parcel's output still worked in IE11 up until this point. Maybe it's ok, but wonder if it should be done in a major version?

mattcompiles: I believe Map is supported in IE11. This support coverage looks good enough to me. But I can remove if you want.
https://caniuse.com/mdn-javascript_builtins_map

packages/core/core/src/applyRuntimes.js
mattcompiles: This is just a placeholder implementation, we'll need to do something smarter here. Sorry if that wasn't clear.

packages/runtimes/js/src/JSRuntime.js
```diff
@@ -632,39 +624,42 @@ function getRegisterCode(
...
-  return (
-    "require('./helpers/bundle-manifest').register(JSON.parse(" +
-    JSON.stringify(JSON.stringify(idToName)) +
-    '));'
-  );
-}
+  let baseUrl =
+    entryBundle.env.outputFormat === 'esmodule' &&
+    entryBundle.env.supports('import-meta-url')
+      ? 'new __parcel__URL__("").toString()' // <-- this isn't ideal. We should use `import.meta.url` directly but it gets replaced currently
```
mattcompiles, line 660: @devongovett Thoughts on this? Seems reasonable that I should be able to use `import.meta.url` without it being replaced? 

I could modify the transform to add a new special parcel symbol (e.g. `__parcel_importMetaUrl__`)? Or potentially add some asset meta data that prevents the stripping?

marcins (approved):

marcins:
Benchmarks are failing in the size comparison code (https://github.com/parcel-bundler/parcel-benchmark-action/blob/master/src/utils/compare-benchmarks.ts#L37-L65) - my guess is this is because it expects all the bundles produced in the baseline to also be present in the target, and you've renamed some of the runtime bundles.

packages/runtimes/js/src/JSRuntime.js
devongovett: hmm yeah, I guess we could add a new `__parcel` thing, but does this work as is for now? if so I'm ok with leaving it as you have it as well.

packages/runtimes/js/src/helpers/bundle-manifest.js
```diff
@@ -1,18 +1,20 @@
...
-function register(pairs) {
-  var keys = Object.keys(pairs);
-  for (var i = 0; i < keys.length; i++) {
-    mapping[keys[i]] = pairs[keys[i]];
+function register(baseUrl, manifest) {
+  for (var i = 0; i < manifest.length - 1; i += 2) {
+    mapping.set(manifest[i], {
+      baseUrl: baseUrl,
+      path: manifest[i + 1],
+    });
```
devongovett, line 8: You could do the `new URL` thing here instead of storing both and doing it every time we resolve?

packages/runtimes/js/src/helpers/bundle-manifest.js
```diff
@@ -1,18 +1,20 @@
...
 }
 
 function resolve(id) {
-  var resolved = mapping[id];
+  var resolved = mapping.get(id);
   if (resolved == null) {
     throw new Error('Could not resolve bundle with id ' + id);
   }
-  return resolved;
+  return new URL(resolved.path, resolved.baseUrl).toString();
```
devongovett, line 17: This one definitely won't work in IE. Probably ok at this point? I guess we can't simply concatenate the strings? `path` could probably start with `../` or something...

devongovett (approved):
looks ok. couple questions but I wouldn't consider them blocking

packages/runtimes/js/src/helpers/bundle-manifest.js
mattcompiles: I originally did this, but I pivoted as creating the URLs is not free CPU wise. We have a huge manifest and creating 100's of URLs in startup isn't ideal. I believe it's better perf wise to create on demand. We could cache in future if need be though but I just kept it simple.

packages/runtimes/js/src/helpers/bundle-manifest.js
mattcompiles: Yeah not ideal. We could just recommend a polyfill for URL if anyone has issues I guess.

PR #9186, opened 2023-08-09
Update parcel-query .stats

celinanperalta:
# ↪️ Pull Request

Currently, parcel-query .stats does not give enough detailed information about bundle types. This PR adds a more granular breakdown of bundle types to the function by separating bundle counts by certain properties. Adds counts for shared, async, entry, and bundle by file type (css, svg, html, etc.)

## 💻 Examples

```
parcel-query
Loading graphs...
See .help. The graphs can be accessed via `assetGraph`, `bundleGraph` and `requestTracker`.
> .stats
╔═════════════════════════════════════════╗
║        # Asset Graph Node Counts        ║
╟────────────────────┬────────────────────╢
║ asset              │ 70731              ║
╟────────────────────┼────────────────────╢
║ dependency         │ 262452             ║
╟────────────────────┼────────────────────╢
║ asset_group        │ 71317              ║
╚════════════════════╧════════════════════╝

╔═════════════════════════════════════════╗
║       # Bundle Graph Node Counts        ║
╟────────────────────┬────────────────────╢
║ dependency         │ 505675             ║
╟────────────────────┼────────────────────╢
║ bundle             │ 3998               ║
╟────────────────────┼────────────────────╢
║ bundle_group       │ 3852               ║
╟────────────────────┼────────────────────╢
║ asset_node_modules │ 24027              ║
╟────────────────────┼────────────────────╢
║ asset_source       │ 45463              ║
╚════════════════════╧════════════════════╝

╔═════════════════════════════════════════╗
║            # Bundles By Type            ║
╟────────────────────┬────────────────────╢
║ entry              │ 41                 ║
╟────────────────────┼────────────────────╢
║ shared             │ 44                 ║
╟────────────────────┼────────────────────╢
║ async              │ 2454               ║
╟────────────────────┼────────────────────╢
║ parallel           │ 35                 ║
╟────────────────────┼────────────────────╢
║ sync               │ 1424               ║
╚════════════════════╧════════════════════╝

╔═════════════════════════════════════════╗
║         # Bundles By Extension          ║
╟────────────────────┬────────────────────╢
║ html               │ 35                 ║
╟────────────────────┼────────────────────╢
║ js                 │ 2573               ║
╟────────────────────┼────────────────────╢
║ svg                │ 991                ║
╟────────────────────┼────────────────────╢
║ css                │ 19                 ║
╟────────────────────┼────────────────────╢
║ png                │ 320                ║
╟────────────────────┼────────────────────╢
║ txt                │ 1                  ║
╟────────────────────┼────────────────────╢
║ woff               │ 26                 ║
╟────────────────────┼────────────────────╢
║ woff2              │ 26                 ║
╟────────────────────┼────────────────────╢
║ gif                │ 7                  ║
╚════════════════════╧════════════════════╝
```

## 🚨 Test instructions

Verify bundle counts are accurate to builds.

## ✔️ PR Todo

### DIFF
--- a/packages/dev/query/src/cli.js
+++ b/packages/dev/query/src/cli.js
@@ -11,6 +11,9 @@ import os from 'os';
 import nullthrows from 'nullthrows';
 import invariant from 'assert';
 
+// $FlowFixMe
+import {table} from 'table';
+
 import {fromProjectPathRelative} from '@parcel/core/src/projectPath';
 import {bundleGraphEdgeTypes} from '@parcel/core/src/BundleGraph.js';
 import {Priority} from '@parcel/core/src/types';
@@ -34,8 +37,9 @@ export function run(input: string[]) {
   }
 
   console.log('Loading graphs...');
-  let {assetGraph, bundleGraph, bundleInfo, requestTracker} =
-    loadGraphs(cacheDir);
+  let {assetGraph, bundleGraph, bundleInfo, requestTracker} = loadGraphs(
+    cacheDir,
+  );
 
   if (requestTracker == null) {
     console.error('Request Graph could not be found');
@@ -547,6 +551,34 @@ export function run(input: string[]) {
     }
   }
 
+  function _getBundlePriority(bundleGraph, bundle) {
+    const bundleNodeId = bundleGraph._graph.getNodeIdByContentKey(bundle.id);
+    const node = bundleGraph._graph
+      .getNodeIdsConnectedTo(bundleNodeId)
+      .map(id => nullthrows(bundleGraph._graph.getNode(id)))
+      .find(node => node.type === 'dependency');
+
+    if (node == null) return null;
+
+    invariant(node.type === 'dependency', 'Not a dependency');
+
+    return node.value.priority;
+  }
+
+  function _printStatsTable(header, data) {
+    const config = {
+      columnDefault: {
+        width: 15,
+      },
+      header: {
+        alignment: 'center',
+        content: header,
+      },
+    };
+
+    console.log(table(data, config));
+  }
+
   // eslint-disable-next-line no-unused-vars
   function stats(_) {
     let ag = {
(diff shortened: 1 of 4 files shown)


packages/dev/query/src/cli.js
```diff
@@ -569,32 +601,64 @@ export function run(input: string[]) {
...
+
     for (let [, n] of bundleGraph._graph.nodes) {
-      if (n.type in bg) {
+      if (n.type == 'bundle_group') {
+        // TODO: Counting priority here seems to work instead of counting on BundleNodes.. Counting DependencyNode priorities themselves
+        // gives different counts as well.
+        let priority = _getBundlePriority(bundleGraph, n);
+        if (priority == Priority.lazy) {
+          bg_type.async++;
+        } else if (priority == Priority.parallel) {
```
AGawrys: Should've clarified this before but BundleGroups can't be parallel, only entry or async. A bundleGroup is a collection of bundles. Connected to it there may be multiple other bundles it pulls in, like a shared bundle or a parallel (aka type change) bundle. 

So the consideration for parallel should be done in the if statement where you check for shared. That might be why your example had 0 parallel bundles.

packages/dev/query/src/cli.js
```diff
@@ -569,32 +601,64 @@ export function run(input: string[]) {
...
-    console.log();
 
-    console.log('# Bundle Graph Node Counts');
-    for (let k in bg) {
-      console.log(k, bg[k]);
-    }
+    invariant(bg.bundle == sum);
+
+    _printStatsTable('# Asset Graph Node Counts', Object.entries(ag));
+    _printStatsTable('# Bundle Graph Node Counts', Object.entries(bg));
```
AGawrys: I notice the BundleGraph table has `asset` count as well as `asset_nodemodules` and `asset_source` I think we can either remove the `asset` row for the bundleGraph table since we already have that count in the `AssetGraph` table, or, you can sum `asset_nodemodules + asset_source` to get the full count. Currently your example said 0 assets but the BundleGraph should contain all the asset nodes present in the AssetGraph

AGawrys (changes requested):
Looks great just 2 changes and I think we'll be set 👍

packages/dev/query/src/cli.js
celinanperalta: Removing `asset` would make more sense to avoid being redundant

AGawrys (approved):

PR #9188, opened 2023-08-09
Add `~` and `/` support to the glob resolver

### LINKED ISSUE #8704: Glob resolver doesn't support `/` and `~`
# 🐛 bug report

These don't work correctly (the glob resolver doesn't replace these special start characters with the correct absolute path before resolving the glob):
```js
import images from "/assets/*.png";
import images from "~/assets/*.png";
```

## 💁 Possible Solution

Handle these cases here:
https://github.com/parcel-bundler/parcel/blob/da6cbb3633d2c0e9fedb8e2af560c82504d05ddc/packages/resolvers/glob/src/GlobResolver.js#L135-L137

This is what the default resolver does:
https://github.com/parcel-bundler/parcel/blob/da6cbb3633d2c0e9fedb8e2af560c82504d05ddc/packages/utils/node-resolver-core/src/NodeResolver.js#L578-L614

## 💻 Code Sample

- `assets` contains png image files.
- Then the imports above should return those instead of an empty object.

## 🌍 Your Environment

| Parcel           | 2.8.2

jondlm:
# ↪️ Pull Request

Fixes #8704 by adding support for `~` and absolute paths with the glob resolver. `~` resolves to the nearest package boundary and `/` to the project root.

> 🧑‍💻 **Tip**: you can view this PR by commit for a smoother review.

## 💻 Examples

This allows folks using the [glob resolver](https://github.com/parcel-bundler/parcel/blob/88e6db442542f56c21899658dda69c652a85349a/packages/resolvers/glob/src/GlobResolver.js) to import globs like:
- `import images from '~/images/*.png` (package boundary)
- `import images from '/dir/images/*.png'` (project root)

## 🚨 Test instructions

I included integration tests that cover this feature.

## ✔️ PR Todo


devongovett:
Thanks for the PR. Absolute and tilde specifiers in Parcel have a different meaning. See the docs here: https://parceljs.org/features/dependency-resolution/#absolute-specifiers. I guess if we add support in the glob resolver they should match the default resolver as well, so absolute should resolve to the project root and tilde should resolve to the nearest node_modules root or project root.

I'm curious about your use case for system-level absolute/tilde paths. We haven't supported that because paths outside of the project are generally not portable between machines.

jondlm:
Ohh, I totally misunderstood that issue! I didn't realize Parcel had special meanings for those identifiers. I just assumed they were system level and there was a good reason for needing it. I don't actually have a use case here. Just trying to knock off some "help wanted" issues to get comfortable contributing.

I'll rework the PR to properly support the Parcel path semantics. Thanks for the feedback!

jondlm:
@devongovett I fixed this one up whenever you have a chance to take a look again. Sorry about the initial mixup!

packages/resolvers/glob/src/GlobResolver.js
```diff
@@ -60,73 +60,114 @@ export default (new Resolver({
...
-            diagnostics: err.diagnostics,
-            invalidateOnFileCreate,
-            invalidateOnFileChange: [...invalidateOnFileChange],
-          };
-        } else {
-          throw err;
-        }
+      // Absolute path. Make the glob relative to the project root.
+      case '/': {
+        specifier = path.resolve(options.projectRoot, specifier.slice(1));
```
mischnic, line 72: ~~On Windows, this will rewrite slashes in the specifier into backslashes:~~ Which is fine for Parcel, apparently
```js
> path.posix.resolve("/src/app","dir") // default path.resolve on Unix
'/src/app/dir'
> path.win32.resolve("/src/app","dir") // default path.resolve on Windows
'\\src\\app\\dir'
```

packages/resolvers/glob/src/GlobResolver.js
```diff
@@ -60,73 +60,114 @@ export default (new Resolver({
...
+      // root - whichever comes first.
+      case '~': {
+        const insideNodeModules = options.projectRoot.includes('node_modules');
+        let dir = path.dirname(sourceFile);
 
-      specifier = path.resolve(path.dirname(result.filePath), rest);
-      if (result.invalidateOnFileChange) {
-        for (let f of result.invalidateOnFileChange) {
-          invalidateOnFileChange.add(f);
+        while (
```
mischnic: `fs.findAnchestorFile(["package.json"], dir, options.projectRoot)` does exactly this

https://github.com/parcel-bundler/parcel/blob/707e5295f6b1874419c2c3d3aa6ec7f10d6a7c81/packages/core/package-manager/src/NodePackageManager.js#L554-L558

mischnic (approved):

jondlm:
Thank you @mischnic for cleaning this one up! My apologies for not getting to it :(

PR #9193, opened 2023-08-15
expose transform and resolveapi from parcel

kiddkai:
# ↪️ Pull Request

Exposes two new API to public use:

* Parcel.transform
* Parcel.resolve

So API consumers can use these two features to reuse the same resolve & transform pipelines
during development.

The implementation is mostly a copy of: https://github.com/parcel-bundler/parcel/compare/v2...parcel-node

But it excluded the features specifically for `jest` and `NodeResolve`.
And it does not transform `import` directly to `require` that only specific for node register.
As an alternative, we may be able to add another asset query to change it to something else.

eg:

```
query: `standalone=true&renameImportTo=__custom__import__`
```

Which provides more customisability, but I am not adding it for now since we still able to post-process the `import()` for now.

## 💻 Examples

The new exposed API allows us to do:

```js
let parcel = createParcel({workerFarm});
    let res = await parcel.transform({
      filePath: path.join(__dirname, 'fixtures/parcel/other.js'),
      query: 'standalone=true',
    });
    let code = await res[0].getCode();

    assert(code.includes('require("./index.js")'));
    assert(code.includes('new URL("index.js", "file:" + __filename);'));
    assert(code.includes('import("index.js")'));
```

And

```js
let res = await parcel.resolve({
        specifier: './other',
        specifierType: 'esm',
        resolveFrom: path.join(__dirname, 'fixtures/parcel/index.js'),
      });
      assert.deepEqual(res, {
        filePath: path.join(__dirname, 'fixtures/parcel/other.js'),
        code: undefined,
        query: undefined,
        sideEffects: undefined,
      });
```

which allows us to develop custom runtime based on v8 or similar js runtime to share the
transform and resolve pipelines during development without bundling the whole project
which benefits huge projects that takes long time to bundle.

## 🚨 Test instructions

* Unit tested.

## ✔️ PR Todo


packages/core/core/src/Parcel.js
```diff
@@ -437,6 +450,78 @@ export default class Parcel {
...
+          options.env?.loc != null
+            ? {
+                ...options.env.loc,
+                filePath: toProjectPath(projectRoot, options.env.loc.filePath),
+              }
+            : undefined,
+      }),
+    });
+
+    let res = await this.#requestTracker.runRequest(request, {force: true});
```
mattcompiles: Should we use `force` here? My understanding is that `force` means the cache is ignored.

packages/core/core/src/Parcel.js
```diff
@@ -437,6 +450,78 @@ export default class Parcel {
...
+            : undefined,
+      }),
+    });
+
+    let req = createPathRequest({
+      dependency,
+      name: 'test',
+    });
+
+    let res = await this.#requestTracker.runRequest(req, {force: true});
```
mattcompiles: Same question here about `force`.

packages/core/core/src/Parcel.js
```diff
@@ -437,6 +450,78 @@ export default class Parcel {
...
+                ...request.env.loc,
+                filePath: toProjectPath(projectRoot, request.env.loc.filePath),
+              }
+            : undefined,
+      }),
+    });
+
+    let req = createPathRequest({
+      dependency,
+      name: 'test',
```
mattcompiles: Not sure what the purpose of this `name` is. Looks like it's normally sourced from the `AssetGraphBuilder` and it could be `Main` or `Runtimes`. 

@devongovett Any input on what this should be?

packages/core/types/index.js
```diff
@@ -645,6 +645,28 @@ export type ASTGenerator = {|
 
 export type BundleBehavior = 'inline' | 'isolated';
 
+export type ParcelTransformOptions = {|
+  filePath: FilePath,
+  code?: string,
+  env?: EnvironmentOptions,
+  pipeline?: ?string,
```
mattcompiles: Not sure if pipeline makes sense as a public option? Shouldn't this be inferred from the `.parcelrc`?

packages/transformers/js/src/JSTransformer.js
```diff
@@ -442,6 +442,7 @@ export default (new Transformer({
       is_esm_output: asset.env.outputFormat === 'esmodule',
       trace_bailouts: options.logLevel === 'verbose',
       is_swc_helpers: /@swc[/\\]helpers/.test(asset.filePath),
+      standalone: asset.query.has('standalone'),
```
mattcompiles, line 445: Not sure about using a query param to drive this behaviour. Maybe this should be part of the environment?

mattcompiles (commented):
Couple of questions but this is looking pretty good and love how simple it is to add so far. 

@devongovett Would love your thoughts on the comments as well.

packages/transformers/js/src/JSTransformer.js
kiddkai: I did think about it, for now using query provides some flexibility & less moving part to change the environment. Kinda like a feature flag for `jsTransform` for now.

packages/core/core/src/Parcel.js
kiddkai: I can update that, but it was originally copied from the intiial branch, I will leave this Q to @devongovett on if this is required?

packages/core/types/index.js
kiddkai: I am happy to remove this `pipline` property.

packages/core/core/src/Parcel.js
kiddkai: I changed it to `.specifeir`

The name seem to be used to construct the `req.id` only. The first part of id is already hashed, so whatever value here doesn't seem to be matter too much. 

Tho, change to specifier should be a lil bit more meaningful.

packages/core/core/src/Parcel.js
kiddkai: Just changed it to a configurable value. Follows the config of `shouldCacheDisabled` from the main config.
I guess this could be helpful to debug transform problems.

packages/core/types/index.js
kiddkai: removed.

packages/core/core/src/Parcel.js
```diff
@@ -437,6 +450,88 @@ export default class Parcel {
...
+      }),
+    });
+
+    let req = createPathRequest({
+      dependency,
+      name: request.specifier,
+    });
+
+    let res = await this.#requestTracker.runRequest(req, {
+      force: shouldDisableCache,
```
mattcompiles: As above

packages/core/core/src/Parcel.js
```diff
@@ -437,6 +450,88 @@ export default class Parcel {
...
+            ? {
+                ...options.env.loc,
+                filePath: toProjectPath(projectRoot, options.env.loc.filePath),
+              }
+            : undefined,
+      }),
+    });
+
+    let res = await this.#requestTracker.runRequest(request, {
+      force: shouldDisableCache,
```
mattcompiles: Turns out I was wrong about this and there's a separate cache layer for transforms so this is safe to hardcode as `true`.

kiddkai: updated

packages/core/core/src/Parcel.js
kiddkai: updated.

mattcompiles (approved):

PR #9222, opened 2023-08-29
Fix shorthand identifier import usage

### LINKED ISSUE #9140: Missing export with `sideEffects: false`
# 🐛 bug report

I suspect an issue where importing and exporting from the same file, along with `sideEffects: false` causes Parcel to drop the import as unused. This may be related to #7622 but it seems distinct enough to merit a separate issue.

Minimal repro ([repository with full code](https://github.com/jondlm/parcel-bug-side-effect-export/)):

`package.json`
```
"sideEffects": false
```

`index.js`
```javascript
import { tokens } from "./tokens";

console.log(tokens.color.blue); // ❌ fails as `tokens.color` is `undefined`
```

`tokens.js`
```javascript
import { color } from "./color";

// 💡 Uncommenting the line below fixes the issue
// console.log(color);

export const tokens = {
  color,
};

// 💡 Alternatively commenting out this line fixes the issue
export { mode } from "./color";
```

`color.js`
```javascript
export const color = {
  blue: "#0196ed",
};

export const mode = "dark";
```

No notable config beyond `sideEffects: false` in the package.json.

## 🤔 Expected Behavior

`tokens.color.blue` should be defined within `index.js`.

## 😯 Current Behavior

Parcel is dropping the `color.js` module thinking it's unused unless I add a `console.log` or uncomment the export at the bottom.

## 💁 Possible Solution

Not sure yet. I'll need to dig into Parcel core to propose something.

## 🔦 Context

We ran into this bug with one of our internal libraries that is side effect free. It was causing a page to not load correctly.

## 💻 Code Sample

https://github.com/jondlm/parcel-bug-side-effect-export

## 🌍 Your Environment

| Parcel           | 2.9.2 |
| Node             | 18.14.0 |
| pnpm         | 8.6.0 |
| Operating System | macOS 13.4.1 |

jondlm:
# ↪️ Pull Request

Fixes #9140 by adding a new SWC visitor to the JS transformer. It visits all identifiers, checks if their IDs match any of the previously detected imports. If so it adds them to the `used_imports` set.

## 💻 Examples

See #9140 for an outline of the bug this fixes.

## 🚨 Test instructions

I added an integration test that fails before this change and passes after. I couldn't find unit tests for the JS transformer so I'm assuming an integration test is appropriate here. I used @lettertwo's new `fsFixture` helper to colocate the test files with the actual test. That worked quite nicely!

## ✔️ PR Todo


packages/transformers/js/core/src/collect.rs
```diff
@@ -757,6 +757,16 @@ impl Visit for Collect {
     }
   }
 
+  fn visit_ident(&mut self, node: &Ident) {
+    // This visitor helps us identify used imports in cases like:
+    //
+    //   import { foo } from "bar";
+    //   const baz = { foo };
+    if self.imports.contains_key(&id!(node)) {
```
jondlm, line 765: I couldn't think of any cases where this new visitor would be problematic since SWC seems to imply that checking the id here helps us uniquely identify the correct use accounting for scope. Let me know if I'm wrong! I am very new to using Rust with SWC.

jondlm:
Perhaps @mischnic could take a look at this one when you have time?

devongovett (approved):

mischnic (approved):
Seems like object shorthand is one of the few cases where there's an actual identifier that isn't/can't be an expression (all of the other cases are handled by `visit_expr`)

PR #9239, opened 2023-09-12
Don't confuse workers with no-opt command line option

danieltroger:
# ↪️ Pull Request

Maybe fixes https://github.com/parcel-bundler/parcel/discussions/9237

## 💻 Examples

See https://github.com/parcel-bundler/parcel/discussions/9237

## 🚨 Test instructions

You can try setting up docker and node like described in the discussion above. But I don't see how this could hurt? Let's just try it and maybe it works.

## ✔️ PR Todo

### DIFF
--- a/packages/core/workers/src/Worker.js
+++ b/packages/core/workers/src/Worker.js
@@ -49,7 +49,7 @@ export default class Worker extends EventEmitter {
 
   async fork(forkModule: FilePath) {
     let filteredArgs = process.execArgv.filter(
-      v => !/^--(debug|inspect|max-old-space-size=)/.test(v),
+      v => !/^--(debug|inspect|no-opt|max-old-space-size=)/.test(v),
     );
 
     for (let i = 0; i < filteredArgs.length; i++) {


mischnic:
> You can try setting up docker and node like described in the discussion above. But I don't see how this could hurt? Let's just try it and maybe it works.

Or: set `NODE_OPTIONS="--opt"` which something in your setup does

danieltroger:
> Or: set `NODE_OPTIONS="--opt"` which something in your setup does

I tried that, the problem is that it yields
`/Users/daniel/.nvm/versions/node/v20.5.0/bin/node: --opt is not allowed in NODE_OPTIONS`


And --no-opt

```
/Users/daniel/.nvm/versions/node/v20.5.0/bin/node: --no-opt is not allowed in NODE_OPTIONS
```

mischnic:
I guess it has be injected in some other way then, e.g. by overriding the `node` binary on PATH or
```
$ node --opt
Welcome to Node.js v18.17.1.
Type ".help" for more information.
> process.execArgv
[ '--opt' ]
```

danieltroger:
I manipulated my path and the argument is there when running node
```
daniel@mmmmmmmmmm utilishared % node                   
Welcome to Node.js v20.5.0.
Type ".help" for more information.
> process.execArgv
[ '--no-opt' ]
> 
```
but `yarn parcel build` still succeeds for me (suggesting that the option is not respected).
However, I did this change with a `yarn patch` and now my Dockerfile seems to get further.

EDIT: yup, my docker image builds successfully now 🎉

mischnic (approved):

PR #9178, opened 2023-08-04
Add support for Chrome Extension manifest V3 side_panel property.

hi-upchen:
# ↪️ Pull Request

This pull request addresses an issue where the existing web extension plugin did not recognize the `side_panel` property in Chrome Extension manifest V3, leading to the HTML files not being compiled and copied to the dist folder. With the recent support for the `side_panel` property in Google Chrome, this PR introduces the necessary changes to enable the plugin to recognize the property, ensuring proper compilation of the HTML files.

The feature aligns with the latest [chrome.sidePanel](https://developer.chrome.com/docs/extensions/reference/sidePanel/) documentation from Chrome, ensuring compliance with the latest standards.

## 💻 Examples

For the manifest
```
{
  "manifest_version": 3,
  "name": "Sample Web Extension",
  "version": "0.0.1",
  "background": {
    "service_worker": "service-worker.ts"
  },
  "action": {
    "default_title": "Click to open panel"
  },
  "side_panel": {
    "default_path": "pages/sidepanel/sidepanel.html"
  },
  "permissions": [
    "sidePanel",
    "tabs"
  ],
  "icons": {
    "16": "assets/images/icon-16.png",
    "48": "assets/images/icon-48.png",
    "128": "assets/images/icon-128.png"
  }
}
```

*Before:* 
The side_panel is not recognized by the `@parcel/config-webextension` so that the html is not bring to the dist folder.
image

*After*
With the update, the HTML file for the `side_panel` settings is now correctly compiled and included in the dist folder.
image

## 🚨 Test instructions

Integration tests for the new functionality have been added.

## ✔️ PR Todo

- [x] Added/updated unit tests for this change
- [x] Filled out test instructions (In case there aren't any unit tests)
- [ ] Included links to related issues/PRs

PR #9257, opened 2023-09-20
Update WebExtensionTransformer.js resolve side_panel.defaul_path

atl-jlandaverde:
Chrome's config for side panels uses side_panel.config instead of sidebar_action.default_panel. 

https://developer.chrome.com/docs/extensions/reference/sidePanel/#use-cases

# ↪️ Pull Request

Chrome uses a different property to configure side panels. It uses side_panel.default_path instead of sidebar_action.default_panel. As a result we can't configure a side panel for a chrome extension

## 💻 Examples

In a manifest.json use the following configuration for a side panel.
```
{
  "name": "My side panel extension",
  ...
  "side_panel": {
    "default_path": "sidepanel.html"
  }
  ...
}
```

## 🚨 Test instructions
Configure you manifest.json for to use a side panel. Check the bundle and it should transform side_panel.default_path 
{
  "name": "My side panel extension",
  ...
  "side_panel": {
    "default_path": "sidepanel.html"
  }
  ...
}

## ✔️ PR Todo

- [x] Added/updated unit tests for this change - No tests for this file
- [ x] Included links to related issues/PRs - Did not find any

### DIFF
--- a/packages/transformers/webextension/src/WebExtensionTransformer.js
+++ b/packages/transformers/webextension/src/WebExtensionTransformer.js
@@ -29,6 +29,7 @@ const DEP_LOCS = [
   ['sandbox', 'pages'],
   ['sidebar_action', 'default_icon'],
   ['sidebar_action', 'default_panel'],
+  ['side_panel', 'default_path'],
   ['storage', 'managed_schema'],
   ['theme', 'images', 'theme_frame'],
   ['theme', 'images', 'additional_backgrounds'],


mischnic:
Could you add this entry to the test case here?
- https://github.com/parcel-bundler/parcel/blob/v2/packages/core/integration-tests/test/integration/webextension-mv3/manifest.json
- and then update the assertion here https://github.com/parcel-bundler/parcel/blob/6dcba3606107eb02f3008e30678538f81a4fc902/packages/core/integration-tests/test/webextension.js#L104
(As far as I can tell, this is only in manifest version 3?)

mischnic:
Thank you for the PR, but I just noticed that we already have a PR for this property that also has tests: #9178

Issue #9292, opened 2023-10-04
Watcher stops watching dependant scss files after syntax error


bdukes:
# 🐛 bug report

When using `parcel watch ./src/styles.scss` and introducing a syntax error in an `scss` file imported by `styles.scss`, the watcher will no longer see changes to the imported file.

## 🎛 Configuration (.babelrc, package.json, cli command)

`package.json`
```js
{
  "name": "parcel-demo",
  "scripts": {
    "start": "parcel watch ./src/index.scss"
  },
  "devDependencies": {
    "@parcel/transformer-sass": "^2.9.3",
    "parcel": "^2.9.3"
  }
}
```

## 🤔 Expected Behavior

Changes to imported file trigger a rebuild.

## 😯 Current Behavior

Changes to imported file trigger a rebuild until there's a syntax error, and then no longer trigger a rebuild.

## 💻 Code Sample

1. Clone https://github.com/bdukes/parcel-issue-9292
2. `npm install`
3. `npm start`
4. Add an extra `}` character to `src/another.scss`
5. Save `src/another.scss`, notice error output
6. Remove extra `}` character
7. Save `src/another.scss`, notice error output does not change
8. Save `src/index.scss`, notice output now changes to show success

## 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Parcel           | 2.9.3 |
| Node             |  18.18.0 |
| npm         |  9.8.1 |
| Operating System | Windows 11 (also reproduced on MacOS Sonoma 14.0) |

mischnic:
Unfortunately a duplicate of https://github.com/parcel-bundler/parcel/issues/7374 . There's not much Parcel can do here without more support from the Sass compiler

Issue #7374, opened 2021-11-28
Error recovery from Sass errors


razvanandreib:
# 🐛 bug report

As I start typing in the scss file , I immediately get this error and it happens everytime:

```
🚨 Build failed.

@parcel/transformer-sass: expected "{".
╷
289 │ .
│  ^
```

And parcel stops to function, i need to close it and npm start all over again.

## 🎛 Configuration (.babelrc, package.json, cli command)
```
{
  "name": "ime-cadastru",
  "version": "1.0.0",
  "source": "./src/pages/cadastru.html",
  "description": "",
  "scripts": {
    "start": "parcel ./src/index.html",
    "build": "parcel build"
  },
  "keywords": [],
  "author": "",
  "license": "ISC",
  "devDependencies": {
    "@parcel/transformer-sass": "^2.0.1",
    "parcel": "^2.0.1",
    "sass": "^1.43.4",
    "sharp": "^0.29.3"
  }
}
```

## 🔦 Context

I try to write css in the scss files and immediatly after i start typing i get this error.

## 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Parcel "@parcel/transformer-sass": "^2.0.1",
    "parcel": "^2.0.1",
    "sass": "^1.43.4",
    "sharp": "^0.29.3"          |
| Node      v17.1.0     |
| npm/Yarn   8.1.2     |
| Operating System Windows 10 |

bensanlau:
+1. I'm also running into same issue. It happens when I'm watching and make a syntax mistake. After rectifying it + save, the bundler does not automatically bundle. I had to close and re-watch.
Strangely, I find that saving another file in your project somehow triggers the build. 
But working on the same file that had the mistake does not trigger anything.

folknor:
This is probably a duplicate of #6685 or one of the 10+ linked issues from that ticket, like #5660.

mischnic:
Can someone post a reproduction with instructions on how to trigger this bug?

razvanandreib:
Sure. will come back with a screenshot.

bensanlau:
Hey @mischnic,

It's my first time logging a bug, but I'll try to be concise:

Set up
1. I create a new project following the [webapp starterkit](https://parceljs.org/getting-started/webapp/), including adding an `index.html` inside the src folder.
2. Instead of calling styles.css, I'm using SCSS and changed the extension. This installs the transformer-sass plugin.
[...]

PR #9315, opened 2023-10-12
Remember related files for Sass on syntax error

bdukes:
# ↪️ Pull Request

When watching Sass files, an error from the Sass compiler will cause Parcel to no longer about which files a Sass entry file imports (i.e. which files will cause the entry file to be invalidated). This is documented in #7374 and #9292.

This PR remembers the previous invalidations and reapplies them to the input file in the case of an error.

Presumably this issue also affects other transformers ([it's documented to also apply to Pug and Elm](https://github.com/parcel-bundler/parcel/issues/7374#issuecomment-1002948172), at least), so perhaps this resolution is at the wrong level (maybe should be [in `RequestTracker`?](https://github.com/parcel-bundler/parcel/blob/75c1921c86017e747d8b46ea172392f78f187814/packages/core/core/src/RequestTracker.js#L991-L1025)), but it does resolve the problem for Sass in the meantime.

## 💻 Examples

See #9292

## 🚨 Test instructions

See #9292

## ✔️ PR Todo

  - For #7374
  - For #9292

### DIFF
--- a/packages/transformers/sass/src/SassTransformer.js
+++ b/packages/transformers/sass/src/SassTransformer.js
@@ -9,6 +9,8 @@ import {promisify} from 'util';
 // E.g: ~library/file.sass
 const NODE_MODULE_ALIAS_RE = /^~[^/\\]/;
 
+const previousIncludedFiles = new Map();
+
 export default (new Transformer({
   async loadConfig({config, options}) {
     let configFile = await config.getConfig(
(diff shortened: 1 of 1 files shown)


mischnic:
The problem with this approach is that it only works for watch mode and only if you don't restart Parcel while there's a Sass error.

bdukes:
Certainly it would be better if it worked from a cold-start, but that is out of our control. This workaround approach, however, resolves 99% of the situations where we run into issues.

PR #9324, opened 2023-10-18
Use Yarn modern

mischnic:
This should help with the constant Yarn install timeouts we have on CI

Using pnpm would probably break lerna (though I haven't tested if the publish scripts still work now with the new Yarn)

"Known issues":
- `"scripts": {"x": "echo x"}` and then `yarn x` takes 1 second for me: https://github.com/yarnpkg/berry/issues/2575
- A noop `yarn install` (e.g. running twice in a row) still takes 3.5s

Bugs preventing the pnpm linker:
- https://github.com/yarnpkg/berry/issues/5165 (only with Yarn 4 though)


mischnic:
No idea why some integration tests fail consistently, they do work fine locally for me.

PR #9109, opened 2023-06-26
Fix duplicate asset references

mattcompiles:
# ↪️ Pull Request

It's currently possible that assets can be duplicated but not wrapped in the packager. This causes the duplicated asset to be evaluated more than once and receive a unique module scope. 

The fix is to ensure that all duplicated assets return true when when checking if an asset is referenced. 

## 🚨 Test instructions

Test added

## ✔️ PR Todo

- [x] Added/updated unit tests for this change
- [x] Filled out test instructions (In case there aren't any unit tests)
- [ ] Included links to related issues/PRs

PR #9348, opened 2023-10-29
Do not wrap duplicated assets when they are in different targets

devongovett:
This fixes a regression from #9109. When an asset is duplicated, that PR caused it to be wrapped so it gets deduplicated at runtime. However, if the duplication occurs in separate targets, this is not necessary. It resulted in `parcelRequire` getting added to library builds, causing downstream errors in poorly written bundlers trying to detect CJS usage (https://github.com/vercel/next.js/issues/57649) (and also unnecessary global variables). The fix is to detect whether the duplication occurs within the same target.

### DIFF
--- a/packages/core/core/src/BundleGraph.js
+++ b/packages/core/core/src/BundleGraph.js
@@ -1122,8 +1122,12 @@ export default class BundleGraph {
   }
 
   isAssetReferenced(bundle: Bundle, asset: Asset): boolean {
-    // If the asset is available in multiple bundles, it's referenced.
-    if (this.getBundlesWithAsset(asset).length > 1) {
+    // If the asset is available in multiple bundles in the same target, it's referenced.
+    if (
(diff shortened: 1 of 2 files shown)


packages/core/core/src/BundleGraph.js
```diff
@@ -1122,8 +1122,12 @@ export default class BundleGraph {
   }
 
   isAssetReferenced(bundle: Bundle, asset: Asset): boolean {
-    // If the asset is available in multiple bundles, it's referenced.
-    if (this.getBundlesWithAsset(asset).length > 1) {
+    // If the asset is available in multiple bundles in the same target, it's referenced.
+    if (
+      this.getBundlesWithAsset(asset).filter(
+        b => b.target.distDir === bundle.target.distDir,
```
devongovett, line 1128: we should probably have a target.id at some point, but this is how the default bundler does it already.

mattcompiles (approved):

PR #9355, opened 2023-11-02
Write cache large blobs in chunks to avoid write file limits in Node

JakeLane:
# ↪️ Pull Request

In Node, there's a limit to the size we can read at once (2 GiB). We are hitting that for large monorepos at Atlassian. This makes the cache unusable for repeated usage, as the node API for readFile will always return an error.

A mitigation for this is to chunk the files in ~2GiB segments. This means the upper limit of blob size will be the buffer size limit.

On my machine, the buffer size limit is:
```
node -e "console.log(require('buffer').constants.MAX_LENGTH / 1.074e+9 + 'GiB')"
3.999038450651769GiB
```

This gives us some headroom to resolve large blob size issues.

### DIFF
--- a/packages/core/cache/src/FSCache.js
+++ b/packages/core/cache/src/FSCache.js
@@ -102,7 +102,7 @@ export class FSCache implements Cache {
     return Buffer.concat(await Promise.all(buffers));
   }
 
-  async setLargeBlob(key: string, contents: Buffer): Promise<void> {
+  async setLargeBlob(key: string, contents: Buffer | string): Promise<void> {
     const chunks = Math.ceil(contents.length / WRITE_LIMIT_CHUNK);
 
     const writePromises: Promise<void>[] = [];
@@ -110,7 +110,12 @@ export class FSCache implements Cache {
       writePromises.push(
         this.fs.writeFile(
           this.#getFilePath(key, i),
-          contents.subarray(i * WRITE_LIMIT_CHUNK, (i + 1) * WRITE_LIMIT_CHUNK),
+          typeof contents === 'string'
+            ? contents.slice(i * WRITE_LIMIT_CHUNK, (i + 1) * WRITE_LIMIT_CHUNK)
+            : contents.subarray(
+                i * WRITE_LIMIT_CHUNK,
+                (i + 1) * WRITE_LIMIT_CHUNK,
+              ),
         ),
       );
     }
--- a/packages/core/cache/src/FSCache.js
+++ b/packages/core/cache/src/FSCache.js
@@ -13,6 +13,8 @@ import {serialize, deserialize, registerSerializableClass} from '@parcel/core';
 // flowlint-next-line untyped-import:off
 import packageJson from '../package.json';
 
+import {WRITE_LIMIT_CHUNK} from './constants';
+
 const pipeline: (Readable, Writable) => Promise<void> = promisify(
   stream.pipeline,
 );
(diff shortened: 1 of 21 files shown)


packages/dev/query/src/index.js
```diff
@@ -120,9 +124,9 @@ export function loadGraphs(cacheDir: string): {|
   return {assetGraph, bundleGraph, requestTracker, bundleInfo};
 }
 
-function loadLargeBlobRequestRequestSync(cacheDir, node) {
+async function loadLargeBlobRequestRequestSync(cache, node) {
```
mattcompiles: I'm not sure why this was originally written to be sync, but if we're making it async then we should rename the function.

packages/core/cache/src/types.js
```diff
@@ -12,7 +12,7 @@ export interface Cache {
   setBlob(key: string, contents: Buffer | string): Promise<void>;
   hasLargeBlob(key: string): Promise<boolean>;
   getLargeBlob(key: string): Promise<Buffer>;
-  setLargeBlob(key: string, contents: Buffer | string): Promise<void>;
```
mattcompiles: Not sure if this ever actually happens, but this is technically a breaking change for this package. Is it definitely required?

packages/core/cache/src/LMDBCache.js
```diff
@@ -91,16 +92,39 @@ export class LMDBCache implements Cache {
...
-    await this.fs.writeFile(path.join(this.dir, key), contents);
+  async setLargeBlob(key: string, contents: Buffer): Promise<void> {
+    const chunks = Math.ceil(contents.length / WRITE_LIMIT_CHUNK);
+
+    const writePromises: Promise<void>[] = [];
+    for (let i = 0; i < chunks; i += 1) {
+      writePromises.push(
+        this.fs.writeFile(
+          this.#getFilePath(key, i),
+          contents.subarray(i * WRITE_LIMIT_CHUNK, (i + 1) * WRITE_LIMIT_CHUNK),
```
mattcompiles: Might be a micro-optimization, but should we avoid this work when `chunks.length === 1` seeing as that's most of the time?

JakeLane: Yeah fair call, there's no copy cost as it references the same memory in the buffer, but there's no guarantee the work is trivial. Will change it when updating the rest of the code

packages/core/cache/src/types.js
JakeLane: It's not required but I figured it would be a nice clean up to do, since it's adding complexity. Is there a convention for tightening APIs like this?

I'll remove the API change anyway, since it's relatively easy to implement

packages/core/cache/src/FSCache.js
```diff
@@ -81,16 +83,44 @@ export class FSCache implements Cache {
...
-    return this.fs.readFile(this._getCachePath(`${key}-large`));
+  async getLargeBlob(key: string): Promise<Buffer> {
+    const buffers: Promise<Buffer>[] = [];
+    for (let i = 0; await this.fs.exists(this.#getFilePath(key, i)); i += 1) {
+      const file: Promise<Buffer> = this.fs.readFile(this.#getFilePath(key, i));
+
+      buffers.push(file);
+    }
+
+    return Buffer.concat(await Promise.all(buffers));
```
lettertwo, line 102: Will this not incur a `buffer.constants.MAX_LENGTH` violation? Or is the problem that `fs.readFile` has it's own, lower limit?

JakeLane: Yeah, `readFile` is limited to 2GiB but Buffers are higher and dependent on the V8 version.

lettertwo: ah, makes sense. I guess we're still gonna crash if we try to put together a 4GiB file, but (🤞) that's a new limit we won't hit any time soon.

lettertwo (approved):
LGTM! 🚀

packages/dev/query/src/index.js
mischnic: If I recall correctly, I made `loadGraphs` sync just to make it easier to use from the Node REPL. So no big deal

mischnic (approved):

PR #9360, opened 2023-11-02
Reduce redundancy in the RequestGraph's file nodes

gorakong:
# ↪️ Pull Request

This PR reduces redundancy in the RequestGraph's file nodes (which account for close to 50% of the RequestGraph's nodes) by getting rid of the `value` object in each of them, since these can be derived from their IDs. On a large app, this nets us a 6.3% reduction in the size of the serialized RequestGraph. 

## 💻 Examples
Before:
```
  {
    id: 'node_modules/glob-parent/index.js',
    type: 'file',
    value: { filePath: 'node_modules/glob-parent/index.js' }
  }
 ```
After:
```
{ id: 'node_modules/glob-parent/index.js', type: 'file' }
```

### CODE REVIEW

File: packages/core/core/src/projectPath.js [hunk 1]
```diff
@@ -6,7 +6,7 @@ import {relativePath, normalizeSeparators} from '@parcel/utils';
 /**
  * A path that's relative to the project root.
  */
-export opaque type ProjectPath = string;
+export type ProjectPath = string;
```
gorakong: I'm somewhat iffy about the removal of `opaque` here to appease Flow, but it seemed to be the most straightforward solution. Let me know if there are objections to this change.

JakeLane (approved):

packages/core/core/src/projectPath.js
lettertwo: You can use `toProjectPathUnsafe(node.id)` to turn the ids into a `ProjectPath`. I think that is preferable to changing the type to not be opaque, as doing so causes every other `ProjectPath` type across Parcel to be less sound than it was before.

lettertwo (changes requested):

lettertwo (approved):

thebriando (approved):

devongovett (approved):

AGawrys (approved):

PR #9361, opened 2023-11-03
Add parcel query option to view de/serialization times & graph size

AGawrys:
# ↪️ Pull Request

We'd like to make cache improvements, so having a quick way to display size/serialization times would be helpful. This PR adds `cacheInfo` structure to hold deserialization times / graph size and a new option to parcel-query called `.inspect-cache`
## 💻 Examples
Example of table
inspectCache

## ✔️ Q's

- Are there any other serialization times we need to take into account besides those of the graphs

### DIFF
--- a/packages/dev/query/src/index.js
+++ b/packages/dev/query/src/index.js
@@ -25,6 +25,7 @@ export async function loadGraphs(cacheDir: string): Promise<{|
   bundleGraph: ?BundleGraph,
   requestTracker: ?RequestTracker,
   bundleInfo: ?Map<ContentKey, PackagedBundleInfo>,
+  cacheInfo: ?Map<string, Array<string | number>>,
 |}> {
   function filesBySizeAndModifiedTime() {
     let files = fs.readdirSync(cacheDir).map(f => {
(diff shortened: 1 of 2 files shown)


packages/dev/query/src/index.js
```diff
@@ -24,6 +24,7 @@ export function loadGraphs(cacheDir: string): {|
   bundleGraph: ?BundleGraph,
   requestTracker: ?RequestTracker,
   bundleInfo: ?Map<ContentKey, PackagedBundleInfo>,
+  cacheInfo: ?Map<string, Array<string | number>>,
```
mischnic, line 28: This parameter doesn't seem to be used

AGawrys: Not sure I understand what's wrong, sorry.`cacheInfo` is meant to store the the size and deserialize times of the loaded graphs, and then returned from the `loadGraphs` function to be used in the `cli` function. It should be partially populated during `loadGraph`.

mischnic: Oops, I missed some punctuation and saw this 🙈 
```js
export function loadGraphs(cacheDir: string,
  assetGraph: ?AssetGraph,
  bundleGraph: ?BundleGraph,
  requestTracker: ?RequestTracker,
  bundleInfo: ?Map<ContentKey, PackagedBundleInfo>,
  cacheInfo: ?Map<string, Array<string | number>>,
) {
	...
	let cacheInfo: ?Map<string, Array<string | number>> = new Map();
```

thebriando (approved):

PR #9364, opened 2023-11-06
Write bundle graph to cache on failure

irismoini:
# ↪️ Pull Request

This PR adds logic to write Parcel's bundle graph to the cache if an error occurs during bundling.

Currently Parcel’s `requestGraph` is written to the cache at the end of a build or when interrupted with ctrl-c. Parcel’s `bundleGraph` is stored in `node.value.results` in the `requestGraph` and is written to the cache as part of the `requestGraph`'s write to cache. If an error occurs during bundling however, the `bundleGraph` is never stored in a `requestGraph` node and as a result isn't written to the cache. This PR adds code to also store the `bundleGraph` in a `requestGraph` node if an error occurs during bundling.

Ideally, we'd eventually like to be able to replicate failed builds from the Parcel cache and this PR is a step towards that.

## 🚨 Test instructions

NA
## ✔️ PR Todo


packages/core/integration-tests/test/cache.js
```diff
@@ -6154,6 +6159,84 @@ describe('cache', function () {
...
+            },
+          
+            bundle({bundleGraph, config}) {
+              DefaultBundler[CONFIG].bundle({bundleGraph, config});
+            },
+            optimize() {throw new Error("Intentionally throw error")},
+          });
+        yarn.lock:`;
+
+      let errorThrown = true;
```
lettertwo: Instead of a a try/catch and assertion flow, you could use something like `await assert.rejects(() => bundle(entries, options))`

packages/core/core/src/requests/BundleGraphRequest.js
```diff
@@ -376,6 +376,16 @@ class BundlerRunner {
         await this.runDevDepRequest(devDepRequest);
       }
     } catch (e) {
+      invariant(internalBundleGraph != null);
```
mischnic: A failing assertion here will overwrite the actual error that was caught, you should wrap the `storeResult` call in an `if`. Should behave the same but retain the error message

PR #9331, opened 2023-10-19
Fix for id$exports being undefined

thebriando:
# ↪️ Pull Request

Fixes #8792 where doing an `import * as` from a library with `export *` in some cases would result in a `id$exports` being undefined

## ✔️ PR Todo

- [x] Added/updated unit tests for this change
- [x] Filled out test instructions (In case there aren't any unit tests)
- [x] Included links to related issues/PRs

PR #9365, opened 2023-11-07
REPL

mischnic:
https://pr-9365.repl.parceljs.org/


mischnic:
@devongovett So what you need to do is:

- Create a Vercel token and set it as `REPL_VERCEL_TOKEN`
- Create a new Vercel project (e.g. deploy an empty folder to a new project) for the REPL
You probably need these settings for that project on vercel.com
Bildschirm­foto 2023-11-11 um 16 24 31


- From .vercel/project.json take the project and org is and set them as `REPL_VERCEL_ORG_ID` and `REPL_VERCEL_PROJECT_ID`

That's used here:
https://github.com/parcel-bundler/parcel/blob/0025d0ebb04172d0bf467fde3211b299f297cbac/.github/workflows/ci.yml#L172-L174

For now, this would then deploy as e.g. https://pr-9365.repl.parceljs.org

https://github.com/parcel-bundler/parcel/blob/0025d0ebb04172d0bf467fde3211b299f297cbac/.github/workflows/ci.yml#L179

We can add some deploy-on-merge/nightly/release stuff later

mischnic:
This is what registering the deployment look like in PRs: https://github.com/mischnic/parcel/pull/1

devongovett:
@mischnic I added the tokens and uploaded an empty project to repl.parceljs.org. I guess we have to merge the GitHub actions changes first in order to test it?

mischnic:
No, this should work from the PR. It did for https://github.com/mischnic/parcel/pull/1

devongovett:
Looks like it failed during Vercel deployment:

```js
error An unexpected error occurred: "https://registry.yarnpkg.com/@parcel%2fconfig-repl: Not found".
```

mischnic:
Then you probably haven't disabled building on Vercel itself as I've shown in the screenshot here: https://github.com/parcel-bundler/parcel/pull/9365#issuecomment-1806843513 ?

The switches have to be enabled and the input fields have to be empty. Kind of a weird UI

devongovett:
I have this. If I change the override switches but don't fill in a value into the inputs the save button doesn't become enabled... 
image

devongovett:
changed the build command to "echo build" 😂

mischnic:
You're right. But this trick worked for me 🙈 
1. Change "dist" to anything else
2. Save
3. Turn on all toggles and make it "dist" again. Then you can save and it does store everything

devongovett:
I think it worked! https://pr-9365.repl.parceljs.org

mischnic:
You have this enabled for the project (probably globally and you need to disable it for the project): https://vercel.com/docs/security/deployment-protection/methods-to-protect-deployments/vercel-authentication#access-requests

Bildschirm­foto 2023-11-12 um 22 22 27

devongovett:
ok I turned that off

mischnic:
Great, so we're set now.

Apart from the fact that the `TypeError: a.FSCache is not a constructor` error you get when using the REPL is apparently a regression from https://github.com/parcel-bundler/parcel/pull/9331 ...

packages/core/cache/src/LayeredCache.js
```diff
@@ -0,0 +1,176 @@
...
+
+  // async #addToBatch(key: string, value: Buffer | string) {
+  //   // if (this.#isWorker) {
+  //   // } else {
+  //   this._batch.push([key, value]);
+  //   if (this._batch.length > this.batchSize) {
+  //     await this.flush();
+  //   }
+  //   // }
+  // }
```
devongovett: I guess this commented out code isn't needed?

packages/core/fs/src/MemoryFS.js
```diff
@@ -913,7 +913,7 @@ class Directory extends Entry {
   }
 }
 
-function makeShared(contents: Buffer | string): Buffer {
+export function makeShared(contents: Buffer | string): Buffer {
```
devongovett, line 916: Are these exports use somewhere?

packages/runtimes/hmr-sse/src/loaders/hmr-runtime.js
```diff
@@ -0,0 +1,547 @@
+// @flow
+/* global HMR_HOST, HMR_PORT, HMR_ENV_HASH, HMR_SECURE, chrome, browser, importScripts */
+
+/*::
```
devongovett: how is this runtime different from our default one?

packages/transformers/react-refresh-wrap/src/ReactRefreshWrapTransformer.js
```diff
@@ -3,6 +3,11 @@
 import path from 'path';
 import {Transformer} from '@parcel/plugin';
 
+const FILENAME = // $FlowFixMe[prop-missing]
+  process.env.PARCEL_BUILD_REPL && process.browser
+    ? '/app/__virtual__/@parcel/transformer-react-refresh-wrap/src/ReactRefreshWrapTransformer.js'
+    : __filename;
```
devongovett: Wonder if we should have a lint rule or something to ensure we don't accidentally add `__filename` or `__dirname` again? Also, I guess the default browser replacement that parcel does for those doesn't work in this case?

devongovett (commented):
Looks pretty self contained! Will be nice to have this finally merged. 🥳

packages/core/fs/src/MemoryFS.js
mischnic: Yes: https://github.com/parcel-bundler/parcel/blob/8fe737fc8aa09793a4e1c05fa2a9db36ac4574a5/packages/dev/repl/src/parcel/ExtendedMemoryFS.js#L3
That implements many of the callback and fd based FS operations so that Yarn works

packages/runtimes/hmr-sse/src/loaders/hmr-runtime.js
mischnic: It used `new EventSource` instead of `new Websocket`. I've merged that into the regular runtime plugin now with a flag to switch between the two.

packages/core/cache/src/LayeredCache.js
mischnic: This whole file isn't used at the moment (and not exported in `cache/src/index`. Same for `RemoteCache.js`). I had experimented with batching writes to the remote cache.

I've deleted all of these files for now

packages/transformers/react-refresh-wrap/src/ReactRefreshWrapTransformer.js
mischnic: (The problem with the current approach for normal Parcel usage is that runtime assets have a filepath of `__filepath` and then load `require("./loaders/browser/js-loader`)` from the real filesystem in node_modules. But that filesystem doesn't exist for the REPL, so `REPLRuntimesResolver` tries to intercept these dependencies and return the contents of the file. The problem is how to identify these dependencies)

It would be possible to implement this without prefixing all runtime asset paths with `/app/__virtual__` (and leave it as is).
But then from the [`REPLRuntimesResolver`](https://github.com/parcel-bundler/parcel/blob/8fe737fc8aa09793a4e1c05fa2a9db36ac4574a5/packages/resolvers/repl-runtimes/src/REPLRuntimesResolver.js) perspective (which provides the dependencies of the runtime assets such as the loaders, etc... at runtime), it looks like a dependency with
```js
resolveFrom: '/app/packages/runtimes/js/src/runtime-70a029dfdf59c5ac.js'
specifier: './helpers/bundle-manifest'
```
and this "internal" dependency is not easily distinguishable from a "user" dependency that should just go through the usual resolver AFAICT. Any ideas?

`__filename` is `packages/runtimes/js/src/JSRuntime.js` when building the REPL. Which then gets concatenated onto the project root `/app` to result in the above `resolveFrom`

packages/runtimes/hmr/src/HMRRuntime.js
```diff
@@ -4,6 +4,15 @@ import {Runtime} from '@parcel/plugin';
...
 import path from 'path';
 
+// Without this, the hmr-runtime.js is transpiled with the React Refresh swc transform because it
+// lives in `/app/packages/runtimes/...` and thus the `config` in the JSTransformer is actually the
+// user's package.json, and hmr-runtime.js is transpiled as a JSX asset.
+const FILENAME =
+  // $FlowFixMe
+  process.env.PARCEL_BUILD_REPL && process.browser
+    ? '/' + __filename
+    : __filename;
```
mischnic, line 14: 😬

packages/transformers/react-refresh-wrap/src/ReactRefreshWrapTransformer.js
mischnic: I changed this now to simply check for these `/app/packages/runtimes/js/src/` paths in the resolver.

devongovett (approved):

PR #9366, opened 2023-11-07
Cache bundle graph on failure

irismoini:
# ↪️ Pull Request
This PR adds logic to write Parcel's bundle graph to the cache if an error occurs during bundling.

Currently Parcel’s `requestGraph` is written to the cache at the end of a build or when interrupted with ctrl-c. Parcel’s `bundleGraph` is stored in `node.value.results` in the `requestGraph` and is written to the cache as part of the `requestGraph`'s write to cache. If an error occurs during bundling however, the `bundleGraph` is never stored in a `requestGraph` node and as a result isn't written to the cache. This PR adds code to also store the `bundleGraph` in a `requestGraph` node if an error occurs during bundling.

Ideally, we'd eventually like to be able to replicate failed builds from the Parcel cache and this PR is a step towards that.

## 💻 Examples

NA

## 🚨 Test instructions

NA

## ✔️ PR Todo

### DIFF
--- a/packages/core/integration-tests/test/cache.js
+++ b/packages/core/integration-tests/test/cache.js
@@ -21,6 +21,8 @@ import {
   getParcelOptions,
   assertNoFilePathInCache,
   findAsset,
+  bundle,
+  fsFixture,
 } from '@parcel/test-utils';
 import {md} from '@parcel/diagnostic';
 import fs from 'fs';
@@ -29,6 +31,9 @@ import {createWorkerFarm} from '@parcel/core';
 import resolveOptions from '@parcel/core/src/resolveOptions';
 import logger from '@parcel/logger';
 import sinon from 'sinon';
+import {version} from '@parcel/core/package.json';
+import v8 from 'v8';
+import {hashString} from '@parcel/rust';
 
 let inputDir: string;
 let packageManager = new NodePackageManager(inputFS, '/');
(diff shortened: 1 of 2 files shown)


packages/core/integration-tests/test/cache.js
```diff
@@ -6154,6 +6159,75 @@ describe('cache', function () {
...
+
+      let bundleGraphCacheKey = hashString(
+        `${version}:BundleGraph:${
+          JSON.stringify(resolvedOptions.entries) ?? ''
+        }${resolvedOptions.mode}`,
+      );
+
+      let bundleGraphPath = path.join(
+        resolvedOptions.cacheDir,
+        bundleGraphCacheKey + '-0',
```
irismoini: Not sure if this is the best way to determine the bundleGraphPath. Currently being determined [here ](https://github.com/parcel-bundler/parcel/blob/817c7786ef57bfd4810fbcb791c2eb90c78bcbc2/packages/core/cache/src/FSCache.js#L110). Which means if that changes, this code will break, but I have no way of accessing this function directly, so this is currently hardcoded.

lettertwo: Maybe `getFilePath()` should be added to the Cache interface? /cc @JakeLane

JakeLane: Yeah I think that would be reasonable.

Though I think it would be more maintainable for the integration test if we used the cache interface to load the blob instead, as we're duplicating business logic here.

mischnic: Is the goal of the this change to be able to access the graphs via parcel-query? You could use that in the test

irismoini: @mischnic For now we're just looking to confirm that the bundleGraph is written to cache if an error occurs during bundling w/ the ultimate goal of being able to replicate errors from the parcel cache.
@JakeLane I made a change based on what I thought you meant, lmk if you meant something else!

JakeLane (approved):

PR #9368, opened 2023-11-09
Progress messages for writing to cache

thebriando:
# ↪️ Pull Request
Writing to cache on shutdown can take a long time for large graphs, which makes it looks like Parcel is hanging after hitting `ctrl-c`. This moves the shutdown message to `Parcel.js` so it logs out right away, and also adds a progress bar indicator for the blobs written to cache.

This behavior will only happen for large graphs (>5000 nodes) but this number can be changed if needed.

## 💻 Examples

https://github.com/parcel-bundler/parcel/assets/18522836/c27b96b7-a1e7-4173-a8d3-3a615c8b0a4e

## ✔️ PR Todo


lettertwo:
Should the progress bar rendering be the responsibility of the `CLIReporter`? Having the `RequestTracker` do it is circumventing the reporter plugins entirely.

Maybe there should be a new `ReporterEvent` that represents this progress, and the `CLIReporter` can handle those events by rendering the progress bar.

thebriando:
I ran into some problems in the unit tests with using the `report` api but it should work now with the change in #9396

packages/core/core/src/Parcel.js
```diff
@@ -176,8 +176,8 @@ export default class Parcel {
     this.#initialized = false;
 
     await Promise.all([
-      this.#disposable.dispose(),
       await this.#requestTracker.writeToCache(),
+      this.#disposable.dispose(),
```
thebriando: We can't dispose the reporter yet because we're sending events in the `writeToCache` function now so I had to change the order here

mischnic: Isn't
```js
await Promise.all([
      await this.#requestTracker.writeToCache(), // this is just undefiend
      this.#disposable.dispose() // actually a promise
])
```
the same as
```js
await this.#requestTracker.writeToCache();
await this.#disposable.dispose();
```
anyway?

AGawrys (approved):

PR #9369, opened 2023-11-09
Mark previously deferred assets as dirty for symbol prop

mattcompiles:
# ↪️ Pull Request

This PR fixes an issue in incremental symbol propagation + lazy mode builds where you have a lazy asset that depends on an asset which is also lazy itself. The issue occurs because `propagateSymbolsDown` is skipped but `propagateSymbolsUp` runs after the lazy bundle is requested. This can cause an incorrect "X does not export 'Y'" error to show as the `usedSymbolsDown` property has not been populated correctly. 

```mermaid
flowchart LR
    Main -.-> Async
    Main -.-> SharedAsync
    Async --> SharedAsync
```

This fix here is to mark previously deferred assets and their dependencies as changed/dirty so symbol propagation knows to re-process them.  

## ✔️ PR Todo

### DIFF
--- a/packages/core/integration-tests/test/lazy-compile.js
+++ b/packages/core/integration-tests/test/lazy-compile.js
@@ -8,10 +8,23 @@ import {
   assertBundles,
   removeDistDirectory,
   run,
+  fsFixture,
+  overlayFS,
 } from '@parcel/test-utils';
 
 const findBundle = (bundleGraph, nameRegex) => {
-  return bundleGraph.getBundles().find(b => nameRegex.test(b.name));
+  const result = bundleGraph.getBundles().find(b => nameRegex.test(b.name));
+
+  if (!result) {
+    throw new Error(
+      `Couldn't find bundle matching '${nameRegex}'. Available bundles are ${bundleGraph
+        ?.getBundles()
+        .map(b => b.name)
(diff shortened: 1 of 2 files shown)


marcins (approved):
Nice one! Love those problems that take days of digging only to result in a tiny change to fix.

packages/core/integration-tests/test/lazy-compile.js
```diff
@@ -261,4 +274,73 @@ describe('lazy compile', function () {
...
+        main.js:
+          export function shared() {
+            return import('./shared');
+          }
+          export function async() {
+            return import('./async');
+          }
+
+        async.js:
+          // Trigger more deps so a full 'propagateSymbolsUp' pass is executed
```
mischnic, line 294: Do you know why this only happens if the full pass runs? Feels odd to me

mattcompiles: There are other circumstances that can cause this to occur outside of a full pass, however this was the easiest way to reflect in a test. The reason is fails though is because the full pass only applies to propagateSymbolsUp. However, in this case there are nodes that are not re-visited by propagateSymbolsDown and contain stale data.

PR #9372, opened 2023-11-10
Write BundleGraph to cache before shutdown 

AGawrys:
# ↪️ [WIP] Write BundleGraph to cache earlier

In a large application, users must wait upwards of 10 seconds to shut down parcel, due to serialization of large graphs. This PR adds methods `storeResultDeep` and `writeNodeToCache` that store the result in the request tracker and then write the singular node/graph into the cache. 

## 🚨 Q's

- [ ] Can we ensure that users are not waiting an extra 3 seconds during re-bundling times on save, this would be semi counter productive if they wait extra. 
- [ ] Move assetgraph serialization to earlier

## ✔️ PR Todo

### DIFF
--- a/packages/core/core/src/RequestTracker.js
+++ b/packages/core/core/src/RequestTracker.js
@@ -877,7 +877,7 @@ export default class RequestTracker {
     }
   }
   storeResultDeep(nodeId: NodeId, result: mixed, cacheKey: ?string) {
-    // Store result in graph the immediately write to cache
+    // Store result in graph, then immediately write to cache
     this.storeResult(nodeId, result, cacheKey);
     this.writeNodeToCache(nodeId);
   }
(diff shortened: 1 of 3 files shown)


packages/core/core/src/RequestTracker.js
```diff
@@ -1123,6 +1131,25 @@ export default class RequestTracker {
...
+    let node = this.graph.getNode(nodeId);
+    if (!node || node.type !== 'request') {
+      return;
+    }
+
+    let resultCacheKey = node.value.resultCacheKey;
+    if (resultCacheKey != null && node.value.result != null) {
+      this.options.cache.setLargeBlob(
+        resultCacheKey,
+        serialize(node.value.result),
```
JakeLane, line 1147: Does this slow down the build speed? If we're storing nodes in the cache with a heavy serialisation phase, we'll be blocking the event loop for a while. That might be ok if it occurs after the build is available in the dev server

JakeLane:
Very cool! This will help with the ticket I'm working on right now, associated with writing to disk cache when the build has completed and we have idle.

packages/core/core/src/RequestTracker.js
AGawrys: Yeah totally agree, I think we couldn't merge this in as in. Your approach with writing to the disk during idle time makes a lot more sense 😅

PR #9373, opened 2023-11-10
fix: exclude `use-openssl-ca` from threads worker args

### LINKED ISSUE #6569: Can't run Parcel with NODE_OPTIONS set to --use-openssl-ca
# 🐛 bug report

Can't run Parcel commands when `NODE_OPTIONS` is set to `--use-openssl-ca`.

`${PROJECT_DIR}/package.json`
```json
{
  "name": "bulma-playground",
  "version": "1.0.0",
  "description": "",
  "main": "dist/index.js",
  "scripts": {
    "test": "echo \"Error: no test specified\" && exit 1",
    "start": "parcel serve ./src/index.html",
    "build": "parcel build ./src/index.html",
  },
  "author": "",
  "license": "ISC",
  "devDependencies": {
    "bulma": "^0.9.3",
    "parcel": "^2.0.0-beta.3.1"
  }
}
```

`~/.npmrc`
```ini
node-options=--use-openssl-ca
```

## 🤔 Expected Behavior

Parcel will build or serve the project without error.

## 😯 Current Behavior

One of two options:
- Don't pass `process.env` to the `env` worker options. This is the default already per the Node documentation. When omitted, it prevents the Node `Worker::Worker` constructor from parsing the `NODE_OPTIONS` environment again allowing the worker to run.
- Filter out `NODE_OPTIONS` since not all per-process options are valid for V8 isolate creation used for thread workers. This isn't really the best course since the prior option will more transparently propagate the environment to the worker.

## 🔦 Context

On development workstation, we trust an in-house certificate authority/CA on local web server containers to ensure HTTPS works as expected. Node will not connect to those servers without the OpenSSL CA trust, where this CA is trusted, being used. To enable this, we use the `.npmrc` file to set `NODE_OPTIONS` to `--use-openssl-ca`.

This bug would prevent anyone using Parcel when the OpenSSL CA trust is enabled via `NODE_OPTIONS` with the `--use-openssl-ca`.

## 💻 Code Sample

[6XGate/bugs-parcel-with-openssl-failure](https://github.com/6XGate/bugs-parcel-with-openssl-failure)

1. Install everything `npm install`
2. Try building with parcel `npm run build`
3. Try running the test script `node test.js`
4. Try running the test script via npm `npm run testjs`

- Step 2 will fail.
- Step 3 and 3 will work.

## 🌍 Your Environment

| Parcel           |2.0.0-beta.3.1
| Node             |12.22.2
| npm/Yarn         |7.19.1
| Operating System |Ubnuntu/20.04

viceice:
# ↪️ Pull Request

`use-openssl-ca` should not send to `Worker` because it's not supported.
- closes #6569

## 💻 Examples

We use a node wrapper wich calls `node --use-openssl-ca ...` and that fails with default threads worker

## ✔️ PR Todo

### DIFF
--- a/packages/core/workers/src/threads/ThreadsWorker.js
+++ b/packages/core/workers/src/threads/ThreadsWorker.js
@@ -26,7 +26,7 @@ export default class ThreadsWorker implements WorkerImpl {
     onError: ErrorHandler,
     onExit: ExitHandler,
   ) {
-    this.execArgv = execArgv;
+    this.execArgv = execArgv.filter(arg => !/use-openssl-ca/.test(arg));
     this.onMessage = onMessage;
     this.onError = onError;
     this.onExit = onExit;


mischnic:
We already have handling of various Node flags here:
https://github.com/parcel-bundler/parcel/blob/d58e3361ad16c90bb44a2cad8edae31df66c7578/packages/core/workers/src/Worker.js#L50-L65

So we should put this flag there as well

viceice:
That only filters cli args but doesn't filter the `NODE_OPTIONS` env, that's why i added it to the threads worker. It should be passed to the process worker, which can otherwise fail is the `use-openssl-ca` flag is missing

mischnic (dismissed):

mischnic:
However I still get
```
Error: Initiated Worker with invalid NODE_OPTIONS env variable: --use-openssl-ca is not allowed in NODE_OPTIONS
```
with your change because that flag is being passed along here
https://github.com/parcel-bundler/parcel/blob/d58e3361ad16c90bb44a2cad8edae31df66c7578/packages/core/workers/src/threads/ThreadsWorker.js#L38

viceice:
I can try to fix it on Monday

viceice:
sorry, but very busy with other open source projects 🤗

Issue #9377, opened 2023-11-11
Transformer "Invalid specifier type: custom"


grishy:
# 🐛 bug report

I wanted to add a custom resolve for files added to Transformer. But when I try to specify the "custom" type, I get an error. The typescript says that it is possible to use this type.

## 🎛 Configuration (.babelrc, package.json, cli command)

```js
{
  "extends": "@parcel/config-default",
  "transformers": {
    "entrypoint": [
      "./src/parcel-plugins/parcel-transformer.mjs"
    ]
  },
  "resolvers": [
    "./src/parcel-plugins/parcel-resolver.mjs",
    "..."
  ]
}

```

## 🤔 Expected Behavior

I will be able to add deps to file and after that handle it in resolver.

## 😯 Current Behavior

```
npm run build

> grishy-blog@1.0.0 build
> parcel build

🚨 Build failed.

@parcel/core: Failed to resolve '/Users/grishy/projects/lab/blog/content/posts/hello-world/index.md' from './entrypoint'
@parcel/resolver-default: Invalid specifier type: custom

  Error: Invalid specifier type: custom
```

## 💁 Possible Solution

Remove from docs `custom` or allow to handle it.  
Docs:
```ts
export type SpecifierType = "commonjs" | "esm" | "url" | "custom";
// ...
  /**
   * How the specifier should be interpreted.
   *   - esm: An ES module specifier. It is parsed as a URL, but bare specifiers are treated as node_modules.
   *   - commonjs: A CommonJS specifier. It is not parsed as a URL.
   *   - url: A URL that works as in a browser. Bare specifiers are treated as relative URLs.
   *   - custom: A custom specifier. Must be handled by a custom resolver plugin.
   */
  readonly specifierType: SpecifierType;
```

## 💻 Code Sample

```js
// parcel-transformer.mjs
import { Transformer } from '@parcel/plugin';
import utils from '@parcel/utils';
import fs from "@parcel/fs";
import path from "path";

const root = process.cwd();
const nodeFS = new fs.NodeFS();

export default new Transformer({
  async transform({ asset, logger }) {
    logger.info({
      message: `Tranform: ${asset.filePath}`
    });

    const globPath = path.join(root, "./content/posts/**/index.md");
    let files = await utils.glob(globPath, nodeFS, {
      onlyFiles: true,
    });

    for (let file_path of files) {
      logger.info({
        message: `Add post: ${file_path}`
      });

      // const relativePath = path.relative(root, file_path);
[...]

PR #9378, opened 2023-11-11
Change default resolver error message for custom specifiers

mischnic:
See #9377

https://github.com/parcel-bundler/parcel/blob/d1f778f79e8c8fbab6ebc56e40da936f59441439/packages/core/types-internal/src/index.js#L548

Custom is not really invalid, it's just not supported by the default resolver. Might be a more helpful error message


grishy:
Thx!
I think it will help in the future

devongovett (approved):

PR #9379, opened 2023-11-12
Simplify invalidateOnFileChange() in Transformers

grishy:
# ↪️ Pull Request

In [docs](https://parceljs.org/plugin-system/transformer/#DependencyOptions)/code:

```
invalidateOnFileChange(FilePath): void
```

You don't need to add await before

## ✔️ PR Todo

### DIFF
--- a/packages/transformers/posthtml/src/PostHTMLTransformer.js
+++ b/packages/transformers/posthtml/src/PostHTMLTransformer.js
@@ -104,15 +104,10 @@ export default (new Transformer({
       plugins: config.plugins,
     });
 
-    if (res.messages) {
-      await Promise.all(
-        res.messages.map(({type, file: filePath}) => {
-          if (type === 'dependency') {
-            return asset.invalidateOnFileChange(filePath);
-          }
-          return Promise.resolve();
-        }),
-      );
+    for (let {type, file: filePath} of res.messages) {
+      if (type === 'dependency') {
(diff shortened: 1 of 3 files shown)


mischnic:
There's actually one more case here which could just be a for loop without any Promises: https://github.com/parcel-bundler/parcel/blob/d58e3361ad16c90bb44a2cad8edae31df66c7578/packages/transformers/posthtml/src/PostHTMLTransformer.js#L108-L115

grishy:
Oh, I see. I will update it, if you want in this PR.

packages/transformers/posthtml/src/PostHTMLTransformer.js
```diff
@@ -104,15 +104,10 @@ export default (new Transformer({
...
-            return asset.invalidateOnFileChange(filePath);
-          }
-          return Promise.resolve();
-        }),
-      );
+    for (let {type, file: filePath} of res.messages) {
+      if (type === 'dependency') {
+        asset.invalidateOnFileChange(filePath);
+      }
     }
```
mischnic: Let's keep the `if`, not sure if `messages` might be unset sometimes
```suggestion
    if (res.messages) {
      for (let {type, file: filePath} of res.messages) {
        if (type === 'dependency') {
          asset.invalidateOnFileChange(filePath);
        }
      }
    }
```

mischnic (approved):
Thanks!

PR #9383, opened 2023-11-14
Reduce redundancy in the RequestGraph's Request, Env, and Option nodes 

gorakong:
# ↪️ Pull Request

Continued from https://github.com/parcel-bundler/parcel/pull/9360. This PR reduces redundancy in the RequestGraph's Request, Env, and Option nodes by getting rid of the `value` object in each of them, since these can be derived from their IDs. On a large app, this nets us a further ~7.1% reduction in the size of the serialized RequestGraph.

### Request Nodes

Before:
```
  {
    id: 'f20cf057afed92f9:Main',
    type: 'request',
    value: { id: 'f20cf057afed92f9:Main', type: 'path_request' },
    invalidateReason: 0
  }
```

After (Request Nodes still maintain a `request` type, but their subtypes can be distinguished via `requestType`):
```
 {
    id: 'f20cf057afed92f9:Main',
    type: 'request',
    requestType: 'path_request',
    invalidateReason: 0
  }
```

### Env Nodes
Before:
```
  {
    id: 'env:NODE_ENV',
    type: 'env',
    value: { key: 'NODE_ENV', value: 'production' }
  }
```
After:
```
 { id: 'env:NODE_ENV', type: 'env', value: 'production' }
```
### Option Nodes
Before:
```
  {
    id: 'option:shouldBuildLazily',
    type: 'option',
    value: { key: 'shouldBuildLazily', hash: 'false' }
  },
  ```
After:
```
{ id: 'option:shouldBuildLazily', type: 'option', hash: 'false' }
```

### CODE REVIEW

File: packages/core/core/src/RequestTracker.js [hunk 1]
```diff
@@ -84,45 +88,61 @@ type SerializedRequestGraph = {|
...
+  | 'parcel_config_request'
+  | 'path_request'
+  | 'dev_dep_request'
+  | 'asset_request'
+  | 'config_request'
+  | 'write_bundles_request'
+  | 'package_request'
+  | 'write_bundle_request'
+  | 'validation_request'
+  | string;
```
gorakong: I'm iffy about this `string` here. I added it to appease Flow because we use a `mock_request` type in the RequestTracker's unit tests, but I wonder if I should just flow-ignore those instead.

mattcompiles: I agree, let's flow ignore the test rather than the main code.

packages/core/core/src/utils.js
```diff
@@ -235,3 +236,7 @@ export function toInternalSymbols<T: {|loc: ?SourceLocation|}>(
     ]),
   );
 }
+
+export function keyFromEnvOrOptionContentKey(contentKey: ContentKey): string {
+  return contentKey.slice(contentKey.indexOf(':') + 1);
```
mattcompiles: I think it'd be better to have an explicit function for both `env` and `option`, that way you don't need to search the string dynamically. For example, env nodes are always `contentKey.slice('env:'.length)` which saves the `indexOf` check. You could also co-locate them with the `nodeFrom*` functions so they can stay in sync.

packages/core/core/src/RequestTracker.js
```diff
@@ -84,45 +88,60 @@ type SerializedRequestGraph = {|
...
   +type: 'request',
-  value: StoredRequest,
+  +requestType?: RequestType,
   invalidateReason: InvalidateReason,
+  result?: mixed,
+  resultCacheKey?: ?string,
+  hash?: string,
 |};
+
+export type RequestType =
```
JakeLane: 💯

packages/core/core/test/RequestTracker.test.js
```diff
@@ -16,13 +16,15 @@ describe('RequestTracker', () => {
     let tracker = new RequestTracker({farm, options});
     await tracker.runRequest({
       id: 'abc',
+      // $FlowFixMe[incompatible-call]
```
JakeLane: nit: It might be better to build a mock typecast so there's no any type created by the suppression, which would mean the test won't break if the types change

Something like:
```
async function mockRunRequest(
      tracker: RequestTracker,
      request: {...Request, type: Request.type | 'mock_request', ...},
      opts?: ?RunRequestOpts,
    ): Promise<TResult> {
      return tracker.runRequest(request, opts);
    }
```

JakeLane (approved):

packages/core/core/test/RequestTracker.test.js
gorakong: Good call @JakeLane! Let me know what you think about how I mocked `runRequest` (wasn't sure if that was the best way)

gorakong: @JakeLane just FYI I ended up changing all the `mock_request`s to be one of the types in the `RequestType` union. When I tried mocking the types (https://github.com/parcel-bundler/parcel/pull/9383/commits/53fdf808757413c9fb5d3dbf2de2e4f0f7ab91a5), the only way I found that appeased Flow was to do it in a way that essentially did the same thing (as above) 😅

PR #9384, opened 2023-11-15
 Break up request graph cache serialisation and run after build completion

JakeLane:
# ↪️ Pull Request

Currently in Jira Frontend, we have a major problem with developers becoming confused when shutting down parcel, as it looks like the process has crashed. This is because the request tracker is serialised and written to cache, when the parcel cli is shutdown.

To avoid this, we can utilise the idle time in the watch process by serialising after the build has completed. It's also important to break up the serialisation phase, as we need to be able to interrupt at any time for a new build. I picked a sane constant value to split up the nodes in groups that took around ~10ms to process on my machine, which should free up the event loop well enough for most developers.

## 🚨 Test instructions

1. Start a parcel build in watch mode
2. Wait for build to complete
3. Observe cache serialising to disk
4. Close parcel (observe near-instant shutdown)


packages/core/core/src/Parcel.js
```diff
@@ -361,6 +358,7 @@ export default class Parcel {
         createValidationRequest({optionsRef: this.#optionsRef, assetRequests}),
         {force: assetRequests.length > 0},
       );
+      await this.#requestTracker.writeToCache();
```
JakeLane: This is after the build has reported as complete. Open to suggestions for a more suitable location.

packages/core/core/src/RequestTracker.js
```diff
@@ -1080,34 +1082,72 @@ export default class RequestTracker {
   }
 
   async writeToCache() {
-    let cacheKey = getCacheKey(this.options);
-    let requestGraphKey = hashString(`${cacheKey}:requestGraph`);
-    let snapshotKey = hashString(`${cacheKey}:snapshot`);
+    const cacheKey = getCacheKey(this.options);
+    const requestGraphKey = `requestGraph:${hashString(cacheKey)}`;
```
JakeLane: I changed these as it's easier to debug what's happening if the cache folder has useful names

devongovett:
Won't this end up spending _more_ time/cpu serializing in general because it runs after every build instead of only when shutting down Parcel? Sure it starts earlier when you actually are shutting down but that's overall a pretty rare event vs doing a regular rebuild, in which case there's no need to save to disk.

JakeLane:
> Won't this end up spending _more_ time/cpu serializing in general because it runs after every build instead of only when shutting down Parcel? Sure it starts earlier when you actually are shutting down but that's overall a pretty rare event vs doing a regular rebuild, in which case there's no need to save to disk.

Yeah, it's definitely a compromise as we'll need to use CPU more often, but it shouldn't impact active waiting time for devs by slowing down the shutdown of the process. The parcel build should be usable to the developer while this serialisation is happening (requests will be served) and it's interruptible, so new changes to source should cancel these writes.

I'm realising now my implementation isn't ideal here, as it doesn't get interrupted in the way I expected. I'll push up an update tomorrow which should make more sense.

devongovett:
Even better would be to only serialize the parts of the graph that changed now that you have chunking...

JakeLane:
> Even better would be to only serialize the parts of the graph that changed now that you have chunking...

I wasn't able to come up with a way to know if it's safe to skip serialisation of a chunk efficiently :/

Ideally we need to know if a node has changed since last build and since it's possible for a node to be deleted, we'd have to have some sort of tracking for changed nodes.

packages/core/cache/src/FSCache.js
```diff
@@ -105,25 +105,34 @@ export class FSCache implements Cache {
...
+              : contents.subarray(
+                  i * WRITE_LIMIT_CHUNK,
+                  (i + 1) * WRITE_LIMIT_CHUNK,
+                ),
+          ),
+        );
+      }
+    }
+
+    // If there's already a file following this chunk, it's old and should be removed
```
marcins: Does it matter if it had extra chunks after that, or does that just become "garbage"?

i.e. if you had chunks 0, 1, 2, 3 (unlikely given the sizes - but still..), and next time have 0,1, you'll delete 2 but leave 3 dangling?

packages/core/cache/src/FSCache.js
```diff
@@ -105,25 +105,34 @@ export class FSCache implements Cache {
...
+    if (await this.fs.exists(this.#getFilePath(key, chunks))) {
```
marcins: Note that async existence checking is long deprecated in Node, and can lead to race conditions: https://nodejs.org/api/fs.html#fsexistspath-callback (i.e. the recommended approach would be to just try and `unlink` and ignore failure)

packages/core/cache/src/LMDBCache.js
```diff
@@ -111,27 +112,47 @@ export class LMDBCache implements Cache {
     return Buffer.concat(await Promise.all(buffers));
   }
 
-  async setLargeBlob(key: string, contents: Buffer | string): Promise<void> {
+  async setLargeBlob(
```
marcins, line 113: Is there any elegant way to share the implementation between the two Cache types as (AFAICT) they're identical?

packages/core/core/src/RequestTracker.js
```diff
@@ -846,6 +846,8 @@ export class RequestGraph extends ContentGraph<
   }
 }
 
+const NODES_PER_BLOB = 2 ** 14;
```
marcins, line 872: How did we arrive at this number?

packages/core/core/src/RequestTracker.js
```diff
@@ -1167,15 +1202,10 @@ export default class RequestTracker {
...
     );
-    report({
-      type: 'cache',
-      phase: 'write',
-      total,
-      size: this.graph.nodes.length,
-    });
-    report({type: 'cache', phase: 'end', total, size: this.graph.nodes.length});
 
     await Promise.all(promises);
```
marcins: How many concurrent promises will be typical here for a Jira cache? `Promise.all` with a large set, especially when writing files or doing other async IO stuff, can have sub-optimal performance. If the concurrency is large enough, you might have better results using `async/queue` with a concurrency limit set.

marcins (commented):
Just left some general feedback / questions for now.

packages/core/cache/src/LMDBCache.js
mischnic: Instantiate a `FSCache` instance inside lmdb cache and forward calls of `setLargeBlob` to that?

packages/core/core/src/RequestTracker.js
mischnic: (We also have `PromiseQueue` in the utils for exactly this use case you haven't seen it in the codebase yet.)

marcins: > (We also have PromiseQueue in the utils for exactly this use case you haven't seen it in the codebase yet.)

Ah, yeah, it's even touched in this PR  😅 

I was thinking of Jira where I have used `async/queue` in the past for this..

packages/core/cache/src/FSCache.js
JakeLane: Yeah I left it dangling as the intention was purely to cut the edge case of joining two different chunks together. In retrospect, it's not complex to just delete everything so I'll add that

packages/core/cache/src/LMDBCache.js
JakeLane: That sounds good to me, will clean this up

packages/core/core/src/RequestTracker.js
JakeLane: I profiled locally on my machine on how long it took to serialise `n` nodes. I tuned `n` with binary search until I reached approximately ~50 ms serialisation time per blob. The goal is to free up the event loop for a reasonable amount of time for user perception.

I'll document this on the constant so it can be tuned in the future if required.

packages/core/core/src/RequestTracker.js
```diff
@@ -1108,75 +1115,124 @@ export default class RequestTracker {
...
   }
 
-  async writeToCache() {
+  async writeToCache(signal?: AbortSignal) {
     let cacheKey = getCacheKey(this.options);
-    let requestGraphKey =
-      hashString(`${cacheKey}:requestGraph`) + '-RequestGraph';
-    let snapshotKey = hashString(`${cacheKey}:snapshot`);
+    let requestGraphKey = `requestGraph-${hashString(cacheKey)}`;
+    let snapshotKey = `snapshot-${hashString(cacheKey)}`;
```
mattcompiles: Not a biggy, but no point hashing the cache key twice.

packages/core/core/src/RequestTracker.js
```diff
@@ -1108,75 +1115,124 @@ export default class RequestTracker {
...
-      let resultCacheKey = node.resultCacheKey;
-      if (resultCacheKey != null && node.result != null) {
-        promises.push(
-          this.options.cache.setLargeBlob(
-            resultCacheKey,
-            serialize(node.result),
+      queue
+        .add(() =>
+          serialiseAndSet(
+            `requestGraph-nodes-${i}-${hashString(cacheKey)}`,
```
mattcompiles: Same here, we could re-use the hashed cache key from earlier.

packages/core/core/src/RequestTracker.js
```diff
@@ -1209,20 +1265,41 @@ async function loadRequestGraph(options): Async<RequestGraph> {
...
+
+    let i = 0;
+    let nodePromises = [];
+    while (
+      await options.cache.hasLargeBlob(
+        `requestGraph-nodes-${i}-${hashedCacheKey}`,
+      )
+    ) {
+      nodePromises.push(
+        getAndDeserialize(`requestGraph-nodes-${i}-${hashedCacheKey}`),
```
mattcompiles: We generate this string at least 3 different times. Could move it to a function so it's clear they're tied together?

packages/core/core/src/RequestTracker.js
```diff
@@ -1108,75 +1115,124 @@ export default class RequestTracker {
...
-    report({type: 'cache', phase: 'end', total, size: this.graph.nodes.length});
+    queue
+      .add(() =>
+        this.options.inputFS.writeSnapshot(
+          this.options.projectRoot,
+          snapshotPath,
+          opts,
+        ),
+      )
+      .catch(() => {});
```
mattcompiles: What's the point of this `.catch`?

packages/core/core/src/RequestTracker.js
```diff
@@ -1108,75 +1115,124 @@ export default class RequestTracker {
...
+        cacheableNodes[i] = newNode;
+      } else {
+        cacheableNodes[i] = node;
+      }
+    }
+
+    for (let i = 0; i * NODES_PER_BLOB < cacheableNodes.length; i += 1) {
+      if (
+        this.cachedRequestsLastChunk !== null &&
+        i < this.cachedRequestsLastChunk
```
mattcompiles: The assumption here is that nodes never change or get re-ordered?

packages/core/core/src/RequestTracker.js
```diff
@@ -1108,75 +1115,124 @@ export default class RequestTracker {
...
+      ) {
         continue;
```
mattcompiles: Maybe it'd be nicer, to calculate `i` with this expression rather than just calling `continue` at the start of the loop?

JakeLane: Refactored this a bit, let me know what you reckon

packages/core/core/src/RequestTracker.js
JakeLane: Yep added a comment here. This idea was from @devongovett in the sync

packages/core/core/src/RequestTracker.js
JakeLane: This is to ensure we don't crash parcel if we interrupt a serialisation with the watcher (e.g. a new build is triggered)

mattcompiles: Is this required even though we're not awaiting the result?

mischnic: This is the "handler" to make sure it's not an ["unhandled promise rejection"](https://nodejs.org/api/process.html#process_event_unhandledrejection)

JakeLane: I'll bring this back, I think that's the main reason it's required to have

mattcompiles (approved):
Good job bud 👏. This was a tricky one to get through. It all makes logical sense to me now. Happy to merge assuming you've tested it extensively.

JakeLane:
- [x] Discuss mutability of request nodes with @mattcompiles

packages/core/core/src/RequestTracker.js
```diff
@@ -269,6 +273,7 @@ export class RequestGraph extends ContentGraph<
       optionNodeIds: this.optionNodeIds,
       unpredicatableNodeIds: this.unpredicatableNodeIds,
       invalidateOnBuildNodeIds: this.invalidateOnBuildNodeIds,
+      cachedRequestChunks: this.cachedRequestChunks,
```
mattcompiles, line 276: Cool idea, this means we'll be able to skip work between different sessions of Parcel.

packages/core/core/src/RequestTracker.js
```diff
@@ -345,6 +351,9 @@ export class RequestGraph extends ContentGraph<
     for (let parentNode of parentNodes) {
       this.invalidateNode(parentNode, reason);
     }
+
+    // If the node is invalidated, the cached request chunk on disk needs to be re-written
+    this.cachedRequestChunks.delete(Math.floor(nodeId / NODES_PER_BLOB));
```
mattcompiles: I think we may need to do this from `Request.completeRequest` as well? I think it's possible that a node is invalidated but its request hasn't been re-run yet. Once it is re-run we'd want to update it's result in the cache.

mattcompiles (approved):
Approval for the new invalidation changes.

PR #9385, opened 2023-11-15
Allow background.type for MV2

mischnic:
No need to be overly restrictive here. Though we don't actually use that value in the transformer right now

https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/background


101arrowz:
If we're going to support module-type background scripts, we should probably manually handle `background.scripts` like we do for `background.service_worker`, rather than leaving it in `DEP_LOCS`. We'd need to set the `env.sourceType` to the `type` property to `program.background.type == 'module' ? 'module' : 'script'`, just like we have for service worker extensions.

PR #9386, opened 2023-11-15
VSC Extension JSON schema

mischnic:
Add autocomplete/popup descriptions/always-on schema validation for .parcelrc and `package.json#targets`

https://github.com/parcel-bundler/parcel/assets/4586894/53813bce-3833-494e-972f-716010dbffc3

Plus descriptions

Bildschirm­foto 2023-11-15 um 21 15 20

Two aspects that are suboptimal are (could be solved in the future with a custom LSP for the JSON as well): 
- No autocompletion for package names (which should in theory also be restricted to the `parcel-{type}-*` pattern)
- You get two errors for invalid parcelrc/targets: one from Parcel and then a warning from the JSON schema.

Bildschirm­foto 2023-11-15 um 21 23 04


celinanperalta (approved):
lgtm! nice addition

thebriando (approved):

packages/utils/parcelforvscode/schemas/package-targets.schema.json
```diff
@@ -0,0 +1,167 @@
...
+      },
+      "additionalProperties": {
+        "type": "string"
+      }
+    },
+    "target": {
+      "type": "object",
+      "properties": {
+        "context": {
+          "description": "The context property defines what type of environment to build for. This tells Parcel what environment-specific APIs are available, e.g. the DOM, Node filesystem APIs.",
```
lukaw3d, line 31: Some of this JSON duplicates what https://github.com/parcel-bundler/parcel/blob/9d6da2c/packages/core/core/src/TargetDescriptor.schema.js does, but descriptions seem novel.

I think if you add descriptions into TargetDescriptor.schema.js, you can just generate this file. (my script for generating json from TargetDescriptor.schema.js https://github.com/lukaw3d/parcel-bundler-json-schemas/blob/main/main.js)

packages/utils/parcelforvscode/schemas/parcelrc.schema.json
```diff
@@ -0,0 +1,94 @@
+{
```
lukaw3d, line 1: And this is related to https://github.com/parcel-bundler/parcel/blob/9d6da2c/packages/core/core/src/ParcelConfig.schema.js

PR #9391, opened 2023-11-16
Check Asset Graph for source asset when dependency retargeting

imbrian:
# ↪️ Pull Request

Removing this check caused a regression when upgrading a project from 2.10.2, where the content key did not yet exist [here](https://github.com/parcel-bundler/parcel/blob/v2/packages/core/core/src/BundleGraph.js#L316).

## ✔️ PR Todo

### DIFF
--- a/packages/core/core/src/BundleGraph.js
+++ b/packages/core/core/src/BundleGraph.js
@@ -310,7 +310,10 @@ export default class BundleGraph {
                       local,
                       loc: reexportAllLoc,
                     });
-                    if (node.value.sourceAssetId != null) {
+                    if (
+                      node.value.sourceAssetId != null &&
+                      assetGraph.hasContentKey(node.value.sourceAssetId)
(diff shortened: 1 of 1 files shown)


mischnic:
Can you figure out what the dependency in question is and why it doesn't have it's parent asset?

For that rewriting to work, both he `symbol.set` before the if and the if have to execute, otherwise you will get invalid code. So this mostly silences the build failure but probably doesn't actually work at runtime.

imbrian:
> Can you figure out what the dependency in question is and why it doesn't have it's parent asset?
> 
> For that rewriting to work, both he `symbol.set` before the if and the if have to execute, otherwise you will get invalid code. So this mostly silences the build failure but probably doesn't actually work at runtime.

Yeah, I'd like to be able to reproduce this with a test case before considering the PR. Just opening draft, since I spoke to @thebriando about it yesterday.

PR #9396, opened 2023-11-20
Move registerCoreWithSerializer to its own file

thebriando:
# ↪️ Pull Request

This moves the `registerCoreWithSerializer` function to a different module to prevent circular dependency problems

## ✔️ PR Todo

### DIFF
--- a/packages/core/core/src/registerCoreWithSerializer.js
+++ b/packages/core/core/src/registerCoreWithSerializer.js
@@ -1,0 +1,32 @@
+// @flow
+import {Graph} from '@parcel/graph';
+import {ctorToName, registerSerializableClass} from './serializer';
+import AssetGraph from './AssetGraph';
+import BundleGraph from './BundleGraph';
+import ParcelConfig from './ParcelConfig';
+import {RequestGraph} from './RequestTracker';
+import Config from './public/Config';
+import packageJson from '../package.json';
+
+export function registerCoreWithSerializer() {
+  const packageVersion: mixed = packageJson.version;
+  if (typeof packageVersion !== 'string') {
+    throw new Error('Expected package version to be a string');
+  }
+
+  // $FlowFixMe[incompatible-cast]
+  for (let [name, ctor] of (Object.entries({
+    AssetGraph,
+    Config,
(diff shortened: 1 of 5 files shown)


packages/core/core/src/registerCoreWithSerializer.js
```diff
@@ -0,0 +1,32 @@
...
+  for (let [name, ctor] of (Object.entries({
+    AssetGraph,
+    Config,
+    BundleGraph,
+    Graph,
+    ParcelConfig,
+    RequestGraph,
+    // $FlowFixMe[unclear-type]
+  }): Array<[string, Class<any>]>)) {
+    if (ctorToName.has(ctor)) {
```
devongovett: Was this function somehow getting called multiple times now? Looks like this wasn't there before

thebriando: Ah we had a different check for this before https://github.com/parcel-bundler/parcel/pull/9396/files#diff-9575fdb466517de49a2fcba34666f6aafe312d93c8c66978f8d7e6b27464b792L51-L55

and I thought I needed to change it bc the function is called in `Parcel.js` and `worker.js` and I was running into this error https://github.com/parcel-bundler/parcel/blob/0ab3f0f96ae3123048501acbc5e6a19d2fa83b7e/packages/core/core/src/serializer.js#L11-L13

but the error was happening bc I had an unused import that was calling the function again. I'll change it back

packages/core/core/src/serializer.js
```diff
@@ -5,7 +5,7 @@ import {serializeRaw, deserializeRaw} from './serializerCore';
 export {serializeRaw, deserializeRaw} from './serializerCore';
 
 const nameToCtor: Map<string, Class<any>> = new Map();
-const ctorToName: Map<Class<any>, string> = new Map();
+export const ctorToName: Map<Class<any>, string> = new Map();
```
devongovett:
```suggestion
const ctorToName: Map<Class<any>, string> = new Map();
```

don't need to export that anymore

devongovett (approved):
lgtm

PR #9403, opened 2023-11-24
Change inline-requires to only run when optimizing

marcins:
# ↪️ Pull Request

Just some minor cleanup.

Changes the plugin so that rather than always running in production mode, will instead only run when doing an optimized build.

Also removes some of the verbose logging, replacing measurement with tracing.

Fixes the example config in `README.md`.

## 🚨 Test instructions

Tested manually that it still works with prod builds.

## ✔️ PR Todo

- [n/a ] Added/updated unit tests for this change

### DIFF
--- a/packages/optimizers/inline-requires/README.md
+++ b/packages/optimizers/inline-requires/README.md
@@ -8,12 +8,9 @@ Add this optimizer to run _first_ (before minification), for JS bundles.
 
 ```json
 {
-    "optimizers": {
-        "*.js": {
-            "parcel-optimizer-inline-requires",
-            "..."
-        }
-    }
+  "optimizers": {
+    "*.js": ["@parcel/optimizer-inline-requires", "..."]
+  }
 }
 ```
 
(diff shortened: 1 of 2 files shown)


mattcompiles (approved):
Nice one

mischnic:
Previously `options.mode !== 'production'` coincidentally also ensured `scopeHoist == true` for most cases. But you should probably add a test for scope hoisting explicitly, that should fix the tests

mischnic:
As far as I understand it, the optimizer only works with scope-hoisting, so checking `bundle.env.scopeHoisting` would probably be good as well. But you probably simply never use it in that situation 🤷‍♂️

marcins:
> As far as I understand it, the optimizer only works with scope-hoisting, so checking `bundle.env.scopeHoisting` would probably be good as well. But you probably simply never use it in that situation 🤷‍♂️

It works regardless of scope hoisting or not - it's actually probably _less effective_ with scope hoisting because there's fewer `parcelRequire` calls to defer. A more "complete" solution would also need to handle scope hoisted assets and defer their execution too.

PR #9404, opened 2023-11-24
Fix CI

mischnic:
Revert #9401 and bump only lightningcss. Bumping all deps had caused:
- In the tests, we somehow do `watcher.getEventsSince(dirThatDoesntExist)`. `@parcel/watcher@>=2.3.0` throws an error here. The dir in question is e.g. `parcel/packages/core/integration-tests/test/input/cbong0yqey6`
- Bumping `@swc/helpers` breaks some assertions that do `output.includes("=>")`
- potentially more

We should bump the whole lockfile at some point anyway though

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -2,16 +2,6 @@
 # yarn lockfile v1
 
 
-"@aashutoshrathi/word-wrap@^1.2.3":
-  version "1.2.6"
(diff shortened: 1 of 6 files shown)


yarn.lock
marcins, line 1: Ah, 🤦 - I missed that this file had changed so significantly when reviewing the change because GH didn't show the diff. Nice catch! Explains why everything was failing on my branch too since I had merged these changes in.

marcins (approved):

Issue #9405, opened 2023-11-25
Vue 3.3.0 broke custom languages/preprocessors in scripts


mischnic:
When bumping Vue to 3.3.0 or newer (specifically `@vue/compiler-sfc`) in the Parcel monorepo, the Parcel integration tests fail 
```sh
$ yarn test:integration -g 'should produce a vue bundle using preprocessors'

  4) vue
       should produce a vue bundle using preprocessors:
     SyntaxError: [vue/compiler-sfc] Missing semicolon. (3:6)

packages/core/integration-tests/test/integration/vue-preprocessors/pre-processors.vue
30 |  <script lang="coffee">
31 |  module.exports =
32 |    data: ->
   |        ^
33 |      msg: 'Hello from coffee!'
34 |  </script>
```

This change was introduced in Vue 3.3.0-alpha.10: https://github.com/vuejs/core/blob/main/CHANGELOG.md#330-alpha10-2023-04-17

PR #9406, opened 2023-11-25
Bump some deps

mischnic:
Fixed:
- Bump Babel, didn't appear to cause any problems
- The new swc helpers added a comment containing `=>`, so minify the result before we check if a `=>` was emitted in the code
- tsc 5.2 apparently emits `namespace` instead of `module` now

Pinned dependencies:
- Pin Vue because the transformer broke with 3.3.0 and custom script preprocessors: https://github.com/parcel-bundler/parcel/issues/9405
- In the tests, we somehow do `watcher.getEventsSince(dirThatDoesntExist)`. `@parcel/watcher@>=2.3.0` throws an error here. The dir in question is e.g. `parcel/packages/core/integration-tests/test/input/cbong0yqey6`

### DIFF
--- a/package.json
+++ b/package.json
@@ -73,7 +73,9 @@
   },
   "packageManager": "yarn@1.22.19",
   "_": "See issue #9405 for the Vue problem",
+  "__": "The watcher complains about non-existing dirs in the integration tests on >=2.3.0",
   "resolutions": {
+    "@parcel/watcher": "~2.2.0",
     "@vue/compiler-sfc": "~3.2.47",
     "vue": "~3.2.47"
   }
--- a/package.json
+++ b/package.json
@@ -71,5 +71,10 @@
     "type": "opencollective",
     "url": "https://opencollective.com/parcel"
   },
-  "packageManager": "yarn@1.22.19"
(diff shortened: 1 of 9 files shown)


mischnic:
The problem with the watcher (if it were bumped, this PR just pins it to a working version), is that with the OverlayFS, that `input/<hash>` dir only exists on the MemoryFS side, not the NodeFS, then `MemoryFS#getEventsSince` tries to get events for both FSs, but the folder exists in one of them.

One solution would be this
```js
class OverlayFS
  async getEventsSince(
    dir: FilePath,
    snapshot: FilePath,
    opts: WatcherOptions,
  ): Promise<Array<Event>> {
    let writableEvents = this.writable.existsSync(path.dirname(snapshot))
      ? await this.writable.getEventsSince(dir, snapshot, opts)
      : [];
    let readableEvents = this.readable.existsSync(path.dirname(snapshot))
      ? await this.readable.getEventsSince(dir, snapshot, opts)
      : [];
    return [...writableEvents, ...readableEvents];
  }
}
```

marcins (approved):

PR #6789, opened 2021-08-28
Revert adding COEP/COOP headers

mischnic:
Revert #6404

There are many more people using third-party resources than SharedArrayBuffer... https://github.com/parcel-bundler/parcel/issues/6561

Related: https://github.com/parcel-bundler/website/issues/884

Issue #6561, opened 2021-07-07
Cross Origin Resource Policy with Parcel nightly build version (local env)


jasonhieu1905:
# 🐛 bug report

Having an **<img** tag include external src causes the CORS issue.

```
<img
                src="https://images.unsplash.com/photo-1593642532842-98d0fd5ebc1a?crop=entropy&cs=srgb&fm=jpg&ixid=Mnw3NjY4NXwxfDF8YWxsfDF8fHx8fHwyfHwxNjI1NjQzNDQ1&ixlib=rb-1.2.1&q=85"
                width="32px"
                height="32px"
            />
```

## 😯 Current Behavior

![3](https://user-images.githubusercontent.com/13063615/124722177-fd3ea780-df33-11eb-9fa3-d01f6d37d6cc.PNG)

## 💁 Possible Solution
added **crossOrigin** to **img** tag
```
<img
                src="https://images.unsplash.com/photo-1593642532842-98d0fd5ebc1a?crop=entropy&cs=srgb&fm=jpg&ixid=Mnw3NjY4NXwxfDF8YWxsfDF8fHx8fHwyfHwxNjI1NjQzNDQ1&ixlib=rb-1.2.1&q=85"
                width="32px"
                height="32px"
                crossOrigin="anonymous"
            />
```

## 🌍 Your Environment

Parcel          nightly version 2.0.0-nightly.753 |  local server.

mischnic:
You can do what https://github.com/parcel-bundler/parcel/issues/6499#issue-926561639 describes, to override/unset these headers.

jasonhieu1905:
@mischnic : thanks for the response, i am just curious why we should do that. It's not happened in previous version. and the issue depends on different image source.

mischnic:
Things like `SharedArrayBuffer` will require these headers very soon: https://web.dev/coop-coep/ (in Firefox & Chrome)

jasonhieu1905:
I did not get that. How could I apply proxy without backend server. I am using react with parcel.

mischnic:
If you add a `.proxyrc.js` in your project, Parcel will use it for its dev server. This way, you can override the headers

barisbikmaz:
I have the same problem. We're building office addins. For that we need to include an external office.js file which dynamically loads other js files.

I can use crossorigin="" on the office.js file which works

```
<script
  type="text/javascript"
  src="https://appsforoffice.microsoft.com/lib/1/hosted/office.js"
[...]

PR #9410, opened 2023-11-27
added support for SharedArrayBuffers

dioptre:
# ↪️ Pull Request

Adding required headers to use SharedArrayBuffers in parcel serve

## 💻 Examples

https://stackoverflow.com/questions/70535752/enable-sharedarraybuffer-on-localhost

## 🚨 Test instructions

const shared = new SharedArrayBuffer(1024);

### DIFF
--- a/packages/reporters/dev-server/src/Server.js
+++ b/packages/reporters/dev-server/src/Server.js
@@ -35,6 +35,8 @@ import fresh from 'fresh';
 
 export function setHeaders(res: Response) {
   res.setHeader('Access-Control-Allow-Origin', '*');
+  res.setHeader('Cross-Origin-Opener-Policy', 'same-origin');
+  res.setHeader('Cross-Origin-Embedder-Policy', 'require-corp');
   res.setHeader(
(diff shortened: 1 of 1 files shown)


mischnic:
We already tried this before: https://github.com/parcel-bundler/parcel/pull/6404 and then reverted again in https://github.com/parcel-bundler/parcel/pull/6789 because many more people rely on some third party code/asset than on SAB: https://github.com/parcel-bundler/parcel/issues/6561

https://github.com/parcel-bundler/website/issues/884

lettertwo:
Maybe there's a path forward here along the lines of this comment from @Banou26  https://github.com/parcel-bundler/parcel/issues/6561#issuecomment-907495414

> One thing you could do if you want to make it a default behavior is detect the use of those features in the code and then apply the COOP/COEP headers just like Parcel already does with service workers or webworkers's usage of their register function serviceWorkerContainer.register.
If we use SharedArrayBuffer or performance., then apply the COOP/COEP headers, otherwise, don't.

PR #9420, opened 2023-12-05
Unstable File Invalidations

gorakong:
# ↪️ Pull Request

This PR allows passing in file paths through FS Event objects as an option through Parcel's node API. These are then converted to events and invalidated in the `RequestTracker`.  This should support the ability to invalidate `node_module` packages in CI when combined with a list that is generated by comparing an old vs new copy of the `yarn.lock` (of the project being built).

## 💻 Examples

Example usage:

```
const bundler = new Parcel({
  entries: 'src/entry.tsx',
  shouldDisableCache: false,
  unstableFileInvalidations: [
    {path: 'node_modules/react-dom/profiling.js', type: 'create'},
    {
      path: 'node_modules/react/cjs/react.production.min.js',
      type: 'create',
    },
    {
      path: 'src/foo.tsx',
      type: 'update',
    },
  ]
});
```

## 🚨 Test instructions
(These reference the example above):
- Did a build of `src/entry.tsx` without `unstableFileInvalidations`
- Then did another build of the same entry (without clearing the cache) with `unstableFileInvalidations` passed in as an option
- Verified that both `node_modules/react-dom/profiling.js`, `node_modules/react/cjs/react.production.min.js`, and `src/foo.tsx` are all rebuilt

### DIFF
--- a/packages/core/parcel/src/cli.js
+++ b/packages/core/parcel/src/cli.js
@@ -101,6 +101,9 @@ const commonOptions = {
     },
     [],
   ],
+  '--node-module-invalidations <path>': [
+    'Comma separated list of file paths of node_module packages to be invalidated',
+  ],
 };
 
 var hmrOptions = {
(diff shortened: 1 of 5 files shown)


packages/core/parcel/src/cli.js
```diff
@@ -213,6 +216,7 @@ if (!args[2] || !program.commands.some(c => c.name() === args[2])) {
 program.parse(args);
 
 function runCommand(...args) {
+  console.log('hello run1');
```
JakeLane: hello

JakeLane (commented):
Awesome work!

mattcompiles (approved):

Issue #4332, opened 2020-03-15
Parcel 2 doesn't detect changes in linked dependencies.


samuelgozi:
# 🐛 bug report

When working on a dependency locally by linking it, Parcel 2 doesn't detect changes and therefore doesn't rebuild. You also need to remove the cache manually in order for it to work on re-runs.

## 🎛 Configuration (.babelrc, package.json, cli command)

Package.json
```json
{
	"name": "parcel-tamplate",
	"version": "1.1.0",
	"description": "My personal Parcel Template",
	"main": "public/index.html",
	"source": "src/index.html",
	"author": "Samuel Elgozi <samu.elgozi@gmail.com>",
	"license": "none",
	"private": true,
	"scripts": {
		"dev": "parcel src/index.html",
		"build": "parcel build src/index.html"
	},
	"targets": {
		"main": {
			"includeNodeModules": true
		}
	},
	"devDependencies": {
		"@babel/plugin-proposal-class-properties": "^7.8.3",
		"babel-eslint": "^10.0.2",
		"eslint": "^6.0.1",
		"less": "^3.9.0",
		"parcel": "^2.0.0-alpha.3.2"
	},
	"dependencies": {
		"firebase-auth-lite": "^0.4.1"
	}
}

```

The commands I use:
```
parcel src/index.html
```

## 🤔 Expected Behavior

Like in version 1, Parcel should detect changes in linked dependencies, invalidate the cache and rebuild.

## 😯 Current Behavior

No changes are made, which require manual removal of `.parcel-cache` and then building again.

## 🔦 Context

I use Parcel to work on npm modules often. Without detecting the changes in linked packages there is no reason for me to use parcel. In fact, it becomes painful to do so...

## 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Parcel           | 2.0.0-alpha.3.2 |
| Node             | v13.10.1|
| Yarn         |1.22.1
| Operating System | MacOs Catalina 10.15.3 |

DeMoorJasper:
Could you create a repo with some instructions on how I'd reproduce this? Would help a lot in fixing this

samuelgozi:
Ok, here you go: https://github.com/samuelgozi/Parcel2-issue-4332
Further instructions can be found in the repo's README file.

DeMoorJasper:
@samuelgozi thanks for creating the repo, I can reproduce it pretty consistently, will try and figure out why it happens and hopefully fix it

[...]

PR #9424, opened 2023-12-07
Configurable watch root

jondlm:
related to: #4332

# ↪️ Pull Request

Adds a new `--watch-dir` CLI flag that allows users to specify a directory higher up than the project root. This allows Parcel in watch mode to detect changes to symlinked dependencies.

This is especially useful for monorepos that have lockfiles for each of the projects and where all projects can't share the same project root for various reasons.

## 💻 Examples

Suppose we have a project shaped like this:
```
.
├── package.json
└── packages
    ├── app
    │   ├── index.js              <- imports lib via the symlink in node_modules
    │   ├── node_modules
    │   │   └── lib -> ../../lib
    │   ├── package.json
    │   └── yarn.lock
    └── lib
        ├── index.js
        ├── package.json
        └── yarn.lock
```

We can run `parcel server --watchDir ../..` from within `packages/app` and changes to `packages/lib/index.js` will trigger HMR updates, etc.

## 🚨 Test instructions
I have included an integration test that exercises this new behavior. I confirmed that without the `watchDir` option Parcel won't detect changes but it _does_ with the proper watch directory set.

## ✔️ PR Todo


packages/core/parcel/src/cli.js
```diff
@@ -73,6 +73,8 @@ const commonOptions = {
   '--config <path>':
     'specify which config to use. can be a path or a package name',
   '--cache-dir <path>': 'set the cache directory. defaults to ".parcel-cache"',
+  '--watch-dir <path>':
```
jondlm, line 76: I put this on `commonOptions` but maybe it only makes sense on `watch` and `serve`?

mischnic: Indeed

jondlm: I'll move it!

devongovett: I think it needs to be where you have it because this should also be used for caching one-shot builds.

packages/core/core/src/Parcel.js
```diff
@@ -389,7 +389,7 @@ export default class Parcel {
     let resolvedOptions = nullthrows(this.#resolvedOptions);
     let opts = getWatcherOptions(resolvedOptions);
     let sub = await resolvedOptions.inputFS.watch(
-      resolvedOptions.projectRoot,
+      resolvedOptions.watchDir,
       (err, events) => {
         if (err) {
```
devongovett, line 394: we should probably also update this so it works for one-shot builds https://github.com/parcel-bundler/parcel/blob/e4f0a69ac764ef927749a1392e2834af8b0bef99/packages/core/core/src/RequestTracker.js#L1217

idk if watchRoot is the most obvious name in that case but not sure about another one. projectRoot is a different thing and affects more stuff so would need to be something different.

packages/core/parcel/src/cli.js
jondlm: Ahh okay, I'll move it back

packages/core/core/src/Parcel.js
jondlm: Done! Yeah the name because less intuitive when used in this context :|

devongovett (approved):

PR #9431, opened 2023-12-11
Publish bundle-stats-cli and parcel-query

thebriando:
# ↪️ Pull Request
Publishes `bundle-stats-cli` and `parcel-query`

## ✔️ PR Todo


packages/dev/query/package.json
```diff
@@ -1,7 +1,18 @@
 {
   "name": "parcel-query",
```
thebriando: Should `parcel-query` be published under the `@parcel` namespace?

mischnic: Yeah, might be a good idea for consistency

mischnic:
- The `parcel` and `create-react-app` CLIs are listed here so that their `bin`  is rewritten from `src/` to `lib/`
https://github.com/parcel-bundler/parcel/blob/7a0bf13328831cd15ec3108490f5fb7a3fdfca6d/gulpfile.js#L32-L35

- `parcel-query` needs
```
  "bin": {
    "parcel-query": "...."
  },
```

- For `bundle-stats`, ~`cli.js`~ `bin.js` has to be moved into `src` like in the other CLIs and also needs an if-check before including babel-register, like https://github.com/parcel-bundler/parcel/blob/ba1d813030d38441b4b0b11672f9a0d56502e92f/packages/dev/query/src/bin.js#L1

packages/dev/bundle-stats-cli/package.json
```diff
@@ -1,7 +1,18 @@
 {
   "name": "@parcel/bundle-stats-cli",
```
mischnic:
```suggestion
  "name": "@parcel/bundle-stats",
```

This would have been the only CLI with "cli" in it's name 🤷

mischnic:
Oops, I meant

> For bundle-stats, `bin.js` has to be moved into `src`...

gulpfile.js
```diff
@@ -32,6 +32,8 @@ const paths = {
   packageJson: [
     'packages/core/parcel/package.json',
     'packages/utils/create-react-app/package.json',
+    'packages/dev/query/package.json',
+    'packages/dev/bundle-stats-cli/package.json',
```
thebriando, line 36: Hmm it works when I run `parcel-query` from src, but after running `yarn build` i get these lib/src errors

```
> parcel-query --help
/Users/bdo/parcel/packages/core/core/src/BundleGraph.js:3
import type {
^^^^^^

SyntaxError: Cannot use import statement outside a module
    at internalCompileFunction (node:internal/vm:73:18)
    at wrapSafe (node:internal/modules/cjs/loader:1178:20)
    at Module._compile (node:internal/modules/cjs/loader:1220:27)
    at Module._extensions..js (node:internal/modules/cjs/loader:1310:10)
    at Module.load (node:internal/modules/cjs/loader:1119:32)
    at Module._load (node:internal/modules/cjs/loader:960:12)
    at Module.require (node:internal/modules/cjs/loader:1143:19)
    at require (node:internal/modules/cjs/helpers:110:18)
    at Object.<anonymous> (/Users/bdo/parcel/packages/dev/query/lib/cli.js:70:20)
    at Module._compile (node:internal/modules/cjs/loader:1256:14)

Node.js v18.17.0
```

mischnic: You need to handle these imports in `deep-imports.js` just like the ones for `AssetGraph` and so on

currently it does `query/lib/cli.js` --hardcoded-import-> `core/src/BundleGraph.js`, but it should also be `lib` when built

Speaking of that, should `@parcel/core` be a peerdependency of parcel-query?

mischnic:
Also this should be just `import ... from "@parcel/query"` (both regarding the package name change and removing the subpath)

``` 
Error ------------------------ packages/dev/bundle-stats-cli/src/cli.js:16:26

Cannot resolve module `parcel-query/src/index.js`. [cannot-resolve-module]

   16| import {loadGraphs} from 'parcel-query/src/index.js';
                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

mischnic:
Once these two problems are fixed, it should be good to go 👍

packages/dev/query/src/deep-imports.js
```diff
@@ -1,33 +1,73 @@
...
+        .fromProjectPathRelative,
     }
   : {
       AssetGraph: require('@parcel/core/src/AssetGraph.js').default,
-      BundleGraph: require('@parcel/core/src/BundleGraph.js').default,
+      // $FlowFixMe(incompatible-cast)
+      BundleGraph: require('@parcel/core/src/BundleGraph.js'),
       RequestTracker: require('@parcel/core/src/RequestTracker.js'),
+      LMDBCache: require('@parcel/cache/src/LMDBCache.js').LMDBCache,
+      // $FlowFixMe(incompatible-cast)
```
thebriando: Added these $FlowFixMe's for cases where there's `import type` instead of `import typeof`, I was running into some incompatible call errors that I wasn't sure how to resolve..

mischnic: I've pushed a fix for that. The problem was that various values you reexported in deep-imports were actually types that don't actually exist at runtime, which is why flow complained as they aren't actually there. 

It's fine to do deep type imports here, because they are erased at build time, so don't refer to the `src` folder once published.

mischnic (approved):

PR #9438, opened 2023-12-13
Bump rimraf version to ^5.05

adelchan07:
Bump rimraff version in response to a security vulnerability within Atlassian products

# ↪️ Pull Request

## ✔️ PR Todo


mischnic:
- We might as well upgrade to the latest version then (v5)
- A breaking change of v4  was the the methods return promises now, so the `promisify` in `packages/utils/create-react-app/src/cli.js` shouldn't be done anymore
- (There is still a dependency on the old version in the monorepo root)

mischnic:
The flow types in `flow-typed/npm/rimraf_v2.x.x.js` are wrong now, it should be something like this now
```ts
  declare module.exports: {
    (f: string, opts?: Options): Promise<boolean>,
```

And the gulpfile has to be changed: 
```diff
diff --git gulpfile.js gulpfile.js
index 4b524e694..61d885410 100644
--- gulpfile.js
+++ gulpfile.js
@@ -59,7 +59,10 @@ class TapStream extends Transform {
 }
 
 exports.clean = function clean(cb) {
-  rimraf('packages/*/*/lib/**', cb);
+  rimraf('packages/*/*/lib/**').then(
+    () => cb,
+    err => cb(err),
+  );
 };

```

mischnic (approved):

Issue #7425, opened 2021-12-10
Typescript's parameter decorators not working


enrico-calzavara:
# 🐛 bug report
Typescript's parameters decorators are not correctly transpiled using parcel's default transpiler (SWC).

## 🎛 Configuration (.babelrc, package.json, cli command)

main.ts
```ts
import {Container, inject, injectable} from "inversify";

interface Warrior {
    fight(): string;
    sneak(): string;
}

interface Weapon {
    hit(): string;
}

interface ThrowableWeapon {
    throw(): string;
}

@injectable()
class Katana implements Weapon {
    public hit() {
        return "cut!";
    }
}

@injectable()
class Shuriken implements ThrowableWeapon {
    public throw() {
        return "hit!";
    }
}

@injectable()
class Ninja implements Warrior {

    private _katana: Weapon;
    private _shuriken: ThrowableWeapon;

    public constructor(
        @inject("Weapon") katana: Weapon,
        @inject("ThrowableWeapon") shuriken: ThrowableWeapon
    ) {
        this._katana = katana;
        this._shuriken = shuriken;
    }

    public fight() { return this._katana.hit(); }
    public sneak() { return this._shuriken.throw(); }

}

const myContainer = new Container();
myContainer.bind<Warrior>("Warrior").to(Ninja);
myContainer.bind<Weapon>("Weapon").to(Katana);
myContainer.bind<ThrowableWeapon>("ThrowableWeapon").to(Shuriken);

export { myContainer };
```
package.json
```json
{
  "name": "parcel2-app",
  "version": "1.0.0",
  "browserslist": [
    "defaults"
  ],
  "description": "",
  "scripts": {
    "build": "npx parcel build index.html",
    "test": "echo \"Error: no test specified\" && exit 1"
  },
  "author": "",
  "license": "ISC",
  "dependencies": {
    "inversify": "^4.13.0",
    "reflect-metadata": "^0.1.13"
  },
  "devDependencies": {
    "parcel": "^2.0.1",
    "typescript": "4.1"
  }
}
```
tsconfig.json
```json
{
  "compilerOptions": {
    "emitDecoratorMetadata": true,
    "experimentalDecorators": true,
    "isolatedModules": true,
    "lib": ["es6", "dom"],
    "module": "commonjs",
    "moduleResolution": "Node",
    "sourceMap": true,
    "target": "es5",
    "types": ["reflect-metadata"]
  },
  "exclude": [
    "node_modules"
  ]
}
```

[...]

PR #9470, opened 2024-01-05
Enable Stage 3 decorators when experimental decorators are off

jrandolf:
# ↪️ Pull Request

This PR turns on Stage 3 decorators by default when experimental decorators are off. A priori, this is a non-breaking change since experimental decorator users are unaffected and non-experimental users do not have decorators in the first place.

## ✔️ PR Todo

## Related issues

- https://github.com/parcel-bundler/parcel/issues/7425


mischnic:
Our approach so far has been to not enable proposals *by default* before they're finished (stage 4), or browsers have themselves enabled them by default via `shipped_proposals`:
https://github.com/jun-sheaf/parcel/blob/e283c44bd27ca10330ec80c785fae7e7002aab02/packages/transformers/js/core/src/lib.rs#L296

Because what would be breaking is a change in the proposal later on.

jrandolf:
At stage 3, it's expected the community is implementing decorators, not just browser vendors. In fact, if the community doesn't, then decorators will **eventually** be scrapped. That's the meaning of stage 3.

See https://tc39.es/process-document/.

PR #9472, opened 2024-01-08
Revert more CI changes to centos job

mischnic:
That stupid CentOS image...

I've actually verified that it runs successfully now


devongovett (approved):
Did you cancel the CI jobs or did it fail for some reason?

mischnic:
I cancelled the CI job to free up runners for testing the nightly workflow (which I then also cancelled after the centos job ran through)

PR #9487, opened 2024-01-13
Statically evaluate constants referenced by macros

devongovett:
This PR enables macro calls to include statically analyzable constants. For example:

```js
import { foo } from './macro' with {type: 'macro'};

const common = {
  foo: 'bar'
};

const result1 = foo({...common, bar: 'test'});
const result2 = foo({...common, bar: 'other'});
```

Constants must be declared with `const` to ensure that they cannot be mutated, as this would be much harder to analyze. Object and array patterns are supported in the declaration in addition to simple identifiers. The result of a macro may also be stored in a const variable, and later used in another macro call.

This also implements support for member expressions and optional chaining to access properties of statically known objects and arrays.

It may be helpful to review with whitespace disabled FYI.

### DIFF
--- a/packages/core/integration-tests/test/macros.js
+++ b/packages/core/integration-tests/test/macros.js
@@ -604,4 +604,120 @@ describe('macros', function () {
       assert(res.includes(`output${i}="2a2300bbd7ea6e9a"`));
     }
   });
+
+  it('should throw a diagnostic when a constant is mutated', async function () {
+    await fsFixture(overlayFS, dir)`
+      index.js:
+        import { hashString } from "@parcel/rust" with { type: "macro" };
+        const object = {foo: 'bar'};
+        object.foo = 'test';
+        output = hashString(object.foo);
+    `;
+
+    try {
+      await bundle(path.join(dir, '/index.js'), {
+        inputFS: overlayFS,
+        mode: 'production',
+      });
+    } catch (err) {
+      assert.deepEqual(err.diagnostics, [
+        {
+          message: 'Could not statically evaluate macro argument',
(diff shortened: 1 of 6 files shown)


mischnic:
> Constants must be declared with const to ensure that they cannot be mutated, as this would be much harder to analyze.

Is this also handled?
```js
const common = {
  foo: 'bar'
};
common.foo = 'baz'; // <---
const result1 = foo({...common, bar: 'test'});
```

devongovett:
@mischnic I was just thinking about that as well. Added some code to make assignments to constant objects an error if that constant is later accessed by a macro. In addition, if a constant object (or a property of a constant object that is also an object) is passed as an argument of a function call, it is also marked as potentially mutated.

packages/core/integration-tests/test/macros.js
```diff
@@ -604,4 +604,120 @@ describe('macros', function () {
...
+        const bar = 'bar';
+        const object = {foo: bar};
+        doSomething(bar); // ok (string)
+        doSomething(object.foo); // ok (evaluates to a string)
+        doSomething(object); // error (object could be mutated)
+        output = hashString(object.foo);
+
+        const object2 = {foo: bar, obj: {}};
+        doSomething(object2.obj); // error (object could be mutated)
+        output2 = hashString(object2);
```
devongovett, line 691: The above could potentially be fine if `object2.obj` is not accessed in a macro, e.g. if `object2.foo` were passed here. But tracking mutations for individual values would be a bit more complicated than only tracking it for constants as a whole.

mischnic (approved):

PR #9489, opened 2024-01-14
Add library bundler

### LINKED ISSUE #7254: Option to disable bundling for building libraries
### Discussed in https://github.com/parcel-bundler/parcel/discussions/6845

<div type='discussions-op-text'>

<sup>Originally posted by **Thoud** August 31, 2021</sup>
Hey everyone,

currently I'm trying to create a React component library which gets published to a registry and imported by another project.
My goal is to have a library which builds the source code in the same file structure as in src.

The idea behind this is that the project which imports the library has the possibility to only import a single component out of there instead of importing the whole thing. In addition to that it would be nice if the components (when importing each other) would not get bundled up into huge files with all their imported files but instead being build into individual files and only being referenced (just like node_modules but on a component level). Here a short example of what I mean:

Component B imports Component A and C.
All three files get compiled to individual files but instead of dumping the code from Component A and C into B and then compile Component B into a big file , it would be great if they would get compiled separately but in a way that component B references the other two, so that no code duplication occurs.

I hope that my explanation makes any sense, if not please let me know which parts are unclear and I will give my best to clarify them.
I'm happy for any help or hints to the solution :D

</div>

# 🙋 feature request

Add an option to disable bundling, only transpile. While I'm aware that this feature is very big, I still wanted to request it to hear your thoughts.

## 🤔 Expected Behavior

When building a library I expect running `npx parcel build [input]` only to compile source, not bundle it.

## 😯 Current Behavior

Parcel always bundles (with the exception of being able to exclude node modules).

## 💁 Possible Solution

Parcel could add a new option (e.g. `--no-bundle`) that could also be the enabled via `package.json` somehow, possibly as the default for libraries. This feature would also normalize glob inputs, so maybe Parcel could also start recognizing glob values for the `source` field:

```json
"source": "./src/**/*.ts"
```

but I haven't thought it through.

## 🔦 Context

When building a library, bundling is unwanted, I want to compile the source files in the same structure as they are in the source directory.

devongovett:
Fixes #7254

This PR implements a separate bundler plugin for libraries that maps each input file to a separate output bundle, i.e. not bundling at all. This is desirable for libraries because bundlers handle separate files better for tree shaking (via the `sideEffects` field in package.json). It's especially helpful for component libraries that import CSS, because only the CSS for used components is included rather than everything.

This is currently implemented as a separate bundler plugin (`@parcel/bundler-library`) that you'd need to manually install and configure to use instead of the default bundler. Maybe this could be the default behavior in v3 though.

This is not fully complete (e.g. doesn't handle dynamic import in libraries which isn't too common), but it is able to successfully build React Spectrum.


packages/core/core/src/BundleGraph.js
```diff
@@ -1142,19 +1146,24 @@ export default class BundleGraph {
...
         },
         this._graph.getNodeIdByContentKey(dep.id),
         bundleGraphEdgeTypes.references,
       );
+
+      if (bundle) {
+        resolved = potential.find(a => a.type === bundle.type);
+      }
+      resolved ||= potential[0];
     }
```
devongovett, line 1168: The changes in this file are to handle the case where a dependency resolves to _multiple_ bundles. In that case, we need to prioritize the one with the same type as the parent bundle. We also prioritize reference edges over bundle group resolution

packages/packagers/js/src/ScopeHoistingPackager.js
```diff
@@ -134,25 +134,10 @@ export class ScopeHoistingPackager {
...
-          }
-        }
-
-        this.externals.set(relativeBundlePath(this.bundle, b), symbols);
+      for (let b of this.bundleGraph.getReferencedBundles(this.bundle, {
+        recursive: false,
+      })) {
+        this.externals.set(relativeBundlePath(this.bundle, b), new Map());
       }
     }
```
devongovett, line 142: This logic ensures that we add imports for all referenced bundles. But we now wait to determine which symbols to import until they are used (below in `addExternal`). This way, we only include symbols user _in this bundle_ instead of across all bundles.

This also switches to only including bundles non-recursively, otherwise we'd end up hoisting imports for things like CSS even if the library had `"sideEffects": false`. For a library, there's no need to hoist even with side effects. We leave that to the application bundler consuming the library.

packages/packagers/js/src/ScopeHoistingPackager.js
```diff
@@ -371,13 +354,20 @@ export class ScopeHoistingPackager {
...
         asset,
         exportAs,
         symbol,
         exportSymbol,
       } of this.bundleGraph.getExportedSymbols(entry)) {
         if (typeof symbol === 'string') {
+          // Only export default if the module has a namespace (e.g. commonjs).
+          if (hasNamespace && exportAs !== '*') {
+            continue;
+          }
```
devongovett: Previously, if a commonjs module was the entry point of a bundle, this could result in invalid identifiers being added to ESM exports, e.g. `exports["foo-bar"] = "foo bar"` (`foo-bar` is not a valid identifier in JS). Now we only export `default` and the importing code uses that as a namespace object.

packages/packagers/js/src/ScopeHoistingPackager.js
```diff
@@ -964,7 +1039,7 @@ ${code}
     } else if (!symbol) {
       invariant(false, 'Asset was skipped or not found.');
     } else {
-      return symbol;
+      return replacements?.get(symbol) || symbol;
     }
```
devongovett, line 1036: Need to check `replacements` in case a symbol got aliased to an external.

packages/packagers/js/src/ScopeHoistingPackager.js
```diff
@@ -324,7 +309,6 @@ export class ScopeHoistingPackager {
 
       if (
         asset.meta.shouldWrap ||
-        this.isAsyncBundle ||
```
devongovett, line 327: This should be handled by the isAssetReferenced check below. The library bundler makes every bundle "async", but we don't want to wrap those.

packages/core/integration-tests/test/output-formats.js
```diff
@@ -1299,7 +1299,6 @@ describe('output formats', function () {
       );
 
       let ns = await run(b);
-      assert.deepEqual(ns.test, true);
```
devongovett, line 1302: Related to https://github.com/parcel-bundler/parcel/pull/9489#discussion_r1512008209. Before we used to output the namespace both on default and without, which is what node does too. But if there are exports with invalid identifiers this is not possible. Not sure what to do here. Filter the exports? Seems better to only export on `default` but idk if this will break something.

devongovett:
used the `export {a as "foo-bar"}` syntax for invalid identifiers (https://github.com/tc39/ecma262/pull/2154). still not supported by typescript, but hopefully TS won't be parsing the JS in node_modules anyway. https://github.com/microsoft/TypeScript/pull/49297

mischnic (approved):

devongovett:
@mischnic ugh, looks like the string export syntax breaks webpack 4, which is still very widely used unfortunately. might need to revert that...

mischnic:
Looks like we don't have a choice...

PR #9505, opened 2024-01-25
feat: add getCurrentPackageManager

bestlyg:
# ↪️ Pull Request

## 💻 Examples

my repo is monorepo by pnpm-workspace protocal and is installed with pnpm.
It will create pnpm-lock yaml in root directory.
When I run parcel build with types in subpackage, it will auto install @parcel/transformer-typescript-types with yarn.

I check the source code and find the result is that the library of package manager will detect current package manager by lockfile.
If the current work directory has no lockfile, it will run in ['yarn', 'pnpm', 'npm'] order.
It is not my expect. I want to it works for pnpm.

So I want to add a function named getCurrentPackageManager, it will detect the current package manager with `process.env.npm_config_user_agent`.

## ✔️ PR Todo

### DIFF
--- a/packages/core/package-manager/src/installPackage.js
+++ b/packages/core/package-manager/src/installPackage.js
@@ -174,13 +174,16 @@ async function determinePackageInstaller(
     return new Yarn();
   }
 
-  const name = getCurrentPackageManager().name;
-  if (name === 'npm') {
-    return new Npm();
-  } else if (name === 'yarn') {
-    return new Yarn();
-  } else if (name === 'pnpm') {
-    return new Pnpm();
+  const currentPackageManager = getCurrentPackageManager();
+  if (currentPackageManager) {
+    const name = currentPackageManager.name;
+    if (name === 'npm') {
+      return new Npm();
+    } else if (name === 'yarn') {
+      return new Yarn();
+    } else if (name === 'pnpm') {
+      return new Pnpm();
+    }
   }
 
   if (await Yarn.exists()) {
--- a/packages/core/package-manager/src/installPackage.js
+++ b/packages/core/package-manager/src/installPackage.js
@@ -174,16 +174,13 @@ async function determinePackageInstaller(
     return new Yarn();
   }
 
-  const currentPackageManager = getCurrentPackageManager();
-  if (currentPackageManager) {
-    const name = currentPackageManager.name;
-    if (name === 'npm') {
-      return new Npm();
-    } else if (name === 'yarn') {
-      return new Yarn();
-    } else if (name === 'pnpm') {
-      return new Pnpm();
-    }
+  let currentPackageManager = getCurrentPackageManager()?.name;
+  if (currentPackageManager === 'npm') {
+    return new Npm();
+  } else if (currentPackageManager === 'yarn') {
+    return new Yarn();
+  } else if (currentPackageManager === 'pnpm') {
+    return new Pnpm();
   }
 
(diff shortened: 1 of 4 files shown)


mischnic:
`projectRoot` in that function either contains a lockfile or a `.git` (and only if none of these files exist in any parent directory, is is the current working directory). So in your case, that should already be the correct folder and find your lockfile.

bestlyg:
> `projectRoot` in that function either contains a lockfile or a `.git` (and only if none of these files exist in any parent directory, is is the current working directory). So in your case, that should already be the correct folder and find your lockfile.

Thanks for your reply.
I make a [demo](https://github.com/bestlyg/parcel-demo) to explain the issue.
reprocude steps:
1. `pnpm i` in root directory.
2. `cd packages/core`
3. `pnpm build`

you can find that the parcel will install dependencies with yarn.
image

The `projectRoot` is correct owing to it contains`.git`.
Special case: I setted the lockfile=false in `.npmrc`, because I do not need lockfile in my actual project.
The `determinePackageInstaller` function only reolve lockfile.
so I want to add the function, it can help parcel find current package manager that is called.

mischnic:
👍 

- You need to run `yarn flow` and fix all type errors. The function declaration should be this
```js
export default function getCurrentPackageManager(
  userAgent: ?string = process.env.npm_config_user_agent,
): ?{|name: string, version: string|} {
  if (userAgent == null) {
    return;
  }

```
- Instead of `log:agent`, just hardcode some current pnpm/npm/yarn user agent strings and check them. The CI probably doesn't have pnpm installed for example

bestlyg:
> 👍
> 
> * You need to run `yarn flow` and fix all type errors. The function declaration should be this
> 
> ```js
> export default function getCurrentPackageManager(
>   userAgent: ?string = process.env.npm_config_user_agent,
> ): ?{|name: string, version: string|} {
>   if (userAgent == null) {
>     return;
>   }
> ```
> 
> * Instead of `log:agent`, just hardcode some current pnpm/npm/yarn user agent strings and check them. The CI probably doesn't have pnpm installed for example

Thanks for your reply.
I have updated the type of `getCurrentPackageManager ` and hardcode the test file.
I have run `yarn flow` and `yarn test` and reported no error.

mischnic:
There are some eslint errors in packages/core/package-manager/test/getCurrentPackageManager.test.js

bestlyg:
> There are some eslint errors in packages/core/package-manager/test/getCurrentPackageManager.test.js

Thanks for remind.
I have fixed it.

packages/core/package-manager/src/installPackage.js
```diff
@@ -173,6 +174,18 @@ async function determinePackageInstaller(
...
+  if (currentPackageManager) {
+    const name = currentPackageManager.name;
+    if (name === 'npm') {
+      return new Npm();
+    } else if (name === 'yarn') {
+      return new Yarn();
+    } else if (name === 'pnpm') {
+      return new Pnpm();
+    }
+  }
```
mischnic:
```suggestion
  let currentPackageManager = getCurrentPackageManager()?.name;
  if (currentPackageManager === 'npm') {
    return new Npm();
  } else if (currentPackageManager === 'yarn') {
    return new Yarn();
  } else if (currentPackageManager === 'pnpm') {
    return new Pnpm();
  }
```

mischnic (approved):
Thanks!

PR #9512, opened 2024-01-30
Allow constants to be used in macros inside a call expression

devongovett:
This fixes a small bug where if a constant were passed to a macro, and the result of that macro was immediately called as a function, a "could not statically evaluate macro argument" error would be thrown. This was due to `in_call` being set to true by the outer call expression, and not resetting it to false when calling a known macro function.


devongovett:
merging to prepare for another refactor

PR #9518, opened 2024-02-01
Add lazy/eager cache key to avoid invalid change when switching modes

JakeLane:
# ↪️ Pull Request

Simple change to consider whether lazy mode is enabled/disabled. This is required because launching with `--lazy`, then re-launching without `--lazy` will cause build errors.

## 🚨 Test instructions

1. `yarn parcel --lazy`
2. Shut down server
3. `yarn parcel`
4. Observe no build error

### DIFF
--- a/packages/core/core/src/PackagerRunner.js
+++ b/packages/core/core/src/PackagerRunner.js
@@ -665,7 +665,8 @@ export default class PackagerRunner {
         bundleGraph.getHash(bundle) +
         JSON.stringify(configResults) +
         JSON.stringify(globalInfoResults) +
-        this.options.mode,
+        this.options.mode +
+        `${this.options.shouldBuildLazily ? 'lazy' : 'eager'}`,
     );
   }
 
--- a/packages/core/core/src/PackagerRunner.js
+++ b/packages/core/core/src/PackagerRunner.js
@@ -666,7 +666,7 @@ export default class PackagerRunner {
         JSON.stringify(configResults) +
         JSON.stringify(globalInfoResults) +
         this.options.mode +
-        `${this.options.shouldBuildLazily ? 'lazy' : 'eager'}`,
+        (this.options.shouldBuildLazily ? 'lazy' : 'eager'),
     );
   }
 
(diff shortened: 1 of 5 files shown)


JakeLane:
There's also a cache key in `packages/core/core/src/UncommittedAsset.js`, which I did not change. Let me know if this is acceptable, I assumed there was no need but my mental model may be incorrect.

packages/core/core/src/PackagerRunner.js
```diff
@@ -665,7 +665,8 @@ export default class PackagerRunner {
         bundleGraph.getHash(bundle) +
         JSON.stringify(configResults) +
         JSON.stringify(globalInfoResults) +
-        this.options.mode,
+        this.options.mode +
+        `${this.options.shouldBuildLazily ? 'lazy' : 'eager'}`,
```
marcins: /nit but this doesn't really need to be a template string.. (but I get that you probably just copied what you did in all the other places, no big deal..)

packages/core/integration-tests/test/cache.js
```diff
@@ -6788,4 +6790,43 @@ describe('cache', function () {
     let res = await run(build.bundleGraph);
     assert.deepEqual(res, {default: 'foo'});
   });
+
+  it('invalidates correctly when switching from lazy to eager modes', async function () {
```
marcins, line 6794: I assume this was a failing test before your changes?

marcins (approved):
Nice! Was the approach just to find everywhere that `mode` was used in a cache key and add it there?

packages/core/integration-tests/test/cache.js
JakeLane: yep, it was!

JakeLane:
> Nice! Was the approach just to find everywhere that `mode` was used in a cache key and add it there?

Yep, I also looked at the Parcel version variable

packages/core/core/src/PackagerRunner.js
JakeLane:
```suggestion
        this.options.shouldBuildLazily ? 'lazy' : 'eager',
```

mattcompiles (approved):

PR #8151, opened 2022-05-27
Apply unresolved mark to inserted `undefined` identifiers

mischnic:
Fixes #8143

- swc already has a type alias for `(JsWord, SyntaxContext)`, so just use that instead of our own
- Apply unresolved mark to inserted `undefined` identifiers. swc's DCE pass relies on this

PR #8029, opened 2022-04-29
Bump swc

mischnic:
Closes https://github.com/parcel-bundler/parcel/issues/7911

- [x] Some changes are needed for https://github.com/swc-project/swc/pull/4436 (when we add previously undefined variables)
- [x] There were some breaking changes already, but let's wait for https://github.com/swc-project/swc/issues/4488
- [x] Add test for https://github.com/parcel-bundler/parcel/issues/7911
- [x] https://github.com/swc-project/swc/issues/4608 is blocking the tests

PR #9520, opened 2024-02-03
Remove decl_collector pass and use SWC's unresolved_mark instead

devongovett:
The decl_collector pass stored all declarations in a HashSet so we could determine if a variable is a global (i.e. undeclared). This is used to find `require` calls and much more. Turns out we don't really need this anymore because SWC has an `unresolved_mark` that it sets on these identifiers in its resolver. Removing this saves us a pass, reduces some memory usage, and simplifies the code by reducing the number of places we need to keep up to date.


mischnic:
Are we sure this is always correct? I just remember the multiple times where swc's unresolved and global mark don't really do what they say they do: https://github.com/parcel-bundler/parcel/pull/8151 and https://github.com/parcel-bundler/parcel/pull/8029

devongovett:
I think it may have been improved. I noticed that SWC's minifier and Turbopack both use this same check. I did also try to remove the second pass of the resolver after preset env that you added before as well, but that was still broken. Otherwise it seems to pass all of our tests here and we can test on some apps too.

mischnic (approved):

PR #9532, opened 2024-02-15
Remove the worker dev dep build cache in packaging

mattcompiles:
# ↪️ Pull Request

There is currently a possible race condition in requests where it's possible for them to run in a different order on the main thread then they were sent from the worker thread. This causes an issue with dev dep requests as they have a build cache that reduces the information on the request if it's already run before to reduce overall serialisation time. 

This PR removes the build cache from just the PackageRunner dev dep requests and leaves it on the Transformation ones. The thinking here is that the Packaging ones don't get a lot of benefit from the cache but seem to be the only ones that fail. This isn't ideal but I think it should make the error pretty much disappear for very little performance loss. 

## ✔️ PR Todo


thebriando (approved):

mattcompiles:
Closing in favour of #9636

PR #9547, opened 2024-02-27
Expose watcher ignore and watcher backend options

Nikola-3:
# ↪️ Pull Request

Adds CLI options to set watcher backend and paths/globs for watcher to ignore. Note ignoring only means relevant events are not passed from OS to js runtime, the events will still be watched natively as there is no way to completely ignore them.

## 💻 Examples

Now it is possible to run:
* `parcel build ./index.html --watcher-backend fsevents`
* `parcel build ./index.html --watcher-ignore ./.yarn ./**/.git`

## ✔️ PR Todo


packages/core/parcel/src/cli.js
```diff
@@ -472,10 +477,15 @@ async function normalizeOptions(
...
     return input.split(',').map(value => value.trim());
   };
+
   return {
     shouldDisableCache: command.cache === false,
     cacheDir: command.cacheDir,
     watchDir: command.watchDir,
+    baseWatcherOptions: {
+      backend: command.watcherBackend,
+      ignore: command.watcherIgnoreDirs ?? ['.git', '.hg'],
```
mattcompiles: I feel like `.git` and `.hg` should always be ignored regardless. Maybe just leave them hardcoded in `getWatcherOptions`?

packages/core/core/src/resolveOptions.js
```diff
@@ -189,6 +189,9 @@ export default async function resolveOptions(
     shouldTrace: initialOptions.shouldTrace ?? false,
     cacheDir,
     watchDir,
+    baseWatcherOptions: initialOptions.baseWatcherOptions ?? {
```
marcins: /nit or /bikeshed - should this just be called `watcherOptions`? I know it's not _all_ the watcher options, but from a consumer of the API point of view it's the options you can specify.

packages/core/types/index.js
```diff
@@ -285,6 +285,11 @@ export type DetailedReportOptions = {|
   assetsPerBundle?: number,
 |};
 
+export type WatcherOptions = {|
+  ...BaseWatcherOptions,
+  ignore: BaseWatcherOptions['ignore'],
```
marcins: huh, neat, I didn't realise you could reference part of a type like that with Flow!

packages/core/core/src/resolveOptions.js
Nikola-3: The reason I didn't go for `watcherOptions` is because I feel like it's somewhat misleading/suggestive precisely since there are other options (e.g. `watchDir`) that configure the watcher. Reading `base...` implies there may be others. Happy to change if you still feel `watcherOptions` is good enough.

packages/core/types/index.js
Nikola-3: I was hoping this would make it non-nullable, but apparently, the preferred way is to wrap the type with `$NonMaybeType<MaybeType>`. It's not as neat though, the vscode server doesn't display the particular type.

packages/core/parcel/src/cli.js
```diff
@@ -75,6 +75,11 @@ const commonOptions = {
   '--cache-dir <path>': 'set the cache directory. defaults to ".parcel-cache"',
   '--watch-dir <path>':
     'set the root watch directory. defaults to nearest lockfile or source control dir.',
+  '--watcher-ignore-dirs [path]': `list of directories watcher should not be tracking for changes. defaults to ['.git', '.hg']`,
```
devongovett: I think they don't need to be directories, the watcher also accepts globs. so maybe just `--watcher-ignore`? or maybe `--watch-ignore` for consistency with `--watch-dir`?

packages/core/core/test/test-utils.js
```diff
@@ -18,6 +18,7 @@ cache.ensure();
 export const DEFAULT_OPTIONS: ParcelOptions = {
   cacheDir: path.join(__dirname, '.parcel-cache'),
   watchDir: __dirname,
+  baseWatcherOptions: {},
```
devongovett: Should we just call it `watcherOptions`? Why the base?

packages/core/parcel/src/cli.js
```diff
@@ -75,6 +75,11 @@ const commonOptions = {
...
+  '--watcher-backend': new commander.Option(
+    '--watcher-backend <name>',
+    'set watcher backend',
+  ).choices(['fs-events', 'watchman', 'inotify', 'windows', 'brute-force']),
```
devongovett: wonder if the allowed options should depend on the platform parcel is running on. e.g. inotify is only available on linux, fs-events is only available on Mac, etc.

mattcompiles (commented):
@Nikola-3 We'll need to add the watcher backend to the cache key as they have different snapshot formats.

packages/core/core/test/test-utils.js
Nikola-3: I wanted to make it clear that these are not the only options that exist that can be set for the watcher, e.g. the `watchDir` option above it

mattcompiles (approved):

JakeLane (approved):

mattcompiles (approved):

PR #9548, opened 2024-02-27
Add repl deploy to release workflow

devongovett:
Copied from the ci workflow but using `--prod` instead of a PR alias domain

### DIFF
--- a/.github/workflows/tag-release.yml
+++ b/.github/workflows/tag-release.yml
@@ -216,3 +216,75 @@ jobs:
         env:
           NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
       - run: yarn release
+
+  # Deployment steps taken from https://github.com/colinwilson/static-site-to-vercel/blob/master/.github/workflows/deploy-preview.yml
+  repl_build:
+    name: Build REPL
+    runs-on: ubuntu-latest
+    needs: [unit_tests]
+    permissions:
+      deployments: write
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          cache: yarn
+          node-version: 20
+      - uses: dtolnay/rust-toolchain@master
+        with:
+          toolchain: stable
+          targets: wasm32-unknown-unknown
+      - name: Install wasm-opt
+        run: |
+          curl -L -O https://github.com/WebAssembly/binaryen/releases/download/version_116/binaryen-version_116-x86_64-linux.tar.gz
+          tar -xf binaryen-version_116-x86_64-linux.tar.gz
+          echo "$PWD/binaryen-version_116/bin" >> $GITHUB_PATH
+      - uses: Swatinem/rust-cache@v2
+        with:
+          key: wasm
+      - name: Bump max inotify watches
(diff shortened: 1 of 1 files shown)


.github/workflows/tag-release.yml
```diff
@@ -216,3 +216,75 @@ jobs:
...
+      #     name: REPL
+      #     path: 'packages/dev/repl/dist'
+      - name: Start Deployment
+        uses: bobheadxi/deployments@v1
+        id: deployment
+        with:
+          step: start
+          token: ${{ secrets.GITHUB_TOKEN }}
+          ref: ${{ github.head_ref }}
+          env: Preview
```
mischnic: We should make that something like `Production`, though this isn't really exposed that much in the Github UI: https://github.com/parcel-bundler/parcel/deployments

.github/workflows/tag-release.yml
```diff
@@ -216,3 +216,75 @@ jobs:
...
+          github-comment: false
+          working-directory: packages/dev/repl
+          vercel-args: '--prod'
+      - name: Update Deployment Status
+        uses: bobheadxi/deployments@v1
+        if: always()
+        with:
+          step: finish
+          token: ${{ secrets.GITHUB_TOKEN }}
+          env: Preview
```
mischnic:
```suggestion
          env: Production
          override: false
      - name: Deploy to Vercel
        uses: amondnet/vercel-action@v25
        id: vercel-action
        with:
          vercel-token: ${{ secrets.REPL_VERCEL_TOKEN }}
          vercel-org-id: ${{ secrets.REPL_VERCEL_ORG_ID }}
          vercel-project-id: ${{ secrets.REPL_VERCEL_PROJECT_ID }}
          github-comment: false
          working-directory: packages/dev/repl
          vercel-args: '--prod'
      - name: Update Deployment Status
        uses: bobheadxi/deployments@v1
        if: always()
        with:
          step: finish
          token: ${{ secrets.GITHUB_TOKEN }}
          env: Production
```

mischnic (approved):

.github/workflows/tag-release.yml
```diff
@@ -216,3 +216,75 @@ jobs:
         env:
           NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
       - run: yarn release
+
+  # Deployment steps taken from https://github.com/colinwilson/static-site-to-vercel/blob/master/.github/workflows/deploy-preview.yml
+  repl_build:
+    name: Build REPL
+    runs-on: ubuntu-latest
+    needs: [unit_tests]
```
MonicaOlejniczak: Will this work cross-workflow?

.github/workflows/tag-release.yml
```diff
@@ -216,3 +216,75 @@ jobs:
         env:
           NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
       - run: yarn release
+
+  # Deployment steps taken from https://github.com/colinwilson/static-site-to-vercel/blob/master/.github/workflows/deploy-preview.yml
+  repl_build:
```
MonicaOlejniczak, line 221: How come this isn't in the nightly workflow as well?

devongovett: probably don't want to deploy to production on every nightly, just when we do stable releases.

Issue #9539, opened 2024-02-21
[RFC] Moving to trunk based development and introducing feature flags


marcins:
# 💬 RFC

We have a lot of complex features / pieces of work planned for Parcel over the foreseeable future. In order to be able to meet our goals of: being able to work on features in parallel,  deliver value incrementally, and provide the ability to dogfood new / risky / complex features at scale before making them generally available, we need to shift the approach to how we do feature development and releases in Parcel.

The summary of this proposal is:
* we shift to a "trunk based development" approach - where changes are merged early and often, rather than living in long running feature branches - note that the current ParcelDB work is out of scope for this, as it is too far along to spend the effort to separate it out now, but future work of the scope / scale of PDB would definitely be in scope.
* we implement a feature flag system that allows control of the state of these features, and allows for a combination of configuration and CLI / API overrides for them to be controlled by external systems for internal use cases.
* we shift from a "nightly" build to a "snapshot" build that's driven off merges to the main (`v2`) branch - this means we can "immediately" (once a build runs) consume a release rather than waiting for a nightly to run.

### Trunk based development

The main philosophy shift here is that rather than working on a feature on a branch until it is done / tested / ready for release, features are able to be released piecemeal onto the main branch and are protected by feature flags (see below for feature flags description).

This means that where changes are isolated in scope within a module, they are conditionally executed based on feature flag outcomes. Where changes are more broad and would require peppering conditionals throughout the code then a parallel implementation needs to be created and conditionally included / abstracted at a higher level. How this is achieved will be specific to particular use cases so is not prescribed by this RFC.

(Open question - what about changes that require new dependencies? Are there any issues with this? This doesn't apply to dependency bumps, those would not be possible to feature flag easily)

The advantages of this approach are that integration costs are lower - you're not needing to constantly keep a branch in sync with what's happening on the main branch. Parallel work from a baseline piece of work on the main branch is easier to achieve.

The disadvantage is additional overhead for complex changes of maintaining an abstraction or conditional code.

We believe the advantages for parallelising work and shipping changes sooner for internal usage outweigh the disadvantages.

### Feature flags

We will introduce the concept of feature flags in Parcel. This will allow us to ship inert code in open source, that we are actively testing and stabilising in our internal large scale use cases. Without the ability to do this, in combination with the improvements to the development model above, it may become untenable for us to work in the open source repository as the source of truth.

[...]

PR #9551, opened 2024-02-28
Add feature flags to ParcelOptions

marcins:
# ↪️ Pull Request

This PR is the first part of implementing #9539. It adds a strongly typed `FeatureFlags` to `ParcelOptions`, which are also available in `PluginOptions`. 

Initially, this change also implemented the ability to set feature flags via `.parcelrc`, but this was removed as it was a) deemed initially unnecessary if CLI options are available, b) a philosophical shift in what `.parcelrc` is for, so now it's a lot smaller.

This change has an `exampleFeature` feature flag that doesn't get used in the code, but is mostly there to differentiate from an empty feature flags object, and to allow for some integration testing.

In the integration tests, it does appear that the `RequestGraph` correctly gets invalidated by feature flag changes because of `invalidateOnOptionChange` - however I'm not sure if that's only applying to transformations or other phases as well, I will validate this and follow up if required to ensure feature flags correctly invalidate other parts as well. Admittedly this is an area where I don't have a deep understanding.

A follow-up change will add support for setting feature flag values via CLI options.

## 💻 Examples

With trunk based development, feature flags will be used to merge changes to `v2` that are not intended / ready for most people to use, but we want to test internally. This is as opposed to what might happen today which is a long running branch, and eventually merging when either ready, or behind some `unstable` options. See #9539 for more details.

## 🚨 Test instructions

There is an integration test.

## ✔️ PR Todo

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,7 @@
-on: pull_request
-
 name: Continuous Integration
+
+on:
+  pull_request:
 
 permissions:
   contents: read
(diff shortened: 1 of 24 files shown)


mattcompiles (approved):
Nice one bud

MonicaOlejniczak (approved):

PR #9552, opened 2024-02-29
Update CI caching

MonicaOlejniczak:
# ↪️ Pull Request

These changes update the CI workflow Rust caching so that it can be better reused across jobs, halving the storage used from 1849MB → 920MB. The specific changes include:
* Remove linting cache as it takes the same amount of time as without a cache, and only wastes a bit of space given the 10GB limit in Github
* Remove unused `components: rustfmt` installation from test jobs
* Reuse the cache between integration and unit tests by adding a `shared-key` — this overrides the default per job cache and is more effective since we are using the same set of OS + targets for both jobs
* Update workflows to make them consistent by placing `name` at the top and using `${{ }}` over `${{}}` for expressions
* Use `runner.os` over `matrix.os` as it is always available

## 💻 Examples

Observe the CI workflow behaviour, given the following [example](https://github.com/parcel-bundler/parcel/actions/runs/8088762629/job/22103450027):
* [Linting](https://github.com/parcel-bundler/parcel/actions/runs/8088762629/job/22103450027) still takes 30s without a cache, the same as this [branch](https://github.com/parcel-bundler/parcel/actions/runs/8088323312/job/22102139137) that has caching
* Same cache is found for unit and integration tests across node versions
  * `v0-rust-ubuntu-latest-a554fcb7-c1764bc1`
  * `v0-rust-macos-latest-4be4b021-c1764bc1`
  * `v0-rust-windows-latest-569c1dff-c1764bc1`
* Linux step only runs on ubuntu jobs
* Build native release takes ~1-3min, the same as the branch [here](https://github.com/parcel-bundler/parcel/actions/runs/8088323312/job/22102140018)
* Observe caches [here](https://github.com/parcel-bundler/parcel/actions/caches) which are ~200-250MB for each target
* [Missed cache](https://github.com/parcel-bundler/parcel/actions/runs/8075525378/job/22062579729) takes ~2x as long to build native release compared to [cache hit](https://github.com/parcel-bundler/parcel/actions/runs/8088762629/job/22103452301)

*Note:* REPL fails here because it is tested on push not pull request, and windows has been consistently failing on main branch

## 🚨 Test instructions

Look at the CI step for this pull request, or otherwise test on your own branch by adding a push trigger

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,7 @@
-on: pull_request
-
 name: Continuous Integration
+
+on:
+  pull_request:
 
 permissions:
   contents: read
@@ -18,7 +19,6 @@ jobs:
         with:
           toolchain: stable
           components: rustfmt
-      - uses: Swatinem/rust-cache@v2
       # use `--frozen-lockfile` to fail immediately if the committed yarn.lock needs updates
       # https://yarnpkg.com/lang/en/docs/cli/install/#toc-yarn-install-frozen-lockfile
       - run: yarn --frozen-lockfile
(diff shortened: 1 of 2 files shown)


mattcompiles (approved):
Nice one

.github/workflows/ci.yml
```diff
@@ -79,29 +80,30 @@ jobs:
             packages/core/rust/*.node
 
   integration_tests:
-    name: Integration tests (${{matrix.os}}, Node ${{matrix.node}})
+    name: Integration tests (${{ matrix.os }}, Node ${{ matrix.node }})
```
marcins, line 83: Are these whitespace changes just some auto-formatting somewhere?

.github/workflows/ci.yml
```diff
@@ -79,29 +80,30 @@ jobs:
...
     strategy:
       matrix:
         node: [18, 20]
         os: [ubuntu-latest, macos-latest, windows-latest]
       # These tend to be quite flakey, so one failed instance shouldn't stop
       # others from potentially succeeding
       fail-fast: false
-    runs-on: ${{matrix.os}}
+    runs-on: ${{ matrix.os}}
```
marcins: /nit this whitespace change seems inconsistent - no trailing space.

marcins: .. related, maybe we should have prettier run on these  `.yml` files? I think that's just adding it to https://github.com/parcel-bundler/parcel/blob/5d8d7f034a85814d43eba4c05f60b463c2304768/package.json#L70

marcins (approved):
Love it, thanks!

.github/workflows/ci.yml
MonicaOlejniczak: Not auto-formatting, it was done manually and matches other workflow files. It's the style the Github documentation uses everywhere.

.github/workflows/ci.yml
MonicaOlejniczak: Thanks, I missed this. Prettier doesn't seem to follow the style, and just keeps everything as is

PR #9553, opened 2024-02-29
Add dev release using a reusable workflow

MonicaOlejniczak:
# ↪️ Pull Request

These changes introduce a new dev release workflow that can be manually triggered in Github Actions. It aims to give contributors a way to validate changes not yet on the main branch.

The `dev-release` workflow calls a reusable `release` workflow that performs the majority of the work, and will soon begin to **replace** the `tag-release` and `nightly-release` workflows in the same way. Other additions include:
* Merge `build-apple-silicon` into the `build` job (now renamed to `build-macos-windows`)
* Remove `JEMALLOC_SYS_WITH_LG_PAGE` that existed in `build-apple-silicon` as the underlying issue should be fixed [here](https://github.com/swc-project/swc/pull/6541)
* Add Rust caching to non-dockerized jobs
  * `aarch64-apple-darwin`
  * `macos-latest` and `windows-latest` (this will match the CI cache names, though I did not validate whether they'll get used or if it breaches Githubs cache isolation)
  * `arm-unknown-linux-gnueabihf`
  * `aarch64-unknown-linux-gnu`

## 💻 Examples

Observe the successful workflow build [here](https://github.com/parcel-bundler/parcel/actions/runs/8088762649/job/22103855246)

## 🚨 Test instructions

Test on your own branch by adding a push trigger with your branch name to `dev-release`, as you cannot manually trigger an workflow until it is on the main branch

### DIFF
--- a/.github/workflows/dev-release.yml
+++ b/.github/workflows/dev-release.yml
@@ -1,0 +1,15 @@
+# This workflow builds and releases all packages with a dev dist tag. Once published,
+# the packages can be installed to streamline the testing and validation of changes, both
+# locally and within CI, that have not yet been approved or merged into the main branch.
+name: dev-release
+
+on:
+  workflow_dispatch:
+
+jobs:
+  build-and-release:
+    name: Build and release dev
+    uses: ./.github/workflows/release.yml
+    secrets: inherit
+    with:
+      release-command: yarn dev:release
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,225 @@
+name: release
+
+on:
+  workflow_call:
+    inputs:
+      release-command:
+        description: 'The command that will release packages as part of the final step'
+        required: true
+        type: string
+    secrets:
+      GHCR_TOKEN:
+        required: true
+      NPM_TOKEN:
+        required: true
+
+permissions:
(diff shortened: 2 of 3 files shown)


mattcompiles (approved):
Awesome work 👏 🍻

marcins:
> Add Rust caching to non-dockerized jobs

Which jobs are dockerized, and is it just a limitation of GH Actions that you can't do the caching for those jobs? Does that have any material impact on timings?

marcins (approved):
Nice work!

MonicaOlejniczak:
@marcins I read that caching is not very effective for docker containers in one of the actions documentation, but did not test it out to confirm. I can verify the behaviour as a follow up, but also am wary of blowing up the small cache size in Github.

devongovett:
FYI I think you still need `JEMALLOC_SYS_WITH_LG_PAGE`. If you try to require the built binary on an Apple Silicon Mac you'll get the following error:

```
> require('./parcel-node-bindings.darwin-arm64.node')
<jemalloc>: Unsupported system page size
memory allocation of 216 bytes failed
zsh: abort      node
```

MonicaOlejniczak:
@devongovett Okay, guess I'll find out soon enough! It's easy enough to add back in

PR #9554, opened 2024-02-29
Add CLI support to set Feature Flag values

marcins:
# ↪️ Pull Request

Follows up on #9551 and adds support to the CLI to pass `featureFlags` into `InitialParcelOptions`.

The approach to this is a single `-F` or `--feature-flag` flag that takes a `name=value` argument. Validation is done to ensure the name is a valid feature flag, and `value` matches the expected type - at the moment this is either `string` or `boolean`, though I'd expect `boolean` to be the most common.

## 💻 Examples

`parcel build --feature-flag makeItFaster=true src/index.js`

## 🚨 Test instructions

Manually tested to ensure that validation and setting the arguments in `InitialParcelOptions` works as expected.

## ✔️ PR Todo

### DIFF
--- a/packages/core/parcel/src/cli.js
+++ b/packages/core/parcel/src/cli.js
@@ -104,7 +104,7 @@ const commonOptions = {
     },
     [],
   ],
-  '-F, --feature-flag <name=value>': [
+  '--feature-flag <name=value>': [
     'sets the value of a feature flag',
     (value, previousValue) => {
       let [name, val] = value.split('=');
--- a/packages/core/parcel/src/cli.js
+++ b/packages/core/parcel/src/cli.js
@@ -110,7 +110,6 @@ const commonOptions = {
       let [name, val] = value.split('=');
       if (name in DEFAULT_FEATURE_FLAGS) {
         let featureFlagValue;
-        console.log(typeof DEFAULT_FEATURE_FLAGS[name]);
         if (typeof DEFAULT_FEATURE_FLAGS[name] === 'boolean') {
           if (val !== 'true' && val !== 'false') {
             throw new Error(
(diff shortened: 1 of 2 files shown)


packages/core/parcel/src/cli.js
```diff
@@ -103,6 +104,31 @@ const commonOptions = {
     },
     [],
   ],
+  '-F, --feature-flag <name=value>': [
+    'sets the value of a feature flag',
+    (value, previousValue) => {
+      let [name, val] = value.split('=');
+      if (name in DEFAULT_FEATURE_FLAGS) {
+        let featureFlagValue;
+        console.log(typeof DEFAULT_FEATURE_FLAGS[name]);
```
marcins: Oops, accidentally left this in.. 🤦

packages/core/parcel/src/cli.js
```diff
@@ -103,6 +104,30 @@ const commonOptions = {
     },
     [],
   ],
+  '-F, --feature-flag <name=value>': [
```
mattcompiles: Nitpick: Honestly prefer without the `-F` shorthand. People aren't going to type this by hand anyway so explicit is better.

mattcompiles: Also can you set multiple flags?

MonicaOlejniczak: I think it might need to be `--feature-flag <name=value...>` as per [this](https://github.com/tj/commander.js/blob/master/examples/options-variadic.js)?

But the input would be like `parcel build --feature-flag foo=true bar=false` whereas I normally prefer `parcel build --feature-flag foo=true --feature-flag bar=false` but not sure how easy that would be

marcins: The latter is what works out the box with `commander` (i.e. `parcel build --feature-flag foo=true --feature-flag bar=false`), that was my intention rather than being able to set multiple flags with one option (I would expect most of time time it'll be some sort of launcher script configuring these anyway, and if you're setting a flag by hand it's usually a single one.

MonicaOlejniczak: Perfect 👍

MonicaOlejniczak (approved):

PR #9559, opened 2024-03-01
Replace nightly with canary releases

MonicaOlejniczak:
# ↪️ Pull Request

Replaces the nightly release with a canary release, which will trigger on every push to v2 instead of running every night. Other notable changes include:
* Upgrade lerna from `^3.22.1` to `^6.6.2` to utilise the `--summary-file` option.

Regarding lerna, there are no notable breaking changes between [versions](https://github.com/lerna/lerna/blob/main/CHANGELOG.md):
* Node v10 / v12 is not supported
* Internally lifecycle scripts are called with `@npmcli/run-script` over `npm-lifecycle`
* Some exports have been removed

## 💻 Examples

This change helps us move to trunk based development with a feature flag system as described in https://github.com/parcel-bundler/parcel/issues/9539

A working [dev release](https://github.com/parcel-bundler/parcel/actions/runs/8136996165/job/22234643543) enables installation of `parcel@dev-test` which points to the latest `parcel@dev` version

## 🚨 Test instructions

Change the target push branch to the current branch


mattcompiles (approved):
Amazing work 🚀 
I'm happy but let's make sure @devongovett is aware before merging. Not sure if he has any external reliance on nightly.

marcins:
> Amazing work 🚀 I'm happy but let's make sure @devongovett is aware before merging. Not sure if he has any external reliance on nightly.

Would it be possible to also tag each canary release as a nightly? So `yarn install parcel@nightly` would just give you the latest canary?

marcins (approved):

MonicaOlejniczak:
@marcins I can have a look, is that just to make this less of a breaking change?

devongovett:
so it's essentially the same just more often? I agree about adding a dist-tag for nightly as well for people currently using that.

marcins:
> so it's essentially the same just more often? I agree about adding a dist-tag for nightly as well for people currently using that.

Yeah, it's so the `nightly` tag still works and points to something. I don't know if that's possible to just do as an "alias" easily (i.e. point the nightly dist-tag to the latest canary) since it isn't a single package, or if it's easier to just release as both `nightly` and `canary` from the same build.

.. or I guess another option is to just keep `nightly` as is?

Running another lerna publish to nightly will take a bit longer as you need to run the command again with a different dist-tag - I'm not sure how much of the work Lerna can re-use, but looks like that part currently takes ~4 min - I guess you could always do it in parallel?

From the npm point of view, I don't think there's any issues with doing this.

MonicaOlejniczak:
@devongovett Yeah, it's the same but more frequent

@marcins It can be aliased, but it's a bit tedious to do. I've written a custom script that does this in the latest commit.

mattcompiles (approved):

scripts/tag-release.mjs
```diff
@@ -0,0 +1,38 @@
...
+import {execSync} from 'node:child_process';
+import {readFile} from 'node:fs/promises';
+import {URL} from 'node:url';
+
+import program from 'commander';
+
+async function main(tag) {
+  let publishSummary = JSON.parse(
+    await readFile(
+      new URL('../lerna-publish-summary.json', import.meta.url).pathname,
```
marcins, line 12: I haven't seen this kind of pattern of using an URL in a `readFile`, what does it do over a `readFile(path.resolve(__dirname, "../lerna-publish-summary.json"))`?

marcins (approved):

scripts/tag-release.mjs
MonicaOlejniczak: Neither, only just used it for the first time. It's the equivalent to your example in esm land, and will be an absolute path due to accessing `pathname`

Example: `/parcel/lerna-publish-summary.json`

PR #9562, opened 2024-03-04
Log project root create, init reporters before the request tracker

mattcompiles:
# ↪️ Pull Request

This PR makes two changes:

- Add a new log message for when the project root create event happens. I've also added a `trackableEvent` property to the diagnostic which allows us to easily locate the log event in our reporter. I chose not to open the `Diagnostic` type info can of worms and just flow ignore the error. Happy to hear other suggestions though. 

- Initialise the reporters before the request tracker. This allows any log events that occur during `RequestTracker.init` to hit the reporters.

### DIFF
--- a/packages/core/core/src/Parcel.js
+++ b/packages/core/core/src/Parcel.js
@@ -141,17 +141,21 @@ export default class Parcel {
     this.#watchEvents = new ValueEmitter();
     this.#disposable.add(() => this.#watchEvents.dispose());
 
-    this.#requestTracker = await RequestTracker.init({
-      farm: this.#farm,
-      options: resolvedOptions,
-    });
-
     this.#reporterRunner = new ReporterRunner({
       config: this.#config,
       options: resolvedOptions,
       workerFarm: this.#farm,
     });
     this.#disposable.add(this.#reporterRunner);
+
+    logger.verbose({
+      origin: '@parcel/core',
+      message: 'Intializing request tracker...',
+    });
+    this.#requestTracker = await RequestTracker.init({
+      farm: this.#farm,
+      options: resolvedOptions,
+    });
 
     this.#initialized = true;
   }
--- a/packages/core/core/src/RequestTracker.js
+++ b/packages/core/core/src/RequestTracker.js
@@ -16,6 +16,7 @@ import type {
   InternalFileCreateInvalidation,
   InternalGlob,
 } from './types';
+import logger from '@parcel/logger';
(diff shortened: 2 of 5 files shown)


packages/core/core/src/Parcel.js
```diff
@@ -141,18 +141,22 @@ export default class Parcel {
...
-    });
-
     this.#reporterRunner = new ReporterRunner({
       config: this.#config,
       options: resolvedOptions,
       workerFarm: this.#farm,
     });
     this.#disposable.add(this.#reporterRunner);
 
+    logger.verbose({
```
marcins, line 151: Is this needed? I mean, it's verbose so it doesn't hurt, but just wondering whether it was a bit of debug code leftover.

packages/core/core/src/RequestTracker.js
```diff
@@ -759,6 +760,13 @@ export class RequestGraph extends ContentGraph<
       // this means the project root was moved and we need to
       // re-run all requests.
       if (type === 'create' && filePath === '') {
+        // $FlowFixMe(incompatible-call) `trackableEvent` isn't part of the Diagnostic interface
+        logger.verbose({
+          origin: '@parcel/core',
+          message:
+            'Watcher reported project root create event. Invalidate all nodes.',
+          trackableEvent: 'project_root_create',
```
marcins, line 768: /yakshave maybe `Diagnostic` needs a weakly typed `meta` property that can be used to send aribtrary data for custom reporters.

Though I just wrote that, and realised it doesn't make sense for core to emit something with `meta` that's designed for undefined random plugins to consume..

Another option might be a new log type like `event` for events from core that aren't strictly log messages but downstream consumers might care about. Though then you'd end up with `ReporterEvent` of type `Event`.

¯\_(ツ)_/¯ I'm probably overthinking it, I mean this is fine for now.

marcins (approved):

packages/core/core/src/Parcel.js
mattcompiles: Sorry meant to leave a comment about this. I added this cause the reporters to initialize. That way we get accurate init timings from the reporter. We could remove this once we have a better method for tracking timing data in Parcel (e.g. performance marks, etc).

packages/core/core/src/RequestTracker.js
mattcompiles: Yeah, I'm totes open to something like this. Just wanted to get this out the door though. If we do a lot of this in future we can add something more typed.

PR #9563, opened 2024-03-04
Fix macros integration test for windows

MonicaOlejniczak:
# ↪️ Pull Request

Fixes CI by handling the windows `\` escaping in the failing macros integration test, as there is no trivial way (that I know of) to fix the underlying problem in a generic way when emitting an error from JavaScript `NodePackageManager` → Rust `macro` → JavaScript `JSTransformer`.

## 💻 Examples

None

## 🚨 Test instructions

CI will no longer fail on respective test. Note that there are still test failures due to other issues.

### DIFF
--- a/packages/core/integration-tests/test/macros.js
+++ b/packages/core/integration-tests/test/macros.js
@@ -396,34 +396,41 @@ describe('macros', function () {
         mode: 'production',
       });
     } catch (err) {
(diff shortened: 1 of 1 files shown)


mattcompiles (approved):

marcins (approved):
Thank you for fixing this!

devongovett (approved):

PR #9569, opened 2024-03-08
Update CI workflow

MonicaOlejniczak:
# ↪️ Pull Request

Updates the CI workflow
* **Add v2 branch push trigger to CI:** This lets branches that originate from v2 to consume the same cache as per Github [cache isolation](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#restrictions-for-accessing-a-cache)
* **Update repl build:** It no longer downloads the artifact from unit tests, and simply builds it itself like in the tag-release workflow, to increase parallelism of jobs

## 💻 Examples

None

## 🚨 Test instructions

Add custom push branches to each workflow and observe changes

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,6 +2,9 @@ name: Continuous Integration
 
 on:
   pull_request:
+  push:
+    branches:
+      - v2
 
 permissions:
   contents: read
@@ -112,8 +115,8 @@ jobs:
   # Deployment steps taken from https://github.com/colinwilson/static-site-to-vercel/blob/master/.github/workflows/deploy-preview.yml
   repl_build:
     name: Build REPL
+    if: ${{ github.event_name == 'pull_request' }}
     runs-on: ubuntu-latest
-    needs: [unit_tests]
     permissions:
       deployments: write
     steps:
@@ -137,11 +140,8 @@ jobs:
       - name: Bump max inotify watches
         run: echo fs.inotify.max_user_watches=524288 | sudo tee -a /etc/sysctl.conf && sudo sysctl -p;
       - run: yarn --frozen-lockfile
-      - name: Download @parcel/rust Linux Binaries artifact
-        uses: actions/download-artifact@v3
-        with:
-          name: Rust Linux Binaries
-          path: packages/core/rust
+      - name: Build native packages
+        run: yarn build-native-release
       - run: yarn build
       - run: yarn build-native-wasm
       - run: yarn workspace @parcel/repl build


.github/workflows/ci.yml
```diff
@@ -112,8 +115,8 @@ jobs:
   # Deployment steps taken from https://github.com/colinwilson/static-site-to-vercel/blob/master/.github/workflows/deploy-preview.yml
   repl_build:
     name: Build REPL
+    if: ${{ github.event_name == 'pull_request' }}
     runs-on: ubuntu-latest
-    needs: [unit_tests]
```
marcins, line 116: Was this due a dependency on unit tests, or to avoid building the REPL if the unit tests weren't going to pass? I know you mentioned it re-uses the rust binaries, but was that just an optimisation since this build waited for unit tests to pass anyway?

Does the deploy for the REPL only happen if the other tests pass?

marcins (approved):

.github/workflows/ci.yml
MonicaOlejniczak: I'll need @devongovett to confirm the original intention, but assumed it's not needed since that's how the release flow works now

mischnic: > I know you mentioned it re-uses the rust binaries, but was that just an optimisation since this build waited

Yes, that was the idea when I added the REPL to CI. I think the REPL for PRs should always be built, not only if all tests pass

mischnic:
> Add v2 branch push trigger to CI: This lets branches that originate from v2 to consume the same cache as per Github [cache isolation](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#restrictions-for-accessing-a-cache)

But that is already the case anyway with the canary workflow that runs on every push to v2?

The Rust builds on the PR are already as fast as they can be, it already only rebuilds the crates in the workspace.

MonicaOlejniczak:
@mischnic The canary and CI workflow don't all share the same cache targets, so technically the ubuntu-latest unit and integration tests don't get built as fast as possible. You can see in the first run [here](https://github.com/parcel-bundler/parcel/actions/runs/8196892430/job/22417964998?pr=9569) that everything gets built in `yarn build-native-release`, but other jobs use the [cache](https://github.com/parcel-bundler/parcel/actions/runs/8196892430/job/22417964617?pr=9569). I could modify the release workflow so that the `build-linux-gnu-x64` job is also cached since it emits the same target name, but I'm not sure how equivalent the outputs are internally when one runs in a centos image while the others use whatever Github runs.

The other motivation for this change, is that it is typical for a CI workflow to also run on the main branch. Let me know what you think, and if you'd just prefer to only have the repl change.

As an aside, I did play around with using [actions/cache](https://github.com/actions/cache) over the rust cache action. Using the recommended options there, it avoids building `swc_core` and its related dependencies within the workspace crates. But it does blow out the cache size a lot, and doesn't significantly improve build times.

mischnic (approved):
I see 👍

PR #9573, opened 2024-03-11
Handle dynamic import type references in TS types transformer

### LINKED ISSUE #8908: Generated TypeScript definitions file has broken import("")
Originally reported at https://github.com/bvaughn/react-error-boundary/issues/133

Parcel version 2.8.3

# Repro
The easiest way to repro this is to:
1. Check out this repo: https://github.com/bvaughn/parceljs-typescript-bug
2. Run `yarn install && yarn build`
3. Look at `dist/index.d.ts`

---

### Source

```ts
// index.ts
export * from "./ErrorBoundary";
export * from "./ErrorBoundaryContext";
```

```ts
// ErrorBoundaryContext.ts
import { createContext } from "react";

export type ErrorBoundaryContextType = {};

export const ErrorBoundaryContext = createContext<ErrorBoundaryContextType | null>(null);
```

```ts
// ErrorBoundary.ts
import { Component, createElement, PropsWithChildren } from "react";
import { ErrorBoundaryContext } from "./ErrorBoundaryContext";

export class ErrorBoundary extends Component<PropsWithChildren> {
  render() {
    const { children } = this.props;

    return createElement(
      ErrorBoundaryContext.Provider,
      {
        value: {},
      },
      children
    );
  }
}
```

## Output
```ts
// dist/index.d.ts
import { Component, PropsWithChildren } from "react";
export type ErrorBoundaryContextType = {};
export const ErrorBoundaryContext: import("react").Context<ErrorBoundaryContextType>;
export class ErrorBoundary extends Component<PropsWithChildren> {
  render(): import("react").FunctionComponentElement<
    import("react").ProviderProps<
      import("ErrorBoundaryContext").ErrorBoundaryContextType
    >
  >;
}
```

## Problem

Note the return type of the render method of `ErrorBoundary` – it tries to _import_ `ErrorBoundaryContext` as though it were 
a separate package/module, rather than a variable that's already in scope.

devongovett:
Fixes #8908

Sometimes TSC generates code like `import('types').Foo` to reference a type from another module inline, without hoisting an import statement. The TS types transformer didn't handle these at all, resulting in invalid output. Now we register these imports and replace them with a temporary identifier during the initial collect phase so they work just like regular import statements.


mischnic (approved):

bvaughn:
👏🏼

PR #9574, opened 2024-03-11
Bump swc

### LINKED ISSUE #9607: Importing D3 is causing TheLayout to produce invalid unicode escape sequence on D3.random
# 🐛 bug report

Importing D3 into my react app that is being migrated from CRA produces invalid unicode escape sequence when passing in the params for D3.random.

```import * as d3 from "d3";```

causes TheLayout to produce
```js
normal: function(\xb5, \u03C3) {
              var n = arguments.length;
              if (n < 2) \u03C3 = 1;
              if (n < 1) \xb5 = 0;
              return function() {
                  var x, y, r;
                  do {
                      x = Math.random() * 2 - 1;
                      y = Math.random() * 2 - 1;
                      r = x * x + y * y;
                  }while (!r || r > 1);
                  return \xb5 + \u03C3 * x * Math.sqrt(-2 * Math.log(r) / r);
              };
          },
```

```js
{
  "aliases": {
    "src/*": "./src",
    "alias": {
      "d3": "d3/dist/d3.js"
    }
  },
  "dependencies": {
    "@azure/msal-browser": "^2.21.0",
    "@azure/msal-react": "^1.2.0",
    "@coreui/coreui": "^3.4.0",
    "@coreui/icons": "^2.0.0-rc.0",
    "@coreui/icons-react": "^1.0.2",
    "@coreui/react": "^3.4.5",
    "@coreui/react-chartjs": "^1.0.1",
    "@coreui/utils": "^1.3.1",
    "@emotion/react": "^11.11.1",
    "@emotion/styled": "^11.11.0",
    "@mui/icons-material": "^5.14.1",
    "@mui/lab": "^5.0.0-alpha.168",
    "@mui/material": "^5.15.13",
    "@parcel/config-default": "^2.12.0",
    "@parcel/core": "^2.12.0",
    "@reduxjs/toolkit": "^1.6.1",
    "axios": "1.6.4",
    "bootstrap": "5.1.1",
    "classnames": "^2.2.6",
    "core-js": "^3.8.3",
    "cronstrue": "^2.23.0",
    "d3": "^7.9.0",
    "date-fns": "^2.29.3",
    "detect-file-encoding-and-language": "^2.1.0",
    "email-validator": "^2.0.4",
    "jwt-decode": "^3.1.2",
    "lodash": "^4.17.21",
    "node-sass": "^8.0.0",
    "plotly.js": "^2.6.2",
    "prop-types": "^15.7.2",
    "python-format-js": "^1.4.0",
    "query-string": "^6.13.8",
    "react": "^18.2.0",
    "react-app-polyfill": "^2.0.0",
    "react-autocomplete": "^1.8.1",
    "react-beautiful-dnd": "^13.1.1",
    "react-bootstrap": "^2.5.0",
    "react-circular-progressbar": "^2.1.0",
    "react-code-blocks": "^0.0.8",
    "react-color": "^2.19.3",
    "react-data-grid": "^7.0.0-beta.19",
    "react-data-table-component": "^7.4.4",
    "react-date-picker": "^10.6.0",
    "react-datepicker": "^6.1.0",
    "react-dom": "^18.2",
    "react-dropzone": "^12.0.4",
    "react-grid-layout": "^1.4.4",
    "react-hot-toast": "^2.4.0",
    "react-icons": "^4.1.0",
    "react-intl": "^5.20.10",
    "react-joyride": "^2.3.0",
    "react-loading-overlay-ts": "^2.0.0",
    "react-pdf": "^7.3.3",
    "react-pivottable": "^0.11.0",
    "react-player": "^2.8.2",
    "react-plotly.js": "^2.5.1",
    "react-polling": "^1.0.9",
    "react-query": "^3.21.1",
    "react-redux": "^7.2.4",
    "react-router-dom": "^5.2.0",
    "react-select": "^5.7.3",
    "react-spinners": "^0.9.0",
    "react-step-progress-bar": "^1.0.3",
    "react-textarea-autosize": "^8.4.1",
    "react-toastify": "^9.1.3",
    "react-tooltip": "^5.14.0",
    "react-use-websocket": "3.0.0",
    "reactflow": "^11.9.4",
    "reactjs-crontab": "^4.5.0",
    "redux": "^4.2.0",
    "title-case": "^3.0.3",
    "uplot": "^1.6.18",
    "uplot-react": "^1.1.1"
  },
  "devDependencies": {
    "@babel/plugin-proposal-private-property-in-object": "7.21.11",
    "@parcel/packager-raw-url": "2.12.0",
    "@parcel/transformer-sass": "2.12.0",
    "@parcel/transformer-webmanifest": "2.12.0",
    "auto-changelog": "~2.2.1",
    "buffer": "^5.5.0||^6.0.0",
    "eslint": "^7.31.0",
    "eslint-plugin-prettier": "^4.2.1",
    "eslint-plugin-react": "^7.24.0",
    "husky": "^8.0.0",
    "lint-staged": "^13.0.3",
    "parcel": "^2.12.0",
    "prettier": "^2.7.1",
    "process": "^0.11.10",
    "react-scripts": "^5.0.0"
  },
  "scripts": {
    "startdev": "REACT_APP_STAGE=dev react-scripts start",
    "start": "REACT_APP_STAGE=local parcel src/index.html --port 3000 ",
    "build": "REACT_APP_STAGE=prod DISABLE_ESLINT_PLUGIN=true react-scripts build",
    "test": "react-scripts test",
    "test:cov": "npm test -- --coverage --watchAll=false",
    "test:debug": "react-scripts --inspect-brk test --runInBand",
    "eject": "react-scripts eject",
    "changelog": "auto-changelog --starting-version 3.0.0 --commit-limit false --hide-credit",
    "lint": "eslint --ext .js --ext .jsx --format compact",
    "prepare": "husky install",
    "pre-commit": "lint-staged"
  },
  "lint-staged": {
    "src/**/*.{js,jsx}": [
      "eslint --ext .jsx --ext .js ./src --fix"
    ],
    "./src/**": [
      "prettier --write ."
    ]
  },
  "husky": {
    "hooks": {
      "pre-commit": "lint-staged"
    }
  },
  "eslintConfig": {
    "extends": "react-app"
  },
  "browserslist": {
    "production": [
      ">0.2%",
      "not dead",
      "not ie <= 10",
      "not op_mini all"
    ],
    "development": [
      "ie 11",
      "last 1 chrome version",
      "last 1 firefox version",
      "last 1 safari version"
    ]
  },
  "resolutions": {
    "react-code-blocks/**/prismjs": "^1.21.0",
    "react-scripts/**/nth-check": "^2.0.1",
    "react-code-blocks/**/react-syntax-highlighter": "^15.4.4",
    "react-code-blocks/**/lowlight": "^1.0.0",
    "react-code-blocks/**/highlight.js": "^10.4.1",
    "react-scripts/**/postcss": "8.4.31",
    "react-scripts/**/follow-redirects": "^1.15.6",
    "axios/follow-redirects": "^1.15.6",
    "autoprefixer": "10.4.5",
    "ip": "2.0.1"
  },
  "jest": {
    "collectCoverageFrom": [
      "src/**/*.{js,jsx}",
      "!**/*index.js",
      "!src/serviceWorker.js",
      "!src/polyfill.js"
    ]
  },
  "engines": {
    "node": ">=10",
    "npm": ">=6"
  }
}

```

## 🤔 Expected Behavior

I would expect it to produce something along the lines of the other D3.random functions
```js
d3.random = {
                        logNormal: function() {
                            var random = d3.random.normal.apply(d3, arguments);
                            return function() {
                                return Math.exp(random());
                            };
                        },
                        bates: function(m) {
                            var random = d3.random.irwinHall(m);
                            return function() {
                                return random() / m;
                            };
                        },
                        irwinHall: function(m) {
                            return function() {
                                for(var s = 0, j = 0; j < m; j++)s += Math.random();
                                return s;
                            };
                        }
                    };
```

## 😯 Current Behavior

```js
d3.random = {
                        normal: function(\xb5, \u03C3) {
                            var n = arguments.length;
                            if (n < 2) \u03C3 = 1;
                            if (n < 1) \xb5 = 0;
                            return function() {
                                var x, y, r;
                                do {
                                    x = Math.random() * 2 - 1;
                                    y = Math.random() * 2 - 1;
                                    r = x * x + y * y;
                                }while (!r || r > 1);
                                return \xb5 + \u03C3 * x * Math.sqrt(-2 * Math.log(r) / r);
                            };
                        },
```

Screenshot 2024-04-02 113940

## 🔦 Context

I am simply trying to convert from CRA to ParcelV2 on an established project

## 💻 Code Sample

Seems either
```yarn install d3```
or 
```import * as d3 from "d3";```

## 🌍 Your Environment

| Parcel           | 2.12.0
| Node             | 18.19.1
| npm/Yarn         | 1.22.22
| Operating System | Windows 11 Pro 64-bit (10.0, Build 22631) (22621.ni_release.220506-1250)

### LINKED ISSUE #9632: Non-ASCII characters in identifiers result in invalid escape sequences
# :scroll: related issues I've found
- https://github.com/parcel-bundler/parcel/issues/9607
- https://github.com/parcel-bundler/parcel/issues/9370 (closed, but still happening?)

# 🐛 bug report

Non-ASCII characters in identifiers result in invalid escape sequences when the code is bundled/optimized. This is a problem not only when using greek letters (e.g. in a _mathy_ domain: `function p(µ, σ) {}`), but also when programming in languages other than english (e.g. when using accents in spanish, like in `módulo`).

This started happening in version [2.10.0](https://github.com/parcel-bundler/parcel/releases/tag/v2.10.0) onwards, with [2.9.3](https://github.com/parcel-bundler/parcel/releases/tag/v2.9.3) it doesn't do that.

The babel configuration is the default one (no `babel.config.json` file or anything like that).

The `package.json` file contains:
```json
{
  "scripts": {
    "start": "parcel index.html"
  },
  "devDependencies": {
    "parcel": "2.10.0"
  }
}
```

## 🤔 Expected Behavior

With this HTML file as input (`index.html`):

```html
<!doctype html>
<html lang="es">
<head>
    <meta charset="UTF-8">
    <title>Mi documento</title>
    <script type="module">
        const módulo = 123;
        console.log(módulo);
    </script>
</head>
<body></body>
</html>
```

My expectations are:
- When running `npm start` and browsing the localhost page, I should see `123` printed in the console.
- When inspecting the generated JS code, a correct escape sequence is inserted in place of the `ó` in `módulo`. For example:
  ```javascript
  const m\u{0f3}dulo = 123;
  console.log(m\u{0f3}dulo);
  ```
  Alternatively, I should see the code without modifications:
  ```javascript
  const módulo = 123;
  console.log(módulo);
  ```

## 😯 Current Behavior

- I see an error in the console: `Uncaught SyntaxError: invalid escape sequence` in Firefox, `Uncaught SyntaxError: Invalid or unexpected token` in Chrome.
- The generated JS code is invalid, as the escape sequence is invalid for an identifier:
  ```javascript
  const m\xf3dulo = 123;
  console.log(m\xf3dulo);
  ```

## 💁 Possible Solution

-

## 🔦 Context

I was trying to use parcel in a project which uses spanish as the language for the code (on variable names, etc.).

## 💻 Code Sample

See "Expected behavior"

## 🌍 My Environment

| Parcel           | 2.10.0
| Node             | 20.11.1
| npm/Yarn         | 10.5.0
| Operating System | Linux

mischnic:
Closes https://github.com/parcel-bundler/parcel/issues/9632
Closes https://github.com/parcel-bundler/parcel/issues/9607

There's a major breaking change, the LHS of assignments is now a special enum.

~~There's one place which I haven't been able to fix yet. Putting this up in case someone else wants to bump swc sooner rather than later for some reason~~


mischnic:
Even in release builds, there is a segfault now in the EnvReplacer, the stack trace does many `fold_expr -> fold_bin_expr -> fold_expr -> ...`. Happens on this file, which has a string concatenation expression containing 750 `+`: https://github.com/highlightjs/highlight.js/blob/105a11a13eedbf830c0e80cc052028ceb593837f/src/languages/isbl.js

Usually, this is only a problem (and has happened for a long time) in non-release builds.

But only on Linux and Windows, macOS works fine.

mischnic:
The same asset also works fine in a standalone Rust projects (with the exact code as the JSTransformer, just without the Node process).
So I guess the stack occupied by Node plus the recursion in swc triggers the crash, and with just Rust the stack space is enough. And macOS probably also has more stack

marcins:
> The same asset also works fine in a standalone Rust projects (with the exact code as the JSTransformer, just without the Node process). So I guess the stack occupied by Node plus the recursion in swc triggers the crash, and with just Rust the stack space is enough. And macOS probably also has more stack

Not really an area I have any experience in, but a quick search shows that is possible to increase the stack size with some linker flags: https://www.reddit.com/r/rust/comments/872fc4/how_to_increase_the_stack_size/

Is that worth experimenting with?

EDIT: read something else that mentioned threads on linux specifically have a smaller stack, maybe we need to increase the Rayon thread stack size (that's where this code is running, right?): https://docs.rs/rayon/latest/rayon/struct.ThreadPoolBuilder.html#method.stack_size

mischnic:
I want to try bisecting which swc version introduced this, because not much changed regarding the AST depth

So increasing the stack size would  maybe just be a workaround until some future version.

mischnic:
As it turns out, this also occurs in Deno and Turbopack: https://github.com/swc-project/swc/issues/8840

jdlm-stripe:
@mischnic let me know if I can help at all on this. We're blocked on some of the bugs now fixed in 1.4.15.

mischnic:
@jdlm-stripe Yeah that would be great! I've resolved the merge conflicts

The main problem is that somehow the stack is still overflowing on Windows, so I suspect that the `maybe_grow_default` added in this PR don't increase the stack enough.

That is declared here (effectively equivalent to `stacker::maybe_grow`): https://github.com/swc-project/swc/blob/main/crates/swc_ecma_utils/src/stack_size.rs
You can also pass it a different threshold and new stack size.

`stacker::remaining_stack()` can be used to query the remaining stack size.

For a reproduction, see https://github.com/parcel-bundler/parcel/pull/9574#issuecomment-2001981011 (just running `parcel build isbl.js` was enough for me to cause a native crash)

I've been putting this off since I always have to work in a Windows VM for this, maybe you happen to have a Windows machine.
Barring any new findings, I think we'll just have to increase the stack size (use larger values than the constants in `maybe_grow_default`). (This shouldn't be the end of the world, since it will only ever happen few times per thread and once the file with the deepest AST was traversed, the stack size will never be increased again.)

Let me know if you have any questions

jondlm:
I'm making some progress on this one. I happen to have a windows machine so I was able to spin up a working dev environment on that. So far I can reproduce the stack overflow with exit code `-1073741571`. I'm still cranking away on it but just wanted to give a quick update.

Side note: I have another PR I'll put up with some other build related fixes for windows. Just some npm scripts issues that were easy to fix.

jondlm:
@mischnic your hunch was right. I put up a fix that works for me here: https://github.com/parcel-bundler/parcel/pull/9693/files

I don't have write permissions to the repo otherwise I would have pushed a commit onto this PR.

mischnic:
I've rebased onto v2 and bumped swc again (but not much changed ac92e7a8ec4e1dcd343da4f9df847626cd76d812) and now Windows CI appears to work (even without your PR), but maybe it's also not deterministic

mischnic:
And indeed, now it's failing again

jondlm:
Ugh yeah. It's definitely non deterministic. The `maybe_grow` I found worked on my machine apparently doesn't cut it for CI. I suspect the data in the stack is larger (maybe longer paths or something).

I think the CI failure is happening on something other than the `EnvReplacer` because it's happening on `yarn build`. So I'm trying to figure out why we're not getting back traces. Any ideas on why the rayon threads (even with `RUST_BACKTRACE=1`) don't report anything? I think [this is the relevant code for that](https://github.com/parcel-bundler/parcel/blob/1b53f9192c8632bf6b9e629a635aac976cad22c0/crates/node-bindings/src/transformer.rs#L36-L42). Perhaps it has something to do with the napi env and the way we're threading.

mischnic:
Does `yarn build` work for you locally?  
`PARCEL_WORKERS=0` might be worth a try

> I think the CI failure is happening on something other than the EnvReplacer because it's happening on yarn build

But EnvReplacer also runs in that case? I don't quite understand what you mean

> why we're not getting back traces

I think Rust simply doesn't emit "nice" panics for stack traces. You get get a segfault/invalid instruction.

marcins:
You'd need to compile with debugging information to get nice stack traces (debug=true in the Cargo profile). 

Though that might affect the outcome, we had an internal rayon panic we saw quite a bit internally go away when we started building the canary release with debuginfo (even though it gets extracted to a separate file and stripped from the binary).

jondlm:
Weirdly I couldn't reproduce even with `PARCEL_WORKERS=0` on my machine. After a bunch of fiddling it looks like [CI passes](https://github.com/parcel-bundler/parcel/actions/runs/8993511707/job/24705435189?pr=9693) with a ["red zone" of 32KiB](https://github.com/parcel-bundler/parcel/pull/9693/commits/e500f3c440a554579075a55a50196d73003f9c31#diff-85cd083e0ee2ec3af34aea26211ce4b9da3398831ab9c0fab1768941499a852dR171) in the main fold in the EnvReplacer. Hopefully that number holds 🙏 

And yeah it looks like you're right about stack overflows not reporting a back trace. I should have known better than to trust gpt lol.

jondlm (approved):
Not sure if my review is enough here but this LGTM.

mischnic:
No, Github doesn't see it as a approving review...

jondlm:
@devongovett when you have a moment could you review this PR? Normally I wouldn't bother with a ping but it's currently a blocker for us at Stripe. I think it's good to go it just needs an approving review.

packages/transformers/js/core/src/modules.rs
```diff
@@ -594,7 +595,7 @@ impl Fold for ESMFold {
     if self.in_export_decl {
       self.create_export(
         node.key.sym.clone(),
-        Expr::Ident(node.key.clone()),
+        Expr::Ident(node.key.id.clone()),
```
marcins: Just sharing what I shared in Discord here, it seems that the change arround `AssignPatProp` means that after visiting here it will visit into `fold_binding_intent`, this results in a duplicate export being added for exports of the form `export const { Foo } = bar;` in non-scope hoisting mode - this breaks at runtime.

This change fixed the issue for me, but not sure if there's some edge case that it might not be correct for.  It would be good to start with a failing integration test.

```diff
diff --git a/packages/transformers/js/core/src/modules.rs b/packages/transformers/js/core/src/modules.rs
index 65235f3c3..bd13d10b6 100644
--- a/packages/transformers/js/core/src/modules.rs
+++ b/packages/transformers/js/core/src/modules.rs
@@ -598,9 +598,11 @@ impl Fold for ESMFold {
         Expr::Ident(node.key.id.clone()),
         DUMMY_SP,
       );
+      node
+    } else {
+      node.fold_children_with(self)
     }
 
-    node.fold_children_with(self)
   }
 
   modules_visit_fn!(fold_function, Function);
```

marcins (changes requested):
Found a new bug that affects non-scope hoisted builds, needs to be fixed before merging.

mischnic:
@marcins @jondlm I fixed that problem, there's a new dev release published at `parcel@2.0.0-dev.1606` if you want to test it again.

https://github.com/parcel-bundler/parcel/actions/runs/9077140040/job/24942042946

marcins (approved):
I've tested our dev server and production builds of Jira and they appear to be working correctly now! 🎉

jondlm:
We also tested it and it's working for us. Thank you so much @mischnic & @marcins!

PR #9580, opened 2024-03-13
Write tmp files to the OS tmp directory

mattcompiles:
# ↪️ Pull Request

This PR attempts to utilize the OS tmp directory for temporary files when streaming content to files. This greatly reduces the amount of events the watcher needs to process during the build. We only use the tmp directory when the following are true:
- The process has read/write access to the tmp directory
- The tmp directory is on the same partition as the target directory

## 🚨 Test instructions

We've tested this approach extensively internally. 

## ✔️ PR Todo

### DIFF
--- a/packages/core/fs/src/NodeFS.js
+++ b/packages/core/fs/src/NodeFS.js
@@ -238,7 +238,7 @@ function shouldUseOsTmpDir(filePath) {
     const tmpDirStats = nativeFS.statSync(tmpDir);
     const filePathStats = nativeFS.statSync(filePath);
     // Check the tmpdir is on the same partition as the target directory.
-    // This is required to allow renaming to occur.
+    // This is required to ensure renaming is an atomic operation.
     useOsTmpDir = tmpDirStats.dev === filePathStats.dev;
   } catch (e) {
     // We don't have read/write access to the OS tmp directory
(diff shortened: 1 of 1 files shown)


JakeLane (approved):

packages/core/fs/src/NodeFS.js
```diff
@@ -222,11 +224,43 @@ try {
   //
 }
 
+let useOsTmpDir;
+function shouldUseOsTmpDir(filePath) {
+  if (useOsTmpDir != null) {
+    return useOsTmpDir;
```
marcins, line 230: I guess there's a tiny risk the `/tmp` directory becomes un-usable between the first check and the first write, but that seems pretty unlikely?

packages/core/fs/src/NodeFS.js
```diff
@@ -222,11 +224,43 @@ try {
...
+  try {
+    const tmpDir = tmpdir();
+    nativeFS.accessSync(
+      tmpDir,
+      nativeFS.constants.R_OK | nativeFS.constants.W_OK,
+    );
+    const tmpDirStats = nativeFS.statSync(tmpDir);
+    const filePathStats = nativeFS.statSync(filePath);
+    // Check the tmpdir is on the same partition as the target directory.
+    // This is required to allow renaming to occur.
```
marcins: Technically it's required to ensure renaming is an atomic operation - you can still rename between filesystems, it'll just be a copy/delete which isn't atomic.

marcins (approved):

packages/core/fs/src/NodeFS.js
mattcompiles: I can't even imagine how that would happen. I'm comfortable with a single check.

PR #9584, opened 2024-03-14
Use a non-circular type to fix TS generation

mattcompiles:
# ↪️ Pull Request

The `build-ts` script is currently failing due to a circular referencing type. This means releases are blocked until this is fixed. I decided to just use `any` as the type for now as a workaround. Shout out if there's something better I could do. 

## ✔️ PR Todo

### DIFF
--- a/packages/core/diagnostic/src/diagnostic.js
+++ b/packages/core/diagnostic/src/diagnostic.js
@@ -48,6 +48,7 @@ export type DiagnosticCodeFrame = {|
 
 /** A JSON object (as in "map") */
 type JSONObject = {
+  // $FlowFixMe
   [key: string]: any,
 };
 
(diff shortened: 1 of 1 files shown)


JakeLane (approved):

JakeLane:
Surprised this is required? https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#more-recursive-type-aliases

mattcompiles:
@JakeLane Yeah same. There's a whole `flow-to-ts` thing going on as well so it's confusing to dig into.

marcins:
> The build-ts script is currently failing due to a circular referencing type.

Should we run `build-ts` as part of branch builds to ensure it still works?

PR #9590, opened 2024-03-20
Add support for globs in package.json#source

### LINKED ISSUE #3821: [RFC][V2] Support glob at package.json#source
# 💬 RFC

Currently, `EntryResolver` expects `package.json#source` to be the path of a file.

I'm building an [SSR plugin for Parcel v2](https://github.com/brillout/parcel-ssr) and I'd like my users to be able to define multiple entry points. For example:

~~~
pages/
pages/package.json
pages/Landing.page.js
pages/about/About.page.js
pages/product/Product.page.js
...
~~~

where `pages/package.json#source` would be `**/*.page.js`. But globs are not supported for `package.json#source`.

It seems that currently the only way to have multiple entries is to have a `package.json` for each entry. But it would be silly to ask my users to create a new `package.json` each time they create a new page.

This is somewhat related to the discussion at #3302.

devongovett:
Closes #3821

This implements support for globs in the package.json "source" field, both top-level and inside targets. Multiple globs are also supported when listed in an array.

### DIFF
--- a/packages/core/core/src/requests/EntryRequest.js
+++ b/packages/core/core/src/requests/EntryRequest.js
@@ -204,36 +204,47 @@ export class EntryResolver {
                 ? target.source
                 : [target.source];
               let i = 0;
-              for (let relativeSource of targetSources) {
-                let source = path.join(entry, relativeSource);
+              for (let source of targetSources) {
+                let sources;
+                if (isGlob(source)) {
+                  sources = await glob(source, this.options.inputFS, {
+                    onlyFiles: true,
+                    cwd: entry,
+                  });
+                } else {
+                  sources = [source];
+                }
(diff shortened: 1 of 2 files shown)


mischnic:
Should this introduce a glob invalidation? Similar to 
https://github.com/parcel-bundler/parcel/blob/90ca165c23b471811302a5d75e8678359cf59aca/packages/core/core/src/requests/EntryRequest.js#L69-L73

devongovett:
gonna merge this to unblock some stuff, lmk if you see anything else I should change!

buvinghausen:
Thanks @devongovett let me know once you get this out to npm so I can remove my janky hacks from our stack!

devongovett:
feel free to try the latest canary tag now. otherwise it'll be out in the next regular release.

PR #9597, opened 2024-03-25
Add config key invalidation

mattcompiles:
# ↪️ Pull Request

Currently, when using config stored inside of package.json keys (utilised by the default JS transformer, bundler and JS packager) you end up invalidation your results when anything in the root package.json changes. This causes massive over invalidation, especially in large monorepos that store a huge list of deps and unrelated config in the root. 

This PR attempts to address this by adding a new type of node to the RequestGraph called a `ConfigKeyNode`. This node acts similar to the existing `FileNode`, however it also stores the package key and a content hash of it's value. When a plugin uses the `config.loadConfigFrom` API and it resolves to something using the `packageKey` option this new type of invalidation will be used. When the config request executes, it will calculate a content hash for the package value and store it in the request graph. Then when a file change is detected for the relevant package.json, the file will be read and the content hash validated against the stored one before causing any invalidation to occur. 

Things to note:
- This feature is currently behind a feature flag so we can test it at scale before defaulting the behaviour
- This does not affect the invalidate on file create systems, they should continue to work as is
- The content hash is build cached, so we should only calculate them once per key, per build
- The JS transformer and JS packager have been updated to take advantage of the new API

## ✔️ PR Todo


packages/core/core/src/InternalConfig.js
```diff
@@ -20,6 +20,10 @@ type ConfigOpts = {|
   env?: Environment,
   result?: ConfigResult,
   invalidateOnFileChange?: Set<ProjectPath>,
+  invalidateOnPackageKeyChange?: Array<{|
```
marcins: Would it make sense to make the invalidation part less “package key” specific and more generic “config key”? I’m thinking for plugin authors it could apply to other config files as well? The current implementation of `pacakgeKey` in getConfigFrom assumes `pacakge.json`, but the invalidation at least at the moment supports generic JSON, and could support other formats.

packages/core/core/src/RequestTracker.js
```diff
@@ -886,6 +956,40 @@ export class RequestGraph extends ContentGraph<
         // to requests as invalidations for future requests.
         this.removeNode(nodeId);
       }
+
+      let packageKeyNodes = this.packageKeyNodes.get(_filePath);
+      if (packageKeyNodes && (type === 'delete' || type === 'update')) {
+          for (let nodeId of packageKeyNodes) {
+          let isInvalid = type === 'delete';
```
marcins, line 962: /nit this indentation is off - should prettier be run against this change?

packages/core/core/src/requests/ConfigRequest.js
```diff
@@ -100,12 +106,48 @@ export async function loadPluginConfig<T: PluginWithLoadConfig>(
...
+  if (cachedValue) {
+    return cachedValue;
+  }
+
+  let conf = await readConfig(
+    options.inputFS,
+    fromProjectPath(options.projectRoot, filePath),
+  );
+
+  if (conf == null || conf.config[packageKey] == null) {
```
marcins: The `packageKey` being removed could be a valid scenario couldn't it? Like if I remove custom config for a transformer that I no longer need because the defaults are file.

packages/core/feature-flags/src/types.js
```diff
@@ -3,4 +3,5 @@
 export type FeatureFlags = {|
   // This feature flag mostly exists to test the feature flag system, and doesn't have any build/runtime effect
   +exampleFeature: boolean,
+  +packageKeyInvalidation: boolean,
```
marcins: Originally I had a "description" for feature flags as part of a richer FF object export, but decided it was unnecessary - however, it would be good to document what the feature flag is for in a comment here at least.

packages/core/integration-tests/test/cache.js
```diff
@@ -1312,6 +1312,128 @@ describe('cache', function () {
...
+              }),
+            );
+          },
+        });
+
+        assert.equal(await run(b.bundleGraph), 'default');
+        assert.equal(b.changedAssets.size, 0);
+      });
+
+      it('should invalidate when package.json config keys change', async function () {
```
marcins, line 1377: Related to comment earlier, should invalidate when config key is removed as well.

packages/packagers/js/src/index.js
```diff
@@ -32,34 +31,40 @@ const CONFIG_SCHEMA: SchemaEntity = {
...
-    if (pkg?.contents[packageKey]) {
+    if (conf?.contents) {
       validateSchema.diagnostic(
         CONFIG_SCHEMA,
         {
-          data: pkg?.contents[packageKey],
-          source: await options.inputFS.readFile(pkg.filePath, 'utf8'),
-          filePath: pkg.filePath,
+          data: conf?.contents,
+          source: await options.inputFS.readFile(conf.filePath, 'utf8'),
```
marcins, line 44: I think we talked about this during pairing, and it's definitely different for the transformer that does this, but this will only read the `package.json` here once when the packager is initialised, not per package?

marcins (changes requested):
Couple of minor things, but generally looks good!

packages/core/core/src/InternalConfig.js
mattcompiles: Agreed. Updated the naming to be config key, rather than package key 👍

packages/core/core/src/RequestTracker.js
mattcompiles: Is it? Prettier is passing for me and I have it setup. Not sure what you're referring to?

packages/packagers/js/src/index.js
mattcompiles: Once per build per worker I believe. However we can definitely improve this overreading in a separate PR.

packages/core/core/src/requests/ConfigRequest.js
mattcompiles: Ahh yeah you're right, good catch.

packages/core/feature-flags/src/types.js
mattcompiles: Done 👍

packages/core/integration-tests/test/cache.js
mattcompiles: Added a test

packages/core/core/src/RequestTracker.js
marcins: Weird, the `let isInvalid` is showing the same indentation as the `for .. {` above, I'd expect it to be indented another level. Anyway, not going to argue with Prettier!

EDIT: huh, indentation is fine when I view the whole file.. must just be something about how GH renders the diff.

marcins (approved):

PR #9598, opened 2024-03-25
Add sentry support

Nikola-3:
# ↪️ Pull Request

Canary builds now release and parcel rust code can be conditionally compiled with support for sentry in rust. This is useful for providing error logging in rust at runtime.

## ✔️ PR Todo


.github/workflows/canary-release.yml
```diff
@@ -15,6 +15,7 @@ jobs:
     uses: ./.github/workflows/release.yml
     secrets: inherit
     with:
+      profile: canary
```
MonicaOlejniczak, line 18: Should this be enabled for dev releases as well? Regardless, it feels like the option should be more generic (i.e. a profiling build) and use the same release command, but pass in a env variable / cli option that you read

scripts/build-native.js
```diff
@@ -26,7 +28,8 @@ async function build() {
     console.log(`Building ${pkg}...`);
     await new Promise((resolve, reject) => {
       let args = [
-        (wasm ? 'wasm:' : '') + (release ? 'build-release' : 'build'),
+        (wasm ? 'wasm:' : '') +
```
JakeLane: can we split this out into an if statement? Nested ternaries are hard to read

packages/core/core/src/Parcel.js
```diff
@@ -100,6 +100,13 @@ export default class Parcel {
 
     await initSourcemaps;
     await initRust?.();
+    try {
+      await initSentry?.();
+    } catch (e) {
+      // Fallthrough
+      // eslint-disable-next-line no-console
+      console.warn(e);
```
JakeLane: It's fine to continue startup if sentry fails, but we should log a bit more effectively here. `logger` is available and we should provide some context to the error

crates/node-bindings/src/init_sentry.rs
```diff
@@ -0,0 +1,87 @@
...
+  }
+
+  let Ok(sentry_dsn) = std::env::var("PARCEL_SENTRY_DSN") else {
+    return Err(Error::from_reason(
+      "Sentry enable but PARCEL_SENTRY_DSN environment variable not provided.",
+    ));
+  };
+
+  let sentry_tags = if let Ok(sentry_tags_raw) = std::env::var("PARCEL_SENTRY_TAGS") {
+    println!("{}", sentry_tags_raw);
```
JakeLane: do we want to log these?

crates/node-bindings/src/init_sentry.rs
```diff
@@ -0,0 +1,87 @@
...
+  Lazy::new(|| Arc::new(Mutex::new(None)));
+const TIMEOUT: Duration = Duration::from_secs(2);
+
+#[napi]
+fn init_sentry() -> Result<(), Status> {
+  if std::env::var("PARCEL_ENABLE_SENTRY").is_err() {
+    return Ok(());
+  }
+
+  println!("Initialising Sentry in rust...");
```
yamadapc: ideally this would use `log` - https://crates.io/crates/log

crates/node-bindings/Cargo.toml
```diff
@@ -7,12 +7,20 @@ edition = "2021"
...
+[features]
+canary = []
+
 [dependencies]
 napi-derive = "2.12.5"
 parcel-js-swc-core = { path = "../../packages/transformers/js/core" }
 parcel-resolver = { path = "../../packages/utils/node-resolver-rs" }
 dashmap = "5.4.0"
 xxhash-rust = { version = "0.8.2", features = ["xxh3"] }
+sentry = "0.32.2"
```
yamadapc: nit:

if you only use these dependencies for sentry you can make them optional then add them into the `canary = []` array to only compile them when this feature is enabled.

you'd first make `once_cell = { version = "1.19.0", optional = true }` then add to the array; see "optional dependencies" https://doc.rust-lang.org/cargo/reference/features.html#optional-dependencies

scripts/build-native.js
alshdavid: Haha this is on me. Nik said the same thing - I agreed but was feeling lazy.

yamadapc (approved):
lgtm ; despite comments

MonicaOlejniczak (approved):

packages/core/core/src/Parcel.js
```diff
@@ -101,6 +101,13 @@ export default class Parcel {
 
     await initSourcemaps;
     await initRust?.();
+    try {
+      await initSentry?.();
+    } catch (e) {
+      // Fallthrough
+      // eslint-disable-next-line no-console
```
mattcompiles: No need for the eslint disable
```suggestion
```

JakeLane (approved):

PR #9609, opened 2024-04-02
Update integration test script

MonicaOlejniczak:
# ↪️ Pull Request

This change updates the integration test script so that `--experimental-vm-modules` is enabled as a CLI option rather than using the `NODE_OPTIONS` environment variable. Doing this enables debuggers, like VSCode, to attach breakpoints using default launch settings.

## 💻 Examples

N/A

## 🚨 Test instructions

* Import `vm` in any plugin or core code, like a transformer
* Ensure extra exports exist, e.g. `SourceTextModule` when running integration tests locally and for CI

### DIFF
--- a/packages/core/integration-tests/package.json
+++ b/packages/core/integration-tests/package.json
@@ -8,7 +8,7 @@
     "url": "https://github.com/parcel-bundler/parcel.git"
   },
   "scripts": {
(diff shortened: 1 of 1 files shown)


alshdavid (approved):

mattcompiles (approved):
Nice find 👏

PR #9610, opened 2024-04-03
Fix circular dependencies on @parcel/types

yamadapc:
This change is removing circular dependencies between certain packages and `@parcel/types`.

This is achieved by moving some types into that package, which is renamed to `@parcel/types-internal`; additionally another `@parcel/types` package is added to include `WorkerFarm` into the parcel options types.

This is not an ideal set-up. The ideal set-up is that there is no centralized types package. Instead the types are declared in `fs-types`, `core-types` etc. packages that correspond to the current logic packages.

## ✔️ PR Todo

- [x] Test documentation generation


packages/dev/bundle-stats-cli/package.json
```diff
@@ -16,7 +16,7 @@
   "main": "lib/cli.js",
   "source": "src/cli.js",
   "bin": {
-    "parcel-bundle-stats": "src/bin.js"
+    "parcel-bundle-stats": "lib/bin.js"
```
yamadapc: revert these

packages/dev/query/package.json
```diff
@@ -14,7 +14,7 @@
     "url": "https://github.com/parcel-bundler/parcel.git"
   },
   "bin": {
-    "parcel-query": "src/bin.js"
+    "parcel-query": "lib/bin.js"
```
yamadapc: revert these

yamadapc:
As of the latest commit, a new `@parcel/types-internal` package is added.

The idea here is to have the following dependency graph:

```
@parcel/core --> @parcel/types --> @parcel/types-internal
                                \-- @parcel/workers ---/
```

And avoid the cycle between `@parcel/workers` and `@parcel/types`, while still keeping the same API.

Ideally the types would not be centralized in a package, and rather there'd be multiple types packages

packages/core/fs/package.json
```diff
@@ -49,7 +49,6 @@
   },
   "dependencies": {
     "@parcel/rust": "2.12.0",
-    "@parcel/types": "2.12.0",
```
mattcompiles, line 52: Not sure this should have been removed as it's imported in `index.d.ts`

yamadapc: You're right; I shouldn't have removed this.

packages/core/profiler/package.json
```diff
@@ -20,12 +20,13 @@
     "node": ">= 12.0.0"
   },
   "scripts": {
-    "build-ts": "flow-to-ts src/*.js --write && tsc --emitDeclarationOnly --declaration --esModuleInterop --target es2015 --moduleResolution node16 --module node16 src/*.ts && mkdir -p lib && mv src/*.d.ts lib/. && rm src/*.ts && node build-ts.js",
+    "build-ts": "flow-to-ts src/*.js --write && rm -f ./src/*.d.ts && tsc --emitDeclarationOnly --declaration --esModuleInterop --target es2015 --moduleResolution node16 --module node16 src/*.ts && mkdir -p lib && mv src/*.d.ts lib/. && rm src/*.ts && node build-ts.js",
```
mattcompiles, line 23: What's context on this change?

marcins:
> This is not an ideal set-up. The ideal set-up is that there is no centralized types package. Instead the types are declared in fs-types, core-types etc. packages that correspond to the current logic packages.

@yamadapc how would that look for a consumer of Parcel types / API? Would they need to know where the types are defined, or would there still be a meta-package `@parcel/types`?

packages/core/types-internal/src/index.js
```diff
@@ -288,7 +327,7 @@ export type DetailedReportOptions = {|
 
 declare type GlobPattern = string;
 
-export type InitialParcelOptions = {|
+export type InitialParcelOptionsInternal<WorkerFarm> = {|
```
marcins, line 338: Nice.. this was the main piece I was missing from my attempt at removing circular deps, how to avoid moving all the `WorkerFarm` related types our of WorkerFarm (and the complexity with there being a `class` as a type..)

packages/core/types/src/index.js
```diff
@@ -0,0 +1,8 @@
+// @flow strict-local
+
+import type WorkerFarm from '@parcel/workers';
```
marcins, line 3: Does this mean `@parcel/types` still has a dependency on `@parcel/workers`? Does that bring in any other heavy deps?

packages/core/types-internal/src/index.js
```diff
@@ -1576,24 +1615,6 @@ export type BundleResult = {|
   +type?: string,
 |};
 
-export type GlobInvalidation = {|
```
mattcompiles, line 1579: Looks like this is no longer exported

packages/core/types-internal/src/index.js
```diff
@@ -1576,24 +1615,6 @@ export type BundleResult = {|
...
-  glob: Glob,
-|};
-
-export type FileInvalidation = {|
```
mattcompiles, line 1583: Looks like this is no longer exported

packages/core/types-internal/src/index.js
```diff
@@ -1576,24 +1615,6 @@ export type BundleResult = {|
...
-  filePath: FilePath,
-|};
-
-export type FileAboveInvalidation = {|
```
mattcompiles, line 1587: Looks like this is no longer exported

packages/core/types-internal/src/index.js
mattcompiles, line 338: ~This is a breaking change for this package. Is there a way we can retain the old type?~

Ignore me, I can see this is the new package

packages/dev/repl/src/parcel/BrowserPackageManager.js
```diff
@@ -4,7 +4,7 @@ import type {FileSystem} from '@parcel/fs';
 import type {
   PackageManager,
   Invalidations,
-  ResolveResult,
+  PackageManagerResolveResult,
```
mattcompiles: This is also technically a breaking change yeah?

packages/core/profiler/package.json
yamadapc: Good question ; let me remove this, I'm not sure

yamadapc: Okay the issue here is that `flow-to-ts` generates both `*.d.ts` and `*.ts` files.

Then when we run `tsc --emitDeclarationOnly ...` it fails because the `*.d.ts` files are already there.

This is broken on `v2` (`build-ts` fails on master due to this) & should be fixed by this change

packages/core/types/src/index.js
yamadapc: It does not bring heavy dependencies (that should be fine) but yes

packages/core/types-internal/src/index.js
yamadapc: Fixed thanks

packages/dev/repl/src/parcel/BrowserPackageManager.js
yamadapc: Yeah... It is a breaking change. We can rename the re-export.

yamadapc: I have fixed the breaking changes.

yamadapc:
> how would that look for a consumer of Parcel types / API? Would they need to know where the types are defined, or would there still be a meta-package @parcel/types?

There could still be `@parcel/types` for public use, but we wouldn't use it internally.

I think depending on types only rather than implementation is a relatively niche use-case ; I'd think most of the time you need to bring in the implementation to do things at least as a peerDependency; so it's guaranteed to be brought in anyway to development/distribution cases, but I'm not familiar with how these packages are being used.

Internally (in this repository) I think where we need to express a dependency on types only (interfaces) then it can be split into multiple packages (`fs-types`, `cache-types` etc.).

packages/core/package-manager/index.d.ts
```diff
@@ -1,8 +1,18 @@
-import type {FilePath} from '@parcel/types';
+import type {FilePath, PackageInstaller, PackageManager} from '@parcel/types';
+import type {PackageManagerResolveResult} from '@parcel/types';
```
mattcompiles: Not a biggy, but this could be merged onto the above line.

mattcompiles (approved):
Good job getting through this bud. Looks good to me 👏

mattcompiles (approved):

PR #9612, opened 2024-04-03
Remove src to lib rewrites on package.json files

yamadapc:
This commit modifies the strategy to local src/lib rewrites, so that the package.json files are never modified.

Intead, the package.json files always point at `bin/...`; which in turn requires `lib/bin.js`.

During development, `lib/bin.js` can just require `src/bin.js` and rely on require.extensions register. But on distribution `lib/**` can be pre-built.

By maintaining a stable `bin/...` entry-point, the following benefits are achieved:

* The git workspace does not become dirty after a build
* It's not possible to get into a state where `node_modules/.bin/parcel` points at a non existant `lib/` file, which can only be fixed by removing node_modules and yarn installing again

## ✔️ PR Todo

- [x] Verify distributed assets from CI


packages/core/parcel/package.json
```diff
@@ -14,9 +14,12 @@
     "type": "git",
     "url": "https://github.com/parcel-bundler/parcel.git"
   },
-  "bin": "src/bin.js",
+  "bin": "bin/parcel.js",
   "main": "lib/bin.js",
   "source": "src/bin.js",
+  "scripts": {
+    "dev:prepare": "rm -rf ./lib/ && mkdir -p lib && cp ./bin/dev-bin.js ./lib/bin.js"
```
MonicaOlejniczak: /nit Although this works in CI because Actions has these global commands available, it wouldn't work in windows. That's why most of the scripts use `rimraf` and you'd need equivalents for the others too. I'm not sure if windows is well supported for contributing in general though.

package.json
```diff
@@ -12,7 +12,7 @@
   ],
   "scripts": {
     "build": "yarn build-bundles && cross-env NODE_ENV=production PARCEL_BUILD_ENV=production gulp",
-    "build-bundles": "rimraf --glob packages/*/*/lib && cross-env NODE_ENV=production PARCEL_BUILD_ENV=production PARCEL_SELF_BUILD=true parcel build packages/core/{fs,codeframe,package-manager,utils} packages/reporters/{cli,dev-server} packages/utils/{parcel-lsp,parcel-lsp-protocol}",
+    "build-bundles": "rimraf --glob packages/*/*/lib && lerna run dev:prepare && cross-env NODE_ENV=production PARCEL_BUILD_ENV=production PARCEL_SELF_BUILD=true parcel build packages/core/{fs,codeframe,package-manager,utils} packages/reporters/{cli,dev-server} packages/utils/{parcel-lsp,parcel-lsp-protocol}",
```
MonicaOlejniczak, line 15: Were you going to ensure that `dev:prepare` does not run during the release? It gets triggered via `prepublishOnly` FYI

yamadapc: This is fine & required as long as `build` will run afterwards.

In order for parcel to self-build, the production build sequence needs to be:

* Link `node_modules/.bin/parcel -> packages/core/parcel/bin/parcel.js` ; this needs to point at the unbuilt file at this point
* `yarn build-bundles` go and run parcel over `@parcel/fs` and such; parcel needs to be pointed at an unbuilt file, because if it pointed at a pre-built dir. we would have a task cycle
* now `yarn build / gulp` will run and all the `lib` directories written in `dev:prepare` should be overwritten. 

We can add extra validation to make sure we don't publish the wrong assets though.

MonicaOlejniczak: Yeah, it's more about making sure the files get overwritten. Otherwise the build / release should fail imo

packages/core/parcel/package.json
```diff
@@ -14,9 +14,13 @@
     "type": "git",
     "url": "https://github.com/parcel-bundler/parcel.git"
   },
-  "bin": "src/bin.js",
+  "bin": "bin/parcel.js",
   "main": "lib/bin.js",
   "source": "src/bin.js",
+  "scripts": {
+    "prepublishOnly": "./ensure-no-dev-lib.sh",
```
MonicaOlejniczak: Could you just have this on the root `prebublishOnly` script and check the entire repo? It would mean you only need one bash script too

yamadapc: I don't know if the root hook is called when children are published, but I'd guess not?

MonicaOlejniczak: AFAIK the entire publish pipeline currently relies on the root script to build the packages since it's not explicitly done in the github release workflows cc @mattcompiles

yamadapc: I think the root script will build the packages, but at the npm/yarn publish level this would be run by the package manager

yamadapc: Having per package scripts is a bit nicer because you can run them separately with built-in tooling (+ possibly supercharge with turbo-repo)

MonicaOlejniczak: I personally don't like the repetition, but it's not a big thing. /nit you could at least share the script in root `scripts` folder since it's all the same

yamadapc: Yeah; that's okay. 

So maybe I can symlink these files. Let me try

yamadapc: There is no repetition anymore; only symlinks

mattcompiles:
@yamadapc Is my understanding correct that to use Parcel via parcel link you'll now need to run `yarn lerna run dev:prepare` beforehand? Otherwise there won't be anything present in `lib/bin.js`?

yamadapc:
>  @yamadapc Is my understanding correct that to use Parcel via parcel link you'll now need to run yarn lerna run dev:prepare beforehand? Otherwise there won't be anything present in lib/bin.js?

That's right. We could add a postinstall hook or such that runs `dev:prepare` on local.

But IMO this is better because before you also would get corrupted workspaces on builds linked or otherwise.

I have a feeling at the moment that you can run `yarn build` once and after that your workspace is unable to build itself again. Also there are those dirty files & yarn over a dirty workspace will break node_modules links forever so then you need to clean-up.

I'm running benchmarks in https://github.com/parcel-bundler/parcel-benchmark-action with prod. builds and in the current setup the workspace gets corrupted every time I build.

So this change would help; there also is no way to commit the dirty files.

The specific workflow for parcel-link might change very slightly; but it should be alright.

mattcompiles (approved):

PR #9614, opened 2024-04-03
skip joining on absolute paths

Nikola-3:
# ↪️ Pull Request

## ✔️ PR Todo


mischnic:
You could also use `path.resolve` which should do exactly that

yamadapc (approved):
LGTM; using `path.resolve` would also work, but `projectRoot` needs to be an absolute path; otherwise it'll change the return value of this function

mattcompiles (approved):

PR #9618, opened 2024-04-04
Implements a `getFeatureFlag` to avoid option drilling

marcins:
# ↪️ Pull Request

Sometimes it's inconvenient to drill options into a new place just to access feature flags, passing feature flag values around also can make cleanup of a feature more complicated. This change implements a `getFeatureFlag` export from` @parcel/feature-flags` that allows you to get a feature flag value from anywhere.

Code is added to initialise the values in core, as well as in workers.

## ✔️ PR Todo

### DIFF
--- a/packages/core/core/src/Parcel.js
+++ b/packages/core/core/src/Parcel.js
@@ -53,6 +53,7 @@ import {
   fromProjectPathRelative,
 } from './projectPath';
 import {tracer} from '@parcel/profiler';
+import {setFeatureFlags} from '@parcel/feature-flags';
 
 registerCoreWithSerializer();
 
@@ -107,6 +108,8 @@ export default class Parcel {
     this.#resolvedOptions = resolvedOptions;
     let {config} = await loadParcelConfig(resolvedOptions);
     this.#config = new ParcelConfig(config, resolvedOptions);
+
+    setFeatureFlags(resolvedOptions.featureFlags);
 
     if (this.#initialOptions.workerFarm) {
       if (this.#initialOptions.workerFarm.ending) {
--- a/packages/core/core/src/worker.js
+++ b/packages/core/core/src/worker.js
@@ -26,6 +26,7 @@ import {clearBuildCaches} from './buildCache';
 import {init as initSourcemaps} from '@parcel/source-map';
 import {init as initRust} from '@parcel/rust';
 import WorkerFarm from '@parcel/workers';
+import {setFeatureFlags} from '@parcel/feature-flags';
 
 import '@parcel/cache'; // register with serializer
 import '@parcel/package-manager';
@@ -78,6 +79,9 @@ async function loadConfig(cachePath, options) {
   );
   config = new ParcelConfig(processedConfig, options);
   parcelConfigCache.set(cachePath, config);
+
+  setFeatureFlags(options.featureFlags);
+
   return config;
 }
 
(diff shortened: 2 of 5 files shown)


mattcompiles (approved):

packages/core/core/src/Parcel.js
```diff
@@ -108,6 +109,8 @@ export default class Parcel {
     let {config} = await loadParcelConfig(resolvedOptions);
     this.#config = new ParcelConfig(config, resolvedOptions);
 
+    setFeatureFlags(resolvedOptions.featureFlags);
```
MonicaOlejniczak, line 112: Since this is all global, wouldn't it break when you run parallel parcel builds if they overwrite the other's ffs? Or would it always be encapsulated in different processes that it wouldn't be a problem? Either way, might want to test that.

yamadapc (approved):
lgtm

packages/core/core/src/Parcel.js
marcins: > parallel parcel builds

Unless I'm misunderstanding, if you're running more than one instance of Parcel it'll have it's own workers.

Within a single execution of Parcel (and it's child workers) it's assumed the feature flag values are static.

marcins: > would it always be encapsulated in different processes
Yes, it will, every Parcel run is isolated.

mischnic: I agree with @MonicaOlejniczak. Each build will have a separate `Parcel` instance and the CLI has a separate process for each build.
But if you use the JS API, then you might have multiple parallel builds in the same process which would share the feature flags stored in `packages/core/feature-flags/src/index.js`.

marcins: Wouldn't this be true of any of the build caches? Are any of those "unsafe" to share between instances?

And just to clarify what you mean, do you mean a case where you might:
```
const a = new Parcel({ ... featureFlags: { flag: true }});
const b = new Parcel({ ... featureFlags: { flag: false }});
Promise.all([a.run(), b.run()]);
```

Not sure how you'd solve for that, except not having FFs be globally accessible, because then you're back to drilling a client or instance id or something down to where you need the feature flag.

mischnic: > Wouldn't this be true of any of the build caches? Are any of those "unsafe" to share between instances?

That is indeed true, though `createBuildCache` is only used internally and not in the public API so far. So this could be fixed without breaking changes.

@devongovett Any thoughts?

For example these just key on the plugin name but not the project root
https://github.com/parcel-bundler/parcel/blob/a24a30d2492b329a076841ed512184dc5dc16bc9/packages/core/core/src/requests/PathRequest.js#L211-L219

https://github.com/parcel-bundler/parcel/blob/a24a30d2492b329a076841ed512184dc5dc16bc9/packages/core/core/src/PackagerRunner.js#L229-L237

> And just to clarify what you mean, do you mean a case where you might:

Yes 

> because then you're back to drilling a client or instance id or something

That's why the `ParcelOptions` are passed around in many places (for project root, filesystem, ...), which has been fine so far.

marcins: > That's why the ParcelOptions are passed around in many places (for project root, filesystem, ...), which has been fine so far.

I guess feature flags are "special" because you might need them somewhere deep down and passing them through many layers can be clumsy. This also makes it more difficult to cleanup a feature than just searching for `getFeatureFlag('myflag')` and removing all the conditions.

The other aspect of feature flags is that while they're part of the API by necessity, we're not expecting that they'll be used by the general population - we certainly don't have a use-case (admittedly _today_) where we'd be running parallel builds in a single process with different feature flag values.

Perhaps a warning (or error?) if `setFeatureFlag` is called multiple times with different values would suffice?

PR #9622, opened 2024-04-04
Fix tracing problems

yamadapc:
This PR fixes issues with the tracing implementation

Namely:

* Makes use of BEGIN and END events rather than COMPLETE events only; this improves the output on some cases
* Adds more instrumentation

### DIFF
--- a/packages/core/core/src/requests/ConfigRequest.js
+++ b/packages/core/core/src/requests/ConfigRequest.js
@@ -26,7 +26,7 @@ import PublicConfig from '../public/Config';
 import {optionsProxy} from '../utils';
 import {getInvalidationHash} from '../assetUtils';
 import {hashString, Hash} from '@parcel/rust';
-import {PluginTracer} from '@parcel/profiler';
+import {measureAsyncFunction, PluginTracer} from '@parcel/profiler';
 import {requestTypes} from '../RequestTracker';
 import {fromProjectPath, fromProjectPathRelative} from '../projectPath';
 import {createBuildCache} from '../buildCache';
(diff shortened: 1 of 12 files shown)


packages/core/core/src/requests/ConfigRequest.js
```diff
@@ -74,37 +74,43 @@ export type ConfigRequest = {
...
+export function loadPluginConfig<T: PluginWithLoadConfig>(
   loadedPlugin: LoadedPlugin<T>,
   config: Config,
   options: ParcelOptions,
 ): Promise<void> {
-  let loadConfig = loadedPlugin.plugin.loadConfig;
-  if (!loadConfig) {
-    return;
-  }
+  return measureAsyncFunction(
```
devongovett, line 82: what's the goal of measuring internal functions like this? This would be captured by the normal v8 profiler already. I thought the goal of tracing is to capture _high level_ timings for user-facing things like plugins. If we start tracing internal things in Parcel core this gets muddier and harder for end users to understand.

yamadapc: The goal is to capture high-level timings, so lower resolution than what the profiler will spit out (& lower or close to none overhead).

We could trace more internal bits though. It should not be granular.

I don't really care about the measurements added in this diff. More about the fixes.

For the `measureAsyncFunction` / `measureFunction` helpers ; they prevent mistakes where `.end` isn't currently called.

That was the case on `packages/core/core/src/Transformation.js`, where you can see I added `measurement.end()` calls that were previously not there. For those branches we had missing traces. Ideally we'd replace all `measurement.end()` calls with function wrappers, since then we can't not call `.end()`.

packages/reporters/tracer/src/TracerReporter.js
```diff
@@ -66,6 +74,7 @@ export default (new Reporter({
       case 'buildSuccess':
       case 'buildFailure':
         nullthrows(tracer).flush();
+        writeStream?.write(']');
```
marcins, line 77: Doesn't hurt, but not really necessary for a trace - the format is designed to work when the writer died before it could cleanly finish.

packages/core/types/index.js
```diff
@@ -1979,7 +1979,8 @@ export type ValidationEvent = {|
  * @section reporter
  */
 export type TraceEvent = {|
-  +type: 'trace',
```
marcins, line 1982: Is this a breaking change?

packages/core/types/index.js
```diff
@@ -1979,7 +1979,8 @@ export type ValidationEvent = {|
...
+  +type: 'trace.start' | 'trace.end',
+  +id: string,
   +ts: number,
   +duration: number,
```
marcins, line 1985: There's a flow error in your calls to `.trace()` because the trace event expects a duration which now won't be relevant with `begin`/`end` events.

PR #9627, opened 2024-04-08
Fix missing content key in symbol prop when asset is removed

mattcompiles:
# ↪️ Pull Request

This PR addresses a bug in incremental symbol propagation where assets are requested for propagation even if they are no longer contained within the AssetGraph. This can happen in the following circumstance:
- Build fails during the `AssetGraphRequest` but finds some new Assets for propagation
- Next build runs but the previously discovered Assets are no longer in the AssetGraph (e.g. removed import)

## 🚨 Test instructions

Integration test added

## ✔️ PR Todo

### DIFF
--- a/packages/core/integration-tests/test/symbol-propagation.js
+++ b/packages/core/integration-tests/test/symbol-propagation.js
@@ -1,0 +1,33 @@
+import assert from 'assert';
+import path from 'path';
+import {bundler, run, overlayFS, fsFixture} from '@parcel/test-utils';
+
+describe('symbol propagation', () => {
+  it('should handle removed assets from previous failed builds', async () => {
+    await fsFixture(overlayFS, __dirname)`
+        broken.js:
+            module.exports = require('./missing.js');
+        working.js:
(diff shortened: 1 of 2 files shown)


packages/core/integration-tests/test/symbol-propagation.js
```diff
@@ -0,0 +1,33 @@
...
+    });
+
+    await overlayFS.writeFile(
+      path.join(__dirname, 'index.js'),
+      `module.exports = require('./working.js');`,
+    );
+
+    let {bundleGraph} = await b.run();
+
+    assert(await run(bundleGraph), 'ITS WORKING');
```
marcins, line 31: I'm assuming without your change this test fails with the "missing key in symbol propagation" error we've been seeing?

marcins (approved):
Nice find and fix!

packages/core/integration-tests/test/symbol-propagation.js
mattcompiles: That's right 😄

PR #9633, opened 2024-04-10
Collect request stats

benjervis:
# ↪️ Pull Request

In order to get a picture of the cache performance in our application, we want to get the request stats and the number of visited `asset_group`s out of each Parcel build.

These get `logger`ed, to be read by a custom reporter.


packages/core/core/src/RequestTracker.js
```diff
@@ -1227,6 +1231,36 @@ export default class RequestTracker {
...
+      messageLines.push(`${requestTypeName}\t\t${count}`);
+    }
+
+    let message = messageLines.join('\n');
+
+    logger.verbose({
+      origin: '@parcel/core',
+      message,
+      meta: {
+        requestStats: formattedStats,
```
marcins: Won't formatted stats be more difficult to deal with as metrics on the other end?

I don't think `RequestTracker` should be responsible for formatting output, it should send the raw data, and the reporter that deals with it can format them for output if needed (or send the raw values to a backend).

benjervis: Might be poor variable naming, `formattedStats` is just the stats but with the requestType numeric codes translated in to the names (i.e. `{ 'asset_graph_request': 1487, ...` rather than `{ 3: 1487, ...`). It is still just raw data though

marcins: Ah, yes, my bad - I missed the `formattedStats` vs `messageLines` being two different things used differently. All good!

marcins (approved):

packages/core/core/src/RequestTracker.js
```diff
@@ -1227,6 +1231,36 @@ export default class RequestTracker {
     }
   }
 
+  flushStats() {
+    let requestTypeEntries = {};
+    type RequestTypes = Array<$Keys<typeof requestTypes>>;
+
+    for (let key of (Object.keys(requestTypes): RequestTypes)) {
```
MonicaOlejniczak: /nit you could just create this mapping at the top of the file so it's done once, and it could be more idiomatic using `Object.entries`

packages/core/core/src/Parcel.js
```diff
@@ -385,6 +385,7 @@ export default class Parcel {
         },
       };
 
+      this.#requestTracker.flushStats();
```
MonicaOlejniczak: I don't know if we want to go down this rabbit hole, but it's pretty common for success events to store stats regarding the build. Instead of having `flushStats`, you could add the stats directly to build success under a new key and then get the reporter to log it.

That way the `RequestTracker` does not have to care about how to report the info, it just gives it to whatever wants to consume it. This is similar to what @marcins noted before, but can be extended to other stats we want to measure.

mattcompiles: Agree this would be nicer. We opted not to do it that way to avoid adding Parcel public API that would be a breaking change to remove or modify in future. Thoughts?

MonicaOlejniczak: As an aside, I'm looking at adding high-level (and potentially more granular) perf metrics and am starting to explore options. Having a stats object or something similar to that would be useful.

In regards to the public API, you could name the field as `unstable_stats` to make no guarantees to consumers. Or in a similar fashion, expose that there is a stats object but type it as a record with any value. That way, we are saying we will provide it but make no guarantees about what's in there. We'd probably need something internal to give us types we know about to make logs or w/e on them, but it doesn't need to be exposed.

packages/core/core/src/RequestTracker.js
mattcompiles: True about top of the file. However we used keys because the flow types to get entries happy was super ugly.

devongovett:
What's the difference between this and tracing? They seem to have similar goals. Maybe the tracer could track requests in a similar way?

packages/reporters/tracer/src/TracerReporter.js
```diff
@@ -66,14 +66,14 @@ export default (new Reporter({
       case 'buildSuccess':
       case 'buildFailure':
         nullthrows(tracer).flush();
+        tracer = null;
```
mattcompiles, line 69: @marcins This actually fixes a race condition that showed up in the tests where it's possible that the tracer isn't cleared quick enough.

mattcompiles:
@devongovett I believe tracing has a performance overhead and is used for in depth analysis of one-off builds. This is something we want to measure across all builds. The `visitedAssetGroupsCount` is pretty unideal but allows us to create an exact cache ratio for transforms and packaging.

PR #9636, opened 2024-04-12
Suspend incomplete requests

benjervis:
Some investigations during a previous PR (#9633) revealed that the long standing DevDepRequest error might be caused by a worker farm-related race condition.

As a possible fix for it, we've added a registry of incomplete requests where we suspend the processing of `runRequest` until the details have been returned from worker land.

It's currently locked behind a feature flag until we've done a bit more testing in the wild.

### DIFF
--- a/packages/core/core/src/RequestTracker.js
+++ b/packages/core/core/src/RequestTracker.js
@@ -54,6 +54,7 @@ import {report} from './ReporterRunner';
 import {PromiseQueue} from '@parcel/utils';
 import type {Cache} from '@parcel/cache';
 import {getConfigKeyContentHash} from './requests/ConfigRequest';
+import {getFeatureFlag} from '@parcel/feature-flags';
 
 export const requestGraphEdgeTypes = {
   subrequest: 2,
@@ -132,6 +133,7 @@ type Request<TInput, TResult> = {|
   id: string,
   +type: RequestType,
   input: TInput,
+  ensureCache?: boolean,
   run: ({|input: TInput, ...StaticRunOpts<TResult>|}) => Async<TResult>,
 |};
 
@@ -1027,6 +1029,10 @@ export default class RequestTracker {
   farm: WorkerFarm;
   options: ParcelOptions;
   signal: ?AbortSignal;
+  incompleteRequests: Map<
+    Request<mixed, mixed>['id'],
+    {|deferred: Deferred<void>, promise: Promise<void>, timerId: TimeoutID|},
+  > = new Map();
 
   constructor({
     graph,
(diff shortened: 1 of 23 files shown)


marcins (approved):
Seems reasonable to me, especially behind an FF so we can validate if that's the cause of our issues.

Note though you have a Flow error that'll prevent merging.

packages/core/core/src/requests/DevDepRequest.js
```diff
@@ -190,14 +190,22 @@ export async function runDevDepRequest<TResult>(
...
-      for (let filePath of nullthrows(devDepRequest.invalidateOnFileChange)) {
+      for (let filePath of nullthrows(
+        devDepRequest.invalidateOnFileChange,
+        'DevDepRequest missing invalidateOnFileChange',
+      )) {
         api.invalidateOnFileUpdate(filePath);
         api.invalidateOnFileDelete(filePath);
       }
 
       for (let invalidation of nullthrows(
```
JakeLane, line 206: We should probably look at adding a lint rule to ensure `nullthrows` always has a description

mattcompiles: Could just change the type definition. It'd be a bit of work though 😅

packages/core/core/src/RequestTracker.js
```diff
@@ -1029,6 +1031,10 @@ export default class RequestTracker {
   options: ParcelOptions;
   signal: ?AbortSignal;
   stats: Map<RequestType, number> = new Map();
+  incompleteRequests: Map<
```
devongovett, line 1034: We already have `RequestGraph#incompleteNodeIds` and also `incompleteNodePromises`. Could you use those here?

bjornwah:
Any progress on this? We're having a lot of "Unexpected undefined" which come from the nullthrows() stuff in runDevDepRequest. It is extra problematic if running stuff in parallel or running other heavy background tasks. That probably affects the timing of stuff?

About the nullthrows; the type 'DevDepRequest' is defined to have optional invalidateOnFileCreate/invalidateOnFileChange/invalidateOnStartup so I find it a bit odd that we even throw for that case. 

That being said the issue is elsewhere - where this object is being constructed/populated. I'm struggling to follow where the code goes from [`here`](https://github.com/parcel-bundler/parcel/blob/56ffb67bf22ad0e218696c50840d9268c609f129/packages/core/core/src/requests/AssetRequest.js#L122) where the devDepRequests are constructed, but that thing sometimes returns incomplete devDepRequests.

bjornwah:
We have now abandoned parcel altogether because of this issue :'(

PR #9637, opened 2024-04-12
Add measure tracing API

MonicaOlejniczak:
# ↪️ Pull Request

This change is similar to https://github.com/parcel-bundler/parcel/pull/9622 but instead introduces non-breaking changes that:
* Introduce a more flexible `createTraceMeasurement` API, by only instantiating relevant event data when tracing is enabled at the consumer side
* Add testing to prevent consumers from ending a measurement, using a new `traceStart` event

Additionally, some arguments have been updated to be more meaningful.

## 💻 Examples

N/A

## 🚨 Test instructions

Run with tracing enabled

### DIFF
--- a/packages/core/core/src/PackagerRunner.js
+++ b/packages/core/core/src/PackagerRunner.js
@@ -396,47 +396,51 @@ export default class PackagerRunner {
 
     let packager = await this.config.getPackager(bundle.name);
     let {name, resolveFrom, plugin} = packager;
-    let measurement;
     try {
-      measurement = tracer.createMeasurement(name, 'packaging', bundle.name, {
-        type: bundle.type,
-      });
-      return await plugin.package({
-        config: configs.get(name)?.result,
-        bundleConfig: bundleConfigs.get(name)?.result,
-        bundle,
-        bundleGraph: new BundleGraph<NamedBundleType>(
-          bundleGraph,
-          NamedBundle.get.bind(NamedBundle),
-          this.options,
-        ),
-        getSourceMapReference: map => {
-          return this.getSourceMapReference(bundle, map);
+      return await tracer.measure(
+        {
+          name,
+          args: {bundle: {name: bundle.name, type: bundle.type}},
+          categories: ['packaging'],
         },
-        options: this.pluginOptions,
-        logger: new PluginLogger({origin: name}),
-        tracer: new PluginTracer({origin: name, category: 'package'}),
-        getInlineBundleContents: async (
-          bundle: BundleType,
-          bundleGraph: BundleGraphType<NamedBundleType>,
-        ) => {
-          if (bundle.bundleBehavior !== 'inline') {
-            throw new Error(
-              'Bundle is not inline and unable to retrieve contents',
-            );
-          }
+        () =>
+          plugin.package({
+            config: configs.get(name)?.result,
+            bundleConfig: bundleConfigs.get(name)?.result,
+            bundle,
+            bundleGraph: new BundleGraph<NamedBundleType>(
+              bundleGraph,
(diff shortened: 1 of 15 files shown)


packages/core/core/src/PackagerRunner.js
```diff
@@ -396,47 +396,52 @@ export default class PackagerRunner {
...
-            bundleGraphToInternalBundleGraph(bundleGraph),
-            configs,
-            bundleConfigs,
-          );
-
-          return {contents: res.contents};
+      return await tracer.measureAsync(
+        {
+          name,
+          args: {bundle: bundle.name},
```
mattcompiles: This also used to track bundle type as well. Did you intend to drop that?

MonicaOlejniczak: Nope, will add it back in. Thanks for picking that up

JakeLane (approved):

packages/core/profiler/src/Tracer.js
```diff
@@ -64,44 +65,44 @@ export default class Tracer {
...
+  measure<T>(
+    {args = {}, categories, name}: MeasurementOptions,
+    fn: () => T,
+  ): T {
+    if (!this.enabled) {
+      return fn();
     }
-  }
 
-  createMeasurement(
```
marcins, line 76: The original reasoning for having an API that didn't involve passing an object was to avoid the overhead of unnecessary object allocation in the case when tracing was disabled (which is most of the time).

Have you checked that the impact of performance on builds with tracing disabled of this change? Does timing/memory usage look okay? I just want to make sure we're not introducing unnecessary GC thrashing that might affect builds the size of Jira's.

MonicaOlejniczak: I created a crude benchmark that can be run with `node --trace-gc --trace-gc_verbose test.mjs` with this, commenting in one test at a time. I will revert the existing change so that it uses something more performant with more flexibility.

```js
let enabled = false;

class Measurement {
  constructor(args) {
    this.args = args;
    this.start = performance.now();
  }

  end() {
    const end = performance.now();

    // tracer.trace();
  }
}

async function measure(args, fn) {
  if (!enabled) {
    return fn();
  }

  const measurement = new Measurement(args);
  let result;
  try {
    result = await fn();
  } finally {
    measurement.end();
  }

  return result;
}

for (let i = 0; i < 100_000_000; i++) {
  // fastest -> slowest
  await measure('world', async () => {});
  // await measure(enabled && { hello: 'world' }, async () => {});
  // await measure(enabled ? { hello: 'world' } : null, async () => {});
  // await measure({ hello: 'world' }, async () => {});
  // await measure(() => ({ hello: 'world' }), async () => {});
  // await measure({ get hello() { return 'world' }}, async () => {});
}

// for (let i = 0; i < 1_000_000_000; i++) {
//   measure({ hello: 'world' }, () => {});
//   measure('world', () => {});
// }
```

packages/core/core/src/PackagerRunner.js
```diff
@@ -398,9 +398,14 @@ export default class PackagerRunner {
     let {name, resolveFrom, plugin} = packager;
     let measurement;
     try {
-      measurement = tracer.createMeasurement(name, 'packaging', bundle.name, {
-        type: bundle.type,
-      });
+      measurement =
+        tracer.enabled &&
```
marcins, line 402: I know moving this check to the consumer avoids the function call at all, but is the only reason to change the `create` signature from individual params to an object? I'm still not sure what the benefit is apart from the naming?

MonicaOlejniczak: It's the naming and flexibility mostly, yeah

PR #9196, opened 2023-08-16
Do not reverse runtimes

devongovett:
Looks like this came from https://github.com/parcel-bundler/parcel/pull/8837 but later the first reverse got removed in https://github.com/parcel-bundler/parcel/pull/8981. Second one seemed to have caused module not found errors at runtime when there are multiple entry script tags in an HTML file.

PR #8837, opened 2023-02-17
Split large runtime manifest into separate bundles

mattcompiles:
# ↪️ Pull Request

To prevent unnecessary cache invalidation, Parcel currently [inlines manifests into entry bundles](https://parceljs.org/features/production/#cascading-invalidation). However, in projects where the entries are very large this still leads to large amounts of unnecessary cache invalidation. To address this, this PR adds a new configurable feature to Parcel that can split runtime manifest assets out into their own bundles.

Basic implementation is as follows:
- Allow runtimes to return a new value (`moveToSeparateBundle`) that signifies a runtime asset should have it's own bundle
- When processing the runtime assets, if `moveToSeparateBundle` is set then create a new bundle inside the current bundle group and add the asset to that bundle instead. 

To allow for this change, `createBundle` needed to be moved to the internal bundle graph. 

## Improvement

After making a small change to a commonly used dynamic import in a very large project here's the different in outputs. 

**No manifest splitting**
35 / 3290 (1.1%) bundles changed
32490 / 217051 (15%) bytes changed

**With manifest splitting**
35 / 3322 (1.1%)  bundles changed
2376 / 217062 (1.1%) bytes changed

## Questions/Discussions

- Currently the config is called `splitManifestThreshold`. Not sure this is a great name. We could also just hard code a value if we don't think the config is useful. 
- I've set a pretty high/safe value as the default (100,000 bytes). Are we happy wtih this? We could get a bit more scientific about it. 
- I think it'd be useful to name these bundles so we know what they are (e.g. `runtime-manifest.HASH.js`). Maybe we should add a new property to `Bundle` to let namer plugins know this is a runtime/manifest bundle?

## 🚨 Test instructions

I've added a couple of new tests but happy to add more if we feel there's anything missing. 

## ✔️ PR Todo

- [x] Added/updated unit tests for this change
- [x] Filled out test instructions (In case there aren't any unit tests)
- [ ] Included links to related issues/PRs

PR #9646, opened 2024-04-15
Fix cannot find module error with runtimes in multiple bundles

devongovett:
Closes #9196

This fixes a cannot find module error that occurs when you have an HTML page with two script tags, each with an async import. It occurs because runtimes are processed in reverse order, so the bundle manifest gets put in the second script tag rather than the first, even though the first also has a dependency on it.

This behavior was introduced in #8837 and subsequently updated in #8981. I believe the intent was to place parallel runtime bundles before the bundles that used them, but it inadvertently also changed the order of other runtimes as well. The fix is to handle parallel runtimes separately, by inserting them at the beginning of the list of connections instead of the end, while treating other runtimes normally.

Added test and existing tests pass, but would be good to verify in applications.


mattcompiles:
Looks good @devongovett. I'll test this and get back to you.

mattcompiles:
Hey @devongovett 
I tested this locally and I'm getting a bunch of `Could not resolve bundle with id` errors. When I search the dist directory though I can't see any bundles that actually define the bundle ID though, only references. 🤔

devongovett:
Oh no! What are we missing here? I thought the logic made sense...

mattcompiles:
@devongovett Comparing the outputs I can see that there's runtime bundles being missed in the manifests which causes the error. I'm 95% sure this is because you removed the reverse topological processing of the bundles. 

i.e.
```js
  // As manifest bundles may be added during runtimes we process them in reverse topological
  // sort order. This allows bundles to be added to their bundle groups before they are referenced
  // by other bundle groups by loader runtimes
  let bundles = [];
  bundleGraph.traverseBundles({
    exit(bundle) {
      bundles.push(bundle);
    },
  });
```

devongovett:
@mattcompiles but the reverse order was what caused the other issue that this PR is supposed to solve... so not sure what we should do here. Maybe it somehow needs to be done in two passes: first generate all the runtimes (including creating new bundles), and then generate the manifests once all bundles are known?

Are those missing runtime bundles generated by a custom plugin or the JS runtime? If the JS runtime, I think the only use of parallel runtimes is for the manifest itself. Can manifest bundles point to other manifest bundles? Are those the ones that are missing?

I guess we are missing a test here because the tests that were added in the original PRs for this feature seem to be passing. So that's why I'm wondering if maybe it's a custom plugin in Jira?

mattcompiles:
@devongovett 

> Are those missing runtime bundles generated by a custom plugin or the JS runtime?

There's no Jira specific code here. The code that decides whether a new bundle should be inserted is here: https://github.com/parcel-bundler/parcel/blob/331bed8a62b7073f4748ca8c8d150068f1fdde9a/packages/runtimes/js/src/JSRuntime.js#L723C6-L723C7

> Can manifest bundles point to other manifest bundles? Are those the ones that are missing?

Yes, this is the problem that's occurring. If you don't run reverse order, you end up inserting bundles in children, but the parent manifests have already been written.

> I guess we are missing a test here because the tests that were added in the original PRs for this feature seem to be passing.

There does seem to be a test missing here, I guess that's my bad. I'll see if I can replicate the issue in a test.

mattcompiles:
I've tried creating a test to replicate the scenario I'm seeing locally but no luck so far. I do know that it's related to manual shared bundles that get assigned a runtime bundle though. I'll keep thinking about we can replicate it in a test. I'd also be happy to pair on it sometime if it would help. Don't want to block your fixes going out.

devongovett:
Interesting. Maybe some kind of circular bundle reference? It sorta makes sense that you need all the bundles that will be referenced to have been created before you can correctly build the manifest. But on the other hand, bundles within the same bundle group get executed in order, so shared runtimes need to be placed in earlier bundles. Kind of a catch-22. That's why I mentioned needing two passes - one to create the bundles and a second to actually generate the manifests. But that would be a pretty big change to the API for runtimes...

PR #9648, opened 2024-04-16
DFS refactor work 

yamadapc:
* Using iterative DFS implementation for all cases
* Added unit-tests
---
I need help with:

* running a benchmark

I will do clean-up around:

* safe roll-out (like a feature-flag or more tests)
* remove old method

Missing from this diff is:

* Replacing `getChildren` with a few generator/callback based options


packages/core/graph/src/Graph.js
```diff
@@ -11,6 +11,7 @@ import type {
 import {BitSet} from './BitSet';
 
 import nullthrows from 'nullthrows';
+import {getFeatureFlag} from '@parcel/feature-flags/src';
```
JakeLane:
```suggestion
import {getFeatureFlag} from '@parcel/feature-flags';
```

packages/core/graph/src/Graph.js
```diff
@@ -289,11 +335,17 @@ export default class Graph<TNode, TEdgeType: number = 1> {
     ) {
       return this.dfsFast(enter, startNodeId);
```
JakeLane, line 336: Would be nice to have a comment here to explain why there are two different dfs implementations

packages/core/graph/src/Graph.js
```diff
@@ -449,15 +501,117 @@ export default class Graph<TNode, TEdgeType: number = 1> {
...
+          this._visited = visited;
+          return context;
+        }
+      } else {
+        let {nodeId, context} = command;
+        if (!this.hasNode(nodeId) || visited.has(nodeId)) continue;
+        visited.add(nodeId);
+
+        skipped = false;
+        let enter = typeof visit === 'function' ? visit : visit.enter;
```
JakeLane: Can we move this comparison outside of the loop?

packages/core/graph/src/Graph.js
yamadapc: Added JSDoc comments

packages/core/graph/src/Graph.js
yamadapc: Done, thank you

packages/core/graph/src/Graph.js
```diff
@@ -449,15 +501,122 @@ export default class Graph<TNode, TEdgeType: number = 1> {
...
+      {nodeId: traversalStartNode, context: null},
+    ];
+    const enter = typeof visit === 'function' ? visit : visit.enter;
+    while (queue.length !== 0) {
+      const command = queue.pop();
+
+      if (command.exit != null) {
+        let {nodeId, context, exit} = command;
+        let newContext = exit(nodeId, command.context, actions);
+        if (typeof newContext !== 'undefined') {
```
MonicaOlejniczak, line 552: /nit `!= null` checks are usually preferred in the codebase from what I can tell

MonicaOlejniczak (approved):

packages/core/graph/src/Graph.js
```diff
@@ -449,15 +501,122 @@ export default class Graph<TNode, TEdgeType: number = 1> {
...
+        if (skipped) {
+          continue;
+        }
+
+        if (stopped) {
+          this._visited = visited;
+          return context;
+        }
+
+        if (typeof visit !== 'function' && visit.exit) {
```
marcins, line 588: We assign `enter` to a const up on line 545 - any reason we don't do the same for `exit`? I'm guessing a `typeof` isn't a huge overhead anyway..

marcins (approved):

JakeLane (approved):

devongovett:
Do you have performance numbers for this? Where did this work come from?

PR #9658, opened 2024-04-19
Start to add rust backed request execution under a feature-flag by rewriting ConfigRequest

yamadapc:
Does not achieve anything in particular other than set patterns for exposing/integrating rust code under a flag.

While implementing unit-tests for the `ConfigRequest` I found a bug where `invalidateOnEnvChange` would not be picked-up on its own.

There are exhaustive tests for the config request current/previous implementations added.

For the hashing section, there is a slight missing part regarding sorting of object keys.


crates/node-bindings/src/core/requests/mod.rs
```diff
@@ -0,0 +1,95 @@
...
+  env: Env,
+  config_request: ConfigRequest,
+  api: JsObject,
+  options: JsObject,
+) -> napi::Result<()> {
+  // Technically we could move `env` to JSRequestAPI but in order to
+  // be able to use env on more places we rc it.
+  let env = Rc::new(env);
+  let api = JSRequestApi::new(env, api);
+  let input_fs = OsFileSystem::default();
```
yamadapc: missing this

crates/node-bindings/src/core/requests/config_request/mod.rs
```diff
@@ -0,0 +1,134 @@
...
+
+  if config_request.invalidate_on_build {
+    api.invalidate_on_build()?;
+  }
+
+  Ok(())
+}
+
+#[napi(object)]
+struct RequestOptions {}
```
yamadapc, line 154: missing this

crates/node-bindings/src/core/requests/config_request/mod.rs
```diff
@@ -0,0 +1,134 @@
...
+  pub invalidate_on_build: bool,
+}
+
+fn get_config_key_content_hash(
+  file_path: &str,
+  config_key: &str,
+  input_fs: &impl FileSystem,
+  project_root: &str,
+) -> napi::Result<String> {
+  todo!("")
```
yamadapc: missing this

.mocharc.json
```diff
@@ -1,5 +1,5 @@
 {
-  "spec": "packages/*/!(integration-tests)/test/*.js",
+  "spec": "packages/*/!(integration-tests)/test/{*.js,**/*.{test,spec}.js}",
   "require": ["@parcel/babel-register", "@parcel/test-utils/src/mochaSetup.js"],
   // TODO: Remove this when https://github.com/nodejs/node/pull/28788 is resolved
```
marcins, line 4: Unrelated to your change, but I was curious and it looks like this issue was closed in 2019 :P

devongovett:
I actually want to remove ConfigRequest completely. It's not a real request at the moment actually since it doesn't do any work. The configs have actually already been loaded, and the ConfigRequest is fake just to setup the request graph structure. In reality, it isn't necessary at all - the invalidations could just be added to the parent request. The config loading doesn't get invalidated separately right now anyway.

In addition, it looks like there's a lot of bridging here. Since ConfigRequest doesn't actually do any work other than mutating the graph, now we're calling into rust only to call back into JS several times. Maybe ConfigRequest was just an example of the setup but perhaps it would be more efficient to do the request work in rust and send back all the invalidations at once as part of the result rather than calling back to JS via the RequestApi a whole bunch of times?

yamadapc:
That makes sense; the purpose of this PR is:

* Start migrating
* Show JS delegation / interop patterns we can use
* Write tests

>  Maybe ConfigRequest was just an example of the setup but perhaps it would be more efficient to do the request work in rust and send back all the invalidations at once as part of the result rather than calling back to JS via the RequestApi a whole bunch of times?

Yes this is just an example so it doesn't do anything interesting.

That said, I feel strongly we should not optimise anything without measuring. The benchmarks added in https://github.com/yamadapc/napi-benchmark/tree/main show there's very low overhead to call JavaScript functions from rust (nanosecond cost for calls in of themselves).

Optimising without benchmarks or measurements has a high risk of implementing something that is at the same time hard to understand and that potentially isn't faster than a naive implementation.

.mocharc.json
yamadapc: I'm not going to fix this one here

.mocharc.json
```diff
@@ -1,5 +1,5 @@
 {
-  "spec": "packages/*/!(integration-tests)/test/*.js",
+  "spec": "packages/*/!(integration-tests)/test/{*.js,**/*.{test,spec}.js}",
```
yamadapc, line 2: There are failing unit-tests on this PR that fail because they were not running.

.mocharc.json
yamadapc: Actually let's resolve it

.mocharc.json
yamadapc: I have fixed it here - https://github.com/parcel-bundler/parcel/pull/9658/commits/33387e0ec7f970550b789ebc3be2cbc3011cb40d

yamadapc:
This test fails on windows ; but only on the JS implementation... 

Screenshot 2024-04-22 at 6 00 53 PM

devongovett:
In a real request that we'd actually want to port to Rust we are going to be calling into JS from a background thread. That involves locking the Rust thread until the JS thread returns each time we call into JS.

Plus, I'd really like to change the way the request tracker works – have it owned by Rust rather than the other way around – and simplify how it's implemented, the API to it, etc. We could certainly continue bolting on some Rust pieces to the existing JS version but I'm concerned it'll create more complexity and technical debt that we have to rip out later.

yamadapc:
> Plus, I'd really like to change the way the request tracker works – have it owned by Rust rather than the other way around – and simplify how it's implemented

I'm keen to reimplement requests so that there's multi-threading throughout, but I think we need to move a bit further with what we're replacing.

> In a real request that we'd actually want to port to Rust we are going to be calling into JS from a background thread. That involves locking the Rust thread until the JS thread returns each time we call into JS.

We can call naively into JS with little downside as far as I can measure.

What problem would be solved by locking a Rust thread but calling into JS from a background thread?

- - - 

What could work is:

* Rewrite parts of the tool incrementally, make sure we test and document the codebase in the process
* Second - Implement request tracking using actors & queues, aim to achieve ideal parallelism, might add some scheduling here & fine tune

In my mind also it'll help to:

* Share little state ; use either logically persistent structures or some CoW strategy . likely when passing structures to worker threads we want to have a logical copy, not a shared instance, to avoid race conditions
* Don't use locks to avoid contention if possible, instead use actors & channels ; can still lock/etc. if that is simpler and better, but generally we can avoid shared state to prevent races, instead send logical copies (for example through ref. counting or similar approaches)

devongovett:
If you want to call an async JS function from a background thread, and you want to get the return value from that JS function on the Rust side, you're gonna need to block until the promise returned by the JS function resolves. In the past I've implemented this with channels: [see here](https://github.com/parcel-bundler/parcel/blob/v2/crates/macros/src/napi.rs). But basically the Rust thread cannot do any other work in the meantime until the promise resolves. You could write everything with a callbacks or some kind of coroutine but it will be harder unless the entire codebase is written that way.

alshdavid:
> What problem would be solved by locking a Rust thread but calling into JS from a background thread?

Essentially, for every 1 JavaScript execution context/worker, napi demands 1 dedicated Rust thread. You cannot pass JavaScript references between threads (with some exceptions) so data is typically handled through channels. If you want to do non blocking JS behaviours, that must be factored into the way you write the napi bindings and is quite verbose.

alshdavid (approved):

PR #9663, opened 2024-04-22
Convert internal asset, dependency, and environment fields to numbers

devongovett:
This converts some fields on our internal Asset, Dependency, and Environment types from strings/booleans to numbers/bitfields. This is mostly in preparation for the conversion to Rust, where we will want more efficient serialization, but in the interim this should also reduce the size of some objects in the JS implementation too. This was started on the Parcel DB branch but should be useful even when using something like Serde too.

Currently, for backward compatibility, some `asset.meta` and `dependency.meta` fields are proxied to internal bitflags transparently to plugins. We may eventually want to expose more official fields for these things but that can be discussed at a later time.


packages/core/core/src/public/Asset.js
```diff
@@ -123,7 +124,55 @@ class BaseAsset {
   }
 
   get meta(): Meta {
-    return this.#asset.value.meta;
+    return new Proxy(this.#asset.value.meta, {
```
yamadapc, line 127: It is possible the use of proxies offsets any improvement this would have as far as runtime is concerned

devongovett: The main goal of this PR is to reduce the size of these data structures, in memory and when serialized, and also align them with how a representation could work in Rust. I don't think there should be much difference in performance in terms of time. This came from the ParcelDB branch where I believe Atlassian saw a fairly significant performance improvement though (likely not because of this in particular but it's a good sign). But if you have a better idea on how to implement this backwards compatibility let me know.

PR #9664, opened 2024-04-22
Optimise the Inline Requires plugin

marcins:
# ↪️ Pull Request

In our large application build, the Inline Requires plugin adds about 25% to the build time (6 min) - so we have looked at some quick optimisation wins. 

While the ideal solution for inline requires would be to implement it in the `ScopeHoistingPackager`, a good time to do that would be when we move the packager to native code rather than working on a complex change to it now.

With a combination of tracing, profiling, and eyeballing (thanks @yamadapc!) we identified a few small performance improvements to the current implementation

* remove potentially expensive object allocation for comparing argument values and replace with a verbose `if` (this saves some GC overhead and provides a 10-15% perf improvement in our tests)
* remove unnecessary usage of an array from an earlier iteration where a boolean would suffice (this saves GC overhead and reduces time spent in the visitor)
* remove `verbose` logging - even when not logging, there was still significant overhead from this due to the volume (another 10-15%)

The main contributor now to this plugin's run time is the SWC `print` of the AST back to JS. We have considered a couple of other potential optimisations (that really still just avoid moving the code into the pacakger where it should be...):
* replace the SWC code with regex matching and replacement - not sure if this would be faster without trying it, but it would complicate source maps for sure.
* move the JS code to Rust - this will avoid serialising/deserialising the AST that SWC has to do with the `parse`/`print` pattern.
* implement minification support in the plugin - this means that we can skip running the SWC minifier, and just have this plugin produce the minified output - saving another `print`/`parse` that plugin has to do.

## 🚨 Test instructions

Verified application still works (unit tests for the visitor pass)

## ✔️ PR Todo

### DIFF
--- a/packages/optimizers/inline-requires/src/InlineRequires.js
+++ b/packages/optimizers/inline-requires/src/InlineRequires.js
@@ -2,27 +2,26 @@
 import {Optimizer} from '@parcel/plugin';
 import {parse, print} from '@swc/core';
 import {RequireInliningVisitor} from './RequireInliningVisitor';
-import type {SideEffectsMap} from './types';
 import nullthrows from 'nullthrows';
 import SourceMap from '@parcel/source-map';
 
-let publicIdToAssetSideEffects = null;
+let assetPublicIdsWithSideEffects = null;
 
 type BundleConfig = {|
-  publicIdToAssetSideEffects: Map<string, SideEffectsMap>,
+  assetPublicIdsWithSideEffects: Set<string>,
 |};
 
 // $FlowFixMe not sure how to anotate the export here to make it work...
 module.exports = new Optimizer<empty, BundleConfig>({
   loadBundleConfig({bundle, bundleGraph, tracer}): BundleConfig {
-    if (publicIdToAssetSideEffects !== null) {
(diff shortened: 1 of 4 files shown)


yamadapc (approved):

marcins:
Thought of another idea while folding laundry, but not sure if it'd make a difference - there's no need to maintain a `Map` of public asset ids -> side effects + filePath. The filePath is only used as a convenience for logging, so this could easily be a `Set` of only public asset ids that have side effects. This turns a map.has + map.get to work out if the asset has side-effects, into just a set.has.

https://github.com/parcel-bundler/parcel/blob/db2984d5ea3d3dbc395d16cdbcf8f427a8054286/packages/optimizers/inline-requires/src/InlineRequires.js#L22

JakeLane (approved):

marcins:
~I want to cut another `dev` release with the asset map -> set change and test it before merging this.. currently blocked by the OpenSSL cross-compile issue though, and I don't want to "revert" the Sentry PR in this branch again just to test.~

Just waiting for a dev release to build now, so I can test this change in CI.

PR #9669, opened 2024-04-24
Update rustfmt

alshdavid:
This PR adds 
```toml
imports_granularity = "Item"
group_imports = "StdExternalCrate"
``` 
to the Rustfmt.

https://rust-lang.github.io/rustfmt/?version=v1.7.0&search=imports_granularity#imports_granularity
https://rust-lang.github.io/rustfmt/?version=v1.7.0&search=#group_imports

This will change Rust imports so that they are split into individual line items that are grouped (in order) as `std`, `extern` and `crate`

example:

```rust
use swc_core::ecma::ast::{Atom, Span};
crate::{Baz, Buzz};
use std::foo::{Foo, Bar};
```

becomes

```rust
use std::foo::Bar;
use std::foo::Foo;

use swc_core::ecma::ast::Atom;
use swc_core::ecma::ast::Span;

use crate::Baz;
use crate::Buzz;
```

The import groupings rule enforces the import grouping convention used by the Rust community whist the import granularity rule makes it easier to read/tidy-up Rust imports and makes it easier to see changes in PRs

Most of the code changes are the formatting rules being applied.

Notable changes in the PR are:
- [package.json](https://github.com/parcel-bundler/parcel/pull/9669/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R24)
- [rustfmt.toml](https://github.com/parcel-bundler/parcel/pull/9669/files#diff-a947c9d2523f34e53b9cdcd6411174a3148f23ea03ad8a61c1aee7b3ea005440R2)

An example case where the applied rule improves readability
image

### DIFF
--- a/crates/json-comments-rs/src/lib.rs
+++ b/crates/json-comments-rs/src/lib.rs
@@ -389,9 +389,10 @@ fn in_line_comment(c: &mut u8) -> State {
 
 #[cfg(test)]
 mod tests {
-  use super::*;
   use std::io::ErrorKind;
   use std::io::Read;
+
+  use super::*;
 
   fn strip_string(input: &str) -> String {
     let mut out = String::new();
--- a/crates/json-comments-rs/src/lib.rs
+++ b/crates/json-comments-rs/src/lib.rs
@@ -41,10 +41,10 @@
 //! # }
 //! ```
 //!
-use std::{
-  io::{ErrorKind, Read, Result},
-  slice::IterMut,
-};
+use std::io::ErrorKind;
+use std::io::Read;
+use std::io::Result;
+use std::slice::IterMut;
 
 #[derive(Eq, PartialEq, Copy, Clone, Debug)]
 enum State {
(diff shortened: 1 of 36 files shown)


yamadapc (approved):
I don't think we should change default formatting settings.

Most rust code that exists will use the default settings. This is great because it means a lot of rust code has consistent style.

For example, the following projects do not change these settings

* rust standard library https://github.com/rust-lang/rust/blob/master/rustfmt.toml
* linux https://github.com/torvalds/linux/blob/e67572cd2204894179d89bd7b984072f19313b03/rust/alloc/boxed.rs
* tokio asynchronous runtime https://github.com/tokio-rs/tokio/

alshdavid:
> This is great because it means a lot of rust code has consistent style.

I agree in general however in this case - with the import style - we have inconsistent usage throughout the codebase. 

I believe that the addition of these style rules will improve code consistency within our codebase in a way that remains congruent with the styling and conventions used broadly by the Rust community.

MonicaOlejniczak (approved):

mischnic:
Apart from causing loads of merge conflicts, this is also not being enforced with the current setup 
```
Warning: can't set `imports_granularity = Item`, unstable features are only available in nightly channel.
Warning: can't set `group_imports = StdExternalCrate`, unstable features are only available in nightly channel.
```
because we have it set to stable
https://github.com/parcel-bundler/parcel/blob/3ff10bce2d244905cd53d860bccf571d759cb145/rust-toolchain#L1

PR #9679, opened 2024-05-01
Enable producing native debug symbols in canary/dev releases

marcins:
# ↪️ Pull Request

Added `debug = true` to the canary release profile, this produces binaries with symbol debug information included. Existing `strip` operations take care of stripping these back out for the `@parcel/rust` package itself, however before we `strip` we now use platform specific tools to extract debug symbols out to a separate file / artifact.

At "build and release" time these are combined into a single `debug-symbols` artifact for ease of use. Any additional work is gated to only the dev/canary releases - there's a minor impact to build time, especially uploading the combined artifact, but it's pretty minor.

This has no impact on final `@parcel/rust` binary size (as the debug info is stripped), and it will be on our end when we bump to a canary to download  the debug symbols artifact and upload it to Sentry.

Example dev-release with these changes: https://github.com/parcel-bundler/parcel/actions/runs/8902598798

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -273,11 +273,24 @@ jobs:
         uses: actions/download-artifact@v3
         with:
           path: artifacts
-      - name: Move artifacts
-        run: for d in artifacts/*/*/*; do cp $d/*.node packages/$(basename $(dirname $d))/$(basename $d); done
+      - name: Move bindings
+        run: for d in artifacts/bindings-*/*/*; do cp $d/*.node packages/$(basename $(dirname $d))/$(basename $d); done
+      - name: Move debug symbols
+        run: |
+          mkdir debug-symbols
+          find artifacts -name "*.debug" -exec cp {} debug-symbols/ \;
+          find artifacts -name "*.node" -path "**/DWARF" -exec cp {} debug-symbols/ \;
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -35,7 +35,7 @@ jobs:
             os: macos-latest
             target: aarch64-apple-darwin
 
-          - name: macos-latest
+          - name: x86_64-apple-darwin
```
marcins, line 38: We talked about this briefly in sync this morning.. maybe it doesn't belong in this PR, but hopefully it doesn't conflict.

mattcompiles (approved):

yamadapc (approved):

marcins:
Just leaving some notes here from some "offline" conversations:
* Github has a default 90 day retention for artifacts - if the size of the new artifacts proves problematic, we have a few options: 
 * remove the "convenience" single symbols artifact, just use the separate ones (which will complicate our consumption side slightly)
 * change the retention for the separate artifacts to 1 day, or work out if we can delete them after the combined artifact is produced
 * look into whether we can publish these directly or elsewhere

yamadapc:
I'm removing this from the merge queue because https://github.com/parcel-bundler/parcel/pull/9658 needs to be merged first.

PR #9680, opened 2024-05-01
Allow actions to be triggered by merge queue

benjervis:
In order to use merge queues, GitHub Actions needs to be able to re-run CI in the background to make sure that things are ok to merge, which means the actions need to be triggerable on `merge_group`.

Have added this trigger to the non-manual actions we have (which is just `ci` and `canary-release`).

### DIFF
--- a/.github/workflows/canary-release.yml
+++ b/.github/workflows/canary-release.yml
@@ -4,7 +4,6 @@
 name: canary-release
 
 on:
-  merge_group:
   push:
     branches:
       - v2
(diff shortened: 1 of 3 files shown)


mattcompiles (approved):

.github/workflows/canary-release.yml
```diff
@@ -4,6 +4,7 @@
 name: canary-release
 
 on:
+  merge_group:
```
yamadapc: You don't want this here

yamadapc:
Let's use https://github.com/parcel-bundler/parcel/pull/9658

PR #9681, opened 2024-05-01
Move crates around and add a start of AssetRequest, EntryRequest and non-workable AssetGraphRequest

yamadapc:
This PR makes two changes on top of https://github.com/parcel-bundler/parcel/pull/9658

1. Add asset request, entry request and asset graph request - boilerplate ; non-tested properly yet
2. Split parcel-filesystem and parcel-napi-helpers into separate crates
3. I still haven't split core because I'd like to get this diff merged first


yamadapc:
Let me split this PR

PR #9683, opened 2024-05-01
Add rust entry request

yamadapc:
* Adds rust entry-request implementation
* Adds `ProjectPath` newtype and uses it in `ConfigRequest`
* Still pending integrate to JavaScript and write "EntryRequest" tests


crates/node-bindings/src/core/js_requests/request_options.rs
```diff
@@ -0,0 +1,43 @@
...
+
+pub fn input_fs_from_options(
+  env: Rc<Env>,
+  options: &JsObject,
+) -> napi::Result<JSDelegateFileSystem> {
+  let Some(input_fs) = options
+    .get("inputFS")?
+    .map(|input_fs| JSDelegateFileSystem::new(env, input_fs))
+  else {
+    // We need to make the `FileSystem` trait object-safe, so we can use dynamic
```
yamadapc, line 36: Note here;

At the moment `FileSystem` has a few generics which make it non object safe.

We want to simplify it so that we can make it object-safe. Making it object safe will allow us to pass around a `Box<dyn FileSystem>`, which would do dynamic dispatch.

This is good because we can have some fallbacks. For example, if JavaScript doesn't give us a filesystem, it is a sensible option to use the `OsFileSystem` implementation, since that's eventually what we will want, rather than JavaScript providing us a real `FileSystem`.

crates/node-bindings/src/core/project_path.rs
```diff
@@ -0,0 +1,76 @@
+use std::path::Path;
+use std::path::PathBuf;
+
+use napi::bindgen_prelude::FromNapiValue;
+use napi::bindgen_prelude::ToNapiValue;
+use napi::sys::napi_env;
+use napi::sys::napi_value;
+
+/// Similar to opaque type ProjectPath in Rust.
```
yamadapc, line 9: This file demonstrates how to new-type wrap values if needed.

Generally I'd avoid doing this.

crates/node-bindings/src/core/requests/entry_request/mod.rs
```diff
@@ -0,0 +1,237 @@
+//! Entry request corresponds to `EntryRequest.js` and is used to resolve entry points.
```
yamadapc, line 1: There are a few missing things here.

alshdavid (approved):

PR #9690, opened 2024-05-02
Retry integration tests 2 times on CI

yamadapc:
As raised in #9689, integration tests are currently flaky. This makes the use of a merge queue harder since with merge queues individual builds can't be retried.

As a temporary work-around this commit adds retries to the integration tests' mocha invocation.

This is very subpar and we must come-up with a plan to address the tests' unreliability.


marcins (approved):

mattcompiles:
@yamadapc I think we already do this via the config. We added it the other day. Not sure if it really helped though. 
https://github.com/parcel-bundler/parcel/blob/3e6a58d21a33f4b59831aea95a32a85808baa6df/packages/core/integration-tests/.mocharc.json#L7

PR #9693, opened 2024-05-03
Fix windows stack overflow with larger stack red_zone

jondlm:
Fixing the stack overflow issues with windows on the `bump-swc` PR #9574.


mischnic:
Thank you for taking a look, somehow Windows is still failing on CI though?

jondlm:
Looks like it's passing now with a larger red zone of 32 KiB.

PR #9694, opened 2024-05-04
Fix windows build

jondlm:
# ↪️ Pull Request

Fixes the windows build by using cross platform [`shx` alternatives](https://github.com/shelljs/shx) to plain commands like `cp` that were failing CI. 

## 💻 Examples

From my windows box:
image

## 🚨 Test instructions

`yarn build` on a windows machine


yamadapc:
Could you use an unix shell instead of having these changes?

I suppose this worked on CI previously because it has `cp` and `rm` on windows (unix like bash shell)?

jondlm:
Huh yeah, I thought I was seeing failures on CI related to this but I think I misread them. I was only seeing them locally. Yeah I can use an alternate shell that has them available. I'll close this.

PR #9696, opened 2024-05-05
Pin rust to v1.77

yamadapc:
We can revert this change when napi-rs/napi-rs#2083 is merged and a new `napi` version is released.


rust-toolchain
```diff
@@ -1 +1 @@
-stable
+1.77.0
```
marcins: The most recent working releases were built with 1.77.2 - will this use the latest 1.77 patch?

yamadapc: No; I've changed it to 1.77

mattcompiles (approved):

marcins (approved):

PR #9699, opened 2024-05-06
Bump napi-rs

mischnic:
Bump napi-rs to include the fix for https://github.com/napi-rs/napi-rs/issues/2085 ~~and revert Rust version change (#9696,  #9697) again~~

- [x] It's still not fixed completely in napi-rs
- [x] Waiting for napi-rs release

----

Rust 1.77 enabled dev assertions in the stdlib, which makes it more strict in dev builds.
There was a bug in napi-rs that was previously unnoticed because there were no assertions:

```
thread '<unnamed>' panicked at library/core/src/panicking.rs:156:5:
unsafe precondition(s) violated: NonNull::new_unchecked requires that the pointer is non-null
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
thread caused non-unwinding panic. aborting.
```


.github/workflows/ci.yml
```diff
@@ -21,7 +21,7 @@ jobs:
           cache: yarn
       - uses: dtolnay/rust-toolchain@master
         with:
-          toolchain: 1.77
+          toolchain: stable
```
marcins: I'm not sure if we should stay on a pinned release of Rust, and update it manually? Or just stay on `stable` and pin only when it breaks?

mischnic: > Or just stay on stable and pin only when it breaks?

Yes, that's what I would have suggested (= what we've done so far). Requires less maintenance from our side as well

marcins:
Do you have any clues as to why integration tests are still failing? I cloned this branch locally and built it and it seems to work okay for a quick smoke test (building a smaller part of our application as a production build locally). I've double checked and it's build with Rust 1.78 and running with Node v18.20.1 (on an M1 Mac)

mischnic:
My only guess is that it's the same stack overflow as https://github.com/parcel-bundler/parcel/pull/9574

But we'll see what happens when rebasing this after merging the swc bump.

mischnic:
Now it's again crashing only on Windows, so probably the same stack problem as with the swc PR

marcins:
It looks like everything is working again now - likely as you suspected, the SWC stack changes might've fixed this.

Are there any blockers to taking this out of Draft and getting it shipped? For context, keen on being able to get a Rust 1.78 based build as the `RwLock` implementation has changed to a Rust based one from the existing pthread one - keen to see if this fixes the issues we see internally with lock based crashes in the resolver coming from the macOS implementation.

EDIT: I pulled in latest v2 changes and still getting Windows integration test failures with `error Command failed with exit code 3221225725.`

mischnic:
https://github.com/parcel-bundler/parcel/pull/9758 has already upgraded napi in Cargo.lock

PR #9700, opened 2024-05-07
Add parcel config crate

MonicaOlejniczak:
# ↪️ Pull Request

This change introduces a new `parcel_config` crate that ports some of the logic in https://github.com/parcel-bundler/parcel/compare/v2...core-rs2 and ultimately expands on the implementation there.
* Parcel config handles .parcelrc file loading, config merging, and performs minor validation
* Added a basic `package_manager` crate and trait to add as a dependency of the config

## 🚨 Test instructions

`cargo test -p parcel_config`

### DIFF
--- a/crates/parcel_config/src/partial_parcel_config.rs
+++ b/crates/parcel_config/src/partial_parcel_config.rs
@@ -1,0 +1,692 @@
+use std::collections::HashSet;
+use std::rc::Rc;
+
+use derive_builder::Builder;
+use indexmap::IndexMap;
+
+use super::parcel_config::PluginNode;
+use super::parcel_rc::ParcelRcFile;
+use crate::config_error::ConfigError;
+
+/// An intermediate representation of the .parcelrc config
+///
+/// This data structure is used to perform configuration merging, to eventually create a compelete ParcelConfig.
+///
+#[derive(Clone, Debug, Default, Builder, PartialEq)]
+#[builder(default)]
+pub struct PartialParcelConfig {
+  pub bundler: Option<PluginNode>,
+  pub compressors: IndexMap<String, Vec<PluginNode>>,
+  pub namers: Vec<PluginNode>,
+  pub optimizers: IndexMap<String, Vec<PluginNode>>,
+  pub packagers: IndexMap<String, PluginNode>,
+  pub reporters: Vec<PluginNode>,
+  pub resolvers: Vec<PluginNode>,
+  pub runtimes: Vec<PluginNode>,
+  pub transformers: IndexMap<String, Vec<PluginNode>>,
+  pub validators: IndexMap<String, Vec<PluginNode>>,
+}
+
+impl TryFrom<&ParcelRcFile> for PartialParcelConfig {
+  type Error = ConfigError;
+
+  fn try_from(parcel_rc: &ParcelRcFile) -> Result<PartialParcelConfig, ConfigError> {
+    // TODO Add validation here: multiple ..., plugin name format, reserved pipelines, etc
+
+    let resolve_from = Rc::new(parcel_rc.path.clone());
+
+    let to_entry = |package_name: &String| PluginNode {
+      package_name: String::from(package_name),
+      resolve_from: Rc::clone(&resolve_from),
+    };
+
+    let to_vec = |maybe_plugins: Option<&Vec<String>>| {
+      maybe_plugins
+        .map(|plugins| plugins.iter().map(to_entry).collect())
+        .unwrap_or(Vec::new())
+    };
+
+    let to_pipelines = |map: Option<&IndexMap<String, Vec<String>>>| {
+      map
+        .map(|plugins| {
+          plugins
+            .iter()
+            .map(|(pattern, plugins)| {
+              (
+                String::from(pattern),
+                plugins.iter().map(to_entry).collect(),
+              )
+            })
+            .collect()
+        })
+        .unwrap_or(IndexMap::new())
+    };
+
+    let to_pipeline = |map: Option<&IndexMap<String, String>>| {
+      map
+        .map(|plugins| {
+          plugins
+            .iter()
+            .map(|(pattern, package_name)| (String::from(pattern), to_entry(package_name)))
+            .collect()
+        })
+        .unwrap_or(IndexMap::new())
+    };
+
+    Ok(PartialParcelConfig {
+      bundler: parcel_rc
+        .contents
+        .bundler
+        .as_ref()
+        .map(|package_name| PluginNode {
+          package_name: String::from(package_name),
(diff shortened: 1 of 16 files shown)


crates/parcel_filesystem/src/lib.rs
```diff
@@ -43,6 +43,34 @@ pub trait FileSystem {
   ) -> Result<PathBuf> {
     self.canonicalize_base(path)
   }
+  fn find_ancestor_file<P: AsRef<Path>>(
```
yamadapc: I reckon this should be a function that receives filesystem, instead of a trait function.

So 
```rust
  fn find_ancestor_file(
    fs: &impl FileSystem,
    filenames: Vec<String>,
    from: impl AsRef<Path>,
    root: impl AsRef<Path>,
  ) -> Option<PathBuf> {
    for dir in from.as_ref().ancestors() {
      // Break if we hit a node_modules directory
      if let Some(filename) = dir.file_name() {
        if filename == "node_modules" {
          break;
        }
      }

      for name in &filenames {
        let fullpath = dir.join(name);
        if fs.is_file(&fullpath) {
          return Some(fullpath);
        }
      }

      if dir == root.as_ref() {
        break;
      }
    }

    None
  }
```

crates/package_manager/Cargo.toml
```diff
@@ -0,0 +1,9 @@
+[package]
+authors = ["Monica Olejniczak <monica.j.olejniczak@gmail.com>"]
+name = "package_manager"
```
yamadapc: nit: we probably want to keep everything prefixed with `parcel_`

crates/parcel_filesystem/src/lib.rs
MonicaOlejniczak: Which crate would you have this fn live in then? Feels like it should be outside of filesystem if it takes it in as a a param imo

crates/parcel_config/src/parcel_rc_config.rs
```diff
@@ -0,0 +1,781 @@
...
+
+#[derive(Default)]
+pub struct LoadConfigOptions<'a> {
+  pub additional_reporters: Vec<PluginNode>,
+  pub config: Option<&'a str>,
+  pub fallback_config: Option<&'a str>,
+}
+
+/// Loads and validates .parcel_rc config
+pub struct ParcelRcConfig<'a, T, U> {
```
yamadapc: I suppose this is a "parcelrc config loader"?

crates/parcel_config/src/parcel_rc_config.rs
```diff
@@ -0,0 +1,781 @@
...
+    let dir = if is_cwd_inside_project_root {
+      &cwd
+    } else {
+      project_root
+    };
+
+    dir.join("index")
+  }
+
+  fn load_config(&self, path: PathBuf) -> Result<(PartialParcelConfig, Vec<PathBuf>), ConfigError> {
```
yamadapc: Would you mind documenting this module ; what it does, how it traverses directories and merges configuration

crates/parcel_config/src/parcel_rc_config.rs
```diff
@@ -0,0 +1,781 @@
...
+        .map_err(|source| ConfigError::UnresolvedConfig {
+          config_type: String::from("config"),
+          from: resolve_from.clone(),
+          source: Box::new(source),
+          specifier: String::from(config),
+        }),
+      None => self.find_config(project_root, &resolve_from),
+    };
+
+    let mut _used_fallback = false;
```
yamadapc: nit: unused variable; should remove

crates/parcel_config/src/partial_parcel_config.rs
```diff
@@ -0,0 +1,693 @@
...
+                ))
+                .build()
+                .unwrap()
+            );
+          }
+        }
+      };
+    }
+
+    test_pipeline_map!(compressors);
```
yamadapc, line 683: I'd personally write helper functions rather than macro it unless not possible because macros mess-up a lot of IDE features

crates/parcel_config/src/pipeline.rs
```diff
@@ -0,0 +1,310 @@
...
+pub struct PipelineMap {
+  map: IndexMap<String, Vec<PluginNode>>,
+}
+
+impl PipelineMap {
+  pub fn new(map: IndexMap<String, Vec<PluginNode>>) -> Self {
+    Self { map }
+  }
+
+  pub fn get(&self, path: &Path, pipeline: &Option<impl AsRef<str>>) -> Vec<PluginNode> {
```
yamadapc: imo you could document what this structure does

crates/parcel_filesystem/src/lib.rs
yamadapc: I'd leave this where you're consuming it if it's not shared yet.

Not sure why it breaks on node_modules

yamadapc (approved):
Thanks, the PR looks good. 

I left some comments regarding documentation and nits.

crates/parcel_config/src/partial_parcel_config.rs
MonicaOlejniczak: Yeah I get that, but there's far too much repetition / not enough coverage without the use of macros.

crates/parcel_config/src/parcel_rc_config.rs
MonicaOlejniczak: It's in the old impl so I left it there, but feel it the logic can be improved later once that piece gets used. But I can remove it in the meantime

crates/parcel_config/src/parcel_rc_config.rs
MonicaOlejniczak: Yeah, I can rename it

crates/package_manager/Cargo.toml
MonicaOlejniczak: Sure

crates/parcel_filesystem/src/lib.rs
MonicaOlejniczak: Ok, it's only used in package_manager and config within the JS code atm

crates/parcel_config/src/parcel_rc_config.rs
MonicaOlejniczak: Sure

crates/parcel_config/src/pipeline.rs
MonicaOlejniczak: Will do

crates/parcel_filesystem/src/lib.rs
yamadapc: If it's used on two other crates you may want to just export from this one

PR #9709, opened 2024-05-09
Handle invalid clockspec error and log more when loading cache fails

Nikola-3:
# ↪️ Pull Request

## ✔️ PR Todo

### DIFF
--- a/packages/core/core/src/RequestTracker.js
+++ b/packages/core/core/src/RequestTracker.js
@@ -1598,29 +1598,7 @@ async function loadRequestGraph(options): Async<RequestGraph> {
     } catch (e) {
       // Prevent logging fs events took too long warning
       clearTimeout(timeout);
-
-      if (e.message && e.message.includes('invalid clockspec')) {
-        const snapshotContents = options.inputFS.readFileSync(
-          snapshotPath,
-          'utf-8',
-        );
-        logger.warn({
-          origin: '@parcel/core',
-          message: `Error reading clockspec from snapshot, building with clean cache.`,
-          meta: {
-            snapshotContents: snapshotContents,
-            trackableEvent: 'invalid_clockspec_error',
-          },
-        });
-      } else if (!(e instanceof FSBailoutError)) {
-        logger.warn({
-          origin: '@parcel/core',
-          message: `Unexpected error loading cache from disk, building with clean cache.`,
-          meta: {
-            trackableEvent: 'cache_load_error',
-          },
-        });
-      }
+      logErrorOnBailout(options, snapshotPath, e);
       // This error means respondToFSEvents timed out handling the invalidation events
       // In this case we'll return a fresh RequestGraph
       return new RequestGraph();
@@ -1628,4 +1606,32 @@ async function loadRequestGraph(options): Async<RequestGraph> {
   }
 
   return new RequestGraph();
+}
+function logErrorOnBailout(
+  options: ParcelOptions,
+  snapshotPath: string,
+  e: Error,
+): void {
+  if (e.message && e.message.includes('invalid clockspec')) {
+    const snapshotContents = options.inputFS.readFileSync(
+      snapshotPath,
+      'utf-8',
+    );
+    logger.warn({
+      origin: '@parcel/core',
+      message: `Error reading clockspec from snapshot, building with clean cache.`,
+      meta: {
+        snapshotContents: snapshotContents,
+        trackableEvent: 'invalid_clockspec_error',
+      },
+    });
+  } else if (!(e instanceof FSBailoutError)) {
+    logger.warn({
+      origin: '@parcel/core',
+      message: `Unexpected error loading cache from disk, building with clean cache.`,
+      meta: {
+        trackableEvent: 'cache_load_error',
+      },
+    });
+  }
 }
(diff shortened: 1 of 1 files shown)


packages/core/core/src/RequestTracker.js
```diff
@@ -65,6 +66,10 @@ export const requestGraphEdgeTypes = {
   dirname: 7,
 };
 
+class FSBailoutError extends Error {
+  name: string = 'FSBailoutError';
```
yamadapc, line 69: afaik you don't need to set `name` ;

packages/core/core/src/RequestTracker.js
```diff
@@ -1543,55 +1548,94 @@ async function loadRequestGraph(options): Async<RequestGraph> {
...
     } catch (e) {
+      if (timeout != null) {
+        clearTimeout(timeout);
+      }
+      let additionalMessage;
+      let logEvent;
+      if (e.message && e.message.includes('invalid clockspec')) {
+        // Note this assumes only one snapshot file per cache and will potentially break
+        // if multiple snapshot files exist in the cache directory
+        const snapshotFile = readdirSync(options.cacheDir).find(file =>
```
yamadapc: you probably want to log something before doing this because it can also fail

packages/core/core/src/RequestTracker.js
```diff
@@ -1543,55 +1548,94 @@ async function loadRequestGraph(options): Async<RequestGraph> {
...
+          },
+        };
+        if (snapshotFile != null) {
+          additionalMessage = readFileSync(
+            path.join(options.cacheDir, snapshotFile),
+          ).toString('utf-8');
+          logEvent.meta.additionalMessage = `Watchman failed reading clockspec, snapshot: ${additionalMessage}`;
+        }
+      }
+      if (!(e instanceof FSBailoutError)) {
```
yamadapc: I suppose we also want to log that we bailed out?

packages/core/core/src/RequestTracker.js
```diff
@@ -1543,55 +1548,94 @@ async function loadRequestGraph(options): Async<RequestGraph> {
...
+        logEvent = {
+          origin: '@parcel/core',
+          message: `Error reading clockspec from snapshot, building with clean cache.`,
+          meta: {
+            additionalMessage: ``,
+            trackableEvent: 'invalid_clockspec_error',
+          },
+        };
+        if (snapshotFile != null) {
+          additionalMessage = readFileSync(
```
yamadapc: why do you need to find the snapshot file? isn't it always written to the same location?

packages/core/core/src/RequestTracker.js
```diff
@@ -1543,55 +1548,94 @@ async function loadRequestGraph(options): Async<RequestGraph> {
...
         options.unstableFileInvalidations || events,
         options,
         10000,
       );
       return requestGraph;
     } catch (e) {
+      if (timeout != null) {
+        clearTimeout(timeout);
+      }
+      let additionalMessage;
```
yamadapc: you can extract this to a function but maybe it can be simpler (avoid the guessing snapshot path part & just log more lines rather than assigning to a variable etc.)

packages/core/core/src/RequestTracker.js
Nikola-3: It's written to `snapshot-<hash>.txt` but I don't know if the hash persists. I guess I can look through the code and see if the files location is available to be passed to `loadRequestGraph`

packages/core/core/src/RequestTracker.js
Nikola-3: That happens elsewhere, hence the FSBailoutError addition

packages/core/core/src/RequestTracker.js
Nikola-3: I guess it is redundant

packages/core/core/src/RequestTracker.js
yamadapc: fair enough ; okay, in that case you could look for `snapshot-*.txt` rather than any txt file?

yamadapc (approved):
please consider extracting the new logging block into a function

packages/core/core/src/RequestTracker.js
```diff
@@ -1543,55 +1548,84 @@ async function loadRequestGraph(options): Async<RequestGraph> {
...
+      if (e.message && e.message.includes('invalid clockspec')) {
+        logEvent = {
+          origin: '@parcel/core',
+          message: `Error reading clockspec from snapshot, building with clean cache.`,
+          meta: {
+            additionalMessage: ``,
+            trackableEvent: 'invalid_clockspec_error',
+          },
+        };
+        additionalMessage = readFileSync(snapshotPath).toString('utf-8');
```
mattcompiles: You shouldn't used `readFileSync` from `fs` directly, you should use `options.inputFS.readFileSync`

packages/core/core/src/RequestTracker.js
```diff
@@ -1543,55 +1548,84 @@ async function loadRequestGraph(options): Async<RequestGraph> {
...
+        logEvent.meta.additionalMessage = `Watchman failed reading clockspec, snapshot: ${additionalMessage}`;
```
mattcompiles: Why not create this before the `logEvent`? Feels messy doing it after.

mattcompiles (approved):
Love your work 🍻

yamadapc (approved):

PR #9714, opened 2024-05-14
Add core rust types

MonicaOlejniczak:
# ↪️ Pull Request

Adds the types from https://github.com/parcel-bundler/parcel/compare/v2...core-rs2 into the core crate and splits up some of the files into logical boundaries

**Notes:**
- Bitflags have been removed from the implementation
- These types are documented using information from the public website and the repository
- Does not integrate with napi, this can be done later as needed

## 🚨 Test instructions

`cargo test -p parcel_core`


crates/parcel_core/src/types.rs
alshdavid, line 1: Currently this file exports members so they are behave like this style of JS export
```javascript
export * as asset from './asset'
// etc
```

You could combine these reexports to be the equivalent of
```javascript
export * from './asset'
```

By writing them like this
```rust
mod asset;
// etc

pub use self::asset::*;
// etc 
```

MonicaOlejniczak: Yeah I can flatten it out but might still want namespaces for nested modules

alshdavid: Fair enough

alshdavid (approved):

mattcompiles (approved):

devongovett:
> Bitflags have been removed from the implementation

Why? I already have another PR open to use these in the JS implementation too.

Issue #9345, opened 2023-10-25
`parcel [serve] --https` does not use the port specified in `.env` properly


rayinaway:
# 🐛 bug report

With the `--https` flag set, if the development server port is set via a _`.env`_ file, it gets picked for serving/accessing static assets, but not for the HMR connection. Or at least the browser tries to connect to the default port and fails. As a result, HMR doesn't work.

Works fine if the port is specified in the CLI parameter. I haven't performed a particularly deep dive into how Parcel works with HTTPS and ports, but at first glance, it seems like the value just gets lost somewhere on the way. If this is the intended behavior, my bad! But I'd say it's a bit confusing and could use some documenting.

## 🎛 Configuration

All defaults.

## 🤔 Expected Behavior

HMR works, no errors.

## 😯 Current Behavior

HMR does not work, the following error in the console:

```
runtime-15d8632956e6e29d.js:86 WebSocket connection to 'wss://localhost:1234/' failed: 
9p9IN @ runtime-15d8632956e6e29d.js:86
newRequire @ index.63aff760.js:71
(anonymous) @ index.63aff760.js:122
(anonymous) @ index.63aff760.js:145
runtime-15d8632956e6e29d.js:174 [parcel] 🚨 Connection to the HMR server was lost
ws.onclose @ runtime-15d8632956e6e29d.js:174
```

## 💁 Possible Solution

If there is a separate configuration for the HTTPS/WSS servers, it could be that the port value doesn't get passed everywhere it should.

## 💻 Code Sample

https://github.com/rayinaway/https-hmr-parcel-bug-repro

There are 3 package scripts there: `serve` serves over HTTP at the specified port; `serve-https` demonstrates the issue; `serve-https-port` serves over HTTPS at the port specified in the CLI parameter.

## 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Parcel           | 2.10.0
| Node             | 21.0.0
| npm         | 10.2.0
| Operating System | Windows 10

rayinaway:
Did some digging. As far as I understand, there are 2 possible issues, and resolving any one of them should be enough to fix the thing.

[...]

PR #9715, opened 2024-05-14
quick fix for HTTPS port from env not working properly by allowing env `HMR_PORT` alongside env `PORT`

700software:
# ↪️ Pull Request
quick fix for #9345

However, I do not like the fact that `normalizeOptions` is a) responsible for invoking `getPort` as a fallback & b) unaware of `env.PORT`.

This will at least solve the immediate issue #9345 and maybe a better refactor of the sequence of these checks can be done later.

## 💻 Examples

Steps to repro: (from `v2` branch)

run two parcel projects, both with `PORT=` and `HMR_PORT=` in the `.env` **and** the `--https` option (no `--port` option)
the two parcel projects you start up should have different non-conflicting port numbers in their `.env`

You will notice that the 2nd one givs an error about port 1234 being taken despite the fact that you have `PORT` specified in env. The reason for this is the first project, while it respects `PORT` for the https server, is silently running and HMR server on port 1234.

So what you will find is the 2nd parcel project you start up is picking a port number randomly with `getPort`.

However, if you use this PR, you will see that you can run multiple `--https` parcel projects with no port conflicts by relying only on `.env` configuration and no need to set `--port` (which in our case we are trying to avoid)

## ✔️ PR Todo

### DIFF
--- a/packages/core/core/src/resolveOptions.js
+++ b/packages/core/core/src/resolveOptions.js
@@ -146,6 +146,7 @@ export default async function resolveOptions(
   };
 
   let port = determinePort(initialOptions.serveOptions, env.PORT);
+  let hmrPort = determinePort(initialOptions.hmrOptions, env.HMR_PORT);
 
   return {
     config: getRelativeConfigSpecifier(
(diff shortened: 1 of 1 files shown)


packages/core/core/src/resolveOptions.js
```diff
@@ -146,6 +146,7 @@ export default async function resolveOptions(
   };
 
   let port = determinePort(initialOptions.serveOptions, env.PORT);
+  let hmrPort = determinePort(initialOptions.hmrOptions, env.HMR_PORT);
```
benedictjohannes, line 149: Hi, I suggest using `env.HMR_PORT?.length ? env.HMR_PORT : env.PORT`. As such, when only setting `env.PORT`, it would allow for HMR to use the same port when `env.HMR_PORT` isn't set. This is aligned with the behavior of the CLI, where `--port` sets both the main and HMR port.

700software: Good idea.

benedictjohannes: Hi @700software , are you interested to implement my suggestion to your PR?

PR #9724, opened 2024-05-20
Un-pin Rust from 1.77 back to stable

marcins:
# ↪️ Pull Request

We had pinned the Rust version at 1.77 due to issues with 1.78 and Parcel. With the SWC bump now merged it seems that Parcel is stable with Rust 1.78, so this change reverts the "pinning".

Rather than just pin to 1.78 we've gone with `stable` as we had before - usually Rust bumps don't cause issues, so it's probably generally fine to stay on latest stable rust and only pin if there's an issue found.

## 🚨 Test instructions

Tested a dev build internally against our very large web application.

## ✔️ PR Todo


JakeLane (approved):

yamadapc (approved):

mischnic:
So you're saying that the SWC bump fixed the Rust version problem. And the NAPI bump only made it worse?

marcins:
> So you're saying that the SWC bump fixed the Rust version problem. And the NAPI bump only made it worse?

It seems the SWC bump fixes issues with using Rust 1.78, though if it's just the Windows integration tests being flaky, then I've seen that with all combinations lately so I don't think anything has fixed those reliably.

PR #9726, opened 2024-05-20
Implement transactional large blob cache writes

yamadapc:
This PR adds unit-tests for FS/LMDB caches and implements transactional writes. This prevents a large blob write from failing partially or from broken entries from being read.

The implementation of this is in `LMDBCache.js`. Large blobs are stored through an entity `LargeBlobEntry`, which contains a randomly generated key for the `FSCache` large blob entries used. Since `FSCache` large blob keys are generated through this mechanism, it's not possible for a partially written entry to be read. Furthermore, it's not possible for the entry to become corrupted.

In addition to this, changes were made to the `RequestTracker` cache writing.

Previously, `parcel-query` worked by listing directory files under the cache directory and finding the latest large blob keys on disk. This was very coupled with the cache implementation in use. To improve this and fix breakage caused by my changes, a new `RequestCacheInfo` entry is written with the latest request graph cache key.

`parcel-query` is changed to read this entry, then scan the request graph nodes to find the asset and bundle graph cache entries.

Some `parcel-query` tests were patched as it was found that they only previously worked due to stale `.parcel-cache` directories being present (e.g.: on v2 on a clean workspace, running only the `parcel-query` integration tests will fail; they only pass in CI because other tests populate a directory).


yamadapc:
This breaks parcel-query because it directly reads the cache filesystem and assumes it's in a certain state.

JakeLane (approved):

PR #9727, opened 2024-05-20
Public rust crate for Parcel API and moving napi bindings to node_bindings

alshdavid:
In this PR I move the napi bindings out of the `parcel_core` package, the idea is the `parcel_core` package will be agnostic to the caller context where context specific behaviour would be consumed by the core abstracted behind traits (like `Resolver`, etc).

# Test instructions
```
yarn build-native-release
```


MonicaOlejniczak (approved):

crates/parcel_core/src/lib.rs
alshdavid, line 1: So the idea is:
```
cargo add parcel_core
```
Which is eventually used something like
```rust
use parcel_core::Parcel;
use parcel_core::ParcelOptions;
use parcel_core::ParcelRunOptions;

fn main() {
  let parcel = Parcel::new(ParcelOptions { /* ... */ });

  let report = parcel.run(ParcelRunOptions {  /* ... */  });
  println!("{:#?}", report);
}
```

For Napi bindings:
```rust
use parcel_core::Parcel;
use parcel_core::ParcelOptions;
use parcel_core::ParcelRunOptions;
use parcel_napi::ParcelNapiConnector;

#[napi]
fn main() {
  let parcel_napi = ParcelNapiConnector::new();
  // napi glue code

  let parcel = Parcel::new(ParcelOptions { 
    plugin_connectors: vec![parcel_napi],
  });

  let report = parcel.run(ParcelRunOptions {  /* ... */  });
  println!("{:#?}", report);
}
```

PR #9735, opened 2024-05-23
Exit process with error code when reporters fail

MonicaOlejniczak:
# ↪️ Pull Request

This change modifies the process exit code when a reporter fails, ensuring that builds do not pass when something goes wrong in a reporter (e.g. generating a file).

Adding this exit code is the simplest mechanism for failing the build, as it does not require changing the existing API or introducing new APIs.

## 🚨 Test instructions

`yarn test:integration`

### DIFF
--- a/packages/core/core/src/ReporterRunner.js
+++ b/packages/core/core/src/ReporterRunner.js
@@ -113,6 +113,7 @@ export default class ReporterRunner {
           });
         } catch (reportError) {
           INTERNAL_ORIGINAL_CONSOLE.error(reportError);
+          process.exitCode = 1;
         } finally {
           measurement && measurement.end();
         }
--- a/packages/core/core/src/ReporterRunner.js
+++ b/packages/core/core/src/ReporterRunner.js
@@ -12,7 +12,6 @@ import {
   NamedBundle,
 } from './public/Bundle';
 import WorkerFarm, {bus} from '@parcel/workers';
-import ParcelConfig from './ParcelConfig';
 import logger, {
   patchConsole,
   unpatchConsole,
@@ -24,8 +23,8 @@ import BundleGraph from './BundleGraph';
 import {tracer, PluginTracer} from '@parcel/profiler';
 
 type Opts = {|
-  config: ParcelConfig,
   options: ParcelOptions,
+  reporters: Array<LoadedPlugin<Reporter>>,
   workerFarm: WorkerFarm,
 |};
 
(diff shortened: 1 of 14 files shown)


packages/core/core/src/ReporterRunner.js
```diff
@@ -113,6 +113,7 @@ export default class ReporterRunner {
           });
         } catch (reportError) {
           INTERNAL_ORIGINAL_CONSOLE.error(reportError);
+          process.exitCode = 1;
```
devongovett: I wonder if it is unexpected to do this for people using the Parcel programmatic API rather than our CLI? Should we store our own property here (rather than setting the global exit code), and then throw/emit an error from the main Parcel object that the CLI could pick up? We do already set an exit code in the CLI when an error is thrown.

packages/core/integration-tests/test/integration/reporters-failure/.parcelrc
marcins, line 1: You could simplify this fixutre by using `fsFixture` - example of an integration test with a reporter using fsFixture: https://github.com/parcel-bundler/parcel/blob/89a69f59b306e30d23136bb118dc9f6f95e15963/packages/core/integration-tests/test/api.js#L62-L109

packages/core/core/src/ReporterRunner.js
marcins: What if we just removed this `catch` and let the error propagate naturally?

packages/core/integration-tests/test/integration/reporters-failure/.parcelrc
MonicaOlejniczak: I initially had an fs fixture, but want to test the CLI as well which requires the files to live on disk

packages/core/core/src/ReporterRunner.js
MonicaOlejniczak: I was trying to avoid propagating the errors because not all of the calls are awaited on, and doing so could regress performance.

As for storing something that the CLI can pickup, I can try doing that but it will only get emitted at the end of the build unless the aforementioned un-awaited cases are changed. 

Alternatively, I could only propagate and await on buildSuccess events as that's the main use-case anyway.

Anyways the latest commit throws any received errors just before returning the build success event. I also initialise the reporters earlier so errors are seen earlier + adding to them becomes simpler.

packages/core/integration-tests/test/integration/reporters-failure/test-reporter/index.js
```diff
@@ -0,0 +1,9 @@
+const { Reporter } = require('@parcel/plugin');
+
+module.exports = new Reporter({
+  async report({ event }) {
+    if (event.type === 'buildSuccess') {
```
yamadapc, line 5: it'd be nice to test for cases where there is a delay here

yamadapc (approved):

PR #9736, opened 2024-05-24
Fix issue in async bundle runtime where it would not always add the runtime code

mattcompiles:
# ↪️ Pull Request

This fixes an issue with the `unstable_asyncBundleRuntime` feature that unlocks usage of `<script type=module async>` by allowing those bundles to be executed out of order. The current expectation is any bundle loaded by an HTML file should contain the extra runtime code, however we were seeing some manual shared bundles that had left out the extra runtime code. 

Previously we were relying on `BundleGraph.hasParentBundleOfType` to return false when one of it's parents bundles were of type `html`. However, in practice this doesn't always seem to be true when you have a complex manual shared bundle setup. 

I have now moved the check to use `BundleGraph.getReferencingBundles` which models this problem more accurately. This API instead reads the `reference` type edges in the bundle graph which models the following:

https://github.com/parcel-bundler/parcel/blob/d988abeb1c666ad2649d4c2b43a108eba0d7a226/packages/core/core/src/BundleGraph.js#L56-L65

This edge type is also the point of truth used by the `HTMLPackager` for which files to include which is exactly our use case.

## ✔️ PR Todo

### DIFF
--- a/packages/packagers/js/src/ScopeHoistingPackager.js
+++ b/packages/packagers/js/src/ScopeHoistingPackager.js
@@ -24,7 +24,6 @@ import ThrowableDiagnostic, {
 import globals from 'globals';
 import path from 'path';
 
-import logger from '@parcel/logger';
 import {ESMOutputFormat} from './ESMOutputFormat';
 import {CJSOutputFormat} from './CJSOutputFormat';
 import {GlobalOutputFormat} from './GlobalOutputFormat';
--- a/packages/packagers/js/src/ScopeHoistingPackager.js
+++ b/packages/packagers/js/src/ScopeHoistingPackager.js
@@ -24,6 +24,7 @@ import ThrowableDiagnostic, {
 import globals from 'globals';
 import path from 'path';
 
+import logger from '@parcel/logger';
 import {ESMOutputFormat} from './ESMOutputFormat';
 import {CJSOutputFormat} from './CJSOutputFormat';
 import {GlobalOutputFormat} from './GlobalOutputFormat';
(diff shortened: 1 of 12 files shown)


packages/packagers/js/src/ScopeHoistingPackager.js
```diff
@@ -264,14 +263,17 @@ export class ScopeHoistingPackager {
   }
 
   shouldBundleQueue(bundle: NamedBundle): boolean {
+    let referencingBundles = this.bundleGraph.getReferencingBundles(bundle);
+    let hasHtmlReference = referencingBundles.some(b => b.type === 'html');
```
marcins, line 267: What would happen if we stopped using HTML as an entry point and instead used JS entry points?

packages/core/integration-tests/test/integration/bundle-queue-runtime/yarn.lock
MonicaOlejniczak, line 1: Should be deleted?

MonicaOlejniczak (approved):

packages/packagers/js/src/ScopeHoistingPackager.js
MonicaOlejniczak: Yeah, worthwhile adding a test for it. But otherwise the premise seems fine

mattcompiles: Well it just wouldn't work lol. TBH so many other things would break if we did that this would be just one of many.

JakeLane (approved):

packages/packagers/js/src/ScopeHoistingPackager.js
jondlm: I'm curious about this. We have hacks in place internally to help with what I think is the same problem y'all are trying to address with the async bundle runtime. The only rub is that we use JS entry points. Anecdotally I've felt like things work a little funky with JS entrypoints and I'm wondering if we should shift to using an html entrypoint. Any more context on why html entries are your preferred approach?

PR #9740, opened 2024-05-27
Ensure const modules in async bundles are wrapped

marcins:
# ↪️ Pull Request

This fixes a case where an async module only exports a constant, and it doesn't end up getting registered.

We had a use case where there were lots of async bundles containing just string exports for a PDF renderer, and a big switch case that would dynamically import fonts on demand. These small async bundles weren't registering their exports.

Pseudo example based on the real pattern - where the real values could vary in size between several hundred bytes and 60KB+, hence deferred loading:

```js
// fonts.ts
export const fonts = {
   'abc': () => import('./abc'),
   'def': () => import('./def')
}

// abc.ts
export const value = "abc";


// consumer.ts
import { fonts } from './fonts';

const { value } = await fonts['abc']();
```

## 🚨 Test instructions

* Added new integration test that was failing before the fix.
* Verified internally that the affected async bundles correctly register their assets

## ✔️ PR Todo

### DIFF
--- a/packages/core/integration-tests/test/bundler.js
+++ b/packages/core/integration-tests/test/bundler.js
@@ -1247,6 +1247,14 @@ describe('bundler', function () {
 
     // This will fail when the async bundle does not export it's constant
     await run(b);
+
+    // Asset should not be inlined
+    const index = b.getBundles().find(b => b.name.startsWith('index'));
+    const contents = overlayFS.readFileSync(index.filePath, 'utf8');
+    assert(
+      !contents.includes('async value'),
+      'async value should not be inlined',
+    );
   });
   describe('manual shared bundles', () => {
     const dir = path.join(__dirname, 'manual-bundle');
--- a/packages/core/integration-tests/test/bundler.js
+++ b/packages/core/integration-tests/test/bundler.js
@@ -1214,6 +1214,40 @@ describe('bundler', function () {
     ]);
   });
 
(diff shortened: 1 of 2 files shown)


packages/core/integration-tests/test/bundler.js
```diff
@@ -1214,6 +1214,40 @@ describe('bundler', function () {
...
+          shouldScopeHoist: true,
+          sourceMaps: false,
+          shouldOptimize: false,
+        },
+        inputFS: overlayFS,
+      },
+    );
+
+    // This will fail when the async bundle does not export it's constant
+    await run(b);
```
mattcompiles, line 1249: Does the asset still get inlined in this case? It wouldn't be ideal if you ended up with something like:
```js
import('./async').then(m => console.log('async value'));
```

MonicaOlejniczak (approved):

packages/core/integration-tests/test/bundler.js
marcins: Good question, it doesn't seem to - I assume the static analysis can't track the usage to that level anyway?

From the bundle in the integration test:
```js
(parcelRequire("8Vicm")).then((m)=>console.log(m.value));
```

marcins: I've added additional verification in the integration test.

mattcompiles (approved):

PR #9748, opened 2024-05-29
Add a healthcheck endpoint to the devserver

yamadapc:
This assists in implementing a few self-healing mechanisms.

### DIFF
--- a/packages/reporters/dev-server/src/Server.js
+++ b/packages/reporters/dev-server/src/Server.js
@@ -497,7 +497,7 @@ export default class Server {
     });
 
     app.use((req, res, next) => {
-      if (req.url === '/healthcheck') {
+      if (req.url === '/__parcel_healthcheck') {
         res.statusCode = 200;
         res.write(`${Date.now()}`);
         res.end();
--- a/packages/reporters/dev-server/src/Server.js
+++ b/packages/reporters/dev-server/src/Server.js
@@ -495,6 +495,17 @@ export default class Server {
       setHeaders(res);
       next();
     });
+
+    app.use((req, res, next) => {
+      if (req.url === '/healthcheck') {
+        res.statusCode = 200;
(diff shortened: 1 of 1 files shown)


packages/reporters/dev-server/src/Server.js
```diff
@@ -495,6 +495,17 @@ export default class Server {
       setHeaders(res);
       next();
     });
+
+    app.use((req, res, next) => {
+      if (req.url === '/healthcheck') {
```
devongovett: Could conflict with a user's URL (eg static page or API proxy), no? Maybe we should prefix with `__parcel_` like we do for the HMR endpoint?

devongovett (commented):
Curious what you're planning on using this for

packages/reporters/dev-server/src/Server.js
yamadapc: Let me do that :+1:

yamadapc:
@devongovett We sometimes have 'connectivity issues' impact the dev-loop due to broken tunnels; we want the dev-server for our main application to health-check and recover from that.

packages/reporters/dev-server/src/Server.js
```diff
@@ -495,6 +495,17 @@ export default class Server {
       setHeaders(res);
       next();
     });
+
+    app.use((req, res, next) => {
+      if (req.url === '/__parcel_healthcheck') {
```
Nikola-3, line 500: Is the `__` in `__parcel_healthcheck` meant to make it "private"?

Nikola-3 (approved):

packages/reporters/dev-server/src/Server.js
yamadapc: Yes, we're just trying to prevent clashes and be consistent with `__parcel_hmr`

PR #9751, opened 2024-05-29
Main function for ParcelV3

alshdavid:
# ↪️ Pull Request

This adds a main function to kick off native code for the main thread and worker threads

Currently the main functions don't do anything but they will tie together the Rust and JavaScript worlds

## 🚨 Test instructions

```
yarn build-native && yarn test:integration
```

## ✔️ PR Todo

### DIFF
--- a/packages/core/core/src/native/main/main.js
+++ b/packages/core/core/src/native/main/main.js
@@ -6,7 +6,8 @@ export class ParcelNative {
     napi.mainBootstrap((_, e) => this.#onEvent(e));
   }
 
-  async #onEvent(event: any) {
+  // eslint-disable-next-line no-unused-vars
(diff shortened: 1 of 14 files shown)


packages/core/core/src/native/main/main.js
alshdavid, line 1: I expect to have

```javascript
async #onEvent(event: ParcelNativeMessage) {
  if ("SpawnWorker" in event) {
    // store this on some global state
    new Worker(path.join(__dirname, '..', 'worker', 'worker.js')) 
    return
  }
    // If event === "spawn worker"
}
```

Though in future `ParcelNative` would be replaced with bindings for a `Parcel` struct coming from the `parcel_core` package which replaces the `Parcel` class in `../../../Parcel.js`

PR #9752, opened 2024-05-29
Added rust linting check

alshdavid:
# ↪️ Pull Request

Added check for rust linting to the the pipeline build

## 🚨 Test instructions

```
yarn lint
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
     "link-all": "node scripts/link-all.js packages",
     "unlink-all": "node scripts/unlink-all.js packages",
     "check": "flow check",
-    "lint": "eslint . && prettier \"./packages/*/*/{src,bin,test}/**/*.{js,json,md}\" --list-different && cargo fmt --all -- --check",
+    "lint": "eslint . && prettier \"./packages/*/*/{src,bin,test}/**/*.{js,json,md}\" --list-different && cargo +nightly fmt --all -- --check",
     "prepublishOnly": "yarn adjust-versions && yarn build && yarn build-ts",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -25,7 +25,7 @@
     "link-all": "node scripts/link-all.js packages",
     "unlink-all": "node scripts/unlink-all.js packages",
     "check": "flow check",
-    "lint": "eslint . && prettier \"./packages/*/*/{src,bin,test}/**/*.{js,json,md}\" --list-different && cargo fmt --all -- --check",
+    "lint": "eslint . && prettier \"./packages/*/*/{src,bin,test}/**/*.{js,json,md}\" --list-different && cargo +nightly fmt --all -- --check",
```
marcins, line 28: Why is `+nightly` needed? We install Rust stable in CI so this fails to work.

alshdavid: Using nightly formatting features (basically all of the fmt options are nightly - something about not wanting to make formatting stable). This doesn't affect the compiler, only using nightly for the formater. 

The command fails when it gets to `cargo fmt` but CI doesn't fail (I guess we don't have `set -e` in the script?).

marcins: Yeah, sorry, that's what I meant - I think you'll need to update CI to _also_ install the `nightly` tooling (since currently it only installs the `stable` tooling)..

yamadapc: We should remove the nightly formatting options so we can avoid having to install it.

It's not worth the hassle, there's no benefit from those linter options and they are not standards across the community.

Having to install two toolchains is a downside.

yamadapc (changes requested):
Please remove the formatting options that require nightly

PR #9754, opened 2024-05-29
Reenable REPL deployment

mischnic:
Let's see if adding `scope: parcel` fixes it.

My guess is that this somehow broke when Vercel migrated all personal projects into a team a few months back.


MonicaOlejniczak (approved):
Will the scope need to be added to the tag release workflow too?

yugiio (approved):

marcins (approved):

PR #9756, opened 2024-05-30
Remove nightly rustfmt options

yamadapc:
We shouldn't be keeping nightly rustfmt options in the repository. This PR removes them to avoid having multiple toolchains as a requirement for development.

### DIFF
--- a/rustfmt.toml
+++ b/rustfmt.toml
@@ -1,3 +1,1 @@
 tab_spaces = 2
-imports_granularity = "Item"
-group_imports = "StdExternalCrate"


alshdavid:
Normally I'd agree, but there really is value in the formatting rules. There is also no risk associated with using a `nightly` formatter, the worst that could happen is your code is formatted incorrectly (we don't build our project with nightly).

Keep in mind that most of the rustfmt options will remain under nightly for the foreseeable future as the development team does not want to extend the stability guarantee to formatting being applied in a consistent fashion, especially as the language itself adds features.

If we agree that formatting is valuable and that we want to have it in Rust, using nightly Rust is the only way and likely will be the only way for the foreseeable future

yamadapc:
I propose we remove this because:

* Depending on nightly-only formatting requires developers to install nightly and set it up as the default formatter, which won't always be the case adding friction
* The options are non-standard and are not widely used 
* Nightly toolchain will need to be installed onto CI agents
* Nightly toolchain formatting might change every day, so our CI pipeline verifying formatting may (exaggerating this) produce different results on a daily basis
* Spending time on formatting is not ideal ; adding these flags has already wasted time from developers who had to fix conflicts with every file changing
* Splitting imports it not necessarily the best style, for example does not work with feature-gates

devongovett:
I prefer the default style fwiw.

devongovett (approved):

PR #9757, opened 2024-05-30
Sanitise user input in reporter runner

Nikola-3:
# ↪️ Pull Request
We were getting internal issues with the markdown reporter reaching an `Unexpected undefined` state when parsing the message in the diagnostic. The message is presumably coming from an untyped plugin where we allow the diagnostic to be just the string containing the message. This check should handle that case.

## ✔️ PR Todo

### DIFF
--- a/packages/core/core/src/ReporterRunner.js
+++ b/packages/core/core/src/ReporterRunner.js
@@ -21,6 +21,7 @@ import logger, {
 import PluginOptions from './public/PluginOptions';
 import BundleGraph from './BundleGraph';
 import {tracer, PluginTracer} from '@parcel/profiler';
+import {anyToDiagnostic} from '@parcel/diagnostic';
 
 type Opts = {|
   options: ParcelOptions,
@@ -85,7 +86,16 @@ export default class ReporterRunner {
     this.report(event);
   };
 
-  async report(event: ReporterEvent) {
+  async report(unsanitisedEvent: ReporterEvent) {
+    let event: ReporterEvent = unsanitisedEvent;
+    if (event.diagnostics) {
+      // Sanitise input before passing to reporters
+      // $FlowFixMe too complex to narrow down by type
+      event = {
+        ...event,
+        diagnostics: anyToDiagnostic(event.diagnostics),
+      };
+    }
     for (let reporter of this.reporters) {
       let measurement;
       try {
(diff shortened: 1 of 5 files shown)


packages/reporters/cli/src/CLIReporter.js
```diff
@@ -216,6 +216,8 @@ async function writeDiagnostic(
   let columns = getTerminalWidth().columns;
   let indent = 2;
   let spaceAfter = isError;
+  diagnostics =
```
marcins: Easy!

Did you validate this works? Are there any tests we can add a test for this to?

marcins (approved):

packages/reporters/cli/src/CLIReporter.js
Nikola-3: Well, the diagnostic is defined in all the cases where it fails since it would have failed earlier in the call stack with a different error to what we see so this should work unless someone is passing something funky... in which case I could add an error log so we can see what the diagnostic is in case it does fail?
In terms of tests, I guess I can copy one of the tests and change the diagnostic to a string.

mattcompiles: This feels like a strange fix. I think we should either be allowing a Diagnostic to be just a string, or warn in the case that its invalid. This code is handling a scenario that shouldn't really happen.

mattcompiles:
@Nikola-3 Could we try for slightly more descriptive PR title btw?

packages/reporters/cli/src/CLIReporter.js
MonicaOlejniczak: Agree, I think you can add something like this to fix the underlying problem and adjust the `Diagnostifiable` type accordingly.

```diff
  export function anyToDiagnostic(input: Diagnostifiable): Array<Diagnostic> {
    if (Array.isArray(input)) {
-      return input;
+      return input.map(anyToDiagnostic);
    }
    // ...
  }
```

Alternatively, update the reporter runner so that it uses `anyToDiagnostic` like so:

```diff
- this.errors.push(reportError);
+ this.errors.push(anyToDiagnostic(reportError));
```

packages/core/core/src/ReporterRunner.js
```diff
@@ -85,7 +86,16 @@ export default class ReporterRunner {
     this.report(event);
   };
 
-  async report(event: ReporterEvent) {
+  async report(unsanitisedEvent: ReporterEvent) {
+    let event: ReporterEvent = unsanitisedEvent;
+    if (event.diagnostics) {
```
MonicaOlejniczak, line 91: Is this really needed? Shouldn't the `anyToDiagnostic` fix this case?

mattcompiles: `anyToDiagnostic` is only used for build failures, not things that come from the logger. The thinking here was to ensure all diagnostics were valid before going out to the reporters.

PR #9758, opened 2024-05-30
Main function for ParcelV3 and FileSystem implementation

alshdavid:
# ↪️ Pull Request

This PR adds:
- A `Parcel` struct from the `parcel_core` crate
- A `ParcelNapi` struct from the `node_bindings` crate which re-exports the core crate implementation to JS
- A Napi FS implementation that plugs into the `Parcel` core options and bridges between JS and Rust
- Integration tests proving that the FS methods are being called. 

## 💻 Examples

The core package (conceptually, published via cargo) exposes a Rust API for building a project with Parcel via Rust (without JS plugins).

```rust
use parcel_core::Parcel;

fn main() {
  let parcel = Parcel::new(ParcelOptions { /*..*/ });
  let report = parcel.build(BuildOptions { /*..*/ });
}
```

The napi bindings are a consumer for this core crate and re-exports it with a JavaScript bindings

```javascript
import { Parcel } from 'parcel'

const parcel = new Parcel({ /*..*/ })
const report = await parcel.build({ /*..*/ })
```

The job of the binding-layer is to coerce values to/from Rust/JS types and handle threading. The plan is that, initially, the `Parcel.build()` method will return an `AssetGraph`. Over time we will change this as we move further along the build process until we return a `BuildReport`.

## 🚨 Test instructions

```
yarn build-native
yarn test:integration
```

## ✔️ PR Todo

### DIFF
--- a/crates/node-bindings/src/file_system/file_system_napi.rs
+++ b/crates/node-bindings/src/file_system/file_system_napi.rs
@@ -13,7 +13,7 @@ use serde::Serialize;
 // TODO error handling
 
 pub struct FileSystemNapi {
-  read_file_fn: Box<dyn Fn(PathBuf) -> String + Send + Sync>,
+  read_file_fn: Box<dyn Fn((PathBuf, String)) -> String + Send + Sync>,
   is_file_fn: Box<dyn Fn(PathBuf) -> bool + Send + Sync>,
   is_dir_fn: Box<dyn Fn(PathBuf) -> bool + Send + Sync>,
 }
@@ -44,7 +44,10 @@ impl FileSystemNapi {
 // thread or they will cause JavaScript to deadlock
 impl FileSystem for FileSystemNapi {
   fn read_to_string(&self, path: &Path) -> std::io::Result<String> {
-    Ok((*self.read_file_fn)(path.to_path_buf()))
+    Ok((*self.read_file_fn)((
+      path.to_path_buf(),
+      "utf8".to_string(),
+    )))
   }
 
   fn is_file(&self, path: &Path) -> bool {
--- a/crates/node-bindings/src/file_system/file_system_napi.rs
+++ b/crates/node-bindings/src/file_system/file_system_napi.rs
@@ -4,44 +4,37 @@ use std::sync::mpsc::channel;
 use std::sync::mpsc::Sender;
 use std::thread;
 
-use crossbeam_channel::Receiver;
 use napi::threadsafe_function::ThreadsafeFunctionCallMode;
 use napi::Env;
 use napi::JsFunction;
 use napi::JsObject;
 use napi::JsUnknown;
 use parcel_filesystem::FileSystem;
+use serde::de::DeserializeOwned;
+use serde::Serialize;
 
 pub struct FileSystemNapi {
-  tx_read_to_string: Sender<(PathBuf, Sender<String>)>,
-  tx_read_to_string: Sender<(PathBuf, Sender<String>)>,
+  tx_read_file_sync: Sender<(PathBuf, Sender<String>)>,
+  tx_is_file_sync: Sender<(PathBuf, Sender<bool>)>,
+  tx_is_dir_sync: Sender<(PathBuf, Sender<bool>)>,
 }
-
-/*
-{
-    canonicalize: path => this.options.fs.realpathSync(path),
-    read: path => this.options.fs.readFileSync(path),
-    isFile: path => this.options.fs.statSync(path).isFile(),
-    isDir: path => this.options.fs.statSync(path).isDirectory(),
-}
-*/
 
 impl FileSystemNapi {
   pub fn new(env: &Env, js_file_system: JsObject) -> napi::Result<Self> {
-    Ok(Self { tx_read_to_string })
+    Ok(Self {
+      tx_read_file_sync: create_js_callback(&env, &js_file_system, "readFileSync")?,
+      tx_is_file_sync: create_js_callback(&env, &js_file_system, "isFile")?,
+      tx_is_dir_sync: create_js_callback(&env, &js_file_system, "isDir")?,
+    })
   }
-
-  fn call_js_fs<T>(&self, method: &str, args: T) {}
 }
 
 impl FileSystem for FileSystemNapi {
-  //   fn canonicalize(&self, path: &Path, _cache: &DashMap<PathBuf, Option<PathBuf>>) {}
-
   fn read_to_string(&self, path: &Path) -> std::io::Result<String> {
     let (tx, rx) = channel();
 
     self
-      .tx_read_to_string
+      .tx_read_file_sync
       .send((path.to_path_buf(), tx))
       .unwrap();
 
@@ -49,29 +42,39 @@ impl FileSystem for FileSystemNapi {
   }
 
   fn is_file(&self, path: &Path) -> bool {
-    self.call_js_fs("isFile", (path.to_path_buf()));
+    let (tx, rx) = channel();
 
-    todo!();
+    self.tx_is_file_sync.send((path.to_path_buf(), tx)).unwrap();
+
+    rx.recv().unwrap()
   }
 
   fn is_dir(&self, path: &Path) -> bool {
-    todo!()
+    let (tx, rx) = channel();
+
+    self.tx_is_dir_sync.send((path.to_path_buf(), tx)).unwrap();
+
+    rx.recv().unwrap()
   }
 }
 
-fn create_js_callback<T, U>(env: &Env) -> Sender<(T, Sender<U>)> {
-  let (tx, rx) = channel::<(T, Sender<U>)>();
+fn create_js_callback<
+  Params: Send + Serialize + 'static,
+  Response: Send + DeserializeOwned + 'static,
+>(
+  env: &Env,
+  js_file_system: &JsObject,
+  method_name: &str,
+) -> napi::Result<Sender<(Params, Sender<Response>)>> {
+  let (tx, rx) = channel::<(Params, Sender<Response>)>();
 
-  let jsfn_read_to_string: JsFunction =
-    js_file_system.get_property(env.create_string("readFileSync")?)?;
+  let jsfn: JsFunction = js_file_system.get_property(env.create_string(method_name)?)?;
 
-  let tsfn_read_to_string = env.create_threadsafe_function(
-    &jsfn_read_to_string,
+  let tsfn = env.create_threadsafe_function(
+    &jsfn,
     0,
-    |ctx: napi::threadsafe_function::ThreadSafeCallContext<PathBuf>| {
-      let path = ctx.env.to_js_value(&ctx.value)?;
-      let encoding = ctx.env.create_string("utf8")?.into_unknown();
-      Ok(vec![path, encoding])
+    |ctx: napi::threadsafe_function::ThreadSafeCallContext<Params>| {
+      Ok(vec![ctx.env.to_js_value(&ctx.value)?])
     },
   )?;
 
@@ -79,13 +82,13 @@ fn create_js_callback<T, U>(env: &Env) -> Sender<(T, Sender<U>)> {
 
   thread::spawn(move || {
     while let Ok((path, rx_result)) = rx.recv() {
-      tsfn_read_to_string.call_with_return_value(
+      tsfn.call_with_return_value(
         Ok(path),
         ThreadsafeFunctionCallMode::Blocking,
         move |result: JsUnknown| {
           let env = unsafe { Env::from_raw(unsafe_env as _) };
 
-          let result = env.from_js_value::<String, JsUnknown>(result)?;
+          let result = env.from_js_value::<Response, JsUnknown>(result)?;
           rx_result.send(result).unwrap();
 
           return Ok(());
@@ -94,5 +97,5 @@ fn create_js_callback<T, U>(env: &Env) -> Sender<(T, Sender<U>)> {
     }
   });
 
-  tx
+  Ok(tx)
 }
--- a/crates/node-bindings/src/file_system/file_system_napi.rs
+++ b/crates/node-bindings/src/file_system/file_system_napi.rs
@@ -1,16 +1,11 @@
 use std::path::Path;
 use std::path::PathBuf;
-use std::sync::mpsc::channel;
-use std::sync::mpsc::Sender;
-use std::thread;
 
 use napi::bindgen_prelude::FromNapiValue;
-use napi::threadsafe_function::ThreadsafeFunction;
 use napi::threadsafe_function::ThreadsafeFunctionCallMode;
 use napi::Env;
 use napi::JsFunction;
 use napi::JsObject;
-use napi::JsUnknown;
 use parcel_filesystem::FileSystem;
 use serde::de::DeserializeOwned;
 use serde::Serialize;
@@ -69,8 +64,6 @@ fn create_js_thread_safe_method<
   js_file_system: &JsObject,
   method_name: &str,
 ) -> Result<impl Fn(Params) -> Response, napi::Error> {
-  let (tx, rx) = channel::<(Params, Sender<Response>)>();
-
   let jsfn: JsFunction = js_file_system.get_property(env.create_string(method_name)?)?;
 
   let threadsafe_function = env.create_threadsafe_function(
@@ -81,7 +74,7 @@ fn create_js_thread_safe_method<
     },
   )?;
   let result = move |params| {
-    let (tx, mut rx) = std::sync::mpsc::sync_channel(1);
+    let (tx, rx) = std::sync::mpsc::sync_channel(1);
     threadsafe_function.call_with_return_value(
       Ok(params),
       ThreadsafeFunctionCallMode::Blocking,
--- a/crates/node-bindings/src/file_system/file_system_napi.rs
+++ b/crates/node-bindings/src/file_system/file_system_napi.rs
@@ -4,6 +4,8 @@ use std::sync::mpsc::channel;
 use std::sync::mpsc::Sender;
 use std::thread;
 
+use napi::bindgen_prelude::FromNapiValue;
+use napi::threadsafe_function::ThreadsafeFunction;
 use napi::threadsafe_function::ThreadsafeFunctionCallMode;
 use napi::Env;
 use napi::JsFunction;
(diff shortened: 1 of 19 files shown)


crates/node-bindings/src/core/js_requests/request_options.rs
alshdavid, line 1: The js_requests should probably be removed, omitting that from this PR to keep the changes minimal

crates/parcel_filesystem/src/in_memory_file_system.rs
alshdavid, line 1: Needs to be `Send + Sync`

crates/parcel_filesystem/src/js_delegate_file_system.rs
alshdavid, line 1: This will be removed later, leaving the implementation as stubbed to avoid needing to remove its dependants in this PR

packages/core/integration-tests/test/parcel-v3.js
alshdavid, line 1: Need to create a wrapper around `ParcelNapi` that cleans up the API for JS consumers

packages/core/integration-tests/test/parcel-v3.js
```diff
@@ -0,0 +1,42 @@
+// @flow
+
+import assert from 'assert';
+import path from 'path';
+import {bundle, run} from '@parcel/test-utils';
+import * as napi from '@parcel/rust';
+import {inputFS} from '@parcel/test-utils';
+
+describe.only('parcel-v3', function () {
```
mattcompiles: Left a .only in

alshdavid: That's one way to make sure the integration tests pass :laughing:, Fixed

crates/parcel_filesystem/src/js_delegate_file_system.rs
```diff
@@ -1,115 +1,28 @@
...
-  Ok(PathBuf::from(result_string))
-}
-
-// Convert arbitrary errors to io errors. This is wrong; the `FileSystem` trait should use
-// `anyhow::Result`
-fn run_with_errors<T>(block: impl FnOnce() -> anyhow::Result<T>) -> Result<T, std::io::Error> {
-  let result = block();
-  result.map_err(|err| std::io::Error::new(std::io::ErrorKind::Other, err.to_string()))
-}
+pub struct JSDelegateFileSystem {}
```
yamadapc: Why do this? Better to just delete it

crates/parcel_filesystem/src/lib.rs
```diff
@@ -4,7 +4,6 @@ use std::path::PathBuf;
 
 use dashmap::DashMap;
 
-/// FileSystem implementation that delegates calls to a JS object
 pub mod js_delegate_file_system;
```
yamadapc: If the idea is just to be able to compile ouside of napi you can just feature-gate this

crates/parcel_filesystem/src/lib.rs
```diff
@@ -25,7 +24,7 @@ pub mod os_file_system;
 ///       it should not be in the trait
 /// * [ ] Do not use io results, instead use anyhow or this error
 ///
-pub trait FileSystem {
+pub trait FileSystem: Send + Sync {
```
yamadapc: There's no need to force this here; you should enforce on the usage site that the type also is `Send + Sync` if you need it

crates/parcel_filesystem/src/in_memory_file_system.rs
```diff
@@ -16,29 +18,29 @@ enum InMemoryFileSystemEntry {
 /// In memory implementation of the `FileSystem` trait, for testing purpouses.
 #[derive(Debug)]
 pub struct InMemoryFileSystem {
-  files: RefCell<HashMap<PathBuf, InMemoryFileSystemEntry>>,
-  current_working_directory: RefCell<PathBuf>,
+  files: Arc<RwLock<HashMap<PathBuf, InMemoryFileSystemEntry>>>,
```
yamadapc: There's no need to force send/sync over this; you could implement a lock based FS that wraps any possible implementation for example and implements Send/Sync;

crates/node-bindings/src/file_system/file_system_napi.rs
```diff
@@ -0,0 +1,106 @@
...
+use napi::JsObject;
+use napi::JsUnknown;
+use parcel_filesystem::FileSystem;
+use serde::de::DeserializeOwned;
+use serde::Serialize;
+
+// TODO error handling
+
+#[derive(Clone)]
+pub struct FileSystemNapi {
```
yamadapc, line 15: Really cool

crates/parcel_filesystem/src/in_memory_file_system.rs
yamadapc: You probably don't need to force it to be Send + Sync; 

Just add the Send+Sync constraint only where it's needed

crates/parcel_filesystem/src/js_delegate_file_system.rs
alshdavid: The JS requests depend on it and I didn't want to make the PR too big, next PR I will

crates/parcel_filesystem/src/lib.rs
alshdavid: I did that initially but I feel it's better that the consumer defines their implementations that satisfy Parcel's defined interface/traits as there doesn't seem to be a benefit to this abstraction.

Essentially:

`parcel_core` should export the `FileSystem` trait 
`parcel_napi` should create an implementation that satisfies that trait

In future:
`parcel_wasm` (...etc) should have their own implementations

crates/parcel_filesystem/src/lib.rs
alshdavid: I tried this initially however Rust would not compile with a `Arc<RwLock<Box<dyn FileSystem>>>>` that was sent between futures. I had to at least implement `Sync` - which forced the change to `Mutex`. Either way, better to internalize lock usage imo.

Have `FileSystem: Send + Sync + Clone`

yamadapc: You'd want this: `Arc<RwLock<Box<dyn FileSystem + Send>>>>`

Then you don't have to change any of the JS delegate stuff for example (I think)

yamadapc: We don't need to enforce the send/sync bounds here because it's technically valid to implement the FS API for single-threaded use. 

We can add the Send/Sync bounds where they are needed (e.g.: at the usage)

alshdavid: I'll try this

crates/node-bindings/src/file_system/file_system_napi.rs
```diff
@@ -0,0 +1,106 @@
...
+
+  let tsfn = env.create_threadsafe_function(
+    &jsfn,
+    0,
+    |ctx: napi::threadsafe_function::ThreadSafeCallContext<Params>| {
+      Ok(vec![ctx.env.to_js_value(&ctx.value)?])
+    },
+  )?;
+
+  let unsafe_env = env.raw() as usize;
```
alshdavid: https://github.com/napi-rs/napi-rs/issues/2128

crates/parcel_core/src/parcel/parcel.rs
```diff
@@ -1,12 +1,22 @@
-// TODO eventual public API for Parcel
-pub struct Parcel {}
+use std::sync::Arc;
+
+use parcel_filesystem::os_file_system::OsFileSystem;
+use parcel_filesystem::FileSystem;
+
+pub struct Parcel {
+  pub fs: Arc<dyn FileSystem>,
```
yamadapc:
```
fs: Arc<dyn FileSystem + Sync + Send>
```

crates/node-bindings/src/core/js_requests/request_options.rs
alshdavid: https://github.com/parcel-bundler/parcel/pull/9759

crates/parcel_filesystem/src/in_memory_file_system.rs
alshdavid: Fixed

crates/parcel_filesystem/src/lib.rs
alshdavid: Updated

crates/parcel_filesystem/src/lib.rs
alshdavid: Updated

crates/parcel_filesystem/src/js_delegate_file_system.rs
alshdavid: Updated

crates/parcel_filesystem/src/in_memory_file_system.rs
alshdavid: Fixed

crates/parcel_core/src/parcel/parcel.rs
alshdavid: Fixed

crates/node-bindings/src/file_system/file_system_napi.rs
```diff
@@ -0,0 +1,106 @@
...
+
+  thread::spawn(move || {
```
yamadapc: Wait, why do you have this?

yamadapc (changes requested):
There should be no new thread just to call napi thread-safe functions; you can just call the function directly

crates/node-bindings/src/file_system/file_system_napi.rs
```diff
@@ -0,0 +1,106 @@
...
+    while let Ok((path, rx_result)) = rx.recv() {
+      tsfn.call_with_return_value(
+        Ok(path),
+        ThreadsafeFunctionCallMode::Blocking,
+        move |result: JsUnknown| {
```
yamadapc: won't it work if you change this to be `result: Response`? then no need to have end around

crates/node-bindings/src/file_system/file_system_napi.rs
yamadapc: I am not sure if we need a thread just to forward calls here & wait; but this is fine I suppose, if we can avoid the env parts

This is a case where async/await will make things much cleaner

crates/node-bindings/src/file_system/file_system_napi.rs
```diff
@@ -0,0 +1,106 @@
+use std::path::Path;
```
yamadapc, line 1: We can do:

```
use std::path::Path;
use std::path::PathBuf;
use std::sync::mpsc::channel;
use std::sync::mpsc::Sender;
use std::thread;
use napi::bindgen_prelude::FromNapiValue;

use napi::threadsafe_function::{ThreadsafeFunction, ThreadsafeFunctionCallMode};
use napi::Env;
use napi::JsFunction;
use napi::JsObject;
use napi::JsUnknown;
use parcel_filesystem::FileSystem;
use serde::de::DeserializeOwned;
use serde::Serialize;

// TODO error handling

pub struct FileSystemNapi {
    // tx_read_file_sync: Sender<(PathBuf, Sender<String>)>,
    // tx_is_file_sync: Sender<(PathBuf, Sender<bool>)>,
    // tx_is_dir_sync: Sender<(PathBuf, Sender<bool>)>,
    read_file_fn: Box<dyn Fn(PathBuf) -> String + Send + Sync>,
    is_file_fn: Box<dyn Fn(PathBuf) -> bool + Send + Sync>,
    is_dir_fn: Box<dyn Fn(PathBuf) -> bool + Send + Sync>,
}

impl FileSystemNapi {
    pub fn new(env: &Env, js_file_system: JsObject) -> napi::Result<Self> {
        Ok(Self {
            read_file_fn: Box::new(create_js_thread_safe_method(&env, &js_file_system, "readFileSync")?),
            is_file_fn: Box::new(create_js_thread_safe_method(&env, &js_file_system, "isFile")?),
            is_dir_fn: Box::new(create_js_thread_safe_method(&env, &js_file_system, "isDir")?),
        })
    }
}

// These methods must be run off the nodejs main/worker
// thread or they will cause JavaScript to deadlock
impl FileSystem for FileSystemNapi {
    fn read_to_string(&self, path: &Path) -> std::io::Result<String> {
        Ok((*self.read_file_fn)(path.to_path_buf()))
    }

    fn is_file(&self, path: &Path) -> bool {
        (*self.is_file_fn)(path.to_path_buf())
    }

    fn is_dir(&self, path: &Path) -> bool {
        (*self.is_dir_fn)(path.to_path_buf())
    }
}

fn create_js_thread_safe_method<
    Params: Send + Serialize + 'static,
    Response: Send + DeserializeOwned + 'static + FromNapiValue,
>(
    env: &Env,
    js_file_system: &JsObject,
    method_name: &str,
) -> Result<impl Fn(Params) -> Response, napi::Error> {
    let (tx, rx) = channel::<(Params, Sender<Response>)>();

    let jsfn: JsFunction = js_file_system.get_property(env.create_string(method_name)?)?;

    let threadsafe_function = env.create_threadsafe_function(
        &jsfn,
        0,
        |ctx: napi::threadsafe_function::ThreadSafeCallContext<Params>| {
            Ok(vec![ctx.env.to_js_value(&ctx.value)?])
        },
    )?;
    let result = move |params| {
        let (tx, mut rx) = std::sync::mpsc::sync_channel(1);
        threadsafe_function.call_with_return_value(Ok(params), ThreadsafeFunctionCallMode::Blocking, move |result: Response| {
            let _ = tx.send(result);
            Ok(())
        });
        rx.recv().unwrap()
    };

    Ok(result)
}

```

package.json
```diff
@@ -70,7 +70,7 @@
   },
   "lint-staged": {
     "*.{js,json,md,ts}": "prettier --write",
-    "*.rs": "rustfmt"
+    "*.rs": "rustfmt +nightly"
```
yamadapc: revert this please

yamadapc (approved):

crates/node-bindings/src/file_system/file_system_napi.rs
```diff
@@ -0,0 +1,93 @@
...
+
+  let threadsafe_function = env.create_threadsafe_function(
+    &jsfn,
+    0,
+    |ctx: napi::threadsafe_function::ThreadSafeCallContext<Params>| {
+      Ok(vec![ctx.env.to_js_value(&ctx.value)?])
+    },
+  )?;
+  let result = move |params| {
+    let (tx, rx) = std::sync::mpsc::sync_channel(1);
```
devongovett, line 80: FWIW, I previously used `crossbeam_channel` when implementing this pattern elsewhere due to crashes in the standard library (possibly similar to the ones you saw with native OS locks). crossbeam didn't have this problem.

crates/node-bindings/src/file_system/file_system_napi.rs
```diff
@@ -0,0 +1,93 @@
...
+    threadsafe_function.call_with_return_value(
+      Ok(params),
+      ThreadsafeFunctionCallMode::Blocking,
+      move |result: Response| {
+        let _ = tx.send(result);
```
devongovett, line 85: FYI, napi-rs doesn't work well with errors here. If the JS function throws an error, the return value callback will never be called (see [here](https://github.com/napi-rs/napi-rs/blob/cd78f79bfa2cbe7be9b2f0b1974bd0d389b3c8c3/crates/napi/src/threadsafe_function.rs#L467)), and a fatal error will be thrown. Looks like this may be fixed in the upcoming napi-rs v3.

I had to wrap all of the JS functions in a try...catch and return the error instead. Then in rust you can do something like this:

```rust
let result = if result.is_error()? {
  Err(napi::Error::from(result))
} else {
  env.from_js_value(result)
};
let _ = tx.send(result);
```

Not super ideal...

crates/node-bindings/src/file_system/file_system_napi.rs
```diff
@@ -0,0 +1,93 @@
...
+  Params: Send + Serialize + 'static,
+  Response: Send + DeserializeOwned + 'static + FromNapiValue,
+>(
+  env: &Env,
+  js_file_system: &JsObject,
+  method_name: &str,
+) -> Result<impl Fn(Params) -> Response, napi::Error> {
+  let jsfn: JsFunction = js_file_system.get_property(env.create_string(method_name)?)?;
+
+  let threadsafe_function = env.create_threadsafe_function(
```
devongovett, line 72: Note that unless you use `ErrorStrategy::Fatal` on the threadsafe function, the JS callback will always receive an extra 1st parameter for the error (usually `null`). Could be unexpected on the JS side.

crates/node-bindings/src/file_system/file_system_napi.rs
```diff
@@ -0,0 +1,93 @@
...
+    let (tx, rx) = std::sync::mpsc::sync_channel(1);
+    threadsafe_function.call_with_return_value(
+      Ok(params),
+      ThreadsafeFunctionCallMode::Blocking,
+      move |result: Response| {
+        let _ = tx.send(result);
+        Ok(())
+      },
+    );
+    rx.recv().unwrap()
```
devongovett, line 89: This will only work if the file system call occurs on a background thread, otherwise this will deadlock. I tried replacing the existing JsFileSystem in the resolver with this, and it didn't work because in that case we call functions from the same thread. We could possibly check if the `result` function is called from the main thread (e.g. if `std::thread::current().id()` is the same as when it was created), and call the JS function directly instead of using the thread safe function. That way the same rust function could always be called regardless of which thread it is called from.

crates/node-bindings/src/file_system/file_system_napi.rs
```diff
@@ -0,0 +1,93 @@
...
+use napi::JsFunction;
+use napi::JsObject;
+use parcel_filesystem::FileSystem;
+use serde::de::DeserializeOwned;
+use serde::Serialize;
+
+// TODO error handling
+
+pub struct FileSystemNapi {
+  read_file_fn: Box<dyn Fn((PathBuf, String)) -> String + Send + Sync>,
```
devongovett, line 16: FYI, using a tuple here will result in the JS function being called with a single array argument rather than separate arguments. We could possibly use a trait to work around this. Unfortunately specialization is not stable in Rust so it's a bit annoying.

```rust
pub trait JsArgs: Serialize {
  fn to_js_args(&self, env: &Env) -> napi::Result<Vec<JsUnknown>> {
    Ok(vec![env.to_js_value(&self)?])
  }
}

impl<A: Serialize, B: Serialize> JsArgs for (A, B) {
  fn to_js_args(&self, env: &Env) -> napi::Result<Vec<JsUnknown>> {
    Ok(vec![env.to_js_value(&self.0)?, env.to_js_value(&self.1)?])
  }
}

impl JsArgs for String {}
impl JsArgs for PathBuf {}
impl JsArgs for bool {}
```

PR #9764, opened 2024-06-04
Added new crates for plugins and tidied up crates

alshdavid:
# ↪️ Pull Request

- Added `parcel` crate
- Moved code from `parcel_core` to `parcel`
- Added `parcel_plugin_resolver` with a `todo!()` implementation
- Added `parcel_plugin_transformer_js` with a `todo!()` implementation
- Added `#![deny(unused_crate_dependencies)]` to crates
- Removed unused dependencies
- Moved `parcel_napi_utils` into `node-bindings` crate
- Updated `@napi/cli` to fix incorrectly generated bindings

## ✔️ PR Todo


yamadapc (approved):

crates/parcel/src/parcel.rs
MonicaOlejniczak, line 1: This breaks the REPL build, since `OsFileSystem` is not available for wasm targets

> unresolved import `parcel_filesystem::os_file_system`
 --> crates/parcel/./src/parcel.rs:3:24
 | use parcel_filesystem::os_file_system::OsFileSystem;
 |                        ^^^^^^^^^^^^^^ could not find `os_file_system` in `parcel_filesystem`

devongovett:
Do we need so many crates? What's the difference between parcel and parcel_core? If the JS transformer and resolver are only going to be called from parcel core, do they need to be in separate crates?

Would be nice to keep parcel simpler than current v2, which has way too many packages IMO. Makes it harder to follow the code and onboard new contributors.

alshdavid:
> What's the difference between parcel and parcel_core?

`parcel` is essentially `libparcel` - where all the packages are re-exported and (theoretically) available from crates.io for rust consumers

`parcel_core` is "Parcel types + extras"

This separation is necessary to avoid a crazy import topology given the many-crates approach.  

> Do we need so many crates? 
> If the JS transformer and resolver are only going to be called from parcel core, do they need to be in separate crates?
> 
> Would be nice to keep parcel simpler than current v2, which has way too many packages IMO. Makes it harder to follow the code and onboard new contributors.

We'll talk more about this tomorrow - but - there are different opinions on this within the team. I am on team fewer crates but more meaningful separations. The argument for more crates is that it makes it easier to work with Parcel. 

It doesn't really matter too much in the end, as long as we don't have circular imports

crates/parcel/src/lib.rs
```diff
@@ -0,0 +1,5 @@
+#![deny(unused_crate_dependencies)]
+
+pub mod parcel;
+pub use parcel::*;
+pub use parcel_filesystem as file_system;
```
yamadapc, line 5: why don't we consume directly?

alshdavid: The `parcel` crate is everything you need to use and integrate with Parcel from Rust. 

The fact that we choose to split crates up is an internal implementation detail that consumers shouldn't be concerned about.

devongovett:
I still think there should be fewer crates, fewer files, less abstraction, etc. We should be simplifying parcel in this rewrite, not making it even more complicated. 😅

PR #9767, opened 2024-06-04
Add native node package manager with resolve support

mattcompiles:
# ↪️ Pull Request

This PR adds a native node package manager for use in the parcel config request. It also changes the node resolver implementation to use a `FileSystemRef` type rather than a FileSystem generic. 

Happy to add any more tests where we feel is appropriate.


crates/parcel_config/src/config_error.rs
```diff
@@ -26,6 +26,6 @@ pub enum ConfigError {
     from: PathBuf,
     specifier: String,
     #[source]
-    source: Box<dyn std::error::Error>,
+    source: Box<anyhow::Error>,
```
yamadapc: you can unbox it

yamadapc (approved):

crates/parcel_package_manager/src/node_package_manager.rs
```diff
@@ -0,0 +1,33 @@
...
+      ),
+    }
+  }
+}
+
+impl<'a> PackageManager for NodePackageManager<'a> {
+  fn resolve(&self, specifier: &str, from: &std::path::Path) -> anyhow::Result<crate::Resolution> {
+    let res = self.resolver.resolve(specifier, from, SpecifierType::Cjs);
+
+    match res.result? {
```
yamadapc, line 28: for what it's worth you could write this just with `??`

```
let (Resolution::Path(pathbuf), _) = res.result?.map_err(|err| anyhow!(...))?;
Ok(...)
```

MonicaOlejniczak (approved):

devongovett:
This is modifying the existing resolver which is used in v2 today. So I guess now would be the time to double check that moving to dyn doesn't have any performance impact.

mattcompiles:
@devongovett I tested this across a bunch of Jira builds are I couldn't detect any noticeable difference in performance.

MonicaOlejniczak (approved):

crates/node-bindings/src/resolver.rs
```diff
@@ -233,38 +192,38 @@ pub struct JsInvalidations {
 #[napi]
 pub struct Resolver {
   mode: u8,
-  #[cfg(not(target_arch = "wasm32"))]
-  resolver: parcel_resolver::Resolver<'static, EitherFs<JsFileSystem, OsFileSystem>>,
-  #[cfg(target_arch = "wasm32")]
-  resolver: parcel_resolver::Resolver<'static, JsFileSystem>,
+  resolver: parcel_resolver::Resolver<'static>,
   #[cfg(not(target_arch = "wasm32"))]
   invalidations_cache: parcel_dev_dep_resolver::Cache,
```
alshdavid, line 197: Does this need to be conditionally compiled out? Can we supply an empty `Cache` for `wasm` builds?

yamadapc (approved):

alshdavid (approved):

crates/node-bindings/src/resolver.rs
mattcompiles: My change actually removes the need for some of the conditions. We can revisit the other stuff later as it's not related to my changes.

PR #9768, opened 2024-06-04
WIP Renamed node-bindings and added a dedicated entry point for wasm

alshdavid:
# ↪️ Pull Request

- Moved `node-bindings` to `parcel_napi`
  - Put existing code under `src/parcel_v2`
  - Will remove `parcel_v2` or migrate code into new context as we progress
- Created `parcel_wasm` package to be an entry point for `wasm` consumers
-  Moved `wasm` code from `node-bindings` into `parcel_wasm`

## ✔️ PR Todo


devongovett:
Why? Isn't parcel_wasm mostly a duplicate of the napi bindings? There's only a few differences that were covered by conditional compilation before.

crates/parcel_wasm/src/function_ref.rs
```diff
@@ -0,0 +1,31 @@
...
+use napi::JsFunction;
+use napi::Ref;
+
+pub struct FunctionRef {
+  pub env: Env,
+  pub reference: Ref<()>,
+}
+
+// We don't currently call functions from multiple threads, but we'll need to change this when we do.
+unsafe impl Send for FunctionRef {}
```
yamadapc: Why do you need this if you don't call it from multiple threads?

crates/parcel_wasm/src/resolver.rs
```diff
@@ -0,0 +1,313 @@
...
+  mode: u8,
+  resolver: parcel_resolver::Resolver<'static, FileSystemWasm>,
+}
+
+#[napi]
+impl Resolver {
+  #[napi(constructor)]
+  pub fn new(project_root: String, options: JsResolverOptions, env: Env) -> Result<Self> {
+    let fs = {
+      let fsjs = options.fs.unwrap();
```
yamadapc: you can just fail here ; `options.fs.unwrap_or(Err...)?`

crates/parcel_wasm/src/file_system.rs
```diff
@@ -0,0 +1,97 @@
...
+use napi::JsBoolean;
+use napi::JsBuffer;
+use napi::JsFunction;
+use napi::JsString;
+use napi_derive::napi;
+use parcel_resolver::FileSystem;
+
+use crate::function_ref::FunctionRef;
+
+pub struct FileSystemWasm {
```
yamadapc: https://github.com/parcel-bundler/parcel/pull/9759/files#diff-02822a9f6bd8b5da0af7e92d669a665c2183ff1e763618817085dc5f9eebd3c6 

was a safe implementation of this

crates/parcel_wasm/src/resolver.rs
```diff
@@ -0,0 +1,313 @@
...
+  }
+
+  fn resolve_internal(
+    &self,
+    options: ResolveOptions,
+  ) -> napi::Result<(parcel_resolver::ResolveResult, bool, u8)> {
+    let mut res = self.resolver.resolve_with_options(
+      &options.filename,
+      Path::new(&options.parent),
+      match options.specifier_type.as_ref() {
```
yamadapc: nit, assign these to variables so that we have names

crates/parcel_wasm/src/resolver.rs
```diff
@@ -0,0 +1,313 @@
...
+          ))
+        }
+        _ => {
+          return Err(napi::Error::new(
+            napi::Status::InvalidArg,
+            format!("Invalid specifier type: {}", options.specifier_type),
+          ))
+        }
+      },
+      if let Some(conditions) = options.package_conditions {
```
yamadapc: same here

crates/parcel_wasm/src/resolver.rs
```diff
@@ -0,0 +1,313 @@
...
+    Vec<Either3<FilePathCreateInvalidation, FileNameCreateInvalidation, GlobCreateInvalidation>>,
+  pub invalidate_on_startup: bool,
+}
+
+#[napi]
+pub struct Resolver {
+  mode: u8,
+  resolver: parcel_resolver::Resolver<'static, FileSystemWasm>,
+}
+
```
yamadapc: you could break up a lot of these functions

crates/parcel_nodejs/src/parcel_v2/transformer.rs
```diff
@@ -0,0 +1,40 @@
...
+  let result = parcel_js_swc_core::transform(config, None)?;
+  env.to_js_value(&result)
+}
+
+#[napi]
+pub fn transform_async(opts: JsObject, env: Env) -> napi::Result<JsObject> {
+  let call_macro = if opts.has_named_property("callMacro")? {
+    let func = opts.get_named_property::<JsUnknown>("callMacro")?;
+    if let Ok(func) = func.try_into() {
+      Some(create_macro_callback(func, env)?)
```
yamadapc, line 20: I'm not sure if this function is sound ; can we talk about what it's doing & try to do a safe implementation?

yamadapc (changes requested):
I don't understand why we're duplicating all this code

crates/parcel_wasm/src/function_ref.rs
alshdavid: Good point, I'm just moving existing code and not making changes to the sources

alshdavid:
> Why? Isn't parcel_wasm mostly a duplicate of the napi bindings? There's only a few differences that were covered by conditional compilation before.

I'm thinking of abstracting those peices out on their own and share them between `parcel_nodejs` and `parcel_wasm` (which function as entry points for those targets). 

For instance, the Resolver implementation can be moved into `parcel_plugin_resolver/napi_bindings/parcel-v2` and gated via crate features `wasm` or `nodejs`.

This would avoid duplication, allowing the Resolver code to remain largely unchanged whist splitting the wasm and nodejs entrypoints up.

In the end state the binding layer will be significantly lighter so this will be replaced

devongovett:
But why do we even need separate entry points for them in the first place if they are both using napi?

PR #9772, opened 2024-06-05
Add plugin loading

MonicaOlejniczak:
# ↪️ Pull Request

These changes introduce a `Plugins` struct which is essentially a port of `ParcelConfig` that lazily loads builtin Rust plugins and JavaScript plugins. The exact mechanism for loading JS plugins and injecting env is not dealt with in this pull request.

**Details**
* Each plugin is instantiated with a `PluginConfig` which represents common dependencies plugins need to access
* The way the plugin config is built may be tweaked later to account for logging and tracing sources
* `load_config` has been removed in the plugin traits to reduce complexity, and instead this functionality is expected to live in the native rust plugin / adapter constructor. This is equivalent to current usages, as the config is always called shortly after instantiation for all plugin types. The loading of native rust plugins that are not built into parcel may require a slightly different API / trait later but that is likely a fair way off.

## 🚨 Test instructions


crates/parcel/src/adapter/bundler_adapter.rs
alshdavid, line 1: I'd imagine we would move the "adapter plugins" to be their own crates to stay consistent with the multi-crate structure?

```
crates\
  parcel_core
  parcel_plugin_resolver
  parcel_plugin_resolver_rpc
  parcel_plugin_transformer
  parcel_plugin_transformer_rpc
```
Though I still think we should internalize everything within the `parcel` crate :stuck_out_tongue:

MonicaOlejniczak: Not everything _has_ to be in another crate, even with a multi-crate architecture. My thinking was to strike a balance between the two approaches, and have the adapters in the parcel crate. Otherwise there will be a lot of crates, that don't really serve any value.

crates/parcel/src/adapter.rs
```diff
@@ -0,0 +1,28 @@
...
+mod resolver_adapter;
+pub use resolver_adapter::*;
+
+mod runtime_adapter;
+pub use runtime_adapter::*;
+
+mod transformer_adapter;
+pub use transformer_adapter::*;
+
+pub trait Adapter {}
```
yamadapc: what is adapter?

mattcompiles (approved):

crates/parcel/src/adapter/bundler_adapter.rs
```diff
@@ -0,0 +1,35 @@
...
+use parcel_core::plugin::PluginContext;
+
+use super::Adapter;
+
+#[derive(Debug)]
+pub struct BundlerAdapter {
+  name: String,
+}
+
+impl BundlerAdapter {
```
yamadapc: So is this NapiBundler?

yamadapc (approved):

crates/parcel/src/adapter.rs
MonicaOlejniczak: Removed the concept for now, this will be revisited once we actually start loading JS plugins

devongovett:
Since Parcel plugins are stateless, do we need to "load" them ahead of time? Could we just load them as we run them? As you have it here, this will result in a bunch of `Vec<Box<_>>` being allocated and passed around, but if those don't store anything beyond the name of the plugin (which we already get from the config), then do we need to do this at all?

MonicaOlejniczak:
@devongovett The plan was to initialise plugins with config, and come up with a solution for using the plugin name as the source in logging and tracing. These are yet to be implemented, and still being iterated on. The exact solution may change once we progress further with JS plugins and the overall details. If you have other ideas, let me know.

devongovett:
Seems like napi bridge could just be a single function that receives a message (what plugin to apply, and the inputs e.g. asset), and returns a response. Then the plugins could all be loaded and executed on the JS side rather than in rust. That would avoid needing to implement a lot of complex stuff to go back and forth (e.g. passing asts around). For example here's what I wrote on my branch to run JS transformer plugins: https://github.com/parcel-bundler/parcel/blob/8ff653c54e967613f81ff86e5ba26c434118b63b/packages/core/core/src/requests/AssetGraphRequestRust.js#L238-L363 It's just one function to run a transformer plugin and all of the individual steps (e.g. load config, parse, transform, generate, etc.) are run from there. The plugin is loaded as part of running it rather than ahead of time as well.

PR #9773, opened 2024-06-05
Moved resolver-rs into parcel_plugin_resolver/core

alshdavid:
# ↪️ Pull Request

- Moved `packages/utils/resolver-rs` into `crates/parcel_plugin_resolver`
- Moved `packages/utils/dev-dep-resolver` into `crates/node-bindings` 
- Moved `SpecifierType` from `resolver` into `parcel_core/plugins/resolver` (to avoid circular dep)
- Moved `ExportsCondition` from `resolver` into `parcel_core/types` (to avoid circular dep)
- Moved `IncludeNodeModules` from `resolver` into `parcel_core/types/environment` (to avoid circular dep)
- Re-exporting `resolver` from `parcel`
- Moved tests from `packages/utils/resolver-core/test` into `crates/parcel_plugin_resolver/test`

Going from this:
image

To this
image


crates/parcel/src/lib.rs
alshdavid, line 1: Expected usage to be something like:

```rust
use parcel::Parcel;
use parcel::ParcelInitOptions;
use parcel::ParcelBuildOptions;
use parcel::plugins::Resolver;

let parcel = Parcel::new(ParcelInitOptions {
  override_resolvers: &[Resolver::default()], // not actually used like this - but essentially
  ..Default::default()
})

let report = parcel.build(ParcelBuildOptions::default());
```

devongovett:
I would like the resolver crate to remain independent. It is useful even outside Parcel so ideally should be a standalone library and not  parcel-specific plugin. I believe it is already used in a few places externally as well via git deps, but I was planning on publishing it to crates.io at some point too.

yamadapc:
Won't be merged as is

PR #9774, opened 2024-06-05
Clean-up existing cache entries on overwrite

yamadapc:
The current changes after #9726 is leaving garbage files behind when entries are overwritten, which happens often.


mattcompiles:
@yamadapc I'm not against doing this fix, but I do think there'd much much less complexity with the same benefit by rolling back the large blob chunking feature instead. It currently never activates so I don't think it has a lot of value.

yamadapc:
>  I'm not against doing this fix, but I do think there'd much much less complexity with the same benefit by rolling back the large blob chunking feature instead. It currently never activates so I don't think it has a lot of value.

Let's look at this but fix `v2` current state first.

mattcompiles (approved):

PR #9776, opened 2024-06-06
Native resolver plugin

mattcompiles:
# ↪️ Pull Request

This PR implements the default resolver for the rust core. The actual resolver code was ported from @devongovett's [core-rs2 branch](https://github.com/parcel-bundler/parcel/blob/core-rs2/crates/core/src/requests/path_request.rs#L276). I've left a bunch of `TODO` comments in the code with open questions or things we'll need to address later like invalidations. I think there's also some opportunities to improve the `Resolutions` type and change it to be an enum for things like externals.

I've also updated the `InMemoryFileSystem` to use `RwLock`, allowing it to be used inside an `Arc`. This is required for testing against the `FileSystemRef` type.

### DIFF
--- a/crates/parcel_plugin_resolver/src/resolver.rs
+++ b/crates/parcel_plugin_resolver/src/resolver.rs
@@ -233,7 +233,7 @@ fn excluded_resolution() -> Resolution {
     side_effects: true,
     can_defer: false,
     code: None,
-    meta: None,
+    meta: JSONObject::default(),
     pipeline: None,
     priority: None,
     query: None,
--- a/crates/parcel_plugin_resolver/src/resolver.rs
+++ b/crates/parcel_plugin_resolver/src/resolver.rs
@@ -317,7 +317,7 @@ mod test {
         can_defer: false,
         code: None,
         is_excluded: false,
-        meta: None,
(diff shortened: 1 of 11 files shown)


alshdavid (approved):

crates/parcel_plugin_resolver/src/resolver.rs
```diff
@@ -1,19 +1,328 @@
...
+        pipeline: None,
+        priority: None,
+        query: None,
+      }),
+      (parcel_resolver::Resolution::Builtin(builtin), _invalidations) => {
+        self.resolve_builtin(ctx, builtin)
+      }
+      (parcel_resolver::Resolution::Empty, _invalidations) => Ok(Resolution {
+        file_path: self
+          .project_root
```
yamadapc, line 192: is project_root the parcel root?

yamadapc (approved):

crates/parcel_plugin_resolver/src/resolver.rs
mattcompiles: No, I should have put a todo here as well. This is a temporary hack, we'll need a better way to resolve the empty file.

PR #9777, opened 2024-06-07
Added start of rpc implemention for Rust <-> JS plugins

alshdavid:
# ↪️ Pull Request

- Added `parcel_plugin_rpc` crate
- Moved/renamed RPC plugins to `plugin_parcel_rpc`
- Added stub implementation for Napi RPC host
- Added `Option<Arc<dyn RpcHost>>` to `ParcelOptions`
- Initializing `RpcHostNodejs` from napi context and passing it into `Parcel`
- Updated tests for `Plugins` to ignore struct properties

### DIFF
--- a/crates/parcel/src/parcel.rs
+++ b/crates/parcel/src/parcel.rs
@@ -3,13 +3,16 @@ use std::sync::Arc;
 use parcel_filesystem::os_file_system::OsFileSystem;
 use parcel_filesystem::FileSystemRef;
 use parcel_package_manager::NodePackageManager;
+use parcel_plugin_rpc::RpcHostRef;
 
 pub struct Parcel {
   pub fs: FileSystemRef,
+  pub rpc: Option<RpcHostRef>,
 }
 
 pub struct ParcelOptions {
   pub fs: Option<FileSystemRef>,
+  pub rpc: Option<RpcHostRef>,
 }
 
 impl Parcel {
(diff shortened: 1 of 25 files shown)


crates/parcel/src/parcel.rs
```diff
@@ -3,22 +3,29 @@ use std::sync::Arc;
...
+use parcel_plugin_rpc::RpcHostRef;
 
 pub struct Parcel {
   pub fs: FileSystemRef,
+  pub rpc: Option<RpcHostRef>,
 }
 
 pub struct ParcelOptions {
   pub fs: Option<FileSystemRef>,
+  pub rpc: Option<RpcHostRef>,
```
alshdavid, line 15: The difference with Mach is I used a `HashMap<String, RpcHostRef>` to allow mutliple plugin execution runtimes to run simultaniously which you could select from the `.parcelrc` 
```yaml
resolvers:
  "nodejs:./plugins/resolver.js" # nodejs is the default when "nodejs:" omitted
  "wasm:./plugins/resolver.wasm"
  "rlib:./plugins/resolver.rlib"
```

crates/parcel/src/plugins.rs
```diff
@@ -57,7 +56,7 @@ impl<'a> Plugins<'a> {
   }
 
   pub fn bundler(&self) -> Result<Box<dyn BundlerPlugin>, anyhow::Error> {
-    Ok(Box::new(NapiBundlerPlugin::new(
+    Ok(Box::new(PluginBundlerRpc::new(
```
MonicaOlejniczak: Naming should be `RpcBundlerPlugin` and for the rest too

alshdavid: Updated

MonicaOlejniczak (approved):

PR #9778, opened 2024-06-07
Added Ping/Pong RPC message circuit for Rust <-> Js plugin interaction

alshdavid:
# ↪️ Pull Request

- Tidied up the `ParcelNapi` constructor a little
- Moved `Parcel` (within `ParcelNapi`) into its own thread
  - Interactions with `Parcel` now occur over a channel
- Added `rpc` callback to `ParcelNapi` Js options object
- Added `napi` to `parcel_plugin_rpc` behind `nodejs` feature
- Added `Ping`/`Pong` circuit to prove a round trip works
- Added napi bindings to `RpcHostNodejs` (not an `unsafe` in sight :rofl:)
- Updated the `RpcHostMessage` enum to reflect changes
- Updated `parcel-v3` test to test the `Ping/Pong` circuit

## 🚨 Test instructions

```
yarn build-native && yarn test:integration
```

## ✔️ PR Todo

### DIFF
--- a/crates/node-bindings/Cargo.toml
+++ b/crates/node-bindings/Cargo.toml
@@ -14,63 +14,49 @@ rustls = ["sentry/rustls"]
 openssl = ["sentry/native-tls"]
 
 [dependencies]
-napi-derive = "2.16.3"
+parcel = { path = "../parcel" }
 parcel-js-swc-core = { path = "../../packages/transformers/js/core" }
 parcel-resolver = { path = "../../packages/utils/node-resolver-rs" }
-parcel_plugin_rpc = { path = "../parcel_plugin_rpc", features = ["nodejs"] }
-parcel = { path = "../parcel" }
-dashmap = "5.4.0"
-xxhash-rust = { version = "0.8.2", features = ["xxh3"] }
-log = "0.4.21"
-tracing = "0.1.40"
-tracing-subscriber = "0.3.18"
 
+anyhow = "1.0.82"
+dashmap = "5.4.0"
 glob = "0.3.1"
+log = "0.4.21"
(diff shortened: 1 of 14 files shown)


crates/node-bindings/src/parcel/parcel.rs
```diff
@@ -1,62 +1,138 @@
...
+            response.send(parcel.fs.is_dir(&path)).is_err()
+          }
+          ParcelMessage::RpcPing { response } => response
+            .send(parcel.rpc.as_ref().unwrap().ping().unwrap())
+            .is_err(),
+        } {
+          return;
+        }
+      }
+    });
```
alshdavid: Running Parcel in its own thread is required because `RpcHostNodejs` will block the main thread waiting for a response. 

We can flatten this out eventually - though any methods we run will need to be on their own thread - something like this:

```rust
#[napi]
impl ParcelNapi {
  #[napi]
  pub fn build(&self) {
    let tsfn = /**/
    let parcel = self.parcel.clone();
    thread::spawn(move || {
       tsfn.call(Ok(parcel.build()));
    });
  }
}
```

With a JS shim that returns a promise

MonicaOlejniczak (approved):

crates/node-bindings/Cargo.toml
alshdavid, line 1: Sorted this file alphabetically and added features to `parcel` so it wouldn't break the wasm target

crates/node-bindings/src/parcel/parcel.rs
alshdavid: Actually the `#[napi] async fn` runs on a Tokio worker thread so it should be fine

PR #9786, opened 2024-06-11
Add watch logging

mattcompiles:
# ↪️ Pull Request

This PR just adds some logging when the watcher callback is fired. We're adding this due to some recent watcher related incidents that this information would have helped diagnose.


MonicaOlejniczak (approved):

yamadapc:
Can you add the first file change event to the log message @mattcompiles ?

yamadapc (approved):

PR #9787, opened 2024-06-11
Alsh added multiple Nodejs workers and workers RPC

alshdavid:
- Added `RpcConnection` trait to describe an active worker (or worker pool the case of napi)
- Added `RpcHost` to act as a factory for creating an `RpcConnection`
- Added `RpcConnectionNodejs` to represent the connection to one Nodejs worker
- Added `RpcConnectionNodejsMulti` to load balance across multiple Nodejs workers
  - Currently uses round-robin to schedule work
- Added JavaScript class `ParcelV3` that adds the JS side of the bindings 
  - This will eventually replace `Parcel`, offloading everything to Rust
- Added JavaScript class `ParcelWorker` that handles the bindings for Nodejs workers
- Updated tests to use the new wrapper
  - Not sure how to tests threads in the integration tests, but I tested them in isolation and they work as expected

### DIFF
--- a/crates/node-bindings/src/parcel/parcel.rs
+++ b/crates/node-bindings/src/parcel/parcel.rs
@@ -2,6 +2,7 @@ use std::path::PathBuf;
 use std::sync::Arc;
 
 use napi::Env;
+use napi::JsNumber;
 use napi::JsObject;
 use napi_derive::napi;
 use parcel::rpc::nodejs::RpcHostNodejs;
@@ -28,7 +29,10 @@ impl ParcelNapi {
     let fs = FileSystemNapi::from_options(&env, &options)?;
 
     // Set up Nodejs plugin bindings
-    let rpc_host_nodejs = RpcHostNodejs::new(&env, options.get_named_property("rpc")?)?;
+    let node_workers: JsNumber = options.get_property(env.create_string("nodeWorkers")?)?;
+    let node_workers = node_workers.get_uint32()?;
+    let rpc_host_nodejs =
+      RpcHostNodejs::new(&env, options.get_named_property("rpc")?, node_workers)?;
 
     // Initialize Parcel
     let parcel = Parcel::new(ParcelOptions {
@@ -39,6 +43,19 @@ impl ParcelNapi {
     Ok(Self {
       parcel: Arc::new(parcel),
     })
+  }
+
+  #[napi]
+  pub async fn build(&self) -> napi::Result<()> {
+    self.parcel.build().unwrap();
+    Ok(())
+  }
+
+  #[napi]
+  pub fn default_thread_count(env: Env) -> napi::Result<JsNumber> {
+    let cpus = num_cpus::get();
+    let cpus = env.create_int32(cpus as i32)?;
+    Ok(cpus)
   }
 
   // Temporary, for testing
--- a/crates/parcel/src/parcel.rs
+++ b/crates/parcel/src/parcel.rs
@@ -1,8 +1,10 @@
 use std::sync::Arc;
 
+use anyhow;
 use parcel_filesystem::os_file_system::OsFileSystem;
 use parcel_filesystem::FileSystemRef;
 use parcel_package_manager::NodePackageManager;
+use parcel_plugin_rpc::RpcConnectionRef;
 use parcel_plugin_rpc::RpcHostRef;
 
 pub struct Parcel {
@@ -27,5 +29,15 @@ impl Parcel {
       fs,
       rpc: options.rpc,
     }
+  }
+
+  pub fn build(&self) -> anyhow::Result<()> {
+    let mut _rpc_connection = None::<RpcConnectionRef>;
+
+    if let Some(rpc_host) = &self.rpc {
+      _rpc_connection = Some(rpc_host.start()?);
+    }
+
+    Ok(())
   }
 }
--- a/crates/parcel_plugin_rpc/src/nodejs/napi.rs
+++ b/crates/parcel_plugin_rpc/src/nodejs/napi.rs
@@ -1,17 +1,12 @@
-use std::sync::mpsc::Receiver;
 use std::sync::mpsc::Sender;
-use std::thread;
 
-use napi::threadsafe_function::ThreadsafeFunction;
-use napi::threadsafe_function::ThreadsafeFunctionCallMode;
 use napi::Env;
 use napi::JsUnknown;
-use napi::Status;
 use serde::de::DeserializeOwned;
 
 // Generic method to create a "resolve" javascript function to
 // return the value from the thread safe function
-pub fn create_callback<Returns: DeserializeOwned + 'static>(
+pub fn create_done_callback<Returns: DeserializeOwned + 'static>(
   env: &Env,
   reply: Sender<Returns>,
 ) -> napi::Result<JsUnknown> {
@@ -30,20 +25,4 @@ pub fn create_callback<Returns: DeserializeOwned + 'static>(
     .into_unknown();
 
   Ok(callback)
-}
-
-pub fn wrap_threadsafe_function<T: Send>(
-  threadsafe_function: ThreadsafeFunction<T>,
-  rx: Receiver<T>,
-) {
-  thread::spawn(move || {
-    while let Ok(msg) = rx.recv() {
-      if !matches!(
-        threadsafe_function.call(Ok(msg), ThreadsafeFunctionCallMode::NonBlocking),
-        Status::Ok
-      ) {
-        return;
-      };
-    }
-  });
 }
--- a/crates/parcel_plugin_rpc/src/nodejs/napi.rs
+++ b/crates/parcel_plugin_rpc/src/nodejs/napi.rs
@@ -1,0 +1,49 @@
+use std::sync::mpsc::Receiver;
+use std::sync::mpsc::Sender;
+use std::thread;
+
+use napi::threadsafe_function::ThreadsafeFunction;
+use napi::threadsafe_function::ThreadsafeFunctionCallMode;
+use napi::Env;
+use napi::JsUnknown;
+use napi::Status;
+use serde::de::DeserializeOwned;
+
+// Generic method to create a "resolve" javascript function to
+// return the value from the thread safe function
+pub fn create_callback<Returns: DeserializeOwned + 'static>(
+  env: &Env,
+  reply: Sender<Returns>,
+) -> napi::Result<JsUnknown> {
+  let callback = env
+    .create_function_from_closure("callback", move |ctx| {
+      let response = ctx
+        .env
+        .from_js_value::<Returns, JsUnknown>(ctx.get::<JsUnknown>(0)?)?;
+
+      if reply.send(response).is_err() {
+        return Err(napi::Error::from_reason("Unable to send rpc response"));
+      }
+
+      ctx.env.get_undefined()
+    })?
+    .into_unknown();
+
+  Ok(callback)
+}
+
+pub fn wrap_threadsafe_function<T: Send>(
+  threadsafe_function: ThreadsafeFunction<T>,
+  rx: Receiver<T>,
+) {
+  thread::spawn(move || {
+    while let Ok(msg) = rx.recv() {
+      if !matches!(
+        threadsafe_function.call(Ok(msg), ThreadsafeFunctionCallMode::NonBlocking),
+        Status::Ok
+      ) {
+        return;
+      };
+    }
+  });
+}
--- a/crates/parcel_plugin_rpc/src/nodejs/rpc_conns_nodejs.rs
+++ b/crates/parcel_plugin_rpc/src/nodejs/rpc_conns_nodejs.rs
@@ -6,12 +6,12 @@ use crate::RpcConnection;
 
 /// Connection to multiple Nodejs Workers
 /// Implements round robin messaging
-pub struct RpcConnectionsNodejs {
+pub struct RpcConnectionNodejsMulti {
   current_index: Mutex<usize>, // TODO use AtomicUsize
   conns: Vec<RpcConnectionNodejs>,
 }
 
-impl RpcConnectionsNodejs {
+impl RpcConnectionNodejsMulti {
   pub fn new(conns: Vec<RpcConnectionNodejs>) -> Self {
     Self {
       current_index: Default::default(),
@@ -30,7 +30,7 @@ impl RpcConnectionsNodejs {
   }
 }
 
-impl RpcConnection for RpcConnectionsNodejs {
+impl RpcConnection for RpcConnectionNodejsMulti {
   fn ping(&self) -> anyhow::Result<()> {
     let next = self.next_index();
     self.conns[next].ping()
--- a/crates/parcel_plugin_rpc/src/nodejs/rpc_conns_nodejs.rs
+++ b/crates/parcel_plugin_rpc/src/nodejs/rpc_conns_nodejs.rs
@@ -1,0 +1,38 @@
+use parking_lot::Mutex;
+
+use super::RpcConnectionNodejs;
+
+use crate::RpcConnection;
+
+/// Connection to multiple Nodejs Workers
+/// Implements round robin messaging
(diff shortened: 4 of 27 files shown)


crates/parcel_plugin_rpc/src/nodejs/napi.rs
```diff
@@ -0,0 +1,49 @@
...
+      }
+
+      ctx.env.get_undefined()
+    })?
+    .into_unknown();
+
+  Ok(callback)
+}
+
+pub fn forward_threadsafe_function_to_channel<T: Send>(
```
yamadapc: This does the opposite of what the name says

yamadapc: Are you sure we need this, vs. just calling the function?

crates/node-bindings/src/parcel/parcel.rs
```diff
@@ -41,6 +45,19 @@ impl ParcelNapi {
     })
   }
 
+  #[napi]
+  pub async fn build(&self) -> napi::Result<()> {
+    self.parcel.build().unwrap();
+    Ok(())
+  }
```
alshdavid, line 52: The plan is that the first milestone will be to return an `AssetGraph` which it will hand over to the existing bundler implementation. We will "eat" our way along the bundling phases and eventually this will complete the entire build.

crates/parcel/src/parcel.rs
```diff
@@ -28,4 +30,14 @@ impl Parcel {
...
+
+  pub fn build(&self) -> anyhow::Result<()> {
+    let mut _rpc_connection = None::<RpcConnectionRef>;
+
+    if let Some(rpc_host) = &self.rpc {
+      _rpc_connection = Some(rpc_host.start()?);
+    }
+
+    Ok(())
+  }
```
alshdavid, line 42: This initializes workers only once the build has started. The workers use channels for transferring data so they can be initialized lazily as their messages are buffered

crates/parcel_plugin_rpc/src/nodejs/worker_init.rs
```diff
@@ -0,0 +1,60 @@
...
+
+use crate::RpcConnectionMessage;
+
+enum WorkerInitMessage {
+  Subscribe(Sender<Sender<RpcConnectionMessage>>),
+  Register(Sender<RpcConnectionMessage>),
+}
+
+// This allows for workers to notify RPC Connections that they are ready
+static WORKER_INIT: Lazy<Sender<WorkerInitMessage>> = Lazy::new(|| {
```
yamadapc, line 21: shouldn't be static, but I'm not sure what this is achieving

can we just call the thread-safe function on the Rpc worker? why do we need channels at all if thread-safe function already wraps a queue?

crates/parcel_plugin_rpc/src/nodejs/rpc_conns_nodejs.rs
```diff
@@ -0,0 +1,38 @@
...
+  fn next_index(&self) -> usize {
+    let mut current_index = self.current_index.lock();
+    if *current_index >= self.conns.len() - 1 {
+      *current_index = 0;
+    } else {
+      *current_index = *current_index + 1;
+    }
+    current_index.clone()
+  }
+}
```
alshdavid, line 31: I tried to use an `AtomicUsize` for this initially but I wasted too much time trying to figure out the `fetch_update`. 

I assume we will change this in future to schedule work on workers based on if they are busy so for now round robin with a Mutex<usize> is fine

crates/parcel_plugin_rpc/src/nodejs/rpc_host_nodejs.rs
alshdavid, line 1: The host is essentially a worker factory. Workers must be unique to a build as they will contain some state, like config, loaded plugins and such.

The `RpcHost` will create a new "worker farm" every time you call `RpcHost.start()`

This means that

```javascript
const parcel = new ParcelV3({ nodeWorkers: 2 })

await Promise.all([
  parcel.build(),
  parcel.build(),
])
``` 
Will spawn 4 workers (2 per build) which terminate at the end of their respective builds

crates/parcel_plugin_rpc/src/nodejs/rpc_host_nodejs.rs
```diff
@@ -93,6 +70,17 @@ impl RpcHost for RpcHostNodejs {
...
+
+  fn start(&self) -> anyhow::Result<RpcConnectionRef> {
+    // JavaScript workers will have already been created
+    // This waits for the workers to connect
+    let mut connections = vec![];
+    for _ in 0..self.node_workers {
+      connections.push(RpcConnectionNodejs::new())
+    }
+
+    Ok(Arc::new(RpcConnectionsNodejs::new(connections)))
```
alshdavid: It takes 20ms to spawn 8 workers in parallel or 100ms sequentially. In this case the build does not block on worker start up, workers are initialized in parallel and messages to them are buffered within their respective channels until they are ready.

This allows us to kick off the build before the workers are ready and have them start in the background

crates/parcel_plugin_rpc/src/nodejs/worker_init.rs
alshdavid, line 1: Workers in Javascript run on a seperate threads that don't share memory in Js land. To tell them to do work I need to call their callback functions from the main bundler thread however Workers are initialized in JavaScript making it difficult to pass Rust-sourced channels to them.

Napi has the concept of `External`s, which can be used to pass a channel into JavaScript and back to Rust - however this type cannot be sent to a Worker as it's not serializable. One solution to this is to convert the channel into a raw pointer and pass the pointer to the Nodejs main thread, the Nodejs Worker and back to Rust however that requires the use of `unsafe` to cast the raw pointer back into a channel.

An alternative (the approach I've taken here) is to use a global cross-thread broadcast channel that allows the Rust manager to produce a channel for the Worker and have the Worker pick it up later (after it bootstraps).

It's a bit janky but at least it's not unsafe :shrug:

crates/parcel_plugin_rpc/src/rpc_conn_message.rs
```diff
@@ -0,0 +1,15 @@
...
+  },
+}
+
+impl RpcConnectionMessage {
+  pub fn get_id(&self) -> u32 {
+    match self {
+      Self::Ping { response: _ } => 0,
+    }
+  }
+}
```
alshdavid: adding an index to the enum is useful for casting the event type into JavaScript

packages/core/core/src/index.js
```diff
@@ -18,3 +18,5 @@ export {
   INTERNAL_RESOLVE,
   INTERNAL_TRANSFORM,
 } from './Parcel';
+
+export * from './parcel-v3';
```
alshdavid, line 22: Is this okay? I don't know if it can be conditionally included

packages/core/core/src/parcel-v3/worker/index.js
alshdavid, line 1: This is the entry point for worker threads.

I had to add Flow pre-processing due to Nodejs workers starting in a new JavaScript context. I believe we have a build step to update the cli entry point - I am not sure how to extend that to include this.

Eventually the JavaScript layer will be pretty light, we might actually be able to get away with no build step? Just jsdoc and vanilla JS? :thinking:

crates/parcel_plugin_rpc/src/nodejs/worker_init.rs
alshdavid: See comment above

crates/parcel_plugin_rpc/src/nodejs/napi.rs
alshdavid: Might not be needed, will play around with it

alshdavid: If the threadsafe function is called directly from the `RpcConnection` it runs on the bundler thread and blocks until JS returns a result. If I send multiple messages they will load-balance across the workers however running them on the bundler thread effectively serializes the requests to the workers.

To solve this I have 1 rust thread for 1 js worker thread where I am calling the threadsafe function within its own dedicated thread. This ensures the request/response of the tsfn does not block the caller thread.

This would also be the case with async Rust or rayon, however it would block the entire worker

yamadapc: I think it makes a lot of use-cases a lot easier to just call the function, because in a lot of cases you'll call the method and want to wait for the response. If you need the thread to be unblocked then there can be an async point at the call site, but most of the time I think you'd want to wait.

devongovett:
tbh it seems like an open question to me on whether we even need multiple JS workers for the first version. If we do most work in Rust and JS plugins are fairly rare, we could just run JS plugins on the main thread and not even bother with workers. It does add a bit of complexity, and until we have an end-to-end build running, we won't know if it is worth it performance wise. Might be good to start simple and measure first to see if we need it?

crates/parcel_plugin_rpc/src/nodejs/napi.rs
alshdavid: Even if we use an async runtime and make the call site `async`, the threadsafe function will block the Tokio worker (because it isn't async). 

```rust
impl RpcConnectionNodejs {
  pub async fn ping() {
    self.threadsafe_function.call(()).unwrap() // Will block the Tokio worker
  }
}
```

Wrapping it in a channel + dedicated thread allows us to use async channels and allow the call site to not be blocking.

```rust
impl RpcConnectionNodejs {
  pub async fn ping() {
    let (tx, mut rx) = tokio::oneshot();
    self.tx_threadsafe_function.send(RpcMessage::Ping{ response: tx });
    rx.await
  }
}
```

yamadapc: Right but you can just spawn if you don't want to block

alshdavid:
> tbh it seems like an open question to me on whether we even need multiple JS workers for the first version. If we do most work in Rust and JS plugins are fairly rare, we could just run JS plugins on the main thread and not even bother with workers.

Haha fair point, we initially didn't think we needed them but realized it would be better to ensure the main thread is never blocked on plugin work because that can lead to shutdown issues. Additionally, workers will have some state so it's a good idea to sandbox plugins to a single build invocation to avoid leaking state between multiple builds that are run programmatically - workers are the easiest way to achieve that.

e.g.
```javascript
const parcel = new Parcel()

await parcel.build()
await parcel.build() // avoid sharing worker state between these two builds
``` 

So with that in mind, we think that we will always need at least 1 worker per build. Supporting multiple workers is really just a freebie of the abstraction I'm introducing and isn't drastically more complex than supporting 1 worker.

> It does add a bit of complexity, and until we have an end-to-end build running, we won't know if it is worth it performance wise. Might be good to start simple and measure first to see if we need it?

At the moment Atlassian's use case still heavily relies on lots of JavaScript plugins so I ran a benchmark on a Jira sized project in Mach to determine the performance difference between multiple workers vs running plugins on only the main. I saw a consistent halving of build times with every worker added up until 16 workers (on a 16 core machine).

The addition of a single JS resolver plugin added ~30 seconds to the build, halving with each worker added.

devongovett:
I wonder if there is a way to avoid starting up workers for builds that don't have any JS plugins. They do have a fairly significant startup cost if I remember correctly. Could we start them lazily?

Also does what you said mean that you are restarting the whole worker on every build and not persisting them like we do in v2 today? I wonder how much of a performance impact that has as well.

alshdavid:
> They do have a fairly significant startup cost if I remember correctly.

It seems to take ~20ms to launch 8 of them but when they are launched sequentially it takes 100ms

> Could we start them lazily?

Values are emitted to them over a per-worker channel that is initialized before the worker starts so they are buffered until the worker is ready - I guess that means we don't actually need to wait for a worker to start before beginning bundling.

> Also does what you said mean that you are restarting the whole worker on every build and not persisting them like we do in v2 today? I wonder how much of a performance impact that has as well.

Yes, this is also a quirk of napi and libuv. The workers have a callback used to send events from Rust to the worker to do things like call plugins. That callback cannot be unrefed (otherwise the worker will shut down immediately).

If I start the workers and persist them, Nodejs will assume a long-lived async action is running and the Nodejs process will not shut down. So the workers need to live/die along with their respective command (`build`, `watch` etc).

> I wonder if there is a way to avoid starting up workers for builds that don't have any JS plugins. 

I do this in Mach, before the build starts I look at the plugin config to see if any plugins require the Nodejs RPC layer and only init workers if they are needed for a build. I expect we will do this in Parcel as well

devongovett:
> I look at the plugin config to see if any plugins require the Nodejs RPC layer and only init workers if they are needed for a build. 

Not sure that will be possible with the default parcel config though. It has lots of plugins defined in it that might not actually be used during a build. For example, the coffee script transformer. If you don't have any .coffee files in your project it'll never run but it is still in the config if you need it.

alshdavid:
Hmm, true. That's a good point. 

I designed the `RpcHost` trait to support lazy init by letting the caller run `RpcHost.start()`. That could be triggered by the Rpc plugins on the first invocation of their working methods (like `RpcTransformer.transform()`). That would defer the bootstrapping of the workers until the first time they are actually needed

crates/parcel_plugin_rpc/src/nodejs/napi.rs
alshdavid: > Right but you can just spawn if you don't want to block

This is essentially what I am doing but rather than spawning a thread per invocation, which would require a new OS thread and cloning the tfsn, I put the tsfn in a single long-lived thread and talk to it with a channel.

The other benefit is the channels buffer messages that will be sent to the worker prior to the worker being initialized. This allows for the workers to be initialized lazily.

alshdavid:
I've updated the PR to lazily initialize the workers and removed the threadsafe function channel wrapper from `RpcHostNodejs`

crates/node-bindings/src/parcel/parcel.rs
```diff
@@ -41,6 +45,19 @@ impl ParcelNapi {
...
+
+  #[napi]
+  pub fn default_thread_count(env: Env) -> napi::Result<JsNumber> {
+    let cpus = num_cpus::get();
+    let cpus = env.create_int32(cpus as i32)?;
+    Ok(cpus)
+  }
```
alshdavid, line 59: The nodejs worker count and rust thread count are decoupled from each other.

Parcel will default to spawning the same number of nodejs worker threads as the specified Rust threads. Users can optionally configure more/less Nodejs workers.

```javascript
new Parcel({
  threads: undefined, // defaults to available processors
  nodeWorkers: undefined // defaults to the value of "threads",
})
```

packages/core/core/src/index.js
mattcompiles: Why do we need to export this from core?

packages/core/core/src/parcel-v3/Parcel.js
```diff
@@ -0,0 +1,85 @@
...
+  nodeWorkers?: number,
+  fs?: FileSystem,
+|};
+
+export class ParcelV3 {
+  _internal: napi.ParcelNapi;
+  #nodeWorkerCount: number;
+
+  constructor({
+    threads = napi.ParcelNapi.defaultThreadCount(),
```
mattcompiles, line 19: Wouldn't it be cleaner to do the defaulting internally?

mattcompiles:
@alshdavid What I'm missing here is what is needed to add a new communication point between Rust -> JS. Like which files do I need to modify? A small breakdown of what you'd need to add to make a new one would be cool.

packages/core/core/src/parcel-v3/Parcel.js
alshdavid: The JavaScript class needs access to this value. I could make it a property of the `ParcelNapi` struct and access it from there - mind if I do that in a follow up PR?

mattcompiles: Yeah this isn't blocking or anything

yamadapc (approved):

PR #9789, opened 2024-06-12
Add node.js only parcel watcher watchman back-end

yamadapc:
Added a commit over https://github.com/parcel-bundler/parcel/compare/v2...jondlm:parcel:jdlm-watchman-drop-in-replacement

~~Migrated to typescript but need to write some tests and integrate with the rest of the codebase.~~

~~The ignore patterns section likely is one thing that needs to be changed carefully though we could support only a subset of the current ignore options and delegate to watchman.~~

---

**Update**

Ignore support added. Right now globs and directories are supported.


yamadapc:
cc @jdlm-stripe

package.json
```diff
@@ -8,11 +8,12 @@
   },
   "private": true,
   "workspaces": [
-    "packages/*/*"
+    "packages/*/*",
+    "packages/utils/parcel-watcher-watchman-js"
```
marcins: Doesn't this already get matched by `packages/*/*`?

MonicaOlejniczak: Yeah, looks like it can be removed

package.json
```diff
@@ -25,11 +26,12 @@
     "link-all": "node scripts/link-all.js packages",
     "unlink-all": "node scripts/unlink-all.js packages",
     "check": "flow check",
+    "flow": "lerna run build-ts && flow",
     "lint": "eslint . && prettier \"./packages/*/*/{src,bin,test}/**/*.{js,json,md}\" --list-different && cargo fmt --all -- --check",
     "prepublishOnly": "yarn adjust-versions && yarn build && yarn build-ts",
-    "test:unit": "cross-env NODE_ENV=test mocha --timeout 5000 && cargo test",
-    "test:integration": "yarn workspace @parcel/integration-tests test",
-    "test:integration-ci": "yarn workspace @parcel/integration-tests test-ci",
+    "test:unit": "lerna run build-ts && cross-env NODE_ENV=test mocha --timeout 5000 && cargo test",
```
MonicaOlejniczak: You should be able to remove the explicit `build-ts` (which is quite heavy handed) by modifying the parcel babel preset

yarn.lock
```diff
@@ -2560,6 +2560,13 @@
   resolved "https://registry.yarnpkg.com/@types/color-name/-/color-name-1.1.1.tgz#1c1261bbeaa10a8055bbc5d8ab84b7b2afc846a0"
   integrity sha512-rr+OQyAjxze7GgWrSaJwydHStIhHq2lvY3BOC2Mj7KnzI7XK0Uw1TOOdI9lDoajEbSWLiYgoo4f1R51erQfhPQ==
 
+"@types/fb-watchman@2.0.4":
+  version "2.0.4"
+  resolved "https://packages.atlassian.com/api/npm/npm-remote/@types/fb-watchman/-/fb-watchman-2.0.4.tgz#3a02379aa90d86ceb965a2beaebd7e79faa81e82"
```
MonicaOlejniczak: This needs to be regenerated with the yarnpkg registry. I normally use `yarn install --registry https://registry.yarnpkg.com/`

packages/utils/parcel-watcher-watchman-js/.gitignore
```diff
@@ -0,0 +1 @@
+/lib
```
MonicaOlejniczak: Lib should already be ignored, and if not it should be fixed in the root `.gitignore` file. Or is this specific to testing?

packages/utils/parcel-watcher-watchman-js/package.json
```diff
@@ -0,0 +1,22 @@
+{
+  "name": "@parcel/watcher-watchman-js",
+  "version": "2.12.0",
```
MonicaOlejniczak, line 3: Maybe confirm with @devongovett what version new packages should be on?

package.json
yamadapc: I'd rather shift towards split compile and run lifecycle than the current mix of babel register etc.

It's quite confusing at the moment that there is the "dev" register mode and another mode that is pre-built (hence need the dev-prepare steps to fix-up the repository).

packages/utils/parcel-watcher-watchman-js/package.json
yamadapc: Confirmed this is what the version should be set to

yamadapc:
I'll just make it easy and use JSDoc syntax (?)

alshdavid (approved):

yarn.lock
```diff
@@ -3891,6 +3891,13 @@ browserslist@^4.0.0, browserslist@^4.16.0, browserslist@^4.16.6, browserslist@^4
     node-releases "^2.0.13"
     update-browserslist-db "^1.0.13"
 
+bser@2.1.1:
+  version "2.1.1"
+  resolved "https://registry.npmjs.org/bser/-/bser-2.1.1.tgz#e6787da20ece9d07998533cfd9de6f5c38f4bc05"
```
MonicaOlejniczak: Shouldn't these be from the yarn registry?

packages/utils/parcel-watcher-watchman-js/src/index.js
```diff
@@ -0,0 +1,8 @@
+import {createWrapper} from './wrapper';
+
+const wrapper = createWrapper();
```
MonicaOlejniczak: The function seems a bit unnecessary, why not just inline the `new ParcelWatcherWatchmanJS()` code?

packages/utils/parcel-watcher-watchman-js/src/wrapper.js
```diff
@@ -0,0 +1,181 @@
...
+  ): Promise<void>;
+  writeSnapshot(
+    dir: FilePath,
+    snapshot: FilePath,
+    opts?: Options,
+  ): Promise<FilePath>;
+}
+
+let client: watchman.Client | null = null;
+function getClient() {
```
MonicaOlejniczak: Can't we just init this in the `ParcelWatcherWatchmanJS` constructor and create a helper for the commands?

mattcompiles: I can refactor a bit, honestly I just left this as I inherited it.

yarn.lock
mattcompiles: Weird. Fixed now.

MonicaOlejniczak (approved):

packages/core/fs/package.json
```diff
@@ -48,6 +48,8 @@
     "check-ts": "tsc --noEmit index.d.ts"
   },
   "dependencies": {
+    "@parcel/feature-flags": "2.12.0",
+    "@parcel/watcher-watchman-js": "2.12.0",
```
devongovett: How can we make this dependency opt-in? Peer dep? try...catch around a require?

mattcompiles: The package is so small (less than 50KB) that I wasn't going to worry about it. I can try make it opt in though.

PR #9791, opened 2024-06-14
Minor transformer plugin changes

yamadapc:
Minor changes to transformer API in preparation for https://github.com/parcel-bundler/parcel/tree/issue/rust-transformer

This removes unnecessary methods from transformer trait & makes minor changes so we get a context type passed in, separate to an input type. This context type will be used to interact with the request API.

### DIFF
--- a/crates/parcel_core/src/plugin/transformer_plugin.rs
+++ b/crates/parcel_core/src/plugin/transformer_plugin.rs
@@ -1,14 +1,13 @@
+use parcel_filesystem::FileSystemRef;
 use std::fmt::Debug;
 use std::fs::File;
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
+use std::rc::Rc;
 
 use parcel_resolver::SpecifierType;
 
-use super::PluginConfig;
-use crate::types::Asset;
-use crate::types::SourceMap;
-
-pub struct AST {}
+use crate::types::{Asset, SourceCode};
+use crate::types::{Dependency, SourceMap};
 
 pub struct GenerateOutput {
   pub content: File,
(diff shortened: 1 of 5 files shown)


crates/parcel_core/src/plugin/transformer_plugin.rs
```diff
@@ -26,103 +25,70 @@ pub struct ResolveOptions {
...
-  ///
-  /// This function is called when the next transformer AST cannot be reused, or this is the last
-  /// transformer in a pipeline.
-  ///
-  fn generate(&self, asset: Asset, ast: AST) -> Result<GenerateOutput, anyhow::Error>;
+/// * File-paths that have just been discovered
+/// * Outputs of previous transformation steps, which are in-place modified
+/// * These two scenarios are distinguished
+pub enum TransformationInput {
+  FilePath(PathBuf),
```
mattcompiles, line 34: There's other info we'll need outside of the file path, e.g. sideEffects, isSource, env. Anything that's specific to this transform or derived from resolution. We can add this as need be, but maybe we should name the type clearer?

crates/parcel_core/src/plugin/transformer_plugin.rs
```diff
@@ -26,103 +25,70 @@ pub struct ResolveOptions {
...
+  Asset(Asset),
```
mattcompiles, line 35: I like the idea of separating the actual Asset from the transform inputs as some stuff isn't needed by the plugins. But we're still using the main Asset type here. Maybe we should have 3 asset types: 
- Pre all transforms
- Post a single transform
- A finalised Asset

crates/parcel_core/src/types/asset.rs
```diff
@@ -76,12 +101,13 @@ pub struct Asset {
   /// This can be used to find assets during packaging, or to create dependencies between multiple
   /// assets returned by a transformer by using the unique key as the dependency specifier.
   ///
+  /// TODO: Make this non-nullable and disallow creating assets without it.
```
mattcompiles, line 104: If we drop asset groups (which I think we should) then we might be able to remove this entirely?

crates/parcel_core/src/plugin/transformer_plugin.rs
yamadapc: We can add that as we actually consume it

crates/parcel_core/src/types/asset.rs
yamadapc: We should remove this

crates/parcel_core/src/plugin/transformer_plugin.rs
yamadapc: We can add that as we actually consume it

crates/parcel_core/src/plugin/transformer_plugin.rs
```diff
@@ -26,103 +25,70 @@ pub struct ResolveOptions {
 /// A function that enables transformers to resolve a dependency specifier
 pub type Resolve = dyn Fn(PathBuf, String, ResolveOptions) -> Result<PathBuf, anyhow::Error>;
 
-/// Compile a single asset, discover dependencies, or convert the asset to a different format
```
MonicaOlejniczak, line 29: /nit The first line of docs should summarise the intent and only be on a single line if it later appears in docs

crates/parcel_core/src/plugin/transformer_plugin.rs
```diff
@@ -26,103 +25,70 @@ pub struct ResolveOptions {
...
-///
-/// Many transformers are wrappers around other tools such as compilers and preprocessors, and are
-/// designed to integrate with Parcel.
+/// Transformers may run against:
```
MonicaOlejniczak, line 28: /nit The first line of docs should summarise the intent and only be on a single line if it later appears in docs

crates/parcel_core/src/plugin/transformer_plugin.rs
```diff
@@ -26,103 +25,70 @@ pub struct ResolveOptions {
...
+}
 
-    transformers.push(Box::new(TestTransformerPlugin {}));
+#[derive(Debug)]
+pub struct TransformResult {
+  pub asset: Asset,
+  pub dependencies: Vec<Dependency>,
+  /// The transformer signals through this field that its result should be invalidated
+  /// if these paths change.
+  pub invalidate_on_file_change: Vec<PathBuf>,
```
MonicaOlejniczak, line 79: I personally find the existing invalidations API tedious to work with and don't think it should be modelled like this long-term. There is a common `Invalidation` type that the request tracker works off (though it is currently a stub), which plugins could return and it would enable all types of invalidations. Whether we want to scope them down per plugin is a different question though. Anyways, just food for thought.

mattcompiles (approved):

crates/parcel_core/src/types/asset.rs
```diff
@@ -92,11 +118,42 @@ impl Asset {
...
     hasher.finish()
   }
+
+  /// Build a new empty asset
+  pub fn new_empty(file_path: PathBuf, source_code: Rc<SourceCode>) -> Self {
+    let asset_type =
+      FileType::from_extension(file_path.extension().and_then(|s| s.to_str()).unwrap_or(""));
+
+    // TODO: rest of this
+    Self {
```
MonicaOlejniczak, line 128: /nit If you derive from default, you can make this shorter by using `..Asset::default()`

MonicaOlejniczak (approved):

crates/parcel_core/src/plugin/transformer_plugin.rs
yamadapc: But that's just an empty enum?

I see two ways to model this:

1. Dynamic - we have a method on "Run...Context" that pushes this out. This is the approach on current `RequestAPI`

* Advantage, incremental, works for error types

```
context.invalidate_on_file_change(...)
```
2. Static - we have a return value ; this is what is written here, and is the same as `enum Invalidation`

* Advantage, clear typing, less exit points

---

I'm not sure which is better, but think this is quite easy to change as things take form.

crates/parcel_core/src/plugin/transformer_plugin.rs
devongovett: shouldn't the input of a transformer always be an Asset? Otherwise transformers would need to be aware of whether they are running first or later in a pipeline. In v2 transformers always receive an asset as input, and core constructs the initial asset to pass to the first transformer.

crates/parcel_core/src/plugin/transformer_plugin.rs
devongovett: I expanded this on the `core-rs2` branch like this: https://github.com/parcel-bundler/parcel/blob/103a063b8f526690cf9eb514b75c9da1ff96960a/crates/core/src/request_tracker.rs#L113-L122

I like the idea of just returning them as part of plugin results rather than mutating an input.

PR #9792, opened 2024-06-14
Replace AHash with xxhash_rust

yamadapc:
Use stable hashing algorithm for stored identifiers; add comment explaining why this is.

```
/// Parcel needs to use a hasher for generating certain identifiers used in caches.
///
/// The hashes don't need to be incredibly fast, but they should be stable across
/// runs, machines, platforms and versions.
///
/// These hashes will likely end-up being written to disk, either within output
/// JavaScript files or internal caches.
```

Stacked over https://github.com/parcel-bundler/parcel/pull/9791

### DIFF
--- a/crates/parcel_core/src/plugin/transformer_plugin.rs
+++ b/crates/parcel_core/src/plugin/transformer_plugin.rs
@@ -1,14 +1,13 @@
+use parcel_filesystem::FileSystemRef;
 use std::fmt::Debug;
 use std::fs::File;
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
+use std::rc::Rc;
 
 use parcel_resolver::SpecifierType;
 
-use super::PluginConfig;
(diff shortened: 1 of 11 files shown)


crates/parcel_core/src/plugin/transformer_plugin.rs
```diff
@@ -1,14 +1,13 @@
+use parcel_filesystem::FileSystemRef;
 use std::fmt::Debug;
 use std::fs::File;
-use std::path::PathBuf;
```
MonicaOlejniczak: This shouldn't be in this pull request

yamadapc: this is stacked ; I haven't written these changes manually, I'm just running RustRover organize imports

MonicaOlejniczak (approved):

PR #9795, opened 2024-06-14
Improve JS transformer documentation

yamadapc:
Write documentation for hoisting struct.

### DIFF
--- a/packages/transformers/js/core/src/hoist.rs
+++ b/packages/transformers/js/core/src/hoist.rs
@@ -220,7 +220,7 @@ pub struct HoistResult {
   pub exported_symbols: Vec<ExportedSymbol>,
   /// Symbols re-exported from other modules.
   ///
-  /// If a symbol is re-exported from another module, parcel will remove delete the export statement
+  /// If a symbol is re-exported from another module, parcel will remove the export statement
   /// from the asset.
   ///
   /// For example, if an input file is:
--- a/packages/transformers/js/core/src/hoist.rs
+++ b/packages/transformers/js/core/src/hoist.rs
@@ -119,6 +119,7 @@ struct Hoist<'a> {
   /// See [`HoistResult::exported_symbols`]
   exported_symbols: Vec<ExportedSymbol>,
   re_exports: Vec<ImportedSymbol>,
+  /// See [`HoistResult::self_references`]
   self_references: HashSet<JsWord>,
   /// See [`HoistResult::dynamic_imports`]
   dynamic_imports: HashMap<JsWord, JsWord>,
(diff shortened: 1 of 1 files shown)


packages/transformers/js/core/src/hoist.rs
```diff
@@ -60,51 +60,266 @@ pub fn hoist(
...
+  /// var $abc$export$6a5cdcad01c973fa;
+  /// var $abc$export$ce14ccb78c97a7d4;
+  /// $abc$export$6a5cdcad01c973fa = 10;
+  /// $abc$export$ce14ccb78c97a7d4 = function() {
+  ///     return $abc$export$6a5cdcad01c973fa;
+  /// };
+  /// ```
+  pub self_references: HashSet<JsWord>,
+  /// TODO: What is this?
+  pub wrapped_requires: HashSet<String>,
```
mischnic, line 280: A non-top-level `require` that needs to be wrapped. A top-level require is anything matching
```js
const/var/let ... = require("something");
```
so the standard CJS import pattern.

packages/transformers/js/core/src/hoist.rs
```diff
@@ -60,51 +60,266 @@ pub fn hoist(
...
+  ///         exported: "something",
+  ///         local: "$abc$export$hashfashdfashdfahsdfh_something",
+  ///         ...
+  ///     },
+  /// ]
+  /// ```
+  pub exported_symbols: Vec<ExportedSymbol>,
+  /// Symbols re-exported from other modules.
+  ///
+  /// If a symbol is re-exported from another module, parcel will remove delete the export statement
```
mischnic:
```suggestion
  /// If a symbol is re-exported from another module, parcel will remove the export statement
```

mischnic (approved):

packages/transformers/js/core/src/hoist.rs
yamadapc: So is this the mangled variable put to replace the `require(...)` statement?

mischnic: No, that should be handled with the other maps (`imports`).
This is to mark the dependencies with the specifier listed in `wrapped_requires` with `dep.meta.shouldWrap = true`, so that the packager ensure that the require is not executed immediately (as for ES6 `import "...";`).

yamadapc: Got it, so this would be the `source` string?

In `require('something')`, it'd be `'something'`?

mischnic: Exactly

PR #9797, opened 2024-06-17
Clean-up dfsFast feature-flag

yamadapc:
Clean-up feature-flag controlling DFS refactor from recursive to iterative implementation. No significant performance difference has been observed in production.

### DIFF
--- a/packages/core/graph/src/Graph.js
+++ b/packages/core/graph/src/Graph.js
@@ -3,7 +3,6 @@
 import {fromNodeId} from './types';
 import AdjacencyList, {type SerializedAdjacencyList} from './AdjacencyList';
 import type {Edge, NodeId} from './types';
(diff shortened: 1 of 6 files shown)


packages/core/graph/src/Graph.js
```diff
@@ -506,7 +500,7 @@ export default class Graph<TNode, TEdgeType: number = 1> {
    *
    * This replaces `dfs` and will replace `dfsFast`.
```
marcins, line 500: What will be the approach for replacing `dfsFast`? Is it just changing the call sites? Will that require a separate flag / testing to ensure functionality is equivalent and performance doesn't regress?

marcins (approved):

packages/core/graph/src/Graph.js
yamadapc: We will need another flag & a bit more code

PR #9798, opened 2024-06-17
V3: Added default thread count and build options

alshdavid:
- Moved setting the default `nodeWorkers` option to Rust
- Setting default Rust `threads` count to be available CPUs
- Setting default `node_worker_count` to be the value set for the Rust `threads`
- Updated `build` to run asynchronously without blocking + parsing the options/return values
- Added `BuildOptions` and `BuildResult` to `Parcel.build()`

### DIFF
--- a/crates/node-bindings/src/parcel/parcel.rs
+++ b/crates/node-bindings/src/parcel/parcel.rs
@@ -1,11 +1,13 @@
 use std::path::PathBuf;
 use std::sync::Arc;
+use std::thread;
 
 use napi::Env;
-use napi::JsNumber;
 use napi::JsObject;
+use napi::JsUnknown;
 use napi_derive::napi;
 use parcel::rpc::nodejs::RpcHostNodejs;
+use parcel::BuildOptions;
 use parcel::Parcel;
 use parcel::ParcelOptions;
 
@@ -13,6 +15,7 @@ use crate::file_system::FileSystemNapi;
 
 #[napi]
 pub struct ParcelNapi {
+  pub node_worker_count: u32,
   parcel: Arc<Parcel>,
 }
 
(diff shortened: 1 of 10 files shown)


crates/node-bindings/src/parcel/parcel.rs
```diff
@@ -25,30 +37,86 @@ impl ParcelNapi {
...
+        &callback,
+        0,
+        |ctx: ThreadSafeCallContext<anyhow::Result<BuildResult>>| match ctx.value {
+          Ok(_) => Ok(vec![ctx.env.create_object()]),
+          Err(err) => Err(napi::Error::from_reason(format!("{:?}", err))),
+        },
+      )?;
+
+    // Asynchronously wait for a response from the build
+    env.spawn_future(async move {
```
devongovett: I don't think you need tokio or thread safe function callbacks for this. You can use `env.create_deferred` to create a promise, and return that.

https://github.com/parcel-bundler/parcel/blob/c03a29ede215967a9ec4e472a1161eed0a913af6/crates/node-bindings/src/core.rs#L128-L137

crates/node-bindings/src/parcel/parcel.rs
```diff
@@ -25,30 +37,86 @@ impl ParcelNapi {
...
+    // Parcel Core Options
+    let mut parcel_options = ParcelOptions::default();
+
     // Wrap the JavaScript-supplied FileSystem
-    let fs = FileSystemNapi::from_options(&env, &options)?;
+    parcel_options.fs = FileSystemNapi::from_options(&env, &options)?;
+
+    // Assign Rust thread count from JavaScript
+    let js_threads = options.get_named_property::<JsUnknown>("threads")?;
+    parcel_options.threads = match js_threads.get_type()? {
```
devongovett: Does this need to be configurable? Can we just use rayon's default?

alshdavid: In the past I have found it helpful to control the number of threads for debugging and benchmarking. You can pass this number to Rayon, Tokio when configuring them. I think it's fine to have?

crates/node-bindings/src/parcel/parcel.rs
alshdavid: Nice, will use the deferred API!

alshdavid: Updated

crates/node-bindings/src/parcel/parcel.rs
```diff
@@ -25,37 +28,58 @@ impl ParcelNapi {
...
 
+    // Parcel Core Options
+    let mut parcel_options = ParcelOptions::default();
+
     // Wrap the JavaScript-supplied FileSystem
-    let fs = FileSystemNapi::from_options(&env, &options)?;
+    parcel_options.fs = FileSystemNapi::from_options(&env, &options)?;
+
+    // Assign Rust thread count from JavaScript
+    let js_threads = options.get_named_property::<JsUnknown>("threads")?;
```
yamadapc: why don't we type `options` statically?

alshdavid: Are we able to pass these properties in as options? I tried this initially but it failed to compile
```rust
#[napi]
pub struct ParcelNapiOptions {
  fs: Option<FileSystemNapi>,
  rpc: JsFunction,
}
```

yamadapc: Yes you should be able to pass `JsFunction` and `Option<JsObject>`

yamadapc: then `options.fs.map(FileSystemNapi::new)`

alshdavid: Updated

yamadapc (approved):

PR #9799, opened 2024-06-17
Alsh: Moved the FileSystem to use the RPC host

alshdavid:
This will be refactored, it's a bit messy and needs a bit of work but the concept will remain the same

- Added `RPCFileSystem`
  - This can be reused for wasm, etc, as long as the `RpcHost` is implemented for that context
- Removed `FileSystemNapi`


packages/core/core/src/parcel-v3/Parcel.js
```diff
@@ -44,6 +47,15 @@ export class ParcelV3 {
       // Ping
       case 0:
         return;
+      case 1:
```
yamadapc, line 50: Consider whether this is strictly better than the multi-threaded func. delegation on `file_system_napi.rs` ; 

One is centralized in one file, while the other can interface with any JS object.

packages/core/core/src/parcel-v3/Parcel.js
```diff
@@ -14,12 +14,15 @@ export type ParcelV3Options = {|
 export class ParcelV3 {
   _internal: napi.ParcelNapi;
   #nodeWorkerCount: number;
+  #fs: FileSystem | void;
```
yamadapc, line 17: Don't make it nullable if it's invalid for it to be missing

crates/parcel_plugin_rpc/src/filesystem/filesystem.rs
```diff
@@ -0,0 +1,44 @@
...
+}
+
+impl RpcFileSystem {
+  pub fn new(rpc_host: RpcHostRef) -> Self {
+    Self { rpc_host }
+  }
+}
+
+impl FileSystem for RpcFileSystem {
+  fn read_to_string(&self, path: &Path) -> io::Result<String> {
```
yamadapc, line 18: We needed to make all these fallible with anyhow errors

crates/parcel_plugin_rpc/src/filesystem/filesystem.rs
```diff
@@ -0,0 +1,44 @@
...
+        }
+      }
+    }
+  }
+
+  fn is_file(&self, path: &Path) -> bool {
+    self
+      .rpc_host
+      .fs_is_file(path)
+      .expect("Un expected panic in RPC FS operation")
```
MonicaOlejniczak, line 35: This should capture the original error, otherwise it will be difficult to debug later on

packages/core/core/src/parcel-v3/Parcel.js
MonicaOlejniczak: Yeah, I'm finding these changes hard to grok

crates/parcel_plugin_rpc/src/nodejs/rpc_host_nodejs.rs
```diff
@@ -90,4 +106,31 @@ impl RpcHost for RpcHostNodejs {
 
     Ok(Arc::new(RpcConnectionNodejsMulti::new(connections)))
   }
+
+  fn fs_read_to_string(&self, path: &Path) -> anyhow::Result<String> {
```
MonicaOlejniczak, line 110: Can these be internal to the file system, so that some of the logic is a bit more self-contained? This would require `call_rpc` being public on the trait, rather than these specific messages. I wonder if a bit more could be done so that the rpc host doesn't become monolithic too 🤔

alshdavid: I totally agree, I've been thinking about this all day and I ended up coming up with [this solution](https://github.com/parcel-bundler/parcel/pull/9802).

Rather than using an enum to describe all messages (which will end up monolithic because of all the mappings needed) - I updated the `RpcHost` to just have an 
```rust
rpc_host.send<R: Deserialize>(
  event_name: string, 
  params: Serialize,
) -> R
``` 
method backed by `serde` that consumers can use to send values to/from JavaScript without the need for explicit type casting.

The idea is the `FileSystemRpc`, `CacheRpc`, etc, can take the `rpc_host` as a dep and call the `send` method directly, internalizing the details of its own messages.

[Example usage](https://github.com/parcel-bundler/parcel/pull/9802/files#diff-196ca92d27b1b31f5a31b178551598f095ea1bc91d12c2abca57f77deef6b211R41)  

For more advanced cases there is a method to manually map params/results at the call site that looks something like 
```rust
rpc_host.send_with<R>(
  event_name: string,
  map_params: Fn(&Env) -> napi::Result<JsUnknown>,
  map_return: Fn(&Env) -> napi::Result<R>
) -> R
```

alshdavid:
Will come back to this

PR #8927, opened 2023-04-03
Fixup collect.used_imports and dev symbol translation

mischnic:
Instead of cleared symbols on the (currently) non-analyzed imports/requires, add `*` instead. This is what scopehoisting does as a fallback as well. (Because cleared dependencies are slightly broken: https://github.com/parcel-bundler/parcel/pull/8928)

Also fix a bug where `symbol_result.imports` didn't list a named import when it was used in a member expression:
```js
import { b } from "foo";
b.run();
```

PR #9800, opened 2024-06-17
Rust only JS transformer plug-in

yamadapc:
Adds transformer plug-in pulled from @devongovett's work on https://github.com/parcel-bundler/parcel/compare/v2...core-rs2

I've tried massaging the source but haven't gone fully through it.

Since we're missing a few fields still on `asset`, `dependency` there are some minor parts commented out.

I've questions marked as `TODO`. I think with better understanding it'll be easier to write tests and break this up further.


crates/parcel_plugin_transformer_js/src/transformer.rs
```diff
@@ -1,24 +1,894 @@
-use parcel_core::plugin::PluginContext;
```
yamadapc, line 1: The bulk of the changes

crates/parcel_plugin_transformer_js/src/transformer.rs
```diff
@@ -1,24 +1,894 @@
...
+        dependency.symbols.push(symbol);
+      }
+    }
+
+    for symbol in hoist_result.re_exports {
+      if let Some(dependency) = dependency_by_specifier.get_mut(&symbol.source) {
+        if is_re_export_all_symbol(&symbol) {
+          let loc = Some(convert_loc(asset_file_path.clone(), &symbol.loc));
+          dependency.symbols.push(make_export_all_symbol(loc));
+          // TODO: Why isn't this added to the asset.symbols array?
```
yamadapc, line 143: Why isn't this added to the `asset.symbols` array?

crates/parcel_plugin_transformer_js/src/transformer.rs
```diff
@@ -1,24 +1,894 @@
...
+
+    for specifier in hoist_result.wrapped_requires {
+      if let Some(dep) = dependency_by_specifier.get_mut(&swc_core::atoms::JsWord::new(specifier)) {
+        dep.flags.set(DependencyFlags::SHOULD_WRAP, true);
+      }
+    }
+
+    // for (name, specifier) in hoist_result.dynamic_imports {
+    //   if let Some(dep) = dependency_by_specifier.get_mut(&specifier) {
+    //     dep.promise_symbol = Some((&*name).into());
```
yamadapc, line 182: What is `promise_symbol`? I suppose this is the mangled name inserted to replace `import(...)`?

crates/parcel_plugin_transformer_js/src/transformer.rs
```diff
@@ -1,24 +1,894 @@
...
+    }
+
+    // For all other imports and requires, mark everything as imported (this covers both dynamic
+    // imports and non-top-level requires.)
+    for dep in dependency_by_specifier.values_mut() {
+      if dep.symbols.is_empty() {
+        dep.symbols.push(Symbol {
+          exported: "*".into(),
+          // TODO: What does it mean?
+          local: "".into(), // format!("${}$", dep.placeholder.as_ref().unwrap_or(&dep.specifier)).into(),
```
yamadapc: What is this?

crates/parcel_plugin_transformer_js/src/transformer.rs
```diff
@@ -1,24 +1,894 @@
...
+            .iter()
+            .find(|candidate| candidate.exported == &*symbol.imported);
+          let existing_flags = existing
+            .map(|e| e.flags & SymbolFlags::IS_WEAK)
+            .unwrap_or(SymbolFlags::IS_WEAK);
+
+          // TODO: What is this? There is no such mangling in the transformer
+          let re_export_name = existing
+            .map(|sym| sym.local.clone())
+            .unwrap_or_else(|| format!("${:016x}$re_export${}", asset_id, symbol.local).into());
```
yamadapc, line 159: What happened here? Because this symbol is not in the transformer.

crates/parcel_plugin_transformer_js/src/transformer.rs
```diff
@@ -1,24 +1,894 @@
...
+    // import_attributes: Vec::new(),
+    pipeline: None,
+    meta: JSONObject::new(),
+    // resolver_meta: JSONObject::new(),
+    package_conditions: ExportsCondition::empty(),
+    // custom_package_conditions: Vec::new(),
+    // TODO:
+    is_entry: false,
+    needs_stable_name: false,
+    is_optional: false,
```
yamadapc: I reckon these fields are deprecated by `AssetFlags` and such

crates/parcel/src/request_tracker/request_tracker.rs
```diff
@@ -27,13 +27,13 @@ impl<T: Clone> RequestTracker<T> {
     }
   }
 
-  pub fn run_request(&mut self, request: Box<&dyn Request<T>>) -> anyhow::Result<T> {
+  pub fn run_request(&mut self, request: &impl Request<T>) -> anyhow::Result<T> {
```
yamadapc: There's no need to box this until we store it

Cargo.lock
```diff
@@ -3154,6 +3162,18 @@ dependencies = [
  "vergen",
 ]
 
+[[package]]
```
MonicaOlejniczak: We should probably try to have only one version

crates/parcel/src/requests/asset_graph_request.rs
```diff
@@ -0,0 +1,51 @@
...
+  todo!()
+}
+
+/// This is part of the asset request ultimately
+fn run_transformer_pipeline(
+  transformers: Vec<Box<dyn TransformerPlugin>>,
+  asset: Asset,
+) -> anyhow::Result<()> {
+  let mut input = TransformationInput::Asset(asset);
+  let mut transform_context = RunTransformContext::new(Arc::new(OsFileSystem::default()));
```
MonicaOlejniczak: This should probably have a todo to use the input fs later

crates/parcel_config/src/map/named_pipelines_map.rs
```diff
@@ -40,14 +40,14 @@ pub struct NamedPattern<'a> {
 /// ```
 ///
 #[derive(Debug, Default, Deserialize, PartialEq, Serialize)]
-pub struct NamedPipelinesMap(
+pub struct NamedPipelinesMap {
   /// Maps patterns and named patterns to a series of plugins, called pipelines
-  IndexMap<String, Vec<PluginNode>>,
-);
+  inner: IndexMap<String, Vec<PluginNode>>,
```
MonicaOlejniczak, line 48: Won't this need to be flattened in serde? Or is it automatic with `inner`?

crates/parcel_core/src/types/asset.rs
```diff
@@ -127,29 +133,40 @@ impl Asset {
...
+}
+
+// TODO: All of these should have documentation
+bitflags! {
+  #[derive(Debug, Clone, Copy, Default, PartialEq)]
+  pub struct AssetFlags: u32 {
+    const IS_SOURCE = 1 << 0;
+    // replaced with `Asset::side_effects` for now
+    // const SIDE_EFFECTS = 1 << 1;
+    const IS_BUNDLE_SPLITTABLE = 1 << 2;
```
MonicaOlejniczak: `is_bundle_splittable` should be removed from the asset struct since it's now a flag

crates/parcel_core/src/types/asset.rs
```diff
@@ -127,29 +133,40 @@ impl Asset {
...
+  pub fn set_interpreter(&mut self, shebang: impl Into<serde_json::Value>) {
+    self.meta.insert("interpreter".into(), shebang.into());
+  }
+}
+
+// TODO: All of these should have documentation
+bitflags! {
+  #[derive(Debug, Clone, Copy, Default, PartialEq)]
+  pub struct AssetFlags: u32 {
+    const IS_SOURCE = 1 << 0;
```
MonicaOlejniczak: `is_source` should be removed from the asset struct since it's now a flag

crates/parcel/src/requests/asset_graph_request.rs
```diff
@@ -0,0 +1,51 @@
+use crate::request_tracker::{Request, RequestResult, RunRequestContext, RunRequestError};
```
yamadapc: I can remove this

crates/parcel_core/src/types/dependency.rs
```diff
@@ -142,6 +152,21 @@ impl Hash for Dependency {
   }
 }
 
+bitflags! {
+  #[derive(Default, Debug, Clone, Copy, Hash, PartialEq)]
+  pub struct DependencyFlags: u8 {
+    const ENTRY    = 1 << 0;
```
MonicaOlejniczak: `is_entry` `is_optional` `needs_stable_name` should be removed from the parent asset struct since they are now a flag

crates/parcel_plugin_transformer_js/src/lib.rs
```diff
@@ -1,4 +1,4 @@
-#![deny(unused_crate_dependencies)]
+// #![deny(unused_crate_dependencies)]
```
MonicaOlejniczak: Just remove the line

yamadapc: - [x] Revert

packages/transformers/js/core/src/utils.rs
```diff
@@ -289,6 +289,12 @@ pub enum SourceType {
   Module,
 }
 
+impl Default for SourceType {
```
MonicaOlejniczak, line 292: Default should be module as it's part of environment which is the default there in JS, you can also import this from core types if you remove the `repr` representation

crates/parcel_plugin_transformer_js/src/transformer.rs
```diff
@@ -1,24 +1,947 @@
...
+    )?;
+
+    // TODO handle errors properly
+    if let Some(errors) = transformation_result.diagnostics {
+      return Err(anyhow!(format!("{:#?}", errors)));
+    }
+
+    let asset = Asset::new_empty(input.file_path().to_path_buf(), source_code);
+    let config = parcel_js_swc_core::Config::default();
+    let options = ParcelOptions::default();
```
MonicaOlejniczak, line 70: This or just `core_path` should be initialised in new

crates/parcel_plugin_transformer_js/src/transformer.rs
```diff
@@ -1,24 +1,947 @@
...
+}
+
+fn convert_result(
+  mut asset: Asset,
+  transformer_config: &parcel_js_swc_core::Config,
+  result: parcel_js_swc_core::TransformResult,
+  options: &ParcelOptions,
+) -> Result<TransformResult, Vec<Diagnostic>> {
+  let asset_file_path = asset.file_path.to_path_buf();
+  let asset_environment = asset.env.clone();
```
MonicaOlejniczak, line 86: We should probably optimise the target / env structures soon to minimise storage

crates/parcel_plugin_transformer_js/src/transformer.rs
```diff
@@ -1,24 +1,947 @@
...
+      &asset_file_path,
+      asset_environment,
+      dependency_flags,
+      asset_id,
+    );
+    dependency_by_specifier.insert(dependency.specifier.as_str().into(), dependency);
+  }
+
+  if let Some(hoist_result) = result.hoist_result {
+    asset.flags.set(AssetFlags::HAS_SYMBOLS, true);
```
MonicaOlejniczak: Is `has_symbols` needed? I find it weird, why can't we check for the `symbols` len?

crates/parcel_plugin_transformer_js/src/transformer.rs
MonicaOlejniczak: Yeah they are

crates/parcel_plugin_transformer_js/src/transformer.rs
```diff
@@ -1,24 +1,947 @@
...
+}
+
+fn make_esm_helpers_dependency(
+  options: &ParcelOptions,
+  asset_file_path: &PathBuf,
+  asset_environment: Environment,
+  dependency_flags: DependencyFlags,
+  asset_id: u64,
+) -> Dependency {
+  Dependency {
```
MonicaOlejniczak, line 446: Can default a lot of these for less LOC, especially the None values

crates/parcel_plugin_transformer_js/src/transformer.rs
MonicaOlejniczak: Yeah they are

packages/transformers/js/core/src/utils.rs
yamadapc: Cool, great

crates/parcel_plugin_transformer_js/src/transformer.rs
devongovett: There is actually a difference between a module having no exports and an asset not having analyzed the symbols at all. In the latter case, we must assume that any import to that module exists rather than throwing a "does not export" error.

What we should probably do is make `asset.symbols` an `Option<Vec<Symbol>>` instead. This flag is a holdover from the old `db2` branch.

crates/parcel_plugin_transformer_js/src/transformer.rs
devongovett: This represents an `export * from '...'` statement. There can be multiple of those in the same file so they need to be on each dependency rather than on the asset.

The representation of symbols really needs to change though tbh. At the moment this is mostly done for compatibility with the representation in JS. There, we actually use a `Map` rather than an array for symbols so there cannot be multiple symbols of the same name. That eventually needs to change.

crates/parcel_plugin_transformer_js/src/transformer.rs
devongovett: This represents an `export {foo} from '...'` statement. In that case there is no "local" symbol used in the source code, the asset only passes through the exports and does not use them itself. So the synthetic local name here is only so the asset symbol and dependency symbol can be matched up.

Another case of a non-ideal symbol representation that we should change eventually but we need for compatibility with JS packager for now.

crates/parcel_plugin_transformer_js/src/transformer.rs
devongovett: https://github.com/parcel-bundler/parcel/pull/8927

crates/parcel/src/requests/asset_graph_request.rs
yamadapc: I'm removing this file as it was for just so I understood things slightly better.

crates/parcel_plugin_transformer_js/src/transformer.rs
yamadapc: Wouldn't all re-exports have the same issue? Is it easier to just collect on both vectors so we get consistent outputs?

crates/parcel_plugin_transformer_js/src/transformer.rs
yamadapc: Sorry, I mean that I don't understand what it means for `Symbol::local` to be an empty string.

Is it supposed to be like that or just in a partial state?

crates/parcel_plugin_transformer_js/src/transformer.rs
yamadapc: I guess why don't we do this inside the SWC transformer though?

crates/parcel_config/src/map/named_pipelines_map.rs
yamadapc: Good call-out!

I'll add `#[serde(transparent)]`

crates/parcel_plugin_transformer_js/src/transformer.rs
devongovett: oh I don't think it should be empty. shouldn't it be the commented out value? Looks like that's what I had here: https://github.com/parcel-bundler/parcel/blob/c03a29ede215967a9ec4e472a1161eed0a913af6/crates/core/src/transformers/js_transformer.rs#L866

crates/parcel_plugin_transformer_js/src/transformer.rs
devongovett: You could try changing the JS implementation here and running the tests to see what breaks: https://github.com/parcel-bundler/parcel/blob/c03a29ede215967a9ec4e472a1161eed0a913af6/packages/transformers/js/src/JSTransformer.js#L905

I suspect it's because there could be both an `export *` dependency _and_ a namespace (`*`) symbol defined within the asset. e.g.

```js
export * from './foo.js';

exports[someNonStaticName] = 2;
```

In this case, the module is not statically analyzable, so it has a `*` namespace symbol in the asset. However it also exports * from foo.js. Since the JS implementation uses a map to represent symbols, these would conflict if they both defined the `*` symbol on the asset.

We can probably change the representation of symbols later to fix this but for now I went with compatibility with the JS implementation (which we need for the packager etc.)

crates/parcel_plugin_transformer_js/src/transformer.rs
devongovett: probably could be, I just ported the JS version https://github.com/parcel-bundler/parcel/blob/c03a29ede215967a9ec4e472a1161eed0a913af6/packages/transformers/js/src/JSTransformer.js#L907-L911

crates/parcel_plugin_transformer_js/src/transformer.rs
yamadapc: I suppose this should be in `context` ;

yamadapc: Won't fix in this diff

crates/parcel_plugin_transformer_js/src/transformer.rs
yamadapc: Won't fix in this diff but yes

crates/parcel_plugin_transformer_js/src/transformer.rs
yamadapc: I've documented but won't fix in this diff

crates/parcel_core/src/types/dependency.rs
yamadapc: I've decided to inline the flags onto the struct for now.

crates/parcel/src/requests/asset_request.rs
```diff
@@ -141,7 +142,7 @@ fn run_pipeline(
     if is_different_asset_type {
       let next_pipeline = plugins.transformers(transform_input.file_path(), None)?;
 
-      if next_pipeline != pipeline {
+      if next_pipeline.hash() != pipeline_hash {
```
mattcompiles, line 145: Why did you make this change?

mattcompiles (approved):

crates/parcel/src/requests/asset_request.rs
yamadapc: Because the plug-in trait needs to take a mutable ref. (to enable mutability further one)

With partial-eq, we need to keep the entire `pipeline` as a read ref. so it can't be mutable as internally in the equality check it could have access to all the data within the `pipeline` object. 

If we externalise the check we're able to make this requirement looser.

PR #9803, opened 2024-06-17
Improve constant module bundling error message

marcins:
# ↪️ Pull Request

Improves the error message when a constant module cannot be found during packaging.

Before:
```
@parcel/packager-js: Constant module not found in bundle
```

After:
```
@parcel/packager-js: Constant module ../platform/packages/analytics/analytics-gas-types/src/index.ts referenced from
../platform/packages/analytics/analytics-listeners/src/fabric/process-event-payload.tsx not found in bundle
shared~vendor~atlaskit.HASH_REF_bb77a59655d528f0.js
```

### DIFF
--- a/packages/packagers/js/src/ScopeHoistingPackager.js
+++ b/packages/packagers/js/src/ScopeHoistingPackager.js
@@ -924,7 +924,13 @@ ${code}
     if (resolved.meta.isConstantModule) {
       invariant(
         this.bundle.hasAsset(resolved),
-        'Constant module not found in bundle',
+        `Constant module ${path.relative(
+          this.options.projectRoot,
+          resolved.filePath,
+        )} referenced from ${path.relative(
+          this.options.projectRoot,
+          parentAsset.filePath,
+        )} not found in bundle ${this.bundle.name}`,
       );
       return false;
     }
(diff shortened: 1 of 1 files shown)


packages/packagers/js/src/ScopeHoistingPackager.js
```diff
@@ -924,7 +924,13 @@ ${code}
     if (resolved.meta.isConstantModule) {
       invariant(
         this.bundle.hasAsset(resolved),
-        'Constant module not found in bundle',
+        `Constant module ${path.relative(
```
mattcompiles: This is probably fine as it only runs for constant modules, however I'd normally favour a `if(){throw err}` here as it means we only generate the error message when it's going to be thrown.

mattcompiles (approved):

packages/packagers/js/src/ScopeHoistingPackager.js
marcins: Yeah, that's fair - ideally `message` could be a `() => string` function that's only invoked if the invariant fails (but then I guess the function is still created?)

Easy enough to change, I'll swap it over.

mattcompiles (approved):

MonicaOlejniczak (approved):

PR #9808, opened 2024-06-18
Parcel V3 - Native asset request

mattcompiles:
This PR adds the initial logic for a native AssetRequest. Highly inspired by @devongovett's [core-rs2 branch](https://github.com/parcel-bundler/parcel/blob/core-rs2/crates/core/src/requests/asset_request.rs). 

I haven't added any tests as yet and I'm a little unsure about whether they're worth it. Unlike the path request, this request needs access to the entire plugins struct which is currently painful to mock/simulate. Happy to be guided here if peeps have bright ideas to solve this. 

There's still a bunch of TODO's left to resolve. We can address them now through the PR comments or come back to them later. I think the code will be easier to revisit once the transformers are in a working place.


crates/parcel_core/src/types/asset.rs
```diff
@@ -150,9 +155,20 @@ impl Asset {
   }
 }
 
+impl Hash for Asset {
```
yamadapc: Are we unable to derive?

crates/parcel/src/requests/asset_request.rs
```diff
@@ -0,0 +1,168 @@
...
+}
+
+impl<'a> Request<AssetResult> for AssetRequest<'a> {
+  fn run(
+    &self,
+    request_context: RunRequestContext<AssetResult>,
+  ) -> Result<RequestResult<AssetResult>, RunRequestError> {
+    request_context.report(ReporterEvent::BuildProgress(BuildProgressEvent::Building(
+      AssetBuildEvent {
+        // TODO: Should we try avoid a clone here?
```
yamadapc, line 67: Generally it'll be fine to clone.

If we find either copying memory or, more likely, memory allocation are a bottleneck, we can optimise easily for all read-only types using references or interning, but without having to spread lifetimes throughout.

crates/parcel_core/src/plugin/transformer_plugin.rs
```diff
@@ -2,45 +2,59 @@ use parcel_filesystem::FileSystemRef;
...
   pub fn file_path(&self) -> &Path {
     match self {
-      TransformationInput::FilePath(path) => path,
+      TransformationInput::InitialAsset(raw_asset) => raw_asset.file_path.as_path(),
       TransformationInput::Asset(asset) => asset.file_path(),
     }
   }
 
-  pub fn read_source_code(&self, fs: FileSystemRef) -> anyhow::Result<Rc<SourceCode>> {
+  pub fn read_source_code(self, fs: FileSystemRef) -> anyhow::Result<Rc<SourceCode>> {
```
yamadapc: why does it consume self? you'd want to store `Option<Rc<SourceCode>>` in the initial asset to prevent consuming things here.

yamadapc (approved):

crates/parcel_core/src/plugin/transformer_plugin.rs
mattcompiles: This was a typo actually I think 😅

crates/parcel/src/plugins.rs
```diff
@@ -175,14 +182,36 @@ impl<'a> Plugins<'a> {
...
+    })
   }
 
   pub fn validators(&self, _path: &Path) -> Result<Vec<Box<dyn ValidatorPlugin>>, anyhow::Error> {
     todo!()
   }
 }
 
+pub struct TransformerPipeline {
+  pub transformers: Vec<Box<dyn TransformerPlugin>>,
```
devongovett, line 197: Still think this could just be a list of plugin names from the config that could be "loaded" on demand as they are run.

crates/parcel/src/requests/asset_request.rs
devongovett: you'll want interning, it makes a big difference.

crates/parcel/src/requests/asset_request.rs
```diff
@@ -0,0 +1,168 @@
...
+        .file_path
+        .extension()
+        .and_then(|s| s.to_str())
+        .unwrap_or(""),
+    );
+    let mut transform_ctx = RunTransformContext::new(self.file_system.clone());
+
+    let result = run_pipeline(
+      pipeline,
+      TransformationInput::InitialAsset(InitialAsset {
```
devongovett, line 86: Can the input to a transformer always be an asset? Ideally transformers shouldn't know or care that they are first in a pipeline

crates/parcel/src/requests/asset_request.rs
```diff
@@ -0,0 +1,168 @@
...
+      &self.plugins,
+      &mut transform_ctx,
+    )?;
+
+    // Write the Asset source code to the cache, this is read later in packaging
+    // TODO: Clarify the correct content key
+    let content_key = result.asset.id().to_string();
+    self
+      .cache
+      .set_blob(&content_key, result.asset.source_code.bytes())?;
```
devongovett: `source_code` is confusing since it's not source anymore it's compiled.

crates/parcel/src/requests/asset_request.rs
```diff
@@ -0,0 +1,168 @@
...
+pub struct AssetRequest<'a> {
+  pub env: Arc<Environment>,
+  pub file_path: PathBuf,
+  pub code: Option<String>,
+  pub pipeline: Option<String>,
+  pub side_effects: bool,
+  // TODO: move the following to RunRequestContext
+  pub cache: CacheRef,
+  pub file_system: FileSystemRef,
+  pub plugins: Arc<Plugins<'a>>,
```
devongovett, line 40: I think this isn't quite ideal because it represents the entire config? So if you change the `.parcelrc` but only the `packagers` section for example, it'll invalidate all of the asset requests too. In my branch I only pass through a reference to the transformers section to avoid this.

crates/parcel/src/requests/asset_request.rs
```diff
@@ -0,0 +1,168 @@
...
+
+impl<'a> Hash for AssetRequest<'a> {
+  fn hash<H: Hasher>(&self, state: &mut H) {
+    // TODO: Just derive this once the contextual params are moved to RunRequestContext
+    self.file_path.hash(state);
+    self.code.hash(state);
+    self.pipeline.hash(state);
+    self.env.hash(state);
+    self.side_effects.hash(state);
+  }
```
devongovett, line 51: Missing plugins? Should invalidate the request when the config changes.

crates/parcel/src/requests/asset_request.rs
mattcompiles: We were thinking that plugins would be tracked via invalidations rather than via the request id/hash. This is why I haven't added them to the hash

crates/parcel/src/requests/asset_request.rs
mattcompiles: The idea here is to abstract this away via methods, but definitely happy to iterate on this design.

crates/parcel/src/requests/asset_request.rs
mattcompiles: Agreed, renamed to `Code`

crates/parcel/src/requests/asset_request.rs
mattcompiles: Agreed, we will address this later either with invalidations or by including the transformers in the hash.

crates/parcel/src/requests/asset_request.rs
mattcompiles: Will address interning in a later PR

crates/parcel_core/src/types/asset.rs
mattcompiles: I removed this as we no longer need it to be hashable

PR #9809, opened 2024-06-19
Port target request to Rust

MonicaOlejniczak:
# ↪️ Pull Request

These changes add the bulk of the target request implementation, but do not yet handle invalidations; initial options like serve mode; browserslist config file; etc. Most of the logic is the same, with notable changes being:
* `PluginConfig` has been renamed to `ConfigLoader` so that the target request can read and parse the `package.json` file in a consistent manner. The `load_package_json_config` function has been modified so that the file is deserialized with serde given the generic parameter.  This is more flexible, and if the previous API is needed for JavaScript plugins at a later point in time it can be added then.
* Context is not inferred for builtin targets when consumers do not provide one, for simplicity and to better align with official standards for the fields
 
## 🚨 Test instructions

### DIFF
--- a/crates/parcel/src/requests/target_request.rs
+++ b/crates/parcel/src/requests/target_request.rs
@@ -1,0 +1,1112 @@
+use std::collections::HashMap;
+use std::ffi::OsStr;
+use std::hash::Hash;
+use std::path::Path;
+use std::path::PathBuf;
+
+use anyhow::anyhow;
+use package_json::BrowserField;
+use package_json::BrowsersList;
+use package_json::BuiltInTargetDescriptor;
+use package_json::ModuleFormat;
+use package_json::PackageJson;
+use package_json::SourceMapField;
+use package_json::TargetDescriptor;
+use parcel_core::config_loader::ConfigLoader;
+use parcel_core::types::engines::Engines;
+use parcel_core::types::BuildMode;
+use parcel_core::types::DefaultTargetOptions;
+use parcel_core::types::Entry;
+use parcel_core::types::Environment;
+use parcel_core::types::EnvironmentContext;
+use parcel_core::types::OutputFormat;
+use parcel_core::types::SourceType;
+use parcel_core::types::Target;
+use parcel_core::types::TargetSourceMapOptions;
+use parcel_resolver::IncludeNodeModules;
+
+use crate::request_tracker::Request;
+use crate::request_tracker::RequestResult;
+use crate::request_tracker::RunRequestContext;
+use crate::request_tracker::RunRequestError;
+
+mod package_json;
+
+/// Infers how and where source code is outputted
+///
+/// Targets will be generated from the project package.json file and input Parcel options.
+///
+pub struct TargetRequest {
+  // TODO Either pass in package.json directly or make config available on the req context
+  pub config: ConfigLoader,
+  pub default_target_options: DefaultTargetOptions,
+  pub env: Option<HashMap<String, String>>,
+  pub exclusive_target: Option<String>,
+  pub mode: BuildMode,
+}
+
+impl Hash for TargetRequest {
+  fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+    self.default_target_options.hash(state);
+    self.exclusive_target.hash(state);
+    self.mode.hash(state);
+  }
+}
+
+#[derive(Debug, Clone, PartialEq)]
+pub struct Targets(Vec<Target>);
+
+struct BuiltInTarget<'a> {
+  descriptor: BuiltInTargetDescriptor,
+  dist: Option<PathBuf>,
+  extensions: Vec<&'a str>,
+  name: &'a str,
+}
+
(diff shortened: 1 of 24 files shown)


crates/parcel/src/requests/target_request.rs
```diff
@@ -0,0 +1,1107 @@
...
+use parcel_resolver::IncludeNodeModules;
+
+use crate::request_tracker::Request;
+use crate::request_tracker::RequestResult;
+use crate::request_tracker::RunRequestContext;
+use crate::request_tracker::RunRequestError;
+
+mod package_json;
+
+pub struct TargetRequest {
```
yamadapc, line 39: can we write down what is target request

crates/parcel/src/requests/target_request.rs
```diff
@@ -0,0 +1,1107 @@
...
+  pub config: ConfigLoader,
+  pub exclusive_target: Option<String>,
+  pub options: ParcelOptions,
+}
+
+impl Hash for TargetRequest {
+  fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
```
yamadapc, line 49: why can't you derive?

crates/parcel/src/requests/target_request.rs
```diff
@@ -0,0 +1,1107 @@
...
+  }
+
+  fn default_dist_dir(&self, package_path: &Path) -> PathBuf {
+    package_path
+      .parent()
+      .unwrap_or_else(|| &package_path)
+      .join("dist")
+  }
+
+  fn fallback_output_format(&self, context: EnvironmentContext) -> OutputFormat {
```
yamadapc: we should prefer free functions for things that don't take `self`

crates/parcel/src/requests/target_request.rs
```diff
@@ -0,0 +1,1107 @@
...
+        }
+      }
+    }
+
+    Ok(inferred_output_format)
+  }
+
+  fn load_package_json(&self) -> Result<(PathBuf, PackageJson), anyhow::Error> {
+    // TODO Invalidations
+    let (package_path, mut package_json) = self.config.load_package_json_config::<PackageJson>()?;
```
yamadapc, line 231: IMO anything that does any reading should be coming either as an input to the request or it should be done through that context parameter.

That is because we will be able to implement read-tracking through the context parameter to not have to track invalidations manually

crates/parcel/src/requests/target_request.rs
MonicaOlejniczak: imo this is personal preference, but can change it 🤷

crates/parcel/src/requests/target_request.rs
MonicaOlejniczak: Sure

crates/parcel/src/requests/target_request.rs
MonicaOlejniczak: Because config is not hashable, I'd like to clean up how this is passed in but at a later time for all the requests

crates/parcel/src/requests/target_request.rs
MonicaOlejniczak: Although I agree in premise, I think we should do a bit of sparring on this before proceeding and apply it to all requests. TargetRequest also does some processing on the pacakge.json file that is specific to this code, so that should be factored in.

crates/parcel/src/requests/target_request.rs
yamadapc: More or less ; it is a bit of a style preference, but there are many advantages to having free functions - https://www.youtube.com/watch?v=WLDT1lDOsb4

* On lots of languages, non-free function denies some optimisations (not sure if in rust)
* To test the function you do not need to create self
* ...

crates/parcel_plugin_transformer_js/src/transformer.rs
```diff
@@ -764,7 +764,6 @@ mod test {
     let asset_1 = Asset::new_empty("mock_path".into(), source_code.clone());
     let asset_2 = Asset::new_empty("mock_path".into(), source_code);
     // This nº should not change across runs/compilation
-    assert_eq!(asset_1.id(), 4127533076662631483);
```
yamadapc, line 101: What happened here? This needs to be stable

yamadapc (approved):

crates/parcel_plugin_transformer_js/src/transformer.rs
MonicaOlejniczak: It's not that value for me locally which is why I removed it, and the next line has stability between ids of the same content

```
thread 'transformer::test::test_asset_id_is_stable' panicked at crates/parcel_plugin_transformer_js/src/transformer.rs:767:5:
assertion `left == right` failed
  left: 17693125794769022406
  right: 4127533076662631483
```

MonicaOlejniczak: Ah it's because I changed env, so the hashed id changed

PR #9811, opened 2024-06-19
Add simple reporting to request tracker

MonicaOlejniczak:
# ↪️ Pull Request

These changes fill out the `report` fn on the request context using a naive implementation to flesh out the empty placeholder. Later on we should decide if we need the equivalent of the `ReporterRunner` from JavaScript, and determine if all / certain events should be run sync / async.

## 🚨 Test instructions

### DIFF
--- a/crates/parcel/src/request_tracker/request_tracker.rs
+++ b/crates/parcel/src/request_tracker/request_tracker.rs
@@ -1,5 +1,8 @@
 use anyhow::anyhow;
+use parcel_core::plugin::ReporterEvent;
+use parcel_core::plugin::ReporterPlugin;
 use std::collections::HashMap;
+use std::sync::Arc;
 
 use petgraph::graph::NodeIndex;
 use petgraph::stable_graph::StableDiGraph;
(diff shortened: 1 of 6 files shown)


crates/parcel/src/request_tracker/request_tracker.rs
```diff
@@ -14,19 +17,36 @@ use super::RunRequestError;
...
       request_index: HashMap::new(),
     }
   }
 
+  pub fn report(&self, event: ReporterEvent) {
+    let event = Arc::new(event);
+
+    for reporter in self.reporters.iter() {
+      // TODO Decide how to handle errors
+      let _ = reporter.report(Arc::clone(&event));
```
yamadapc: We should just pass the event by reference instead of using `Arc`

`Arc` is for multi-threaded scenarios ; for single-threaded reference counting we can use `Rc`

But for this use-case we can be passing the events by reference into the reporters.

MonicaOlejniczak: I was trying to factor in multi-threading use-case from the get go, but realised I can just do that anyway later with a borrow to reference so I will fix it shortly.

yamadapc (approved):

PR #9813, opened 2024-06-20
V3: Added JsCallable

alshdavid:
Added `JsCallable` utility to simplify the usage of napi JavaScript functions from Rust. 

These must be run in their own thread. This works with async functions, sync functions, and promises.

Please leave your thoughts and comments 🙏

### Usage

```rust
#[napi]
pub fn example(callback: JsFunction) -> napi::Result<()> {
  let callable = JsCallable::new(callback)?;

  thread::spawn(move || {
    let result = callable.call_with_return_serde::<bool>(42)?;
    println!("From thread {:?}", result); // "true"
  });

  Ok(())
}
```

Or with the `FileSystemNapi` wrapper

```rust
pub struct FileSystemNapi {
  read_file_fn: JsCallable,
}

impl FileSystemNapi {
  pub fn new(js_file_system: &JsObject) -> napi::Result<Self> {
    Ok(Self {
      read_file_fn: JsCallable::new_from_object_prop("readFileSync", &js_file_system)?,
    })
  }
}

impl FileSystem for FileSystemNapi {
  fn read_to_string(&self, path: &Path) -> std::io::Result<String> {
    self
      .read_file_fn
      .call_with_return_serde((path.to_path_buf(), "utf8"))
      .map_err(|e| std::io::Error::other(e))
  }
}
```

### Before / After hooks

The `call` and `call_with_return` methods take mapping functions to let you cast Rust types to napi types and back. This is useful if you need to supply a callback as a parameter or do some custom deserialization logic (often needed for enums)

```rust
/*
callable.call_with_return(
  map_params_fn,
  map_return_fn,
)
*/

let value_to_send = 42;

let result = callable.call_with_return(
  move |env| { 
    // Runs before the target function to send values into JS
    let v = env.create_int32(value_to_send)?.into_unknown();
    Ok(vec![v])
  },
  move |env, returned| {
    // Runs after target function to return values back
    let value = env.from_js_value::<bool, _>(returned)?;
    Ok(value)
  }
)?;
```

### Mappers

Included are helper functions `map_params_serde` and `map_return_serde` which handle the type conversion automatically using serde

```rust
// Two params
callable.call_with_return(
  map_params_serde((42, "hello")),
  map_return_serde::<bool>(),
)

// One param
callable.call_with_return(
  map_params_serde("hello),
  map_return_serde::<bool>(),
)

// No params
callable.call_with_return(
  map_params_serde(()),
  map_return_serde::<bool>(),
)
```


crates/node-bindings/src/file_system/file_system_napi.rs
```diff
@@ -1,94 +1,47 @@
...
   fn is_dir(&self, path: &Path) -> bool {
-    (*self.is_dir_fn)(path.to_path_buf())
+    self
+      .is_dir_fn
+      .call_with_return(path.to_path_buf())
+      .expect("TODO handle error case")
   }
 }
-
-fn create_js_thread_safe_method<
```
yamadapc, line 62: The idea was to use:
```
/// ## Safety
/// This function will do unsafe casting of a raw env pointer in order to access it
/// from the thread-safe function block. This is safe because even though the NAPI API
/// does not expose env there, that block runs on the main JS thread.
fn create_async_js_thread_safe_method<
  Params: Send + Serialize + 'static,
  Response: Send + DeserializeOwned + 'static + FromNapiValue,
>(
  env: &Env,
  js_object: &JsObject,
  method_name: &str,
) -> Result<impl (Fn(Params) -> Response) + Send, napi::Error> {
  let jsfn: JsFunction = js_object.get_property(env.create_string(method_name)?)?;

  let threadsafe_function = env.create_threadsafe_function(
    &jsfn,
    0,
    |ctx: napi::threadsafe_function::ThreadSafeCallContext<Params>| {
      Ok(vec![ctx.env.to_js_value(&ctx.value)?])
    },
  )?;

  let result = move |params| {
    let (tx, rx) = std::sync::mpsc::sync_channel(1);
    threadsafe_function.call_with_return_value(
      Ok(params),
      ThreadsafeFunctionCallMode::Blocking,
      move |result: CustomUnknownWeAreToSmart| {
        let env = result.env;
        // This is safe because this block runs on the main-thread.
        let env = unsafe { napi::Env::from_raw(env) };
        let promise = Promise::<Result>::from_js_unknown(result.value);
        let _ = env.spawn_future(async move {
            let result = result.await?;
            let _ = tx.send(result);
            Ok(())
        })?;
        Ok(())
      },
    );
    rx.recv().unwrap()
  };

  Ok(result)
}
```

+  the hack to get env passed in
```
struct CustomUnknownWeAreTooSmart {
   env: napi_env,
   value: JsUnknown
}

impl FromNapiValue for CustomUnknownWeAreTooSmart {
  unsafe fn from_napi_value(env: sys::napi_env, napi_val: sys::napi_value) -> Result<Self> {
    Ok(Self {
       env: env,
       value: JsUnknown::from_napi_value(env, napi_val),
    })
  }
}
```

crates/node-bindings/src/helpers/js_callable/js_callable.rs
```diff
@@ -0,0 +1,288 @@
...
+  callback: usize,
+}
+
+impl JsCallable {
+  pub fn new(callback: JsFunction) -> napi::Result<Self> {
+    let initial_thread = std::thread::current().id();
+
+    // Store the threadsafe function on the struct
+    let tsfn: ThreadsafeFunction<Box<dyn ErasedSerialize>, ErrorStrategy::Fatal> = callback
+      .create_threadsafe_function(0, |ctx: ThreadSafeCallContext<Box<dyn ErasedSerialize>>| {
```
yamadapc: You don't need to box it ; or use type erasure if the type is moved up as it was

crates/node-bindings/src/helpers/js_callable/js_callable.rs
```diff
@@ -0,0 +1,288 @@
...
+        let result = ctx.env.to_js_value(&ctx.value)?;
+        if result.is_array()? {
+          // SAFETY: type assertion above
+          let result = panic::catch_unwind::<_, napi::Result<Array>>(|| unsafe {
```
yamadapc: You don't need unsafe ; use `napi::Array::from_unknown`

crates/node-bindings/src/helpers/js_callable/js_callable.rs
```diff
@@ -0,0 +1,288 @@
...
+
+    let env = value.1;
+    let callback = value.cast::<JsFunction>()?;
+
+    let mut args = vec![];
+
+    let result = env.to_js_value(&params)?;
+    if result.is_array()? {
+      // SAFETY: type assertion above
+      let result = panic::catch_unwind::<_, napi::Result<Array>>(|| unsafe {
```
yamadapc: there's no need to use unsafe to do this

crates/node-bindings/src/helpers/js_callable/js_callable.rs
```diff
@@ -0,0 +1,288 @@
...
+use napi::JsFunction;
+use napi::JsObject;
+use napi::JsUnknown;
+use napi::NapiRaw;
+use serde::de::DeserializeOwned;
+use serde::Serialize;
+
+thread_local! {
+  /// Storage for napi JavaScript functions on the local thread
+  static LOCAL_FUNCTIONS: (RefCell<usize>, RefCell<HashMap<usize, Rc<JsValue>>>) = Default::default();
```
yamadapc: I think everything looks pretty good, except for the unsafe & type-erasure this introduces.

But IMO it's much much simpler to just have 2 forms of delegation:

* One is using multi-threaded function & delegates using the `create_js_thread_safe_method` and `create_async_js_thread_safe_method`, which are combined around 40 lines of code with no unsafe
* The other can use normal delegation - https://github.com/parcel-bundler/parcel/blob/25b7f904fa7511b0a88b7c90252d4d66d51014cf/crates/parcel_filesystem/src/js_delegate_file_system.rs#L44

If we found that there is a need to write an abstraction that works for both cases, we can use it.

yamadapc: Also we chatted re. `env` not being safe to store across calls;

The contents of this hashmap are thread-local copies of the functions and env pointer.

My understanding is this needs to be `Ref` because these values may be GCed across calls.

yamadapc (changes requested):

crates/node-bindings/src/helpers/js_callable/js_callable.rs
```diff
@@ -0,0 +1,288 @@
...
+        }
+        Ok(())
+      },
+    );
+
+    Ok(())
+  }
+
+  /// Unwrap promise return value. Cannot be run on the main thread
+  pub fn await_promise<Response: Send + DeserializeOwned + 'static>(
```
yamadapc: This looks great to avoid async/await :+1:

crates/node-bindings/src/helpers/js_callable/js_callable.rs
alshdavid: Fixed

crates/node-bindings/src/helpers/js_callable/js_callable.rs
alshdavid: Updated

crates/node-bindings/src/helpers/js_callable/js_callable.rs
alshdavid: > If we found that there is a need to write an abstraction that works for both cases, we can use it.

Perhaps you're right about this. We will probably never call a function from the main JavaScript thread

> My understanding is this needs to be Ref because these values may be GCed across calls.

Good point, I'll double check this

crates/node-bindings/src/helpers/js_callable/js_value.rs
```diff
@@ -0,0 +1,34 @@
...
+use napi::JsUnknown;
+use napi::NapiRaw;
+use napi::NapiValue;
+use std::panic;
+
+pub struct JsValue(pub JsUnknown, pub Env);
+
+impl JsValue {
+  pub fn cast<T: NapiValue>(&self) -> napi::Result<T> {
+    if let Ok(result) = panic::catch_unwind::<_, T>(|| unsafe { self.0.cast() }) {
```
yamadapc: there is `T::from_unknown` which isn't unsafe

crates/node-bindings/src/helpers/js_callable/local_functions.rs
```diff
@@ -0,0 +1,47 @@
...
+use std::rc::Rc;
+
+use super::JsValue;
+use napi::bindgen_prelude::FromNapiValue;
+use napi::Env;
+use napi::JsFunction;
+
+thread_local! {
+  /// Storage for napi JavaScript functions on the local thread
+  static LOCAL_FUNCTIONS: (RefCell<usize>, RefCell<HashMap<usize, Rc<JsValue>>>) = Default::default();
```
yamadapc: V8 may GC this value we can't hold it;

In order to keep it we need to use - https://napi.rs/docs/compat-mode/concepts/ref

crates/node-bindings/src/helpers/js_callable/serde.rs
```diff
@@ -0,0 +1,34 @@
+use napi::bindgen_prelude::Array;
+use napi::bindgen_prelude::FromNapiValue;
+use napi::Env;
+use napi::JsUnknown;
+use serde::de::DeserializeOwned;
+use serde::Serialize;
+
+pub fn map_params_serde<Params: Serialize + Send + Sync + 'static>(
+  params: Params,
+) -> Box<dyn FnOnce(&Env) -> napi::Result<Vec<JsUnknown>> + Send + 'static> {
```
yamadapc: Why do we return `Vec<...>` instead of `JsUnknown`?

crates/node-bindings/src/helpers/js_callable/js_value.rs
```diff
@@ -0,0 +1,34 @@
...
+
+impl FromNapiValue for JsValue {
+  unsafe fn from_napi_value(env: napi_env, value: napi_value) -> napi::Result<Self> {
+    let value = JsUnknown::from_napi_value(env, value)?;
+    let env = Env::from_raw(env);
+    Ok(Self(value, env))
+  }
+}
+
+impl NapiRaw for JsValue {
```
yamadapc: This is unused

crates/node-bindings/src/helpers/js_callable/js_callable.rs
```diff
@@ -0,0 +1,163 @@
...
+    map_return: impl Fn(&Env, JsUnknown) -> napi::Result<Return> + Send + 'static,
+  ) -> napi::Result<Return> {
+    if self.initial_thread == std::thread::current().id() {
+      self.call_local_with_return(map_params, map_return)
+    } else {
+      self.call_thread_with_return(map_params, map_return)
+    }
+  }
+
+  pub fn call_local(
```
yamadapc: I don't think we have any use-cases that require calling JS threads from Rust that runs on the main JavaScript thread.

crates/node-bindings/src/file_system/file_system_napi.rs
```diff
@@ -1,94 +1,52 @@
...
+use crate::helpers::js_callable::map_return_serde;
+use crate::helpers::js_callable::JsCallable;
 
 // TODO error handling
 
 pub struct FileSystemNapi {
-  read_file_fn: Box<dyn Fn((PathBuf, String)) -> String + Send + Sync>,
-  is_file_fn: Box<dyn Fn(PathBuf) -> bool + Send + Sync>,
-  is_dir_fn: Box<dyn Fn(PathBuf) -> bool + Send + Sync>,
+  read_file_fn: JsCallable,
```
yamadapc, line 11: Can you move the result/param generics back to the variable itself;

crates/node-bindings/src/helpers/js_callable/local_functions.rs
alshdavid: Removed

crates/node-bindings/src/helpers/js_callable/js_callable.rs
alshdavid: Removed

crates/node-bindings/src/file_system/file_system_napi.rs
alshdavid: I tried this initially but chose not to for 3 reasons:

1. JsFunctions are overloadable, so it's more accurate for the type signature to be defined at the call site
2. Having mapping functions for before/after function invocation better represents the behaviour we want. Rather than opting into serde for everything, we can choose serde or manually map types - which is desirable in some cases (like converting enums)
3. Functions that take multiple arguments do so in the form of `Vec<JsUnknown>` (`NapiValue` but whatever). 

This can be problematic for a generic as I need to constrain the type to be iterable but also include tuples.

```rust
// T = impl Serialize

JsCallable<usize>
// This translates to: 
//    function(arg0: number): void

JsCallable<(usize, String)>
// This translates to: 
//    function(arg0: [number, string]): void
```

```rust
// T = Iterator<Item = impl Serialize>

JsCallable<Vec<usize>>
// This translates to: 
//    function(...args: number[]): void

JsCallable<(usize, String)> // This is not valid
```
I couldn't figure this out, so I just left the typing at the function call

yamadapc: I don't think
```
JsCallable<(usize, String)>
```

Translates to `[number, string]` ; we are already keeping the generic at the construction site on the current `create_js_thread_safe_method` implementation

crates/node-bindings/src/helpers/js_callable/js_value.rs
```diff
@@ -0,0 +1,27 @@
...
+use napi::Env;
+use napi::JsUnknown;
+use napi::NapiValue;
+use std::panic;
+
+pub struct JsValue(pub JsUnknown, pub Env);
+
+impl JsValue {
+  pub fn cast<T: NapiValue>(&self) -> napi::Result<T> {
+    if let Ok(result) = panic::catch_unwind(|| unsafe { self.0.cast() }) {
```
yamadapc: Please do not use unsafe

yamadapc (changes requested):

crates/node-bindings/src/file_system/file_system_napi.rs
alshdavid: Added a comment, please confirm

crates/node-bindings/src/helpers/js_callable/js_callable.rs
alshdavid: Removed local functions

crates/node-bindings/src/helpers/js_callable/serde.rs
alshdavid: To describe multiple arguments for the target JavaScript function. This is how the Napi API works so it just maps directly to it

crates/node-bindings/src/helpers/js_callable/js_value.rs
alshdavid: Removed

crates/node-bindings/src/helpers/js_callable/js_value.rs
alshdavid: Removed

crates/node-bindings/src/file_system/file_system_napi.rs
alshdavid: That's okay, wanted to explain my reasoning. Can I mark this comment as resolved?

yamadapc (approved):

crates/node-bindings/src/file_system/file_system_napi.rs
yamadapc: I'm resolving even though I think it'd be slightly better to move the generic here:

* 2 less allocations per call
* actually less code

PR #9814, opened 2024-06-20
Fix dev packager global reference

MonicaOlejniczak:
# ↪️ Pull Request

When running parcel without scope hoisting, `global` references are not passed through correctly in the `DevPackager` module. This results in runtime errors when any code directly uses `global.` without any checks of its existence since `var global = arguments[3]` is always undefined. These changes amend the pass through of the global object, so that `arguments[3]` exists under a privately namespaced variable to prevent collisions with other constants named global.

## 🚨 Test instructions

`yarn test:integration`

### DIFF
--- a/packages/packagers/js/src/DevPackager.js
+++ b/packages/packagers/js/src/DevPackager.js
@@ -116,7 +116,7 @@ export class DevPackager {
         let output = code || '';
         wrapped +=
           JSON.stringify(this.bundleGraph.getAssetPublicId(asset)) +
-          ':[function(require,module,exports) {\n' +
+          ':[function(require,module,exports,__globalThis) {\n' +
           output +
           '\n},';
         wrapped += JSON.stringify(deps);
--- a/packages/core/integration-tests/test/globals.js
+++ b/packages/core/integration-tests/test/globals.js
@@ -1,8 +1,10 @@
 import assert from 'assert';
 import path from 'path';
-import {bundle, run} from '@parcel/test-utils';
 
-describe('global alias', function () {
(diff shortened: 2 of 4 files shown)


mattcompiles (approved):

packages/packagers/js/src/DevPackager.js
```diff
@@ -116,7 +116,7 @@ export class DevPackager {
         let output = code || '';
         wrapped +=
           JSON.stringify(this.bundleGraph.getAssetPublicId(asset)) +
-          ':[function(require,module,exports) {\n' +
+          ':[function(require,module,exports,__globalThis) {\n' +
```
mischnic, line 119: Where will that `__globalThis` be referenced?

MonicaOlejniczak: It is referenced [here](https://github.com/parcel-bundler/parcel/blob/v2/packages/transformers/js/core/src/global_replacer.rs#L139-L154)

I should have added it to the description, my bad :)

devongovett: that's referenced via `arguments[3]` though. Wouldn't this mean that if someone declared `let __globalThis` in their code they'd get an error stating that it is already defined? Doesn't look like this variable is used directly.

MonicaOlejniczak: Yes, that's correct. I opted for `__globalThis` since `__global` is more likely to be defined. I'd prefer to use `globalThis` directly instead of specifying an argument since it's well supported, but it would be inconsistent with the parcel runtime.

PR #9815, opened 2024-06-20
Refactor transformer

yamadapc:
Refactor rust transformer implementation into a few files and add tests/documentation.


crates/parcel_plugin_transformer_js/src/transformer/conversion/symbol.rs
```diff
@@ -0,0 +1,170 @@
+use std::path::Path;
+
+use parcel_core::types::Symbol;
+
+use crate::transformer::conversion::loc::convert_loc;
+
+/// Convert `CollectImportedSymbol`, `ImportedSymbol` and into `Symbol`
+macro_rules! convert_symbol {
```
mattcompiles, line 8: Curious as to why this is a macro and not a function?

yamadapc: Because the input types are different so for it to be a function I'd have to introduce a trait, implement, generic etc etc etc

yamadapc: I don't love it;

Maybe it's better to repeat the code

mattcompiles: No that's fine and makes sense. Was just asking to learn.

mattcompiles (approved):

PR #9817, opened 2024-06-21
Clean-up some APIs and add tracing subscriber into log file

yamadapc:
Adding a tracing subscriber that will write to a temporary file. This will skip reporting system for lower level logs as we can simplify internal development logs by using the `tracing` ecosystem.


crates/parcel_core/src/plugin/reporter_plugin/composite_reporter_plugin.rs
```diff
@@ -0,0 +1,47 @@
...
+
+use crate::plugin::{ReporterEvent, ReporterPlugin};
+
+/// A reporter plugin that delegates to multiple other reporters.
+#[derive(Debug)]
+pub struct CompositeReporterPlugin {
+  reporters: Vec<Box<dyn ReporterPlugin>>,
+}
+
+impl CompositeReporterPlugin {
```
MonicaOlejniczak, line 11: I like this concept 👍

MonicaOlejniczak (approved):

devongovett:
What does this trace?

yamadapc:
Nothing because we don't have working builds.
We can instrument things with tracing as we progress, as well as log information.

PR #9821, opened 2024-06-25
Request tracker multi threading

mattcompiles:
# ↪️ Pull Request

This PR adds makes all requests run within the rayon threadpool and adds a new request tracker API called `queue_request`. This API is definitely not finalised but it's hopefully a step in the right direction. It's inspired by @devongovett's work on the `core-rs2` branch, but attempts to internalise the threading to the request tracker which allows us to:
- Keep the request handling behaviour centralised to the request tracker instead of leaking to the request impl
- Share the threading implementation with other requests in future

We had to make a lot of changes to get this working so I'm opening this up for feedback but I'm definitely happy to simplify, update stuff or move some stuff out to separate PRs. It was just getting so big that we wanted to share the progress sooner rather than later. 

Probably a good idea to chat about this approach at core sync as well 😃


crates/parcel_core/src/plugin/reporter_plugin/composite_reporter_plugin.rs
```diff
@@ -0,0 +1,47 @@
+use std::fmt::{Debug, Display, Formatter};
```
MonicaOlejniczak: Can you rebase so these changes aren't in your commits?

crates/parcel/src/request_tracker/request.rs
```diff
@@ -1,67 +1,126 @@
...
 }
 
-dyn_hash::hash_trait_object!(<T: Clone> Request<T>);
+dyn_hash::hash_trait_object!(Request);
 
-#[derive(Debug, PartialEq)]
-pub struct RequestResult<Req> {
-  pub result: Req,
+#[derive(Debug, Clone, PartialEq)]
+pub struct ResultAndInvalidations {
```
MonicaOlejniczak, line 124: I detest the naming. Maybe we should just get rid of this struct if we plan to automatically track invalidations through `fs`

crates/parcel/src/request_tracker/mod.rs
```diff
@@ -1,6 +1,6 @@
 mod request;
 mod request_graph;
-mod request_tracker;
+pub mod request_tracker;
```
MonicaOlejniczak: How come it's public? I thought everything should go through the `parcel` function or whatever that will be

crates/parcel/src/lib.rs
```diff
@@ -2,7 +2,15 @@ pub use parcel::*;
 pub use parcel_filesystem as file_system;
 pub use parcel_plugin_rpc as rpc;
 
+#[allow(dead_code)]
```
yamadapc: ideally we don't allow dead code / unused variables

crates/parcel/src/test_utils.rs
```diff
@@ -0,0 +1,69 @@
...
+impl Default for RequestTrackerTestOptions {
+  fn default() -> Self {
+    Self {
+      fs: Arc::new(InMemoryFileSystem::default()),
+      search_path: PathBuf::default(),
+      project_root: PathBuf::default(),
+    }
+  }
+}
+pub(crate) fn request_tracker(options: RequestTrackerTestOptions) -> RequestTracker {
```
MonicaOlejniczak, line 47: /nit While test utils are fine in principle, I'm not fond of having all of them in a single file and prefer they live closer to the definition (e.g. request tracker utils in the request tracker module)

crates/parcel/src/request_tracker/request.rs
yamadapc: What would be a better name? We changed it when pairing as since this already wraps the `RequestResult` (previously `Req` here was actually a generic for `Req...uestResult`).

Furthermore, the actual types are `Result<ResultAndInvalidations>`, which is already confusing. Previously it was `Result<RequestResult<RequestResult>>`.

crates/parcel/src/request_tracker/request_tracker.rs
```diff
@@ -46,27 +75,92 @@ impl<T: Clone> RequestTracker<T> {
     }
   }
 
-  /// Run a request that has no parent. Return the result.
+  // Run a request that has no parent. Return the result.
```
yamadapc: `///` is a doc comment ; `//` is not

For all comments over functions we should have `///` - https://doc.rust-lang.org/rustdoc/what-is-rustdoc.html

crates/parcel/src/requests.rs
```diff
@@ -1,3 +1,19 @@
...
 mod target_request;
+
+#[derive(Clone, Debug, PartialEq)]
+pub enum RequestResult {
+  Path(PathResolution),
+  Asset(AssetResult),
+  Target(Targets),
+  // The following are test request types only used in the test build
+  #[cfg(test)]
+  Sub(String),
```
yamadapc: nit: Prefix this with `Test` ; also don't abbreviate.

You can also use a single case rather than multiple

yamadapc (approved):

crates/parcel/src/request_tracker/request.rs
mattcompiles: Totally and this was really just porting what we already have. I agree we should rethink some of these types.

crates/parcel/src/request_tracker/mod.rs
mattcompiles: I'm not 100% actually, I inherited this from the changes Pedro made. I'll see if I can make it private again.

crates/parcel/src/lib.rs
mattcompiles: Agreed, I didn't mean to add these TBH. I thought they came from your PR which I merged. Maybe I borked something when merging.

MonicaOlejniczak (approved):

crates/parcel/src/request_tracker/request.rs
MonicaOlejniczak: I don't have a good answer, most of the options I can think of aren't intuitive either. That's why I'm advocating for the autotracking. It's fine to leave as is.

crates/parcel/src/lib.rs
yamadapc: I think no ; I had removed these pragmas on [another PR](https://github.com/parcel-bundler/parcel/pull/9817/files#diff-9397decafc95795cd6fc3281edcca679248cf887cadf0643b415a840cce6b957L8)

crates/parcel/src/request_tracker/request_tracker.rs
```diff
@@ -1,75 +1,185 @@
...
+  /// than 1. For this reason the minimum number of threads our thread-pool uses is 4.
+  ///
+  /// There are multiple ways we can fix this in our implementation:
+  /// * Use async, so we get cooperative multi-threading and don't need to worry about this
+  /// * Whenever we block a thread, block using recv_timeout and then use [`rayon::yield_now`] so
+  ///   other tasks get a chance to tick on our thread-pool. This is a very poor implementation of
+  ///   the cooperative threading behaviours async will grant us.
+  /// * Don't use rayon for multi-threading here and use a custom thread-pool implementation which
+  ///   ensures we always have more threads than concurrently running requests
   #[allow(unused)]
```
devongovett, line 95: Could be simpler to only queue requests from the main thread (outside the thread pool). That way you never have this problem, and it's also faster to check the request graph there directly (no need for any locks or channels).

We only have two types of requests that need to queue other requests: AssetGraphRequest and WriteBundlesRequest (eventually). In fact these might not even need to be requests at all, they could just be functions that run as part of the main build. Requests are best for things that we'd want to cache the result of, and I'm not sure we need to store the entire AssetGraph at all. That is essentially derived from the results of other requests that are stored.

yamadapc: I've added consideration for this option, which is fair

PR #9822, opened 2024-06-26
V3: Using JsCallable for worker callbacks

alshdavid:
- Added `parcel_napi_helpers` crate to hold common napi utils (needed for RPC)
  - Moved utils there
- Renamed `RpcConnection` to `RpcWorker` and renamed implementations 
- Updated `NodejsWorker` to use `JsCallable`
- Updated JavaScript worker to pass a callable object
- Removed RPC from Javascript main


crates/parcel_napi_helpers/src/get_function.rs
```diff
@@ -0,0 +1,31 @@
...
+  let function_class: JsUnknown = env.get_global()?.get_named_property("Function")?;
+  let is_function = method.instanceof(function_class)?;
+  if !is_function {
+    return Err(napi::Error::from_reason(format!(
+      "[napi] Method is not a function: {}",
+      field_name
+    )));
+  }
+
+  let method_fn = unsafe { JsFunction::from_napi_value(env.raw(), method.raw()) }?;
```
yamadapc, line 29: I think when adding this I was unaware of `JsFunction::from_unknown` which skips the need for unsafe

packages/core/core/src/parcel-v3/plugins/Resolver.js
```diff
@@ -0,0 +1,9 @@
+// @flow
+
+export class ResolverNapi {
```
yamadapc, line 3: don't we have an existing resolver interface?

yamadapc (approved):

packages/core/core/src/parcel-v3/plugins/Resolver.js
alshdavid: I'll probably rename this, it's more like a `ResolverController` that will be used to orchestrate calling resolvers from the Rust core

PR #9827, opened 2024-07-01
Rust - LMDB Cache back-end

yamadapc:
* Adds a Rust LMDB wrapper and benchmark
* Set-up `bincode` for our types
* RKYV had been added but was removed on `71e7fc269881a44bab0d44b054e53ce8beabc477`
* Adds `docs/rfcs/parcel-v3/0001_cache_implementation.md` which discusses some cache related ideas and measurements taken from this branch


crates/parcel_core/src/types/asset.rs
```diff
@@ -47,11 +48,20 @@ impl From<String> for Code {
   }
 }
 
+#[derive(
+  Default, PartialEq, Clone, Debug, Deserialize, Serialize, bincode::Encode, bincode::Decode,
```
devongovett, line 52: Doesn't bincode already support serde? Not sure why there's a separate Encode/Decode derive

docs/rfcs/parcel-v3/0001_cache_implementation.md
```diff
@@ -0,0 +1,283 @@
...
+The issue is discussed on https://github.com/AltSysrq/lmdb-zero/issues/8 and https://github.com/meilisearch/heed/issues/198.
+
+### Orders of magnitude for LMDB write performance from rust
+
+This has been tested on an EC2 `c7i.8xlarge` instance using criterion.
+
+- Serializing a single request - **165ns** with `bincode` and 288ns with `rkyv`
+- Read a single request and deserialize - **821ns**
+- Write 1 request with async writes - **15204ns** / **15us**
+- Write 1 request with sync writes - **3000000ns** / **3000us** / **3ms**
```
devongovett, line 239: That's expensive! That means if you have 100,000 nodes and 500,000 edges in the RequestGraph (not unreasonable, my current benchmark has 87k nodes and 400k edges, and I'm sure Jira has 10x more than that), just writing them alone will take 30 minutes.

Even with async writes, that would be 9 seconds added to the build time just to write request nodes to the cache. That's 11x slower than performing a complete build from scratch with no cache.

devongovett:
I don't think we need a database for this. We don't need any of the features that a database offers, e.g. durability, fault tolerance, transactions, indexes, etc. We don't care about data loss - a cache can simply be rebuilt from scratch if something goes wrong. All of those features have significant overhead and added complexity.

The history here is that we started with a file system based cache in Parcel. We used this cache not only as an actual cache, but as temporary storage within a build to pass data from one phase to another (e.g. transformation to packaging), which may be on different threads. In the JS implementation of Parcel, this was necessary due to the lack of shared memory between threads, as well as v8 heap size limits that we ran into while building large apps. We eventually moved to LMDB because it was faster than the FS cache due to batching, but avoiding disk IO during builds entirely would be much faster still.

In Rust, neither of these are problems anymore. We can keep as much as we want in memory (even more than the total system physical memory if necessary - virtual memory is already paged by the OS), and we don't need to write to disk in order to share memory between threads. Avoiding writing to disk until Parcel shuts down (or maybe during idle time) will be much faster - I think your benchmarks show the cost well. This is worth doing because the common case is a dev server that does many builds over its lifetime, none of which need to be written to disk, so we should optimize that dev loop. Not only is this faster, it's also simpler to implement, and avoids problems of inconsistent state that could occur when data structures are sharded.

crates/parcel_core/src/types/asset.rs
yamadapc: I think I need to enable the serde feature to use the serde traits.

docs/rfcs/parcel-v3/0001_cache_implementation.md
yamadapc: Note we'd write only request nodes so those estimates are 6x higher than reality.

It's quite expensive though, I think there is likely still some tuning that could speed it up a lot ; or something my benchmark does incorrectly.

yamadapc:
We might care about transactions / consistency on some types of entries, but generally I agree that most of the time we just want to write the transformation outputs (or other intermediary assets) into files. That's anyhow the way to support remote caching.

To do that, we'd need individual cache entries for each of these intermediary assets.

crates/parcel_core/src/types/asset.rs
MonicaOlejniczak: Can you update the PR to enable the feature?

yamadapc: Hey not everything is supported in the serde layer, which is only a compatibility mode.

It also performs marginally worse. Because of this we could keep the derive as it's likely a better way of interfacing with bincode, even when the serde feature-flag is enabled (which still requires `Compat` wrappers around structs - https://github.com/bincode-org/bincode/tree/trunk/src/features/serde)

yamadapc: Refer to - https://github.com/parcel-bundler/parcel/pull/9827/commits/c1719ad0aa14ea2cee11bdbb03667f9d468b52bd

However I've reverted it since it was crashing with https://github.com/bincode-org/bincode/blob/trunk/src/features/serde/mod.rs#L70-L73

MonicaOlejniczak: Okay, thanks for looking into it

crates/parcel_filesystem/src/lib.rs
```diff
@@ -53,3 +57,63 @@ pub trait FileSystem {
   fn is_file(&self, path: &Path) -> bool;
   fn is_dir(&self, path: &Path) -> bool;
 }
+
+impl<'a, Fs: FileSystem> FileSystem for &'a Fs {
```
MonicaOlejniczak, line 61: Why is this needed? I don't really understand it

crates/parcel/benches/parcel_benchmarks.rs
```diff
@@ -0,0 +1,162 @@
...
+  let cache = LMDBCache::new(LMDBCacheOptions {
+    async_writes: false,
+    ..Default::default()
+  })
+  .unwrap();
+  benchmark_suite(c, "sync writes", cache);
+}
+
+fn benchmark_suite(c: &mut Criterion, name: &str, cache: LMDBCache) {
+  let _ = std::fs::remove_dir_all(".parcel-cache/rust-cache").unwrap();
```
MonicaOlejniczak, line 62: This should probably use `.join` along with the other usages, otherwise it won't be cross-platform

crates/parcel/benches/parcel_benchmarks.rs
```diff
@@ -0,0 +1,162 @@
...
+
+  c.bench_function(
+    &format!(
+      "{} - write request to cache one at a time using bincode to serialize",
+      name
+    ),
+    |b| {
+      let mut scratch = vec![];
+      scratch.resize(10000, 0); // pre-alloc
```
MonicaOlejniczak, line 71: Isn't this equivalent to `vec![0; 10000]` on a single line?

crates/parcel/src/cache/mod.rs
```diff
@@ -0,0 +1,119 @@
...
+/// * We don't need to allocate when returning the entries
+pub struct LMDBCache {
+  environment: heed::Env,
+  database: Database,
+}
+
+impl LMDBCache {
+  #[allow(unused)]
+  pub fn new(options: LMDBCacheOptions) -> anyhow::Result<Self> {
+    let rust_cache_path = Path::new(".parcel-cache/rust-cache");
```
MonicaOlejniczak, line 29: Should use `join`

docs/rfcs/parcel-v3/0001_cache_implementation.md
MonicaOlejniczak: Considering the cache is still unused, let's iterate on the implementation and keep the discussion open over the next few weeks as we get closer to having an E2E

crates/parcel/src/cache/mod.rs
yamadapc: I think this works on windows - https://doc.rust-lang.org/std/path/struct.Path.html

crates/parcel_filesystem/src/lib.rs
yamadapc: This is here to support the following cases - https://github.com/parcel-bundler/parcel/pull/9827/files/58b162d26553a1f9f39e6434ea96dce4fe351135#diff-4ccc502f9a5e2b717a497747f4dd20b4a48b43e51f6f5476a9db347f9756bbbcR177

```
    let child = MockFileSystem::new();
    let _tracking = TrackingFileSystem::new(&child);
    let child: FileSystemRef = Arc::new(MockFileSystem::new());
    let _tracking = TrackingFileSystem::new(&child);
    let _tracking = TrackingFileSystem::new(child);
```

We'd want to support `TrackingFileSystem::new(&other_fs)` because the way things will look like could be by borrowing fs.

This allows `Arc<dyn FileSystem>` to be `Arc<MockFileSystem>` but also `Arc<&MockFileSystem>` or for `TrackingFileSystem<Fs: FileSystem>` to be `TrackingFileSystem<Arc<dyn FileSystem>>` but also `TrackingFileSystem<MockFileSystem>` or `TrackingFileSystem<&MockFileSystem>`.

I can remove these blanket implementations from this diff

devongovett:
I'm not sure remote caching will be beneficial anymore tbh. If done per file, the network latency alone will be an order of magnitude slower than the CPU time needed to simply recompile that file from scratch (a couple milliseconds). Even if you downloaded the entire cache at once up front it's questionable whether recompiling without a cache would be faster depending on your internet speed.

yamadapc:
That's not how it'll work, I don't think. But that is a good point.

I reckon remote caching would work through downloading large incremental intermediary artifacts that are combined into the output files. That would surely be faster than rebuilding locally.

Currently there's a large amount of complexity related to cache writes, which is working around the fact there's a single cache entry. We'd like to avoid this complexity while also addressing a number of pitfalls it causes.

---

Regarding what you wrote regarding cache write times, I think we'll be able to address this better once we have an E2E build. My estimate was quite a bit lower, but I also thought since we don't need to wait we can offload any writes we do want to make.

I do agree regarding storing transformation outputs as files.

crates/parcel_filesystem/src/lib.rs
MonicaOlejniczak: Happy to keep it, makes sense now

crates/parcel/src/cache/mod.rs
MonicaOlejniczak: Interesting, I've had a lot of issues with it before but maybe it's from mixing the two

devongovett:
> That would surely be faster than rebuilding locally.

I'm not so sure. You might have to download a lot of data. Say you start with no cache. Just downloading a large cache folder over a typical internet connection might be slower than building from scratch.

> once we have an E2E build

Just pointing out that we have had this for a while on the core-rs2 branch. You're welcome to test stuff there. Not sure why we need to rewrite everything again... 😕

yamadapc:
We've decided to split the code from the markdown RFC on this branch.

Therefore I'll remove all code from this PR, and we can keep it open as a draft while we gather information and discuss how we'd like to implement persistence of "requests" or transformation outputs.

Pending information would be:

* Performance of builds past rewrites
* Combined size of all serialized requests for large builds
* Combined runtime impact of reading and writing requests individually as opposed to all in one go

It'll be possible to gather this information once E2E builds are working on `v2` with the `v3` feature-flag turned-on.

yamadapc:
It will also be useful to clarify the meaning and process related to RFC documents added into this project. Which is something we can do soon.

PR #9829, opened 2024-07-01
Fix building parcel with a debug build of itself

yamadapc:
This commit fixes a segmentation fault in parcel when building itself using a debug build of its native extensions.

This would previously crash due to usage of `Fold` and its recursive nature.

All expressions will recurse in the typeof replacer code, so SWC will crash if a module has large enough expressions, which is a case that does happen within parcel own dependencies.

This should improve performance and is the recommended usage by SWC
* https://github.com/parcel-bundler/parcel/discussions/9828 We should replace all SWC transformations from `Fold` to `VisitMut`.

parcel-bundler/parcel/discussions/9828

### DIFF
--- a/packages/transformers/js/core/src/lib.rs
+++ b/packages/transformers/js/core/src/lib.rs
@@ -77,8 +77,8 @@ use swc_core::ecma::transforms::optimization::simplify::expr_simplifier;
 use swc_core::ecma::transforms::proposal::decorators;
 use swc_core::ecma::transforms::react;
 use swc_core::ecma::transforms::typescript;
-use swc_core::ecma::visit::FoldWith;
 use swc_core::ecma::visit::VisitWith;
+use swc_core::ecma::visit::{FoldWith, VisitMutWith};
 use typeof_replacer::*;
 use utils::error_buffer_to_diagnostics;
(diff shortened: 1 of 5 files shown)


packages/transformers/js/core/src/lib.rs
```diff
@@ -388,7 +384,12 @@ pub fn transform(
                   ),
                 );
 
-                module.fold_with(&mut passes)
+                let mut module = module.fold_with(&mut passes);
+                if config.source_type != SourceType::Script {
```
MonicaOlejniczak: Should this be done before `passes` so that the other optimisations (e.g. dead branch, paren remover, etc) are guaranteed to execute?

MonicaOlejniczak (approved):
👏 for figuring this out! It looks like this broke a couple of tests legitimately, it's probably something subtle that needs to be fixed prior to merging.

packages/transformers/js/core/src/lib.rs
yamadapc: Yes I think you're right; I've put it back in the same place surrounded with an `as_folder` which will wrap the visit-mut with a top-level Fold instance that just implements fold_module by calling into the visitor

PR #9830, opened 2024-07-02
Remove SWC fold from globals replacer

yamadapc:
Continuing with https://github.com/parcel-bundler/parcel/discussions/9828 and https://github.com/parcel-bundler/parcel/pull/9829

### DIFF
--- a/packages/transformers/js/core/src/global_replacer.rs
+++ b/packages/transformers/js/core/src/global_replacer.rs
@@ -1,18 +1,18 @@
 use std::path::Path;
-use swc_core::ecma::utils::stack_size::maybe_grow_default;
 
 use indexmap::IndexMap;
 use path_slash::PathBufExt;
+use swc_core::common::sync::Lrc;
 use swc_core::common::Mark;
 use swc_core::common::SourceMap;
 use swc_core::common::SyntaxContext;
 use swc_core::common::DUMMY_SP;
-use swc_core::ecma::ast::ComputedPropName;
-use swc_core::ecma::ast::{self};
+use swc_core::ecma::ast::{self, Expr};
+use swc_core::ecma::ast::{ComputedPropName, Module};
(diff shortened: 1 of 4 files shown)


packages/transformers/js/core/src/global_replacer.rs
```diff
@@ -22,171 +22,225 @@ use crate::utils::is_unresolved;
 use crate::utils::SourceLocation;
 use crate::utils::SourceType;
 
+/// Replaces a few node.js constants with literals or require statements.
+/// This duplicates some logic in [`NodeReplacer`]
```
MonicaOlejniczak, line 26: Honestly, all of these "globals" are node specific. Could just move it all there, if you're keen for a refactor.

MonicaOlejniczak (approved):

packages/transformers/js/core/src/global_replacer.rs
yamadapc: It seems there's overlap though; e.g.: there're two implementations of the same thing on these two files.

packages/transformers/js/core/src/global_replacer.rs
```diff
@@ -22,171 +22,225 @@ use crate::utils::is_unresolved;
...
-impl<'a> Fold for GlobalReplacer<'a> {
-  fn fold_expr(&mut self, node: ast::Expr) -> ast::Expr {
+impl VisitMut for GlobalReplacer<'_> {
+  fn visit_mut_expr(&mut self, node: &mut Expr) {
     use ast::Expr::*;
-    use ast::Ident;
     use ast::MemberExpr;
     use ast::MemberProp;
 
-    // Do not traverse into the `prop` side of member expressions unless computed.
```
yamadapc, line 43: * [x] check that `something.__dirname` won't get incorrectly transformed

PR #9832, opened 2024-07-08
Fix bugs in cache writes

yamadapc:
* Fix bug where cache info entries would be written even if write was aborted
* Clean-up entries partially written if we're aborting
* Prevent concurrent cache writes
* Degrade gracefully on missing entries (which is sort of what previously happened)

### DIFF
--- a/packages/core/core/src/RequestTracker.js
+++ b/packages/core/core/src/RequestTracker.js
@@ -1516,18 +1516,31 @@ export default class RequestTracker {
         snapshotPath,
         opts,
       );
+
+      await storeRequestTrackerCacheInfo(this.options.cache, {
+        allLargeBlobKeys: Array.from(allLargeBlobKeys),
+        requestGraphKey,
+        snapshotKey,
+        timestamp: Date.now(),
+      });
+      report({
+        type: 'cache',
+        phase: 'end',
+        total,
+        size: this.graph.nodes.length,
+      });
     } catch (err) {
       // If we have aborted, ignore the error and continue
       if (!signal?.aborted) throw err;
+      logger.warn({
+        origin: '@parcel/core',
+        message: 'Aborted writing to cache, ignoring error',
+        meta: {
+          errorMessage: err.message,
+          errorStack: err.stack,
+        },
+      });
     }
-
-    await storeRequestTrackerCacheInfo(this.options.cache, {
-      allLargeBlobKeys: Array.from(allLargeBlobKeys),
-      requestGraphKey,
-      snapshotKey,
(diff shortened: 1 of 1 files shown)


packages/core/core/src/RequestTracker.js
```diff
@@ -1157,12 +1161,14 @@ export default class RequestTracker {
       return result;
     } else if (node.resultCacheKey != null && ifMatch == null) {
       let key = node.resultCacheKey;
-      invariant(this.options.cache.hasLargeBlob(key));
-      let cachedResult: T = deserialize(
-        await this.options.cache.getLargeBlob(key),
-      );
-      node.result = cachedResult;
-      return cachedResult;
+      try {
```
yamadapc, line 1164: because we're clearing entries on write, reading an asset graph while a new version for the same key is getting written will trigger an error here

packages/core/core/src/RequestTracker.js
```diff
@@ -1516,18 +1542,36 @@ export default class RequestTracker {
...
+        origin: '@parcel/core',
+        message: 'Aborted writing to cache, ignoring error',
+        meta: {
+          errorMessage: err.message,
+          errorStack: err.stack,
+        },
+      });
+
+      for (let key of allLargeBlobKeys) {
+        await this.options.cache.deleteLargeBlob(key);
```
mattcompiles, line 1572: Why do we need to delete all the large blobs on every bailout? The whole point of the bailout is to make responsive to user changes, this is potentially a lot of FS work we have to wait for?

packages/core/core/src/RequestTracker.js
yamadapc: If we don't clean-up then we don't need to handle this, but the state might still be no good as you read a potentially stale file out of disk.

packages/core/core/src/RequestTracker.js
yamadapc: We don't need to but I'm pretty sure this can complete very fast (e.g.: thousands of ops a second & we only need to delete tens to one hundred file)

JakeLane: also should we add this lint rule? https://eslint.org/docs/latest/rules/no-await-in-loop

yamadapc: I'm not a fan of this rule because a lot of the time you're not looking for concurrency & you never want uncontrolled concurrency

yamadapc:
We'll revert transactional large blobs then aim to remove large blobs or large blob sharding.

PR #9835, opened 2024-07-09
Add changes required for AssetGraphRequest

MonicaOlejniczak:
# ↪️ Pull Request

These minor changes better enable the asset graph request:
* Add `Entry` type to `EntryRequest`
* Add `named_pipelines` method to `Plugins` and update the internal `named_pipelines` to return a `Vec<String>` to avoid specifying lifetimes
* Return `entry` from `TargetRequest`, as this is the easiest way to create the connection in the AssetGraph, without async / await or storing requests in a complicated way
* Update `TargetRequest` to take in an `Entry` type
* Use dedicated `SourceField` type for `package.json` definition so there are less conflicts with Entry types

## 🚨 Test instructions

`cargo test`

### DIFF
--- a/crates/parcel/src/requests/target_request.rs
+++ b/crates/parcel/src/requests/target_request.rs
@@ -8,6 +8,7 @@ use package_json::BrowsersList;
 use package_json::BuiltInTargetDescriptor;
 use package_json::ModuleFormat;
 use package_json::PackageJson;
+use package_json::SourceField;
 use package_json::SourceMapField;
 use package_json::TargetDescriptor;
 use parcel_core::config_loader::ConfigFile;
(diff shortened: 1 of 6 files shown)


crates/parcel/src/requests/target_request.rs
```diff
@@ -42,22 +43,23 @@ mod package_json;
...
 pub struct TargetRequest {
   pub default_target_options: DefaultTargetOptions,
+  pub entry: Entry,
   pub env: Option<HashMap<String, String>>,
-  pub exclusive_target: Option<String>,
   pub mode: BuildMode,
 }
 
 impl Hash for TargetRequest {
   fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
```
yamadapc, line 52: Why don't we derive?

yamadapc (approved):

crates/parcel/src/requests/target_request.rs
MonicaOlejniczak: The `env` uses `HashMap` does not derive `Hash`

yamadapc: We should wrap it with a type we implement hash in then but still derive here

PR #9836, opened 2024-07-09
Revert transactional large blob entries

yamadapc:
This reverts a set of changes that added transaction semantics to large blob entries.

This means files that are spread amongst multiple files may be corrupted. We're revert to avoid other bugs related to this diff, possibly related to concurrency races and so on that will remain on the codebase.

Namely, we had issues where "large blobs" were missing when code expects it to be present.

Most of the time when reading "AssetGraph" large blob entry and sometimes at runtime possibly related to that entry being cleaned-up.


packages/core/cache/test/LMDBCache.test.js
```diff
@@ -1,94 +0,0 @@
-// @flow
-
-import {DEFAULT_FEATURE_FLAGS, setFeatureFlags} from '@parcel/feature-flags';
```
mattcompiles, line 3: Are all these tests broken now? Seems like most of them will still pass?

packages/core/core/test/test-utils.js
```diff
@@ -54,7 +53,9 @@ export const DEFAULT_OPTIONS: ParcelOptions = {
     sourceMaps: false,
   },
   featureFlags: {
-    ...DEFAULT_FEATURE_FLAGS,
+    exampleFeature: false,
```
mattcompiles, line 56: Should this be `DEFAULT_FEATURE_FLAGS`

packages/core/cache/src/FSCache.js
```diff
@@ -141,12 +132,12 @@ export class FSCache implements Cache {
...
-                  i * this.#writeLimitChunk,
-                  (i + 1) * this.#writeLimitChunk,
+                  i * WRITE_LIMIT_CHUNK,
+                  (i + 1) * WRITE_LIMIT_CHUNK,
                 )
               : contents.subarray(
-                  i * this.#writeLimitChunk,
-                  (i + 1) * this.#writeLimitChunk,
+                  i * WRITE_LIMIT_CHUNK,
+                  (i + 1) * WRITE_LIMIT_CHUNK,
```
mattcompiles, line 140: Did we want to revert the chunking here too seeing as that was the potential issue you were concerned about?

flow-libs/lmdb.js.flow
```diff
@@ -1,669 +0,0 @@
-// @flow
```
mattcompiles, line 1: Feels like a shame to revert this?

mattcompiles (approved):
I'll approve as I know this is a revert, but it feels like we're reverting code we'd like to keep too.

mattcompiles:
Didn't realise auto-merge was on 😬

yamadapc:
It's simpler to just revert the commits.

Issue #9651, opened 2024-04-16
Use v4 of actions/upload-artifact or actions/download-artifact


mischnic:
Our CentOS image used to build against a old glibc version doesn't support v4 of actions/upload-artifact or actions/download-artifact:
- https://github.com/parcel-bundler/parcel/pull/9471
- https://github.com/parcel-bundler/parcel/pull/9472

The v3 of these actions will stop working in November 2024, so we'll have to fix it until then: https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/

mischnic:
This could be used instead of the ancient Centos image: https://github.com/corsix/polyfill-glibc

Issue #5599, opened 2021-01-06
Error: /lib64/libc.so.6: version `GLIBC_2.18' not found in Red Hat Enterprise Linux 7.9


guyk-yahoo:
# 🐛 bug report
With Parcel1 everything works.
After migrating to Parcel2 -- I'm getting the following error - running on Red Hat Enterprise Linux 7.9

From the looks of it - (running `/lib64/libc.so.6 --version`) this distro contains `GNU C Library (GNU libc) stable release version 2.17, by Roland McGrath et al.`
```
parcel build src/public/js/main.ts --target browser --no-scope-hoist
14:56:16 
14:56:16 Error: /lib64/libc.so.6: version `GLIBC_2.18' not found (required by /......./node_modules/@parcel/fs-search/fs-search.linux-x64-gnu.node)
14:56:16     at Object.Module._extensions..node (internal/modules/cjs/loader.js:1122:18)
14:56:16     at Module.load (internal/modules/cjs/loader.js:928:32)
14:56:16     at Function.Module._load (internal/modules/cjs/loader.js:769:14)
14:56:16     at Module.require (internal/modules/cjs/loader.js:952:19)
14:56:16     at require (/......./node_modules/v8-compile-cache/v8-compile-cache.js:159:20)
14:56:16     at Object.<anonymous> (/......./node_modules/@parcel/fs-search/index.js:16:18)
14:56:16     at Module._compile (/......./node_modules/v8-compile-cache/v8-compile-cache.js:192:30)
14:56:16     at Object.Module._extensions..js (internal/modules/cjs/loader.js:1092:10)
14:56:16     at Module.load (internal/modules/cjs/loader.js:928:32)
14:56:16     at Function.Module._load (internal/modules/cjs/loader.js:769:14)
14:56:16 npm ERR! code ELIFECYCLE
14:56:16 npm ERR! errno 1
14:56:16 npm ERR! my_project@0.0.43-pr43.b22265458 parcel: `parcel build src/public/js/main.ts --target browser --no-scope-hoist`
14:56:16 npm ERR! Exit status 1
14:56:16 npm ERR! 
14:56:16 npm ERR! Failed at the my_project@0.0.43-pr43.b22265458 parcel script.
14:56:16 npm ERR! This is probably not a problem with npm. There is likely additional logging output above.
```

## 🎛 Configuration (.babelrc, package.json, cli command)

```js
{
[...]

PR #9841, opened 2024-07-11
Fix canary releases

yamadapc:
This PR switches just one runner image to use `node:...debian` docker image to test it out.

Long-term we'd want all linux builds to run in debian, but this PR is just switching the x86 build to try to fix the canary release - https://github.com/parcel-bundler/parcel/actions/runs/9864446097/job/27296209033

- - -

## Context

Currently the GitHub actions configuration is complex due to supporting building the native add-ons for many target OSs and architectures.

To support this, there are a number of different runner jobs running on different images/os:

* macos-latest is used for arm / x86 builds of the macOS targets
* windows-latest is used for windows builds
* linux builds
  * CentOS 7 is used for a x64 build
  * Ubuntu 20.04 is used for armhf 32-bit, arm64
  * What looks like Alpine linux in napi-rs/napi-rs/nodejs-rust is used for MUSL x86 64-bit and MUSL arm64

While this is currently set-up and working, maintaining this set-up becomes hard. When adding new native libraries for example that need to be linked dynamically against the main binary, this becomes an issue.

## Trying debian

Rather than using multiple OSes we could strive for having a single Linux distribution cross compile for all targets. For example, debian has good support for cross-compilation with popular libraries being distributed as binary for many architectures:

* https://wiki.debian.org/Multiarch/HOWTO

## OpenSSL

* To avoid having to cross-compile OpenSSL at the moment we're using rustls - SSL is needed for the sentry client used only on canary releases and was added on https://github.com/parcel-bundler/parcel/pull/9667

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -90,7 +90,7 @@ jobs:
     name: linux-gnu-x64
     runs-on: ubuntu-20.04
     container:
-      image: docker.io/mischnic/centos7-node16
+      image: node:22.4-bookworm
     steps:
       - uses: actions/checkout@v3
       - name: Install yarn
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -93,8 +93,6 @@ jobs:
       image: node:22.4-bookworm
     steps:
       - uses: actions/checkout@v3
-      - name: Install yarn
-        run: npm install --global yarn@1
       - name: Install Rust
         uses: actions-rs/toolchain@v1
         with:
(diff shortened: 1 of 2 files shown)


mattcompiles (approved):

mischnic:
For those who don't remember/know why it was there: it was using CentOS so that the dynamically linked minimum glibc version is low enough for RHEL/CentOS: https://github.com/parcel-bundler/parcel/issues/5599

There's another possible solution if someone still needs this older glibc here: https://github.com/parcel-bundler/parcel/issues/9651

But if all of these old RHEL machines are decommissioned now, then even better.

PR #9849, opened 2024-07-15
Use debian bullseye to support GLIBC 2.31

yamadapc:
Use older debian version to support slightly older GLIBC base version.

Follow-up on https://github.com/parcel-bundler/parcel/pull/9841

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -90,7 +90,7 @@ jobs:
     name: linux-gnu-x64
     runs-on: ubuntu-20.04
     container:
-      image: node:22.4-bookworm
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -290,6 +290,10 @@ jobs:
         with:
           name: debug-symbols
           path: debug-symbols/**
+      - name: Build node packages
```
MonicaOlejniczak: Either these or the scripts in `prepublishOnly` need to be removed

MonicaOlejniczak (changes requested):

MonicaOlejniczak (approved):

PR #9850, opened 2024-07-16
Add initial Rust integration

MonicaOlejniczak:
# ↪️ Pull Request

These changes update the Parcel V3 entrypoint to use the various pieces that have been implemented in the last several weeks.
* Add logic to infer the project root, that is roughly equivalent to the JavaScript implementation
* Add `find_ancestor_directory` and `find_ancestor` to support the inferred project root
* Add `PackageManagerNapi` binding
* Convert all options passed in from napi to native `ParcelOptions` core type and fix incorrect serde annotations
* Load config, plugins, and initialise the request tracker in the parcel entrypoint
* Update flow types and replace previously unused `FileSystem` type (it is essentially a duplicate of `ResolverOptions`)
* Update `FileSystemNapi` to handle `canonicalize` and `cwd`
* Update v3 integration test with new inputs, use existing package manager and fs fixtures
* Update napi error output so more details are reported
* Update tracing setup to report original error alongside the user-facing message

## 🚨 Test instructions

`cargo test`


crates/parcel/src/project_root.rs
```diff
@@ -0,0 +1,254 @@
...
+    return None;
+  }
+}
+
+/// Finds the common path prefix shared between all input paths
+fn common_path(paths: &[PathBuf]) -> Option<PathBuf> {
+  let components = paths.iter().map(|path| path.components()).collect();
+  let mut common_path: Option<PathBuf> = None;
+
+  for component in CommonComponents(components) {
```
yamadapc: This would be quite a bit clearer written as normal iterator lambdas or just a loop rather than a custom iterator

I'm not entirely sure if I follow the logic in `CommonComponents` & think this is worth writing a unit-test to show it works

crates/parcel/src/project_root.rs
```diff
@@ -0,0 +1,254 @@
...
+    common_path = common_path
+      .map(|p| p.join(component))
+      .or(Some(PathBuf::from(&component)));
+  }
+
+  common_path
+}
+
+pub fn infer_project_root(fs: FileSystemRef, entries: Option<Entry>) -> PathBuf {
+  let cwd = fs.cwd().expect("Expected fs.cwd() to exist");
```
yamadapc: can't we fail instead of panicking?

crates/parcel/src/request_tracker/request_tracker.rs
```diff
@@ -35,33 +35,32 @@ use super::{RunRequestContext, RunRequestMessage};
 ///
 /// This will be used to trigger cache invalidations.
 pub struct RequestTracker {
-  graph: RequestGraph<RequestResult>,
-  reporter: Arc<CompositeReporterPlugin>,
-  request_index: HashMap<u64, NodeIndex>,
   cache: CacheRef,
```
yamadapc, line 38: This looks fine really but alphabetical ordering of fields is not a good **rule** in native because the order here is the order in memory and the memory layout can matter for certain things (not this though)

yamadapc (changes requested):
I think it looks good to me ; the only thing that jumped out was the project_root part.

It'd be nice to have tests for `search.rs` since it appears there weren't any

crates/parcel/src/project_root.rs
MonicaOlejniczak: I had it in a loop before, but it was ick until I refined it some more. I'll put it back in a loop since it's not so bad now along with some tests.

crates/parcel/src/project_root.rs
MonicaOlejniczak: Yeah, I had different signatures before because reasons but can move this to a result now

crates/parcel/src/request_tracker/request_tracker.rs
MonicaOlejniczak: Sorry, I'm a bit obsessive over it. Isn't this only relevant to non-Rust representations though?

MonicaOlejniczak:
@yamadapc Thanks for the review! I've addressed your feedback now, and will add the search tests in a follow-up pull request

crates/parcel/src/request_tracker/request_tracker.rs
yamadapc: I wouldn't think so; though maybe without repr(C) there's no guarantee the compiler won't mess around with the layout.

Anyhow I'm just commenting, this is not a request to change or anything like that.

yamadapc (approved):

PR #9852, opened 2024-07-17
Add env replacer based on VisitMut

yamadapc:
#9828

This is aimed at improving moving SWC usage to visit for all parcel
use-cases as we are facing memory errors due to stack-overflow. Errors
include segmentation faults and bus errors.

This is part of #9828. Adds exhaustive tests for the folder.
Update transformer to use the VisitMut impl.

This replaces the Fold env replacer with a VisitMut implementation.
All tests pass on both old/new versions.

The following behaviours of the fold implementation should be revised in
the future as they might be unwanted:

* `process.browser = ...` assignment expressions are replaced with
  `process.browser = true`, which is not a safe replacement
* expression replacements `const x = ({ x } = process.env);` are
  supported but insert an extra trailing object `const x = (x = 'asdf', {});
* when `replace_env` replacement is off the transformer still replaces
  several cases that aren't the base declarations case

Overall could simplify this so that less cases are supported, and then
shift it to be a generalised inline variable replacement implementation.

### DIFF
--- a/packages/transformers/js/core/src/env_replacer.rs
+++ b/packages/transformers/js/core/src/env_replacer.rs
@@ -209,8 +209,7 @@ impl<'a> EnvReplacer<'a> {
   /// If an expression matches `process.browser = ...` then the RHS is replaced with
   /// `true` when `is_browser` is set to true.
   ///
-  /// This doesn't make sense so it's deprecated and will not run anymore on canary builds or
-  /// future versions.
+  /// This likely doesn't make sense so it should be deprecated in the future.
   fn replace_browser_assignment(&mut self, node: &Expr) -> Option<Expr> {
     let Expr::Assign(ref assign) = node else {
       return None;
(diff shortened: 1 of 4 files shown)


packages/transformers/js/core/src/env_replacer.rs
```diff
@@ -187,19 +191,48 @@ impl<'a> Fold for EnvReplacer<'a> {
...
-    }
+    };
   }
 }
 
 impl<'a> EnvReplacer<'a> {
+  /// If an expression matches `process.browser = ...` then the RHS is replaced with
+  /// `true` when `is_browser` is set to true.
+  ///
+  /// This doesn't make sense so it's deprecated and will not run anymore on canary builds or
```
MonicaOlejniczak: Where is the logic to omit this code from canary builds and future versions?

yamadapc: I haven't gated it yet

MonicaOlejniczak (approved):

PR #9853, opened 2024-07-17
Implement minimal plugin options

MonicaOlejniczak:
# ↪️ Pull Request

Adds the plugin options needed for the Rust based resolver plugin

### DIFF
--- a/crates/parcel/src/parcel.rs
+++ b/crates/parcel/src/parcel.rs
@@ -84,8 +84,11 @@ impl Parcel {
       config,
       PluginContext {
         config: Arc::clone(&config_loader),
-        // TODO options and logger
-        options: Arc::new(PluginOptions::default()),
+        options: Arc::new(PluginOptions {
+          mode: self.options.mode.clone(),
+          project_root: self.project_root.clone(),
+        }),
+        // TODO Initialise actual logger
         logger: PluginLogger::default(),
       },
     );


crates/parcel_core/src/plugin.rs
```diff
@@ -48,3 +48,12 @@ pub struct PluginOptions {
   pub mode: BuildMode,
   pub project_root: PathBuf,
 }
+
+impl From<&ParcelOptions> for PluginOptions {
+  fn from(options: &ParcelOptions) -> Self {
+    PluginOptions {
+      mode: options.mode.clone(),
```
yamadapc: Can you just inline this?

`From` is for conversion between types ;

I don't think that is really the case here - https://doc.rust-lang.org/std/convert/trait.From.html#when-to-implement-from

MonicaOlejniczak: There will be a conversion later when more of the options are ported over. The `PluginOptions` is a subset of `ParcelOptions` effectively as you can see in the JavaScript implementation [here](https://github.com/parcel-bundler/parcel/blob/v2/packages/core/core/src/public/PluginOptions.js) and the `From` will be a convenience then.

I could create a constructor or inline it instead of using `From`, which might be okay. But I do like chaining options using the destructuring syntax.

yamadapc: I'd rather use a constructor, because it seems you're constructing an invalid options state (without the root) then immediately patching it. Also it's used once.

If there's a full conversion then we can refactor to trait implementation.

We should always start by writing raw methods and functions.

From docs above

> The conversion is value-preserving: the conceptual kind and meaning of the resulting value is the same, even though the Rust type and technical representation might be different
> The conversion is obvious: it’s the only reasonable conversion between the two types. Otherwise it’s better to have it be a named method or constructor

MonicaOlejniczak: Yeah, I'm not fussed tbh and ended up inlining it

yamadapc (approved):

PR #9858, opened 2024-07-21
Move all dependencies for v3 rust prototype behind a feature flag

devongovett:
This is in preparation for the next v2 release. I noticed that the `@parcel/rust` npm package has grown 25% since v2.12.0 due to the addition of the rust core prototype. This moves all of that work behind a Cargo feature, which is only enabled in canary builds and development builds.

### DIFF
--- a/crates/node-bindings/Cargo.toml
+++ b/crates/node-bindings/Cargo.toml
@@ -8,38 +8,55 @@ edition = "2021"
 crate-type = ["cdylib"]
 
 [features]
-canary = ["sentry", "whoami", "rustls"]
-
+canary = ["sentry", "whoami", "rustls", "v3-prototype"]
+v3-prototype = [
+  "parcel",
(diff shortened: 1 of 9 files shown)


crates/node-bindings/Cargo.toml
```diff
@@ -8,50 +8,79 @@ edition = "2021"
...
+  "parcel_package_manager",
+  "parcel_plugin_transformer_js",
+  "parcel_napi_helpers",
+  "anyhow",
+  "glob",
+  "log",
+  "mockall",
+  "num_cpus",
+  "parking_lot",
+  "toml",
```
MonicaOlejniczak, line 24: Toml doesn't look used, can remove that one

MonicaOlejniczak (approved):

devongovett:
Closing in favor of #9936.

PR #9859, opened 2024-07-21
changelog for v2.13.0

devongovett:
I was unsure whether to include "Add config key invalidation" #9597 or not since it is still behind a feature flag. Are we moving that out of a flagged state or leaving it for now @mattcompiles?


mattcompiles:
Hey @devongovett. Thanks for the reminder. Just took a look at the stats for the config key invalidation feature. We're seeing an increased cache hit ratio for requests leading to roughly 8% faster builds when a cache is present. I think it's safe to roll this out for everyone. I'll make a PR to clean up the feature flag today.

mattcompiles:
PR up 
https://github.com/parcel-bundler/parcel/pull/9864

PR #9860, opened 2024-07-22
Add retriable dynamic imports behind a feature flag

alshdavid:
This PR adds the ability to retry dynamic imports. This will be re-implemented when we revisit the runtime during the native rewrite.

## 💻 Examples

```typescript
while true {
  try {
    await import('./foo.js').then(console.log)
    break
  } catch (error) {
    await new Promise(res => setTimeout(res, 1000))
  }
}
```

```
npx parcel build --feature-flag importRetry=true
```

### DIFF
--- a/packages/runtimes/js/src/JSRuntime.js
+++ b/packages/runtimes/js/src/JSRuntime.js
@@ -475,14 +475,23 @@ function getLoaderRuntime({
     loaderCode = `(${loaderCode})`;
   }
 
+  if (mainBundle.type === 'js') {
+    let parcelRequire = bundle.env.shouldScopeHoist
+      ? 'parcelRequire'
+      : 'module.bundle.root';
+    loaderCode += `.then(() => ${parcelRequire}('${bundleGraph.getAssetPublicId(
+      bundleGraph.getAssetById(bundleGroup.entryAssetId),
+    )}'))`;
+  }
+
   if (needsEsmLoadPrelude && options.featureFlags.importRetry) {
-    const assetId = bundleGraph.getAssetById(bundleGroup.entryAssetId);
     loaderCode = `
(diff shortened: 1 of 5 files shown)


packages/runtimes/js/src/JSRuntime.js
```diff
@@ -475,6 +475,24 @@ function getLoaderRuntime({
     loaderCode = `(${loaderCode})`;
   }
 
+  if (needsEsmLoadPrelude && options.featureFlags.importRetry) {
+    const assetId = bundleGraph.getAssetById(bundleGroup.entryAssetId);
+    loaderCode = `
+      Object.defineProperty(module, 'exports', { get: () => {
```
mattcompiles, line 489: Is there a reason for this module definition stuff?

packages/runtimes/js/src/JSRuntime.js
```diff
@@ -475,6 +475,24 @@ function getLoaderRuntime({
...
+        let load = require('./helpers/browser/esm-js-loader-retry');
```
MonicaOlejniczak, line 490: How come this isn't using the logic from below? Specifically the non-scope hoisting behaviour of `module.bundle.root`

packages/runtimes/js/src/JSRuntime.js
alshdavid: At the call site we call `parcelRequire` directly, expecting it to return a promise:

```javascript
const abc = parcelRequire("abc")
```

The value resolved by the `Promise` is cached so subsequent calls will not rerun the constructor. Using a getter ensures the promise is recreated - essentially hacking it to be 

```javascript
const abc = parcelRequire("abc")()
```

That said, I should probably only remove the cache entry if there is a failure in a request. Will try to do that

mattcompiles (approved):

PR #9861, opened 2024-07-22
Worker thread initialization via JsTransferable

alshdavid:
- Added `JsTransferable` type to allow sending values from the main thread to workers
  - This is similar to Napi's `External` type but can be sent to worker thread via `postMessage` 
  - This is needed to inject things into Node worker threads (like filesystem)
- Transfering a Rust channel to the Nodejs worker so it can notify the main thread that a worker has started

## 💻 Examples

```rust
// Main thread
#[napi]
fn create_foo() -> napi::Result<JsTransferable<String>> {
  let my_string = "hello world".to_string();
  Ok(JsTransferable::new(my_string))
}
```

```javascript
// Main thread
import napi from 'napi.node'

let my_string_ptr = napi.createFoo();
workerThread.postMessage(my_string_ptr);
```

```javascript
// Worker thread
import napi from 'napi.node'
import { parentPort } from 'node::worker_threads'

parentPort.addEvent('message', ({ data: my_string_ptr }) => {
  napi.useFoo(my_string_ptr)
})
```

```rust
// Worker thread
#[napi]
fn use_foo(my_string_ptr: JsTransferable<String>) -> napi::Result<()> {
  let my_string = my_string_ptr.take()?;
  println!("{my_string}");
  Ok(())
}
```


packages/core/rust/index.js.flow
```diff
@@ -50,18 +52,25 @@ export type ParcelNapiOptions = {|
   threads?: number,
 |};
 
+export type ParcelBuildOptions = {|
+  registerWorker: (channel: Transferable) => any,
```
MonicaOlejniczak: returns void?

crates/parcel_plugin_rpc/src/nodejs/rpc_host_nodejs.rs
```diff
@@ -22,7 +30,10 @@ impl RpcHost for RpcHostNodejs {
     let mut connections = vec![];
 
     for _ in 0..self.node_workers {
-      connections.push(get_worker())
+      let Ok(worker) = self.rx_worker.lock().recv() else {
```
MonicaOlejniczak: Do we need to lock on every iteration?

crates/parcel_napi_helpers/src/transferable.rs
```diff
@@ -0,0 +1,81 @@
...
+    let id = COUNTER.fetch_add(1, Ordering::Relaxed);
+
+    VALUES.lock().unwrap().insert(id.clone(), Box::new(value));
+    Self {
+      id,
+      _value: Default::default(),
+    }
+  }
+
+  /// Take value out of Transferable. It can no longer be accessed
```
MonicaOlejniczak: The first line of any doc comment should be one sentence

```suggestion
  /// Take the value out of Transferable, so it can no longer be accessed
```

crates/parcel_napi_helpers/src/transferable.rs
```diff
@@ -0,0 +1,81 @@
...
+  pub fn take(self) -> napi::Result<T> {
+    let Some(value) = VALUES.lock().unwrap().remove(&self.id) else {
+      return Err(napi::Error::from_reason("Value does not exist"));
```
MonicaOlejniczak: Are we able to make these errors more identifiable later with a human readable name? I feel the stack traces won't be useful, but could be wrong

crates/parcel_napi_helpers/src/transferable.rs
```diff
@@ -0,0 +1,81 @@
...
+  Lazy::new(|| Default::default());
+
+/// Creates an external reference to a Rust value and
+/// makes it transferable across Nodejs workers
+///
+/// This is to get around the limitations of what can be transferred
+/// between workers in Nodejs
+///
+/// https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Transferable_objects
+pub struct JsTransferable<T> {
```
MonicaOlejniczak, line 27: Can you add a unit test or demonstrate some integration with an existing structure?

alshdavid: I'm currently using`JsTransferable` to transfer a `channel::Sender` from the main thread into the nodejs worker which it uses to notify the main thread that it's ready to be used. 

Can add a test though, that will be helpful for documenting things

crates/parcel_plugin_rpc/src/nodejs/rpc_host_nodejs.rs
```diff
@@ -1,18 +1,26 @@
...
 use crate::RpcHost;
 use crate::RpcWorkerRef;
 
-use super::worker_init::get_worker;
+use super::NodejsWorker;
 use super::NodejsWorkerFarm;
 
 pub struct RpcHostNodejs {
   node_workers: usize,
+  rx_worker: Arc<Mutex<Receiver<NodejsWorker>>>,
```
yamadapc: why is there a lock around this?

crates/node-bindings/src/parcel/worker.rs
```diff
@@ -1,10 +1,20 @@
...
-pub fn register_worker(env: Env, worker: JsObject) -> napi::Result<JsUndefined> {
+pub fn register_worker(
+  env: Env,
+  channel: JsTransferable<Sender<NodejsWorker>>,
+  worker: JsObject,
+) -> napi::Result<JsUndefined> {
   let worker = NodejsWorker::new(worker)?;
-  NodejsWorker::register_worker(worker);
+  let tx_worker = channel.take()?;
+  if tx_worker.send(worker).is_err() {
```
yamadapc, line 36: ~~I don't think it's safe to send this across threads - https://github.com/parcel-bundler/parcel/blob/445b9bc0ca57922a20977ee81b50e7a07e35b6ab/crates/parcel_plugin_rpc/src/nodejs/rpc_conn_nodejs.rs#L11~~

yamadapc: This is safe due to `JsCallable` being thread-safe

packages/core/rust/index.js.flow
alshdavid: Updated

crates/parcel_napi_helpers/src/transferable.rs
alshdavid: Yeah that's a fair point. I've updated the PR to add a little more information (id and type) in the error messages

crates/parcel_plugin_rpc/src/nodejs/rpc_host_nodejs.rs
alshdavid: The `RpcHostNodejs` needs to be `Send + Sync`. I don't need the `Arc` though so I've removed that

crates/parcel_plugin_rpc/src/nodejs/rpc_host_nodejs.rs
alshdavid: Moved the lock above the loop. Was being a bit lazy there :sweat_smile:

yamadapc (approved):

PR #9864, opened 2024-07-23
Cleanup config key invalidation feature for release

mattcompiles:
# ↪️ Pull Request

This PR removes the `configKeyInvalidation` feature flag in preparation for release as it has been tested successfully. 

See the original PR for details: https://github.com/parcel-bundler/parcel/pull/9597

### DIFF
--- a/packages/core/integration-tests/test/incremental-bundling.js
+++ b/packages/core/integration-tests/test/incremental-bundling.js
@@ -5,6 +5,7 @@ import {
   inputFS,
   overlayFS,
   run,
+  fsFixture,
 } from '@parcel/test-utils';
 import assert from 'assert';
 import path from 'path';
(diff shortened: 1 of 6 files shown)


packages/core/integration-tests/test/incremental-bundling.js
```diff
@@ -712,21 +713,33 @@ console.log('index.js');`,
 
     it('changing bundler options', async () => {
       let subscription;
-      let fixture = path.join(__dirname, '/integration/incremental-bundling');
       try {
-        let b = bundler(path.join(fixture, 'index.js'), {
+        await fsFixture(overlayFS, __dirname)`
```
mattcompiles, line 717: I migrated this test as it was incorrectly sending file `create` events rather than `update` events causing the config to not invalidate. This is a bug when using overlayFS with a normal read FS but an in memory write FS.

MonicaOlejniczak (approved):
lgtm

PR #9866, opened 2024-07-23
Parcel3 - Fixed hanging process and tracer

alshdavid:
*Fixes*

Process would panic if `new ParcelV3()` was called more than one time in a single process - this throws in integration tests where a single Nodejs process creates multiple Parcel instances.
- Ignore tracer errors when Parcel is initialized multiple times
- Refactored tracer initializer
- By default, the tracer now outputs to stdout in debug mode and to a file in release

Process would hang on exit
- Added `into_unref()` to `JsCallable` to tell Nodejs not to hold the process open waiting for callbacks to be dropped
- Updated call sites of `JsCallable` to ensure nothing held the process open


crates/node-bindings/src/parcel/tracer.rs
```diff
@@ -0,0 +1,178 @@
...
+    directory: String,
+    /// A prefix for the log file names.
+    prefix: String,
+    /// The maximum number of rotated files to keep.
+    max_files: u32,
+  },
+}
+
+/// Telling Napi how to convert from a JS type to the enum
+impl FromNapiValue for TracerMode {
```
MonicaOlejniczak: Can't you just annotate the enum with napi since it's already in node bindings?

crates/node-bindings/src/parcel/tracer.rs
```diff
@@ -0,0 +1,178 @@
...
+
+/// A reference counted guard that holds the tracer
+/// open until all instances are dropped
+pub struct Tracer {
+  count: Arc<AtomicUsize>,
+  worker_guard: Arc<Option<WorkerGuard>>,
+}
+
+impl Tracer {
+  /// The Tracer can only be initialized once per process instance,
```
MonicaOlejniczak: This isn't a function doc comment, should just be `//`

crates/node-bindings/src/parcel/tracer.rs
alshdavid: Unfortunately structured enums are not supported by napi-rs and need to be manually mapped. An enum with internal values also has weird mappings.

I'm thinking of writing a macro to standardize enum serialization/deserialization so that an enum always serializes to/from `{ key: string, ...data: EnumProps }`

crates/node-bindings/src/parcel/tracer.rs
alshdavid: :+1: Updated

crates/node-bindings/src/parcel/tracer.rs
MonicaOlejniczak: I would just use serde then 🤷

crates/node-bindings/src/parcel/tracer.rs
```diff
@@ -0,0 +1,178 @@
...
+}
+
+/// Needs to exist for deserialization
+impl ToNapiValue for TracerMode {
+  unsafe fn to_napi_value(
+    _env: napi::sys::napi_env,
+    _val: Self,
+  ) -> napi::Result<napi::sys::napi_value> {
+    // we don't send this value back to JS
+    unimplemented!()
```
yamadapc: Can you remove if it's not used

crates/node-bindings/src/parcel/tracer.rs
```diff
@@ -0,0 +1,178 @@
...
+/// A reference counted guard that holds the tracer
+/// open until all instances are dropped
+pub struct Tracer {
+  count: Arc<AtomicUsize>,
+  worker_guard: Arc<Option<WorkerGuard>>,
+}
+
+impl Tracer {
+  // The Tracer can only be initialized once per process instance,
+  //subsequent invocations will reuse an existing instance
```
yamadapc: I don't understand why you need to handle this;

crates/node-bindings/src/parcel/tracer.rs
```diff
@@ -0,0 +1,178 @@
...
+    global_tracer.replace(tracer.clone());
+
+    Ok(tracer)
+  }
+}
+
+/// Increment reference count
+impl Clone for Tracer {
+  fn clone(&self) -> Self {
+    self.count.fetch_add(1, Ordering::Relaxed);
```
yamadapc: Instead of doing this you can make factory functions return `Arc` and keep a `Weak` in the global

But I'm not sure why we need to handle to build this wrapper

crates/node-bindings/src/parcel/parcel/tracing_setup.rs
```diff
@@ -1,73 +0,0 @@
...
-}
-
-/// This is a guard, when it's dropped certain resources related to tracing might flush to disk or
-/// be released. It should be kept alive for the duration of the program.
-pub struct ParcelTracingGuard {
-  guard: Option<tracing_appender::non_blocking::WorkerGuard>,
-}
-
-/// Set-up tracing based on JavaScript provided options and return a guard.
-pub fn setup_tracing(options: &Option<ParcelTracingOptions>) -> anyhow::Result<ParcelTracingGuard> {
```
yamadapc, line 47: This already will only be initialized once because it'll fail if you initialise twice, we can just ignore the failure

crates/node-bindings/src/parcel/tracer.rs
alshdavid: Do we envision needing multiple different tracers running? For example during integration tests where they may need to output to files with different prefixes or something

yamadapc: No I don't think so, but `tracing` itself won't allow more than one per process

crates/node-bindings/src/parcel/tracer.rs
alshdavid: Updated

crates/node-bindings/src/parcel/tracer.rs
alshdavid: Updated

crates/node-bindings/src/parcel/tracer.rs
alshdavid: Updated

crates/node-bindings/src/parcel/tracer.rs
alshdavid: Removed this

crates/node-bindings/src/parcel/parcel/tracing_setup.rs
alshdavid: Updated to ignore the failure

crates/parcel_napi_helpers/src/js_callable/js_callable.rs
```diff
@@ -58,6 +58,11 @@ impl JsCallable {
     Self::new(jsfn)
   }
 
+  pub fn into_unref(mut self, env: &Env) -> napi::Result<Self> {
+    self.threadsafe_function.unref(env)?;
```
yamadapc, line 62: I don't think this is right, we should do it on drop if we need to

yamadapc: https://nodejs.org/api/n-api.html#napi_unref_threadsafe_function

Nevermind, I guess this is okay

packages/core/integration-tests/test/parcel-v3.js
```diff
@@ -52,7 +52,7 @@ describe('parcel-v3', function () {
     assert.equal(output(), 3);
   });
 
-  it('should run the main-thread bootstrap function', async function () {
+  it.only('should run the main-thread bootstrap function', async function () {
```
yamadapc: You don't want this

crates/parcel_napi_helpers/src/js_callable/js_callable.rs
alshdavid: :facepalm:  Thanks

packages/core/integration-tests/test/parcel-v3.js
alshdavid: :facepalm:  Thanks

yamadapc (approved):

PR #9869, opened 2024-07-24
Fix small HMR issues affecting usage at scale

JakeLane:
# ↪️ Pull Request

A couple of quick fixes for the HMR dev loop. This is required to get a major size project at Atlassian to work somewhat correctly.

- Fix HMR not using HTTPs from dev server correctly (#6061)
- When update event is too large, send a reload event instead of crashing
- New error when react refresh runtime and transformers have mismatched react refresh dependency versions


packages/reporters/dev-server/src/HMRServer.js
```diff
@@ -272,9 +272,19 @@ export default class HMRServer {
   }
 
   broadcast(msg: HMRMessage) {
-    const json = JSON.stringify(msg);
-    for (let ws of this.wss.clients) {
-      ws.send(json);
+    if (msg.type === 'update' && msg.assets.length > 10000) {
```
marcins: Is this just an arbitrary threshold that should be well under the limit?

Also, should this be extracted to a `const` rather than a magic number as it's referenced in a few places?

marcins (approved):

packages/reporters/dev-server/src/HMRServer.js
JakeLane: yeah, it's mostly arbitrary, just the most round number, which was around 75% of the limit when I was local testing.

Happy to move this to a constant 👍

packages/runtimes/hmr/src/HMRRuntime.js
```diff
@@ -41,7 +41,9 @@ export default (new Runtime({
           port != null &&
             // Default to the HTTP port in the browser, only override
             // in watch mode or if hmr port != serve port
-            (!options.serveOptions || options.serveOptions.port !== port)
+            (!options.serveOptions ||
+              options.serveOptions.port !== port ||
+              forcePortOverride)
```
devongovett: If the dev server and HMR server are on the same port, won't there be an error that the port is already in use? How does this option work?

packages/transformers/react-refresh-wrap/src/helpers/helpers.js
```diff
@@ -37,7 +38,16 @@ var enqueueUpdate = debounce(function () {
...
+  window.__REACT_REFRESH_VERSION_TRANSFORMER = version;
   window.$RefreshReg$ = function (type, id) {
+    if (
+      window.__REACT_REFRESH_VERSION_TRANSFORMER !==
+      window.__REACT_REFRESH_VERSION_RUNTIME
+    ) {
+      throw new Error(
+        `react-refresh versions did not match between transformer and runtime. Please check your dependencies. Transformer: ${window.__REACT_REFRESH_VERSION_TRANSFORMER}, Runtime: ${window.__REACT_REFRESH_VERSION_RUNTIME}`,
+      );
+    }
```
devongovett, line 50: Maybe better to enforce this at the package manager level rather than at runtime? Eg yarn constraints or resolutions

packages/reporters/dev-server/src/HMRServer.js
```diff
@@ -272,9 +273,19 @@ export default class HMRServer {
   }
 
   broadcast(msg: HMRMessage) {
-    const json = JSON.stringify(msg);
-    for (let ws of this.wss.clients) {
-      ws.send(json);
+    if (msg.type === 'update' && msg.assets.length > BROADCAST_MAX_ASSETS) {
+      // Split up message if too large
```
devongovett: FYI splitting could result in different behavior in the browser. The logic in hmr-runtime.js does various checks over all assets in a message. Maybe it's ok just pointing it out in case.

packages/transformers/react-refresh-wrap/src/helpers/helpers.js
JakeLane: Yeah the challenge here is that multiple repos were onboarding to parcel and didn't understand the problem, so yarn mitigations are too late. Handling this at build time would be more ideal I think, not sure if there's prior art of managing dependencies like this in Parcel though.

packages/reporters/dev-server/src/HMRServer.js
JakeLane: I didn't bother looking into if it worked correctly and just refactored this by introducing a partial message and gets joined together. Hopefully that preserves the current behaviour sufficently.

packages/runtimes/hmr/src/loaders/hmr-runtime.js
```diff
@@ -139,12 +140,19 @@ if ((!parent || !parent.isParcelRequire) && typeof WebSocket !== 'undefined') {
 
     var data /*: HMRMessage */ = JSON.parse(event.data);
 
-    if (data.type === 'update') {
+    if (data.type === 'partial-update') {
```
marcins: Is there a potential edge case where we "lose" the update message and are left with incomplete / incorrect data here? Should a groups of `partial-update` messages have some sort of unique identifier? We know how many partial-updates we're going to send so they could even contain an `x of y` type thing.

Not sure what the delivery reliability for websockets is like, and whether it's worth the additional complexity.

packages/runtimes/hmr/src/HMRRuntime.js
JakeLane: This was allowing the logic on https://github.com/parcel-bundler/parcel/blob/045662f5839450c60f6a757961d601e7500bbb3a/packages/reporters/dev-server/src/ServerReporter.js#L45 to be forced through, which seemed to fix the configuration in Jira.

I've followed up with a new change which I believe is the root cause, where the server was using HTTPS but the client wasn't.

packages/runtimes/hmr/src/loaders/hmr-runtime.js
JakeLane: I initially didn't want to add that complexity because I assumed that since WebSockets are TCP-based, we can expect correct order and delivery. That may be a poor assumption, and I am happy to refactor if required.

marcins: Yeah, a quick look shows that websocket messages should arrive in the correct order (and won't be "lost" unless the actual connection dies).. I think it's fine as is, and if we have actual issues we can work out a suitable solution then.

yamadapc (approved):

PR #9870, opened 2024-07-25
Use request context in path request

MonicaOlejniczak:
# ↪️ Pull Request

Simplifies the path request instantiation by using the dependencies available in the request context, instead of manually injecting named pipelines and resolvers

* Add `mockall` dependency for testing this change
* Create a `Plugins` trait so that a mock instance can be passed into the request tracker, and move the existing implementation to `ConfigPlugins`
* Consolidate `reporters` into a single composite reporter within `ConfigPlugins` to remove odd coupling in request tracker
* Update path request to use request context and remove those dependencies from the input
* Use `PluginsRef` over `&Plugins` in the request tracker, so mock instances can be passed in for tests

## 🚨 Test instructions

`cargo test`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1719,6 +1719,7 @@ version = "0.1.0"
 dependencies = [
  "anyhow",
  "dyn-hash",
+ "mockall",
  "num_cpus",
  "parcel-resolver",
  "parcel_config",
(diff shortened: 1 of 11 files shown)


mattcompiles (approved):
Nice one, I hit this too but you beat me to it. 😄

yamadapc (approved):
lgtm ;

nit: I think there's a macro that can just be a function

PR #9872, opened 2024-07-25
Add native asset graph request

mattcompiles:
This PR adds the initial implementation for the native Asset graph request. It's it highly inspired from @devongovett's work on the [core-rs2](https://github.com/parcel-bundler/parcel/tree/core-rs2) branch. Outside of the asset graph request it adds/changes the following:
- Parcel options are now available on the request context 
- Cache has been removed from the request context as we are no longer utilising in the Asset request.
- Entries are now always normalised to a vector 
- Request Ids are now available on the queue_request receiver channel


crates/parcel/src/requests/asset_graph_request.rs
```diff
@@ -0,0 +1,235 @@
...
+    }
+
+    let mut visited = HashSet::new();
+    let mut asset_request_to_asset = HashMap::new();
+    let mut waiting_asset_requests = HashMap::<u64, HashSet<NodeIndex>>::new();
+    let mut request_id_to_dep_node_index = HashMap::<RequestId, NodeIndex>::new();
+
+    // This allows us to defer PathRequests that are not yet known to be used as their requested
+    // symbols are not yet referenced in any discovered Assets.
+    macro_rules! on_undeferred {
```
yamadapc: why isn't this a function?

crates/parcel_core/src/asset_graph.rs
```diff
@@ -0,0 +1,345 @@
...
+      AssetGraphNode::Asset(idx) => Some(*idx),
+      _ => None,
+    }
+  }
+
+  /// Propagates the requested symbols from an incoming dependency to an asset,
+  /// and forwards those symbols to re-exported dependencies if needed.
+  /// This may result in assets becoming un-deferred and transformed if they
+  /// now have requested symbols.
+  pub fn propagate_requested_symbols<F: FnMut(NodeIndex, Arc<Dependency>)>(
```
yamadapc, line 164: We should test everything in this file ; including this part

crates/parcel_core/Cargo.toml
```diff
@@ -17,6 +17,7 @@ derive_builder = "0.20.0"
 dyn-hash = "0.x"
 nodejs-semver = "4.0.0"
 mockall = "0.12.1"
+petgraph = { version = "*", features = ["serde-1"] }
```
yamadapc: prob. want to lock this

crates/parcel_core/src/asset_graph.rs
```diff
@@ -0,0 +1,345 @@
...
+        value: SerializedDependency,
+        has_deferred: bool,
+      },
+    }
+
+    #[derive(serde::Serialize)]
+    struct SerializedAssetGraph {
+      nodes: Vec<SerializedAssetGraphNode>,
+      // TODO: somehow make this a typed array?
+      edges: Vec<u32>,
```
yamadapc, line 339: I think we can make it `Buffer` then pass it https://napi.rs/docs/concepts/typed-array

crates/parcel/src/requests/asset_graph_request.rs
```diff
@@ -0,0 +1,235 @@
...
+}
+
+impl Request for AssetGraphRequest {
+  fn run(
+    &self,
+    mut request_context: RunRequestContext,
+  ) -> Result<ResultAndInvalidations, RunRequestError> {
+    let mut graph = AssetGraph::new();
+    let (tx, rx) = channel();
+    // TODO: Move this out later
```
yamadapc: move to context? where is it moving?

crates/parcel/src/requests/asset_graph_request.rs
```diff
@@ -0,0 +1,235 @@
...
+        break;
+      }
+
+      let Ok(result) = rx.recv() else {
+        break;
+      };
+
+      work_count -= 1;
+
+      match result {
```
yamadapc, line 97: we should probably also split and unit-test this

crates/parcel/src/requests/asset_graph_request.rs
```diff
@@ -0,0 +1,235 @@
...
+              .entry(id)
+              .and_modify(|nodes| {
+                nodes.insert(node);
+              })
+              .or_insert_with(|| HashSet::from([node]));
+          }
+        }
+        other => {
+          // This is an error...
+          todo!("{:?}", other);
```
yamadapc: nit nit : assuming these are the testing branches we could just type them out

crates/parcel_core/src/asset_graph.rs
```diff
@@ -0,0 +1,345 @@
...
+      graph,
+      assets: Vec::new(),
+      dependencies: Vec::new(),
+    }
+  }
+
+  pub fn add_asset(&mut self, parent_idx: NodeIndex, asset: Asset) -> NodeIndex {
+    let idx = self.assets.len();
+
+    self.assets.push(AssetNode {
```
yamadapc, line 90: Q: Why is storage split vs. just storing in the graph? 

petgraph will just have an array of nodes internally ; 

https://github.com/petgraph/petgraph/blob/master/src/graph_impl/mod.rs#L348

crates/parcel/src/requests/asset_graph_request.rs
mattcompiles: I updated the comment. We were thinking the work count tracking might be better off as part of the `request_context.queue_request` API.

crates/parcel/src/requests/asset_graph_request.rs
mattcompiles: It was to get around borrow rules. I originally had it as a closure but the compiler was complaining. Happy to make it a function if we can make it work.

yamadapc: Let me try to open a diff against this branch

yamadapc: https://github.com/parcel-bundler/parcel/pull/9873

crates/parcel_core/src/asset_graph.rs
```diff
@@ -0,0 +1,345 @@
...
+        }
+      }
+    }
+  }
+}
+
+#[derive(serde::Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct SerializedAsset {
+  id: u64,
```
MonicaOlejniczak: Can you make this and the `SerializedDependency` id a `String` type and add `.to_string()` in the conversion? It's what breaks the memory fs, since the number is too big for JS

yamadapc: We could reduce the size of these numbers ; to other benefits

Since they will never grow this large

MonicaOlejniczak: Yeah, either is fine. Not sure what the motivation Devon had for making it u64 before.

yamadapc (approved):

PR #9873, opened 2024-07-26
Fix compilation and use function instead of macro

yamadapc:
This fixes compilation of the native asset request branch and uses a function instead of a macro for the undeferred lambda.

Macros are a bit harder to use with editor integration / intellisense.


yamadapc:
We will update files directly rather than merge

PR #9876, opened 2024-07-29
Add asset graph request tests - part 1 of X

yamadapc:
Fix in-memory FS and add a few tests


crates/parcel_plugin_transformer_js/src/transformer.rs
```diff
@@ -60,7 +60,7 @@ impl TransformerPlugin for ParcelJsTransformerPlugin {
 
     let asset = Asset::new_empty(input.file_path().to_path_buf(), source_code);
     let config = parcel_js_swc_core::Config::default();
-    let options = ParcelOptions::default();
+    let options = context.options();
```
yamadapc: This was a bug as options weren't forwarded properly.

crates/parcel_filesystem/src/in_memory_file_system.rs
```diff
@@ -22,39 +22,26 @@ pub struct InMemoryFileSystem {
 
 impl InMemoryFileSystem {
```
yamadapc, line 23: This didn't respect CWD when reading/writing files

crates/parcel/src/requests/asset_graph_request.rs
```diff
@@ -344,15 +344,208 @@ impl AssetGraphBuilder {
...
+        .assets
+        .get(0)
+        .unwrap()
+        .asset
+        .file_path,
+      PathBuf::from("/parcel_tests/entry.js")
+    );
+  }
+
+  fn setup_core_modules(fs: &InMemoryFileSystem, core_path: &Path) {
```
yamadapc, line 529: We can do better solution but it's nice to see it running E2E regardless

crates/parcel/src/request_tracker/request_tracker.rs
```diff
@@ -96,7 +96,12 @@ impl RequestTracker {
   #[allow(unused)]
   pub fn run_request(&mut self, request: impl Request) -> anyhow::Result<RequestResult> {
     let thread_pool = rayon::ThreadPoolBuilder::new()
+      .thread_name(|count| format!("RequestTracker-{}", count))
       .num_threads(num_cpus::get() + 4)
+      .panic_handler(|failure| {
+        tracing::error!("Lost thread from thread-pool. This is a bug in parcel. Builds may stall.");
```
yamadapc, line 102: This actually did not fix things ; atm we get silent panics from these threads

mattcompiles (approved):

crates/parcel/src/request_tracker/request_tracker.rs
```diff
@@ -255,7 +260,7 @@ impl RequestTracker {
     match request_node {
       RequestNode::Root => Err(diagnostic_error!("Impossible")),
       RequestNode::Incomplete => Err(diagnostic_error!("Impossible")),
-      RequestNode::Error(error) => Err(anyhow!(error.clone())),
+      RequestNode::Error(error) => Err(anyhow!(error.to_string())),
```
MonicaOlejniczak, line 263: I believe this will butcher the error output, specifically stack traces will be missing when enabled as an env var

Can we go back to cloning it or find another alternative?

crates/parcel_filesystem/src/in_memory_file_system.rs
```diff
@@ -22,39 +22,26 @@ pub struct InMemoryFileSystem {
...
   /// Change the current working directory. Used for resolving relative paths.
-  pub fn set_current_working_directory(&self, cwd: PathBuf) {
+  pub fn set_current_working_directory(&self, cwd: &Path) {
+    let cwd = self.canonicalize_impl(cwd);
```
MonicaOlejniczak, line 26: Can we fix the trait in these changes? I'm not a fan of the `_base` fn for the default usage, we should just have `canonicalize` and the one that needs a cache should be named explicitly... something like `canonicalize_with_cache`

crates/parcel/src/request_tracker/request_tracker.rs
yamadapc: It's the other way around actually ; before this diff `RequestNode::Error` is a string, 

Now I'm carrying the inner error but we still need to convert to string here unfortunately. We needed a clonable error type to fix it.

I was changing because atm strack-traces are missing.

crates/parcel_filesystem/src/in_memory_file_system.rs
yamadapc: There're a few things I think we could change in `trait FileSystem`, but that's really not the intention of this PR.

I just wanted a non-failable version of canonicalize for use internally.

MonicaOlejniczak: Yeah, I understand that. A follow up at some point would be nice.

crates/parcel/src/request_tracker/request_tracker.rs
MonicaOlejniczak: Oh right, I thought that was fixed. We should revisit this at some point.

yamadapc:
There's something broken on windows that causes the paths to be stripped of the prefix

crates/parcel_filesystem/src/in_memory_file_system.rs
```diff
@@ -69,7 +56,7 @@ impl FileSystem for InMemoryFileSystem {
           result = vec![Component::Prefix(prefix)];
         }
         Component::RootDir => {
-          result = vec![Component::RootDir];
+          result.push(Component::RootDir);
```
yamadapc, line 59: This was a bug as on windows the prefix would be stripped

yamadapc:
> There's something broken on windows that causes the paths to be stripped of the prefix

Fixed by https://github.com/parcel-bundler/parcel/pull/9876#discussion_r1697734943

PR #9877, opened 2024-07-29
Integrate asset graph

MonicaOlejniczak:
# ↪️ Pull Request

Integrates the asset graph request with a working commonjs test using the `parcelV3` feature flag

## 🚨 Test instructions

`cargo test`

### DIFF
--- a/crates/parcel_core/src/types/asset.rs
+++ b/crates/parcel_core/src/types/asset.rs
@@ -20,7 +20,7 @@ pub struct AssetId(pub NonZeroU32);
 
 /// The source code for an asset.
 #[derive(PartialEq, Default, Clone, Debug, Deserialize, Serialize)]
-#[serde(rename_all = "camelCase")]
+#[serde(rename_all = "camelCase", transparent)]
 pub struct Code {
   inner: String,
 }
--- a/packages/core/core/src/CommittedAsset.js
+++ b/packages/core/core/src/CommittedAsset.js
@@ -6,10 +6,12 @@ import type {Asset, Dependency, ParcelOptions} from './types';
 import {Readable} from 'stream';
 import SourceMap from '@parcel/source-map';
 import {bufferStream, blobToStream, streamFromPromise} from '@parcel/utils';
+import {getFeatureFlag} from '@parcel/feature-flags';
 import {generateFromAST} from './assetUtils';
 import {deserializeRaw} from './serializer';
 
 export default class CommittedAsset {
+  key: ?string;
   value: Asset;
   options: ParcelOptions;
   content: ?Promise<Buffer | string>;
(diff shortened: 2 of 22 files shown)


packages/core/core/src/CommittedAsset.js
```diff
@@ -21,16 +23,19 @@ export default class CommittedAsset {
 
   constructor(value: Asset, options: ParcelOptions) {
     this.value = value;
+    this.key = getFeatureFlag('parcelV3')
+      ? this.value.id
+      : this.value.contentKey;
```
mattcompiles, line 28: This is probably fine for now but I'm not sure if not having a contentKey is going to cause problems anywhere else. I believe id is a hash of the filePath, where as content key might be more specific and include contents as well.

yamadapc: I reckon the content key is a cached ID. That means when the inputs change, the id may change but not this content key.

crates/parcel_core/src/types/asset.rs
```diff
@@ -20,7 +20,7 @@ pub struct AssetId(pub NonZeroU32);
 
 /// The source code for an asset.
 #[derive(PartialEq, Default, Clone, Debug, Deserialize, Serialize)]
-#[serde(rename_all = "camelCase")]
+#[serde(rename_all = "camelCase", transparent)]
```
mattcompiles, line 23: TIL

packages/core/core/src/Parcel.js
```diff
@@ -115,6 +117,23 @@ export default class Parcel {
       this.#initialOptions,
     );
     this.#resolvedOptions = resolvedOptions;
+
+    if (resolvedOptions.featureFlags.parcelV3) {
+      // eslint-disable-next-line no-unused-vars
+      let {entries, inputFS, outputFS, ...options} = this.#initialOptions;
+
+      this.#parcelV3 = new ParcelV3({
```
mattcompiles: Nitpick, but do we really need to store on `this`? Looks like it could be local to this method?

packages/utils/node-resolver-rs/src/lib.rs
```diff
@@ -477,7 +477,7 @@ impl<'a> ResolveRequest<'a> {
           && self.resolver.flags.contains(Flags::EXPORTS)
         {
           // An internal package #import specifier.
-          let package = self.find_package(self.from.parent().unwrap())?;
+          let package = self.find_package(self.from.parent().unwrap_or_else(|| self.from))?;
```
mattcompiles, line 480: Any context on this one? This changes behaviour outside of the feature flag right?

packages/core/core/src/Parcel.js
MonicaOlejniczak: I can make it local, I think I had it designed a different way initially where it was necessary then

packages/utils/node-resolver-rs/src/lib.rs
MonicaOlejniczak: It should be pretty safe. It was to pass some of the integration tests but I can't recall which one's atm. The original code can cause a legitimate failure when `self.from` is the root directory. We shouldn't fail when `from` is the root, as `self.from.parent()` will be `None` in that case.

packages/core/core/src/CommittedAsset.js
MonicaOlejniczak: It's just to get it working as a first pass, happy for it to change when necessary

mattcompiles (approved):

PR #9883, opened 2024-07-31
Fix JsCallable error handling for fs

MonicaOlejniczak:
# ↪️ Pull Request

When the Rust `JsCallable` calls a JavaScript function that throws an error, the process panics instead of returning the error as a result.

> thread 'FATAL ERROR: threadsafe_function.rs:749 Failed to convert return value in ThreadsafeFunction callback into Rust value: GenericFailure, Error: ENOENT: no such file or directory, open '/parcel/packages/core/integration-tests/test/integration/node_require_browser/node_modules/testmodule/package.json'
>
> <unnamed>' panicked at /parcel/crates/parcel_napi_helpers/src/js_callable/js_callable.rs:140:15: called `Result::unwrap()` on an `Err` value: RecvError

This occurs because the napi value, invoked by the threadsafe function, never returns `is_error()` as true. Therefore, the logic will fallback to the `map_return` [branch](https://github.com/parcel-bundler/parcel/blob/v2/crates/parcel_napi_helpers/src/js_callable/js_callable.rs#L142) and fail when converting an JavaScript Error type to the expected Rust value (usually a string or boolean). There doesn't seem to be any alternative API to make napi return the thrown error as an error, since the callback is never called by napi when an error is thrown as per [here](https://github.com/napi-rs/napi-rs/blob/327998d0aca2302e516fb2f06ee9ca59f8fc2799/crates/napi/src/threadsafe_function.rs#L571).

```rust
d.and_then(|d| D::from_napi_value(d.0.env, d.0.value).and_then(cb))
```

These changes introduce a `jsCallable` on the JavaScript side that must be used with `JsCallable` on the Rust side, so that errors are returned instead of thrown. I have also split out the worker / worklet tests from `javascript.js` file, to better group those tests and reduce the size of the latter file as part of working through the integration test suite.

## 🚨 Test instructions

`cargo test`


mattcompiles (approved):

alshdavid:
Nice work 🥇

PR #9884, opened 2024-08-01
Forward missing env and transformer inputs

MonicaOlejniczak:
# ↪️ Pull Request

These changes primarily forward through missing inputs from the transformer plugin to the transformer, as well as the asset environment. In doing so, the target and dependency env have been updated to an arc which is consistent with other definitions.

**Future improvements:**
* Actually share the dependency env instead of creating new instances, or intern it
* Pass through remaining inputs to the transformer based on config or otherwise

## 🚨 Test instructions

`cargo test`

### DIFF
--- a/crates/parcel_plugin_transformer_js/src/transformer.rs
+++ b/crates/parcel_plugin_transformer_js/src/transformer.rs
@@ -2,7 +2,8 @@ use anyhow::{anyhow, Error};
 
 use parcel_core::plugin::TransformerPlugin;
 use parcel_core::plugin::{RunTransformContext, TransformResult, TransformationInput};
-use parcel_core::types::Asset;
+use parcel_core::types::engines::EnvironmentFeature;
+use parcel_core::types::{Asset, BuildMode, FileType, LogLevel, OutputFormat, SourceType};
 
 mod conversion;
 #[cfg(test)]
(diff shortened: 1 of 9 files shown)


alshdavid (approved):

crates/parcel_plugin_transformer_js/src/transformer.rs
```diff
@@ -69,7 +87,22 @@ impl TransformerPlugin for ParcelJsTransformerPlugin {
...
+        .and_then(|s| s.to_str())
+        .unwrap_or_default(),
+    );
+
+    let asset = Asset {
+      asset_type,
+      code: source_code,
+      env: env.clone(),
+      file_path: file_path.to_path_buf(),
+      ..Asset::default()
```
mattcompiles, line 103: I'm not 100% sure, but if there's any properties that can be set by previous transformers this will override them with the defaults.

mattcompiles, line 103: I'm not 100% sure, but if there's any properties that can be set by previous transformers this will override them with the defaults.

MonicaOlejniczak: I see, it's what the previous code expands out to so will need to be fixed later.

PR #9886, opened 2024-08-01
Create Parcel V3 javascript integration test suite

mattcompiles:
This PR copies the javascript integration test suite, running all the tests with the `parcelV3` feature flag enabled. Eventually we should probably run the entire suite with the flag enabled but while we're so early it's a bit easier to keep things simple. 

Let's remove the skips as we fix the tests 👍


MonicaOlejniczak (approved):

devongovett:
Wouldn't it be easier to work on a branch instead of copying files? Or maybe add a function to conditionally skip tests based on a feature flag. Otherwise whenever we make changes to the tests we'll have to do it in two places.

PR #9889, opened 2024-08-02
Add test setup for v3

MonicaOlejniczak:
# ↪️ Pull Request

Adds V3 tests to CI via a new testing API that utilises a `PARCEL_V3` environment variable to switch between which set of tests are to be run

Note that the `parcelVersion` logic may need to be refined later as needed

## 🤔 Considerations
**Globals**

I initially tried to override the globally defined `describe` and `it` rather than having to directly import the wrappers, but there are a couple of issues:
* when the globals are defined in another file such as the test utils, it either results in a circular reference or mocha is unable to attach the `it` handler due to its intertwined suite [context](https://github.com/mochajs/mocha/issues/2765#issuecomment-293457842)
* the globals must be inlined in the test file itself, which even if generated will not have a good developer experience (lack of findable definitions, and slower performance due to code injection)

It also wasn't possible to use a [root hook plugin](https://mochajs.org/#defining-a-root-hook-plugin) or [global fixture](https://mochajs.org/#global-fixtures) to inject the values.

## 💻 Examples

**Run tests in both v2 and v3**

```javascript
describe('parcel', () => {
  it('bundles');
});
```

---

**Run tests in v2 only**

```javascript
describe.v2('parcel', () => {
  it('bundles');
});

it.v2('test');
```

---

**Run tests in v3 only**

```javascript
describe.v3('parcel', () => {
  it('bundles');
});

it.v3('test');
```

---

**Run tests selectively in v2, v3, or both**

```javascript
describe('parcel', () => {
  it('bundles for both v2 and v3');

  it.v2('bundles for v2');

  it.v3('bundles for v3');
});
```

## 🚨 Test instructions

```sh
yarn test:integration

PARCEL_V3=false yarn test:integration

PARCEL_V3=true yarn test:integration
```


packages/core/integration-tests/test/javascript.js
```diff
@@ -717,7 +765,7 @@ describe('javascript', function () {
     assert.equal(await output(), 7);
   });
 
-  it('should not duplicate a module which is already in a parent bundle', async function () {
+  it.skip('should not duplicate a module which is already in a parent bundle', async function () {
```
mattcompiles: Did you mean to skip these?

MonicaOlejniczak: Nope! Probably a bad remnant from where I branched off, I'll double check for others and fix it

mattcompiles (approved):

PR #9891, opened 2024-08-02
Native: Update meta forwarding in JS conversion to match JSTransformer.js

mattcompiles:


crates/parcel_plugin_transformer_js/src/transformer/conversion.rs
```diff
@@ -507,6 +514,18 @@ fn convert_dependency(
     )),
     _ => {
       let mut env = asset.env.clone();
+      base_dependency.set_kind(format!("{}", transformer_dependency.kind));
```
mattcompiles, line 517: This one is a bit dodgy. Happy to write a match if we'd prefer, however it'd just map to the exact enum values.

crates/parcel_core/src/types/dependency.rs
```diff
@@ -157,6 +158,40 @@ impl Dependency {
...
+    self.meta.insert("webworker".into(), true.into());
+  }
+
+  pub fn set_kind(&mut self, kind: impl Into<serde_json::Value>) {
+    self.meta.insert("kind".into(), kind.into());
+  }
+
+  pub fn set_should_wrap(&mut self, should_wrap: bool) {
+    self.meta.insert("shouldWrap".into(), should_wrap.into());
+    self.should_wrap = should_wrap;
```
MonicaOlejniczak, line 176: Thoughts on constructing meta during serialization of the asset graph for fields also directly stored on the dep, or do you think that will be too expensive?

mattcompiles: I think either is probably fine. I opted to do it like this as I think it's simpler to debug and understand. Rather than having a bunch of magic properties appear as it's serialised.

MonicaOlejniczak (approved):

PR #9892, opened 2024-08-02
Improve crash reporting integration

yamadapc:
_All of these changes are specific to canary builds and opt-in. They're meant to track internal reliability of parcel at Atlassian_

---

We lack visibility on native crashes coming from parcel, but reports of crashes are frequent.

This commit makes the following changes to error monitoring integration:

* Tracing is set-up for every build not just v3 builds. This is so we can get visibility on the mainline rust code. This is disabled by default.
* Tracing/Sentry code are moved to a new crate
* Crash reporting integration is added with `minidumper`, to allow us to get more information about crashes.

---

`minidumper` is picked rather than the [`sentry-native`](https://blog.sentry.io/debugging-a-segfault-in-rust/#) SDK because providing the SDK for all supported parcel architectures is not something we'd like to maintain.

Previously written Rust / build wrappers for that SDK (which is a C dependency) have been deprecated in favour of this approach. https://github.com/daxpedda/sentry-contrib-native.


crates/parcel_monitoring/src/tracer.rs
```diff
@@ -1,16 +1,15 @@
...
 use tracing_appender::non_blocking::WorkerGuard;
 
+use crate::from_env::{optional_var, FromEnvError};
+
 #[derive(Debug, Deserialize, Serialize)]
 #[serde(rename_all = "camelCase", tag = "mode")]
-pub enum TracerMode {
-  /// Disable the Tracer
-  Disabled,
+pub enum TracingOptions {
```
MonicaOlejniczak: How come you favour `Options` over `Mode`? The latter makes more sense to me, since there's only one variant to be used at a time

yamadapc: Changed back

packages/core/core/src/Parcel.js
```diff
@@ -104,9 +108,9 @@ export default class Parcel {
     await initSourcemaps;
     await initRust?.();
     try {
-      initSentry?.();
+      initializeMonitoring?.();
```
MonicaOlejniczak, line 111: /nit I think a class instance would work better here, e.g.

```javascript
let monitoring = new ParcelMonitoring();

process.on('exit', () => {
  monitoring.close();
});
```

crates/parcel_monitoring/src/sentry_integration.rs
```diff
@@ -0,0 +1,173 @@
...
+  let guard = sentry::init((options.sentry_dsn, sentry_client_options));
+
+  configure_scope(|scope| {
+    scope.set_user(Some(sentry::User {
+      username: Some(username()),
+      ..Default::default()
+    }));
+  });
+
+  configure_scope(|scope| {
```
MonicaOlejniczak: Can you collapse this in the previous `configure_scope` or does the closure / borrowing complain?

crates/parcel_monitoring/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+Parcel uses
+[crash-handling](https://github.com/EmbarkStudios/crash-handling/blob/e2891a4c6a8d43374ec63d791c7e6d42ff2e6545/README.md)
+utilities to write [minidumps](https://github.com/EmbarkStudios/crash-handling/tree/main/minidumper) on crashes.
+
+When this feature is enabled, the parcel process will try to IPC with a server process using the
+[`minidumper`](https://github.com/EmbarkStudios/crash-handling/tree/main/minidumper) library. The server should write
+and report the crash.
+
+A default server implementation is not provided, but an example can be seen on `examples/sample_usage.rs`.
```
MonicaOlejniczak, line 50: Can it be run? Instructions?

MonicaOlejniczak (approved):

crates/parcel_monitoring/README.md
yamadapc: Yes it can be ran with `cargo run --example sample_usage` ;

crates/parcel_monitoring/src/sentry_integration.rs
yamadapc: Done ;

packages/core/core/src/Parcel.js
yamadapc: This maps to a function so I don't think there's benefit to wrapping it in a class.

MonicaOlejniczak: I was also suggesting to change what it calls to a struct

yamadapc: There's no reason to do that.

PR #9894, opened 2024-08-05
Fix MUSL canary builds

yamadapc:
Fixes MUSL canary builds by removing crash-reporting functionality from MUSL targets.

The reason why crash-reporting isn't working is not fully understood but relates to how crash-handler is implemented interposing the `pthread_create` value. https://github.com/EmbarkStudios/crash-handling/blob/e1f1241580d5204880fa47802b867b6bcb975f98/crash-handler/src/unix/pthread_interpose.rs#L36-L41

`crash-reporting` has been added as a part of https://github.com/parcel-bundler/parcel/pull/9892 and will only target canary releases at this point.


yamadapc:
I believe the issue comes from https://github.com/EmbarkStudios/crash-handling crates not compiling on MUSL.

yamadapc:
The problem is - https://github.com/EmbarkStudios/crash-handling/blob/e1f1241580d5204880fa47802b867b6bcb975f98/crash-handler/src/unix/pthread_interpose.rs#L31-L42

JakeLane (approved):

PR #9902, opened 2024-08-09
Make optimisations to resolver implementation

yamadapc:
Replaced by https://github.com/parcel-bundler/parcel/pull/9903

The main change in this diff is removing an extra `is_dir` / `is_file` call which I had added.

We ran benchmarks replaying all imports of a large application against multiple resolver implementations. In practice, the previous code had around 10% worse performance than the shared structures code. The difference in performance goes away by removing those extra `is_dir` and `is_file` calls. There is no difference achieved from preventing copies.

I'll attach performance profiles captured on macOS.

After the fixes in this commit, the new resolver crate out-performs the older one by 26-27%.

That is due to a change on top of the FS calls revert, which I've found through profiling. Hashes have been replaced for both implementations' filesystem canonicalize cache, which is the hottest by a faster hashing function.

The worse culprit of CPU time at the moment is hashing file-paths on this benchmark. This might be due to how the file-paths are passed into the benchmark, but we're constantly re-hashing file-path components to store in these cache hash-maps.

The total time of resolving all the application level imports on multiple threads (~500k) on this application is:

* 2.13 seconds for M1 Pro (new branch)
* 2.88 seconds for M1 Pro (old branch)

This is consistent across runs and based on an average of 50 executions over this entire list.

On a large Linux EC2 instance the results are similar, but resolution is even faster.

After path hashing, the next hottest path are the file-system calls.

DashMap does not appear to have a huge impact on performance on its own, and is showing-up on crash reports. I'm adding it back hoping it's not the culprit.

- - -

Note, this is relatively incredibly fast and therefore not a focus for performance optimisation.

We're still experiencing crashes from this area, which we'll investigate further.

### DIFF
--- a/crates/parcel_filesystem/Cargo.toml
+++ b/crates/parcel_filesystem/Cargo.toml
@@ -7,6 +7,8 @@ description = "FileSystem wrapper trait for use in Parcel codebase."
 [dependencies]
 dashmap = "5.5.3"
 mockall = "0.12.1"
+gxhash = "3.4.1"
+anyhow = "1.0.86"
 
 [dev-dependencies]
 assert_fs = "1.1.1"
--- a/crates/parcel_filesystem/Cargo.toml
+++ b/crates/parcel_filesystem/Cargo.toml
@@ -8,7 +8,6 @@ description = "FileSystem wrapper trait for use in Parcel codebase."
 dashmap = "5.5.3"
 mockall = "0.12.1"
 xxhash-rust = { version = "0.8.2", features = ["xxh3"] }
-gxhash = "3.4.1"
 anyhow = "1.0.86"
 
 [dev-dependencies]
--- a/crates/parcel_filesystem/Cargo.toml
+++ b/crates/parcel_filesystem/Cargo.toml
@@ -7,6 +7,7 @@ description = "FileSystem wrapper trait for use in Parcel codebase."
 [dependencies]
 dashmap = "5.5.3"
 mockall = "0.12.1"
+xxhash-rust = { version = "0.8.2", features = ["xxh3"] }
 gxhash = "3.4.1"
 anyhow = "1.0.86"
 
--- a/packages/utils/node-resolver-rs/src/cache.rs
+++ b/packages/utils/node-resolver-rs/src/cache.rs
@@ -16,19 +16,20 @@ use crate::tsconfig::TsConfig;
 use crate::tsconfig::TsConfigWrapper;
 use crate::ResolverError;
 
+type DefaultHasher = xxhash_rust::xxh3::Xxh3Builder;
+
 pub struct Cache {
   pub fs: FileSystemRef,
   /// These map paths to parsed config files. They aren't really 'static, but Rust doens't have a good
   /// way to associate a lifetime with owned data stored in the same struct. We only vend temporary references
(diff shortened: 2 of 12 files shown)


packages/utils/node-resolver-rs/src/cache.rs
```diff
@@ -86,15 +89,13 @@ impl JsonError {
...
-    let packages = HashMap::new();
-    let tsconfigs = HashMap::new();
     Self {
       fs,
-      packages: RwLock::new(packages),
-      tsconfigs: RwLock::new(tsconfigs),
-      is_file_cache: DashMap::new(),
-      is_dir_cache: DashMap::new(),
-      realpath_cache: DashMap::new(),
+      packages: DashMap::with_hasher(gxhash::GxBuildHasher::default()),
```
MonicaOlejniczak, line 94: Last I checked this won't compile on musl and some other platforms

crates/parcel_filesystem/Cargo.toml
```diff
@@ -7,6 +7,8 @@ description = "FileSystem wrapper trait for use in Parcel codebase."
 [dependencies]
 dashmap = "5.5.3"
 mockall = "0.12.1"
+gxhash = "3.4.1"
```
MonicaOlejniczak: Last I checked this won't compile on musl and some other platforms

You also need to enable aes

packages/utils/node-resolver-rs/src/cache.rs
```diff
@@ -22,11 +22,14 @@ pub struct Cache {
...
-  is_file_cache: DashMap<PathBuf, bool>,
-  is_dir_cache: DashMap<PathBuf, bool>,
-  realpath_cache: DashMap<PathBuf, Option<PathBuf>>,
+  packages: DashMap<PathBuf, Arc<Result<Arc<PackageJson>, ResolverError>>, gxhash::GxBuildHasher>,
+  tsconfigs:
+    DashMap<PathBuf, Arc<Result<Arc<TsConfigWrapper>, ResolverError>>, gxhash::GxBuildHasher>,
+  // In particular just the is_dir_cache spends around 8% of the time on a large project resolution
+  // hashing paths. Instead of using a hashmap we should try a trie here.
+  is_dir_cache: DashMap<PathBuf, bool, gxhash::GxBuildHasher>,
+  is_file_cache: DashMap<PathBuf, bool, gxhash::GxBuildHasher>,
```
devongovett: Something I found before but never got around to committing was that making the keys `OsString` instead of `PathBuf` made the hashing significantly faster. I think that should be equivalent as long as the path separators are normalized (but even then it shouldn't matter for correctness, would just result in multiple entries for the same file).

yamadapc: Most of the hashing time is spent in the function body of the `Path` hash function (rather than the hasher).

So I think it might be faster to just use strings, but the hashes will be different if any path key isn't normalized yes.

crates/parcel_filesystem/Cargo.toml
yamadapc: Changing to xxhash doesn't change the results very much so I think I'll replace with that.

JakeLane (approved):

yamadapc:
I suppose the REPL deployment is failing due to this being on my own github fork rather than this repository (which was an accident.

I will re-open this PR.

PR #9906, opened 2024-08-12
Initiate watchman watch session if not already started

mattcompiles:
Previous to this change the watchman watcher would error our if the watchman watch session wasn't already running.

### DIFF
--- a/packages/utils/parcel-watcher-watchman-js/src/wrapper.js
+++ b/packages/utils/parcel-watcher-watchman-js/src/wrapper.js
@@ -3,6 +3,7 @@ import * as fs from 'fs';
 import * as path from 'path';
 import * as watchman from 'fb-watchman';
 import {isGlob} from '@parcel/utils';
+import logger from '@parcel/logger';
 import type {
   Options,
   Event,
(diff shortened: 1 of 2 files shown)


packages/utils/parcel-watcher-watchman-js/src/wrapper.js
```diff
@@ -41,26 +42,73 @@ export interface Watcher {
...
-  commandAsync(args: any[]): Promise<any> {
+  init(watchDir: string): Promise<void> {
+    if (this.initPromise) {
+      return this.initPromise;
+    }
+
+    const client = this.client;
+
+    this.initPromise = new Promise((resolve, reject) => {
+      client.capabilityCheck({optional: [], required: []}, error => {
```
yamadapc: Could we write with proper promises code?

packages/utils/parcel-watcher-watchman-js/src/wrapper.js
```diff
@@ -41,26 +42,73 @@ export interface Watcher {
...
+        });
+      });
+    });
+
+    return this.initPromise;
+  }
+
+  async commandAsync(
+    command: string,
+    watchDir: string,
```
yamadapc: Not all watchman commands receive these two parameters ; you can just make it variadic if you don't want the array wrapper on the calls

packages/utils/parcel-watcher-watchman-js/src/wrapper.js
yamadapc: That is; write

```
function capabilityCheckAsync(client, opts) {
   // ...
}
```

And then use that + commandAsync here ;

packages/utils/parcel-watcher-watchman-js/src/wrapper.js
yamadapc: Either keep as it was or make `args` variadic (`...args: `)

yamadapc (approved):

JakeLane (approved):

packages/utils/parcel-watcher-watchman-js/src/wrapper.js
mattcompiles: I changed it because we need access to watchdir to ensure things are initialised. All the commands we actually use in the client follow this pattern. I can change though if you prefer.

yamadapc: Understand. Is this what the C++ wrapper does? It spawns `watch ...` for whatever directory it got commands for?

mattcompiles: I guess so. This is just how watchman works.

PR #9907, opened 2024-08-12
Disable jemalloc on canary macOS builds

yamadapc:
We get a few memory errors around jemalloc code.

This points to:

* Memory corruption caused by some other faulty code, which I believe could be lmdb-js
* Allocator has problems

I want to remove it on canary builds to see if it addresses our problems. This will also provide an understanding of what impact it has on performance if any.

### DIFF
--- a/crates/node-bindings/Cargo.toml
+++ b/crates/node-bindings/Cargo.toml
@@ -59,6 +59,8 @@ getrandom = { version = "0.2", features = ["custom"], default-features = false }
 napi = { version = "2.16.4", features = ["serde-json"] }
 
 [target.'cfg(all(target_os = "macos", not(feature = "canary")))'.dependencies]
+# jemallocator is disabled on canary builds to experiment with its performance
+# and reliability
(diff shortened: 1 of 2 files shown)


crates/node-bindings/Cargo.toml
```diff
@@ -58,7 +58,7 @@ rayon = "1.7.0"
 getrandom = { version = "0.2", features = ["custom"], default-features = false }
 napi = { version = "2.16.4", features = ["serde-json"] }
 
-[target.'cfg(target_os = "macos")'.dependencies]
+[target.'cfg(all(target_os = "macos", not(feature = "canary")))'.dependencies]
```
mattcompiles, line 61: Maybe just leave a comment here to specify why it's being disabled for canary?

yamadapc: Added a comment

yamadapc:
I see I opened this from my fork again by accident. I'll re-open

PR #9910, opened 2024-08-13
Forward tsconfig transformer options

MonicaOlejniczak:
# ↪️ Pull Request

Passes in the relevant tsconfig options from the transformer plugin into the transformer crate. Refactors some of the signatures to distinguish once off operations / setup as well to be consistent with other plugins.

## 🚨 Test instructions

`yarn build-native && cargo test && yarn test:integration`


mattcompiles (approved):
Nice

PR #9912, opened 2024-08-14
WIP: Fixing repl

alshdavid:
There were two issues:
- `env.from_js_value()` was failing at runtime for wasm builds (I'm not sure why atm)
  - Changed this from `env.from_js_value` to manually constructing the target struct
- `@parcel/node-resolver-core/ResolverBase` had a conditional statement to select the old vs new resolver which was failing at runtime for wasm builds.
  - Changed this from a `BaseResolver` property to a `getBaseResolver()` function

### DIFF
--- a/crates/node-bindings/src/transformer.rs
+++ b/crates/node-bindings/src/transformer.rs
@@ -5,7 +5,12 @@ use napi_derive::napi;
 
 #[napi]
 pub fn transform(opts: JsObject, env: Env) -> napi::Result<JsUnknown> {
+  #[cfg(target_arch = "wasm32")]
+  console_log::init_with_level(Level::Debug);
+
+  log::info!("transform called");
   let config: parcel_js_swc_core::Config = env.from_js_value(opts)?;
+  log::info!("transform called");
 
   let result = parcel_js_swc_core::transform(config, None)?;
   env.to_js_value(&result)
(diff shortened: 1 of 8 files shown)


crates/node-bindings/src/transformer.rs
```diff
@@ -1,11 +1,49 @@
+use napi::bindgen_prelude::Buffer;
 use napi::Env;
 use napi::JsObject;
 use napi::JsUnknown;
 use napi_derive::napi;
 
 #[napi]
 pub fn transform(opts: JsObject, env: Env) -> napi::Result<JsUnknown> {
-  let config: parcel_js_swc_core::Config = env.from_js_value(opts)?;
+  let config = parcel_js_swc_core::Config {
```
alshdavid, line 9: @devongovett Any idea on why `from_js_value` fails to deserialize the value for wasm but not napi? The error was `expected $name, got -1`. 

Is it okay to commit manually constructing the struct like this rather than using serde?

yamadapc: Did we change things in `Config` since the repl broke?

yamadapc: Is it possible to log the value getting passed?

devongovett: Looks like that error comes from here: https://github.com/napi-rs/napi-rs/blob/8b6cc10d779f94601775910e90d22067987fb0f7/crates/napi/src/bindgen_runtime/js_values/arraybuffer.rs#L702

Could be due to the upgrade of napi-rs somehow being incompatible with napi-wasm. I will investigate.

devongovett:
Closing in favor of #9936.

PR #9914, opened 2024-08-15
Fix completed asset graph edge ordering

MonicaOlejniczak:
# ↪️ Pull Request

These changes fix a bug in the asset graph request when requests have already been completed. Previously the edge was being added from the child to the parent, causing resolution failures in tests within the dev packager when nodes could not be found in the adjacency list. Now, the edge is correctly formed as a parent to child relationship.

## 🚨 Test instructions

`PARCEL_V3=true yarn test:integration`


mattcompiles (approved):
👏

yamadapc (approved):

PR #9968, opened 2024-09-29
Update dependencies

### LINKED ISSUE #9739: Please upgrade `sharp` to version >0.32.6 due to **security vulnerability**
# 🙋 feature request

When using image optimization, parcel currently installs version `^0.31.1` of `sharp`, as hardcoded here: https://github.com/parcel-bundler/parcel/blob/38635b59ded912d6d6a1f8b174299a392ffcbd02/packages/transformers/image/src/loadSharp.js#L5

Unfortunately, all versions of `sharp` below `0.32.6` have a security vulnerability in the libwebp dependency, more information can be found here: https://github.com/advisories/GHSA-54xq-cgqr-rpm3
Dependabot and `pnpm` are constantly reporting this vulnerability, but I can't upgrade to a version above `0.31.3` since that breaks my parcel build (`Error: Could not find module "sharp" satisfying ^0.31.1.`).

## 🤔 Expected Behavior

Parcel installs `sharp` version `0.32.6` or above.

## 😯 Current Behavior

Parcel install `sharp` version `0.31.3`.

## 💁 Possible Solution

Bump up the version in this line: https://github.com/parcel-bundler/parcel/blob/38635b59ded912d6d6a1f8b174299a392ffcbd02/packages/transformers/image/src/loadSharp.js#L5

And maybe check the [changelog of `sharp`](https://sharp.pixelplumbing.com/changelog) to make sure nothing breaks.

## 🔦 Context

I'm simply trying to make use of the image optimization and resizing feature. I have the following line in my HTML and I'm trying to get it to work: ``

devongovett:
Fixes #9739


Vresod:
I think this fix is important enough to warrant an immediate minor release.

If not that, do you expect that the fix for #9739 will be in a release-version of Parcel anytime soon?

PR #9969, opened 2024-09-30
Load svgo version from project

### LINKED ISSUE #8948: Unmet peer dependency svgo 3 in optimizer-htmlnano

### LINKED ISSUE #9660: To update svgo to version 3.0 in @parcel/optimizer-htmlnano.
> # 🐛 bug report
> The deprecated warning for stable@0.1.8 is displayed during Parcel installation.
> 
> ```
> Downloading @parcel/rust@2.12.0: 65.58 MB/65.58 MB, done
> Downloading @swc/core-win32-x64-msvc@1.4.16: 16.65 MB/16.65 MB, done
>  WARN  1 deprecated subdependencies found: stable@0.1.8  <-----------
> Packages: +172
> ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
> Progress: resolved 210, reused 0, downloaded 172, added 172, done
> node_modules/.pnpm/@swc+core@1.4.16_@swc+helpers@0.5.10/node_modules/@swc/core: Running postinstall script, done in 69ms
> node_modules/.pnpm/msgpackr-extract@3.0.2/node_modules/msgpackr-extract: Running install script, done in 153ms
> node_modules/.pnpm/lmdb@2.8.5/node_modules/lmdb: Running install script, done in 188ms
> 
> devDependencies:
> + parcel 2.12.0
> 
>  WARN  Issues with peer dependencies found
> .
> └─┬ parcel 2.12.0
>   └─┬ @parcel/config-default 2.12.0
>     └─┬ @parcel/optimizer-htmlnano 2.12.0
>       └─┬ htmlnano 2.1.0
>         └── ✕ unmet peer svgo@^3.0.2: found 2.8.0 in @parcel/optimizer-htmlnano
> 
> Done in 10.3s
> ```
> 
> ## 🎛 Configuration (.babelrc, package.json, cli command)
> ```shell
> pnpm i parcel -D
> ```
> 
> ## 🤔 Expected Behavior
> Do not display the deprecated warning.
> 
> ## 😯 Current Behavior
> The console outputs "WARN 1 deprecated subdependencies found: stable@0.1.8."
> 
> ## 💁 Possible Solution
> https://github.com/parcel-bundler/parcel/blob/91b47ff5823d3c91cd46937bca83da4a8e29abb5/packages/optimizers/htmlnano/package.json#L27
> 
> The svgo version used here is v2.4.0, and its dependency, stable v0.1.8, has been deprecated. It's now possible to update to the latest version of svgo, v3.2.0, to address the deprecation warning encountered during Parcel installation.
> 
> 
> 
> ## 🌍 Your Environment
> | Software            | Version(s) |
> | -----------         | -----------|
> | Parcel              | ^2.12.0    |
> | Node                | v20.11.0   |
> | pnpm                | v9.0.4     |
> | Operating System    | windows 10 |

devongovett:
Closes #9045, closes #9935, fixes #8948, fixes #9660

This is another attempt at upgrading SVGO to v3. In order to avoid breaking changes, this moves the dependency from the Parcel plugin to the project. To handle the migration, we attempt to detect the version of SVGO needed by inspecting the config file and determining if any options that were removed in v3 are present. If so, we'll log a warning and install v2. Otherwise we will install v3.

This means that new projects, and existing projects with no config or a config compatible with v3, will be upgraded automatically by installing v3 into their project. Existing projects that are using deprecated options in v2 will install v2 into their project. From then on, Parcel will load svgo relative to the project root (as it probably should have all along).

### DIFF
--- a/packages/core/integration-tests/test/html.js
+++ b/packages/core/integration-tests/test/html.js
@@ -14,6 +14,7 @@ import {
   fsFixture,
 } from '@parcel/test-utils';
 import path from 'path';
+import Logger from '@parcel/logger';
 
 describe('html', function () {
   beforeEach(async () => {
@@ -677,6 +678,77 @@ describe('html', function () {
         '<svg version=1.1 baseprofile=full width=300 height=200 xmlns=http://www.w3.org/2000/svg><rect width=100% height=100% fill=red></rect><circle cx=150 cy=100 r=80 fill=green></circle><text x=150 y=125 font-size=60 text-anchor=middle fill=white>SVG</text></svg>',
       ),
     );
+  });
+
+  it('should detect the version of SVGO to use', async function () {
+    // Test is outside parcel so that svgo is not already installed.
+    await fsFixture(overlayFS, '/')`
+      htmlnano-svgo-version
+        index.html:
+          <!DOCTYPE html>
+          <html>
+            <body>
+              <svg><rect id="test" /></svg>
+            </body>
+          </html>
+
+        .htmlnanorc:
+          {
+            "minifySvg": {
+              "full": true
+            }
+          }
+
+        yarn.lock:
+    `;
+
+    let messages = [];
+    let loggerDisposable = Logger.onLog(message => {
+      if (message.level !== 'verbose') {
+        messages.push(message);
+      }
+    });
+
+    try {
+      await bundle(path.join('/htmlnano-svgo-version/index.html'), {
+        inputFS: overlayFS,
+        defaultTargetOptions: {
+          shouldOptimize: true,
+        },
+        shouldAutoinstall: false,
+      });
+    } catch (err) {
+      // autoinstall is disabled
+      assert.equal(
+        err.diagnostics[0].message,
+        'Could not resolve module "svgo" from "/htmlnano-svgo-version/index"',
+      );
+    }
+
+    loggerDisposable.dispose();
+    assert(
+      messages[0].diagnostics[0].message.startsWith(
+        'Detected deprecated SVGO v2 options in',
+      ),
+    );
+    assert.deepEqual(messages[0].diagnostics[0].codeFrames, [
+      {
+        filePath: path.normalize('/htmlnano-svgo-version/.htmlnanorc'),
+        codeHighlights: [
+          {
+            message: undefined,
+            start: {
+              line: 3,
+              column: 5,
+            },
+            end: {
+              line: 3,
+              column: 16,
+            },
+          },
+        ],
+      },
+    ]);
   });
 
   it('should not minify default values inside HTML in production mode', async function () {
--- a/packages/core/integration-tests/test/html.js
+++ b/packages/core/integration-tests/test/html.js
@@ -15,6 +15,7 @@ import {
 } from '@parcel/test-utils';
 import path from 'path';
 import Logger from '@parcel/logger';
+import {md} from '@parcel/diagnostic';
 
 describe('html', function () {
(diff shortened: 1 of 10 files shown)


jakec-dev:
This seems to be a breaking change. We are no longer able to build as it creates error:

[BuildError: Could not resolve module "svgo" from "/builds/path/tot/index"] {
  diagnostics: [
    {
      message: 'Could not resolve module "svgo" from "/builds/path/to/index"',
      hints: [
        'Autoinstall is disabled, please install this package manually and restart Parcel.'
      ],
      origin: '@parcel/optimizer-svgo'
    }
  ]
  
  Issue only started happening after updating to 2.13.0

jakec-dev:
Can confirm after rolling back to 2.12.0 with no other changes fixes the issue. 

If SVGO is being removed from dependencies then should it not be put in as a peerDependency?

devongovett:
The hint in the diagnostic you posted suggests that you turned off auto install, and so you need to manually install svgo into your project to fix the issue. Did you follow the suggestion?

Peer dependencies are not an option because we need to detect which major version of svgo to install based on the config file in order to prevent an actually breaking change (config file no longer works) while also allowing people to upgrade to the new version. Adding a peer dep of `"svgo": "^2 || ^3"` would result in v3 always being installed by default, breaking existing projects with old config files. That's why we had to rely on auto install for this.

jakec-dev:
We haven't disabled auto install though. Our config is simply:

{
  "resolvers": [
    "@parcel/resolver-glob",
    "..."
  ]
}

initiated as

```ts
const bundler = new Parcel({
  entries: path.resolve(__dirname, '../src/client/index.html'),
  defaultConfig: '@parcel/config-default',
  mode,
  targets: {
    client: {
      context: 'browser',
      distDir: 'dist/client',
      isLibrary: false,
      scopeHoist: true,
      sourceMap: mode === 'development',
    },
  },
  shouldDisableCache: !isDev,
  ...(isDev && {
    hmrOptions: {
      port: 3000,
    },
  }),
  defaultTargetOptions: {
    publicUrl,
  },
});
```

Also as per my original comment, this is a breaking change not a minor update. To resolve this we now have to pin our version in package.json to `2.12.0`. We can't use `^2.12.0` otherwise npm updates package-lock.json to `2.13.0` which is a breaking change.

devongovett:
Ah, didn't know you were using Parcel programatically instead of via the CLI. In that case, you can enable auto install via the `shouldAutoInstall` option. Or, just install svgo in your project manually and it should be fixed. No need to pin anything.

Sorry it broke your build, but I hope you can appreciate the pickle we are in here. We can't _not_ update svgo because it's already broken, in particular due already being updated in htmlnano, causing peer dependency errors on install (#8948). So we have to bump it, we can't use peer dependencies, we're not going to release Parcel v3.0 just for this, so trying to auto install the right version is our best option.

kflorence:
Just wanted to note that this also broke my build upon upgrading from 2.12 -> 2.13. It seems that the svgo dependency must have not needed to be auto installed previously, and now it does, so it broke on the production build where auto install is disabled. Adding the svgo dependency to `devDependencies` fixed the problem.

devongovett:
Yes, that is the expected upgrade path.

PR #10057, opened 2024-12-23
Add RSC-based static site generator

devongovett:
This adds plugins for a React Server Components-based static site generator. In this setup, Parcel executes server components during the build and produces static HTML. Client components are hydrated at runtime.

It's implemented in two plugins:

* `@parcel/namer-static` renames entry bundles to have a `.html` extension, so that the following packager runs.
* `@parcel/packager-react-static` runs the JS assets in its bundle at build time, and uses React's static APIs to generate HTML.

Core changes are to support packagers returning multiple streams for different versions of the bundle. The react static packager returns both a `.html` file, as well as a `.rsc` file containing the raw RSC stream. This can be used for client-side navigations instead of reloading the whole page. Each `PackagedBundle` now has a `files` property which lists all of the files produced for that bundle, including source maps, and files produced by compressor plugins as well.

Page components receive a list of all of the pages (based on the entry bundles), along with the current page, from the packager as props. This lets them generate navigation links.


packages/examples/react-static/package.json
```diff
@@ -0,0 +1,22 @@
+{
+  "name": "parcel-rsc-static-example",
+  "version": "0.0.0",
+  "private": true,
+  "source": "pages/*.tsx",
+  "targets": {
+    "default": {
+      "context": "react-server",
+      "scopeHoist": false
```
devongovett, line 9: Slightly unfortunate: because we want to skip the JS packager and run the assets during the static packager directly, we have to turn off scope hoisting. It's easy to emulate the dev packager's output with node's vm but hard to handle the structure generated by the scope hoisting transformer.

PR #10063, opened 2024-12-28
perf: remove utility-types dependency

timdeschryver:
# ↪️ Pull Request

I want to propose to remove the `utility-types` dependency.

As part of the [e18e](https://e18e.dev/) initiative (ecosystem cleanup), we try to reduce the burden of old or big libs and modernize the ecosystem.
We noticed `utility-types` is added as a dependency within `parcel`, but it isn't used anymore.

With this change we reduce the traffic with 706.92MB a month, whenever it gets installed as part of `@parcel/types-internal`.

For reference see https://github.com/es-tooling/ecosystem-cleanup/issues/137.

## 💻 Examples

N/A

## 🚨 Test instructions

N/A

## ✔️ PR Todo

### DIFF
--- a/packages/core/fs/package.json
+++ b/packages/core/fs/package.json
@@ -63,8 +63,7 @@
     "@parcel/watcher-watchman-js": "2.13.3",
     "graceful-fs": "^4.2.4",
     "ncp": "^2.0.0",
-    "nullthrows": "^1.1.1",
-    "utility-types": "^3.10.0"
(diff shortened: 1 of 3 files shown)


devongovett:
It is actually still used, just not in the source code. When the build converts the types from Flow to TypeScript it gets referenced in the output. https://app.unpkg.com/@parcel/types-internal@2.14.3/files/lib/index.d.ts

PR #10073, opened 2025-01-19
Use import maps for bundle manifest

devongovett:
This refactors Parcel's [bundle manifest](https://parceljs.org/features/production/#cascading-invalidation), which is used to avoid cascading cache invalidation, to use native HTML [import maps](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script/type/importmap) when possible. This happens when using HTML entry points, ES module output format, and when `import.meta.resolve` is supported. In this case, a `<script type="importmap">` is automatically injected into the HTML entry point. This means it won't invalidate the entry JS bundle when it otherwise doesn't change, which benefits many apps that have large entries. In addition, it is not a separate HTTP request to load the manifest since it is embedded in the HTML itself. It also removes additional runtime code to resolve the manifest, instead relying on `import.meta.resolve`.

For React Server Components, we cannot currently use native import maps because of client side navigations, where a new HTML file is not requested. Eventually, multiple `<script type="importmap">` elements will be supported (https://github.com/whatwg/html/pull/10528) ([coming Chrome 133](https://chromestatus.com/feature/5121916248260608)), at which point React could potentially inject them. In the meantime, this adds two new APIs: `parcelRequire.extendImportMap` registers mappings, and `parcelRequire.resolve` (and `parcelRequire.load`) resolves them. An import map is added as part of the RSC client reference, and registered prior to loading the client bundles. Dynamic imports, URL dependencies, workers, etc. then use this mapping. In effect, this is the same as native import maps (it is injected into the initial HTML), but the map can be extended during client side navigations (or whenever new RSCs are loaded) too.


bertrand-caron-cascade:
Can this behaviour be disabled? This new behaviour leads to all kind of violations for websites using strict [CSP](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/CSP).

oschrenk:
Can this be disabled? Or any documentation for an upgrade path?

This currently breaks our build as instead of images it seems that `[Object object]` is being requested instead of the images.

PR #10101, opened 2025-03-08
Add optional peer dependency in @parcel/rust

nanokind:
# ↪️ Pull Request

Add missing dependencies `napi-wasm` for `@parcel/rust` which is used in `packages/core/rust/browser.js`

https://github.com/parcel-bundler/parcel/blob/6f68c78b1fa2b9653e298fd893e43e875ff6bdb8/packages/core/rust/browser.js#L1

Previously `napi-wasm` was put in `devDependencies`, this PR moves it to `dependencies`

## 💻 Examples

The original issue is that when I try to bundle `@parcel/fs` in a browser environment like parcel repl did, it says "Could not resolve napi-wasm" since `napi-wasm` is only declared in `devDependencies`

## ✔️ PR Todo

### DIFF
--- a/packages/core/rust/package.json
+++ b/packages/core/rust/package.json
@@ -33,6 +33,14 @@
     "@napi-rs/cli": "^2.15.2",
     "napi-wasm": "^1.1.2"
   },
+  "peerDependencies": {
+    "napi-wasm": "^1.1.2"
+  },
+  "peerDependenciesMeta": {
+    "napi-wasm": {
+      "optional": true
+    }
+  },
   "scripts": {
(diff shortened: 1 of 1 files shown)


devongovett:
Hmm I think we wanted to avoid installing an extra dependency that isn't needed in most cases, except if you're doing a browser build. Maybe it should be some kind of optional peer dependency or something?

nanokind:
Sounds reasonable to mark it as optional peer dependency. Code has been updated.

devongovett (approved):

PR #10118, opened 2025-03-20
Support Ranges & Integer Validation In Config Schemas

### LINKED ISSUE #10117: Support Ranges & Integer Validation In Config Schemas
# 🙋 feature request

Add support config schemas with ranges & restrictions to only integers for config schemas.

## 🤔 Expected Behavior

The feature would allow specifying optional number properties `min` and `max` values for a schema entity of type `number` for restricting the range, as well as an optional boolean property `integer` to indicate if it must be an integer (the default, undefined case is the same as `false`).

## 😯 Current Behavior

Currently, the config schema only supports declaring a schema with a predefined set of numbers, via the `enum` property:
https://github.com/parcel-bundler/parcel/blob/86bddb4031136f4b1b76e1a9a4c8a53ff4280b42/packages/core/utils/src/schema.js#L47-L51

However, this property is not that useful, as all the possible values must be listed out manually.

## 💁 Possible Solution

I'm going to open a PR in a few which adds support, because it's not that many lines of change.

## 🔦 Context

An example usecase is the @parcel/transformer-image plugin:
https://github.com/parcel-bundler/parcel/blob/86bddb4031136f4b1b76e1a9a4c8a53ff4280b42/packages/transformers/image/src/validateConfig.js#L6-L50

For many of these properties, there are range restrictions on them. However, there are too many values to realistically list all of them. Take `quality` for example, it must be in the range of 1-100, however listing all integers from 1 to 100 would be rather tedious.
So, it would be much better if instead it could be specified as the following:
```javascript
properties: {
  quality: {
    type: 'number',
    min: 1,
    max: 100,
    integer: true,
  },
  ...
}
```

This would restrict it to only integers in the range of 1-100.

solonovamax:
# ↪️ Pull Request

Adds support for config schemas with ranges & integer restrictions.

Closes #10117

## 💻 Examples

Example usage:
```javascript
const JPEG_OUTPUT_SCHEMA: SchemaEntity = {
  type: 'object',
  properties: {
    quality: {
      type: 'number',
      min: 1,
      max: 100,
      integer: true,
    }
  }
}
```
this restricts the `quality` parameter to be an integer in the range of 1-100, and it will fail with an error if it is not.

## 🚨 Test instructions

I did not add any tests for it because there wasn't really any existing tests for the config schema stuff.
Testing it manually should be quite easy, just make a config schema with a restriction on the range or a restriction to only integers, and then make a config which violates that restriction.

## ✔️ PR Todo

### DIFF
--- a/packages/bundlers/default/src/DefaultBundler.js
+++ b/packages/bundlers/default/src/DefaultBundler.js
@@ -1658,6 +1658,7 @@ const CONFIG_SCHEMA: SchemaEntity = {
           },
           split: {
             type: 'number',
+            min: 1,
           },
         },
         required: ['name', 'assets'],
@@ -1666,12 +1667,15 @@ const CONFIG_SCHEMA: SchemaEntity = {
     },
     minBundles: {
       type: 'number',
+      min: 1,
     },
     minBundleSize: {
       type: 'number',
+      min: 1,
     },
     maxParallelRequests: {
       type: 'number',
+      min: 1,
     },
     disableSharedBundles: {
       type: 'boolean',
--- a/packages/core/utils/src/schema.js
+++ b/packages/core/utils/src/schema.js
@@ -47,6 +47,9 @@ export type SchemaString = {|
 export type SchemaNumber = {|
   type: 'number',
   enum?: Array<number>,
+  min?: number,
+  max?: number,
+  integer?: boolean,
   __type?: string,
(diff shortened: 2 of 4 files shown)


codeCraft-Ritik (commented):
The API design (min, max, integer) is intuitive and easy to use

PR #10191, opened 2025-06-22
Support tree shaking without scope hoisting

devongovett:
Small transform to remove unused exports in the non-scope hoisting packager. Basically just removes the `parcelHelpers.export` call, or assignment to `exports`, which lets the minifier delete the code if it is side effect free.

This is mostly an experiment to see how much scope hoisting itself really helps, vs tree shaking but still wrapping all modules in function scopes. In reality, very few modules are actually scope hoisted because they get shared between bundles and therefore get wrapped anyway (in an app I tested < 5% of the modules were not wrapped). Tree shaking without scope hoisting is much simpler, and if we can get most of the size benefits and only maintain a single packager that would be nice.

The ReactStaticPackager also doesn't work with scope hoisted bundles since it executes modules in a dev-packager like environment at build time, so this also enables tree shaking in those builds.


yisar:
How about ESModule without scope hoist?

Issue #10172, opened 2025-05-29
`<svg>` tag makes the next siblings disappear


byF:
# 🐛 bug report

If you try to parcel an HTML file such as this:

```html
<!doctype html>
<html lang="en">
  <head>
    <meta charset="UTF-8" />
  </head>
  <body>
    <div>
      <svg></svg>
      <p>aaa</p>
      <p>bbb</p>
    </div>
  </body>
</html>
```

the HTML file produced by Parcel will have two newlines instead of paragraphs, i.e.

```html
<div>
  <svg></svg>


</div>
```

This happens only for the tags which follow after the `svg` tag.

## 🎛 Configuration (.babelrc, package.json, cli command)

Default settings, running `parcel build test.html`

## 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Parcel           | 2.15.2
| Node             | 22.15.0
| npm/Yarn         | 11.3.0
| Operating System | macOS Sequoia 15.5

rferrerasm:
I have a similar problem in my project.

When I tried to migrate from 2.13 to 2.15 I thought there was a problem with the `svg` tag.

However, when creating a basic project to report it, I can reproduce the problem with a very basic html file and without `svg`s. The sample project is https://github.com/rferrerasm/parcel-minimal-example.

I get a partially minifyed, truncated `index.html`:
```
<!DOCTYPE html><html lang=en><meta charset=utf-8><meta name=viewport content="width=device-width, initial-scale=1.0"><title>Title</title><body>
    <p>
      Body.
    </p>
  


```

I am using Windows 11. I have tried Node 20, 22 and 24 all with the same results.

sailerinteractive:
Im having the same issue. It seems like it's only happening when building a file not in development mode. Maybe related to https://github.com/parcel-bundler/parcel/pull/10090

joshfester:
I'm getting this issue as well. I had to go back to 2.14.4 to get it working. From 2.15.0 and up it doesn't work

TjeuKayim:
A `Containerfile` with minimal reproduction:

```Dockerfile
FROM node:24-alpine
WORKDIR /app
RUN npm install -g parcel@latest
RUN echo '<div><p>Before</p><svg><circle/></svg><p>After</p><p>Also removed</p></div>' > index.html && \
    parcel build index.html && \
    echo "=== Original ===" && cat index.html && \
    echo "=== Built ===" && cat dist/index.html && \
[...]

PR #10234, opened 2025-11-05
Preserve siblings after <svg>

TjeuKayim:
# ↪️ Pull Request

This pull-request fixes issue #10172 where an `<svg>` tag makes the next siblings disappear.

## 💻 Examples

Let's say you have the following HTML:

```
<svg><circle/></svg>
<p>aaa</p>
<p>bbb</p>
```

You would expect `parcel build` to keep the `<p>` elements, but it actually removed those elements. What adds to the confusion is that the bug did not happen with `parcel serve` and `parcel build --no-optimize`. So, it will be nice to have this bug fixed.

## ✔️ PR Todo

### DIFF
--- a/crates/html/src/optimize.rs
+++ b/crates/html/src/optimize.rs
@@ -389,10 +389,14 @@ pub fn optimize<'arena>(
           remove_whitespace(node, &options);
 
           if should_optimize_svg {
+            // Detach the node to prevent Oxvg from altering following siblings.
+            let parent = node.parent.take();
+            let previous_sibling = node.previous_sibling.take();
+            let next_sibling = node.next_sibling.take();
+
             // Synthesize a fake document node to act as the root of the SVG.
             let document = arena.alloc(Node::new(NodeData::Document, 0));
-            document.first_child.set(Some(node));
-            document.last_child.set(Some(node));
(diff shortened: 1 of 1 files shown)


crates/html/src/optimize.rs
```diff
@@ -402,6 +406,11 @@ pub fn optimize<'arena>(
               Err(_err) => {}
               Ok(()) => {}
             }
+
+            // Reattach in original position.
+            node.parent.set(parent);
+            node.previous_sibling.set(previous_sibling);
+            node.next_sibling.set(next_sibling);
```
TjeuKayim, line 413: alternative approach:

```rs
            // Remember the original position
            let stub = arena.alloc(Node::new(NodeData::Document, 0));
            node.insert_after(stub);

            // Synthesize a fake document node to act as the root of the SVG.
            let document = arena.alloc(Node::new(NodeData::Document, 0));
            document.append(node);

            _ = options.minify_svg.into_jobs(OxvgKind::Html).run(
              &&*document,
              &oxvg_ast::visitor::Info::<crate::oxvg::Element>::new(arena),
            );

            // Reattach
            stub.insert_after(node);
            stub.detach();
```

PR #10337, opened 2026-07-18
core: normalize codeFrame.code on thrown build errors

### LINKED ISSUE #5115: Normalize code frames on thrown errors so that `code` always exists
# 🙋 feature request

When using the Parcel API and an error is thrown, there's no guarantee that `error.diagnostics.*.codeFrame.code` exists. If it doesn't then the consumer of the API has to duplicate this logic to retrieve the code pertaining to the code frame:

https://github.com/parcel-bundler/parcel/blob/c4e494f3b9fd0eac672edc1a27c6c9216182dd6b/packages/core/utils/src/prettyDiagnostic.js#L52-L55

This is kind of a pain. I would love to see Parcel normalize these errors so that when a consumer receives a code frame on an error, the `code` property always exists and the consumer doesn't have to bother with reading source files directly.

## 🤔 Expected Behavior
I would expect the `code` property to always exists on any code frame being reported by Parcel.

## 😯 Current Behavior
The `code` property only sometimes exists on a code frame.

## 💁 Possible Solution
If a plugin doesn't provide the `code` property, Parcel could add it before the error makes its way to the Parcel consumer.

## 🔦 Context
This is described in further detail in https://github.com/parcel-bundler/parcel/discussions/5081.

Souptik96:
Fixes #5115.

## Problem

When using the Parcel API and a build fails, the thrown `BuildError`'s diagnostics may have code frames where `codeFrame.code` is `undefined` -- it's only guaranteed to be filled in when a reporter (like the CLI) calls `prettyDiagnostic()`, which reads the file from disk as a fallback purely for rendering. A consumer catching the thrown error and inspecting `error.diagnostics[i].codeFrames[j].code` directly has to duplicate that same file-read fallback themselves.

## Fix

Added `normalizeCodeFrames()` to `@parcel/diagnostic` (`packages/core/diagnostic/src/diagnostic.js`). It walks a list of diagnostics and, for any code frame missing `code`, fills it in by reading the file (via an injected `readFile` function, resolving relative paths against an optional `projectRoot` -- mirroring the existing logic in `prettyDiagnostic.js`). It leaves already-populated code untouched, doesn't mutate its input, and swallows read failures (e.g. a deleted file) by leaving `code` unset rather than throwing.

Called it at both `BuildError` throw sites in `packages/core/core/src/Parcel.js` (`run()` and the `requestBundle` callback inside `_build()`), using the resolved options' `inputFS`/`projectRoot` -- the same values `prettyDiagnostic` already receives via `PluginOptions`.

This is a pure JS/Flow change; no native/Rust code is touched.

## Testing

Added `packages/core/diagnostic/test/normalizeCodeFrames.test.js` covering:
- diagnostics without `codeFrames` pass through untouched
- an already-set `codeFrame.code` is left alone (and `readFile` is never called)
- missing `code` is filled in via `readFile`
- a relative `filePath` is joined with `projectRoot` before reading
- a failing `readFile` leaves `code` unset instead of throwing
- the input diagnostics array/objects are not mutated

I wasn't able to run the monorepo's own `yarn test:unit` in this environment (would require a full `yarn install` across the workspace), so I additionally hand-verified the exact control flow in `normalizeCodeFrames` against a plain-Node reproduction of the same logic to confirm all six scenarios behave as asserted in the test file.

### DIFF
--- a/packages/core/core/src/Parcel.js
+++ b/packages/core/core/src/Parcel.js
@@ -18,7 +18,10 @@ import type {FarmOptions, SharedReference} from '@parcel/workers';
 import type {Diagnostic} from '@parcel/diagnostic';
 
 import invariant from 'assert';
-import ThrowableDiagnostic, {anyToDiagnostic} from '@parcel/diagnostic';
+import ThrowableDiagnostic, {
+  anyToDiagnostic,
+  normalizeCodeFrames,
+} from '@parcel/diagnostic';
 import {assetFromValue} from './public/Asset';
 import {PackagedBundle} from './public/Bundle';
 import BundleGraph from './public/BundleGraph';
@@ -174,7 +177,14 @@ export default class Parcel {
     await this._end();
 
     if (result.type === 'buildFailure') {
-      throw new BuildError(result.diagnostics);
+      let resolvedOptions = nullthrows(this.#resolvedOptions);
+      throw new BuildError(
+        await normalizeCodeFrames(
+          result.diagnostics,
+          filePath => resolvedOptions.inputFS.readFile(filePath, 'utf8'),
+          resolvedOptions.projectRoot,
+        ),
+      );
     }
 
     return result;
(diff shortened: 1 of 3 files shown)


Souptik96:
Closing this from my side for now.

I've hit GitHub's per-user open-pull-request limit, so I'm withdrawing my least-active PRs to free capacity. Nothing here is abandoned on principle, and this one has had no maintainer feedback yet, so it's the cheapest to withdraw. The change (normalising `codeFrame.code` on thrown build errors) still applies to `main` and the branch on my fork is intact â€” happy to reopen if a maintainer wants to pick it up.

PR #10348, opened 2026-07-30
Inline process.platform imports in browser builds

### LINKED ISSUE #8033: Add support for `import { platform } from "process"`
# 🙋 feature request

I'm trying to bundle [`ansi-colors`](https://www.npmjs.com/package/ansi-colors) for the web, which [references](https://github.com/doowb/ansi-colors/blob/a4794363369d7b4d1872d248fc43a12761640d8e/symbols.js#L4-L5) [`process.platform`](https://nodejs.org/api/process.html#processplatform).

Tthis is currently [not supported](https://github.com/parcel-bundler/parcel/issues/7904#issuecomment-1110195182), which means a Node.JS `import` statement is currently left behind in web-targeted output:

    import{platform as e}from"process";

## 🤔 Expected Behavior

The bundler should perform static analysis on `process.platform` properties and resolve expressions at compile-time - similar to how `process.env` gets resolved, perhaps even reusing the same code.

## 🔦 Context

#7904 was recently resolved and received [an enhancement](https://github.com/parcel-bundler/parcel/pull/7954) to better support `ansi-colors`, but didn't get all the way there.

## 💻 Example

See [this project](https://github.com/mindplay-dk/inertion/tree/parcel-ansi-colors) for reference.

wanxiankai:
# ↪️ Pull Request

Closes #8033.

Browser builds currently resolve `import {platform} from "process"` through the full browser process polyfill, even though that polyfill intentionally does not define `platform`.

This change replaces named `platform` imports with a local `undefined` binding in browser contexts. Mixed imports keep their remaining specifiers, while Node contexts are left unchanged.

## 💻 Examples

```js
import {platform} from 'process';

const target = platform === undefined ? 'browser' : platform;
```

The browser bundle no longer includes `process/browser.js`, and Node builds still receive the real `process.platform` value.

## 🚨 Test instructions

- `cargo test -p parcel-js-swc-core -p parcel-html env_replacer::test`
- `/usr/bin/env NODE_ENV=test PARCEL_BUILD_ENV=test yarn mocha --experimental-vm-modules packages/core/integration-tests/test/javascript.js --grep 'process.platform imports'`
- `yarn build-ts`
- `yarn flow check`
- `yarn lint`

## ✔️ PR Todo

### DIFF
--- a/packages/core/integration-tests/test/integration/process-platform/index.js
+++ b/packages/core/integration-tests/test/integration/process-platform/index.js
@@ -1,0 +1,3 @@
+import {platform} from 'process';
+
+module.exports = platform === undefined ? 'browser' : platform;
--- a/packages/core/integration-tests/test/javascript.js
+++ b/packages/core/integration-tests/test/javascript.js
@@ -1767,6 +1767,53 @@ describe('javascript', function () {
     assert.equal(output(), true);
   });
 
+  it('should replace process.platform imports for target browser', async function () {
+    let b = await bundle(
+      path.join(__dirname, '/integration/process-platform/index.js'),
+      {
+        targets: {
+          main: {
+            context: 'browser',
+            distDir: path.join(
+              __dirname,
+              '/integration/process-platform/dist.js',
+            ),
+          },
+        },
+      },
+    );
+
+    assertBundles(b, [
+      {
+        type: 'js',
+        assets: ['index.js'],
+      },
+    ]);
+
+    let output = await run(b);
+    assert.equal(output, 'browser');
+  });
+
+  it('should preserve process.platform imports for target node', async function () {
+    let b = await bundle(
+      path.join(__dirname, '/integration/process-platform/index.js'),
+      {
+        targets: {
+          main: {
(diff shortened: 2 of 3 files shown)


codeCraft-Ritik (commented):
"Nice catch on this optimization! Quick question: Does this transformation also cover scenarios where someone uses a namespace import, like import * as process from 'process', and then accesses process.platform? Will that still pull in the polyfill, or is it also optimized away?" (Note: Based on the description "replaces named platform imports", it might only cover import { platform }, so this is a valid technical question).

wanxiankai:
Thanks for asking! This change intentionally only handles named imports such as `import {platform} from 'process'`. A namespace import like `import * as process from 'process'` is left unchanged, so it will still pull in the browser process polyfill; `process.platform` remains `undefined` because that polyfill does not define it.

Optimizing the namespace form would require analyzing every use of the namespace binding (and preserving the import if any other process members are used), so I kept this PR scoped to the named-import case. That could be explored separately if the maintainers would like broader coverage.

